PR #257, opened 2021-08-16
State threadsafe

alonmuroch:
In this PR, the multiple locks in code were removed and now that functionality is encapsulated within the `threadsafe` package.


utils/threadsafe/int32.go
```diff
@@ -0,0 +1,31 @@
+package threadsafe
```
amir-blox, line 1: using `atomic` package might give us better performance as it doesn't use locks.
https://pkg.go.dev/sync/atomic#Value

amir-blox: Performance depends on the number of running validators, therefore adding `atomic` doesn't benefit much.

PR #262, opened 2021-08-17
move vars & mock to separate file

### LINKED ISSUE #82: Move tests var's to separate file
ex: 
rsa_encryption_test.go

### LINKED ISSUE #85: Add Tests
The following places in code needs more tests:
1. node/node.go - unit tests can be added, need to decouple code by breaking node.go::Start
2. slotqueue - add unit tests
3. Tests are commented out - need to remove or uncomment:
    - `ibft/spectesting/algorithm_test.go` 
    - `utils/threshold/threshold_test.go`

nivBlox:
resolve #82 
resolve #85


eth1/mock/client.go
```diff
@@ -0,0 +1,31 @@
+package mock
```
amir-blox, line 1: Could be better to place it in `./eth1/test_utils.go` as this is the approach we have in other packages

nivBlox: done

amir-blox (approved):

PR #269, opened 2021-08-23
Incorrect size issue

### LINKED ISSUE #264: Signing duty fails on bad input value
Marshaling of the decided value breaks on bad size:
`could not execute duty {"error": "failed to sign input data: failed to marshal attestation: incorrect size"}`

https://github.com/bloxapp/ssv/blob/64b938fc15b5d3d3573d9180bf17ffdde114092d/validator/signature.go#L44

nivBlox:
resolve #264 
- fix Incorrect size issue 
- improve operator key logs

**1,** Adding same duty twice (caused by appending all new duties without filter exist one's) causing  ibft instance to start twice in parallel (with the same seqNum and identifier) and once one of the instances getting decided msg the sec one is returning empty result struct and when we try to UnmarshalSSZ (signature.go:44) decided value its throwing "Incorrect size" error

**2,** add more informative logs about the creation/use of operator logs


operator/duties/fetcher.go
```diff
@@ -40,7 +40,7 @@ type DutyFetcher interface {
 // newDutyFetcher creates a new instance
 func newDutyFetcher(logger *zap.Logger, beaconClient beaconDutiesClient, indicesFetcher validatorsIndicesFetcher, network core.Network) DutyFetcher {
 	df := dutyFetcher{
-		logger:         logger.With(zap.String("component", "operator/dutyFetcher")),
+		logger:         logger.With(zap.String("component", "Aoperator/dutyFetcher")),
```
amir-blox, line 43: typo `Aoperator`

PR #271, opened 2021-08-24
Prometheus collectors and Grafana

amir-blox:
This PR introduces prometheus metrics as detailed in https://github.com/bloxapp/ssv/blob/prometheus_grafana/metrics/README.md#collected-metrics

* adding metrics with prometheus client
* remove previous metrics
* added local setup for monitoring (`prometheus` + `grafana`)


ibft/ibft.go
```diff
@@ -93,6 +109,10 @@ func New(role beacon.RoleType, identifier []byte, logger *zap.Logger, storage co
 
 // Init sets all major processes of iBFT while blocking until completed.
 func (i *ibftImpl) Init() {
+	pubkey := i.ValidatorShare.PublicKey.SerializeToHexStr()
+	metricsIBFTsOnInit.WithLabelValues(pubkey).Inc()
```
nivBlox, line 113: @amir-blox 
so prometheus will be init by default? maybe should be with flag?

ibft/instance.go
```diff
@@ -236,11 +260,15 @@ func (i *Instance) BumpRound() {
 	i.processChangeRoundQuorumOnce = sync.Once{}
 	i.processPrepareQuorumOnce = sync.Once{}
 	i.processCommitQuorumOnce = sync.Once{}
-	i.State.Round.Set(i.State.Round.Get() + 1)
+	newRound := i.State.Round.Get() + 1
+	i.State.Round.Set(newRound)
+	metricsIBFTRound.WithLabelValues(string(i.State.Lambda.Get())).Set(float64(newRound))
```
nivBlox, line 265: we can use i.State.Round.Get() here instead

nivBlox (changes requested):

nivBlox:
We need to check if it will affect operators memory (prometheus will collect data event if not exposed)

ibft/instance.go
amir-blox: calling `i.State.Round.Get()` will acquire the lock again

amir-blox:
@nivBlox Prometheus use a single variable to store values and therefore have minimal memory footprint.
https://www.robustperception.io/memory-usage-of-prometheus-client-libraries

PR #285, opened 2021-09-01
Refactor Commit and Message Processing

alonmuroch:
Fixing the following issues:
* Align commit to QBFT (spec https://github.com/bloxapp/ssv/issues/282)
* Commit quorum execute once (https://github.com/bloxapp/ssv/issues/283)
* Commit quorum aggregate error handling (https://github.com/bloxapp/ssv/issues/284)
* Change to FIFO msg processing (https://github.com/bloxapp/ssv/issues/286)


ibft/commit.go
```diff
@@ -77,34 +76,41 @@ func (i *Instance) uponCommitMsg() pipeline.Pipeline {
...
 					zap.Int("got_votes", len(sigs)))
 
-				if aggMsg := i.aggregateMessages(sigs); aggMsg != nil {
-					i.decidedMsg = aggMsg
-					// mark instance commit
-					i.ProcessStageChange(proto.RoundState_Decided)
-					i.Stop()
+				aggMsg, err := i.aggregateMessages(sigs)
+				if err != nil {
+					i.Logger.Error("could not aggregate commit messages after quorum", zap.Error(err))
```
amir-blox, line 81: shouldn't we break here if aggregation failed?

ibft/ibft.go
```diff
@@ -53,52 +54,61 @@ type IBFT interface {
...
 // Init sets all major processes of iBFT while blocking until completed.
 func (i *ibftImpl) Init() {
+	i.logger.Info("iBFT implementation init started")
 	i.processDecidedQueueMessages()
 	i.processSyncQueueMessages()
 	i.listenToSyncMessages()
 	i.waitForMinPeerOnInit(2) // minimum of 3 validators (me + 2)
-	i.SyncIBFT()
+	if err := i.SyncIBFT(); err != nil {
+		i.logger.Error("crashing.. ", zap.Error(err))
```
amir-blox, line 106: @alonmuroch same here, just logging the error is probably not sufficient

ibft/commit.go
amir-blox: In addition, round timer could be stopped while the aggregation fails
https://github.com/bloxapp/ssv/pull/285/files#diff-186423815922aeefaa19db97036033f00d2454e9dc88c19fbe33a1435ae5a5e6R73

In general the timer is being stopped by `instance.Stop()`, the call from this function to `stopRoundTimer` looks redundant

alonmuroch: We shouldn't stop if aggregation fails as you've decided.

not sure I understand the round timer comment

ibft/change_round.go
```diff
@@ -135,13 +133,12 @@ func (i *Instance) changeRoundQuorum(round uint64) (quorum bool, t int, n int) {
...
 
 func (i *Instance) roundChangeInputValue() ([]byte, error) {
-	quorum, msgs := i.PrepareMessages.QuorumAchieved(i.State.PreparedRound.Get(), i.State.PreparedValue.Get())
-
 	// prepare justificationMsg and sig
 	var justificationMsg *proto.Message
 	var aggSig []byte
 	ids := make([]uint64, 0)
-	if quorum {
+	if i.isPrepared() {
```
alonmuroch, line 140: Should we use highest prepared/ prepared messages? combination of the 2?

PR #288, opened 2021-09-02
Merge stage to main

amir-blox:
Resolve https://github.com/bloxapp/ssv/issues/235


amir-blox:
Will be opened with more changes

PR #291, opened 2021-09-05
Remove fatal when sync fails

amir-blox:


ibft/ibft.go
```diff
@@ -94,7 +99,10 @@ func (i *ibftImpl) Init() {
 	i.processSyncQueueMessages()
 	i.listenToSyncMessages()
 	i.waitForMinPeerOnInit(2) // minimum of 3 validators (me + 2)
-	i.SyncIBFT()
+	if err := i.SyncIBFT(); err != nil {
+		i.logger.Error("crashing.. ", zap.Error(err))
```
alonmuroch, line 103: maybe the error should not be crashing if it doesn't actually crash?

alonmuroch (changes requested):

ibft/ibft.go
amir-blox: OK, will be fixed in the following PR

PR #301, opened 2021-09-12
Refactor Resync

amir-blox:
Resolve https://github.com/bloxapp/ssv/issues/298

* Added a new flag for eth1 data cleanup (`CleanRegistryData` / `CLEAN_REGISTRY_DATA`)
  * Using this flag will clear saved sync offset, i.e. the ETH1 sync will start from the beginning
  * `4E706F` is the initial sync offset for [registry contract](https://goerli.etherscan.io/address/0x687fb596F3892904F879118e2113e1EEe8746C2E)
* Change `ETH1SyncOffset` behavior to be a secondary value, i.e. it will be used in cases where no previous sync offset exist  (e.g. first start, resync)


exporter/node.go
```diff
@@ -104,9 +103,26 @@ func New(opts Options) Exporter {
...
 
+	if err := e.init(opts); err != nil {
+		e.logger.Fatal("failed to init", zap.Error(err))
+	}
+
 	return &e
 }
 
+func (exp *exporter) init(opts Options) error {
+	if opts.CleanRegistryData {
```
alonmuroch, line 114: @amir-blox why not put the delete operation in the cli package?

exporter/storage/storage.go
```diff
@@ -29,6 +31,11 @@ func NewExporterStorage(db basedb.IDb, logger *zap.Logger) Storage {
 	return &es
 }
 
+// Clean clears all information
+func (es *exporterStorage) Clean() error {
```
alonmuroch, line 35: @amir-blox this deletes the entire storage?

operator/node.go
```diff
@@ -68,9 +68,23 @@ func New(opts Options) Node {
...
 	}
+
+	if err := ssv.init(opts); err != nil {
+		ssv.logger.Fatal("failed to init", zap.Error(err))
+	}
+
 	return ssv
 }
 
+func (n *operatorNode) init(opts Options) error {
```
alonmuroch, line 79: @amir-blox why not put it in the cli package?

operator/storage.go
```diff
@@ -40,6 +40,10 @@ func (s *storage) SaveSyncOffset(offset *eth1.SyncOffset) error {
 	return s.db.Set(prefix, syncOffsetKey, offset.Bytes())
 }
 
+func (s *storage) cleanSyncOffset() error {
```
alonmuroch, line 43: @amir-blox cleaning the offset doesn't clean the actual shares right? doesn't it mean the node could potentially still use old non deleted shares?

validator/controller.go
```diff
@@ -107,6 +102,21 @@ func (c *controller) ListenToEth1Events(cn pubsub.SubjectChannel) {
 	}
 }
 
+// initShares initializes shares
+func (c *controller) initShares(options ControllerOptions) {
+	if options.CleanRegistryData {
```
alonmuroch, line 107: @amir-blox maybe move to the cli package?

alonmuroch (changes requested):

exporter/node.go
amir-blox: The delete is not "generic" - it uses some prefix, meaning we need to perform this from the actual `collection`/`storage` where the prefix is defined

operator/storage.go
amir-blox: Theoretically yes as the sync offset and shares uses different `collection` (shares cleanup is on validator storage), 
but it is the same underlying DB and therefore it could happen if the DB returns error for one operation and success for the other.

exporter/storage/storage.go
amir-blox: It will delete exporter storage - validators/operators information, which is saved in a different collection than shares

operator/node.go
amir-blox: Same comment as above:

The delete is not "generic" - it uses some prefix, meaning we need to perform this from the actual collection/storage where the prefix is defined

validator/controller.go
amir-blox: Same comment as above:

The delete is not "generic" - it uses some prefix, meaning we need to perform this from the actual collection/storage where the prefix is defined

exporter/node.go
alonmuroch: @amir-blox it can cause issues in the future if we need something else deleted and the logic is spread around in multiple places.
I guess for now it will do

operator/storage.go
alonmuroch: @amir-blox maybe delete offset and shares?

amir-blox: I can also fail to start if we can't delete something

PR #302, opened 2021-09-12
fix timeout timer go rountine leak

alonmuroch:


ibft/roundtimer/timer.go
```diff
@@ -9,78 +9,100 @@ import (
...
+	defer t.syncLock.Unlock()
+
+	t.stopped = false
 
 	if t.timer != nil {
 		t.timer.Stop()
 		t.timer.Reset(d)
 	} else {
-		t.timer = time.NewTimer(d)
+		t.timer = time.AfterFunc(d, func() {
```
amir-blox, line 65: Using `timer.AfterFunc` could lead to leaks as it is not GC'ed:
https://www.fatalerrors.org/a/use-with-caution-time.after-can-cause-memory-leak-golang.html (this suggest to use time ticker instead) 
https://medium.com/@oboturov/golang-time-after-is-not-garbage-collected-4cbc94740082

alonmuroch: @amir-blox it seems AfterFunc does get GC'd .. only the After function isn't

Issue #298, opened 2021-09-11
handleValidatorAddedEvent issues


alonmuroch:
When receiving a new validator share we check against the storage if it exists. If it doesn't we save the share 

1. we update share if resync flag on instead of deleting the db.. that could cause override issues 
2. If the share is does not exist we call validatorsMap.UpdateValidatorShare which does nothing as the validator doesn't exist yet
3. If the share does exist we get it from the map but then call start on it, this can cause multiple start calls on the same validator

Expected behaviour
1) resync should delete the eth1 storage not override shares
2) only new shares should get saved, create a validator instance in the map and start the validator

PR #303, opened 2021-09-12
Change indices fetching to work synchronously

amir-blox:
Resolve https://github.com/bloxapp/ssv/issues/296
Resolve https://github.com/bloxapp/ssv/issues/297


amir-blox:
Fetching indices synchronously on validator added introduces several issues:
* The sync is considered successful if all events were parsed, but what if we fail to fetch indices? should fail the sync so the data will arrive again on next sync
* Overloading beacon requests

alonmuroch:
@amir-blox I think it solves #298 as well

PR #314, opened 2021-09-20
Validator Metadata

alonmuroch:
Added share status to indicate deposited/ exited/ slashed to reflect in grafana.

Open question - I've added aa field into the share struct, will that require all nodes to resync?

@amir-blox how do we reflect validator status in grafana?


amir-blox:
The new field is now a struct of its own, called `ValidatorMetadata`: 
https://github.com/bloxapp/ssv/pull/314/files#diff-ce1dc86d2cccc8c95bd3f27446a0ff8cdee922091abfd6fb97597be4524b20c6R18)
We can easily identify if not exist (nil) or fetched with bad values, unlike before when the default state of the var.

It is possible to upgrade w/o resync.

amir-blox:
**Note** exporter deployment was changed to support beacon node configuration: 
https://github.com/bloxapp/ssv/pull/314/files#diff-6d35f97fbb2f0815d0c21cd72d2562d46da7453ab25c04ae4c53976d81afe342L141

config-map of exporter should include the `BeaconNodeAddr` config

amir-blox:
Changes:
* added `ValidatorMetadata` to hold information from beacon
  * added `ValidatorMetadata` and beacon node to exporter
* changed the flow of updating metadata and reporting validator status
* improves validator status metric (+ dashboard in a different PR) to identify `exited` and `not found` validators
* removed exporter metrics, using now the same metrics as ssv operator node

PR #324, opened 2021-09-24
F+1 catch bump round fix

alonmuroch:
This solves an issue of bumping round (while f+1 catch up) directly by calling state.round ++ which doesn't execute the necessary bump round logic. 
This caused f+1 to sometimes not execute the change round quorum as it didn't reset:
i.processChangeRoundQuorumOnce = sync.Once{}
	i.processPrepareQuorumOnce = sync.Once{}
	i.processCommitQuorumOnce = sync.Once{}

### DIFF
--- a/ibft/partial_change_round.go
+++ b/ibft/partial_change_round.go
@@ -55,7 +55,7 @@ func (i *Instance) uponChangeRoundPartialQuorum(msgs []*network.Message) (bool, 
 
 	// TODO - could have a race condition where msgs are processed in a different thread and then we trigger round change here
 	if foundPartialQuorum {
-		i.State.Round.Set(lowestChangeRound)
+		i.bumpToRound(lowestChangeRound)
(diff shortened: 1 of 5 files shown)


amir-blox (approved):

ibft/partial_change_round.go
```diff
@@ -55,7 +55,7 @@ func (i *Instance) uponChangeRoundPartialQuorum(msgs []*network.Message) (bool,
 
 	// TODO - could have a race condition where msgs are processed in a different thread and then we trigger round change here
 	if foundPartialQuorum {
-		i.State.Round.Set(lowestChangeRound)
+		i.bumpToRound(lowestChangeRound)
 		metricsIBFTRound.WithLabelValues(string(i.State.Lambda.Get()),
```
amir-blox, line 59: reporting to metric is redundant as `bumpToRound` calls it

PR #344, opened 2021-10-04
Network Improvements

amir-blox:
Trying to fix https://github.com/bloxapp/ssv/issues/310:
* Added [flood publish](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing) so each peer will broadcast **his** messages to all known peers. 
* Introduce `main` topic, a central pubsub topic for publishing decided messages across the network

In addition libp2p User Agent was added - `app:version:operatorPubKey`


network/p2p/p2p_decided.go
```diff
@@ -26,6 +26,15 @@ func (n *p2pNetwork) BroadcastDecided(topicName []byte, msg *proto.SignedMessage
 
 	n.logger.Debug("Broadcasting decided message", zap.String("lambda", string(msg.Message.Lambda)), zap.Any("topic", topic), zap.Any("peers", topic.ListPeers()))
 
+	// publishing on main topic as well
```
alonmuroch, line 29: @amir-blox should this be in network? why not in the node logic?

alonmuroch:
@amir-blox you will add commit msg handling to this PR or a different one?

utils/commons/build_data.go
```diff
@@ -0,0 +1,19 @@
...
+
+// SetBuildData updates local vars for build data
+func SetBuildData(app string, ver string) {
+	appName = app
+	version = ver
+}
+
+// GetBuildData returns build data
+func GetBuildData() string {
+	return fmt.Sprintf("%s:%s", appName, version)
```
alonmuroch, line 18: @amir-blox didn't we want to add operator pub key somehow?

network/p2p/p2p_decided.go
amir-blox: After `ExecuteDuty`?  it will happen after signature collection.. 
Other places could be duty_execution or IBFT but then I need to restore the message to send

utils/commons/build_data.go
amir-blox: I remember we decided to avoid it for now and wait for a proper way of authenticating peers, but yes we could easily add this now until we are ready

amir-blox:
@alonmuroch Commit msg handling (exporter) is in https://github.com/bloxapp/ssv/pull/345 for the operator nodes I'll create another PR

PR #363, opened 2021-10-10
[WIP] Fork support - step 1

alonmuroch:


alonmuroch:
Fork support in a nutshell:
- We add a fork interface to the following structs: Operator node, Validator, ibft controller, ibft instance, network, storage 
- each has a set of dedicated functions that dynamically change implementation depending on the fork.
- for example ibft instance changes pipelines depending on fork
- The forks are nested: operator node holds ibft controller, network and storage forks. IBFT controller fork holds ibft instance fork.
- the main node fork is set (currently hard forked) in the CLI package.
- That fork instance is then propagated to al other components

PR #398, opened 2021-11-01
Fix docker port publishing

yura-pakhuchiy:
SSV uses UDP port 12000, not TCP port


amir-blox (approved):

amir-blox:
@yura-pakhuchiy thanks, it will be merged to `main` branch in the coming days

PR #400, opened 2021-11-02
Update issue templates

nivBlox:
Adding issue template for Blox team and contributes


nivBlox:
@amir-blox @lior-blox @alonmuroch 
FYI

PR #401, opened 2021-11-02
Refactor network package

amir-blox:
- Resolve https://github.com/bloxapp/ssv/issues/390

### DIFF
--- a/network/p2p/discovery_discv5.go
+++ b/network/p2p/discovery_discv5.go
@@ -1,0 +1,279 @@
+package p2p
+
+import (
+	"crypto/ecdsa"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p/discover"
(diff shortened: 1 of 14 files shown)


network/p2p/discovery_discv5.go
```diff
@@ -0,0 +1,279 @@
...
+func (n *p2pNetwork) listenForNewNodes() {
+	defer n.logger.Debug("done listening for new nodes")
+	iterator := n.dv5Listener.RandomNodes()
+	//iterator = enode.Filter(iterator, s.filterPeer)
+	defer iterator.Close()
+	n.logger.Debug("starting to listen for new nodes")
+	for {
+		// Exit if service's context is canceled
+		if n.ctx.Err() != nil {
+			break
```
nivBlox, line 198: worth adding log here

nivBlox: background ctx. if fails all is down

PR #416, opened 2021-11-08
close Sync stream fix

alonmuroch:


network/p2p/utils.go
```diff
@@ -17,22 +17,26 @@ import (
...
-	if err != nil {
-		return nil, err
-	}
-	dst := make([]byte, hex.EncodedLen(len(rawbytes)))
-	hex.Encode(dst, rawbytes)
-	if err := file.WriteFile(defaultKeyPath, dst); err != nil {
-		return nil, err
+
+	if saveToFile {
+		defaultKeyPath := defaultDataDir()
```
amir-blox, line 27: default data dir is located in the OS home folder, which is not mounted currently

alonmuroch: @amir-blox this should be fixed in a different PR

PR #420, opened 2021-11-09
Sync stream compatibility

amir-blox:
- Added version to sync protocols:
  - `/sync/highest_decided/0.0.1`
  - `/sync/decided_by_range/0.0.1`
  - `/sync/last_change_round/0.0.1`
- Added the old `/sync/0.0.1` protocol for backwards compatibility


amir-blox:
Will be fixed in https://github.com/bloxapp/ssv/pull/422

PR #427, opened 2021-11-14
Slashing protection v2

alonmuroch:
Add slashing protection to value check before and during consensus phases


y0sher:
@moshe-blox is this up to date in anyway? we're leaving it for ref? we want to do something with it anytime?

PR #431, opened 2021-11-16
Decided message signers fix

amir-blox:
Resolve https://github.com/bloxapp/ssv/issues/430

IBFT messages were purged once the instance decided or stopped, the behavior changed in this PR:
- when instance finished w/ decided - ctrl starts to listen for late commit messages with timeout and purge messages once done
- when instance finished w/o decided - ctrl purges messages

### DIFF
--- a/ibft/controller/controller_decided.go
+++ b/ibft/controller/controller_decided.go
@@ -64,6 +64,16 @@ func (i *Controller) ProcessDecidedMessage(msg *proto.SignedMessage) {
 		return
 	}
 	if known {
+		// if decided is known, check for a more complete message (more signers)
+		if ignore, _ := i.checkDecidedMessageSigners(msg); !ignore {
+			if err := i.ibftStorage.SaveDecided(msg); err != nil {
+				i.logger.Error("can't update decided message", zap.Error(err))
+				return
+			}
+			i.logger.Debug("decided was updated")
+			ibft.ReportDecided(i.ValidatorShare.PublicKey.SerializeToHexStr(), msg)
+			return
+		}
 		i.logger.Debug("decided is known, skipped")
 		return
 	}
@@ -117,6 +127,22 @@ func (i *Controller) decidedMsgKnown(msg *proto.SignedMessage) (bool, error) {
 		return false, errors.Wrap(err, "could not get decided instance from storage")
 	}
 	return found, nil
+}
+
+// checkDecidedMessageSigners checks if signers of existing decided includes all signers of the newer message
+func (i *Controller) checkDecidedMessageSigners(msg *proto.SignedMessage) (bool, error) {
+	decided, found, err := i.ibftStorage.GetDecided(msg.Message.Lambda, msg.Message.SeqNumber)
+	if err != nil {
+		return false, errors.Wrap(err, "could not get decided instance from storage")
+	}
+	if !found {
+		return false, nil
+	}
+	// decided message should have at least 3 signers, so if the new decided has 4 signers -> override
+	if len(decided.SignerIds) < i.ValidatorShare.CommitteeSize() && len(msg.SignerIds) > len(decided.SignerIds) {
+		return false, nil
+	}
+	return true, nil
 }
 
 // decidedForCurrentInstance returns true if msg has same seq number is current instance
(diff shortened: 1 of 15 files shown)


ibft/controller/controller_decided.go
```diff
@@ -119,6 +129,22 @@ func (i *Controller) decidedMsgKnown(msg *proto.SignedMessage) (bool, error) {
 	return found, nil
 }
 
+// checkDecidedMessageSigners checks if signers of existing decided includes all signers of the newer message
+func (i *Controller) checkDecidedMessageSigners(msg *proto.SignedMessage) (bool, error) {
```
alonmuroch, line 133: @amir-blox tests for those conditions?

ibft/controller/controller_running_instance.go
```diff
@@ -107,33 +122,41 @@ func (i *Controller) instanceStageChange(stage proto.RoundState) (bool, error) {
...
 			if netMsg := i.msgQueue.PopMessage(idxKey); netMsg != nil && netMsg.SignedMessage != nil {
 				if netMsg.SignedMessage.Message == nil || netMsg.SignedMessage.Message.Type != proto.RoundState_Commit {
 					// not a commit message -> skip
 					continue
 				}
 				logger := i.logger.With(zap.Uint64("seq", netMsg.SignedMessage.Message.SeqNumber),
-					zap.String("identifier", string(netMsg.SignedMessage.Message.Lambda)))
-				if err := runningInstance.CommitMsgValidationPipeline().Run(netMsg.SignedMessage); err != nil {
+					zap.Uint64s("signers", netMsg.SignedMessage.SignerIds))
+				if err := instance.CommitMsgValidationPipelineV0(identifier, seq, i.ValidatorShare).Run(netMsg.SignedMessage); err != nil {
```
alonmuroch, line 144: @amir-blox better use the fork and not the implementation directly

ibft/controller/controller_running_instance.go
```diff
@@ -107,33 +122,41 @@ func (i *Controller) instanceStageChange(stage proto.RoundState) (bool, error) {
...
+
 	f := func(stopper tasks.Stopper) (interface{}, error) {
 	loop:
 		for {
 			if stopper.IsStopped() {
 				break loop
 			}
-			idxKey := msgqueue.IBFTMessageIndexKey(runningInstance.State().Lambda.Get(), runningInstance.State().SeqNumber.Get())
 			if netMsg := i.msgQueue.PopMessage(idxKey); netMsg != nil && netMsg.SignedMessage != nil {
 				if netMsg.SignedMessage.Message == nil || netMsg.SignedMessage.Message.Type != proto.RoundState_Commit {
```
alonmuroch, line 138: why check msg nil? pipeline should take care of it no?

ibft/instance/commit.go
```diff
@@ -65,12 +76,17 @@ func (i *Instance) CommitMsgValidationPipeline() pipeline.Pipeline {
 
 // CommitMsgValidationPipelineV0 is version 0
 func (i *Instance) CommitMsgValidationPipelineV0() pipeline.Pipeline {
+	return CommitMsgValidationPipelineV0(i.State().Lambda.Get(), i.State().SeqNumber.Get(), i.ValidatorShare)
+}
+
+// CommitMsgValidationPipelineV0 is version 0 of commit message validation
+func CommitMsgValidationPipelineV0(identifier []byte, seq uint64, share *storage.Share) pipeline.Pipeline {
```
alonmuroch, line 83: that seems redundant ..

alonmuroch (changes requested):

ibft/instance/commit.go
amir-blox: its a function so the exporter could use it as well

alonmuroch: we have the instance which has that function why create another?

amir-blox: late commit is done on controller level when the instance finished and on exporter, in both cases we don't have an instance..

amir-blox:
The ibft instance is not available in late commit scenarios:
1. ibft controller after the instance was cleared
2. exporter

Therefore in this PR, a static commit message validation function was added to overcome the missing instance object.
It should be refactored in the future to be accessed via the fork

PR #433, opened 2021-11-17
Node Type (User agent)

nivBlox:
Adding node type (Unknown, operator, exporter) to user agent (temp solution until network refactoring)

* report to Operator table in grafana under "type" header
* once fetching peers for topic it will filter all exporter node's type

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ vendor/
 
 .env*
 /config/share*
+/config/config.yaml
 
 # User-specific stuff:
 .idea/**/workspace.xml
--- a/ibft/sync/speedup/speedup.go
+++ b/ibft/sync/speedup/speedup.go
@@ -59,10 +59,10 @@ func (s *Speedup) Start() ([]*proto.SignedMessage, error) {
 				Lambda: s.identifier,
 			})
 			if err != nil {
-				s.logger.Error("error fetching latest change round", zap.Error(err))
+				s.logger.Error("error fetching latest change round", zap.Error(err), zap.Any("peer", peer))
 			} else if err := s.lastMsgError(msg); err != nil {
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ vendor/
 
 .env*
 /config/share*
+/config/config.yaml
```
amir-blox, line 19: This will remove the config example.. need to change the filename (e.g. `config.example.yaml`) or remove this

ibft/sync/speedup/speedup.go
```diff
@@ -59,10 +59,10 @@ func (s *Speedup) Start() ([]*proto.SignedMessage, error) {
 				Lambda: s.identifier,
 			})
 			if err != nil {
-				s.logger.Error("error fetching latest change round", zap.Error(err))
+				s.logger.Error("error fetching latest change round", zap.Error(err), zap.Any("peer", peer))
```
amir-blox, line 62: should be `zap.String`

network/p2p/config.go
```diff
@@ -40,6 +40,28 @@ type Config struct {
 	OperatorPrivateKey *rsa.PrivateKey
 	// ReportLastMsg whether to report last msg metric
 	ReportLastMsg bool
+	// NodeType added to ENR in order to differentiate exporters peers from others
```
amir-blox, line 43: remove `added to ENR` from comment

.gitignore
nivBlox: Added config.example.yaml

ibft/sync/speedup/speedup.go
nivBlox: done

network/p2p/config.go
nivBlox: done

PR #465, opened 2021-12-09
Read only mode for operator node

amir-blox:
Enables to pass a config flag to turn off duty execution in an operator node, means that the node won't take part in validator's consensus


alonmuroch:
looks good

nivBlox (approved):

PR #470, opened 2021-12-12
Refactor internal msg listeners

amir-blox:
- extract listeners from p2p package for reuse and cleaning in general
- added container to manage thread-safe listeners registration, de-registration and invocation
- resolved high memory+goroutines in the previous implementation

### DIFF
--- a/network/commons/listeners/container.go
+++ b/network/commons/listeners/container.go
@@ -19,6 +19,7 @@ type Container interface {
 	GetListeners(msgType network.NetworkMsg) []*Listener
 }
 
+// listeners is the data structure where listeners are saved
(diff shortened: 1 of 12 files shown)


network/commons/listeners/container.go
```diff
@@ -0,0 +1,102 @@
...
+	"crypto/rand"
+	"fmt"
+	"github.com/bloxapp/ssv/network"
+	"go.uber.org/zap"
+	"math/big"
+	"sync"
+	"time"
+)
+
+// Container is an interface for of listeners container
```
nivBlox, line 14: remove 'for'

PR #471, opened 2021-12-12
Sync Backoff

amir-blox:
- added `HistorySyncBackOff` flag to control the work load coming from sync requests


alonmuroch:
@amir-blox backoff suggests that it increments somehow (say if many requests are sent). It seems it's constant no?

nivBlox (approved):

amir-blox:
@alonmuroch @lior-blox I will change the name to `HistorySyncRateLimit`

amir-blox:
fixed in https://github.com/bloxapp/ssv/pull/474

PR #487, opened 2021-12-26
Peers Limit (part 2)

amir-blox:
Trying to resolve https://github.com/bloxapp/ssv/issues/477, by reducing network overload / lower the number of connected peers.
The idea is to connect only with relevant peers (in some validator's committee) once the node reached peers limit.
It is done with filters that are executed by connection handler for every new connection.
In addition, discovery will also filter irrelevant peers if needed by reading operator-id and node type directly from peer's ENR (https://github.com/bloxapp/ssv/pull/495)

- Improved `PeersIndex`
  - uses libp2p's peerstore as storage
  - index both incoming connections (by User Agent) and discovered nodes (ENR)
  - added cache for storing pruned peers
- Added a network notifications handler that indexes and filters p2p connections
  - Added `ConnectionFilter` function interface
  - Added filter for irrelevant operators
  - Added filter for unknown user agent (Resolve https://github.com/bloxapp/ssv/issues/500)
- `ConnectionGater` [interface](https://github.com/libp2p/go-libp2p-core/blob/master/connmgr/gater.go) was implemented
- Discovery: added a function to try new nodes, that will filter irrelevant nodes if we reached peers limit
- An exporter will try to connect to all nodes, and they will always accept it
  - TODO: need the same for bootnode, but it requires to change ENR across the network and therefore won't be implemented now


amir-blox:
There are some issues with the idea to filter peers based on the user agent:
1. we need to connect to some node in order to get its user agent for the operator public key (and disconnect if it doesn't share committees)
2. there is a race in this scenario as the peer indexing is async, and we actually need to wait for the peer to be indexed rather than connected

Instead, we can use the ENR to store the value, and therefore identify operators w/o connection (https://github.com/bloxapp/ssv/pull/488).
The drawback is that other peers need to update version as we previously set a substring of that value (in 'pk' field, which was not in use and therefore we decided to wait with implementing something else).

amir-blox:
> There are some issues with the idea to filter peers based on their user agent:
> 
> 1. we need to connect to some node in order to get its user agent for the operator public key (and disconnect if it doesn't share committees)
> 2. there is a race in this scenario as the peer indexing is async, and we actually need to wait for the peer to be indexed rather than connected
> 
> Instead, we can use the ENR to store the value, and therefore identify operators w/o connection (#488). The drawback is that other peers need to update version as we previously set a substring of that value (in 'pk' field).

This was fixed in https://github.com/bloxapp/ssv/pull/487/commits/3cd5ebc97f38026937dd430d8a62a2e167aebe01, with a fallback to check the user agent for non updated peers

PR #510, opened 2022-01-19
Exporter: fix validators order

amir-blox:
Resolve https://github.com/bloxapp/ssv/issues/509

Set the index of a validator before making request to beacon for it's data, which could take time and breaks the order

### DIFF
--- a/exporter/registry_handlers.go
+++ b/exporter/registry_handlers.go
@@ -7,6 +7,7 @@ import (
 	"github.com/bloxapp/ssv/exporter/api"
 	"github.com/bloxapp/ssv/exporter/storage"
 	"github.com/bloxapp/ssv/validator"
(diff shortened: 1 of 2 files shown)


amir-blox:
Seems like because we use goroutines to process events, the order gets messed up.
A better place in code to determine internal index is right when we are parsing the event:
https://github.com/bloxapp/ssv/blob/09a841012b53b0e38aaaa3ae1c39eca7d10d6d6c/eth1/goeth/goETH.go#L379

amir-blox:
As a more robust fix seems to require some changes.

The options that were discussed:
1. Save validators by the block number + block index (incremental index within the block) and avoid maintaining additional index
2. Govern the insertion of new validators and set the internal index accordingly

PR #515, opened 2022-01-25
[WIP] migrations

olegshmuelov:


migrations/migrations.go
```diff
@@ -0,0 +1,78 @@
...
+	migrationsPrefix   = []byte("migrations-")
+	migrationCompleted = []byte("migrationCompleted")
+)
+
+type Eth1Syncer interface {
+	// SyncEth1Events TODO: implement
+	SyncEth1Events() error
+}
+
+type environment struct {
```
nivBlox, line 38: Lets call it "Options" to align with the project semantic

storage/basedb/storage.go
```diff
@@ -14,6 +15,14 @@ type Options struct {
 	Ctx       context.Context
 }
 
+// ITxn interface for badger transaction like functions
+type ITxn interface {
```
amir-blox, line 19: you can remove the starting `I` (`ITxn` > `Txn`)

`IDb` is there from historical reasons..

storage/kv/badger.go
```diff
@@ -231,6 +234,45 @@ func (b *BadgerDb) listRawKeys(prefix []byte, txn *badger.Txn) [][]byte {
...
+
+type badgerTxn struct {
+	txn *badger.Txn
+}
+
+func (t *badgerTxn) Set(prefix []byte, key []byte, value []byte) error {
+	return t.txn.Set(append(prefix, key...), value)
+}
+
+func (t *badgerTxn) Get(prefix []byte, key []byte) (obj basedb.Obj, found bool, err error) {
```
moshe-blox, line 257: @amir-blox Can we re-use this snippet in `IDb.Get` or would it just complicate or hurt performance? (because of interfaces)

amir-blox: Good idea.

`Get` is usually called to fetch a single item,`GetMany` should be called if there are multiple items to read at once (e.g. get decided in range), therefore we shouldn't see a real impact on performance when reusing `Txn.Get`.

PR #516, opened 2022-01-27
Networking Specification

amir-blox:


docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+Multiplexing of protocols over channels is achieved using [yamux](https://github.com/libp2p/go-libp2p-yamux) protocol.
+
+### Messaging
+
+Messages in the network are formatted with `protobuf` (NOTE: `v0` messages are encoded/decoded with JSON),
+and being transported p2p with one of the following methods:
+
+**Streams** 
+
+Streams are used for direct messages between peers.
```
alonmuroch, line 64: Is that the right way to do it? I'm pretty sure Prysm does that the same just maybe worth double checking

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+### Connection Scoring
+
+Peer's connection score is determined after a successful handshake,
+and peers with low score will be pruned.
+
+Connection scores are based on the following properties:
+
+- Shared subnets / committees (`25`)
+- Static nodes such as `exporter` (`10000`)
```
alonmuroch, line 534: Not sure this can be part of scoring as a malicious peer can simply say its an exporter when its not

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+- `opportunisticGraftThreshold`: 5
+
+[Score function](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#the-score-function) is running pariodically and will determine the score of peers. During heartbeat, the score it checked and bad peers are handled accordingly.
+
+
+### Message Scoring
+
+Message scorers track on operators' behavior w.r.t incoming IBFT messages:
+
+- Invalid message signature (`-100`)
```
alonmuroch, line 523: maybe add late arrivals of msgs?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+- Message from operator w/o shared committees (`-1000`)
```
alonmuroch, line 524: why is that a factor?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+### Network Discovery
+
+[discv5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md) 
+is used in `SSV.Network` as the discovery system.
+
+More information is available in [Networking > Discovery](#discovery)
+
+
+### Peer Scoring
```
alonmuroch, line 117: I'd love to see some reference to eth/eth2 scoring methodology

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+As part of the algorithm, nodes are exchanging messages with other nodes in the committee. \
+Once the committee reaches consensus, the nodes will publish the decided message across the network.
+
+Consensus messages are being sent in the network over pubsub topics (see [subnets](#subnets))
+
+More information regarding the protocol can be found in [iBFT annotated paper (By Blox)](/ibft/IBFT.md)
+
+### Message Structure
+
+`SignedMessage` is a wrapper for IBFT messages, it holds a message and its signature with a list of signer IDs:
```
alonmuroch, line 149: I think we should reference the in-dev QBFT and SSV specs
https://github.com/bloxapp/ssv/blob/ssv_spec/docs/spec/qbft/messages.go

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+  ```
+</details>
+
+**NOTE:** 
+- all pubsub messages in the network are wrapped with libp2p's message structure
+- `signer_ids` must be sorted, to allow hashing the entire message
+
+---
+
+## Sync Protocols
```
alonmuroch, line 209: I think we can leave sync messages to:
1) get highest sequence for a specific QBFT instance
2) get historical decided.

We can have a flag in the node for saving/ not saving historical decided messages.
Both exporter and a regular node (operator and not operator) can comply with those api/ rpc calls.

Should we make public what the node supports in the handshake?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+The following error codes exist: **TBD**
+- `EntryNotFoundError` is returned when no results were found
+- `InternalError` is returned upon internal error
+
+### Protocols
+
+SSV nodes use the following stream protocols:
+
+### 1. Highest Decided
```
alonmuroch, line 279: this one we will need for all types of nodes

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+        "signer_ids": [1,2,4]
+      }
+    ],
+    "type": 0,
+    "identifier": "..."
+  }
+  ```
+</details>
+
+### 2. Decided By Range
```
alonmuroch, line 319: this one can be optional depending if the node supports it or not

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+  ```
+</details>
+
+---
+
+## Handshake protocol
+
+`/ssv/handshake/0.0.1`
+
+The handshake protocol allows peers to identify, by exchanging signed information. \
```
alonmuroch, line 422: I'm not sure I get the handshake messages. The way I see ethereum doing it is with status, goodbye and ping messages. Is that what we are trying to get here?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+The main purpose is for broadcasting messages to a group (AKA subnet) of nodes. \
+In addition, the following are achieved as well:
+
+- subscriptions metadata helps to get liveliness information of nodes
+- pubsub scoring enables to prune bad/malicious peers based on network behavior and application-specific rules
+
+The following parameters are used for initializing pubsub:
+
+- `floodPublish` was turned on for better reliability, as peer's own messages will be propagated to a larger set of peers
```
alonmuroch, line 485: we should keep it on?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+- subscriptions metadata helps to get liveliness information of nodes
+- pubsub scoring enables to prune bad/malicious peers based on network behavior and application-specific rules
+
+The following parameters are used for initializing pubsub:
+
+- `floodPublish` was turned on for better reliability, as peer's own messages will be propagated to a larger set of peers 
+  (see [Flood Publishing](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing)) 
+- `peerOutboundQueueSize` / `validationQueueSize` were increased to `600`, to avoid dropped messages on bad connectivity or slow processing
+- `directPeers` includes the exporter peer ID, to ensure it gets all messages
```
alonmuroch, line 488: I don't think we should have privileged nodes in the network, the exporter should be able to do its job without being a preferred node

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+- `subscriptionFilter` was injected to ensure a peer will connect to relevant topics, see [SubscriptionFilter](https://github.com/libp2p/go-libp2p-pubsub/blob/master/subscription_filter.go) interface
+- (fork `v1`) `msgID` is a custom function that calculates a `msg-id` based on the message content hash.
```
alonmuroch, line 490: this needs better explanation. We basically say that each network message some deterministic msg ID which prevents re-sending the same messages time and time again?

docs/specs/NETWORKING.md
alonmuroch: Also, as this is the network spec we should maybe define a network message struct in which different types of messages are set?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+- `peerOutboundQueueSize` / `validationQueueSize` were increased to `600`, to avoid dropped messages on bad connectivity or slow processing
+- `directPeers` includes the exporter peer ID, to ensure it gets all messages
+- `subscriptionFilter` was injected to ensure a peer will connect to relevant topics, see [SubscriptionFilter](https://github.com/libp2p/go-libp2p-pubsub/blob/master/subscription_filter.go) interface
+- (fork `v1`) `msgID` is a custom function that calculates a `msg-id` based on the message content hash. 
+The default function uses the `sender` + `msg_seq` which we don't track, and enforces signature / verification for each message. 
+As all the messages are being verified using the share key, it would be redundant to it also in the pubusb level.
+Moreover, the default `msg-id` duplicates messages, causing it to be processed more than once, in case it was sent by multiple peers (e.g. decided message).
+- (fork `v1`) `signPolicy` was set to `StrictNoSign` (required for custom `msg-id`) to avoid producing and verifying message signatures in the pubsub router
```
alonmuroch, line 494: it means messages are not signed with the network private key?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+[Ethereum Node Records](https://github.com/ethereum/devp2p/blob/master/enr.md) is a format that holds peer information.
+Records contain a signature, sequence (for republishing record) and arbitrary key/value pairs. 
+
+`ENR` structure in `SSV.Network` consists of the following key/value pairs:
+
+| Key         | Value                                                          | Status          |
+|:------------|:---------------------------------------------------------------|:----------------|
+| `id`        | name of identity scheme, e.g. "v4"                             | Done            |
+| `secp256k1` | compressed secp256k1 public key, 33 bytes                      | Done            |
```
alonmuroch, line 574: what key is it?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+| `ip`        | IPv4 address, 4 bytes                                          | Done            |
+| `tcp`       | TCP port, big endian integer                                   | Done            |
+| `udp`       | UDP port, big endian integer                                   | Done            |
+| `type`      | node type, integer; 1 (operator), 2 (exporter), 3 (bootnode)   | Done (`v0.1.9`) |
+| `oid`       | operator id, 32 bytes                                          | Done (`v0.1.9`) |
+| `version`   | fork version, integer                                          | -               |
+| `subnets`   | bitlist, 0 for irrelevant and 1 for assigned subnet            | -               |
+
+#### Discovery Alternatives
```
alonmuroch, line 583: no need for alternatives in a spec, let's decided on a specific spec and go with it

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+As `ENR` has a size limit (`< 300` bytes), and therefore discv5 won't support multiple key/value pairs.
+
+Notes:
+- discv5 specifies [Topic Index](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-rationale.md#the-topic-index) that help to lookup relevant nodes in a smaller set of nodes, currently not fully implemented
+- See [Ethereum specs > phase 0 > p2p interface](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#why-are-we-using-discv5-and-not-libp2p-kademlia-dht) 
+details why discv5 was chosen over libp2p Kad DHT in Ethereum.
+
+In `v0` discv5 is used, `v1` TBD, this section will be updated once that work is complete.
+
+### Network ID
```
alonmuroch, line 598: There are 2 use cases for network ID.
1) in networking which is just to not spam and connect with peers from the wrong network
2) in signing messages so that if we by accident have the same private key for testnet and mainnet we won't have valid testnet sigs on mainnet.

Where are we using the network ID?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+Global/main topic is not needed in this solution and therefore will be reduced.
+
+**Validators Mapping**
+
+Validator's public key is mapped to a subnet using a hash function, 
+which helps to distribute validators across subnets in a balanced way:
+
+`hash(validatiorPubKey) % num_of_subnets`
+
+Deterministic mapping is ensured as long as the number of subnets doesn't change,
```
alonmuroch, line 646: I think we should have some number simulation behind this. 
Like say SSV has 100k/ 1M validators and 128 subnets, how many msgs a second each subnet will have.
The number needs to make sense

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+**TBD** A dynamic number of subnets (e.g. `log(num_of_peers)`) which requires a different approach.
+
+
+### Peers Connectivity
+
+In a fully-connected network, where each peer is connected to all other peers in the network,
+running nodes will consume many resources to process all network related tasks e.g. parsing, peers management etc.
+
+To lower resource consumption, the number of connected peers is limited, currently set to `250`. \
```
alonmuroch, line 657: peer limit should be via flag and not part of spec

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+In comparison to `v0`, the number of topics will be reduced and the number of messages sent over the network should grow. \
+As messages will be propagated to a larger set of nodes, we can expect better reliability (arrival of messages to all operators in the committee).
+
+In addition, a larger group of operators provides:
+- redundancy of decided messages across multiple nodes
+- better security for subnets as more nodes will validate messages and can score bad/malicious nodes that will be pruned accordingly.
+
+Global/main topic is not needed in this solution and therefore will be reduced.
+
+**Validators Mapping**
```
alonmuroch, line 639: Not sure where we should write it but having this mapping means that as an operator node, regardless of my shares, I'll save highest decided (and potentially historical decided if flag is set) for all the subnets I'm in at

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+**validator topic mapping**
+
+Validator public key is used as the topic name:
+
+`bloxstaking.ssv.<hex(validator-public-key)>`
+
+**message encoding/decoding**
+
+JSON is used for encoding/decoding of messages.
```
alonmuroch, line 697: Shouldn't we adopt SSZ like in eth2?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+**validator topic mapping**
+
+Validator public key hash is used to determine the validator's subnet which is the topic name:
+
+`bloxstaking.ssv.<hash(validatiorPubKey) % num_of_subnets>`
+
+### NAT port map
+
+libp2p enables to configure a `NATManager` that will attempt to open a port in the network's firewall using `UPnP`.
```
alonmuroch, line 715: Should be part of the spec?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+The following measures are used to protect against malicious peers and denial of service attacks:
+- Connection score is determined during discovery process, ensures that the node will try to connect only with relevant peers.
+- Connection filters determine the score for inbound connections, and terminates low scored connections
+- Connection gater protects against peers which were pruned in the past and tries to reconnect again before backoff timeout (5 min). 
+it kicks in in an early stage, before the other components processes the request to avoid resources consumption.
+
+
+### High Availability
+
+As it participates in a decentralized p2p network, HA for an ssv node is not trivial. \
```
alonmuroch, line 737: not sure it's part of this spec

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+
+### High Availability
+
+As it participates in a decentralized p2p network, HA for an ssv node is not trivial. \ 
+The reason is that running multiple instances might lead to slashing or 
+disturb the consensus as it creates ambiguity and conflicts.
+
+Ideas TBD:
+
+#### Hub Node
```
alonmuroch, line 743: I think simply enabling an operator admin key to register multiple operators is an easier solution no?

docs/specs/NETWORKING.md
```diff
@@ -1,3 +1,769 @@
...
+but it can't have multiple instances running because it is part of a p2p network (see above).
+
+A possible implementation could be a node with a proxied network layer 
+that uses websocket to communicate with worker nodes.
+
+The drawback is that the hub node becomes a point of failure, 
+and it's a single instance and therefore will have downtime, 
+but less than a full-blown stateful node.
+
+#### Subnet Partitions
```
alonmuroch, line 757: what is this for?

alonmuroch (changes requested):

docs/specs/NETWORKING.md
alonmuroch: We can follow this
https://github.com/bloxapp/ssv/blob/ssv_spec/docs/spec/types/messages.go#L64-L68

docs/specs/NETWORKING.md
amir-blox: Why to reduce the last round change?

Handshake won't be needed, libp2p supports this case OOTB with [protocol negotiation](https://github.com/libp2p/specs/blob/master/connections/README.md#protocol-negotiation).

amir-blox: Also, regards operators storing decided messages, do you refer to their validators or all the validators in the subnets they are part of?

docs/specs/NETWORKING.md
amir-blox: Pubsub is based on streams, and the network key is used to establish secured connections/streams ([go-libp2p-noise](https://github.com/libp2p/go-libp2p-noise)).

The idea is to reduce to work done in pubsub level, as pubsub router by default signs and verifies messages (https://github.com/libp2p/specs/blob/master/pubsub/README.md#message-signing)

docs/specs/NETWORKING.md
amir-blox: I'll add a table with 4/5 rows to show numbers
Calculated the best case where each operator get every message (once) for all committees in subnet:
`msgs_in_subnet = msgs_per_committee * operators_per_subnet * committees_per_subnet`

amir-blox:
TODO: complete handshake/identify protocol

PR #522, opened 2022-02-01
Exporter: websocket pings in high load

amir-blox:


amir-blox:
Was fixed as part of #525

PR #523, opened 2022-02-01
add operator pk auto generate flag

nivBlox:
Add flag (default false) to whether generate operator private key if there is no exist one or passed from config.

If flag is false and no key exist or passed, panic will raise

### DIFF
--- a/utils/migrationutils/migrator_test.go
+++ b/utils/migrationutils/migrator_test.go
@@ -2,15 +2,13 @@ package migrationutils
 
 import (
 	"github.com/stretchr/testify/require"
-	"io/ioutil"
 	"os"
 	"path"
 	"testing"
 )
 
 func TestMigrate(t *testing.T) {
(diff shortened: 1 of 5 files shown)


utils/migrationutils/migrator_test.go
```diff
@@ -2,13 +2,15 @@ package migrationutils
...
 	"github.com/stretchr/testify/require"
+	"io/ioutil"
 	"os"
 	"path"
 	"testing"
 )
 
 func TestMigrate(t *testing.T) {
-	tmpPath := path.Join(t.TempDir(), "xxx")
+	tmpPath, err := ioutil.TempDir("", "xxx")
```
amir-blox, line 12: This will return the OS's default dir for temporary files (see `os.TempDir`)

Use `t.TempDir()` instead, which will ensure that whatever you do is cleaned and won't be conflicted with other tests

PR #536, opened 2022-02-21
Pubsub Topics Management

amir-blox:
- added `TopicManager` for orchestrating topics, subscriptions and their life cycle
- reduced lock for topics
  - using `sync.Map` instead (our scenario is few writes, multiple reads)
  - added an atomic state to topics


amir-blox:
Will be changed in network `v1`

PR #537, opened 2022-02-21
Exporter 2

amir-blox:


amir-blox:
Was merged in #539 and https://github.com/bloxapp/ssv/commit/f95d68683e47edd2cfe5db7107208aa04f2e4054

PR #540, opened 2022-02-23
Async to Sync processing ETH1 events

olegshmuelov:
Resolve https://github.com/bloxapp/ssv/issues/461

Move async to sync processing the ETH1 events
on node startup, all events should be scanned and processed before we can start their instance.

### DIFF
--- a/validator/controller.go
+++ b/validator/controller.go
@@ -137,9 +137,9 @@ func (c *controller) ListenToEth1Events(feed *event.Feed) {
 	defer sub.Unsubscribe()
 	for {
 		select {
-		case event := <-cn:
-			if err := c.ProcessEth1Event(*event); err != nil {
-				c.logger.Error("could not process event", zap.Error(err))
+		case e := <-cn:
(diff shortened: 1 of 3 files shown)


validator/controller.go
```diff
@@ -285,40 +306,28 @@ func (c *controller) GetValidatorsIndices() []spec.ValidatorIndex {
 }
 
 // handleValidatorAddedEvent handles registry contract event for validator added
-func (c *controller) handleValidatorAddedEvent(validatorAddedEvent abiparser.ValidatorAddedEvent) error {
-	pubKey := hex.EncodeToString(validatorAddedEvent.PublicKey[:])
+func (c *controller) handleValidatorAddedEvent(validatorAddedEvent abiparser.ValidatorAddedEvent) (*validatorstorage.Share, error) {
```
amir-blox, line 309: Let's change the name of the function, e.g. `createValidatorShare`

olegshmuelov: we'll stick with the current name for now.
this area will be refactored in eth1_v2_events scope

amir-blox (approved):

PR #542, opened 2022-03-08
config: add `OwnerAddress` field to `example_share.yaml`

### LINKED ISSUE #541: config: `OwnerAddress` is missing in `example_share.yaml`
**Describe the bug**
If `example_share.yaml` is copied and its placeholders are filled, it's not clear that `OwnerAddress` also needs to be added. Without it, the node outputs the `failed to create share object: empty share` error

**To Reproduce**
1. Copy `example_share.yaml`
2. Fill the copied file with your values
3. Leave the file without the `OwnerAddress` field
4. Run the node
5. See error

**Expected behavior**
Node doesn't output an error if `example_share.yaml` is copied and its placeholders are filled

**Additional context**
https://github.com/bloxapp/ssv/blob/main/validator/storage/share_opts.go#L23

nkryuchkov:
Closes #541 

As per issue description

### DIFF
--- a/config/example_share.yaml
+++ b/config/example_share.yaml
@@ -9,6 +9,7 @@ ssv:
       - NodeID: <node-id>
         PublicKey: <validator-pubkey>
         ShareKey: <share-priv-key>
+        OwnerAddress: <owner-address>
         Committee:
           <share-pubkey-node1>: <node1-id>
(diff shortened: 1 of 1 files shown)


nivBlox (approved):

nivBlox:
@nkryuchkov please squash and merge

PR #546, opened 2022-03-10
ibft/controller: fix panic in `ProcessSyncMessage`

nkryuchkov:
Fixes https://github.com/bloxapp/ssv/issues/538

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,7 +33,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_BUILD_REF $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
   only:
-    - fix/panic-in-ProcessSyncMessage
+    - stage
 
 Deploy ssv exporter to blox-infra-stage cluster:
   stage: deploy
@@ -68,7 +68,7 @@ Deploy ssv nodes to blox-infra-stage cluster:
     - .k8/scripts/deploy-ssv-nodes-yamls-on-stage-k8s.sh $DOCKER_REPO_INFRA_STAGE $CI_BUILD_REF ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
     - .k8/scripts/deploy-ssv-node-v1-yamls-on-stage-k8s.sh $DOCKER_REPO_INFRA_STAGE $CI_BUILD_REF ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT_V1 $SSV_NODES_MEM_LIMIT_V1
   only:
-    - fix/panic-in-ProcessSyncMessage
+    - stage
 
 
 #blox-infra-prod
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,7 +33,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_BUILD_REF $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
   only:
-    - stage
+    - fix/panic-in-ProcessSyncMessage
 
 Deploy ssv exporter to blox-infra-stage cluster:
   stage: deploy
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -33,7 +33,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_BUILD_REF $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
   only:
-    - stage
+    - fix/panic-in-ProcessSyncMessage
```
nivBlox, line 36: lets change back to stage

.gitlab-ci.yml
```diff
@@ -68,7 +68,7 @@ Deploy ssv nodes to blox-infra-stage cluster:
     - .k8/scripts/deploy-ssv-nodes-yamls-on-stage-k8s.sh $DOCKER_REPO_INFRA_STAGE $CI_BUILD_REF ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
     - .k8/scripts/deploy-ssv-node-v1-yamls-on-stage-k8s.sh $DOCKER_REPO_INFRA_STAGE $CI_BUILD_REF ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT_V1 $SSV_NODES_MEM_LIMIT_V1
   only:
-    - stage
+    - fix/panic-in-ProcessSyncMessage
```
nivBlox, line 71: lets change back to stage

ibft/controller/controller_sync.go
```diff
@@ -33,9 +34,18 @@ func (i *Controller) processSyncQueueMessages() {
 
 // ProcessSyncMessage processes sync messages
 func (i *Controller) ProcessSyncMessage(msg *network.SyncChanObj) {
+	i.logger.Info("ProcessSyncMessage: processing message")
```
nivBlox, line 37: i think we can remove this log, no?

ibft/controller/controller_sync.go
```diff
@@ -33,9 +34,18 @@ func (i *Controller) processSyncQueueMessages() {
...
 	currentInstaceSeqNumber := int64(-1)
 	if i.currentInstance != nil {
+		if i.currentInstance.State() == nil {
+			i.logger.Warn("ProcessSyncMessage: i.currentInstance != nil but i.currentInstance.State() == nil")
+		} else if i.currentInstance.State().SeqNumber == nil {
+			i.logger.Warn("ProcessSyncMessage: i.currentInstance != nil && i.currentInstance.State() != nil " +
+				"but i.currentInstance.State().SeqNumber == nil")
+		}
+	}
+	if i.currentInstance != nil && i.currentInstance.State() != nil && i.currentInstance.State().SeqNumber != nil {
```
nivBlox, line 48: what happens when if is false?

ibft/controller/controller_sync.go
```diff
@@ -33,9 +34,18 @@ func (i *Controller) processSyncQueueMessages() {
 
 // ProcessSyncMessage processes sync messages
 func (i *Controller) ProcessSyncMessage(msg *network.SyncChanObj) {
+	i.logger.Info("ProcessSyncMessage: processing message")
 	var lastChangeRoundMsg *proto.SignedMessage
 	currentInstaceSeqNumber := int64(-1)
 	if i.currentInstance != nil {
+		if i.currentInstance.State() == nil {
```
amir-blox, line 41: Would be better to get state once (`state := i.currentInstance.State()`) and then check that variable.

ibft/controller/controller_sync.go
```diff
@@ -33,9 +34,18 @@ func (i *Controller) processSyncQueueMessages() {
...
+			i.logger.Warn("ProcessSyncMessage: i.currentInstance != nil but i.currentInstance.State() == nil")
```
amir-blox, line 42: We should have a meaningful log here, e.g. `could not process message: IBFT instance doesn't have an active state`

Same for the log in line 44

nkryuchkov: I added these both logs just to try to catch in Kibana what exactly causes the panic, so I'm going to remove both of them before merging the PR

ibft/controller/controller_sync.go
nkryuchkov: @amir-blox Thanks, good idea

.gitlab-ci.yml
nkryuchkov: @nivBlox Sure, I kept it for a case if I still can find the panic in Kibana logs (to add more code). I don't see any logs in Kibana that could trigger this panic, so it seems that this panic is not reproduced in our deployment and panics in Kibana are caused by some other part of code. Therefore, I think we can remove debug logs now.

ibft/controller/controller_sync.go
nkryuchkov: Sure, I needed it for debugging and will delete it before merging

ibft/controller/controller_sync.go
nkryuchkov: @nivBlox If either of `i.currentInstance` or `i.currentInstance.State()` or `i.currentInstance.State().SeqNumber` is `nil`, we cannot do `i.currentInstance.State().SeqNumber.Get()`, so we don't update `lastChangeRoundMsg` and `currentInstanceSeqNumber`. Did I miss anything?

nivBlox (approved):

ibft/controller/controller_sync.go
nkryuchkov: I have removed the debug logs, so the amount of usages of `i.currentInstance.State()` now is the same as it was before the PR (3). But if I change the code so that it gets state once, I'll need to add an extra nested `if`, so I think we can leave it as it was if we don't keep debug logs

nkryuchkov: done

PR #557, opened 2022-03-20
fixtures: remove package

### LINKED ISSUE #86: remove fixtures package and use threshold instead
We can use the threshold package to generate keys and share sets, no need to have an additional dedicated package with hardcoded values.

nkryuchkov:
Fixes #86.


validator/test_utils.go
```diff
@@ -2,32 +2,44 @@ package validator
...
-	"time"
 )
 
 var (
 	refAttestationDataByts = _byteArray("000000000000000000000000000000003a43a4bf26fb5947e809c1f24f7dc6857c8ac007e535d48e6e4eca2122fd776b0000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000003a43a4bf26fb5947e809c1f24f7dc6857c8ac007e535d48e6e4eca2122fd776b")
 
-	//refSk                  = _byteArray("2c083f2c8fc923fa2bd32a70ab72b4b46247e8c1f347adc30b2f8036a355086c")
-	refPk = fixtures.RefPk
+	// refSk = _byteArray("2c083f2c8fc923fa2bd32a70ab72b4b46247e8c1f347adc30b2f8036a355086c")
+	refPk = _byteArray("a9cf360aa15fb1d1d30ee2b578dc5884823c19661886ae8b892775ccb3bd96b7d7345569a2aa0b14e4d015c54a6a0c54")
```
amir-blox, line 28: Did you kept the hard-coded keys because the tests failed?

nkryuchkov: @amir-blox yes, I don't know how to easily generate `refAttestationSplitSigs`, `refAttestationSig`, `refSigRoot` from a generated `refSk`, so I left the hard-coded values here

PR #576, opened 2022-05-16
Peer ID of outgoing sync messages

amir-blox:
This PR helps to be prepared for the new network version, where we reduced the `FromPeerID` field and therefore needs to mitigate it in older nodes.

The solution in this case is to save a reference of the sending `peer.ID`, specifically for highest decided requests, which might be complemented with additional requests for sync, and therefore we need the `peer.ID` to know who to call.

**TODO:** fix the lint error which was ignored in this PR


amir-blox:
Was mitigated in #582

PR #586, opened 2022-06-09
Merge exporter into operator

nkryuchkov:


amir-blox:
The exporter functionality after the refactoring is to expose an API and additional means to improve observability.
Therefore we can keep `exporter` folder with the following structure:
- `exporter`
  - `api`
    - `query_handlers.go` (+`query_handlers_test.go`) - high level handlers for the API
    - `readers.go` - helpers to read data
    - the other files in that package can be decoupled by extracting WebSocket stuff into a dedicated folder, but we can do it later, for now we can keep these files together
  - `metrics.go` - declares metrics and exports reporting of operators, to be used from `./operator/node.go`
  - `README.md`

operator/node.go
```diff
@@ -39,35 +45,54 @@ type Options struct {
...
 
 	forkVersion forksprotocol.ForkVersion
+
+	ws        api.WebSocketServer
+	wsAPIPort int
 }
 
 // New is the constructor of operatorNode
 func New(opts Options) Node {
+	qbftStorage := qbftstorage.New(opts.DB, opts.Logger, "attestation", opts.ForkVersion)
```
nivBlox, line 76: lets change "attestation" to `message.RoleTypeAttester.String()`

nkryuchkov: @nivBlox done

nkryuchkov:
@amir-blox done

PR #592, opened 2022-06-15
Cleanup and Preparations

amir-blox:
- align env vars
- change logs to `Warn` instead of `Error`
- clean redundant logs

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -157,9 +157,6 @@ var StartNodeCmd = &cobra.Command{
 		cfg.SSVOptions.Beacon = beaconClient
 		cfg.SSVOptions.ETHNetwork = eth2Network
 		cfg.SSVOptions.Network = p2pNet
-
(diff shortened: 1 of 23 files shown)


cli/operator/node.go
```diff
@@ -201,9 +198,9 @@ var StartNodeCmd = &cobra.Command{
 
 		validatorCtrl := validator.NewController(cfg.SSVOptions.ValidatorOptions)
 		cfg.SSVOptions.ValidatorController = validatorCtrl
-		if cfg.ReadOnlyMode {
-			cfg.SSVOptions.DutyExec = duties.NewReadOnlyExecutor(Logger)
-		}
+		//if cfg.ReadOnlyMode {
```
nivBlox, line 201: why not removing?

amir-blox: You can lookup `p2p host was configured` log,
I've added it as a field (`selfPeer`) to all p2p logs

nivBlox (approved):

PR #614, opened 2022-07-04
Stop ibft instance on share removal

olegshmuelov:


amir-blox:
@olegshmuelov where do you trigger the stop/close validator?

olegshmuelov:
> 

onShareRemove - is called when a validator was removed from one of the events below
 - handleValidatorRemovedEvent
 - handleOperatorRemovedEvent
 - handleAccountLiquidatedEvent

PR #616, opened 2022-07-05
Peers Distribution in Subnets

amir-blox:
This PR introduces some changes to apply balanced peers distribution across subnets.
The idea is to find the best set of peers, in a way that the node will have enough peers for all of its subnets.

Changes:

- find best peers based on the amount of subnets and the state of these subnets 
(subnet with less peers weight more)

- tag peers before calling trim by conn manager
TODO: switch back to libp2p's connection manager, as currently the trimming is done manually


amir-blox:
Details will be available in the [SIP](https://github.com/bloxapp/SIPs/pull/2) once approved

PR #633, opened 2022-07-14
V2 Alignment of Config Defaults

amir-blox:
Config alignment for V2:

- modified the default value of following configs so `v2` network won't merge with `v1`:
  - `NETWORK_ID`
  - `BOOTNODES` (**TODO**)
- changed pubsub topic prefix


amir-blox:
Note that this is a breaking change, need to update the other branches once this is merged

PR #643, opened 2022-07-25
Use round-robin leader selector

nkryuchkov:


protocol/v1/qbft/instance/leader/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	specqbft "github.com/bloxapp/ssv-spec/qbft"
+	spectypes "github.com/bloxapp/ssv-spec/types"
+
+	beaconprotocol "github.com/bloxapp/ssv/protocol/v1/blockchain/beacon"
+	qbftprotocol "github.com/bloxapp/ssv/protocol/v1/qbft"
+)
+
+// RoundRobin robin leader selection will always return the same leader
+type RoundRobin struct {
```
nivBlox, line 12: @nkryuchkov why not using implementation from spec?

nivBlox (changes requested):

PR #654, opened 2022-08-01
Use []byte instead of MsgID for identifier in qbft

nkryuchkov:


automation/qbft/scenarios/sync_failover.go
```diff
@@ -146,7 +146,8 @@ loop:
 	}
 	r.logger.Info("node #4 synced", zap.Int64("highest decided", int64(nextSeq)-1))
 
-	decides, err := ctx.Stores[3].GetDecided(message.ToMessageID(msgs[1].Message.Identifier), 0, nextSeq)
+	messageID := message.ToMessageID(msgs[1].Message.Identifier)
```
nivBlox, line 149: @nkryuchkov 
we need to change the `msgs[1].Message` to use spec message struct

exporter/api/query_handlers.go
```diff
@@ -39,7 +40,7 @@ func HandleDecidedQuery(logger *zap.Logger, qbftStorage qbftstorage.QBFTStore, n
 	identifier := spectypes.NewMsgID(pkRaw, message.RoleTypeFromString(string(nm.Msg.Filter.Role)))
```
nivBlox, line 40: lets rename identifier to msgID

nivBlox (changes requested):

automation/qbft/scenarios/sync_failover.go
nkryuchkov: @nivBlox Do you mean to change `*specqbft.SignedMessage` to `*specqbft.Message`?

nivBlox: @nkryuchkov yes

exporter/api/query_handlers.go
nkryuchkov: done

PR #658, opened 2022-08-01
Avoid Broadcasting Decided of Late Commit

amir-blox:
Reducing the number of decided messages that are propagated in the network.
We expect fullnodes to pick up the late commits on their own and therefore the complete decided messages are redundant (we might get 1-3 duplicates).


amir-blox:
As expected, seems like the exporter manage to pick up late commit on its own.

```
022-08-02T09:12:29.165280Z [35mDEBUG[0m controller/proccess_message.go:86 decided message was updated after late commit processing 
{
   ...
   "who": "ProcessLateCommitMsg", "seq": 3355, "identifier": "...", 
   "signers": [3], "updated_signers": [1,2,4,3]
}
```

PR #670, opened 2022-08-04
Align errors to spec

nkryuchkov:
Depends on https://github.com/bloxapp/ssv/pull/649 and branch `align-error-messages` of `ssv-spec`

Fixes:
- proposal.WrongHeight()
- proposal.WrongSignature()
- prepare.InvalidPrepareData()
- prepare.WrongSignature()
- commit.InvalidCommitData()
- commit.WrongHeight()
- commit.WrongSignature()
- roundchange.WrongHeight()
- roundchange.WrongSig()
- roundchange.JustificationMultiSigners()
- roundchange.JustificationInvalidSig()
- roundchange.JustificationInvalidPrepareData()


protocol/v1/qbft/validation/changeround/validate_justification.go
```diff
@@ -52,49 +52,8 @@ func (p *validateJustification) Run(signedMessage *specqbft.SignedMessage) error
...
-			return errors.Wrap(err, "change round could not get pubkey")
-		}
-		aggregated := pks.Aggregate()
-
-		if err = rcj.Signature.Verify(rcj, types.GetDefaultDomain(), spectypes.QBFTSignatureType, aggregated.Serialize()); err != nil {
-			return errors.Wrap(err, "change round could not verify signature")
+		if err := p.validateRoundChangeJustification(rcj, data, signedMessage); err != nil {
+			return fmt.Errorf("round change justification invalid: %w", err)
 		}
 	}
```
nivBlox, line 59: @nkryuchkov 
Regarding - `protocol/v1/qbft/validation/changeround/validate_justification.go:128`

this quorum check is from spec?

nivBlox (changes requested):

protocol/v1/qbft/instance/instance.go
```diff
@@ -253,29 +252,31 @@ func (i *Instance) Stopped() bool {
 
 // ProcessMsg will process the message
 func (i *Instance) ProcessMsg(msg *specqbft.SignedMessage) (bool, error) {
-	var pp pipelines.SignedMessagePipeline
-	var errPrefix string // TODO(nkryuchkov): make similar in ssv-spec
+	if err := msg.Validate(); err != nil {
```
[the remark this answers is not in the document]
nkryuchkov: @nivBlox What removal do you mean? I added this validation from spec: https://github.com/bloxapp/ssv-spec/blob/main/qbft/instance.go#L86

protocol/v1/qbft/validation/changeround/validate_justification.go
nkryuchkov: @nivBlox Not exactly. Here is how the check looked like in ssv:
```go
func (i *Instance) changeRoundQuorum(msgs []*specqbft.SignedMessage) (quorum bool, t int, n int) {
	quorum = len(msgs)*3 >= i.ValidatorShare.CommitteeSize()*2
	return quorum, len(msgs), i.ValidatorShare.CommitteeSize()
}
```

In spec it looks like this:
```go
// HasQuorum returns true if a unique set of signers has quorum
func HasQuorum(share *types.Share, msgs []*SignedMessage) bool {
	uniqueSigners := make(map[types.OperatorID]bool)
	for _, msg := range msgs {
		for _, signer := range msg.GetSigners() {
			uniqueSigners[signer] = true
		}
	}
	return share.HasQuorum(len(uniqueSigners))
}

func (share *Share) HasQuorum(cnt int) bool {
	return uint64(cnt) >= share.Quorum
}
```

So the most significant difference was checking length of messages instead of length of unique signers, so I added this check to `changeRoundQuorum` in ssv and renamed it to `HasQuorum`. There's another difference in quorum calculation, ssv does that with `len(msgs)*3 >= i.ValidatorShare.CommitteeSize()*2` but spec does this with `uint64(cnt) >= share.Quorum`, however, ssv's `Share` doesn't have the `Quorum` field, so it's impossible just to copy `HasQuorum` from spec, therefore I kind of merged it into what ssv had by checking unique signers length instead of messages length:

```go
func (i *Instance) changeRoundQuorum(msgs []*specqbft.SignedMessage) (quorum bool, t int, n int) {
	uniqueSigners := make(map[types.OperatorID]bool)
	for _, msg := range msgs {
		for _, signer := range msg.GetSigners() {
			uniqueSigners[signer] = true
		}
	}
	quorum = len(uniqueSigners)*3 >= i.ValidatorShare.CommitteeSize()*2
	return quorum, len(uniqueSigners), i.ValidatorShare.CommitteeSize()
}
```

nivBlox (approved):

nivBlox:
approved. let's run locally before merging

PR #671, opened 2022-08-05
Align qbft commit message processing with spec

nkryuchkov:
Depends on https://github.com/bloxapp/ssv/pull/670 and branch `align-error-messages` of `ssv-spec`

Fixes:
- commit.FutureRound()
- commit.PastRound()
- commit.WrongData1()
- commit.WrongData2()
- commit.NoPrevAcceptedProposal()


protocol/v1/qbft/instance/commit.go
```diff
@@ -17,8 +17,26 @@ import (
 
 // CommitMsgPipeline - the main commit msg pipeline
 func (i *Instance) CommitMsgPipeline() pipelines.SignedMessagePipeline {
+	validationPipeline := i.CommitMsgValidationPipeline()
 	return pipelines.Combine(
-		i.CommitMsgValidationPipeline(),
+		signedmsg.CheckProposal(i.State()),
```
nivBlox, line 22: @nkryuchkov it's redundant as we check it again in `protocol/v1/qbft/instance/commit.go:39`

protocol/v1/qbft/instance/partial_change_round.go
```diff
@@ -30,6 +31,8 @@ func (i *Instance) uponChangeRoundPartialQuorum() pipelines.SignedMessagePipelin
 			i.ResetRoundTimer()
 			//i.ProcessStageChange(qbft.RoundStateChangeRound)
 
+			i.State().ProposalAcceptedForCurrentRound.Store((*specqbft.SignedMessage)(nil))
```
nivBlox, line 34: why reseting when bump round?

protocol/v1/qbft/types.go
```diff
@@ -38,33 +38,36 @@ type State struct {
...
-	InputValue    atomic.Value // []byte
-	Round         atomic.Value // specqbft.Round
-	PreparedRound atomic.Value // specqbft.Round
-	PreparedValue atomic.Value // []byte
+	Height                          atomic.Value // specqbft.Height
+	InputValue                      atomic.Value // []byte
+	Round                           atomic.Value // specqbft.Round
+	PreparedRound                   atomic.Value // specqbft.Round
+	PreparedValue                   atomic.Value // []byte
+	ProposalAcceptedForCurrentRound atomic.Value // *specqbft.SignedMessage
```
nivBlox, line 46: @nkryuchkov this is just to verify propose on commit?

protocol/v1/qbft/validation/signedmsg/msg_round.go
```diff
@@ -1,18 +1,62 @@
...
+	"github.com/bloxapp/ssv/protocol/v1/qbft"
 	"github.com/bloxapp/ssv/protocol/v1/qbft/pipelines"
 )
 
 // ValidateRound validates round
 func ValidateRound(round specqbft.Round) pipelines.SignedMessagePipeline {
 	return pipelines.WrapFunc("round", func(signedMessage *specqbft.SignedMessage) error {
 		if round != signedMessage.Message.Round {
-			return errors.Errorf("message round (%d) does not equal state round (%d)", signedMessage.Message.Round, round)
+			return fmt.Errorf("round is wrong")
```
nivBlox, line 17: we need the round info in the log. 
let's map it instead or check prefix for err in tests

nivBlox (changes requested):

protocol/v1/qbft/instance/commit.go
nkryuchkov: @nivBlox Yes, the proposal is checked twice right now
1) here, because spec does this as the first check upon commit message
2) in `ValidateProposal` (`protocol/v1/qbft/instance/commit.go:39`) to avoid panic if `ValidateProposal` is called when `ProposalAcceptedForCurrentRound` is `nil` (it's called only when `ProposalAcceptedForCurrentRound` is not `nil`, however, if `ValidateProposal` is reused or called in tests, this validation might be needed)

So if we remove the check in 1) (here), the order of checks will be different than in spec, so if we remove the redundancy, I would prefer to do that in 2). Would you agree with that?

protocol/v1/qbft/instance/partial_change_round.go
nkryuchkov: @nivBlox It's according to spec: https://github.com/bloxapp/ssv-spec/blob/main/qbft/round_change.go#L72
However, [there's a to-do ](https://github.com/bloxapp/ssv-spec/blob/main/qbft/round_change.go#L71) to check if we actually need to do that, so I'm not sure if we need to keep that or not.

protocol/v1/qbft/types.go
nkryuchkov: @nivBlox it's also used [on prepare](https://github.com/bloxapp/ssv-spec/blob/main/qbft/prepare.go#L14) and [on proposal](https://github.com/bloxapp/ssv-spec/blob/main/qbft/proposal.go#L93).

protocol/v1/qbft/validation/signedmsg/msg_round.go
nkryuchkov: @nivBlox What if we add a log with the round info (probably even outside `ValidateRound`) and still return `round is wrong`? If that doesn't sound good to you, I will map it.

protocol/v1/qbft/instance/commit.go
nivBlox: As disscused, removing `signedmsg.CheckProposal(i.State()),` as we check it again on other func

nkryuchkov: @nivBlox I've checked if removing `signedmsg.CheckProposal(i.State())` breaks something and it does: the `no previous accepted proposal` test expects the `did not receive proposal for this round` error, but if we remove the check here, the error will be wrapped: `invalid commit message: did not receive proposal for this round`, so the test will fail.

protocol/v1/qbft/validation/signedmsg/msg_round.go
nkryuchkov: Added a log outside `ValidateRound`

protocol/v1/qbft/instance/pre_prepare.go
```diff
@@ -27,6 +27,9 @@ func (i *Instance) PrePrepareMsgPipeline() pipelines.SignedMessagePipeline {
 				return fmt.Errorf("could not get proposal data: %w", err)
 			}
 			i.containersMap[specqbft.ProposalMsgType].AddMessage(signedMessage, proposalData.Data)
+
+			i.State().ProposalAcceptedForCurrentRound.Store(signedMessage)
+
```
nivBlox, line 32: lets move it to be inside `UponPrePrepareMsg`

nkryuchkov: @nivBlox Good idea, done

nivBlox (approved):

PR #679, opened 2022-08-11
Align qbft prepare message processing with spec

nkryuchkov:
Depends on https://github.com/bloxapp/ssv/pull/671 and branch `align-error-messages` of `ssv-spec`

Fixes:
- prepare.MultiSigner()
- prepare.NoPreviousProposal()
- prepare.OldRound()
- prepare.FutureRound()
- prepare.WrongData()

### DIFF
--- a/protocol/v1/qbft/instance/forks/genesis/fork.go
+++ b/protocol/v1/qbft/instance/forks/genesis/fork.go
@@ -10,6 +10,7 @@ import (
 	"github.com/bloxapp/ssv/protocol/v1/qbft/instance/forks"
 	"github.com/bloxapp/ssv/protocol/v1/qbft/pipelines"
 	"github.com/bloxapp/ssv/protocol/v1/qbft/validation/changeround"
+	"github.com/bloxapp/ssv/protocol/v1/qbft/validation/prepare"
 	"github.com/bloxapp/ssv/protocol/v1/qbft/validation/preprepare"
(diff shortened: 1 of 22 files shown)


protocol/v1/qbft/instance/forks/genesis/fork.go
```diff
@@ -56,6 +57,7 @@ func (g *ForkGenesis) PrepareMsgValidationPipeline(share *beacon.Share, state *q
 		signedmsg.ValidateLambdas(identifier[:]),
 		signedmsg.ValidateSequenceNumber(state.GetHeight()),
 		signedmsg.AuthorizeMsg(share),
+		prepare.ValidatePrepareMsg(),
```
nivBlox, line 60: lets rename it to be more generic as it just checking signers

nivBlox (changes requested):

protocol/v1/qbft/instance/forks/genesis/fork.go
nkryuchkov: @nivBlox I will rename it but it will not be generic for other message types because spec uses the `prepare msg allows 1 signer` error message for this check (same for other types).

nkryuchkov: done

nivBlox (approved):

PR #693, opened 2022-08-23
Align proposal leftovers

nkryuchkov:


protocol/v1/qbft/instance/change_round.go
```diff
@@ -18,8 +18,14 @@ import (
 
 // ChangeRoundMsgPipeline - the main change round msg pipeline
 func (i *Instance) ChangeRoundMsgPipeline() pipelines.SignedMessagePipeline {
+	validationPipeline := i.ChangeRoundMsgValidationPipeline()
```
nivBlox, line 21: if tests are passing, why we need to align this error?

protocol/v1/qbft/instance/proposal.go
```diff
@@ -13,8 +13,17 @@ import (
 
 // ProposalMsgPipeline is the main proposal msg pipeline
 func (i *Instance) ProposalMsgPipeline() pipelines.SignedMessagePipeline {
+	validationPipeline := i.proposalMsgValidationPipeline()
+
+	// TODO: Add value check.
 	return pipelines.Combine(
-		i.proposalMsgValidationPipeline(),
+		pipelines.WrapFunc(validationPipeline.Name(), func(signedMessage *specqbft.SignedMessage) error {
+			if err := validationPipeline.Run(signedMessage); err != nil {
```
nivBlox, line 21: again, why need to align?

protocol/v1/qbft/validation/changeround/validate_justification.go
```diff
@@ -26,6 +27,14 @@ func Validate(share *beacon.Share) pipelines.SignedMessagePipeline {
 	}
 }
 
+// ValidateWithRound is the constructor of validateJustification with round
+func ValidateWithRound(share *beacon.Share, round specqbft.Round) pipelines.SignedMessagePipeline {
```
nivBlox, line 31: why we need with round? where in spec need it?

nivBlox (changes requested):

protocol/v1/qbft/instance/change_round.go
nkryuchkov: Because I don't like how the below looks like 
```go
	case specqbft.ProposalMsgType:
		if err := i.ProposalMsgPipeline().Run(msg); err != nil {
			if errors.Is(err, signedmsg.ErrWrongRound) {
				// NOTE: These 4 checks will be replaced by one in a future PR.
				i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
			}
			return false, fmt.Errorf("invalid proposal message: %w", err)
		}
	case specqbft.PrepareMsgType:
		if err := i.PrepareMsgPipeline().Run(msg); err != nil {
			if errors.Is(err, signedmsg.ErrWrongRound) {
				// NOTE: These 4 checks will be replaced by one in a future PR.
				i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
			}
			return false, err
		}
	case specqbft.CommitMsgType:
		if err := i.CommitMsgPipeline().Run(msg); err != nil {
			if errors.Is(err, signedmsg.ErrWrongRound) {
				// NOTE: These 4 checks will be replaced by one in a future PR.
				i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
			}
			return false, err
		}
	case specqbft.RoundChangeMsgType:
		if err := i.ChangeRoundMsgPipeline().Run(msg); err != nil {
			if errors.Is(err, signedmsg.ErrWrongRound) {
				// NOTE: These 4 checks will be replaced by one in a future PR.
				i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
			}
			return false, fmt.Errorf("invalid round change message: %w", err)
		}
```
		
In two cases I return just `err`, in other two a wrapped error, I did that just to pass tests, but in this PR I want to have the message processing same for all types, so I moved wrapped error inside. In my opinion, it would be much easier to understand.

protocol/v1/qbft/validation/changeround/validate_justification.go
nkryuchkov: https://github.com/bloxapp/ssv-spec/blob/75fdec56196df1c1f494fc54f6a10df838c27ce2/qbft/round_change.go#L218 here spec checks the round from the message, not message's justification, so we need to pass it from message

nivBlox (approved):

PR #694, opened 2022-08-24
Align prepare leftovers

nkryuchkov:


protocol/v1/qbft/instance/instance.go
```diff
@@ -257,44 +258,24 @@ func (i *Instance) ProcessMsg(msg *specqbft.SignedMessage) (bool, error) {
...
-		if err := i.ChangeRoundMsgPipeline().Run(msg); err != nil {
-			if errors.Is(err, signedmsg.ErrWrongRound) {
-				// NOTE: These 4 checks will be replaced by one in a future PR.
-				i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
-			}
-			return false, fmt.Errorf("invalid round change message: %w", err)
-		}
-	default:
+	pipelineMap := map[specqbft.MessageType]pipelines.SignedMessagePipeline{
+		specqbft.ProposalMsgType:    i.ProposalMsgPipeline(),
```
nivBlox, line 262: `ProcessMsg` func should be light. init a map and hold all pipelines i think is over kill. 
why no keep switch case and just handle the errors at the end instead of in each case?

protocol/v1/qbft/instance/prepare.go
```diff
@@ -37,8 +37,6 @@ func (i *Instance) PrepareMsgPipeline() pipelines.SignedMessagePipeline {
 func (i *Instance) PrepareMsgValidationPipeline() pipelines.SignedMessagePipeline {
 	return pipelines.Combine(
 		i.fork.PrepareMsgValidationPipeline(i.ValidatorShare, i.State()),
-		signedmsg.ValidateRound(i.State().GetRound()),
```
nivBlox, line 40: moved so somewhere else? if so, where?

protocol/v1/qbft/instance/prepare.go
```diff
@@ -102,34 +100,48 @@ func (i *Instance) uponPrepareMsg() pipelines.SignedMessagePipeline {
...
-		return nil
+
+		var errorPrp error
+		i.processPrepareQuorumOnce.Do(func() {
+			i.Logger.Info("prepared instance",
+				zap.String("identifier", hex.EncodeToString(i.State().GetIdentifier())), zap.Any("round", i.State().GetRound()))
+
+			// set prepared state
+			i.State().PreparedValue.Store(prepareData.Data) // passing the data as is, and not get the specqbft.PrepareData cause of msgCount saves that way
+			i.State().PreparedRound.Store(i.State().GetRound())
```
nivBlox, line 118: why use state round? a bit confusing

protocol/v1/qbft/instance/prepare.go
```diff
@@ -102,34 +100,48 @@ func (i *Instance) uponPrepareMsg() pipelines.SignedMessagePipeline {
...
-					i.Logger.Info("could not broadcast commit message", zap.Error(err))
-					errorPrp = e
-				}
-			})
-			return errorPrp
+		if quorum, _, _ := signedmsg.HasQuorum(i.ValidatorShare, i.containersMap[specqbft.PrepareMsgType].ReadOnlyMessagesByRound(i.State().GetRound())); !quorum {
+			return nil
+		}
+
+		if i.didSendCommitForHeightAndRound() {
```
nivBlox, line 107: We doing that check with `processPrepareQuorumOnce`. 
i think let's comment this out and add todo to check if need to replace

nivBlox (changes requested):

protocol/v1/qbft/instance/instance.go
nkryuchkov: @nivBlox 
The map looked to me easier to read than a switch. But a switch is fine too. However, I'd prefer to handle all errors in the end to avoid having 4 same checks like this:
```go
if errors.Is(err, signedmsg.ErrWrongRound) {
	// NOTE: These 4 checks will be replaced by one in a future PR.
	i.Logger.Debug(fmt.Sprintf("message round (%d) does not equal state round (%d)", msg.Message.Round, i.State().GetRound()))
}
```

protocol/v1/qbft/instance/prepare.go
nkryuchkov: @nivBlox `protocol/v1/qbft/instance/forks/genesis/fork.go:58`

protocol/v1/qbft/instance/prepare.go
nkryuchkov: @nivBlox according to spec https://github.com/bloxapp/ssv-spec/blob/main/qbft/prepare.go#L52

protocol/v1/qbft/instance/instance.go
nivBlox: @nivBlox yes thats what i meant

protocol/v1/qbft/instance/prepare.go
nkryuchkov: @nivBlox done

protocol/v1/qbft/instance/instance.go
nkryuchkov: @nivBlox ok, I'll change to a switch now

nkryuchkov: @nivBlox done

exporter/api/server.go
```diff
@@ -71,7 +71,12 @@ func (ws *wsServer) Start(addr string) error {
 		zap.String("addr", addr),
 		zap.Strings("endPoints", []string{"/query", "/stream"}))
 
-	err := http.ListenAndServe(addr, ws.router)
+	server := http.Server{
+		Addr:              addr,
```
nivBlox, line 75: @nkryuchkov what is this change for?

monitoring/metrics/handler.go
```diff
@@ -98,7 +100,13 @@ func (mh *metricsHandler) Start(mux *http.ServeMux, addr string) error {
 	})
 
 	go func() {
-		if err := http.ListenAndServe(addr, mux); err != nil {
+		server := &http.Server{
```
nivBlox, line 103: Why change for?

utils/boot_node/node.go
```diff
@@ -105,7 +108,13 @@ func (n *bootNode) Start(ctx context.Context) error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/p2p", handler.httpHandler)
 
-	if err := http.ListenAndServe(fmt.Sprintf(":%d", 5000), mux); err != nil {
+	server := http.Server{
```
nivBlox, line 111: what is the change for?

nivBlox (changes requested):

exporter/api/server.go
nkryuchkov: @nivBlox Alternatively, I can add a `// nolint: gosec` comment if you'd prefer

nivBlox: please add nolint and open a task on @amir-blox

nkryuchkov: @nivBlox done

nivBlox (approved):

PR #695, opened 2022-08-24
Align commit leftovers

nkryuchkov:


protocol/v1/qbft/instance/commit.go
```diff
@@ -80,40 +78,102 @@ upon receiving a quorum Qcommit of valid ⟨COMMIT, λi, round, value⟩ message
...
+			i.decidedMsg = agg
+			// mark instance commit
+			i.ProcessStageChange(qbft.RoundStateDecided)
+		})
+
+		return onceErr
 	})
 }
 
+func aggregateCommitMsgs(msgs []*specqbft.SignedMessage) (*specqbft.SignedMessage, error) {
```
nivBlox, line 111: can we use spec func?

nkryuchkov: @nivBlox no, it's not exported

nivBlox (approved):

PR #696, opened 2022-08-25
Align change round leftovers

nkryuchkov:


protocol/v1/qbft/instance/change_round.go
```diff
@@ -104,26 +105,22 @@ func (i *Instance) uponChangeRoundFullQuorum() pipelines.SignedMessagePipeline {
...
+				PrepareJustification:     highest.RoundChangeJustification,
+			}
 
-			if notPrepared {
-				proposalData = &specqbft.ProposalData{
-					Data:                     i.State().GetInputValue(),
-					RoundChangeJustification: i.containersMap[specqbft.RoundChangeMsgType].ReadOnlyMessagesByRound(i.State().GetRound()),
-				}
+			if prepared {
+				proposalData.Data = i.State().GetInputValue()
```
nivBlox, line 120: by spec if highest is prepared, need to use it's value. if not use from state

`// Chose proposal value.
		// If justifiedRoundChangeMsg has no prepare justification chose state value
		// If justifiedRoundChangeMsg has prepare justification chose prepared value
		valueToPropose := instanceStartValue
		if highestRCData.Prepared() {
			valueToPropose = highestRCData.PreparedValue
		}
`

nivBlox: qbft/round_change.go:45

protocol/v1/qbft/instance/change_round.go
```diff
@@ -104,26 +105,22 @@ func (i *Instance) uponChangeRoundFullQuorum() pipelines.SignedMessagePipeline {
 				return
 			}
 
-			notPrepared, highest, e := i.HighestPrepared(signedMessage.Message.Round)
+			prepared, highest, e := i.HighestPrepared(i.State().GetRound())
```
nivBlox, line 108: Spec is usin the msg round and not the state 

qbft/round_change.go:30

protocol/v1/qbft/instance/commit.go
```diff
@@ -13,14 +13,12 @@ import (
...
 	"github.com/bloxapp/ssv/protocol/v1/qbft/instance/msgcont"
 	"github.com/bloxapp/ssv/protocol/v1/qbft/pipelines"
-	"github.com/bloxapp/ssv/protocol/v1/qbft/validation/signedmsg"
 )
 
 // CommitMsgPipeline - the main commit msg pipeline
 func (i *Instance) CommitMsgPipeline() pipelines.SignedMessagePipeline {
 	validationPipeline := i.CommitMsgValidationPipeline()
 	return pipelines.Combine(
-		signedmsg.ProposalExists(i.State()),
```
nivBlox, line 23: Why remove this?

protocol/v1/qbft/instance/msgcont/inmem/inmem.go
```diff
@@ -138,7 +138,8 @@ func (c *messagesContainer) OverrideMessages(msg *specqbft.SignedMessage, data [
 }
 
 func (c *messagesContainer) PartialChangeRoundQuorum(stateRound specqbft.Round) (found bool, lowestChangeRound specqbft.Round) {
-	lowestChangeRound = specqbft.Round(100000) // just a random really large round number
+	seenLowestChangeRound := false
```
nivBlox, line 141: optimization?

protocol/v1/qbft/instance/msgcont/inmem/inmem_test.go
```diff
@@ -53,7 +53,7 @@ func TestFindPartialChangeRound(t *testing.T) {
 				protocoltesting.SignMsg(t, secretKeys, uids[1:2], generateConsensusMsg(0)),
 			},
 			false,
-			100000,
+			0,
```
nivBlox, line 56: why we need to change here?

protocol/v1/qbft/instance/prepare.go
```diff
@@ -21,7 +21,6 @@ func (i *Instance) PrepareMsgPipeline() pipelines.SignedMessagePipeline {
 	validationPipeline := i.PrepareMsgValidationPipeline()
 
 	return pipelines.Combine(
-		signedmsg.ProposalExists(i.State()),
```
nivBlox, line 24: why removing?

nivBlox (changes requested):

protocol/v1/qbft/instance/change_round.go
nkryuchkov: @nivBlox Thanks, I missed the inversion

protocol/v1/qbft/instance/commit.go
nkryuchkov: @nivBlox spec uses different errors texts upon prepare and commit if proposal doesn't exist, so it's not possible to use a generic function. It's checked inside `ValidateProposal`. This issue appeared in this PR because I changed error mapping to reversed.

protocol/v1/qbft/instance/msgcont/inmem/inmem.go
nkryuchkov: @nivBlox I don't like having a hardcoded value like `100000`, so this is to get rid of a hardcoded value

protocol/v1/qbft/instance/msgcont/inmem/inmem_test.go
nkryuchkov: @nivBlox I made `PartialChangeRoundQuorum`return `0` instead of `100000` if not found

nivBlox: let's make sure it's not affecting the qbft process

protocol/v1/qbft/instance/change_round.go
nkryuchkov: @nivBlox Spec has these lines:
```go
noPrevProposal := state.ProposalAcceptedForCurrentRound == nil && state.Round == newRound
prevProposal := state.ProposalAcceptedForCurrentRound != nil && newRound > state.Round
```
, where `state` is `i.State` (instance state) and `newRound` is `signedRoundChange.Message.Round` (`signedRoundChange` is the message for which `uponRoundChange` is called). 

Therefore, you're right that we need to pass message round into `HighestPrepared`, but we need to get messages inside `HighestPrepared` according to state round, and the lines above should be rewritten to these ones:
```go
noPrevProposal := i.State().GetProposalAcceptedForCurrentRound() == nil && i.State().GetRound() == round
prevProposal := i.State().GetProposalAcceptedForCurrentRound() != nil && round > i.State().GetRound()
```
Thanks

protocol/v1/qbft/instance/change_round.go
nkryuchkov: @nivBlox I wanted to handle prepared first and then not prepared, then reverted this change and forgot to put the exclamation sign back, thanks

protocol/v1/qbft/instance/prepare.go
nkryuchkov: same as https://github.com/bloxapp/ssv/pull/696#discussion_r955067698

protocol/v1/qbft/instance/msgcont/inmem/inmem_test.go
nivBlox: i mean, if 0 instead of 1000 is ok where we using this func

nkryuchkov: @nivBlox yes, it's used only in `uponChangeRoundPartialQuorum` and `0` is better than `100000` there because of check that round returned from `uponChangeRoundPartialQuorum` must be higher than the round in state.

protocol/v1/qbft/instance/change_round.go
nivBlox: ok got it

nivBlox (approved):

PR #710, opened 2022-09-13
Find Peers Enablement

amirylm:
**WIP**

Enable finding peers in case there is a low peer count on some topic.

- finalized discovery interface on top of discv5 (WIP)
- expose a function to connect with peers for a given validator (WIP)
- call connect peers upon failed attempts to find peers of topic (resolve https://github.com/bloxapp/ssv/issues/93)


amir-blox:
On hold, to be continued in future versions

PR #720, opened 2022-10-13
Delete shares by operator ID instead of operator PK on operator removal

nkryuchkov:
On an operator removal event, validator shares are deleted by operator PK. It's possible to create a new operator with a PK of an existing operator, and then remove the created operator, which will cause validator shares of the already existing operator to be removed. Validator shares should be deleted by operator ID (index) instead of PK.

### DIFF
--- a/operator/validator/event_handler.go
+++ b/operator/validator/event_handler.go
@@ -93,6 +93,7 @@ func (c *controller) handleOperatorRemovalEvent(
 	//	}
 	//}
 
+	// TODO: check by operator ID, not operator public key
 	isOperatorEvent := strings.EqualFold(od.PublicKey, c.operatorPubKey)
 	logFields := make([]zap.Field, 0)
 	if isOperatorEvent || c.validatorOptions.FullNode {
@@ -109,7 +110,7 @@ func (c *controller) handleOperatorRemovalEvent(
 		return logFields, nil
 	}
 
-	shares, err := c.collection.GetOperatorValidatorShares(od.PublicKey, false)
+	shares, err := c.collection.GetOperatorIDValidatorShares(event.OperatorId, false)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get all operator validator shares")
 	}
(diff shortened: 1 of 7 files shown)


pk910:
Heya,
Great to see a fast fix for this 👍 
I think the issue might have other effects that haven't been found yet and requires more work to be fixed consistently (see the TODO comments in your commit).
Wouldn't it be a better approach to just prevent duplicate registrations with the same pubkey on smart contract level?
That way you can avoid going through all problems that occur from a non unique pubkey.

operator/validator/event_handler.go
```diff
@@ -93,6 +93,7 @@ func (c *controller) handleOperatorRemovalEvent(
 	//	}
 	//}
 
+	// TODO: check by operator ID, not operator public key
```
olegshmuelov, line 96: Just to note:
in order to achieve this, the controller should be aware of the operator ID.
In some cases (when the node is not synced events wise), the controller will not know the operator ID until the OperatorRegistrationEvent of the current operator will be handled.

operator/validator/storage.go
```diff
@@ -147,6 +148,29 @@ func (s *Collection) GetOperatorValidatorShares(operatorPubKey string, enabled b
 	return res, err
 }
 
+// GetOperatorIDValidatorShares returns all not liquidated validator shares belongs to operator ID.
+func (s *Collection) GetOperatorIDValidatorShares(operatorID uint32, enabled bool) ([]*beaconprotocol.Share, error) {
```
amir-blox, line 152: Not sure that we have cases where we actually need the entire share object as result (currently we just need the public key).

Not urgent as this won't happen too often, so let's add a TODO here to check regards returning a slice of public keys instead of shares

amir-blox (approved):

operator/validator/event_handler.go
nkryuchkov: @olegshmuelov thanks

nkryuchkov:
@pk910 Hey, thanks for the suggestion. I think we need to keep the fix on the operator level anyway to not rely on the smart contract level. Adding a fix on a smart contract level might be a good option.

PR #736, opened 2022-11-07
Add metrics for attestation submission and message processing durations

nkryuchkov:
Added metrics:

- Attestation flow
    - Getting attestation data duration
    - Proposal stage duration
    - Prepare stage duration
    - Commit stage duration
    - Post consensus signatures collection duration
    - Attestation submission duration
    - Consensus duration
    - Full submission flow duration

- Message processing:
    - Message processing duration


protocol/v1/qbft/instance/commit.go
```diff
@@ -80,6 +82,11 @@ func (i *Instance) uponCommitMsg() pipelines.SignedMessagePipeline {
 			i.decidedMsg = agg
 			// mark instance commit
 			i.ProcessStageChange(qbft.RoundStateDecided)
+			messageID := message.ToMessageID(i.GetState().GetIdentifier())
+			metricsDurationStageCommit.
```
nivBlox, line 86: lets extract to a separate func

protocol/v1/qbft/instance/metrics.go
```diff
@@ -16,13 +23,24 @@ var (
 		Name: "ssv:validator:ibft_round",
 		Help: "IBFTs round",
 	}, []string{"identifier", "pubKey"})
+	metricsDurationStageProposal = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ssv:validator:duration_stage_proposal",
```
nivBlox, line 27: i think we can combine into 1 with type and pubkey 
no need for metric for each type

protocol/v1/validator/validator.go
```diff
@@ -107,6 +108,8 @@ func (v *Validator) Close() error {
 
 // Start starts the validator
 func (v *Validator) Start() error {
+	v.startTime = time.Now()
```
nivBlox, line 111: why this is for?

nivBlox (changes requested):

protocol/v1/qbft/instance/commit.go
nivBlox: yeah cause this is not part of the really pipeline so need to be more clear and easy to read

protocol/v1/qbft/controller/metrics.go
```diff
@@ -21,17 +27,18 @@ var (
 		Name: "ssv:validator:running_ibfts_count",
 		Help: "Count running IBFTs by validator pub key",
 	}, []string{"pubKey"})
+	metricsDurationAttestation = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "ssv:validator:duration_attestation",
+		Help:    "Attestation duration (seconds)",
+		Buckets: []float64{0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 60},
```
nivBlox, line 33: you think we need those high buckets? like the 10 and 60

nivBlox:
@nkryuchkov please update grafana files with the new metrics. once done we can approve pr

protocol/v1/qbft/controller/metrics.go
nkryuchkov: I thought it might be useful to check when consensus is stuck. Until 5s or 10s could mean just 1 missed round. But I don't mind removing 10 and 60

nivBlox: above 1.5s it's already not good so i think the heights bucket should be 1.5

nivBlox (approved):

PR #739, opened 2022-11-14
align instance container

nivBlox:


protocol/v2/commons/spec_adapters.go
```diff
@@ -29,6 +29,14 @@ func (sa *storageAdapter) GetHighestDecided(identifier []byte) (*qbft2.SignedMes
 	return sa.store.GetLastDecided(identifier)
 }
 
+func (sa *storageAdapter) SaveInstanceState(state *qbft2.State) error {
```
amir-blox, line 32: do we want to keep with the adapter? we can remove it now

amir-blox (approved):

PR #743, opened 2022-11-18
Deprecate v1

nkryuchkov:


operator/validator/event_handler.go
```diff
@@ -59,7 +58,7 @@ func (c *controller) handleOperatorRegistrationEvent(event abiparser.OperatorReg
 	}
 
 	logFields := make([]zap.Field, 0)
-	if strings.EqualFold(eventOperatorPubKey, c.operatorPubKey) || c.validatorOptions.FullNode {
+	if strings.EqualFold(eventOperatorPubKey, c.operatorPubKey) || c.validatorOptions.Mode == validator.ModeRW {
```
amir-blox, line 61: Validator mode (R - non committee, RW - in committee) is a different notion than `fullnode` which is a flag in the node level that adds persistency of the entire decided history of all validators (explorer), and applies on all validators.

amir-blox:
`TestQBFTMapping/qbft_controller_late_commit_past_instance` is failing due to wrong generated jsons, will be fixed in followup commits

PR #744, opened 2022-11-22
Fixes

amir-blox:
* save share public key when parsing events
* fix concurrent map read write


protocol/v2/ssv/msgqueue/queue.go
```diff
@@ -187,14 +187,18 @@ func (q *queue) Peek(n int, idx Index) []*spectypes.SSVMessage {
 
 // WithIterator looping through all indexes and return true when relevant and pop
 func (q *queue) WithIterator(n int, peek bool, iterator func(index Index) bool) []*spectypes.SSVMessage {
+	q.itemsLock.RLock()
+
 	for k := range q.items {
 		if iterator(k) {
+			q.itemsLock.RUnlock()
```
nkryuchkov, line 194: It will be locked once and can be unlocked more than once if `iterator` returns `true` for more than one element. Can we copy the solution from https://github.com/bloxapp/ssv/pull/723?

nkryuchkov (changes requested):

PR #745, opened 2022-11-23
Local events (shares)

olegshmuelov:
This PR will change the way we load local shares to run the node locally.


eth1/local_events.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (e *operatorRemovalEventYAML) toEventData() (interface{}, error) {
+	return abiparser.OperatorRemovalEvent{
+		OperatorId:   e.OperatorId,
+		OwnerAddress: common.HexToAddress(e.OwnerAddress),
+	}, nil
+}
+
+func (e *validatorRegistrationEventYAML) toEventData() (interface{}, error) {
+	var toByteArr = func(orig []string) ([][]byte, error) {
```
amir-blox, line 70: We can extract to function instead of defining it inline

eth1/local_events.go
```diff
@@ -0,0 +1,189 @@
+package eth1
```
lior-blox, line 1: @olegshmuelov lets add tests for it please

amir-blox (approved):

PR #753, opened 2022-11-30
Panic fixes

amir-blox:
* fix nil dereference
* fix msg q cleaning
* updated local net script exec flag


protocol/v2/qbft/instance/instance.go
```diff
@@ -145,12 +145,18 @@ func (i *Instance) SetConfig(config types.IConfig) {
 
 // GetHeight interface implementation
 func (i *Instance) GetHeight() specqbft.Height {
-	return i.State.Height
+	if s := i.State; s != nil {
```
lior-blox, line 148: @amir-blox @olegshmuelov in which case this is happening?
is it relevant to spec as well?

lior-blox (changes requested):

protocol/v2/qbft/instance/instance.go
amir-blox: @lior-blox changed the fix https://github.com/bloxapp/ssv/pull/753/commits/44410e9aae2e697cc485f186ed0d0fe9183b3b42

PR #756, opened 2022-12-02
Constant round duration

nkryuchkov:


ibft/storage/store.go
```diff
@@ -155,85 +153,6 @@ func (i *ibftStorage) CleanAllInstances(msgID []byte) error {
...
-		if err := sm.Decode(msg); err != nil {
-			return res, err
-		}
-		res = append(res, sm)
-	}
-	return res, nil
-}
-
-// CleanLastChangeRound cleans last change round message of some validator, should be called upon controller init
-func (i *ibftStorage) CleanLastChangeRound(identifier []byte) error {
```
nkryuchkov, line 207: We don't need to remove this as it should be run in the migration. We should take all shares, go through them and run this function for each share's identifier.

migrations/migration_9_remove_change_round_sync.go
```diff
@@ -0,0 +1,19 @@
+package migrations
+
+import (
+	"context"
+)
+
+var migrationRemoveChangeRoundSync = Migration{
+	Name: "migration_8_remove_change_round_sync",
```
nkryuchkov, line 8: Fix the copy-paste issue

migrations/migration_9_remove_change_round_sync.go
```diff
@@ -0,0 +1,19 @@
...
+	Run: func(ctx context.Context, opt Options, key []byte) error {
+		stores := opt.getRegistryStores()
```
nkryuchkov, line 10: Clean change round storage instead

network/p2p/p2p_sync.go
```diff
@@ -47,33 +47,7 @@ func (n *p2pNetwork) SyncHighestDecided(mid spectypes.MessageID) error {
...
-			n.msgRouter.Route(spectypes.SSVMessage{
-				MsgType: spectypes.SSVConsensusMsgType,
-				MsgID:   mid,
-				Data:    raw,
-			})
-		})
-	}()
-
-	return nil
+	panic("deprecated; remove after specqbft.Syncer is adjusted")
```
nkryuchkov, line 50: NOTE: Open an issue in spec to remove this as well. Link this PR in the issue

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -2,24 +2,14 @@ package roundtimer
...
-type RoundTimeout func(round specqbft.Round) time.Duration
-
-// DefaultRoundTimeout returns the default timeout function (base^round seconds).
-func DefaultRoundTimeout(base float64) RoundTimeout {
-	return func(round specqbft.Round) time.Duration {
-		roundTimeout := math.Pow(base, float64(round))
-		return time.Duration(float64(time.Second) * roundTimeout)
-	}
-}
+const DefaultRoundDuration = 2 * time.Second
```
nkryuchkov, line 12: Change as it's in SIP (use `RoundTimeout` name):
```go
// RoundTimeout returns the number of seconds until next timeout for a give round
func (i *Instance) RoundTimeout(round Round) uint64 {
	return 2
}
```

ibft/storage/store.go
```diff
@@ -155,85 +153,6 @@ func (i *ibftStorage) CleanAllInstances(msgID []byte) error {
...
-	n, err := i.db.DeleteByPrefix(prefix)
-	if err != nil {
-		return errors.Wrap(err, "failed to remove last change round")
-	}
-	i.logger.Debug("removed last change round", zap.Int("count", n),
-		zap.String("identifier", hex.EncodeToString(identifier)))
-	return nil
-}
-
-func (i *ibftStorage) CleanAllChangeRound() error {
```
nkryuchkov, line 223: We don't need to remove this as it should be run in the migration. We should take all shares, go through them and run this function for each share's identifier.

network/p2p/p2p_sync.go
nkryuchkov: cc @nivBlox

migrations/migration_9_remove_change_round_sync.go
```diff
@@ -0,0 +1,28 @@
...
+		validatorStorage := opt.validatorStorage()
+		shares, err := validatorStorage.GetAllValidatorShares()
+		if err != nil {
+			return err
+		}
+
+		qbftStorage := opt.qbftStorage()
+		for _, share := range shares {
+			messageID := spectypes.NewMsgID(share.ValidatorPubKey, spectypes.BNRoleAttester)
+			if err := qbftStorage.CleanAllInstances(messageID[:]); err != nil {
```
nivBlox, line 21: need to clean all change rounds not instances

migrations/migrations.go
```diff
@@ -65,10 +70,15 @@ func (o Options) validatorStorage() validatorstorage.ICollection {
 		Logger: o.Logger,
 	})
 }
+
 func (o Options) nodeStorage() operatorstorage.Storage {
 	return operatorstorage.NewNodeStorage(o.Db, o.Logger)
 }
 
+func (o Options) qbftStorage() qbftstorageprotocol.QBFTStore {
+	return qbftstorage.New(o.Db, o.Logger, spectypes.BNRoleAttester.String(), forksprotocol.GenesisForkVersion)
```
nivBlox, line 79: please add a todo for support multi duty types and not only attester.

nivBlox (approved):

ibft/storage/store.go
lior-blox: @nkryuchkov Can we move it to the migration file it self? so we wont have leftovers in the code?

amir-blox: We can keep this function instead to avoid redundant calls to read the shares, plus a call per identifier to clean.

lior-blox (changes requested):

ibft/storage/store.go
nkryuchkov: @amir-blox It's implemented wrong: It deletes by `append(i.prefix, []byte(lastChangeRoundKey)...)`, however, `SaveLastChangeRoundMsg` does `append(i.prefix, identifier, []byte(lastChangeRoundKey)...`. I don't see a quick way to fix this

amir-blox (approved):

PR #761, opened 2022-12-05
Priority queue

moshe-blox:
## Package `queue`
- Design in `protocol/v2/ssv/queue/README.md` (h/t @nivBlox)
- `State` is a summary of the validator's state (height/slot/running instance...)
- `SSVMessagePrioritizer.Prior` returns whether message A is of higher priority than message B within the given `State`
- `TestMessagePrioritizerSlice` is given slices of correctly-ordered messages, shuffles them repeatedly, and tests that `SSVMessagePrioritizer` reproduces the correct order. Otherwise it dumps human-readable table of the incorrect order.

## Package `validator`

- Replace usage of old `msgqueue` with `queue`


protocol/v2/ssv/queue/queue.go
```diff
@@ -0,0 +1,81 @@
...
+// Pop removes & returns the highest priority message with the given BeaconRole.
+// Returns nil if no message is found.
+func (q *PriorityQueue) Pop(role types.BeaconRole) *DecodedSSVMessage {
+	var (
+		highest        *DecodedSSVMessage
+		highestElement *list.Element
+	)
+	for e := q.messages.Front(); e != nil; e = e.Next() {
+		msg := e.Value.(*DecodedSSVMessage)
+		if msg.MsgID.GetRoleType() != role {
```
nivBlox, line 65: we will have queue for each role so why need to pop by role?

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -6,25 +6,32 @@ import (
...
 
 // HandleMessage handles a spectypes.SSVMessage.
 func (v *Validator) HandleMessage(msg *spectypes.SSVMessage) {
-	fields := []zap.Field{
+	v.logger.Debug("got message, pushing to queue",
 		zap.Int("queue_len", v.Q.Len()),
 		zap.String("msgType", message.MsgTypeToString(msg.MsgType)),
 		zap.String("msgID", msg.MsgID.String()),
+	)
+	decodedMsg, err := queue.DecodeSSVMessage(msg)
```
nivBlox, line 25: heavy process? if so, maybe worth to do it async in order not to block network

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -40,57 +47,41 @@ func (v *Validator) StartQueueConsumer(msgID spectypes.MessageID, handler Messag
...
-		//}
-		lastHeight := v.GetLastHeight(msgID)
+			// Construct a representation of the current state.
+			runner := v.DutyRunners.DutyRunnerForMsgID(msgID)
+			state := &queue.State{
+				HasRunningInstance: runner != nil && runner.HasRunningDuty() &&
+					runner.GetBaseRunner().State.RunningInstance != nil,
+				Height: v.GetLastHeight(msgID),
+				Slot:   0,
+				Quorum: v.Share.Quorum,
```
nivBlox, line 76: could be constant at init

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -40,57 +47,41 @@ func (v *Validator) StartQueueConsumer(msgID spectypes.MessageID, handler Messag
...
 
-		// clean all old messages. (when stuck on change round stage, msgs not deleted)
-		cleaned := v.Q.Clean(func(index msgqueue.Index) bool {
-			return index.H <= (lastHeight - 2) // remove all msg's that are 2 heights old. not post consensus & decided
-		})
-		if cleaned > 0 {
-			logger.Debug("indexes cleaned from queue", zap.Int64("count", cleaned))
+			// Pop the highest priority message and handle it.
+			msg := v.Q.Pop(msgID.GetRoleType())
+			handler(msg.SSVMessage)
```
nivBlox, line 84: check if msg is nil?

protocol/v2/ssv/validator/startup.go
```diff
@@ -19,6 +18,10 @@ func (v *Validator) Start() error {
 		if !ok {
 			return nil
 		}
+
+		// Initialize the message queue.
+		v.Q = queue.New(nil)
```
nivBlox, line 23: why nil?

nivBlox (changes requested):

protocol/v2/ssv/validator/startup.go
moshe-blox: There's no point passing a `MessagePrioritizer` at the beginning because we do so each time with `Sort` inside `ConsumeQueue`

protocol/v2/ssv/validator/msgqueue_consumer.go
moshe-blox: Yeah it's heavy, but decoding is expected to move to P2P layer (for validation).

In the short term, if this will cause slowdown, we can move decoding to a separate goroutine (as temporary solution until decoding moves to P2P layer).

moshe-blox: Yeah it's heavy, but decoding is expected to move to P2P layer (for validation).

In the short term, if this will cause slowdown, we can move decoding to a separate goroutine (as temporary solution until decoding moves to P2P layer).

nivBlox: @lior-blox @amir-blox FYI

amir-blox:
Continued in #801

PR #766, opened 2022-12-11
Protocol packages utils restructuring

amir-blox:
This PR changes the packages containing utilities, test mocks, etc.

- `testing` dir that exports testing utilities from some  package (added for `ssv`, `qbft`)
- added mock-p2p-network with messaging


operator/validator/controller.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rsa"
 	"encoding/hex"
+	commons2 "github.com/bloxapp/ssv/protocol/v2/qbft"
```
nkryuchkov, line 7: Could you please rename it from `commons2` to something else? E. g. `qbftprotocol`

protocol/v2/message/consensus_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"sort"
 	"testing"
 
-	testing2 "github.com/bloxapp/ssv/protocol/v2/testing"
+	qbfttesting "github.com/bloxapp/ssv/protocol/v2/testing"
```
nkryuchkov, line 7: Why is it `qbfttesting`, not `ssvtesting`?

protocol/v2/p2p/testing/utils.go
```diff
@@ -0,0 +1,408 @@
+package testing
+
+import (
+	"context"
+	crand "crypto/rand"
```
nkryuchkov, line 5: Why `crand`?

protocol/v2/p2p/testing/utils.go
```diff
@@ -0,0 +1,408 @@
...
+	"encoding/hex"
+	"encoding/json"
+	"github.com/bloxapp/ssv/protocol/v2/message"
+	protocolp2p "github.com/bloxapp/ssv/protocol/v2/p2p"
```
nkryuchkov, line 9: I think we need to decide at this point what names we use for importing packages inside `protocol`: If `protocol` is a prefix or a postfix. I used it as a postfix in a few places, but probably having it as a prefix is better, also we already have a `spec` prefix.

@lior-blox @nivBlox Your thoughts?

protocol/v2/p2p/testing/utils.go
```diff
@@ -0,0 +1,408 @@
...
+		if err := m.SendStreamMessage("last_changeround", pi, msg); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+//
+//func (m *mockNetwork) GetHistory(mid spectypes.MessageID, from, to specqbft.Height, targets ...string) ([]SyncResult, specqbft.Height, error) {
```
nkryuchkov, line 322: Do we need to keep it?

protocol/v2/p2p/testing/utils.go
amir-blox: to differentiate from `math/rand`

protocol/v2/message/consensus_test.go
amir-blox: I'll change to `protocoltesting`

nkryuchkov (approved):

PR #768, opened 2022-12-12
Revised logging

moshe-blox:
This PR tries to make logs:
- Quiet, to avoid spamming log indexers (such as Kibana)
- User-friendly, to help operators understand how their node is doing (or failing) with digestible messages

See more information in https://github.com/bloxapp/ssv/issues/676

**Changes**

- Package-level loggers (instead of passing loggers by dependency injection)
- Log levels are configurable per package with the `DebugServices` parameter (using https://github.com/ipfs/go-log)

From now on, the following rules should be applied to logs:

- Levels reserved for operators:
  - ERROR: operator should take action
  - WARN: operator should pay attention
- DEBUG is reserved for developers:
  - Non-critical errors (such as 'failed to broadcast' or 'invalid message from peer')
  - High-level tracing of consensus flow (instead of per p2p message)

### DIFF
--- a/protocol/v2/qbft/instance/round_change.go
+++ b/protocol/v2/qbft/instance/round_change.go
@@ -15,7 +15,7 @@ func (i *Instance) uponRoundChange(
 	roundChangeMsgContainer *specqbft.MsgContainer,
 	valCheck specqbft.ProposedValueCheckF,
 ) error {
-	if err := validRoundChange(i.State, i.config, signedRoundChange, i.State.Height, signedRoundChange.Message.Round); err != nil {
+	if err := validRoundChange(i.State, i.config, signedRoundChange, i.State.Height, i.State.Round); err != nil {
 		return errors.Wrap(err, "round change msg invalid")
 	}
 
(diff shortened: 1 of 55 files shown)


protocol/v2/qbft/instance/round_change.go
```diff
@@ -78,6 +83,9 @@ func (i *Instance) uponChangeRoundPartialQuorum(newRound specqbft.Round, instanc
 		return errors.Wrap(err, "failed to create round change message")
 	}
 
+	i.logger.Debug("got change round partial quorum, broadcasting change round message",
```
amir-blox, line 86: We might want to log this every 10 rounds similar to timeout

moshe-blox: @amir-blox Modulo on `new_round` or `old_round? Problem is they both seem to so skip numbers so modulo isn't reliable, no?

amir-blox: `newRound`, otherwise we see multiple logs in this flow..
it's fine that we won't see all the logs

lior-blox (changes requested):

amir-blox (approved):

PR #771, opened 2022-12-18
Round Change: linear timeout 

amir-blox:
- Changed from constant timeout (`2` sec) to a linear based timeout (`round*2` sec), with a limit on the round value (up to `10`)
- Sort by round descending to process newer messages first
- Fixed validate change round to ignore smaller rounds rather than not equal (**TBD** spec changes)


protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -11,9 +11,19 @@ import (
...
-func RoundTimeout(specqbft.Round) time.Duration {
-	return 2 * time.Second
+var (
+	quickTimeoutThreshold = specqbft.Round(8)
+	quickTimeout          = 2 * time.Second
+	slowTimeout           = 2 * time.Minute
+)
+
+// RoundTimeout returns the number of seconds until next timeout for a give round.
+// if the round is smaller than 3 -> 2s; otherwise -> 2m
```
amir-blox, line 21: change. to 8

amir-blox: and link to SIP

PR #772, opened 2022-12-19
Create a regular simulation

nkryuchkov:
The PR adds an integration test runner that allows custom scenarios for nodes connected to a local network and running duties. It also adds a regular scenario that runs 4 nodes and starts executing duties. 

Starting from now, the integration tests will run in the CI on creating a PR into `spec-align-qbft`


automation/qbft/scenarios/regular.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return startErr
+}
+
+func (r *regularScenario) PostExecution(ctx *runner.ScenarioContext) error {
+	for i, v := range r.validators {
+		var pk [48]byte
+		copy(pk[:], v.Share.ValidatorPubKey)
+
+		if err := v.StartDuty(&spectypes.Duty{
```
amir-blox, line 110: shouldn't we start the duty within Execute() ?

automation/qbft/runner/scenario.go
```diff
@@ -0,0 +1,50 @@
...
+	NumOfOperators() int
+	// NumOfBootnodes returns the desired number of bootnodes for the test.
+	// zero in case we want mdns
+	NumOfBootnodes() int
+	// NumOfFullNodes returns the desired number of full nodes for the test.
+	NumOfFullNodes() int
+}
+
+// Scenario represents a testplan for a specific scenario
+type Scenario interface {
```
amir-blox, line 40: We should have a cleaner interface, e.g. following spec tests with
`Name() string` and `Run(t *testing.T)`

automation/qbft/runner/scenario.go
```diff
@@ -0,0 +1,50 @@
...
+	LocalNet    *p2pv1.LocalNet
+	Stores      []*storage.QBFTStores
+	KeyManagers []spectypes.KeyManager
+	DBs         []basedb.IDb
+}
+
+// Bootstrapper bootstraps the given scenario.
+type Bootstrapper func(ctx context.Context, logger *zap.Logger, scenario Scenario) (*ScenarioContext, error)
+
+type scenarioCfg interface {
```
amir-blox, line 29: We should define something more verbose to allow us to specify the cases, these configs are just a subset of possible networking parameters.
Besides that we should define things like a slice of duties to execute and the expected states in each node

automation/qbft/scenarios/regular.go
```diff
@@ -0,0 +1,125 @@
...
+			ValidatorIndex:          spec.ValidatorIndex(i),
+			CommitteeIndex:          spectestingutils.TestingAttesterDuty.CommitteeIndex,
+			CommitteesAtSlot:        spectestingutils.TestingAttesterDuty.CommitteesAtSlot,
+			CommitteeLength:         spectestingutils.TestingAttesterDuty.CommitteeLength,
+			ValidatorCommitteeIndex: spectestingutils.TestingAttesterDuty.ValidatorCommitteeIndex,
+		}); err != nil {
+			return err
+		}
+	}
+
```
amir-blox, line 123: We should add assertions (e.g. check the controller state of each node)

automation/commons/shares.go
```diff
@@ -0,0 +1,129 @@
...
+		l := logger.With(zap.String("w", fmt.Sprintf("node-%d", i)))
+		val := validatorprotocol.NewValidator(ctx, options)
+		val.DutyRunners = validator.SetupRunners(ctx, l, options)
+		validators = append(validators, val)
+	}
+	return share, sks, validators, nil
+}
+
+// CreateShare creates a new beacon.Share
+func CreateShare(operators [][]byte) (*ssvtypes.SSVShare, map[uint64]*bls.SecretKey, error) {
```
amir-blox, line 96: let's avoid creating more utilities and use a single one.

following guidelines in https://github.com/bloxapp/ssv/pull/766

automation/qbft/scenarios/regular.go
```diff
@@ -0,0 +1,125 @@
...
+	var wg sync.WaitGroup
+	var startErr error
+	for _, val := range r.validators {
+		wg.Add(1)
+		go func(val *validator.Validator) {
+			defer wg.Done()
+			if err := val.Start(); err != nil {
+				startErr = errors.Wrap(err, "could not start validator")
+			}
+			<-time.After(time.Second * 3)
```
nkryuchkov, line 97: @amir-blox What do you think of getting rid of the dependency on the timeout in this PR?

amir-blox: Let's do it in a separate PR, we better do it once and cross the project

automation/qbft/scenarios/integration_test_type.go
```diff
@@ -0,0 +1,364 @@
...
+
+	dbs := make(map[spectypes.OperatorID]basedb.IDb)
+	for _, operatorID := range it.OperatorIDs {
+		db, err := storage.GetStorageFactory(basedb.Options{
+			Type:   "badger-memory",
+			Path:   "",
+			Logger: zap.L(),
+		})
+		if err != nil {
+			logger.Panic("could not setup storage", zap.Error(err))
```
amir-blox, line 74: let's return an error

nkryuchkov: done @amir-blox

amir-blox (approved):

PR #780, opened 2022-12-26
Fix data races

nkryuchkov:


amir-blox:
Fixed in https://github.com/bloxapp/ssv/pull/791

PR #785, opened 2023-01-03
Full node

moshe-blox:


protocol/v2/qbft/controller/instance_container.go
```diff
@@ -0,0 +1,110 @@
...
+func (c storageInstanceContainer) FindInstance(height specqbft.Height) *instance.Instance {
+	// Search in underlying.
+	if inst := c.InstanceContainer.FindInstance(height); inst != nil {
+		return inst
+	}
+
+	// Search in storage.
+	storedInst, err := c.config.GetStorage().GetInstance(c.messageID, height)
+	if err != nil {
+		c.logger.Error("could not load instance from storage", zap.Error(err))
```
amir-blox, line 125: `Warn` or even `Debug` level

protocol/v2/qbft/controller/controller.go
```diff
@@ -36,17 +36,22 @@ func NewController(
...
+	fullNode bool,
 ) *Controller {
-	return &Controller{
+	c := &Controller{
 		Identifier:          identifier,
 		Height:              specqbft.FirstHeight,
 		Domain:              domain,
 		Share:               share,
-		StoredInstances:     InstanceContainer{},
+		StoredInstances:     NewInstanceContainer(DefaultInstanceContainerCapacity),
```
amir-blox, line 46: shouldn't we use a larger capacity for fullnodes?

protocol/v2/qbft/controller/decided.go
```diff
@@ -45,31 +45,34 @@ func (c *Controller) UponDecided(msg *specqbft.SignedMessage) (*specqbft.SignedM
 	} else { // decide previously, add if has more signers
 		signers, _ := inst.State.CommitContainer.LongestUniqueSignersForRoundAndValue(msg.Message.Round, msg.Message.Data)
 		if len(msg.Signers) > len(signers) {
+			//TODO: we dont broadcast late decided anymore, so we should check if we can remove this
 			inst.State.CommitContainer.AddMsg(msg)
+			// TODO: update storage
 		}
 	}
 
+	if msg.Message.Height != c.Height {
```
amir-blox, line 54: It's sufficient to check previous decided here, no need to check height..
We can also move this block to where it was, below future check as we want to be aligned with spec.

The only difference here should be `SaveInstance` as we want to support fullnode

operator/validator/controller.go
```diff
@@ -63,7 +64,8 @@ type ControllerOptions struct {
 	Beacon                     beaconprotocol.Beacon
 	ShareEncryptionKeyProvider ShareEncryptionKeyProvider
 	CleanRegistryData          bool
-	FullNode                   bool `yaml:"FullNode" env:"FULLNODE" env-default:"false" env-description:"Flag that indicates whether the node saves decided history or just the latest messages"`
+	FullNode                   bool `yaml:"FullNode" env:"FULLNODE" env-default:"false" env-description:"Save decided history rather than just highest messages"`
+	FullNodeNonCommittee       bool `yaml:"FullNodeNonCommittee" env:"FULLNODE_NON_COMMITTEE" env-default:"false" env-description:"Save decided history for non committee validators as well"`
```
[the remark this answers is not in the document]
moshe-blox: `FullNode` by default only saves history for validators.
`FullNodeNonCommittee` is a special flag to sync history for all non committee validators

lior-blox (approved):

PR #792, opened 2023-01-08
Streams Deadline

amir-blox:
Fixing some issues with libp2p streams, that lead to disconnections and multiple warning logs.

- remove pending cache for handshake 
(caused deadlines in case of multiple handshake attempts)
- logs alignment
- added resource manager to apply stream limits (TODO: check in scale)
see [libp2p: dos mitigation](https://docs.libp2p.io/concepts/security/dos-mitigation/#leverage-the-resource-manager-to-limit-resource-usage-go-libp2p-only)

### DIFF
--- a/network/p2p/test_utils.go
+++ b/network/p2p/test_utils.go
@@ -77,7 +77,7 @@ func CreateAndStartLocalNet(pctx context.Context, loggerFactory LoggerFactory, f
 		go func(node network.P2PNetwork, i int) {
 			logger := loggerFactory(fmt.Sprintf("node-%d", i))
 			if err := node.Start(); err != nil {
-				logger.Error("could not start node", zap.Error(err))
+				logger.Warn("could not start node", zap.Error(err))
 				wg.Done()
 				return
 			}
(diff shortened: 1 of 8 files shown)


network/p2p/test_utils.go
```diff
@@ -94,7 +94,7 @@ func CreateAndStartLocalNet(pctx context.Context, loggerFactory LoggerFactory, f
 					logger.Fatal("could not find enough peers", zap.Int("n", n), zap.Int("found", len(peers)))
 					return
 				}
-				// logger.Debug("found enough peers", zap.Int("n", n), zap.Int("found", len(peers)))
+				logger.Debug("found enough peers", zap.Int("n", n), zap.Int("found", len(peers)))
```
nivBlox, line 97: not a too frequently log?

network/peers/connections/conn_handler.go
```diff
@@ -78,11 +78,10 @@ func (ch *connHandler) Handle() *libp2pnetwork.NotifyBundle {
 			return errors.New("reached peers limit")
 		}
 		if !ch.checkSubnets(conn) && conn.Stat().Direction != libp2pnetwork.DirOutbound {
-			_logger.Debug("disconnecting incoming connection after subnets check")
 			disconnect(net, conn)
 			return errors.New("peer doesn't share enough subnets")
 		}
-		_logger.Debug("new connection is ready",
+		logger.Debug("new connection is ready",
```
nivBlox, line 84: not a too frequently log?

network/p2p/test_utils.go
amir-blox: Only in local scenario (mdns)

nivBlox (approved):

PR #793, opened 2023-01-09
Round change & f+1 decided integration tests; validation functions

nkryuchkov:
Includes:
- round change integration test
- f+1 decided integration test
- validation functions for expected instances instead of structures for regular and round change scenarios (https://github.com/bloxapp/ssv/pull/799 implements them for f+1 decided scenario)


integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +61,160 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
-					},
-				},
-			}},
-			RoundChangeContainer: &specqbft.MsgContainer{Msgs: map[specqbft.Round][]*specqbft.SignedMessage{}},
-		},
-		DecidedMessage: &specqbft.SignedMessage{
-			Message: &specqbft.Message{
+				Data:       prepareData,
+			})
+			if !isMessageExistInRound(expectedMsg, actual.State.PrepareContainer.Msgs[specqbft.FirstRound]) {
```
nkryuchkov, line 116: @egorka-sereda 

If there're duplicated messages in the container (e.g. two from 1, one from 2, one from 3, one from 4), the test would pass, but needs to fail. 

Instead, could you please build a slice of 4 expected messages, sort them by the first signer, sort the actual result and then check if they are the same?

BTW, we might consider passing testing.T and using https://github.com/stretchr/testify because it has nice functions for assertions, which will remove a lot of code. For this case, https://pkg.go.dev/github.com/stretchr/testify/assert#Subset would be nice to use.

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +61,160 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
 				Identifier: identifier[:],
-				Data:       spectestingutils.PrepareDataBytes(consensusData),
+				Data:       commitData,
+			})
+			if isMessageExistInRound(expectedMsg, actual.State.CommitContainer.Msgs[specqbft.FirstRound]) {
+				foundedMsgsCounter++
+			}
+		}
+
+		if !actual.State.Share.HasQuorum(foundedMsgsCounter) {
```
nkryuchkov, line 132: @egorka-sereda 

If there're two messages from 1 and one message from 2, it will pass but should fail. We need to make sure that:
- each actual message exists in a set of possible messages
- it is not duplicated
- there are 3 actual messages (I think we won't have a case when we have 4 actual messages, let's check if it's 3)

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +61,160 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
+					Round:      specqbft.FirstRound,
+					Identifier: identifier[:],
+					Data:       spectestingutils.PrepareDataBytes(consensusData),
+				},
+			},
+		}
+
+		expectedStateRoot, err := expectedStoredInstance.State.GetRoot()
+		if err != nil {
+			return fmt.Errorf("error during geting root from expected state: %w", err)
```
nkryuchkov, line 173: @egorka-sereda 

Errors are usually chained, so if we have a similar message somewhere inside or outside, we will get something like `error during A: error during B: actual error`, which will be very long. Can we make it shorter? Something like this:
```suggestion
			return fmt.Errorf("get expected state root: %w", err)
```

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +61,160 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
+		}
+
+		actualStateRoot, err := actual.State.GetRoot()
+		if err != nil {
+			return fmt.Errorf("error during geting root from actual state: %w", err)
```
nkryuchkov, line 178: @egorka-sereda same as above

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +61,160 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
+		}
+
+		if !bytes.Equal(expectedStateRoot, actualStateRoot) {
+			return fmt.Errorf("expected and actual state roots doesn't matching, expected state = %+v, actual = %+v", expectedStoredInstance.State, actual.State)
```
nkryuchkov, line 182: @egorka-sereda 

- `expected and actual state roots doesn't matching` => `expected and actual state roots don't match`
- if we output both states here, the output will be unreadable because encoded states are very long, I think it's fine to return something like below and put a breakpoint and check states manually if test fails.
```suggestion
			return fmt.Errorf("expected and actual state roots don't match")
```

integration/qbft/scenarios/regular.go
nkryuchkov: @amir-blox Your thoughts about using `testify`?

integration/qbft/scenarios/integration_test_type.go
```diff
@@ -36,12 +37,20 @@ import (
...
-	ExpectedInstances map[spectypes.OperatorID][]*protocolstorage.StoredInstance
-	StartDutyErrors   map[spectypes.OperatorID]error
+	Name               string
+	OperatorIDs        []spectypes.OperatorID
+	Identifier         spectypes.MessageID
+	ValidatorDelays    map[spectypes.OperatorID]time.Duration
+	InitialInstances   map[spectypes.OperatorID][]*protocolstorage.StoredInstance
+	Duties             map[spectypes.OperatorID][]scheduledDuty
+	ExpectedInstances  map[spectypes.OperatorID][]*protocolstorage.StoredInstance
+	InstanceValidators map[spectypes.OperatorID][]func(*protocolstorage.StoredInstance) error
```
amir-blox, line 47: Are we keeping both `ExpectedInstances` and `InstanceValidators`?

nkryuchkov: @amir-blox `ExpectedInstances` is used in round change and f+1 decided, `InstanceValidators` is used in regular. This is an intermediate version to be able to run all tests and make sure they all pass. Afterwards, I will get rid of `ExpectedInstances`

integration/qbft/scenarios/integration_test_type.go
```diff
@@ -157,8 +167,11 @@ func (it *IntegrationTest) Run() error {
 
 	var eg errgroup.Group
```
amir-blox, line 168: using an errgroup here results that not all validators starts

egorka-sereda: @nkryuchkov

amir-blox (approved):

amir-blox (approved):

amir-blox:
In addition to the changes detailed in the PR description, we also added [avoid prehandshake](https://github.com/bloxapp/ssv/pull/793/commits/430005c66d907acd607f60874bd5dd2d15408105)

amir-blox (approved):

amir-blox (approved):

integration/qbft/scenarios/round_change.go
```diff
@@ -0,0 +1,285 @@
...
+		}).Encode()
+		if err != nil {
+			return fmt.Errorf("encode proposal data: %w", err)
+		}
+
+		prepareData, err := (&specqbft.PrepareData{
+			Data: consensusData,
+		}).Encode()
+		if err != nil {
+			panic(err)
```
amir-blox, line 106: return err instead of panic

integration/qbft/scenarios/round_change.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		commitData, err := (&specqbft.CommitData{
+			Data: consensusData,
+		}).Encode()
+		if err != nil {
+			panic(err)
```
amir-blox, line 113: return err instead of panic

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +62,152 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
-		Data: consensusData,
-	}).Encode()
-	if err != nil {
-		panic(err)
-	}
+		prepareData, err := (&specqbft.PrepareData{
+			Data: encodedConsensusData,
+		}).Encode()
+		if err != nil {
+			panic(err)
```
amir-blox, line 85: return err instead of panic

integration/qbft/scenarios/regular.go
```diff
@@ -57,115 +62,152 @@ func Regular(role spectypes.BeaconRole) *IntegrationTest {
...
-		Data: consensusData,
-	}).Encode()
-	if err != nil {
-		panic(err)
-	}
+		commitData, err := (&specqbft.CommitData{
+			Data: encodedConsensusData,
+		}).Encode()
+		if err != nil {
+			panic(err)
```
amir-blox, line 92: return err instead of panic

amir-blox (approved):

PR #797, opened 2023-01-09
TimeoutF To Runner

nivBlox:
Move register timeout to runner and inject timeoutF from validator

### DIFF
--- a/protocol/v2/ssv/validator/validator.go
+++ b/protocol/v2/ssv/validator/validator.go
@@ -48,13 +48,6 @@ func NewValidator(pctx context.Context, options Options) *Validator {
 
 	logger := logger.With(zap.String("validator", hex.EncodeToString(options.SSVShare.ValidatorPubKey)))
 
-	queues := make(map[spectypes.BeaconRole]msgqueue.MsgQueue)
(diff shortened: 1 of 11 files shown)


amir-blox:
What about spec changes? do we want to differ or patch these changes into spec?

nivBlox:
> 

differ, as those are implamantion concern

protocol/v2/ssv/validator/validator.go
```diff
@@ -65,9 +58,19 @@ func NewValidator(pctx context.Context, options Options) *Validator {
...
 		Signer:      options.Signer,
-		Queues:      queues,
+		Queues:      make(map[spectypes.BeaconRole]msgqueue.MsgQueue), // populate below
 		state:       uint32(NotStarted),
 	}
+
+	indexers := msgqueue.WithIndexers(msgqueue.SignedMsgIndexer(), msgqueue.DecidedMsgIndexer(), msgqueue.SignedPostConsensusMsgIndexer(), msgqueue.EventMsgMsgIndexer())
+	for _, dutyRunner := range options.DutyRunners {
+		// set timeout F
+		dutyRunner.SetTimeoutF(v.onTimeout)
```
lior-blox, line 68: @nivBlox Lets move the SetTimoutF to base runner rather then on each

nivBlox: done

amir-blox (approved):

PR #801, opened 2023-01-11
New Msg Q

amir-blox:
Finalizing #761


amir-blox:
Leftover: enable `TestMessageConsumer`

PR #808, opened 2023-01-15
Metrics and Dashboards

amirylm:


protocol/v2/ssv/runner/proposer.go
```diff
@@ -155,6 +172,10 @@ func (r *ProposerRunner) ProcessPostConsensus(signedMsg *specssv.SignedPartialSi
 		return nil
 	}
 
+	metricsPostConsensusDuration.
+		WithLabelValues(hex.EncodeToString(r.GetShare().ValidatorPubKey), spectypes.BNRoleProposer.String()).
```
amir-blox, line 176: we are doing too many hex encoding, you can do it once and reuse

nkryuchkov:
Still to be done:
- [ ] ETH1 registry events
- [ ] Change representation of durations, make them easier to understand
- [ ] Add more info to Validators 
- [ ] Add more info to Operator data
- [ ] Fix versions in Network Discovery, Row 3

protocol/v2/ssv/runner/proposer.go
nkryuchkov: @amir-blox I agree that we do it many times, but I'm not sure how to have a good design for reusing, could you please suggest one?

Dockerfile
```diff
@@ -35,7 +35,9 @@ RUN go get -d -v ./...
 
 ARG APP_VERSION
 
-RUN CGO_ENABLED=1 GOOS=linux go install -tags="blst_enabled,jemalloc,allocator" -ldflags "-X main.Version=`if [ ! -z "${APP_VERSION}" ]; then echo $APP_VERSION; else git describe --tags $(git rev-list --tags --max-count=1); fi` -linkmode external -extldflags \"-static -lm\"" ./cmd/ssvnode
+RUN git fetch --tags
+RUN if [ ! -z "${APP_VERSION}" ]; then echo $APP_VERSION; else echo `git describe --always --tags $(git rev-list --tags --max-count=1)`; fi
```
amir-blox, line 39: Why do we have 2 RUN commands

eth1/goeth/goETH.go
```diff
@@ -118,11 +118,13 @@ func (ec *eth1Client) HealthCheck() []string {
 	}
 	if sp != nil {
 		reportNodeStatus(statusSyncing)
+		metricsEth1LastSyncedBlock.Set(float64(sp.CurrentBlock))
```
amir-blox, line 121: this current block is the latest block that the ETH node is familiar with, we need to use the latest block that contains a registry event (e.g. register validator)

Dockerfile
nkryuchkov: @amir-blox The first RUN fetches information about tags, which is needed for `git describe`. The second RUN debug-prints the version, so we can check it in the pipeline, it will make debugging easier if passing version gets broken

protocol/v2/ssv/runner/proposer.go
amir-blox: Define it once, and use the variables instead of invoking the same functions again:
```go
pkHex, beaconRole := hex.EncodeToString(r.GetShare().ValidatorPubKey), spectypes.BNRoleProposer.String()

metric.WithLabelValue(phHex, beaconRole)...
```

Dockerfile
nkryuchkov: removed the second one

protocol/v2/ssv/runner/proposer.go
nkryuchkov: thanks, done

.gitlab-ci.yml
```diff
@@ -32,7 +32,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_BUILD_REF $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_BUILD_REF
   only:
-    - stage
+    - monitoring-upgrade
```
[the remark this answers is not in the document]
nkryuchkov: undone

nkryuchkov (approved):

lior-blox (approved):

nivBlox (approved):

moshe-blox (approved):

lior-blox:
@nkryuchkov let's include this fix or close it in this PR
#686

nkryuchkov:
> @nkryuchkov let's include this fix or close it in this PR #686

@lior-blox #686 it's already included into #808: https://github.com/bloxapp/ssv/blob/monitoring-upgrade/monitoring/grafana/dashboard_ssv_node.json#L1861

moshe-blox (changes requested):

moshe-blox (approved):

lior-blox (approved):

PR #824, opened 2023-01-30
Fix PriorityQueue data race

nkryuchkov:
Using `atomic.Pointer` instead of `unsafe.Pointer` in message queue


lior-blox (approved):

lior-blox:
closed due to the refactoring of queue #837

PR #841, opened 2023-02-08
Create stores for all role types on creating a new operator node

nkryuchkov:


exporter/api/query_handlers.go
```diff
@@ -4,21 +4,20 @@ import (
 	"encoding/hex"
 	"fmt"
 
-	spectypes "github.com/bloxapp/ssv-spec/types"
-
 	specqbft "github.com/bloxapp/ssv-spec/qbft"
+	spectypes "github.com/bloxapp/ssv-spec/types"
 	"go.uber.org/zap"
```
nivBlox, line 9: please sort the import

exporter/api/query_handlers.go
```diff
@@ -37,10 +36,18 @@ func HandleDecidedQuery(logger *zap.Logger, qbftStorage qbftstorage.QBFTStore, n
...
+		res.Data = []string{"malformed role"}
+		nm.Msg = res
+		return
+	}
+
+	msgID := spectypes.NewMsgID(pkRaw, beaconRole)
 	from := specqbft.Height(nm.Msg.Filter.From)
 	to := specqbft.Height(nm.Msg.Filter.To)
-	instances, err := qbftStorage.GetInstancesInRange(msgID[:], from, to)
+	instances, err := qbftStorage.Get(beaconRole).GetInstancesInRange(msgID[:], from, to)
```
nivBlox, line 50: add check in case no role storage exist

protocol/v2/message/msg.go
```diff
@@ -31,20 +32,19 @@ func MsgTypeToString(mt spectypes.MsgType) string {
 }
 
 // BeaconRoleFromString returns BeaconRole from string
-func BeaconRoleFromString(s string) spectypes.BeaconRole {
+func BeaconRoleFromString(s string) (spectypes.BeaconRole, error) {
```
nivBlox, line 35: why we need error here? let's return unknown role instead

nivBlox (changes requested):

exporter/api/query_handlers.go
nkryuchkov: @nivBlox 

I usually sort imports according this rule:
1. std packages
2. packages from other public repositories
3. packages from the current repository

Do you want to sort them according to another rule?

protocol/v2/message/msg.go
nkryuchkov: @nivBlox 
We need it to avoid panic if role in request is malformed. We don't have a `BeaconRole` for an unknown role.

exporter/api/query_handlers.go
nkryuchkov: @nivBlox 
done

exporter/api/query_handlers.go
nivBlox: yeah exactly

protocol/v2/message/msg.go
nivBlox: so we can return unknown role and the one that used this func should handle this case

nivBlox: in that case let's keep it with error

exporter/api/query_handlers.go
```diff
@@ -44,10 +44,18 @@ func HandleDecidedQuery(logger *zap.Logger, qbftStorage *storage.QBFTStores, nm
 		return
 	}
 
+	roleStorage := qbftStorage.Get(beaconRole)
+	if roleStorage == nil {
```
nivBlox, line 48: can you add test for this case?

nivBlox (changes requested):

exporter/api/query_handlers.go
nkryuchkov: @nivBlox done

nivBlox (approved):

PR #844, opened 2023-02-14
Helm charts

systemblox:
Deploy helm charts


.gitlab-ci.yml
```diff
@@ -35,48 +35,279 @@ Build stage Docker image:
     - apk add --no-cache py-pip
     - pip install pyyaml==5.3.1
     - pip install awscli
-    - docker build -t $IMAGE_NAME:$CI_BUILD_REF -f Dockerfile .
+    - docker build -t $IMAGE_NAME:$CI_BUILD_REF-$CI_COMMIT_REF_NAME -f Dockerfile .
```
y0sher, line 38: thanks for that 👍

.gitlab-ci.yml
```diff
@@ -35,48 +35,279 @@ Build stage Docker image:
...
-    - .k8/scripts/deploy-ssv-node-v3-yamls-on-stage-k8s.sh $DOCKER_REPO_INFRA_STAGE $CI_BUILD_REF ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
+    - mkdir ~/.kube/
+    - echo $STAGE_KUBECONFIG | base64 -d > kubeconfig
+    - mv kubeconfig ~/.kube/
+    - export KUBECONFIG=~/.kube/kubeconfig
+    - kubectl config get-contexts
+    - IMAGE_TAG=$CI_BUILD_REF-$CI_COMMIT_REF_NAME
+
+# SSV cluster A node 1 TEST
+    - helm upgrade
```
y0sher, line 60: can't we put all this into a script file with arguments?

.k8/helm3/ssv-node-cluster-b/node-b-1/templates/node-b-1-deployment.yaml
```diff
@@ -2,10 +2,10 @@
 apiVersion: v1
```
y0sher, line 2: I would really rather that we not have so many files, e.g for each cluster and node. if we could have a general file for a cluster and node and run it multiple times with different arguments (which name will be one of them). it will be much cleaner

PR #850, opened 2023-02-22
IO1-2633: Fix race conditions in integration tests

aaron-blox:
Fixes for race conditions in https://github.com/bloxapp/ssv/pull/830

See comments for explanation of race conditions

### DIFF
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -73,12 +73,11 @@ func (b *BaseRunner) baseStartNewDuty(runner Runner, duty *spectypes.Duty) error
 	}
 
 	// potentially incomplete locking of b.State. runner.Execute(duty) has access to
-	// b.State but currently this does not happen
-	func() {
-		b.mtx.Lock() // writes to b.State
-		defer b.mtx.Unlock()
-		b.State = NewRunnerState(b.Share.Quorum, duty)
-	}()
+	// b.State but currently does not write to it
+	state := NewRunnerState(b.Share.Quorum, duty)
+	b.mtx.Lock() // writes to b.State
+	b.State = state
+	b.mtx.Unlock()
 
 	return runner.executeDuty(duty)
 }
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -1,6 +1,8 @@
 package runner
 
 import (
+	"sync"
+
 	logging "github.com/ipfs/go-log"
 	"go.uber.org/zap"
 
@@ -49,6 +51,7 @@ type Runner interface {
 }
 
 type BaseRunner struct {
+	mtx            sync.RWMutex
 	State          *State
 	Share          *spectypes.Share
 	QBFTController *controller.Controller
@@ -68,7 +71,15 @@ func (b *BaseRunner) baseStartNewDuty(runner Runner, duty *spectypes.Duty) error
 	if err := b.canStartNewDuty(); err != nil {
 		return err
 	}
-	b.State = NewRunnerState(b.Share.Quorum, duty)
+
+	// potentially incomplete locking of b.State. runner.Execute(duty) has access to
+	// b.State but currently this does not happen
+	func() {
+		b.mtx.Lock() // writes to b.State
+		defer b.mtx.Unlock()
+		b.State = NewRunnerState(b.Share.Quorum, duty)
+	}()
+
 	return runner.executeDuty(duty)
 }
 
@@ -228,6 +239,9 @@ func (b *BaseRunner) decide(runner Runner, input *spectypes.ConsensusData) error
 
 // hasRunningDuty returns true if a new duty didn't start or an existing duty marked as finished
 func (b *BaseRunner) hasRunningDuty() bool {
+	b.mtx.RLock() // reads b.State
+	defer b.mtx.RUnlock()
+
 	if b.State == nil {
 		return false
 	}
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ full-test:
 .PHONY: integration-test
 integration-test:
 	@echo "Running integration tests"
-	@go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./integration/...
+	@go test -tags blst_enabled -count=1 -timeout 20m ${COV_CMD} -p 1 -v ./integration/...
 
 .PHONY: unit-test
 unit-test:
@@ -58,7 +58,7 @@ unit-test:
 .PHONY: spec-test
 spec-test:
 	@echo "Running spec tests"
-	@go test -tags blst_enabled -timeout 15m ${COV_CMD} -race -p 1 -v `go list ./... | grep spectest`
+	@go test -tags blst_enabled -timeout 15m ${COV_CMD} -race -count=1 -p 1 -v `go list ./... | grep spectest`
 
 #Test
 .PHONY: docker-spec-test
--- a/integration/qbft/tests/round_change_test.go
+++ b/integration/qbft/tests/round_change_test.go
@@ -1,15 +1,18 @@
 package tests
 
 import (
+	"testing"
+
 	"github.com/bloxapp/ssv-spec/qbft"
 	spectypes "github.com/bloxapp/ssv-spec/types"
 	"github.com/bloxapp/ssv/protocol/v2/qbft/roundtimer"
 	protocolstorage "github.com/bloxapp/ssv/protocol/v2/qbft/storage"
 	"github.com/stretchr/testify/require"
-	"testing"
 )
 
 func TestRoundChange4CommitteeScenario(t *testing.T) {
+	t.SkipNow() // TODO: test is flakey
+
 	roundChange := &Scenario{
 		Committee: 4,
 		Duties: map[spectypes.OperatorID]DutyProperties{
@@ -33,6 +36,7 @@ func TestRoundChange4CommitteeScenario(t *testing.T) {
 	//roundChange.Run(t, spectypes.BNRoleSyncCommitteeContribution) todo implement sync committee contribution role support
 }
 
+// nolint: unused
 func roundChangeValidator() func(t *testing.T, committee int, actual *protocolstorage.StoredInstance) {
 	return func(t *testing.T, committee int, actual *protocolstorage.StoredInstance) {
 		require.Equal(t, int(qbft.FirstHeight), int(actual.State.Height), "height not matching") //int conversion needs to show correct output from require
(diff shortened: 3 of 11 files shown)


aaron-blox:
## BaseRunner Race Condition
Base runner is is embedded/used/invoked by each duty runner. `hasRunningDuty()` has a race condition with `baseStartNewDuty()` which accessing the `State` field.

This is the call hierarchy for `hasRunningDuty()`:
image

It is called synchronously in `ConsumeQueue()` and asynchronously by the `Validator` `onTimeoutFunc()` in `NewValidator()`. The `onTimeoutFunc()` is called to set the BaseRunner `TimeoutF` field

image

This field is used to set the `done()` function of the `roundtimer.Timer`

image

image

the `done()` function is executed when a round times out

image

the `waitForRound()` method which calls `done()` is invoked in a goroutine which is invoked by a few different code paths
image

**So in summary, this race condition on the `State` field occurs when a round times out while the routine to consume messages from the queue is still proceeding.**

aaron-blox:
# RoundTimer race condition
The RoundTimer has a race condition between the `OnTimeout` and the `waitForRound` methods accessing the `done` field.

aaron-blox:
# RoundTimer race condition 

**tl;dr message processing and round timeout routines access the `RoundTimer` `done` field without any synchronization**

The race condition in `RoundTimer` between the `OnTimeout` and `waitForRound` methods has the same cause as the `BaseRunner` race condition, aka the RoundTimer `waitForRound()` method which calls `done()` is invoked in a goroutine which is not synchronised with any other routine. 

The call hierarchy for `waitForRound` was explained in the previous comment. That call path is in contention with the various code paths with update or set the `RoundTimer` `done` field. These various code paths have a choke point at `Validator.ProcessMessage`

image

aaron-blox:
# Validator race condition

The validator has a race condition between the `ConsumeQueue`, `HandleMessage`, and `Stop` methods. They try to access the `Queues` field of the validator without synchronization.

The `Stop` method gets called directly during the test on the main goroutine of the test
The `Stop` function gets called during the controller `onShareRemove` method which is invoked when processing various Eth1 events.

image

The `ConsumeQueue` method gets called in a goroutine started by the `Validator` `Start` method

image

The `HandleMessage` method gets called 
* by the `StartNetworkHandlers` `controller` method in a goroutine. `StartNetworkHandlers` is called by `operatorNode.Start`
image

image

* Via the p2p network message handler for `ssv-spec.SSVMessage`

image

aaron-blox:
Potential other issue. Validator Stop is called and after the roundtimeout happens anyway

aaron-blox:
we can remove the mutxes from the timer and add a new mutex to the baserunner when it sets the timer done function

aaron-blox:
in the future we could do an event loop. if we do we can get rid of goroutines for waiting for a timer. in the eventloop we just wait for the timeout to hit then do timeout stuff

moshe-blox:
> Potential other issue. Validator Stop is called and after the roundtimeout happens anyway

Just to add the solution we thought of:
-`Stop` write-locks and sets `Validator.state` to 'notStarted'
- `roundTimeout` read-locks & returns if `Validator.state` is not 'started'

aaron-blox:
> we can remove the mutxes from the timer and add a new mutex to the baserunner when it sets the timer done function

I don't think this actually will work, or I forgot part of the solution 

if we make that change, we control when the done func is set in relation to other things the baserunner is doing, but we still are not able to control when the done func read by the timer

the timer's `waitForRound` reads t.done when the timer ends, that read is still not coordinated with anything

aaron-blox:
Go 1.20 upgrade done in docker files/CI
Unit tests enabled i& pass in CI, unstable test is skipped

aaron-blox:
We can not use the NewBaseRunner func and instead make baserunner mutex not a pointer

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -39,19 +39,13 @@ func NewAggregatorRunner(
...
-			QBFTController: qbftController,
-			logger:         logger.With(zap.String("who", "BaseRunner")),
-		},
-		beacon:   beacon,
-		network:  network,
-		signer:   signer,
-		valCheck: valCheck,
-		logger:   logger.With(zap.String("who", "AggregatorRunner")),
-		metrics:  metrics.NewConsensusMetrics(share.ValidatorPubKey, spectypes.BNRoleAggregator),
+		BaseRunner: NewBaseRunner(logger, spectypes.BNRoleAggregator, beaconNetwork, share, qbftController),
```
lior-blox, line 42: @aaron-blox let's  revert to be more align to spec

protocol/v2/ssv/runner/runner.go
```diff
@@ -60,14 +63,22 @@ type BaseRunner struct {
...
+		Share:          share,
+		QBFTController: qbftController,
 	}
 }
 
 // baseStartNewDuty is a base func that all runner implementation can call to start a duty
 func (b *BaseRunner) baseStartNewDuty(runner Runner, duty *spectypes.Duty) error {
+	b.mtx.Lock() // writes to b.State
+	defer b.mtx.Unlock()
+
```
aaron-blox, line 81: Try to lock, ask canStartNewDuty & assign the new state, and then unlock (before executeDuty which is a heavy call [uses io])

aaron-blox:
We cant do minimal locking of v.Queues like this in HandleMessge because the queue gets deleted during processing after Stop() is called on the validator

```go
func (v *Validator) HandleMessage(msg *spectypes.SSVMessage) {
	id++
	var q queueContainer

	func() {
		v.mtx.RLock() // read v.Queues
		defer v.mtx.RUnlock()
		var ok bool
		q, ok = v.Queues[msg.MsgID.GetRoleType()]
		if ok {
			v.logger.Error("missing queue for role type", zap.String("role", msg.MsgID.GetRoleType().String()))
			return
		}
		fmt.Printf("%v: found queue\n", id)
	}()

	decodedMsg, err := queue.DecodeSSVMessage(msg)
	if err != nil {
		v.logger.Warn("failed to decode message",
			zap.Error(err),
			zap.String("msg_type", message.MsgTypeToString(msg.MsgType)),
			zap.String("msg_id", msg.MsgID.String()),
		)
		return
	}

	fmt.Printf("%v: %v, %v\n", id, q.Q == nil, decodedMsg == nil)
	if pushed := q.Q.TryPush(decodedMsg); !pushed {
		msgID := msg.MsgID.String()
		v.logger.Warn("dropping message because the queue is full",
			zap.String("msg_type", message.MsgTypeToString(msg.MsgType)),
			zap.String("msg_id", msgID))
	}
}
```

Logs summary:
```log
Stop() called, clearing the queue
444: found queue
444: true, false
2023-03-02T10:18:14.041+0200	DEBUG	ssv/protocol/ssv/validator	validator/msgqueue_consumer.go:133	queue consumer is closed	{"validator": "8e80066551a81b318258709edaf7dd1f63cd686a0e4db8b29bbb7acfe65608677af5a527d9448ee47835485e02b50bc0", "identifier": "8e80066551a81b318258709edaf7dd1f63cd686a0e4db8b29bbb7acfe65608677af5a527d9448ee47835485e02b50bc003000000"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x38 pc=0x105641ad8]

goroutine 1103 [running]:
panic({0x108f151e0, 0x109b0b730})
	/usr/local/go/src/runtime/panic.go:987 +0x3c0 fp=0xc0060cd340 sp=0xc0060cd280 pc=0x104e59dd0
runtime.panicmem()
	/usr/local/go/src/runtime/panic.go:260 +0x48 fp=0xc0060cd360 sp=0xc0060cd340 pc=0x104e58358
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:835 +0x1b8 fp=0xc0060cd390 sp=0xc0060cd360 pc=0x104e72958
github.com/bloxapp/ssv/protocol/v2/ssv/validator.(*Validator).HandleMessage(0xc005deb900, 0xc074a4b200)
	/Users/aaron/github/bloxapp/ssv/protocol/v2/ssv/validator/msgqueue_consumer.go:57 +0x288 fp=0xc0060cd6b0 sp=0xc0060cd3a0 pc=0x105641ad8
github.com/bloxapp/ssv/integration/qbft/tests.(*msgRouter).Route(0xc000010950, {0x0, {0x8e, 0x80, 0x6, 0x65, 0x51, 0xa8, 0x1b, 0x31, ...}, ...})
	/Users/aaron/github/bloxapp/ssv/integration/qbft/tests/msg_router.go:13 +0x9c fp=0xc0060cd6e0 sp=0xc0060cd6b0 pc=0x10600d5bc
github.com/bloxapp/ssv/network/p2p.(*p2pNetwork).handlePubsubMessages(0xc0036362c0, {0x1?, 0x15c?}, 0xc07480fc40)
	/Users/aaron/github/bloxapp/ssv/network/p2p/p2p_pubsub.go:197 +0x4b0 fp=0xc0060cd950 sp=0xc0060cd6e0 pc=0x106003070
github.com/bloxapp/ssv/network/p2p.(*p2pNetwork).handlePubsubMessages-fm({0xc000cba3f7, 0x9}, 0x2?)
	<autogenerated>:1 +0x50 fp=0xc0060cd9a0 sp=0xc0060cd950 pc=0x10600d3a0
github.com/bloxapp/ssv/network/topics.(*topicsCtrl).listen(0xc0031579e0, 0xc002be05a0)
	/Users/aaron/github/bloxapp/ssv/network/topics/controller.go:239 +0x404 fp=0xc0060cdd60 sp=0xc0060cd9a0 pc=0x105ff1424
github.com/bloxapp/ssv/network/topics.(*topicsCtrl).start(0xc0031579e0, {0xc000cba3f7, 0x9}, 0xc002be05a0)
	/Users/aaron/github/bloxapp/ssv/network/topics/controller.go:191 +0x74 fp=0xc0060cdf80 sp=0xc0060cdd60 pc=0x105ff0c34
github.com/bloxapp/ssv/network/topics.(*topicsCtrl).Subscribe.func2()
	/Users/aaron/github/bloxapp/ssv/network/topics/controller.go:143 +0x5c fp=0xc0060cdfd0 sp=0xc0060cdf80 pc=0x105ff032c
runtime.goexit()
	/usr/local/go/src/runtime/asm_arm64.s:1172 +0x4 fp=0xc0060cdfd0 sp=0xc0060cdfd0 pc=0x104e8ff94
created by github.com/bloxapp/ssv/network/topics.(*topicsCtrl).Subscribe
	/Users/aaron/github/bloxapp/ssv/network/topics/controller.go:143 +0x3e0
```

aaron-blox:
No idea why spectests timeout in CI, locally they run on-time

```
race condition fix - with locks: 307.022s        coverage: 89.9% of statements
stage (control case):                 320.969s        coverage: 89.9% of statements
race condition fix - no locks:    308.406s        coverage: 89.9% of statements
```

aaron-blox:
squashed all commits to change base to current stage

protocol/v2/ssv/runner/runner.go
```diff
@@ -68,7 +71,15 @@ func (b *BaseRunner) baseStartNewDuty(runner Runner, duty *spectypes.Duty) error
...
 	}
-	b.State = NewRunnerState(b.Share.Quorum, duty)
+
+	// potentially incomplete locking of b.State. runner.Execute(duty) has access to
+	// b.State but currently this does not happen
+	func() {
+		b.mtx.Lock() // writes to b.State
+		defer b.mtx.Unlock()
+		b.State = NewRunnerState(b.Share.Quorum, duty)
+	}()
```
moshe-blox, line 81: Can we simplify to this?

```go
state := NewRunnerState(b.Share.Quorum, duty)
b.mtx.Lock()
b.State = state
b.mtx.Unlock()
```

So that we don't need func + defer, and also this reduces the lock's scope a bit

moshe-blox (changes requested):

PR #853, opened 2023-02-22
Instance compaction improvements & goroutine leak fix

moshe-blox:
- Call `instance.Compact` after processing a message regardless of whether an error was returned
- Fixed goroutine leak caused by incorrect queue usage (`queue.Push`)


protocol/v2/ssv/runner/runner.go
```diff
@@ -156,6 +150,17 @@ func (b *BaseRunner) baseConsensusMsgProcessing(runner Runner, msg *specqbft.Sig
 	return true, decidedValue, nil
 }
 
+// Compacts the given message's associated instance if it's either...
```
lior-blox, line 153: @moshe-blox move to a separate file

lior-blox (approved):

lior-blox (approved):

PR #868, opened 2023-03-03
IO1-2505: logging: composite key for duty processing

nkryuchkov:
Create a composite key to set in logs which is unique per duty


protocol/v2/qbft/controller/controller.go
```diff
@@ -12,11 +12,8 @@ import (
 
 	"github.com/bloxapp/ssv/protocol/v2/qbft"
 	"github.com/bloxapp/ssv/protocol/v2/qbft/instance"
-	ipfslog "github.com/ipfs/go-log"
 )
 
-var logger = ipfslog.Logger("ssv/protocol/qbft/controller").Desugar()
```
aaron-blox, line 18: I dont want to get rid of this logger until we do this https://bloxxx.atlassian.net/browse/IO1-2503
we will be using the ipfs log in order to achieve the path based configuration

protocol/v2/qbft/instance/instance.go
```diff
@@ -4,18 +4,14 @@ import (
...
-
 	specqbft "github.com/bloxapp/ssv-spec/qbft"
 	spectypes "github.com/bloxapp/ssv-spec/types"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
 
 	"github.com/bloxapp/ssv/protocol/v2/qbft"
 )
 
-var logger = ipfslog.Logger("ssv/protocol/qbft/instance").Desugar()
```
aaron-blox, line 17: same comment as above

protocol/v2/ssv/runner/runner.go
```diff
@@ -1,22 +1,17 @@
...
 	spectypes "github.com/bloxapp/ssv-spec/types"
 	ssz "github.com/ferranbt/fastssz"
-
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 
 	"github.com/bloxapp/ssv/protocol/v2/qbft/controller"
 )
 
-var logger = logging.Logger("ssv/protocol/ssv/runner").Desugar()
```
aaron-blox, line 18: same comment as above

protocol/v2/ssv/validator/task_unique_id.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"fmt"
+
+	spectypes "github.com/bloxapp/ssv-spec/types"
+
+	"github.com/bloxapp/ssv/protocol/v2/ssv/runner"
+)
+
+func getTaskUniqueID(runner runner.Runner, duty *spectypes.Duty) string {
```
aaron-blox, line 11: move to logging package and name it
```suggestion
func TaskID(runner runner.Runner, duty *spectypes.Duty) string {
```

so we use it like `logging.TaskID`, maybe even naming it `logging.DutyID` is better because task is very generic

protocol/v2/ssv/validator/task_unique_id.go
```diff
@@ -0,0 +1,14 @@
...
+	epoch := runner.GetBaseRunner().BeaconNetwork.EstimatedEpochAtSlot(duty.Slot)
+	return fmt.Sprintf("<%v-e<%v>-s<%v>-v<%v>", duty.Type.String(), epoch, duty.Slot, duty.ValidatorIndex)
```
aaron-blox, line 13: please remove the `<` `>`

protocol/v2/ssv/validator/validator.go
```diff
@@ -2,22 +2,20 @@ package validator
...
+	"go.uber.org/zap"
 
 	"github.com/bloxapp/ssv/ibft/storage"
+	"github.com/bloxapp/ssv/protocol/v2/message"
 	"github.com/bloxapp/ssv/protocol/v2/ssv/queue"
 	"github.com/bloxapp/ssv/protocol/v2/ssv/runner"
 	"github.com/bloxapp/ssv/protocol/v2/types"
 )
 
-var logger = ipsflog.Logger("ssv/protocol/ssv/validator").Desugar() // TODO REVIEW
```
aaron-blox, line 19: same comment as first comment

protocol/v2/ssv/validator/validator.go
```diff
@@ -77,21 +75,36 @@ func NewValidator(pctx context.Context, cancel func(), options Options) *Validat
...
 
+	if br := dutyRunner.GetBaseRunner(); br != nil && br.State != nil {
+		var duty *spectypes.Duty
+		if dv := br.State.DecidedValue; dv != nil && dv.Duty != nil {
+			duty = dv.Duty
+		} else if d := br.State.StartingDuty; d != nil {
+			duty = d
+		}
+
+		logger = logger.With(zap.String("taskUniqueID", getTaskUniqueID(dutyRunner, duty)))
```
aaron-blox, line 105: why do we only set the logger only in this `if` condition?

protocol/v2/ssv/validator/task_unique_id.go
aaron-blox: and add a comment with an example of what the ID looks like

aaron-blox: and please write a unit test

protocol/v2/qbft/instance/instance.go
nkryuchkov: @aaron-blox There's an issue that fields which are necessary for logging are encoded in `StartValue`, so we need to decode it to get them, which breaks the abstraction since `StartValue`'s type is `[]byte` and technically can be anything using any encoding

protocol/v2/ssv/validator/validator.go
nkryuchkov: @aaron-blox To avoid nil pointer dereference in tests

aaron-blox: what is nil? duty? do we process messages with a nil duty? if we do, we should still set a trace ID for them

nkryuchkov: an example is `preconsensus.NoRunningDuty` in spec tests (state is nil), we don't start a duty, process a message and its validation fails, but we set a trace ID before the validation

protocol/v2/ssv/validator/task_unique_id.go
nkryuchkov: done

protocol/v2/ssv/validator/task_unique_id.go
nkryuchkov: done

protocol/v2/ssv/validator/validator.go
```diff
@@ -81,20 +82,26 @@ func (v *Validator) StartDuty(logger *zap.Logger, duty *spectypes.Duty) error {
 	if dutyRunner == nil {
 		return errors.Errorf("duty type %s not supported", duty.Type.String())
 	}
+
+	logger = logger.With(fields.DutyID(dutyRunner, duty))
```
nkryuchkov, line 86: this won't be included in post-consensus and duty submission logs

PR #871, opened 2023-03-06
Named loggers refactor

aaron-blox:


egorka-sereda:
moved https://github.com/bloxapp/ssv/pull/874 because of commits history problem after force-pushing stage

PR #882, opened 2023-03-09
set log name key

aaron-blox:


aaron-blox:
closing, is implemented in https://github.com/bloxapp/ssv/pull/883

PR #888, opened 2023-03-13
Log instance id when round change message is created

GalRogozinski:
There is some ambiguity in the log because theoretically several instances can be at the same height and round.
I decided to hex encode it because this is what I saw our explorer uses.


GalRogozinski:
Replaced by #890

PR #890, opened 2023-03-14
Log pubkey and role

GalRogozinski:
Currently my debugging is blocked because validator pubkey and role is not printing.

@nkryuchkov claims that logging should work without this due to logic in `validator/startup.go`. Until the bug is fixed this can be a temporary solution that later should be reverted.

This PR replaces #888


GalRogozinski:
apparently bug got fixed
Closing

PR #906, opened 2023-03-20
Stage candidate

aaron-blox:
Mostly logging changes


logging/fields/fields.go
```diff
@@ -195,23 +208,10 @@ func Count(val int) zap.Field {
...
 }
 
-func DutyID(runner runner.Runner, duty *spectypes.Duty) zap.Field {
-	return zap.Stringer(FieldDutyID, funcStringer{
-		fn: func() string {
-			epoch := runner.GetBaseRunner().BeaconNetwork.EstimatedEpochAtSlot(duty.Slot)
-			return fmt.Sprintf("T:%v::E:%v::S:%v::V:%v", duty.Type.String(), epoch, duty.Slot, duty.ValidatorIndex)
-		},
-	})
+func ConsensusTime(val time.Time) zap.Field {
```
moshe-blox, line 215: Computing durations on the logging side is not ideal, it's better to pass the already computed duration on the caller's side, because it's more predictable & precise

So just `func ConsensusTime(dur time.Duration)` would be enough 👍

Also, sometimes you wanna stop counting the duration now, but only log it later, which is impossible if you have to pass time.Time

protocol/v2/ssv/runner/runner_validations.go
```diff
@@ -15,6 +15,10 @@ func (b *BaseRunner) ValidatePreConsensusMsg(runner Runner, signedMsg *spectypes
 		return errors.New("no running duty")
 	}
 
+	// zap.L().Debug("got partial sig message (pre-consensus)", fields.PubKey(b.Share.ValidatorPubKey), fields.Role(b.BeaconRoleType),
```
moshe-blox, line 18: This is just a log I added to debug something, it's useless now, let's remove it :D

Ditto for the log below it (post-consensus)

PR #918, opened 2023-03-29
MEV support 

nkryuchkov:
Supersedes https://github.com/bloxapp/ssv/pull/870 and https://github.com/bloxapp/ssv/pull/856
Depends on https://github.com/bloxapp/ssv-spec/pull/226


beacon/goclient/proposer.go
```diff
@@ -34,20 +34,71 @@ func (gc *goClient) GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.Commi
 
 	switch beaconBlock.Version {
 	case spec.DataVersionPhase0:
+		if b := beaconBlock.Bellatrix.Body; b != nil && b.ExecutionPayload != nil {
+			gc.log.Info("got beacon block",
```
y0sher, line 38: is there a case where the body or payload are empty? shouldn't we log or do something in this case?

beacon/goclient/proposer.go
```diff
@@ -123,6 +174,116 @@ func (gc *goClient) SubmitBeaconBlock(block *spec.VersionedBeaconBlock, sig phas
...
+			Signature: sig,
+		},
+	}
+
+	gc.postponedRegistrations = append(gc.postponedRegistrations, signedReg)
+
+	return nil
+}
+
+type ValidatorRegistration struct {
```
y0sher, line 254: seems to be unused.

nkryuchkov: @y0sher Yes. This PR and the spec PR need a minor cleanup in the validator registration submission before being merged. `SubmitValidatorRegistrationBatched` is also to be removed.

beacon/goclient/proposer.go
nkryuchkov: I need to check this and I'm not sure if this log will remain

beacon/goclient/goclient.go
```diff
@@ -132,11 +142,13 @@ func New(logger *zap.Logger, opt beaconprotocol.Options) (beaconprotocol.Beacon,
...
-		client:         httpClient.(*http.Service),
-		indicesMapLock: sync.Mutex{},
-		graffiti:       opt.Graffiti,
+		log:               logger,
+		ctx:               opt.Context,
+		network:           network,
+		client:            httpClient.(*http.Service),
+		graffiti:          opt.Graffiti,
+		operatorID:        operatorID,
+		batchRegistration: true,
```
y0sher, line 151: it there a reason to not do batch registration? if so we should probably create a ticket/comment about adding configuration for this. otherwise let's remove this.

beacon/goclient/proposer.go
```diff
@@ -16,12 +16,19 @@ import (
 
 	apiv1bellatrix "github.com/attestantio/go-eth2-client/api/v1/bellatrix"
 	apiv1capella "github.com/attestantio/go-eth2-client/api/v1/capella"
+
+	"github.com/bloxapp/ssv/logging/fields"
+)
+
+const (
+	batchSize = 500
+	// TODO: This is a reasonable default, but we should probably make this configurable.
```
y0sher, line 25: ask @Tomblox1 to make tickets for this

beacon/goclient/proposer.go
```diff
@@ -38,16 +45,49 @@ func (gc *goClient) GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.Commi
...
+func (gc *goClient) GetBlindedBeaconBlock(slot phase0.Slot, graffiti, randao []byte) (ssz.Marshaler, spec.DataVersion, error) {
+	sig := phase0.BLSSignature{}
+	copy(sig[:], randao[:])
+
+	reqStart := time.Now()
+	beaconBlock, err := gc.client.BlindedBeaconBlockProposal(gc.ctx, slot, sig, graffiti)
+	if err != nil {
+		return nil, 0, err
+	}
+	metricsProposerDataRequest.Observe(time.Since(reqStart).Seconds())
```
y0sher, line 71: maybe add a label to this metric about this request being for blinded block

beacon/goclient/proposer.go
```diff
@@ -38,16 +45,49 @@ func (gc *goClient) GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.Commi
...
+
+	if beaconBlock == nil {
```
y0sher, line 73: I would add this to `GetBeaconBlock` as well.

beacon/goclient/proposer.go
```diff
@@ -38,16 +45,49 @@ func (gc *goClient) GetBeaconBlock(slot phase0.Slot, committeeIndex phase0.Commi
...
+		return nil, 0, fmt.Errorf("block is nil")
+	}
+
+	switch beaconBlock.Version {
+	case spec.DataVersionBellatrix:
+		if b := beaconBlock.Bellatrix.Body; b != nil && b.ExecutionPayloadHeader != nil {
```
y0sher, line 79: if `Body` is `nil` shouldn't we return an error?

beacon/goclient/proposer.go
```diff
@@ -123,6 +163,86 @@ func (gc *goClient) SubmitBeaconBlock(block *spec.VersionedBeaconBlock, sig phas
...
+		return gc.handleBatchValidatorRegistration(pubkey, feeRecipient, sig)
+	}
+
+	return gc.client.SubmitValidatorRegistrations(gc.ctx, []*api.VersionedSignedValidatorRegistration{
+		gc.createValidatorRegistration(pubkey, feeRecipient, sig),
+	})
+}
+
+func (gc *goClient) handleBatchValidatorRegistration(pubkey []byte, feeRecipient bellatrix.ExecutionAddress, sig phase0.BLSSignature) error {
+	gc.batchMu.Lock()
```
y0sher, line 177: should we lock all this function for sure? think if locking only when accessing a state that should be protected is possible.

beacon/goclient/signing.go
```diff
@@ -11,6 +12,21 @@ import (
 )
 
 func (gc *goClient) DomainData(epoch phase0.Epoch, domain phase0.DomainType) (phase0.Domain, error) {
+	if domain == types.DomainApplicationBuilder { // no domain for DomainApplicationBuilder. need to create.  https://github.com/bloxapp/ethereum2-validator/blob/v2-main/signing/keyvault/signer.go#L62
```
y0sher, line 15: is `DomainApplicationBuilder` created? if no is this code used?

docs/MEV_EDGE_CASES.md
```diff
@@ -0,0 +1,48 @@
+# MEV edge cases
```
y0sher, line 1: add a general paragraph about our MEV support, the CLI flag we need to set, a short explanation about how to set the beacon node to use MEV. do we support MEV-boost? do we NOT support certain MEV relays or all are supported? if there are any other known issues I would add them here.

ekm/signer_key_manager_test.go
```diff
@@ -35,7 +34,7 @@ func testKeyManager(t *testing.T) spectypes.KeyManager {
 	db, err := getBaseStorage(logger)
 	require.NoError(t, err)
 
-	km, err := NewETHKeyManagerSigner(db, beacon2.NewNetwork(core.PraterNetwork, 0), types.GetDefaultDomain(), logger)
```
y0sher, line 38: add tests with blinded on?

operator/duties/controller.go
```diff
@@ -241,6 +244,55 @@ func (dc *dutyController) handleSlot(logger *zap.Logger, slot phase0.Slot) {
 		go dc.onDuty(logger, &duties[i])
 	}
 
+	dc.handleSyncCommittee(logger, slot, syncPeriod)
```
y0sher, line 247: this isn't related to MEV, you just took out the sync committee code to a different func right?

operator/duties/controller.go
```diff
@@ -241,6 +244,55 @@ func (dc *dutyController) handleSlot(logger *zap.Logger, slot phase0.Slot) {
...
+	dc.handleValidatorRegistration(logger, slot)
+}
+
+func (dc *dutyController) handleValidatorRegistration(logger *zap.Logger, slot phase0.Slot) {
```
y0sher, line 251: does validatorRegistration count as a duty in eth? otherwise why is this code here?

y0sher: also this is added as part of MEV, should we do validatorregistration only when MEV is enabled?

operator/fee_recipient/controller.go
```diff
@@ -74,44 +71,61 @@ func (rc *recipientController) listenToTicker(logger *zap.Logger, slots chan pha
...
+		if end > len(shares) {
+			end = len(shares)
 		}
-		if err := g.Wait().ErrorOrNil(); err != nil {
-			logger.Warn("failed to submit proposal preparation", zap.Error(err))
-		} else {
-			logger.Debug("proposal preparation submitted", zap.Int32("count", atomic.LoadInt32(&counter)))
+		batch := shares[start:end]
+
+		count, err := rc.submit(logger, batch)
```
y0sher, line 101: is it possible that some were sent successfully and we still get error? if so it's not counted in submitted

y0sher: ok it seems unlikely, resolving this.

beacon/goclient/goclient.go
nkryuchkov: The submission process is much simpler if not batched, so it might be useful for debugging if the registration-related code is suspected in having bugs. However, straightforward submission is quick to implement, so I agree and will get rid of this

beacon/goclient/proposer.go
nkryuchkov: Done, I'll resolve this once @Tomblox1 confirms

beacon/goclient/proposer.go
nkryuchkov: I don't want to add another label to `ssv_beacon_data_request_duration_seconds` because it's generic for all duty types. I can create a new `role` value for blinded blocks, but that would require fixing the dashboards. Also, we might want to think what we want to show in Grafana. Therefore, I'd extract this out of scope of this PR and define first what changes in Grafana are needed. @y0sher Do you agree?

beacon/goclient/proposer.go
nkryuchkov: I didn't find any cases when it's `nil` when I did a quick look, but let's log it

protocol/v2/ssv/runner/proposer.go
```diff
@@ -237,7 +240,8 @@ func (r *ProposerRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *spe
 		logger.Info("✅ successfully submitted block proposal",
 			fields.Slot(signedMsg.Message.Slot),
 			fields.Height(r.BaseRunner.QBFTController.Height),
-			fields.Round(r.GetState().RunningInstance.State.Round))
+			fields.Round(r.GetState().RunningInstance.State.Round),
+			fields.Blinded(decidedBlindedBlock))
```
y0sher, line 244: doesn't `decidedBlindedBlock` says whether it's decided or not? `fields.Blinded` is about blinded or not.

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -9,10 +9,20 @@ import (
...
+	"github.com/bloxapp/ssv/logging/fields"
+	"github.com/bloxapp/ssv/protocol/v2/qbft/controller"
+	"github.com/bloxapp/ssv/protocol/v2/types"
+)
+
+// TODO: This is a reasonable default, but we should probably make this configurable.
+//
+//	Discussion here: https://github.com/ethereum/builder-specs/issues/17
+const (
+	gasLimit = 30_000_000
```
y0sher, line 25: this shows up in a few different places around the code, same as the batch size. even if we're not adding it as a configuration right now, I'd put the consts in one file and read it from there to prevent confusion when one is changed.

beacon/goclient/proposer.go
y0sher: Agree, @Tomblox1 , lets make sure we know we need to revisit metrics especially for distinguishing blinded blocks.

beacon/goclient/proposer.go
y0sher: if you didn't find any cases then I'm not sure why we have this check?

nkryuchkov: I'm not completely sure if it's never `nil` and if it is, this will panic. Also, it would trigger the linter (dereferencing a pointer without checking if it's `nil` is indeed bad).

But I agree we better log this, despite the code would look bulkier.

protocol/v2/ssv/runner/proposer.go
nkryuchkov: It says if the decided value has a blinded or regular block (also it should be called only after the instance has decided). The name is copied from the spec, so I don't want to change it only here

beacon/goclient/proposer.go
nkryuchkov: IO1-3274

beacon/goclient/signing.go
nkryuchkov: It's created in spec and used for signing in `ValidatorRegistrationRunner`. It's from https://github.com/bloxapp/ssv/pull/856 (this PR is based on it). It should be actually `spectypes`, not `types`

operator/duties/controller.go
nkryuchkov: > does validatorRegistration count as a duty in eth? otherwise why is this code here?
It's not a direct duty, but kind of an indirect one: Validators need to do that regularly to use MEV. Operators need to reach a consensus of partial signatures to build a full signature, so its designed like a duty and fits here well.

> also this is added as part of MEV, should we do validatorregistration only when MEV is enabled?
I think yes

ekm/signer_key_manager_test.go
nkryuchkov: The behavior would be same. But I'll better pass `true` to test that it works well in this case. The tests here make sure that signing once works and signing twice is slashable

protocol/v2/ssv/runner/proposer.go
nkryuchkov: Although I changed the variable name to be more clear

docs/MEV_EDGE_CASES.md
nkryuchkov: Please let me know if the added information is sufficient

y0sher (approved):

moshe-blox (approved):

storage/kv/badger.go
```diff
@@ -49,6 +49,8 @@ func New(logger *zap.Logger, options basedb.Options) (basedb.IDb, error) {
 	opt.Logger = newLogger(zap.NewNop())
 	if logger != nil && options.Reporting {
 		opt.Logger = newLogger(logger)
+	} else {
+		opt.Logger = newLogger(zap.NewNop()) // TODO: we should allow only errors to be logged
```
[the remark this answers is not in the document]
nkryuchkov: I'm not sure, it was done in the Niv's commits, I didn't check this specific line. Do you want to delete this?

PR #920, opened 2023-04-02
propose log

egorka-sereda:
add log when send a propose message


protocol/v2/qbft/instance/instance.go
```diff
@@ -70,6 +70,9 @@ func (i *Instance) Start(logger *zap.Logger, value []byte, height specqbft.Heigh
 				logger.Warn("❗ failed to create proposal", zap.Error(err))
 				// TODO align spec to add else to avoid broadcast errored proposal
 			} else {
+				logger.Debug("📢 broadcasting proposal message",
+					zap.Uint64("round", uint64(i.State.Round)),
+					zap.Any("proposal-signers", proposal.Signers))
```
y0sher, line 75: please change any to uint since there is only one singer at this case.

y0sher:
I think eventually we'll include this as part of a bigger PR with logs. so you can close this pr for now.

y0sher:
closing for now. 
Improving our logs is a bigger task.

PR #925, opened 2023-04-16
Message Validation on consensus messages

GalRogozinski:


GalRogozinski:
Replaced by https://github.com/bloxapp/ssv-spec/pull/188

PR #926, opened 2023-04-16
Write sturctured logs to a file

y0sher:
This PR adds functionality that lets you write structured logs ( json formatted logs `{ "L": level, "M": msg, "T": time, "field": field}`. 
This way indexing services like Elastic can easily index these lines and give us nice features in Kibana or KQL to search the logs.

## note about log file rotation
currently I added in-code log file rotation using `lumberjack` lib, the best practice is that the env will take care of it and our app won't be concerned with compressing and moving files. this is temporary so we won't explode our cloud nodes with endless logs. 

we'll need to do some research and consult @zevzek and come with a better rotation solution.

## reminder 
- [ ] - change deploy code back before merging


lior-blox:
@y0sher can you please fix lint?

y0sher:
created a better PR

PR #936, opened 2023-04-24
feat: add support for reading operator key from a file

SpontaneousOverthrow:
The main idea here is to be able to read `operatorPrivateKey` from a file, which is the standard way to access secrets in k8s.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/bloxapp/ssv/logging"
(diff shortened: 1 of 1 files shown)


cli/operator/node.go
```diff
@@ -254,7 +256,15 @@ func setupDb(logger *zap.Logger, eth2Network beaconprotocol.Network) (basedb.IDb
 
 func setupOperatorStorage(logger *zap.Logger, db basedb.IDb) (operatorstorage.Storage, *registrystorage.OperatorData) {
 	nodeStorage := operatorstorage.NewNodeStorage(db)
-	operatorPubKey, err := nodeStorage.SetupPrivateKey(logger, cfg.OperatorPrivateKey, cfg.GenerateOperatorPrivateKey)
+	operatorPrivateKey := cfg.OperatorPrivateKey
+	if operatorPrivateKey == "" && !cfg.GenerateOperatorPrivateKey {
```
y0sher, line 260: Hi @SpontaneousOverthrow !
Thanks for the PR and contributing to SSV! 

Now to the review, don't you think it will be a better way to check if `OperatorPrivateKeyPath` is set?
Generally we need to decide on an order, let's say the config entry `OperatorPrivateKey` is highest priority, second is the file, and third and last is to generate a new key.

so to make things clearer - 
```
1. Check whether `cfg.OperatorPrivateKey` if it is use the key.
2. Check whether `cfg.OperatorPrivateKeyPath` is set, if it is use the key.
3. If both of the above are empty then we have no choice but to generate a new key. 
```

last point (doesn't have to be part of this PR) is maybe it'll be a good practice to save the key to a file even if we generated it, so it can be re-used in the future? or when for example migrating to a new server..

@SpontaneousOverthrow , @lior-blox your thoughts here will be appreciated.

y0sher (changes requested):

SpontaneousOverthrow:
I agree with all points above, so I made changes. If you ok with this logic, I will update tests also.

y0sher:
@SpontaneousOverthrow Thanks for making the changes. sorry for taking long to replay.
I realized thinking with @lior-blox that you can set up the operator private key inside a share.yaml file this is how we do it in our k8s. 
though this is not the same as done here where the operator private key has it's own dedicated file. if you can explain shortly this specific use case, we can go ahead and merge.

SpontaneousOverthrow:
The main point here is to be able to store the configuration file in GitHub (for instance), so that every engineer has the ability to read and update it. On the other hand, the secret key file would be stored in Google Secret Manager (for instance) with limited access scope.

y0sher:
Hi @SpontaneousOverthrow , thank you for contributing both the code and the idea, we have implemented this with a keystore file, meaning you have to pass in file with encrypted operator key and a password file. this is similar to how keys are loaded in ethereum. please checkout this PR - https://github.com/bloxapp/ssv/pull/1019 - soon to be merged.
closing this for now!

PR #938, opened 2023-05-01
Peer Scoring Refactors

moshe-blox:
This PR changes how we score peers by subnets within the recurring task which disconnects the lowest scored peer every minute.

Previously, we seemed to have not favored peers who share many subnets with us, which caused us to disconnect from them and desaturate our subnets.

With this PR, we also strongly boost the scores of subnets for which we have too few connections, and attempt to not disconnect from the peers with whom we share them.


moshe-blox:
I think the scoring algorithm would be much simpler if we discard the limit of maximum peers per subnet.

```python
def score_subnet(conns, min_conns):
  # Exponential scoring when below minimum.
  if conns < min_conns:
    return 1 + ((min_conns - conns) / min_conns)^2
  
  # Inverse scoring when above minimum.
  return 1 / ((conns - min_conns) / min_conns)
```

network/peers/conn_manager.go
```diff
@@ -64,48 +72,122 @@ func (c connManager) TrimPeers(ctx context.Context, logger *zap.Logger, net libp
...
 	return GetTopScores(peerScores, n)
 }
+
+type peerLog struct {
+	Peer          peer.ID
+	Score         PeerScore
+	SharedSubnets int
+}
+
+func (c connManager) logPeerScores(peerLogs []peerLog, mySubnets records.Subnets, subnetConnections []int) {
```
y0sher, line 132: this code is a bit complicated and thats not a good thing. Also not something we going to change right now but this peer-balancing according to subnets is an attack vector, because we report about ourselves the subnets we are in. And you can send whatever subnets you want, this affects scoring and can help an attacker eclipse our node.
it's pretty easy actually as in the handshake you send me the subnets you are in, I can understand you want these subnets so I send you back that I'm in all of them. and when balancing you will kick out peers that are not present in all of them like me (of course if I keep connecting with more peers that share all the subnets)
it's worth to research about how libp2p recommends to do peer scoring as subnets are just topics eventually.

y0sher (approved):
Approving, though I think we need to revisit the whole concept of how and why we use Subnets and how we balance peers inside them. this is a delicate matter that we can see just like this PR can result in missing duties when handled wrongly.

We need to eventually understand how much value we're getting from this, is it a premature optimization? we need to be able to test and know how much it helps.

y0sher:
> I think the scoring algorithm would be much simpler if we discard the limit of maximum peers per subnet, for example:
> 
> ```python
> def score_subnet(conns, min_conns):
>   # Exponential scoring when below minimum.
>   if conns < min_conns:
>     return 1 + ((min_conns - conns) / min_conns)^2
>   
>   # Inverse scoring when above minimum.
>   return 1 / ((conns - min_conns) / min_conns)
> ```
> 
> Do we even need a maximum? @y0sher @lior-blox



> I think the scoring algorithm would be much simpler if we discard the limit of maximum peers per subnet, for example:
> 
> ```python
> def score_subnet(conns, min_conns):
>   # Exponential scoring when below minimum.
>   if conns < min_conns:
>     return 1 + ((min_conns - conns) / min_conns)^2
>   
>   # Inverse scoring when above minimum.
>   return 1 / ((conns - min_conns) / min_conns)
> ```
> 
> Do we even need a maximum? @y0sher @lior-blox

we have max peers, and we want them to spread across the subnets so if we have max peers and a subnet with "max" or a lot of peers and other subnet with a low number of peers we need to strive to balance it and replace some from the high number of peers subnets with peers that are in the lower number peers subnet.

lior-blox (approved):

network/peers/conn_manager.go
moshe-blox: > this code is a bit complicated and thats not a good thing. Also not something we going to change right now but this peer-balancing according to subnets is an attack vector, because we report about ourselves the subnets we are in. And you can send whatever subnets you want, this affects scoring and can help an attacker eclipse our node. it's pretty easy actually as in the handshake you send me the subnets you are in, I can understand you want these subnets so I send you back that I'm in all of them. and when balancing you will kick out peers that are not present in all of them like me (of course if I keep connecting with more peers that share all the subnets) it's worth to research about how libp2p recommends to do peer scoring as subnets are just topics eventually.

Good call, let's create a GitHub issue to rethink subnet negotiation?

PR #946, opened 2023-05-10
Logging/verbose qbft messages

y0sher:
Added and changed a few log lines in qbft instance

### DIFF
--- a/logging/fields/fields.go
+++ b/logging/fields/fields.go
@@ -1,6 +1,7 @@
 package fields
 
 import (
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"net"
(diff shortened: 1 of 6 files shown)


y0sher:
> Looks great! But we missing one: got change round quorum
@andrew-blox 
so we have two cases
1. we `got justified round change` - https://github.com/bloxapp/ssv/pull/946/files#diff-12db8357df0766f566c37820d564228fec5c8730880178e829c1fc6a787df674R66
2. we got partial quorum so `broadcasting round change message` - https://github.com/bloxapp/ssv/pull/946/files#diff-12db8357df0766f566c37820d564228fec5c8730880178e829c1fc6a787df674R102
I can't find any other places this log should be added because we are not checking for that in the code.

andrew-blox:
I will explain why i need it.
We want to know consensus times, prepare times, etc...
I have log of starting instance which is great for round 1 for me to find consensus times and what not

y0sher:
First log indeed means that we start a new instance. But probably there is another scenario, I'll get back to you about it soon.

lior-blox (approved):

moshe-blox (approved):

PR #947, opened 2023-05-10
ship logs into a file and stdout at the same time.

y0sher:
rotate log files using `lumberjack`. send all debug logs into a file in json format.

### Old pr description
This PR adds functionality that lets you write structured logs ( json formatted logs { "L": level, "M": msg, "T": time, "field": field}.
This way indexing services like Elastic can easily index these lines and give us nice features in Kibana or KQL to search the logs.
note about log file rotation

currently I added in-code log file rotation using lumberjack lib, the best practice is that the env will take care of it and our app won't be concerned with compressing and moving files. this is temporary so we won't explode our cloud nodes with endless logs.

we'll need to do some research and consult @zevzek and come with a better rotation solution.


cli/config/config.go
```diff
@@ -16,6 +16,7 @@ type GlobalConfig struct {
 	LogLevel       string `yaml:"LogLevel" env:"LOG_LEVEL" env-default:"info" env-description:"Defines logger's log level'"`
 	LogFormat      string `yaml:"LogFormat" env:"LOG_FORMAT" env-default:"console" env-description:"Defines logger's encoding, valid values are 'json' (default) and 'console''"`
 	LogLevelFormat string `yaml:"LogLevelFormat" env:"LOG_LEVEL_FORMAT" env-default:"capitalColor" env-description:"Defines logger's level format, valid values are 'capitalColor' (default), 'capital' or 'lowercase''"`
+	LogFilePath    string `yaml:"LogFilePath" env:"LOG_FILE_PATH" env-default:"./debug.log" env-description:"Defines a file path to write logs into"`
```
lior-blox, line 19: add data folder to the path

moshe-blox (approved):

PR #948, opened 2023-05-12
Avoid executing duties if eth1 node is syncing

nkryuchkov:


nkryuchkov:
The PR was tested, worked as expected

cli/operator/node.go
```diff
@@ -122,6 +122,7 @@ var StartNodeCmd = &cobra.Command{
 		cfg.SSVOptions.Context = ctx
 		cfg.SSVOptions.DB = db
 		cfg.SSVOptions.BeaconNode = el
+		cfg.SSVOptions.Eth1Client = cl
```
y0sher, line 125: I know this was in the code beforehand, but this is funny, or some one was confused. 
EL - execution layer - ETH1 (geth)
CL - consensus layer - ETH2 (beacon node)

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -240,6 +243,10 @@ func (r *AggregatorRunner) expectedPostConsensusRootsAndDomain() ([]ssz.HashRoot
 // 4) Once consensus decides, sign partial aggregation data and broadcast
 // 5) collect 2f+1 partial sigs, reconstruct and broadcast valid SignedAggregateSubmitRequest sig to the BN
 func (r *AggregatorRunner) executeDuty(logger *zap.Logger, duty *spectypes.Duty) error {
+	if !r.eth1.IsReady() {
+		panic("eth1 node isn't ready, there's a risk of getting slashed")
```
y0sher, line 247: let's use logger.fatal

moshe-blox:
Eth1 nodes operate under the guidance of a connected Eth2 node. Eth2 dictates to the Eth1 which blocks have passed the fork choice rule, and we might guess that it means also which blocks to *sync*.

I suspect that it's possible that if the Beacon node isn't synced or down, Eth1 might not report that it's "syncing".

I think we should research/test this. @lior-blox

nkryuchkov:
@moshe-blox Should we then check both?

moshe-blox:
By the way, I think that the `optimistic` field of the Eth2 /syncing route already tells whether the Eth1 node is 'sufficiently' synced.

Although, I'm not sure if this matches our requirements for Eth1's recency. For example, the Eth2 might be fine if the Eth1 is up to 64 blocks back, but we aren't.

nkryuchkov:
The spec says that [optimistic node must not execute a duty](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#validator-assignments), so I think we should check if both eth1 and eth2 are not syncing. 

I'll also add a check that eth2 is ready to this PR according to the following logic: 
```go
func (v *Validator) assertEthNodes(logger *zap.Logger) {
	if !v.Eth1.IsReady() || !v.Beacon.IsReady() {
		logger.Panic("eth nodes must be ready, otherwise there's a risk of getting slashed",
			zap.Bool("eth1_ready", v.Eth1.IsReady()),
			zap.Bool("eth2_ready", v.Beacon.IsReady()),
		)
	}
}

// IsReady returns if beacon node is currently ready: responds to requests, not in the syncing state, not optimistic
// (for optimistic see https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#block-production).
func (gc *goClient) IsReady() bool {
	ctx, cancel := context.WithTimeout(gc.ctx, healthCheckTimeout)
	defer cancel()

	syncState, err := gc.client.NodeSyncing(ctx)
	return err == nil && syncState != nil && !syncState.IsSyncing && !syncState.IsOptimistic
}

// IsReady returns if eth1 is currently ready: responds to requests and not in the syncing state.
func (ec *eth1Client) IsReady() bool {
	ctx, cancel := context.WithTimeout(ec.ctx, healthCheckTimeout)
	defer cancel()

	sp, err := ec.conn.SyncProgress(ctx)
	return err == nil && sp == nil
}
```

Beacon node API reference: https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus

@moshe-blox  @y0sher @lior-blox Please let me know if you disagree or want to add something

beacon/goclient/goclient.go
```diff
@@ -158,6 +158,17 @@ func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.Op
...
 }
 
+// IsReady returns if beacon node is currently ready: responds to requests, not in the syncing state, not optimistic
+// (for optimistic see https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#block-production).
+func (gc *goClient) IsReady() bool {
+	ctx, cancel := context.WithTimeout(gc.ctx, healthCheckTimeout)
+	defer cancel()
+
+	syncState, err := gc.client.NodeSyncing(ctx)
+	// TODO: also check if !syncState.ElOffline when github.com/attestantio/go-eth2-client supports it
```
y0sher, line 168: @Tomblox1

protocol/v2/ssv/validator/validator.go
```diff
@@ -62,6 +62,11 @@ func NewValidator(pctx context.Context, cancel func(), options Options) *Validat
 		dutyIDs:     hashmap.New[spectypes.BeaconRole, string](),
 	}
 
+	// Changing specssv.BeaconNode requires modifications in spec.
+	if beaconStatusChecker, ok := options.Beacon.(NodeStatusChecker); ok {
```
y0sher, line 66: if this returns false v.Beacon stays nil, we should somehow mitigate otherwise if there's no option for this to happen this check is redundant.

moshe-blox:
> The spec says that [optimistic node must not execute a duty](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#validator-assignments), so I think we should check if both eth1 and eth2 are not syncing.
> 
> I'll also add a check that eth2 is ready to this PR according to the following logic:
> 
> ```go
> func (v *Validator) assertEthNodes(logger *zap.Logger) {
> 	if !v.Eth1.IsReady() || !v.Beacon.IsReady() {
> 		logger.Panic("eth nodes must be ready, otherwise there's a risk of getting slashed",
> 			zap.Bool("eth1_ready", v.Eth1.IsReady()),
> 			zap.Bool("eth2_ready", v.Beacon.IsReady()),
> 		)
> 	}
> }
> 
> // IsReady returns if beacon node is currently ready: responds to requests, not in the syncing state, not optimistic
> // (for optimistic see https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#block-production).
> func (gc *goClient) IsReady() bool {
> 	ctx, cancel := context.WithTimeout(gc.ctx, healthCheckTimeout)
> 	defer cancel()
> 
> 	syncState, err := gc.client.NodeSyncing(ctx)
> 	return err == nil && syncState != nil && !syncState.IsSyncing && !syncState.IsOptimistic
> }
> 
> // IsReady returns if eth1 is currently ready: responds to requests and not in the syncing state.
> func (ec *eth1Client) IsReady() bool {
> 	ctx, cancel := context.WithTimeout(ec.ctx, healthCheckTimeout)
> 	defer cancel()
> 
> 	sp, err := ec.conn.SyncProgress(ctx)
> 	return err == nil && sp == nil
> }
> ```
> 
> Beacon node API reference: https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus
> 
> @moshe-blox @y0sher @lior-blox Please let me know if you disagree or want to add something

Hi @nkryuchkov, thanks, this looks good.

We discussed this whole thing again, and we think to make another upgrade:

1. Since we only care about the readiness of both nodes and not one in particular, we should probably just expose a single function to check readiness
2. Checking readiness in `StartDuty` is a bit problematic, because there can be many parallel calls to `StartDuty` and this would overload the nodes. Instead, we should probably check this periodically in the background in a dedicated goroutine.
3. After panicing, the node would restart immediately (in most setups), at which point the Eth nodes are probably still "syncing" and therefore we'd remain in a crash loop until they aren't. To prevent that, we should wait for the nodes to be ready before starting SSV. This should happen at a very early stage, at the very least before we start the validators, but if we have access to the Eth1/Eth2 nodes even earlier, then why not even before?

I've made some skeleton interface/implementation which should support both checking for readiness (`Ready`) and also waiting (`Wait`) for the both nodes to be ready in a single call:
```go
// Existing eth1Client
func (e *eth1Client) NodeSyncing() (bool, error) {
	// ...
}

// Existing beaconClient
func (e *beaconClient) NodeSyncing() (bool, error) {
	// ...
}

type NodeSyncing interface {
	NodeSyncing() (bool, error)
}

type NodeReadinessProber struct {
	interval time.Duration
	nodes    []NodeSyncing

	ready bool
	cond  sync.Cond
	mu    sync.RWMutex
}

func NewNodeReadinessProber(interval time.Duration, nodes ...NodeSyncing) *NodeReadinessProber {
	return &NodeReadinessProber{
		interval: interval,
		nodes:    nodes,
	}
}

func (n *NodeReadinessProber) Run(ctx context.Context) error {
	ticker := time.NewTicker(n.interval)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-ticker.C:
			if err := n.probe(); err != nil {
				// Log and continue.
			}
		}
	}
}

var errSyncing = errors.New("node is syncing")

func (n *NodeReadinessProber) probe(ctx context.Context) error {
	// Query all nodes in parallel.
	ctx, cancel := context.WithTimeout(ctx, n.interval)
	defer cancel()

	ready := true
	var wg sync.WaitGroup
	for _, node := range n.nodes {
		wg.Add(1)
		go func(node NodeSyncing) {
			defer wg.Done()
			if syncing, err := node.NodeSyncing(); err != nil {
				// Log and continue.
			} else if syncing {
				// Quit early.
				ready = false
				cancel()
			}
		}(node)
	}
	wg.Wait()

	n.mu.Lock()
	defer n.mu.Unlock()
	n.ready = ready
	if n.ready {
		n.cond.Broadcast()
	}
	return nil
}

func (n *NodeReadinessProber) Ready() bool {
	n.mu.RLock()
	defer n.mu.Unlock()

	return n.ready
}

func (n *NodeReadinessProber) Wait() {
	n.mu.RLock()
	defer n.mu.RUnlock()

	for !n.ready {
		n.cond.Wait()
	}
}
```

You can turn `NodeReadinessProber` (or w/e you wanna call it) into an interface if you'd like, to make it testable.

Please let me know your thoughts.

nkryuchkov:
@moshe-blox Thanks for the proposal. Implemented with minor changes. Please review

protocol/v2/ssv/validator/validator.go
```diff
@@ -86,6 +86,10 @@ func NewValidator(pctx context.Context, cancel func(), options Options) *Validat
 
 // StartDuty starts a duty for the validator
 func (v *Validator) StartDuty(logger *zap.Logger, duty *spectypes.Duty) error {
+	if ready, err := v.NodeChecker.IsReady(v.ctx); err != nil || !ready {
+		logger.Panic("nodes must be ready, otherwise there's a risk of getting slashed")
+	}
+
```
moshe-blox, line 92: Panic & message are correct, and [the issue i mentioned here](https://github.com/bloxapp/ssv/pull/948#issuecomment-1586248162) about `StartDuty` is no longer relevant, however I forgot to mention another issue — waiting for `StartDuty` to panic might cause us to never panic at all (for example, if the nodes crash/stuck and stop providing duties).

For that reason, we should find some other place to make this check & panic.

There are a few ways to do this:

1. After starting the prober, start another goroutine to do this check every X seconds
2. Change `ready bool` to `ready chan bool` and `IsReady()` to `IsReady() (<-chan bool)`, and send update to `ready` at every probe. With this, we don't need the `Wait` at all, we can just listen to `IsReady()` until its `true`, and then start the node and start a goroutine to listen to it until its `false` to panic
3. Accept "callback" in `NewProber` and we'll pass one that panics if ready is false

Would love to hear what you think is the safest/simplest.

moshe-blox (changes requested):

protocol/v2/ssv/validator/validator.go
y0sher: the above lines in startDuty might not be needed because we added the periodic check. 
@moshe-blox do you think there is anything else that should be done?

y0sher: @nkryuchkov ,
I feel like we can keep the wait. but the bottom line is that we can't panic only at `StartDuty` because probably if we got a duty, we are synced, and if we are not synced we're not getting any duties.
So its not the optimal place to do this check, it might be good thing IMO to go with moshe's option 3, pass a callback into the prober constructor that will determine what do to when the nodes are not ready. we need somehow to make it active only after the first wait though.

nkryuchkov: @y0sher I see, thanks. Implemented. Could you please review again?

y0sher (approved):
lgtm, just like we discussed

cli/operator/node.go
```diff
@@ -156,8 +167,13 @@ var StartNodeCmd = &cobra.Command{
 			go startMetricsHandler(cmd.Context(), logger, db, cfg.MetricsAPIPort, cfg.EnableProfile)
 		}
 
-		metrics.WaitUntilHealthy(logger, cfg.SSVOptions.Eth1Client, "execution client")
-		metrics.WaitUntilHealthy(logger, cfg.SSVOptions.BeaconNode, "consensus client")
+		nodeChecker.Wait()
+
+		nodeUnreadyHandler := func() {
+			logger.Panic("nodes must be ready, otherwise there's a risk of getting slashed")
```
moshe-blox, line 173:
```suggestion
			logger.Fatal("Ethereum nodes are out of sync. Usage of unsynchronized nodes can lead to undefined behavior, such as wrongly marking removed validators as active. Ensure the nodes are synced to resume.")
```

moshe-blox (approved):

cli/operator/node.go
```diff
@@ -156,8 +167,13 @@ var StartNodeCmd = &cobra.Command{
 			go startMetricsHandler(cmd.Context(), logger, db, cfg.MetricsAPIPort, cfg.EnableProfile)
 		}
 
-		metrics.WaitUntilHealthy(logger, cfg.SSVOptions.Eth1Client, "execution client")
-		metrics.WaitUntilHealthy(logger, cfg.SSVOptions.BeaconNode, "consensus client")
+		nodeChecker.Wait()
+
+		nodeUnreadyHandler := func() {
+			logger.Fatal("Ethereum nodes are out of sync. Usage of unsynchronized nodes can lead to undefined behavior, such as wrongly marking removed validators as active. Ensure the nodes are synced to resume.")
```
lior-blox, line 173: @nkryuchkov I think this can be a comment.
In the log we should state which node is out of sync (execution layer or consensus layer)
so it will be more informative.
so something like the execution node is out of sync -

moshe-blox, line 173:
```suggestion
			logger.Fatal("Ethereum nodes are out of sync. Ensure the nodes are synced to resume.")
```

y0sher (approved):

moshe-blox (approved):

moshe-blox (approved):

PR #953, opened 2023-05-17
Network configs

nkryuchkov:
Adjacent spec PR: https://github.com/bloxapp/ssv-spec/pull/241


cli/operator/node.go
```diff
@@ -83,7 +83,15 @@ var StartNodeCmd = &cobra.Command{
 			log.Fatal("could not create logger", err)
 		}
 
-		eth2Network, forkVersion := setupSSVNetwork(logger)
+		beaconNetworks := []spectypes.BeaconNetworkDescriptor{networkconfig.PraterNetwork, networkconfig.MainNetwork}
+		for _, bn := range beaconNetworks {
+			spectypes.RegisterBeaconNetwork(bn)
```
y0sher, line 88: how can we register 2 networks?

beacon/goclient/aggregator.go
```diff
@@ -92,7 +92,7 @@ func isAggregator(committeeCount uint64, slotSig []byte) (bool, error) {
 
 // waitOneThirdOrValidBlock waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot)
 func (gc *goClient) waitToSlotTwoThirds(slot phase0.Slot) {
-	oneThird := gc.network.SlotDurationSec() / 3 /* one third of slot duration */
+	oneThird := gc.network.SlotDuration() / 3 /* one third of slot duration */
```
y0sher, line 95: why this change?

beacon/goclient/types.go
```diff
@@ -9,5 +13,5 @@ var (
 	EpochsPerSyncCommitteePeriod         uint64       = 256
 	TargetAggregatorsPerCommittee        uint64       = 16
 	FarFutureEpoch                       phase0.Epoch = 1<<64 - 1
-	GenesisForkVersion                                = phase0.Version{0x0, 0x0, 0x10, 0x20} // prater
+	GenesisForkVersion                                = networkconfig.PraterNetwork.ForkVersion // TODO: pass from caller
```
y0sher, line 16: So this could ideally look like this.
```go
GenesisForkVersion = networkconfig.ForkVersion
```
and when initializing we'll do something like 
```go
networkconfig.SetNetwork(PraterNetwork)
```
where `PraterNetwork` is a struct holding all values incl `ForkVersion`

beacon/goclient/aggregator.go
nkryuchkov: It returns a `time.Duration`, not seconds. But this still may be changed depending on what changes I'll do in the future

cli/operator/node.go
```diff
@@ -288,29 +290,34 @@ func setupOperatorStorage(logger *zap.Logger, db basedb.IDb) (operatorstorage.St
...
-		cfg.P2pNetworkConfig.NetworkID = format.DomainType(types.GetDefaultDomain()).String()
-	} else {
-		// we have some custom network id, overriding default domain
-		domainType, err := format.DomainTypeFromString(cfg.P2pNetworkConfig.NetworkID)
-		if err != nil {
-			logger.Fatal("failed to parse network id", zap.Error(err))
-		}
-		types.SetDefaultDomain(spectypes.DomainType(domainType))
+func setupSSVNetwork(logger *zap.Logger) (beaconprotocol.Network, forksprotocol.ForkVersion, error) {
+	beaconNetwork, err := networkconfig.GetNetworkByName(cfg.P2pNetworkConfig.NetworkID)
```
y0sher, line 294: I'm not sure it's good that we'll use `cfg.P2pNetworkConfig.NetworkID` as the config param for network name.
Network name is the name of the network for the configuration.
NetworkID was originally used in P2P to make sure nodes from different networks won't try to connect each other.
So for sure we want the Network name to be on higher level in the config than P2P since now it represents more than what it used to. 
The network configuration should include a NetworkID that is a byte that is used for the same purpose as the previous NetworkID.

nkryuchkov: > So for sure we want the Network name to be on higher level in the config than P2P since now it represents more than what it used to.

okay, I agree

> The network configuration should include a NetworkID that is a byte that is used for the same purpose as the previous NetworkID.

Why do we need to keep it in the config file if it's configured in the network config?

nkryuchkov: done

.k8/yamls-v1/ssv-node-v2-1-deployment.yml
```diff
@@ -119,16 +119,12 @@ spec:
           value: "false"
         - name: DISCOVERY_TYPE_KEY
           value: "discv5"
-        - name: NETWORK
```
y0sher, line 122: I don't understand why in some we add network = jato and in some we delete network entirely ?

go.mod
```diff
@@ -184,6 +184,6 @@ replace github.com/google/flatbuffers => github.com/google/flatbuffers v1.11.0
 
 replace github.com/dgraph-io/ristretto => github.com/dgraph-io/ristretto v0.1.1-0.20211108053508-297c39e6640f
 
-replace github.com/bloxapp/ssv-spec => github.com/bloxapp/ssv-spec v0.0.0-20230521155754-79016edc0974 // #mev-capella
+replace github.com/bloxapp/ssv-spec => github.com/nkryuchkov/ssv-spec v0.0.0-20230523123751-7a59c345f536 // network-configs
```
y0sher, line 187: if we keep it diverged from main spec branch its better to push to spec's remote origin. though lets wait with it before we merge.

networkconfig/config.go
```diff
@@ -0,0 +1,20 @@
+package networkconfig
+
+import (
+	"fmt"
+
+	spectypes "github.com/bloxapp/ssv-spec/types"
+)
+
```
y0sher, line 8: @nkryuchkov 
this isn't important as part of this PR, but we'll eventually want to be able to add new network config on the fly. So a function that adds a config to `SupportedConfigs` (will probably require mutex).
then maybe load it from a json file for example. this will be useful for local testnet.
@Tomblox1 please create ticket.

protocol/v2/types/domain.go
```diff
@@ -2,10 +2,13 @@ package types
 
 import (
 	spectypes "github.com/bloxapp/ssv-spec/types"
+
+	"github.com/bloxapp/ssv/networkconfig"
 )
 
+// TODO: get rid of singleton, pass domain as a parameter
```
y0sher, line 9: @nkryuchkov @Tomblox1 please create a ticket

y0sher (approved):
Approved, please address my comments.

networkconfig/config.go
Tomblox1: https://bloxxx.atlassian.net/browse/IO1-3440
@nkryuchkov @y0sher

protocol/v2/types/domain.go
nkryuchkov: It's left like that because this change would require a lot of modifications, which would increase the size of the PR, increase the code review complexity, and delay merging of the PR. Also it looks like something that could be extracted and done separately

network/p2p/p2p_setup.go
```diff
@@ -189,7 +192,7 @@ func (n *p2pNetwork) setupPeerServices(logger *zap.Logger) error {
 	}
 
 	filters := []connections.HandshakeFilter{
-		connections.NetworkIDFilter(n.cfg.NetworkID),
+		connections.NetworkIDFilter(domain),
```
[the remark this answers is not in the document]
nkryuchkov: yes, `n.cfg` is `cfg.P2pNetworkConfig`, here's the code that used to initialize domain according to `cfg.P2pNetworkConfig.NetworkID`:
```go
func setupSSVNetwork(logger *zap.Logger) (beaconprotocol.Network, forksprotocol.ForkVersion) {
	if len(cfg.P2pNetworkConfig.NetworkID) == 0 {
		cfg.P2pNetworkConfig.NetworkID = format.DomainType(types.GetDefaultDomain()).String()
	} else {
		// we have some custom network id, overriding default domain
		domainType, err := format.DomainTypeFromString(cfg.P2pNetworkConfig.NetworkID)
		if err != nil {
			logger.Fatal("failed to parse network id", zap.Error(err))
		}
		types.SetDefaultDomain(spectypes.DomainType(domainType))
	}
```

.k8/yamls-v1/ssv-node-v2-1-deployment.yml
nkryuchkov: We used to have `NETWORK` (beacon network name, e.g. `prater`) and `NETWORK_ID` (domain, e.g. `0x00003011`). But `NETWORK_ID` was set only on stage nodes, it was not set on prod nodes, which used a default value. 

Now we don't need to configure both as they are stored in the network descriptor. So we remove them. But instead, we need to set a network name. We set it only on stage nodes because the prod nodes use the default value (`mainnet`), this logic is similar to what we had with `NETWORK_ID`. For setting a network name we reuse the `NETWORK` field.

moshe-blox (approved):

y0sher (approved):
good job

lior-blox (approved):

PR #954, opened 2023-05-17
Bootnode Configuration

moshe-blox:
- Configurable ports
- Persistent `enode` database
- Deployment of a 2nd bootnode


Makefile
```diff
@@ -13,13 +13,15 @@ ifndef $(BUILD_PATH)
...
-# node command builder
+# Node command.
 NODE_COMMAND=--config=${CONFIG_PATH}
-
 ifneq ($(SHARE_CONFIG),)
   NODE_COMMAND+= --share-config=${SHARE_CONFIG}
 endif
 
+# Bootnode command.
+BOOTNODE_COMMAND=--config=${CONFIG_PATH}
```
y0sher, line 23: NODE_COMMAND is the same so this const is kinda funny

y0sher (approved):
I was under the impression boot node is just a regular node that doesn't try to do bootstrap or discovery by itself but instead waits for incoming Peer discovery requests and saves all of them then return to whoever requests. 

Hence all the added configuration should be a part of a regular node as well no?

lior-blox (approved):

PR #955, opened 2023-05-18
fix: messages.go url from ssv-spec repo

SpontaneousOverthrow:
current url is broken (404 response code)


y0sher (approved):
thanks for that.
@moshe-blox please review

moshe-blox (approved):

PR #960, opened 2023-05-23
Conditional Popping

moshe-blox:
### Problem
Sometimes messages are being processed prematurely which leads to them being rejected, for example:
1. Prepares/commits for a proposal that wasn't yet received
2. Proposal for a duty that wasn't yet started

This PR refactors `priorityQueue` to support popping messages with a filter, and uses this new ability to avoid popping certain messages too early, specifically:
1. **If no duty is running, pop only ExecuteDuty messages**
  *This keeps pre-consensus/consensus messages in the queue, until we're ready to process them.*
2. **If no proposal was accepted for the current round, skip prepare & commit messages for the current height and round**
  *This helps handle network-layer asynchrony. While prepares & commits are sent after proposals, they're sometimes received before.*

### DIFF
--- a/protocol/v2/ssv/queue/queue.go
+++ b/protocol/v2/ssv/queue/queue.go
@@ -101,17 +101,20 @@ func (q *priorityQueue) Pop(ctx context.Context, prioritizer MessagePrioritizer,
 
 	// Wait for a message to be pushed.
 Wait:
-	select {
-	case msg := <-q.inbox:
-		if q.head == nil {
-			q.head = &item{message: msg}
-		} else {
-			q.head = &item{message: msg, next: q.head}
+	for {
+		select {
+		case msg := <-q.inbox:
+			if q.head == nil {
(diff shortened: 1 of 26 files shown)


protocol/v2/ssv/queue/queue.go
```diff
@@ -85,13 +94,23 @@ func (q *priorityQueue) Pop(ctx context.Context, prioritizer MessagePrioritizer)
...
 	select {
 	case msg := <-q.inbox:
-		q.head = &item{message: msg}
+		if q.head == nil {
+			q.head = &item{message: msg}
+		} else {
+			q.head = &item{message: msg, next: q.head}
+		}
+		if !filter(msg) {
+			goto Wait
```
y0sher, line 112: this usually achieved like this in golang- 
```go
Loop:
for {
  select {
    case msg := <-q.inbox:
      //do stuff
      break Loop
  }
}
```

PR #963, opened 2023-05-23
Align contract v0.3.1-rc.0 

olegshmuelov:


moshe-blox (approved):

operator/validator/event_handler.go
```diff
@@ -123,6 +142,53 @@ func (c *controller) handleValidatorAddedEvent(
...
+	// Calculate the expected length of constructed shares based on the number of operator IDs,
+	// signature length, public key length, and encrypted key length.
+	operatorCount := len(event.OperatorIds)
+	signatureOffset := phase0.SignatureLength
+	pubKeysOffset := phase0.PublicKeyLength*operatorCount + signatureOffset
+	sharesExpectedLength := encryptedKeyLength*operatorCount + pubKeysOffset
+
+	if sharesExpectedLength != len(event.Shares) {
+		return nil, errors.Errorf("%s event shares length is not correct", abiparser.ValidatorAdded)
+	}
```
olegshmuelov, line 179: The error should be follows
```
eturn nil, &MalformedEventError{
	Err: errors.Wrap(err, "failed to verify signature"),
}
```

olegshmuelov: In general we have to ditignuish between Malformed and sytem errors,
and increment the Nonce only for malformed.
This way we can ensure the validity of the Nonce

operator/validator/event_handler.go
```diff
@@ -123,6 +142,53 @@ func (c *controller) handleValidatorAddedEvent(
...
+
+	err = c.eventHandler.SaveEventData(event.TxHash)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to save event data")
+	}
+
+	err = c.eventHandler.BumpNonce(event.Owner)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to bump the nonce")
+	}
```
olegshmuelov, line 168: the saving the txHash and bumping the nonce should happend only after event was handled
By handle I mean:
- 1. the event was handled succuessfully 
- 2. there was malformed event (only!) see https://github.com/bloxapp/ssv/pull/963/files#r1204377224

olegshmuelov: Also the todo comment from last commit was lost
The SaveEventData & BumpNonce shoud be executed as one transaction (atomic).
Not for this PR but lets keep the track of todos

registry/storage/recipients.go
```diff
@@ -97,6 +108,58 @@ func (s *recipientsStorage) GetRecipientDataMany(logger *zap.Logger, owners []co
...
+	} else if found {
+		// Bump the nonce
+		*rData.Nonce++
+	}
+
+	raw, err := json.Marshal(rData)
+	if err != nil {
+		return errors.Wrap(err, "could not marshal recipient data")
+	}
+	return s.db.Set(s.prefix, buildRecipientKey(rData.Owner), raw)
```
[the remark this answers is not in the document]
olegshmuelov: yeah we should definitely use lock here
For both BumpNonce & GetNextNonce
```go
s.lock.RLock()
defer s.lock.RUnlock()
```

PR #965, opened 2023-05-24
[Audit Issue E] Handle event message processing errors

nkryuchkov:
Return errors returned by `OnTimeout` and `OnExecuteDuty` instead of just logging them


y0sher (approved):
LGTM, @moshe-blox

moshe-blox (approved):

moshe-blox:
at first i thought this could cause issues, but then i saw that if this func returns an error we ignore it as well: https://github.com/bloxapp/ssv/blob/90874aefde6f266d476f6c74ff3ed066ed136b58/protocol/v2/ssv/validator/msgqueue_consumer.go#L161-L165

PR #966, opened 2023-05-26
[Audit Suggestion 2] Check eth1 events type assertion

nkryuchkov:
Avoid panic if event type mismatches


y0sher (changes requested):
We could probably reduce some code here if we use the switch type of go this way - 
```golang
switch e.Data.(type) {
 case abiParser.OperatorRemovedEvent:
    // here we know for sure e.Data is OperatorRemovedEvent
 default:
     // return malformed event error unknown event
}
```
@olegshmuelov , @moshe-blox  if you have any objections please state them.

nkryuchkov:
@y0sher But then we wouldn't check if `e.Name` contains the same event. Although we could probably remove it

y0sher:
> @y0sher But then we wouldn't check if `e.Name` contains the same event. Although we could probably remove it

Ok than let's keep it like this for now, maybe change to what I suggested as part of a bigger change.

y0sher (approved):

moshe-blox (approved):

moshe-blox:
> @y0sher But then we wouldn't check if e.Name contains the same event. Although we could probably remove it

I noticed that event types are already determined by the names, so this check is already being done before fireEvent is called, therefore we don't need to check the event names in Eth1EventHandler (because it's already checked before fireEvent is called in [handleEvent](https://github.com/bloxapp/ssv/blob/ed7b79a6780126cc1999ef25da8ee0057328108b/eth1/goeth/goETH.go#L345)).

Can you confirm @nkryuchkov?

I've made a commit to only check the types, does it make sense for you?

nkryuchkov:
@moshe-blox I think then we need to remove the `Name` field because it's used only in `HandleEventResult` for logging

PR #971, opened 2023-05-29
Dependency Upgrades

moshe-blox:
SSV has some considerably outdated dependencies, for example: `libp2p` is at `v0.24.2` while the latest is at `v0.27.3` (for reference — Prysm is at `v0.26.2`)

This PR...
- Upgrades `libp2p` and `libp2p-pubsub` to their latest versions
- Removes unused package `grpcex` (and with it a dependency we don't need)
- Upgrades Go from 1.19 to 1.20
- `TODO: upgrade more packages`


network/peers/connections/mock/mock_id_service.go
```diff
@@ -35,3 +35,8 @@ func (m IDService) Close() error {
 	//TODO implement me
 	panic("implement me")
 }
+
+func (m IDService) Start() {
```
y0sher, line 39: don't this API change mean we should Start() IDService? did we check that it works without?

y0sher (changes requested):

network/p2p/p2p_sync.go
```diff
@@ -259,7 +259,9 @@ func (n *p2pNetwork) makeSyncRequest(logger *zap.Logger, peers []peer.ID, mid sp
 		logger := logger.With(fields.PeerID(pid))
 		raw, err := n.streamCtrl.Request(logger, pid, protocol, encoded)
 		if err != nil {
-			if err != multistream.ErrNotSupported {
+			// TODO: is this how to check for ErrNotSupported?
```
nkryuchkov, line 262: Can we have a test for this to be sure?

network/peers/connections/mock/mock_id_service.go
moshe-blox: > don't this API change mean we should Start() IDService? did we check that it works without?

correct, good catch

network/p2p/p2p_sync.go
moshe-blox: > Can we have a test for this to be sure?

I should definitely verify that this would work, however I could not find the exact lines at which this error is returned from :\

moshe-blox: Because this condition only affects logging, it's not critical, I'll research this after merging. 

https://github.com/bloxapp/ssv/issues/1013

y0sher (approved):
don't forget to revert deployments

PR #973, opened 2023-05-29
Log duration in TestThrottle

nkryuchkov:
Log duration in `TestThrottle`, it might be helpful because this test is flaky locally


y0sher (approved):

y0sher:
Either that or round the result in the test. @moshe-blox lets approve and merge, this test is flaky right now.

lior-blox (approved):

PR #974, opened 2023-05-29
Validator metadata panic

y0sher:
## The bug

We had a panic happen at stage env that was a result of `ValidatorRemoved` event that was fired during 
`UpdateValidatorMetadata` this resulted in the validator not being in the `validatorMap` and yet we tried to activate it using a share from the DB, which wasn't returning an error but did return an empty share (nil) which panicked when we tried to use (log it in this case)

## Solution
on `UpdateValidatorMetadata` a new check on `Share` will make sure we wont start the share if it is empty.

## Notes

We should revamp this area, to make sure validator events are interlocked with routinely updating the validator metadata. This will probably happen as part of a bigger change in the way we receive, handle and store all the information from ETH1 node.

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -523,6 +523,9 @@ func (c *controller) UpdateValidatorMetadata(logger *zap.Logger, pk string, meta
 		if err != nil {
 			return errors.Wrap(err, "could not get share")
 		}
+		if share == nil {
+			return errors.New("share was not found")
+		}
 		started, err := c.onShareStart(logger, share)
 		if err != nil {
 			return errors.Wrap(err, "could not start validator")
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -318,25 +318,12 @@ func (c *controller) handleRouterMessages(logger *zap.Logger) {
 	}
 }
 
-// getShare returns the share of the given validator public key
-// TODO: optimize
-func (c *controller) getShare(pk spectypes.ValidatorPK) (*types.SSVShare, error) {
-	share, found, err := c.sharesStorage.GetShare(pk)
-	if err != nil {
-		return nil, errors.Wrapf(err, "could not read validator share [%s]", pk)
-	}
-	if !found {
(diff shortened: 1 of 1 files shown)


olegshmuelov (commented):
It appears that we have a degree of redundancy between the [getShare method](https://github.com/bloxapp/ssv/blob/515888efa3562b482a620a0c8f654d69f92567df/operator/validator/controller.go#L323) and the [sharesStorage.GetShare](https://github.com/bloxapp/ssv/blob/515888efa3562b482a620a0c8f654d69f92567df/registry/storage/shares.go#L74) method from the registry store. Both seem to perform similar operations.

To streamline our codebase and reduce redundancy, I suggest we consider directly using the `sharesStorage.GetShare` method from the registry store instead of `getShare`. The `sharesStorage.GetShare` method includes a 'found' parameter that can help us identify whether the share exists, providing additional functionality.

As a result, we may deprecate the `getShare` method.

y0sher:
> It appears that we have a degree of redundancy between the [getShare method](https://github.com/bloxapp/ssv/blob/515888efa3562b482a620a0c8f654d69f92567df/operator/validator/controller.go#L323) and the [sharesStorage.GetShare](https://github.com/bloxapp/ssv/blob/515888efa3562b482a620a0c8f654d69f92567df/registry/storage/shares.go#L74) method from the registry store. Both seem to perform similar operations.
> 
> To streamline our codebase and reduce redundancy, I suggest we consider directly using the `sharesStorage.GetShare` method from the registry store instead of `getShare`. The `sharesStorage.GetShare` method includes a 'found' parameter that can help us identify whether the share exists, providing additional functionality.
> 
> As a result, we may deprecate the `getShare` method.

I agree, it seems that is was planned that `getShare` will be an optimized version, rather than querying the db every time. though this didn't happen yet. I can deprecate `getShare` and hopefully we`ll optimize this whole area in the future.

moshe-blox (approved):

olegshmuelov (approved):

PR #975, opened 2023-05-29
Improved Nonce Handling for 'Validator Added' Events"

y0sher:
### Description
This PR enhances the handling of 'Validator Added' events within the SSV node. It introduces several key changes to ensure the integrity of the registration process and the consistency of the nonce incrementing logic:

- Added checks to prevent a validator from registering with the same public key but with a different owner address. The system now verifies that a validator's share exists and that the owner's address matches before processing the 'Validator Added' event.
- If a validator's share doesn't exist, it is created and validated. The nonce is incremented only after successful creation and validation.
- The error handling mechanism has been improved to catch and address issues that may arise during the validation process. In the event of a system error, the function returns immediately without incrementing the nonce.
- The system now properly handles a malformed 'Validator Added' event. In such cases, the nonce is still incremented to maintain the order and consistency of the event queue.
- Added unit tests for recipient&events data stores


operator/validator/event_handler.go
```diff
@@ -157,15 +156,24 @@ func (c *controller) handleValidatorAddedEvent(
...
-	}
+	var valid, malformed bool
 
-	err = c.eventHandler.BumpNonce(event.Owner)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to bump the nonce")
-	}
+	defer func() {
+
+		if valid || malformed {
```
moshe-blox, line 163: Why do we need this condition, shouldn't we always save & bump?

y0sher: no, that's the point of this PR, only save and bump when the event is malformed or valid.
if for example we saved+bumped and then had a crash/db error in the middle, we wont to re-run this event.
if at any point we have a "system error"  we should re-run the whole thing.

Like we talked, this is unlikely to happen and if we have DB/system error we should probably quit/crash. but for example in our stage env there are a lot of redeploys, if we were redeployed during event handling we would just re-run it! saving the saving the event and bumping the nocne before finishing might result in bad data in this case.

@olegshmuelov please review as well

moshe-blox (approved):

olegshmuelov (commented):
1. we can move the `defer` statement to the beginning of the method. This ensures the cleanup tasks are executed regardless of where a potential panic or error may occur in the function.

2. We need to ensure data consistency. We need modify the code so that the `valid` flag is only set to true once we have successfully saved the share in the database."

3. we can simplify the error handling process. Instead of passing a `malformed` parameter, please incorporate the following into the `defer` statement. It uses `errors.As()` to check if the error is of type `*abiparser.MalformedEventError`. This way, the block executes if `valid` is `true` or if we have a malformed event error:
   ```
   var malformedEventErr *abiparser.MalformedEventError
   if valid || errors.As(err, &malformedEventErr) {
       ...
   }
   ```

operator/validator/event_handler.go
```diff
@@ -175,7 +183,8 @@ func (c *controller) handleValidatorAddedEvent(
 	sharesExpectedLength := encryptedKeyLength*operatorCount + pubKeysOffset
 
 	if sharesExpectedLength != len(event.Shares) {
-		return nil, errors.Errorf("%s event shares length is not correct", abiparser.ValidatorAdded)
+		malformed = true
+		return nil, &abiparser.MalformedEventError{Err: errors.Errorf("%s event shares length is not correct", abiparser.ValidatorAdded)}
 	}
```
olegshmuelov, line 188: we have to set the err and only then return it
```go
if sharesExpectedLength != len(event.Shares) {
  err = &abiparser.MalformedEventError{
	  Err: errors.Errorf(
		  "%s event shares length is not correct: expected %d, got %d",
		  abiparser.ValidatorAdded,
		  sharesExpectedLength,
		  len(event.Shares),
	  ),
  }
  return nil, err
}
```

operator/validator/event_handler.go
```diff
@@ -184,19 +193,13 @@ func (c *controller) handleValidatorAddedEvent(
 
 	// verify sig
 	if err = verifySignature(event.Signature, event.Owner, event.PublicKey, nonce); err != nil {
+		malformed = true
 		return nil, &abiparser.MalformedEventError{
 			Err: errors.Wrap(err, "failed to verify signature"),
 		}
 	}
```
olegshmuelov, line 200: ditto
```go
if err = verifySignature(event.Signature, event.Owner, event.PublicKey, nonce); err != nil {
  err = &abiparser.MalformedEventError{Err: errors.Wrap(err, "failed to verify signature")}
  return nil, err
}
```

operator/validator/event_handler.go
```diff
@@ -184,19 +193,13 @@ func (c *controller) handleValidatorAddedEvent(
...
 
 	pubKey := hex.EncodeToString(event.PublicKey)
-	// TODO: check if need
-	if ongoingSync {
-		if _, ok := c.validatorsMap.GetValidator(pubKey); ok {
-			logger.Debug("validator was loaded already")
-			return nil, nil
-		}
-	}
```
olegshmuelov, line 199: I would say it's critical to remove it since we have to test it first

operator/validator/event_handler.go
y0sher: why?

operator/validator/event_handler.go
y0sher: ok we can leave it then

operator/validator/event_handler.go
```diff
@@ -6,18 +6,17 @@ import (
...
 	"github.com/bloxapp/ssv/eth1/abiparser"
 	"github.com/bloxapp/ssv/exporter"
 	"github.com/bloxapp/ssv/protocol/v2/ssv/validator"
 	"github.com/bloxapp/ssv/protocol/v2/types"
 	registrystorage "github.com/bloxapp/ssv/registry/storage"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/herumi/bls-eth-go-binary/bls"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
```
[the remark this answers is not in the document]
y0sher: I agree, it was goimports formatter that automatically updated this.

GalRogozinski:
A little question.
I just took a sneak peek. I see that "malformed" also means "invalid" (when you check sigs for example).

So can you explain in the description what is the third state an event can have besided valid and malformed?

operator/validator/event_handler.go
olegshmuelov: I've noticed that there is no handling for a malformed event after the function [`c.onShareCreate(logger, event)`](https://github.com/bloxapp/ssv/blob/c1b73a0581cde68280778cfe465c7a5ada338d21/operator/validator/event_handler.go#L209) is called. Specifically, it appears that `malformed = true` is not set under these circumstances.

This lack is problematic because if `onShareCreate` returns an error, it's unclear whether it's due to a malformed event or a system error. This could result in the nonce being falsely incremented, which could lead to corrupted data

To better manage this, I suggest using the `errors.As()` function as I outlined in [(section 3)](https://github.com/bloxapp/ssv/pull/975#pullrequestreview-1452505021). This function can be used to check if an error is of a specific type, in this case, a malformed event error. By implementing this change, we can better differentiate between error types and prevent false nonce increments.

operator/validator/event_handler.go
```diff
@@ -141,30 +142,25 @@ func (c *controller) handleValidatorAddedEvent(
...
 	event abiparser.ValidatorAddedEvent,
 	ongoingSync bool,
-) ([]zap.Field, error) {
-	_, found, err := c.eventHandler.GetEventData(event.TxHash)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get event data")
+) (logFields []zap.Field, err error) {
+	var valid bool
+	defer func() {
+		err = c.handleValidatorAddedEventDefer(valid, err, event)
```
y0sher, line 148: I rather the code be here than jumping to a different func. I know it looks a bit less clean but its easier to follow.

olegshmuelov:
> A little question. I just took a sneak peek. I see that "malformed" also means "invalid" (when you check sigs for example).
> 
> So can you explain in the description what is the third state an event can have besided valid and malformed?

In the context of this code, "valid" and "malformed" are two states that an event can have. An event is considered "valid" if it passes all the checks and requirements defined in our validation process. This includes correct formatting, accurate information, and successful signature verification.

On the other hand, an event is considered "malformed" if it fails any of the checks or requirements. For example, if the event's signature does not verify correctly, or if the length of the shares does not match the expected length, the event is considered "malformed". This doesn't necessarily mean that the event is inherently "invalid", but rather that it does not conform to the expected format or requirements.

The third state an event could be in is "system error". This state is not related to the content of the event itself, but to the environment in which the event is processed. If there's a problem with the system, like a database error or a network issue, that prevents the event from being processed correctly, we can encounter a "system error" state. It's important to note that in the case of a system error, the function returns immediately without incrementing the nonce.

olegshmuelov (approved):

registry/storage/events_test.go
```diff
@@ -0,0 +1,55 @@
...
+		require.NoError(t, err)
+		require.Nil(t, nonExistingEvent)
+		require.False(t, found)
+	})
+
+	t.Run("create and get event", func(t *testing.T) {
+		edToCreate := &storage.EventData{}
+		copy(edToCreate.TxHash[:], "0x1")
+		err := storageCollection.SaveEventData(edToCreate.TxHash)
+		require.NoError(t, err)
```
nkryuchkov, line 34: Could you please add a new line below for better readability?

registry/storage/recipients_test.go
```diff
@@ -120,6 +121,169 @@ func TestStorage_SaveAndGetRecipientData(t *testing.T) {
 			require.Equal(t, r.FeeRecipient, recipients[r.Owner])
 		}
 	})
+
+	t.Run("create recipient should not initializing nonce", func(t *testing.T) {
+		rdToCreate := &storage.RecipientData{
+			Owner: common.BytesToAddress([]byte("0x11111")),
+		}
+		rd, err := storageCollection.SaveRecipientData(rdToCreate)
+		require.NoError(t, err)
```
nkryuchkov, line 130: I'd add a new line below to separate saving/getting actions

operator/validator/event_handler.go
```diff
@@ -230,7 +249,33 @@ func (c *controller) handleValidatorAddedEvent(
...
+	var malformedEventErr *abiparser.MalformedEventError
+
+	if valid || errors.As(err, &malformedEventErr) {
+		saveErr := c.eventHandler.SaveEventData(event.TxHash)
+		if saveErr != nil {
+			wrappedErr := errors.Wrap(saveErr, "could not save event data")
+			if err == nil {
+				return wrappedErr
+			}
+			return errors.Wrap(err, wrappedErr.Error())
```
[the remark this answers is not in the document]
y0sher: Agree with this point @olegshmuelov, maybe wrapping doesn't suit this case.

operator/validator/event_handler.go
```diff
@@ -198,30 +201,46 @@ func (c *controller) handleValidatorAddedEvent(
...
+		err = &abiparser.MalformedEventError{
+			Err: errors.Errorf(
+				"validator share already exists with different owner address: expected %s, got %s",
+				validatorShare.OwnerAddress.String(),
+				event.Owner.String(),
+			),
+		}
+		return nil, err
 	}
 	if !found {
```
y0sher, line 221: I assume `found && event.Owner == validatorShare.Owner` is prevented on the contract level?

operator/validator/event_handler.go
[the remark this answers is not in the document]
olegshmuelov: the [Join](https://pkg.go.dev/errors#Join) LGTM
but we have to upgrade to go 1.20 first
there is an open PR for this
https://github.com/bloxapp/ssv/pull/971

operator/validator/event_handler.go
olegshmuelov: Right, it's handled on the contract side

PR #979, opened 2023-05-31
Log Panics

moshe-blox:
This PR recovers from panics to log them with zap, and then re-panics to preserve existing behavior.

Example:
```bash
2023-05-31T20:11:20.101180Z     PANIC   Recovered from panic    {"panic": "not implemented", "stackTrace": "...full-stack-trace..."}
```

### DIFF
--- a/logging/global.go
+++ b/logging/global.go
@@ -3,6 +3,7 @@ package logging
 import (
 	"io"
 	"os"
+	"runtime/debug"
 	"time"
 
 	"gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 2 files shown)


logging/global.go
```diff
@@ -93,3 +95,16 @@ func SetGlobalLogger(levelName string, levelEncoderName string, logFormat string
 
 	return nil
 }
+
+func CapturePanic(logger *zap.Logger) {
+	if r := recover(); r != nil {
+		// defer logger.Sync()
```
nkryuchkov, line 101: does this line need to be kept?

lior-blox: @moshe-blox

PR #980, opened 2023-06-01
Operate Exiting Validators

moshe-blox:
This PR expands the meaning of "active validators" to "validators who should be attesting".

*Condition was changed from `status == 'active_ongoing'` to `status in ['active_ongoing', 'active_exiting']`.*


lior-blox (approved):

olegshmuelov (commented):
Could be we have to add the `active_slashed` status as well
Not sure if the validator should perform his duties during his way to be exited

olegshmuelov:
> active_slashed



> Could be we have to add the `active_slashed` status as well Not sure if the validator should perform his duties during his way to be exited

nm the `active_slashed` validator should not perform duties till he exited.

olegshmuelov (approved):

PR #990, opened 2023-06-06
Refactor Subscriptions

moshe-blox:


go.mod
```diff
@@ -38,7 +38,9 @@ require (
 	github.com/stretchr/testify v1.8.1
 	github.com/wealdtech/go-eth2-util v1.6.3
 	go.opencensus.io v0.24.0
+	go.uber.org/multierr v1.9.0
```
nkryuchkov, line 41: What do you think about leaving a TODO to replace it with `errors.Join` once we update Go to 1.20? It should happen soon because we already have a PR for it

moshe-blox: @nkryuchkov Good call, now that Go 1.20 is merged, I updated this PR :-)

PR #991, opened 2023-06-06
Set read/write timeouts for HTTP servers

nkryuchkov:
Fix `gosec`'s `G114`


y0sher (approved):
Approved for now, though this is lacking graceful shutdowns. this is something we'll want to handle better across all the node. 
@Tomblox1 please add ticket - Node graceful shutdown.

it'll handle also shutting down these http servers by calling `httpServer.Close()` when a shutdown signal is sent.

Tomblox1:
https://bloxxx.atlassian.net/browse/IO1-3539
@nkryuchkov @y0sher

nkryuchkov:
@y0sher I wanted to keep only adding timeouts here, because lack of them may be vulnerable. I think we need to do graceful shutdowns, but it's a bit different issue. Also, we have many other components that need to be shut down gracefully. So I'd gather them all in the graceful shutdown issue.

nkryuchkov:
@y0sher It seems I misunderstood you. Sorry for the confusion.

PR #992, opened 2023-06-06
network/discovery: return error instead of negative value in nsToSubnet

nkryuchkov:


y0sher (approved):

moshe-blox:
Since we're already on it, let's maybe refactor `nsToSubnet` to validate the number is in range?
Range should be between 0 and `dvs.fork.Subnets()`.
If you refactor `nsToSubnet` to a method, it will have access to `fork`.

What I'm worried about is that malicious nodes pass a number out of range, and our node crashes because we later (somewhere in the code) try to access a slice element which doesn't exist (such as `subnets[18238123]`)

nkryuchkov:
@moshe-blox Is the condition `0 <= subnet < dvs.fork.Subnets()`?

nkryuchkov:
@moshe-blox done

moshe-blox:
yup, bc it's a uint you don't need to check for <0 and iirc 0 is a valid subnet

moshe-blox (approved):

PR #993, opened 2023-06-06
eth1/goeth, monitoring/metrics, network/p2p: crash if setup fails

nkryuchkov:
- If metrics server cannot be started, it likely means a configuration error which should not be hidden
- If streaming smart contract events fails, then `listenToSubscription` doesn't happen anymore, which means it's better to restart and try again. Probably we could add some code to handle this case instead.
- If `n.cfg.Subnets` cannot be parsed, then config is malformed, and crashing would show this


y0sher (approved):

moshe-blox:
> * If metrics server cannot be started, it likely means a configuration error which should not be hidden
> * If streaming smart contract events fails, then `listenToSubscription` doesn't happen anymore, which means it's better to restart and try again. Probably we could add some code to handle this case instead.
> * If `n.cfg.Subnets` cannot be parsed, then config is malformed, and crashing would show this

+1

We shouldn't ignore wrong configuration or bad environment (occupied ports), we must crash

moshe-blox (approved):

PR #995, opened 2023-06-06
Remove duplicated imports

nkryuchkov:
Some imports are duplicated, e.g.:
```go
	"github.com/bloxapp/ssv-spec/qbft"
```


y0sher (approved):
Good change, we have too many identical names in the node and spec codebase and it is very easy to get confused.

PR #996, opened 2023-06-06
Fix builtin shadowing

nkryuchkov:
`copy` is a builtin and should not be shadowed


y0sher (approved):
@moshe-blox , do we need to do something to align this with spec?

moshe-blox (approved):

PR #998, opened 2023-06-06
Fix error handling issues

nkryuchkov:
- Fix `%W` verb
- Error type assertion doesn't work with wrapped errors, `errors.As` should be used instead
- Comparison using `==` or `!=` doesn't work with wrapped errors, `errors.Is` should be used instead

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -464,6 +464,7 @@ linters:
     - goconst
     - unconvert
     - gosec
+    - errorlint
 
 #  disable:
 #    - maligned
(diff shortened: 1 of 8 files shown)


y0sher (approved):
didn't know about `%w` that's cool you don't really need Wrap

nkryuchkov:
@y0sher yes, https://github.com/pkg/errors was discontinued in favor of Go 1.13 error handling (`fmt.Errorf` with `%w`, `errors.Is`, `errors.As`)

moshe-blox (approved):

PR #999, opened 2023-06-07
Fix Runtime Subscriptions

moshe-blox:
This PR fixes a case where `p2pNetwork.Subscribe` is being called in runtime as new validators join, but the subnets record isn't updated neither in the node's metadata and the new topics aren't being registered to.

This is a temporary PR until https://github.com/bloxapp/ssv/pull/990 for until is ready.

### DIFF
--- a/eth1/goeth/goETH.go
+++ b/eth1/goeth/goETH.go
@@ -174,7 +174,12 @@ func (ec *eth1Client) fireEvent(log types.Log, name string, data interface{}) {
 
 // streamSmartContractEvents sync events history of the given contract
 func (ec *eth1Client) streamSmartContractEvents(logger *zap.Logger) error {
-	logger.Debug("streaming smart contract events")
+	currentBlock, err := ec.conn.BlockNumber(ec.ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to get current block")
+	}
+	logger.Debug("streaming smart contract events",
+		zap.Uint64("current_block", currentBlock))
 
 	contractAbi, err := abi.JSON(strings.NewReader(ec.contractABI))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


network/p2p/p2p.go
```diff
@@ -213,51 +213,61 @@ func (n *p2pNetwork) isReady() bool {
...
-	n.activeValidators.Range(func(pkHex string, status validatorStatus) bool {
-		if status == validatorStatusInactive {
-			return true
+	// TODO: this is a temporary fix to update subnets when validators are added/removed,
+	// there is a pending PR to replace this: https://github.com/bloxapp/ssv/pull/990
+	for {
+		time.Sleep(time.Second * 4)
+
+		start := time.Now()
+		logger := logger.Named(logging.NameP2PNetwork)
```
y0sher, line 222: should not be inside the loop

network/p2p/p2p.go
```diff
@@ -213,51 +213,61 @@ func (n *p2pNetwork) isReady() bool {
...
 		}
-		subnet := n.fork.ValidatorSubnet(pkHex)
-		if _, ok := visited[subnet]; ok {
+		newSubnets := make([]byte, n.fork.Subnets())
+		n.activeValidators.Range(func(pkHex string, status validatorStatus) bool {
+			if status == validatorStatusInactive {
+				return true
+			}
+			subnet := n.fork.ValidatorSubnet(pkHex)
+			if _, ok := visited[subnet]; ok {
```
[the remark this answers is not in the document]
moshe-blox: deleted

just to clarify, this PR didn't change the code here, except for wrapping it in a `for` loop with a `Sleep`

eth1/goeth/goETH.go
```diff
@@ -174,7 +174,12 @@ func (ec *eth1Client) fireEvent(log types.Log, name string, data interface{}) {
 
 // streamSmartContractEvents sync events history of the given contract
 func (ec *eth1Client) streamSmartContractEvents(logger *zap.Logger) error {
-	logger.Debug("streaming smart contract events")
+	currentBlock, err := ec.conn.BlockNumber(ec.ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to get current block")
+	}
+	logger.Debug("streaming smart contract events",
+		zap.Uint64("current_block", currentBlock))
```
moshe-blox, line 183: added this as an unrelated small change so we can debug a block-skipping bug

PR #1003, opened 2023-06-07
Fix/merge lint errors

lior-blox:


monitoring/metrics/handler.go
```diff
@@ -86,20 +86,18 @@ func (mh *metricsHandler) Start(logger *zap.Logger, mux *http.ServeMux, addr str
 	mux.HandleFunc("/database/count-by-collection", mh.handleCountByCollection)
 	mux.HandleFunc("/health", mh.handleHealth)
 
-	go func() {
```
y0sher, line 89: This means start never returns unless there's an error. I'm not sure we want this.

moshe-blox (approved):

PR #1015, opened 2023-06-12
Go 1.20 in GitHub actions & Dockerfiles

moshe-blox:


.github/workflows/full-test.yml
```diff
@@ -19,9 +19,9 @@ jobs:
         run: sudo apt-get update && sudo apt-get install make
 
       - name: Setup Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
-          go-version: 1.19
+          go-version: "1.20.x"
```
y0sher, line 24: why not just `1.20` ?

y0sher (dismissed):

.github/workflows/full-test.yml
moshe-blox: > why not just `1.20` ?

there was an issue with that, i forgot which exactly, but it either didn't get the latest or not worked at all

moshe-blox:
Regarding the `-race` flag, I found that we're not the only ones that started having issues with it in Go 1.20: https://github.com/golang/go/issues/58867

y0sher (approved):

lior-blox (approved):

PR #1020, opened 2023-06-16
Eth1 refactoring

nkryuchkov:
https://github.com/bloxapp/ssv/blob/eth1-refactor/eth/design.md


eth1_refactor/eth1client/eth1client.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+			return false, false
+		}
+		return true, false
+	}, ec.reconnectionInitialInterval, ec.reconnectionMaxInterval+(ec.reconnectionInitialInterval))
+
+	logger.Info("reconnected")
+}
+
+func (ec *Eth1Client) setLastFetchedBlock(block uint64) {
```
y0sher, line 264: I', thinking it might be better to store this in another component and keep Eth1 as just a client without logic.

nkryuchkov: @y0sher so we extract `streamLogsToChan` to another component and keep the Eth1Client with `SubscribeNewHead` and `FilterLogs` methods instead?

y0sher: No, just pass the block number into the `SteamLogs` function. last fetched block should be saved and retrieved from the DB. (what used to be syncoffset). 
so the updating of it should be a part of the DB tx. 
Eth1 should be passed the required block to start from every time, it should not care about saving this.

nkryuchkov: I see what you mean. Yes, we can actually extract it to a variable in `streamLogsToChan` and pass it to `StreamLogs`

docs/design/eth1.md
```diff
@@ -0,0 +1,380 @@
...
+TBD
+
+
+### EventDataHandler
+
+Gets batched events from EventDispatcher (EventBatcher). Updates SyncOffset (LastProcessedBlock). Notifies about tasks from events.
+
+#### Has methods:
+
+- `HandleOngoingEventBatch(eventBatchCh <-chan EventBatch)`
```
nkryuchkov, line 293: consider `HandleOngoingBlockEvents`

docs/design/eth1.md
```diff
@@ -0,0 +1,380 @@
...
+
+Goes over the channel and sends each EventBatch for execution to processEvent asyncronously. After each EventBatch it notifies subscribers about a new task.
+
+- `HandlePastEventBatches(eventBatches []EventBatch)`
```
nkryuchkov, line 297: Just write to DB, no need to notify about new tasks. Instead, need to block

nkryuchkov: Consider passing one eventBatch (one `BlockEvents`)

nkryuchkov: Summary: When we do past events, we don't need to trigger anything after this, because we can just start the node and the DB state would contain the changes from past events processing

docs/design/eth1.md
nkryuchkov: Take task logic out of this func. Instead of these 2 funcs, have 1 (receives chan). Task is triggered outside of EventDataHandler

nkryuchkov:
```go
func HandleEventLog(blockEvents <-chan BlockEvents) {
//    store to db one by one, no tasks processing/triggering here
}
```

docs/design/eth1.md
```diff
@@ -0,0 +1,380 @@
...
+
+Close the loop and event processing
+
+- `AddSubscriber(subscriber EventSubscriber)`
+
+Adds a subscriber for batched events. Consider replacing it with another pattern as we'll have only 1 subscriber
+
+- `ongoingEventBatcher(events <-chan Event)`
+
+Batches ongoing events from channel and streams batches to subscribers (HandleOngoingEventBatch) as `<-chan EventBatch` (EventBatch represents a batch of events for one block)
```
nkryuchkov, line 271: It should also dispatch tasks (pass to `TaskExecutor`) after DB changes are done by `HandleOngoingEventBatch`

docs/design/eth1.md
```diff
@@ -0,0 +1,380 @@
...
+
+#### Has methods:
+
+##### Exported:
+
+- `FetchPastLogs(ctx context.Context, fromBlock uint64) ([]ethtypes.Log, error)`
+
+Fetches past blocks.
+
+- `StreamLogs(ctx context.Context) <-chan ethtypes.Log`
```
nkryuchkov, line 207: Instead of FilterLogs call custom func that calls FilterLogs multiple times and batches

eth1_refactor/eventdispatcher/event_dispatcher.go
```diff
@@ -0,0 +1,74 @@
+package eventdispatcher
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+)
+
```
y0sher, line 9: I'd move all the interfaces above to here, having them in separate files doesn't really feel clean. And in practice they are used here, so it makes sense to have them here.

eth1_refactor/taskexecutor/task_executor.go
```diff
@@ -0,0 +1,27 @@
+package taskexecutor
```
y0sher, line 1: thought: still not sure, but it might be good to move this package to be next to validatorcontroller, or just import it from there.

cli/operator/node.go
```diff
@@ -116,7 +116,13 @@ var StartNodeCmd = &cobra.Command{
...
 
 		cfg.ETH2Options.Context = cmd.Context()
-		el, cl := setupNodes(logger, operatorData.ID, networkConfig, slotTicker)
+
+		cfg.ETH2Options.Graffiti = []byte("SSV.Network")
+		cfg.ETH2Options.GasLimit = spectypes.DefaultGasLimit
+		cfg.ETH2Options.Network = networkConfig.Beacon
+
+		el := setupEth2(logger, operatorData.ID, slotTicker)
+		cl := setupEth1(logger, networkConfig.RegistryContractAddr)
```
y0sher, line 125: that should be opposite, eht1 = el, eth2 = cl

eth/eventdatahandler/handlers.go
```diff
@@ -0,0 +1,550 @@
...
+	od := &registrystorage.OperatorData{
+		PublicKey:    event.PublicKey,
+		OwnerAddress: event.Owner,
+		ID:           event.OperatorId,
+	}
+
+	// throw an error if there is an existing operator with the same public key and different operator id
+	if edh.operatorData.ID != 0 && bytes.Equal(edh.operatorData.PublicKey, event.PublicKey) && edh.operatorData.ID != event.OperatorId {
+		edh.logger.Warn("malformed OperatorAdded event: operator registered with the same operator public key")
+		return nil
```
y0sher, line 50: shouldn't this be an error then?

eth/sharestorage/share_storage.go
```diff
@@ -0,0 +1,56 @@
+package sharestorage
+
+import (
+	"github.com/cornelk/hashmap"
+
+	"github.com/bloxapp/ssv/protocol/v2/types"
+	ssvtypes "github.com/bloxapp/ssv/protocol/v2/types"
+)
+
+type ShareStorage struct {
```
y0sher, line 10: name is kinda misleading, there's no storage here.

eth/eventbatcher/event_batcher.go
```diff
@@ -0,0 +1,84 @@
...
+			currentBlockEvents.Events = append(currentBlockEvents.Events, event)
+		}
+
+		if len(currentBlockEvents.Events) != 0 {
+			blockEvents <- currentBlockEvents
+		}
+	}()
+
+	return blockEvents
+}
```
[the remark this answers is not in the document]
nkryuchkov: I still haven't thought about logging/metrics/etc in this package, therefore I left it as a struct

eth/eventdatahandler/handlers.go
nkryuchkov: Handlers are designed in a way that returned error interrupts the flow. So it returns only critical errors like DB

eth/eventdatahandler/handlers.go
```diff
@@ -0,0 +1,550 @@
...
+	isOperatorShare := validatorShare.BelongsToOperator(edh.operatorData.ID)
+	if isOperatorShare {
+		edh.metrics.ValidatorInactive(event.PublicKey)
+	}
+
+	return nil
+}
+
+func (edh *EventDataHandler) saveEvent(txn eventdb.RW, event *contract.ContractValidatorAdded) error {
+	if err := txn.SaveEventData(event.Raw.TxHash); err != nil {
```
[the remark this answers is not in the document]
nkryuchkov: @y0sher Is this what you mean? https://github.com/bloxapp/ssv/pull/1020/commits/3bcdef0ad7056728a57cdf83d5ab5c20c3b26e1a

cli/operator/node.go
```diff
@@ -159,36 +189,100 @@ var StartNodeCmd = &cobra.Command{
...
+			eventBatcher,
+			eventDataHandler,
+			eventdispatcher.WithLogger(logger),
+			eventdispatcher.WithMetrics(metricsReporter),
+			eventdispatcher.WithNodeProber(nodeProber),
+		)
+
+		txn := eventDB.ROTxn()
+		defer txn.Discard()
+
```
y0sher, line 258: why do we need it here? if thats the only way to access `GetLastProcessedBlock`, why not call `txn.Discard()` after getting it ?

nkryuchkov: We need `txn.Discard()` because we must call it when we finish working with the transaction. Actually I could add a `GetLastProcessedBlock` method to event DB, but I think that would make event DB API less clean.

I agree that we can call `txn.Discard()` earlier. Fixed.

eth/eventdatahandler/task.go
```diff
@@ -0,0 +1,101 @@
...
+
+	ssvtypes "github.com/bloxapp/ssv/protocol/v2/types"
+)
+
+type Task interface {
+	Execute() error
+}
+
+type AddValidatorTask struct {
+	executor  TaskExecutor
```
y0sher, line 14: this could probably be a thinner interface containing only `AddValidator` or even just a function. 
this could applies to all tasks below, it might be useful for tests.

nkryuchkov: Actually functions need type aliases to look good which is not shorter than interfaces, so let's proceed with interfaces. Fixed

moshe-blox:
hey @nkryuchkov 

can we try to integrate [this PR](https://github.com/bloxapp/ssv/pull/1016) into Eth1 refactoring?

cli/operator/node.go
```diff
@@ -159,36 +187,100 @@ var StartNodeCmd = &cobra.Command{
...
-		nodeChecker.Wait()
+		nodeProber.Wait()
 
-		nodeUnreadyHandler := func() {
+		nodeProber.SetUnreadyHandler(func() {
 			logger.Fatal("Ethereum node(s) are either out of sync or down. Ensure the nodes are ready to resume.")
+		})
+
+		metricsReporter.SSVNodeHealthy()
+
```
y0sher, line 225: I would add comment that the code beneath is meant to handle events from execution client. or move all of it to a function that specifics it .

eth/contract/generate.go
```diff
@@ -0,0 +1,3 @@
+package contract
+
```
y0sher, line 2: would add a comment here about which version of the abi we use, maybe link the file in repo or specify the version of the contract.

eth/sharemap/share_map.go
y0sher, line 1: don't we have this implemented already?

eth/eventdatahandler/event_data_handler.go
```diff
@@ -0,0 +1,536 @@
...
+
+		logger.Info("task execution finished", fields.Count(len(tasks)))
+	}
+
+	return lastProcessedBlock, nil
+}
+
+func (edh *EventDataHandler) processBlockEvents(blockEvents eventbatcher.BlockEvents) ([]Task, error) {
+	txn := edh.eventDB.RWTxn()
+	defer txn.Discard()
```
y0sher, line 146: should `Discard` be called after `Commit` as well ?

eth/eventdatahandler/handlers.go
```diff
@@ -0,0 +1,560 @@
+package eventdatahandler
```
y0sher, line 1: @pavelkrolevets / @nkryuchkov 
TODO: might be too big to be part of this PR, we need to add tests to each of the handlers , should probably be a mock DB where we test that after running the handler we check that the DB state is as expected.

docs/design/eth1.md
```diff
@@ -0,0 +1,381 @@
+# [ETH1 - Information Retrival] Design
```
lior-blox, line 1: @nkryuchkov lets maybe have it in the eth package?

nkryuchkov: @lior-blox I agree. Moved

eth/eventdatahandler/event_data_handler.go
nkryuchkov: I doesn't have to be but can be. 

```go
// Discard discards a created transaction. This method is very important and must be called. Commit
// method calls this internally, however, calling this multiple times doesn't cause any issues. So,
// this can safely be called via a defer right when transaction is created.
```

I preferred to keep `Discard` because it's a must for every transaction and I wanted to explicitly show that transaction would be finalized correctly. I think `Commit` calling `Discard` is implementation detail, so if we remove `Discard`, the one who reads the code would need to check docs to make sure that `Commit` is enough.

nkryuchkov:
> hey @nkryuchkov
> 
> can we please try to integrate [this PR](https://github.com/bloxapp/ssv/pull/1016) into Eth1 refactoring?

@moshe-blox Sure. Done

eth/eventdatahandler/event_data_handler.go
y0sher: yeah I understand and agree lets keep it.

eth/eventbatcher/event_batcher_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/bloxapp/ssv/eth/eventbatcher"
+)
+
+func TestEventBatcher_BatchEvents(t *testing.T) {
```
y0sher, line 13: shouldn't we write a test that makes sure our batch size limit is enforced?

cli/operator/node.go
nkryuchkov: Extracted to a function

eth/eventdatahandler/event_data_handler.go
```diff
@@ -0,0 +1,536 @@
...
+) (*EventDataHandler, error) {
+	abi, err := contract.ContractMetaData.GetAbi()
+	if err != nil {
+		return nil, fmt.Errorf("get contract ABI: %w", err)
+	}
+
+	filterer, err := client.Filterer()
+	if err != nil {
+		return nil, fmt.Errorf("create contract filterer: %w", err)
+	}
```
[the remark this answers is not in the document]
nkryuchkov: I agree. Done

eth/sharemap/share_map.go
nkryuchkov: We have `sharesStorage` which is a wrapper for both share DB and share map. I cannot reuse it because DB is implemented not in a way I need to. Therefore I needed to have my own share map as well. However, I thought that the original share map would not be used but it is used when updating validator metadata. So this is an issue that needs to be solved before merging

moshe-blox: isn't it better to try supporting transactions in the extisting sharesStorage implementation?

for example, you could ask for badger.Txn in every method, and only if given nil then use the `db` directly, and for convenience this behavior can be encapsulated in a method

for that you'd likely need to implement a struct that fulfills our database interface (`IDb` iirc?) on top of a badger.Txn, so all our database structs (such as sharesStorage) can use transactions like they use databases w/out code changes

wdyt?

eth/eventbatcher/event_batcher.go
```diff
@@ -0,0 +1,46 @@
+package eventbatcher
```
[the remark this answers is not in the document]
nkryuchkov: Added logs

eth/eventbatcher/event_batcher_test.go
nkryuchkov: We don't have a limit like that right now. Why do we need it? We want to process all block events in one transaction. Do you just want to forbid creating enormously big batches?

eth/sharemap/share_map.go
nkryuchkov: @moshe-blox yes, that's likely the best way for now

eth/eventdatahandler/event_data_handler.go
```diff
@@ -0,0 +1,525 @@
...
+		edh.logger.Warn("unknown local event name", fields.Name(event.Name))
+		return nil
+	}
+}
+
+// filterSupersedingTasks filters out tasks that are superseded by other tasks:
+// - opposite tasks (start-stop, stop-start, liquidate-reactivate, reactivate-liquidate)
+// - superseding tasks (updateFee-updateFee)
+// TODO: check if start-stop can be superseded by reactivate-liquidate
+func (edh *EventDataHandler) filterSupersedingTasks(tasks []Task) []Task {
```
y0sher, line 421: I must say this whole process kinda scares me, if we end up removing some task that needs to happen or not removing one that shouldn't we risk being slashed. for example aren't there are more consequences to removing a pair of start-stop? for example other events in between that might happen ( updateFee?) what should we do with these.
I understand why we do this I just think we need to thoroughly make sure we do the right thing and have tests in place.

eth/eventdispatcher/event_dispatcher.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return ed
+}
+
+// Start starts EventDispatcher.
+// It fetches historical logs since fromBlock and passes them for processing.
+// Then it asynchronously runs a loop which retrieves data from ExecutionClient event stream and passes them for processing.
+// Start blocks until historical logs are processed.
+func (ed *EventDispatcher) Start(ctx context.Context, fromBlock uint64) error {
+	ed.logger.Info("starting event dispatcher")
```
y0sher, line 71: please revisit logs in this func, currently its all `Info`

eth/executionclient/execution_client.go
```diff
@@ -0,0 +1,391 @@
...
+	if client == nil {
+		return nil, nil, ErrNotConnected
+	}
+
+	currentBlock, err := client.BlockNumber(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("get current block: %w", err)
+	}
+
+	lastBlock := currentBlock - ec.finalizationOffset
```
y0sher, line 109: so finalizationOffset is constant? 
I think we initially thought to do it with reading the finalized checkpoint from consensus layer and go by it. @moshe-blox you think constant will be enough?

eth/executionclient/execution_client.go
```diff
@@ -0,0 +1,391 @@
...
+
+	heads := make(chan *ethtypes.Header)
+
+	sub, err := client.SubscribeNewHead(ctx, heads)
+	if err != nil {
+		return fromBlock, fmt.Errorf("subscribe heads: %w", err)
+	}
+
+	for {
+		select {
```
y0sher, line 308: check for ctx cancellation as well

y0sher (approved):
Please take a look on my comments. Good job, this is a big and important PR! some logs and metrics are still missing.

eth/executionclient/execution_client.go
nkryuchkov: As I understood, yes. If it shouldn't be, let's discuss the requirements. Added a TODO for this

eth/eventdatahandler/event_data_handler.go
nkryuchkov: I made it optional

eth/eventdatahandler/handlers.go
nkryuchkov: added a TODO

nkryuchkov:
> Please take a look on my comments. Good job, this is a big and important PR! some logs and metrics are still missing.

Let's discuss what's missing because I could have forgotten something. What I know is missing is new metrics (I kept the existing ones) but I think we should better decide what we want to monitor

eth/sharemap/share_map.go
nkryuchkov: @y0sher @moshe-blox done

moshe-blox (changes requested):

eth/executionclient/execution_client.go
moshe-blox: > so finalizationOffset is constant? I think we initially thought to do it with reading the finalized checkpoint from consensus layer and go by it. @moshe-blox you think constant will be enough?

For prater, yes. For mainnet, I think it's too risky because if there is a reorg, everything will break (unless no validators were added in the reorged blocks).

Like we said, it's possible to not wait so long for finalization and instead handle reorgs, but since it's too complex, we preferred to just wait for blocks to be finalized before reacting on them.

`finalizationOffset` is a good idea for us to test on stage without having to wait 2 epochs, but we should definitely wait for actual finalization in a permissionless mainnet environment.

@lior-blox do you think we should do this in a separate PR?

eth/eventdatahandler/handlers.go
```diff
@@ -0,0 +1,561 @@
...
+
+	logger.Debug("processed event")
+	return nil
+}
+
+func (edh *EventDataHandler) handleValidatorAdded(txn basedb.Txn, event *contract.ContractValidatorAdded) (*ssvtypes.SSVShare, error) {
+	logger := edh.logger.With(
+		zap.String("owner_address", event.Owner.String()),
+		zap.Uint64s("operator_ids", event.OperatorIds),
+		zap.String("operator_pub_key", ethcommon.Bytes2Hex(event.PublicKey)),
```
y0sher, line 126: @nkryuchkov , @pavelkrolevets this log is wrong. should be validator pubkey or pubkey. please go over tihs file and make sure logging is correct.

moshe-blox:
```go
func (edh *EventDataHandler) processBlockEvents(blockEvents eventbatcher.BlockEvents) ([]Task, error) {
	txn := edh.nodeStorage.RWTxn()
	defer txn.Discard()

	var tasks []Task
	for _, event := range blockEvents.Events {
		task, err := edh.processEvent(txn, event)
		if err != nil {
			return nil, err
		}

		if task != nil {
			tasks = append(tasks, task)
		}
	}

	if err := edh.nodeStorage.SaveLastProcessedBlock(txn, new(big.Int).SetUint64(blockEvents.BlockNumber)); err != nil {
		return nil, fmt.Errorf("set last processed block: %w", err)
	}

	if err := txn.Commit(); err != nil {
		return nil, fmt.Errorf("commit transaction: %w", err)
	}

	return tasks, nil
}
```
^ this is exactly what we need 🙂

good job 💪

eth/eventdatahandler/handlers.go
pavelkrolevets: @y0sher this is a typo in the log - `event.PublicKey` is a public key of the validator. Fixed.

moshe-blox (changes requested):

eth/executionclient/execution_client.go
```diff
@@ -0,0 +1,394 @@
...
+					logCount++
+				}
+			}
+		}
+
+		ec.logger.Info("fetched historical blocks",
+			zap.Uint64("from", startBlock),
+			zap.Uint64("to", endBlock),
+			fields.Count(logCount),
+		)
```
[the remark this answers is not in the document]
moshe-blox: updated by Lior's PR

eth/eventdispatcher/event_dispatcher.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("handle historical block events: %w", err)
+	}
+	ed.metrics.LastBlockProcessed(lastProcessedBlock)
+
+	if err := <-errorStream; err != nil {
+		return fmt.Errorf("error occurred while fetching historical logs: %w", err)
+	}
+
+	ed.logger.Info("finished processing registry events in batches",
+		zap.Uint64("last_processed_block", lastProcessedBlock))
```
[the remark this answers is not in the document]
y0sher: nice catch, we should probably do it.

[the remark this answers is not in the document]
nkryuchkov: Well, it used to be like this, but I thought that if logs are requested in the future, then the future ones would be marked as processed, although they wouldn't be processed. Also, if 80 is returned, the next batch would be 80-180, so if there are no logs between 80-100, we'd just get logs in range 100-180 and proceed to the next batch.

I agree that we should pass the correct input, but an extra check wouldn't hurt. Also, if we look at the `FetchHistoricalLogs` as at a part of an independent component, it should handle any input correctly. If we pass the `[0;1000000000]` interval to it, then it processes, e.g. `[0;1000]`, then returns that the last processed block is `1000000000`, I don't think it's a correct behavior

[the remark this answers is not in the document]
moshe-blox: > if logs are requested in the future, then the future ones would be marked as processed, although they wouldn't be processed

you mean that they would be processed further down in the pipeline (after we already updated the `lastProcessedBlock` in the database)?

yes, we should only update this inside the event handler, in the same goroutine where we update it when there are actually events in the block

cli/operator/node.go
```diff
@@ -580,7 +580,7 @@ func setupEventHandling(
 		}()
 
 		// TODO: revert
-		select {}
```
y0sher: can you explain?

moshe-blox: removed

moshe-blox (approved):

PR #1031, opened 2023-06-25
kv: fix lack of badger DB report

nkryuchkov:
`report` is being used as `report(logger)` instead of `report(logger)()`, so it should report but not return a function that reports


moshe-blox (approved):

moshe-blox (approved):

y0sher (approved):
Eventually we should strive to not pass loggers to all the functions..

PR #1033, opened 2023-06-27
Eth1 execution client tests

pavelkrolevets:
- added Eth1 chain simulator with events
- added test for FetchHistoricalLogs method of the client
- added test for StreamLogs method of the client


eth/executionclient/execution_client.go
```diff
@@ -268,3 +334,17 @@ func (ec *ExecutionClient) reconnect(ctx context.Context) {
 
 	logger.Info("reconnected")
 }
+
+// func spitToRanges(start uint64, stop uint64, n uint64) []struct{} {
```
nkryuchkov, line 338: Do we need to keep this? Can we delete it?

eth/executionclient/execution_client_test.go
```diff
@@ -36,58 +181,71 @@ func TestExecutionClient(t *testing.T) {
...
 
+	// Generate test chain before we read historical logs
+	generateInitialTestChain(t, done, blockStream, 1008)
+	for blocks := range processedStream {
+		t.Log("Processed blocks: ", len(blocks))
+	}
+	// Fetch all logs history starting from block 0
+	logs, lastBlock, err := client.FetchHistoricalLogs(ctx, 0)
+	require.NoError(t, err)
+	require.NotEmpty(t, logs)
```
nkryuchkov, line 192: Can we check the exact log length and the last block?

nkryuchkov (approved):
LGTM, great job! Just a couple minor questions

eth/executionclient/execution_client.go
```diff
@@ -117,6 +124,66 @@ func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock ui
...
 }
 
+type FetchedLogs struct {
+	logs      []ethtypes.Log
+	lastBlock uint64
+	err       error
+}
+
+// Calls FilterLogs multiple times and batches results to avoid fetching enormous amount of events
+func (ec *ExecutionClient) batchedFilterLogs(ctx context.Context, client *ethclient.Client, query ethereum.FilterQuery) <-chan FetchedLogs {
```
nkryuchkov, line 134: We need to think if we need to test this particular method or testing FetchHistoricalLogs is enough

eth/executionclient/execution_client.go
```diff
@@ -117,6 +124,66 @@ func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock ui
...
+
+	logsChan := make(chan FetchedLogs)
+
+	go func() {
+		defer close(logsChan)
+
+		q := math.Floor(float64((query.ToBlock.Uint64() - query.FromBlock.Uint64()) / ec.logBatchSize))
```
nkryuchkov, line 141: Why do we need `q` to be `float64`? Can we just do 
```go
q := query.ToBlock.Uint64() - query.FromBlock.Uint64()) / ec.logBatchSize
```
instead?

eth/executionclient/execution_client_test.go
```diff
@@ -2,30 +2,175 @@ package executionclient
...
+/*
+Example contract to test event emission:
+
+	pragma solidity >=0.7.0 <0.9.0;
+	contract Callable {
+		event Called();
+		function Call() public { emit Called(); }
+	}
+*/
+const callableAbi = "[{\"anonymous\":false,\"inputs\":[],\"name\":\"Called\",\"type\":\"event\"},{\"inputs\":[],\"name\":\"Call\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
```
nkryuchkov, line 148: Will we need it? Can we delete it?

eth/executionclient/execution_client.go
pavelkrolevets: I wanted to be sure that any uin64 division will not be rounded to upper bound. Looks like Go uint division should give a floor as well. Lets go with 
```
q := query.ToBlock.Uint64() - query.FromBlock.Uint64()) / ec.logBatchSize
```

eth/executionclient/execution_client_test.go
pavelkrolevets: Yes, lets remove it. I saved in case we would need to use ABI

eth/executionclient/execution_client.go
pavelkrolevets: Yes, lets remove as it wasnt needed at the moment

PR #1039, opened 2023-06-29
Domain Type Discovery

moshe-blox:
This PR introduces domain type-based peer discovery to our discovery service.

The code now checks for domain type compatibility when discovering peers. A mismatched domain type results in the node being skipped, ensuring nodes connect only with peers in the same domain. The domain type of each node is stored in the Ethereum Node Record (ENR).

We also adjust how we decorate node records, moving away from a `map[string]interface{}`-based approach to using type-safe decorators for the different parameters.

Incidentally, the field `operatorID` has been removed from the node record because it exposes the operator's IP address (plus, it was unused).

This upgrade ensures better network segmentation and can help to prevent potential cross-talk between nodes operating in different network domains, which we've recently witnessed on JatoV1 & JatoV2.


moshe-blox:
image

Something wrong is going on with the GitHub actions, says "nsubnets in unused" when a line below it's being used 😅

network/discovery/dv5_service.go
```diff
@@ -148,15 +153,33 @@ func (dvs *DiscV5Service) Bootstrap(logger *zap.Logger, handler HandleNewPeer) e
...
 	dvs.discover(dvs.ctx, func(e PeerEvent) {
+		logger := logger.With(
+			fields.ENR(e.Node),
+			fields.PeerID(e.AddrInfo.ID),
+		)
+
+		// Get the peer's domain type, skipping if it mismatches ours.
+		nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record())
+		if err != nil {
+			logger.Debug("could not read domain type")
```
nkryuchkov, line 164: Should we log the error?

network/discovery/dv5_service.go
```diff
@@ -148,15 +153,33 @@ func (dvs *DiscV5Service) Bootstrap(logger *zap.Logger, handler HandleNewPeer) e
...
+		if nodeDomainType != dvs.domainType {
+			logger.Debug("skipping node with different domain type")
+			return
+		}
+
+		// Get the peer's subnets, skipping if it has none.
 		nodeSubnets, err := records.GetSubnetsEntry(e.Node.Record())
 		if err != nil {
-			logger.Debug("could not read subnets", fields.ENR(e.Node))
+			logger.Debug("could not read subnets")
```
nkryuchkov, line 175: Should we log the error?

moshe-blox:
TODO: check the domain type in handshake as well as discovery (in case the discovery record was out of date)

y0sher:
> TODO: check the domain type in handshake as well as discovery (in case the discovery record was out of date)

are you planning to do it as part of this PR or should we create a ticket for it?

network/discovery/dv5_service.go
```diff
@@ -331,15 +354,18 @@ func (dvs *DiscV5Service) createLocalNode(logger *zap.Logger, discOpts *Options,
...
-	err = f.DecorateNode(localNode, map[string]interface{}{
-		"operatorID": opts.OperatorID,
-		"subnets":    opts.Subnets,
-	})
+	err = f.DecorateNode(
+		localNode,
+
+		// Satisfy decorations of forks supported by this node.
+		genesis.DecorateWithDomainType(dvs.domainType),
+		genesis.DecorateWithSubnets(opts.Subnets),
```
y0sher, line 362: should genesis be hardcoded? shouldn't it be derived from the domain type? probably relates to what @olegshmuelov is working on right now.

moshe-blox: now that forks are removed, these became package methods

y0sher (dismissed):
Good! this should solve the connections with different network IDs?

moshe-blox:
> Good! this should solve the connections with different network IDs?

yes

however, you'd still get messages from peers with different domain type until the handshake fails and the connection is closed, but i guess that's a different PR, no?

network/discovery/dv5_service.go
```diff
@@ -124,15 +128,33 @@ func (dvs *DiscV5Service) Bootstrap(logger *zap.Logger, handler HandleNewPeer) e
...
+		nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record())
+		if err != nil {
+			logger.Debug("could not read domain type", zap.Error(err))
+			return
+		}
+		if nodeDomainType != dvs.domainType {
+			logger.Debug("skipping node with different domain type")
+			return
+		}
+
```
lior-blox, line 146: @moshe-blox 
Due to it being not backward compatible and requiring a fork, let's not enforce the check.
and add it once we ready with fork support

moshe-blox: done

moshe-blox:
I just noticed that discovery rejects peers without shared subnets only when we're at the peer limit, however the handshake rejects them in either case (iirc).

So we should either match this condition in the handshake as well or remove it altogether.

network/discovery/dv5_service.go
moshe-blox: https://github.com/bloxapp/ssv/pull/1039/files#diff-f608add482b1df6a8efdebfbc895809e63ffca366b76ffd8192666b0f132ea9cR21

y0sher (approved):

network/discovery/node_record.go
```diff
@@ -0,0 +1,32 @@
...
+	return func(node *enode.LocalNode) error {
+		return records.SetSubnetsEntry(node, subnets)
+	}
+}
+
+// DecorateNode will enrich the local node record with more entries, according to current fork
+func DecorateNode(node *enode.LocalNode, decorations ...NodeRecordDecoration) error {
+	for _, decoration := range decorations {
+		if err := decoration(node); err != nil {
+			return errors.Wrap(err, "failed to decorate node record")
```
nkryuchkov, line 28: let's use `fmt.Errorf` for new code because `github.com/pkg/errors` is discontinued

nkryuchkov (dismissed):

network/discovery/node_record.go
moshe-blox: done

network/discovery/dv5_service.go
```diff
@@ -121,37 +127,60 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+	nodeSubnets, err := records.GetSubnetsEntry(e.Node.Record())
+	if err != nil {
+		return fmt.Errorf("could not read subnets: %w", err)
+	}
+	if bytes.Equal(zeroSubnets, nodeSubnets) {
+		return errors.New("zero subnets")
+	}
+
+	dvs.subnetsIdx.UpdatePeerSubnets(e.AddrInfo.ID, nodeSubnets)
+	if !dvs.limitNodeFilter(e.Node) {
```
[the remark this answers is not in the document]
moshe-blox: done

nkryuchkov (approved):

nkryuchkov (dismissed):

nkryuchkov (approved):

PR #1044, opened 2023-07-06
Duty Scheduler

olegshmuelov:
## 🎯 Refactoring of Duty Scheduler for SSV Node

This pull request introduces substantial refactoring of the Duty Scheduler in the SSV Project. The changes aim to enhance the current architecture by optimizing duty scheduling for upcoming epochs/periods 🚀, improving the execution speed of Attester and Sync Committee duties 🏃‍♂️, handling blockchain reorganizations effectively 🔄, and proactively scheduling duties for upcoming active validators 🔮.

The refactoring restructures the Duty Scheduler and Duty Fetcher, into one primary component that manage validator duties in the Ethereum 2.0 protocol.

**🔑 Key Improvements**:
- The Duty Scheduler now not only manages current epoch duties, but also plans and schedules duties for upcoming epochs and periods .
- The scheduler has been improved to handle re-organizations in blockchain state and to 
- Foresee and schedule duties for validators expected to become active in the beacon chain in the next epoch or sync committee period.
- The scheduler now efficiently handles (un)/scheduling duties in response to contract events

These enhancements aim to provide a more robust, efficient and proactive duty scheduling system, effectively handling upcoming duties and responding promptly to changes in the blockchain state and contract events. Further discussions and testing are required to ensure the efficiency and reliability of the new design 🧪.


operator/duties/attester.go
```diff
@@ -0,0 +1,265 @@
...
+		ValidatorCommitteeIndex: duty.ValidatorCommitteeIndex,
+	}
+}
+
+func (h *AttesterHandler) shouldExecute(logger *zap.Logger, duty *eth2apiv1.AttesterDuty) bool {
+	currentSlot := h.network.Beacon.EstimatedCurrentSlot()
+	// execute task if slot already began and not pass 1 epoch
+	if currentSlot >= duty.Slot && currentSlot-duty.Slot <= 32 {
+		return true
+	} else if currentSlot+1 == duty.Slot {
```
nkryuchkov, line 232:
```suggestion
	}
	if currentSlot+1 == duty.Slot {
```

operator/duties/base_handler.go
```diff
@@ -0,0 +1,86 @@
...
+func (h *baseHandler) IndicesChangeChannel() chan bool {
+	return h.indicesChange
+}
+
+func (h *baseHandler) ReorgChannel() chan ReorgEvent {
+	return h.reorg
+}
+
+type Duties[K ~uint64, D any] struct {
+	m map[K]map[phase0.Slot][]D
```
nkryuchkov, line 66: Should the map be locked with a mutex?

operator/duties/proposer.go
```diff
@@ -0,0 +1,181 @@
...
+//  2. If necessary, fetch duties for the current epoch.
+func (h *ProposerHandler) HandleDuties(ctx context.Context, logger *zap.Logger) {
+	logger = logger.With(zap.String("handler", h.Name()))
+	logger.Info("starting duty handler")
+
+	h.fetchFirst = true
+
+	for {
+		select {
+		case slot := <-h.ticker:
```
nkryuchkov, line 60: Let's add a case for `<-ctx.Done`

operator/node.go
```diff
@@ -156,7 +155,9 @@ func (n *operatorNode) Start(logger *zap.Logger) error {
 	go n.reportOperators(logger)
 
 	go n.feeRecipientCtrl.Start(logger)
-	n.dutyCtrl.Start(logger)
+	if err := n.dutyScheduler.Run(n.context, logger); err != nil {
+		return errors.Wrap(err, "failed to run duty scheduler")
```
nkryuchkov, line 159: Let's not use `"github.com/pkg/errors"` as it's discontinued
```suggestion
		return fmt.Errorf("run duty scheduler: %w", err)
```

operator/duties/validatorregistration.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/bloxapp/ssv/logging/fields"
+)
+
+const validatorRegistrationEpochInterval = uint64(10)
+
+type ValidatorRegistrationHandler struct {
+	baseHandler
+
+	validatorsPassedFirstRegistration map[string]struct{}
```
nkryuchkov, line 18: Do we need a mutex or https://github.com/cornelk/hashmap ?

operator/duties/validatorregistration.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (h *ValidatorRegistrationHandler) Name() string {
+	return spectypes.BNRoleValidatorRegistration.String()
+}
+
+func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context, logger *zap.Logger) {
+	logger = logger.With(zap.String("handler", h.Name()))
+	logger.Info("starting duty handler")
+
+	for slot := range h.ticker {
```
nkryuchkov, line 35: Let's check for `<-ctx.Done()`

operator/duties/scheduler.go
```diff
@@ -0,0 +1,344 @@
...
+		With(fields.CurrentSlot(s.network.Beacon)).
+		With(fields.Slot(duty.Slot)).
+		With(fields.Epoch(s.network.Beacon.EstimatedEpochAtSlot(duty.Slot))).
+		With(fields.PubKey(duty.PubKey[:])).
+		With(fields.StartTimeUnixMilli(s.network.Beacon, duty.Slot))
+}
+
+// waitOneThirdOrValidBlock waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot)
+func (s *Scheduler) waitOneThirdOrValidBlock(slot phase0.Slot) {
+	s.waitMutex.Lock()
```
nkryuchkov, line 292: Can we have a test for this method to make sure that locking `waitMutex` here and in the goroutine doesn't lead to a deadlock?

operator/duties/attester.go
```diff
@@ -0,0 +1,265 @@
...
+// On Indices Change:
+//  1. Execute duties.
+//  2. Reset duties for the current epoch.
+//  3. Fetch duties for the current epoch.
+//  4. If necessary, fetch duties for the next epoch.
+//
+// On Ticker event:
+//  1. Execute duties.
+//  2. If necessary, fetch duties for the next epoch.
+func (h *AttesterHandler) HandleDuties(ctx context.Context, logger *zap.Logger) {
```
nkryuchkov, line 61: Let's pass logger to NewAttesterHandler instead of passing it to each of its methods

operator/duties/base_handler.go
```diff
@@ -0,0 +1,86 @@
...
+	beaconNode          beaconprotocol.BeaconNode
+	network             networkconfig.NetworkConfig
+	validatorController ValidatorController
+	executeDuties       ExecuteDutiesFunc
+	ticker              chan phase0.Slot
+
+	reorg         chan ReorgEvent
+	indicesChange chan bool
+
+	fetchFirst     bool
```
nkryuchkov, line 35: Should this be atomic?

logging/fields/fields.go
```diff
@@ -65,6 +65,7 @@ const (
 	FieldPubKey              = "pubkey"
 	FieldRole                = "role"
 	FieldRound               = "round"
+	FieldEpoch               = "epoch"
```
nkryuchkov, line 68: let's keep in sorted alphabetically

operator/duties/scheduler.go
```diff
@@ -0,0 +1,344 @@
...
+
+	// Initialize the reorg channel
+	s.reorg = make(chan ReorgEvent)
+
+	s.waitCond = sync.NewCond(&s.waitMutex)
+
+	// Subscribe to head events.  This allows us to go early for attestations if a block arrives, as well as
+	// re-request duties if there is a change in beacon block.
+	if err := s.beaconNode.Events(ctx, []string{"head"}, s.HandleHeadEvent(logger)); err != nil {
+		return errors.Wrap(err, "failed to subscribe to head events")
```
nkryuchkov, line 101:
```suggestion
		return fmt.Errorf("failed to subscribe to head events: %w", err)
```

operator/duties/attester.go
olegshmuelov: done

operator/duties/proposer.go
olegshmuelov: done

operator/duties/validatorregistration.go
olegshmuelov: done

logging/fields/fields.go
olegshmuelov: done

operator/duties/scheduler.go
olegshmuelov: done

operator/duties/base_handler.go
olegshmuelov: no, the code designed to not write in parallel

beacon/goclient/sync_committee_contribution.go
```diff
@@ -91,3 +91,14 @@ func (gc *goClient) GetSyncCommitteeContribution(slot phase0.Slot, selectionProo
 func (gc *goClient) SubmitSignedContributionAndProof(contribution *altair.SignedContributionAndProof) error {
 	return gc.client.SubmitSyncCommitteeContributions(gc.ctx, []*altair.SignedContributionAndProof{contribution})
 }
+
+// waitOneThirdOrValidBlock waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot)
+func (gc *goClient) waitOneThirdOrValidBlock(slot phase0.Slot) {
```
y0sher, line 96: I don't see where does a valid block comes in this func. its only waitOneThird

go.mod
```diff
@@ -151,6 +151,7 @@ require (
 	github.com/rivo/uniseg v0.4.3 // indirect
 	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
 	github.com/sirupsen/logrus v1.9.0 // indirect
+	github.com/sourcegraph/conc v0.3.0 // indirect
```
y0sher, line 154: very nice lib, good find.

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+	logger = logger.With(zap.String("handler", h.Name()))
+	logger.Info("starting duty handler")
+
+	h.fetchCurrentEpoch = true
+	if h.shouldFetchNexEpoch(h.network.Beacon.EstimatedCurrentSlot()) {
+		h.fetchNextEpoch = true
+	}
+	h.fetchFirst = true
```
y0sher, line 68: this could be just 
```
if h.shouldFetchNexEpoch(h.network.Beacon.EstimatedCurrentSlot()) {
		h.fetchNextEpoch = true
	}
```
if you would set `h.fetchCurrentEpoch` and `h.fetchFirst` to true in `NewAttesterHandler`

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+			if h.fetchFirst {
+				h.processFetching(ctx, logger, currentEpoch, slot)
+				h.fetchFirst = false
+				h.processExecution(logger, currentEpoch, slot)
+			} else {
+				h.processExecution(logger, currentEpoch, slot)
+				if h.indicesChanged {
+					h.duties.Reset(currentEpoch)
+					h.indicesChanged = false
+				}
```
y0sher, line 89: shouldn't we check if indicies changed before execution? also shouldn't we check it even if `fetchFirst` is true?

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+				h.processFetching(ctx, logger, currentEpoch, slot)
```
y0sher, line 90: it is possible that  we fetched stuff that should be executed on this slot? in other words should we run execution again after fetching?

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+			// Get next epoch's attester duties, but wait until half-way through the epoch
+			// This allows us to set them up at a time when the beacon node should be less busy.
+			if uint64(slot)%h.network.Beacon.SlotsPerEpoch() == h.network.Beacon.SlotsPerEpoch()/2-2 {
+				h.fetchNextEpoch = true
+			}
+
+			// last slot of epoch
+			if uint64(slot)%h.network.Beacon.SlotsPerEpoch() == h.network.Beacon.SlotsPerEpoch()-1 {
+				h.duties.Reset(currentEpoch)
+			}
```
y0sher, line 102: why not set a var 
```suggestion

                        slotsPerEpoch := network.Beacon.SlotsPerEpoch()
			// Get next epoch's attester duties, but wait until half-way through the epoch
			// This allows us to set them up at a time when the beacon node should be less busy.
			if uint64(slot)%slotsPerEpoch == slotsPerEpoch/2-2 {
				h.fetchNextEpoch = true
			}

			// last slot of epoch
			if uint64(slot)%slotsPerEpoch == slotsPerEpoch-1 {
				h.duties.Reset(currentEpoch)
			}
```

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+			} else {
+				h.processExecution(logger, currentEpoch, slot)
+				if h.indicesChanged {
+					h.duties.Reset(currentEpoch)
+					h.indicesChanged = false
+				}
+				h.processFetching(ctx, logger, currentEpoch, slot)
+			}
+
+			// Get next epoch's attester duties, but wait until half-way through the epoch
```
y0sher, line 93: rephrasing suggestion
```suggestion
			// If we're half-way through the epoch, get next epoch's attester duties.
```

operator/duties/attester.go
```diff
@@ -0,0 +1,270 @@
...
+	duties, err := h.beaconNode.AttesterDuties(ctx, epoch, indices)
+	if err != nil {
+		return fmt.Errorf("failed to fetch attester duties: %w", err)
+	}
+
+	for _, d := range duties {
+		h.duties.Add(epoch, d.Slot, d)
+	}
+
+	h.prepareDutiesResultLog(logger, epoch, duties, start)
```
y0sher, line 189: name is kinda funny since you log in this func and not just prepare

operator/duties/validatorregistration.go
olegshmuelov: no need, there no parallel access to the map

operator/duties/scheduler.go
olegshmuelov: the unit tests covering it

beacon/goclient/sync_committee_contribution.go
olegshmuelov: the purpose of this func to release the waiting 1/3 of the slot time on timeout or on valid block arrival

moshe-blox (changes requested):

ekm/eth_key_manager_signer.go
```diff
@@ -66,7 +66,7 @@ func NewETHKeyManagerSigner(logger *zap.Logger, db basedb.IDb, network networkco
 	}
 
 	slashingProtector := slashingprotection.NewNormalProtection(signerStore)
-	beaconSigner := signer.NewSimpleSigner(wallet, slashingProtector, core.Network(network.Beacon.BeaconNetwork))
+	beaconSigner := signer.NewSimpleSigner(wallet, slashingProtector, core.Network(network.Beacon.GetBeaconNetwork()))
```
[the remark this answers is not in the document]
olegshmuelov: GetBeaconNetwork
should return the spec network

GetNetwork
should return the SSV network

all this area should be refactored starting from the spec

operator/duties/attester.go
olegshmuelov: no, the new fetched duties should be executed only on the next slot

ekm/eth_key_manager_signer.go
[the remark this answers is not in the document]
olegshmuelov: see https://github.com/bloxapp/ssv/pull/1044#discussion_r1269319253

operator/duties/proposer.go
```diff
@@ -0,0 +1,177 @@
...
+				if h.shouldExecute(logger, d) {
+					toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleProposer))
+				}
+			}
+			h.executeDuties(logger, toExecute)
+		}
+	}
+}
+
+func (h *ProposerHandler) fetchDuties(ctx context.Context, logger *zap.Logger, epoch phase0.Epoch) error {
```
y0sher, line 133: I think fetchDuties should only fetch duties, return them and the calling function should decide what to do with it (ex: add to the map, log etc..). otherwise this function name should be different.

olegshmuelov: we want to have one place to have it all (the fetch, the log the subscribe and so on)
we can rename it to fetchAndProcessDuties

operator/duties/attester.go
```diff
@@ -0,0 +1,262 @@
...
+		ValidatorIndex:          duty.ValidatorIndex,
+		CommitteeIndex:          duty.CommitteeIndex,
+		CommitteeLength:         duty.CommitteeLength,
+		CommitteesAtSlot:        duty.CommitteesAtSlot,
+		ValidatorCommitteeIndex: duty.ValidatorCommitteeIndex,
+	}
+}
+
+func (h *AttesterHandler) shouldExecute(logger *zap.Logger, duty *eth2apiv1.AttesterDuty) bool {
+	currentSlot := h.network.Beacon.EstimatedCurrentSlot()
```
y0sher: when you call this function you usually already know the current slot, so passing it in instead of calling `h.network.Beacon.EstimatedCurrentSlot()` can reduce times you call it and reduce the dependency in this function.

olegshmuelov: We explicitly estimate the current slot at the exact time of the function call to ensure our operation is based on the most accurate and recent data

beacon/goclient/sync_committee_contribution.go
y0sher: Again I can't see how this function is waiting to the block arrival. you might use it in another place to do so, so this function name should include only what this function does.

operator/duties/attester.go
```diff
@@ -0,0 +1,261 @@
...
+			toExecute := make([]*spectypes.Duty, 0, len(duties)*2)
+			for _, d := range duties {
+				if h.shouldExecute(d) {
+					toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleAttester))
+					toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleAggregator))
+				}
+			}
+			h.executeDuties(h.logger, toExecute)
+		}
+	}
```
y0sher, line 174: this might be just code-style but I prefer less nested code. feel free to dismiss..
```suggestion
	slotMap, ok := h.duties.m[epoch];
	if !ok
	   return
	}
	duties, ok := slotMap[slot]
	if !ok {
	    return
	}
	toExecute := make([]*spectypes.Duty, 0, len(duties)*2)
	for _, d := range duties {
		if h.shouldExecute(d) {
			toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleAttester))
			toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleAggregator))
		}
	}
	h.executeDuties(h.logger, toExecute)
```

beacon/goclient/sync_committee_contribution.go
olegshmuelov: was referring to diff func :)
np renamed it to `waitForOneThirdSlotDuration`

operator/duties/scheduler.go
```diff
@@ -0,0 +1,341 @@
...
+	go fanOut(ctx, s.indicesChg, indicesChangeChs)
+	go fanOut(ctx, s.reorg, reorgChs)
+
+	if ready != nil {
+		ready <- struct{}{}
+	}
+	return handlerPool.Wait()
+}
+
+// TODO: use a fan out package or Prysm's event.Feed, because sends in this implementation are blocking
```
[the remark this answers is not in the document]
y0sher: two approaches can be taken, 
- in, and subscribers channels should be buffered
- add a default clause to the select.

I guess the first one is the best because the default will mean we will skip events and that will cause issues.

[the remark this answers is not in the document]
y0sher: we could also fall into default and retry only the subscribers that were blocking, still I think a small buffer will be ok since none of these event should happen a lot.

operator/duties/scheduler.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// SlotTicker handles the "head" events from the beacon node.
+func (s *Scheduler) SlotTicker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case slot := <-s.ticker:
+			delay := s.network.SlotDurationSec() / 3 /* a third of the slot duration */
```
y0sher, line 206: maybe this should be a const?

operator/duties/scheduler.go
```diff
@@ -0,0 +1,342 @@
...
+		With(fields.CurrentSlot(s.network.Beacon.EstimatedCurrentSlot())).
+		With(fields.Slot(duty.Slot)).
+		With(fields.Epoch(s.network.Beacon.EstimatedEpochAtSlot(duty.Slot))).
+		With(fields.PubKey(duty.PubKey[:])).
+		With(fields.StartTimeUnixMilli(s.network.Beacon.GetSlotStartTime(duty.Slot)))
+}
+
+// waitOneThirdOrValidBlock waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot)
+func (s *Scheduler) waitOneThirdOrValidBlock(slot phase0.Slot) {
+	// Wait for the event or signal
```
y0sher, line 336: aren't we waiting only for the block here?

operator/validator/controller.go
```diff
@@ -151,6 +157,8 @@ type controller struct {
 	// nonCommittees is a cache of initialized nonCommitteeValidator instances
 	nonCommitteeValidators *ttlcache.Cache[spectypes.MessageID, *nonCommitteeValidator]
 	nonCommitteeMutex      sync.Mutex
+
+	indicesChange chan bool
```
y0sher, line 161: this could be an empty struct `indicesChange chan struct{}` since we never use the boolean?

operator/validator/event_handler.go
y0sher, line 1: we don't want to signal removed validators as well?

operator/validator/event_handler.go
```diff
@@ -379,12 +383,16 @@ func (c *controller) handleClusterReactivatedEvent(
...
 		for _, share := range toEnable {
-			_, err = c.onShareStart(logger, share)
+			started, err = c.onShareStart(logger, share)
 			if err != nil {
 				logger.Warn("could not start validator", zap.String("pubkey", hex.EncodeToString(share.ValidatorPubKey)), zap.Error(err))
 			}
 		}
+		if started {
+			c.indicesChange <- true
+		}
```
y0sher, line 395: what if last one in range wasn't started, but 2 befores did. we still need to shoot an event to the channel. this won't work. last one is what matters in this code. I suggest maybe stop changing started after one has returned true.

y0sher (changes requested):
Please go through my comments. generally good work, I love the thorough testing 🥇

operator/duties/scheduler.go
olegshmuelov: do you mean move the `3` to be a const?

operator/duties/scheduler.go
olegshmuelov: no, we are waiting for the block from head event or for the one third of the slot release from scheduler slot ticker

operator/validator/controller.go
olegshmuelov: see https://github.com/bloxapp/ssv/pull/1044#discussion_r1273232868

operator/validator/event_handler.go
olegshmuelov: good catch

operator/duties/scheduler.go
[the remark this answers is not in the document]
moshe-blox: the problem with buffering is that if you were stuck for say 12 seconds (the max timeout), then after 12 sec after the slot has already advanced, you resume reading from slotTicker, but then you get the previous slot from the ticker bc it was buffered

the default clause may work, we have this prototype:
```go
func fanOut[T any](ctx context.Context, in <-chan T, subscribers []chan<- T) {
    for {
        select {
        case <-ctx.Done():
            return
        case item := <-in:
            outbox := make([]chan<- T, len(subscribers))
            copy(outbox, subscribers)

            for len(outbox) > 0 {
                for i, subscriber := range outbox {
                    // Fan out the message to all subscribers.
                    select {
                    case <-ctx.Done():
                        return
                    case subscriber <- item:
                        // Sent -> Remove the subscriber from the outbox.
                        outbox = append(outbox[:i], outbox[i+1:]...)
                    default:
                    }
                }
            }
        }
    }
}
```

however bc it gets a bit complex, we're considering just using Prysm's event.Feed which does a similar thing already

operator/duties/scheduler.go
moshe-blox: @y0sher @olegshmuelov 
this is called `INTERVALS_PER_SLOT` in the Beacon spec

https://github.com/prysmaticlabs/prysm/blob/2674e204cab117b1e902c78d72d8b6faaec7e6a0/config/params/mainnet_config.go#L119

olegshmuelov: adjusted in https://github.com/bloxapp/ssv/pull/1044/commits/4c8d7399dc6536a36faba4c4e20aa4db22adf148

y0sher (approved):
Glad to see this new design and the added testing. good job!

moshe-blox (approved):

andrew-blox:
Manual and Unit test plans can be found here:
https://docs.google.com/spreadsheets/d/15dB_kIRkc43cckAGyttlKTnJKFJtFTT8HdUHIjcUL_I/edit#gid=1806251730

andrew-blox (approved):

operator/duties/scheduler.go
```diff
@@ -0,0 +1,360 @@
...
+	s.pool = pool.New().WithContext(ctx).WithCancelOnError()
+	subscriptionCtx, subscribed := context.WithCancel(ctx)
+	s.pool.Go(func(ctx context.Context) error {
+		defer subscribed()
+		err := s.beaconNode.Events(ctx, []string{"head"}, s.HandleHeadEvent(logger))
+		if err != nil {
+			return fmt.Errorf("failed to subscribe to head events: %w", err)
+		}
+		return nil
+	})
```
y0sher, line 136: what if an error is returned? subscribed is still called and we act like we have head events. I think there's no value in this code being inside a goroutine, it just adds complexity.

operator/validator/controller_test.go
```diff
@@ -73,66 +75,81 @@ func TestHandleNonCommitteeMessages(t *testing.T) {
...
+			ActivationEpoch: phase0.Epoch(101),
 		}),
 
 		"3": newValidator(&beacon.ValidatorMetadata{
-			Balance: 0,
-			Status:  3, // ValidatorStateActiveOngoing
-			Index:   3,
+			Balance:         0,
+			Status:          3, // ValidatorStateActiveOngoing
+			Index:           3,
```
y0sher, line 103: why not use current epoch?

y0sher (changes requested):
last small comment and lets merge

operator/duties/scheduler.go
olegshmuelov: see [only call subscribed if there was no error](https://github.com/bloxapp/ssv/pull/1044/commits/0ed04167602bfd2f7592b0ada5fd6357aff7b35a)

operator/validator/controller_test.go
olegshmuelov: I chose to use the hardcoded value `phase0.Epoch(100)` for the following reasons:

1. **Ensuring a Known State**: Tests should ideally be deterministic, meaning they always produce the same results given the same input. Using a hardcoded value helps to ensure this, as it prevents the test's behavior from changing based on the state of an external variable like "currentEpoch".

2. **Readability**: By using a specific value, it can be clearer to someone reading the test why a specific scenario is being tested.

3. **Isolation**:  If "currentEpoch" is set elsewhere in the code, changing its value could potentially break this test, even if there's nothing wrong with the code being tested. Using a hardcoded value avoids this problem.

Hope this clarifies the decision :)

operator/duties/scheduler.go
olegshmuelov: There is a convention in golang to use the Start function to initiate an operation in a non-blocking manner, often launching goroutines to do the work in the background.

the subscription to head events is running in a goroutine because it needs to be non-blocking. If the subscription were moved out of the goroutine, it could potentially block the rest of the Start function, which goes against the convention.

y0sher: This convention adheres to long actions. Subscribing to headevents should be a short action, and since we don't want to continue if we can't subscribe, it's better to do it in a synchronous matter. After [only call subscribed if there was no error](https://github.com/bloxapp/ssv/pull/1044/commits/0ed04167602bfd2f7592b0ada5fd6357aff7b35a), the situation is now worse, if there was an error all our handler go routines will hang waiting to the subscribe context to be done but it won't event be done, the handler will never run.
When you write a start function, it's best to run in the main start go routine stuff that are mandatory for the function to run, then return error if they fail, even if it blocks for a short time. after that you can start any long running tasks in go routine and unblock.

y0sher: See this exmaple for a start function

```golang

func Start(input string) error {

    if input != GoodInput {
        return ErrBadInput
    }
    if err := CheckRequiredService(); err != nil {
        return err
    }

    go ServiceEventLoop()

    return nil

}

File: operator/validator/controller_test.go (hunk 30 above)
Comment (y0sher) [reply operator/validator/controller_test.go:103]: Honestly since you defined the currentEpoch variable in the test and your testing against it, I think its better to use it. In the future if we'll want to change the value you'll need to edit many places in stead of one.

1. this is incorrect, currectEpoch is an internal test variable. its basically like hardcoded
2. its easier to see differences when using this variable
3. this is incorrect, its set a few lines above and not "elsewhere in the code". 
your reusing the same hardcoded value when you defined a var for this purpose.

File: operator/duties/scheduler.go (hunk 29 above)
Comment (olegshmuelov) [reply operator/duties/scheduler.go]: > the situation is now worse, if there was an error all our handler go routines will hang waiting to the subscribe context to be done but it won't ever be done, the handler will never run.

if there is an error subscribing 
`s.pool = pool.New().WithContext(ctx).WithCancelOnError()`
should cancel the context on error 

and the `Wait`
```
if err := n.dutyScheduler.Wait(); err != nil {
  logger.Fatal("duty scheduler exited with error", zap.Error(err))
}
```
should catch it

y0sher (approved):
Lets merge this!

moshe-blox (approved):

PR #1045, opened 2023-07-09
Unidirectional Handshake & SSV API

moshe-blox:


network/topics/sub_filter.go
```diff
@@ -43,13 +44,25 @@ func (sf *subFilter) CanSubscribe(topic string) bool {
...
 // there are too many subscriptions.
-func (sf *subFilter) FilterIncomingSubscriptions(pi peer.ID, subs []*ps_pb.RPC_SubOpts) ([]*ps_pb.RPC_SubOpts, error) {
+func (sf *subFilter) FilterIncomingSubscriptions(pi peer.ID, subs []*ps_pb.RPC_SubOpts) (res []*ps_pb.RPC_SubOpts, err error) {
+	defer func() {
+		zap.L().Debug(
+			"FilterIncomingSubscriptions",
+			fields.PeerID(pi),
+			zap.Int("before_subscriptions_len", len(subs)),
+			zap.Int("after_subscriptions_len", len(res)),
+			zap.Any("before_subscriptions", subs),
```
y0sher, line 54: This log is super verbose, we should probably not print all the subs array. either make it smaller or remove.

network/topics/controller.go
```diff
@@ -110,6 +110,9 @@ func (ctrl *topicsCtrl) Close() error {
 
 // Peers returns the peers subscribed to the given topic
 func (ctrl *topicsCtrl) Peers(name string) ([]peer.ID, error) {
+	if name == "" {
```
y0sher, line 113: this is only for the API

network/peers/connections/handshaker_test.go
```diff
@@ -68,53 +52,37 @@ func TestHandshakeTestData(t *testing.T) {
...
-			MockNodeInfo: td.NodeInfo,
+		beforeHandshake := time.Now()
+		td.Handshaker.filters = func() []HandshakeFilter {
+			return []HandshakeFilter{
+				func(senderID peer.ID, nodeInfo records.AnyNodeInfo) error {
+					return fmt.Errorf("peer filtered")
+				},
+			}
 		}
+		require.Error(t, td.Handshaker.Handshake(logging.TestLogger(t), td.Conn))
```
y0sher, line 67: I would make sure the err is `peer filtered`

PR #1053, opened 2023-07-10
hotfix: eth1 sync

moshe-blox:
- Sync new blocks that were added while we were syncing history
- SaveEventData for all events, to never replay any event when syncing the same block again


operator/validator/event_handler.go
```diff
@@ -39,29 +40,47 @@ func splitBytes(buf []byte, lim int) [][]byte {
 
 // Eth1EventHandler is a factory function for creating eth1 event handler
 func (c *controller) Eth1EventHandler(logger *zap.Logger, ongoingSync bool) eth1.SyncEventHandler {
-	return func(e eth1.Event) ([]zap.Field, error) {
+	return func(e eth1.Event) (logs []zap.Field, err error) {
```
y0sher, line 43: don't we want to not save when the error was just DB err for example? so we could run it again?

y0sher (dismissed):

operator/validator/event_handler.go
moshe-blox: as discussed in Discord, it's a tough call, both ways are bad and this way requires less code changes & testings

PR #1055, opened 2023-07-12
fix: don't overwrite nonce when saving fee recipient address

moshe-blox:


registry/storage/recipients_test.go
```diff
@@ -66,6 +66,36 @@ func TestStorage_SaveAndGetRecipientData(t *testing.T) {
 		require.NotNil(t, rdFromDB)
 	})
 
+	t.Run("save/get/save fee recipient address without overwriting nonce", func(t *testing.T) {
```
y0sher, line 69: 👍

y0sher (dismissed):

PR #1058, opened 2023-07-13
feat: (`SyncHighestDecided`) wait for minimum peers before requesting highest decided

moshe-blox:


network/p2p/p2p_sync.go
```diff
@@ -75,8 +73,8 @@ func (n *p2pNetwork) LastDecided(logger *zap.Logger, mid spectypes.MessageID) ([
 	if !n.isReady() {
 		return nil, p2pprotocol.ErrNetworkIsNotReady
 	}
-	pid, peerCount := n.fork.ProtocolID(p2pprotocol.LastDecidedProtocol)
-	peers, err := n.getSubsetOfPeers(logger, mid.GetPubKey(), peerCount, allPeersFilter)
+	pid, maxPeers := n.fork.ProtocolID(p2pprotocol.LastDecidedProtocol)
+	peers, err := waitSubsetOfPeers(logger, n.getSubsetOfPeers, mid.GetPubKey(), 3, maxPeers, time.Second*24, allPeersFilter)
```
y0sher, line 77: IMO `3` and `time.Second*24` should be consts or at least vars, solely for the purpose of them being named.

network/p2p/p2p_sync.go
```diff
@@ -300,3 +288,42 @@ func (n *p2pNetwork) peersWithProtocolsFilter(protocols ...libp2p_protocol.ID) f
...
+	}
+	if minPeers < 0 || maxPeers < 0 {
+		return nil, fmt.Errorf("minPeers and maxPeers should not be negative")
+	}
+	if timeout <= 0 {
+		return nil, fmt.Errorf("timeout should be positive")
+	}
+
+	// Wait for minPeers with a deadline.
+	deadline := time.Now().Add(timeout)
```
y0sher, line 311: could've been done with context.Deadline as well. though I don't see a real issue with current approach. once we'll want to introduce shutdown we'll need to pass a context down and combine it with this.

network/p2p/p2p_sync.go
```diff
@@ -300,3 +288,42 @@ func (n *p2pNetwork) peersWithProtocolsFilter(protocols ...libp2p_protocol.ID) f
...
+	for {
+		peers, err := getSubsetOfPeers(logger, vpk, maxPeers, filter)
+		if err != nil {
+			return nil, err
+		}
+		if len(peers) >= minPeers || minPeers == 0 {
```
y0sher, line 317: I don't get why we need to `minPeers == 0` condition if it is never used.

network/p2p/p2p_sync.go
moshe-blox: this func doesn't accept context (like all other funcs in this struct)

what needs to be done is a huge refactor to every blocking func/method in SSV to accept context and pass it down correctly

though, for graceful shutdown the only important piece is the database (imo), so we can make that refactor a bit smaller

network/p2p/p2p_sync.go
moshe-blox: leftover from another condition — deleted

y0sher (approved):

lior-blox (approved):

PR #1059, opened 2023-07-13
feat: (api) provide status, index and owner address at `/v1/validators`

moshe-blox:
- Respond with the validator's status, index and Eth1 owner address in `/v1/validators`
- Exporter nodes now fetch metadata for all validators so they can provide their statuses & indices

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -699,6 +699,7 @@ func (c *controller) UpdateValidatorMetaDataLoop(logger *zap.Logger) {
 
 		filters := []registrystorage.SharesFilter{registrystorage.ByNotLiquidated()}
 		if !c.validatorOptions.Exporter {
+			// If we're not an exporter node, fetch only for validators of our operator.
 			filters = append(filters, registrystorage.ByOperatorID(c.operatorData.ID))
 		}
 		shares := c.sharesStorage.List(filters...)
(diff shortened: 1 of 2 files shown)


operator/validator/controller.go
```diff
@@ -687,7 +697,12 @@ func (c *controller) UpdateValidatorMetaDataLoop(logger *zap.Logger) {
 	for {
 		time.Sleep(c.metadataUpdateInterval)
 
-		shares := c.sharesStorage.List(registrystorage.ByOperatorID(c.operatorData.ID), registrystorage.ByNotLiquidated())
+		filters := []registrystorage.SharesFilter{registrystorage.ByNotLiquidated()}
+		if !c.validatorOptions.Exporter {
+			// If we're not an exporter node, fetch only for validators of our operator.
+			filters = append(filters, registrystorage.ByOperatorID(c.operatorData.ID))
```
y0sher, line 703: we don't care anymore about `ByNotLiquidated`?

operator/validator/controller.go
y0sher, line 1: I would've separated this to a different PR. this is not api code.

moshe-blox: i just see it as a consequence of this change: in order to provide this API as an exporter, you must fetch the data first

that said, if it was a big change, i'd also want it as a different PR

y0sher (approved):

lior-blox (approved):

PR #1061, opened 2023-07-14
Eth1 refactor more tests

pavelkrolevets:
- add more tests for executor

### DIFF
--- a/eth/eventdatahandler/event_data_handler_test.go
+++ b/eth/eventdatahandler/event_data_handler_test.go
@@ -111,7 +111,7 @@ func setupDataHandler(t *testing.T, ctx context.Context, logger *zap.Logger) (*E
 		Context:         ctx,
 		DB:              db,
 		RegistryStorage: nodeStorage,
-		KeyManager: keyManager,
+		KeyManager:      keyManager,
 	})
 
 	cl := executionclient.New("test", ethcommon.Address{})
--- a/eth/eventdatahandler/event_data_handler_test.go
+++ b/eth/eventdatahandler/event_data_handler_test.go
@@ -111,6 +111,7 @@ func setupDataHandler(t *testing.T, ctx context.Context, logger *zap.Logger) (*E
 		Context:         ctx,
 		DB:              db,
 		RegistryStorage: nodeStorage,
+		KeyManager: keyManager,
 	})
 
 	cl := executionclient.New("test", ethcommon.Address{})
--- a/eth/eventdatahandler/task_executor_test.go
+++ b/eth/eventdatahandler/task_executor_test.go
@@ -25,135 +25,89 @@ const rawValidatorAdded = `{
 		"transactionHash": "0x921a3f836fb873a40aa4f83097e52b69225334c49674dc262b2bb90d27e3a801"
 	  }`
 
-func TestExecuteStartValidatorTask(t *testing.T) {
-	logger, logs := setupLogsCapture()
+func TestExecuteTask(t *testing.T) {
+	logger, observedLogs := setupLogsCapture()
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-
 	edh, err := setupDataHandler(t, ctx, logger)
 	require.NoError(t, err)
-
-	logValidatorAdded := unmarshalLog(t, rawValidatorAdded)
-	validatorAddedEvent, err := edh.filterer.ParseValidatorAdded(logValidatorAdded)
-	if err != nil {
-		t.Fatal("parse ValidatorAdded", err)
-	}
-
-	share := &ssvtypes.SSVShare{
-		Share: spectypes.Share{
-			ValidatorPubKey: validatorAddedEvent.PublicKey,
-		},
-	}
-	task := NewStartValidatorTask(edh.taskExecutor, share)
-	require.NoError(t, task.Execute())
-	if logs.Len() == 0 {
-		t.Errorf("No logs")
-	} else {
-		entry := logs.All()[len(logs.All())-1]
+	t.Run("test AddValidator task execution", func(t *testing.T) {
+		logValidatorAdded := unmarshalLog(t, rawValidatorAdded)
+		validatorAddedEvent, err := edh.filterer.ParseValidatorAdded(logValidatorAdded)
+		if err != nil {
+			t.Fatal("parse ValidatorAdded", err)
+		}
+		share := &ssvtypes.SSVShare{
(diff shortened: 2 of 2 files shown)


eth/eventdatahandler/event_data_handler_test.go
```diff
@@ -111,6 +111,7 @@ func setupDataHandler(t *testing.T, ctx context.Context, logger *zap.Logger) (*E
 		Context:         ctx,
 		DB:              db,
 		RegistryStorage: nodeStorage,
+		KeyManager: keyManager,
```
nkryuchkov, line 114: not formatted

eth/eventdatahandler/task_executor_test.go
```diff
@@ -5,12 +5,13 @@ import (
 	"testing"
 
 	spectypes "github.com/bloxapp/ssv-spec/types"
-	ethtypes "github.com/ethereum/go-ethereum/core/types"
-	"github.com/stretchr/testify/require"
-	"go.uber.org/zap/zaptest"
-
 	"github.com/bloxapp/ssv/eth/eventbatcher"
 	ssvtypes "github.com/bloxapp/ssv/protocol/v2/types"
+	ethcommon "github.com/ethereum/go-ethereum/common"
```
nkryuchkov, line 10: Let's agree on a import sorting rule to avoid import reformatting after each other. I'd suggest having 3 blocks:

1) standard library
2) 3rd party dependencies
3) project dependencies

However, I'm not sure how to setup it this way in VSCode. Ignore this comment if it's complex to set up

eth/eventdatahandler/task_executor_test.go
```diff
@@ -45,10 +46,99 @@ func TestExecuteTask(t *testing.T) {
 	}
 	task := NewStartValidatorTask(edh.taskExecutor, share)
 	require.NoError(t, task.Execute())
+	if logs.Len() == 0 {
```
nkryuchkov, line 49: require.NotZero would be shorter and would eliminate else branch

eth/eventdatahandler/task_executor_test.go
```diff
@@ -24,8 +25,8 @@ const rawValidatorAdded = `{
 		"transactionHash": "0x921a3f836fb873a40aa4f83097e52b69225334c49674dc262b2bb90d27e3a801"
 	  }`
 
-func TestExecuteTask(t *testing.T) {
-	logger := zaptest.NewLogger(t)
+func TestExecuteStartValidatorTask(t *testing.T) {
```
nkryuchkov, line 28: What do you think about having one test for all tasks with sub-tests for each of them?

eth/eventdatahandler/task_executor_test.go
```diff
@@ -45,10 +46,99 @@ func TestExecuteTask(t *testing.T) {
 	}
 	task := NewStartValidatorTask(edh.taskExecutor, share)
 	require.NoError(t, task.Execute())
+	if logs.Len() == 0 {
+		t.Errorf("No logs")
+	} else {
+		entry := logs.All()[len(logs.All())-1]
+		require.Equal(t, "validator wasn't started", entry.Message)
```
nkryuchkov, line 53: Let's also think how to test the happy flow

eth/eventdatahandler/task_executor_test.go
```diff
@@ -24,8 +25,8 @@ const rawValidatorAdded = `{
 		"transactionHash": "0x921a3f836fb873a40aa4f83097e52b69225334c49674dc262b2bb90d27e3a801"
 	  }`
 
-func TestExecuteTask(t *testing.T) {
-	logger := zaptest.NewLogger(t)
+func TestExecuteStartValidatorTask(t *testing.T) {
+	logger, logs := setupLogsCapture()
```
nkryuchkov, line 29: what do you think of `observedLogs`?
```suggestion
	logger, observedLogs := setupLogsCapture()
```

eth/eventdatahandler/task_executor_test.go
```diff
@@ -45,10 +46,99 @@ func TestExecuteTask(t *testing.T) {
...
+}
+
+func TestExecuteUpdateFeeRecipientTask(t *testing.T) {
+	logger, logs := setupLogsCapture()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	edh, err := setupDataHandler(t, ctx, logger)
+	require.NoError(t, err)
+	task := NewFeeRecipientTask(edh.taskExecutor, ethcommon.HexToAddress("0x0000000000000000000000000000000000000001"), ethcommon.HexToAddress("0x0000000000000000000000000000000000000002"))
```
nkryuchkov, line 130: Would `0x1`/`0x2` work? If it would, I think it could be a bit easier to read

eth/eventdatahandler/task_executor_test.go
pavelkrolevets: I prefer separated tests as it follows better KISS pattern, but yeah lets keep everything in the same style.

nkryuchkov: alright, let's keep them separated then

eth/eventdatahandler/task_executor_test.go
pavelkrolevets: Ethereum address is 20 bytes hash with hex 40 chars so I prefer to use full length, but yes it will work as well

eth/eventdatahandler/task_executor_test.go
pavelkrolevets: Not sure here what would be the happy flow - we need to set proper env vars probably to have an event with the valid sig

nkryuchkov: We need to add what's missing to make it start, e.g. share metadata

eth/eventdatahandler/task_executor_test.go
```diff
@@ -25,54 +26,113 @@ const rawValidatorAdded = `{
...
 		defer close(eventsCh)
 		for _, event := range events {
 			eventsCh <- event
 		}
 	}()
-
 	lastProcessedBlock, err := edh.HandleBlockEventsStream(eb.BatchEvents(eventsCh), true)
 	require.Equal(t, uint64(0x89EBFF), lastProcessedBlock)
 	require.NoError(t, err)
+	var oservedLogsFlow []string
```
nkryuchkov, line 121: observedLogsFlow

PR #1063, opened 2023-07-16
(feat) height to slots, stop instance after 2 epochs.

y0sher:


protocol/v2/qbft/instance/instance.go
```diff
@@ -22,7 +22,9 @@ type Instance struct {
 
 	processMsgF *spectypes.ThreadSafeF
 	startOnce   sync.Once
-	StartValue  []byte
+
+	forceStop  bool
```
nkryuchkov, line 26: Should we make it atomic?

y0sher: currently instance isn't thread safe. so there's no reason to protect `forceStop` only.

moshe-blox (approved):

lior-blox (dismissed):

nkryuchkov (dismissed):

scripts/differ/differ
nkryuchkov, line 1: @y0sher Binary is committed. Please remove

PR #1066, opened 2023-07-18
Message validation

nkryuchkov:


network/p2p/ATTESTER.md
```diff
@@ -0,0 +1,74 @@
...
+
+Node can react to a message in three ways:
+- `Accept` - message is valid, accepted and processed further.
+- `Ignore` - message is invalid, ignored and not processed further, violation is noted.
+- `Reject` - message is rejected, ignored and not processed further, peer is banned.
+
+## Attester assertions
+
+- Message validator maintains a state for each validator.
+- Validator attests only once per epoch.
```
lior-blox, line 13: lets add a comment need to check if upon reorg can be more than one duty per epoch

network/p2p/ATTESTER.md
```diff
@@ -0,0 +1,74 @@
...
+- Validator submits messages for slot N within slots [N, N+32).
+- Validator submits messages for slot N in within [0, 384) seconds after slot start.
+- Message round is in range [1, 12].
```
lior-blox, line 16: lets explain why 12

network/p2p/ATTESTER.md
```diff
@@ -0,0 +1,74 @@
+# Attester message validation
+
+## Message reactions
+
+Node can react to a message in three ways:
+- `Accept` - message is valid, accepted and processed further.
+- `Ignore` - message is invalid, ignored and not processed further, violation is noted.
+- `Reject` - message is rejected, ignored and not processed further, peer is banned.
+
+## Attester assertions
```
lior-blox, line 10: we need to add to each assertion, what is the action if it fails it

network/p2p/ATTESTER.md
nkryuchkov: added an explanation

network/p2p/ATTESTER.md
nkryuchkov: added

network/p2p/ATTESTER.md
nkryuchkov: @lior-blox Added, but please review because I'm not sure about correctness

network/p2p/ATTESTER.md
MatheusFranco99: What you mean as the necessity for a second attestation is like the following scenario?
- validator V attests for block B at slot S
- a fork happens in slot S-1 and B is not in the canonical chain
- V should make another attestation for the canonical block of slot S?

But wouldn't this have the risk of being considered a slashable attestation?
- (Slashable Offences section) https://ethos.dev/beacon-chain#:~:text=An%20attestation%20contains%20both%20an,chain%20in%20a%20single%20epoch
- https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#is_slashable_attestation_data

network/p2p/AGGREGATOR_RULES.md
```diff
@@ -0,0 +1,25 @@
...
+  - Slots [N+32; N+42) are ignored. Further ones are rejected.
+- Message round is in range [1, 12].
+  - Given quick round duration is 2 seconds, slow round duration is 120 seconds, last quick round is 8. 8 quick rounds take 16 seconds. As submission must be no later than 384 seconds, there are 368 seconds left for slow rounds. 368 / 120 = 3.0666, so there are 4 slow rounds if rounded up. Therefore, maximal possible round is 8 + 4 = 12.
+  - Violation is rejected.
+- If message slot is equal to current slot, validator submits up to f+6 messages for each slot-round pair:
+  - 1 pre-consensus
+  - 1 proposal
+  - 1 prepare
+  - 1 commit
+  - f+1 aggregated commit/decided
```
MatheusFranco99, line 17: Quick question here that serves for every duty. Do we enforce that N == 3f+1? If not, then we assume that N>=3f+1 and the number of commits could be N - 2f, with f = floor((N-1)/3).

nkryuchkov: @MatheusFranco99 Do you mean committee size by `N`? If so, then we should enforce it

MatheusFranco99: Yes, I meant N = committee size. Oh ok, thanks for clarifying it

GalRogozinski: In my opinion, even though we currently enforce N, It will be nice if possible to make the implementation more robust.
This is because I personally I don't know why can't $N > 3f + 1$ and maybe someday we change this?

nkryuchkov: @GalRogozinski I don't mind

@lior-blox @moshe-blox your thoughts?

GalRogozinski: Maybe a reason why we enforce it actually is that if we make $N$ larger there are larger chances that $F$ will become unintentionally larger as well in practice?

GalRogozinski: I think you may actually lose security in practice. Because the value of f is also a function of N in real life.
You pick a random committee node that has some (small) probability of being malicious.

If it isn't malicious you have no security gain. If it is then that's not good news.

network/p2p/GENERAL_RULES.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Semantic assertions
+
+- Message validator maintains a state for each validator.
+- Generally, if violation happens by a small margin, the message is ignored. 
+  - If an assertion has a rule for ignoring, the rule has higher priority than this one.
+- Generally, If violation happens by a large margin, the message is rejected. 
+  - If an assertion has a rule for rejecting, the rule has higher priority than this one.
+- If violation count reaches a certain threshold, all further messages from the validator for current round are rejected.
+- Each round resets violation count to 0.
```
nkryuchkov, line 18: Handle case if node is a bit malicious (ignored) each round but does it repetitively: should it be punished?

network/p2p/GENERAL_RULES.md
```diff
@@ -0,0 +1,69 @@
...
+- Validator attests only once per epoch.
+  - TODO: Need to check if upon reorg there can be more than one duty per epoch.
+  - First violation is ignored. Further ones are rejected.
+- Message round is equal to estimated current round.
+- Violation by [1; ceil(MAX_POSSIBLE_ROUND * 0.2)] rounds is ignored. Violations above that are rejected.
```
nkryuchkov, line 23: - future rounds should be punished
- earlier ones are to be discussed
- test if clock drift may cause any false positives (if so, consider checking Prysm code regarding clock drift)

cli/operator/node.go
```diff
@@ -112,7 +114,8 @@ var StartNodeCmd = &cobra.Command{
 		cfg.P2pNetworkConfig.Permissioned = permissioned
 		cfg.P2pNetworkConfig.WhitelistedOperatorKeys = append(cfg.P2pNetworkConfig.WhitelistedOperatorKeys, networkConfig.WhitelistedOperatorKeys...)
 
-		p2pNetwork := setupP2P(forkVersion, operatorData, db, logger, networkConfig)
+		messageValidator := validation.NewMessageValidator(networkConfig.Beacon)
```
lior-blox, line 117: @nkryuchkov maybe better this to be setup inside the setupP2p and not here?

nkryuchkov: @lior-blox that may be temporary, `NewMessageValidator` will receive more args and it would be easier to do it here. I'll leave a TODO to consider moving it to `setupP2P` depending on amount of args

message/validation/message_counts.go
```diff
@@ -0,0 +1,56 @@
+package validation
```
lior-blox, line 1: add description

message/validation/signer_state.go
```diff
@@ -0,0 +1,22 @@
+package validation
```
lior-blox, line 1: add description

message/validation/validator.go
```diff
@@ -0,0 +1,346 @@
+package validation
```
lior-blox, line 1: add description

message/validation/message_counts.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/bloxapp/ssv/protocol/v2/ssv/queue"
+)
+
+// maxMessageCounts is the maximum number of acceptable messages from a signer within a slot & round.
+func maxMessageCounts(committeeSize int) MessageCounts {
+	return MessageCounts{
+		PreConsensus:  1,
+		Proposals:     1,
+		Prepares:      1,
+		Commits:       committeeSize + 1,
```
lior-blox, line 16: @nkryuchkov need to add the decide messages. 
based on @MatheusFranco99 calculation

lior-blox, line 16: commits messages:
signer commit is only 1
and then the rest are decided

message/validation/message_counts.go
```diff
@@ -0,0 +1,56 @@
...
+		switch m.Message.MsgType {
+		case specqbft.ProposalMsgType:
+			c.Proposals++
+		case specqbft.PrepareMsgType:
+			c.Prepares++
+		case specqbft.CommitMsgType:
+			c.Commits++
+		}
+	case *spectypes.SignedPartialSignatureMessage:
+		if m.Message.Type == spectypes.PostConsensusPartialSig {
```
lior-blox, line 41: @nkryuchkov should be switch case

message/validation/validator.go
```diff
@@ -0,0 +1,346 @@
...
+	maxPartialSignatureMsgSize = math.MaxUint64 // TODO: define
+)
+
+var (
+	ErrUnknownValidator = errors.New("unknown validator")
+	ErrInvalidRole      = errors.New("invalid role")
+	ErrEarlyMessage     = errors.New("early message")
+	ErrLateMessage      = errors.New("late message")
+	ErrInvalidSigners   = errors.New("invalid signers")
+)
```
lior-blox, line 39: @nkryuchkov maybe better to move it to some other file

message/validation/validator.go
```diff
@@ -0,0 +1,346 @@
...
+
+	// Validate message counts within the current round.
+	if !signerState.MessageCounts.Record(msg, maxMessageCounts(len(signedMsg.Signers))) {
+		return errors.New("too many messages")
+	}
+
+	return nil
+}
+
+func (mv *MessageValidator) knownValidator(pubKey []byte) bool {
```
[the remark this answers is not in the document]
lior-blox: also we need to have a function in our storage for that IMO.
@moshe-blox WDYT?

message/validation/validator.go
```diff
@@ -0,0 +1,346 @@
...
+func NewMessageValidator(network beaconprotocol.Network, ctrl validator.Controller) *MessageValidator {
+	return &MessageValidator{
+		network: network,
+		index:   make(map[ConsensusID]*ConsensusState),
+		ctrl:    ctrl,
+	}
+}
+
+func (mv *MessageValidator) ValidateMessage(ssvMessage *spectypes.SSVMessage) (*queue.DecodedSSVMessage, error) {
+	if err := mv.validateSSVMessage(ssvMessage); err != nil {
```
lior-blox, line 66: first check its in the current domain

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+	// lateMessageMargin is the duration past a message's TTL in which it is still considered valid.
+	lateMessageMargin = time.Second * 3
+
+	// clockErrorTolerance is the maximum amount of clock error we expect to see between nodes.
+	clockErrorTolerance = time.Millisecond * 50
+
+	maxMessageSize             = 8388608        // TODO: define
+	maxConsensusMsgSize        = maxMessageSize // TODO: define
+	maxPartialSignatureMsgSize = maxMessageSize // TODO: define
+	allowedRoundsInFuture      = 2
```
MatheusFranco99, line 35: I think SignedPartialSignatureMessage max size is in bytes:
8 (OperatorID) +
96 (Signature) +
8 (PartialSignatureMessages.Type) +
8 (PartialSignatureMessages.Slot) + 
13 * (
    8 (PartialSignatureMessage.Signer) +
    96 (PartialSignatureMessage.PartialSignature) +
    32 (PartialSignatureMessage.SigningRoot)
    )
= 1888

SSVMessage with Data as SignedPartialSignatureMessage would be:
8 + 56 + 1888 = 1952

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+
+const (
+	// lateMessageMargin is the duration past a message's TTL in which it is still considered valid.
+	lateMessageMargin = time.Second * 3
+
+	// clockErrorTolerance is the maximum amount of clock error we expect to see between nodes.
+	clockErrorTolerance = time.Millisecond * 50
+
+	maxMessageSize             = 8388608        // TODO: define
+	maxConsensusMsgSize        = maxMessageSize // TODO: define
```
MatheusFranco99, line 33: Upper bound for the SignedMessage (consensus msg) is 2^23 bytes (8388608 bytes)

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+	if !bytes.Equal(ssvMessage.MsgID.GetDomain(), mv.netCfg.Domain[:]) {
+		return nil, fmt.Errorf("wrong domain, got %v, expected %v", hex.EncodeToString(ssvMessage.MsgID.GetDomain()), hex.EncodeToString(mv.netCfg.Domain[:]))
+	}
+
+	if !mv.validRole(ssvMessage.MsgID.GetRoleType()) {
+		return nil, ErrInvalidRole
+	}
+
+	share := mv.shareStorage.Get(ssvMessage.MsgID.GetPubKey())
+	if share != nil {
```
MatheusFranco99, line 88: Shouldn't be "if share == nil" ?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+		return nil, ErrUnknownValidator // ERR_VALIDATOR_ID_MISMATCH
+	}
+
+	if !share.BeaconMetadata.IsAttesting() {
+		return nil, fmt.Errorf("validator is not active")
+	}
```
MatheusFranco99, line 94: Does this encompass the case for a liquidated validator? I mean, if the validator is liquidated, will this raise the error?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+
+	return nil
+}
+func (mv *MessageValidator) validateEventMessage(msg *queue.DecodedSSVMessage) error {
+	signedMsg, ok := msg.Body.(*ssvtypes.EventMsg)
+	if !ok {
+		return fmt.Errorf("expected event message")
+	}
+
+	_ = signedMsg // TODO: validate it
```
MatheusFranco99, line 210: Since no Event messages are expected from the network, may we already raise an error here?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+	default:
+		panic("unknown role")
+	}
+}
+
+func (mv *MessageValidator) validSigners(share *ssvtypes.SSVShare, msg *queue.DecodedSSVMessage) error {
+	// TODO: consider having different validation functions for each message type
+	if !share.BeaconMetadata.IsAttesting() {
+		return fmt.Errorf("validator not active")
+	}
```
MatheusFranco99, line 392: Wasn't this already checked in ValidateMessage function?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+		if len(m.Signers) == 0 {
+			return ErrNoSigners
+		}
+
+		if len(m.Signers) == 1 {
+			qbftState := &specqbft.State{
+				Height: m.Message.Height,
+				Share:  &share.Share,
+			}
+			leader := specqbft.RoundRobinProposer(qbftState, m.Message.Round)
```
MatheusFranco99, line 411: I think this works but, even though there's a RoundRobinProposer in spec, in [proposal.go](https://github.com/bloxapp/ssv/blob/main/protocol/v2/qbft/instance/proposal.go#L214) a _proposer_ function is defined. If it changes, we might be in trouble. Would be better, if possible, to avoid this duplicated code. I don't know how. What do you think?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+			}
+		} else if m.Message.MsgType != specqbft.CommitMsgType {
+			return fmt.Errorf("non-decided with multiple signers, len: %d", len(m.Signers))
+		}
+
+		if !slices.IsSorted(m.Signers) {
+			return fmt.Errorf("signers not sorted") // ERR_SIGNERS_NOT_SORTED
+		}
+
+		// TODO: if decided, check if previous decided had less or equal signers
```
MatheusFranco99, line 423: Also if len >= quorum and len <= committee size.

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	hashedFullData, err := specqbft.HashDataRoot(signedMsg.FullData)
+	if err != nil {
+		return fmt.Errorf("hash data root: %w", err)
+	}
+
+	if hashedFullData != signedMsg.Message.Root {
+		return fmt.Errorf("root doesn't match full data hash")
+	}
```
MatheusFranco99, line 325: Prepares and Commits don't use the FullData field. Is this checked just for proposal?

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+
+	if len(signedMsg.Signers) <= signerState.LastDecidedQuorumSize {
+		return fmt.Errorf("decided must have more signers than previous decided")
+	}
```
MatheusFranco99, line 329: This should be checked only for Decideds, right?

message/validation/message_counts.go
```diff
@@ -0,0 +1,121 @@
...
+	case *spectypes.SignedPartialSignatureMessage:
+		switch m.Message.Type {
+		case spectypes.RandaoPartialSig, spectypes.SelectionProofPartialSig, spectypes.ContributionProofs, spectypes.ValidatorRegistrationPartialSig:
+			if c.PreConsensus > 0 {
+				return fmt.Errorf("pre-consensus is not expected")
+			}
+		case spectypes.PostConsensusPartialSig:
+			if c.PostConsensus > 0 { // TODO: do we need this check?
+				return fmt.Errorf("post-consensus is not expected")
+			}
```
MatheusFranco99, line 67: I guess we need, to prevent against an attacker that keeps sending duplicated post-consensus.

message/validation/validator.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+func (mv *MessageValidator) lateMessage(slot phase0.Slot, role spectypes.BeaconRole) bool {
+	// Note: this is just an example, should be according to ssv-spec.
+	var ttl phase0.Slot
+	switch role {
+	case spectypes.BNRoleProposer, spectypes.BNRoleSyncCommittee, spectypes.BNRoleSyncCommitteeContribution:
+		ttl = 1 + 1
+	case spectypes.BNRoleAttester, spectypes.BNRoleAggregator:
+		ttl = 32 + 2
```
MatheusFranco99, line 227: Both should be + 2 (so 1 + 2 and 32 + 2). My spec was confusing and wrong here, I'll update there. Sorry for that.
The +2 is a common slot spread that Prysm suggested, so it can even be a const variable (config, global, as you want and if you want).

message/validation/validator.go
nkryuchkov: yes, `IsAttesting` means it's either `ValidatorStateActiveOngoing` or `ValidatorStateActiveExiting`

message/validation/validator.go
nkryuchkov: `proposer` depends on `specqbft.State` and it's initialized in the validator controller, so I think it's too complex to use

MatheusFranco99: Indeed it is.

message/validation/validator.go
```diff
@@ -0,0 +1,510 @@
...
+				Height: m.Message.Height,
+				Share:  &share.Share,
+			}
+			leader := specqbft.RoundRobinProposer(qbftState, m.Message.Round)
+			if m.Signers[0] != leader {
+				err := ErrSignerNotLeader
+				err.got = m.Signers[0]
+				err.want = leader
+				return err
+			}
```
MatheusFranco99, line 455: We should do this verification only for ProposalMsgType.

message/validation/partial_validation.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	// TODO: check running duty
+
+	if err := mv.validPartialSignatures(share, signedMsg); err != nil {
+		return err
+	}
+
+	role := msg.GetID().GetRoleType()
```
MatheusFranco99, line 37: Maybe, verify here that the role is valid.

message/validation/validation.go
```diff
@@ -0,0 +1,262 @@
...
+		// this should be caught by the earlyMessage check.
+		err := ErrSlotAlreadyAdvanced
+		err.want = signerState.Slot
+		err.got = msgSlot
+		return err
+	}
+
+	// Advance slot & round, if needed.
+	if signerState.Slot < msgSlot {
+		signerState.Reset(msgSlot, 0) // TODO: 0 or specqbft.FirstRound?
```
MatheusFranco99, line 173: I think specqbft.FirstRound

message/validation/partial_validation.go
MatheusFranco99, line 37: We should also verify that the PartialSigMsgType is consistent with the role. This is duty specific rules, so:
- Attestation allows PostConsensusPartialSig
- Attestation Egg allows SelectionProofPartialSig and PostConsensusPartialSig
- Proposal allows RandaoPartialSig and PostConsensusPartialSig
- Sync Committee allows PostConsensusPartialSig
- Sync Committee Aggregation allows ContributionProofs and PostConsensusPartialSig
- ValidatorRegistration allows ValidatorRegistrationPartialSig

message/validation/consensus_validation.go
```diff
@@ -0,0 +1,361 @@
...
+	for _, signer := range m.Signers {
+		if err := mv.commonSignerValidation(signer, share); err != nil {
+			return err
+		}
+
+		if _, ok := seen[signer]; ok {
+			return ErrDuplicatedSigner
+		}
+		seen[signer] = struct{}{}
+	}
```
moshe-blox, line 359: since the slice is sorted, we can make this cheaper check without a map allocation:

```suggestion
	var prevSigner spectypes.OperatorID
	for _, signer := range m.Signers {
		if err := mv.commonSignerValidation(signer, share); err != nil {
			return err
		}
		if signer == prevSigner {
			return ErrDuplicatedSigner
		}
		prevSigner = signer
	}
```
(signer is already verified to not be 0 so don't need to check for it)

message/validation/consensus_validation.go
```diff
@@ -0,0 +1,361 @@
...
+func (mv *MessageValidator) waitAfterSlotStart(role spectypes.BeaconRole) time.Duration {
+	switch role {
+	case spectypes.BNRoleSyncCommitteeContribution:
+		return mv.netCfg.Beacon.SlotDurationSec() // TODO: do we need this?
+	case spectypes.BNRoleAttester, spectypes.BNRoleAggregator, spectypes.BNRoleSyncCommittee, spectypes.BNRoleProposer, spectypes.BNRoleValidatorRegistration:
+		return 0
+	default:
+		panic("unknown role")
+	}
+}
```
moshe-blox, line 300: if this won't be used outside of tests, i think it's better to move this to the test file (as a func that requires `netCfg` for example)

message/validation/consensus_validation.go
```diff
@@ -0,0 +1,361 @@
...
+
+func (mv *MessageValidator) currentEstimatedRound(sinceSlotStart time.Duration) specqbft.Round {
+	if currentQuickRound := specqbft.FirstRound + specqbft.Round(sinceSlotStart/roundtimer.QuickTimeout); currentQuickRound <= roundtimer.QuickTimeoutThreshold {
+		return currentQuickRound
+	}
+
+	sinceFirstSlowRound := sinceSlotStart - (time.Duration(roundtimer.QuickTimeoutThreshold) * roundtimer.QuickTimeout)
+	estimatedRound := roundtimer.QuickTimeoutThreshold + specqbft.FirstRound + specqbft.Round(sinceFirstSlowRound/roundtimer.SlowTimeout)
+	return estimatedRound
+}
```
moshe-blox, line 289: same as `waitAfterSlotStart`, unless you think it will be used outside of test?

message/validation/event_validation.go
```diff
@@ -0,0 +1,19 @@
...
+// event_validation.go contains methods for validating event messages
+
+func (mv *MessageValidator) validateEventMessage(msg *queue.DecodedSSVMessage) error {
+	_, ok := msg.Body.(*ssvtypes.EventMsg)
+	if !ok {
+		return fmt.Errorf("expected event message")
+	}
+
+	return ErrEventMessage
+}
```
moshe-blox, line 19: events are internal messages (such as round timeout or execute duty) and aren't supposed to be sent by peers
so i think we can remove this and just return an error when receiving an event message as early as possible in the validation pipeline

message/validation/partial_validation.go
```diff
@@ -0,0 +1,190 @@
...
+			return fmt.Errorf("deserialize pk: %w", err)
+		}
+		sig := &bls.Sign{}
+		if err := sig.Deserialize(signature); err != nil {
+			return fmt.Errorf("deserialize signature: %w", err)
+		}
+
+		if !sig.VerifyByte(&pk, root[:]) {
+			return fmt.Errorf("wrong signature")
+		}
```
moshe-blox, line 165: i think this can use the `BatchVerifier.AggregateVerify` method as well to increase performance (only difference is here we pass only one public key, if im not wrong)

message/validation/consensus_validation.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	state := mv.consensusState(consensusID)
+
+	if !nonCommittee {
+		for _, signer := range signedMsg.Signers {
+			if err := mv.validateSignerBehavior(state, signer, share, msg); err != nil {
+				return fmt.Errorf("bad signer behavior: %w", err)
+			}
+		}
+	}
```
moshe-blox, line 109: why do we skip this check (and the one below) if the validator is non-committee? did i miss it in the [knowledge-base](https://github.com/bloxapp/knowledge-base)?

moshe-blox (changes requested):
great progress ✌️

left a few comments and questions

message/validation/validation.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+}
+
+func (mv *MessageValidator) SetValidatorGetter(f validatorGetterFunc) {
+	mv.validatorGetter = f
+}
+
+func (mv *MessageValidator) SetNonCommitteeValidatorGetter(f nonCommitteeValidatorGetterFunc) {
+	mv.nonCommitteeValidatorGetter = f
+}
```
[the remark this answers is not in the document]
moshe-blox: update: i'm not sure why we the validation needs to distinguish between committee/non-committee (https://github.com/bloxapp/ssv/pull/1066#discussion_r1306640643)

message/validation/consensus_validation.go
nkryuchkov: It was unused because I disabled the estimated round check, but now it's used again

message/validation/consensus_validation.go
nkryuchkov: same as waitAfterSlotStart

moshe-blox:
Message validation queue metrics are not yet implemented.

I've found an example by Prysm of how to implement these by hooking to pubsub's trace: https://github.com/libp2p/go-libp2p-pubsub/issues/534#issuecomment-1612478109

message/validation/consensus_validation.go
```diff
@@ -0,0 +1,428 @@
...
+func (mv *MessageValidator) validateSignerBehavior(
+	state *ConsensusState,
+	signer spectypes.OperatorID,
+	share *ssvtypes.SSVShare,
+	msg *queue.DecodedSSVMessage,
+) error {
+	signedMsg, ok := msg.Body.(*specqbft.SignedMessage)
+	if !ok {
+		panic("validateSignerBehavior should be called on signed message")
+	}
```
moshe-blox, line 202: why not requiring this type in the func parameter instead?

nkryuchkov: @moshe-blox Because I need some data from SSVMessage. Although I see that this data can be passed, so I'll try to get rid of it

nkryuchkov: @moshe-blox Fixed in https://github.com/bloxapp/ssv/pull/1066/commits/ee711ef998274a049d711366c9f0cd49fb49ee40

nkryuchkov:
Spec alignment will be checked and fixed by @moshe-blox

integration/qbft/tests/msg_router.go
```diff
@@ -1,17 +1,18 @@
...
 
 type msgRouter struct {
 	validator *protocolvalidator.Validator
 }
 
-func (m *msgRouter) Route(logger *zap.Logger, message spectypes.SSVMessage) {
-	m.validator.HandleMessage(logger, &message)
+func (m *msgRouter) Route(logger *zap.Logger, message *queue.DecodedSSVMessage) {
+	m.validator.HandleMessage(logger, message)
 }
```
y0sher, line 16: I thought we are not even running these tests, also this seems like code duplication? we have router in p2p. maybe I'm missing something.

nkryuchkov: We run integration tests in `.github/workflows/integration-test.yml`. Or did you mean something else?

MatheusFranco99 (approved):
Approved since spec alignment will be fixed later

MatheusFranco99 (approved):

message/validation/errors.go
```diff
@@ -0,0 +1,100 @@
+package validation
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Error struct {
```
y0sher, line 8: error usually contains the text inside, not sure why you need both innerErr and text. you also just use composition and put error inside.
I understand how you use `Error()` here in tests to print stuff pretty, but this could be achieved by a test helper 

func ErrorMessage(e Error) string {
	var sb strings.Builder
	sb.WriteString(e.Error())

	if e.got != nil {
		sb.WriteString(fmt.Sprintf(", got %v", e.got))
	}
	if e.want != nil {
		sb.WriteString(fmt.Sprintf(", want %v", e.want))
	}
        // this is not needed since error will always be assigned as its the new `text`
	//if e.innerErr != nil {
	//       sb.WriteString(fmt.Sprintf(": %s", e.innerErr.Error()))
	//}

	return sb.String()
}

nkryuchkov: Also given we put hex encoded signature/PK into the error (I replaced them with `...` in the example above), we'd have a separate error shown in Grafana for each signature/PK, so we can easily have hundreds of them, which would make the dashboard useless

message/validation/partial_validation.go
```diff
@@ -0,0 +1,253 @@
...
+		return msgSlot, ErrPartialSignatureTypeRoleMismatch
+	}
+
+	if err := mv.validatePartialMessages(share, signedMsg); err != nil {
+		return msgSlot, err
+	}
+
+	consensusState := mv.consensusState(msgID)
+
+	state := mv.consensusState(msgID)
```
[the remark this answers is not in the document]
nkryuchkov: Thanks, pushing a fix

network/syncing/syncer.go
```diff
@@ -92,11 +96,32 @@ func (s *syncer) SyncHighestDecided(
...
+
+		ssvMessage := &spectypes.SSVMessage{
 			MsgType: spectypes.SSVConsensusMsgType,
 			MsgID:   id,
 			Data:    raw,
-		})
+		}
+
+		// TODO: consider uncommenting
+		if s.msgValidator != nil {
```
y0sher, line 107: when is `msgValidator` nil?

operator/duties/attester.go
y0sher, line 1: I would ask  @olegshmuelov to review any duties related code.

network/syncing/syncer.go
nkryuchkov: It seems like now it's never nil, removing this check

operator/validator/router.go
```diff
@@ -1,34 +1,36 @@
...
 type messageRouter struct {
-	ch    chan spectypes.SSVMessage
-	msgID commons.MsgIDFunc
+	logger *zap.Logger
+	ch     chan *queue.DecodedSSVMessage
+	msgID  commons.MsgIDFunc
 }
 
-func (r *messageRouter) Route(logger *zap.Logger, message spectypes.SSVMessage) {
+func (r *messageRouter) Route(logger *zap.Logger, message *queue.DecodedSSVMessage) {
```
y0sher, line 26: logger is now not used, should we take it out? I know its a pain in the a** to do so though. also we might want to pass in context instead, this way we can cancel the send and we can attach some data to it in the future.

protocol/v2/blockchain/beacon/network.go
```diff
@@ -82,6 +83,11 @@ func (n Network) GetSlotStartTime(slot phase0.Slot) time.Time {
 	return start
 }
 
+// GetSlotEndTime returns the end time for the given slot
+func (n Network) GetSlotEndTime(slot phase0.Slot) time.Time {
+	return n.GetSlotStartTime(slot + 1) // TODO: check if it's correct
```
y0sher, line 88: did we check?

nkryuchkov: Added a test

operator/validator/router.go
nkryuchkov: Done

moshe-blox (approved):

MatheusFranco99 (dismissed):

protocol/v2/types/crypto.go
```diff
@@ -47,6 +55,43 @@ func VerifyByOperators(s spectypes.Signature, data spectypes.MessageSignature, d
 		return errors.Wrap(err, "could not compute signing root")
 	}
 
+	// verify
+	// if res := sign.FastAggregateVerify(pks, computedRoot[:]); !res {
+	// 	return errors.New("failed to verify signature")
+	// }
```
y0sher, line 53: is there a reason to leave that comment?

nkryuchkov: It could be useful but we decided to disable batch verifier so I'm deleting it

protocol/v2/types/crypto.go
```diff
@@ -82,3 +127,469 @@ func VerifyReconstructedSignature(sig *bls.Sign, validatorPubKey []byte, root [3
 func rootHex(r [32]byte) string {
 	return hex.EncodeToString(r[:])
 }
+
+var Verifier = NewBatchVerifier(runtime.NumCPU(), 10, time.Millisecond*5)
+
+func init() {
+	go Verifier.Start()
+}
+
```
y0sher: I'm not a fan of this approach, it makes testing (especially integration) and replacing harder, I would want to see this being created at the top level and passed down to components that need this. also there's nothing terminating the goroutine. 
this could be taken out to other task not as part of this pr. but should be done IMO.

nkryuchkov: I totally agree and I don't like it either. However, I took it as-is from Moshe's branch and didn't touch it because I was not sure that it will be kept because if verification performance issues. Now we decided to disable batch verification, so I deleted this.

protocol/v2/types/batch_verifier.go
```diff
@@ -0,0 +1,475 @@
...
+				zap.Int("failed_batches", stats.FailedBatches),
+				zap.Int("failed_requests", stats.FailedRequests),
+				zap.String("request_total_durations", formatGauge(stats.RequestTotalDurations)),
+				zap.String("request_pending_durations", formatGauge(stats.RequestPendingDurations)),
+				zap.String("batch_durations", formatGauge(stats.BatchDurations)),
+				zap.String("batch_pending_durations", formatGauge(stats.BatchPendingDurations)),
+			)
+		}
+	}()
+
```
y0sher, line 293: I would move this code to a separate func, maybe add a todo about making this optional

nkryuchkov: Generally, I agree, but I think we don't need to touch the batch verifier anymore because we'll likely remove it. I'd rather delete the file but @moshe-blox wanted to keep it

protocol/v2/types/batch_verifier.go
```diff
@@ -0,0 +1,475 @@
+package types
```
y0sher, line 1: Generally its hard for me to understand the gain we get from this batchverifier without comparison to previous state. also a lot of locks and stuff here, I think we need to revisit this code and optimize.

nkryuchkov: @moshe-blox FYI: I'm going to delete the batch verifier because it doesn't pass the review. If we need it in the future we'll be able to copy it from this PR

nkryuchkov: @y0sher deleted

message/validation/validation.go
```diff
@@ -0,0 +1,559 @@
...
+	ssvtypes "github.com/bloxapp/ssv/protocol/v2/types"
+	registrystorage "github.com/bloxapp/ssv/registry/storage"
+)
+
+const (
+	// lateMessageMargin is the duration past a message's TTL in which it is still considered valid.
+	lateMessageMargin = time.Second * 3
+
+	// clockErrorTolerance is the maximum amount of clock error we expect to see between nodes.
+	clockErrorTolerance = time.Millisecond * 50
```
y0sher, line 41: sounds small skew considering we aren't doing any enforcing on clocks. should at least add to our docs that clocks need to be synced with some ext service like NTP.

message/validation/validation.go
```diff
@@ -0,0 +1,559 @@
...
+	Signers *hashmap.Map[spectypes.OperatorID, *SignerState]
+}
+
+// GetSignerState retrieves the state for the given signer.
+// Returns nil if the signer is not found.
+func (cs *ConsensusState) GetSignerState(signer spectypes.OperatorID) *SignerState {
+	signerState, ok := cs.Signers.Get(signer)
+	if !ok {
+		return nil
+	}
```
y0sher, line 71: this code works, but checking is meaningless since if `!ok` `singerState` will be `nil` anyway.

message/validation/validation.go
```diff
@@ -0,0 +1,559 @@
...
+	defer func() {
+		sinceStart := time.Since(start)
+		mv.metrics.MessageValidationDuration(sinceStart, validationDurationLabels...)
+	}()
+
+	decodedMessage, descriptor, err := mv.validateP2PMessage(pmsg, time.Now())
+	round := specqbft.Round(0)
+	if descriptor.Consensus != nil {
+		round = descriptor.Consensus.Round
+	}
```
y0sher, line 262: it's a bit weird that you are not checking the err first, I assume descriptor can have value even if the err is not nil, but this is also not a very common usage.

protocol/v2/types/batch_verifier.go
y0sher: oh I didn't know its out of use for now. so ignore all my comments on the batch verifier.

y0sher (approved):
Added just some minor comments. This is a lot of code, lets get it merged soon.
Generally I'm not a big fan of this approach, we're replicating the consensus state in a way and keeping the messages twice. the code is pretty well organized though and functions as a "separate entity" which is good because one of my concerns was bloating the p2p layer with consensus logic.
anyway I hope one day we would just relay on the protocol to validate its own messages since it should already have all tools in place. there are some extra stuff we validate by tracking the messages but IMO eventually it should be splitted from consensus logic.

message/validation/validation.go
nkryuchkov: I agree it looks sketchy but the purpose of the descriptor is for logging, so `round` is used in error handling

y0sher (approved):

MatheusFranco99 (dismissed):

moshe-blox (approved):

MatheusFranco99 (approved):

y0sher (approved):

PR #1069, opened 2023-07-23
fix: operator & validator pubkey logs in eth1-refactor

moshe-blox:
- log field was called "operator_pub_key" instead of "validator" in ValidatorAdded
- operator pubkeys were logged as hex instead of base64
- log wasn't acknowledging it encountered its own OperatorAdded or ValidatorAdded events

### DIFF
--- a/eth/eventdatahandler/handlers.go
+++ b/eth/eventdatahandler/handlers.go
@@ -42,7 +42,7 @@ var (
 func (edh *EventDataHandler) handleOperatorAdded(txn basedb.Txn, event *contract.ContractOperatorAdded) error {
 	logger := edh.logger.With(
 		fields.OperatorID(event.OperatorId),
-		zap.String("operator_pub_key", ethcommon.Bytes2Hex(event.PublicKey)),
+		fields.OperatorPubKey(event.PublicKey),
 		zap.String("owner_address", event.Owner.String()),
 		zap.String("event_type", OperatorAdded),
 	)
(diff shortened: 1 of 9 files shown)


eth/eventdatahandler/handlers.go
```diff
@@ -74,12 +74,10 @@ func (edh *EventDataHandler) handleOperatorAdded(txn basedb.Txn, event *contract
 
 	if bytes.Equal(event.PublicKey, edh.operatorData.PublicKey) {
 		edh.operatorData = od
+		logger = logger.With(zap.Bool("own_operator", true))
 	}
 
 	edh.metrics.OperatorPublicKey(od.ID, od.PublicKey)
-	logger.Debug("report operator public key",
```
lior-blox, line 80: @moshe-blox we dont want to show all the logs in debug?

eth/eventdatahandler/handlers.go
```diff
@@ -208,12 +206,12 @@ func (edh *EventDataHandler) handleValidatorAdded(txn basedb.Txn, event *contrac
 	isOperatorShare := validatorShare.BelongsToOperator(edh.operatorData.ID)
 	if isOperatorShare {
 		edh.metrics.ValidatorInactive(event.PublicKey)
-		logger.Debug("processed event")
-		return validatorShare, nil
+		ownShare = validatorShare
+		logger = logger.With(zap.Bool("own_validator", true))
```
lior-blox, line 210: @moshe-blox maybe committee or cluster instead of own?

operator/validator/task_executor.go
```diff
@@ -112,6 +104,5 @@ func (c *controller) UpdateFeeRecipient(owner, recipient common.Address) error {
 		return fmt.Errorf("update validators map: %w", err)
 	}
 
-	logger.Info("executed task")
```
moshe-blox, line 115: when you see the previous log of 'executing task' and no error log after it, it means the task was executed (so 'executed task' log is not necessary)

nkryuchkov: @moshe-blox then maybe better "executed task" and "failed to execute task: ..."? I'd keep "executing task" as debug though to debug a possible deadlock when executing a task

nkryuchkov (approved):

PR #1070, opened 2023-07-24
Small fixes

lior-blox:


nodeprobe/nodeprobe.go
```diff
@@ -114,15 +114,12 @@ func (p *Prober) probe(ctx context.Context) {
 	p.ready.Store(allNodesReady.Load())
 
 	if !p.ready.Load() {
-		p.logger.Debug("not all nodes are ready")
+		p.logger.Warn("not all nodes are ready")
```
nkryuchkov, line 117: Should this be ERROR? Logs above use error when not ready:

```go
			if err != nil {
				logger.Error("failed to check if node is ready", zap.Error(err))
			} else if !ready {
				logger.Error("node is not ready")
			} else {
				logger.Info("node is ready")
			}
```

nodeprobe/nodeprobe.go
```diff
@@ -114,15 +114,12 @@ func (p *Prober) probe(ctx context.Context) {
...
 		if h := p.unreadyHandler.Load(); h != nil {
 			(*h)()
 		}
 		return
 	}
-
-	p.logger.Info("all nodes are ready")
```
nkryuchkov, line 124: Why don't we need this log?

nodeprobe/nodeprobe.go
lior-blox: @nkryuchkov changed to error

nodeprobe/nodeprobe.go
lior-blox: We should refrain from documenting normal operations; instead, focus on logging occurrences of issues or anomalies, just as it will do if they are not ok.

for example, this log is written every minute and clutters the output

moshe-blox (approved):

PR #1071, opened 2023-07-24
update storage interfaces

pavelkrolevets:
Changes according to https://github.com/bloxapp/ssv/pull/1020#discussion_r1271399377

### DIFF
--- a/operator/storage/storage.go
+++ b/operator/storage/storage.go
@@ -27,7 +27,7 @@ var (
 type Storage interface {
 	// TODO: de-anonymize the sub-storages, like Shares() below
 
-	ROTxn() basedb.Txn
+	ROTxn() basedb.Reader
 	RWTxn() basedb.Txn
 
 	SaveLastProcessedBlock(txn basedb.Txn, offset *big.Int) error
@@ -46,7 +46,7 @@ type Storage interface {
 
 type storage struct {
 	logger *zap.Logger
-	db     basedb.IDb
+	db     basedb.Database
 
 	operatorStore  registrystorage.Operators
 	recipientStore registrystorage.Recipients
@@ -55,7 +55,7 @@ type storage struct {
 }
 
 // NewNodeStorage creates a new instance of Storage
-func NewNodeStorage(logger *zap.Logger, db basedb.IDb) (Storage, error) {
+func NewNodeStorage(logger *zap.Logger, db basedb.Database) (Storage, error) {
 	stg := &storage{
 		logger:         logger,
 		db:             db,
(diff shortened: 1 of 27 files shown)


operator/storage/storage.go
```diff
@@ -27,7 +27,7 @@ var (
 type Storage interface {
 	// TODO: de-anonymize the sub-storages, like Shares() below
 
-	ROTxn() basedb.Txn
+	ROTxn() basedb.Reader
```
nkryuchkov, line 30: Should `RWTxn` be `ReadWriter`?

storage/basedb/storage.go
```diff
@@ -14,37 +14,41 @@ type Options struct {
...
+	Commit() error
+	Discard()
+}
+
+// Database interface for Badger DB
+type Database interface {
+	RWTxn() Txn
+	ROTxn() Reader // TODO: afaik there is no effect for Commit/Discard on read-only transactions so a `Reader` is sufficient?
+	ReadWriter
+	// TODO: consider moving these functions into Reader and ReadWriter interfaces?
```
nkryuchkov, line 51: Comment seems outdated

storage/kv/txn.go
```diff
@@ -81,3 +81,19 @@ func (t badgerTxn) GetAll(prefix []byte, handler func(int, basedb.Obj) error) er
 func (t badgerTxn) Delete(prefix []byte, key []byte) error {
 	return t.txn.Delete(append(prefix, key...))
 }
+
+func (t badgerTxn) CountByCollection(prefix []byte) (int64, error) {
+	return t.db.CountByCollection(prefix)
```
nkryuchkov, line 86: I think this should be a part of the transaction (same for methods below)

pavelkrolevets: Do you mean on *badger.Txn? but it doesnt implement these methods. Can you please clarify a bit?

operator/storage/storage.go
nkryuchkov: I think Discard for both RO and RW, Commit only for RW @pavelkrolevets

storage/kv/txn.go
nkryuchkov: Yes, it doesn't, therefore @moshe-blox left them in the `basedb.IDb` interface (now `basedb.Database`). Can we remove them from `badgerTxn` and move these methods from transactions to `basedb.Database`?

pavelkrolevets: Moved back to Database interface

moshe-blox:
merged this draft and completed in #1020

PR #1072, opened 2023-07-25
fix: don't increment connections after ignoring one

moshe-blox:


network/peers/connections/conn_handler.go
```diff
@@ -71,13 +71,14 @@ func (ch *connHandler) Handle(logger *zap.Logger) *libp2pnetwork.NotifyBundle {
 		delete(ongoingHandshakes, pid)
 	}
 
+	var ignoredConnection = errors.New("ignored connection")
```
y0sher, line 74: convention is to use `ErrIgnoredConnection`. this way from outside you could check if the returned err is `connections.ErrIgnoredConnection`. currently its not clear its err from the name and its not exported.

y0sher: @moshe-blox though if this is used only internal to the package (like I can see now..) then its doesn't matter much..

y0sher (approved):

network/peers/connections/conn_handler.go
moshe-blox: @y0sher it'd be nice to return a bool to signify that connection was accepted, but yeah this func is internal so doesn't matter much

PR #1074, opened 2023-07-25
add encryption functionallty to ekm signer

guym-blox:


cli/operator/node.go
```diff
@@ -97,7 +99,13 @@ var StartNodeCmd = &cobra.Command{
 		}
 		nodeStorage, operatorData := setupOperatorStorage(logger, db)
 
-		keyManager, err := ekm.NewETHKeyManagerSigner(logger, db, networkConfig, cfg.SSVOptions.ValidatorOptions.BuilderProposals)
+		operatorKey, _, err := nodeStorage.GetPrivateKey()
```
y0sher, line 102: check err

ekm/signer_key_manager_test.go
```diff
@@ -47,6 +58,52 @@ func testKeyManager(t *testing.T) spectypes.KeyManager {
...
+	encryptionKey := fmt.Sprintf("%x", hash)
+	threshold.Init()
+	sk := bls.SecretKey{}
+	sk.SetByCSPRNG()
+	index := 0
+	logger := logging.TestLogger(t)
+	db, err := getBaseStorage(logger)
+	require.NoError(t, err)
+	signerStorage := NewSignerStorage(db, networkconfig.TestNetwork.Beacon, logger)
+	err = signerStorage.SetEncryptionKey(encryptionKey)
```
y0sher, line 76: check no err

cli/operator/node.go
y0sher: this is still not solved

y0sher:
Some tests fail, you probably need to fix up old tests to pass in a password.

cli/operator/node.go
```diff
@@ -97,7 +99,12 @@ var StartNodeCmd = &cobra.Command{
 		}
 		nodeStorage, operatorData := setupOperatorStorage(logger, db)
 
-		keyManager, err := ekm.NewETHKeyManagerSigner(logger, db, networkConfig, cfg.SSVOptions.ValidatorOptions.BuilderProposals)
+		operatorKey, _, _ := nodeStorage.GetPrivateKey()
```
y0sher, line 102: I would say lets check for error but after `setupOperatorStorage` we should have it for sure.

y0sher (approved):

moshe-blox:
@guym-blox do you think there is a point in encrypting/decrypting in `OpenWallet` and `SaveWallet` too?

cli/operator/node.go
guym-blox: this is why I removed it,
in that step the private key must be set or the node won't run

moshe-blox (approved):

moshe-blox:
gj 💪

PR #1078, opened 2023-07-31
adding migration to encrypt shares

guym-blox:


migrations/migration_encrypt_shares.go
```diff
@@ -0,0 +1,47 @@
...
+			return fmt.Errorf("failed to list accounts: %w", err)
+		}
+
+		operatorKey, found, err := nodeStorage.GetPrivateKey()
+		if err != nil {
+			return fmt.Errorf("failed to get private key: %w", err)
+		}
+		if !found {
+			return nil
+		}
```
y0sher, line 31: just for the sake of order, I would do it like this 
```suggestion
		nodeStorage, err := opt.nodeStorage(logger)
		if err != nil {
			return fmt.Errorf("failed to get node storage: %w", err)
		}
		operatorKey, found, err := nodeStorage.GetPrivateKey()
		if err != nil {
			return fmt.Errorf("failed to get private key: %w", err)
		}
		if !found {
			return nil
		}
		signerStorage := opt.signerStorage(logger)
		accounts, err := signerStorage.ListAccounts()
		if err != nil {
			return fmt.Errorf("failed to list accounts: %w", err)
		}
```

y0sher (changes requested):
small change please look

y0sher (approved):

migrations/migration_encrypt_shares.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/x509"
+	"fmt"
+
+	"go.uber.org/zap"
+)
+
+var encryptSharesMigration = Migration{
```
moshe-blox, line 12:
```suggestion
var migrationEncryptShares = Migration{
```

just to keep the naming convention of other migrations

this is just so that IDEs would autocomplete nicely when you start typing "migration"

migrations/migrations.go
```diff
@@ -24,6 +24,7 @@ var (
 	defaultMigrations = Migrations{
 		migrationExample1,
 		migrationExample2,
+		encryptSharesMigration,
```
moshe-blox, line 27:
```suggestion
		migrationEncryptShares,
```

migrations/migration_encrypt_shares.go
```diff
@@ -0,0 +1,46 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to set encryption key: %w", err)
+		}
+		for _, account := range accounts {
+			err := signerStorage.SaveAccount(account)
+			if err != nil {
+				return fmt.Errorf("failed to save account %s: %w", account, err)
+			}
+		}
+		return nil
```
moshe-blox, line 44: this needs to save the accounts and set the migration as completed in a single database transaction

moshe-blox (changes requested):

moshe-blox (dismissed):

migrations/migration_2_encrypt_shares.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/bloxapp/ssv/storage/basedb"
+
+	"go.uber.org/zap"
+)
+
+var migrationEncryptShares = Migration{
+	Name: "migration_2_encrypt_shares",
+	Run: func(ctx context.Context, logger *zap.Logger, opt Options, key []byte) error {
+		return opt.Db.Update(func(txn basedb.Txn) error {
+			err := txn.Set(migrationsPrefix, key, migrationCompleted)
```
y0sher, line 18: this is done twice? in return as well. I think lets leave the return only?

y0sher (changes requested):

migrations/migration_2_encrypt_shares.go
y0sher: could've just used defer? but since its all a tx it doesn't really matter.

y0sher (approved):

moshe-blox (approved):

PR #1079, opened 2023-08-01
 Prevent Unnecessary Round Bumps by Validating Timeout Messages

olegshmuelov:
### Description:

In the existing implementation, there's a possibility that the processing of a delayed timeout message can cause an additional, unnecessary round bump. This situation can occur if the instance receives a round justification message leading to a +1 round bump, and then shortly after handles a round timeout message from the same round, causing another +1 round bump and thereby leading to an undesired +2 round jump.

To resolve this issue, this PR introduces a mechanism to associate each timeout message with a specific round number. Upon receiving a timeout message, the instance now checks if the round number in the message matches its current round. If the message's round number is lower than the current round, the timeout message is outdated and is discarded, preventing the extra round bump.

These changes ensure that each round bump is justified and prevent the processing of delayed timeout messages that could disrupt the round synchronization in the distributed system.


y0sher:
@olegshmuelov this looks good and like it could solve the issue. but I'm very hesitant about adding in code that is not tested. I think unless we could prove this help, we don't know if we need this.
@moshe-blox wdyt?

y0sher:
It seems this will be on hold until we can get a test done. I don't see urgency to merge this without tests.

y0sher (approved):

moshe-blox (approved):

PR #1081, opened 2023-08-06
fix: notify `DutyScheduler` after starting new validators

moshe-blox:
`UpdateValidatorsMetadataLoop` wasn't seeing that validators were started after calling `UpdateValidatorsMetadataBatch` because it was asynchronous.

Along the way, I realized this function is unnecessary so I removed it and now call `UpdateValidatorsMetadata` (which is synchronous) instead.

- Removed unnecessary `UpdateValidatorsMetadataBatch` (go-eth2-client already batches, and even so from my experience both Prysm & Lighthouse can respond with 600k validators in one go)
- Removed unnecessary `batch` function and relevant tests
- Removed unnecessary usage of `utilsprotocol.Queue`
- Refactored `Validator.Start` to return whether it was really started or not


moshe-blox:
TODO: Should we batch at all? If so then we should do so in every call to update metadata and not just in loop.

operator/validator/controller.go
```diff
@@ -718,54 +718,58 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
+		return !ok || time.Since(last) > c.metadataUpdateInterval
+	})
 
-		// Filter for validators which haven't been updated recently.
-		filters = append(filters, func(s *ssvtypes.SSVShare) bool {
-			last, ok := lastUpdated[string(s.ValidatorPubKey)]
-			return !ok || time.Since(last) > c.metadataUpdateInterval
-		})
+	for {
+		time.Sleep(interval)
```
y0sher, line 753: should we wait at first iteration? if not we can move this to end of loop

operator/validator/controller.go
```diff
@@ -159,6 +156,7 @@ type controller struct {
 	nonCommitteeMutex      sync.Mutex
 
 	recentlyStartedValidators atomic.Int64
+	metadataLastUpdated       *hashmap.Map[string, time.Time]
```
y0sher, line 159: I don't see why this needs to be `hashmap` (threadsafe). this is used only within the same loop in only one go routine.

operator/validator/controller.go
```diff
@@ -718,54 +718,58 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
+		time.Sleep(interval)
+		start := time.Now()
 
 		// Get the shares to fetch metadata for.
 		shares := c.sharesStorage.List(nil, filters...)
 		var pks [][]byte
 		for _, share := range shares {
 			pks = append(pks, share.ValidatorPubKey)
-			lastUpdated[string(share.ValidatorPubKey)] = time.Now()
+			c.metadataLastUpdated.Set(string(share.ValidatorPubKey), time.Now())
```
y0sher, line 761: Look at the comment I put on `metadataLastUpdated` on the top, this wasn't thread safe before, no need to add hashmap and make it thread safe now. map seems to be enough since it only used in this loop.

protocol/v2/ssv/validator/startup.go
```diff
@@ -9,53 +9,56 @@ import (
...
+func (v *Validator) Start(logger *zap.Logger) (started bool, err error) {
 	logger = logger.Named(logging.NameValidator).With(fields.PubKey(v.Share.ValidatorPubKey))
 
-	if atomic.CompareAndSwapUint32(&v.state, uint32(NotStarted), uint32(Started)) {
-		n, ok := v.Network.(p2p.Subscriber)
-		if !ok {
-			return nil
+	if !atomic.CompareAndSwapUint32(&v.state, uint32(NotStarted), uint32(Started)) {
+		return false, nil
+	}
```
y0sher, line 25: 👍

operator/validator/controller.go
```diff
@@ -718,54 +718,58 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
 	}
-	if err := v.Start(c.logger); err != nil {
+	started, err := v.Start(c.logger)
+	if err != nil {
 		c.metrics.ValidatorError(v.Share.ValidatorPubKey)
 		return false, errors.Wrap(err, "could not start validator")
 	}
-	c.recentlyStartedValidators.Add(1)
+	if started {
+		c.recentlyStartedValidators.Add(1)
```
y0sher, line 727: I would say this also doesn't need to be atomic but it seems it is accessed from startValidator and UpdateMetadataLoop, though I don't see how they are run in parallel. hence we could keep this atomic.

y0sher (changes requested):
thanks for addressing the comments, lets restore gitlab ci file and merge

y0sher (approved):

olegshmuelov (approved):

PR #1092, opened 2023-08-14
[Audit Suggestion 3] Prevent Badger InMemory From Running in Production.

nkryuchkov:
Looking at the code it's not easy to make sure prod/stage don't run in-memory BadgerDB despite the default DB is not in-memory. The PR creates an explicit function for creating in-memory badger which should be used only in tests


moshe-blox (dismissed):
much better than before, thanks 👍

PR #1097, opened 2023-08-14
[Audit Issue D] Crash on config breaking changes

nkryuchkov:
Crash if node was already run with different network config or whether to use local events changed


moshe-blox:
maybe let's create a single struct for this?

for example `struct ConfigLock { ... }` (or `FrozenConfig` 😅) which would contain these 2 fields, and then we only need to have 2 funcs in the db for this (GetConfigLock/SaveConfigLock)

this would be useful especially when we'll add more params to it in the future

also, you'd be able to just encode them to JSON to check equality (of the FrozenConfig saved in db vs. the one that the node started up with)

nkryuchkov:
@moshe-blox I thought about this approach first but then I realized that the error logs would be clearer for user with the current approach. If a config just doesn't match, it would be not so clear for user what's wrong. If we compare each field and return an error based on that, that may look complex. I'll try to check the second approach

moshe-blox:
> @moshe-blox I thought about this approach first but then I realized that the error logs would be clearer for user with the current approach. If a config just doesn't match, it would be not so clear for user what's wrong. If we compare each field and return an error based on that, that may look complex. I'll try to check the second approach

agreed that errors messages need to be clear so the user can quickly fix, but you can still do this with only 1 struct

the point is to not have to create & maintain a new DB func for each field

for this you only need:
- a `ConfigLock` struct (for example)
- a `createConfigLock` to create it from the full config struct
- two db funcs to get & save it
- a comparison function between 2 `ConfigLock`s that returns a descriptive error, such as `ConfigDiff{Field: "Network", Frozen: "mainnet", Given: "prater"}` with an `Error() string` function that would format it nicely ("database was created with <Field>=<Frozen>, but <Given> was given")

nkryuchkov:
@moshe-blox Yeah, I'm working on something like this

nkryuchkov:
@moshe-blox done, please review

PR #1098, opened 2023-08-14
[Audit Issue C] Limit decoded share size

nkryuchkov:
Max share size seems to be 1245. Limiting max allowed input size to twice of that

### DIFF
--- a/protocol/v2/types/ssvshare.go
+++ b/protocol/v2/types/ssvshare.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	MaxPossibleShareSize = 1245
-	MaxAllowedShareSize  = MaxPossibleShareSize * 2 // Leaving some room for protocol updates and calculation mistakes.
+	MaxAllowedShareSize  = MaxPossibleShareSize * 8 // Leaving some room for protocol updates and calculation mistakes.
 )
 
 // SSVShare is a combination of spectypes.Share and its Metadata.
--- a/protocol/v2/types/ssvshare.go
+++ b/protocol/v2/types/ssvshare.go
@@ -15,7 +15,10 @@ import (
 	beaconprotocol "github.com/bloxapp/ssv/protocol/v2/blockchain/beacon"
 )
 
-const MaxPossibleShareSize = 1245
+const (
+	MaxPossibleShareSize = 1245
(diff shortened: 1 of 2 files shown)


protocol/v2/types/ssvshare.go
```diff
@@ -15,6 +15,11 @@ import (
 	beaconprotocol "github.com/bloxapp/ssv/protocol/v2/blockchain/beacon"
 )
 
+const (
+	MaxPossibleShareSize = 1245
+	MaxAllowedShareSize  = MaxPossibleShareSize * 2 // Leaving some room for protocol updates and calculation mistakes.
```
moshe-blox, line 20: i think we can be even more forgiving and allow up to 16kb per share (8 * 1024)

the attack vector they mentioned is if we try to load megabytes of data, so we don't really need to be precise, but rather just avoid huge decoding datas

moshe-blox (changes requested):

protocol/v2/types/ssvshare.go
moshe-blox: i think we'll likely forget to update these anyway, no matter how we call the consts

alternatively, you can just increase the factor from 2 to 8 for example, just to give more room for mistakes

after all, we don't even think this is a realistic attack vector 😅

nkryuchkov: I just don't understand why we need to tie to an exact number. But I don't mind increasing the factor. With the current approach, on share size increase the test checking the share size would break and we'd immediately see it and adjust the value

moshe-blox: the bad scenario with small allowable size is that some user creates a share with a long graffiti (or a new field we allow) and we don't test this, but then in production it will not decode, so i prefer to allow big sizes as long as they don't cause us to go OOM

nkryuchkov: I understand your concern but we don't have such fields, the graffiti is 32 bytes. Also, `MaxPossibleShareSize` is calculated by creating a share of max possible size

nkryuchkov: @moshe-blox Increased factor to 8

nkryuchkov: By the way, setting a constant is less readable in my opinion. If someone looks at the 16kb limit, it's not clear if the share fits it and requires calculating the share size

moshe-blox (approved):

PR #1099, opened 2023-08-14
[Audit Issue F] Add missing Check on Quorum for the RoundChange Justification

nkryuchkov:
Port of https://github.com/bloxapp/ssv-spec/pull/207


GalRogozinski:
@nkryuchkov 
How did the test pass before?
See https://github.com/bloxapp/ssv-spec/pull/207/files#diff-1303c5099a54727ffafeb48b9fa2548d1081b5570765c0fbc3bffff6c4d019b3

nkryuchkov:
> @nkryuchkov How did the test pass before? See https://github.com/bloxapp/ssv-spec/pull/207/files#diff-1303c5099a54727ffafeb48b9fa2548d1081b5570765c0fbc3bffff6c4d019b3

@GalRogozinski Because the implementation calls spec's `CreateRoundChange` which is fixed: https://github.com/bloxapp/ssv/blob/main/protocol/v2/qbft/spectest/create_msg_type.go#L99

moshe-blox (approved):

GalRogozinski (approved):
I don't understand what are the numbers in the spec-allignment script though

Explanation?

nkryuchkov:
@GalRogozinski As far as I understand, those are hashes of files that differ from spec, so if we have a non-whitelisted change, a CI check would fail.

Not sure if my explanation is correct, perhaps @moshe-blox could explain it better because he authored it

PR #1100, opened 2023-08-14
[Audit Suggestion 1] Do not use panics on fork

nkryuchkov:
Use Fatal instead of Panic, do that in parent method

### DIFF
--- a/operator/node.go
+++ b/operator/node.go
@@ -151,7 +151,7 @@ func (n *operatorNode) Start(logger *zap.Logger) error {
 	go n.ticker.Start(logger)
 	go func() {
 		if err := n.listenForCurrentSlot(logger); err != nil {
-			logger.Fatal(err.Error())
+			logger.Fatal("unexpected error while listening for current slot", zap.Error(err))
 		}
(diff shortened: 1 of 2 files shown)


operator/node.go
```diff
@@ -149,7 +149,11 @@ func (n *operatorNode) Start(logger *zap.Logger) error {
 	}()
 
 	go n.ticker.Start(logger)
-	go n.listenForCurrentSlot(logger)
+	go func() {
+		if err := n.listenForCurrentSlot(logger); err != nil {
+			logger.Fatal(err.Error())
```
moshe-blox, line 154:
```suggestion
			logger.Fatal("failed listening for current slot", zap.Error(err))
```

wdyt?

moshe-blox (changes requested):

operator/node.go
nkryuchkov: I didn't do that to keep the same error, but after looking at it again I think logs would look good with your suggestion. I agree

moshe-blox (approved):

PR #1102, opened 2023-08-14
[Audit Suggestion 7] Pin package versions in Dockerfile

nkryuchkov:


nkryuchkov:
@moshe-blox I also doubt we need this

lior-blox:
@moshe-blox let’s ask the audit team what they suggest in that case and if it’s indeed best practice

shareef-dweikat:
Actually i have an example from the JS world, but still relevant to our case here. Almost 5 years ago the maintainer of a famous NPM package called event-stream passed permissions to an unknown developer to maintain the package. It turned out the new developer is a bad guy. He pushed a malicious code to the library. A lot of projects where affected because they updated their node_modules without carefully checking the code of the new released versions!

PR #1108, opened 2023-08-20
feat: retry standard block after Prysm fails to retrieve blinded

moshe-blox:
This PR retries to get non-blinded block after Prysm failed to retrieve a blinded block.


protocol/v2/ssv/runner/proposer.go
```diff
@@ -105,20 +109,33 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
 		// get block data
 		obj, ver, err = r.GetBeaconNode().GetBlindedBeaconBlock(duty.Slot, r.GetShare().Graffiti, fullSig)
 		if err != nil {
-			// Prysm currently doesn’t support MEV.
-			// TODO: Check Prysm MEV support after https://github.com/prysmaticlabs/prysm/issues/12103 is resolved.
-			return errors.Wrap(err, "failed to get Beacon block")
+			// Prysm workaround: when Prysm can't retrieve an MEV block, it responds with an error
+			// saying the block isn't blinded, implying to request a standard block instead.
+			// https://github.com/prysmaticlabs/prysm/issues/12103
+			if strings.Contains(err.Error(), "Prepared beacon block is not blinded") {
```
lior-blox, line 115: @moshe-blox for any other err we dont want to fallback?

protocol/v2/ssv/runner/proposer.go
```diff
@@ -390,3 +412,43 @@ func (r *ProposerRunner) GetRoot() ([32]byte, error) {
...
 }
+
+type blockSummary struct {
+	Hash    phase0.Hash32
+	Blinded bool
+	Version spec.DataVersion
+}
+
+func summarizeBlock(block any) (summary blockSummary, err error) {
+	if block == nil {
```
lior-blox, line 423: @moshe-blox maybe better to take it out to a separate file?

protocol/v2/ssv/runner/proposer.go
moshe-blox: > @moshe-blox for any other err we dont want to fallback?

we could, but the problem with this is that then you will start retrying in other clients which don't have this issue and don't benefit from retrying

moshe-blox: maybe we should change the condition to if the client is Prysm

protocol/v2/ssv/runner/proposer.go
moshe-blox: its a small function used in this file for logging info about the proposal, i added some comments on it now

moshe-blox: also, spec-alignment script automatically skips it (bc there's no such func in spec)

protocol/v2/ssv/runner/proposer.go
lior-blox: I like that more I think (client is prysm).
But why not to fallback in all clients?
I mean if you get an error for blinded you don't continue anyhow

protocol/v2/ssv/runner/proposer.go
lior-blox: ok...

y0sher (approved):

protocol/v2/ssv/runner/proposer.go
moshe-blox: > I like that more I think (client is prysm). But why not to fallback in all clients? I mean if you get an error for blinded you don't continue anyhow

you continue to sign the leader's block if you're not the leader, so this would cause an unnecessary delay with other clients

even the Prysm check isn't ideal because it would cause the same unnecessary delay when the error isn't "prepared beacon block is not blinded"

moshe-blox: correction: this may cause a delay to the round change

protocol/v2/ssv/runner/proposer.go
```diff
@@ -105,20 +109,33 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
+			return errors.Wrap(err, "failed to get beacon block")
 		}
 	}
 
-	logger.Debug("🧊 got beacon block",
-		zap.Duration("took", time.Since(start)))
+	// Log essentials about the retrieved block.
+	blockSummary, summarizeErr := summarizeBlock(obj)
+	logger.Info("🧊 got beacon block proposal",
+		zap.String("block_hash", blockSummary.Hash.String()),
```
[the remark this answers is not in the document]
moshe-blox: theres already a slot in the logger

PR #1116, opened 2023-08-23
Deprecate current fork support implementation

olegshmuelov:


moshe-blox (approved):

moshe-blox:
@olegshmuelov gj, code approved 👍
just need to test that this can reach consensus with old nodes, by deploying to 2/4 nodes for example

PR #1122, opened 2023-08-29
Added operator data before eh test run to increase the code coverage 

AKorpusenko:


eth/eventhandler/event_handler_test.go
```diff
@@ -203,6 +205,23 @@ func TestHandleBlockEventsStream(t *testing.T) {
...
+		operatorEthKey, err := crypto.HexToECDSA("42e14d227125f411d6d3285bb4a2e07c2dba2e210bd2f3f4e2a36633bd61bfe6")
+		require.NoError(t, err)
+		operatorEthAdr := crypto.PubkeyToAddress(operatorEthKey.PublicKey)
+
+		eh.operatorData.SetOperatorData(&registrystorage.OperatorData{
+			PublicKey:    ops[0].pub,
+			ID:           ops[0].id,
+			OwnerAddress: operatorEthAdr,
+		})
+		_, err = eh.nodeStorage.SetupPrivateKey(base64.StdEncoding.EncodeToString(ops[0].priv))
```
y0sher, line 218: I don't get why you set this twice, once here and once in `setupEventHandler`

eth/eventhandler/event_handler_test.go
```diff
@@ -577,9 +596,11 @@ func setupEventHandlerWithMockedCtrl(t *testing.T, ctx context.Context, logger *
 		Ctx: ctx,
 	})
 	require.NoError(t, err)
+	_, operatorSk, err := rsaencryption.GenerateKeys()
+	require.NoError(t, err)
 
 	storageMap := ibftstorage.NewStores()
-	nodeStorage, _ := setupOperatorStorage(logger, db)
+	nodeStorage, _ := setupOperatorStorage(logger, db, &testOperator{priv: operatorSk, id: 0})
```
y0sher, line 603: if we would ever want the storage to have an operator in test that use `setupEventHandlerWithMockedCtrl`. we would probably need to change this. also 0 is not a valid operator key.
if these test didn't have valid operator to this day we will need to add in the future.

eth/eventhandler/event_handler_test.go
AKorpusenko: Fixed

eth/eventhandler/event_handler_test.go
```diff
@@ -59,7 +59,12 @@ func TestHandleBlockEventsStream(t *testing.T) {
 	logger := zaptest.NewLogger(t)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	eh, err := setupEventHandler(t, ctx, logger)
+
+	// Create operators rsa keys
+	ops, err := createOperators(4)
```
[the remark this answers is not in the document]
AKorpusenko: Fixed

eth/eventhandler/local_events_test.go
```diff
@@ -17,6 +17,10 @@ import (
 )
 
 func TestHandleLocalEvent(t *testing.T) {
+	// Create operators rsa keys
+	ops, err := createOperators(4)
```
y0sher, line 21: says 4 but only 1 is use

eth/eventhandler/event_handler_test.go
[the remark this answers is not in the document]
y0sher: is still says 4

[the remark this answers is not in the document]
AKorpusenko: I need all 4 validators here. This is correct

PR #1131, opened 2023-09-04
fix: jammed `indicesChange` channel

moshe-blox:
One of the subscribers of that channel was not reading messages from it, which caused the `event.Feed` to get stuck on it's `Send` method for eternity.

### DIFF
--- a/operator/duties/scheduler_test.go
+++ b/operator/duties/scheduler_test.go
@@ -270,13 +270,12 @@ func TestScheduler_Regression_IndiciesChangeStuck(t *testing.T) {
 	err := s.Start(ctx, logger)
 	require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


olegshmuelov (approved):

y0sher (approved):
Great job, this works! 

A good rule of thumb for the future is, always create the channel you're reading from. with this approach the reader will create the channel and set the rules (buffered/non, goroutine,none) and will "request" the sender to send on this channel knowing that it listens. this could prevent such cases in the future.

y0sher (dismissed):

y0sher (dismissed):

y0sher (approved):

olegshmuelov (approved):

PR #1140, opened 2023-09-10
Fix: Use Share PublicKey for Share Removal

olegshmuelov:
### Problem

Previously, shares were being removed based on the validator's public key, which was incorrect.

### Solution

This fix introduces the proper way to delete a share by using the share's public key instead of the validator's public key. This ensures that the correct share is removed from the EKM (ETH Key Manager) storage.

UPD: based on @moshe-blox [comment](https://github.com/bloxapp/ssv/pull/1140#pullrequestreview-1618840674)
added a logic that interact with database to be as part of the tx in eventhandler.

### DIFF
--- a/ekm/eth_key_manager_signer.go
+++ b/ekm/eth_key_manager_signer.go
@@ -43,9 +43,16 @@ type ethKeyManagerSigner struct {
 	builderProposals  bool
 }
 
+// StorageProvider provides the underlying KeyManager storage.
+type StorageProvider interface {
+	ListAccounts() ([]core.ValidatorAccount, error)
+	RetrieveHighestAttestation(pubKey []byte) (*phase0.AttestationData, bool, error)
+	RetrieveHighestProposal(pubKey []byte) (phase0.Slot, bool, error)
+}
+
 // NewETHKeyManagerSigner returns a new instance of ethKeyManagerSigner
 func NewETHKeyManagerSigner(logger *zap.Logger, db basedb.Database, network networkconfig.NetworkConfig, builderProposals bool, encryptionKey string) (spectypes.KeyManager, error) {
-	signerStore := NewSignerStorage(db, network.Beacon.GetNetwork(), logger)
+	signerStore := NewSignerStorage(db, network.Beacon, logger)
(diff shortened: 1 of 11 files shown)


moshe-blox (changes requested):
Code looks good 💪

Unrelated to this fix, I think there's another bug: `RemoveShare` is called from a task and not from event handler. This means that it's not part of the DB transaction and task failures are ignored (whereas event handling failures are retried).

I think we should do this in this PR because this would make the primary fix testable in `TestHandleBlockEventsStream` (which doesn't run tasks) by adding a test that removes a validator and checks that the SP data is gone as well.

Wdyt? @y0sher @olegshmuelov

y0sher:
> Code looks good 💪
> 
> Unrelated to this fix, I think there's another bug: `RemoveShare` is called from a task and not from event handler. This means that it's not part of the DB transaction and task failures are ignored (whereas event handling failures are retried).
> 
> I think we should do this in this PR because this would make the primary fix testable in `TestHandleBlockEventsStream` (which doesn't run tasks) by adding a check on the SP data in the already existing test that removes a validator.
> 
> Wdyt? @y0sher @olegshmuelov

I strongly agree that any code that plays with database should be as part of the tx in `eventhandler`.

olegshmuelov:
### Concern: Migration for Existing Shares

While this PR fixes the issue of shares being incorrectly removed by the validator's public key, it's worth noting that there may be existing shares in the ETH Key Manager storage that were not properly deleted due to this bug.

I recommend considering the creation of a migration script that scans for and deletes these improperly removed shares, particularly those that were not deleted by the `ValidatorRemove` event. This will ensure data integrity and consistency in the ETH Key Manager storage moving forward.

@moshe-blox @y0sher

y0sher (dismissed):
@olegshmuelov I agree about the migration point, though I think even without the migration there is no hard of having old shares in your DB. so we should make it a point to release the new version with this migration but it dosen't have to be a part of this task. I can create a task for it. 
@moshe-blox wdyt?

moshe-blox (approved):

y0sher (approved):

PR #1142, opened 2023-09-11
fix generate operator keys command

guym-blox:


cli/generate_operator_keys.go
```diff
@@ -5,119 +5,121 @@ import (
...
+	return skPem, pk, nil
+}
+
+func encryptAndSavePrivateKey(sk []byte, passwordFilePath string, logger *zap.Logger) {
+	absPath, err := filepath.Abs(passwordFilePath)
+	if err != nil {
+		logger.Fatal("Failed to read absolute path of password file", zap.Error(err))
+	}
+	// #nosec G304
+	passwordBytes, err := os.ReadFile(absPath)
```
y0sher, line 99: generally, I would take all I/O (file reading/writing) to different function, either the top function or just separate functions. this makes a more reusable and testable code.

cli/generate_operator_keys.go
```diff
@@ -5,119 +5,121 @@ import (
...
+	skPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(rsaKey),
+	})
+
+	operatorPublicKey, err := rsaencryption.ExtractPublicKey(rsaKey)
+	if err != nil {
+		return nil, nil, err
+	}
+	pk, err := base64.StdEncoding.DecodeString(operatorPublicKey)
```
y0sher, line 86: don't we already have tested methods for all this stuff in `rsaencryption`? if we do its better to reuse them instead of rewriting the code.

y0sher (changes requested):
Generally the change works, It'll be great to make it more future-proof.

cli/generate_operator_keys.go
```diff
@@ -1,123 +1,125 @@
...
+	if err != nil {
+		logger.Fatal(fmt.Sprintf("Failed to read absolute path of %s", filePath), zap.Error(err))
+	}
+	// #nosec G304
+	contentBytes, err := os.ReadFile(absPath)
+	if err != nil {
+		logger.Fatal(fmt.Sprintf("Failed to read file %s", filePath), zap.Error(err))
+	}
+	return contentBytes, err
+}
```
y0sher, line 121: this function is weird, you don't need logger if you return errors. return the error and log at the top level instead.

cli/generate_operator_keys.go
```diff
@@ -1,123 +1,125 @@
...
-						Type:  "RSA PRIVATE KEY",
-						Bytes: x509.MarshalPKCS1PrivateKey(rsaKey),
-					},
-				)
-
-				operatorPublicKey, _ := rsaencryption.ExtractPublicKey(rsaKey)
-				publicKey, _ := base64.StdEncoding.DecodeString(operatorPublicKey)
-				sk = skPem
-				pk = publicKey
+				logger.Fatal("Failed to read private key from file", zap.Error(err))
```
y0sher, line 37: I meant more for something like this -
```go
if privateKeyFilePath != "" {
   keybytes, err := readFile(privateKeyFilePath)
   if err != nil {
    logger.Fatal("failed to read file", zap.Error(err))
   }
  sk, err := readPrivateKey(keybytes)
  if err != nil {
    logger.Fatal("failed to parse public key", zap.Error(err))
  }
 pk = sk.PublicKey
}
  ```

File: cli/generate_operator_keys.go [hunk 5]
```diff
@@ -1,123 +1,125 @@
...
-			}
-
-			err = os.WriteFile("encrypted_private_key.json", encryptedJSON, 0600)
-			if err != nil {
-				logger.Fatal("Failed to write encrypted private key to file", zap.Error(err))
-			}
-
-			logger.Info("private key encrypted and stored in encrypted_private_key.json")
+		if passwordFilePath != "" {
+			encryptAndSavePrivateKey(sk, passwordFilePath, logger)
```
y0sher, line 46: please use similar structure to what I described above.
one function that encrypts and returns either encrypted data or error, and then use the writeFile func to write the returned data.

y0sher (approved):

moshe-blox (approved):

PR #1143, opened 2023-09-11
e2e tests for eth execution layer package

AKorpusenko:


eth/eventhandler/event_handler_test_utils.go
```diff
@@ -0,0 +1,299 @@
...
+
+func setupEventHandler(
+	t *testing.T,
+	ctx context.Context,
+	logger *zap.Logger,
+	operator *testOperator,
+	ownerAddress *ethcommon.Address,
+	useMockCtrl bool,
+) (*EventHandler, *mocks.MockController, error) {
+	db, err := kv.NewInMemory(logger, basedb.Options{
```
y0sher, line 165: if this is a copy in some way of what used in `cli/operator/node.go`. I think it'll be best to extract `setupEventHandler` to a function that receives most of what it needs (for example a validator controller interface that could be either a mock or actual validator controller instead of useMock bool) and then try to reuse code between `node.go` and the e2e tests, this will make sure we're testing the same functionality used even if code changes in the future.

eth/eth_test/common_test.go
```diff
@@ -0,0 +1,178 @@
...
+	logger := zaptest.NewLogger(t)
+
+	// Create operators RSA keys
+	ops, err := createOperators(operatorsCount, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	validators := make([]*testValidatorData, validatorsCount)
+	shares := make([][]byte, 10)
```
nkryuchkov, line 80: Does the exact value mean anything or is it just an arbitrary number?

eth/eth_test/common_test.go
```diff
@@ -0,0 +1,178 @@
...
+	rpcServer, _ := sim.Node.RPCHandler()
+	// Expose handler on a test server with ws open
+	httpSrv := httptest.NewServer(rpcServer.WebsocketHandler([]string{"*"}))
+	addr := "ws:" + strings.TrimPrefix(httpSrv.URL, "http:")
+
+	parsed, _ := abi.JSON(strings.NewReader(simcontract.SimcontractMetaData.ABI))
+	auth, _ := bind.NewKeyedTransactorWithChainID(testKeyAlice, big.NewInt(1337))
+	contractAddr, _, _, err := bind.DeployContract(auth, parsed, ethcommon.FromHex(simcontract.SimcontractMetaData.Bin), sim)
+	if err != nil {
+		t.Errorf("deploying contract: %v", err)
```
nkryuchkov, line 116: Should it be logged outside of the function? I feel like we could get rid of having `t` here, however, `setupEventHandler` uses it, so `setupEventHandler` and this function use two different approaches: The former asserts errors, the latter returns them. I'd align them to have similar approach to error handling

eth/eth_test/common_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if validatorCtrl == nil {
+		return nil, fmt.Errorf("error: validatorCtrl is empty")
+	}
+
+	// Adding testAddresses to the genesis block mostly to specify some balances for them
+	sim := simTestBackend(testAddresses)
+
+	// Create JSON-RPC handler
+	rpcServer, _ := sim.Node.RPCHandler()
```
nkryuchkov, line 107: I think we need to keep all errors checked unless there's a specific issue with the error here (in this case it needs to be documented)

eth/eth_test/common_test.go
AKorpusenko: Correct. I've removed this. Using the `validatorsCount` and `operatorsCount` params instead

AKorpusenko: Fixed

eth/eth_test/common_test.go
AKorpusenko: Fixed

eth/eth_test/common_test.go
AKorpusenko: Fixed

eth/eth_test/common_test.go
```diff
@@ -0,0 +1,205 @@
...
+		executionclient.WithFollowDistance(0),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	err = client.Healthy(ctx)
+	if err != nil {
+		// Always returns nil error
+		client.Close()
```
nkryuchkov, line 161: Why are these two closed not before each return? Maybe `defer` would suit here better?

eth/eth_test/utils_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	storageMap := ibftstorage.NewStores()
+	nodeStorage, operatorData := setupOperatorStorage(logger, db, operator, ownerAddress)
+	testNetworkConfig := networkconfig.TestNetwork
+
+	keyManager, err := ekm.NewETHKeyManagerSigner(logger, db, testNetworkConfig, true, "")
+	require.NoError(t, err)
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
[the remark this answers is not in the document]
AKorpusenko: Fixed

eth/eth_test/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("error: contractCode is empty")
+	}
+
+	// Create a client and connect to the simulator
+	e.execClient, err = executionclient.New(
+		ctx,
+		addr,
+		contractAddr,
+		executionclient.WithLogger(logger),
+		executionclient.WithFollowDistance(0),
```
moshe-blox, line 172: This means follow distance is never tested, even though it's always used in practice (value is 8 in production). Can we please try increasing this to 1 or 2 and adjust the tests accordingly?

I imagine that the major difference is that the tests would have to produce an additional `followDistance` number of empty blocks after a block with transactions to make the execution client sync the transactions. Probably we can just do `for i := 0; i < followDistance; i++ { Commit() }` each time after producing a block with a tx that we want to test. This can probably be automated in a special `CloseFollowDistance` method under `TestEnv`, so the test may commit its transactions in however many blocks it wants and then we may call this method to make sure we sync them.

Additionally, when confirming sync height, instead of comparing `lastProcessedBlockNumber` to `currentBlockNumber`, you may have to compare to `currentBlockNumber - followDistance`.

moshe-blox (changes requested):
Very well-written!

I really like the commenting and utilities you made for the tests.

Left a few comments, only 1 is significant, however I think it's quite important to test.

eth/eth_test/eth_e2e_test.go
```diff
@@ -0,0 +1,300 @@
+package eth_test
```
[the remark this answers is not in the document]
nkryuchkov: @moshe-blox actually, the underscore is used in package names in test files 😄 So if a package named `x` then its tests' package may be either `x` if tests use non-exported symbols or `x_test` for black-box testing approach. But for regular packages (which can be imported) the underscore should be avoided. Therefore, both `eth` and `eth_test` are fine here (BTW, this location is unique in a way that it has no code, only tests, so it's hard to say what package name it must have)

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	operatorPubKey, err := nodeStorage.SetupPrivateKey(base64.StdEncoding.EncodeToString(operator.priv))
+	if err != nil {
+		logger.Fatal("could not setup operator private key", zap.Error(err))
+	}
+
+	_, found, err := nodeStorage.GetPrivateKey()
+	if err != nil || !found {
+		logger.Fatal("failed to get operator private key", zap.Error(err))
```
nkryuchkov, line 271: if err is nil and not found, then error would be `failed to get operator private key error=nil` which looks confusing

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func setupOperatorStorage(
+	logger *zap.Logger,
+	db basedb.Database,
+	operator *testOperator,
+	ownerAddress *ethcommon.Address,
+) (operatorstorage.Storage, *registrystorage.OperatorData) {
+	if operator == nil {
+		logger.Fatal("empty test operator was passed", zap.Error(fmt.Errorf("empty test operator was passed")))
```
nkryuchkov, line 256: I think the error field is redundant here 

```suggestion
		logger.Fatal("empty test operator was passed")
```

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func generateSharesData(validatorData *testValidatorData, operators []*testOperator, owner ethcommon.Address, nonce int) ([]byte, error) {
+	var pubkeys []byte
+	var encryptedShares []byte
+
+	for i, op := range operators {
+		rsakey, err := rsaencryption.ConvertPemToPublicKey(op.pub)
+		if err != nil {
+			return nil, fmt.Errorf("cant convert publickey: %w", err)
```
nkryuchkov, line 114: `can't` or `cannot` (although, I'd remove this word), `public key` or `pubkey`

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		rawshare := validatorData.operatorsShares[i].sec.SerializeToHexStr()
```
nkryuchkov, line 117: let's follow the camel case naming rules in this function

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/bloxapp/ssv/utils/threshold"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/golang/mock/gomock"
+	"github.com/herumi/bls-eth-go-binary/bls"
+	"go.uber.org/zap"
+	"math/big"
+
+	"testing"
```
nkryuchkov, line 35: Why is only this one in a separate block? I suggest having 3 blocks:
- std lib
- 3rd party libs
- github.com/bloxapp/ssv

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+type testValidatorData struct {
+	masterKey        *bls.SecretKey
+	masterPubKey     *bls.PublicKey
+	masterPublicKeys bls.PublicKeys
+	operatorsShares  []*testShare
+}
+
+type testOperator struct {
+	id   uint64
+	pub  []byte // rsa pub
```
nkryuchkov, line 47: If this needs to be commented, I'd consider renaming the struct fields, rsaPub/rsaPriv or rsaPK/rsaSK. But I'd rather remove comments

eth/ethtest/common_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	sim.Commit()
+
+	// Check contract code at the simulated blockchain
+	contractCode, err := sim.CodeAt(ctx, contractAddr, nil)
+	if err != nil {
+		return fmt.Errorf("getting contract code: %w", err)
+	}
+	if len(contractCode) == 0 {
+		return fmt.Errorf("error: contractCode is empty")
```
nkryuchkov, line 167: why do we need `error: ` prefix?

eth/ethtest/common_test.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	// Expose handler on a test server with ws open
+	httpSrv := httptest.NewServer(rpcServer.WebsocketHandler([]string{"*"}))
+	e.httpSrv = httpSrv
+
+	addr := "ws:" + strings.TrimPrefix(httpSrv.URL, "http:")
+
+	parsed, err := abi.JSON(strings.NewReader(simcontract.SimcontractMetaData.ABI))
+	if err != nil {
+		return fmt.Errorf("parsing contract abi: %w", err)
```
nkryuchkov, line 146:
```suggestion
		return fmt.Errorf("parsing contract ABI: %w", err)
```

eth/ethtest/common_test.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Adding testAddresses to the genesis block mostly to specify some balances for them
+	sim := simTestBackend(testAddresses)
+
+	// Create JSON-RPC handler
+	rpcServer, err := sim.Node.RPCHandler()
+	e.rpcServer = rpcServer
+	if err != nil {
+		return fmt.Errorf("creatingt rpc server: %w", err)
```
nkryuchkov, line 136:
```suggestion
		return fmt.Errorf("create RPC server: %w", err)
```

eth/eth_test/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+func simTestBackend(testAddresses []*ethcommon.Address) *simulator.SimulatedBackend {
+	genesis := core.GenesisAlloc{}
+
+	for _, testAddr := range testAddresses {
+		genesis[*testAddr] = core.GenesisAccount{Balance: big.NewInt(10000000000000000)}
+	}
+
+	return simulator.NewSimulatedBackend(
+		genesis, 10000000,
```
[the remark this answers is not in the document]
AKorpusenko: Fixed. Thanks

eth/eth_test/eth_e2e_test.go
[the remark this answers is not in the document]
AKorpusenko: Renamed

eth/eth_test/common_test.go
AKorpusenko: Added this. Thanks. Good point

eth/ethtest/utils_test.go
AKorpusenko: this is exact copy of what we got in `node.go`. [link](https://github.com/bloxapp/ssv/blob/stage/cli/operator/node.go#L433)

eth/ethtest/utils_test.go
AKorpusenko: This one has to be set in project formatting options. Default golang `gofmt` is used here

eth/ethtest/utils_test.go
AKorpusenko: Agree. Fixed

eth/ethtest/common_test.go
AKorpusenko: No need here. Removed

eth/ethtest/common_test.go
AKorpusenko: Fixed. Thanks

eth/ethtest/common_test.go
AKorpusenko: Fixed. Thanks

eth/ethtest/utils_test.go
AKorpusenko: Fixed. Thanks

eth/ethtest/utils_test.go
AKorpusenko: Fixed. Thanks

eth/eventhandler/event_handler_test.go
```diff
@@ -1395,7 +1395,7 @@ func createNewValidator(ops []*testOperator) (*testValidatorData, error) {
 	num := uint64(len(ops))
 	validatorData.operatorsShares = make([]*testShare, num)
 
-	// derive a `hareCount` number of shares
+	// derive a `shareCount` number of shares
```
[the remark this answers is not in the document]
AKorpusenko: Fixed. Thanks for attention

eth/ethtest/common_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	auth, err := bind.NewKeyedTransactorWithChainID(testKeyAlice, big.NewInt(1337))
+	if err != nil {
+		return err
+	}
+
+	contractAddr, _, _, err := bind.DeployContract(auth, parsed, ethcommon.FromHex(simcontract.SimcontractMetaData.Bin), sim)
+	if err != nil {
+		return fmt.Errorf("deploying contract: %w", err)
```
nkryuchkov, line 158: I think initial form of verb is easier to read and as far as I know it's usually used. Let's use "deploy"

eth/ethtest/utils_test.go
```diff
@@ -0,0 +1,300 @@
...
+	for i, op := range operators {
+		rsaKey, err := rsaencryption.ConvertPemToPublicKey(op.rsaPub)
+		if err != nil {
+			return nil, fmt.Errorf("can't convert public key: %w", err)
+		}
+
+		rawShare := validatorData.operatorsShares[i].sec.SerializeToHexStr()
+		cipherText, err := rsa.EncryptPKCS1v15(rand.Reader, rsaKey, []byte(rawShare))
+		if err != nil {
+			return nil, errors.New("cant encrypt share")
```
nkryuchkov, line 120: `can't`

also, is there any reason to hide `err`? if not, lets's wrap it

eth/ethtest/utils_test.go
nkryuchkov: I think it might be a good time to fix both 🙂

eth/eventhandler/event_handler_test.go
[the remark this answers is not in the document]
nkryuchkov: comment: `shareCount`
code: `sharesCount`

AKorpusenko:
https://github.com/bloxapp/ssv/pull/1143#discussion_r1343714963

@nkryuchkov, this should be done in a separate PR, due the current task assumes to provide just a e2e tests. This is wrong to make any changes in prod code behavior, even in the error handlers

nkryuchkov (approved):

nkryuchkov:
https://github.com/bloxapp/ssv/pull/1143#issuecomment-1745470112

ok, since it's only tests, I agree

eth/eth_test/common_test.go
moshe-blox: > Added this. Thanks. Good point

Thanks!

moshe-blox (approved):
gj 👌

PR #1144, opened 2023-09-13
Implement Slashing Protection Data Update for Reactivated Validators

olegshmuelov:
#### Description:

This PR addresses the issue of outdated slashing protection data for reactivated validators in the EKM (Ethereum Key Manager). Specifically, it aims to prevent the signing of slashable attestations or proposals by ensuring that the slashing protection data is updated appropriately when a validator is reactivated.

#### Changes:

- Added logic to update slashing protection data for reactivated validators.
- The slashing protection data is "bumped" to the next epoch & slot only if the stored values are lower. This ensures that we do not overwrite higher stored values.

```go
# Pseudo-code to demonstrate the bump logic
storedEpoch = max(nextEpoch, storedEpoch)
```

#### Testing:

- Added unit tests to verify that the slashing protection data is updated correctly.

### DIFF
--- a/ekm/eth_key_manager_signer.go
+++ b/ekm/eth_key_manager_signer.go
@@ -343,8 +343,6 @@ func (km *ethKeyManagerSigner) BumpSlashingProtection(pubKey []byte) error {
 
 // updateHighestAttestation updates the highest attestation data for slashing protection.
 func (km *ethKeyManagerSigner) updateHighestAttestation(pubKey []byte, slot phase0.Slot) error {
-	var highAtt *phase0.AttestationData
-
 	// Retrieve the highest attestation data stored for the given public key.
 	retrievedHighAtt, found, err := km.RetrieveHighestAttestation(pubKey)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


eth/eventhandler/handlers.go
```diff
@@ -422,6 +423,13 @@ func (eh *EventHandler) handleClusterReactivated(txn basedb.Txn, event *contract
 		return nil, fmt.Errorf("could not process cluster event: %w", err)
 	}
 
+	// bump slashing protection for operator reactivated validators
+	for _, share := range toReactivate {
+		if err := eh.keyManager.(ekm.StorageProvider).BumpSlashingProtection(share.SharePubKey); err != nil {
```
y0sher, line 428: I don't really like this cast but I understand otherwise it will require spec change. I thought maybe to use add share and add the functionality in add share to bump if the share exist, it will not return error if it exists on validatorAdded though which might be an issue. wdyt? @moshe-blox @olegshmuelov

ekm/eth_key_manager_signer.go
```diff
@@ -334,28 +324,125 @@ func (km *ethKeyManagerSigner) RemoveShare(pubKey string) error {
...
 	}
-	account := wallets.NewValidatorAccount("", key, nil, "", nil)
-	if err := km.wallet.AddValidatorAccount(account); err != nil {
-		return errors.Wrap(err, "could not save new account")
+
+	currentEpoch := km.storage.BeaconNetwork().EstimatedEpochAtSlot(slot)
+
+	// Validate the retrieved highest attestation data.
+	if found && retrievedHighAtt != nil {
+		if retrievedHighAtt.Source.Epoch < currentEpoch-1 && retrievedHighAtt.Target.Epoch < currentEpoch {
```
moshe-blox, line 358: Compare against `computeMinimalAttestationSP`

ekm/eth_key_manager_signer.go
```diff
@@ -334,28 +324,125 @@ func (km *ethKeyManagerSigner) RemoveShare(pubKey string) error {
...
+
+	// Retrieve the highest proposal slot stored for the given public key.
+	retrievedHighProp, found, err := km.RetrieveHighestProposal(pubKey)
+	if err != nil {
+		return fmt.Errorf("could not retrieve highest proposal: %w", err)
 	}
+
+	// Validate the retrieved highest proposal slot.
+	if found && retrievedHighProp != 0 {
+		if retrievedHighProp < slot {
```
moshe-blox, line 391: Compare against `computeMinimalProposerSP`

eth/eventhandler/handlers.go
moshe-blox: the problem is that addshare currently does not bump the SP data if the account already exists (as it always did)

moshe-blox (approved):

moshe-blox (approved):

ekm/signer_key_manager_test.go
```diff
@@ -36,15 +35,22 @@ const (
 	pk2Str = "8796fafa576051372030a75c41caafea149e4368aebaca21c9f90d9974b3973d5cee7d7874e4ec9ec59fb2c8945b3e01"
 )
 
-func testKeyManager(t *testing.T) spectypes.KeyManager {
+func testKeyManager(t *testing.T, network *networkconfig.NetworkConfig) spectypes.KeyManager {
```
y0sher, line 38: This is never used, network is always `nil`

olegshmuelov: there is an option to pass the network in order to simulate the current slot.
will be relevant for future slashing protection tests.

PR #1146, opened 2023-09-19
Standalone Slashing Protection Database

olegshmuelov:
This PR introduces a significant update to the SSV node by decoupling the slashing protection data from the node's main database. Previously, the slashing protection data was stored within the node's database, making it susceptible to data loss in case of database removal or migration. With this update, the slashing protection data is now stored in a separate BadgerDB instance, enhancing its persistence and robustness.

### 1. New Configuration Parameter

A new configuration parameter has been added to specify the path for the slashing protection database. This allows for greater flexibility and control over where the slashing protection data is stored.

### 2. Separate BadgerDB Instance

A dedicated BadgerDB instance is now used for storing slashing protection data. The `SyncWrites` option is set to `true`, ensuring that all writes to the database are immediately flushed to disk. This enhances the reliability of the slashing protection mechanism.

### 3. Modified Node Startup Process

The node startup process has been enhanced to include a verification step that checks the consistency and presence of slashing protection data for each signer account. The `MatchSlashingProtectionData` function iterates through all accounts, ensuring that the SP data for them are present and valid. If any of these checks fail, an error is returned, preventing the node from starting. This verification step is crucial for ensuring the integrity of the slashing protection mechanism, particularly across restarts, thereby minimizing the risk of data loss and potential slashing events.

### 4. Database Migration

A database migration process has been implemented to transfer existing slashing protection data from the node's main database to the new dedicated slashing protection database. This ensures that existing data is not lost during this transition, and allows for a smooth upgrade experience for existing node operators.


cli/operator/create_slashing_protection_db.go
```diff
@@ -0,0 +1,151 @@
+package operator
```
y0sher, line 1: @moshe-blox Running a CLI command as a prerequisite sounds pretty harsh to me in terms of UX. maybe we can generate it on our on if it doesn't exist? of course we'll never want to overwrite existing db. so either newly create or point by cli param? IMO we need some more thought on this.

cli/operator/node.go
```diff
@@ -60,9 +60,16 @@ type KeyStore struct {
...
+type DBOptions struct {
+	Path       string        `yaml:"Path" env:"DB_PATH" env-default:"./data/db" env-description:"Path for storage"`
+	Reporting  bool          `yaml:"Reporting" env:"DB_REPORTING" env-default:"false" env-description:"Flag to run on-off db size reporting"`
+	GCInterval time.Duration `yaml:"GCInterval" env:"DB_GC_INTERVAL" env-default:"6m" env-description:"Interval between garbage collection cycles. Set to 0 to disable."`
+}
+
 type config struct {
 	global_config.GlobalConfig `yaml:"global"`
-	DBOptions                  basedb.Options                   `yaml:"db"`
+	DBOptions                  DBOptions                        `yaml:"db"`
```
y0sher, line 71: can't we just use `basedb.Options`?

olegshmuelov: The introduction of the new `DBOptions` struct serves to segregate the configurations for the node database and the slashing protection database. This separation is essential as `basedb.Options` could be employed for either database. By having a distinct `DBOptions` struct, we can maintain clearer and more organized configurations in our yaml/environment settings specifically for the slashing protection and node databases.

ekm/slashing_protection_storage.go
```diff
@@ -0,0 +1,185 @@
...
+	highestAttPrefix      = prefix + "highest_att-"
+	highestProposalPrefix = prefix + "highest_prop-"
+
+	// versionPrefix is the version prefix key of slashing protection DB
+	versionPrefix = "version"
+	// GenesisVersion is the genesis version of the slashing protection DB
+	GenesisVersion = "0x0"
+)
+
+type SpStorage interface {
```
y0sher, line 27: usually abbreviations are written with all caps, like DB, so you could do SPStorage

ekm/slashing_protection_storage.go
```diff
@@ -0,0 +1,185 @@
...
+		return "", found, errors.Wrap(err, "failed to get version")
+	}
+	if !found {
+		return "", found, nil
+	}
+
+	return string(obj.Value), found, nil
+}
+
+func NewSlashingProtectionStorage(db basedb.Database, logger *zap.Logger, prefix []byte) SpStorage {
```
y0sher, line 69: I don't see a point returning the interface here. you could just make the struct be public, and set any functions you don't want used as private.

ekm/slashing_protection_storage.go
```diff
@@ -0,0 +1,185 @@
...
+func (s *spStorage) RemoveHighestAttestation(pubKey []byte) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	return s.db.Delete(s.objPrefix(highestAttPrefix), pubKey)
+}
+
+func (s *spStorage) SaveHighestProposal(pubKey []byte, slot phase0.Slot) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
y0sher, line 139: I don't get why do we need mutex here, I'm assuming the DB is thread safe. anyway if we decide to use it anyway you should wrap only the critical section (database write/read) in mutex. the validations and marshaling shouldn't be serialized.

If there was a write here that is based on a read from the database, I would understand the mutex used wanting to lock these actions togteher. but I can't see anything like this here.

ekm/storage.go
```diff
@@ -0,0 +1,31 @@
...
+// Storage represents the interface for ssv node storage
+type Storage interface {
+	SignerStorage
+	SpStorage
+}
+
+type storage struct {
+	SignerStorage
+	SpStorage
+}
```
y0sher, line 23: these are the same, why can't we just leave `storage`, we can change the name to `Storage`

migrations/migration_4_standalone_slashing_data.go
```diff
@@ -0,0 +1,69 @@
...
+				highAtt, found, err := legacySPStorage.RetrieveHighestAttestation(sharePubKey)
+				if err != nil {
+					return err
+				}
+				if !found {
+					return fmt.Errorf("highest attestation not found for share %s", hex.EncodeToString(sharePubKey))
+				}
+				if highAtt == nil {
+					return fmt.Errorf("highest attestation is nil for share %s", hex.EncodeToString(sharePubKey))
+				}
```
y0sher, line 39: @olegshmuelov , @moshe-blox  in case of error, shouldn't we skip `continue` and try next account?

y0sher (changes requested):
Generally good job! code looks good. some comments left :)

migrations/migration_4_standalone_slashing_data.go
olegshmuelov: No, in this case, an error signifies a critical issue with the slashing data for the specific share. Automatically moving to the next account without addressing the error could potentially lead to incomplete or incorrect data migration, which is why manual intervention is preferred.

ekm/storage.go
olegshmuelov: The `NewEKMStorage` function is designed to return interface, as it interacts with an external package, especially when setting the storage with `options.SetStorage(ekmStorage)`

ekm/slashing_protection_storage.go
olegshmuelov: pls see [comment](https://github.com/bloxapp/ssv/pull/1146/files/465528b438cf08d623bcf426fa06195dc0e19610#r1342156357)
in addition to my previous comment I think it may offer more level of abstraction and testability.
But still if u insist we can change it

ekm/slashing_protection_storage.go
olegshmuelov: This code was not introduced in this pull request but was moved from another file. The mutex usage here was from the original design for ensuring atomic operations with the database. Your point on limiting the scope of mutex lock is valid so we can discuss it further. @moshe-blox wdyt?

cli/operator/node.go
y0sher: honestly I don't understand your point. Since the struct is similar, I don't see how is this better than what we had before, maybe I'm missing something. can you explain further ?

olegshmuelov: Our node should handle two different databases - one for the node and one for slashing protection. Each of these databases has its own set of configurations. 

The `basedb.Options` struct could technically be used for both, but to keep configurations separate and self-explanatory, I introduced two new structs - `DBOptions` for the node database and `SlashingProtectionOptions` for the slashing protection database.

Here’s an example with the `SlashingProtectionOptions` struct:
```go
type SlashingProtectionOptions struct {
    DBPath string yaml:"DBPath" env:"SP_DB_PATH" env-description:"Path for slashing protection db"
}
```
This struct allows us to define a specific path for the slashing protection database through yaml or environment variables, making it explicit what this configuration is for. 

Similarly, `DBOptions` is intended for configuring the node database. By having separate structs, we can avoid any potential confusion or misconfiguration that might arise from using a single struct for both databases. It also provides a clearer, more organized way to handle configurations for each database independently, making the code more readable and maintainable.

cli/operator/node.go
```diff
@@ -351,25 +364,31 @@ func setupGlobal(cmd *cobra.Command) (*zap.Logger, error) {
...
-func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.BadgerDB, error) {
-	db, err := kv.New(logger, cfg.DBOptions)
+func setupDB(ctx context.Context, logger *zap.Logger, network beaconprotocol.BeaconNetwork, spDB *kv.BadgerDB) (*kv.BadgerDB, error) {
+	options := basedb.Options{
+		Path:       cfg.DBOptions.Path,
+		Reporting:  cfg.DBOptions.Reporting,
+		GCInterval: cfg.DBOptions.GCInterval,
+	}
+
+	db, err := kv.New(ctx, logger, options)
```
nkryuchkov, line 374: As far as I know, passing context to a constructor is not a common pattern unless it's something specific. Instead, context is usually passed to methods that may be canceled. Could you please explain why we need to pass it here?

cli/operator/node.go
```diff
@@ -402,6 +421,38 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to open slashing protection DB: %w", err)
+	}
+
+	// If the sp db was created with cli command, it should include "genesis" version of the db
+	// this way we can validate the db was created via the cli command. if not, we should fail
+	storage := ekm.NewSlashingProtectionStorage(db, logger, []byte(network.GetBeaconNetwork()))
+	value, found, err := storage.GetVersion()
+	if err != nil || !found || value != ekm.GenesisVersion {
+		return nil, fmt.Errorf("slashing protection database was not created via cli command, please create it using the 'create-slashing-protection-db' command: %w", err)
```
nkryuchkov, line 450: I think a separate error for cases if `err` is `nil` would look better and would help understand what exactly is wrong

ekm/slashing_protection_storage.go
```diff
@@ -0,0 +1,186 @@
...
+	if pubKey == nil {
+		return errors.New("pubKey must not be nil")
+	}
+
+	if slot == 0 {
+		return errors.New("invalid proposal slot, slot could not be 0")
+	}
+
+	var data []byte
+	data = ssz.MarshalUint64(data, uint64(slot))
```
nkryuchkov, line 151:
```suggestion
	data := ssz.MarshalUint64(data, uint64(slot))
```

olegshmuelov: should be
```
data := ssz.MarshalUint64(nil, uint64(slot))
```

nkryuchkov (dismissed):

nkryuchkov (approved):

migrations/migration_4_standalone_slashing_data.go
```diff
@@ -0,0 +1,67 @@
...
+				if !found {
+					return fmt.Errorf("highest proposal not found for share %s", hex.EncodeToString(sharePubKey))
+				}
+				if highProposal == 0 {
+					return fmt.Errorf("highest proposal is 0 for share %s", hex.EncodeToString(sharePubKey))
+				}
+
+				if err := spStorage.SaveHighestProposal(sharePubKey, highProposal); err != nil {
+					return fmt.Errorf("failed to save highest proposal for share %s: %w", hex.EncodeToString(sharePubKey), err)
+				}
```
moshe-blox, line 60: Do you think it can be helpful?
```go
				// Ensure the data is saved and can be read.
				highestAttestation, found, err := spStorage.RetrieveHighestAttestation(sharePubKey)
				if err != nil {
					return err
				}
				if !found || !reflect.DeepEqual(highestAttestation, highAtt) {
					return fmt.Errorf("highest attestation not found for share %s", hex.EncodeToString(sharePubKey))
				}
```
(also for proposal)

moshe-blox (changes requested):

cli/operator/create_slashing_protection_db.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		options := basedb.Options{
+			Path:       dbPath,
+			SyncWrites: true,
+		}
+		db, err := kv.New(cmd.Context(), nil, options)
+		if err != nil {
+			logger.Panic("failed to create slashing protection db", zap.Error(err))
+		}
```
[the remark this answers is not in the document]
olegshmuelov: added IsEmpty() method to badgerDB

moshe-blox (dismissed):

cli/operator/node.go
y0sher: I'm ok with this approach but IMO this isn't a lot better than just having the same struct type used twice for different purposes.

y0sher (dismissed):

cli/operator/create_slashing_protection_db.go
y0sher: So if we're going through with this, this should be represented in our README and docs. @olegshmuelov , @moshe-blox , @lior-blox , @arielssv
we should probably have a diagram or something about how do we want to act in any of the cases (db exists, db not exist and more) also from product perspective ( @arielssv )and make sure our code is aligned with it.

y0sher (approved):

y0sher (dismissed):

y0sher:
@moshe-blox please look

PR #1149, opened 2023-09-20
New slot ticker

MatusKysel:
New feed-less, gorountine-less slot ticker that is based on golang Timer

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_COMMIT_SHA $DOCKER_REPO_INFRA_STAGE:$CI_COMMIT_SHA
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_COMMIT_SHA
   only:
-    - new-slot-ticker
+    - stage
 
 Deploy nodes to stage:
   stage: deploy
@@ -58,7 +58,7 @@ Deploy nodes to stage:
     # +--------------------+
     # |  Deploy SSV nodes  |
     # +--------------------+
-    # - .k8/stage/scripts/deploy-cluster-1--4.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
+    - .k8/stage/scripts/deploy-cluster-1--4.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
     - .k8/stage/scripts/deploy-cluster-5--8.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
     # - .k8/stage/scripts/deploy-cluster-9--12.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
     # - .k8/stage/scripts/deploy-cluster-13--16.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE $SSV_NODES_CPU_LIMIT $SSV_NODES_MEM_LIMIT
@@ -69,7 +69,7 @@ Deploy nodes to stage:
     # █▓▒░ Keep commented unless you're testing the bootnode ░▒▓█
     # - .k8/stage/scripts/deploy-boot-nodes.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE blox-infra-stage kubernetes-admin@blox-infra stage.ssv.network $K8S_API_VERSION $STAGE_HEALTH_CHECK_IMAGE 1000m 1000m
   only:
-    - new-slot-ticker
+    - stage
 
 Deploy exporter to stage:
   stage: deploy


operator/fee_recipient/controller_test.go
```diff
@@ -52,32 +52,47 @@ func TestSubmitProposal(t *testing.T) {
...
+			2,                                        // should not call submit
+			20,                                       // should not call submit
+			phase0.Slot(network.SlotsPerEpoch()) / 2, // halfway through epoch
+			63,                                       // should not call submit
+		}
+
+		for _, s := range slots {
+			mockTimeChan <- time.Now()
+			mockSlotChan <- s
+			time.Sleep(time.Millisecond * 500)
```
nkryuchkov, line 89: Can the duration be reduced? I think it might be too much for the test unless there's any reason for this (let's add a comment in this case)

operator/slot_ticker/slotticker.go
```diff
@@ -6,83 +6,82 @@ import (
...
+		select {
+		case <-s.timer.C:
+		default:
 		}
-	}()
+	}
+	slotNumber := uint64(timeSinceGenesis / s.slotDuration)
+	nextSlotStartTime := s.genesisTime.Add(time.Duration(slotNumber+1) * s.slotDuration)
+	s.timer.Reset(time.Until(nextSlotStartTime))
+	s.slot = phase0.Slot(slotNumber + 1)
```
nkryuchkov, line 81: If `Next` and `Slot` OR `Next` and `Next` are called simultaneously, there might be a risk of read/write or write/write data race here.

operator/fee_recipient/controller_test.go
nkryuchkov: Oh, I didn't notice that. Then I agree it's fine to keep the behavior

operator/slot_ticker/slotticker.go
nkryuchkov: Well, I think we could keep it as is if we make sure that `Next` and `Slot` are used correctly from the outside, but it would be better to add a comment for it

beacon/goclient/goclient.go
```diff
@@ -147,7 +147,7 @@ type goClient struct {
 }
 
 // New init new client and go-client instance
-func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.OperatorID, slotTicker slot_ticker.Ticker) (beaconprotocol.BeaconNode, error) {
+func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.OperatorID, slotTickerProvider slot_ticker.SlotTickerFactory) (beaconprotocol.BeaconNode, error) {
```
nkryuchkov, line 150: What do you think about `slot_ticker.SlotTickerProvider` for consistency with the variable name? That would also get rid of the Java-like `Factory` suffix which is usually not used in Go

nkryuchkov (dismissed):
Great work! Approving. Just one minor question

nkryuchkov (approved):

operator/slot_ticker/slotticker.go
```diff
@@ -6,83 +6,90 @@ import (
...
 )
 
-// The TTicker interface defines a type which can expose a
-// receive-only channel firing slot events.
-type TTicker interface {
-	C() <-chan phase0.Slot
-	Done()
+//go:generate mockgen -package=mocks -destination=./mocks/slotticker.go -source=./slotticker.go
+
+type SlotTickerProvider func() SlotTicker
```
y0sher, line 11: Naming wise, I would call this just `Provider` since it will be used in external packages as `slot_ticker.Provider`, and `slot_ticker.SlotTickerProvider` will be really long and repetitive, which isn't needed.

operator/slot_ticker/slotticker.go
```diff
@@ -6,83 +6,90 @@ import (
...
-// SlotTicker is a special ticker for the beacon chain block.
-// The channel emits over the slot interval, and ensures that
-// the ticks are in line with the genesis time. This means that
-// the duration between the ticks and the genesis time are always a
-// multiple of the slot duration.
-// In addition, the channel returns the new slot number.
-type SlotTicker struct {
-	c    chan phase0.Slot
-	done chan struct{}
+type SlotTickerConfig struct {
```
y0sher, line 23: same comment, this could be just `Config` so end of the day it is used as `slot_ticker.Config`, more compact and more readable.

operator/slot_ticker/slotticker.go
```diff
@@ -6,83 +6,90 @@ import (
...
-	if genesisTime.IsZero() {
-		panic("zero genesis time")
+type slotTicker struct {
+	timer        *time.Timer
+	slotDuration time.Duration
+	genesisTime  time.Time
+	slot         phase0.Slot
+}
+
+func NewSlotTicker(cfgProvider ConfigProvider) SlotTicker {
```
y0sher, line 43: `New` is probably enough here.

y0sher: it's a best practice in go to return concrete types. https://duncanleung.com/go-idiom-accept-interfaces-return-types/

operator/slot_ticker/slotticker.go
nkryuchkov: Let's maybe also rename `slot_ticker` as well as underscore it usually not used in package names?

operator/duties/scheduler_test.go
```diff
@@ -17,35 +17,79 @@ import (
...
 	*zap.Logger,
-	*mockSlotTicker,
+	*mockSlotTickerService,
 	time.Duration,
 	context.CancelFunc,
 	*pool.ContextPool,
 ) {
 	ctrl := gomock.NewController(t)
-	timeout := 100 * time.Millisecond
+	timeout := 200 * time.Millisecond
```
nkryuchkov, line 85: Why do we need this change? Has the timing behavior changed? If so, I think we need to comment it

nkryuchkov (dismissed):

operator/slotticker/slotticker_test.go
moshe-blox, line 1: Very well tested 👌

I'd only add one important test for slot skipping.

For example, a SlotTicker for-loop with some chance for the iteration to take longer than the duration of the slot, and a confirmation that the ticker skipped the slot and sent the next one roughly on-time.

moshe-blox (changes requested):
Good job 👍
Left a few comments

beacon/goclient/proposer.go
```diff
@@ -230,9 +231,11 @@ func (gc *goClient) createValidatorRegistration(pubkey []byte, feeRecipient bell
...
 
-func (gc *goClient) registrationSubmitter(slots <-chan phase0.Slot) {
-	for currentSlot := range slots {
-		gc.submitRegistrationsFromCache(currentSlot)
+func (gc *goClient) startSlotTicker(slotTickerProvider slotticker.Provider) {
+	ticker := slotTickerProvider()
+	for {
+		<-ticker.Next()
+		gc.submitRegistrationsFromCache(ticker.Slot())
 	}
```
y0sher, line 239: I would also check for `ctx.Done()` here using select. since we have ctx in `goClient`. this is only for node shutown, but still I'm not a fan of never-ending loops.

operator/duties/validatorregistration.go
```diff
@@ -36,10 +36,11 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
 		case <-ctx.Done():
 			return
 
-		case slot := <-h.ticker:
+		case <-h.ticker.Next():
 			shares := h.validatorController.GetOperatorShares()
-
+			slot := h.ticker.Slot()
```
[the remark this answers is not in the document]
y0sher: either that or make the slot return with Next() and then just 
`slot := <-h.ticker.Next()`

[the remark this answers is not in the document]
y0sher: After reading the code of `Next` I understand the logic since `Next`'s channel is actually waiting for the slot to arrive. we could still pack the logic to one func that will wait and return a chan with the slot. (it'll still be a chan so you could select on it.)

y0sher (approved):
Well written and commented, added some comments.

cli/operator/node.go
```diff
@@ -214,7 +214,9 @@ var StartNodeCmd = &cobra.Command{
 		cfg.SSVOptions.ValidatorController = validatorCtrl
 		cfg.SSVOptions.Metrics = metricsReporter
 
-		operatorNode = operator.New(logger, cfg.SSVOptions, slotTicker)
+		operatorNode = operator.New(logger, cfg.SSVOptions, func() slotticker.SlotTicker {
```
nkryuchkov, line 217: What do you think of extracting the anonymous function to a variable (e.g. `slotTickerProvider`)? That would make places where it's passed easier to read, remove code duplication and the variable name would show what's the purpose of this function

operator/slotticker/slotticker.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type slotTicker struct {
+	timer        *time.Timer
+	slotDuration time.Duration
+	genesisTime  time.Time
+	slot         phase0.Slot
+}
+
+// `New` returns a goroutine-free `SlotTicker` implementation which is not thread-safe.
```
nkryuchkov, line 43: I have never seen function names wrapped in backticks in Godoc comments, so I'm not sure how it would be rendered and whether it would pass linter that checks comments for exported functions/methods. 

I ran a search for "// \`" in all SSV dependencies and didn't find any places where it's used in the beginning of a Godoc comment.

If backticks have any advantages, I don't mind, but in this case I think we need to discuss them, agree on them and to enforce this rule in all PRs to be consistent

nkryuchkov (dismissed):

nkryuchkov (dismissed):

y0sher (dismissed):

nkryuchkov (approved):

y0sher (approved):

operator/slotticker/slotticker.go
moshe-blox: mb, it shouldn't begin with a backtick

moshe-blox (changes requested):
I really liked the slot skipping test 👌

Left only 1 question.

nkryuchkov (approved):

y0sher (approved):

operator/duties/scheduler.go
```diff
@@ -20,19 +22,34 @@ import (
...
 	"github.com/bloxapp/ssv/networkconfig"
 	"github.com/bloxapp/ssv/operator/duties/dutystore"
+	"github.com/bloxapp/ssv/operator/slotticker"
 	"github.com/bloxapp/ssv/protocol/v2/types"
 )
 
 //go:generate mockgen -package=mocks -destination=./mocks/scheduler.go -source=./scheduler.go
 
+var slotDelayHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
+	Name:    "slot_ticker_delay_milliseconds",
```
nkryuchkov, line 32: According to [Prometheus naming guidelines](https://prometheus.io/docs/practices/naming/):

> ...must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
> ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.

Also, other metrics in the project use the `_seconds` suffix, so we might need to consider making them consistent. On the other hand, I don't foresee any issues if we don't change it.

nkryuchkov (approved):

operator/duties/scheduler.go
```diff
@@ -20,19 +22,34 @@ import (
...
+	Help:    "The delay in milliseconds of the slot ticker",
+	Buckets: []float64{5, 10, 20, 100, 500, 5000}, // Buckets in milliseconds. Adjust as per your needs.
+})
+
+func init() {
```
nkryuchkov, line 37: I think we need to add new metrics to `MetricsReporter` and pass them instead of creating them in `init` for new packages

nkryuchkov: okay, I see it's still something to be discussed, so let's not make the PR depend on this

PR #1157, opened 2023-10-11
feat: ignore messages during handshake

moshe-blox:


message/validation/validation.go
```diff
@@ -252,6 +261,15 @@ func (mv *messageValidator) ValidatePubsubMessage(_ context.Context, _ peer.ID,
 		mv.metrics.MessageValidationDuration(sinceStart, validationDurationLabels...)
 	}()
 
+	// Ignore messages from peers that are not connected.
+	// TODO:
+	// - pass the existing PeerInfoIndex to NewMessageValidator
+	// - add a separate metric to track # of messages ignored here
+	// - add unit tests with a PeerInfoIndex mock
+	if mv.peersIndex != nil && mv.peersIndex.State(pmsg.GetFrom()) != peers.StateConnected {
```
y0sher, line 269: is this what we wanted though? I'm not sure I understand, this is pubsub, we don't get messages from only peers we're connected to.

y0sher: well, this kinda works but I think we better find a way to actually not forward messages into the gossip and msg validation pipeline of peers that haven't gone through the handshakes.

PR #1163, opened 2023-10-17
P2P metrics 

AKorpusenko:


network/topics/metrics.go
```diff
@@ -24,6 +24,12 @@ var (
 		Name: "ssv:p2p:pubsub:score:inspect",
 		Help: "Gauge for negative peer scores",
 	}, []string{"pid"})
+
+	// score.go invalidMessageDeliveries value per topic
```
nkryuchkov, line 28: I think we can remove this. `score.go` looks like a file from libp2p but it can be moved in a next version and there may be several "score.go"'s there. It's easy to follow what's passed to `metricPubSubPeerP4Score`

lior-blox:
@AKorpusenko @MatusKysel please include @zktaiga as a reviewer once you ready

monitoring/metricsreporter/metrics_reporter.go
```diff
@@ -137,13 +139,21 @@ var (
 		Name: "ssv_message_non_committee",
 		Help: "The amount of messages not in committee",
 	}, []string{"ssv_msg_type", "decided"})
+	messagesReceivedFromPeer = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "ssv_messages_received_from_peer",
+		Help: "The amount of messages received from the specific peer",
+	}, []string{"peer_id"})
```
moshe-blox, line 145: Recently we merged a PR to reduce the metrics size (#1184) by removing high cardinality labels such as validator public keys.

I think `peer_id` is a similar issue, as it can potentially have an infinite number of unique values over time as we discover new peers. Unless you think we need to query this by peer, and it's possible to remove a peer from this metric after disconnecting from it, I suggest to remove the `peer_id` label.

monitoring/metricsreporter/metrics_reporter.go
```diff
@@ -188,6 +200,12 @@ func New(opts ...Option) *MetricsReporter {
 		}
 	}
 
+	// reset  metric every hour because it can grow infinitely
+	async.Interval(ctx, 8*time.Hour, func() {
```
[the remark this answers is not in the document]
moshe-blox: Rather than reseting the metrics (which might disrupt the Grafana chart), I think we can just call `DeleteLabelValues` on a `peer_id` after we've disconnected from it. That way, you'd only have 60 (max peers) different values top per metric.

moshe-blox (changes requested):

monitoring/metricsreporter/metrics_reporter.go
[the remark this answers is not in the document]
moshe-blox: Rather than resetting the metrics (which might disrupt the Grafana chart), I think we can just call DeleteLabelValues on a peer_id after we've disconnected from it. That way, you'd only have 60 (max peers) different values top per metric.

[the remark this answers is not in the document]
zktaiga: @nkryuchkov Infinitely growing values (a counter) for a given metric is fine. What we are discussing here as an issue is an infinitely growing amount of labels, e.g.
```
metric{key="peer1"} 0
metric{key="peer2"} 0
...
```

Prometheus doesn't deal well with these case scenarios at large scale.

zktaiga (commented):
Agree with @moshe-blox - I think it's a great use case for `DeleteLabelValues()`. It allows us to have fixed and predictable cardinality. 

I think where possible we should avoid the `Reset()` pattern on a predefined interval, today an interval might seem reasonable but produce a lot of noise in the future.

The rest looks good to me!

GalRogozinski (approved):

zktaiga (approved):
@GalRogozinski I don't see an issue as long as we have verified that the cardinality is of a fixed `peerCount` length.

I'd encourage to create a flag for high-cardinality data though, e.g. `peerID`/`pubKey` though, just not sure that it is in scope for this PR. 

lgtm otherwise

MatheusFranco99 (dismissed):

GalRogozinski (dismissed):

network/topics/scoring.go
```diff
@@ -66,6 +70,8 @@ func scoreInspector(logger *zap.Logger, scoreIdx peers.ScoreIndex) pubsub.Extend
 			logger.Debug("peer scores", fields...)
 
 			metricPubsubPeerScoreInspect.WithLabelValues(pid.String()).Set(peerScores.Score)
+			metricPubSubPeerP4Score.WithLabelValues(pid.String()).Set(p4ScoreSquaresSum)
```
moshe-blox, line 73: This one is not being reset like the one above, so it means it can grow infinitely no?

moshe-blox (changes requested):

network/topics/metrics.go
```diff
@@ -24,20 +24,33 @@ var (
...
-	if err := prometheus.Register(metricPubsubOutbound); err != nil {
-		logger.Debug("could not register prometheus collector")
-	}
-	if err := prometheus.Register(metricPubsubInbound); err != nil {
-		logger.Debug("could not register prometheus collector")
+
+	allMetrics := []prometheus.Collector{
+		metricPubsubTrace,
+		metricPubsubOutbound,
+		metricPubsubInbound,
```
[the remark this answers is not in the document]
moshe-blox: @zktaiga i agree, this should probably be in a separate PR

protocol/v2/ssv/runner/metrics/metrics.go
```diff
@@ -125,6 +138,7 @@ func (cm *ConsensusMetrics) EndPreConsensus() {
 func (cm *ConsensusMetrics) StartConsensus() {
 	if cm != nil {
 		cm.consensusStart = time.Now()
+		cm.metricsDutiesCreated.Inc()
```
GalRogozinski, line 141: just saying that "fake duties" like validator registration or volunary exit won't be measured by this.

Also unless we know postconsensus was successful then we don't know if it was successful or not.

Since we use e2m anyhow for most duties we can simply change the title in grafana from "duties" to "consensus instances" to not confuse the user

moshe-blox: @GalRogozinski the question is which metric do we prefer? would you like to see all duties created or just consensus?

moshe-blox:
Made a few changes:
- Inspect scores and update metrics every minute rather than 5 minutes, but log every 5 minutes still
- Renamed metrics from "duties" to "instances" (because it only tracks QBFTs and not ValidatorRegistration duties, etc..)
- Added "role" label to instances started/decided metrics

monitoring/metricsreporter/metrics_reporter.go
```diff
@@ -152,14 +148,71 @@ var (
 		Name: "ssv_message_validation_rsa_checks",
 		Help: "The amount message validations",
 	}, []string{})
+	pubsubPeerScore = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ssv:p2p:pubsub:score:inspect",
+		Help: "Pubsub peer scores",
+	}, []string{"pid"})
+	pubsubPeerP4Score = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ssv:p2p:pubsub:score:invalid_message_deliveries",
+		Help: "Pubsub peer P4 scores (for invalid message deliveries)",
```
GalRogozinski, line 157: Maybe this is better:

```suggestion
                Help: "Pubsub peer P4 scores (sum of square of counters for invalid message deliveries)",
```

better do this in IDE cause I added spaces manually

moshe-blox: @GalRogozinski done, thx

AKorpusenko:
@moshe-blox 
I've fixed the charts for QBFT instances. Now we can see the start / decided instances counters splitted by roles 
image

moshe-blox:
> @moshe-blox I've fixed the charts for QBFT instances. Now we can see the start / decided instances counters splitted by roles image

@AKorpusenko  Very nice, thanks!

protocol/v2/types/crypto.go
```diff
@@ -7,14 +7,32 @@ import (
...
+	}
+}
+
 // VerifyByOperators verifies signature by the provided operators
 // This is a copy of a function with the same name from the spec, except for it's use of
 // DeserializeBLSPublicKey function and bounded.CGO
 //
 // TODO: rethink this function and consider moving/refactoring it.
 func VerifyByOperators(s spectypes.Signature, data spectypes.MessageSignature, domain spectypes.DomainType, sigType spectypes.SignatureType, operators []*spectypes.Operator) error {
+	metricsSignaturesVerifications.WithLabelValues().Inc()
```
y0sher, line 35: good metric, though is this the only place we verify sigs?

y0sher (dismissed):
Approving this regardless of all my comments, I think we need a overhaul to how we do metrics codebase-wide, so this isn't the place to treat all of that. regarding what this PR is meant to handle I think it looks OK.

monitoring/metricsreporter/metrics_reporter.go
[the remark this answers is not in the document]
moshe-blox: as we discussed over DIscord, this will be done in a future PR to refactor metrics

monitoring/metricsreporter/metrics_reporter.go
```diff
@@ -137,14 +136,84 @@ var (
...
+		Help: "Pubsub peer scores",
+	}, []string{"pid"})
+	pubsubPeerP4Score = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ssv:p2p:pubsub:score:invalid_message_deliveries",
+		Help: "Pubsub peer P4 scores (sum of square of counters for invalid message deliveries)",
+	}, []string{"pid"})
 )
 
-type MetricsReporter struct {
+type MetricsReporter interface {
```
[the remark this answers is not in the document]
moshe-blox: i agree, i think this interface should be split to multiple interfaces, one in each package where it belongs

and that package can just expose the implementation or maybe we can get rid of this package altogether and do the implementation inside each respective package

protocol/v2/types/crypto.go
moshe-blox: @GalRogozinski do we need this to track other signature verification methods as well?

GalRogozinski: We also track RSA sigs in another metric

Are there more places where we do BLS verifications?
This looks like the main function to me

GalRogozinski: well I guess better to add

y0sher (approved):

PR #1175, opened 2023-10-24
Dont send recv msg before handshake

pavelkrolevets:
- added connection gater to p2p options
- now we can to add CIDR subnet for allowing/denying only certain peer connections

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.2
 	github.com/herumi/bls-eth-go-binary v1.29.1
 	github.com/ilyakaznacheev/cleanenv v1.4.2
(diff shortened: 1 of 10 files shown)


pavelkrolevets:
@y0sher is this somenthing you were aiming for?

lior-blox:
@pavelkrolevets @y0sher 
Is it a duplicate with https://github.com/bloxapp/ssv/pull/1157

y0sher:
> @pavelkrolevets @y0sher Is it a duplicate with #1157

@lior-blox , not really, both are trying to achieve same thing though. look at my comment on #1157

@pavelkrolevets , this PR is missing something critical, the gating should gate using our application specific handshake. it can be found under `network/peers/connections`.

PR #1145, opened 2023-09-19
Refactored Monitoring guide

RaekwonIII:
Added simpler Grafana config files

PR #1176, opened 2023-10-27
Fix dashboards and refactor docs

zktaiga:
Currently the dashboards are not generic, they have hardcoded instances as well as datasources. 

This PR makes it so that:
- Datasources are inferred from the Grafana environment, as well as using the datasource dynamically across all Prometheus calls.  
- Instances are inferred from an `instance` label in the `ssv_beacon_status` metric. This is currently k8s specific but we can refactor this in the future so that we take into consideration non-k8s but dockerized environments.


lior-blox:
@zktaiga I changed base branch to stage

zktaiga:
@RaekwonIII @lior-blox This PR contains cherry-picked commits from #1145 + relevant modifications. These are changes  with significant QoL improvements. We can iterate on them and add other improvements as part of subsequent PRs.

RaekwonIII (approved):

systemblox (approved):

PR #1178, opened 2023-10-27
fix: usage of prater in `NewValidatorRegistrationRunner`

moshe-blox:


operator/duties/validatorregistration.go
```diff
@@ -66,10 +57,11 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
...
 				}})
 
-				sent++
-				h.validatorsPassedFirstRegistration[string(share.ValidatorPubKey)] = struct{}{}
+				validators = append(validators, share.BeaconMetadata.Index)
 			}
-			h.logger.Debug("validator registration duties sent", zap.Uint64("slot", uint64(slot)), fields.Count(sent))
+			h.logger.Debug("validator registration duties sent",
+				zap.Uint64("slot", uint64(slot)),
+				zap.Any("validators", validators))
```
lior-blox, line 64: I think we probably should revert this change.
not print all validators. just the count

PR #1185, opened 2023-10-30
P2p enc+metrics

MatusKysel:


network/peers/conn_manager.go
```diff
@@ -74,6 +76,7 @@ func (c connManager) TrimPeers(ctx context.Context, logger *zap.Logger, net libp
 		if !c.connManager.IsProtected(pid, protectedTag) {
 			err := net.ClosePeer(pid)
 			logger.Debug("closing peer", zap.String("pid", pid.String()), zap.Error(err))
+			metrics.DeletePeerInfo(pid)
```
moshe-blox, line 79: this means we only evict when we're the ones closing the connection, but not the other way around, so the metric might keep growing indefinitely

i guess the right place to put this is in `conn_handler.go` in the disconnection callback, unless we'd like to remember peers for a bit after we've disconnected (for debugging), in which case a more complex eviction mechanism is needed

protocol/v2/qbft/controller/controller.go
```diff
@@ -71,6 +84,7 @@ func (c *Controller) StartNewInstance(logger *zap.Logger, height specqbft.Height
 	newInstance := c.addAndStoreNewInstance()
 	newInstance.Start(logger, value, height)
 	c.forceStopAllInstanceExceptCurrent()
+	metricsDutiesCreated.WithLabelValues().Inc()
```
moshe-blox, line 87: not sure if this is just a test or not, but we already have similar metrics inside each Runner (see `r.metrics.Consensus()` for starting instance and `r.metrics.EndConsensus()` for decided)

i'd check if it's possible to work with the existing implementation of Runner metrics instead of adding more metric calls

network/p2p/p2p.go
```diff
@@ -80,7 +82,7 @@ type p2pNetwork struct {
 }
 
 // New creates a new p2p network
-func New(logger *zap.Logger, cfg *Config) network.P2PNetwork {
+func New(logger *zap.Logger, cfg *Config, mr *metricsreporter.MetricsReporter) network.P2PNetwork {
```
moshe-blox, line 85: instead of requiring a whole `MetricsReporter` implementation, i suggest to require instead a package-level `Metrics` interface that includes only the methods that `p2pNetwork` needs to call

that's to be consistent with other packages (such as `protocol/v2/ssv/queue`)

moshe-blox (commented):
(i know this is still a draft, just left a few comments in the meantime)

PR #1191, opened 2023-11-02
Use generic name for metrics port

zktaiga:


.k8/hetzner-stage/ssv-node-44-deployment.yml
```diff
@@ -48,86 +48,90 @@ spec:
...
+        - name: ssv-node-44
+          persistentVolumeClaim:
+            claimName: ssv-node-44
+        - name: ssv-node-44-cm
+          configMap:
+            name: ssv-node-44-cm
       hostNetwork: true
+  ports:
+    - null
+    - null
```
y0sher, line 136: care to explain the `null`s?

.gitlab-ci.yml
y0sher, line 1: I guess you'd revert this file before merging ?

.k8/hetzner-stage/ssv-node-44-deployment.yml
zktaiga: Ah nice catch, wrote a script and didn't realize there were 2 specs per file. Should be good now.

.gitlab-ci.yml
zktaiga: If this works as expected I'll revert that commit to apply it to all nodes.

zktaiga:
Diff is very verbose due to YAML linting, but I'm just adding a convention the same named port across all nodes so we can leverage Prometheus CRDs down the line and not have to hardcode ports.

Tested on stage, ready for review. cc @y0sher @nkryuchkov

nkryuchkov (dismissed):

systemblox (dismissed):

systemblox (approved):

PR #1192, opened 2023-11-06
Fix initial metadata fetch

moshe-blox:


operator/validator/controller.go
```diff
@@ -387,12 +387,37 @@ func (c *controller) StartValidators() {
...
 		c.logger.Info("could not find validators")
 		return
 	}
-	c.setupValidators(shares)
+
+	var ownShares []*ssvtypes.SSVShare
+	var allPubKeys = make([][]byte, 0, len(shares))
+	for _, share := range shares {
+		if share.BelongsToOperator(c.GetOperatorData().ID) {
+			ownShares = append(ownShares, share)
```
y0sher, line 400: I would do this change since `GetOperatorData()` uses a mutex. its a read mutex but still I think we don't have to call it everytime.
```suggestion
	var allPubKeys = make([][]byte, 0, len(shares))
	ownOpID := c.GetOperatorData().ID
	for _, share := range shares {
		if share.BelongsToOperator(ownOpID) {
			ownShares = append(ownShares, share)
```

y0sher (approved):

PR #1198, opened 2023-11-07
Ignore messages if there's no duty

nkryuchkov:
- if there's no duty for attester, proposer, sync committee role, ignore message instead of rejecting
- duty is checked before semantic checks


lior-blox:
This is not relevant, closing for now

PR #1200, opened 2023-11-09
Voluntary exit

nkryuchkov:
- Add what's missing in the new runner structure for voluntary exit
- Update the BeaconNode with the VoluntaryExit calls and create a new duty handler for the voluntary exit duty
- Trigger exit upon exit event from the contract
- Voluntary Exit rules  for msg validation
- Trigger exit upon exit event from the contract test


eth/eventhandler/event_handler.go
```diff
@@ -424,46 +458,53 @@ func (eh *EventHandler) processLocalEvent(txn basedb.Txn, event localevents.Even
 	case OperatorAdded:
 		data := event.Data.(contract.ContractOperatorAdded)
 		if err := eh.handleOperatorAdded(txn, &data); err != nil {
-			return fmt.Errorf("handle OperatorAdded: %w", err)
+			return fmt.Errorf("handle %s: %w", event.Name, err)
```
y0sher, line 461: honestly I would keep this PR clean and not do this here, or at least have it in separate commit.

eth/eventhandler/handlers.go
```diff
@@ -469,6 +469,48 @@ func (eh *EventHandler) handleFeeRecipientAddressUpdated(txn basedb.Txn, event *
...
+		return nil, nil
+	}
+
+	pk := phase0.BLSPubKey{}
+	copy(pk[:], share.ValidatorPubKey)
+
+	ed := &ExitDescriptor{
+		PubKey:         pk,
+		ValidatorIndex: share.BeaconMetadata.Index,
+		Slot:           eh.networkConfig.Beacon.EstimatedCurrentSlot(),
```
y0sher, line 508: we need to see how this plays out, what if its the end of the slot? might be ok since it's not full consensus. Also we need to see if the duty slot should be the same as the signing slot, as again, until we finish and publish this slot might be over, so we might want to sign on exit for next slot even though we run the duty in the current slot.

eth/eventhandler/handlers.go
```diff
@@ -318,7 +317,7 @@ func validatorAddedEventToShare(
 	}
 
 	validatorShare.Quorum, validatorShare.PartialQuorum = ssvtypes.ComputeQuorumAndPartialQuorum(len(committee))
-	validatorShare.DomainType = ssvtypes.GetDefaultDomain()
```
y0sher, line 321: good catch 🥇

nkryuchkov: thanks

eth/eventhandler/handlers.go
```diff
@@ -469,6 +468,48 @@ func (eh *EventHandler) handleFeeRecipientAddressUpdated(txn basedb.Txn, event *
 	return r != nil, nil
 }
 
+type ExitDescriptor struct {
+	PubKey         phase0.BLSPubKey
+	ValidatorIndex phase0.ValidatorIndex
+	Slot           phase0.Slot
+}
+
+func (eh *EventHandler) handleValidatorExited(txn basedb.Txn, event *contract.ContractValidatorExited) (*ExitDescriptor, error) {
```
y0sher, line 477: missing in this function: does the address who triggered this tx really the address that registered this validator?

message/validation/consensus_validation.go
y0sher, line 1: where do we check that this validator is active?

nkryuchkov: https://github.com/bloxapp/ssv/blob/voluntary-exit/message/validation/validation.go#L395

eth/eventhandler/handlers.go
nkryuchkov: Well spotted, thanks. Added

eth/eventhandler/handlers.go
y0sher: it seems that the beacon chain only cares that the slot and epoch of the exit request isn't in the far future, or too near to the validator activation.
see ethereum spec for more info -
```python
//    def process_voluntary_exit(state: BeaconState, signed_voluntary_exit: SignedVoluntaryExit) -> None:
//     voluntary_exit = signed_voluntary_exit.message
//     validator = state.validators[voluntary_exit.validator_index]
//     # Verify the validator is active
//     assert is_active_validator(validator, get_current_epoch(state))
//     # Verify exit has not been initiated
//     assert validator.exit_epoch == FAR_FUTURE_EPOCH
//     # Exits must specify an epoch when they become valid; they are not valid before then
//     assert get_current_epoch(state) >= voluntary_exit.epoch
//     # Verify the validator has been active long enough
//     assert get_current_epoch(state) >= validator.activation_epoch + SHARD_COMMITTEE_PERIOD
//     # Verify signature
//     domain = get_domain(state, DOMAIN_VOLUNTARY_EXIT, voluntary_exit.epoch)
//     signing_root = compute_signing_root(voluntary_exit, domain)
//     assert bls.Verify(validator.pubkey, signing_root, signed_voluntary_exit.signature)
//     # Initiate exit
```

beacon/goclient/voluntary_exit.go
```diff
@@ -2,9 +2,8 @@ package goclient
 
 import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
-	"github.com/pkg/errors"
 )
 
-func (gc *goClient) SubmitVoluntaryExit(voluntaryExit *phase0.SignedVoluntaryExit, sig phase0.BLSSignature) error {
-	return errors.New("not implemented")
+func (gc *goClient) SubmitVoluntaryExit(voluntaryExit *phase0.SignedVoluntaryExit, _ phase0.BLSSignature) error {
```
y0sher, line 7: so maybe this can be `SubmitVoluntaryExit(voluntaryExit *phase0.SignedVoluntaryExit) error` ? of course might require spec change so only do when and if @GalRogozinski approves.

protocol/v2/ssv/runner/voluntary_exit.go
```diff
@@ -90,6 +93,12 @@ func (r *VoluntaryExitRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg
 		return errors.Wrap(err, "could not submit voluntary exit")
 	}
 
+	logger.Debug("voluntary exit submitted successfully",
```
y0sher, line 96: @moshe-blox, @lior-blox  this will probably need to be added to spec-alignment.. let me know who you want to do it.

y0sher (dismissed):
good job, minor comments for the rest of the reviewers.

beacon/goclient/voluntary_exit.go
nkryuchkov: @y0sher I already requested to change this. @MatheusFranco99 has a PR for this: https://github.com/bloxapp/ssv-spec/pull/319

nkryuchkov: https://github.com/bloxapp/ssv-spec/pull/319 was merged, but I think we need to do this change on the next alignment with spec to avoid delaying this PR

protocol/v2/ssv/runner/voluntary_exit.go
moshe-blox: @y0sher done, it was just a metric line

eth/eventhandler/handlers.go
moshe-blox: @y0sher @nkryuchkov 
the scheduling is indeed problematic, as different operators may handle this event at different slots, which may happen surprisingly often when your EL/CL node is just a few seconds behind.

in case 2/4 operators schedule at different slots, they wouldn't reach consensus and force the user to pay fees on another TX

i believe the most reliable way to schedule this duty must be deterministic (without current time, determined from the event only), one proposal could be:
1. when handling the event, pass over the event's block number (rather than current slot)
2. when executing the task, determine the event's slot (requires querying the EL) and schedule the duty for X slots ahead

this means that if X is 4 (for example), then operators can handle the event up to 3 slots late and still execute it together and reach consensus

wdyt? do you have any other ideas that don't require querying the EL?

moshe-blox:
please see https://github.com/bloxapp/ssv/pull/1200#discussion_r1394458535
besides that, looks very good 💪

eth/eventhandler/handlers.go
y0sher: so again this sums up the first part I mentioned where the issue is with scheduling the slot when we do the duty. which we need to somehow be able to determine same slot. using the `CurrentEstimatedSlot` probably won't fit since @moshe-blox says BL/EL nodes might delay and the slots end up different. (even though so far on stage this never happened). 
So I think we must hang to something deterministic about the event to be able to determine a slot. it seems like everything will require contact with EL. we should get this finalized ASAP though, @nkryuchkov ,@lior-blox, @GalRogozinski do you have any idea or leaning towards any approach?

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,99 @@
...
+			currentSlot := h.ticker.Slot()
+
+			h.logger.Debug("🛠 ticker event", fields.Slot(currentSlot))
+
+			var dutiesForExecution []*spectypes.Duty
+
+			h.dutyQueueMu.Lock()
+			for _, duty := range h.dutyQueue {
+				if duty.Slot > currentSlot {
+					break
```
moshe-blox, line 55: I think this should be `continue`, otherwise the subsequent duties that should be executed now would be ignored. I realize this may not happen because `validatorExitCh` should receive messages with increasing block numbers, however it's always safer to make less assumptions about external behavior, especially with channels (in some cases channels don't preserve order of messages).

What do you think about something like this?
```go
var forExecution []*spectypes.Duty
var remaining []*spectypes.Duty
for _, duty := range h.dutyQueue {
	if duty.Slot <= h.currentSlot {
		forExecution = append(forExecution, duty)
	} else {
		remaining = append(remaining, duty)
	}
}
h.dutyQueue = remaining
// ...
```

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,99 @@
...
+
+			if dutyCount != 0 {
+				h.executeDuties(h.logger, dutiesForExecution)
+				h.logger.Debug("executed voluntary exit duties",
+					fields.Slot(currentSlot),
+					fields.Count(dutyCount))
+			}
+
+		case exitDescriptor := <-h.validatorExitCh:
+			time, err := h.executionClient.BlockTime(ctx, exitDescriptor.BlockNumber)
```
y0sher, line 76: time is a name of a package so its preferred not to use it as a var name

eth/executionclient/execution_client.go
```diff
@@ -244,6 +244,15 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 	return nil
 }
 
+func (ec *ExecutionClient) BlockTime(ctx context.Context, blockNumber uint64) (time.Time, error) {
```
moshe-blox, line 247: Just two notes for future-proofing the `ExecutionClient`:
1. Accept int64 instead of uint64 so it's possible to pass special numbers such as `rpc.FinalizedBlockNumber`
2. Consider renaming to `BlockTimeByNumber` or maybe just `BlockByNumber` and return the whole block.

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,99 @@
...
+			if err != nil {
+				h.logger.Warn("failed to get block time from execution client, skipping voluntary exit duty",
+					zap.Error(err))
+				continue
+			}
+
+			slot := h.network.Beacon.EstimatedSlotAtTime(time.Unix()) + voluntaryExitSlotsToPostpone
```
y0sher: (voluntaryExitSlotsToPostpone * SlotTimeInSeconds) .
As far as I understand `voluntaryExitSlotsToPostpone` is just slot number (4). to add this to `time` you need to multiply it by the number of seconds in a slot. Unix timestamp is represented in seconds.

operator/validator/task_executor.go
```diff
@@ -100,17 +101,21 @@ func (c *controller) UpdateFeeRecipient(owner, recipient common.Address) error {
...
-		Type:           spectypes.BNRoleVoluntaryExit,
+
+func (c *controller) ExitValidator(pubKey phase0.BLSPubKey, blockNumber uint64, validatorIndex phase0.ValidatorIndex) error {
+	logger := c.taskLogger("ExitValidator",
+		fields.PubKey(pubKey[:]),
+		fields.BlockNumber(blockNumber),
+		zap.Uint64("validator_index", uint64(validatorIndex)),
+	)
+
+	c.validatorExitCh <- duties.ExitDescriptor{
```
moshe-blox, line 112: This should probably send in a goroutine like we do with `indicesChanges`, to avoid blocking the `EventHandler` when VoluntaryExitHandler is not currently reading from the channel:
```go
	go func() {
			select {
			case c.indicesChange <- struct{}{}:
			case <-time.After(2 * SlotDuration):
				logger.Error("failed to schedule ExitValidator duty!")
			}
		}()
```
I suggest to include a timeout of 1 or 2 slots with an error log (not debug like we usually do, this is a really bad error that should never happen)

moshe-blox (changes requested):

operator/duties/voluntary_exit.go
nkryuchkov: Also, since `voluntaryExitSlotsToPostpone` is typed, it's impossible to add it to a different type without casting

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,99 @@
...
+	PubKey         phase0.BLSPubKey
+	ValidatorIndex phase0.ValidatorIndex
+	BlockNumber    uint64
+}
+
+type VoluntaryExitHandler struct {
+	baseHandler
+	validatorExitCh <-chan ExitDescriptor
+	dutyQueue       []*spectypes.Duty
+	dutyQueueMu     sync.Mutex
```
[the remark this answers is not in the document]
nkryuchkov: done in https://github.com/bloxapp/ssv/pull/1200#discussion_r1406135945

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,93 @@
...
+
+		case exitDescriptor := <-h.validatorExitCh:
+			blockTime, err := h.executionClient.BlockTime(ctx, exitDescriptor.BlockNumber)
+			if err != nil {
+				h.logger.Warn("failed to get block time from execution client, skipping voluntary exit duty",
+					zap.Error(err))
+				continue
+			}
+
+			slot := h.network.Beacon.EstimatedSlotAtTime(blockTime.Unix()) + voluntaryExitSlotsToPostpone
```
moshe-blox, line 79: To avoid re-fetching whole blocks (not very lightweight) when there are more than 1 validator exits in them, we should probably keep a dumb cache (for example `blockSlots map[uint64]phase0.Slot`) in the struct, and for simple eviction — delete any key that is a lower block after caching a new block.

operator/duties/voluntary_exit.go
moshe-blox: > I'm not sure what could be the case when order of messages is not preserved, but okay, let's make it less error-prone.
For example, when multiple goroutines send to a channel, then the order of arrival is not guaranteed to match the order of who started sending first.

nkryuchkov: Well, task executor is what triggers sending to `validatorExitCh`, so slots should be in order. Also, with the new approach we are going to copy duties every slot when they exist.

However, since this is a rare duty, the queue is expected to be small, and any bug here would be hard to find, I agree to change this.

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+
+			slot := h.network.Beacon.EstimatedSlotAtTime(time.Unix()) + voluntaryExitSlotsToPostpone
+
+			duty := &spectypes.Duty{
+				Type:           spectypes.BNRoleVoluntaryExit,
+				PubKey:         exitDescriptor.PubKey,
+				Slot:           slot,
+				ValidatorIndex: exitDescriptor.ValidatorIndex,
+			}
```
y0sher: the duty should run on the slot `slot`, but it should sign on the slot of the blocktime.

nkryuchkov: well spotted, please check it once again

beacon/goclient/voluntary_exit.go
y0sher: @Tomblox1 we might need a task here to not forget to align this once we align the spec.

nkryuchkov: @y0sher I think we won't be able to forget to align this because the code will not compile after updating spec version

eth/contract/contract.go
y0sher, line 1: Before merging I'll check in the contract repo if anything was updated so maybe its best to take a new abi

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,115 @@
...
+			blockSlot, ok := h.blockSlots[exitDescriptor.BlockNumber]
+			if !ok {
+				block, err := h.executionClient.BlockByNumber(ctx, new(big.Int).SetUint64(exitDescriptor.BlockNumber))
+				if err != nil {
+					h.logger.Warn("failed to get block time from execution client, skipping voluntary exit duty",
+						zap.Error(err))
+					continue
+				}
+
+				blockSlot = h.network.Beacon.EstimatedSlotAtTime(int64(block.Time()))
```
y0sher, line 87: so this is to prevent calling execution client multiple times if there is a lot of validator duty exits on the same block? I'd add a comment to make it clear why we have this map.

scripts/spec-alignment/differ.config.yaml
```diff
@@ -11,7 +11,7 @@ ApprovedChanges: ["256a3dc0f1eb7abf","22b66e9a63ba145b","12c1c3a1622fb7cc","1c44
                   "90b8a0c8d2c30e95","e8292a58d2eb08ab","17cf3119ac6879f2","3f31546191c9e6b2","29c96f90edc2458d","f29db2624fd63635","dff6fea2c2d32a5f",
                   "ae1b53fc580ce346","c117bd5db3eeabd6","d06552d71b9ca4cd","4cb333a88af66575","2a580187c312c79a","bf8cf93c55c1eadb","6d877e24991465e4",
                   "b1c8e0148a4a755","2c25abb7c776bd54","a1754e08473bd1fa","4dbab14670fa155d","2a3667a499a23b16","930379d323dd95e8","65efe31656e8814f",
-                  "1270cef2e573f846","aeafb38ca9114f12","2a83e3384b45f2d7","91fbb874b3ce2570","74ad51ca63526e1e","defd8406641d53a5"]
+                  "1270cef2e573f846","aeafb38ca9114f12","2a83e3384b45f2d7","91fbb874b3ce2570","74ad51ca63526e1e","defd8406641d53a5","efa21b9890ec787b"]
```
y0sher, line 14: @moshe-blox please approve this.

y0sher (dismissed):

operator/duties/voluntary_exit.go
nkryuchkov: @y0sher yes, added

y0sher (approved):

operator/duties/voluntary_exit.go
```diff
@@ -116,3 +101,26 @@ func (h *VoluntaryExitHandler) HandleDuties(ctx context.Context) {
 		}
 	}
 }
+
+// blockSlot gets slots happened at the same time as block,
+// it prevents calling execution client multiple times if there are several validator exit events on the same block
+func (h *VoluntaryExitHandler) blockSlot(ctx context.Context, blockNumber uint64) (phase0.Slot, error) {
+	blockSlot, ok := h.blockSlots[blockNumber]
+	if !ok {
```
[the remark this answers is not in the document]
nkryuchkov: sure

moshe-blox (approved):

y0sher:
Since @olegshmuelov is back and this touches some of his areas I would love him to review it as well. approved on my end though.

moshe-blox (dismissed):

operator/duties/voluntary_exit.go
```diff
@@ -0,0 +1,128 @@
...
+
+			h.dutyQueue = append(h.dutyQueue, duty)
+
+			h.logger.Debug("🛠 scheduled duty for execution",
+				zap.Uint64("block_slot", uint64(blockSlot)),
+				zap.Uint64("duty_slot", uint64(dutySlot)),
+				fields.BlockNumber(exitDescriptor.BlockNumber),
+			)
+		}
+	}
```
olegshmuelov, line 102: there is a potential channel blocking issue in the select block of the for-loop, to resolve it necessary to add the following case statements:

```go
case <-h.indicesChange:
	continue

case <-h.reorg:
	continue
```
These additions will ensure that signals from h.indicesChange and h.reorg are correctly handled, preventing any blocking in the loop.

olegshmuelov (changes requested):

operator/duties/voluntary_exit.go
olegshmuelov: Also, just to be sure, are there any particular processes or procedures we need to initiate upon receiving these signals?

y0sher: @olegshmuelov we start a duty. all the rest is taken care by the runner.

moshe-blox: @olegshmuelov Good catch!

moshe-blox: > Also, just to be sure, are there any particular processes or procedures we need to initiate upon receiving these signals?

Not really, it's kind of like ValidatorRegistration

moshe-blox (approved):

y0sher (approved):

olegshmuelov (approved):

PR #1203, opened 2023-11-13
Node health

pavelkrolevets:
#### Health check at GET `/v1/node/health`
- checks BN connection health
- checks execution health
- checks event sync health
- checks amount of connected peers (less than 10 isnt good)
- checks if inbound peer equal to 0 - probably local port isnt reachable by other peers

#### Returns:
```json
{
  "p2p": "bad: no peers are connected",
  "beacon_node": "good",
  "execution_node": "good",
  "event_syncer": "good",
  "advanced": {
    "peers": 0,
    "inbound_conns": 0,
    "outbound_conns": 0,
    "p2p_listen_addresses": [
      "192.168.1.112:13001"
    ]
  }
}
```

#### Example:
```sh
curl -i -H "Accept: application/json" http://5.161.192.132:16052/v1/node/health
```

```sh
curl -i -H "Accept:  text/plain" http://5.161.192.132:16052/v1/node/health
```


api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	}
+	// check peers connection
+	var activePeerCount int
+	prs := h.Network.Peers()
+	for _, p := range h.peers(prs) {
+		if p.Connectedness == "Connected" {
+			activePeerCount++
+		}
+	}
+	switch count := activePeerCount; {
```
nkryuchkov, line 150: I see both `count` and `activePeerCount` being used in the switch-statement, so I'm not sure why we need to create the `count` variable here

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	case count >= 5:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %d  peers are connected", Healthy.String(), activePeerCount)
+	case count < 5:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %s", NotHealthy.String(), "less than 5 peers are connected")
+	case count == 0:
```
nkryuchkov, line 155: Unreachable, overlapped by `count < 5`

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	for _, p := range h.peers(prs) {
+		if p.Connectedness == "Connected" {
+			activePeerCount++
+		}
+	}
+	switch count := activePeerCount; {
+	case count >= 5:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %d  peers are connected", Healthy.String(), activePeerCount)
+	case count < 5:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %s", NotHealthy.String(), "less than 5 peers are connected")
```
nkryuchkov, line 154: `NotHealthy.String()` can be just `NotHealthy`

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	// check peers connection
+	var activePeerCount int
+	prs := h.Network.Peers()
+	for _, p := range h.peers(prs) {
+		if p.Connectedness == "Connected" {
+			activePeerCount++
+		}
+	}
+	switch count := activePeerCount; {
+	case count >= 5:
```
nkryuchkov, line 151: let's extract 5 to a constant

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %d  peers are connected", Healthy.String(), activePeerCount)
```
nkryuchkov, line 152:
```suggestion
		resp.PeersHealthStatus = fmt.Sprintf("%s: %d peers are connected", Healthy.String(), activePeerCount)
```

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	// check event sync health
+	err = h.NodeProber.CheckEventSycNodeHealth(ctx)
+	if err != nil {
+		resp.EventSyncHealthStatus = Healthy.String()
+	} else {
+		resp.EventSyncHealthStatus = fmt.Sprintf("%s: %s", NotHealthy.String(), err.Error())
+	}
+	// check peers connection
+	var activePeerCount int
+	prs := h.Network.Peers()
```
nkryuchkov, line 144:
```suggestion
	peers := h.Network.Peers()
```

nodeprobe/nodeprobe.go
```diff
@@ -136,3 +136,27 @@ func (p *Prober) AddNode(name string, node Node) {
 
 	p.nodes[name] = node
 }
+
+func (p *Prober) CheckBeaconNodeHealth(ctx context.Context) error {
+	err := p.nodes["consensus client"].Healthy(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
```
nkryuchkov, line 145:
```suggestion
	return p.nodes["consensus client"].Healthy(ctx)
```

or let's wrap the error

nodeprobe/nodeprobe.go
```diff
@@ -136,3 +136,27 @@ func (p *Prober) AddNode(name string, node Node) {
...
+
+func (p *Prober) CheckExecutionNodeHealth(ctx context.Context) error {
+	err := p.nodes["execution client"].Healthy(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *Prober) CheckEventSycNodeHealth(ctx context.Context) error {
```
nkryuchkov, line 156:
```suggestion
func (p *Prober) CheckEventSyncNodeHealth(ctx context.Context) error {
```

nodeprobe/nodeprobe.go
```diff
@@ -136,3 +136,27 @@ func (p *Prober) AddNode(name string, node Node) {
 
 	p.nodes[name] = node
 }
+
+func (p *Prober) CheckBeaconNodeHealth(ctx context.Context) error {
+	err := p.nodes["consensus client"].Healthy(ctx)
```
nkryuchkov, line 141: it's not a part of your PR, but I think these string constants are error-prone, so I thought maybe it's worth to leave a TODO that we could e.g. add a method to clients that returns their name or solve this in another way

api/handlers/node.go
nkryuchkov: also I think we need to be consistent in naming as there are `prs`, `peers`, and `peerz` for the same entity

nkryuchkov (changes requested):

api/handlers/node.go
```diff
@@ -1,18 +1,40 @@
...
 
+type HealthStatus int
+
+const (
+	NotHealthy HealthStatus = iota
+	Healthy
+)
+
+func (c HealthStatus) String() string {
+	str := [...]string{"NotHealthy", "Healthy"}
```
y0sher, line 31: confusing name, I'd chose something more bold like "OK" and "BAD"/"ERROR"

api/handlers/node.go
```diff
@@ -65,6 +96,84 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+		}
+	}
+	switch {
+	case activePeerCount >= healthyPeersAmount:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %d  peers are connected", Healthy, activePeerCount)
+	case activePeerCount > 0 && activePeerCount < healthyPeersAmount:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: less than %d peers are connected", NotHealthy, healthyPeersAmount)
+	case activePeerCount == 0:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %s", NotHealthy, "error: no peers are connected")
+	}
```
y0sher, line 161: I would break this into two. one that says GOOD/BAD (so status only), and one that says the number, or maybe reason.

api/handlers/node.go
```diff
@@ -64,7 +92,73 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
 
 func (h *Node) Peers(w http.ResponseWriter, r *http.Request) error {
-	peers := h.Network.Peers()
+	prs := h.Network.Peers()
+	resp := h.peers(prs)
+	return api.Render(w, r, resp)
+}
+
+func (h *Node) Topics(w http.ResponseWriter, r *http.Request) error {
+	allpeers, peerbytpc := h.TopicIndex.PeersByTopic()
```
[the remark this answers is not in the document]
pavelkrolevets: Topics was here before the PR. But yes, we should follow camel case everywhere. Why linter isnt complaining? Maybe we need to make it more strict...

y0sher (approved):
LGTM

y0sher:
@pavelkrolevets Can you please post here pic or text of how the final result looks like?

api/handlers/node.go
```diff
@@ -65,6 +96,72 @@ func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
...
+	for _, p := range h.peers(peers) {
+		if p.Connectedness == "Connected" {
+			activePeerCount++
+		}
+	}
+	switch {
+	case activePeerCount > 0 && activePeerCount < healthyPeersAmount:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %d peers are connected", bad, activePeerCount)
+	case activePeerCount == 0:
+		resp.PeersHealthStatus = fmt.Sprintf("%s: %s", bad, "error: no peers are connected")
```
nkryuchkov, line 148: This can be simplified to
```suggestion
		resp.PeersHealthStatus = fmt.Sprintf("%s: error: no peers are connected", bad)
```

cli/operator/node.go
```diff
@@ -12,13 +12,13 @@ import (
...
 	"time"
 
-	spectypes "github.com/bloxapp/ssv-spec/types"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ilyakaznacheev/cleanenv"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 
+	spectypes "github.com/bloxapp/ssv-spec/types"
```
nkryuchkov, line 21: This should be in the block above

nkryuchkov (approved):

cli/operator/node.go
pavelkrolevets: My IDE moves automatically to bloxapp group :-) Changed manually

y0sher (approved):

nkryuchkov (dismissed):

lior-blox:
@MatusKysel @pavelkrolevets most of the issues that operators encountered was that they forgot to open ports in their server providers.
So a check from an external tool needs to be added.
Here is my suggestion:
as we don't want to choose and hardcode a service URL  (it might be down or changed)
lets add to this route as a query param port checker URL and trigger it.
also please make sure to set context timeout for the reques

nodeprobe/nodeprobe.go
```diff
@@ -136,3 +136,23 @@ func (p *Prober) AddNode(name string, node Node) {
...
+	ctx, cancel := context.WithTimeout(ctx, p.interval)
+	defer cancel()
+	return p.nodes["execution client"].Healthy(ctx)
+}
+
+func (p *Prober) CheckEventSyncerHealth(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, p.interval)
+	defer cancel()
+	return p.nodes["event syncer"].Healthy(ctx)
+}
```
y0sher, line 158: all reset of the places accessing the nodes are using mutex, shouldn't we protect here as well?

moshe-blox:
Please also merge https://github.com/bloxapp/ssv/pull/1141 with this

nodeprobe/nodeprobe.go
pavelkrolevets: added mutex

api/handlers/node.go
```diff
@@ -106,88 +113,64 @@ func (h *Node) Peers(w http.ResponseWriter, r *http.Request) error {
...
-	tpcs := []topicIndexJSON{}
-	for topic, peers := range peerbytpc {
-		tpcs = append(tpcs, topicIndexJSON{TopicName: topic, Peers: peers})
+	peers, byTopic := h.TopicIndex.PeersByTopic()
+
+	resp := AllPeersAndTopicsJSON{
+		AllPeers: peers,
+	}
+	for topic, peers := range byTopic {
+		resp.PeersByTopic = append(resp.PeersByTopic, topicIndexJSON{TopicName: topic, Peers: peers})
```
y0sher, line 122: using peers twice for different stuff, though I think its not a big issue since first one usage is done before scond is started

y0sher (dismissed):

y0sher (commented):
@moshe-blox @pavelkrolevets @MatusKysel what is the status here?

y0sher (dismissed):

y0sher (approved):

PR #1205, opened 2023-11-15
Init fetch

MatusKysel:
Initialize the scheduler by fetching duties for both the Proposer and the Sync Committee. This step will help eliminate message validation errors during initialization, as it ensures that duties are already fetched and in place.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ Build stage Docker image:
     - docker tag $IMAGE_NAME:$CI_COMMIT_SHA $DOCKER_REPO_INFRA_STAGE:$CI_COMMIT_SHA
     - $DOCKER_LOGIN_TO_INFRA_STAGE_REPO && docker push $DOCKER_REPO_INFRA_STAGE:$CI_COMMIT_SHA
   only:
-    - init-fetch
+    - stage
 
 # +---------------------+
 # | STAGE HETZNER NODES |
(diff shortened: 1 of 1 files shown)


operator/node.go
```diff
@@ -157,6 +149,14 @@ func (n *operatorNode) Start(logger *zap.Logger) error {
 		logger.Fatal("duty scheduler exited with error", zap.Error(err))
```
y0sher, line 149: don't we need to move this code block (with wait and errcheck) after all the code below?
Wait basically blocks forever.
Now start will block until after the initial fetch which is what we wanted.

operator/duties/scheduler_test.go
```diff
@@ -275,7 +275,7 @@ func TestScheduler_Run(t *testing.T) {
 	for _, mockDutyHandler := range s.handlers {
 		mockDutyHandler.(*MockdutyHandler).EXPECT().Setup(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
 		mockDutyHandler.(*MockdutyHandler).EXPECT().HandleDuties(gomock.Any()).
-			DoAndReturn(func(ctx context.Context) {
+			DoAndReturn(func(ctx context.Context, w *sync.WaitGroup) {
```
y0sher, line 278: I guess that should reverted now?

operator/duties/proposer.go
```diff
@@ -106,6 +106,15 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
 	}
 }
 
+func (h *ProposerHandler) HandleInitialDuties(ctx context.Context) {
+	slot := h.network.Beacon.EstimatedCurrentSlot()
+	epoch := h.network.Beacon.EstimatedEpochAtSlot(slot)
+	h.processFetching(ctx, epoch, slot)
+	h.fetchFirst = false
+	h.indicesChanged = false
+	h.logger.Info("fetching duties on init done")
```
y0sher, line 115: either add more info to this log or remove it. at least make it debug. IMO we can have one log about initial duties fetched either on `operatorNode` or on the `scheduler`

y0sher (approved):
good job.

y0sher (approved):

PR #1206, opened 2023-11-15
Score parameters for banning peers sending rejected messages

y0sher:
### Description
This PR introduces efficient peer scoring, by adjusting values and params used when a message is rejected, ignored or accepted in message validation, we set a peer score with meaningful value that might result in ban when the peer is acting bad.

https://github.com/bloxapp/ssv/pull/1150 merged using cherry pick - adding a test to see that the scoring is really effective.


y0sher:
Seems like tests are passing locally but failing in ci, might be a resources issue.

network/topics/params/helpers.go
```diff
@@ -7,10 +7,10 @@ import (
...
 )
 
-// scoreByWeight provides the relevant score by the provided weight and threshold.
-func scoreByWeight(maxScore float64, weight, threshold float64) float64 {
-	return maxScore / (weight * threshold * threshold)
-}
+const (
+	oneEpochDuration = (12 * time.Second) * 32
+	slotsPerEpoch    = 32
+)
```
y0sher, line 13: we better use values from the network config, but we'll have to pass them around to do that.

MatheusFranco99: `slotsPerEpoch` was unused, so I removed it.
`oneEpochDuration` is used as the default value if the `Options.Network.OneEpochDuration` is not set.

PR #1208, opened 2023-11-16
Feat/ log branch and commit ver

y0sher:
looks like this now - 
```
2023-11-18T20:14:37.699299Z	[35mINFO[0m	starting ssv-node { "version":  "v1.0.0-rc.2-44-g2409736e-stage-2409736e1e1b22e1c6e882821075eb38cc076dde" }
```

### DIFF
--- a/cli/bootnode/boot_node.go
+++ b/cli/bootnode/boot_node.go
@@ -1,6 +1,8 @@
 package bootnode
 
 import (
+	"fmt"
+	"github.com/bloxapp/ssv/utils/commons"
 	"log"
 
 	"github.com/bloxapp/ssv/logging"
@@ -41,11 +43,15 @@ var StartBootNodeCmd = &cobra.Command{
 				MaxBackups: cfg.LogFileBackups,
 			},
 		)
+
 		if err != nil {
 			log.Fatal(err)
 		}
 
 		logger := zap.L()
+		log.Printf("starting SSV node (version %s)", commons.GetBuildData())
+
+		logger.Info(fmt.Sprintf("starting SSV %v (version %s)", cmd.Short, commons.GetBuildData()))
 
 		bootNode, err := bootnode.New(cfg.Options)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cli/bootnode/boot_node.go
```diff
@@ -41,11 +43,15 @@ var StartBootNodeCmd = &cobra.Command{
 				MaxBackups: cfg.LogFileBackups,
 			},
 		)
+
 		if err != nil {
 			log.Fatal(err)
 		}
 
 		logger := zap.L()
+		log.Printf("starting SSV node (version %s)", commons.GetBuildData())
```
nkryuchkov, line 52: Is there any reason not to use `logger` here? Why do we have two similar logs?

cli/bootnode/boot_node.go
```diff
@@ -41,11 +43,15 @@ var StartBootNodeCmd = &cobra.Command{
...
+
+		logger.Info(fmt.Sprintf("starting SSV %v (version %s)", cmd.Short, commons.GetBuildData()))
```
nkryuchkov, line 54: The `AppName` is `SSV-Node`, so this seems to be printed as `starting SSV SSV-Node`, should we remove the "SSV" duplication?

cli/bootnode/boot_node.go
y0sher: you are right. changed.

cli/bootnode/boot_node.go
y0sher: agree.

Roy-blox: @y0sher  Take a look at this: https://blox-ensemble.monday.com/boards/4755515647/pulses/5536501995
It seems commons.GetBuildData returns "SSV node:' ' " with no data in it.
Regardless of the bug I believe a quick validation to make sure that the data is not empty is in place here

andrew-blox (approved):

PR #1223, opened 2023-12-03
No reason to silence wrong domain ignore error

GalRogozinski:
Should help debugging issues


y0sher:
@GalRogozinski this is silenced because we are still getting connections from other networks (domains) and their messages are slipping into pubsub hence message validation, we'll probably see tons of these logs if we remove the silenced.
if we finish this PR's original premise #1175 then this should not be an issue anymore.

PR #1225, opened 2023-12-04
peer limiter 

MatusKysel:


message/validation/validation.go
```diff
@@ -262,6 +268,8 @@ func (mv *messageValidator) ValidatePubsubMessage(_ context.Context, peerID peer
 				}
 
 				mv.metrics.MessageRejected(valErr.Text(), descriptor.Role, round)
+				mv.peerRateLimiter.RegisterRequest(peerID)
```
GalRogozinski, line 271: We had a discussion now whether we should register here ever "reject" or just rejects that come from invalid signatures.

This is because of the follwing logic:
We want to be somewhat robust to bugs. Currently if the rejections happen across a wide range of topics (subnets) we can endure as much as 291 rejects (worst case) with pure libp2p scoring.

as long as those rejects don't hit the sig validation it isn't this bad. Since it is possible we introduce bugs that causes REJECTs we might want to consider to give more leeway to such scenarios.

I prefer to let @lior-blox @moshe-blox or @y0sher decide about this, since they know SSV code better than me

PR #1226, opened 2023-12-04
feat: rate limit inbound connections by IP

moshe-blox:
Following [Prysm's example](https://github.com/prysmaticlabs/prysm/pull/6780), this PR adds rate limiting in `ConnectionGater` to prevent Syn Flood attacks.

TODO: add unit test


y0sher:
the issue is that conngater is never used so this has no effect.

y0sher:
> the issue is that conngater is never used so this has no effect.

btw conn gater is passed to libp2p options and used in this PR - #1225 
maybe it worth to merge this into there.

moshe-blox:
> > the issue is that conngater is never used so this has no effect.
> 
> btw conn gater is passed to libp2p options and used in this PR - #1221 maybe it worth to merge this into there.

agreed, although maybe we can merge that one and then this one separately?

y0sher (approved):

moshe-blox:
On pause, @MatheusFranco99 is checking potential issue where we blindly reject messages (without validating again after rate limit exceeded) and end up losing potential valid messages.

GalRogozinski (approved):
Test looks ok

Verified by initiating log

y0sher (approved):

PR #1229, opened 2023-12-06
fix: stale operator ID in `p2pNetwork`

moshe-blox:


y0sher (approved):

network/p2p/config.go
```diff
@@ -63,7 +63,7 @@ type Config struct {
 	// OperatorPubKeyHash is hash of operator public key, used for identity, optional
 	OperatorPubKeyHash string
 	// OperatorID contains numeric operator ID
-	OperatorID spectypes.OperatorID
+	OperatorID func() spectypes.OperatorID
```
nkryuchkov, line 66: Please update comment. 

Also, I'd consider renaming to something like `OperatorIDFunc`/`OperatorID`/`GetOperatorID`/`OperatorIDGetter` to explicitly show that this is not a number

nkryuchkov (approved):

PR #1232, opened 2023-12-07
Peer msg rate limiter v1

MatusKysel:
This PR introduces PeerRateLimitManager, a Go-based solution for managing rate limiting on a per-peer basis in networked applications. It features individual rate limiters for each peer, ensuring fair and efficient handling of requests. The manager uses a dual-limiter system (rejectLimiter and ignoreLimiter) for flexible request management and incorporates a blocking mechanism to prevent overload.

### DIFF
--- a/message/validation/ratelimiter/peer_limiter.go
+++ b/message/validation/ratelimiter/peer_limiter.go
@@ -25,7 +25,7 @@ type Config struct {
 
 func DefaultConfig() Config {
 	return Config{
-		InvalidRSAs:     5,
+		InvalidRSAs:     20,
 		InvalidMessages: 70000,
 		BlockDuration:   1 * time.Minute,
 		CacheSize:       100,


GalRogozinski (commented):
@MatusKysel 
How does the limiter refills its tokens?

@MatheusFranco99 
I noticed no decay of the score... what did you change in the scoring exactly?

GalRogozinski:
Not that is is that important to me. But in this PR we use a different lib that does the same thing:
https://github.com/bloxapp/ssv/pull/1226

message/validation/peer_limiter.go
```diff
@@ -0,0 +1,101 @@
...
+	rejectLimiter *rate.Limiter
+	ignoreLimiter *rate.Limiter
+	blockRequests bool
+	lastBlocked   time.Time
+	mu            sync.Mutex // Mutex to protect blockRequests and lastBlocked
+}
+
+func NewRateLimiter(rejectLimit, ignoreLimit rate.Limit) *RateLimiter {
+	return &RateLimiter{
+		rejectLimiter: rate.NewLimiter(rejectLimit, int(rejectLimit)),
```
GalRogozinski, line 22: maybe like you said in our conversation
if this is just for RSA sig errors we should give this a less confusing name

GalRogozinski:
@MatusKysel 
I am thinking about what you explained to me regarding `blast`.
since the reject limiter is locked only on sig validation maybe it will be beneficial to also set this parameter to 20.
So after 20 quick messages we reject w.o waiting for the whole ticker to pass (one minute if I understood correctly)

The reasoning is simple, sig validation is less susceptible to buggy quirks of the node. Either the signature is correct or not.

GalRogozinski (changes requested):
I see tests fail when I run them

GalRogozinski (approved):

cli/operator/node.go
```diff
@@ -187,6 +187,7 @@ var StartNodeCmd = &cobra.Command{
 			validation.WithMetrics(metricsReporter),
 			validation.WithDutyStore(dutyStore),
 			validation.WithOwnOperatorID(operatorData.ID),
+			validation.WithLimiter(validation.NewPeerRateLimitManager(20, 300000, cfg.P2pNetworkConfig.MaxPeers*2, 1*time.Minute)), // rate per second and duration
```
moshe-blox, line 190: 300k is a reasonable limit for invalid messages? @MatheusFranco99 @MatusKysel

message/validation/validation.go
```diff
@@ -229,6 +237,11 @@ func (mv *messageValidator) ValidatorForTopic(_ string) func(ctx context.Context
 // ValidatePubsubMessage validates the given pubsub message.
 // Depending on the outcome, it will return one of the pubsub validation results (Accept, Ignore, or Reject).
 func (mv *messageValidator) ValidatePubsubMessage(_ context.Context, peerID peer.ID, pmsg *pubsub.Message) pubsub.ValidationResult {
+	if mv.peerRateLimiter != nil && !mv.peerRateLimiter.AllowRequest(peerID) {
```
moshe-blox, line 240: I'd suggest instead of allowing peerRateLimiter to be nil, to either always have to pass it (even in tests), or make it an interface and build an additional "no-op" implementation used for tests only.

message/validation/peer_limiter.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (rl *RateLimiter) RegisterRSAError() {
+	rl.rsaLimiter.Allow()
+}
+
+func (rl *RateLimiter) RegisterInvalid() {
+	rl.invalidLimiter.Allow()
+}
+
+type PeerRateLimitManager struct {
```
moshe-blox, line 56: I think just PeerRateLimiter name can be sufficient, as a struct that handles RateLimiter instances for each peer

message/validation/peer_limiter.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	if rl.rsaLimiter.Tokens() < 1.0 || rl.invalidLimiter.Tokens() < 1.0 {
+		rl.blockRequests = true
+		rl.lastBlocked = time.Now()
+		return false
+	}
+
+	return true
+}
```
moshe-blox, line 46:
```suggestion
func (rl *RateLimiter) AllowRequest(blockingTime time.Duration) bool {
	rl.mu.RLock()
	defer rl.mu.RUnlock()

	if rl.blockRequests {
		if time.Since(rl.lastBlocked) > blockingTime {
			rl.mu.Lock()
			rl.blockRequests = false
			rl.mu.Unlock()
		} else {
			return false
		}
	}

	if rl.rsaLimiter.Tokens() < 1.0 || rl.invalidLimiter.Tokens() < 1.0 {
		rl.mu.Lock()
		rl.blockRequests = true
		rl.lastBlocked = time.Now()
		rl.mu.Unlock()
		return false
	}

	return true
}
```

@MatusKysel  Does this make sense to you as a performance optimization to avoid write-locking most of the time?

I'm not sure, just exploring this possibility

moshe-blox (changes requested):
Good job, left just a few comments for your review

cli/operator/node.go
MatheusFranco99: @GalRogozinski please correct me if I'm wrong.
I think we selected 300k because the attack with such an amount didn't produce any impact on the network.
Nonetheless, it's way above any rate of messages expected. Should we decrease it? (I don't have yet a strong opinion on this trade-off)

message/validation/validation.go
```diff
@@ -262,6 +269,11 @@ func (mv *messageValidator) ValidatePubsubMessage(_ context.Context, peerID peer
 				}
 
 				mv.metrics.MessageRejected(valErr.Text(), descriptor.Role, round)
+				if errors.Is(err, ErrRSASignature) {
+					mv.peerRateLimiter.RegisterRSAErrorRequest(peerID)
+				} else {
+					mv.peerRateLimiter.RegisterInvalidRequest(peerID)
+				}
 				return pubsub.ValidationReject
 			}
```
MatheusFranco99, line 278: Increasing the invalid counter here, even though the message was rejected, was a good idea!!
Can you just add some comments explaining why we do that? Basically, to say that even though the message was rejected, it still consumes resources as a message that would be ignored.

message/validation/validation.go
```diff
@@ -88,13 +89,14 @@ type messageValidator struct {
...
 // NewMessageValidator returns a new MessageValidator with the given network configuration and options.
-func NewMessageValidator(netCfg networkconfig.NetworkConfig, opts ...Option) MessageValidator {
+func NewMessageValidator(netCfg networkconfig.NetworkConfig, limiterCacheSize int, opts ...Option) MessageValidator {
 	mv := &messageValidator{
 		logger:                  zap.NewNop(),
 		metrics:                 &nopMetrics{},
 		netCfg:                  netCfg,
 		operatorIDToPubkeyCache: hashmap.New[spectypes.OperatorID, *rsa.PublicKey](),
 		validationLocks:         make(map[spectypes.MessageID]*sync.Mutex),
+		peerRateLimiter:         NewPeerRateLimiter(20, 300000, limiterCacheSize, 1*time.Minute),
```
MatheusFranco99, line 99: This `20` is the rejection limit value, right? As we discussed, I think this value should be higher. The maximum no. of rejections is 219, but it should be a little below that due to delay, right?

MatheusFranco99 (commented):
Good work! Just a comment and a question regarding a hyperparameter.

MatheusFranco99 (approved):

cli/operator/node.go
GalRogozinski: @moshe-blox 

@MatheusFranco99 and I concluded that if we find time to think about a lower bound we will change this.
But it isn't a priority and it is also ok to do it in a later PR

GalRogozinski (dismissed):

message/validation/ratelimiter/peer_limiter.go
```diff
@@ -0,0 +1,124 @@
...
+	BlockDuration time.Duration
+
+	// CacheSize is the number of peers to keep in the cache
+	CacheSize int
+}
+
+func DefaultConfig() Config {
+	return Config{
+		InvalidRSAs:     5,
+		InvalidMessages: 20,
```
GalRogozinski, line 29: InvalidMessages are ignored messages? If so maybe give a clear name
it should be well above 20 for sure because we can have lots of them

I remember we set invalid RSAs to be 20, because IIRC correctly after 20 we also make the peer greylisted on libp2p. why did this change?

GalRogozinski (changes requested):
I spoke with @MatheusFranco99 and he updated the theoretical Max MPS in the google sheets.

It is now 71,902.50 MPS. So I would put a limit of 70K messages for ignored (or `invalid`).

@MatheusFranco99 may later give a formula to calculate this limit as a function of the number of validators. But currently the figure I gave is good enough for this PR

message/validation/ratelimiter/peer_limiter.go
zippoxer: btw i didn't use the correct numbers here, we'll revert it back to 20 RSAs and 300k messages

moshe-blox:
@GalRogozinski thanks, @MatusKysel already put these new numbers in

moshe-blox:
On pause, @MatheusFranco99  is checking potential issue where we blindly reject messages (without validating again after rate limit exceeded) and end up losing potential valid messages.

y0sher:
Consider using this - https://github.com/libp2p/go-libp2p-pubsub/blob/master/blacklist.go

PR #1234, opened 2023-12-11
Duplicate proposal

MatusKysel:
Ignore duplicate proposal instead of reject

### DIFF
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -56,6 +56,7 @@ var (
 	ErrEarlyMessage                        = Error{text: "early message"}
 	ErrLateMessage                         = Error{text: "late message"}
 	ErrTooManySameTypeMessagesPerRound     = Error{text: "too many messages of same type per round"}
+	ErrDuplicatedProposalWithDifferentData = Error{text: "duplicated proposal with different data"}
 	ErrRSADecryption                       = Error{text: "rsa decryption", reject: true}
 	ErrOperatorNotFound                    = Error{text: "operator not found", reject: true}
 	ErrPubSubMessageHasNoData              = Error{text: "pub-sub message has no data", reject: true}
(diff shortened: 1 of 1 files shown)


nkryuchkov (approved):

y0sher:
> There's about 83 messages rejected because of this issue on production as a whole (holesky, prater and mainnet).
> 
> For that reason, I'm not sure if this is an urgent hotfix, and instead let's try to focus on aligning with the potential fix first: [bloxapp/ssv-spec#321](https://github.com/bloxapp/ssv-spec/pull/321)

I agree

message/validation/errors.go
```diff
@@ -56,6 +56,7 @@ var (
 	ErrEarlyMessage                        = Error{text: "early message"}
 	ErrLateMessage                         = Error{text: "late message"}
 	ErrTooManySameTypeMessagesPerRound     = Error{text: "too many messages of same type per round"}
+	ErrDuplicatedProposalWithDifferentData = Error{text: "duplicated proposal with different data"}
```
y0sher, line 59: I'd like @MatheusFranco99 's input here, did duplicate proposals just became ok?

MatheusFranco99: From my perspective, this type of error is clearly malicious and it should stay as rejected for releases.
If we're suspicious that our attacks is causing performance issue due to this rule, maybe we can test the attack without it in a deployment.
If it's really the issue, I think we should adapt/add some message validation rules to prevent against this specific attack, instead of ignoring duplicated proposals.
Maybe I'm missing something because I don't understand exactly how the attack triggers it.

moshe-blox (approved):

message/validation/errors.go
GalRogozinski: From a clear theoretical perspective @MatheusFranco99 is right.

However, currently we have a bug and on at least one instance at least we saw this being triggered wildly somehow. I found it hard to replicate.
However, if this issue gets triggered on mainnet somehow it will pose an issue.

Changing this to `ignore` won't open up any attack vectors and is the safe way to go.

MatheusFranco99 (approved):

moshe-blox:
We're putting this on hold at the moment because this PR might help avoid these errors: https://github.com/bloxapp/ssv/pull/1252

PR #1235, opened 2023-12-11
Enable P3 score in GossipSub score parameters

MatheusFranco99:
We focus the change on two parameters:
- P3 Threshold. Goal: be more resilient.
- P3 Weight. Goal: smaller penalisation not to trigger thresholds (note: the peer will still be removed from the mesh by opportunistic graft).

- Minor fix: use correct max positive score for weight.

### Threshold
- Used to take into account: decay factor, message rate, a dampening factor (1/50).
- Change: divided msg rate by D (the degree/mesh size in GossipSub) to take into account messages first delivered by other peers.

### Weight
- Used to take into account the max positive score, topic weight and the threshold.
- Change: divided weight by (subnet * ⅓). The idea is to force the score to go to 0 if peer is totally idle in at least ⅓ of all subnets.
- With this change, a fully idle peer will go to at most ~ -194.

Idle peer test result: https://blox-ensemble.monday.com/boards/4802754866/pulses/5672886747/posts/2614500337


network/topics/params/topic_score.go
```diff
@@ -210,13 +210,13 @@ func TopicParams(opts Options) (*pubsub.TopicScoreParams, error) {
...
 	meshMessageDeliveriesDecay := scoreDecay(opts.Network.OneEpochDuration*opts.Topic.MeshDeliveryDecayEpochs, decayInterval)
-	meshMessageDeliveriesThreshold, err := decayThreshold(meshMessageDeliveriesDecay, (expectedMessagesPerDecayInterval * opts.Topic.MeshDeliveryDampeningFactor))
+	meshMessageDeliveriesThreshold, err := decayThreshold(meshMessageDeliveriesDecay, (expectedMessagesPerDecayInterval * opts.Topic.MeshDeliveryDampeningFactor / float64(opts.Topic.D)))
 	if err != nil {
 		return nil, errors.Wrap(err, "could not calculate threshold for mesh message deliveries threshold")
 	}
 	var meshMessageDeliveriesWeight float64
 	if meshScoringEnabled {
-		meshMessageDeliveriesWeight = -(opts.maxScore() / (opts.Topic.TopicWeight * math.Pow(meshMessageDeliveriesThreshold, 2)))
+		meshMessageDeliveriesWeight = -(opts.maxScore() / (opts.Topic.TopicWeight * (float64(opts.Network.Subnets) * 1 / 3) * math.Pow(meshMessageDeliveriesThreshold, 2)))
```
GalRogozinski, line 219: why did you choose 1/3?
not criticizing just asking

network/topics/scoring.go
```diff
@@ -33,16 +33,20 @@ func scoreInspector(logger *zap.Logger, scoreIdx peers.ScoreIndex) pubsub.Extend
...
+			meshDeliveryCounters := make([]float64, 0)
 			for topic, snapshot := range peerScores.Topics {
 				if snapshot.InvalidMessageDeliveries != 0 {
 					filtered[topic] = snapshot
 				}
 				if snapshot.InvalidMessageDeliveries > 0 {
 					totalInvalidMessages += math.Sqrt(snapshot.InvalidMessageDeliveries)
 				}
-				if snapshot.MeshMessageDeliveries < 107 {
+				if snapshot.MeshMessageDeliveries < 11.243655244810899 {
```
GalRogozinski, line 45: what is the calculation for this non round number?
can we have fractional message deliveries?

network/topics/scoring.go
```diff
@@ -33,16 +33,20 @@ func scoreInspector(logger *zap.Logger, scoreIdx peers.ScoreIndex) pubsub.Extend
...
 					totalLowMeshDeliveries++
+					totalMeshDeliveriesSquaredDeficitSum += math.Pow(11.243655244810899-snapshot.MeshMessageDeliveries, 2)
```
GalRogozinski, line 47: less my department but since you use this constant more than once better extact it as a `const`
I would even extract consts that I use only once

GalRogozinski (commented):
nice

network/topics/scoring.go
MatheusFranco99: This is approximately the threshold. Just a log for testing purposes. Will be removed afterwards.
Yup, fractional due to decay.

network/topics/scoring.go
MatheusFranco99: This is just for testing. Will be dropped after everybody agrees with the logic.

network/topics/params/topic_score.go
MatheusFranco99: It's actually open for discussion. I just tried not to be too harsh. I first wanted to go with 10% of the subnets, but it seems that 1/3 is already making peers get penalized.

network/topics/params/topic_score.go
```diff
@@ -10,8 +10,9 @@ import (
 
 const (
 	// Network Topology
-	gossipSubD          = 8
-	minActiveValidators = 200
+	gossipSubD                = 8
+	minActiveValidators       = 200
+	msgsPerSecondPerValidator = 0.05
```
GalRogozinski, line 15: how did you arrive to this number?

GalRogozinski (commented):
nice

network/topics/params/topic_score.go
MatheusFranco99: 0.05 comes from the estimation of the number of messages received in the sheets.

GalRogozinski (approved):

y0sher:
did we test this and how does this differs from current scoring?

MatheusFranco99:
@y0sher
Yes, we tested the normal case and the idle peer case.
The result is shown in the description.
It differs by enabling P3, which allows one to prune from idle peers.
Tests were done from 2 to 3 weeks ago. I can do a new test and show you the compiled results. What do you think?

PR #1236, opened 2023-12-11
Increase the number of outbound peers

GalRogozinski:
From a default of 2 increase to 4.

This is the maximal allowed according to the [libp2p docs](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#outbound-mesh-quotas)
This provides better eclipse protection.


MatheusFranco99 (approved):

.gitlab-ci.yml
```diff
@@ -100,7 +100,7 @@ Deploy exporter to hetzner stage:
     - kubectl config get-contexts
     - .k8/hetzner-stage/scripts/deploy-holesky-exporters.sh $DOCKER_REPO_INFRA_STAGE $CI_COMMIT_SHA ssv $APP_REPLICAS_INFRA_STAGE hetzner.stage.k8s.local hetzner.stage.k8s.local stage.ssv.network $K8S_API_VERSION $SSV_EXPORTER_CPU_LIMIT $SSV_EXPORTER_MEM_LIMIT
   only:
-    - stage
+    - dout
```
y0sher, line 103: revert the deployment

GalRogozinski: done

lior-blox (changes requested):
Please add a summary that explains why we need to change this value

GalRogozinski:
@lior-blox done

MatheusFranco99 (approved):

y0sher (approved):
Generally I understand why increasing outbound connections might reduce the chance for an eclipse attack. but at the same time, I don't get how you came up we are vulnerable to it, what made you want to make this change. did you check how this change affect and realize it protects us in any way?

y0sher (approved):

PR #1237, opened 2023-12-11
Fix "too many duties" false positives

nkryuchkov:
- check duty count if message epoch is same as state epoch
    - remove check if new slot is in new epoch


y0sher:
currently by @nkryuchkov this doesn't solve the issue so I put this PR on hold.

PR #1238, opened 2023-12-12
Message validation ConsensusState eviction

nkryuchkov:
Evict state entries older than 2 epochs


message/validation/validation.go
```diff
@@ -569,13 +580,13 @@ func (mv *messageValidator) consensusState(messageID spectypes.MessageID) *Conse
 		Role:   messageID.GetRoleType(),
 	}
 
-	if _, ok := mv.index.Load(id); !ok {
-		cs := &ConsensusState{
-			Signers: hashmap.New[spectypes.OperatorID, *SignerState](),
-		}
-		mv.index.Store(id, cs)
+	if csItem := mv.state.Get(id); csItem != nil {
```
y0sher, line 583: Get extends the time to live of that item, is 2 epochs always enough or we do want to extend the life of an item once touched? if its enough extending can be disabled.

y0sher (approved):

message/validation/validation.go
nkryuchkov: 2 epochs should be enough because messages after 32+1 slots are considered late. Extending the life might be useful for preventing malicious behavior because the TTL would be updated on any action, so the violator would become closer to the punishment threshold on each violation.

message/validation/validation.go
```diff
@@ -101,6 +110,8 @@ func NewMessageValidator(netCfg networkconfig.NetworkConfig, opts ...Option) Mes
 		opt(mv)
 	}
 
+	go mv.state.Start()
```
y0sher, line 113: This should move to a `Start` function of message validation, and `Stop` should also be added on a stop function.

y0sher (changes requested):
one comment

y0sher (approved):

moshe-blox:
@y0sher @nkryuchkov 

what do you think to replace the ttlhash with a map that every slot we lock, unset the key from 64 slots ago, and then unlock?

nkryuchkov:
@moshe-blox but what if there's a slot with no messages?

olegshmuelov (approved):
LGTM

y0sher:
@nkryuchkov @moshe-blox 
I think we should adjust this to fit `alan/no-fork`.

nkryuchkov:
@y0sher Perhaps we need to consider trying to change ttlcache with an array or something like that

PR #1240, opened 2023-12-13
Fix remaining stale operator ID cases

nkryuchkov:
fix passing stale operator ID to
- message validation
- beacon client
- fee recipient controller


y0sher (dismissed):

y0sher:
@moshe-blox

operator/node.go
```diff
@@ -111,7 +111,7 @@ func New(logger *zap.Logger, opts Options, slotTickerProvider slotticker.Provide
 			Network:            opts.Network,
 			ShareStorage:       opts.ValidatorOptions.RegistryStorage.Shares(),
 			RecipientStorage:   opts.ValidatorOptions.RegistryStorage,
-			OperatorData:       opts.ValidatorOptions.OperatorData,
+			GetOperatorID:      func() spectypes.OperatorID { return opts.ValidatorController.GetOperatorData().ID },
```
moshe-blox, line 114: can we maybe try to add GetOperatorID function to `ValidatorController` instead of creating this func here and elsewhere?

then we can just pass `opts.ValidatorController.GetOperatorID``

or require an `OperatorIDProvider` interface and pass `opts.ValidatorController` directly

moshe-blox (dismissed):
looks good, just 1 comment

operator/node.go
nkryuchkov: @moshe-blox done, please review again

olegshmuelov (changes requested):
following your changes, we can adjust
```go
func (eh *EventHandler) validatorAddedEventToShare(
	event *contract.ContractValidatorAdded,
	shareEncryptionKeyProvider ShareEncryptionKeyProvider,
	operatorData *registrystorage.OperatorData,
	sharePublicKeys [][]byte,
	encryptedKeys [][]byte,
) (*ssvtypes.SSVShare, *bls.SecretKey, error)
```
in `eventhandler` pkg
to receive `getOperatorId` instead of `operatorData`

message/validation/validation.go
```diff
@@ -76,7 +76,7 @@ type messageValidator struct {
 	index                   sync.Map
 	nodeStorage             operatorstorage.Storage
 	dutyStore               *dutystore.Store
-	ownOperatorID           spectypes.OperatorID
+	getOwnOperatorID        func() spectypes.OperatorID
```
olegshmuelov, line 79: Suggesting we define type `OperatorIDFunc func() spectypes.OperatorID`
for `func() spectypes.OperatorID` due to its frequent use.
This would improve readability and ease future modifications. wdyt?

olegshmuelov (commented):
Should we merge `stage` into our current branch? There might be additional changes related to `operatorID` we haven't accounted for yet.

y0sher:
@nkryuchkov , lets fix @olegshmuelov 's comments and merge

nkryuchkov:
@olegshmuelov 

> Should we merge stage into our current branch? There might be additional changes related to operatorID we haven't accounted for yet.

I think we should. Done.

> following your changes, we can adjust

Actually I think we don't even need to pass EventHandler's fields to its methods because they also have access to these fields. Simplified.

message/validation/validation.go
nkryuchkov: I didn't do that because it's used in very different packages so I'm not sure what's best location for it

olegshmuelov (approved):

olegshmuelov (approved):

y0sher (approved):

PR #1243, opened 2023-12-14
Finalized Checkpoints: make eth event syncer to use blocks passed finalization checkpoint only

AKorpusenko:


eth/ethtest/utils_test.go
```diff
@@ -298,3 +299,19 @@ func simTestBackend(testAddresses []*ethcommon.Address) *simulator.SimulatedBack
...
+
+func setFinalizedBlocksProducer(sim *simulator.SimulatedBackend) func(ctx context.Context, finalizedBlocks chan<- uint64) error {
+	return func(ctx context.Context, finalizedBlocks chan<- uint64) error {
+		go func() {
+			heads := make(chan *ethtypes.Header)
+			sub, _ := sim.SubscribeNewHead(ctx, heads)
+			defer sub.Unsubscribe()
+
+			for {
+				header := <-heads
```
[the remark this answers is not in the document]
AKorpusenko: Fixed. Thanks

eth/executionclient/execution_client.go
```diff
@@ -47,9 +48,15 @@ type ExecutionClient struct {
...
-func New(ctx context.Context, nodeAddr string, contractAddr ethcommon.Address, opts ...Option) (*ExecutionClient, error) {
+func New(
+	ctx context.Context,
+	nodeAddr string,
+	contractAddr ethcommon.Address,
+	opts ...Option,
+) (*ExecutionClient, error) {
 	client := &ExecutionClient{
 		nodeAddr:                    nodeAddr,
+		finalizedBlocks:             make(chan uint64),
```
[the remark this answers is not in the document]
AKorpusenko: Fixed. Thanks

moshe-blox:
Haven't reviewed yet, but just noting that this likely requires a fork, otherwise operators who didn't upgrade yet would start attesting immediately after a validator was registered, while others will reject their messages for "validator doesn't exist" since they've not seen the event yet.

eth/executionclient/execution_client.go
```diff
@@ -73,20 +78,19 @@ func New(ctx context.Context, nodeAddr string, contractAddr ethcommon.Address, o
...
 func (ec *ExecutionClient) Close() error {
 	close(ec.closed)
+	close(ec.finalizedBlocks)
 	ec.client.Close()
 	return nil
 }
 
 // FetchHistoricalLogs retrieves historical logs emitted by the contract starting from fromBlock.
 func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64) (logs <-chan BlockLogs, errors <-chan error, err error) {
+	// #TODO refactor this to fetch last finalized one
```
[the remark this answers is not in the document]
AKorpusenko: As we need to do a fork, then forget about this. No need to get last finalized epoch on start

eth/executionclient/execution_client.go
```diff
@@ -306,6 +343,43 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
...
+			return fromBlock, context.Canceled
+
+		case <-ec.closed:
+			return fromBlock, ErrClosed
+
+		case toBlock, ok := <-ec.blocksChan:
+			if !ok {
+				return lastBlock, fmt.Errorf("blocksChan is closed")
+			}
+			if toBlock < ec.finalizedCheckpointActivationSlot {
```
AKorpusenko, line 359: @moshe-blox, @y0sher, 
Not sure do we need this or not.
Is panic suitable here?

AKorpusenko:
@moshe-blox, I've reworked this PR to support fork functionality
It's kinda hard to test this using local network because "finalization" is a tricky thing here, so I've tried at least to test the transition from current version  to version with finalization checkpoints

eth/eventhandler/event_handler_test.go
```diff
@@ -308,6 +322,11 @@ func TestHandleBlockEventsStream(t *testing.T) {
 		})
 	})
 
+	// Setting new blocks producer
+	fmt.Println("FORKING ON BLOCK ", lastBlock.NumberU64())
```
y0sher, line 326: remove or use logger

eth/executionclient/defaults.go
```diff
@@ -1,14 +1,16 @@
...
-	DefaultFollowDistance              = 8
-	DefaultHistoricalLogsBatchSize     = 5000
-	defaultLogBuf                      = 8 * 1024
+	DefaultConnectionTimeout                 = 10 * time.Second
+	DefaultReconnectionInitialInterval       = 1 * time.Second
+	DefaultReconnectionMaxInterval           = 64 * time.Second
+	DefaultFollowDistance                    = 8
+	DefaultHistoricalLogsBatchSize           = 5000
+	defaultLogBuf                            = 8 * 1024
+	DefaultFinalizedCheckpointActivationSlot = math.MaxUint64
```
y0sher, line 15: this should be part of network config

eth/executionclient/execution_client.go
```diff
@@ -30,35 +31,44 @@ var (
...
-	reconnectionMaxInterval     time.Duration
-	logBatchSize                uint64
+	logger                            *zap.Logger
+	metrics                           metrics
+	connectionTimeout                 time.Duration
+	reconnectionInitialInterval       time.Duration
+	reconnectionMaxInterval           time.Duration
+	logBatchSize                      uint64
+	followDistance                    uint64 // used for backward compatibility only
+	finalizedCheckpointActivationSlot uint64
```
y0sher, line 45: we should get this from the networkconfig

eth/executionclient/execution_client.go
y0sher: Lets say we're running a new network, where we sync from block 1 to block 8 , but the fork is only set to block 20, wouldn't this trigger this exit?
also how is block number and slot same in this case?

y0sher (changes requested):
can you please explain the relation between follow distance and the finalized checkpoint blocks?  why do we need both?

eth/executionclient/execution_client.go
```diff
@@ -306,6 +336,51 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
...
+			for block := range logStream {
+				logs <- block
+				lastBlock = block.BlockNumber
+			}
+			if err := <-fetchErrors; err != nil {
+				// If we get an error while fetching, we return the last block we fetched.
+				return lastBlock, fmt.Errorf("fetch logs: %w", err)
+			}
+			fromBlock = toBlock + 1
+			ec.metrics.ExecutionClientLastFetchedBlock(fromBlock)
```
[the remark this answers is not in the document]
AKorpusenko: Thanks! Refactored that a bit more. There is no duplication anymore

eth/executionclient/execution_client.go
AKorpusenko: I think passing this as an Option is ok. By default this value contains the "INF" block number

PR #1244, opened 2023-12-17
fix: (SlotTicker) potential double tick

moshe-blox:
This PR fixes a bug where the SlotTicker wakes up earlier than planned, for example 11.999 seconds into the same slot it already ticked on, and ends up ticking the same slot again.

See https://github.com/bloxapp/ssv/pull/1244#discussion_r1431331559 for more explanation.

### DIFF
--- a/operator/slotticker/slotticker.go
+++ b/operator/slotticker/slotticker.go
@@ -21,26 +21,6 @@ type Config struct {
 	GenesisTime  time.Time
 }
 
-type TimerProvider func(d time.Duration) Timer
-
-type Timer interface {
-	Stop() bool
-	Reset(d time.Duration) bool
-	C() <-chan time.Time
-}
-
-type timer struct {
-	*time.Timer
-}
-
-func NewTimer(d time.Duration) Timer {
-	return &timer{time.NewTimer(d)}
-}
-
-func (t *timer) C() <-chan time.Time {
-	return t.Timer.C
-}
-
 type slotTicker struct {
 	logger       *zap.Logger
 	timer        Timer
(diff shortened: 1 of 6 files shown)


operator/slotticker/slotticker.go
```diff
@@ -81,10 +84,15 @@ func (s *slotTicker) Next() <-chan time.Time {
...
 	}
-	slotNumber := uint64(timeSinceGenesis / s.slotDuration)
-	nextSlotStartTime := s.genesisTime.Add(time.Duration(slotNumber+1) * s.slotDuration)
+	currentSlot := phase0.Slot(timeSinceGenesis / s.slotDuration)
+	if currentSlot <= s.slot {
+		// We've already ticked for this slot, so we need to wait for the next one.
+		currentSlot = s.slot
+		s.logger.Warn("double tick", zap.Uint64("slot", uint64(currentSlot)))
+	}
+	nextSlotStartTime := s.genesisTime.Add(time.Duration(currentSlot+1) * s.slotDuration)
```
y0sher, line 93: @MatusKysel , we've seen this happen in e2e, where <-Next(), Slot(), resulted in getting the same slot twice. 
so far I wasn't able to write a test that reproduces it, might be easier if we mock the whole clock. 
So, IMO, this "fixes" the issue but with realizing exactly what caused it to happen. I'd rather we have some test included but I think this should go in anyhow.

y0sher (dismissed):

moshe-blox:
@MatusKysel and I added another test that reproduces the problem with real timers.

@y0sher can you please re-review?

y0sher (approved):
Great job!

PR #1248, opened 2023-12-21
v1.2.2

y0sher:


eth/contract/generate.go
```diff
@@ -1,5 +1,6 @@
 // Package contract generates the contract bindings for v1 contract.
 package contract
 
+// contract.abi is taken from https://github.com/bloxapp/ssv-network/blob/contract-abi/docs/dev/v1.0.0.rc5/abi/SSVNetwork.json
```
y0sher, line 4: is this the current version of the contract to use for next release? @mtabasco , @moshe-blox , @lior-blox , before deploying this we should make sure prod and is aligned with this.

y0sher: @moshe-blox @mtabasco @lior-blox 
*NOTE*: this version doesn't seem to be the version on mainnet / prod. we should probably update here to be the correct version.

y0sher: latest contract version is this - https://github.com/bloxapp/ssv-network/blob/contract-abi/docs/testnet/v1.1.0/abi/SSVNetwork.json
we should probably update.

y0sher: https://github.com/bloxapp/ssv/pull/1254

moshe-blox (approved):

olegshmuelov (approved):

PR #1253, opened 2023-12-26
Message validation estimated round checks

nkryuchkov:
- Rewrite estimated round calculation logic according to changes in https://github.com/bloxapp/ssv/pull/1120
- Enable lowest allowed round check


message/validation/consensus_validation.go
```diff
@@ -147,6 +147,24 @@ func (mv *messageValidator) validateConsensusMessage(
 	return consensusDescriptor, msgSlot, nil
 }
 
+func (mv *messageValidator) allowedRoundRange(estimatedRound specqbft.Round, maxRound specqbft.Round) (specqbft.Round, specqbft.Round) {
+	var lowestAllowed, highestAllowed specqbft.Round
+
+	if lowestAllowed > allowedRoundsInPast {
```
y0sher, line 153: this is here in case `allowedRoundsInPast` is changed? since `lowestAllowed` will always be 0

message/validation/roundthresholds/round_thresholds.go
```diff
@@ -0,0 +1,87 @@
...
+	rt := roundtimer.New(unusedCtx, c.bn, role, nil)
+
+	var cumulativeDuration time.Duration
+	round := specqbft.Round(1)
+	c.thresholds[role] = []time.Duration{}
+
+	for {
+		roundDuration := rt.RoundDuration(round)
+		if roundDuration <= 0 {
+			panic(fmt.Sprintf("invalid round duration for round %d", round))
```
y0sher, line 46: don't panic, either return error or use logger to fatal if this is a really fatal case. can this never happen ?

message/validation/consensus_validation.go
nkryuchkov: this must be `if estimatedRound > allowedRoundsInPast {`

message/validation/roundthresholds/round_thresholds.go
nkryuchkov: Round duration must never be negative/zero, so this case would mean that there's a critical issue in the round timer that must be noticed immediately, so panic fits best in my opinion. I agree that fatal as a workaround for panic would be better

message/validation/roundthresholds/round_thresholds.go
```diff
@@ -0,0 +1,87 @@
...
+	beaconprotocol "github.com/bloxapp/ssv/protocol/v2/blockchain/beacon"
+	"github.com/bloxapp/ssv/protocol/v2/qbft/roundtimer"
+)
+
+// Cache contains thresholds when each round finishes for each role.
+type Cache struct {
+	bn         beaconprotocol.BeaconNetwork
+	thresholds map[spectypes.BeaconRole][]time.Duration
+	mu         sync.Mutex
+}
```
y0sher, line 21: does this cache really needs to be mutex protected? I see Init is called once in the beginning and after that it only gets read. also it seems like that cache is not the right name for this in that case, maybe its a mapping or legend.

y0sher (changes requested):
small changes, lets put some effort into testing this. @nkryuchkov

message/validation/roundthresholds/round_thresholds.go
nkryuchkov: Actually we can remove it, it's a leftover from a previous state of code. Ok, I'll rename it to mapping

nkryuchkov:
I tested this PR and while performance looked good, I saw significant amount of messages violating estimated round rules. Their timestamp looked like the rule was applied correctly, but the reason for the odd timestamp is unknown.

PR #1257, opened 2023-12-28
Authorized endpoints

pavelkrolevets:
* add `/v1/node/sign` route to sign using RSA private key on incoming data

1. Set some secret string - `SecretToken` env variable at SSV
2. Generate JWT token "HS256" at client side (any language)
3. POST body `{data: [hex string]}` to `/v1/node/sign with`  `-H "Authorization: Bearer JWT token"`
4. Get response `{signature: [hex string]}` with RSA sig on data


api/handlers/node.go
```diff
@@ -78,6 +84,14 @@ type healthCheckJSON struct {
 	} `json:"advanced"`
 }
 
+type signRequestJSON struct {
+	Data string `json:"data"`
+}
+
+type signResponseJSON struct {
+	Sig string `json:"sig"`
```
y0sher, line 92: lets change the name to `signature`

y0sher (approved):

api/handlers/node.go
```diff
@@ -78,6 +81,14 @@ type healthCheckJSON struct {
...
 }
 
+type SignRequestJSON struct {
+	Data string `json:"data"`
+}
+
+type SignResponseJSON struct {
+	Signature string `json:"signature"`
+}
+
```
moshe-blox, line 91: For consistency, let's put these as anonymous structs in the request handler method, like in:
https://github.com/bloxapp/ssv/blob/c567481d43c0727a81abbaea4846b8a61f7a1df7/api/handlers/validators.go#L22-L29

There's no reason to expose them outside at the moment. When they are used in tests, we can construct them from there as anonymous structs as well (its really only 1 field).

In the future (SSV API v2) we'll have an API `client` package so we'd be able to just call `Sign()` method from tests, rather than encoding JSONs :)

api/handlers/node.go
```diff
@@ -38,7 +41,7 @@ type connectionJSON struct {
 	Direction string `json:"direction"`
 }
 
-type peerJSON struct {
+type PeerJSON struct {
```
moshe-blox, line 44: Unrelated to this PR, but if it already exposes this then let's give it a nice name, just `Peer` should be enough :)

api/server/server.go
```diff
@@ -19,44 +20,54 @@ type Server struct {
...
+	server := &http.Server{
+		Addr:         s.addr,
+		Handler:      s.SetRoutes(),
+		ReadTimeout:  12 * time.Second,
+		WriteTimeout: 12 * time.Second,
+	}
+	return server.ListenAndServe()
+}
+
+func (s *Server) SetRoutes() chi.Router {
```
moshe-blox, line 53: This should be a private method since it does not need to be called from outside.

I think we can just call it `routes()` since it also returns the routes rather than just 'set' them.

api/handlers/node.go
```diff
@@ -165,10 +177,37 @@ func (h *Node) Health(w http.ResponseWriter, r *http.Request) error {
 	return api.Render(w, r, resp)
 }
 
-func (h *Node) peers(peers []peer.ID) []peerJSON {
-	resp := make([]peerJSON, len(peers))
+func (h *Node) Sign(w http.ResponseWriter, r *http.Request) error {
+	// TODO: there is a limit to amount of data can be signed at once. Or brake down to chunks
```
y0sher, line 181: what happens now if we try to sign a "too" big data?

y0sher: I saw the error

y0sher (dismissed):

api/handlers/node.go
```diff
@@ -47,7 +50,7 @@ type peerJSON struct {
 	Version       string           `json:"version"`
 }
 
-type identityJSON struct {
+type IdentityJSON struct {
```
moshe-blox, line 53: Same as PeerJSON, here `NodeIdentity` is a more appropriate name.

moshe-blox:
Good job! 👏

Left a few comments

moshe-blox (changes requested):

api/handlers/node.go
```diff
@@ -165,10 +169,40 @@ func (h *Node) Health(w http.ResponseWriter, r *http.Request) error {
 	return api.Render(w, r, resp)
 }
 
-func (h *Node) peers(peers []peer.ID) []peerJSON {
-	resp := make([]peerJSON, len(peers))
+func (h *Node) Sign(w http.ResponseWriter, r *http.Request) error {
+	// TODO: there is a limit to amount of data can be signed at once. Or brake down to chunks
+	rawdata, err := io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
pavelkrolevets: if we use not exact then we cant unmarshall json - so I count  `{"hashed_data":"hash 32 bytes"}` as 82 bytes

api/handlers/node.go
```diff
@@ -165,10 +169,40 @@ func (h *Node) Health(w http.ResponseWriter, r *http.Request) error {
 	return api.Render(w, r, resp)
 }
 
-func (h *Node) peers(peers []peer.ID) []peerJSON {
-	resp := make([]peerJSON, len(peers))
+func (h *Node) Sign(w http.ResponseWriter, r *http.Request) error {
+	// TODO: there is a limit to amount of data can be signed at once. Or brake down to chunks
```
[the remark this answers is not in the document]
pavelkrolevets: done

api/server/server.go
```diff
@@ -48,15 +63,13 @@ func (s *Server) Run() error {
...
-		Handler:      router,
-		ReadTimeout:  12 * time.Second,
-		WriteTimeout: 12 * time.Second,
-	}
-	return server.ListenAndServe()
+	// Authorized endpoints
+	router.Group(func(router chi.Router) {
+		router.Use(jwtauth.Verifier(s.tokenAuth))
+		router.Use(jwtauth.Authenticator(s.tokenAuth))
+		router.Post("/v1/operator/sign", api.Handler(s.node.Sign))
```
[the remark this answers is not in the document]
pavelkrolevets: done

moshe-blox:
@MatusKysel @pavelkrolevets 
feature request: an optional query argument to change signing method to PSS

PR #1259, opened 2023-12-29
Added p2p metrics to existing dashboards

AKorpusenko:


y0sher (dismissed):
Approving this but I think @zktaiga and @Roy-blox are more appropriate reviewers.

zktaiga (changes requested):
These current dashboards seem to be expecting datasources which are internal. Can you please export externally via Grafana so it will template the datasource for the end user?

zktaiga (approved):

moshe-blox:
Why are there so many unrelated changes? For example this seems like a significant change:
https://github.com/bloxapp/ssv/pull/1259/files#diff-f244de1db4725c6e264c581853904a0a4e9f36fd4951673f1a41b2c5ba87198dL740-L773

Was the dashboard in the repo out of date or did we make our changes on an out of date dashboard in our Grafana?

@zktaiga can you confirm if the node dashboard here is as up to date as our mainnet? (for example)

- [Mainnet Node Dashboard](https://grafana.bloxinfra.com/d/QNiM324rdoVz/mainnet-node-dashboard?orgId=1&var-instance=ssv-full-node-mainnet-1)
- [Message Validation](https://grafana.bloxinfra.com/d/DppaYPgSk/message-validation?orgId=1)

AKorpusenko:
> Why are there so many unrelated changes? For example this seems like a significant change: https://github.com/bloxapp/ssv/pull/1259/files#diff-f244de1db4725c6e264c581853904a0a4e9f36fd4951673f1a41b2c5ba87198dL740-L773

@moshe-blox, there are several values there on chart "Validation Failure Reason", which are listed and in the current stage version and only one of the values of this chart is visible. I've fixed this and showing all of them by default

Stage:
image


This PR:
image

AKorpusenko:
@moshe-blox, I've mostly just added p2p metrics and added rows for grouping charts (mixing existing and new ones for better accessibility)

monitoring/grafana/dashboard_msg_validation.json
```diff
@@ -737,46 +822,13 @@
...
-                "id": "custom.hideFrom",
-                "value": {
-                  "legend": false,
-                  "tooltip": false,
-                  "viz": true
-                }
-              }
-            ]
-          }
-        ]
```
moshe-blox, line 773: @zktaiga there seem to be a few changes like this that aren't related to what's being done in this PR, are they not consequential?

zktaiga: > @zktaiga there seem to be a few changes like this that aren't related to what's being done in this PR, are they not consequential?

Maybe it should have been mentioned but I've just verified and these are just minor fixes which are just generic QoL improvements. In this particular case, a list was defaulting a specific item menu, and now it just shows all the items in the list.

Overall the change looks good to me.

PR #1266, opened 2024-01-04
test: add new test for slotTicker

MatusKysel:


operator/slotticker/slotticker_test.go
```diff
@@ -178,3 +179,54 @@ func TestSlotSkipping(t *testing.T) {
...
+	}, &mockTimerProvider{C: mockTimerChan})
+
+	// Manually fire the timer twice to simulate rapid ticks
+	mockTimerChan <- time.Now()
+	mockTimerChan <- time.Now()
+
+	// Call Next() twice to process the ticks
+	<-ticker.Next()
+	<-ticker.Next()
+
```
y0sher, line 219: shouldn't we have also different slots between these calls? I would add a check for this. IMO that's the actual test more than checking the logs.

y0sher (approved):

PR #1268, opened 2024-01-08
Don't check round change messages for duplicated full data

nkryuchkov:
This PR fixes a case when node receives proposal or decided messages in round 2 and compares them with full data of round change messages for round 2, resulting in a validation error because full data differs.

Changes:
- Don't check round change messages for duplicated full data


message/validation/consensus_validation.go
```diff
@@ -317,10 +317,17 @@ func (mv *messageValidator) validateBeaconDuty(
 	return nil
 }
 
-func (mv *messageValidator) hasFullData(signedMsg *specqbft.SignedMessage) bool {
+func (mv *messageValidator) needDuplicatedFullDataCheck(signedMsg *specqbft.SignedMessage) bool {
+	return (signedMsg.Message.MsgType == specqbft.ProposalMsgType ||
+		mv.isDecidedMessage(signedMsg)) &&
+		len(signedMsg.FullData) != 0
```
GalRogozinski, line 323: I don't think we should have a case that fullData is empty for such message types

message/validation/consensus_validation.go
```diff
@@ -244,7 +244,7 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 	}
 
 	if msgSlot == signerState.Slot && msgRound == signerState.Round {
-		if mv.hasFullData(signedMsg) && signerState.ProposalData != nil && !bytes.Equal(signerState.ProposalData, signedMsg.FullData) {
+		if mv.needDuplicatedFullDataCheck(signedMsg) && signerState.ProposalData != nil && !bytes.Equal(signerState.ProposalData, signedMsg.FullData) {
```
GalRogozinski, line 247: so before the error occured because the fullData of a different message type (non-proposal)?

nkryuchkov: The error occurred before because `fullData` was checked for:
- proposal
- decided
- round change

Here error doesn't occur and `fullData` is checked for:
- proposal
- decided

So `fullData` duplication check for round change caused issues.

message/validation/consensus_validation.go
nkryuchkov: Should I create a new rule that `fullData` must not be empty for proposal and decided and check it before full data hash check?

GalRogozinski: if full data is missing for such msg types that also is an invalid message imo

message/validation/consensus_validation.go
GalRogozinski: I think the logic is fine.
I am trying to understand why the error showed up though...
You usually saw the error on the first round correct (that's what I remember but not sure)?

It makes sense that maybe on a new round we get a different full data...
But not sure why it happened on a first round

message/validation/consensus_validation.go
nkryuchkov: I agree. Fixed

message/validation/consensus_validation.go
nkryuchkov: I usually saw error happening on proposal or decided messages in round 2 because full data was compared to round change messages for round 2

GalRogozinski: this makes more sense I think

GalRogozinski (approved):
maybe some unit tests should be included or changed?
I am not an SSV TL so I am not insisting in my review

nkryuchkov:
> maybe some unit tests should be included or changed? 

done

message/validation/consensus_validation.go
y0sher: I remember message type used to say commit on most of this error, so it happened when we had round change after a commit?

nkryuchkov: no, it happened on commit after round change, because round change full data was remembered, so commit full data was compared to round change full data

MatheusFranco99 (approved):
Good work!

MatheusFranco99 (approved):
LGTM

olegshmuelov (approved):
LGTM

y0sher:
@nkryuchkov @MatheusFranco99 @moshe-blox 
Is this also part of new message validation? I think we should bring this in to genesis and new impls.

nkryuchkov:
> @nkryuchkov @MatheusFranco99 @moshe-blox 
> 
> Is this also part of new message validation? I think we should bring this in to genesis and new impls.

As I remember it was implemented in the new message validation. It would be nice to have in the old one, but its lifecycle will be shorter than this PR is opened

nkryuchkov:
This PR is for old message validation. The new message validation handles this case by saving full data only for proposal messages: https://github.com/ssvlabs/ssv/blob/v2.3.1/message/validation/consensus_validation.go#L321

PR #1269, opened 2024-01-08
Compute cluster hash wrong calculation

guym-blox:


protocol/v2/types/ssvshare.go
```diff
@@ -67,25 +67,32 @@ func (s *SSVShare) SetFeeRecipient(feeRecipient bellatrix.ExecutionAddress) {
...
-			return nil, err
-		}
+		encodedData = append(encodedData, abiEncodeUint64(id)...)
 	}
 
-	return hash.Sum(nil), nil
+	// Compute the keccak256 hash of the encoded data
+	encrypted := crypto.Keccak256(encodedData)
+
+	return encrypted, nil
```
y0sher, line 88: a bit misleading, its a hash not encryption, change to - hashed / hashedCluster / clusterHash

y0sher (approved):

protocol/v2/types/ssvshare_test.go
```diff
@@ -19,6 +21,24 @@ func TestSSVShare_BelongsToOperator(t *testing.T) {
 	require.True(t, metadata.BelongsToOperator(1))
 	require.False(t, metadata.BelongsToOperator(2))
 }
+func TestSSVShare_ComputeClusterIDHash(t *testing.T) {
+	var (
+		aliceClusterHash = "fec0dc0e4be38e074cdaad0dff2375159010d56425f82d926468341c2f7afd3e"
```
moshe-blox, line 26: Where did we get this excepted output from?

I'm getting a different output from this Solidity code:
```solidity

contract Test {
    // State variables
    address constant myAddress = 0x71562b71999873DB5b286dF957af199Ec94617F7;
    uint[] operatorIds;

    constructor() {
        // Initialize operatorIds
        operatorIds.push(1);
        operatorIds.push(2);
        operatorIds.push(3);
        operatorIds.push(4);
    }

    function get() public view returns (bytes32) {
        return keccak256(abi.encodePacked(myAddress, operatorIds));
    }
}
```

Gives `0xa341933234aa1e6dfd3b8d6677172bdcd0986b1e6afc2e84d321f154d9736717`, unless I got something wrong.

Maybe can ask @mtabasco to give us an expected output for a given address + operator IDs, just to be sure.

moshe-blox (changes requested):

protocol/v2/types/ssvshare_test.go
mtabasco: @moshe-blox please confirm but I think the clusterID and its hashed value only apply here. It's different from the hashed cluster we manage via contracts (that includes `validatorCount`, `active`, etc).

y0sher: @mtabasco yes this is hash cluster id only ,not cluster data

moshe-blox: @mtabasco my Solidity code is based on Alon's comment here: https://github.com/bloxapp/ssv/issues/1262

Is it correct?

moshe-blox (dismissed):
Well done! 💪

y0sher (dismissed):

moshe-blox (approved):

PR #1274, opened 2024-01-12
Operator keys encapsulation

nkryuchkov:
Encapsulate all operator keys operations in a single package.

Note: Since storage and EKM use different hashes, there are two private key methods:
- `StorageHash`
- `EKMHash`


message/validation/rsa.go
```diff
@@ -28,26 +24,17 @@ func (mv *messageValidator) verifyRSASignature(messageData []byte, operatorID sp
 			return e
 		}
 
-		operatorPubKey, err := base64.StdEncoding.DecodeString(string(operator.PublicKey))
+		rsaPubKey, err = keys.PublicKeyFromString(string(operator.PublicKey))
```
y0sher: One point of this PR is that now the rest of the code that is not in `keys` doesn't care about which priv/pub key the operator uses. we can now remove traces of the RSA name from this function.

cli/operator/node.go
```diff
@@ -118,16 +114,32 @@ var StartNodeCmd = &cobra.Command{
 			logger.Fatal("could not setup db", zap.Error(err))
 		}
 
-		nodeStorage, operatorData := setupOperatorStorage(logger, db)
+		var operatorPrivKey keys.OperatorPrivateKey
+		if cfg.KeyStore.PrivateKeyFile != "" {
+			operatorPrivKey, err = keys.PrivateKeyFromFile(cfg.KeyStore.PrivateKeyFile, cfg.KeyStore.PasswordFile)
```
y0sher, line 119: I rather we take out the file logic from the keys package. read the files here and pass raw strings into the function.

nkryuchkov: @y0sher I don't mind. Do you think we should also extract the JSON decryption logic with `keystorev4.New().Decrypt` or something like the below looks fine to you?

```go
if cfg.KeyStore.PrivateKeyFile != "" {
	// nolint: gosec
	encryptedJSON, err := os.ReadFile(cfg.KeyStore.PrivateKeyFile)
	if err != nil {
		logger.Fatal("could not read PEM file", zap.Error(err))
	}

	// nolint: gosec
	keyStorePassword, err := os.ReadFile(cfg.KeyStore.PasswordFile)
	if err != nil {
		logger.Fatal("could not read password file", zap.Error(err))
	}

	operatorPrivKey, err = keys.PrivateKeyFromEncryptedJSON(encryptedJSON, string(keyStorePassword))
	if err != nil {
		logger.Fatal("could not extract operator private key from file", zap.Error(err))
	}
} else {
	operatorPrivKey, err = keys.PrivateKeyFromString(cfg.OperatorPrivateKey)
	if err != nil {
		logger.Fatal("could not decode operator private key", zap.Error(err))
	}
}
```

olegshmuelov (dismissed):

y0sher (dismissed):

cli/operator/node.go
y0sher: I fixed this here https://github.com/bloxapp/ssv/pull/1274/commits/23cb56903dd4b83c5e4071e188e9462e8c123b15

moshe-blox (approved):

y0sher (approved):

PR #1275, opened 2024-01-14
removed any reference for operator id in node metadata

guym-blox:


y0sher (changes requested):
fix lint and test and write a test to make sure this is not a fork. (try to decode/ecoode when expecting the different versions

y0sher (approved):
Great! and no fork!

moshe-blox (approved):

PR #1279, opened 2024-01-18
Use libp2p conn manager to trim peers

nkryuchkov:
As per title

### DIFF
--- a/network/peers/conn_manager.go
+++ b/network/peers/conn_manager.go
@@ -67,20 +67,10 @@ func (c connManager) TagBestPeers(logger *zap.Logger, n int, mySubnets records.S
 }
 
 func (c connManager) TrimPeers(ctx context.Context, logger *zap.Logger, net libp2pnetwork.Network) {
-	allPeers := net.Peers()
-	before := len(allPeers)
-	// TODO: use libp2p's conn manager once ready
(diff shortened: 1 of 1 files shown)


nkryuchkov:
@iurii-ssv as you had recently been working on `TrimPeers`, do you think we might need to trim peers using libp2p or we should close this PR?

iurii-ssv:
> @iurii-ssv as you had recently been working on `TrimPeers`, do you think we might need to trim peers using libp2p or we should close this PR?

Right, we have "our own trimming implementation now" that takes into account various aspects of managing incoming/outgoing connections such that we have the lowest amount of dead/solo subnets possible.

Looking at `func (cm *BasicConnMgr) getConnsToClose()` (that will be called as the result of `c.connManager.TrimOpenConns(ctx)` call) it seems it contains quite a bit of logic that's probably not beneficial to us (probably even harmful) - such as "not trimming peers in the grace period" - and it's probably not a good idea to spread out peer-management & trimming logic even further throughout various parts of code ... so I don't think it's a good idea to use `c.connManager.TrimOpenConns(ctx)` (it would be a fine default if we didn't need our peer-management solution like I mentioned above).

So I think we no longer need this PR and can close it.

PR #1286, opened 2024-01-24
Validator controller refactoring

nkryuchkov:


y0sher:
we're not going through with this for now, might visit this again after CC.

PR #1293, opened 2024-01-28
fix: timeout in go-eth2-client's `Validators` method

moshe-blox:
Latest go-eth2-client sometimes calls `BeaconState` in `Validators` method for faster retrieval of large validator sets. This call is heavy and exceeds the 5s timeout we've had so far.

Although timeout can be passed in each call, it's not obeyed if the client's timeout is lower (for why, see https://github.com/attestantio/go-eth2-client/pull/101)

Changes:
- Increase go-eth2-client's timeout from 5s to 30s, and to preserve current behavior — pass 5s to every call, except for `Validators` which gets the maximum 30s.
- Preflight check for open debug endpoints on Prysm nodes, so we can fail early if Prysm doesn't allow `BeaconState` calls (see https://github.com/attestantio/go-eth2-client/pull/100)
- Added unit test to check the various timeout conditions.

Note: maxTimeout is set to 2 min because I noticed that a Lighthouse v4.5.0 node takes roughly 1.5 min to provide BeaconState.

Update: see https://github.com/bloxapp/ssv/pull/1293#issuecomment-1916536035

### DIFF
--- a/beacon/goclient/goclient.go
+++ b/beacon/goclient/goclient.go
@@ -278,6 +278,7 @@ func (gc *goClient) commonOpts() api.CommonOpts {
 }
 
 func (gc *goClient) checkPrysmDebugEndpoints() error {
+	start := time.Now()
 	address := strings.TrimSuffix(gc.client.Address(), "/")
 	if !strings.HasPrefix(address, "http") {
 		address = fmt.Sprintf("http://%s", address)
@@ -303,5 +304,6 @@ func (gc *goClient) checkPrysmDebugEndpoints() error {
 	if data.JustifiedCheckpoint.Root == (phase0.Root{}) {
 		return fmt.Errorf("no justified checkpoint found")
 	}
+	gc.log.Debug("Prysm debug endpoints are enabled", zap.Duration("took", time.Since(start)))
 	return nil
 }
--- a/beacon/goclient/goclient.go
+++ b/beacon/goclient/goclient.go
@@ -2,6 +2,7 @@ package goclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"strings"
@@ -141,18 +142,26 @@ type goClient struct {
 	registrationMu       sync.Mutex
 	registrationLastSlot phase0.Slot
 	registrationCache    map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration
+	commonTimeout        time.Duration
+	maxTimeout           time.Duration
 }
 
 // New init new client and go-client instance
 func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.OperatorID, slotTickerProvider slotticker.Provider) (beaconprotocol.BeaconNode, error) {
-	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr), fields.Network(string(opt.Network.BeaconNetwork)))
+	const (
+		commonTimeout = time.Second * 5
+		maxTimeout    = time.Second * 30
+	)
 
-	httpClient, err := http.New(opt.Context,
+	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr), fields.Network(string(opt.Network.BeaconNetwork)))
+	dialCtx, cancel := context.WithTimeout(opt.Context, commonTimeout)
+	defer cancel()
+	httpClient, err := http.New(dialCtx,
 		// WithAddress supplies the address of the beacon node, in host:port format.
 		http.WithAddress(opt.BeaconNodeAddr),
 		// LogLevel supplies the level of logging to carry out.
 		http.WithLogLevel(zerolog.DebugLevel),
-		http.WithTimeout(time.Second*5),
+		http.WithTimeout(maxTimeout),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create http client: %w", err)
@@ -167,10 +176,14 @@ func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.Op
 		gasLimit:          opt.GasLimit,
 		operatorID:        operatorID,
 		registrationCache: map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration{},
+		commonTimeout:     commonTimeout,
+		maxTimeout:        maxTimeout,
 	}
 
 	// Get the node's version and client.
-	nodeVersionResp, err := client.client.NodeVersion(opt.Context, &api.NodeVersionOpts{})
+	nodeVersionResp, err := client.client.NodeVersion(opt.Context, &api.NodeVersionOpts{
+		Common: client.commonOpts(),
+	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to get node version: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


beacon/goclient/goclient.go
```diff
@@ -187,6 +200,29 @@ func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.Op
 		zap.String("version", client.nodeVersion),
 	)
 
+	// If it's a Prysm, check that the debug endpoints are enabled, otherwise
+	// the Validators method might fail when calling BeaconState.
+	// TODO: remove this once Prysm enables debug endpoints by default.
+	if client.nodeClient == NodePrysm {
```
y0sher, line 206: just in terms of how it looks inside the `New` function, I would move that part to a separate function.

beacon/goclient/goclient.go
```diff
@@ -141,18 +142,26 @@ type goClient struct {
...
-	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr), fields.Network(string(opt.Network.BeaconNetwork)))
+	const (
+		commonTimeout = time.Second * 5
+		maxTimeout    = time.Second * 120
+	)
 
-	httpClient, err := http.New(opt.Context,
+	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr), fields.Network(string(opt.Network.BeaconNetwork)))
+	dialCtx, cancel := context.WithTimeout(opt.Context, commonTimeout)
+	defer cancel()
```
nkryuchkov, line 158: The context gets canceled at the end of `New`, but it's passed to `httpClient`, which is used in `goClient` for beacon calls. Is this intended behavior?

beacon/goclient/goclient.go
```diff
@@ -187,6 +200,29 @@ func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.Op
...
+	// the Validators method might fail when calling BeaconState.
+	// TODO: remove this once Prysm enables debug endpoints by default.
+	if client.nodeClient == NodePrysm {
+		// BeaconState is quite heavy to call, checking ForkChoice should be enough.
+		resp, err := client.client.(eth2client.ForkChoiceProvider).ForkChoice(opt.Context, &api.ForkChoiceOpts{
+			Common: api.CommonOpts{Timeout: client.maxTimeout},
+		})
+		if err != nil {
+			var apiErr *api.Error
+			if errors.As(err, &apiErr) && apiErr.StatusCode == 404 {
```
nkryuchkov, line 213: 404 is well-known code, however perhaps this would be a bit more readable:

```suggestion
			if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusNotFound {
```

beacon/goclient/goclient.go
moshe-blox: @nkryuchkov only `dialCtx` is canceled and it's passed only to `eth2clienthttp.New` which doesn't keep it for later (like we do in `goClient` which is wrong by the way 😅)

beacon/goclient/goclient.go
moshe-blox: done

beacon/goclient/goclient.go
moshe-blox: done

beacon/goclient/goclient.go
```diff
@@ -141,18 +142,26 @@ type goClient struct {
...
+	maxTimeout           time.Duration
 }
 
 // New init new client and go-client instance
 func New(logger *zap.Logger, opt beaconprotocol.Options, operatorID spectypes.OperatorID, slotTickerProvider slotticker.Provider) (beaconprotocol.BeaconNode, error) {
-	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr), fields.Network(string(opt.Network.BeaconNetwork)))
+	const (
+		commonTimeout = time.Second * 5
+		maxTimeout    = time.Second * 120
+	)
```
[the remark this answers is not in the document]
moshe-blox: done

y0sher (approved):
gj

olegshmuelov (dismissed):

beacon/goclient/goclient.go
nkryuchkov: @moshe-blox As I understand, `eth2clienthttp.New` keeps it for later, not as a struct field, but instead it spawns a goroutine that closes connection on context done:

https://github.com/attestantio/go-eth2-client/blob/master/http/service.go#L138-L143

moshe-blox:
### Update

Since go-eth2-client merged those 2 PRs:
- https://github.com/attestantio/go-eth2-client/pull/102
- https://github.com/attestantio/go-eth2-client/pull/101

I upgraded to the latest commit and we can use `WithTimeout` of 5 seconds again, since it's now possible to pass a higher timeout per-call.

beacon/goclient/goclient.go
moshe-blox: @nkryuchkov woah, good catch 🎯

see https://github.com/bloxapp/ssv/pull/1293#issuecomment-1916536035

beacon/goclient/goclient.go
```diff
@@ -24,9 +26,15 @@ import (
...
-// Don't check for it, check for errors or nil data instead.
-const DataVersionNil spec.DataVersion = math.MaxUint64
+const (
+	// DataVersionNil is just a placeholder for a nil data version.
+	// Don't check for it, check for errors or nil data instead.
+	DataVersionNil spec.DataVersion = math.MaxUint64
+
+	// Client timeouts.
+	commonTimeout = time.Second * 5
+	maxTimeout    = time.Second * 120
```
nkryuchkov, line 36: IMO "max" is not the best description of this timeout, I'd prefer something like `longTimeout` because when I saw "common" and "max" first time I didn't understand what's the purpose of "max" just from its name. But that's subjective, you might disagree here

nkryuchkov (dismissed):

beacon/goclient/goclient.go
moshe-blox: https://github.com/bloxapp/ssv/blob/a25c4fb942b7c1a2d8b9ab0c74821ab19b4e45a9/beacon/goclient/goclient.go#L34-L36

nkryuchkov (dismissed):

nkryuchkov (approved):

beacon/goclient/mock-beacon-responses.json
nkryuchkov, line 1: There's a convention in Go to put fixtures into `testdata` folder (https://dave.cheney.net/2016/05/10/test-fixtures-in-go), so I'd move it there because `goclient` has many files

nkryuchkov (approved):

beacon/goclient/mock-beacon-responses.json
moshe-blox: done

beacon/goclient/goclient_test.go
```diff
@@ -0,0 +1,138 @@
...
+			BeaconStateDelay: longTimeout * 2,
+		})
+		client, err := mockClient(t, ctx, unresponsiveServer.URL, commonTimeout, longTimeout)
+		require.NoError(t, err)
+
+		require.NoError(t, err)
+		_, err = client.(*goClient).GetValidatorData(nil) // Should call BeaconState internally.
+		require.ErrorContains(t, err, "context deadline exceeded")
+
+		duties, err := client.(*goClient).ProposerDuties(ctx, 132502, nil)
```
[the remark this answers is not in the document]
moshe-blox: its in testdata, from a Sepolia BN
moved to constant

nkryuchkov (approved):

PR #1294, opened 2024-01-28
feat: Save e2e logs artifact in gh actions on test failure

guym-blox:
This PR introduces two key enhancements to our CI/CD process within GitHub Actions. Firstly, it implements a feature to automatically save container logs whenever an end-to-end (e2e) test fails. Secondly, it adds flexibility to our e2e testing framework by allowing the execution of tests across a dynamic mix of nodes. Specifically, it enables the configuration of two nodes to run code from the current branch and two additional nodes to operate using a Docker image specified through the workflow dispatch input node_group_1_image_tag.


.github/workflows/e2e-test.yml
```diff
@@ -1,15 +1,17 @@
...
-    workflows: [Lint, Unit tests]
+    workflows: ["Lint", "Unit tests"]
     types: [completed]
     branches: [stage, main]
-  #  pull_request:
-  #    branches:
-  #      - "**"
 
-  workflow_dispatch:
+#  workflow_dispatch:
```
y0sher, line 14: we want to bring this back before merging

.github/workflows/e2e-test.yml
```diff
@@ -1,15 +1,17 @@
 name: E2E Test
 
 on:
+  pull_request:
+    branches:
+      - stage
+      - main
```
y0sher, line 7: maybe lets leave main only here. for stage we'll run manually.

e2e/run.sh
```diff
@@ -1,5 +1,32 @@
...
+
+save_logs() {
+  echo "Creating directory at: $LOG_DIR"
+  mkdir -p "$LOG_DIR"
+  # Saving logs of each container separately
+  docker compose logs ssv-node-1 > "$LOG_DIR/ssv-node-1.txt"
+  docker compose logs ssv-node-2 > "$LOG_DIR/ssv-node-2.txt"
+  docker compose logs ssv-node-3 > "$LOG_DIR/ssv-node-3.txt"
+  docker compose logs ssv-node-4 > "$LOG_DIR/ssv-node-4.txt"
+  docker compose logs beacon_proxy > "$LOG_DIR/beacon_proxy.txt"
```
y0sher, line 27: what about logs catcher?

.github/workflows/e2e-test.yml
```diff
@@ -31,6 +33,20 @@ jobs:
...
           ./run.sh
-
       - name: clean up
         run: docker compose down -v
+
+      - name: Collect docker logs on failure
+        if: failure()
+        uses: jwalton/gh-docker-logs@v2
+        with:
+          dest: './logs'
```
y0sher, line 43: why we need this if we save the logs in the `./run.sh` script?

guym-blox: @y0sher why do we need what?

guym-blox: @y0sher this one: jwalton/gh-docker-logs@v2

is collectiong docker logs and upload them as artifacts.
the one in the run.sh is for local usage

y0sher: well, I think its best we do it the same locally and in gh if possible so we have similar results. but I assume you tested and it looks similar ?

guym-blox: yes,
this library not so smart...
just loop over containers and save the logs

.github/workflows/e2e-test.yml
```diff
@@ -1,15 +1,19 @@
 name: E2E Test
 
 on:
+  pull_request:
+    branches:
+      - main
+
+  workflow_dispatch:
+    branches:
+      - stage
```
y0sher, line 10: we want to be able to run it manually on any branch not just stage, why it as empty.

e2e/run.sh
```diff
@@ -1,15 +1,56 @@
...
+
+  # Special handling for logs_catcher to get the most recent container
+  logs_catcher_container=$(docker ps -a --filter ancestor=logs_catcher:latest --format "{{.Names}}" | head -n 1)
+  if [ ! -z "$logs_catcher_container" ]; then
+    echo "Saving logs for the most recent logs_catcher container: $logs_catcher_container..."
+    docker logs "$logs_catcher_container" > "$LOG_DIR/$logs_catcher_container.txt"
+  fi
+}
+
+export BEACON_NODE_URL=http://bn-h-2.stage.bloxinfra.com:3502/
```
nkryuchkov, line 52: I think the addresses may change, let's create a constant in the beginning of file

nkryuchkov (commented):
There are commits overlapping with https://github.com/bloxapp/ssv/pull/1339. Please rebase and change base branch if needed

y0sher (approved):

PR #1305, opened 2024-02-04
fix: get own sync committee duties after validators setup.

y0sher:
#### Passed E2E

- [x] - [run log](https://github.com/bloxapp/ssv/actions/runs/7775676172)

## Description

This PR makes `AllActiveIndices` (optionally) wait for a channel that is closed only when an initial validators setup has already happened.

It aims to solve a situation where an operator that asked for sync-committee duties before validators setup will think it does not have any validators with sync-committee duties for a whole period, resulting in a lot of missing sync committee duties.

made with @moshe-blox

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -761,40 +761,6 @@ func (c *controller) onShareStart(share *ssvtypes.SSVShare) (bool, error) {
 	return c.startValidator(v)
 }
 
-//
-//func (c *controller) onShareStartValidator() (bool, error) {
-//	if !share.HasBeaconMetadata() { // fetching index and status in case not exist
-//		c.logger.Warn("skipping validator until it becomes active", fields.PubKey(share.ValidatorPubKey))
-//		return false, nil
-//	}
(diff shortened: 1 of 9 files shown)


operator/validator/controller.go
```diff
@@ -631,7 +635,10 @@ func (c *controller) CommitteeActiveIndices(epoch phase0.Epoch) []phase0.Validat
 	return indices
 }
 
-func (c *controller) AllActiveIndices(epoch phase0.Epoch) []phase0.ValidatorIndex {
+func (c *controller) AllActiveIndices(epoch phase0.Epoch, afterMetadata bool) []phase0.ValidatorIndex {
```
nkryuchkov, line 638: The `afterMetadata` variable is named `afterValidatorSetup`/`afterValidatorsSetup` in interfaces, should they have the same name?

nkryuchkov (approved):

moshe-blox (dismissed):

lior-blox (approved):

PR #1309, opened 2024-02-07
fallback to regular block if MEV fails

MatusKysel:


protocol/v2/ssv/runner/proposer.go
```diff
@@ -161,6 +138,25 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
+		// Attempt to fetch the blinded beacon block.
+		obj, ver, err = r.GetBeaconNode().GetBlindedBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
+		if err == nil {
+			return obj, ver, nil
+		}
+		logger.Debug("failed to get blinded beacon block, falling back to standard block")
+	}
+	// Fetch the standard beacon block if blinded block fetching fails or if blinded is false.
+	obj, ver, err = r.GetBeaconNode().GetBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
+	return obj, ver, err
```
nkryuchkov, line 157: I think we could simplify this a bit
```suggestion
	if r.ProducesBlindedBlocks {
		// Attempt to fetch the blinded beacon block.
		obj, ver, err := r.GetBeaconNode().GetBlindedBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
		if err == nil {
			return obj, ver, nil
		}
		logger.Debug("failed to get blinded beacon block, falling back to standard block")
	}
	// Fetch the standard beacon block if blinded block fetching fails or if blinded is false.
	return r.GetBeaconNode().GetBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
```

protocol/v2/ssv/runner/proposer.go
```diff
@@ -161,6 +138,25 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
+	var obj ssz.Marshaler
+	var ver spec.DataVersion
+	var err error
+	if r.ProducesBlindedBlocks {
+		// Attempt to fetch the blinded beacon block.
+		obj, ver, err = r.GetBeaconNode().GetBlindedBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
+		if err == nil {
+			return obj, ver, nil
+		}
+		logger.Debug("failed to get blinded beacon block, falling back to standard block")
```
nkryuchkov, line 153: I might be wrong but I think we may need to consider using `Warn` because an end user might want to know that MEV is actually not used and would likely have default log level

protocol/v2/ssv/runner/proposer.go
```diff
@@ -105,33 +104,11 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
-	} else {
-		// get block data
-		obj, ver, err = r.GetBeaconNode().GetBeaconBlock(duty.Slot, r.GetShare().Graffiti, fullSig)
-		if err != nil {
-			return errors.Wrap(err, "failed to get beacon block")
-		}
+	// get block data
+	obj, ver, err := r.fetchBeaconBlock(logger, duty.Slot, fullSig)
+	if err != nil {
+		return errors.Wrap(err, "failed to get beacon block")
```
nkryuchkov, line 111: It's part of old code, but I'd like to suggest using `fmt.Errorf` for new code/changes as [github.com/pkg/errors]( https://github.com/pkg/errors) is discontinued

protocol/v2/ssv/runner/proposer.go
```diff
@@ -105,33 +104,11 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
-				return errors.Wrap(err, "failed to get blinded beacon block")
-			}
-		}
-	} else {
-		// get block data
-		obj, ver, err = r.GetBeaconNode().GetBeaconBlock(duty.Slot, r.GetShare().Graffiti, fullSig)
-		if err != nil {
-			return errors.Wrap(err, "failed to get beacon block")
-		}
+	// get block data
```
nkryuchkov, line 108: You made no changes here, but I think while we touch this, we could remove this comment as it's obvious

protocol/v2/ssv/runner/proposer.go
```diff
@@ -161,6 +138,25 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
 
+// Function to attempt fetching a beacon block based on the provided conditions.
+// Returns the block (as ssz.Marshaler), the data version, and any error encountered.
+func (r *ProposerRunner) fetchBeaconBlock(logger *zap.Logger, slot phase0.Slot, fullSig []byte) (ssz.Marshaler, spec.DataVersion, error) {
+	var obj ssz.Marshaler
+	var ver spec.DataVersion
+	var err error
+	if r.ProducesBlindedBlocks {
+		// Attempt to fetch the blinded beacon block.
+		obj, ver, err = r.GetBeaconNode().GetBlindedBeaconBlock(slot, r.GetShare().Graffiti, fullSig)
```
nkryuchkov, line 149: Did you check how long it usually takes when it fails? Could this duration be an issue? (I think not, just want to be sure)

protocol/v2/ssv/runner/proposer.go
```diff
@@ -105,33 +104,10 @@ func (r *ProposerRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *spec
...
-		}
-	} else {
-		// get block data
-		obj, ver, err = r.GetBeaconNode().GetBeaconBlock(duty.Slot, r.GetShare().Graffiti, fullSig)
-		if err != nil {
-			return errors.Wrap(err, "failed to get beacon block")
-		}
+	obj, ver, err := r.fetchBeaconBlock(logger, duty.Slot, fullSig)
+	if err != nil {
+		return fmt.Errorf("failed to get beacon block %w", err)
```
nkryuchkov, line 110:
```suggestion
		return fmt.Errorf("failed to get beacon block: %w", err)
```

nkryuchkov (approved):
Approving, but let's make sure that delay when `GetBlindedBeaconBlock` fails doesn't cause any issues

nkryuchkov (approved):

PR #1313, opened 2024-02-13
Feat: Mark invalid shares

AKorpusenko:


eth/eventhandler/handlers.go
```diff
@@ -263,16 +268,18 @@ func (eh *EventHandler) validatorAddedEventToShare(
...
 	publicKey, err := ssvtypes.DeserializeBLSPublicKey(event.PublicKey)
 	if err != nil {
-		return nil, nil, &MalformedEventError{
+		return validatorShare, nil, &MalformedEventError{
 			Err: fmt.Errorf("failed to deserialize validator public key: %w", err),
 		}
 	}
 	validatorShare.ValidatorPubKey = publicKey.Serialize()
 	validatorShare.OwnerAddress = event.Owner
+	validatorShare.Metadata.InvalidSecret = false
```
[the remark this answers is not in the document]
AKorpusenko: removing. thanks 🙂

eth/eventhandler/handlers.go
```diff
@@ -233,11 +233,12 @@ func (eh *EventHandler) handleShareCreation(
 		sharePublicKeys,
 		encryptedKeys,
 	)
-	if err != nil {
-		return nil, fmt.Errorf("could not extract validator share from event: %w", err)
+
+	if err != nil && !share.Metadata.InvalidSecret {
```
moshe-blox, line 237: is the 2nd check needed? as far as i can see that method returns no error when the secret is invalid

moshe-blox (changes requested):

eth/eventhandler/handlers.go
```diff
@@ -209,7 +209,7 @@ func (eh *EventHandler) handleValidatorAdded(txn basedb.Txn, event *contract.Con
 	}
 
 	isOperatorShare := validatorShare.BelongsToOperator(eh.operatorData.GetOperatorData().ID)
-	if isOperatorShare {
+	if isOperatorShare && !validatorShare.InvalidSecret {
 		eh.metrics.ValidatorInactive(event.PublicKey)
 		ownShare = validatorShare
 		logger = logger.With(zap.Bool("own_validator", isOperatorShare))
```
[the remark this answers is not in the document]
AKorpusenko: Agree. Added this

eth/eventhandler/handlers.go
moshe-blox: oh sry, i thought that we're not returning error in `validatorAddedEventToShare` when only the share is invalid, thats what i previously meant in my comment there

can we try to do this?

i'm interested to see how it would look, we can just log the reason for invalid secret instead of returning it as error (for example `logger.Debug("invalid secret error: could not decrypt share private")`)

eth/eventhandler/handlers.go
```diff
@@ -263,16 +273,17 @@ func (eh *EventHandler) validatorAddedEventToShare(
 	sharePublicKeys [][]byte,
 	encryptedKeys [][]byte,
 ) (*ssvtypes.SSVShare, *bls.SecretKey, error) {
-	validatorShare := ssvtypes.SSVShare{}
+	validatorShare := &ssvtypes.SSVShare{}
 
 	publicKey, err := ssvtypes.DeserializeBLSPublicKey(event.PublicKey)
 	if err != nil {
-		return nil, nil, &MalformedEventError{
+		return validatorShare, nil, &MalformedEventError{
```
y0sher, line 280: why did we do this change? is returning an empty share anything better or different than returning nil?

y0sher (commented):
@moshe-blox lets decide if we still want to have this.

zippoxer:
hey @MatusKysel @y0sher 

yes, and i think we ended up saying my suggestion did not work out well, and this PR would be better in the original form where it returned an error on invalid share (rather than only setting the field and not returning error)

PR #1325, opened 2024-02-20
feat: get blinded and non-blinded blocks in parallel

y0sher:


y0sher:
this is irrelevant since we added propsoal V3 instead #1329

PR #1331, opened 2024-02-28
Add degradation tests into CI 

AKorpusenko:


scripts/degradation-tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+}
+func checkLine(line, section string) int {
+	reChange := regexp.MustCompile(`-?\d+\.\d+%`)
+	reTestName := regexp.MustCompile(`^\S+`)
+	matches := reChange.FindAllString(line, -1)
+	testNameMatch := reTestName.FindString(line)
+
+	if testNameMatch == "geomean" {
```
nkryuchkov, line 115: A comment explaining what this string means would be great

nkryuchkov (dismissed):
Great job! Just minor comments

scripts/degradation-tester/main.go
```diff
@@ -0,0 +1,159 @@
...
+	loadConfig(configFilename)
+
+	file, err := os.Open(resultsFilename)
+	if err != nil {
+		fmt.Printf("Error opening results file: %v\n", err)
+		os.Exit(1)
+	}
+	defer file.Close()
+
+	var currentSection string
```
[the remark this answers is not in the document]
AKorpusenko: I need to track what's the current line, so I need it to be declared before for loop

nkryuchkov (approved):

scripts/degradation-tester/main.go
```diff
@@ -0,0 +1,170 @@
...
+
+	scanner := bufio.NewScanner(file)
```
moshe-blox, line 55: try to run benchstat with CSV output format, instead of parsing lines
this could help reduce regex usage

https://cs.opensource.google/go/x/perf/+/5eefbfdb:cmd/benchstat/main.go;l=455

moshe-blox (changes requested):

message/validation/rsa_benchmark_test.go
y0sher, line 1: This test/benchmark needs to be updated after #1274 and #1358

AKorpusenko: Updated

y0sher:
@AKorpusenko I don't understand how you updated the test but the benchmark results didn't change? especially how this passes on CI since on linux we now use openssl which should be 10x faster than stdlib rsa (which you run locally on your mac).

AKorpusenko:
@y0sher, we have to compare the benchmarks only between different GHA workflow runs, right? I've added a `manual-update-benchmarks` GHA workflow for this, but it only can be used after this code will be in master. Before this, I used my own fork of ssv to get linux runs benchmarks. 
In the last commit, I've just made you can see significant performance increase. So it works like we expect. 

Another thing I've noticed, is that we have `queue_test.go` which uses `b.Log` and this logs appear in the benchmarks, what is wrong from my perspective and should be fixed in another PR.

PR #1335, opened 2024-03-05
Don't consider same duty slashable

nkryuchkov:
The PR makes same duty considered non-slashable to fix a bug when there's a delay in receiving attestation data at the beginning of a new slot, causing broadcasting partial signatures and therefore bumping round in slashing protection only by part of the nodes.

The changes compare the current attestation data or block root with the previous one, and if they are the same, then the slashing protection is not called. If they are different, it means either a new slot, or a slashable duty, so in this case the slashing protection is called.

### Overview of the bug:

- The round timeout calculation is based on slot time and round durations, so the first round starts at the same time when slot starts

- Upon new slot, runner sends a request for attestation data to BN. If it takes long (e.g. 1.5s), then the consensus starts closer to the end of the first round. 

- The delay may lead to a case when some nodes get a commit quorum a few ms before round ends and in this case the node broadcasts a partial signature message. The partial signature generation method updates the highest attestation in the slashing protector. 

- The node that got the commit quorum (e.g. Node 1) expects further messages to be either consensus messages with a higher slot or partial signature messages. 

- Meanwhile, other nodes (e.g. Node 2-4) receive the commit quorum a few ms after round 1 ends so these messages get rejected by message validation with the "past round" error. They broadcast round changes and go to a new round. The messages sent by Node 2-4 get rejected by Node 1 as slashable by value check function (which uses slashing protector). But they are enough to reach a consensus of 3 nodes and broadcast partial signatures, which get accepted by Node 1, so at the end all nodes are able to reconstruct partial signatures and successfully submit the attestation

- If 1 or 3 nodes get a commit quorum in a new round, nodes are able to reach a consensus afterwards, although with an an unnecessary delay of one round. 

- If 2 nodes get a commit in a new round, they aren't able to reach a consensus and the duty is missed.


protocol/v2/ssv/valuechecker/sync_committee_contribution.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	if err := cd.Validate(); err != nil {
+		return fmt.Errorf("invalid value: %w", err)
+	}
+
+	if err := vc.checkDuty(&cd.Duty, types.BNRoleSyncCommitteeContribution); err != nil {
+		return fmt.Errorf("duty invalid: %w", err)
+	}
+
+	//contributions, _ := cd.GetSyncCommitteeContributions()
```
nkryuchkov, line 22: Copied from spec

nkryuchkov: Outdated

olegshmuelov (changes requested):
The current comparison in the wrapper focuses solely on epoch data (Source and Target). Given our strict [slashing protection rules](https://github.com/bloxapp/eth2-key-manager/blob/eca3b4b2f95f4ef286cc0489de9552717f5f0aba/slashing_protection/normal_protection.go#L35), it's crucial to compare the whole struct to ensure we're accurately identifying identical attestations. I suggest we expand the comparison to cover the entire attestation data structure, not just the Source and Target.

Additionally, it's important to note that when a validator is registered/reactivated, we update the attestation data without storing any signing root information. This practice could impact the effectiveness of the new changes, as the absence of signing root data might compromise our ability to accurately assess slashable conditions under the revised logic.

nkryuchkov:
After discussing this with @olegshmuelov, we came to conclusion that this approach might be too dangerous and avoiding touching slashing protection would be preferable.

Therefore, we need to consider making the "past round" check tolerant to partial signature messages from the previous round.

nkryuchkov:
I was wrong about partial signature messages in the previous comment. Actually it's decided messages. PoC PR: https://github.com/bloxapp/ssv/pull/1343

nkryuchkov:
On hold, it will likely be superseded by https://github.com/bloxapp/ssv/pull/1343

PR #1339, opened 2024-03-11
Gm/e2e test rsa change

guym-blox:


y0sher:
please fix title and add proper description.

cli/cli.go
```diff
@@ -30,7 +31,8 @@ func Execute(appName, version string) {
 }
 
 func init() {
-	RootCmd.AddCommand(bootnode.StartBootNodeCmd)
 	RootCmd.AddCommand(operator.StartNodeCmd)
 	RootCmd.AddCommand(operator.GenerateDocCmd)
+	RootCmd.AddCommand(bootnode.StartBootNodeCmd)
+	RootCmd.AddCommand(tests.RsaCmd)
```
y0sher, line 37: should be a completely separate command if possible IMO, no need to add it to the RootCmd, make its own.

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,409 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		logger, err := setupGlobal()
+		ctx := context.Background()
+		if err != nil {
+			log.Fatal("could not create logger", err)
+		}
+
+		metricsReporter := metricsreporter.New(
+			metricsreporter.WithLogger(logger),
+		)
```
y0sher, line 96: why do we need actual metrics reporter? are we using the metrics for anything?

cli/tests/rsa_signature.go
y0sher, line 1: There's a lot of stuff created here that isn't used, the metrics, some validator logic which isn't relevant, whatever we can pass a mock or some empty struct its better. Also there's a lot of copied code, some if it as well isn't relevant to the task at all, at least lets try to not copy it, but now that it works you can start to get rid of stuff and eventually leave only whats necessary.

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,409 @@
...
+		cfg.P2pNetworkConfig.OperatorPubKeyHash = format.OperatorID(operatorData.PublicKey)
+		operatorId := 0
+		cfg.P2pNetworkConfig.OperatorID = func() spectypes.OperatorID {
+			operatorId++
+			return spectypes.OperatorID(operatorId)
+		}
+		cfg.P2pNetworkConfig.FullNode = cfg.SSVOptions.ValidatorOptions.FullNode
+		cfg.P2pNetworkConfig.Network = networkConfig
+
+		validatorsMap := validatorsmap.New(ctx)
```
y0sher, line 139: why do we need validators map?

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,409 @@
...
+
+		dutyStore := dutystore.New()
+		cfg.SSVOptions.DutyStore = dutyStore
+
+		messageValidator := validation.NewMessageValidator(
```
y0sher, line 144: we don't really need to validate messages..

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,409 @@
...
+		cfg.SSVOptions.ValidatorOptions.Network = p2pNetwork
+		cfg.SSVOptions.ValidatorOptions.KeyManager = keyManager
+		cfg.SSVOptions.ValidatorOptions.OperatorData = operatorData
+		cfg.SSVOptions.ValidatorOptions.ValidatorsMap = validatorsMap
+		cfg.SSVOptions.ValidatorOptions.RegistryStorage = nodeStorage
+		cfg.SSVOptions.ValidatorOptions.BeaconNetwork = networkConfig.Beacon.GetNetwork()
+		cfg.SSVOptions.ValidatorOptions.ShareEncryptionKeyProvider = nodeStorage.GetPrivateKey
+
+		validatorCtrl = validator.NewController(logger, cfg.SSVOptions.ValidatorOptions)
+		cfg.SSVOptions.ValidatorController = validatorCtrl
```
y0sher, line 172: why is a full controller needed?

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,409 @@
...
+
+	sk, err := rsaencryption.ConvertPemToPrivateKey(string(operatorKeyByte))
+	if err != nil {
+		return nil, err
+	}
+
+	return sk, err
+}
+
+func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.BadgerDB, error) {
```
y0sher, line 287: do we even need the db?

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,269 @@ package logs_catcher
...
+	if _, err := m.waitForStartCondition(ctx, []string{waitFor}, beaconProxyContainer); err != nil {
+		return err
+	}
 
-var secondTargets = []string{"ssv-node-1", "ssv-node-2", "ssv-node-3", "ssv-node-4"}
+	return m.processMode(ctx)
+}
 
-const waitFor = "End epoch finished"
+func (m *Matcher) processMode(ctx context.Context) error {
```
y0sher, line 48: I don't get this func, all it does is to check that the passed mode exits? IMO should be done on a much higher level than in this impl

y0sher: it would make more sense make to extract each mode to a different function and return it based on the mode here.

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,269 @@ package logs_catcher
...
-const nonSlashableMessage = "\"attester_slashable\":false"
+// testDuty performs a generic validation of attestation logs by comparing entries across beacon proxy and SSV node containers.
+// It's designed to handle both non-slashable and slashable attestation log validations.
+func (m *Matcher) testDuty(ctx context.Context) error {
+	var beaconCriteria, nodeCriteria []string
+	var discrepancyCheck func(beaconCount, nodeCount []any) bool
 
-// Take field
-const idField = "pubkey"
+	switch m.mode {
```
y0sher, line 64: very big func, I would split each to a smaller func , also do the switch here or in processMode above, there's no reason to do it twice

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,269 @@ package logs_catcher
...
-
-	// find slashable attestation not signing for each slashable validator
-	if err := matchMessages(ctx, logger, cli, []string{origMessage, slashableMessage}, []string{slashableMatchMessage}, 0); err != nil {
-		return err
-	}
-
-	// find non-slashable validators successfully submitting (all first round + 1 for second round)
-	if err := matchMessages(ctx, logger, cli, []string{origMessage, nonSlashableMessage}, []string{nonSlashableMatchMessage}, 30); err != nil {
-		return err
+func (m *Matcher) organizeLogsByPubKey(logs []map[string]any, containerName string) (map[string][]any, error) {
```
y0sher, line 260: imo logsByPubkey is simpler name and enough

cli/tests/rsa_signature.go
guym-blox: the process of parsing messages includes decode and p2p requires messageValidator

cli/tests/rsa_signature.go
guym-blox: yes we do need for nodeStorage that been using in p2p

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,380 @@
...
+						// #nosec G404
+						localRand := rand.New(rand.NewSource(seed))
+
+						// Generate a random number, for example between 0 and 9
+						// #nosec G404
+						dynamicValue := localRand.Intn(10) // This will give you a random integer between 0 and 9
+
+						// Using fmt.Sprintf to insert the dynamic value into the string
+						signatureString := fmt.Sprintf("sVV0fsvqQlqliKv/ussGIatxpe8LDWhc8uoaM5WpjbiYvvxUr1eCpz0ja7UT%dPGNDdmoGi6xbMC1g/ozhAt4uCdpy0Xdfqbv", dynamicValue)
+
```
y0sher, line 180: why do we need the random here? I don't see how it helps

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,380 @@
...
+							logger.Fatal("failed to encode message", zap.Error(err))
+						}
+
+						if sendMessageCount < 4 {
+							err = p2pNetwork.Broadcast(ssvMessage)
+							if err != nil {
+								logger.Fatal("failed to broadcast message", zap.Error(err))
+							}
+							sendMessageCount++
+						}
```
y0sher, line 196: call cancel to shutdown if we got enough messages

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,380 @@
...
+						// Converting the formatted string to a byte slice
+						signature := []byte(signatureString)
+						mm.Signature = signature
+						incomingMessage.Data, err = mm.Encode()
+						incomingMessage.Body, err = incomingMessage.Encode()
+						if err != nil {
+							logger.Fatal("failed to encode message", zap.Error(err))
+						}
+
+						if sendMessageCount < 4 {
```
y0sher, line 190: move 4 to a constant

cli/tests/rsa_signature.go
y0sher: yes but you could create a small message validator that doesn't goes through all validation logic. not super critical but makes more sense here.

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,380 @@
...
+						// #nosec G404
+						dynamicValue := localRand.Intn(10) // This will give you a random integer between 0 and 9
+
+						// Using fmt.Sprintf to insert the dynamic value into the string
+						signatureString := fmt.Sprintf("sVV0fsvqQlqliKv/ussGIatxpe8LDWhc8uoaM5WpjbiYvvxUr1eCpz0ja7UT%dPGNDdmoGi6xbMC1g/ozhAt4uCdpy0Xdfqbv", dynamicValue)
+
+						// Converting the formatted string to a byte slice
+						signature := []byte(signatureString)
+						mm.Signature = signature
+						incomingMessage.Data, err = mm.Encode()
```
nkryuchkov, line 184: Error is not checked

cli/tests/rsa_signature.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+	var operatorData *registrystorage.OperatorData
+	operatorData, found, err = nodeStorage.GetOperatorDataByPubKey(nil, operatorPubKey)
+	if err != nil {
+		logger.Fatal("could not get operator data by public key", zap.Error(err))
+	}
+	if !found {
+		operatorData = &registrystorage.OperatorData{
+			PublicKey: operatorPubKey,
+			ID:        4,
```
nkryuchkov, line 237: Why 4?

cli/tests/rsa_signature.go
nkryuchkov: I have the same opinion

e2e/crash-logs/ssv-node-1.txt
nkryuchkov, line 1: Why do we need this?

e2e/config/share5.yaml
```diff
@@ -0,0 +1,5 @@
+# Operator 4
```
nkryuchkov, line 1: Operator 5?

e2e/config/share5.yaml
```diff
@@ -0,0 +1,5 @@
...
+# Public key: LS0tLS1CRUdJTiBSU0EgUFVCTElDIEtFWS0tLS0tCk1JSUJJakFOQmdrcWhraUc5dzBCQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBdi82RytFUEhOK2NycGZ1Z01aYTYKQkpPd3ZycjdTNGkvbWprbTU4czN3RXhkVnFQUjZFbXBUa29QQ0tWamlWSTMvUzIvRjdRdm1FNlA1M3N3cEZoOQpCTDZnaHRpRlZtUzBXQVpOejJJODJRT01HdzNxSEk4NmJSS2RuUjdmU2ZjYlFtSlFuRDNyK0xVRngyQngxZENVCndwSTltSDhXNEhrVHE0VkhGTVY5SlRyaWNhYS9SZ0lqd3RjZ25QUkQ3VTBpeU1zeStyS2RpK2xWcHJqNHFqYmwKcEZ6MExGcXVrYWc4cWdxVnpqRVZsVEZSa3VuaWhvNm96TlZhZU1sSnlXRk4rZ1d5YVp5T0xQMjVhNjl2V0hwbQoxeG1nclZuZW9ncEdnc0hNMGZuTHJJRXlTUkNQdjU5UEVwYTZIdGtlODh1ZjkwRFBDa2RaV0xnTXRzdnkwREJDCnJRSURBUUFCCi0tLS0tRU5EIFJTQSBQVUJMSUMgS0VZLS0tLS0K
```
nkryuchkov, line 2: Why is it same as Operator 4's?

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,208 @@ package logs_catcher
...
+}
+
+func NewLogMatcher(logger *zap.Logger, cli DockerCLI, mode SubMode) *Matcher {
+	return &Matcher{mode: mode, cli: cli, logger: logger}
+}
 
-const waitFor = "End epoch finished"
+// Match starts log catching based on the Matcher mode
+func (m *Matcher) Match(pctx context.Context) error {
+	ctx, cancel := context.WithTimeout(pctx, 24*time.Minute) // 24 minutes for 4 epochs
```
nkryuchkov, line 34: Maybe we'll need to run this with custom eth network, so let's calculate this instead

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,208 @@ package logs_catcher
...
 
-	return nil
+	return processedLogs, nil
 }
 
-func Match(pctx context.Context, logger *zap.Logger, cli DockerCLI) error {
-	startctx, startc := context.WithTimeout(pctx, time.Minute*6*4) // wait max 4 epochs
-	_, err := StartCondition(startctx, logger, []string{waitFor}, waitTarget, cli)
+func (m *Matcher) logsByPublicKey(ctx context.Context, containerName string, matchStrings []string) (map[string][]any, error) {
+	logss, err := m.processDockerLogs(ctx, containerName, matchStrings)
```
nkryuchkov, line 162:
```suggestion
	logs, err := m.processDockerLogs(ctx, containerName, matchStrings)
```

e2e/logs_catcher/matcher.go
```diff
@@ -2,118 +2,208 @@ package logs_catcher
...
+		}
+	}
+	if len(nodeOpids) != 4 {
+		return fmt.Errorf("expected 4 unique opids in node %s, got %d", nodeContainer, len(nodeOpids))
+	}
 	return nil
 }
+
+func validOpid(opid int) bool {
+	return opid >= 1 && opid <= 4
```
nkryuchkov, line 208: I think 4 should be constant

y0sher (dismissed):
Please add title and description and see comments by @nkryuchkov

y0sher (changes requested):
Please add title and description and see comments by @nkryuchkov

PR #1341, opened 2024-03-12
refactor: (VoluntaryExit) fetch header instead of block, to get the slot

moshe-blox:


y0sher (approved):
I think we should merge this but needs conflict fix

y0sher:
We encountered cases where the execution client is not able to answer "BlockByNumber" because some unmarshling issue of the txs inside the block, instead of trying to find the issue with exec clients, we can just get the header instead and skip this issue.

iurii-ssv (approved):

PR #1343, opened 2024-03-19
Allow previous round decided messages

nkryuchkov:
Fixes the case described in https://github.com/bloxapp/ssv/pull/1335 using another approach


olegshmuelov (commented):
@GalRogozinski @MatheusFranco99
please take a look
https://github.com/bloxapp/ssv/pull/1335#issuecomment-2006422431

MatheusFranco99 (approved):
Good work! Maybe I should even add a proof in knowledge-base that a *decided* is OK for `round == State.Round - 1` but not ok for `round < State.Round -1`. Thanks for the insight

y0sher (dismissed):

protocol/v2/qbft/instance/instance.go
```diff
@@ -162,7 +161,14 @@ func (i *Instance) BaseMsgValidation(msg *specqbft.SignedMessage) error {
 		return errors.Wrap(err, "invalid signed message")
 	}
 
-	if msg.Message.Round < i.State.Round {
+	// If a node gets a commit quorum before round change and other nodes get it after,
+	// then the node with quorum before wouldn't be able to receive next round messages from others
```
GalRogozinski, line 165: don't get the comment...
a "node with [commit] quorum before" finishes consensus and doesn't need to receive next round messages.

you meant the other way around, of letting nodes that changed round receive commit messages

protocol/v2/qbft/instance/instance.go
```diff
@@ -162,7 +161,14 @@ func (i *Instance) BaseMsgValidation(msg *specqbft.SignedMessage) error {
...
+	// due to slashing protection, unless we allow decided messages from previous round.
```
GalRogozinski, line 166: slashing protection shouldn't be about not signing equivocations?
maybe you meant not breaking safety of consensus?

even if safety is compromised, and nodes agree on different contradicting data, slashing won't occur because you need a quorum signing on the same data

protocol/v2/qbft/instance/instance.go
```diff
@@ -162,7 +161,14 @@ func (i *Instance) BaseMsgValidation(msg *specqbft.SignedMessage) error {
...
+	if decided := msg.Message.MsgType == specqbft.CommitMsgType && len(msg.Signers) > 1; decided {
```
GalRogozinski, line 167: I think we should use `len(msg.Signers) >= quorumSize`

protocol/v2/qbft/instance/instance.go
```diff
@@ -162,7 +161,14 @@ func (i *Instance) BaseMsgValidation(msg *specqbft.SignedMessage) error {
...
+		if msg.Message.Round < i.State.Round-1 {
+			return errors.New("past round")
+		}
```
GalRogozinski, line 170: I would think that theoretically decided messages should be able to be processed at any point

GalRogozinski (dismissed):

protocol/v2/qbft/instance/instance.go
nkryuchkov: I can make them be processed at any point. @MatheusFranco99 what's your opinion on this?

protocol/v2/qbft/instance/instance.go
nkryuchkov: I agree, done

protocol/v2/qbft/instance/instance.go
nkryuchkov: I modified the comment, does it look ok now?

protocol/v2/qbft/instance/instance.go
nkryuchkov: Removed mentioning slashing protection

protocol/v2/qbft/instance/instance.go
nkryuchkov: done

GalRogozinski: @MatheusFranco99 answered you?
I actually am curious about a second point of view

MatheusFranco99: Sry for the delay
From a security point of view/msg validation:
- it's fine if we are limiting the number of decided messages for a round a peer can send. And, if I'm not mistaken, we are
From a protocol point of view:
- it's possible it the receiving node is extremely early

MatheusFranco99 (approved):

y0sher (approved):
lgtm

y0sher:
@GalRogozinski @MatheusFranco99 @nkryuchkov 
I think we should also make a spec PR for that.
for now I'm approving the spec divergence

PR #1344, opened 2024-03-19
Logs compress / upload CLI commands

AKorpusenko:


cli/operator/cmd_compress_logs/cmd_test.go
```diff
@@ -0,0 +1,266 @@
...
+func deleteFiles(paths ...string) error {
+	for _, path := range paths {
+		if err := os.Remove(path); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+//func cleanPath(path string) (string, error) {
```
nkryuchkov, line 258: Can we delete this?

moshe-blox:
@AKorpusenko This should do the job, but I believe we can make this PR smaller and more efficient.

- Instead of searching for log files with a custom function, we can just use `filepath.Match`
- Instead of copying the logs to a temporary directory (requires 2x disk space) and having custom code to archive it recursively (even though its flat), we can just archive the list of files we got in `filepath.Match` directly with a minimal function which only archives a list of file paths (its possible to use io.Copy directly from file reader to archive writer)

Regarding uploading, it can just be a flag. For example, you can just run just `archive` for compressing only or `archive --upload <s3-bucket-url>` to also upload and maybe `archive --upload <s3-bucket-url> --clean` to upload and then delete (to free disk space) which is what we'd recommend most operators to run.

moshe-blox:
@AKorpusenko @MatusKysel i've been thinking about concatenating the log files into one, it'd make it much more convenient for us to grep.

Here's a pseudocode which includes this feature:
```
zipFile, _ := os.Create(fmt.Sprintf("debug-snapshot--operator-%d--%s.zip", operatorID, time.Now().Format("2006-01-0--15-04-05")))
zipWriter := zip.NewWriter(zipFile)

// Write metdata.
metadataFile, _ := zipWriter.Create(`metadata.json`)
metadata := map[string]interface{}{
    "operator_id": operatorID,
    "time": time.Now(),
    "version": ssvVersion,
    "config": config, // TODO: redact sensitive data
}

// Write concatenated debug log.
debugFile, _ := zipWriter.Create(`debug.log`)
for _, logFilename := filepath.Match(filepath.Join(filepath.Dir(logFilePath), `debug*.log`)) {
    copyFile(debugFile, logFilename)
    fmt.Fprintf(debugFile, "\n") // separate log files in case they dont end with a newline
}

// Write metrics snapshot.
metricsFile := zipWriter.Create(`prometheus-metrics`)
metricsOutput := // TODO get metrics snapshot
metricsOutput.WriteTo(metricsFile)

// Save the zip file.
zipWriter.Close()

// TODO: upload it if requested

// TODO: clean if requested
```

Notice the archive filename, its important for us to establish a

PR #1347, opened 2024-03-24
Fix Message Validation and Validator Registration for Newly Active Validators

olegshmuelov:
This PR introduces fixes to two critical areas affecting newly active validators: message validation and validator registration processes. These changes ensure that validators becoming active are not unfairly penalized due to metadata update delays and are correctly registered for duties in a timely manner.

**Changes**
1. **Message Validation Fix:**
	- **Issue**: Validators that had recently become active were having their messages ignored because their metadata had not yet been updated to reflect their new status.

	- **Resolution**: Updated the message validation logic to allow for validators in the process of becoming active (i.e., their status is pending queued but the activation epoch is current or past) to have their messages processed. This was achieved by adjusting the validation checks to consider the estimated current epoch and the validator's activation epoch.

2. **Validator Registration Fix:**
	- **Issue**: In the process of validator registration, validators whose status was pending in the queue but eligible for activation based on the epoch were not being registered to relays.

	- **Resolution**: Modified the HandleDuties function within the ValidatorRegistrationHandler to ensure that validators pending activation but within the eligible epoch window are now correctly registered. This adjustment ensures that all validators due for activation are considered for registration, aligning their duties with their activation status.

### DIFF
--- a/protocol/v2/types/ssvshare.go
+++ b/protocol/v2/types/ssvshare.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"sort"
 
+	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	spectypes "github.com/bloxapp/ssv-spec/types"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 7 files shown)


protocol/v2/types/ssvshare.go
```diff
@@ -61,6 +63,11 @@ func (s *SSVShare) HasBeaconMetadata() bool {
 	return s != nil && s.BeaconMetadata != nil
 }
 
+func (s *SSVShare) IsActive(epoch phase0.Epoch) bool {
```
moshe-blox, line 66: we should probably name this `IsAttesting` to match the existing terminology

protocol/v2/types/ssvshare_test.go
```diff
@@ -105,3 +106,82 @@ func TestValidCommitteeSize(t *testing.T) {
...
+						ActivationEpoch: currentEpoch - 1,
+					},
+				},
+			},
+			Epoch:    currentEpoch,
+			Expected: true,
+		},
+	}
+
+	for _, tc := range tt {
```
nkryuchkov, line 181: Since we are not using Go 1.22 yet, let's add this line

```suggestion
	for _, tc := range tt {
		tc := tc
```

nkryuchkov (approved):
LGTM

protocol/v2/types/ssvshare_test.go
moshe-blox: good catch @nkryuchkov

y0sher (approved):
lgtm

nkryuchkov (approved):

moshe-blox:
https://github.com/bloxapp/ssv/blob/5b1e5f23419d714e66251085d177ba7e6bbb875e/operator/validator/controller.go#L318-L331

we should use the new func here too right?

nkryuchkov (approved):

nkryuchkov (approved):

moshe-blox (approved):

PR #1353, opened 2024-03-28
optimization: use miscrosoft/openssl for RSA

y0sher:
### Description 
This uses  is based on #1274 and uses parts from #1348.

### Changes 

- When available and running unix/windows use the https://github.com/golang-fips/openssl/ which is orders of magnitude faster than golang std lib.
- docker to run from ubuntu golang to have openssl by default


.k8/hetzner-stage/ssv-exporter-holesky.yml
```diff
@@ -88,7 +88,7 @@ spec:
               hostPort: 16073
           env:
             - name: SHARE_CONFIG
-              value: "./data/share.yaml"
+              value: "/data/share.yaml"
```
nkryuchkov, line 91: Why do we need to change this in this PR?

operator/keys/keys.go
```diff
@@ -71,7 +69,8 @@ func GeneratePrivateKey() (OperatorPrivateKey, error) {
 }
 
 type privateKey struct {
-	privKey *rsa.PrivateKey
+	privKey       *rsa.PrivateKey
+	cachedPrivKey any
```
nkryuchkov, line 73: What if we create another `privateKey` in `keys_linux.go` which contains `*openssl.PrivateKeyRSA` and use the new library there without using `any` type? This would also simplify the code for signing.

operator/keys/rsa_darwin.go
nkryuchkov, line 1: `rsa_darwin.go` and `rsa_windows.go` look same. Instead, we can create just one file `rsa.go` containing this build constraint:
```go
//go:build !linux
```

operator/keys/rsa_darwin.go
```diff
@@ -0,0 +1,22 @@
+// go:build darwin
```
nkryuchkov, line 1: `go:build` should follow `//` without the whitespace

operator/keys/rsa_linux_test.go
```diff
@@ -0,0 +1,40 @@
+// go:build linux
```
nkryuchkov, line 1: ditto

protocol/v2/types/signature_benchmark_linux_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/microsoft/go-crypto-openssl/openssl"
+	openssl "github.com/microsoft/go-crypto-openssl/openssl"
```
nkryuchkov, line 11: looks redundant

operator/keys/rsa_linux_test.go
```diff
@@ -0,0 +1,40 @@
...
+	sig, err := rsa.SignPKCS1v15(rand.Reader, key, crypto.SHA256, hashed[:])
+	require.NoError(t, err)
+
+	pk := &privateKey{key, nil}
+	pub := pk.Public().(*publicKey)
+
+	require.NoError(t, VerifyRSA(pub, msg, sig))
+}
+
+func Test_VerifyOpenSSLWithOpenSSL(t *testing.T) {
```
nkryuchkov, line 29: It fails in the CI

PR #1354, opened 2024-03-28
fix: timings in attestation submission log

moshe-blox:


y0sher (approved):

protocol/v2/ssv/runner/attester.go
```diff
@@ -181,7 +183,8 @@ func (r *AttesterRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *spe
 
 		logger.Info("✅ successfully submitted attestation",
 			zap.String("block_root", hex.EncodeToString(signedAtt.Data.BeaconBlockRoot[:])),
-			fields.ConsensusTime(time.Since(r.started)),
+			fields.ConsensusTime(consensusDuration),
+			zap.Float64("submission_time", time.Since(start).Seconds()),
```
nkryuchkov, line 187: This is likely not the only place where we want to log submission time, so shall we create a function for this in the `fields` package?

nkryuchkov (approved):

protocol/v2/ssv/runner/attester.go
moshe-blox: @nkryuchkov done

nkryuchkov (approved):

PR #1355, opened 2024-03-28
 optimization: use miscrosoft/openssl for RSA 

y0sher:
### Description 
This uses  is based on #1274 and uses parts from #1348.

### Changes 

- When available and running unix/windows use the https://github.com/golang-fips/openssl/ which is orders of magnitude faster than golang std lib.
- docker to run from ubuntu golang to have openssl by default


operator/keys/rsa_linux.go
```diff
@@ -0,0 +1,103 @@
...
+	"crypto/rsa"
+	"crypto/sha256"
+	"math/big"
+
+	"github.com/microsoft/go-crypto-openssl/openssl"
+	"github.com/microsoft/go-crypto-openssl/openssl/bbig/bridge"
+)
+
+type privateKey struct {
+	privKey       *rsa.PrivateKey
```
nkryuchkov, line 16: Do we need `privKey` in the linux implementation? For linux, we could create a new `*openssl.PrivateKeyRSA` when we create a new `privateKey`

PR #1356, opened 2024-03-29
OperatorData store

nkryuchkov:
The PR creates an OperatorData store solely responsible for actions with the node's own operator ID. 

Its purpose is to fix nil pointer dereference that sometimes occurs after https://github.com/bloxapp/ssv/pull/1240 was merged


beacon/goclient/goclient.go
```diff
@@ -218,8 +217,9 @@ func New(
 		zap.String("version", client.nodeVersion),
 	)
 
-	// Start registration submitter.
-	go client.registrationSubmitter(slotTickerProvider)
+	if operatorDataStore != nil {
+		go client.registrationSubmitter(slotTickerProvider)
+	}
```
y0sher, line 222: why would it be nil?  and if it will be nil at the point of creation but then it will be created then `registrationSubmitter` won't ever start..

nkryuchkov: At the moment in `mockClient` in tests, but maybe it will be changed

nkryuchkov: The operator ID used to be 0 in `mockClient`, and now the client waits until operator ID is not 0, so it's a workaround for tests

nkryuchkov: Although I think it can be set to any non-zero value in tests instead

operator/datastore/data_store.go
```diff
@@ -0,0 +1,94 @@
...
+// SetOperatorData sets the operator data in a thread-safe manner and marks the operator ID as ready if valid.
+func (ods *operatorDataStore) SetOperatorData(od *registrystorage.OperatorData) {
+	ods.operatorDataMu.Lock()
+	defer ods.operatorDataMu.Unlock()
+
+	ods.operatorData = od
+	if od != nil && od.ID != 0 {
+		ods.setOperatorIDReady()
+	}
+}
```
y0sher, line 88: Initially I thought there will be a deadlock between these two functions if `AwaitOperatorID` was called before `SetOperatorData`, as the mutex will be taken but it won't be release since `SetOperatorData` also needs to take the mutex. Apparently in go `sync.Cond.Wait()` suspends the go-routines' execution and unlocks all the mutexs in it until its being released.

y0sher (dismissed):
Generally I think this makes much more sense than storing operator data inside validator controller and passing it around. It's about time we did this change. gj

y0sher (approved):

moshe-blox (approved):

PR #1360, opened 2024-03-31
fix: ignore sc/proposer msgs before fetching metadata and duties

y0sher:


message/validation/errors.go
```diff
@@ -95,6 +95,7 @@ var (
 	ErrInvalidJustifications               = Error{text: "invalid justifications", reject: true}
 	ErrTooManyDutiesPerEpoch               = Error{text: "too many duties per epoch", reject: true}
 	ErrNoDuty                              = Error{text: "no duty for this epoch", reject: true}
+	ErrDutyStoreNotReady                   = Error{text: "duty store is not ready", reject: false}
```
nkryuchkov, line 98: Let's maybe move this upper where all non-rejected errors are located and keep them sorted by whether they are rejected? It would be easier to navigate among them. Also, we can remove `reject: false` to keep the style consistent

PR #1366, opened 2024-04-04
add logic & test for first validator registration

guym-blox:


moshe-blox:
since we no longer have the "first registration" functionality anymore, then this can be closed

PR #1367, opened 2024-04-07
experiment: cache attestation data to prevent wrong block routes.

y0sher:
### Problem

We see on stage cases where operators submit an attestation but it ends up as missed because (we assume) the block root is different than the slot's proposed block root. 

### Solution

This PR aims to solve this and reduce calls to beacon node by caching AttestationData for Slot-CommitteeIndex, this way we always use same root, even for aggregator and also don't need to call it again.

### TODOs
- [ ] - Unit tests
- [ ] - Audit the concurrency solution (maybe there's and existing lib that does this better?)
- [ ] - Deploy half stage with this and half just stage branch and create a report to compare the amount of misses because of wrong block root.


y0sher:
closed and mentioned : https://github.com/ssvlabs/ssv/issues/1472

PR #1370, opened 2024-04-11
optimization: inc block propgagation

y0sher:
### Description

We found out that sometimes a node receives a head event about a block that arrived early and starts consensus, then other nodes dropping the message because it takes them longer to get the event. 
For this we added the block propagation delay, but we see this delay isn't enough, hence adding 100 more ms.


moshe-blox (approved):

y0sher:
closed and noted in - https://github.com/ssvlabs/ssv/issues/1471

PR #1373, opened 2024-04-17
feat: disable ip rate limit

y0sher:
### Changes

This PR adds a config option to disable IP rate limiting and IP-Colocation scoring. this is used mainly in local and testing environments where we want to enable entities to use the same IP without being banned. (In a cloud environment or a single machine running multiple nodes for example).

### How to use

As added in the stage yamls, you can just set `DISABLE_IP_RATE_LIMIT` env var to `true`, of course this is `false` by default.

### DIFF
--- a/network/topics/params/peer_score.go
+++ b/network/topics/params/peer_score.go
@@ -45,7 +45,7 @@ func PeerScoreThresholds() *pubsub.PeerScoreThresholds {
 }
 
 // PeerScoreParams returns peer score params according to the given options
-func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, disableCoLocation bool, ipWhilelist ...*net.IPNet) *pubsub.PeerScoreParams {
+func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, disableColocation bool, ipWhilelist ...*net.IPNet) *pubsub.PeerScoreParams {
 	if oneEpoch == 0 {
 		oneEpoch = oneEpochDuration
 	}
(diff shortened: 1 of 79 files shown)


nkryuchkov (dismissed):
LGTM.

Do we need to modify the exporter config as well?

network/topics/params/peer_score.go
```diff
@@ -45,7 +45,7 @@ func PeerScoreThresholds() *pubsub.PeerScoreThresholds {
 }
 
 // PeerScoreParams returns peer score params according to the given options
-func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, ipWhilelist ...*net.IPNet) *pubsub.PeerScoreParams {
+func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, disableCoLocation bool, ipWhilelist ...*net.IPNet) *pubsub.PeerScoreParams {
```
nkryuchkov, line 48: Let's keep the same naming style as in `IPColocationFactorWeight`

```suggestion
func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, disableColocation bool, ipWhilelist ...*net.IPNet) *pubsub.PeerScoreParams {
```

network/topics/params/peer_score.go
```diff
@@ -57,6 +57,11 @@ func PeerScoreParams(oneEpoch, msgIDCacheTTL time.Duration, ipWhilelist ...*net.
 	targetVal = targetVal - behaviourPenaltyThreshold
 	behaviourPenaltyWeight := gossipThreshold / (targetVal * targetVal)
 
+	localIPColocationFactorWeight := ipColocationFactorWeight
```
nkryuchkov, line 60: As I understand, it's assumed to be local when `disableCoLocation` is true? If so, the name might be confusing as it's not always local

nkryuchkov (approved):

PR #1375, opened 2024-04-18
fix: find host address using dig

y0sher:
### Fix

In our Makefile we are setting HOST_ADDRESS env to pass to the node in order to advertise itself in discovery. when we introduced #1358 , we switched to using a different image in docker and this image lacks `dig` which is the utility we use to resolve our address against a dns service.

this installs dnsutils which has dig so we could get our address.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get update                                                        && \
   gcc-aarch64-linux-gnu=4:12.2.0-3 \
   bzip2=1.0.8-5+b1 \
   make=4.3-4.1 \
-  dnsutils \
+  dnsutils=1:9.18.24-1 \
   && rm -rf /var/lib/apt/lists/*
 
 RUN go version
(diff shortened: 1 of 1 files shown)


nkryuchkov (dismissed):

zktaiga (changes requested):
Can you please pin the version of dnsutils? Looks good otherwise.

Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apt-get update                                                        && \
   gcc-aarch64-linux-gnu=4:12.2.0-3 \
   bzip2=1.0.8-5+b1 \
   make=4.3-4.1 \
+  dnsutils \
```
zktaiga, line 16:
```suggestion
  dnsutils=1:9.18.24-1 \
```

zktaiga (approved):

moshe-blox (approved):

PR #1380, opened 2024-05-01
remove bloxstaking images and links references

guym-blox:


nkryuchkov (commented):
Not sure if it should be a part of this PR but should we change [this line](https://github.com/ssvlabs/ssv/blob/gm/remove_bloxstaking_references/network/discovery/subnets.go#L12)?

moshe-blox (approved):

PR #1387, opened 2024-05-06
Gm/ssv 137 add pre and post consensus logs

guym-blox:
This pull request proposes the addition of detailed logging to the ProcessPreConsensus,ProcessConsensus and ProcessPostConsensus functions. we aim to enhance visibility into the time consumption of various components during consensus and post-consensus processes.


protocol/v2/ssv/runner/attester.go
```diff
@@ -166,10 +160,15 @@ func (r *AttesterRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *spe
 		}
 		specSig := phase0.BLSSignature{}
 		copy(specSig[:], sig)
-
+		r.metrics.EndPostConsensus()
 		logger.Debug("🧩 reconstructed partial signatures",
-			zap.Uint64s("signers", getPostConsensusSigners(r.GetState(), root)))
+			zap.Uint64s("signers", getPostConsensusSigners(r.GetState(), root)),
+			fields.QuorumTime(r.metrics.GetPostConsensusTime()))
```
y0sher, line 166: don't you think its confusing to use QuorumTime with same key for pre, consensus and post? just for sake of easy search or comparing results, or making averages, it makes sense to add some prefix like pre_/cons_/post_ to it.

protocol/v2/ssv/runner/attester.go
```diff
@@ -166,10 +160,15 @@ func (r *AttesterRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *spe
...
 
+		attestationSubmissionEnd := r.metrics.StartBeaconSubmission()
```
y0sher, line 168: this confused me at first, I think the var name should be or include "EndFunc"

y0sher (dismissed):
@guym-blox and @moshe-blox 
I think we should go ahead and soon merge this (with appropriate changes) into `alan/no-fork` as well.

Approved but please consider my comments.

protocol/v2/ssv/runner/attester.go
moshe-blox: renamed

protocol/v2/ssv/runner/attester.go
moshe-blox: done

protocol/v2/ssv/runner/metrics/metrics.go
```diff
@@ -229,3 +262,21 @@ func (cm *ConsensusMetrics) RoleSubmissionFailed() {
...
+
+// EndBeaconData sends metrics for data duration.
+func (cm *ConsensusMetrics) EndBeaconData() {
+	if cm != nil && cm.beaconData != nil && !cm.beaconDataStart.IsZero() {
+		duration := time.Since(cm.beaconDataStart)
+		cm.beaconDataDuration = duration
+		log.Printf("setting beaconDataDuration to %v (started: %s)", duration, cm.beaconDataStart)
+		cm.beaconData.Observe(duration.Seconds())
+		cm.beaconDataStart = time.Time{}
+	}
```
y0sher, line 281: imo its better to do `if cm == nil { return }`

y0sher (approved):

protocol/v2/ssv/runner/metrics/metrics.go
moshe-blox: > imo its better to do `if cm == nil { return }`

i think this condition should not even exist 😆 

instead of checking for nil, we can just use a `noopConsensusMetrics` implementation 

however, this is how the other methods are implemented so we can leave this for metrics refactor

moshe-blox (approved):

PR #1393, opened 2024-05-13
Fix message validation leftovers

nkryuchkov:


MatheusFranco99 (commented):
Great work!

message/validation/partial_validation.go
```diff
@@ -118,54 +127,69 @@ func (mv *messageValidator) validatePartialSignatureMessageSemantics(
...
+		}
 	}
 
-	signer := signedSSVMessage.GetOperatorIDs()[0]
-	signerState := state.GetSignerState(signer)
-	if signerState == nil {
-		signerState = state.CreateSignerState(signer)
+	if err := mv.validateBeaconDuty(signedSSVMessage.SSVMessage.GetID().GetRoleType(), messageSlot, committeeData.indices); err != nil {
+		return err
 	}
```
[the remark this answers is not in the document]
MatheusFranco99: My older comment (which should be referenced here) disappeared. But I meant to add:
```go
// Rule: For the proposal and sync committee contribution duties, we check if the validator is assigned to it
```
In both `mv.validateBeaconDuty` checks

message/validation/consensus_validation.go
```diff
@@ -80,13 +80,17 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
...
-		if consensusMessage.MsgType != specqbft.CommitMsgType {
+		// Rule: Decided msg with different type than Commit
+		if msgType != specqbft.CommitMsgType {
 			e := ErrNonDecidedWithMultipleSigners
 			e.got = len(signers)
 			return e
 		}
 
+		// Rule: Number of signers must be more than quorum size
 		if uint64(len(signers)) < quorumSize {
```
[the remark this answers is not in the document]
nkryuchkov: @MatheusFranco99 Yes, I left a comment in the KB about this: https://github.com/ssvlabs/knowledge-base-internal/pull/20/files#r1608582666

It used to be here, but I wrote tests for all rules and I couldn't trigger this rule with a test because those two checks already cover it, so I had to remove it. Can we remove it from the KB?

message/validation/consensus_validation.go
```diff
@@ -156,32 +165,62 @@ func (mv *messageValidator) validateQBFTLogic(
...
+					return ErrDuplicatedProposalWithDifferentData
+				}
 			}
 
+			// Rule: Peer must send only 1 proposal, 1 prepare, 1 commit, maxDecidedCount(committeeSize) decided and 1 round-change per round
 			limits := maxMessageCounts(len(committee))
 			if err := signerState.MessageCounts.ValidateConsensusMessage(signedSSVMessage, consensusMessage, limits); err != nil {
 				return err
 			}
 		}
```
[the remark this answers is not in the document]
nkryuchkov: I agree. Done

message/validation/const.go
```diff
@@ -11,11 +11,12 @@ const (
...
-	maxConsensusMsgSize                      = 8388608 // TODO: calculate new value
-	maxPartialSignatureMsgSize               = 1952
-	maxPayloadSize                           = maxConsensusMsgSize
-	maxSignedMsgSize                         = 4 + 56 + maxPayloadSize                // Max possible MsgType + MsgID + Data
-	maxEncodedMsgSize                        = maxSignedMsgSize + maxSignedMsgSize/10 // 10% for encoding overhead
+	// TODO: make sure values are correct
+	maxConsensusMsgSize                      = 6291829 // TODO: 8+8+8+56+32+8+13*65536+13*65536 instead?
+	maxPartialSignatureMsgSize               = 8 + 8 + 1000*(96+32+8+8)
+	maxPayloadSize                           = maxConsensusMsgSize                               // max(maxConsensusMsgSize, maxPartialSignatureMsgSize)
+	maxSignedMsgSize                         = 13*256 + 13*8 + 8 + 56 + maxPayloadSize + 5243144 // Max possible MsgType + MsgID + Data
```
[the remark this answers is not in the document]
nkryuchkov: Yes, even I don't understand what I meant here. Fixed.

message/validation/consensus_validation.go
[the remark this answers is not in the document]
MatheusFranco99: Yes

MatheusFranco99 (approved):

message/validation/common_checks.go
```diff
@@ -13,70 +14,108 @@ func (mv *messageValidator) committeeRole(role spectypes.RunnerRole) bool {
...
-		err.got = fmt.Sprintf("%v (role %v)", state.EpochDuties, msgID.GetRoleType())
-		err.want = fmt.Sprintf("less than %v", dutyLimit)
-		return err
+	// Rule: For a sync committee aggregation duty message, we check if the validator is assigned to it
+	if role == spectypes.RoleSyncCommitteeContribution {
+		period := mv.netCfg.Beacon.EstimatedSyncCommitteePeriodAtEpoch(mv.netCfg.Beacon.EstimatedEpochAtSlot(slot))
+		if mv.dutyStore != nil && mv.dutyStore.SyncCommittee.Duty(period, indices[0]) == nil {
+			return ErrNoDuty
+		}
 	}
```
y0sher, line 120: if `mv.dutyStore` doesn't exist this function will return nil which IMO isn't correct, we should return a separate error for example dutyStore doesn't exist and on the upper level decide if to ignore and just go on or do something else.

message/validation/consensus_state.go
```diff
@@ -14,26 +17,22 @@ type consensusID struct {
...
-	mu      sync.Mutex
+	state    map[spectypes.OperatorID]*treemap.Map // TODO: use *treemap.Map[phase0.Slot, *SignerState] after updating to Go 1.21
+	maxSlots phase0.Slot
+	mu       sync.Mutex
 }
 
-// GetSignerState retrieves the state for the given signer.
-// Returns nil if the signer is not found.
-func (cs *consensusState) GetSignerState(signer spectypes.OperatorID) *SignerState {
+func (cs *consensusState) Get(signer spectypes.OperatorID) *treemap.Map {
```
y0sher, line 25: function name should represent that it creates the item if it doesn't exist, in the name or at least comment.

message/validation/common_checks.go
nkryuchkov: @y0sher We never pass nil duty store to the message validator constructor, this is actually a leftover from the previous message validation when the duty store was optional. So I think we can just remove the check and panic if it's nil

message/validation/consensus_state.go
nkryuchkov: I agree, thanks

message/validation/common_checks.go
nkryuchkov: @y0sher done

message/validation/consensus_validation.go
```diff
@@ -96,8 +100,8 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
 	}
 
 	if len(signedSSVMessage.FullData) != 0 {
-		if consensusMessage.MsgType == specqbft.PrepareMsgType ||
-			consensusMessage.MsgType == specqbft.CommitMsgType && len(signedSSVMessage.GetOperatorIDs()) == 1 {
+		// Rule: Prepare or commit messages must not have full data
+		if msgType == specqbft.PrepareMsgType || msgType == specqbft.CommitMsgType && len(signers) == 1 {
```
y0sher, line 104: I rather write conditions like this with parentheses as it makes things clearer
`if msgType == specqbft.PrepareMsgType || (msgType == specqbft.CommitMsgType && len(signers) == 1) {`

nkryuchkov: @y0sher I don't mind, done

message/validation/consensus_validation.go
```diff
@@ -228,50 +287,65 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
...
+
+	signerCount := len(signedSSVMessage.GetOperatorIDs())
+	if signerCount > 1 {
+		encodedOperators, err := encodeOperators(signedSSVMessage.GetOperatorIDs())
+		if err != nil {
+			return
 		}
 
-		signerState.MessageCounts.RecordConsensusMessage(signedSSVMessage, consensusMessage)
+		signerState.SeenSigners[string(encodedOperators)] = struct{}{}
```
y0sher, line 328: this is a security concern IMO @GalRogozinski @MatheusFranco99 @nkryuchkov @moshe-blox 
if somehow a malicious signer was able to create a operatorID that is unable to be encoded, it might escape from getting into the `SeenSigners` list and send endless messages.
I might be missing an earlier check that prevents this though.

message/validation/consensus_validation.go
```diff
@@ -396,3 +450,13 @@ func (mv *messageValidator) roundRobinProposer(height specqbft.Height, round spe
...
 }
+
+func encodeOperators(operators []spectypes.OperatorID) ([]byte, error) {
+	buf := new(bytes.Buffer)
+	for _, operator := range operators {
+		if err := binary.Write(buf, binary.LittleEndian, operator); err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), nil
```
y0sher, line 461: why not do something like a hash?

message/validation/const.go
```diff
@@ -11,16 +11,27 @@ const (
...
 	lateSlotAllowance                        = 2
 	rsaSignatureSize                         = 256
-	blsSignatureSize                         = 96
 	syncCommitteeSize                        = 512
 	maxSignaturesInSyncCommitteeContribution = 13
 )
+
+var (
+	maxPayloadSize    = max(maxConsensusMsgSize, maxPartialSignatureMsgSize) // not const because of max TODO: const after Go 1.21
+	maxSignedMsgSize  = 13*256 + 13*8 + 8 + 56 + maxPayloadSize + 5243144
```
y0sher, line 26: what is 5243144?

y0sher: add comment at least or const
other numbers here can also be consts

message/validation/errors.go
```diff
@@ -52,31 +52,36 @@ func (e Error) Text() string {
...
+	ErrSlotAlreadyAdvanced                     = Error{text: "signer has already advanced to a later slot"}
+	ErrRoundAlreadyAdvanced                    = Error{text: "signer has already advanced to a later round"}
+	ErrDecidedWithSameSigners                  = Error{text: "decided with same number of signers"}
+	ErrPubSubDataTooBig                        = Error{text: "pub-sub message data too big"}
+	ErrIncorrectTopic                          = Error{text: "incorrect topic"}
+	ErrNonExistentCommitteeID                  = Error{text: "committee ID doesn't exist"}
+	ErrRoundTooHigh                            = Error{text: "round is too high for this role"}
+	ErrValidatorIndexMismatch                  = Error{text: "partial signature validator index not found"}
+	ErrTooManyDutiesPerEpoch                   = Error{text: "too many duties per epoch"}
+	ErrNoDuty                                  = Error{text: "no duty for this epoch"}
```
y0sher, line 71: why is this not reject anymore? @moshe-blox and I was working on fetching all duties ASAP so we'll not reject tons of messages on startup or new validators.

message/validation/errors.go
```diff
@@ -52,31 +52,36 @@ func (e Error) Text() string {
...
-	ErrEarlyMessage                            = Error{text: "early message"}
-	ErrLateMessage                             = Error{text: "late message"}
-	ErrTooManySameTypeMessagesPerRound         = Error{text: "too many messages of same type per round"}
-	ErrRoundTooHigh                            = Error{text: "round is too high for this role" /*, reject: true*/} // TODO: enable reject
+	ErrEarlySlotMessage                        = Error{text: "message was sent before slot starts"}
+	ErrLateSlotMessage                         = Error{text: "current time is above duty's start +34(committee and aggregator) or +3(else) slots"}
+	ErrSlotAlreadyAdvanced                     = Error{text: "signer has already advanced to a later slot"}
+	ErrRoundAlreadyAdvanced                    = Error{text: "signer has already advanced to a later round"}
+	ErrDecidedWithSameSigners                  = Error{text: "decided with same number of signers"}
+	ErrPubSubDataTooBig                        = Error{text: "pub-sub message data too big"}
```
y0sher, line 65: why is this not reject? how could someone send a too big message and it'll be ok?
I can think of a fork scenario where message size decreases and we don't want to reject nodes that haven't updated yet. but this is only a short timespan.

@nkryuchkov @MatheusFranco99

message/validation/message_counts.go
```diff
@@ -39,34 +37,28 @@ func (c *MessageCounts) ValidateConsensusMessage(signedSSVMessage *spectypes.Sig
...
-			err := ErrTooManySameTypeMessagesPerRound
+			err := ErrDuplicatedMessage
 			err.got = fmt.Sprintf("prepare, having %v", c.String())
 			return err
 		}
 	case specqbft.CommitMsgType:
 		if len(signedSSVMessage.GetOperatorIDs()) == 1 {
 			if c.Commit >= limits.Commit {
-				err := ErrTooManySameTypeMessagesPerRound
+				err := ErrDuplicatedMessage
```
y0sher, line 53: duplicate is more correct because there should be only one message per signer per consensus per round?

message/validation/consensus_validation.go
nkryuchkov: `(*bytes.Buffer).Write` never returns an error, but I think it might be better to reject a message in this case for easier understanding. Also, theoretically, `bytes.Buffer` implementation may change in the future or we may decide to change the encoding and forget about the error handling

message/validation/consensus_validation.go
nkryuchkov: The most common case is `8*4=32` bytes, the max possible size with 13 operators is `8*13=104` bytes. Sha256 is 32 bytes, so the difference won't be big, but I don't mind hashing this

message/validation/consensus_validation.go
nkryuchkov: done rejecting

message/validation/consensus_validation.go
nkryuchkov: done

message/validation/const.go
nkryuchkov: it's from spec, done

message/validation/errors.go
nkryuchkov: Aligned to the knowledge base: https://github.com/ssvlabs/knowledge-base-internal/blob/main/ssv/network/p2p/MessageValidation/Rules.md

cc @GalRogozinski

message/validation/errors.go
nkryuchkov: Aligned to the knowledge base: https://github.com/ssvlabs/knowledge-base-internal/blob/main/ssv/network/p2p/MessageValidation/Rules.md

cc @GalRogozinski

message/validation/message_counts.go
nkryuchkov: We don't check decided message count here anymore because other checks override it, so if we'd keep the check here, we wouldn't be able to cover it with tests. Therefore this method checks only duplicates now. Also, it's `ErrDuplicatedMessage` in the knowledge base and we want to align code and KB as much as possible

message/validation/signer_state.go
```diff
@@ -4,34 +4,20 @@ package validation
...
-	s.ProposalData = nil
-	if newEpoch {
-		s.EpochDuties = 1
-	} else {
-		s.EpochDuties++
-	}
-	s.SeenDecidedLengths = make(map[int]struct{})
+func (s *SignerState) Init() {
+	s.Round = specqbft.FirstRound
+	s.SeenSigners = make(map[string]struct{})
```
y0sher, line 20: Shouldn't this also restart `MessageCounts` and `ProposalData`?

message/validation/validation.go
```diff
@@ -180,15 +200,23 @@ func (mv *messageValidator) obtainValidationLock(messageID spectypes.MessageID)
 	return mutex
 }
```
y0sher, line 201: are we planning to really lock by msgid eventually? here there's just a global block, at least lets add a todo or some description.

message/validation/validation.go
```diff
@@ -199,47 +227,56 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
...
-		return nil, nil, ErrValidatorLiquidated
+		return CommitteeData{}, ErrValidatorLiquidated
 	}
 
 	if validator.BeaconMetadata == nil {
-		return nil, nil, ErrNoShareMetadata
+		return CommitteeData{}, ErrNoShareMetadata
 	}
 
+	// Rule: If validator is not active
```
y0sher, line 263: does this cover also checking that he's not slashed?

network/p2p/test_utils.go
y0sher, line 1: comments in this file should be removed?

operator/validator/controller.go
```diff
@@ -1181,13 +1187,27 @@ func SetupCommitteeRunners(ctx context.Context, logger *zap.Logger, options vali
...
 		return crunner.(*runner.CommitteeRunner)
 	}
 }
 
 // SetupRunners initializes duty runners for the given validator
-func SetupRunners(ctx context.Context, logger *zap.Logger, options validator.Options) runner.ValidatorDutyRunners {
+func SetupRunners(
+	ctx context.Context,
+	logger *zap.Logger,
+	options validator.Options,
```
y0sher, line 1209: lets be consistent with `SetupRunners` and `SetupCommitteeRunners` - either pass networkConfig to both functions or in `validator.Options` I guess if `Options` has it we should we it or remove ( if not needed internally)

protocol/v2/ssv/runner/committee.go
```diff
@@ -45,7 +46,9 @@ type CommitteeRunner struct {
 	postStarted   time.Time
 }
 
-func NewCommitteeRunner(beaconNetwork types.BeaconNetwork,
+func NewCommitteeRunner(
+	networkConfig networkconfig.NetworkConfig,
+	beaconNetwork types.BeaconNetwork,
```
y0sher, line 51: I guess `BeaconNetwork` isn't needed anymore now that `NetworkConfig` is passed, since `NetworkConfig` has it. 
either pass in exactly what `CommitteeRunner` uses out of `NetworkConfig` or just pass in `NetworkConfig` as whole and use it.

message/validation/errors.go
y0sher: @moshe-blox @GalRogozinski lets talk about it, resolving for now.

y0sher (changes requested):
Final comments, @nkryuchkov good job on this big PR! looks very good.
Lets do a final check where you merge this with `alan/no-fork` and test on stage please 🙏 

After final comments addressed and this test I'll merge.

message/validation/partial_validation.go
```diff
@@ -23,10 +23,10 @@ func (mv *messageValidator) validatePartialSignatureMessage(
 ) {
 	ssvMessage := signedSSVMessage.SSVMessage
 
-	if len(ssvMessage.Data) > maxPartialSignatureMsgSize {
+	if len(ssvMessage.Data) > genesisValidation.MaxPartialSignatureMsgSize {
 		e := ErrSSVDataTooBig
 		e.got = len(ssvMessage.Data)
-		e.want = maxPartialSignatureMsgSize
```
y0sher, line 31: we should not mix genesis and non-genesis code. this file should use the post-fork constants.

message/validation/signer_state.go
moshe-blox: @y0sher looks like this is not used to reset SignerState, but only after creating it, so I just replaced this to NewSignerState and using this instead of struct literal.

With that said, ResetRound seems to reset everything except SeenSigners so I've fixed it also.

message/validation/validation.go
moshe-blox: @y0sher im seeing here a lock per msgid, not sure if that was changed since you last checked?

operator/validator/controller.go
moshe-blox: as far as i can see we're not passing networkConfig to runners

but i did remove the redundant validator.Options.Domain field (its already in Options.NetworkConfig.Domain)

message/validation/consensus_validation.go
```diff
@@ -228,52 +290,72 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
...
+		}
+
+		signerState.SeenSigners[string(encodedOperators)] = struct{}{}
 	}
 
+	signerState.MessageCounts.RecordConsensusMessage(signedSSVMessage, consensusMessage)
 	return nil
 }
 
+func (mv *messageValidator) pruneOldSlots(stateBySlot *treemap.Map, lastSlot phase0.Slot) {
```
moshe-blox, line 342: what is the reason for not just using a Go map here?

i prefer to not add a dependency if we dont have to

network/p2p/p2p_sync.go
```diff
@@ -188,8 +188,8 @@ func allPeersFilter(id peer.ID) bool {
 
 func waitSubsetOfPeers(
 	logger *zap.Logger,
-	getSubsetOfPeers func(logger *zap.Logger, vpk spectypes.ValidatorPK, maxPeers int, filter func(peer.ID) bool) (peers []peer.ID, err error),
-	vpk spectypes.ValidatorPK,
+	getSubsetOfPeers func(logger *zap.Logger, senderID []byte, maxPeers int, filter func(peer.ID) bool) (peers []peer.ID, err error),
```
moshe-blox, line 191: we should rename all occurrences of senderID to dutyExecutorID and so on (because the spec recently made this change)

the reason i'm not doing this now is to lessen the number of conflicts with other PRs pending merge to alan/no-fork

it's likely better to do this refactor after merging message validation and at least part of fork support

message/validation/validation.go
```diff
@@ -118,77 +121,105 @@ func (mv *messageValidator) handlePubsubMessage(pMsg *pubsub.Message, receivedAt
...
-		return nil, err
+	decodedMessage.SSVMessage = signedSSVMessage.SSVMessage
+
+	if err := mv.validateSSVMessage(signedSSVMessage.SSVMessage); err != nil {
+		return decodedMessage, err
 	}
 
-	committee, validatorIndices, err := mv.getCommitteeAndValidatorIndices(signedSSVMessage.SSVMessage.GetID())
+	committeeData, err := mv.getCommitteeAndValidatorIndices(signedSSVMessage.SSVMessage.GetID())
 	if err != nil {
```
moshe-blox, line 139: this would be a quite heavy operation to perform on every incoming message (imagine constructing slices of 500 indices at a rate of dozens/hundreds of times per second)

i would not rush to optimize this now, but we should definitely do so before finalizing Alan message validation

caching would not be ideal, as the validator indices may change, instead we can leverage the ValidatorStore to keep track of committees' indices and return them in Committee methods (which already return a Committee struct that we should add an Indices filter to)

the reason the ValidatorStore approach is faster (if done correctly) is because it listens to added/removed shares and doesn't have to recompute everything on each method call

network/p2p/p2p_test.go
```diff
@@ -409,38 +342,44 @@ func (r *dummyRouter) Route(_ context.Context, _ *queue.DecodedSSVMessage) {
...
+
+	signedSSVMsg := &spectypes.SignedSSVMessage{
+		Signatures:  [][]byte{[]byte("sVV0fsvqQlqliKv/ussGIatxpe8LDWhc9uoaM5WpjbiYvvxUr1eCpz0ja7UT1PGNDdmoGi6xbMC1g/ozhAt4uCdpy0Xdfqbv")},
+		OperatorIDs: []spectypes.OperatorID{1, 3, 4},
+		SSVMessage:  ssvMsg,
+	}
+
+	// TODO: remove or use?
+	//signedSSVMsg, err := spectypes.SSVMessageToSignedSSVMessage(ssvMsg, 1, dummySignSSVMessage)
+	//require.NoError(t, err)
```
moshe-blox, line 374: why cant we call this instead of producing our own here?

message/validation/validation.go
```diff
@@ -118,77 +121,105 @@ func (mv *messageValidator) handlePubsubMessage(pMsg *pubsub.Message, receivedAt
...
+}
+
 func (mv *messageValidator) obtainValidationLock(messageID spectypes.MessageID) *sync.Mutex {
 	// Lock this SSV message ID to prevent concurrent access to the same state.
 	mv.validationMutex.Lock()
 	mutex, ok := mv.validationLocks[messageID]
 	if !ok {
 		mutex = &sync.Mutex{}
 		mv.validationLocks[messageID] = mutex
+		// TODO: Clean the map when mutex won't be needed anymore. Now it's a mutex leak...
```
moshe-blox, line 199: we could use the ttlcache package i think we already import, however if we make sure to not instantiate this before verifying the committee/pk & role are correct, then we should not have more than `NumValidators * NumRoles` records in there (so at least we have an upper bound that should prevent mem leaks, and we can estimate the max memory usage to determine if TTL is even worth the effort)

message/validation/signer_state.go
```diff
@@ -4,41 +4,31 @@ package validation
...
-	Slot               phase0.Slot
-	Round              specqbft.Round
-	MessageCounts      MessageCounts
-	ProposalData       []byte
-	EpochDuties        int
-	SeenDecidedLengths map[int]struct{}
+	Round         specqbft.Round
+	MessageCounts MessageCounts
+	ProposalData  []byte
+	SeenSigners   map[string]struct{}
```
moshe-blox, line 15: can use [32]byte instead of string here (to prevent unnecessary conversions)

message/validation/consensus_validation.go
moshe-blox: @MatusKysel yes, but why do we even need this package at all? why are Go maps not enough? i strongly prefer to not add such big dependencies unless they're really worth the risks and other trade-offs.

For example the `ttlcache` package has a clear benefit where we use it (we TTL to reduce memory usage), and its a very small dependency (less risks) and very easy to reason about what its doing without having to go to their docs.

moshe-blox: if we use treemap in order to be able to tell the `Min` for pruning, i think instead we can just create our own struct which holds a standard slot map with a `minimumValue` field and `Put(slot, state)/Get(slot)/Prune(maxSlot)` methods, and when Put is called it can update the minimum value to later be able to prune from it to the maxSlot.

its probably 20 lines of code to avoid importing `gods` which is probably thousands.

message/validation/message_counts.go
```diff
@@ -108,13 +100,8 @@ func (c *MessageCounts) RecordConsensusMessage(signedSSVMessage *spectypes.Signe
...
 	case specqbft.CommitMsgType:
-		switch {
-		case len(signedSSVMessage.GetOperatorIDs()) == 1:
+		if len(signedSSVMessage.GetOperatorIDs()) == 1 {
 			c.Commit++
-		case len(signedSSVMessage.GetOperatorIDs()) > 1:
-			c.Decided++
-		default:
-			panic("unexpected signers length 0") // 0 length should be checked before
 		}
```
moshe-blox, line 105: what happens if there are more than 1 signers? does this mean you can keep sending messages with more than 1 signers as an attack to avoid being blocked by the counter?

message/validation/common_checks.go
```diff
@@ -13,70 +14,108 @@ func (mv *messageValidator) committeeRole(role spectypes.RunnerRole) bool {
...
+	}
+
+	return nil
+}
+
+func (mv *messageValidator) dutyLimit(msgID spectypes.MessageID, validatorIndices []phase0.ValidatorIndex) (int, bool) {
 	switch msgID.GetRoleType() {
 	case spectypes.RoleAggregator, spectypes.RoleValidatorRegistration, spectypes.RoleVoluntaryExit:
-		dutyLimit = 2
+		return 2, true
```
moshe-blox, line 90: VoluntaryExit is risky to put here because we don't limit the rate at which nodes can start this duty, its dictated purely by contract events.

We should probably hook up to the event syncer in some way to ask if such duty exists (if it received an event for this validator and this slot). 

Regarding ValidatorRegistration, its basically scheduled once per 10 epochs for each validator, however it might be necessary to allow 2 per epoch, since we're also performing it immediately after validators register, so this helps avoid ignoring/rejecting cases where registration and scheduling happened to fall on the same epoch for some validator. With that said, we should likely come up with a better solution long term.

moshe-blox (changes requested):

moshe-blox:
Well done @nkryuchkov, very clear comments and much easier to follow the code after this refactor! 💪 

I've made some refactors, and we're leaning towards merging this as it is and to keep the rest of the comments to be resolved by a future PR.

network/p2p/test_utils.go
nkryuchkov: As far as I remember to fix the compilation errors to be able to test the PR

operator/validator/controller.go
nkryuchkov: good catch, thanks

message/validation/consensus_validation.go
nkryuchkov: @moshe-blox treemap is used because there's a loop removing min (stale) slots that are older than `maxSlotsInState`. I think it almost always removes only one slot, but a slot gap is theoretically possible, so keeping slots sorted would help here and just one minimum value is not enough. 

The `gods` library is quite well known and takes just 1.0M, while the total size of the repo and all dependencies is 1.6G.

But I don't mind getting rid of it for simplicity and even having a loop over all values because `maxSlotsInState` is just `34` now and `pruneOldSlots` is not called often

network/p2p/p2p_test.go
```diff
@@ -32,65 +26,6 @@ import (
...
-	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
-	require.NoError(t, err)
-
-	rsaPubKey, ok := pub.(*rsa.PublicKey)
-	require.True(t, ok)
-
-	require.NoError(t, rsa.VerifyPKCS1v15(rsaPubKey, crypto.SHA256, messageHash[:], decodedMsg.Signature[:]))
-	require.Equal(t, testMessage, decodedMsg.Data)
-}
-
```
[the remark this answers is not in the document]
nkryuchkov: `operator/keys` is already well covered

network/p2p/test_utils.go
nkryuchkov: Before the PR the code looked like this:

```go
cfg.MessageValidator = nil //validation.NewMessageValidator(networkconfig.TestNetwork)
```

As the PR changes the signature of `NewMessageValidator`, I adjusted the commented-out code to the new signature and added dependencies. 

However, I don't remember why we assign `nil` now, probably to make some test pass. But probably this should not be part of this PR. I'll think about this once again

network/p2p/test_utils.go
```diff
@@ -159,7 +179,13 @@ func (ln *LocalNet) NewTestP2pNetwork(ctx context.Context, nodeIndex int, keys t
...
 		cfg.MessageValidator = options.MessageValidatorProvider(nodeIndex)
 	} else {
-		cfg.MessageValidator = nil //validation.NewMessageValidator(networkconfig.TestNetwork, validation.WithSelfAccept(selfPeerID, true))
+		cfg.MessageValidator = nil //validation.New(
+		//networkconfig.TestNetwork,
+		//nodeStorage.ValidatorStore(),
+		//dutyStore,
+		//signatureVerifier,
+		//validation.WithSelfAccept(selfPeerID, true),
+		//)
```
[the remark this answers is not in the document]
nkryuchkov: https://github.com/ssvlabs/ssv/pull/1393#discussion_r1668444958

network/p2p/p2p_test.go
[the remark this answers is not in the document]
nkryuchkov: Yes, this test was not p2p-related and the `operator/keys` package is well-covered

network/p2p/p2p_test.go
nkryuchkov: Probably because of the signature, but I reverted this for now

message/validation/validation.go
nkryuchkov: left a comment

message/validation/message_counts.go
nkryuchkov: We don't need to check the message count for decided messages here as it would be unreachable because we already check that the number of signers is correct and the same signers were not seen

message/validation/validation.go
nkryuchkov: @y0sher we use the global lock only for access to map with locks by message ID, I'm not sure I understand what you mean

message/validation/common_checks.go
nkryuchkov: @moshe-blox I agree, this sounds good to me but it looks like something worth a separate PR because this PR doesn't change this logic

message/validation/validation.go
nkryuchkov: @moshe-blox Let's change this in another PR because this PR just adds a TODO?

message/validation/common_checks.go
moshe-blox: @nkryuchkov +1
thanks

message/validation/validation.go
```diff
@@ -6,24 +6,30 @@ package validation
...
 	"github.com/ssvlabs/ssv/monitoring/metricsreporter"
+	"github.com/ssvlabs/ssv/network/commons"
 	"github.com/ssvlabs/ssv/networkconfig"
 	"github.com/ssvlabs/ssv/operator/duties/dutystore"
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/queue"
 	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
 	"github.com/ssvlabs/ssv/registry/storage"
 )
 
+const MaxPartialSignatureMsgSize = 1952
```
nkryuchkov, line 31: it should've been 144020, we need to make sure all constants are correct

message/validation/consensus_validation.go
nkryuchkov: we decided to consider using an array instead of treemap

message/validation/common_checks.go
nkryuchkov: @moshe-blox and I decided that VoluntaryExit changes won't be implemented now, so only ValidatorRegistration will be done

PR #1407, opened 2024-05-28
Use Alan shares in genesis message validation

nkryuchkov:
- Use Alan shares in genesis message validation
- Fix broken unit tests

### DIFF
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -10,9 +10,7 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	genesisspecqbft "github.com/ssvlabs/ssv-spec-pre-cc/qbft"
 	genesisspectypes "github.com/ssvlabs/ssv-spec-pre-cc/types"
-	spectypes "github.com/ssvlabs/ssv-spec-pre-cc/types"
 	alanspecqbft "github.com/ssvlabs/ssv-spec/qbft"
-	alanspectypes "github.com/ssvlabs/ssv-spec/types"
 	"golang.org/x/exp/slices"
 
 	"github.com/ssvlabs/ssv/protocol/v2/qbft/roundtimer"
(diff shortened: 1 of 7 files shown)


message/validation/genesis/consensus_validation.go
```diff
@@ -10,32 +10,22 @@ import (
...
 	var consensusDescriptor ConsensusDescriptor
 
-	if mv.operatorDataStore != nil && mv.operatorDataStore.OperatorIDReady() {
-		if mv.inCommittee(share) {
-			mv.metrics.InCommitteeMessage(alanspectypes.MsgType(spectypes.SSVConsensusMsgType), mv.isDecidedMessage(signedMsg))
-		} else {
-			mv.metrics.NonCommitteeMessage(alanspectypes.MsgType(spectypes.SSVConsensusMsgType), mv.isDecidedMessage(signedMsg))
-		}
-	}
-
```
y0sher, line 38: can you example how come its not needed anymore? OperatorIDReady used to wait here until we have our operator ID (from db or sync)

nkryuchkov: It's only used for metric if message belongs to committee which we don't use as far as I know (some time ago I added it temporarily to debug something and didn't delete it). Also, we don't have this metric in the new message validation, so I deleted it here too. If you think we need it, I can add it to both

message/validation/genesis/validation.go
```diff
@@ -46,7 +45,7 @@ const (
 	clockErrorTolerance = time.Millisecond * 50
 
 	maxMessageSize             = maxConsensusMsgSize
-	maxConsensusMsgSize        = 8388608
+	maxConsensusMsgSize        = 6291829
```
y0sher, line 48: does this also relates to post-conesnsus messages? afaik they got bigger

nkryuchkov: It's the latest value from stage: https://github.com/ssvlabs/ssv/blob/stage/message/validation/validation.go#L47

y0sher: @MatheusFranco99 doesn't need to be calculated again for CC??

PR #1422, opened 2024-06-06
Fix incorrect removeIndex in loop 

AKorpusenko:
## Bug description
`Committee.StartDuty` calls `FilterCommitteeDuty` which deletes some duties from the `duty.BeaconDuties`. But this is done wrong and we accidentally caught this because with message: `panic: runtime error: slice bounds out of range [8:7]`. 

## Fix
Slice items can't be randomly deleted correctly. 
Changed func `FilterCommitteeDuty` to collect which ids should be removed and then run `removeIndices` to get the modified filtered slice. 
Checking the `indicesToRemove` not to be out of range or contain duplicates.


protocol/v2/ssv/validator/committee.go
```diff
@@ -150,10 +151,18 @@ func (c *Committee) PushToQueue(slot phase0.Slot, dec *queue.DecodedSSVMessage)
...
+	for _, index := range indicesToRemove {
+		// Panic on incorrect indices
+		if index < 0 || index >= len(s) {
+			panic(fmt.Sprintf("index %d out of range of %v", index, indicesToRemove))
+		}
+		s = append(s[:index], s[index+1:]...)
+	}
+
+	return s
 }
```
y0sher, line 166: 1. don't panic. - might be a security concern. return error, especially as this part of normal node operation.
2. It's better to not touch the original passed slice as its part of the `CommitteeRunner` and we "don't know" about its thread safety and such. create a new slice instead.

y0sher (changes requested):
please see my comment.

y0sher (approved):

PR #1424, opened 2024-06-07
fix validator index in spec share

MatusKysel:
This change resolves issue when new validator are registered and index in spectypes share is not updated from metadata.


operator/duties/committee.go
```diff
@@ -139,7 +139,7 @@ func (h *CommitteeHandler) processExecution(period uint64, epoch phase0.Epoch, s
 	vsmap := make(map[phase0.ValidatorIndex]spectypes.CommitteeID, 0)
 	vs := h.validatorProvider.SelfParticipatingValidators(epoch)
 	for _, v := range vs {
-		vsmap[v.ValidatorIndex] = v.CommitteeID()
```
y0sher, line 142: if you updated `ValidatorIndex` in shares.go why do you need to change here to `BeaconMetadata.Index`?

y0sher (approved):

PR #1428, opened 2024-06-10
fix: (alan) use a lock for writing to the queues

y0sher:
We were writing and creating the queue without acquiring the lock, resulting in concurrent writes and panics.


AKorpusenko (approved):
Nice catch!

PR #1434, opened 2024-06-16
feat: (bootnode) filter nodes by domain type

moshe-blox:


y0sher (dismissed):
lgtm, also looked at the bootnode code.

network/discovery/dv5_service.go
```diff
@@ -198,6 +199,29 @@ func (dvs *DiscV5Service) initDiscV5Listener(logger *zap.Logger, discOpts *Optio
 	if err != nil {
 		return err
 	}
+	dv5Cfg.NodeFilter = func(n *enode.Node) bool {
+		logger := zap.L().With(fields.ENR(n))
+		if n == nil {
+			logger.Debug("discoveryNodeFilter: nil node")
```
moshe-blox, line 205: all debug logs can be safely removed after testing

go.mod
```diff
@@ -229,3 +229,5 @@ replace github.com/google/flatbuffers => github.com/google/flatbuffers v1.11.0
 replace github.com/dgraph-io/ristretto => github.com/dgraph-io/ristretto v0.1.1-0.20211108053508-297c39e6640f
 
 replace github.com/attestantio/go-eth2-client => github.com/y0sher/go-eth2-client v0.0.0-20240305092513-92d154b203a6
+
+replace github.com/ethereum/go-ethereum => github.com/moshe-blox/go-ethereum v0.0.0-20240616101439-ced204b13167
```
nkryuchkov, line 233: As I understand, it's not temporary, so it will stay here, right? If so, I don't think it's a good idea to maintain a fork of `go-ethereum`. 

If we must make a change, I'd prefer to create an issue or a PR in go-ethereum (if there's one already, we need a comment) or to fork just the changed package from `go-ethereum` and import it as well as the original `go-ethereum`.

Anyway, even if it's temporary, I think we need a detailed comment here explaining why it was done.

nkryuchkov (changes requested):
Could you please add a short PR description?

PR #1435, opened 2024-06-16
refactor: low discovery interval

moshe-blox:
This PR changes the discovery mechanism to only sleep after discovering a relevant peer, rather than after discovering _any_ peer. Plus, the interval is reduced from 1s to 100ms.


y0sher (approved):
lgtm

PR #1436, opened 2024-06-19
Added stop committee runner queue consumption trigger

AKorpusenko:


protocol/v2/ssv/runner/committee.go
```diff
@@ -43,6 +43,8 @@ type CommitteeRunner struct {
 	started       time.Time
 	consensusDone time.Time
 	postStarted   time.Time
+
+	stopCh chan struct{}
```
y0sher, line 47: why is this here if it has no meaning for the committee runner itself?

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -72,7 +72,12 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.DecodedSSVMessa
 
 // ConsumeQueue consumes messages from the queue.Queue of the controller
 // it checks for current state
-func (v *Committee) ConsumeQueue(logger *zap.Logger, slot phase0.Slot, handler MessageHandler) error {
+func (v *Committee) ConsumeQueue(
+	logger *zap.Logger,
+	slot phase0.Slot,
+	handler MessageHandler,
+	stopCh <-chan struct{},
```
y0sher, line 79: should've probably just used a context dedicated to this slot queue.

y0sher (changes requested):
create a context when you create the queue and pass it in.

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -72,11 +72,32 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.DecodedSSVMessa
...
+	defer func() {
+		v.mtx.Lock()
+		defer v.mtx.Unlock()
+
+		var ok bool
+		q, ok = v.Queues[slot]
+		if !ok {
+			logger.Warn("committee queue not found for slot while stopping the consumer", fields.Slot(slot))
+			return
+		}
```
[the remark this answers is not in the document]
AKorpusenko: Added. Thanks

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -72,11 +72,33 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.DecodedSSVMessa
...
 	var q queueContainer
+
+	// in case of any error try to call the ctx.cancel to prevent the ctx leak
+	defer func() {
+		var ok bool
+
+		v.mtx.Lock()
+		q, ok = v.Queues[slot]
+		v.mtx.Unlock()
+
```
y0sher, line 90: `StartDuty` is already locking, so this is potential "deadlock" or just double locking, although this happens in a goroutine and the lock is released in start duty. Still I think its better to just pass the Queue `q` from `StartDuty` as we're already using the mutex to get it there.

by passing it to this function you don't need to use the mutex here at all.

operator/validator/controller.go
```diff
@@ -864,6 +864,9 @@ func (c *controller) onShareStop(pubKey spectypes.ValidatorPK) {
 	if ok {
 		vc.RemoveShare(v.Share.Share.ValidatorIndex)
 		if len(vc.Shares) == 0 {
+			for _, slotQueue := range vc.Queues {
+				slotQueue.StopQueueF()
+			}
```
y0sher, line 869: you can't do this here, accessing Queues requires mutex acquiring. best is to have a function inside the committee struct that handles this.

y0sher (changes requested):
concurrency issues..

operator/validator/controller.go
```diff
@@ -864,6 +864,7 @@ func (c *controller) onShareStop(pubKey spectypes.ValidatorPK) {
 	if ok {
 		vc.RemoveShare(v.Share.Share.ValidatorIndex)
 		if len(vc.Shares) == 0 {
+			vc.StopQueues()
```
y0sher, line 867: `Stop` or `Close` is more appropriate

protocol/v2/ssv/validator/events.go
```diff
@@ -42,13 +43,22 @@ func (c *Committee) handleEventMessage(logger *zap.Logger, msg *queue.DecodedSSV
 		if err != nil {
 			return err
 		}
+
 		c.mtx.Lock()
-		dutyRunner := c.Runners[slot] // TODO: err check , runner exist?
+		q := c.Queues[slot]
+		dutyRunner, exists := c.Runners[slot]
 		c.mtx.Unlock()
+		if !exists {
```
y0sher, line 51:
```suggestion
		q, exists := c.Queues[slot]
		dutyRunner, exists := c.Runners[slot]
		c.mtx.Unlock()
		if !exists || !qexists {
```

queue should be present if there's a runner but lets be on the safe side anyway.

y0sher (changes requested):
small change

y0sher (approved):

protocol/v2/ssv/validator/events.go
```diff
@@ -42,13 +43,23 @@ func (c *Committee) handleEventMessage(logger *zap.Logger, msg *queue.SSVMessage
...
+		if !rExists || !qExists {
+			logger.Error("no committee runner or queue found for slot", fields.Slot(slot), fields.MessageID(msg.MsgID))
+			return nil
+		}
+
 		if err := dutyRunner.GetBaseRunner().QBFTController.OnTimeout(logger, *eventMsg); err != nil {
 			return fmt.Errorf("timeout event: %w", err)
 		}
+
+		q.StopQueueF()
```
moshe-blox, line 61: why are we stopping the queue on timeout message?

AKorpusenko:
Replaced with PR https://github.com/ssvlabs/ssv/pull/1556 based on the `stage` branch

PR #1446, opened 2024-07-02
align genesis runners and qbft with main

MatusKysel:
Alignment of runners and qbft package with main branch (mainnet)

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -266,6 +266,9 @@ var StartNodeCmd = &cobra.Command{
 		cfg.SSVOptions.ValidatorOptions.RecipientsStorage = nodeStorage
 		cfg.SSVOptions.ValidatorOptions.GasLimit = cfg.ConsensusClient.GasLimit
 
+		genesisKeyManager := &ekm.GenesisKeyManagerAdapter{KeyManager: keyManager}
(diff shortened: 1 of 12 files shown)


protocol/genesis/ssv/runner/aggregator.go
```diff
@@ -285,26 +270,20 @@ func (r *AggregatorRunner) executeDuty(logger *zap.Logger, duty *genesisspectype
 	signedPartialMsg := &genesisspectypes.SignedPartialSignatureMessage{
 		Message:   msgs,
 		Signature: signature,
-		Signer:    r.operatorSigner.GetOperatorID(),
+		Signer:    r.GetShare().Committee[0].Signer,
```
moshe-blox, line 273: this is the signer of the message, it should be the current operator which is running this runner rather than the first operator in the committee

previously shares had this weird `OperatorID` field which is basically that, however it was removed in Alan shares

so instead, we should either pass the operator id as a field, or expose it via the KeyManager interface (it means we need to define our own GenesisKeyManager interface which extends the current and adds GetOperatorID() as is present in OperatorSigner in Alan)

if keyManager doesn't already has access to the operator ID internally, then i guess it'd be simpler to just pass operator ID as a new field to runners (in their constructor).

PR #1449, opened 2024-07-05
reject non-Alan peers

pavelkrolevets:


moshe-blox:
Domain type already contains the fork version (in the 4th byte), so the domain type for Alan is different than the genesis one,  and there is already rejection of connections due to differing domain type so in effect we already reject non-Alan peers, without having to add a new 'fork version' field.

What is missing for fork support is being able to switch our domain type in runtime.

PR #1450, opened 2024-07-08
Alan/switch domain at runtime

pavelkrolevets:
- get domain based on beacon fork and epoch. After we reach epoch to switch to Alan fork, the domain will be changed at runtime.


networkconfig/config.go
```diff
@@ -75,3 +74,25 @@ func (n NetworkConfig) SlotsPerEpoch() uint64 {
...
+		if n.AlanForked(n.Beacon.EstimatedCurrentSlot()) {
+			return spectypes.AlanMainnet
+		}
+		return spectypes.GenesisMainnet
+	case [4]byte{0x01, 0x01, 0x70, 0x00}:
+		if n.AlanForked(n.Beacon.EstimatedCurrentSlot()) {
+			return spectypes.JatoAlanTestnet
+		}
+		return spectypes.JatoTestnet
+	default:
```
pavelkrolevets, line 92: @MatusKysel @moshe-blox  is this what you described? 
Im not sure herewhich beacon fork related to JatoTestnet/JatoAlanTestnet
Also should we default to mainnet or better return error?

pavelkrolevets: Another question: we need to switch a domain type at ENR after we reach the Alan epoch. I can use `func DecorateWithDomainType(domainType spectypes.DomainType)` to update it, but do we have a worker which can trigger it?

moshe-blox:
@pavelkrolevets 

Few notes:
- Let's rename `Domain()` to `DomainType()` so it's less ambiguous
- Let's make sure every domain type consumer is using the method instead of storing its own copy of domain type, so nothing breaks on the fork epoch. By "domain type consumer" I mean every struct or long-running method/function which accepts a domain type (for example, a function which spawns a long-running goroutine should also call DomainType method instead of working with a potentially outdated copy)
- For DomainType consumers which don't have access to NetworkConfig and wouldn't really need all of it, it's preferable to pass a `DomainTypeProvider` interface which would only have the `DomainType` method

Please don't forget about discovery which still seems to be holding a copy: https://github.com/ssvlabs/ssv/blob/e865d6915005aab1d1a241bd2635ce8dfeddb1af/network/discovery/dv5_service.go#L60

networkconfig/config.go
moshe-blox: @pavelkrolevets DomainType changes according to SSV forks, not Beacon forks

So you could use the domain type value passed to NetworkConfig on creation (such as in holesky-stage.go) and only edit the 4th byte it represents the 'fork version'

For example, if we're before AlanForkEpoch, then the version is 0, if we're after then its 1

moshe-blox: @pavelkrolevets 

Let's create a 'ForkListener' struct which uses a SlotTicker in the same way its used in `duties/attester.go` for example, and make it wait for the Alan fork epoch so that it could trigger the ENR update such as is done here: https://github.com/ssvlabs/ssv/blob/e865d6915005aab1d1a241bd2635ce8dfeddb1af/network/discovery/dv5_service.go#L271

pavelkrolevets: @moshe-blox great, thank you for clarification, ill update approach. Now everything clear.

networkconfig/config.go
```diff
@@ -75,3 +75,14 @@ func (n NetworkConfig) SlotsPerEpoch() uint64 {
...
+// Domain returns current domain type of the network
+func (n NetworkConfig) DomainType() spectypes.DomainType {
+	if n.AlanForked(n.Beacon.EstimatedCurrentSlot()) {
+		forkDomain := make([]byte, 4)
+		copy(forkDomain, n.domainType[:2])
+		copy(forkDomain[2:], []byte{0x1})
+		return spectypes.DomainType(forkDomain)
+	}
+	return n.domainType
+}
```
pavelkrolevets, line 88: @moshe-blox changed to this. So after we reach Alan fork 4th byte will be changed to 0x1

pavelkrolevets: @moshe-blox @MatusKysel does it make sense to branch out from `alan/no-fork` and redo the changes? will be much cleaner at new PR?

operator/duties/scheduler.go
```diff
@@ -151,6 +152,7 @@ func NewScheduler(opts *SchedulerOptions) *Scheduler {
 			NewSyncCommitteeHandler(dutyStore.SyncCommittee),
 			NewVoluntaryExitHandler(opts.ValidatorExitCh),
 			NewCommitteeHandler(dutyStore.Attester, dutyStore.SyncCommittee),
+			NewForkListener(opts.P2PNetwork),
```
pavelkrolevets, line 155: moved `ForkListener` to scheduler, not sure that its very elegant, but less changes anyway....

operator/duties/fork_listener.go
```diff
@@ -0,0 +1,58 @@
...
+		case <-f.indicesChange:
+			slot := f.network.Beacon.EstimatedCurrentSlot()
+			if f.network.AlanForked(slot) {
+				if err := f.p2pnet.UpdateDomainTypeAtFork(f.logger); err != nil {
+					f.logger.Warn("could not update ENR at fork", zap.Error(err))
+				}
+			}
+		}
+	}
+}
```
pavelkrolevets, line 54: @moshe-blox @MatusKysel

networkconfig/config.go
pavelkrolevets: Ok closing, reopening at new PR

PR #1454, opened 2024-07-11
Alan/switch domain at runtime

pavelkrolevets:
* add `DomainType()` to `NetworkConfig` which returns fork domain type based on reaching fork epoch
* add fork listener to `operator/duties/scheduler.go` to trigger ENR update after reaching fork epoch
* add `network/discovery/dv5_service.go UpdateDomainTypeAtFork()` to update/publish ENR  at fork epoch


operator/duties/fork_listener.go
```diff
@@ -0,0 +1,48 @@
+package duties
```
y0sher, line 1: I don't think this is part of this package.

networkconfig/holesky-stage.go
```diff
@@ -11,7 +11,8 @@ import (
 var HoleskyStage = NetworkConfig{
 	Name:                 "holesky-stage",
 	Beacon:               beacon.NewNetwork(spectypes.HoleskyNetwork),
-	Domain:               [4]byte{0x00, 0x00, 0x31, 0x99}, // TODO: (Alan) revert
+	GenesisDomainType:    [4]byte{0x00, 0x00, 0x31, 0x12},
+	AlanDomainType:       [4]byte{0x00, 0x00, 0x31, 0x99},
```
y0sher, line 15: why not `0x13` (leave space for next forks)

networkconfig/holesky-e2e.go
```diff
@@ -10,7 +10,8 @@ import (
 var HoleskyE2E = NetworkConfig{
 	Name:                 "holesky-e2e",
 	Beacon:               beacon.NewNetwork(spectypes.HoleskyNetwork),
-	Domain:               spectypes.DomainType{0x0, 0x0, 0xee, 0x1},
+	GenesisDomainType:    spectypes.DomainType{0x0, 0x0, 0xee, 0x0},
+	AlanDomainType:       spectypes.DomainType{0x0, 0x0, 0xee, 0x1},
```
y0sher, line 14: probably `CurrentDomainType` and `NextDomainType` are better names and can be used in future forks as well

moshe-blox: You're right in essence, although I think it could be useful to give the domain type the name of the fork, so we can later add a 3rd one (or just rename the Alan one)

networkconfig/holesky-stage.go
moshe-blox: done

operator/duties/fork_listener.go
moshe-blox: deleted

PR #1459, opened 2024-07-14
update go and libp2p

cryptix:
I saw this depended on Go 1.20 because of an old quic-go / libp2p version. So this updates those.


y0sher:
this is indeed a needed update, but we need to do this on `stage` or (we are working on big feature now on `alan/no-fork`).

y0sher:
@cryptix I'm closing this PR and will review if you go forward with what I mentioned above. thanks 🙏

PR #1460, opened 2024-07-14
fix: pass only relevant duty validator shares into the slot committee runner

y0sher:
We were copying all shares when only the shares that are relevant for this slot duty should go into the runner.

### DIFF
--- a/protocol/v2/ssv/validator/committee.go
+++ b/protocol/v2/ssv/validator/committee.go
@@ -119,6 +119,10 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 		}
 	}
 
+	if len(duty.BeaconDuties) == 0 {
+		return errors.New("CommitteeDuty has no valid beacon duties")
+	}
+
 	r := c.CreateRunnerFn(duty.Slot, validatorShares)
 	// Set timeout function.
 	r.GetBaseRunner().TimeoutF = c.onTimeout
--- a/protocol/v2/ssv/validator/committee.go
+++ b/protocol/v2/ssv/validator/committee.go
@@ -98,12 +98,12 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 		return errors.New(fmt.Sprintf("CommitteeRunner for slot %d already exists", duty.Slot))
 	}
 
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/validator/committee.go
```diff
@@ -98,12 +98,12 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 		return errors.New(fmt.Sprintf("CommitteeRunner for slot %d already exists", duty.Slot))
 	}
 
-	var sharesCopy = make(map[phase0.ValidatorIndex]*spectypes.Share, len(c.Shares))
-	for k, v := range c.Shares {
-		sharesCopy[k] = v
+	validatorShares := make(map[phase0.ValidatorIndex]*spectypes.Share, len(duty.BeaconDuties))
+	for _, bd := range duty.BeaconDuties {
+		validatorShares[bd.ValidatorIndex] = c.Shares[bd.ValidatorIndex]
```
moshe-blox, line 103: should probably add a safety check that if we're missing one of the shares then to return an error (or less risky to just log an error & continue)

moshe-blox: although maybe we should remove the duty, log an error & continue instead

y0sher: we delete when we call `onShareStop` but it might be right after we sent a duty. so its best to remove the duty like you said.

y0sher: @moshe-blox please take a look on recent commit

protocol/v2/ssv/validator/committee.go
```diff
@@ -98,12 +98,28 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
...
+	}
+	// Remove beacon duties that don't have a share
+	if len(toRemove) > 0 {
+		newDuties, err := removeIndices(duty.BeaconDuties, toRemove)
+		if err != nil {
+			logger.Warn("could not remove beacon duties", zap.Error(err), zap.Ints("indices", toRemove))
+		} else {
+			duty.BeaconDuties = newDuties
+		}
 	}
```
moshe-blox, line 121: would add here to return err if BeaconDuties is empty

moshe-blox (approved):

PR #1462, opened 2024-07-15
message validation: optimize constructing validator indices

nkryuchkov:
Move slice construction to the validator store and keep it updated there


registry/storage/validatorstore.go
y0sher, line 1: lets stay consistent and use either `share.ValidatorIndex` or `share.BeaconMetadata.Index`, ideally  `share.ValidatorIndex`

y0sher (changes requested):

registry/storage/validatorstore.go
nkryuchkov: @y0sher I agree, done

y0sher (approved):

PR #1468, opened 2024-07-15
message validation: check voluntary exit existence

nkryuchkov:
According to https://github.com/ssvlabs/ssv/pull/1393#discussion_r1667687976


operator/duties/dutystore/voluntary_exit.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"sync"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
+
+type VoluntaryExitDuties struct {
+	mu sync.RWMutex
+	m  map[phase0.Slot]map[phase0.BLSPubKey]int
```
y0sher, line 11: we need to clean this map at some point, maybe remove the duties after they are executed? or that's too short?

y0sher (changes requested):
clean the new duty store you added.

operator/duties/dutystore/voluntary_exit.go
nkryuchkov: done, let's clean them after 1 epoch

moshe-blox:
@nkryuchkov @y0sher 
another refactor is required for this to work, since the event handler currently doesnt fire tasks if its not your validator:
https://github.com/ssvlabs/ssv/blob/adaa86b1da7a742c27467622427e089adb8a1148/eth/eventhandler/handlers.go#L514-L516

we should probably try something like this:
image

and in DutyScheduler always add to duties map but only execute our own duties

y0sher (approved):
lgtm.

PR #1469, opened 2024-07-16
Refresh P2P scoring parameters

MatheusFranco99:
# Overview

The GossipSub scoring parameters are dependent on the topics' message rate estimation which, on the other hand, is dependent on the number and distribution of validators in the topic. Thus, such parameters should be updated once the network changes (e.g. when validators are added or removed).

To accomplish this, this PR:
- Adds a `UpdateScoreParams` function, to `topics.Controller`, that updates the parameters.
- Adds a `UpdateScoreParams` function, to `P2PNetwork`, that calls `topicsController. UpdateScoreParams` once per epoch and is called asynchronously in `operatorNode.Start`.

### TODO
- This solution updates the parameters every epoch. It would be better to update it once a validator addition or removal event happens. This should be done in a later PR.


network/topics/pubsub.go
```diff
@@ -108,7 +109,7 @@ func (cfg *PubSubConfig) initScoring() {
 }
 
 // NewPubSub creates a new pubsub router and the necessary components
-func NewPubSub(ctx context.Context, logger *zap.Logger, cfg *PubSubConfig, metrics Metrics) (*pubsub.PubSub, Controller, error) {
+func NewPubSub(ctx context.Context, logger *zap.Logger, cfg *PubSubConfig, metrics Metrics, getCommittees func() []*storage.Committee) (*pubsub.PubSub, Controller, error) {
```
moshe-blox, line 112: if you define an interface here for just that method it's cleaner and also on the caller side you can just pass the ValidatorStore as it is (without declaring an anonymous func)
example:
```suggestion
type CommitteesProvider interface {
    Committees() []*storage.Committee
}

// NewPubSub creates a new pubsub router and the necessary components
func NewPubSub(ctx context.Context, logger *zap.Logger, cfg *PubSubConfig, metrics Metrics, committeesProvider CommitteesProvider) (*pubsub.PubSub, Controller, error) {
```

MatheusFranco99: Awesome, thanks!

moshe-blox (approved):

PR #1477, opened 2024-07-16
feature: (alan) Dual subnets

y0sher:
### Description

Committees are registered their own subnets, this is different than previous distribution was based on validator pubkey. In order to fork we include both subnets in discovery and stop broadcasting on old topics one the fork has passed. 

### Notes 

This is based on https://github.com/ssvlabs/ssv/pull/1454 ( will change base to `alan/no-fork` after its merged) and https://github.com/ssvlabs/ssv/pull/1466

### TODO
- [x] - scoring strategy update at fork 
- [x] - Trigger scoring update on the fork epoch


network/p2p/p2p.go
```diff
@@ -260,57 +273,99 @@ func (n *p2pNetwork) UpdateSubnets(logger *zap.Logger) {
...
+				// TODO: alan - optimization to get active committees only
+				share := n.nodeStorage.Shares().Get(nil, pbBytes)
+				if share == nil {
+					return false
+				}
+				cid := share.CommitteeID()
+				subnet := commons.CommitteeSubnet(cid)
+				updatedSubnets[subnet] = byte(1)
+				return true
+			})
```
MatheusFranco99, line 302: Isn't this implementing the alan logic
```go
subnet := commons.CommitteeSubnet(cid)
```
while inside "genesis if" (`!n.cfg.Network.PastAlanFork()`) ?

network/p2p/p2p.go
```diff
@@ -251,6 +258,12 @@ func (n *p2pNetwork) isReady() bool {
 func (n *p2pNetwork) UpdateSubnets(logger *zap.Logger) {
 	// TODO: this is a temporary fix to update subnets when validators are added/removed,
 	// there is a pending PR to replace this: https://github.com/ssvlabs/ssv/pull/990
+	var (
+		topicScoreUpdateInterval = 32 * n.cfg.Network.SlotDurationSec()
+		topicScoreUpdateTicker   = time.NewTicker(topicScoreUpdateInterval)
+	)
+	defer topicScoreUpdateTicker.Stop()
+
```
MatheusFranco99, line 266: I believe this was going to be used for updating the topic score parameters. But there is already a function for it (the one below this one) and these variables are unused, so we can remove it, right?

network/p2p/p2p_pubsub.go
```diff
@@ -157,62 +143,89 @@ func (n *p2pNetwork) Subscribe(pk spectypes.ValidatorPK) error {
...
+		return fmt.Errorf("could not find share for validator %s", hex.EncodeToString(pk[:]))
+	}
+
+	err := n.subscribeCommittee(share.CommitteeID())
+	if err != nil {
+		return fmt.Errorf("could not subscribe to committee: %w", err)
+	}
+	if !n.cfg.Network.PastAlanFork() {
+		return n.subscribeValidator(pk)
+	}
```
MatheusFranco99, line 158: Just to confirm, we are subscribing to the committee-related subnet even though we may not already be on post-fork, right? Is this for safety?

y0sher: Since not subscribing to committee subnets and then subscribing right at the fork epoch might result in us not having enough peers at the fork epoch, we subscribe in advance to both.

network/p2p/p2p.go
y0sher: yeah lets remove

network/p2p/p2p.go
y0sher: you're right, we should take this part out of the if and only include in the if the pre-fork behavior

y0sher: should look lie this:

```go
		n.activeValidators.Range(func(pkHex string, status validatorStatus) bool {
			subnet := commons.ValidatorSubnet(pkHex)
			updatedSubnets[subnet] = byte(1)
			return true
		})

```

network/p2p/p2p_pubsub.go
MatheusFranco99: Got it. Great! Thanks for the clarification

network/p2p/p2p.go
MatheusFranco99: Removing

network/p2p/p2p.go
MatheusFranco99: Okk, updating

PR #1478, opened 2024-07-17
update go and libp2p

cryptix:
#1459 take 2 - hope this time on the correct branch!

* previous version of transitive dependency quic-go was holding back builds on newer Go versions
* update Dockerfile to recent Go version


cryptix:
I tried `docker build` but I'm not familiar enough with your process here to understand why these version pins are so specific nor versed enough in debian to re-create these.

https://gist.github.com/cryptix/b2cce81f0b2c409c3d8efeeb606ccd7f

nkryuchkov:
Hi @cryptix, thank you for the contribution. We decided to update to Go 1.22, it will be done at https://github.com/ssvlabs/ssv/pull/1560

PR #1479, opened 2024-07-17
fix: (alan) bring back slashing check from spec

y0sher:
### Description 

On `validator/controller.go` we were creating the functions to start a new `committeeRunner`, like before every runner uses a value check to make sure its running consensus on valid and non slashable value. We commented this part out and used our own version that doesn't use slashing check, this was because value check are usually implemented in the spec and the spec wasn't ready when this code was written.

### Changes

Now that the spec is ready, the code is aligned to use the `beaconVoteValCheck` from the spec which includes a value check for all the provided shares.

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1185,7 +1185,7 @@ func hasNewValidators(before []phase0.ValidatorIndex, after []phase0.ValidatorIn
 func SetupCommitteeRunners(
 	ctx context.Context,
 	options validator.Options,
-) func(slot phase0.Slot, shares map[phase0.ValidatorIndex]*spectypes.Share) *runner.CommitteeRunner {
+) validator.CommitteeRunnerFunc {
 	buildController := func(role spectypes.RunnerRole, valueCheckF specqbft.ProposedValueCheckF) *qbftcontroller.Controller {
 		config := &qbft.Config{
 			BeaconSigner:      options.Signer,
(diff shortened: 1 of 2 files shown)


operator/validator/controller.go
```diff
@@ -1250,7 +1214,11 @@ func SetupCommitteeRunners(
 	return func(slot phase0.Slot, shares map[phase0.ValidatorIndex]*spectypes.Share) *runner.CommitteeRunner {
 		// Create a committee runner.
 		epoch := options.NetworkConfig.Beacon.GetBeaconNetwork().EstimatedEpochAtSlot(slot)
-		valCheck := TempBeaconVoteValueCheckF(options.Signer, slot, options.SSVShare.Share.SharePubKey, epoch) // TODO: (Alan) fix slashing check (committee is not 1 pubkey)
+		sharepbs := make([]spectypes.ShareValidatorPK, len(shares))
+		for i, s := range shares {
+			sharepbs[i] = s.SharePubKey
+		}
+		valCheck := specssv.BeaconVoteValueCheckF(options.Signer, slot, sharepbs, epoch)
```
GalRogozinski, line 1221: It is best to pass only shares of validators doing the attestation duty...
since only they are slashable

y0sher: Done, please check updated code.

protocol/v2/ssv/validator/committee.go
```diff
@@ -98,6 +100,7 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 		return errors.New(fmt.Sprintf("CommitteeRunner for slot %d already exists", duty.Slot))
 	}
 
+	attestingValidators := make([]spectypes.ShareValidatorPK, 0, len(duty.BeaconDuties))
```
GalRogozinski, line 103:
```suggestion
	slashableValidators := make([]spectypes.ShareValidatorPK, 0, len(duty.BeaconDuties))
```
I think this name is better

GalRogozinski (approved):
LGTM

PR #1482, opened 2024-07-17
Message Validation Review

MatheusFranco99:
I've review comments with a `MSGVALIDATIONREVIEW` tag.

### DIFF
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -207,7 +207,6 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 ) error {
 	signerState := state.GetSignerState(signer)
 
-	// MSGVALIDATIONREVIEW: validateJustifications is also called in the ending of this function.
 	if signerState == nil {
 		return mv.validateJustifications(share, signedMsg)
 	}
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -45,6 +45,8 @@ func (mv *messageValidator) validateConsensusMessage(
 		return consensusDescriptor, msgSlot, e
 	}
 
+	// MSGVALIDATIONREVIEW: it seems that we are missing the "invalid identifier" check. I.e., the SSVMessage.Identifier should be equal to the qbftMessage.Identifier
+
 	if err := mv.validateSignatureFormat(signedMsg.Signature); err != nil {
 		return consensusDescriptor, msgSlot, err
 	}
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -45,8 +45,6 @@ func (mv *messageValidator) validateConsensusMessage(
 		return consensusDescriptor, msgSlot, e
 	}
 
-	// MSGVALIDATIONREVIEW: it seems that we are missing the "invalid identifier" check. I.e., the SSVMessage.Identifier should be equal to the qbftMessage.Identifier
-
 	if err := mv.validateSignatureFormat(signedMsg.Signature); err != nil {
 		return consensusDescriptor, msgSlot, err
 	}
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -131,6 +131,8 @@ func (mv *messageValidator) validateConsensusMessage(
 			signerState.Reset(msgRound)
 		}
 
+		// MSGVALIDATIONREVIEW: I think we should add a condition here that the msg is a proposal msg.
+		// Notice that a RC message also has FullData and would set the signer's state ProposalData
 		if mv.hasFullData(signedMsg) && signerState.ProposalData == nil {
 			signerState.ProposalData = signedMsg.FullData
 		}
@@ -203,6 +205,7 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 ) error {
 	signerState := state.GetSignerState(signer)
 
+	// MSGVALIDATIONREVIEW: validateJustifications is also called in the ending of this function.
 	if signerState == nil {
 		return mv.validateJustifications(share, signedMsg)
 	}
--- a/message/validation/genesis/message_counts.go
+++ b/message/validation/genesis/message_counts.go
@@ -50,6 +50,11 @@ func (c *MessageCounts) ValidateConsensusMessage(msg *specqbft.SignedMessage, li
 			return err
 		}
 	case specqbft.CommitMsgType:
+		// MSGVALIDATIONREVIEW: we should use "else if len(msg.Signers) > 1" and then "else {panic()}" or "else {return Error...}" to cover
+		// the case where len(msg.Signers) == 0.
+		// Ofc that this should be checked in previous rules. However, it may be better to be independent in that sense
+		// and, if we have problems in this other rule, this bad message would pass unnoticed through the check here
+		// e.g. below, in RecordConsensusMessage, there is a similar check
 		if len(msg.Signers) == 1 {
 			if c.Commit >= limits.Commit {
 				err := ErrTooManySameTypeMessagesPerRound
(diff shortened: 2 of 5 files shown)


message/validation/partial_validation.go
```diff
@@ -112,6 +118,8 @@ func (mv *messageValidator) validatePartialSignatureMessageSemantics(
 			return err
 		}
 
+		// MSGVALIDATIONREVIEW: since we don't expect operators to be synced on the committees' validators set,
+		// should we still keep this rule?
 		// Rule: Validator index must match with validatorPK or one of CommitteeID's validators
 		if !slices.Contains(validatorIndices, message.ValidatorIndex) {
 			e := ErrValidatorIndexMismatch
```
MatheusFranco99, line 125: @lior-blox @moshe-blox 
In `PartialSignatureMessage`, we have a `ValidatorIndex` field and there is the following rule for it:
> Validator index must match with validatorPK or one of CommitteeID's validators

But this assumes that the nodes are synced on the committees $\to$ []validators state. However, we have had discussions in which we said that we actually don't support that. Should I drop this rule?

message/validation/genesis/validation.go
```diff
@@ -488,6 +499,7 @@ func (mv *messageValidator) validateSSVMessage(msg *queue.DecodedSSVMessage, rec
 		case spectypes.DKGMsgType:
 			return nil, descriptor, ErrDKGMessage
 		}
+		// MSGVALIDATIONREVIEW: we should add here a "default:" with another error, no?
```
MatheusFranco99, line 502: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/partial_validation.go
```diff
@@ -104,6 +104,12 @@ func (mv *messageValidator) validatePartialSignatureMessageSemantics(
 	}
 
 	for _, message := range partialSignatureMessages.Messages {
+		// MSGVALIDATIONREVIEW: add signature size check as below:
```
MatheusFranco99, line 107: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/partial_validation.go
```diff
@@ -112,6 +118,8 @@ func (mv *messageValidator) validatePartialSignatureMessageSemantics(
 			return err
 		}
 
+		// MSGVALIDATIONREVIEW: since we don't expect operators to be synced on the committees' validators set,
```
MatheusFranco99, line 121: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/consensus_validation.go
```diff
@@ -45,6 +45,8 @@ func (mv *messageValidator) validateConsensusMessage(
 		return consensusDescriptor, msgSlot, e
 	}
 
+	// MSGVALIDATIONREVIEW: it seems that we are missing the "invalid identifier" check. I.e., the SSVMessage.Identifier should be equal to the qbftMessage.Identifier
```
MatheusFranco99, line 48: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/consensus_validation.go
```diff
@@ -131,6 +133,8 @@ func (mv *messageValidator) validateConsensusMessage(
 			signerState.Reset(msgRound)
 		}
 
+		// MSGVALIDATIONREVIEW: I think we should add a condition here that the msg is a proposal msg.
```
MatheusFranco99, line 136: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/consensus_validation.go
```diff
@@ -203,6 +207,7 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 ) error {
 	signerState := state.GetSignerState(signer)
 
+	// MSGVALIDATIONREVIEW: validateJustifications is also called in the ending of this function.
```
MatheusFranco99, line 210: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/message_counts.go
```diff
@@ -50,6 +50,11 @@ func (c *MessageCounts) ValidateConsensusMessage(msg *specqbft.SignedMessage, li
 			return err
 		}
 	case specqbft.CommitMsgType:
+		// MSGVALIDATIONREVIEW: we should use "else if len(msg.Signers) > 1" and then "else {panic()}" or "else {return Error...}" to cover
```
MatheusFranco99, line 53: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/message_counts.go
```diff
@@ -77,6 +82,7 @@ func (c *MessageCounts) ValidateConsensusMessage(msg *specqbft.SignedMessage, li
 	return nil
 }
 
+// MSGVALIDATIONREVIEW: below checks should use >= (and not >)
```
MatheusFranco99, line 85: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/genesis/partial_validation.go
```diff
@@ -33,6 +33,8 @@ func (mv *messageValidator) validatePartialSignatureMessage(
 		return msgSlot, err
 	}
 
+	// MSGVALIDATIONREVIEW: It seems to me that we are not doing here the "validateSlotTime" check (that the message isn't sent before the duty start time or after its final slot + 3 slots)
```
MatheusFranco99, line 36: Being fixed here: https://github.com/ssvlabs/ssv/pull/1488

message/validation/partial_validation.go
nkryuchkov: It has `ssz-size:"96"` assertion, so `ErrWrongBLSSignatureSize` will never be returned. However, I don't mind adding a comment

message/validation/partial_validation.go
nkryuchkov: I'm not sure I agree because when there are actions with a share, we update the state, so I don't understand why we don't expect this. Could you please add more details?

message/validation/genesis/consensus_validation.go
nkryuchkov: The idea was that if there's no state, then we don't need all state-related checks because it's the first message for the new state and all checks would pass, but we still need to check justifications

message/validation/partial_validation.go
MatheusFranco99: Oh true. Thanks! I'll drop the check and leave a comment

message/validation/partial_validation.go
MatheusFranco99: This may be for the following case. Let's say we have operators A and B. Operator A is part of Committee C, which just added a new validator V.

For our current state, we can't assume that operator B must be up to date with this information. So, whenever A sends a partial signature message on behalf of committee C that includes the index of validator V, B may not be aware of this (so it can't validate).

Even in the protocol (in-committee), when processing partial signature messages, we simply do "continue" in the iteration if we don't know the validator index

message/validation/genesis/consensus_validation.go
MatheusFranco99: Oh ok. Got it. Will update

message/validation/genesis/validation.go
```diff
@@ -449,6 +452,8 @@ func (mv *messageValidator) validateSSVMessage(msg *queue.DecodedSSVMessage, rec
 
 	descriptor.SSVMessageType = spectypes.MsgType(ssvMessage.MsgType)
 
+	// MSGVALIDATIONREVIEW: a nil nodeStorage prevents us form doing any validation. Shouldn't "nodeStorage != nil" be enforced?
```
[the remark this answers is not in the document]
MatheusFranco99: Got it. Thanks
Removing the comment

message/validation/genesis/consensus_validation.go
MatheusFranco99: Yes, I agree. I've added a comment in the fix PR

PR #1486, opened 2024-07-18
feat: (Alan) message validation forking

y0sher:
### Description 

- Execute validation on the appropriate message validation based on the current epoch.
- Ignore malformed messages on first slot of the epoch


message/validation/validation.go
```diff
@@ -96,6 +97,12 @@ func (mv *messageValidator) Validate(_ context.Context, peerID peer.ID, pmsg *pu
 
 	decodedMessage, err := mv.handlePubsubMessage(pmsg, time.Now())
 	if err != nil {
+		if mv.netCfg.Beacon.EstimatedCurrentSlot() == mv.netCfg.Beacon.FirstSlotAtEpoch(mv.netCfg.AlanForkEpoch) {
+			fullErr, ok := err.(Error)
```
nkryuchkov, line 101: This should work, but if we do a refactor that wraps ErrMalformedPubSubMessage into another error (we'll unlikely do it though), the check will fail and this will be complex to find out. I'd consider using `errors.As` instead

message/validation/validation.go
```diff
@@ -96,6 +97,12 @@ func (mv *messageValidator) Validate(_ context.Context, peerID peer.ID, pmsg *pu
...
+			if ok && errors.Is(fullErr.innerErr, ErrMalformedPubSubMessage) {
```
nkryuchkov, line 102: `innerErr` never contains ErrMalformedPubSubMessage, it has an underlying error. We need to check if `fullErr.text == ErrMalformedPubSubMessage.text`. I wouldn't say I like that we cannot use a single `errors.Is/As` check but this is how errors are designed now. Perhaps, we could consider a better solution in the future.

nkryuchkov, line 102:
```suggestion
			var fullErr Error
			if errors.As(err, &valErr) && fullErr.text == ErrMalformedPubSubMessage.text {
```

message/validation/validation.go
```diff
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"github.com/pkg/errors"
```
nkryuchkov, line 10: It's been discontinued for years, let's avoid using it for new code

cli/operator/node.go
```diff
@@ -92,6 +94,26 @@ var globalArgs global_config.Args
 
 var operatorNode operator.Node
 
+type ForkingMessageValidation struct {
```
nkryuchkov, line 97: I wouldn't probably leave this comment for another file, but I'd extract this into another file (even in the same package is fine) because `node.go` is already quite big and complex to navigate, although I understand this is temporary

message/validation/validation.go
```diff
@@ -96,6 +97,12 @@ func (mv *messageValidator) Validate(_ context.Context, peerID peer.ID, pmsg *pu
 
 	decodedMessage, err := mv.handlePubsubMessage(pmsg, time.Now())
 	if err != nil {
+		if mv.netCfg.Beacon.EstimatedCurrentSlot() == mv.netCfg.Beacon.FirstSlotAtEpoch(mv.netCfg.AlanForkEpoch) {
```
nkryuchkov, line 100: It might be easier to understand and ensure the absence of bugs if we do this in `decodeSignedSSVMessage`. We could do something like this:

```go
func (mv *messageValidator) decodeSignedSSVMessage(pMsg *pubsub.Message) (*spectypes.SignedSSVMessage, error) {
	// Rule: Pubsub.Message.Message.Data decoding
	signedSSVMessage := &spectypes.SignedSSVMessage{}
	if err := signedSSVMessage.Decode(pMsg.GetData()); err != nil {
		e := ErrMalformedPubSubMessage
		e.innerErr = err
		
		if mv.netCfg.Beacon.EstimatedCurrentSlot() == mv.netCfg.Beacon.FirstSlotAtEpoch(mv.netCfg.AlanForkEpoch) {
			e.reject = false
		}
				
		return nil, e
	}

	return signedSSVMessage, nil
}
```

nkryuchkov (changes requested):

message/validation/validation.go
y0sher: you meant `fullErr` and not `valErr` right?

nkryuchkov: Yes, sorry

nkryuchkov:
superseded by https://github.com/ssvlabs/ssv/pull/1497 and https://github.com/ssvlabs/ssv/pull/1489

nkryuchkov (approved):

nkryuchkov:
@y0sher please review my changes, I can't request your review because you opened the PR

PR #1487, opened 2024-07-18
(alan/no-fork) Committee runner: Data race fix

AKorpusenko:
## Data race problem

When committee tests were added to ssv we didn't note that the `ProcessMessage` is executed right after `StartDuty`. This is totally fine for spec, but in out implementation this causes the data race:
- We run `go ConsumeQueue()` in `StartDuty` which reads the base runner state. 
- After calling the `StartDuty`, the is called `ProcessMessage` for the same `Committee` in main goroutine
- Result => 2 goroutines share the same state => data race 

## Fix description

This happens only in tests, because spec doesn't have the queue running inside `StartDuty`. 
So let's do the same thing. Remove `go ConsumeQueue` from the `StartDuty` and move it to another func `StartConsumeQueue` to `Committee` and run it from `DutyExecutor`

### DIFF
--- a/protocol/v2/ssv/validator/committee.go
+++ b/protocol/v2/ssv/validator/committee.go
@@ -236,8 +236,8 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.DecodedSSVMess
 			return errors.Wrap(err, "invalid qbft Message")
 		}
 		c.mtx.Lock()
-		defer c.mtx.Unlock() // read c.Runners map + ProcessConsensus r/w state
 		runner, exists := c.Runners[phase0.Slot(qbftMsg.Height)]
+		c.mtx.Unlock()
 		if !exists {
 			return errors.New("no runner found for message's slot")
 		}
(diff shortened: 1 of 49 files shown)


protocol/v2/ssv/validator/committee.go
```diff
@@ -236,8 +236,8 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.DecodedSSVMess
 			return errors.Wrap(err, "invalid qbft Message")
 		}
 		c.mtx.Lock()
+		defer c.mtx.Unlock() // read c.Runners map + ProcessConsensus r/w state
```
y0sher, line 239: this is not smart, locking all consumers from getting a runner for the whole time where a runner is processing a consensus message. if the same runner can't be accesses in parallel (although they shouldn't be accessed in parallel as ConsumeQueue will pop 1 message at a time and process it.)
We might need to have a lock per runner.

Can you explain the data race further so we know whats the best next step?

y0sher (changes requested):
we need to get deeper understanding of what state needs to be protected here. I think just increasing the mutex scope to include the runner itself is wrong in this case.

committee.mtx should protect the map of the runners and the queue only.

AKorpusenko:
@y0sher, I found a better way to do this. Please check

protocol/v2/ssv/validator/committee.go
AKorpusenko: Removed this. Found better way to prevent data race

y0sher (approved):
Great!

PR #1488, opened 2024-07-19
Fix message validation (review)

MatheusFranco99:
# Overview

This PR applies the suggestions of the message validation review: https://github.com/ssvlabs/ssv/pull/1482

#### Alan
- Update the "validator index match ValidatorPK or CommitteeID" rule to be checked only for validator duties.

#### Genesis
- Add the `ProposalMsgType` check before updating `ProposalData` in the state.
- Add slot time and beacon duty checks in PartialSignatureMessages validation.
- Fix > to >= in `MessageCount`.
- Add assertion and `ErrNoSigners` error in `MessageCount` for completeness.
- Add default check for a wrong message type in `ValidateSSVMessage`.

### DIFF
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -82,6 +82,7 @@ var (
 	ErrUnexpectedConsensusMessage              = Error{text: "unexpected consensus message for this role", reject: true}
 	ErrNoSigners                               = Error{text: "no signers", reject: true}
 	ErrWrongRSASignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
+	ErrWrongBLSSignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
 	ErrZeroSigner                              = Error{text: "zero signer ID", reject: true}
 	ErrSignerNotInCommittee                    = Error{text: "signer is not in committee", reject: true}
 	ErrDuplicatedSigner                        = Error{text: "signer is duplicated", reject: true}
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -99,7 +99,7 @@ var (
 	ErrPartialSignatureTypeRoleMismatch        = Error{text: "partial signature type and role don't match", reject: true}
 	ErrNonDecidedWithMultipleSigners           = Error{text: "non-decided with multiple signers", reject: true}
 	ErrDecidedNotEnoughSigners                 = Error{text: "not enough signers in decided message", reject: true}
-	ErrDuplicatedProposalWithDifferentData     = Error{text: "duplicated proposal with different data", reject: true}
+	ErrDifferentProposalData                   = Error{text: "different proposal data", reject: true}
 	ErrMalformedPrepareJustifications          = Error{text: "malformed prepare justifications", reject: true}
 	ErrUnexpectedPrepareJustifications         = Error{text: "prepare justifications unexpected for this message type", reject: true}
 	ErrMalformedRoundChangeJustifications      = Error{text: "malformed round change justifications", reject: true}
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -104,7 +104,6 @@ var (
 	ErrUnexpectedPrepareJustifications         = Error{text: "prepare justifications unexpected for this message type", reject: true}
 	ErrMalformedRoundChangeJustifications      = Error{text: "malformed round change justifications", reject: true}
 	ErrUnexpectedRoundChangeJustifications     = Error{text: "round change justifications unexpected for this message type", reject: true}
-	ErrDeserializePublicKey                    = Error{text: "deserialize public key", reject: true}
 	ErrNoPartialSignatureMessages              = Error{text: "no partial signature messages", reject: true}
 	ErrNoValidators                            = Error{text: "no validators for this committee ID", reject: true}
 	ErrNoSignatures                            = Error{text: "no signatures", reject: true}
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -82,7 +82,6 @@ var (
 	ErrUnexpectedConsensusMessage              = Error{text: "unexpected consensus message for this role", reject: true}
 	ErrNoSigners                               = Error{text: "no signers", reject: true}
 	ErrWrongRSASignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
-	ErrWrongBLSSignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
 	ErrZeroSigner                              = Error{text: "zero signer ID", reject: true}
 	ErrSignerNotInCommittee                    = Error{text: "signer is not in committee", reject: true}
 	ErrDuplicatedSigner                        = Error{text: "signer is duplicated", reject: true}
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -248,7 +248,7 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 
 	if msgSlot == signerState.Slot && msgRound == signerState.Round {
 		if mv.hasFullData(signedMsg) && signerState.ProposalData != nil && !bytes.Equal(signerState.ProposalData, signedMsg.FullData) {
-			return ErrDuplicatedProposalWithDifferentData
+			return ErrDifferentProposalData
 		}
 
 		limits := maxMessageCounts(len(share.Committee))
--- a/message/validation/genesis/consensus_validation.go
+++ b/message/validation/genesis/consensus_validation.go
@@ -4,7 +4,6 @@ package validation
 
 import (
 	"bytes"
-	"encoding/hex"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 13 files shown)


message/validation/errors.go
```diff
@@ -82,6 +82,7 @@ var (
 	ErrUnexpectedConsensusMessage              = Error{text: "unexpected consensus message for this role", reject: true}
 	ErrNoSigners                               = Error{text: "no signers", reject: true}
 	ErrWrongRSASignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
+	ErrWrongBLSSignatureSize                   = Error{text: "wrong RSA signature size", reject: true}
```
nkryuchkov, line 85: typo
```suggestion
	ErrWrongBLSSignatureSize                   = Error{text: "wrong BLS signature size", reject: true}
```

nkryuchkov: New message validation has 100% test coverage, could you please add a test for this one?

But since it has ssz size assertion, I'm not sure if it's possible, so probably we'll need to drop this check and replace it with a comment

message/validation/genesis/consensus_validation.go
```diff
@@ -131,7 +140,7 @@ func (mv *messageValidator) validateConsensusMessage(
 			signerState.Reset(msgRound)
 		}
 
-		if mv.hasFullData(signedMsg) && signerState.ProposalData == nil {
+		if mv.hasFullData(signedMsg) && signerState.ProposalData == nil && signedMsg.Message.MsgType == genesisspecqbft.ProposalMsgType {
```
nkryuchkov, line 143: I'd check `signedMsg.Message.MsgType == genesisspecqbft.ProposalMsgType` first for readability

message/validation/genesis/consensus_validation.go
```diff
@@ -203,10 +212,6 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 ) error {
 	signerState := state.GetSignerState(signer)
 
-	if signerState == nil {
```
nkryuchkov, line 206: https://github.com/ssvlabs/ssv/pull/1482#discussion_r1687192683

message/validation/genesis/message_counts.go
```diff
@@ -56,13 +56,14 @@ func (c *MessageCounts) ValidateConsensusMessage(msg *specqbft.SignedMessage, li
...
-		}
-		if len(msg.Signers) > 1 {
+		} else if len(msg.Signers) > 1 {
 			if c.Decided >= limits.Decided {
 				err := ErrTooManySameTypeMessagesPerRound
 				err.got = fmt.Sprintf("decided, having %v", c.String())
 				return err
 			}
+		} else {
+			return ErrNoSigners
```
nkryuchkov, line 66: If we do multiple checks of the same things, perhaps it will be complex to follow. We already check this in `validConsensusSigners`

nkryuchkov (commented):
If we touch the genesis message validation, we also need to add the fixes from open PRs  (I'm not sure how many of them we have)

message/validation/genesis/message_counts.go
MatheusFranco99: I guess having the assumption is fine. But I also think it's good to make these assertions, even when they should be impossible due to previous logic. What do you guys think @y0sher @moshe-blox ?

message/validation/errors.go
MatheusFranco99: Yup, replacing it with a comment. Thanks!

message/validation/genesis/consensus_validation.go
MatheusFranco99: Updated. Thanks!

message/validation/genesis/consensus_validation.go
MatheusFranco99: Agreed. Changed.

message/validation/genesis/message_counts.go
nkryuchkov: If we decide to make them, I think we'll need to make them consistent and add to all methods

message/validation/genesis/consensus_validation.go
```diff
@@ -203,6 +205,8 @@ func (mv *messageValidator) validateSignerBehaviorConsensus(
 ) error {
 	signerState := state.GetSignerState(signer)
 
+	// If signer state is nil, this is the first message for the signer and
+	// the next rules can't be checked.
```
nkryuchkov, line 209: I'd reword this because those rules (that check rules against state) can be checked but there's no reason for that because the state is empty

MatheusFranco99: Oh okok. Thanks!!

MatheusFranco99: Fixed

nkryuchkov:
If we want to finetune genesis message validation, I think we should also cherry-pick open PRs for it:
- https://github.com/ssvlabs/ssv/pull/1268
- https://github.com/ssvlabs/ssv/pull/1238

nkryuchkov (approved):

MatheusFranco99:
@nkryuchkov I agree. We should also merge these PRs

message/validation/genesis/message_counts.go
y0sher: I agree with @nkryuchkov here. either we set some basic assertions we do all the time, or we count on previous code to have done this assertions.
I think we should leave it only in the most proper place and go forward assuming it was already checked. (as long as we know all messages go through this flow. )

message/validation/genesis/consensus_validation.go
```diff
@@ -131,8 +131,10 @@ func (mv *messageValidator) validateConsensusMessage(
 			signerState.Reset(msgRound)
 		}
 
-		if mv.hasFullData(signedMsg) && signerState.ProposalData == nil {
-			signerState.ProposalData = signedMsg.FullData
+		if signedMsg.Message.MsgType == genesisspecqbft.ProposalMsgType {
```
y0sher, line 134: what about round change?

message/validation/genesis/message_counts.go
y0sher: I think we can return an error but we need to emphasize this is a not expected "bug" error and not a common error that can happen with a valid message. maybe even panic or fatal here. 
we should comment about this case and use specialized error.

message/validation/validation.go
```diff
@@ -232,6 +232,8 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
 		}, nil
 	}
 
+	// NOTE: do we have to deserialize the public key here? we already check for existence in the validator store.
```
y0sher, line 235: @MatheusFranco99

y0sher (changes requested):
please address my comments

message/validation/genesis/message_counts.go
nkryuchkov: I agree, but I'd prefer fatal

y0sher:
> If we want to finetune genesis message validation, I think we should also cherry-pick open PRs for it:
> 
>     * [Don't check round change messages for duplicated full data #1268](https://github.com/ssvlabs/ssv/pull/1268)
> 
>     * [Message validation ConsensusState eviction #1238](https://github.com/ssvlabs/ssv/pull/1238)

where is this included?

nkryuchkov:
> > If we want to finetune genesis message validation, I think we should also cherry-pick open PRs for it:
> > ```
> > * [Don't check round change messages for duplicated full data #1268](https://github.com/ssvlabs/ssv/pull/1268)
> > 
> > * [Message validation ConsensusState eviction #1238](https://github.com/ssvlabs/ssv/pull/1238)
> > ```
> 
> where is this included?

@y0sher what do you mean?

message/validation/genesis/message_counts.go
nkryuchkov: Since I'm taking over this PR, I'm going to delete this because in my opinion if we add assertion checks, this is not the only place to do it, so we could do it in a new PR

nkryuchkov:
> > > > If we want to finetune genesis message validation, I think we should also cherry-pick open PRs for it:
> > > > ```
> > > > * [Don't check round change messages for duplicated full data #1268](https://github.com/ssvlabs/ssv/pull/1268)
> > > > 
> > > > * [Message validation ConsensusState eviction #1238](https://github.com/ssvlabs/ssv/pull/1238)
> > > > ```
> > > 
> > > 
> > > where is this included?
> > 
> > 
> > @y0sher what do you mean?
> 
> @nkryuchkov you and @MatheusFranco99 mentioned you want to also include these PRs. I was asking what did you end up doing with it? should include in this PR? other PR?

Let's do it in a new PR to avoid increasing scope of this PR while tests are broken

message/validation/genesis/consensus_validation.go
nkryuchkov: @GalRogozinski could you please confirm if it's correct?

GalRogozinski: What you said is correct

If you look at https://github.com/ssvlabs/ssv-spec/blob/ccf408d1ecd87a4ab631885409d679f10c3fd080/qbft/round_change.go#L429 you should see that we add fullData to RC messages. The reason is because that if a node didn't see this data in the previous round, then on the next round it is the leader and is obliged to propose the same data, then it can get it from the RC message.

But this data must be justified in order to accept it as correct

GalRogozinski: So if you have a quorum of valid RC messages that carry a justification of Prepares (quorum) for this data, then you know that the full data they carry must be the correct proposal data even if your node missed it

nkryuchkov: @GalRogozinski thanks, so it means that this particular check should be done for both `genesisspecqbft.ProposalMsgType` and `genesisspecqbft.RoundChangeMsgType`?

GalRogozinski: I see that on lines 210-212 you check for justifications if the signerState is nil...
so not sure
give me a call

GalRogozinski: Per our conversation... the if check was added to avoid an attack where any node can send an RC message that changes the `mv` state.

The code is fine the way it is at the moment

moshe-blox (approved):

PR #1489, opened 2024-07-20
(alan) Added forking mechanism for msg validator 

AKorpusenko:


y0sher:
https://github.com/ssvlabs/ssv/pull/1486

cli/operator/node.go
```diff
@@ -219,26 +219,21 @@ var StartNodeCmd = &cobra.Command{
...
-		} else {
-			messageValidator = genesisvalidation.New(
+		messageValidator := validation.New(
+			networkConfig,
+			validatorStore,
+			dutyStore,
+			signatureVerifier,
+			validation.WithLogger(logger),
+			validation.WithMetrics(metricsReporter),
+			validation.WithGenesisValidator(genesisvalidation.New(
```
nkryuchkov, line 229: As a small optimization, we can check if the fork happened and don't even create the genesis message validator in this case

networkconfig/config.go
```diff
@@ -77,6 +77,5 @@ func (n NetworkConfig) GetGenesisTime() time.Time {
 }
 
 func (n NetworkConfig) AlanFork() bool {
-	// TODO: implement
-	return true
+	return n.Beacon.EstimatedCurrentEpoch() >= n.AlanForkEpoch
 }
```
y0sher, line 81: we should have this or the other not both.

y0sher:
I honestly like [this](https://github.com/ssvlabs/ssv/pull/1486) approach better because it doesn't change the new message validation code.
It's really a matter of preference I guess but maybe @nkryuchkov, @moshe-blox and @MatusKysel  can chime to finalize this.

networkconfig/config.go
AKorpusenko: Agree. I was not sure does it matter to provide slot as param or not for `AlanForked(slot)`

cli/operator/node.go
AKorpusenko: Agree. Maybe it even worths to do it inside the `WithGenesisValidator` option.

nkryuchkov:
@y0sher I agree with both approaches but since this PR is closed I finished your PR https://github.com/ssvlabs/ssv/pull/1486

PR #1491, opened 2024-07-21
review: alan/no-fork review comments

y0sher:


message/validation/common_checks.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	// Rule: For a sync committee aggregation duty message, we check if the validator is assigned to it
+	if role == spectypes.RoleSyncCommitteeContribution {
+		period := mv.netCfg.Beacon.EstimatedSyncCommitteePeriodAtEpoch(mv.netCfg.Beacon.EstimatedEpochAtSlot(slot))
+		if mv.dutyStore.SyncCommittee.Duty(period, indices[0]) == nil {
+			return ErrNoDuty
+		}
+	}
```
y0sher, line 115: @MatheusFranco99 @nkryuchkov 

Why we use `indices[0]` here? how do we know this is the correct validator index from the whole committee validator list?

y0sher: this should probably be checked with either all the indices or the right index that really has duty.

MatheusFranco99: In this function, I think it's assumed that, if the role is either Proposer or Sync Committee Contribution, the indices list will only have one element. Right @nkryuchkov ?

MatheusFranco99: It's checked somewhere else but we can add here an extra check to see that the length of the list is 1, for independence, if you like

registry/storage/validatorstore.go
```diff
@@ -0,0 +1,366 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Update byValidatorIndex
+	if share.HasBeaconMetadata() {
+		c.byValidatorIndex[share.BeaconMetadata.Index] = share
+	}
+
+	// Update byCommitteeID
+	for _, committee := range c.byCommitteeID {
```
y0sher, line 309: like you did in shareAdded -
```suggestion
	for _, committee := range c.byCommitteeID {
	if committee.ID != share.CommitteeID() {
	  continue
	}
```

registry/storage/validatorstore.go
```diff
@@ -0,0 +1,366 @@
...
+				break
+			}
+		}
+
+		for i, index := range committee.Indices {
+			if index == share.ValidatorIndex {
+				committee.Indices[i] = share.ValidatorIndex
+				break
+			}
+		}
```
y0sher, line 322: @nkryuchkov this doesn't seem right. how can this work? 
if the index is already the correct index from the share why we even need to run this code?
if the index isn't correct - on fresh sync before the metadata update, the index will be 0. then this update code just won't work.

Probably best to set this above, where you found the validator pubkey, and use the same slice index to update the validator index.

registry/storage/validatorstore.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/1495

registry/storage/validatorstore.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/1495

message/validation/common_checks.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/1496

y0sher: its not what I meant, I didn't see the part where we are passing the 1 len list. I was now able to track it down to the code at `getCommitteeAndValidatorIndices` where we populate the list with only one validator when the message belongs to a propose duty.

PR #1493, opened 2024-07-22
Custom DomainType and Trusted Peers

moshe-blox:


network/p2p/p2p.go
```diff
@@ -164,9 +167,53 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
 		return nil
 	}
 
-	logger.Info("starting")
+	connector := make(chan peer.AddrInfo, connectorQueueSize)
+	go func() {
+		// Wait for own subnets to be subscribed to and updated.
+		// TODO: wait more intelligently with a channel.
+		time.Sleep(8 * time.Second)
```
y0sher, line 174: what if we are freshly syncing? this might take much longer

network/p2p/p2p.go
```diff
@@ -164,9 +167,53 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+		ID:    n.host.ID(),
+		Addrs: n.host.Addrs(),
+	})
+	if err != nil {
+		logger.Fatal("could not get my address", zap.Error(err))
+	}
+	maStrs := make([]string, len(ma))
+	for i, ma := range ma {
+		maStrs[i] = ma.String()
+	}
```
y0sher, line 210: too much mas 🤣 
```suggestion
	for i, ima := range ma {
		maStrs[i] = ima.String()
	}
```

network/p2p/p2p.go
```diff
@@ -164,9 +167,53 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+					return fmt.Errorf("could not parse trusted peer: %w", err)
+				}
+				trustedPeers[addrInfo.ID] = append(trustedPeers[addrInfo.ID], addrInfo.Addrs...)
+			}
+		}
+		go func() {
+			for id, addrs := range trustedPeers {
+				connector <- peer.AddrInfo{ID: id, Addrs: addrs}
+			}
+		}()
```
y0sher, line 197: this is like extra safety here to do go routine on a buffered channel..

registry/storage/shares.go
```diff
@@ -82,6 +82,7 @@ func (s *sharesStorage) load() error {
 		if err := val.Decode(obj.Value); err != nil {
 			return fmt.Errorf("failed to deserialize share: %w", err)
 		}
+		val.DomainType = types.GetDefaultDomain()
```
y0sher, line 85: isn't this going to be different that the custom domain?

network/p2p/p2p.go
moshe-blox: @y0sher  IIRC we should be starting p2p after we've finished syncing history

network/p2p/p2p.go
moshe-blox: @y0sher i think the code should not assume that the buffer size is higher than the number of trusted peers

y0sher (dismissed):

cli/operator/node.go
```diff
@@ -539,7 +541,17 @@ func setupSSVNetwork(logger *zap.Logger) (networkconfig.NetworkConfig, error) {
 	if err != nil {
 		return networkconfig.NetworkConfig{}, err
 	}
-
+	if strings.HasPrefix(cfg.SSVOptions.CustomDomainType, "0x") {
```
y0sher, line 544: do we want to consider a case where its not starting with "0x"?

network/p2p/p2p.go
y0sher: but subscribing to subnet might happen later after p2p already started. for example when downloading metadata

cli/operator/node.go
moshe-blox: @y0sher you're right we should err in this case rather than silently ignore

network/p2p/p2p.go
moshe-blox: on second thought, we could simply set the buffer size to `min(256, len(trustedPeers))`, but not sure that'd be worth another round of testing 😅

y0sher (approved):

PR #1494, opened 2024-07-22
(alan/no-fork) Minor spec tests fixes: Removed skip blinded + wrapped qbft timeout tests

AKorpusenko:


protocol/v2/ssv/spectest/ssv_mapping_test.go
```diff
@@ -78,11 +78,6 @@ func prepareTest(t *testing.T, logger *zap.Logger, name string, test interface{}
 
 	switch testType {
 	case reflect.TypeOf(&tests.MsgProcessingSpecTest{}).String():
-		// TODO: fix blinded test
-		if strings.Contains(testName, "propose regular decide blinded") || strings.Contains(testName, "propose blinded decide regular") {
-			logger.Info("skipping blinded block test", zap.String("test", testName))
-			return nil
-		}
```
GalRogozinski, line 85: nice, works out of the box :-)

protocol/v2/qbft/spectest/qbft_mapping_test.go
```diff
@@ -108,8 +108,11 @@ func TestQBFTMapping(t *testing.T) {
 			err = pre.Decode(preByts)
 			require.NoError(t, err)
 			typedTest.Pre = pre
+			t.Run(typedTest.Name, func(t *testing.T) { // using only spec struct so no need to run our version (TODO: check how we choose leader)
+				t.Parallel()
+				RunTimeout(t, typedTest)
+			})
```
GalRogozinski, line 114: I know `t.parallel()` pattern is used everywhere. Any idea why we are not simply doing it once on the first line of QBFT Mapping? If there is no reason maybe we can make the code better if @MatusKysel agrees

I don't understand the TODO... leader is chosen in round robin

AKorpusenko: @GalRogozinski TODOs  are here before my changes. I just left this as it was

AKorpusenko: `t.parallel()` -- should be called inside every `t.Run()` calling it once outside the for loop makes the tests run slower. Same for `ssv_mapping_test.go`.  Tests this 

About the TODOs. I left it as it was before

GalRogozinski (approved):

PR #1495, opened 2024-07-22
validatorStore: fix bug with indices on share update

nkryuchkov:
as per title

### DIFF
--- a/registry/storage/validatorstore.go
+++ b/registry/storage/validatorstore.go
@@ -307,6 +307,9 @@ func (c *validatorStore) handleShareUpdated(share *types.SSVShare) {
 
 	// Update byCommitteeID
 	for _, committee := range c.byCommitteeID {
+		if committee.ID != share.CommitteeID() {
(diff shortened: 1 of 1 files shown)


registry/storage/validatorstore.go
```diff
@@ -310,12 +310,6 @@ func (c *validatorStore) handleShareUpdated(share *types.SSVShare) {
 		for i, validator := range committee.Validators {
```
y0sher, line 310:
```suggestion
                if committee.ID != share.CommitteeID() {
                    continue
                }
		for i, validator := range committee.Validators {
```

y0sher (changes requested):
one change to make it a bit a bit more efficient

registry/storage/validatorstore.go
nkryuchkov: well spotted, thanks

y0sher (approved):

PR #1496, opened 2024-07-22
message validation: add comments when getting first validator index

nkryuchkov:
Improve clarity of the check by adding a comment

### DIFF
--- a/message/validation/common_checks.go
+++ b/message/validation/common_checks.go
@@ -98,10 +98,16 @@ func (mv *messageValidator) validateBeaconDuty(
 	slot phase0.Slot,
 	indices []phase0.ValidatorIndex,
 ) error {
+	if len(indices) != 1 {
+		return ErrProposalSCCSeveralIndices
+	}
+
+	index := indices[0]
+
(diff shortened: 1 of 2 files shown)


nkryuchkov:
@MatheusFranco99 Should we add `ErrProposalSCCSeveralIndices` to the knowledge base?

message/validation/common_checks.go
```diff
@@ -98,18 +98,24 @@ func (mv *messageValidator) validateBeaconDuty(
 	slot phase0.Slot,
 	indices []phase0.ValidatorIndex,
 ) error {
+	if len(indices) != 1 {
+		return ErrProposalSCCSeveralIndices
+	}
+
+	index := indices[0]
```
y0sher, line 105: from what I've seen in the code, we are getting the indicies from the committeeInfo struct which is created by getting all the committee indices from the validatorStore, I don't userstand how can you be sure that the first index is the correct validator that has the proposal duty?

y0sher:
not needed.

nkryuchkov:
@y0sher Let's add a comment instead

y0sher (approved):

PR #1500, opened 2024-07-23
Improve p2p scoring log

MatheusFranco99:
# Overview

This PR improves the p2p scoring log. Namely, it adds the explicit counters for p1, p2, p4, p6 and p7.

In the `network/topics/pubsub.go:NewPubSub` function, the order of some initialization calls was changed so that `scoreInspector` could access topics' parameters and overall score parameters.

### DIFF
--- a/network/topics/scoring.go
+++ b/network/topics/scoring.go
@@ -1,7 +1,6 @@
 package topics
 
 import (
-	"math"
 	"time"
 
 	"github.com/ssvlabs/ssv/logging/fields"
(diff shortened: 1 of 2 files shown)


network/topics/scoring.go
```diff
@@ -26,52 +25,128 @@ func DefaultScoringConfig() *ScoringConfig {
...
+		// Use a "scope cache" for getting a topic's score parameters
+		// otherwise, the factory method would be called multiple times for the same topic
+		topicScoreParamsCache := make(map[string]*pubsub.TopicScoreParams)
+		getScoreParamsForTopic := func(topic string) *pubsub.TopicScoreParams {
+			if topicParams, exists := topicScoreParamsCache[topic]; exists {
+				return topicParams
+			}
+			topicScoreParamsCache[topic] = topicScoreParamsFactory(topic)
+			return topicScoreParamsCache[topic]
+		}
```
y0sher, line 51: can score params for a certain topic change? if so the cache will not reflect that. I know for fact they change in fork so we can include special code for that like in #1477 , if there aren't any other cases then we can resolve this.

MatheusFranco99: The p2 parameters may change upon validators' addition/removal.

This cache has only the scope of one execution of the function. So, for the next execution, there will be a new cache and it will use the updated parameters.

I added this cache since we would generate the same topic parameters repeatedly for each peer and waste resource (I mean that in a single function execution).

Does the code seem aligned with my purpose?

PR #1501, opened 2024-07-23
fix: (alan) don't submit empty attestation or sc messages

y0sher:
### Description 

We saw a few cases where a response from the beacon node indicated an empty message sent.

### Fix

Don't submit if there is no data to send.

### DIFF
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -419,8 +419,6 @@ func (cr *CommitteeRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *t
 		}
 	}
 
-	var submissionStart time.Time
-
 	logger = logger.With(durationFields...)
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/runner/committee.go
```diff
@@ -418,44 +418,52 @@ func (cr *CommitteeRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *t
 			}
 		}
 	}
+
+	var submissionStart time.Time
```
nkryuchkov, line 422: Can we do `submissionStart := time.Now()` in both `if`s and delete this line?

nkryuchkov (approved):

protocol/v2/ssv/runner/committee.go
y0sher: good idea, done.

nkryuchkov (approved):

PR #1509, opened 2024-07-25
Spec alignment review

MatheusFranco99:
# Overview

This PR adds suggestions from the spec alignment review. More than the minor suggested changes, I posted below some questions on major differences.

## Differences

1. In the runners (`AggregatorRunner`, `CommitteeRunner`, ...), an `OperatorSigner` interface is used which is also defined in ssv. In spec, we changed it to a structure.
2. In `BaseRunner`, there is a `DomainTypeProvider` attribute that doesn't exist in spec. In spec, we get the `DomainType` from `Share`.

Note: The `VerifySignature` method from `Config` (of QBFT) was dropped since it was unaligned to spec and not being used. Plus, it's a dangerous variable.


protocol/v2/qbft/controller/controller.go
```diff
@@ -151,6 +151,7 @@ func (c *Controller) UponExistingInstanceMsg(logger *zap.Logger, msg *specqbft.P
 		logger.Debug("❌ failed to broadcast decided message", zap.Error(err))
 	}
 
+	// TODO: spec checks if prevDecided (and returns if so) right after inst.ProcessMsg. Should we align?
```
y0sher, line 154: can you explain further? we also check here for prevDecided.. whats the diff?

y0sher: @nkryuchkov is it on purpose on just not aligned?

protocol/v2/ssv/runner/committee.go
```diff
@@ -539,12 +539,14 @@ func findValidators(
 
 // Unneeded since no preconsensus phase
 func (cr CommitteeRunner) expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error) {
+	// TODO: spec uses a panic here since this function should never be used. Should we align?
```
y0sher, line 542: panic inside our code isn't a good practice. we could use logger.Fatal. anyway not worried about this as like you said, should never be used.

protocol/v2/qbft/controller/controller.go
nkryuchkov: I only applied what was changed since the previous alignment: https://github.com/ssvlabs/ssv-spec/compare/tmp-160724-spec-align...alan?expand=1

I didn't touch the rest because I don't know if it's intentional or not.

Regarding this particular change, it looks to me like we need to align it but carefully test it afterwards.

protocol/v2/ssv/runner/committee.go
MatheusFranco99: Agree

protocol/v2/qbft/controller/controller.go
moshe-blox: the reason the implementation is returning later here is most likely so it can broadcast multiple decideds (and not just the first one at quorum), in order for the exporter to more reliably tell whether all operators signed or not.

since the exporter doesn't longer use decided messages to determine participation, then this misalignment should no longer be needed.

moshe-blox:
1. This difference is intended because the implementation has two different implementations for signing with RSA (one for Linux and one for the rest)

protocol/v2/qbft/controller/controller.go
y0sher: > the reason the implementation is returning later here is most likely so it can broadcast multiple decideds (and not just the first one at quorum), in order for the exporter to more reliably tell whether all operators signed or not.
> 
> since the exporter doesn't longer use decided messages to determine participation, then this misalignment should no longer be needed.
> 
> we can align this to spec imo

@nkryuchkov

nkryuchkov:
> 1. This difference is intended because the implementation has two different implementations for signing with RSA (one for Linux and one for the rest)

Also because the implementation has `GetOperatorIdF` function instead of `OperatorID` field because it doesn't know operator ID on start, therefore the implementation cannot use the structure from spec

nkryuchkov:
> 3. `Config` for QBFT has a `VerifySignature` method that returns a boolean. This doesn't exist in spec. Why does this method exist?

It was used to disable signature checks in message validation IIRC, but if it's always false now, let's remove it

protocol/v2/qbft/controller/controller.go
nkryuchkov: done

y0sher: @moshe-blox fixed it

protocol/v2/qbft/instance/instance.go
```diff
@@ -113,6 +113,10 @@ func (i *Instance) Broadcast(logger *zap.Logger, msg *spectypes.SignedSSVMessage
 	if !i.CanProcessMessages() {
 		return errors.New("instance stopped processing messages")
 	}
+	// TODO: In spec, the below assignment doesn't exist.
+	// In fact, the msg should already have the correct identifier. Why are we setting it again?
+	// Also, even though it doesn't change anything, we shouldn't change the message's content
```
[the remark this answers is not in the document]
MatheusFranco99: Oh true. I didn't realise it. Yup, at least for the scope of this PR, in spec we also use msgID in Broadcast, but from the msg itself.

Regarding changing the Broadcast function, we would also need to do a change in spec. Maybe it's not the scope of this PR but we can add it later on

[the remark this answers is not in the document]
MatheusFranco99: @y0sher Can I align this to spec, to use the ID from the msg?

protocol/v2/ssv/runner/runner.go
```diff
@@ -179,6 +179,7 @@ func (b *BaseRunner) baseConsensusMsgProcessing(logger *zap.Logger, runner Runne
 	}
 
 	decidedMsg, err := b.QBFTController.ProcessMsg(logger, msg)
+	// TODO: Should we remove this function since it doesn't do anything?
```
MatheusFranco99, line 182: @y0sher Can I remove this `compactInstanceIfNeeded` function?

nkryuchkov: @MatheusFranco99 let's remove it as it's empty right now

MatheusFranco99: Removed

protocol/v2/qbft/instance/instance.go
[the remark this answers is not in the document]
MatheusFranco99: For now, I'm aligning it. Afterwards, if we find it necessary, we can go back to copying the ID from the state.

nkryuchkov (approved):

protocol/v2/ssv/runner/committee.go
moshe-blox: @MatheusFranco99 why isnt just returning an error enough? tests should fail if they reach here, unless they dont check error for some reason which is an error in the test already? 🤔

MatheusFranco99: Yeah, I agree

protocol/v2/qbft/instance/instance.go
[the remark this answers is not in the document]
y0sher: there's no special reason to use the msgid from the state, the message is ok.

PR #1511, opened 2024-07-25
Alan - Spec alignment

MatheusFranco99:
# Overview

This PR updates the approved changes between ssv and spec, for the spec-alignment script.

Also, it performs a single alignment change.


moshe-blox:
image
is this not a significant diff?

moshe-blox:
image
do we still need the validator stopping logic in the implementation?

moshe-blox:
image
should return error in implementation

moshe-blox:
image
is this not a significant change?

MatheusFranco99:
@moshe-blox 

> is this not a significant diff?

(Regarding the msgID attributes) No. We don't have DomainTypeProvider in spec and the CommitteeID is, actually, better as implemented in the node. Although they result the same, if one needs to change, it should be spec. 

> do we still need the validator stopping logic in the implementation?

This is outside of the scope of spec. And it seems that it's being used in the node.

> should return error in implementation (and probably without attester)

I've raised this issue before (this is the function that is never called, remember?). I agree we can change to DomainError and return the error. May I add that?

> is this not a significant change? i think the implementation is sorting commit signers here to avoid creating duplicates, however that should probably also happen in spec if deemed needed, or likely removed from implementation if not.

This difference already existed before. Since the problem is in spec, I preferred approving it here and creating the issue in spec. What do you think?

moshe-blox:
> I agree we can change to DomainError and return the error. May I add that?
yes please

> This difference already existed before. Since the problem is in spec, I preferred approving it here and creating the issue in spec. What do you think?
+1

agreed on the rest

moshe-blox (approved):

moshe-blox (approved):

PR #1536, opened 2024-07-29
Add spec types wrappers

AKorpusenko:


protocol/v2/types/value_check.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		//contributions, _ := cd.GetSyncCommitteeContributions()
+		//
+		//for _, c := range contributions {
+		//	// TODO check we have selection proof for contribution
+		//	// TODO check slot == duty slot
+		//	// TODO check beacon block root somehow? maybe all beacon block roots should be equal?
+		//
+		//}
```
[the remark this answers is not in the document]
AKorpusenko: @GalRogozinski approved we can just remove this for now

protocol/v2/ssv/spectest/ssv_mapping_test.go
```diff
@@ -96,18 +95,6 @@ func prepareTest(t *testing.T, logger *zap.Logger, name string, test interface{}
 			typedTest.Tests = append(typedTest.Tests, msgProcessingSpecTestFromMap(t, subtest.(map[string]interface{})))
 		}
 
-		return &runnable{
-			name: typedTest.TestName(),
-			test: func(t *testing.T) {
-				typedTest.Run(t)
-			},
-		}
-	case reflect.TypeOf(&partialsigmessage.MsgSpecTest{}).String(): // no use of internal structs so can run as spec test runs
```
AKorpusenko, line 105: @MatusKysel FYI: we don't have any cases of `partialsigmessage.MsgSpecTest`type tests

GalRogozinski: can you add this change to the description..

protocol/v2/qbft/spectest/create_msg_type.go
```diff
@@ -14,7 +14,25 @@ import (
...
+	ExpectedRoot                                     string
+	ExpectedState                                    spectypes.Root `json:"-"` // Field is ignored by encoding/json"
+	ExpectedError                                    string
+}
+
+func (test *CreateMsgSpecTest) TestName() string {
+	return "qbft create message " + test.Name
+}
+
+func (test *CreateMsgSpecTest) RunCreateMsg(t *testing.T) {
```
GalRogozinski, line 35: did you debug and check that all `create` functions run ssv code?

GalRogozinski: I see you are still calling spec logic underneath
https://github.com/bloxapp/ssv/blob/005616572e8845320b43195d9112b4e99abdb379/protocol/v2/qbft/spectest/create_msg_type.go#L96

GalRogozinski (changes requested):

protocol/v2/qbft/spectest/create_msg_type.go
AKorpusenko: @GalRogozinski, I don't get it what for do we need to copy this to implementation right now. By doing this we just increase amount of copy paste code which has to be maintained in 2 repos. This significantly increases a chance to miss something and create a bug, increases amount of work in future and brings no profit.

Here I don't see any business logic we hide when constructing the qbft messages. We use:
- `CreateProposal`
- `CreatePrepare`
- `CreateCommit` 
just to construct a message and sign it. Thats it.

@MatusKysel **FYI**

protocol/v2/ssv/spectest/ssv_mapping_test.go
AKorpusenko: Added

protocol/v2/qbft/spectest/create_msg_type.go
GalRogozinski: https://github.com/bloxapp/ssv/blob/7fe53a03491b405d7d2b54f76f58274c645d4125/protocol/v2/qbft/instance/proposal.go#L274-L315
it seems that you have CreateProposal and other Create functions in ssv already...
so by calling the spec impls you are bypassing ssv which isn't good please change

GalRogozinski (approved):
lgtm

moshe-blox:
why is PartialSigContainer under `types` package in implementation when its under `ssv` package in spec?

let's make sure the others are aligned as well

AKorpusenko:
> why is PartialSigContainer under `types` package in implementation when its under `ssv` package in spec? and RoundRobinProposer is under `types` when its under `qbft` in spec?
> 
> let's make sure everything else is placed correctly as well

@moshe-blox, our implementation of `ReconstructSignature` differ from what we have in spec (no partial sig param, only signatures slice) and I can't place the `PartialSigContainer` without having the import cycles

y0sher:
> > why is PartialSigContainer under `types` package in implementation when its under `ssv` package in spec? and RoundRobinProposer is under `types` when its under `qbft` in spec?
> > let's make sure everything else is placed correctly as well
> 
> @moshe-blox, our implementation of `ReconstructSignature` differ from what we have in spec (no partial sig param, only signatures slice) and I can't place the `PartialSigContainer` without having the import cycles

I agree with @moshe-blox on that end, lets try to keep it organized so even if we copy from the spec its the easiest possible to keep it updated, if you can't do it due to import cycle, IMO you can create a "specwrap" package and put everything under it.

moshe-blox:
re-reviewed, a few things:
- @AKorpusenko the value check funcs were in the wrong package (validator), so i moved them to ssv.
- @GalRogozinski @AKorpusenko the PartialSigContainer implementation is slightly different from the spec, which commit of the spec did you use to copy from? Example: [spec](https://github.com/ssvlabs/ssv-spec/blob/ccf408d1ecd87a4ab631885409d679f10c3fd080/ssv/partial_sig_container.go#L24) vs. [impl](https://github.com/ssvlabs/ssv/blob/be92171eef713707324fd55c95c9dc61676feb6a/protocol/v2/ssv/partial_sig_container.go#L22) ([]byte vs Signature)
- @AKorpusenko why did we rename `spectypes` import to `types` in partial_sig_container.go?

AKorpusenko:
@moshe-blox 
> * @AKorpusenko why did we rename `spectypes` import to `types` in partial_sig_container.go, did you maybe mean to use our types package instead?

I added `partial_sig_container.go` usage `threshold.ReconstructSignatures` and `types.VerifyReconstructedSignature`. Now this file uses `spectypes` only for types, but the implementation is used from ssv.

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -21,7 +20,9 @@ import (
 	qbftcontroller "github.com/ssvlabs/ssv/protocol/v2/qbft/controller"
 	qbftctrl "github.com/ssvlabs/ssv/protocol/v2/qbft/controller"
 	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
+	"github.com/ssvlabs/ssv/protocol/v2/ssv"
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/queue"
+	"github.com/ssvlabs/ssv/protocol/v2/types"
 	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
```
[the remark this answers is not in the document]
AKorpusenko: Fixed

moshe-blox (approved):

PR #1547, opened 2024-07-30
Resurrect Validator Registration Handler

olegshmuelov:
Revive deleted scheduler validator registration handler 
closes https://github.com/ssvlabs/ssv/issues/1548


olegshmuelov:
@moshe-blox 
we need to decide if we want to keep the `BuilderProposals` flag in order to init the handler only if the flag is passed

moshe-blox (approved):

moshe-blox:
> @moshe-blox we need to decide if we want to keep the `BuilderProposals` flag in order to init the handler only if the flag is passed

this isn't an option anymore, only the BN now determines whether to build MEV or not

PR #1549, opened 2024-07-31
chore: added `duty_id` field to committee execution flow

AKorpusenko:
## Description 

- Logger instance we passed to StartDuty and to the runner instance was different from one we pass into the `StartConsumeQueue`. Fixed that
- Also added common field to use by default - duty_id in human readable format: `COMMITTEE-X_Y_Z-%e-%s`


protocol/v2/ssv/validator/committee.go
```diff
@@ -87,7 +87,7 @@ func (c *Committee) RemoveShare(validatorIndex phase0.ValidatorIndex) {
 func (c *Committee) StartConsumeQueue(logger *zap.Logger, duty *spectypes.CommitteeDuty) error {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-
+	logger.Info("Starting committee consume queue")
```
moshe-blox, line 90: this is unnecessary as there's already a starting log inside ConsumeQueue

protocol/v2/ssv/validator/duty_executer.go
```diff
@@ -34,6 +34,8 @@ func (c *Committee) OnExecuteDuty(logger *zap.Logger, msg *types.EventMsg) error
 		return fmt.Errorf("failed to get execute committee duty data: %w", err)
 	}
 
+	logger = logger.With(fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(executeDutyData.Duty.Slot), executeDutyData.Duty.Slot)))
```
moshe-blox, line 37: won't this only show the duty ID in the duty start flow? what about the flow of processing incoming messages? 

do you see this field in "got commit quorum" log, for example?

moshe-blox (changes requested):

protocol/v2/ssv/validator/committee.go
AKorpusenko: Removed 👌

protocol/v2/ssv/validator/duty_executer.go
AKorpusenko: @moshe-blox, I've just tested this on stage. It works fine. I see logs for `"got commit quorum"`

AKorpusenko: As we pass the logger with `duty_id` field added into ConsumeQueue -> ProcessMessage, we also have in `ProcessConsensus` and `ProcessPostConsensus` as well

nkryuchkov (approved):

protocol/v2/ssv/validator/duty_executer.go
y0sher, line 37: @AKorpusenko do we still have `slot` added?

y0sher (approved):

moshe-blox (approved):

PR #1553, opened 2024-08-02
fix: update both Genesis & Alan shares

MatusKysel:
This change allows to update genesis validator shares as well as new shares on any update


nkryuchkov (approved):

moshe-blox:
👍 

let's merge both functions to one UpdateShares that receives both shares (to enforce whole updates at syntax level), and let's also lock during the update.

AKorpusenko (approved):
Seems to be correct ✔️

moshe-blox (approved):

PR #1556, opened 2024-08-05
Added stop old committee queue and runners 

AKorpusenko:
…ed old queues & runners stop.

### DIFF
--- a/protocol/v2/ssv/validator/committee.go
+++ b/protocol/v2/ssv/validator/committee.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/pkg/errors"
@@ -19,6 +20,10 @@ import (
 	"github.com/ssvlabs/ssv/protocol/v2/message"
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/queue"
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/runner"
+)
+
+var (
+	maxRunnerAgeSlot = phase0.Slot(32)
 )
 
 type CommitteeRunnerFunc func(slot phase0.Slot, shares map[phase0.ValidatorIndex]*spectypes.Share, slashableValidators []spectypes.ShareValidatorPK) *runner.CommitteeRunner
@@ -88,13 +93,24 @@ func (c *Committee) StartConsumeQueue(logger *zap.Logger, duty *spectypes.Commit
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
+	// Setting the cancel function separately due the queue could be created in HandleMessage
+	q, found := c.Queues[duty.Slot]
+	if !found {
+		return errors.New(fmt.Sprintf("no queue found for slot %d", duty.Slot))
+	}
+
+	// required to stop the queue consumer when timeout message is received by handler
+	queueCtx, cancelF := context.WithDeadline(c.ctx, time.Unix(c.BeaconNetwork.EstimatedTimeAtSlot(duty.Slot+maxRunnerAgeSlot), 0))
+	q.StopQueueF = cancelF // DO we still need this?
+
 	r := c.Runners[duty.Slot]
 	if r == nil {
 		return errors.New(fmt.Sprintf("no runner found for slot %d", duty.Slot))
 	}
 
 	go func() {
-		err := c.ConsumeQueue(logger, duty.Slot, c.ProcessMessage, r)
+		defer cancelF()
+		err := c.ConsumeQueue(queueCtx, q, logger, duty.Slot, c.ProcessMessage, r)
 		logger.Error("failed consuming queue", zap.Error(err))
 	}()
 	return nil
@@ -292,6 +308,60 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
 	}
 	return nil
 
+}
+func (c *Committee) StopOldRunners(logger *zap.Logger, currentSlot phase0.Slot) error {
+	// TODO: Should we keep this check to prevent overflow case for uint64? Guess no, but lets discuss.
+	if maxRunnerAgeSlot > currentSlot { //
+		return nil
+	}
+
+	firstTooOldSlot := currentSlot - maxRunnerAgeSlot
+	var oldRunnersIds []phase0.Slot
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot, q := range c.Queues {
+		if slot <= firstTooOldSlot {
+			if q.StopQueueF == nil {
+				c.logger.Error("⚠️ can't stop committee queue StopQueueF is nil",
+					fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(slot), slot)),
+					fields.Slot(slot),
+				)
+				continue
+			}
+			logger.Info("stopping old runner queue", zap.Uint64("slot", uint64(slot)))
+			q.StopQueueF()
+			oldRunnersIds = append(oldRunnersIds, slot)
+		}
+	}
+
+	//time.Sleep(100 * time.Millisecond) // TODO: if we deleting the runners, should we do a time.sleep call here to make sure the ProcessMessage finished it work?
+	//												+ If yes, then should we unlock the mtx?
+
+	// TODO: Should we iterate through the Runners map also and delete the old ones from there? Otherwise we get the memory leak issue
+	for _, slot := range oldRunnersIds {
+		logger.Info("stopping old runner", zap.Uint64("slot", uint64(slot)))
+		delete(c.Runners, slot)
+	}
+
+	return nil
+}
+
+func (c *Committee) Stop() {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot, q := range c.Queues {
+		if q.StopQueueF == nil {
+			c.logger.Error("⚠️ can't stop committee queue StopQueueF is nil",
+				fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(slot), slot)),
+				fields.Slot(slot),
+			)
+			continue
+		}
+		q.StopQueueF()
+	}
 }
 
 func (c *Committee) Encode() ([]byte, error) {
--- a/protocol/v2/ssv/validator/committee_queue.go
+++ b/protocol/v2/ssv/validator/committee_queue.go
@@ -75,28 +75,24 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.SSVMessage) {
 
 // ConsumeQueue consumes messages from the queue.Queue of the controller
 // it checks for current state
-func (v *Committee) ConsumeQueue(logger *zap.Logger, slot phase0.Slot, handler MessageHandler, runner *runner.CommitteeRunner) error {
-	if runner == nil {
-		return fmt.Errorf("could not get duty runner for slot %d", slot)
-	}
-
-	ctx, cancel := context.WithCancel(v.ctx)
-	defer cancel()
-
-	var q queueContainer
-
-	err := func() error {
-		v.mtx.RLock() // read v.Queues
-		defer v.mtx.RUnlock()
-		var ok bool
-		q, ok = v.Queues[slot]
-		if !ok {
-			return errors.New(fmt.Sprintf("queue not found for slot %d", slot))
+func (v *Committee) ConsumeQueue(
+	ctx context.Context,
+	q queueContainer,
+	logger *zap.Logger,
+	slot phase0.Slot,
+	handler MessageHandler,
+	runner *runner.CommitteeRunner,
+) error {
+	// in case of any error try to call the ctx.cancel to prevent the ctx leak
+	defer func() {
+		if q.StopQueueF == nil {
+			logger.Error("⚠️ committee queue consumer StopQueueF is nil", fields.Slot(slot))
+			return
 		}
-		return nil
+		q.StopQueueF()
 	}()
-	if err != nil {
-		return err
+	if runner == nil {
+		return errors.New(fmt.Sprintf("duty runner for slot %d is nil", slot))
 	}
 
 	state := *q.queueState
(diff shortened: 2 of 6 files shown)


protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +309,60 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+					fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(slot), slot)),
+					fields.Slot(slot),
+				)
+				continue
+			}
+			logger.Info("stopping old runner queue", zap.Uint64("slot", uint64(slot)))
+			q.StopQueueF()
+			oldRunnersIds = append(oldRunnersIds, slot)
+		}
+	}
```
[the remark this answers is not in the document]
AKorpusenko: Removed this

protocol/v2/ssv/validator/duty_executer.go
```diff
@@ -42,5 +42,9 @@ func (c *Committee) OnExecuteDuty(logger *zap.Logger, msg *types.EventMsg) error
 		return fmt.Errorf("could not start committee consume queue: %w", err)
 	}
 
+	if err := c.StopOldRunners(logger, executeDutyData.Duty.Slot); err != nil {
+		return fmt.Errorf("could not start committee duty: %w", err)
+	}
```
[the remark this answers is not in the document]
AKorpusenko: Good point. Thanks
To prevent too frequent mtx lock I moved this to `Committee.StartDuty()` for cases when we create a new runner, a new goroutine is started to clean up the old runners.

protocol/v2/ssv/validator/committee.go
```diff
@@ -178,6 +194,13 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 
 	}
 
+	// Stop all old committee runners, when new runner is created
+	go func() {
+		if err := c.StopOldRunners(logger, duty.Slot); err != nil {
+			logger.Error("couldn't stop old committee runners", zap.Uint64("current_slot", uint64(duty.Slot)), zap.Error(err))
+		}
+	}()
```
moshe-blox, line 202: why do we need a goroutine for this?

protocol/v2/ssv/validator/committee.go
```diff
@@ -21,6 +22,10 @@ import (
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/runner"
 )
 
+var (
+	maxRunnerAgeSlot = phase0.Slot(32)
```
moshe-blox, line 26: lets try to come up with consistent naming for this new concept

for example `runnerExpirySlots` and `stopExpiredRunners`

protocol/v2/ssv/validator/committee.go
moshe-blox, line 202: why is a goroutine needed here?

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +316,41 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+
+	firstTooOldSlot := currentSlot - maxRunnerAgeSlot
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot := range c.Runners {
+		if slot <= firstTooOldSlot {
+			logger.Info("stopping old committee runner", zap.Uint64("slot", uint64(slot)))
+			delete(c.Runners, slot)
```
moshe-blox, line 332: dont we also need to delete the queue?

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +316,41 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
 	return nil
 
 }
+func (c *Committee) StopOldRunners(logger *zap.Logger, currentSlot phase0.Slot) error {
+	if maxRunnerAgeSlot > currentSlot {
+		return nil
+	}
+
+	firstTooOldSlot := currentSlot - maxRunnerAgeSlot
```
moshe-blox, line 324: i'd rename to `minValidSlot`

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +316,41 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot := range c.Runners {
+		if slot <= firstTooOldSlot {
```
moshe-blox, line 330: @MatheusFranco99  @MatusKysel 
do we have a message validation rule that would not allow messages from 32 slots ago?
i'm asking because it might be >32 (so 33 slots), in which case we'd need to replace the condition here from <= to <

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -75,28 +75,24 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.SSVMessage) {
...
+	ctx context.Context,
+	q queueContainer,
+	logger *zap.Logger,
+	slot phase0.Slot,
+	handler MessageHandler,
+	runner *runner.CommitteeRunner,
+) error {
+	// in case of any error try to call the ctx.cancel to prevent the ctx leak
+	defer func() {
+		if q.StopQueueF == nil {
```
moshe-blox, line 88: i'm likely missing something here, but why would it ever be nil?

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -75,28 +75,24 @@ func (v *Committee) HandleMessage(logger *zap.Logger, msg *queue.SSVMessage) {
...
+			return
 		}
-		return nil
+		q.StopQueueF()
 	}()
-	if err != nil {
-		return err
+	if runner == nil {
+		return errors.New(fmt.Sprintf("duty runner for slot %d is nil", slot))
 	}
```
moshe-blox, line 96: why would the runner ever be nil?

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -22,6 +22,8 @@ type MessageHandler func(logger *zap.Logger, msg *queue.SSVMessage) error
 // queueContainer wraps a queue with its corresponding state
 type queueContainer struct {
 	Q          queue.Queue
+	StopQueueF context.CancelFunc
```
moshe-blox, line 25: i'd recommend renaming it to just Stop or Cancel for a bit more elegance

moshe-blox (approved):

moshe-blox (changes requested):
few changes, recommendations and questions

protocol/v2/ssv/validator/committee.go
MatheusFranco99: We have a rule for ["late messages"](https://github.com/ssvlabs/ssv/blob/stage/message/validation/common_checks.go#L37-L52). Committee messages are allowed up to 34 slots in the future and messages of other duty types are allowed up to 3.

protocol/v2/ssv/validator/committee.go
AKorpusenko: @moshe-blox Actually, yep, good catch. We need to delete queue even we stoped the consumer of it.

protocol/v2/ssv/validator/committee.go
AKorpusenko: I decided to call it in goroutine, because we have `mtx.Lock()` inside`StopOldRunners`. The problem is that the mutex is already locked by `StartDuty` and it's going to be deadlock.
So I see we have 2 options:
- call inside goroutine and it will take the resource after `StartDuty` execution is finished
- call it synchronously, but remove the lock inside of `StopOldRunners`. 

I prefer to keep `mtx.Lock()` usage `StopOldRunners()` to prevent data race.

Let me know what do you think about it

protocol/v2/ssv/validator/msgqueue_consumer.go
AKorpusenko: Agree. Thanks

protocol/v2/ssv/validator/committee_queue.go
AKorpusenko: Agree. This is extra check. Removed

protocol/v2/ssv/validator/committee_queue.go
AKorpusenko: You are right. This can be replaced with just `defer q.Stop()`. We 100% set ctx.cancel func before the `ConsumeQueue()` call => no presence check required

protocol/v2/ssv/validator/committee.go
AKorpusenko: Agree. Renamed old -> expired

protocol/v2/ssv/validator/committee.go
AKorpusenko: @moshe-blox I set the value `runnerExpirySlots` to 34 slots

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +318,40 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+	}
+
+	minValidSlot := currentSlot - runnerExpirySlots
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot := range c.Runners {
+		if slot <= minValidSlot {
+			logger.Info("stopping expired committee runner", zap.Uint64("slot", uint64(slot)))
```
y0sher, line 334: I think this is debug. going to happen every slot basically.

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +318,40 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+
+	return nil
+}
+
+func (c *Committee) Stop() {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot, q := range c.Queues {
+		c.logger.Error("stopping committee queue",
```
y0sher, line 348: doesn't seem like an error to me

protocol/v2/ssv/validator/committee.go
```diff
@@ -293,6 +318,40 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+	minValidSlot := currentSlot - runnerExpirySlots
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	for slot := range c.Runners {
+		if slot <= minValidSlot {
+			logger.Info("stopping expired committee runner", zap.Uint64("slot", uint64(slot)))
+			delete(c.Runners, slot)
+			delete(c.Queues, slot)
```
y0sher, line 336: didn't you mean to also call Q.Stop here to stop the queue consume goroutine?

y0sher (changes requested):
Good job on this! few comments

y0sher (approved):

protocol/v2/ssv/validator/committee.go
```diff
@@ -177,6 +196,13 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 
 	}
 
+	// Stop all expired committee runners, when new runner is created
+	go func() {
+		if err := c.stopExpiredRunners(logger, duty.Slot); err != nil {
+			logger.Error("couldn't stop expired committee runners", zap.Uint64("current_slot", uint64(duty.Slot)), zap.Error(err))
+		}
+	}()
+
```
moshe-blox, line 205: repeating my previous comment which disappeared: we don't need a goroutine here, we can just remove the lock from stopExpiredRunners and execute it directly here

if we're worried about other callers, then it can be renamed to unsafeStopExpiredRunners() with a comment that a lock is required

protocol/v2/ssv/validator/committee.go
```diff
@@ -87,14 +93,27 @@ func (c *Committee) RemoveShare(validatorIndex phase0.ValidatorIndex) {
...
+	// Setting the cancel function separately due the queue could be created in HandleMessage
+	q, found := c.Queues[duty.Slot]
+	if !found {
+		return errors.New(fmt.Sprintf("no queue found for slot %d", duty.Slot))
+	}
+
+	// required to stop the queue consumer when timeout message is received by handler
+	queueCtx, cancelF := context.WithDeadline(c.ctx, time.Unix(c.BeaconNetwork.EstimatedTimeAtSlot(duty.Slot+runnerExpirySlots), 0))
+	q.Stop = cancelF // DO we still need this?
```
moshe-blox, line 105: i think we dont need this anymore now that we cancel with a deadline?

we can just give the queues the ctx of the Committee and add a new cancel field which we can populate in NewCommittee to later call in Stop()

protocol/v2/ssv/validator/committee.go
```diff
@@ -290,6 +316,40 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
...
+	defer c.mtx.Unlock()
+
+	for slot, q := range c.Queues {
+		c.logger.Debug("stopping committee queue",
+			fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(slot), slot)),
+			fields.Slot(slot),
+		)
+		q.Stop()
+	}
+}
```
moshe-blox, line 352: see comment in StartConsumeQueue: this whole func can just be c.cancel()

moshe-blox (changes requested):
left just a few comments

protocol/v2/ssv/validator/committee.go
```diff
@@ -177,6 +196,13 @@ func (c *Committee) StartDuty(logger *zap.Logger, duty *spectypes.CommitteeDuty)
 
 	}
 
+	// Stop all expired committee runners, when new runner is created
+	go func() {
+		if err := c.stopExpiredRunners(logger, duty.Slot); err != nil {
+			logger.Error("couldn't stop expired committee runners", zap.Uint64("current_slot", uint64(duty.Slot)), zap.Error(err))
```
moshe-blox, line 202: since this func no longer stops anything but just removes from maps, i think we should rename it to and any mention about it from "stop" to "prune"

moshe-blox (approved):

protocol/v2/ssv/validator/committee.go
```diff
@@ -290,6 +315,27 @@ func (c *Committee) ProcessMessage(logger *zap.Logger, msg *queue.SSVMessage) er
 	return nil
 
 }
+func (c *Committee) unsafePruneExpiredRunners(logger *zap.Logger, currentSlot phase0.Slot) error {
+	if runnerExpirySlots > currentSlot {
+		return nil
+	}
```
olegshmuelov, line 321: could be we never reach this piece of code?
since runnerExpirySlots is hardcoded to `phase.Slot(34)`

PR #1557, opened 2024-08-06
fix: stop genesis validators on fork

AKorpusenko:
## Description:

- Added primitive `ForkMonitor()` which is checks is fork is passed or not
- In case it passed the fork epoch, then `CleanGenesisValidators()` is called to stop all genesis validators instances
- Added skipping the initialization of `genesisValidator` in `onShareInit()`


moshe-blox:
made a refactor to stop all validators at once with a parent context, and a bunch of other changes for nil safety and memory freeing

y0sher (approved):
lgtm

AKorpusenko (commented):
@moshe-blox, lgtm 👍

y0sher:
@AKorpusenko please test this in `stage` ON-fork to see it works and doesn't break.

operator/validator/controller.go
```diff
@@ -1242,38 +1256,31 @@ func (c *controller) startCommittee(vc *validator.Committee) (bool, error) {
...
-		case <-c.context.Done():
-			return
-		case <-time.After(timeUntilFork):
-			logger.Info("Stopping genesis validators after alan fork")
-			c.CleanGenesisValidators()
-			logger.Info("✅Successfully stopped genesis validators")
+		case <-c.ctx.Done():
 			return
+		case <-next:
+			next = slotTicker.Next()
```
AKorpusenko, line 1281: @moshe-blox, do we really need this?

operator/validator/controller.go
```diff
@@ -1242,38 +1256,31 @@ func (c *controller) startCommittee(vc *validator.Committee) (bool, error) {
...
 
+	next := slotTicker.Next()
 	for {
 		select {
-		case <-c.context.Done():
-			return
-		case <-time.After(timeUntilFork):
-			logger.Info("Stopping genesis validators after alan fork")
-			c.CleanGenesisValidators()
-			logger.Info("✅Successfully stopped genesis validators")
```
AKorpusenko: @moshe-blox, why did you remove the logs here? It happens once and bring some confidence stopping did happen

moshe-blox: agree, i returned the log

operator/validator/controller.go
moshe-blox: we don't really, but i don't trust very long sleeps to be precise

when we release Alan, the fork epoch could be set to around 2 weeks ahead, and i can't trust that if we schedule a Go timer for 2 weeks from now, it would fire precisely at the exact millisecond we need it to

given that system clocks often correct themselves, in such a long timeframe, i don't know what would be the impact of that

moshe-blox (approved):

PR #1560, opened 2024-08-06
update to Go 1.22

### LINKED ISSUE #1545: upgrade Go to 1.21
Go 1.20 is no longer supported, so this is critical. The reason we prefer Go 1.21 over 1.22 is because that's where the libp2p and go-ethereum dependencies are presently at.

After upgrading Go, we should upgrade the libp2p and go-ethereum dependencies too, and make sure there were no breaking changes or any changes we must make by reading their changelogs or docs.

nkryuchkov:
Closes #1545 

- Update to Go 1.22
- Update libp2p to the latest version
- Fix lack of dump of failure in `StartNewRunnerDutySpecTest`
- Fix bug with slice cap on unmarshalling an instance container
	- When `StoredInstances` is unmarshaled from JSON in https://github.com/ssvlabs/ssv/blob/e269d3be6ba4fd752f61bdafbef2473c4f51c89f/protocol/v2/ssv/spectest/ssv_mapping_test.go#L367-L369 it has `len=1 cap=4` before Go 1.21, but since Go 1.21 it has `len=1 cap=1`
- update `go-ethereum` from v1.13.14 to v1.13.15 as v1.13.14 is vulnerable


moshe-blox:
The difference in `StoredInstances` between spec and implementation is because the spec keeps appending instances without limit which is eventually a memory leak, so we can't revert it back.

Why would `cap` change between Go versions? That doesn't make any sense :/

nkryuchkov:
```go
func TestJSONUnmarshalSliceCap(t *testing.T) {
	exampleJSON := `{"testSlice": [1]}`

	var testStruct struct {
		TestSlice []int `json:"testSlice"`
	}

	require.NoError(t, json.Unmarshal([]byte(exampleJSON), &testStruct))
	require.Equal(t, 1, len(testStruct.TestSlice))
	require.Equal(t, 1, cap(testStruct.TestSlice))
}
```

Go 1.20.11:

```
/opt/homebrew/Cellar/go@1.20/1.20.11/libexec/bin/go tool test2json -t /Users/bloxmac/Library/Caches/JetBrains/GoLand2024.2/tmp/GoLand/___TestJSONUnmarshalSliceCap_in_github_com_ssvlabs_ssv_protocol_v2_ssv_spectest.test -test.v=test2json -test.paniconexit0 -test.run ^\QTestJSONUnmarshalSliceCap\E$
=== RUN   TestJSONUnmarshalSliceCap
    ssv_mapping_test.go:592: 
        	Error Trace:	/Users/bloxmac/GolandProjects/ssv/protocol/v2/ssv/spectest/ssv_mapping_test.go:592
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 4
        	Test:       	TestJSONUnmarshalSliceCap
--- FAIL: TestJSONUnmarshalSliceCap (0.00s)

Expected :1
Actual   :4
<Click to see difference>


FAIL
```

Go 1.21.12:

```
=== RUN   TestJSONUnmarshalSliceCap
--- PASS: TestJSONUnmarshalSliceCap (0.00s)
PASS
```

Go 1.22.5:

```
=== RUN   TestJSONUnmarshalSliceCap
--- PASS: TestJSONUnmarshalSliceCap (0.00s)
PASS
```

nkryuchkov:
The change was introduced in https://github.com/golang/go/commit/ac27b4dd1d72bd6467bf7bfa8bc77a1522efd4cc, so it seems the tests had a bug that they worked by accident when cap was 4 but they shouldn't have worked

protocol/v2/ssv/spectest/multi_start_new_runner_duty_type.go
```diff
@@ -99,8 +99,6 @@ func (test *StartNewRunnerDutySpecTest) RunAsPartOfMultiTest(t *testing.T, logge
 	postRoot, err := test.Runner.GetRoot()
 	require.NoError(t, err)
 
-	require.EqualValues(t, test.PostDutyRunnerStateRoot, hex.EncodeToString(postRoot[:]))
```
moshe-blox, line 102: why was this removed?

moshe-blox (commented):
great work

only 1 comment so far

protocol/v2/ssv/spectest/multi_start_new_runner_duty_type.go
nkryuchkov: There's a similar check inside the if. The idea is that if `PostDutyRunnerStateRoot` differs, then we calculate `diff` and trigger `require` library providing `diff` as the reason. However, with this line it either failed without calculating diff or passed, so the if-block below was unused

y0sher (approved):
lgtm

nkryuchkov:
@MatusKysel `go-ethereum` v1.13.x uses `golang.org/x/exp/slog` and v1.14.x uses `log/slog`, so we need to update to v1.14.x to fix this. However, v1.14.x changes some method signatures and when I fixed them our tests got broken, so I decided that the update to v1.14.x should be out of the scope of this PR. When we update to 1.14, the location you mentioned won't compile so we won't forget about that

PR #1563, opened 2024-08-06
Share Conversion Func Tests

### LINKED ISSUE #1554: add unit tests for share conversion functions
Today we convert from Alan to Genesis share. We should add the reverse function for symmetry and then write tests that prove the conversion works and preserves all fields (except those that are not present on both of the forks [of course]).

At minimum, we should have this test:
1. Create an Alan share with all fields populated
2. Convert to Genesis share
4. Convert share back to Alan share
5. Expect equality with original share
6. Modify final share and expect inequality (to ensure equality check is valid)
7. Modify original share in the same way and expect equality again

olegshmuelov:
closes #1554


nkryuchkov (approved):

protocol/genesis/types/share_test.go
```diff
@@ -0,0 +1,107 @@
...
+	genesisShare := ConvertToGenesisShare(originalShare, &spectypes.CommitteeMember{OperatorID: 1})
+
+	// Step 3: Convert share back to Alan share
+	convertedBackShare := ConvertFromGenesisShare(genesisShare)
+
+	// Step 4: Expect equality with original share
+	require.True(t, sharesEqual(originalShare, convertedBackShare), "The converted back share should be equal to the original share")
+
+	// Step 5: Modify final share and expect inequality
+	modifiedShare := deepCopyShare(convertedBackShare)
```
moshe-blox, line 37: the fact that we need to copy the converted share before modifying to not modify the original has raised a bug in the conversion functions — they don't properly copy all of the fields.

instead of copying, let's fix the conversion funcs to copy any non-value field (such as slices or pointers) before assigning it to the new share.

moshe-blox (changes requested):
very well done!

left a change request for a bug that's not in the test, but i think we should fix it here because the test should prove it's fixed.

protocol/genesis/types/share_test.go
olegshmuelov: done here - https://github.com/ssvlabs/ssv/pull/1563/commits/b58bda9f5596a412202a2057ed3dcf03c9796b54

protocol/genesis/types/share.go
```diff
@@ -35,33 +35,42 @@ func ConvertToGenesisShare(share *spectypes.Share, operator *spectypes.Committee
...
+			SharePubKey: c.PubKey,
+			Signer:      c.OperatorID,
 		})
 	}
 
 	return share
 }
+
+// ConvertToGenesisSSVShare converts an Alan share to a genesis SSV share.
+func ConvertToGenesisSSVShare(alanSSVShare *typesv2.SSVShare, operator *spectypes.CommitteeMember) *SSVShare {
```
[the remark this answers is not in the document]
y0sher: https://github.com/ssvlabs/ssv/pull/1563/files#diff-e7b6091cd20c01cd2f58a6bd7bef1255faefbba9f8831dfbe611867e824b820bR1062

y0sher (approved):

protocol/genesis/types/share.go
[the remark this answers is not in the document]
moshe-blox: yeah i commented on the wrong one

protocol/genesis/types/share.go
```diff
@@ -15,53 +15,44 @@ func ConvertToGenesisShare(share *spectypes.Share, operator *spectypes.Committee
...
 
-	for _, c := range alanShare.Committee {
-		share.Committee = append(share.Committee, &genesisspectypes.Operator{
-			OperatorID: c.Signer,
-			PubKey:     c.SharePubKey,
-		})
+	return &SSVShare{
+		Share:    *genesisShare,
+		Metadata: convertedMetadata,
 	}
```
moshe-blox, line 57: I missed it before, but we're only testing the spectypes.Share conversion and not the entire SSVShare, so this part remains untested.

With that said, I noticed that `Validator`(for whom we convert the SSVShare) only uses the Share.Share field and not the Metadata, so that could be unnecessary to test for.

@y0sher wdyt?

olegshmuelov: My initial intention was to include metadata conversion for completeness. However, since the Validator isn't currently using it and the genesis code will be removed after the fork, it makes sense to remove the metadata conversion to simplify the code. wdyt?

moshe-blox (approved):

protocol/genesis/types/share.go
moshe-blox: I think it's safe to leave it as it's not used in practice even though Validator struct still requires it.

PR #1571, opened 2024-08-07
fix CI

nkryuchkov:


moshe-blox (approved):

integration/qbft/tests/setup_test.go
```diff
@@ -35,8 +35,6 @@ func GetSharedData(t *testing.T) SharedData { //singleton B-)
 }
 
 func TestMain(m *testing.M) {
-	return // TODO: fixme
```
y0sher, line 38: lets comment this test as it should not be running (its all skipped)

y0sher (approved):

PR #1573, opened 2024-08-07
Align Pre-Fork to Main DB

olegshmuelov:
The pre fork version introduced a bug where the DB shareStorage schema was changed without migration.
The fix is to restore the original share storage schema and to use existing mappings.

This fix will require all prefork/postfork stage nodes to be re-synced.


registry/storage/shares.go
```diff
@@ -284,21 +284,18 @@ func specShareToStorageShare(share *types.SSVShare) *storageShare {
...
-		// }
-
 		committee[i] = &spectypes.ShareMember{
 			Signer:      c.OperatorID,
 			SharePubKey: c.PubKey,
 		}
 	}
 
+	var validatorPubKey spectypes.ValidatorPK
+	copy(validatorPubKey[:], share.ValidatorPubKey)
```
y0sher, line 294: I think this can work `validatorPubKey := spectypes.ValidatorPK(share.ValidatorPubKey)`

olegshmuelov: If we ensure that share.ValidatorPubKey is 48 bytes long, then directly converting using `validatorPubKey := spectypes.ValidatorPK(share.ValidatorPubKey)` is safe.
However, if we don’t check for length and the length is not 48 it can lead to runtime error.
a safer approach is:
```go
var validatorPubKey spectypes.ValidatorPK
copy(validatorPubKey[:], share.ValidatorPubKey)
```

lmk wdyt

y0sher: well, if the validator pubkey is not 48 bytes then its a bug. in that case we could panic anyway, or at least return a critical error.. at least the error will surface, with the copy approach the key would not be valid and we won't know. So I guess its best to ensure the size anyway

olegshmuelov: sure, added length check

y0sher (approved):

y0sher:
Please update that you've tested transition from main db to this branch on stage before merging.

olegshmuelov:
> Please update that you've tested transition from main db to this branch on stage before merging.
The PR was tested.

y0sher (approved):

moshe-blox (approved):

PR #1574, opened 2024-08-07
differ: add unit tests

moshe-blox:
This PR adds another unit test to differ.

However, I've noticed that on main branch the method `ProposerRunner.ProcessPreConsensus` is somehow cut and shuffled in the middle.

We should add yet another unit test to reproduce it in differ and fix that problem.


y0sher (approved):

moshe-blox:
i think we can merge this and fix the other bug later :https://github.com/ssvlabs/ssv/issues/1681

y0sher (approved):

PR #1579, opened 2024-08-07
fix: 3 forking bugs

guym-blox:
### Fixes
1. QBFT controller does not use updated domain type
2. Handshake filter does not use use update domain type
3. Message validation does not switch on fork epoch

### DIFF
--- a/networkconfig/holesky-stage.go
+++ b/networkconfig/holesky-stage.go
@@ -16,7 +16,7 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        99999999,
+	AlanForkEpoch:        70452,
 	Bootnodes: []string{
 		"enr:-Ja4QDYHVgUs9NvlMqq93ot6VNqbmrIlMrwKnq4X3DPRgyUNB4ospDp8ubMvsf-KsgqY8rzpZKy4GbE1DLphabpRBc-GAY_diLjngmlkgnY0gmlwhDQrLYqJc2VjcDI1NmsxoQKnAiuSlgSR8asjCH0aYoVKM8uPbi4noFuFHZHaAHqknYNzc3YBg3RjcIITiYN1ZHCCD6E",
 	},
--- a/networkconfig/holesky-stage.go
+++ b/networkconfig/holesky-stage.go
@@ -16,7 +16,7 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        70615,
+	AlanForkEpoch:        70617,
 	Bootnodes: []string{
 		"enr:-Ja4QDYHVgUs9NvlMqq93ot6VNqbmrIlMrwKnq4X3DPRgyUNB4ospDp8ubMvsf-KsgqY8rzpZKy4GbE1DLphabpRBc-GAY_diLjngmlkgnY0gmlwhDQrLYqJc2VjcDI1NmsxoQKnAiuSlgSR8asjCH0aYoVKM8uPbi4noFuFHZHaAHqknYNzc3YBg3RjcIITiYN1ZHCCD6E",
 	},
--- a/networkconfig/holesky-stage.go
+++ b/networkconfig/holesky-stage.go
@@ -11,8 +11,8 @@ import (
 var HoleskyStage = NetworkConfig{
 	Name:                 "holesky-stage",
 	Beacon:               beacon.NewNetwork(spectypes.HoleskyNetwork),
-	GenesisDomainType:    [4]byte{0x00, 0x00, 0x31, 0x10},
-	AlanDomainType:       [4]byte{0x00, 0x00, 0x31, 0x16},
+	GenesisDomainType:    [4]byte{0x00, 0x00, 0x31, 0x12},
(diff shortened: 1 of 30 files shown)


networkconfig/holesky-stage.go
```diff
@@ -11,12 +11,12 @@ import (
...
-	GenesisDomainType:    [4]byte{0x00, 0x00, 0x31, 0x12},
-	AlanDomainType:       [4]byte{0x00, 0x00, 0x31, 0x13},
+	GenesisDomainType:    [4]byte{0x00, 0x00, 0x31, 0x10},
+	AlanDomainType:       [4]byte{0x00, 0x00, 0x31, 0x16},
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        99999999,
+	AlanForkEpoch:        70617,
 	Bootnodes: []string{
```
moshe-blox, line 20: we should revert this back

operator/validator/controller.go
```diff
@@ -1358,8 +1358,10 @@ func SetupCommitteeRunners(
 		}
 		config.ValueCheckF = valueCheckF
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.Operator.CommitteeID[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		qbftCtrl := qbftcontroller.NewController(func() []byte {
+			identifier := spectypes.NewMsgID(options.NetworkConfig.AlanDomainType, options.Operator.CommitteeID[:], role)
+			return identifier[:]
+		}, options.Operator, config, options.OperatorSigner, options.FullNode)
```
moshe-blox, line 1364: this would be clearer to read:
```suggestion
		identifier := func() []byte {
			identifier := spectypes.NewMsgID(options.NetworkConfig.AlanDomainType, options.Operator.CommitteeID[:], role)
			return identifier[:]
		}
		qbftCtrl := qbftcontroller.NewController(identifier, options.Operator, config, options.OperatorSigner, options.FullNode)
```

operator/validator/controller.go
```diff
@@ -1418,8 +1420,10 @@ func SetupRunners(
 		}
 		config.ValueCheckF = valueCheckF
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.SSVShare.Share.ValidatorPubKey[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		qbftCtrl := qbftcontroller.NewController(func() []byte {
+			identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.SSVShare.Share.ValidatorPubKey[:], role)
+			return identifier[:]
+		}, options.Operator, config, options.OperatorSigner, options.FullNode)
```
moshe-blox, line 1426: same as above

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -59,7 +59,9 @@ func NewCommitteeObserver(identifier convert.MessageID, opts CommitteeObserverOp
 
 	// TODO: does the specific operator matters?
 
-	ctrl := qbftcontroller.NewController(identifier[:], opts.Operator, config, opts.OperatorSigner, opts.FullNode)
+	ctrl := qbftcontroller.NewController(func() []byte {
+		return identifier[:]
+	}, opts.Operator, config, opts.OperatorSigner, opts.FullNode)
```
moshe-blox, line 64: same as above

moshe-blox (changes requested):
well done 💪 

please see comments

operator/validator/controller.go
```diff
@@ -1358,8 +1358,11 @@ func SetupCommitteeRunners(
 		}
 		config.ValueCheckF = valueCheckF
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.Operator.CommitteeID[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		identifier := func() []byte {
+			identifier := spectypes.NewMsgID(options.NetworkConfig.AlanDomainType, options.Operator.CommitteeID[:], role)
+			return identifier[:]
+		}
```
y0sher, line 1364: why do we need this to be a func?

y0sher: Also why `options.NetworkConfig.DomainType()` isn't good enough? since buildController is called everytime a new runner is created (on each slot with duty), we should be able to grab the correct domain when the fork epoch has passed.

operator/validator/controller.go
```diff
@@ -1418,8 +1421,11 @@ func SetupRunners(
 		}
 		config.ValueCheckF = valueCheckF
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.SSVShare.Share.ValidatorPubKey[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		identifier := func() []byte {
+			identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType(), options.SSVShare.Share.ValidatorPubKey[:], role)
+			return identifier[:]
+		}
+		qbftCtrl := qbftcontroller.NewController(identifier, options.Operator, config, options.OperatorSigner, options.FullNode)
```
y0sher, line 1428: ditto

protocol/v2/qbft/controller/controller.go
```diff
@@ -37,14 +37,14 @@ type Controller struct {
 }
 
 func NewController(
-	identifier []byte,
+	Identifier func() []byte,
```
y0sher, line 40: A controller identifier should not change after its created, so I don't get why this func is needed.

PR #1616, opened 2024-08-12
fix flaky TestScheduler_Committee_Indices_Changed_Attester_Only

nkryuchkov:
as per title

### DIFF
--- a/operator/duties/committee_test.go
+++ b/operator/duties/committee_test.go
@@ -387,10 +387,11 @@ func TestScheduler_Committee_Indices_Changed_Attester_Only(t *testing.T) {
 	waitForNoActionCommittee(t, logger, fetchDutiesCall, executeDutiesCall, timeout)
 
 	// STEP 4: wait for committee duties to be executed
-	currentSlot.Set(phase0.Slot(2))
 	aDuties, _ := attDuties.Get(0)
-	committeeMap := commHandler.buildCommitteeDuties([]*eth2apiv1.AttesterDuty{aDuties[2]}, nil, 0, currentSlot.Get())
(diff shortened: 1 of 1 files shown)


operator/duties/committee_test.go
```diff
@@ -387,10 +387,11 @@ func TestScheduler_Committee_Indices_Changed_Attester_Only(t *testing.T) {
 	waitForNoActionCommittee(t, logger, fetchDutiesCall, executeDutiesCall, timeout)
 
 	// STEP 4: wait for committee duties to be executed
-	currentSlot.Set(phase0.Slot(2))
 	aDuties, _ := attDuties.Get(0)
-	committeeMap := commHandler.buildCommitteeDuties([]*eth2apiv1.AttesterDuty{aDuties[2]}, nil, 0, currentSlot.Get())
+	const executionSlot = phase0.Slot(2)
+	committeeMap := commHandler.buildCommitteeDuties([]*eth2apiv1.AttesterDuty{aDuties[2]}, nil, 0, executionSlot)
 	setExecuteDutyFuncs(scheduler, executeDutiesCall, len(committeeMap))
+	currentSlot.Set(executionSlot)
```
y0sher, line 394: can you explain this fix? Since currentSlot.Get() is called before any code is executed it should always return 2.

nkryuchkov: @y0sher On step 3 `currentSlot` is 1, on step 4 it is 2. However, we need to set it after `setExecuteDutyFuncs` because setting `currentSlot` to 2 triggers calling `ExecuteCommitteeDuty`, so it must be set before

y0sher (approved):

PR #1623, opened 2024-08-13
dedicated error for unexpected genesis/alan messages

nkryuchkov:
Show dedicated errors if received
- alan message while running genesis
- genesis message while running alan


y0sher:
@Roy-blox @MatheusFranco99 @andrew-blox 
please let us know if this fixes your msg validation concerns..

y0sher (approved):

message/validation/genesis/validation.go
```diff
@@ -334,6 +334,12 @@ func (mv *messageValidator) validateP2PMessage(pMsg *pubsub.Message, receivedAt
 	msg, err := genesisqueue.DecodeGenesisSignedSSVMessage(signedSSVMsg)
 	if err != nil {
 		if errors.Is(err, genesisqueue.ErrDecodeNetworkMsg) {
+			alanSignedSSVMsg := &alanspectypes.SignedSSVMessage{}
+			if err := alanSignedSSVMsg.Decode(pMsg.GetData()); err == nil {
+				e := ErrAlanMessage
+				return nil, Descriptor{}, e
```
y0sher, line 340: decoding to alan message in genesis msg validation is wasteful without need. we don't need to decode here.

nkryuchkov: However, this is the case that we saw on stage 
image

I think it's fine to keep it here, I understand that from the design point of view alan and genesis shouldn't know about each other, but these errors really help us understand what's wrong with the message and I didn't have better ideas on how to do this. Anyway, we'll deprecate genesis message validation soon, so this won't live long

moshe-blox: @nkryuchkov the concern is generating more garbage per incoming message (which is the likely hottest path in the node).

we're already aware that we receive messages from other forks, specifically during the first seconds of connection before handshake fails due to different domain type, so now that your PR confirms this again, do we still need this code? how does it benefit? maybe we can find another solution

nkryuchkov: @moshe-blox 

> @nkryuchkov the concern is generating more garbage per incoming message (which is the likely hottest path in the node, although rejections are less common due to the peers being silenced).

Why would it be garbage? The PR would replace the malformed message log (which isn't helpful) with the log of what exactly is malformed

> do we still need this code? how does it benefit? maybe we can find another solution

Yes, we'll be able to see what exactly is wrong if we get malformed messages. And today's case on the stage nodes confirms this: we have some nodes down after the fork because of malformed messages and we don't see the exact reason in the logs. Why do we need to find another solution and what are the disadvantages of this one?

moshe-blox: @nkryuchkov garbage as in allocating more values for the garbage collector to clean (in this case the decoded 
 message structs and whatever else is allocated by UmarshalSSZ)

nkryuchkov: > @nkryuchkov garbage as in allocating more objects for the garbage collector to clean (in this case the decoded message structs and whatever else is allocated by UmarshalSSZ)

Yes, there would be more allocations, but anyway, we do allocations for all incoming messages, I don't think this particular allocation would make a lot of difference

moshe-blox: > we have some nodes down after the fork because of malformed messages and we don't see the exact reason in the logs
we can assume that "malformed pubsub message" means we got a message from the other fork, and if we're not sure we can confirm by the peer ID that they're indeed running the other fork

if we see a case where it isn't another fork, then we can deploy this change to debug deeper

with that said, this PR can be merged if either:
1. this check is enabled with a flag
2. instead of decoding the whole message, it reads only the domain type from the pubsub message (i assume the position of it in the raw bytes is deterministic)

wdyt @y0sher @nkryuchkov ?

y0sher:
Closing this for now.

PR #1627, opened 2024-08-14
Fix linter errors bumping linter from 1.59.1 => 1.60.1

AKorpusenko:


network/topics/controller.go
```diff
@@ -131,7 +131,7 @@ func (ctrl *topicsCtrl) UpdateScoreParams(logger *zap.Logger) error {
 		}
 	}
 	if len(errs) > 0 {
-		return fmt.Errorf(errs)
+		return fmt.Errorf("%s", errs)
```
nkryuchkov, line 134: can we rewrite error handling here to use https://pkg.go.dev/errors#Join ?

AKorpusenko: Changes are welcome (but maybe in different PR).
Right now I see we need to fix linter ASAP

nkryuchkov: > I guess it's just this, right @nkryuchkov ?

No, because `errs` is string. I think it should be something like this but I don't mind doing it in another PR

```go

func (ctrl *topicsCtrl) UpdateScoreParams(logger *zap.Logger) error {
	if ctrl.scoreParamsFactory == nil {
		return fmt.Errorf("scoreParamsFactory is not set")
	}
	var errs error
	topics := ctrl.ps.GetTopics()
	for _, topicName := range topics {
		topic := ctrl.container.Get(topicName)
		if topic == nil {
			errs = errors.Join(errs, fmt.Errorf("topic %s is not ready; ", topicName)
			continue
		}
		p := ctrl.scoreParamsFactory(topicName)
		if p == nil {
			errs = errors.Join(errs, fmt.Errorf("score params for topic %s is nil; ", topicName)
			continue
		}
		if err := topic.SetScoreParams(p); err != nil {
			errs = errors.Join(errs, fmt.Errorf("could not set score params for topic %s: %d; ", topicName, err)
			continue
		}
	}

	return errs
}
```

integration/qbft/tests/setup_test.go
```diff
@@ -75,10 +76,13 @@ func TestMain(m *testing.M) {
 		Nodes: nodes,
 	}
 
-	m.Run()
+	// exitCode required here only for linter
```
nkryuchkov, line 79: but does a non-zero code also indicate an issue while running the test?

integration/qbft/tests/setup_test.go
```diff
@@ -75,10 +76,13 @@ func TestMain(m *testing.M) {
...
+	exitCode := m.Run()
 
 	//teardown
 	for i := 0; i < len(ln.Nodes); i++ {
 		_ = ln.Nodes[i].Close()
 	}
+
+	os.Exit(exitCode)
```
nkryuchkov, line 87: can we use [require.Zero](https://pkg.go.dev/github.com/stretchr/testify/require#Zero)?

AKorpusenko: @nkryuchkov, you can't use `require.Zero` as it needs `t *testing.T` and we have here `testing.M`

network/topics/controller.go
moshe-blox: let's do it here it should be quite straightforward @nkryuchkov @AKorpusenko

moshe-blox (changes requested):
well done @AKorpusenko 
just 1 change request

network/topics/controller.go
AKorpusenko: @moshe-blox, @nkryuchkov updated this

AKorpusenko:
> well done @AKorpusenko just 1 change request

@moshe-blox, added the usage of `errors.Join` 👌

moshe-blox (approved):

PR #1628, opened 2024-08-14
fix: duties aren't refetched after initial sync

MatusKysel:
This PR fixed broken flow after resync, when we would not setup new validators after metadata was fetched but only update running ones, but as there were no running once after resync we did not run any.


registry/storage/shares.go
```diff
@@ -58,7 +58,6 @@ type sharesStorage struct {
 	db             basedb.Database
 	prefix         []byte
 	shares         map[string]*types.SSVShare
-	operators      map[uint64][]byte
```
y0sher, line 61: not used anymore?

y0sher: I guess we have this data already in `ValidatorStore` and `operatorStorage`, so this is legacy.

y0sher (approved):

moshe-blox (approved):

PR #1633, opened 2024-08-16
bootnode: filter nodes by domain type (stage)

nkryuchkov:
port of https://github.com/ssvlabs/ssv/pull/1434 to stage branch


y0sher:
We decided to go with a different solution.

PR #1651, opened 2024-08-22
fix duty_id log to be with the correct slot

guym-blox:
### Description 

The log that indicates pruning of old committee runners had the current executed dutyid added to it. This PR replaces it with actual duty id of the pruned runner.


protocol/v2/ssv/validator/committee.go
```diff
@@ -324,6 +324,7 @@ func (c *Committee) unsafePruneExpiredRunners(logger *zap.Logger, currentSlot ph
 
 	for slot := range c.Runners {
 		if slot <= minValidSlot {
+			logger = logger.With(fields.DutyID(fields.FormatCommitteeDutyID(c.Operator.Committee, c.BeaconNetwork.EstimatedEpochAtSlot(slot), slot)))
```
y0sher, line 327: did you check this on stage? wouldn't this add 2 duty_id fields since it already has dutyID field?

guym-blox: checked with e2e

y0sher: and can you confirm my question?

y0sher: Did this by passing down a new logger

y0sher (approved):

PR #1656, opened 2024-08-25
fix: (discovery) incorrect domain type check and stale ENR on fork

moshe-blox:
**Changes**
- Fix domain type check
- Reduce discovery interval to match `main`
- Publish ENR only once after adding/removing subnets
- Switch to bootnode hosted on stage

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -23,7 +23,7 @@ import (
 )
 
 var (
-	defaultDiscoveryInterval = time.Second
+	defaultDiscoveryInterval = time.Millisecond * 100
 	publishENRTimeout        = time.Minute
 
 	publishStateReady   = int32(0)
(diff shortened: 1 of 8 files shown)


network/discovery/dv5_service.go
```diff
@@ -152,9 +152,13 @@ func (dvs *DiscV5Service) checkPeer(logger *zap.Logger, e PeerEvent) error {
...
-	if nodeDomainType != dvs.domainType.DomainType() &&
-		nodeDomainType != dvs.domainType.NextDomainType() {
-		return fmt.Errorf("mismatched domain type: %x", nodeDomainType)
+	nodeNextDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyNextDomainType)
+	if err != nil && !errors.Is(err, records.ErrEntryNotFound) {
+		return errors.Wrap(err, "could not read domain type")
+	}
+	if dvs.domainType.DomainType() != nodeDomainType &&
+		dvs.domainType.DomainType() != nodeNextDomainType {
+		return fmt.Errorf("mismatched domain type: %x is neither %x or %x", nodeDomainType, dvs.domainType.DomainType(), dvs.domainType.NextDomainType())
```
y0sher, line 161: just a note we should keep in mind for the future, these messages can pile up and might be hard to search. we might want to have a custom Error type where we can add fields like `want` or `have` so we can print it as {"error message", fields...}. this will make our logs much more machine readable.

y0sher (approved):
good job on this !

nkryuchkov (approved):

PR #1659, opened 2024-08-25
use .validatorProvider.ParticipatingValidators instead of validatorCo…

guym-blox:
…ntroller.AllActiveIndices


y0sher (approved):
lgtm, @olegshmuelov please have a look

moshe-blox:
the reason DS is still using Controller is bc of this:
https://github.com/ssvlabs/ssv/blob/98203ff9e1e5a5c6c480279bb399d3322405092c/operator/validator/controller.go#L861-L864
are we certain this is not needed anymore? @y0sher @olegshmuelov

y0sher:
> the reason DS is still using Controller is bc of this wait:
> 
> https://github.com/ssvlabs/ssv/blob/98203ff9e1e5a5c6c480279bb399d3322405092c/operator/validator/controller.go#L861-L864
> 
> are we certain this is not needed anymore? @y0sher @olegshmuelov

You are right, it might not be enough to use isParticipating, especially for Proposer and SyncCommittee where we need the initial sync and metadata for getting the duties.

y0sher:
We're waiting with this.

Issue #1594, opened 2024-08-08
add ValidatorStore unit tests


moshe-blox:
Bulk add/update tests:

We've recently upgraded the `handleSharesUpdated` method to accept multiple shares, however we did not upgrade unit tests to test it. We should test that when we pass multiple shares, all are updated. Same goes with adding new shares, however such a test might already exist.

Comprehensive index test:
- Add a share with no metadata
- Check that ByIndex(0) does not return it
- Check that both `OperatorValidators` and `Committee` methods return it, and that they don't when given another operator ID and committee ID
- Update it to have a metadata and an index
- Check that `ByIndex` method returns it correctly
- Remove the share 
- Check that `ByIndex` does not return it
- Check that both `OperatorValidators` and `Committee` methods do not return it anymore

If a very similar test already exists, we can probably just upgrade it.

On a side note, we should also make sure that there's at least one test that's testing ValidatorStore through SharesStorage solely (by adding/removing/updating shares) and not by adding validators to ValidatorStore directly.

PR #1668, opened 2024-08-27
Validator Store tests

olegshmuelov:
closes https://github.com/ssvlabs/ssv/issues/1594

### DIFF
--- a/protocol/v2/types/ssvshare.go
+++ b/protocol/v2/types/ssvshare.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"slices"
 	"sort"
+	"sync"
 	"time"
 
 	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
@@ -26,7 +27,9 @@ const (
 type SSVShare struct {
 	spectypes.Share
 	Metadata
-	committeeID *spectypes.CommitteeID
+
+	committeeID     *spectypes.CommitteeID
+	committeeIDOnce sync.Once
 }
 
 // BelongsToOperator checks whether the share belongs to operator.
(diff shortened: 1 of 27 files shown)


protocol/v2/types/ssvshare.go
```diff
@@ -58,16 +60,22 @@ func (s *SSVShare) SetFeeRecipient(feeRecipient bellatrix.ExecutionAddress) {
...
+	// Compute the CommitteeID since it's not yet set.
 	ids := make([]spectypes.OperatorID, len(s.Share.Committee))
 	for i, v := range s.Share.Committee {
 		ids[i] = v.Signer
 	}
 	id := ComputeCommitteeID(ids)
-	s.committeeID = &id
+
+	// Create a new pointer and store it atomically.
+	s.committeeID.Store(&id)
```
y0sher, line 78: committeeID should never change, isn't it better to just have a `New` function to `SSVShare` which sets the committeeID?

y0sher (changes requested):
one comment

registry/storage/mocks/validatorstore.go
```diff
@@ -19,6 +19,158 @@ import (
 	gomock "go.uber.org/mock/gomock"
 )
 
+// MockBaseValidatorStore is a mock of BaseValidatorStore interface.
+type MockBaseValidatorStore struct {
```
moshe-blox, line 23: do we need this mock if its unused?

protocol/v2/types/ssvshare.go
moshe-blox: @y0sher unfortunately Go has weak constructors, so that would pass the responsibility of guaranteeing `Share` is never created without `New` to us, for not much benefit (as far as I can tell)

after we decouple from spec, we can refactor `Share` to an interface backed by a struct with unexported fields to enforce that

registry/storage/validatorstore.go
```diff
@@ -211,8 +224,12 @@ func (c *validatorStore) handleSharesAdded(shares ...*types.SSVShare) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	// Update byValidatorIndex
 	for _, share := range shares {
+		if share == nil {
+			continue
```
moshe-blox, line 229: this should never be allowed
we should make sure to return error in sharesStorage that if a passed share is nil

y0sher (approved):

moshe-blox (approved):
well done 💪

PR #1674, opened 2024-08-28
fix: don't read post-consensus message before decided

### LINKED ISSUE #1671: Rejecting post-consensus messages because no decided value, dropping the message
### Description 

In `CommitteeRunner`
We have witnessed that sometimes, a commit quorum is achieved on some nodes and they are sending post-consensus right away, though our node haven't received or processed the last commit message yet and processes the post-consensus before, ending up throwing it away because there is no decided value yet.

### Task

Filter the messages out of the queue and don't pop post-consensus messages before we have decided value.

y0sher:
Closes #1671

### DIFF
--- a/protocol/v2/ssv/validator/committee_queue.go
+++ b/protocol/v2/ssv/validator/committee_queue.go
@@ -109,17 +109,26 @@ func (v *Committee) ConsumeQueue(
 			}
 		} else if runningInstance != nil && runningInstance.State.ProposalAcceptedForCurrentRound == nil {
 			// If no proposal was accepted for the current round, skip prepare & commit messages
-			// for the current height and round.
+			// for the current round.
 			filter = func(m *queue.SSVMessage) bool {
 				sm, ok := m.Body.(*specqbft.Message)
 				if !ok {
+					return false
+				}
+
+				if sm.Round != state.Round { // allow next round or change round messages.
 					return true
 				}
 
-				if sm.Height != state.Height || sm.Round != state.Round {
-					return true
-				}
 				return sm.MsgType != specqbft.PrepareMsgType && sm.MsgType != specqbft.CommitMsgType
+			}
+		} else if runningInstance != nil && !runningInstance.State.Decided {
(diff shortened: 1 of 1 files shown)


moshe-blox (approved):

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -109,18 +109,24 @@ func (v *Committee) ConsumeQueue(
 			}
 		} else if runningInstance != nil && runningInstance.State.ProposalAcceptedForCurrentRound == nil {
 			// If no proposal was accepted for the current round, skip prepare & commit messages
-			// for the current height and round.
+			// for the current round.
 			filter = func(m *queue.SSVMessage) bool {
 				sm, ok := m.Body.(*specqbft.Message)
 				if !ok {
-					return true
+					return false
```
y0sher, line 116: I'm guessing this is wrong, we do want to prevent post-consensus at this stage like I intended, but this condition also applies to Event messages, while we know `ExecuteDuty` will not happen here as we move it to run through `StartDuty` we can still get a Timeout event message.

y0sher:
> Looks good, but maybe we should have a same for all other queues as I saw same behaviour even with genesis runners

Well its a bit more complicated in genesis (or alan roles that are not committee) since this queue is used for all slots and duty types. this role works for single committee runner.

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -98,29 +97,29 @@ func (v *Committee) ConsumeQueue(
...
-			// for the current height and round.
+			// for the current round.
 			filter = func(m *queue.SSVMessage) bool {
 				sm, ok := m.Body.(*specqbft.Message)
 				if !ok {
-					return true
+					if m.MsgType != spectypes.SSVPartialSignatureMsgType {
+						return true
+					}
+					return false
```
olegshmuelov, line 109: return m.MsgType != spectypes.SSVPartialSignatureMsgType

moshe-blox (approved):

PR #1675, opened 2024-08-29
fix: ekm deadlock

y0sher:
### Description

Updated to version fixed in https://github.com/ssvlabs/eth2-key-manager/pull/108

### TODO

 Update to official tag when its merged.

Thanks to @moshe-blox and @vaclav-ssvlabs, @zktaiga


moshe-blox (approved):
gj 💪

PR #1678, opened 2024-08-30
fix: (queue) nil return condition in Pop

MatusKysel:
There is currently bug in queue that returns nil even though queue has more msg inside.
So basic flow is:
- we receive msgA into the queue that's not matching filter
- we now receive another one msgB that's will match the filer
- after breaking from main loop in `Pop` we read msg again
- we get again non filter matching one msgC
- in `pop` if msgB is true on filter but doesn't have prio over msgC we will return nil

There is test simulating this scenario


AKorpusenko (approved):
Great!

moshe-blox (approved):
well done 💪

PR #1653, opened 2024-08-25
logs: (alan) add dutyID to msg validation ignore/rejects

y0sher:
### Description

In debugging we a missed duty we use DutyID as a way to look on all logs related to this duty, we know that sometimes messages are dropped in the P2P layer but currently its not easy to connect the consensus process messages to the p2p layer logs about rejecting/ignoring messages.

### Change

This change adds a duty ID field to msg validation logs about ignores/rejects, it checks whether the logger is on debug level to do so, since to make up the duty id, a call to validatorStore is required and we want to prevent it if not necessary.

PR #1685, opened 2024-09-02
fix: duties log string for committee duties

y0sher:
### Description 

#1653 introduced a logging bug where duties in the `duties` field are formatted using the runner role, means they all get "COMITTTEE" when they are committee duties. We want to support any duty role type so a string might fix better this parameter to enable flexibility.


nkryuchkov (changes requested):
Can we remove `GenesisFormatDutyID` and use the modified `FormatDutyID` instead?

nkryuchkov (approved):

moshe-blox (approved):

PR #1689, opened 2024-09-02
fix: (EventHandler) update non-committee shares upon liquidation/reactivation

moshe-blox:
Reopened from #1220


y0sher (approved):

moshe-blox:
since this change may cause upgraded nodes to reject non-upgrades nodes because they aren't seeing that a certain validator is liquidated, @y0sher noted that we should activate this new behavior on the fork epoch

eth/eventhandler/handlers.go
```diff
@@ -567,7 +567,7 @@ func (eh *EventHandler) processClusterEvent(
 		}
 	}
 
-	return toUpdate, updatedPubKeys, nil
+	return operatorShares, operatorValidatorPubKeys, nil
```
iurii-ssv, line 570: From the looks of it, `processClusterEvent` func does 2 things:
- updates shares
- returns a list of operator (own) shares & pub-keys

can we split this `func` in 2 then to make it clear these things aren't related ?

iurii-ssv: I guess this applies to all `EventHandler` functions

y0sher (approved):

moshe-blox:
LGTM, just need to finish QA with Andrew on both in-committee and out-of-committee validators to confirm it still works before merging

eth/eventhandler/handlers.go
moshe-blox: agree, but i think just adding one comment above each step would make it clear enough

olegshmuelov (approved):

PR #1699, opened 2024-09-05
chore: smaller format `invalid_messages` logs

AKorpusenko:


network/topics/scoring.go
```diff
@@ -249,3 +253,13 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
 	}
 	return topicCommittees
 }
+
+// truncateStats returns the subnets in a small format topicNum=ti,fmd,mmd,imd
+func truncateStats(filtered map[string]*pubsub.TopicScoreSnapshot) []string {
+	filteredStrings := make([]string, 0, len(filtered))
+	for topic, snapshot := range filtered {
+		formatted := fmt.Sprintf("%s=%0.3f;%0.3f;%0.3f;", topic, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
```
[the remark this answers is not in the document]
AKorpusenko: Done

network/topics/scoring.go
```diff
@@ -249,3 +253,13 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
...
+
+// truncateStats returns the subnets in a small format topicNum=ti,fmd,mmd,imd
+func truncateStats(filtered map[string]*pubsub.TopicScoreSnapshot) []string {
+	filteredStrings := make([]string, 0, len(filtered))
+	for topic, snapshot := range filtered {
+		formatted := fmt.Sprintf("%s=%d;%0.3f;%0.3f;%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
+		filteredStrings = append(filteredStrings, formatted)
+	}
+	return filteredStrings
+}
```
MatheusFranco99, line 265: I think @moshe-blox proposed an even further reduction by setting it all in a single string, i.e. applying a join with " " for the list of strings that this function returns. Is that right @moshe-blox?

network/topics/scoring.go
```diff
@@ -249,3 +253,13 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
 	}
 	return topicCommittees
 }
+
+// truncateStats returns the subnets in a small format topicNum=ti,fmd,mmd,imd
+func truncateStats(filtered map[string]*pubsub.TopicScoreSnapshot) []string {
```
moshe-blox, line 258: just to be a bit more specific (i know long names are not ideal but i think they're very helpful if they're only used once)
```suggestion
func formatInvalidMessageStats(filtered map[string]*pubsub.TopicScoreSnapshot) []string {
```

network/topics/scoring.go
```diff
@@ -249,3 +253,13 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
...
+	filteredStrings := make([]string, 0, len(filtered))
+	for topic, snapshot := range filtered {
+		formatted := fmt.Sprintf("%s=%d;%0.3f;%0.3f;%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
+		filteredStrings = append(filteredStrings, formatted)
+	}
+	return filteredStrings
```
moshe-blox, line 264: Instead of returning []string, which would add about 4 characters per item (quotes and commas), we should try to compose one long string separated with just " " in order to take the least amount of characters:

Using `strings.Builder` is the most efficient way to build big strings in Go:
```suggestion
	var b strings.Builder
	i := 0
	for topic, snapshot := range filtered {
		if i > 0 {
			b.WriteString(" ")
		}
		fmt.Fprintf(b, "%s=%d,%0.3f,%0.3f,%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
		i++
	}
	return b.String()
```
(it allocates much less memory than adding strings to slice and later using strings.Join)

network/topics/scoring.go
```diff
@@ -147,7 +151,7 @@ func scoreInspector(logger *zap.Logger, scoreIdx peers.ScoreIndex, logFrequency
 				zap.Float64("w6_ip_colocation_factor", w6),
 				zap.Float64("p7_behaviour_penalty", p7),
 				zap.Float64("w7_behaviour_penalty", w7),
-				zap.Any("invalid_messages", filtered),
+				zap.Strings("invalid_messages", invalidMessagesStats),
 			}
```
MatheusFranco99, line 155: A little bit above, we include a `p2_first_message_deliveries` field. This variable is a list of structures and it's also very big. It was introduced for debugging for scoring analysis some weeks ago, but I think now we can change it to be the sum through topics. I.e., the summation of `snapshot.FirstMessageDeliveries`*`topicScoreParams.FirstMessageDeliveriesWeight` through topics. If it's confusing, reach me so we can discuss it :)

MatheusFranco99 (changes requested):
Good job! Raising 2 possible improvements

moshe-blox (changes requested):

network/topics/scoring.go
moshe-blox: @MatheusFranco99 yeah i agree: https://github.com/ssvlabs/ssv/pull/1699#discussion_r1745627722

network/topics/scoring.go
AKorpusenko: Nice. Lets do it also

network/topics/scoring.go
```diff
@@ -254,12 +255,16 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
...
-		formatted := fmt.Sprintf("%s=%d;%0.3f;%0.3f;%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
-		filteredStrings = append(filteredStrings, formatted)
+		if i > 0 {
+			b.WriteString(" ")
+		}
+		fmt.Fprintf(&b, "%s=%d,%0.3f,%0.3f,%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
+		i++
 	}
-	return filteredStrings
+	return b.String()
```
moshe-blox, line 269: i forgot, this currently prints 3 decimals points even if the number is round, so that beats the purpose of us trying to reduce the total length

lets try this?

```go
// formatInvalidMessageStats returns the subnets in a small format topicNum=ti,fmd,mmd,imd
func formatInvalidMessageStats(filtered map[string]*pubsub.TopicScoreSnapshot) string {
	fmtFloat := func(n float64) string {
		if math.Trunc(n) == n {
			return strconv.FormatInt(int64(n), 10)
		}
	}
	var b strings.Builder
	i := 0
	for topic, snapshot := range filtered {
		if i > 0 {
			b.WriteString(" ")
		}
		fmt.Fprintf(&b, "%s=%d,%s,%s,%s", topic, fmtFloat(snapshot.TimeInMesh), fmtFloat(snapshot.FirstMessageDeliveries), fmtFloat(snapshot.MeshMessageDeliveries), fmtFloat(snapshot.InvalidMessageDeliveries))
		i++
	}
	return b.String()
}
```

this should print without decimals if the number is round or with 2 decimals if not

on another note i think numbers are always round so maybe we can even convert them to integers instead? @MatheusFranco99

moshe-blox (changes requested):

network/topics/scoring.go
MatheusFranco99: Great!!
TimeInMesh is round but the other counters suffer decay so it may really become real floats 😅 
I agree with adding this, good job!

network/topics/scoring.go
```diff
@@ -249,3 +254,17 @@ func filterCommitteesForTopic(topic string, committees []*storage.Committee) []*
...
+
+// formatInvalidMessageStats returns the subnets in a small format topicNum=ti,fmd,mmd,imd
+func formatInvalidMessageStats(filtered map[string]*pubsub.TopicScoreSnapshot) string {
+	var b strings.Builder
+	i := 0
+	for topic, snapshot := range filtered {
+		if i > 0 {
+			b.WriteString(" ")
+		}
+		fmt.Fprintf(&b, "%s=%d,%0.3f,%0.3f,%0.3f", topic, snapshot.TimeInMesh, snapshot.FirstMessageDeliveries, snapshot.MeshMessageDeliveries, snapshot.InvalidMessageDeliveries)
```
nkryuchkov, line 266: `%d` prints the number of nanoseconds for `time.Duration` which is less readable and longer than `%s`, which represents the duration in a compact format. Or is there any reason to use `%d` instead?

network/topics/scoring_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"testing"
+
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTruncateStats(t *testing.T) {
+	filtered := map[string]*pubsub.TopicScoreSnapshot{
+		"ssv.v2.103": {
+			TimeInMesh:               0,
```
nkryuchkov, line 13: Can we test both zero and non-zero values for all the fields to test all of them?

y0sher (approved):

y0sher (approved):

moshe-blox (approved):

MatheusFranco99 (approved):
Good job!
The P2 log may also be reduced. It's very long and was made that way to debug scoring. If you guys want I can fix it here or in another PR
@y0sher @moshe-blox

moshe-blox:
@MatheusFranco99 ur right i forgot about it
@AKorpusenko can you please do the same formatting for the p2 log?
image

moshe-blox (approved):

Issue #1692, opened 2024-09-03
Data race in TestScheduler_Committee_Early_Block


AKorpusenko:
## Description

There was a failure of `TestScheduler_Committee_Early_Block` on run https://github.com/ssvlabs/ssv/actions/runs/10679665619/job/29599394867

Issue #1666, opened 2024-08-27
High char amount within a log line


zktaiga:
Some log lines are emitting a large amount of chars. This might make it more difficult to generate bloom blocks for log engines.

We should look into encoding this info.

An example of 13k+ chars for a single log line:
```json
{
  "level": "\u001b[35mDEBUG\u001b[0m",
  "time": "2024-08-27T13:49:42.993093Z",
  "name": "P2PNetwork",
  "msg": "peer scores",
  "selfPeer": "16Uiu2HAmJdFEVYpJ6N1LwyUwh3rz1Yu2RAxoaEt1wQXUPQN6yqwL",
  "peer_id": "16Uiu2HAmBzqZrJF37ZAuHz3FkXTYVH2jSy5ANk7dkH3pXXegHcnc",
  "peer_score": "-15600",
  "invalid_messages": {
    "ssv.v2.1": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 1
    },
    "ssv.v2.103": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 3
    },
    "ssv.v2.107": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 3
    },
    "ssv.v2.109": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 1
    },
    "ssv.v2.110": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 1
    },
    "ssv.v2.114": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 1
    },
    "ssv.v2.117": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
      "MeshMessageDeliveries": 0,
      "InvalidMessageDeliveries": 2
    },
    "ssv.v2.118": {
      "TimeInMesh": 0,
      "FirstMessageDeliveries": 0,
[...]

PR #1704, opened 2024-09-05
fix: committee duty id log

olegshmuelov:
Fix committee duty ID logging.

 - The PR introducing a fix to #1711 and his duplicate (https://github.com/ssvlabs/ssv/issues/1692) 
 - this commit https://github.com/ssvlabs/ssv/pull/1704/commits/992a85accb349cd41ac885b272bb10b140e21424
resolves partially  https://github.com/ssvlabs/ssv/issues/1666

### DIFF
--- a/operator/duties/committee.go
+++ b/operator/duties/committee.go
@@ -133,7 +133,7 @@ func (h *CommitteeHandler) appendBeaconDuty(vc validatorCommitteeDutyMap, c comm
 
 	cd, ok := c[committee.id]
 	if !ok {
-		c[committee.id] = &committeeDuty{
+		cd = &committeeDuty{
 			id:          committee.id,
 			operatorIDs: committee.operatorIDs,
 			duty: &spectypes.CommitteeDuty{
@@ -141,6 +141,7 @@ func (h *CommitteeHandler) appendBeaconDuty(vc validatorCommitteeDutyMap, c comm
 				ValidatorDuties: make([]*spectypes.ValidatorDuty, 0),
 			},
 		}
+		c[committee.id] = cd
 	}
 
 	cd.duty.ValidatorDuties = append(c[committee.id].duty.ValidatorDuties, beaconDuty)
--- a/operator/duties/committee.go
+++ b/operator/duties/committee.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ssvlabs/ssv/operator/duties/dutystore"
 )
 
+type validatorCommitteeDutyMap map[phase0.ValidatorIndex]*committeeDuty
 type committeeDutiesMap map[spectypes.CommitteeID]*committeeDuty
 
 type CommitteeHandler struct {
@@ -19,8 +20,6 @@ type CommitteeHandler struct {
 
 	attDuties  *dutystore.Duties[eth2apiv1.AttesterDuty]
 	syncDuties *dutystore.SyncCommitteeDuties
-
-	validatorCommitteeDutyMap map[phase0.ValidatorIndex]*committeeDuty
 }
 
 type committeeDuty struct {
@@ -33,8 +32,6 @@ func NewCommitteeHandler(attDuties *dutystore.Duties[eth2apiv1.AttesterDuty], sy
 	h := &CommitteeHandler{
 		attDuties:  attDuties,
 		syncDuties: syncDuties,
-
-		validatorCommitteeDutyMap: make(map[phase0.ValidatorIndex]*committeeDuty),
 	}
 
 	return h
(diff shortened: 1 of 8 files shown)


operator/duties/committee.go
```diff
@@ -12,23 +12,29 @@ import (
...
 	baseHandler
 
 	attDuties  *dutystore.Duties[eth2apiv1.AttesterDuty]
 	syncDuties *dutystore.SyncCommitteeDuties
 
-	validatorCommitteeIDs map[phase0.ValidatorIndex]spectypes.CommitteeID
+	validatorCommitteeDutyMap map[phase0.ValidatorIndex]*committeeDuty
+}
+
+type committeeDuty struct {
```
nkryuchkov, line 26: Do I understand correctly that 
- we never use `duty` in `validatorCommitteeDutyMap`, although the type is called `committeeDuty`;
- `id` is redundant in `committeeDutiesMap`?

Should we perhaps use different types for them?

olegshmuelov: the `duty` is indeed used in the `ExecuteCommitteeDuties` function, where it's executed via `committee.duty`.

As for the `id` field in `committeeDutiesMap`,  it simplifies access and lookup by avoiding recalculating or passing it around during the building of duties, improving performance.

operator/duties/committee.go
```diff
@@ -91,7 +97,10 @@ func (h *CommitteeHandler) buildCommitteeDuties(attDuties []*eth2apiv1.AttesterD
 	// This approach reduces contention and improves performance, as multiple individual calls would be slower.
 	vs := h.validatorProvider.SelfParticipatingValidators(epoch)
 	for _, v := range vs {
-		h.validatorCommitteeIDs[v.ValidatorIndex] = v.CommitteeID()
+		h.validatorCommitteeDutyMap[v.ValidatorIndex] = &committeeDuty{
+			id:          v.CommitteeID(),
+			operatorIDs: v.OperatorIDs(),
+		}
```
y0sher, line 103: Probably out of this PR scope but still:
Since this is loaded on each slot with committee duties, I don't see a point in storing this? If you would've loaded it each epoch it will make more sense I guess, but then you need to reload probably on indices  change or reorg. So maybe storing this doesn't add much value ( since managing is complicated ). right now I don't see if this map is every evacuated for example. (validator removed, etc..)

y0sher (approved):
LGTM

y0sher (approved):

operator/duties/committee.go
```diff
@@ -90,42 +93,56 @@ func (h *CommitteeHandler) buildCommitteeDuties(attDuties []*eth2apiv1.AttesterD
...
-		h.logger.Error("can't find validator committeeID in validator store", zap.Uint64("validator_index", uint64(beaconDuty.ValidatorIndex)))
+		h.logger.Error("failed to find committee for validator", zap.Uint64("validator_index", uint64(beaconDuty.ValidatorIndex)))
 		return
 	}
 
-	if _, ok := m[committeeID]; !ok {
-		m[committeeID] = &spectypes.CommitteeDuty{
-			Slot:            beaconDuty.Slot,
-			ValidatorDuties: make([]*spectypes.ValidatorDuty, 0),
+	if _, ok := c[committee.id]; !ok {
```
y0sher, line 134:
```suggestion
	cdm, ok := c[committee.id]; 
	if !ok {
```

operator/duties/committee.go
```diff
@@ -90,42 +93,56 @@ func (h *CommitteeHandler) buildCommitteeDuties(attDuties []*eth2apiv1.AttesterD
...
+			operatorIDs: committee.operatorIDs,
+			duty: &spectypes.CommitteeDuty{
+				Slot:            beaconDuty.Slot,
+				ValidatorDuties: make([]*spectypes.ValidatorDuty, 0),
+			},
 		}
 	}
-	m[committeeID].ValidatorDuties = append(m[committeeID].ValidatorDuties, beaconDuty)
+
+	c[committee.id].duty.ValidatorDuties = append(c[committee.id].duty.ValidatorDuties, beaconDuty)
```
y0sher, line 145:
```suggestion
	cdm.duty.ValidatorDuties = append(c[committee.id].duty.ValidatorDuties, beaconDuty)
```

y0sher (approved):

y0sher (approved):

y0sher (approved):

operator/duties/scheduler.go
```diff
@@ -414,18 +414,20 @@ func (s *Scheduler) ExecuteDuties(logger *zap.Logger, duties []*spectypes.Valida
...
 func (s *Scheduler) ExecuteCommitteeDuties(logger *zap.Logger, duties committeeDutiesMap) {
-	for committeeID, duty := range duties {
-		committeeID := committeeID
-		duty := duty
-		logger := s.loggerWithCommitteeDutyContext(logger, committeeID, duty)
+	for _, committee := range duties {
+		duty := committee.duty
+		logger := s.loggerWithCommitteeDutyContext(logger, committee)
+		dutyEpoch := s.network.Beacon.EstimatedEpochAtSlot(duty.Slot)
+		logger.Debug("🔧 executing committee duty", fields.Duties(dutyEpoch, duty.ValidatorDuties))
```
moshe-blox, line 421: is this log necessary given theres the "starting duty processing" log [here](https://github.com/ssvlabs/ssv/blob/f969da086db0b9e33d6eb1ae7b8076dd0be1e1b8/protocol/v2/ssv/validator/committee.go#L207) which can receive the "duties" field also?

moshe-blox: discussed with @olegshmuelov and it would be simpler to rely on this log rather than passing `duties` down

olegshmuelov: The purpose of this PR is to provide visibility at each step from execution to submission, specifically to track which duties are scheduled to be executed in a given slot. If an error occurs between the moment duties are sent for execution and the "starting duty processing" log, we wouldn't have any record of what duties were actually sent for execution. This log ensures we have that information available.

moshe-blox (approved):
well done 👍

Issue #1662, opened 2024-08-26
drop peers with low score or irrelevant subnets


moshe-blox:
On the fork epoch, operator nodes unsubscribe from subnets which are no longer relevant. This means some of the connections made before the fork might become irrelevant, if we no longer share subnets with them.

Beacon nodes running few validators very often change their subnets, so they should a similar same problem.

We should look into how Beacon clients handle this issue before deciding how to go about it for SSV.

### Proposal
We can call `PubSub.ListPeers(topic)` periodically for each topic we're subscribed to and with it populate a map of peer IDs to their subscribed topics, and finally use this map to disconnect from peers with 0 shared subnets, and potentially also in AppSpecificScore to rank each peer according to how many subnets they share with us.

MatheusFranco99:
With a second read, it seems that **indeed GossipSub doesn't disconnect the bad peers**. Basically, it only drops the peer if it's dead or blacklisted.

**Prysm** prunes peers by using a [forever loop in which it checks if there are any bad peers and disconnects if so](https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/sync/rpc_status.go#L29-L52) ([sendGoodbyeAndDisconnect function](https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/sync/rpc_goodbye.go#L55-L94), [disconnect function](https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/p2p/service.go#L338-L341)). It also has a ["sanity" like prune method](https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/p2p/peers/status.go#L580-L636).

An **issue** is that **GossipSub must be alerted** of this peer removal to safely store the peer's score (in case it reconnects later). I'll check how this can be ensured.

Regarding 0-shared subnets, I think GossipSub naturally keeps track of "fanout" topics and "fanout" peers. This means that there could be two forces indefinitely pushing each other. But, definitely, we can drop some in case we are full or near full.

Regarding app-specific score, again this may impact these "fanout" peers who may join later on and I'm concerned with unpredictable behaviours this may produce. I would stick with the disconnection for now.

PR #1707, opened 2024-09-06
feat: (p2p) drop bad or irrelevant peers

MatheusFranco99:
# Overview

This PR adds a mechanism to disconnect from bad peers and irrelevant peers that don't share any subnet.

To do this:
- We added a `GossipSubScoreIndex` interface to the `Index` interface and added an implementation for it (`gossipSubScoreIndex`), which simply stores the scores in a map.
- This structure is filled through the `scoreInspector` function. Notice that this function gains a higher importance than just logging. Also, on every function call, the `gossipSubScoreIndex ` is updated, differently from the logging feature that is executed only for some function calls.
- Once `p2pNetwork.peersBalancing()` is called, in case we have the maximum number of connected peers, we will:
  - disconnect from bad peers, using the `GossipSubScoreIndex`
  - If no peers got disconnected, we then disconnect from irrelevant peers (that have no common subnets)
  - If no peers got disconnected, we then disconnect according to the previous peers balancing approach (weighting subnets and scoring peers)

Closes https://github.com/ssvlabs/ssv/issues/1662.


network/peers/conn_manager.go
```diff
@@ -192,3 +200,43 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+	return net.ClosePeer(peerID)
+}
+
+// DisconnectFromBadPeers will disconnect from bad peers according to the bad peers collector
+func (c connManager) DisconnectFromBadPeers(logger *zap.Logger, net libp2pnetwork.Network, allPeers []peer.ID, badPeersCollector BadPeersCollector) {
+	for _, peerID := range allPeers {
+		if isBad, score := badPeersCollector.IsBad(peerID); isBad {
+			err := c.disconnect(peerID, net)
+			if err != nil {
+				logger.Error("Couldn't disconnect from bad peer", zap.String("peer", string(peerID)), zap.Float64("score", score))
```
nkryuchkov, line 215:
```suggestion
				logger.Error("Couldn't disconnect from bad peer", fields.PeerID(peerID), zap.Float64("score", score))
```

nkryuchkov (approved):

network/peers/bad_peers_collector.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (c *badPeersCollector) RegisterBadPeer(peerID peer.ID, score float64) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	c.badPeers[peerID] = score
+}
+
+func (c *badPeersCollector) IsBad(peerID peer.ID) (bool, float64) {
```
y0sher, line 40: you need to use the mutex to protect here as well.

y0sher (changes requested):
Good job! this looks good. 
1 comment about locking the map.

And another point - 
I'm thinking we should probably use the `badPeersCollector` at more places like - 
- filtering nodes with bad score from discovery
- not accepting incoming connections from peers that are "bad" (not connecting of course as well)

This might be out of the scope of this PR, as it will require a different approach to managing the `badPeers` map ( not clearing it each time scoreInspector runs.. ), keeping the bad peer indicator even after we are not connected to a peer (with some decay). making this essentially a banning mechanism.

@MatheusFranco99 , @moshe-blox let me know what you guys think about if and when we should try going further with it.

moshe-blox:
@MatheusFranco99 i noticed in Prysm that they would blacklist "bad peers" in many layers such as discovery, handshake and connection gater. You can see that if you search for usages of the [Status.IsBad](https://github.com/prysmaticlabs/prysm/blob/b7b017f5b607f04d4e9056a1ec8a6851a9c7da29/beacon-chain/p2p/peers/status.go#L330-L334) method, and maybe usages of other "bad peer" related funcs.

i think probably we should track scores in our PeerIndex because this one is passed to various layers of our code already, and in order to not have to pass all of it to layers which don't require it such as discovery, then it could expose a `BadPeerer` interface or a `PeerScorer` interface with only the methods relevant for what those other layers require.

MatheusFranco99:
@moshe-blox @y0sher 
I've implemented the changes we discussed. Let me know what you guys think

networkconfig/holesky-stage.go
```diff
@@ -16,7 +16,7 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        99999999,
+	AlanForkEpoch:        78286,
```
anatollupacescu, line 19: do we want to move this away from pre-fork on stage?

network/peers/conn_manager.go
```diff
@@ -192,3 +208,60 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+
+	disconnectedPeers := 0
+
+	for _, peerID := range allPeers {
+
+		// Check if peer is bad
+		if c.gossipSubScoreIndex.HasBadGossipSubScore(logger, peerID) {
+
+			// Get GossipSub score for logging
+			gossipSubScore, hasScore := c.gossipSubScoreIndex.GetGossipSubScore(peerID)
```
y0sher, line 223: I think this could be better as one call. 
So something like  `gossipSubScore, hasScore := c.gossipSubScoreIndex.GetGossipSubScore(peerID)` and then calculate if `gossipSubScore` is over a threshold for being bad. 
not too bad but I don't see a reason searching the peer in the map twice..

y0sher (changes requested):
Implementation looks good. But we still missing filtering based on the score in handshake, conngater, and discovery.

networkconfig/holesky-stage.go
MatheusFranco99: This is just a temporary change so I can do some tests in stage 😅 
I'll remove it after the testing is done

network/peers/conn_manager.go
MatheusFranco99: Indeed, I'll fix that
Thanks

y0sher:
@MatheusFranco99 using `connGater` we should be able to stop connection to peers even before libp2p has created gossipsub entry for them, checkout this file `network/peers/connections/conn_gater.go` I think specifically `InterceptSecured` and for handshake yeah it can be done with the `HandshakeFilter`.

MatheusFranco99:
@y0sher Added

networkconfig/holesky-stage.go
y0sher: A note for everyone for future testing. Instead of editing your branch - create a new branch called `test/youbranchname` or `deploy/youbranchname` and deploy it instead polluting your branch with irrelevant changes you'll need to revert..

MatheusFranco99: Got it! Thanks

y0sher (approved):

network/peers/conn_manager.go
```diff
@@ -192,3 +208,57 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+
+	for _, peerID := range allPeers {
+
+		// Check if peer is bad
+		if isBad, gossipSubScore := c.gossipSubScoreIndex.HasBadGossipSubScore(peerID); isBad {
+
+			// Disconnect
+			err := c.disconnect(peerID, net)
+			if err != nil {
+				logger.Error("Couldn't disconnect from bad peer", fields.PeerID(peerID), zap.Float64("GossipSub Score", gossipSubScore))
```
moshe-blox, line 225: for consistency with the rest of the codebase: lower case logs with preference for "failed"
```suggestion
				logger.Error("failed to disconnect from bad peer", fields.PeerID(peerID), zap.Float64("GossipSub Score", gossipSubScore))
```

network/peers/conn_manager.go
```diff
@@ -192,3 +208,57 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+			} else {
+				logger.Debug("Disconnecting from bad peer", fields.PeerID(peerID), zap.Float64("GossipSub Score", gossipSubScore))
```
moshe-blox, line 227: snake case log fields
```suggestion
				logger.Debug("disconnecting from bad peer", fields.PeerID(peerID), zap.Float64("gossip_score", gossipSubScore))
```

network/peers/conn_manager.go
```diff
@@ -192,3 +208,57 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+		peerSubnets := c.subnetsIdx.GetPeerSubnets(peerID)
+
+		// Get shared subnets
+		sharedSubnets := records.SharedSubnets(mySubnets, peerSubnets, len(mySubnets))
+
+		// If there's no common subnet, disconnects
+		if len(sharedSubnets) == 0 {
+			err := c.disconnect(peerID, net)
+			if err != nil {
+				logger.Error("Couldn't disconnect from peer with irrelevant subnets", fields.PeerID(peerID))
```
moshe-blox, line 252:
```suggestion
				logger.Error("failed to disconnect from peer with irrelevant subnets", fields.PeerID(peerID))
```

network/peers/conn_manager.go
```diff
@@ -192,3 +208,57 @@ func scorePeer(peerSubnets records.Subnets, subnetsScores []float64) PeerScore {
...
+			} else {
+				logger.Debug("Disconnecting from peer with irrelevant subnets", fields.PeerID(peerID))
```
moshe-blox, line 254:
```suggestion
				logger.Debug("disconnecting from peer with irrelevant subnets", fields.PeerID(peerID))
```

moshe-blox (changes requested):
very well done 💪

moshe-blox:
@MatheusFranco99 what's the thought process of checking this in InterceptSecured, when the handshake already checks this around this stage? not saying we shouldn't just curious 

btw i've added a missing check in InterceptAddrDial: https://github.com/ssvlabs/ssv/pull/1707/commits/477ee0abc5ff8a76911c190003ac320f4edfd438

moshe-blox (approved):

MatheusFranco99:
@moshe-blox 
Tbh, I don't understand this part very well. I've added it according to the above [comment](https://github.com/ssvlabs/ssv/pull/1707#issuecomment-2341306786). We can adjust it as needed

Issue #1664, opened 2024-08-27
Discovery forking mechanism 


y0sher:
### Description 

We witnessed in mainnet and our stage environment that ETH nodes are filling our discovery routing tables and that we have to go through some eth nodes before finding a SSV node when selecting from the nodes we have in our discovery, this in turn slows down the process to connect to SSV nodes, reducing network connectivity and might hurt our performance. 

### Solution 
Through experiments in our staging ENV we discovered the `ProtocolID` property in DiscV5, while setting this to be `SSVV5` we could block all other protocols that use similar discovery but not running on SSV (mainly the ETH nodes mentioned above).
We implemented this in our stage branch and it works once deployed on both all nodes and bootnode.

### Task 

This task is about getting ready to switch mainnet to use this approach, this is a part of forking to Alan.

- Run two instances of discV5, one of them should work with protocol ID and one without. (Protocol ID can be the domaintype, might need padding)
- use structure similar to discv5's sharedUDPConn to handle the messages passed to `Unhandled` (a config property of discv5) channel which message that could not be handled is passed into. 
- On fork epoch stop listening to the old (no protocol id) version of our discv5, or at least don't start if we are after fork epoch.
- Run 2 bootnodes, one with and one without protocol id and use both. 
- task forking on stage

moshe-blox:
> Protocol ID can be the domaintype
this would be a bit of a problem for us when we want to test multiple domain types in parallel
we can have it NetworkConfig, or if we think its a bit unrelated there, then we can use the domain type unless the config.yaml  overrides it

MatheusFranco99:
We were probably allowing nodes from other networks/protocols due to [this](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md#comparison-with-other-discovery-mechanisms):
> The Node Discovery Protocol doesn't rely on a single operator and places a small amount of trust in every participant. It becomes more resistant to censorship as the size of the network increases and participants of multiple distinct peer-to-peer networks can share the discovery network to further increase its resilience.

A bigger network is desirable for discovery because it makes it more robust (though it's never safe). However, it shouldn't be a problem since we have some hundreds of operators already.

I'm confused about using the ProtocolID as the DomainType, and not simply using a constant tag like `SSV`. On a fork epoch, all handshakes would start again, all topic advertisements, and so on. But the discovery shouldn't suffer this because of protocol changes, no?
Wdyt @y0sher @moshe-blox ?

[...]

PR #1709, opened 2024-09-07
discovery: fork with protocol ID

nkryuchkov:
# Overview

This PR adds the following changes:
- Before the Alan fork, the node runs two discV5 instances, one with protocol ID restriction and one without
- Implement a discovery fork listener that manages pre- and post-fork discV5 listeners
- Implement UDP connection sharing between pre- and post-fork listeners
- Pass protocol ID for the discovery layer from the network config
- After the Alan fork, the pre-fork listener is closed
- Upon a restart, after the Alan fork, only the post-fork discovery service is started

Closes https://github.com/ssvlabs/ssv/issues/1664

## Tests

- [x] `preAndPostForkIterator`:
  - [x] for nonempty pre- and post-fork
  - [x] for empty post-fork
  - [x] for empty pre-fork
  - [x] `Close`
- [x] `ForkListener` considering pre-fork and post-fork cases for:
  - [x] Create
  - [x] `Lookup`
  - [x] `RandomNodes`
  - [x] `AllNodes`
  - [x] `Ping`
  - [x] `LocalNode`
  - [x] `Close`
- [x] DiscV5Service: creation (and initialization)
- [x] DiscV5Service: `Close`
- [x] DiscV5Service: `RegisterSubnets`
- [x] DiscV5Service: `DeregisterSubnets`
- [x] DiscV5Service: `PublishENR`
- [x] DiscV5Service: `Bootstrap`
- [x] DiscV5Service: `Node` for known and unknown peers
- [x] DiscV5Service: `checkPeer` for:
  - [x] Valid peer
  - [x] Peer with no domain
  - [x] Peer with no next domain
  - [x] Peer with a matching domain (current and next)
  - [x] Peer with mismatching domains
  - [x] Peer with no subnets
  - [x] Peer with zero subnets
  - [x] Valid peer but with limit reached
  - [x] Valid peer but without any common subnets
- [x] DiscV5Service listener created type for pre-fork and post-fork


y0sher (changes requested):
Good job @nkryuchkov , @MatheusFranco99 
a few points:
1. Please notice that bootnode has a bit different setup since it has its own config, you might want to use the types you setup here there in order to re-use code. currently bootnode sets up its own discovery using the discv5 lib, you can see the code here - 
https://github.com/ssvlabs/ssv/blob/235ffdc4d1615416487147424f44078ed3ea9bdd/utils/boot_node/node.go
2. I think we can go ahead with stopping to listen to discovery without protocol ID after the fork. at minimum we can do it after there's a restart (so not starting the pre-fork one on a fresh start that is post-fork). For this we probably need to discuss with @moshe-blox @liorrutenberg Also this means a lot of testing :)
3. I think we can add a few UT to see the two connections behave as we expect. at least for the shareConn struct.

MatheusFranco99:
Yes. In order to make more changes, first, it's better to reach an agreement on when we will deprecate the previous discovery. Will we do as [2. of the above comment](https://github.com/ssvlabs/ssv/pull/1709#pullrequestreview-2297393967) @liorrutenberg @moshe-blox ? Or will we keep two discoveries on the fork, and later on deprecate it on a new release?

liorrutenberg:
@MatheusFranco99 I think 2. is good(after restart) to not overcomplicate the on-fork logic

nkryuchkov (commented):
@MatheusFranco99 Great job, thanks for your contribution! I'd just simplify a couple of things

network/discovery/fork_listener.go
```diff
@@ -0,0 +1,160 @@
...
+	if l.netCfg.PastAlanFork() {
+		l.closePreForkListener()
+		return l.postForkListener.Ping(node)
+	}
+
+	err := l.postForkListener.Ping(node)
+	if err != nil {
+		return l.preForkListener.Ping(node)
+	}
+	return err
```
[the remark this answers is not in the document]
MatheusFranco99: Sure, thanks

network/discovery/dv5_filters.go
```diff
@@ -8,9 +8,9 @@ import (
 	"go.uber.org/zap"
 )
 
-// limitNodeFilter checks if limit exceeded
+// limitNodeFilter returns true if the limit is exceeded
 func (dvs *DiscV5Service) limitNodeFilter(node *enode.Node) bool {
-	return !dvs.conns.AtLimit(libp2pnetwork.DirOutbound)
+	return dvs.conns.AtLimit(libp2pnetwork.DirOutbound)
```
y0sher, line 13: I'm not sure about this change, I was under the impression filters that returns true means that peers is ok. And if we're at the limit shouldn't we filter out the peer? maybe I misunderstood filters?

MatheusFranco99: Oh okay, it made more sense to me for this function to return true if we were at the limit. But I'll revert it

network/discovery/fork_listener.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	hasNext := i.postForkIterator.Next()
+	if hasNext {
+		return true
+	}
+
+	i.hasChangedToPreForkIterator = true
+	return i.preForkIterator.Next()
+}
```
moshe-blox, line 160: we should probably Close the post-fork iterator here since it will never be used again

and in that case Close() should only close postFork if we didnt already switch to preFork (to not close twice)

network/discovery/fork_listener.go
```diff
@@ -0,0 +1,160 @@
...
+	"sync"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ssvlabs/ssv/networkconfig"
+)
+
+// forkListener wraps a pre-fork and a post-fork listener.
+// Before the fork, it performs operations on both services.
+// Aftet the fork, it performs operations only on the post-fork service.
+type forkListener struct {
```
moshe-blox, line 13: since we already have a ForkListener in the node, we should probably rename this one so we don't have too many 😅 

maybe `forkingDV5Listener`?
(to parallel a similar mechanism we have called `forkingMessageValidation`)

y0sher (approved):
Code-wise I approve. I think we need @Roy-blox and @andrew-blox to chime in, we should start covering all the cases in stage-tests.
- [ ] - pre-fork -> alan
- [ ] - perf and connectivity without restarting
- [ ] - perf and connectivity after restart
- [ ] - what happens if boot node stops accepting peer with no protocolID (post-fork)

moshe-blox (changes requested):
LGTM 🔥 

left a few comments

i would've been interested to see an approach with a slice of 2, but i guess it doesn't really matter in the grand scheme because this code will likely be removed post-Alan

network/discovery/fork_listener.go
MatheusFranco99: Agree, renaming

network/discovery/fork_listener.go
MatheusFranco99: Agree, adding this

y0sher (approved):

moshe-blox (approved):

PR #1727, opened 2024-09-11
Removed redundant slashableAttestationCheck

AKorpusenko:
## Description

Removed `beacon/goclient/attest_protect.go` due the `slashableAttestationCheck` always returned nil and we rely on slashing check in the signing mechanism.


MatheusFranco99 (approved):

y0sher (approved):

moshe-blox:
> BTW @moshe-blox, I'm not sure about the signing root computation here:
> 
> ```go
>                  _, err := gc.getSigningRoot(attestation.Data)
> 		if err != nil {
> 			return errors.Wrap(err, "failed to get signing root")
> 		}
> ```
> 
> Can this signing root computation can be assumed as a sanity check?

where is it in? can't see it by searching

moshe-blox (approved):

PR #1729, opened 2024-09-11
fix: (eventhandler) pubkey & id uniqueness check in OperatorAdded

AKorpusenko:
## Description

Adde verification mechanism to ensure the uniqueness of the operator's public key and ID

### Validation flow

Handling OperatorAdded event we now check:
- there is no operator in storage with the same ID (otherwise throwing `ErrAlreadyRegisteredWithSameId`)
- there is no operator in storage with the same pubkey (otherwise throwing `ErrAlreadyRegisteredWithSamePubkey`)

### DIFF
--- a/eth/eventhandler/event_handler_test.go
+++ b/eth/eventhandler/event_handler_test.go
@@ -202,6 +202,66 @@ func TestHandleBlockEventsStream(t *testing.T) {
 		}
 	})
 
+	t.Run("test OperatorAdded event handle with duplicated pubkey, but different Id", func(t *testing.T) {
+		op := &testOperator{}
+		op.id = 5
+		op.privateKey = ops[0].privateKey
+
+		encodedPubKey, err := op.privateKey.Public().Base64()
+		require.NoError(t, err)
+
+		// Call the contract method
+		packedOperatorPubKey, err := eventparser.PackOperatorPublicKey(encodedPubKey)
+		require.NoError(t, err)
+		_, err = boundContract.SimcontractTransactor.RegisterOperator(auth, packedOperatorPubKey, big.NewInt(100_000_000))
+		require.NoError(t, err)
+
+		sim.Commit()
+
+		block := <-logs
+		require.NotEmpty(t, block.Logs)
+		require.Equal(t, ethcommon.HexToHash("0xd839f31c14bd632f424e307b36abff63ca33684f77f28e35dc13718ef338f7f4"), block.Logs[0].Topics[0])
+
+		eventsCh := make(chan executionclient.BlockLogs)
+		go func() {
+			defer close(eventsCh)
+			eventsCh <- block
+		}()
+
+		// Check that there is no registered operators
+		operators, err := eh.nodeStorage.ListOperators(nil, 0, 0)
+		require.NoError(t, err)
+		require.Equal(t, len(ops), len(operators))
+
+		// Handle the event
+		lastProcessedBlock, err := eh.HandleBlockEventsStream(eventsCh, false)
+		require.Equal(t, blockNum+1, lastProcessedBlock)
+		require.NoError(t, err)
+		blockNum++
+
+		// Check storage for the new operators
+		operators, err = eh.nodeStorage.ListOperators(nil, 0, 0)
+		require.NoError(t, err)
+		require.Equal(t, len(ops), len(operators))
+
+		// Check if operators in the storage have same attributes
+		for _, log := range block.Logs {
+			operatorAddedEvent, err := contractFilterer.ParseOperatorAdded(log)
+			require.NoError(t, err)
+			fmt.Println(operatorAddedEvent)
+
+			//data, _, err := eh.nodeStorage.GetOperatorData(nil, operatorAddedEvent.OperatorId)
+			//require.NoError(t, err)
+			//require.Equal(t, operatorAddedEvent.OperatorId, data.ID)
+			//require.Equal(t, operatorAddedEvent.Owner, data.OwnerAddress)
+			//
+			//encodedPubKey, err := ops[i].privateKey.Public().Base64()
+			//require.NoError(t, err)
+			//
+			//require.Equal(t, encodedPubKey, data.PublicKey)
+		}
+	})
+
 	t.Run("test OperatorRemoved event handle", func(t *testing.T) {
 
 		// Should return MalformedEventError and no changes to the state
--- a/eth/eventhandler/event_handler_test.go
+++ b/eth/eventhandler/event_handler_test.go
@@ -202,10 +202,10 @@ func TestHandleBlockEventsStream(t *testing.T) {
 		}
 	})
 
-	t.Run("test OperatorAdded event handle with duplicated pubkey, but different Id", func(t *testing.T) {
+	t.Run("test OperatorAdded event handle with the same pubkey, but with a different id", func(t *testing.T) {
 		op := &testOperator{}
 		op.id = 5
-		op.privateKey = ops[0].privateKey
+		op.privateKey = ops[1].privateKey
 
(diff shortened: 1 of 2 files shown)


eth/eventhandler/handlers.go
```diff
@@ -54,13 +54,16 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
 
 	// throw an error if there is an existing operator with the same public key and different operator id
 	operatorData := eh.operatorDataStore.GetOperatorData()
-	if operatorData.ID != 0 && bytes.Equal(operatorData.PublicKey, event.PublicKey) && operatorData.ID != event.OperatorId {
+	_, found, err := eh.nodeStorage.GetOperatorDataByPubKey(txn, event.PublicKey)
+	if err != nil {
+		return fmt.Errorf("could not get operator data by public key: %w", err)
+	}
+	if found || operatorData.ID != 0 && bytes.Equal(operatorData.PublicKey, event.PublicKey) && operatorData.ID != event.OperatorId {
```
y0sher, line 61: why || and not && ? if its found but operator ID is 0, is going to be true condition.

y0sher, line 61: I think the point was to check the operator data from the storage against the event operator data. 
`operatorData` is our *own* operator. we want to compare the event to the operator found in the DB/

eth/eventhandler/event_handler_test.go
```diff
@@ -202,6 +202,92 @@ func TestHandleBlockEventsStream(t *testing.T) {
 		}
 	})
 
+	t.Run("test OperatorAdded event fails for same event data", func(t *testing.T) {
```
y0sher, line 205: it it possible to check that the exact error is returned?

AKorpusenko: Nope. We can't do it because of how we handle the `MalformedEventError`. 
```go
		if err := eh.handleOperatorAdded(txn, operatorAddedEvent); err != nil {
			eh.metrics.EventProcessingFailed(abiEvent.Name)

			var malformedEventError *MalformedEventError
			if errors.As(err, &malformedEventError) {
				return nil, nil
			}
			return nil, fmt.Errorf("handle OperatorAdded: %w", err)
		}
```

eth/eventhandler/handlers.go
AKorpusenko: Agree. Fixed

eth/eventhandler/handlers.go
```diff
@@ -52,15 +53,28 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
...
+	// throw an error if there is an existing operator with the same public key
+	operatorData, foundByPubKey, err := eh.nodeStorage.GetOperatorDataByPubKey(txn, event.PublicKey)
+	if err != nil {
+		return fmt.Errorf("could not get operator data by public key: %w", err)
+	}
+	if foundByPubKey && operatorData.ID != 0 && bytes.Equal(operatorData.PublicKey, event.PublicKey) && operatorData.ID != event.OperatorId {
 		logger.Warn("malformed event: operator registered with the same operator public key",
 			zap.Uint64("expected_operator_id", operatorData.ID))
-		return &MalformedEventError{Err: ErrAlreadyRegistered}
+		return &MalformedEventError{Err: ErrAlreadyRegisteredWithSamePubkey}
```
moshe-blox, line 75: if the contract does not prevent that, should we? @y0sher @GalRogozinski 

btw, if we choose to not prevent that, we should check around in the entire codebase that we never identify operators by their pubkey (via lookup or by maintaining a map of operators by pubkey, for example)

moshe-blox: also what would that mean for nodes which synced before this change comes up?

if there's already an active operator with a duplicate pubkey by the time this releases, then newly synced nodes will not see them while already synced nodes would

moshe-blox: never mind

@mtabasco approves that the contract doesn't allow duplicate registrations so we can proceed :-)

y0sher: > if the contract does not prevent that, should we? @y0sher @GalRogozinski
> 
> btw, if we choose to keep allowing duplicates, we should check around in the entire codebase that we always identify operators by their ID and never by their pubkey (i.e never lookup by pubkey or maintain a map of operators by pubkey, for example)

its like nonce, enforced by the node. 




> also what would that mean for nodes who synced before this change?
> 
> if there's already an active operator with a duplicate pubkey by the time this releases, then newly synced nodes will not see them while already synced nodes would, and it becomes a fork?

I agree about this. but this is a bug, we should check the db before releasing to see if there's any special treatment needed.

y0sher (approved):

eth/eventhandler/handlers.go
AKorpusenko: > I agree about this. but this is a bug, we should check the db before releasing to see if there's any special treatment needed.

@y0sher, let's create a issue for this?

moshe-blox: @y0sher there's no need to support that now that Marco approved it can't happen bc the contract enforces

y0sher: @moshe-blox so this means we can remove this check as whole? so no need for this PR. (we still need a PR to remove it..)

moshe-blox: @y0sher technically if we're certain the contract should not emit such events then we should actually return a normal error to fail the sync entirely (rather than malformedError) in these cases

moshe-blox (changes requested):

eth/eventhandler/handlers.go
moshe-blox: @y0sher technically if we're certain the contract should not emit such events then we should actually return a normal error to fail the sync entirely (rather than malformedError) in these cases

eth/eventhandler/handlers.go
```diff
@@ -52,15 +53,28 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
...
+			zap.Uint64("expected_operator_id", event.OperatorId))
+		return &MalformedEventError{Err: ErrAlreadyRegisteredWithSameId}
+	}
+
+	// throw an error if there is an existing operator with the same public key
+	operatorData, foundByPubKey, err := eh.nodeStorage.GetOperatorDataByPubKey(txn, event.PublicKey)
+	if err != nil {
+		return fmt.Errorf("could not get operator data by public key: %w", err)
+	}
+	if foundByPubKey && operatorData.ID != 0 && bytes.Equal(operatorData.PublicKey, event.PublicKey) && operatorData.ID != event.OperatorId {
```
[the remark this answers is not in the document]
AKorpusenko: @moshe-blox, agree. It's just enough to check the PK does not exists in the db

eth/eventhandler/handlers.go
```diff
@@ -25,7 +25,8 @@ import (
 const encryptedKeyLength = 256
 
 var (
-	ErrAlreadyRegistered            = fmt.Errorf("operator registered with the same operator public key")
+	ErrOperatorPubkeyAlreadyExists  = fmt.Errorf("operator public key already exists")
+	ErrOperatorIDAlreadyExists      = fmt.Errorf("operator ID already exists")
```
anatollupacescu, line 29: I see that this is the current convention, but wouldn't we want eventually to have these as `error.New()`?

AKorpusenko: @anatollupacescu, [pkg/errors](https://github.com/pkg/errors) is deprecated and archived. We trying to use `fmt.Errorf` istead

moshe-blox (approved):

moshe-blox (approved):

PR #1733, opened 2024-09-11
chore: removed startCommittee & startValidator methods

AKorpusenko:
## Description 

Removed `StartCommittee` and `StartValidator` as we don't need them anymore


y0sher (approved):

operator/validator/controller.go
```diff
@@ -1080,45 +1061,13 @@ func (c *controller) setShareFeeRecipient(share *ssvtypes.SSVShare, getRecipient
...
-	started, err := c.validatorStart(v)
-	if err != nil {
-		c.metrics.ValidatorError(v.Share().ValidatorPubKey[:])
-		return false, errors.Wrap(err, "could not start validator")
-	}
-	if started {
-		c.recentlyStartedValidators++
-	}
-
-	return true, nil
```
moshe-blox, line 1107: doesn't this mean we're not starting validators at all right now?

moshe-blox (changes requested):

operator/validator/controller.go
y0sher: I agree we should call Validator.Start here.

y0sher (approved):

moshe-blox (approved):

PR #1734, opened 2024-09-11
fix: minor refactor ECDSAPrivFromInterface to remove ScalarBaseMult

AKorpusenko:
## Description

The `ECDSAPrivFromInterface` function in the codebase is responsible for converting a `crypto.PrivKey` into an `ecdsa.PrivateKey`. However, this function currently relies on the `ScalarBaseMult` method to compute the public key coordinates from the private key bytes. 
The `ScalarBaseMult` method is deprecated and considered a low-level, unsafe API


network/commons/keys.go
```diff
@@ -19,14 +19,24 @@ func ECDSAPrivFromInterface(privkey crypto.PrivKey) (*ecdsa.PrivateKey, error) {
...
+		return nil, errors.Wrap(err, "could not convert ecdsa.PrivateKey")
 	}
 
 	privKey := new(ecdsa.PrivateKey)
 	k := new(big.Int).SetBytes(rawKey)
 	privKey.D = k
 	privKey.Curve = gcrypto.S256() // Temporary hack, so libp2p Secp256k1 is recognized as geth Secp256k1 in disc v5.1.
-	privKey.X, privKey.Y = gcrypto.S256().ScalarBaseMult(rawKey)
+
+	ecdhPrivKey, err := ecdh.P256().NewPrivateKey(rawKey)
```
MatheusFranco99, line 30: Why are we using elliptic curve diffie hellman for that? Is this the standard way?

y0sher (approved):

moshe-blox:
looks good to me, i'd like @MatheusFranco99 approval as well for this crypto

moshe-blox (approved):

MatheusFranco99 (approved):
Overall it looks good to me. But, tbh, I don't understand this btcec library very well. But it seems we have already been using it

PR #1738, opened 2024-09-12
fix: Added missing ownValidator param to ExitValidator

AKorpusenko:


y0sher (approved):

moshe-blox:
does this bug mean exits should not work at the moment? do we have a confirmation for that? @andrew-blox

andrew-blox:
> does this bug mean exits should not work at the moment? do we have a confirmation for that? @andrew-blox

Yes I can confirm for single and bulk on pre and post that they all failed
I tested all of these again with this branch (pre / post - single / bulk for both) and all pass

moshe-blox (approved):

PR #1740, opened 2024-09-15
fix: (discovery) publish latest ENR despite ongoing publish

moshe-blox:


network/discovery/dv5_service.go
```diff
@@ -319,21 +310,47 @@ func (dvs *DiscV5Service) PublishENR(logger *zap.Logger) {
...
-		metricPublishEnrPings.Inc()
-		err := dvs.dv5Listener.Ping(e.Node)
-		if err != nil {
-			if err.Error() == "RPC timeout" {
-				// ignore
-				return
-			}
-			logger.Warn("could not ping node", fields.TargetNodeENR(e.Node), zap.Error(err))
-			return
+	dvs.publishMutex.Lock()
```
y0sher, line 313: I don't get why we need the mutex? if the deadline is only changed/accessed in this func/loop..

moshe-blox: just so there's no race condition between the read & write operations on `publishDeadline`

network/discovery/dv5_service.go
```diff
@@ -319,11 +309,31 @@ func (dvs *DiscV5Service) PublishENR(logger *zap.Logger) {
...
+
+	select {
+	case <-ctx.Done():
+		return
+	case dvs.publishLock <- struct{}{}:
+	}
+	defer func() {
+		// Release lock.
+		<-dvs.publishLock
+	}()
```
y0sher, line 324: the only issue I have here is:
if previous publish is still running, we will wait the timeout, if timeout expired and we return. we will not publish a newer ENR. IMO this function should be able to know the latest ENR and maybe even cancel a previous publish to publish the most updated version.

moshe-blox: the ongoing publish will publish the new ENR because it's updated before we start waiting for lock

i avoided cancelling to keep the code simple, because cancelling requires adding a shared cancel func and a done channel protected by a lock to wait for the running goroutine to exit

despite this solution being less optimization, i'm not worried about there being a lot of waiting goroutines because PublishENR is not called often

moshe-blox: Example of cancellation:
image

y0sher: We concluded that there's no reason to stop the publish in order to publish more recent ENR, since the ENR is updated within discv5 and every round of the previous publish will also publish the latest ENR.

y0sher (approved):

network/discovery/dv5_service.go
olegshmuelov, line 326: To avoid potential deadlocks or indefinite waiting, we can make the lock acquisition non-blocking. This ensures that if another routine is already holding the lock, the current routine doesn't block indefinitely but instead logs a warning and skips the current attempt.

```go
ctx, done := context.WithTimeout(dvs.ctx, publishENRTimeout)
defer done()

select {
case dvs.publishLock <- struct{}{}:
    defer func() {
        <-dvs.publishLock // Release the lock
    }()
    
    // the internal logic should go here (the ENR publishing logic)

case <-ctx.Done():
    logger.Warn("skipped ENR publishing: context timed out while waiting for the publish lock")
    return

default:
    logger.Warn("skipped ENR publishing due to another ongoing operation")
    return
}
```

moshe-blox: @olegshmuelov this would mean that if an ongoing publish is let's say in it's last second (59 out of 60) then the updated record would only be published for 1 second.

in that scenario, the current PR waits 1 second to acquire the lock so that it can continue publishing for 59 more seconds

i updated the comment above the lock to explain the mechanism:
```go
// Acquire publish lock to prevent parallel publishing.
// If there's an ongoing goroutine, it would now start publishing the record updated above,
// and if it's done before the new deadline, this goroutine would pick up where it left off.
```

y0sher: I agree with @moshe-blox here. 
Also on @olegshmuelov suggestion we will rarely wait for the timeout, we will drop into default most times (when lock is acquired).
Waiting is limited to the timeout so I don't see how deadlock can happen currently.

olegshmuelov: Got it, @moshe-blox. That approach makes sense now, and the updated comment clears it up. Looks good!

olegshmuelov (approved):

PR #1725, opened 2024-09-11
fix: duty scheduler indices change & reorg bleeps

olegshmuelov:
close #1724 
close https://github.com/ssvlabs/ssv/issues/1714
close #1807

PR #1741, opened 2024-09-16
fix: (DutyScheduler) duties reset race condition

moshe-blox:
This PR fixes a race condition discovered by @olegshmuelov which happens right after an indices changes and re-orgs, when AttesterHandler and SyncCommitteeHandler reset their respective duty maps prior to re-fetching, during which any reads by CommitteeHandler would end up with zero duties.

The fix changes how we populate the map, opting for `fetch -> set` rather than `reset -> fetch -> add`.

This bug was first fixed in https://github.com/ssvlabs/ssv/pull/1725 using a different approach. Since that PR contains other other improvements & fixes, then this PR doesn't close it, and we should combine both after verifying this approach works.


operator/duties/dutystore/sync_committee.go
```diff
@@ -51,20 +51,13 @@ func (d *SyncCommitteeDuties) Duty(period uint64, validatorIndex phase0.Validato
...
-	defer d.mu.Unlock()
-
-	if _, ok := d.m[period]; !ok {
-		d.m[period] = make(map[phase0.ValidatorIndex]dutyDescriptor[eth2apiv1.SyncCommitteeDuty])
-	}
-
-	d.m[period][validatorIndex] = dutyDescriptor[eth2apiv1.SyncCommitteeDuty]{
-		duty:        duty,
-		inCommittee: inCommittee,
+func (d *SyncCommitteeDuties) Set(period uint64, duties []DutyDescriptor[eth2apiv1.SyncCommitteeDuty]) {
```
y0sher, line 57: you removed the mutex lock here

moshe-blox: @y0sher good catch, fixed
i also pushed a bunch of refactors

operator/duties/dutystore/duties.go
```diff
@@ -70,22 +72,19 @@ func (d *Duties[D]) ValidatorDuty(epoch phase0.Epoch, slot phase0.Slot, validato
...
-	}
-	d.m[epoch][slot][validatorIndex] = dutyDescriptor[D]{
-		duty:        duty,
-		inCommittee: inCommittee,
+	d.m[epoch] = make(map[phase0.Slot]map[phase0.ValidatorIndex]StoreDuty[D])
+	for _, duty := range duties {
+		if _, ok := d.m[epoch][duty.Slot]; !ok {
+			d.m[epoch][duty.Slot] = make(map[phase0.ValidatorIndex]StoreDuty[D])
+		}
+		d.m[epoch][duty.Slot][duty.ValidatorIndex] = duty
```
olegshmuelov, line 87: we can reduce lock time and reduce redundant lookups
```go
func (d *Duties[D]) Set(epoch phase0.Epoch, duties []StoreDuty[D]) {
	newDuties := make(map[phase0.Slot]map[phase0.ValidatorIndex]StoreDuty[D])

	for _, duty := range duties {
		if _, slotExists := newDuties[duty.Slot]; !slotExists {
			newDuties[duty.Slot] = make(map[phase0.ValidatorIndex]StoreDuty[D])
		}
		newDuties[duty.Slot][duty.ValidatorIndex] = duty
	}

	d.mu.Lock()
	defer d.mu.Unlock()

	d.m[epoch] = newDuties
}
```

operator/duties/dutystore/sync_committee.go
```diff
@@ -51,20 +57,16 @@ func (d *SyncCommitteeDuties) Duty(period uint64, validatorIndex phase0.Validato
...
-	}
-
-	d.m[period][validatorIndex] = dutyDescriptor[eth2apiv1.SyncCommitteeDuty]{
-		duty:        duty,
-		inCommittee: inCommittee,
+	d.m[period] = make(map[phase0.ValidatorIndex]StoreSyncCommitteeDuty)
+	for _, duty := range duties {
+		d.m[period][duty.ValidatorIndex] = duty
 	}
 }
```
olegshmuelov, line 71: same here
```go
func (d *SyncCommitteeDuties) Set(period uint64, duties []StoreSyncCommitteeDuty) {
	newDuties := make(map[phase0.ValidatorIndex]StoreSyncCommitteeDuty, len(duties))

	for _, duty := range duties {
		newDuties[duty.ValidatorIndex] = duty
	}

	d.mu.Lock()
	defer d.mu.Unlock()

	d.m[period] = newDuties
}
```

y0sher (approved):

olegshmuelov (approved):

PR #1742, opened 2024-09-16
chore: update dependencies

AKorpusenko:


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22
 
 require (
 	github.com/aquasecurity/table v1.8.0
-	github.com/attestantio/go-eth2-client v0.21.7
+	github.com/attestantio/go-eth2-client v0.21.11
```
moshe-blox, line 7: prefer to not upgrade go-eth2-client without a good reason because they sometimes change roots which may cause forks 

when we do change it it should be probably also changed in spec and tested

@GalRogozinski can attest

moshe-blox (changes requested):

go.mod
AKorpusenko: Got it. Even the [diff](https://github.com/attestantio/go-eth2-client/compare/v0.21.7...v0.21.11) between 0.21.11 and 0.21.7 is really minor (thats why I added this change), updating the spec-ssv repo first makes sense. Agree 100% 
Removed this

moshe-blox (approved):

y0sher (approved):

PR #1743, opened 2024-09-17
feat: duty stopping

moshe-blox:
This PR implements a `CommitteeDutyGuard` shared by `Committee` and `CommitteeRunner` to guarantee exclusive execution of only one duty per validator, as well as non-execution of stopped validators.

Cons: this does not stop consensus even when all duties are no longer valid, rather it only prevents signing and processing of partial signatures at post-consensus.

A future more complete solution could be to remove direct duty access from `CommitteeRunner`, which instead would access them via something like `CommitteeDuties.At(slot)` which would only return the valid ones, and if there are none, we should stop the runner, including at consensus stage.


networkconfig/holesky-stage.go
```diff
@@ -16,7 +16,7 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        99999999,
+	AlanForkEpoch:        0,
```
y0sher, line 19: do we want to merge it to be post-fork?

moshe-blox: @y0sher just testing deployment, will revert soon

protocol/v2/ssv/validator/committee_guard.go
```diff
@@ -0,0 +1,70 @@
...
+	defer a.mu.Unlock()
+
+	duties, ok := a.duties[role]
+	if !ok {
+		return fmt.Errorf("unsupported role %d", role)
+	}
+	runningSlot, exists := duties[validator]
+	if exists && runningSlot >= slot {
+		return fmt.Errorf("duty already running at slot %d", runningSlot)
+	}
```
y0sher, line 39: if `exists && slot > running slot` shouldn't we stop the existing duty before starting this one?

moshe-blox: it does, indirectly

in case the given slot is higher than running slot (`exists && slot > running slot`), then it will replace it in the map, after which `ValidDuty` would return an error upon checking the previously running slot

maybe we should rename the methods or update comments to make it clear

y0sher: I get it now. And it seems that previous approach was doing it pretty similar to this...

y0sher (approved):
This looks better than previous code.

protocol/v2/ssv/runner/committee.go
```diff
@@ -216,9 +216,15 @@ func (cr *CommitteeRunner) ProcessConsensus(logger *zap.Logger, msg *spectypes.S
 	}
 
 	beaconVote := decidedValue.(*spectypes.BeaconVote)
+	validDuties := 0
 	for _, duty := range duty.(*spectypes.CommitteeDuty).ValidatorDuties {
+		if err := cr.DutyGuard.ValidDuty(duty.Type, spectypes.ValidatorPK(duty.PubKey), duty.DutySlot()); err != nil {
+			logger.Debug("invalid duty", fields.Validator(duty.PubKey[:]), fields.BeaconRole(duty.Type), zap.Error(err))
```
olegshmuelov, line 222: use logger.error or warn

protocol/v2/ssv/runner/committee.go
```diff
@@ -573,15 +586,15 @@ func (cr *CommitteeRunner) expectedPostConsensusRootsAndBeaconObjects() (
 		if validatorDuty == nil {
 			continue
 		}
-		_, stopped := cr.stoppedValidators[spectypes.ValidatorPK(validatorDuty.PubKey)]
-		if stopped {
+		if err := cr.DutyGuard.ValidDuty(validatorDuty.Type, spectypes.ValidatorPK(validatorDuty.PubKey), validatorDuty.DutySlot()); err != nil {
+			logger.Debug("invalid duty", fields.Validator(validatorDuty.PubKey[:]), fields.BeaconRole(validatorDuty.Type), zap.Error(err))
```
olegshmuelov, line 590: ditto

protocol/v2/ssv/validator/committee_guard.go
```diff
@@ -0,0 +1,70 @@
...
+	duties[validator] = slot
+	return nil
+}
```
MatheusFranco99, line 42: As I understood, even if `StopValidator` is called for, let's say, a validator V, `StartDuty` may return no error for starting a new duty for validator V, right? Is this the intended behaviour?

MatheusFranco99:
The CommitteeRunner calls `HasSubmittedAllValidatorDuties` to check if it can update the state to `Finished`. Here:
```go
func (cr *CommitteeRunner) ProcessPostConsensus(logger *zap.Logger, signedMsg *spectypes.PartialSignatureMessages) error {

        // ....

	// Check if duty has terminated (runner has submitted for all duties)
	if cr.HasSubmittedAllValidatorDuties(attestationMap, committeeMap) {
		cr.BaseRunner.State.Finished = true
	}
	return nil
}
```
In case there are invalid duties, I believe the `cr.HasSubmittedAllValidatorDuties` function may never return true. I think we should change this behaviour to return true if all valid duties are submitted. What do you think?

protocol/v2/ssv/runner/committee.go
```diff
@@ -240,15 +250,21 @@ func (cr *CommitteeRunner) ProcessConsensus(logger *zap.Logger, msg *spectypes.S
...
 				return errors.Wrap(err, "failed signing sync committee message")
 			}
 			postConsensusMsg.Messages = append(postConsensusMsg.Messages, partialMsg)
+		default:
+			return fmt.Errorf("invalid duty type: %s", duty.Type)
 		}
 	}
+	if validDuties == 0 {
+		return errors.New("no valid duties")
+	}
```
MatheusFranco99, line 267: Perhaps, can we set the Runner as "Finished" here as well?

protocol/v2/ssv/validator/committee_guard.go
GalRogozinski: maybe you guys want to also mention `slot` in the error?
Or maybe your logger does this already?

protocol/v2/ssv/validator/committee_guard.go
```diff
@@ -0,0 +1,70 @@
...
+	duties, ok := a.duties[role]
+	if !ok {
+		return fmt.Errorf("unsupported role %d", role)
+	}
+	runningSlot, exists := duties[validator]
+	if !exists {
+		return fmt.Errorf("duty not found")
+	}
+	if runningSlot != slot {
+		return fmt.Errorf("slot mismatch: duty is running at slot %d", runningSlot)
```
GalRogozinski, line 67: again for the error

GalRogozinski (approved):
lgtm

protocol/v2/ssv/validator/committee_guard.go
moshe-blox: yeah this would be in the context of the logger which tried to start the duty ("failed to handle message" log)

protocol/v2/ssv/runner/committee.go
moshe-blox: @MatheusFranco99 done

protocol/v2/ssv/validator/committee_guard.go
moshe-blox: @GalRogozinski added: https://github.com/ssvlabs/ssv/pull/1743/commits/8c19bc3870a8eb2a58c9f1c506adb37dea8c3a2a

MatheusFranco99 (approved):

PR #1747, opened 2024-09-19
fix: fee recipient update (validator registration duty)

olegshmuelov:
This PR addresses an issue that arises when nodes are running in a mixed cluster of pre-fork and current production versions (the real pre-fork). The problem occurs when the fee recipient is updated—the `GenesisShare` was not being updated due to a deep copy, and as a result, the fee recipient was not correctly reflected in the runner's share.

To resolve this, a temporary solution has been introduced. The PR continues to pass the pre-fork share options struct to `SetupGenesisRunners`, but adds a function to the `ValidatorRegistrationRunner` to dynamically retrieve the fee recipient from the current (pre-fork) share, rather than from a shallow copy (`GenesisShare`).

A long-term solution has been described in [issue #1748], which will involve passing the genesis options directly to `SetupGenesisRunners`, removing unnecessary genesis options from the pre-fork options struct, and identifying further cleanups that can follow these changes.

Additionally, a related issue has been raised with the spec team [issue [#504](https://github.com/ssvlabs/ssv-spec/issues/504)], highlighting the fact that the `ValidatorRegistration` duty currently does not include the fee recipient and uses a dynamic duty struct for all duties. This can lead to a data race during the pre-consensus process when the fee recipient is updated, causing the duty to fail. A spec refactor is needed to prevent this.


moshe-blox (approved):
LGTM 🔥

y0sher (approved):

PR #1749, opened 2024-09-19
fix: genesis options correct use 

### LINKED ISSUE #1748: Improve passing genesis options to SetupGenesisRunners to avoid reliance on pre-fork options
The current solution for handling fee recipient updates in `ValidatorRegistrationRunner` is a temporary workaround that still passes pre/post-fork share options structs to `SetupGenesisRunners`. A better long-term solution would be to pass the appropriate genesis options directly to `SetupGenesisRunners`, eliminating the need for pre-fork options and reducing complexity.

The goal is to refactor how genesis options are handled during the setup, ensuring that the runner always works with the correct, up-to-date genesis data without relying on dynamic retrieval from pre/post-fork share options.

**Steps to improve:**

- Refactor `SetupGenesisRunners` to accept genesis options directly.
- Remove unnecessary genesis options from the pre-fork options struct for better clarity.
- Ensure the runner works with up-to-date genesis data without relying on dynamic retrieval from pre/post-fork share options.
- Look for additional potential cleanups that can be followed by these improvements, especially around the handling of genesis options and any redundant data in the pre/post-fork options structs.

olegshmuelov:
different way to fix https://github.com/ssvlabs/ssv/pull/1747
close #1748


moshe-blox (commented):
LGTM

@andrew-blox we should test that the fee recipient changes on both prefork and postfork with this PR

if it's already been tested, pls lmk

y0sher (approved):
lgtm

PR #1756, opened 2024-09-23
fix: committee role duty limit

olegshmuelov:
This PR fixes the committee role's duty limit by ensuring that unnecessary duty checks are skipped when the number of validators is high. It adjusts duty limit calculations and ensures proper checks for the sync committee based on validator count.

### DIFF
--- a/message/validation/common_checks.go
+++ b/message/validation/common_checks.go
@@ -64,7 +64,7 @@ func (mv *messageValidator) validateDutyCount(
 		return nil
 	}
 
-	if dutyCount > dutyLimit {
+	if dutyCount >= dutyLimit {
 		err := ErrTooManyDutiesPerEpoch
 		err.got = fmt.Sprintf("%v (role %v)", dutyCount, msgID.GetRoleType())
 		err.want = fmt.Sprintf("less than %v", dutyLimit)
--- a/message/validation/common_checks.go
+++ b/message/validation/common_checks.go
@@ -54,17 +54,17 @@ func (mv *messageValidator) messageLateness(slot phase0.Slot, role spectypes.Run
 func (mv *messageValidator) validateDutyCount(
 	msgID spectypes.MessageID,
 	msgSlot phase0.Slot,
-	validatorIndexCount int,
+	validatorIndices []phase0.ValidatorIndex,
 	signerStateBySlot *OperatorState,
 ) error {
 	dutyCount := signerStateBySlot.DutyCount(mv.netCfg.Beacon.EstimatedEpochAtSlot(msgSlot))
 
-	dutyLimit, exists := mv.dutyLimit(msgID, msgSlot, validatorIndexCount)
+	dutyLimit, exists := mv.dutyLimit(msgID, msgSlot, validatorIndices)
 	if !exists {
 		return nil
 	}
 
-	if dutyCount >= dutyLimit {
+	if dutyCount > dutyLimit {
 		err := ErrTooManyDutiesPerEpoch
 		err.got = fmt.Sprintf("%v (role %v)", dutyCount, msgID.GetRoleType())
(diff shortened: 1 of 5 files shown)


MatheusFranco99 (changes requested):
Good job!!

A test is not passing, so perhaps one msg validation test may have broken with the changes

olegshmuelov:
> Good job!!
> 
> A test is not passing, so perhaps one msg validation test may have broken with the changes

please see [revert duty count limit check to inclusive comparison (>=)](https://github.com/ssvlabs/ssv/pull/1756/commits/8d9ff186d3844968cc659e7b8ef511525ef50bc4)
The revert was necessary because we first validate the duty count and only then update the state. This ensures that the first count is 0, which is why the inclusive comparison (>=) is now correct

message/validation/common_checks.go
```diff
@@ -86,7 +86,24 @@ func (mv *messageValidator) dutyLimit(msgID spectypes.MessageID, slot phase0.Slo
 		return 2, true
```
nkryuchkov, line 86: @olegshmuelov @MatheusFranco99 do we need to change it to 1 because now we check `dutyCount > dutyLimit`?

olegshmuelov: > @olegshmuelov @MatheusFranco99 do we need to change it to 1 because now we check `dutyCount > dutyLimit`?

https://github.com/ssvlabs/ssv/pull/1756#issuecomment-2368091077

nkryuchkov:
> > Good job!!
> > A test is not passing, so perhaps one msg validation test may have broken with the changes
> 
> please see [revert duty count limit check to inclusive comparison (>=)](https://github.com/ssvlabs/ssv/pull/1756/commits/8d9ff186d3844968cc659e7b8ef511525ef50bc4) The revert was necessary because we first validate the duty count and only then update the state. This ensures that the first count is 0, which is why the inclusive comparison (>=) is now correct

Can you please add a comment on that?

MatheusFranco99:
> please see [revert duty count limit check to inclusive comparison (>=)](https://github.com/ssvlabs/ssv/pull/1756/commits/8d9ff186d3844968cc659e7b8ef511525ef50bc4)
The revert was necessary because we first validate the duty count and only then update the state. This ensures that the first count is 0, which is why the inclusive comparison (>=) is now correct

Understood it, thanks!

moshe-blox:
@olegshmuelov @y0sher we should add a test to check this case and also the other case where the validators aren't in sync committee

message/validation/common_checks.go
nkryuchkov: but I think the logic in this case is still different from other cases because it's max allowed plus 1 (there may be max 1 aggregator/registration duty, right?). In other cases it's just max allowed

moshe-blox:
@olegshmuelov LGTM 👌

i think we can merge it and add tests in a separate PR

message/validation/common_checks.go
nkryuchkov: confirmed with @MatheusFranco99 that it should be 2, resolving the conversation

protocol/genesis/qbft/instance/prepare.go
```diff
@@ -59,7 +59,7 @@ func (i *Instance) uponPrepare(logger *zap.Logger, signedPrepare *genesisspecqbf
 
 	logger.Debug("📢 broadcasting commit message",
 		fields.Round(specqbft.Round(i.State.Round)),
-		zap.Any("commit_singers", commitMsg.Signers),
+		zap.Any("commit_signers", commitMsg.Signers),
```
MatheusFranco99, line 62: 😅

MatheusFranco99 (approved):

nkryuchkov (approved):
Good catch!

y0sher (approved):

Issue #1558, opened 2024-08-06
rethink committee share management


moshe-blox:
There are most likely race conditions in how we [remove shares from Committee without locks](https://github.com/ssvlabs/ssv/blob/f026fedb7dd0a68980c410c62b063cde6cbfe887/operator/validator/controller.go#L1014), and in how we stop said committee if no shares are left while concurrently we could be adding a new share in the other goroutine which fetches metadata and discovers a newly activated validator?

Also, does the share map hold all shares or only active shares? If all shares then we likely dont stop Committee when there are 0 active but >0 inactive validators

We should rethink this entire subsystem and likely rewrite parts or all of it.

y0sher:
A committee is a set of 4/7/13 operators. The `Committee` struct hold all the `Shares`, a share is a part of a validator private key that is managed by this committee. so for each validator there are 4 `shares` in the committee, 1 for each `operator`. so usually from the operator perspective, there's one share that belongs to you and you use it to send your messages and eventaully sign ethereum duties (block proposals/attestations). and you know all the rest of the shares in order to verify that the operators that are sending messages.

If a validator is removed from SSV, we want to remove his shares from the committee, there could be up to 500 validators in each committee. If all shares (all validators) are removed from this committee, we want to remove that committee. 

Instances of the `Committee` struct are stored in a map that is called `validatormap` under the `validatorController` ( a struct and file that manages most of the business logic of the operator, incl starting and stopping validators and executing their duties using consensus.)

After all shares are removed we want to remove the instance of `Committee` from the `validatormap`. 

So the actual issue is we protect with mutex the shares inside `Committee` so we lock when a `share` is added or removed (`Committee.AddShare`, `Committee.RemoveShare`), but we are not interlocking the committee removal from the `validatormap`. 
So we need to somehow increase the scope of the lock so if `AddShare` is called before the removal of the committee from the validatormap, we end up not removing it. we need to somehow lock these actions together.

anatolie-ssv:
> There are most likely race conditions in how we [remove shares from Committee without locks](https://github.com/ssvlabs/ssv/blob/f026fedb7dd0a68980c410c62b063cde6cbfe887/operator/validator/controller.go#L1014), and in how we stop said committee if no shares are left while concurrently we could be adding a new share in the other goroutine which fetches metadata and discovers a newly activated validator?
> 
> Also, does the share map hold all shares or only active shares? If all shares then we likely dont stop Committee when there are 0 active but >0 inactive validators
[...]

PR #1760, opened 2024-09-24
fix: atomically add and remove shares from a committee

anatolie-ssv:
Share addition/removal takes place under a single lock.

Relevant to this PR: #1861

Close: #1558


operator/validator/controller.go
```diff
@@ -945,36 +935,39 @@ func (c *controller) onShareInit(share *ssvtypes.SSVShare) (*validators.Validato
...
-			zap.String("committee", fields.FormatCommittee(committeeOpIDs)),
-			zap.String("committee_id", hex.EncodeToString(operator.CommitteeID[:])),
-		}...)
+	committeeOpIDs := types.OperatorIDsFromOperators(operator.Committee)
 
-		committeeRunnerFunc := SetupCommitteeRunners(ctx, opts)
+	logger := c.logger.With([]zap.Field{
+		zap.String("committee", fields.FormatCommittee(committeeOpIDs)),
+		zap.String("committee_id", hex.EncodeToString(operator.CommitteeID[:])),
+	}...)
```
y0sher, line 962: it works without explicitly using `zap.Field` 
```suggestion
	logger := c.logger.With(
		zap.String("committee", fields.FormatCommittee(committeeOpIDs)),
		zap.String("committee_id", hex.EncodeToString(operator.CommitteeID[:])))
```

operator/validators/validators_map.go
```diff
@@ -150,16 +150,27 @@ func (vm *ValidatorsMap) PutCommittee(pubKey spectypes.CommitteeID, v *validator
...
 }
 
-// Remove removes a committee instance from the map
-func (vm *ValidatorsMap) RemoveCommittee(pubKey spectypes.CommitteeID) *validator.Committee {
-	if v, found := vm.GetCommittee(pubKey); found {
-		vm.mlock.Lock()
-		defer vm.mlock.Unlock()
+// UpdateCommitteeAtomic allows to perform a mutation on a given committee in an atomic manner
+// returns true if the given committee was found
+func (vm *ValidatorsMap) UpdateCommitteeAtomic(pubKey spectypes.CommitteeID, mutate func(*validator.Committee)) bool {
```
y0sher, line 155: lets add unit tests for this function

y0sher (changes requested):
lets add a test

y0sher (approved):
good job!

y0sher (changes requested):
This solution indeed works at the moment, but there's something I don't like about it. We still have `AddShare `and `RemoveShare` methods in `committee.go` . This means someone can use them without the atomicity you just added. This means if someone add a usage of them in the future, we might have a race condition again.
I can think of 2 options we have here:
1. Emphasize in name and comments that these methods are practically unsafe, and that any consumer should use the atomic function in `ValidatorsMap`
2. Move the atomicity to inside the `Committee` struct, which will inherently stop itself when it has no shares left. this way you can probably check in `validatorsmap` whether the committee has stopped, and if so, remove it from the map. and still keep the functions safe.

I'm leaning more towards 2, what do you guys think? @anatolie-ssv , @MatusKysel , @moshe-blox

anatolie-ssv:
> This solution indeed works at the moment, but there's something I don't like about it. We still have `AddShare `and `RemoveShare` methods in `committee.go` . This means someone can use them without the atomicity you just added. This means if someone add a usage of them in the future, we might have a race condition again. I can think of 2 options we have here:
> 
>     1. Emphasize in name and comments that these methods are practically unsafe, and that any consumer should use the atomic function in `ValidatorsMap`
> 
>     2. Move the atomicity to inside the `Committee` struct, which will inherently stop itself when it has no shares left. this way you can probably check in `validatorsmap` whether the committee has stopped, and if so, remove it from the map. and still keep the functions safe.
> 
> 
> I'm leaning more towards 2, what do you guys think? @anatolie-ssv , @MatusKysel , @moshe-blox

I moved the atomicity functionality to the committee object.

protocol/v2/ssv/validator/committee.go
```diff
@@ -87,6 +90,9 @@ func (c *Committee) RemoveShare(validatorIndex phase0.ValidatorIndex) {
 	if share, exist := c.Shares[validatorIndex]; exist {
 		c.dutyGuard.StopValidator(share.ValidatorPubKey)
 		delete(c.Shares, validatorIndex)
+		if len(c.Shares) == 0 {
+			c.stop()
```
anatolie-ssv, line 94: @y0sher now that we are able to stop the committee should we be checking its state before doing actions like `AddShare`, `ProcessMessage` etc?

anatolie-ssv:
deployed to stage `311-314` starting epoch `83532`

operator/validators/validators_map_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+	"testing"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+
+	specssv "github.com/ssvlabs/ssv-spec/ssv"
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+	tests2 "github.com/ssvlabs/ssv/integration/qbft/tests"
```
nkryuchkov, line 13: can you please rename this?

protocol/v2/ssv/validator/committee.go
iurii-ssv: Perhaps it's better (more straightforward) to just call `Committee.cancel` once we determined that committee can stop (shares dropped to 0), presumably this should be enough to finish all the work `Committee` was doing (while it's `controller`'s job to make sure no new tasks will be assigned to such `Committee` - which it does handle by atomically removing it from `controller.validatorsMap`),

and, from my reading of this code, unless we do something with stopped `Committee` (have some business logic for it in stopped state, like restarting it) there is no reason to even store this `stopped atomic.Bool` (and have corresponding methods `stop` and `Stopped`) - you could simply remove these and adjust the way `UpdateCommitteeAtomic` works (so that instead of checking `Stopped()` func `mutate` will return a `bool` to signal that committee was stopped by this mutation).

operator/validators/validators_map.go
```diff
@@ -150,16 +156,39 @@ func (vm *ValidatorsMap) PutCommittee(pubKey spectypes.CommitteeID, v *validator
...
+	for k, v := range vm.committees {
+		if v.Stopped() {
+			stopped = append(stopped, k)
+		}
+	}
+	for _, k := range stopped {
+		delete(vm.committees, k)
 	}
-	return nil
 }
```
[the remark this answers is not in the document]
anatolie-ssv: just to cleanup, fixed now!

operator/validators/validators_map_test.go
```diff
@@ -0,0 +1,87 @@
...
+		ctx := context.Background()
+		vm := validators.New(ctx)
+		ctx, cancel := context.WithCancel(ctx)
+
+		cmt := validator.NewCommittee(
+			ctx,
+			cancel,
+			zap.NewNop(),
+			tests.NewTestingBeaconNodeWrapped().GetBeaconNetwork(),
+			&new(specssv.Committee).CommitteeMember, nil,
```
nkryuchkov, line 38:
```suggestion
			&spectypes.CommitteeMember{}, nil,
```

operator/validator/controller.go
```diff
@@ -980,10 +976,9 @@ func (c *controller) onShareInit(share *ssvtypes.SSVShare) (*validators.Validato
 
 		committeeOpIDs := types.OperatorIDsFromOperators(operator.Committee)
 
-		logger := c.logger.With([]zap.Field{
+		logger := c.logger.With(
 			zap.String("committee", fields.FormatCommittee(committeeOpIDs)),
-			zap.String("committee_id", hex.EncodeToString(operator.CommitteeID[:])),
-		}...)
+			zap.String("committee_id", hex.EncodeToString(operator.CommitteeID[:])))
```
nkryuchkov, line 981:
```suggestion
			fields.CommitteeID(operator.CommitteeID))
```

operator/validators/validators_map_test.go
```diff
@@ -0,0 +1,87 @@
...
+		vm := validators.New(ctx)
+		ctx, cancel := context.WithCancel(ctx)
+		t.Cleanup(cancel)
+
+		cmt := validator.NewCommittee(
+			ctx,
+			cancel,
+			zap.NewNop(),
+			tests.NewTestingBeaconNodeWrapped().GetBeaconNetwork(),
+			&new(specssv.Committee).CommitteeMember, nil,
```
nkryuchkov, line 68:
```suggestion
			&spectypes.CommitteeMember{}, nil,
```

nkryuchkov (approved):

operator/validator/controller.go
```diff
@@ -698,10 +698,9 @@ func (c *controller) UpdateValidatorsMetadata(data map[spectypes.ValidatorPK]*be
 			if err != nil {
 				c.logger.Warn("could not start validator", zap.Error(err))
 			}
-			vc, found := c.validatorsMap.GetCommittee(v.Share().CommitteeID())
-			if found {
+			_ = c.validatorsMap.UpdateCommitteeAtomic(v.Share().CommitteeID(), func(vc *validator.Committee) {
```
iurii-ssv, line 701: Shouldn't we log an error or something when `UpdateCommitteeAtomic` returns false here (just as a sanity check/thing) ?

operator/validator/controller.go
```diff
@@ -967,8 +957,14 @@ func (c *controller) onShareInit(share *ssvtypes.SSVShare) (*validators.Validato
...
 	}
 
+	var vc *validator.Committee
+
 	// Start a committee validator.
-	vc, found := c.validatorsMap.GetCommittee(operator.CommitteeID)
+	found = c.validatorsMap.UpdateCommitteeAtomic(operator.CommitteeID, func(fvc *validator.Committee) {
+		fvc.AddShare(&share.Share)
+		vc = fvc
+	})
```
iurii-ssv, line 966: Would be cleaner to return `validator.Committee` from `UpdateCommitteeAtomic`, wdyt ?

operator/validators/validators_map.go
```diff
@@ -139,6 +141,10 @@ func (vm *ValidatorsMap) GetCommittee(pubKey spectypes.CommitteeID) (*validator.
 
 	v, ok := vm.committees[pubKey]
 
+	if !ok || v.Stopped() {
+		return nil, false
+	}
```
iurii-ssv, line 146: It doesn't seem like "stopped" committee can be present in `vm.committees` map because we immediately remove stopped committees from `UpdateCommitteeAtomic`, so perhaps no need to add this check here ?

iurii-ssv (approved):
The code should work, I left couple of nits, but mostly [I don't quite understand why we need the concept of "stopped" committee](https://github.com/ssvlabs/ssv/pull/1760#discussion_r1795462674) - if it isn't something we are planning to use I'd rather we simplify code by getting rid of it.

operator/validator/controller.go
```diff
@@ -889,20 +888,11 @@ func (c *controller) onShareStop(pubKey spectypes.ValidatorPK) {
...
-				c.logger.Warn("could not find committee to remove on no validators",
-					fields.CommitteeID(v.Share().CommitteeID()),
-					fields.PubKey(pubKey[:]),
-				)
-				return
-			}
-			deletedCommittee.Stop()
-		}
+	}) {
+		c.logger.Warn("committee not found", fields.PubKey(pubKey[:]))
```
iurii-ssv, line 895: Similar to https://github.com/ssvlabs/ssv/pull/1760/files#r1840850336, probably want to log this as `Error` (not `Warn`) ? Because if committee is not found here - looks like something is wrong with the way we manage it's lifecycle, and we'd need to address it (warnings in logs are pretty easy to miss, compared to errors).

anatolie-ssv: As I remember in past conversations it was decided not to change the log level and keep it as it was previously (line 898)

operator/validator/controller.go
anatolie-ssv: We could, but there are 3 call sites to this function and only this one would benefit from the extra return value. @y0sher @MatusKysel your take here?

moshe-blox:
@anatolie-ssv do you still think this is needed?

anatolie-ssv:
> @anatolie-ssv do you still think this is needed?

Yes, I think it's still needed but as of right now it's incomplete. It's needed because of the semantics of the `ValidatorsMap` solution: it has this hierarchy of `ValidatorsMap` -> `Committee` -> `Share`, but the use case is that we add shares to a committee on the level of the map, hence the need for a mechanism to ensure that it's done atomically. The reason why I think it's incomplete is that I don't fully understand the decisions taken around designing the `Committee` object, specifically what does it imply for a committee to be **stopped** (do we need it in the first place? if yes - do we block adding new shares to a stopped committee? I'd say yes but I don't want to speculate). What I would like to have is a discussion that would also involve @y0sher and see where we want to go with it.

protocol/v2/ssv/validator/committee.go
```diff
@@ -292,8 +298,13 @@ func (c *Committee) unsafePruneExpiredRunners(logger *zap.Logger, currentSlot ph
...
 }
 
-func (c *Committee) Stop() {
+func (c *Committee) Stopped() bool {
+	return c.stopped.Load()
+}
+
+func (c *Committee) stop() {
 	c.cancel()
+	c.stopped.Store(true)
```
y0sher, line 307: Maybe @iurii-ssv is right and its best to just check the context? I tend to agree.
@MatusKysel @anatolie-ssv @moshe-blox wdyt?

moshe-blox: +1

moshe-blox:
@anatolie-ssv oh right i see that we're currently stopping & removing in one goroutine (event syncing -> onShareStop) and creating & starting in another goroutine (metadata fetch -> startValidator())

regarding your question: committees are stopped after all their validators were stopped (after removal/liquidation in contract, or Beacon metadata change to exit/slashed/etc)

when a committee is stopped, the memory it was using should be garbage collected, its goroutines exited, and so on...

protocol/v2/ssv/validator/committee.go
```diff
@@ -292,7 +295,11 @@ func (c *Committee) unsafePruneExpiredRunners(logger *zap.Logger, currentSlot ph
 	return nil
 }
 
-func (c *Committee) Stop() {
+func (c *Committee) Stopped() bool {
+	return c.ctx.Err() == context.Canceled
```
y0sher, line 299: maybe just `ctx.Err() != nil` ?

anatolie-ssv: fixed

operator/validator/controller.go
```diff
@@ -835,7 +828,8 @@ func (c *controller) onShareInit(share *ssvtypes.SSVShare) (*validator.Validator
 	}
 
 	// Start a committee validator.
-	vc, found := c.validatorsMap.GetCommittee(operator.CommitteeID)
+	vc, found := c.validatorsMap.AddShareToCommittee(share)
```
y0sher, line 831: honestly this doesn't look atomic.
if there was no committee here, its not added, goes into `!found` there could be other share added to the same committee and it'll also go to `!found` then only one will be added and it will only have the latest share. 

though it might not be needed if this could never happen in parallel..

y0sher: Either make this atomic as well or make sure it can never go twice to this codepath and add a comment why you're not taking care of atomicity here.

moshe-blox:
IMO this PR has over time become too complex.

I think we should rethink this from scratch, starting with the basic questions of what is exactly running in parallel and what is the simplest thing we can do to ensure safety?

anatolie-ssv:
> IMO this PR has over time become too complex.
> 
> I think we should rethink this from scratch, starting with the basic questions of what is exactly running in parallel and what is the simplest thing we can do to ensure safety?

Closing this one, an investigation with a follow up solution should be added to the backlog.

PR #1761, opened 2024-09-24
feat: prevent incompatible version downgrade

### LINKED ISSUE #1759: feat: Prevent usage of alan database in main
### Description 

`Alan` release is changing the way our database is structured, specifically the `Share`s structure loading and saving, we discovered the change is not backward compatible, although its not easily indicated, so if some operator decides to downgrade their operator version back to pre-alan code, its going to run normally but we'll have performance because of the corrupt database.

### Task

Write a migration to change `ConfigLock` mechanism in the new version, so once a pre-alan version tried to open it, it will fail because of the ConfigLock and will not run on the new database.

olegshmuelov:
close #1759 

This PR includes the following changes:

- **Major Version Compatibility Check**: The version is now stored in the ConfigLock struct, and ValidateCompatibility ensures no downgrade on the major version.
- **Migration**: Adds a migration to store the version in ConfigLock.
- **Local Build Versioning**: Introduces a default version (v0.0.0-dev) for local builds and allows custom version input via the Makefile.

The downgrade incompatibility currently applies only to major version changes. We can extend this to include minor version changes to prevent minor updates from requiring major version increments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,11 @@ docker-integration-test:
 #Build
 .PHONY: build
 build:
-	CGO_ENABLED=1 go build -o ./bin/ssvnode -ldflags "-X main.Commit=`git rev-parse HEAD` -X main.Version=`git describe --tags $(git rev-list --tags --max-count=1)`" ./cmd/ssvnode/
+	CGO_ENABLED=1 go build -o ./bin/ssvnode \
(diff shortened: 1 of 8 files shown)


y0sher:
closing because we have https://github.com/ssvlabs/ssv/pull/1762
Not deleting the branch purposely so maybe we can revive this in the future.

PR #1763, opened 2024-09-27
message validation: rearrange checks

nkryuchkov:
move ErrUnexpectedConsensusMessage ErrRoundTooHigh to be triggered as soon as possible to be able to replicate them in the attack sim

@MatheusFranco99 please adjust the knowledge base

### DIFF
--- a/message/validation/consensus_validation.go
+++ b/message/validation/consensus_validation.go
@@ -132,6 +132,30 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
 		return e
 	}
 
+	role := signedSSVMessage.SSVMessage.GetID().GetRoleType()
+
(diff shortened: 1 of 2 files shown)


MatheusFranco99 (approved):
For me the changes are fine 👍

Regarding knowledge base or spec, nonetheless, I'm afraid I have to disagree with changing it. As I see it, the rules for maximum rounds and validity of the consensus message depending on the duty type refer much more to duty logic instead of semantics.

In the knowledge base, easiness of reasoning about the rules is the most important factor. If I mix rules now, I will need to roll back everything in the future so I can reason about it clearly.

For implementation, on the other hand, I believe it's okay to deviate from spec on these orderings.

y0sher (approved):

nkryuchkov:
@y0sher @moshe-blox Can I merge this?

PR #1632, opened 2024-08-15
feat: (API) decided instances endpoint

moshe-blox:
Example:
```
/v1/exporter/instances?from=9742603&to=9743782&roles=PROPOSER,SYNC_COMMITTEE&pubkeys=82679162c8be185ba4e3d9a8763604c4b4a158b761fb36b0da26ebb3b3e286a3f9fc5a29230a8ad52204cc173c9d9c29,8aa3ba481ae7de589c5a0ce20480e93b1e20fd6b66cee2490c2838cf1384bc2fbec4cdfbf40ef610cd8d313b592c7f28,834a32eb0de3e1ae01d0f57ee3aaf2f46576f184ef9a4901d0218acb4c1e4267dc28b446ec1d2472bf70bf2eb7c34942,a4234749af70593709206ca154de0c676353c8a1993bf33707582e9c3ca4029a7f92dec8db4dce65e44cd7690dabd95e
```

```json
{
   "data":[
      {
         "Signature":"i1j+OW4LRvo2lDHlhAsTPQ6HZCE6Y0kJ3LCUPizh5WW4nZWYac+nPj3ZsYQCI0g6CuCeYSlf1x5EMq7yKgQ5zUEioAsgLL3LmwViw4SyP98+QsarXWUDOf0SOB+keZ9S",
         "Signers":[
            6,
            19,
            23
         ],
         "Message":{
            "MsgType":2,
            "Height":9743624,
            "Round":1,
            "Identifier":"AAAAAKQjR0mvcFk3CSBsoVTeDGdjU8ihmTvzNwdYLpw8pAKaf5LeyNtNzmXkTNdpDavZXgMAAAA=",
            "Root":"Si7z8kPOjiZYx0MJ4Rl2asIWmI6/eSJsztis/GQWxvo=",
            "DataRound":0,
            "RoundChangeJustification":[
               
            ],
            "PrepareJustification":[
               
            ]
         },
         "FullData":"FAAAAAEAAAAAAAAAiAAAAIgAAAADAAAAAAAAAKQjR0mvcFk3CSBsoVTeDGdjU8ihmTvzNwdYLpw8pAKaf5LeyNtNzmXkTNdpDavZXgitlAAAAAAAi2wSAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGwAAABeAAAAAAAAADn9Yhdx0LUhQjZL3DuCxJQ+3+tcL4BRhLDcxfxCANvk"
      },
[...]

PR #1766, opened 2024-09-30
exporter: decideds endpoint; fix signers; fix resolving role; cache domain; fix convert

olegshmuelov:
Changes to Exporter:
- port https://github.com/ssvlabs/ssv/pull/1632
    - add decided instances endpoint (`GET /v1/exporter/decideds`)
    - add `X-SSV-Node-Version` header
- fix role type casting when using the `convert` package
- allow saving partial signatures of length not equal to 3f+1
- cache domain requests to beacon node within one epoch
- make resolving beacon role type by roots work
- fix wrong participants rendering by merging committees on new quorum instead of overwriting


networkconfig/holesky-stage.go
```diff
@@ -16,7 +16,7 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        99999999,
+	//AlanForkEpoch:        99999999,
```
y0sher, line 19: lets not merge this

api/handlers/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+	Slot      uint64 `json:"slot"`
+	PublicKey string `json:"public_key"`
+	Message   struct {
+		Signers []uint64 `json:"Signers"`
+	} `json:"message"`
+}
+
+func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
+	var request struct {
+		From    int           `json:"from"`
```
nkryuchkov, line 32: Just something to think of: if 'from' or 'to' are not passed, the method will use zero values and try to fetch decided messages for slot 0, which might not be clear. However, passing 0 to both 'from' and 'to' looks fine to me

nkryuchkov: Also, I think 'from' and 'to' should be better `uint64`s, same as `phase0.Slot`

api/handlers/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(request.PubKeys) == 0 {
+		return api.BadRequestError(fmt.Errorf("at least one public key is required"))
+	}
+	if len(request.Roles) == 0 {
+		return api.BadRequestError(fmt.Errorf("at least one role is required"))
+	}
+
+	response.Data = []*ParticipantResponse{}
+	for _, role := range request.Roles {
+		roleStorage := e.QBFTStores.Get(convert.RunnerRole(role))
```
nkryuchkov, line 60: It's not introduced in this PR but `convert.RunnerRole` has 
```go
	RoleSyncCommitteeContribution = 3
	RoleSyncCommittee = 4
```

but `spectypes.BeaconRole` has 

```go
	BNRoleSyncCommittee = 3
	BNRoleSyncCommitteeContribution = 4
```

, so those roles seem to be inverted here. However, it isn't not part of the code of this PR

api/handlers/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+		if roleStorage == nil {
+			return fmt.Errorf("role storage doesn't exist: %v", role)
+		}
+		for _, pubKey := range request.PubKeys {
+			msgID := convert.NewMsgID(e.DomainType, pubKey, convert.RunnerRole(role))
+			from := phase0.Slot(request.From)
+			to := phase0.Slot(request.To)
+
+			participantsList, err := roleStorage.GetParticipantsInRange(msgID, from, to)
```
nkryuchkov, line 69: `GetParticipantsInRange` doesn't seem to return quorums with less than 3f+1 participants due to how `SaveParticipants` is implemented, however, it's not because of the code in this PR

nkryuchkov (changes requested):

api/handlers/exporter.go
nkryuchkov: @MatusKysel could you please confirm that it's not intended?

api/server/server.go
```diff
@@ -41,20 +45,24 @@ func (s *Server) Run() error {
...
 	router.Use(middlewareLogger(s.logger))
+	router.Use(middlewareNodeVersion)
 
 	router.Get("/v1/node/identity", api.Handler(s.node.Identity))
 	router.Get("/v1/node/peers", api.Handler(s.node.Peers))
 	router.Get("/v1/node/topics", api.Handler(s.node.Topics))
 	router.Get("/v1/node/health", api.Handler(s.node.Health))
 	router.Get("/v1/validators", api.Handler(s.validators.List))
+	router.Get("/v1/exporter/decideds", api.Handler(s.exporter.Decideds))
+	router.Post("/v1/exporter/decideds", api.Handler(s.exporter.Decideds))
```
nkryuchkov, line 56: why do we need `Post`?

olegshmuelov: This is intended to avoid breaking changes at this stage.

api/handlers/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+	DomainType spectypes.DomainType
+	QBFTStores *ibftstorage.QBFTStores
+}
+
+type ParticipantResponse struct {
+	Role      string `json:"role"`
+	Slot      uint64 `json:"slot"`
+	PublicKey string `json:"public_key"`
+	Message   struct {
+		Signers []uint64 `json:"Signers"`
```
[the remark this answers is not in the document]
nkryuchkov: @olegshmuelov can you please add a comment explaining this?

api/server/server.go
nkryuchkov: @olegshmuelov can you please add a comment explaining this?

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -155,10 +160,12 @@ func (ncv *CommitteeObserver) ProcessMessage(msg *queue.SSVMessage) error {
...
 
-func (ncv *CommitteeObserver) getRole(msg *queue.SSVMessage, root [32]byte) convert.RunnerRole {
+func (ncv *CommitteeObserver) getRole(msg *queue.SSVMessage, slot phase0.Slot, validatorIndex phase0.ValidatorIndex) convert.RunnerRole {
 	if msg.MsgID.GetRoleType() == spectypes.RoleCommittee {
-		_, found := ncv.Roots[root]
-		if !found {
+		beacon := ncv.networkConfig.Beacon
+		period := beacon.EstimatedSyncCommitteePeriodAtEpoch(beacon.EstimatedEpochAtSlot(slot))
+		d := ncv.dutyStore.SyncCommittee.Duty(period, validatorIndex)
+		if d == nil {
```
olegshmuelov, line 168: I was considering a similar approach checking for the existence of a duty.

**Pros:**
 - It’s a straightforward method to determine if a duty exists for a specific role.
 - Simplicity: avoids the need to store roots and differentiate roles based on them.
 
**Cons:**
 - It may introduce reorg/indices change anomalies, though these are very rare edge cases.

If we go with this direction, we’ll need to deprecate the root logic, as it wasn't implemented according to the author's intention. We should also consider implementing the original logic as intended to ensure it works as expected.

olegshmuelov: We need to adjust the logic in the `getRole` function to return roles based on the quorum, since the same quorum of operators can execute both attester and sync committee duties at the same slot.

```go
func (ncv *CommitteeObserver) getRoles(
	msgRole spectypes.RunnerRole,
	slot phase0.Slot,
	validatorIndex phase0.ValidatorIndex,
) []convert.RunnerRole {
	roles := make([]convert.RunnerRole, 0)

	if msgRole == spectypes.RoleCommittee {
		beacon := ncv.networkConfig.Beacon
		epoch := beacon.EstimatedEpochAtSlot(slot)
		period := beacon.EstimatedSyncCommitteePeriodAtEpoch(epoch)
		if d := ncv.dutyStore.SyncCommittee.Duty(period, validatorIndex); d != nil {
			roles = append(roles, convert.RoleSyncCommittee)
		}
		if d := ncv.dutyStore.Attester.ValidatorDuty(epoch, slot, validatorIndex); d != nil {
			roles = append(roles, convert.RoleAttester)
		}
	} else {
		roles = append(roles, convert.RunnerRole(msgRole))
	}
	return roles
}
```

This way, we iterate over the returned roles and try to save participants for each of them, ensuring that duties for both attester and sync committee are correctly handled when they occur at the same slot.

y0sher: @nkryuchkov I agree with @olegshmuelov here. actually on both cases.
- We should be able to determine the roles by the roots. I'm not sure why its not working atm but since attester and sync committee have different signing roots, this should be achievable.
- If for some reason we're taking the `dutyStore` approach (IMO a bit more complex since like @olegshmuelov said, reorgs might occur or maybe the duties might be deleted for some other reason), we need to make sure we cover all roles for the slot.

@nkryuchkov if there's any reason why the roots approach can't be achieved lets discuss..

networkconfig/holesky-stage.go
```diff
@@ -16,8 +16,8 @@ var HoleskyStage = NetworkConfig{
 	GenesisEpoch:         1,
 	RegistrySyncOffset:   new(big.Int).SetInt64(84599),
 	RegistryContractAddr: "0x0d33801785340072C452b994496B19f196b7eE15",
-	AlanForkEpoch:        999999999,
-	DiscoveryProtocolID:  [6]byte{'s', 's', 'v', 'd', 'v', '5'},
+	//AlanForkEpoch:        999999999,
```
y0sher, line 19: should not be in this PR even though stage is going to be post-fork from now..

olegshmuelov: > should not be in this PR even though stage is going to be post-fork from now..

yes, should be reverted once argo sensor is deleted

nkryuchkov (approved):

api/handlers/exporter.go
```diff
@@ -0,0 +1,108 @@
...
+		return api.BadRequestError(fmt.Errorf("at least one role is required"))
+	}
+
+	response.Data = []*ParticipantResponse{}
+	for _, role := range request.Roles {
+		runnerRole := casts.BeaconRoleToConvertRole(spectypes.BeaconRole(role))
+		roleStorage := e.QBFTStores.Get(runnerRole)
+		if roleStorage == nil {
+			return fmt.Errorf("role storage doesn't exist: %v", role)
+		}
```
y0sher, line 65: - maybe its better to check for existence of all roles before this loop. 
-  why not returning `api.BadRequestError` here? 
- maybe we should continue to next role if some of them do exist?

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -106,47 +130,86 @@ func (ncv *CommitteeObserver) ProcessMessage(msg *queue.SSVMessage) error {
...
+			if roleStorage == nil {
+				return fmt.Errorf("role storage doesn't exist: %v", beaconRole)
+			}
+
+			existingQuorum, err := roleStorage.GetParticipants(msgID, slot)
+			if err != nil {
+				return fmt.Errorf("could not get participants %w", err)
+			}
+
+			if len(existingQuorum) > len(quorum) {
```
y0sher, line 166: `>=` ? 
do we want to override previous quorum of same amount? (override 1,2,3 with 1,2,4)

y0sher: though it should not happen..

api/handlers/exporter.go
nkryuchkov: I agree with the first two, thanks. I think we don't need to continue if a role doesn't exist because it would indicate that QBFT stores are not setup correctly

protocol/v2/ssv/validator/non_committee_validator.go
nkryuchkov: We don't need to override it, but now I think we need to merge quorums. If we receive 1,2,3,4,5 and then 2,3,4,5,6,7, then we'll lose 1. Pushed a fix

nkryuchkov (changes requested):
Requesting changes not to forget to revert AlanForkEpoch once it's tested

networkconfig/holesky-stage.go
y0sher: next time just please create a separate `deploy` branch to keep the main PR clean.

protocol/v2/ssv/validator/domain_cache.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/jellydator/ttlcache/v3"
+
+	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
+)
+
+type DomainCache struct {
+	beaconNode beacon.BeaconNode
+	cache      *ttlcache.Cache[domainCacheKey, phase0.Domain]
```
y0sher, line 14: seems to be threadsafe from looking at the code

protocol/v2/ssv/validator/domain_cache.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"time"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/jellydator/ttlcache/v3"
+
+	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
+)
+
+type DomainCache struct {
```
y0sher, line 12: nice addition, we should probably use this in the actual runners and not just exporter! of course in a different PR.

y0sher (approved):
@olegshmuelov @nkryuchkov 
Good job on figuring out all the quirks here! this is a big transition in exporter.

networkconfig/holesky-stage.go
nkryuchkov: reverted

nkryuchkov (approved):

networkconfig/holesky-stage.go
nkryuchkov: reverted

moshe-blox (approved):
LGTM

well done @olegshmuelov @nkryuchkov @y0sher  💪 

let's not forget to let @andrew-blox know that this should be QA'd before release (this will go in v2.0.2 after the fork)

PR #1768, opened 2024-09-30
pubsub: set max msg size

nkryuchkov:
Currently, the max message size supported by the protocol is `6286507` but default pub-sub max message size is 1 MiB according to https://github.com/libp2p/go-libp2p-pubsub/blob/f71345c1ec0ee4b30cd702bb605927f788ff9f36/pubsub.go#L495.

The PR changes the max pub-sub message size to MaxEncodedMsgSize (`6286507` as of now).

@MatheusFranco99 @GalRogozinski `go-libp2p-pubsub` discourages increasing the max size, should we consider reducing it?

### DIFF
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -54,5 +54,7 @@ const (
 const (
 	maxPayloadDataSize = max(maxEncodedConsensusMsgSize, maxEncodedPartialSignatureSize)
 	maxSignedMsgSize   = maxSignaturesSize + maxOperatorIDSize + msgTypeSize + identifierSize + maxPayloadDataSize + maxFullDataSize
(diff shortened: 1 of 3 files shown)


y0sher (approved):

MatheusFranco99 (approved):
I think we can't reduce it because it's tied to the max size of a `BeaconBlock` that can be put in the `FullData` field.

nkryuchkov:
@y0sher @moshe-blox Can I merge this?

moshe-blox:
@nkryuchkov @MatheusFranco99 does the libp2p max msg size count compressed bytes or just the raw msg?

(im not sure if we have compression or not, but we might if its on by default)

PR #1773, opened 2024-10-07
gitignore: .idea folder

iurii-ssv:
I've installed `Gitlink` plugin and now the following file shows up in staged changes: `.idea/GitLink.xml`,

unless there is a specific reason to ignore only certain files in `.idea` folder, seems like the best way to address this would be to just gitignore whole `.idea` folder (as opposed to forcing developer to configure his own local `.git/info/exclude`).


iurii-ssv:
I mean, there is some 2-3 years old configuration stored in there, just doesn't seem like it's a good way to share it (if it is still being used by anybody even):

image

moshe-blox (approved):
lol, i never even noticed we have `.idea` in the repo, makes sense to remove it

nkryuchkov (approved):
@moshe-blox IIRC, we had it for Goland configurations to run 4 ssv nodes, but yes, it is likely outdated

PR #1778, opened 2024-10-07
perf(registry/storage): don't lock in-memory shares storage for database ops 

anatolie-ssv:
The approach taken here is to minimize the time spent with a locked `shares` map.
By separately locking the badger instance we allow access to the in-memory shares while the "heavy lifting" disk access is being executed.

### DIFF
--- a/registry/storage/shares.go
+++ b/registry/storage/shares.go
@@ -210,8 +210,6 @@ func (s *sharesStorage) Save(rw basedb.ReadWriter, shares ...*types.SSVShare) er
 		}
 	}
 
-	s.logger.Debug("save validators to shares storage", zap.Int("count", len(shares)))
-
 	s.storageMtx.Lock()
 	defer s.storageMtx.Unlock()
 
@@ -236,11 +234,11 @@ func (s *sharesStorage) Save(rw basedb.ReadWriter, shares ...*types.SSVShare) er
 		}
 
 		if err := s.validatorStore.handleSharesUpdated(updateShares...); err != nil {
-			return err
+			return fmt.Errorf("handleSharesUpdated: %w", err)
 		}
 
 		if err := s.validatorStore.handleSharesAdded(addShares...); err != nil {
-			return err
+			return fmt.Errorf("handleSharesAdded: %w", err)
 		}
 
 		return nil
--- a/registry/storage/shares.go
+++ b/registry/storage/shares.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/gob"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"sync"
 
@@ -57,8 +58,8 @@ type sharesStorage struct {
 	prefix         []byte
 	shares         map[string]*types.SSVShare
 	validatorStore *validatorStore
+	dbmu           sync.Mutex // parent lock for in-memory mutex
 	mu             sync.RWMutex
-	dbmu           sync.Mutex
 }
 
 type storageOperator struct {
@@ -133,12 +134,7 @@ func NewSharesStorage(logger *zap.Logger, db basedb.Database, prefix []byte) (Sh
 
 // load reads all shares from db.
 func (s *sharesStorage) load() error {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	s.dbmu.Lock()
-	defer s.dbmu.Unlock()
-
+	// not locking since at this point nobody has the reference to this object
 	return s.db.GetAll(append(s.prefix, sharesPrefix...), func(i int, obj basedb.Obj) error {
 		val := &storageShare{}
 		if err := val.Decode(obj.Value); err != nil {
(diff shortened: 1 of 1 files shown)


anatolie-ssv:
@moshe-blox I'm looking at ways to move the interaction with the database in a separate goroutine.
Right now the problem that I'm facing is the inability to propagate the potential database error from the goroutine to the method updating the in-memory shares map. So if there is an error from badger db we would find out about it only after we've made changes to the shares map. I'm not really sure how to get around this issue.

registry/storage/shares.go
```diff
@@ -58,6 +58,7 @@ type sharesStorage struct {
 	shares         map[string]*types.SSVShare
 	validatorStore *validatorStore
 	mu             sync.RWMutex
+	dbmu           sync.Mutex
```
iurii-ssv, line 61: Would be nice to have some comment for each of these mutex(s) clarifying how they are different.

registry/storage/shares.go
```diff
@@ -364,35 +369,17 @@ func (s *sharesStorage) UpdateValidatorsMetadata(data map[spectypes.ValidatorPK]
...
-	}
-
-	return nil
+	return s.Save(nil, shares...)
 }
 
 // Drop deletes all shares.
 func (s *sharesStorage) Drop() error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
```
iurii-ssv, line 378: Can maybe lock `mu` just for these lines (and then also have to `dbmu.Unlock()` before going for `mu.Lock`):
```go
s.shares = make(map[string]*types.SSVShare)
s.validatorStore.handleDrop()
```
Essentially, same optimization you've made for `sharesStorage.Save` above, seems like this is applicable to other func(s) too, eg. `Delete` (and maybe others).

iurii-ssv:
> @moshe-blox I'm looking at ways to move the interaction with the database in a separate goroutine. Right now the problem that I'm facing is the inability to propagate the potential database error from the goroutine to the method updating the in-memory shares map. So if there is an error from badger db we would find out about it only after we've made changes to the shares map. I'm not really sure how to get around this issue.

One pattern I've seen for this is:
- instead of returning an `error` (from func(s) like `Save`) you return `chan error`
- you pass this `chan error` to background go-routine to either close it (important) or send an error through this channel
- the caller of `Save` now has to receive from channel if he intends to handle the potential error (that might eventually come on this `chan error`), he can also handle it by spinning a go-routine of his own that will listen on this `chan error` and maybe log the result or something

It does complicate `sharesStorage` API and usage quite a bit, the question is whether there are real callers that can benefit from this ?

moshe-blox:
@anatolie-ssv oh that's right during event sync we can't ignore DB errors, i forgot!

@iurii-ssv returning `<-chan error` is an interesting idea to simulate async/await behavior, with that said i like Go's approach of blocking by default and delegating async to the caller, which can simply spawn a goroutine if they wish, because it keeps the API trivial as you noted.

i think going for a pure solution might be too complex, because what does it mean if a database save fail — should we undo the changes we've done in the map?

maybe we should try to target the specific issue we have and witnessed causing a slowdown in the real world — that updating of thousands of share metadatas completely locks the storage for many seconds at a time

i'm not aware of other points of pressure around shares than that (so far!)

WDYT?

iurii-ssv:
> i'm not aware of other points of pressure around shares than that (so far!)

Yeah, that's what I was getting at. If the performance of fully blocking implementation is sufficient - there is no need to go for the harder solution.

> with that said i like Go's approach of blocking by default and delegating async to the caller, who can simply spawn a goroutine if they wish, because it keeps the API trivial as you noted, and allows the caller to avoid unnecessary channels when they do need to block.

Yup, the simpler the better. I was just describing a potential solution (with it's tradeoffs) if we need to go down this path sometime in the future.

registry/storage/shares.go
anatolie-ssv: added

registry/storage/shares.go
```diff
@@ -315,95 +318,88 @@ func (s *sharesStorage) storageShareToSpecShare(share *storageShare) (*types.SSV
...
 
-	func() {
-		s.mu.RLock()
-		defer s.mu.RUnlock()
+	s.dbmu.Lock()
+	defer s.dbmu.Unlock()
+
+	err := func() error {
+		s.mu.Lock()
+		defer s.mu.Unlock()
```
iurii-ssv, line 362: Here we can use `RLock ?

iurii-ssv (approved):
LGTM, but it seems we lock `dbmu` for longer times than necessary (I assume it's only goal is to manage access to `sharesStorage.db`).

registry/storage/shares.go
```diff
@@ -57,6 +58,7 @@ type sharesStorage struct {
 	prefix         []byte
 	shares         map[string]*types.SSVShare
 	validatorStore *validatorStore
+	dbmu           sync.Mutex // parent lock for in-memory mutex
 	mu             sync.RWMutex
```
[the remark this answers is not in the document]
moshe-blox: i'd suggest renaming to something like `storageMtx` and `memoryMtx` because it's the fundamental difference between the two

registry/storage/shares.go
anatolie-ssv: re-added `s.shares[key] = share` in the same scope

registry/storage/shares.go
```diff
@@ -210,48 +210,57 @@ func (s *sharesStorage) Save(rw basedb.ReadWriter, shares ...*types.SSVShare) er
...
-		return err
-	}
-	if err := s.validatorStore.handleSharesAdded(addShares...); err != nil {
+		if err := s.validatorStore.handleSharesUpdated(updateShares...); err != nil {
+			return err
+		}
+
+		if err := s.validatorStore.handleSharesAdded(addShares...); err != nil {
+			return err
+		}
```
iurii-ssv, line 244: Would be nice to wrap errors here (because they come from different code branches might be hard to debug without it):
```go
		if err := s.validatorStore.handleSharesUpdated(updateShares...); err != nil {
			return fmt.Erorrf("handleSharesUpdated: %w", err)
		}

		if err := s.validatorStore.handleSharesAdded(addShares...); err != nil {
			return fmt.Erorrf("handleSharesAdded: %w", err)
		}
```

registry/storage/shares.go
```diff
@@ -210,48 +210,57 @@ func (s *sharesStorage) Save(rw basedb.ReadWriter, shares ...*types.SSVShare) er
 		}
 	}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.logger.Debug("save validators to shares storage", zap.Int("count", len(shares)))
```
iurii-ssv, line 213: nit: `save validators to shares storage` -> `save validator shares to shares storage`

moshe-blox (approved):
LGTM 🔥

registry/storage/shares.go
moshe-blox: oops, missed this one 😓 

we can probably do this in next PR to storage

PR #1785, opened 2024-10-10
fix: gosec warnings

MatusKysel:


protocol/v2/types/ssvshare.go
```diff
@@ -157,7 +157,7 @@ func ComputeCommitteeID(committee []spectypes.OperatorID) spectypes.CommitteeID
 	// Convert to bytes
 	bytes := make([]byte, len(committee)*4)
 	for i, v := range committee {
-		binary.LittleEndian.PutUint32(bytes[i*4:], uint32(v)) // nolint:gosec
+		binary.LittleEndian.PutUint32(bytes[i*4:], uint32(v)) // #nosec
```
zippoxer, line 160: let's please disable the specific rule so we don't miss any future rules that should activate on this line

moshe-blox, line 160: let's please disable only the specific rule to avoid missing any future rules

eth/eventhandler/handlers.go
```diff
@@ -95,6 +95,7 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
 	return nil
 }
 
+// This function is currently no-op and it will do nothing. Operator removed event is not used in the current implementation.
```
moshe-blox, line 98: i think we can remove the commented code inside and move this comment there, also clarifying that it will do nothing beyond validating the operator exists

moshe-blox (approved):

y0sher (approved):

Issue #1703, opened 2024-09-05
optimization: (alan) only listen and get metatdata of committee validators


y0sher:
### Description

Mainly because message validation, our nodes gets metadata for all network vaildators (for mainnet now almost 40k), this overloads the ssv-node and the beacon node that has to track this validators, the message validation itself for all these validator messages also overloads the node. 

This can work only in `alan` since previously having 500 validators made you listen to all network subnets hence you needed to know all validators.

### Solution 

- Consider to pull metadata from beacon node only about validators that are part of your listened subnets.
- Reject/ignore all messages that are not part of your subnet validators (this might happen automatically since you won't have their status and duties)

PR #1787, opened 2024-10-10
feat(validator): listen and get metadata only for committee validators

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/1703

Note: each `selfSubnets` call takes approx. 300-900 µs


y0sher (changes requested):
I know this pr is still immature. quoting the issue here to clarify my original intentions - 

> 
    - Consider to pull metadata from beacon node only about validators that are part of your listened subnets.
    - Reject/ignore all messages that are not part of your subnet validators (this might happen automatically since you won't have their status and duties)

So about first point I see you started it, but your using the `ValidatorSubnet` calculation which is pre-alan way to find your subnets. We now use `CommiteeSubnet` and have a different way of dividing our validators to subnets (according to committees).

About second point I don't see that you did anything about message validation, though if we don't have any metadata about this operator's validators than we might just naturally reject/ignore it? please make sure what happens in this case.

nkryuchkov:
> So about first point I see you started it, but your using the `ValidatorSubnet` calculation which is pre-alan way to find your subnets. We now use `CommiteeSubnet` and have a different way of dividing our validators to subnets (according to committees).

Added usage of `CommiteeSubnet`, please review again

> About second point I don't see that you did anything about message validation, though if we don't have any metadata about this operator's validators than we might just naturally reject/ignore it? please make sure what happens in this case.

In this case the message gets ignored:
```go
	ErrNoShareMetadata                         = Error{text: "share has no metadata"}
```

nkryuchkov:
> So about first point I see you started it, but your using the `ValidatorSubnet` calculation which is pre-alan way to find your subnets. We now use `CommiteeSubnet` and have a different way of dividing our validators to subnets (according to committees).

Added post-fork handling, please review again

> About second point I don't see that you did anything about message validation, though if we don't have any metadata about this operator's validators than we might just naturally reject/ignore it? please make sure what happens in this case.

If there's no metadata, a message gets ignored by message validation:
```go
	ErrNoShareMetadata                         = Error{text: "share has no metadata"}
```

operator/validator/controller.go
```diff
@@ -517,12 +521,22 @@ func (c *controller) StartValidators() {
...
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		allPubKeys = append(allPubKeys, share.ValidatorPubKey[:])
+		committeeSubnet := byte(commons.CommitteeSubnet(share.CommitteeID()))
+		if bytes.Contains(c.network.ActiveSubnets(), []byte{committeeSubnet}) {
+			pubKeysToFetch = append(pubKeysToFetch, share.ValidatorPubKey[:])
+		}
+
+		if !c.networkConfig.PastAlanFork() {
```
y0sher, line 534: since this version is going to be released after mainnet fork, there's no reason to include fork code really, we're planning to remove all pre-fork logic in this release anyway.

y0sher (changes requested):
- Remove prefork code, as I explained its not needed anymore
- to make this PR final we need to POC  that it works on stage
- Compare cpu mem usage
- compare how this impacts the overload on beacon node
- We probably need to ignore messages that we don't have duties or we don't see as active (because they're not ours). instead of rejecting them.
- @MatheusFranco99 can you chime in in terms of protocol and networking, other than more well spread messages, any reason why ignoring these message will impact us badly?

MatheusFranco99:
> @MatheusFranco99 can you chime in in terms of protocol and networking, other than more well-spread messages, any reason why ignoring these messages will impact us badly?

@y0sher 
Do you mean ignoring messages from validators that we think are liquidated / not active / exited or whatever similar? I think it's not a problem to ignore instead of reject. We also have ignoring protection against an attack so we should be fine.

But we must not do this if you meant ignoring messages about validators that we are not part of the committee.

nkryuchkov:
> * Remove prefork code, as I explained its not needed anymore

done

> * to make this PR final we need to POC  that it works on stage

I'm still testing it but so far it seems to be working well

image

> * Compare cpu mem usage

It was deployed between 23:46 and 00:28 and since 01:13

CPU dropped from ~2 to ~0.4-0.5:

image

Memory might have dropped a little bit but the difference is not that huge, also it differs between runs:

image

Receive bandwidth dropped from ~1.3 to ~0.4-0.5, transmit bandwidth just a little bit:

image

Although, for another node the transmit bandwidth difference is significant:

image

> * compare how this impacts the overload on beacon node

I don't see any difference

CPU:

image

Mem:

image

> * We probably need to ignore messages that we don't have duties or its validator we don't see as active (because they're not ours). instead of rejecting them.

Well, we ignore messages if there's no metadata, so if they're not ours, we should have no metadata for them, hence the message is ignored

MatheusFranco99:
> Well, we ignore messages if there's no metadata, so if they're not ours, we should have no metadata for them, hence the message is ignored

@nkryuchkov 
When you say "if they're not ours", you mean non SSV validators, right? Or do you mean non-committee validators? Or validators from topics we're not subscribed to? 😅 
I think that's the part I got a bit confused

Great job!

nkryuchkov:
@MatheusFranco99 sorry for the confusion, I meant validators from topics we're not subscribed to

operator/validator/controller.go
```diff
@@ -517,12 +521,15 @@ func (c *controller) StartValidators() {
...
 	var ownShares []*ssvtypes.SSVShare
-	var allPubKeys = make([][]byte, 0, len(shares))
+	var pubKeysToFetch [][]byte
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		allPubKeys = append(allPubKeys, share.ValidatorPubKey[:])
+		committeeSubnet := byte(commons.CommitteeSubnet(share.CommitteeID()))
+		if bytes.Contains(c.network.ActiveSubnets(), []byte{committeeSubnet}) {
```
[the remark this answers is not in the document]
nkryuchkov: great catch, thanks

operator/validator/controller.go
```diff
@@ -517,12 +521,16 @@ func (c *controller) StartValidators() {
...
-	var allPubKeys = make([][]byte, 0, len(shares))
+	var pubKeysToFetch [][]byte
+	activeSubnets := c.network.ActiveSubnets()
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		allPubKeys = append(allPubKeys, share.ValidatorPubKey[:])
+		committeeSubnet := byte(commons.CommitteeSubnet(share.CommitteeID()))
+		if bytes.Contains(activeSubnets, []byte{committeeSubnet}) {
```
y0sher, line 531: I believe we have `SharedSubnets` and some other subnets utility functions. its best to use a function across the code to make changing and bug fixing easier.

operator/validator/controller.go
```diff
@@ -1139,6 +1148,12 @@ func (c *controller) UpdateValidatorMetaDataLoop() {
 			if share.Liquidated {
 				return true
 			}
+
+			committeeSubnet := byte(commons.CommitteeSubnet(share.CommitteeID()))
+			if !bytes.Contains(activeSubnets, []byte{committeeSubnet}) {
```
y0sher, line 1153: ditto

operator/validator/controller.go
```diff
@@ -1131,6 +1139,7 @@ func (c *controller) UpdateValidatorMetaDataLoop() {
 	const batchSize = 512
 	var sleep = 2 * time.Second
 
+	activeSubnets := c.network.ActiveSubnets()
```
moshe-blox, line 1142: for performance reasons, because this would be checked >44k times on Mainnet, we should change it to [SubnetCount]bool and then we can just check the array at the subnet index instead of using bytes.Contains

nkryuchkov: @moshe-blox There was a bug in my implementation, I don't need to use `bytes.Contains` actually, I can just check the index. I don't think we need to have an array in this PR just in this method, but we need to refactor subnets to have array as an underlying type

moshe-blox (approved):
LGTM 🔥 

fixed a bug and pushed: https://github.com/ssvlabs/ssv/pull/1787/commits/ac49a96947d3ae27fb44b491b382d3a71be27525

moshe-blox:
> > @MatheusFranco99 can you chime in in terms of protocol and networking, other than more well-spread messages, any reason why ignoring these messages will impact us badly?
> 
> @y0sher Do you mean ignoring messages from validators that we think are liquidated / not active / exited or whatever similar? I think it's not a problem to ignore instead of reject. We also have ignoring protection against an attack so we should be fine.
> 
> But we must not do this if you meant ignoring messages about validators that we are not part of the committee.

@nkryuchkov @y0sher this means we should change the PR a little

is this the case now in message validation with this PR?

if we receive a message with committee `X` for validator that's actually in committee `Y` (where both aren't our committees), would we reject it for invalid committee or ignore for missing metadata?

operator/validator/controller.go
```diff
@@ -517,12 +520,16 @@ func (c *controller) StartValidators() {
...
+	var pubKeysToFetch [][]byte
+	activeSubnets := c.network.ActiveSubnets()
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		allPubKeys = append(allPubKeys, share.ValidatorPubKey[:])
+		subnet := commons.CommitteeSubnet(share.CommitteeID())
+		if uint64(len(activeSubnets)) >= subnet && activeSubnets[subnet] != 0 {
+			pubKeysToFetch = append(pubKeysToFetch, share.ValidatorPubKey[:])
```
y0sher, line 531: Honestly this sounds a bit cycle-ish for me. We are subscribing to validators'/committee's subnet after a metadata fetch that indicates the validator is active_ongoing. then here we are only fetching validators that show up on p2p active subnets?

Have you tested this on a fresh db sync? Am I missing something? @nkryuchkov @moshe-blox 
I feel like we should probably get the list of validators/committees to fetch the database based on the contract sync. And the p2p active subnets should be a result of this fetch.

So 
contract sync -> validator list -> metadata fetch(validator list) -> active subnets (fetched as active_ongoing).

moshe-blox: @y0sher good point, agree 👌

feels like an "import cycle" 🤣 

@nkryuchkov we should fetch metadata for validators in all subnets of our committees, whether they're active or not (because we don't really know until we fetch)

perhaps instead of calling `c.network.ActiveSubnets()` we should call a func that returns the subnets of all of our committees (whether active or not), and the rest of the logic can remain the same.

i think we might already have this logic in the DynamicMaxPeers PR, so maybe we just need to export it to a func

nkryuchkov: @y0sher @moshe-blox please review if https://github.com/ssvlabs/ssv/pull/1787/commits/fd45bebf3683c7c2c9791f31cefc8536d396c13d is what you meant

operator/validator/controller.go
```diff
@@ -494,20 +502,33 @@ func (c *controller) StartValidators() {
...
 }
 
+func (c *controller) selfSubnets() records.Subnets {
+	myValidators := c.validatorStore.OperatorValidators(c.operatorDataStore.GetOperatorID())
+	mySubnets := make(records.Subnets, networkcommons.SubnetsCount)
+	for _, v := range myValidators {
+		subnet := networkcommons.CommitteeSubnet(v.CommitteeID())
+		if mySubnets[subnet] == 0 {
+			mySubnets[subnet] = 1
+		}
```
y0sher, line 526:
```suggestion
		mySubnets[subnet] = 1
```

operator/validator/controller.go
```diff
@@ -969,13 +990,21 @@ func (c *controller) UpdateValidatorMetaDataLoop() {
...
+
 		// Get the shares to fetch metadata for.
 		start := time.Now()
 		var existingShares, newShares []*ssvtypes.SSVShare
 		c.sharesStorage.Range(nil, func(share *ssvtypes.SSVShare) bool {
 			if share.Liquidated {
 				return true
 			}
+
+			subnet := networkcommons.CommitteeSubnet(share.CommitteeID())
```
y0sher, line 1003: you don't want to fetch metadata for a share if its active? what if the validator's status suddenly changes?

y0sher (changes requested):
check latest comment. we need to make sure we always update all our validators' non liquidated shares

operator/validator/controller.go
```diff
@@ -456,13 +459,18 @@ func (c *controller) StartValidators() {
...
 
+	mySubnets := c.selfSubnets()
+
 	var ownShares []*ssvtypes.SSVShare
-	var allPubKeys = make([][]byte, 0, len(shares))
+	var pubKeysToFetch [][]byte
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
```
y0sher, line 469: wouldn't this do the same as the lines below? just use all the pubkeys from ownshares?

y0sher (approved):
looks good. can you please reply my question

operator/validator/controller.go
nkryuchkov: fixed

operator/validator/controller.go
nkryuchkov: fixed

operator/validator/controller.go
nkryuchkov: @y0sher no, we need `ownShares` to start only own validators and we need `pubKeysToFetch` to fetch metadata

y0sher: yeah I was just trying to say, aren't we collecting it twice? aren't all the pubkeys to fetch already inside ownshares?

operator/validator/controller.go
```diff
@@ -976,6 +995,7 @@ func (c *controller) UpdateValidatorMetaDataLoop() {
 			if share.Liquidated {
 				return true
 			}
+
```
y0sher, line 998: we need to add here the same filter we do in the `startValidators` func. getting all the validators that in my subnets.

y0sher: we need to call the self subnets inside the loop to make sure we get the updated list of subnets for each iteration (to make sure its updated)

y0sher: please also check how much time "selfSubnets" function takes to make sure we are not slowing down this loop too much.

operator/validator/controller.go
y0sher: I see it now. ownShares is not enough because there might be other committees in our subnets even though we are not part of them. resolving this.

operator/validator/controller.go
```diff
@@ -456,13 +459,18 @@ func (c *controller) StartValidators() {
...
 	var ownShares []*ssvtypes.SSVShare
-	var allPubKeys = make([][]byte, 0, len(shares))
+	var pubKeysToFetch [][]byte
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		allPubKeys = append(allPubKeys, share.ValidatorPubKey[:])
+		subnet := networkcommons.CommitteeSubnet(share.CommitteeID())
+		if uint64(len(mySubnets)) >= subnet && mySubnets[subnet] != 0 {
```
y0sher, line 471: I don't think the len check is necessary here because you initialize the mySubnets slice in the same size each time..

operator/validator/controller.go
nkryuchkov: > please also check how much time "selfSubnets" function takes to make sure we are not slowing down this loop too much.

300-900 µs

y0sher (approved):
gj.
please lets check the resource change for calling selfSubnets for every iteration of the metadata loop before merging this.

operator/validator/controller.go
y0sher: sounds good. please add to the pr description

moshe-blox (approved):
LGTM

olegshmuelov:
It seems there's an issue with the current implementation of this PR. While the logic correctly focuses on listening and getting metadata only for committee validators, it does not account for the exporter node. The exporter node requires metadata for all validators, not just the committee ones.

This is a critical oversight, as it may cause incomplete data in the exporter node's operations. We need to update the logic to ensure that the exporter node retrieves metadata for all validators.

Issue #1745, opened 2024-09-18
fix: share CommitteeDutyGuard across all committees


moshe-blox:
This should be a very small change and would help guarantee only one duty is running for a certain validator across all committees, in case we have a race condition when a validator moves between committees.

PR #1788, opened 2024-10-11
CommitteeDutyGuard: share between committees

iurii-ssv:
Hopefully closes https://github.com/ssvlabs/ssv/issues/1745

I've been looking into `CommitteeDutyGuard` logic and got a question (see below).

Decided to take a swing at related issue while I'm at it :)


protocol/v2/ssv/validator/committee_guard.go
```diff
@@ -34,6 +34,7 @@ func (a *CommitteeDutyGuard) StartDuty(role spectypes.BeaconRole, validator spec
 		return fmt.Errorf("unsupported role %d", role)
 	}
 	runningSlot, exists := duties[validator]
+	// TODO: leave a clarifying comment for this condition
 	if exists && runningSlot >= slot {
 		return fmt.Errorf("duty already running at slot %d", runningSlot)
 	}
```
iurii-ssv, line 40: I've been puzzled by this, there is a [short discussion](https://github.com/ssvlabs/ssv/pull/1743/files#r1763483388) about this code line but it doesn't really answer **why it's OK to replace running duty with a new duty started afresh**. 

So I'd want to add a comment here if you could help me out, my default assumption is - we don't want to replace running duty (because nothing really prevents us from doing it over and over again) and rather prefer to finish the one started at the earliest time. But I'm clearly wrong on this ... perhaps this way we are trying to restart a duty that "got stuck" or something ?

iurii-ssv:
Just a note, for operators who serve many validators `CommitteeDutyGuard.mtx` could be somewhat contentious, 

it's unlikely to be an issue, but if it is - we can optimize this (later on) by having a bunch of mutexs (each serving a subset/bucket of validators).

protocol/v2/ssv/validator/committee_guard.go
moshe-blox: Due to our slashing protection mechanism, which would not let you sign an attestation older than the highest signed one. This means that if you sign on a newer duty, you will no longer be able to sign on the older one, therefore we prematurely stop the older to prevent this conflict and save resources.

There isn't an Ethereum slashing rule that strictly prevents you from signing an older attestation, but rather it's just a limitation of our slashing protection implementation, which can potentially be upgraded in the future.

moshe-blox:
@iurii-ssv operators can serve up to 500 validators at the moment, which comes off at an average of 15 per slot (15 at a time), so it should be insignificant given each call should finish within nanoseconds

of course, if this surprisingly pops up on pprof, then this could change

protocol/v2/ssv/validator/committee_guard.go
iurii-ssv: I see, that helps, I've added a clarification comment, thank you!

I guess my worry here was that we can continuously start new duties never letting the previous ones to finish. But I guess in practice there is always a large time-window between 2 duties for the same validator (at least 1 slot), which means it should never be a problem ?

iurii-ssv: Right, updated the comment there.

y0sher (approved):
Great!

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...","2f70337ba7566a69","dd607a44e1341e6b","5210501625ac3de5","f786bf475b5085aa","18a66ed6e613d9c1","e8943e7741f6843d","276a489bd5a00032","ba3bba59f10bf6b","3c50ce0c8089d871","89ee72f6c610ab84","c92b95a85da2cb11","927ea6aed3f98f20","9338904026a0ce37","9683cfa19dc544a3","4d3fa2b8dfcb5f5b", "f19e9a2b295bcfb3", "b10199b2de6f03b8", "1afc17e358f9ca79","4b58762c0b433442","d293ec1bc61bb707","3e88c3b49d093605","4890ff80c88cc41d","5227ff3a225dd20d","81a60407a3a0ba80","db2ad807eb66254a","d308bd7c553ccdcf","bdaf172971637cbe","6ade9202843071fe","2fe8e14083997744","19c9a5362d1e1d3a","5956f803d239f178","92c55a4548a8b760","9a95524213bccfff","2f51a7338b86c229","e96966a281d74505","3ee479b9cbbc3a1d","82b392ba39c6c594","b9d2404e5c570019","24f528d85fb021f2","fe9609a785305d81","b0934079dcd986cc","a9c520a19b26049","d19a9403fd732d94","74a928f5dcb2fdd9","cbbfdb5e68cdac80","10e39d2ceda91f34","f99a004cf6697875","8fa5e8ebf7d223ec","6c80c145ba705243","fbabbc90d0b4178a","ab0c7f24e551ca6","af38a11cb8682c75","b110cba51df9f8d2","c4ff2ed3d20dc419","9295a5bb10efcec7","ab56ea44a75f898a","ff51ef26ab53ba58","df3771e2589008f9","106e5689655bcfc6","f90e0fb6883bff93","667656095cec39ee","9a5597af260c748a", "a347ee50e92f7334"]
```
iurii-ssv, line 1: Corresponding diff:
```
diff --git a/protocol/v2/ssv/validator/committee.go@NewCommittee b/ssv/committee.go@NewCommittee
index a347ee50e92f7334..2222222
--- a/protocol/v2/ssv/validator/committee.go@NewCommittee
+++ b/ssv/committee.go@NewCommittee
@@ -1,25 +1,13 @@
 func NewCommittee(
-	ctx context.Context,
-	cancel context.CancelFunc,
-	beaconNetwork BeaconNetwork,
-	committeeMember *CommitteeMember,
-	createRunnerFn CommitteeRunnerFunc,
-	shares map[phase0.ValidatorIndex]*Share,
-	dutyGuard *CommitteeDutyGuard,
+	committeeMember CommitteeMember,
+	share map[spec.ValidatorIndex]*Share,
+	createRunnerFn CreateRunnerFn,
 ) *Committee {
-	if shares == nil {
-		shares = make(map[phase0.ValidatorIndex]*Share)
-	}
-	return &Committee{
-		logger:			logger,
-		BeaconNetwork:		beaconNetwork,
-		ctx:			ctx,
-		cancel:			cancel,
-		Queues:			make(map[phase0.Slot]queueContainer),
-		Runners:		make(map[phase0.Slot]*runner.CommitteeRunner),
-		Shares:			shares,
+	c := &Committee{
+		Runners:		make(map[spec.Slot]*CommitteeRunner),
 		CommitteeMember:	committeeMember,
 		CreateRunnerFn:		createRunnerFn,
-		dutyGuard:		dutyGuard,
+		Share:			share,
 	}
+	return c
 }
\ No newline at end of file
```

iurii-ssv:
CI [pipelines](https://github.com/ssvlabs/ssv/commits/ci/committee-duty-guard-share-between-committees) passing.

moshe-blox (approved):

moshe-blox:
LGTM 🔥

nkryuchkov (approved):

PR #1791, opened 2024-10-14
fix (main): set discovery interval to 1ms

y0sher:
### Description

Since `RandomNodes()` func has internal way to wait for table to fill, it doesn't really need the sleep interval to not create a busy loop when its empty.

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	defaultDiscoveryInterval = time.Millisecond * 100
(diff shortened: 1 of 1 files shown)


moshe-blox (approved):

nkryuchkov (commented):
If we don't need to wait, can we remove the interval and the ticker?

y0sher:
> If we don't need to wait, can we remove the interval and the ticker?
@nkryuchkov 
We can, but we don't want to do this for upcoming version since its last minute change. we'll do it next versions.

Issue #1786, opened 2024-10-10
Correctly implement handleOperatorRemoved


MatusKysel:
We need to re-implement `handleOperatorRemoved` as it currently doing nothing

```go
	// TODO: In original handler we didn't delete operator data, so this behavior was preserved. However we likely need to.
	// TODO: Delete operator from all the shares.
	//	var shares []Share
	//	for _, s := range nodeStorage.Shares().List() {
	//		// if operator in committee, delete him from it:
	//		//     shares = append(shares, s)
	//	}
	//	nodeStorage.Shares().Save(shares)
	// err = eh.nodeStorage.DeleteOperatorData(txn, od.ID)
	// if err != nil {
	// 	return err
	// }
```

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

nkryuchkov:
Seems to have been fixed by https://github.com/ssvlabs/ssv/pull/2084. @kchojn please reopen if not

PR #1796, opened 2024-10-15
eventhandler: allow creating removed operator

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/1786

### DIFF
--- a/eth/eventhandler/handlers.go
+++ b/eth/eventhandler/handlers.go
@@ -69,7 +69,12 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
 	if err != nil {
 		return fmt.Errorf("could not get operator data by public key: %w", err)
 	}
-	if pubkeyExists {
+
+	eh.removedOperatorsMu.Lock()
+	_, removed := eh.removedOperators[string(od.PublicKey)]
+	eh.removedOperatorsMu.Unlock()
+
+	if pubkeyExists && !removed {
 		logger.Warn("malformed event: operator public key already exists",
 			fields.OperatorPubKey(operatorData.PublicKey))
 		return &MalformedEventError{Err: ErrOperatorPubkeyAlreadyExists}
@@ -79,7 +84,8 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
 	if err != nil {
 		return fmt.Errorf("save operator data: %w", err)
 	}
-	if exists {
+
+	if exists && !removed {
 		logger.Debug("operator data already exists")
 		return nil
 	}
@@ -88,6 +94,10 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
 		eh.operatorDataStore.SetOperatorData(od)
 		logger = logger.With(zap.Bool("own_operator", true))
 	}
+
+	eh.removedOperatorsMu.Lock()
+	delete(eh.removedOperators, string(od.PublicKey))
+	eh.removedOperatorsMu.Unlock()
 
 	eh.metrics.OperatorPublicKey(od.ID, od.PublicKey)
 	logger.Debug("processed event")
@@ -117,7 +127,12 @@ func (eh *EventHandler) handleOperatorRemoved(txn basedb.Txn, event *contract.Co
 		fields.Owner(od.OwnerAddress),
 	)
 
-	// This function is currently no-op and it will do nothing. Operator removed event is not used in the current implementation.
+	// This function is currently just maintaining the set of removed validators, and it will do nothing beyond that.
+	// The set of removed validators allows registering a validator with a pubkey that was once removed.
+	// Operator removed event is not used for anything else in the current implementation.
+	eh.removedOperatorsMu.Lock()
+	eh.removedOperators[string(od.PublicKey)] = struct{}{}
+	eh.removedOperatorsMu.Unlock()
 
 	logger.Debug("processed event")
 	return nil
--- a/eth/eventhandler/event_handler.go
+++ b/eth/eventhandler/event_handler.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"sync"
 	"time"
 
 	"github.com/ssvlabs/ssv/ekm"
@@ -57,14 +58,16 @@ type taskExecutor interface {
 }
 
 type EventHandler struct {
-	nodeStorage       nodestorage.Storage
-	taskExecutor      taskExecutor
-	eventParser       eventparser.Parser
-	networkConfig     networkconfig.NetworkConfig
-	operatorDataStore operatordatastore.OperatorDataStore
-	operatorDecrypter keys.OperatorDecrypter
-	keyManager        ekm.KeyManager
-	beacon            beaconprotocol.BeaconNode
+	nodeStorage        nodestorage.Storage
+	taskExecutor       taskExecutor
+	eventParser        eventparser.Parser
+	networkConfig      networkconfig.NetworkConfig
+	operatorDataStore  operatordatastore.OperatorDataStore
+	operatorDecrypter  keys.OperatorDecrypter
+	keyManager         ekm.KeyManager
+	beacon             beaconprotocol.BeaconNode
+	removedOperators   map[string]struct{}
+	removedOperatorsMu sync.Mutex
 
 	fullNode bool
 	logger   *zap.Logger
@@ -91,6 +94,7 @@ func New(
 		operatorDecrypter: operatorDecrypter,
 		keyManager:        keyManager,
 		beacon:            beacon,
+		removedOperators:  make(map[string]struct{}),
 		logger:            zap.NewNop(),
 		metrics:           nopMetrics{},
 	}


eth/eventhandler/event_handler_test.go
```diff
@@ -1346,6 +1233,118 @@ func TestHandleBlockEventsStream(t *testing.T) {
...
+			require.NoError(t, err)
+			blockNum++
+
+			// Check if the operator wasn't removed successfully
+			operators, err = eh.nodeStorage.ListOperators(nil, 0, 0)
+			require.NoError(t, err)
+			require.Equal(t, len(ops), len(operators))
+		})
+
+		// Receive event, unmarshall, parse, check parse event is not nil or with error, operator id is correct
```
iurii-ssv, line 1273: This kind of description is probably better split in multiple lines and put those right where these things happen, while test description should probably describe more high-level intent for this test

iurii-ssv (approved):
LGTM, just minor suggestions

eth/eventhandler/event_handler_test.go
nkryuchkov: I'd even delete this comment. But I just moved this subtest closer to the end of the file and uncommented the related check. I didn't want to touch any irrelevant code. Also, I think your comment applies to other places in this file, so I think I'd prefer this to be out of the scope of this PR even though I'm touching this one

eth/eventhandler/handlers.go
```diff
@@ -79,11 +80,46 @@ func (eh *EventHandler) handleOperatorAdded(txn basedb.Txn, event *contract.Cont
...
 	if exists {
 		logger.Debug("operator data already exists")
 		return nil
 	}
 
+	var modifiedShares []*ssvtypes.SSVShare
+	for _, share := range eh.nodeStorage.Shares().List(txn, registrystorage.ByOperatorID(event.OperatorId)) {
+		if !share.Liquidated {
+			continue
+		}
```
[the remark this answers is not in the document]
nkryuchkov: We also have the ClusterLiquidated event which may make a share `Liquidated`, so it can be both true and false

eth/eventhandler/event_handler_test.go
iurii-ssv: True, you are right.

eth/eventhandler/handlers.go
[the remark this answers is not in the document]
iurii-ssv: Right, perhaps leaving a comment noting that wouldn't hurt :)

But then I wonder what happens to **validator** in the following scenario:
- 4 operators are running validator
- at some point validator(shares) gets liquidated because he couldn't pay his bills
- then those 4 operators themselves get de-registered from SSV network (doesn't need to happen at the same time, could be many days of delay between each de-registering)
- then those 4 operators re-register again - also reviving this validator (that shouldn't  ever get revived in that manner) ?

well, actually the whole validator won't get revived ^ if I'm reading the code correctly, only the share belonging to the operator referenced in the event will be revived (unliquidated) ... which also means (again, if I got that right) this whole algorithm will only work as expected to handle just 1 operator de-regestering -> re-regestering (won't work in scenario where 2 operators out of N de-register, and only then after that try to re-register again)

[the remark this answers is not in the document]
nkryuchkov: @iurii-ssv I agree about the comment. But I'm not sure I understand your concern about the re-registering logic. We process events one by one, so if 2 operators exist and 2 are trying to re-register, then after processing the first one there would be 3 active operators, so registering the second operator would lead to 4 active operators and share reactivation. If I didn't catch your thought correctly, please follow-up

eth/eventhandler/event_handler_test.go
```diff
@@ -1346,6 +1233,118 @@ func TestHandleBlockEventsStream(t *testing.T) {
...
+			// Check storage for the new operator
+			operators, err = eh.nodeStorage.ListOperators(nil, 0, 0)
+			require.NoError(t, err)
+			require.Equal(t, len(ops)+1, len(operators))
+
+			// Now start the OperatorRemoved event handling
+			// Call the contract method
+			_, err = boundContract.SimcontractTransactor.RemoveOperator(auth, 4)
+			require.NoError(t, err)
+			sim.Commit()
```
iurii-ssv, line 1320: I think we are removing the wrong operator here (wrong ID - not the one this test added) - this is because `createOperators` starts numerating operators with `1` (not `0`) - not a big deal this being the last test ... but still better adjust it to use `op.id` instead

iurii-ssv: Actually, don't bother changing it just yet - I'm preparing a change you'll probably wanna use (and will include this too).

eth/eventhandler/handlers.go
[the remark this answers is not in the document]
iurii-ssv: Well yeah, this operator/share management logic is a bit more involved than I expected ... so I decided to write a test or two to verify whether I understand it correctly ... and got a bit stuck.

There seems to be no "combined" tests that would define interaction between operator and his shares - but rather these aspects are tested in isolation. Here is my sketch for how they could look like (essentially turning my comment from above into code, and adding some more) - [a branch](https://github.com/ssvlabs/ssv/tree/testing/allow-creating-removed-operator) with 1 commit on top of what you have in this PR's branch - see [diff](https://github.com/ssvlabs/ssv/compare/allow-creating-removed-operator...testing/allow-creating-removed-operator). Note also, there are 2 minor fixes in there too, namely (so I recommend to "take as is and adjust it as you see fit"):
- `sharePubKey := validatorData2.operatorsShares[0].sec.GetPublicKey().Serialize()` seems to have been a typo
- and what I mentioned in https://github.com/ssvlabs/ssv/pull/1796#discussion_r1825720014
- plus some minor adjustments to `test ClusterLiquidated + ClusterReactivated events handling`

why did I get stuck ... well, first of all I don't really know what behavior we want to have (to define proper expectations in tests), I did it best I could and encountered that we [increment `operatorId` (in smart contract)](https://github.com/ssvlabs/ssv/blob/stage/eth/simulator/simcontract/simcontract.sol#L53) even for operators who are re-registering (using same pubkey - they end up with different `operatorId`)
- as the result ^ we can't utilize `operatorId` for certain things, for example [when checking whether share belongs to operator or not](https://github.com/ssvlabs/ssv/blob/7e058e953dbf7299ceaa595564819d95fffa7d6a/eth/eventhandler/handlers.go#L427-L440) - meaning, re-registered operator can't find his "old" share
- hence the tests I've added cannot digest that, but I don't think they are the problem; rather it seems we should either not increment `operatorId` in smart contract or not rely/use it like we do
- also, less relevant, but I'm not sure why we only have a part of solidity code but not the rest (the rest I guess is pre-compiled and used in tests ?), and hence can't quite understand how these events compare to what we have in main smart contract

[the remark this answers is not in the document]
nkryuchkov: @iurii-ssv I think that contract is just for simulator, the actual contract is likely this one https://github.com/ssvlabs/ssv-network/blob/583b7b7cb1c1abc5d4c3b13bafca59bf315113b6/contracts/modules/SSVOperators.sol#L26

Can you please create a PR for your changes?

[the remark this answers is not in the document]
iurii-ssv: > Can you please create a PR for your changes?

Yup, why not - https://github.com/ssvlabs/ssv/pull/1842

eth/eventhandler/event_handler_test.go
iurii-ssv: Addressed in https://github.com/ssvlabs/ssv/pull/1842

eth/eventhandler/handlers.go
```diff
@@ -117,7 +157,21 @@ func (eh *EventHandler) handleOperatorRemoved(txn basedb.Txn, event *contract.Co
...
 
-	// This function is currently no-op and it will do nothing. Operator removed event is not used in the current implementation.
+	if err := eh.nodeStorage.DeleteOperatorData(txn, od.ID); err != nil {
+		return fmt.Errorf("delete operator data: %w", err)
+	}
+
+	var modifiedShares []*ssvtypes.SSVShare
+	for _, share := range eh.nodeStorage.Shares().List(txn, registrystorage.ByOperatorID(event.OperatorId)) {
+		share.Liquidated = true
+		modifiedShares = append(modifiedShares, share)
```
y0sher, line 167: this will liquidate the validator even if only one operator was removed.

y0sher (approved):
After internal discussions, this is the current conclusion - 
- we don't want to stop the whole committee ( current pr with liquidated ) because of a single removed operator. 
- we do want to encourage validators to move to full committees without removed operators.
- We don't want to accept messages from removed operators.

Hence the way we want to approach this PR is to check to remove the operator data and check the operator data store in msg validation then ignore their messages.

y0sher (changes requested):
After internal discussions, this is the current conclusion - 
- we don't want to stop the whole committee ( current pr with liquidated ) because of a single removed operator. 
- we do want to encourage validators to move to full committees without removed operators.
- We don't want to accept messages from removed operators.

Hence the way we want to approach this PR is to check to remove the operator data and check the operator data store in msg validation then ignore their messages.

nkryuchkov:
It might be related to changes in https://github.com/ssvlabs/ssv/pull/2084

y0sher:
@nkryuchkov @kchojn 
could you guys please advise if and why this PR is till needed after #2084 
@nkryuchkov please add to the PR description.

Issue #1472, opened 2024-07-16
Use cacheing to ensure consistent block root.


y0sher:
### Description 

We have witnessed cases where we have different block roots we get from the beacon node for the same slot (because of different request times). by caching the results we can ensure higher consistency and efficiency.

See https://github.com/ssvlabs/ssv/pull/1367 for example. needs more work as mentioned in the PR.

iurii-ssv:
@y0sher just to clarify if I understand correctly:

once I implement this feature it would be pretty much impossible to **definitively** test it because we can't prove this type of failure never happens again, the best we can do is:
- to keep monitoring attestation failures and checking there is no root inconsistency anymore
- observe attestation failures rate, it should be same or lower after this feature is deployed

PR #1798, opened 2024-10-16
beacon-client: attestation data request caching

iurii-ssv:
Closes https://github.com/ssvlabs/ssv/issues/1472

Technically, caching solution implemented here is general purpose (in case we need similar caching functionality somewhere else), it could be extracted into separate "util-like" package (for now I guess there is no need for it).

Note, I've also seen/reviewed https://github.com/ssvlabs/ssv/pull/1367 - the approach I've taken is pretty similar but tackles some concurrency nuances in slightly different way.

Note, PR is ready for review (at least to see if it's conceptually fine), but I'll need to:
- [x] add unit tests
- [x] test this on stage


beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
 }
 
-func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.CommitteeIndex) (*phase0.AttestationData, spec.DataVersion, error) {
+// GetAttestationData returns attestation data for a given slot (which is same for all 64 committeeIndex
+// values that identify Ethereum committees chosen to attest on this slot).
+// Note, committeeIndex is an optional parameter that will be used to set AttestationData.Index
+// in the resulting data returned from this function.
+// Note, result returned is meant to be read-only, it's not safe to modify it (because it will be
+// accessed by multiple concurrent readers).
+func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.CommitteeIndex) (
```
iurii-ssv, line 36: Regarding `committeeIndex`,

as far as I can tell (looking at the usage in code), we don't really need `committeeIndex` to even exist as parameter for this func, removing it completely though might be not that simple because `spec` repo is also some definitions with it.

protocol/v2/ssv/runner/committee.go
```diff
@@ -667,11 +666,12 @@ func (cr *CommitteeRunner) expectedPostConsensusRootsAndBeaconObjects(logger *za
...
 	start := time.Now()
 	slot := duty.DutySlot()
+	// We set committeeIndex to 0 for simplicity, there is no need to specify it exactly because
+	// all 64 Ethereum committees assigned to this slot will get the same data to attest for.
 	attData, _, err := cr.GetBeaconNode().GetAttestationData(slot, 0)
 	if err != nil {
 		return errors.Wrap(err, "failed to get attestation data")
 	}
-	//TODO committeeIndex is 0, is this correct?
 	logger = logger.With(
```
iurii-ssv, line 675: See also discussions above and comments in the code answering this TODO.

protocol/v2/blockchain/beacon/duty_data.go
```diff
@@ -1,77 +0,0 @@
-package beacon
```
iurii-ssv, line 1: This file seems to be unused, hence deleting it.

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+func (gc *GoClient) pruneStaleAttestationDataRunner() {
+	pruneStaleAttestationData := func() {
+		// slotRetainCnt defines how many recent slots we want to preserve in attestation data cache.
+		slotRetainCnt := 5 * gc.network.SlotsPerEpoch()
+		gc.attestationDataCache.Range(func(slot phase0.Slot, data *phase0.AttestationData) bool {
+			if uint64(slot) < (gc.recentAttestationSlot.Load() - slotRetainCnt) {
+				gc.attestationDataCache.Delete(slot)
+			}
+			return true
+		})
```
iurii-ssv, line 118: 5 epochs is an arbitrary number from my head, perhaps 2 would be enough, but we can go with 10 also and that won't make much difference.

beacon/goclient/goclient.go
```diff
@@ -141,19 +142,40 @@ var _ NodeClientProvider = (*GoClient)(nil)
...
 	registrationLastSlot phase0.Slot
 	registrationCache    map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration
-	commonTimeout        time.Duration
-	longTimeout          time.Duration
+
+	// attestationReqMuPool helps us prevent the sending of multiple attestation data requests
+	// for the same slot number (to avoid doing unnecessary work).
+	// It's a pool of mutexes (not a single mutex) to allow for some parallelism when requests
+	// targeting different slots are made.
+	attestationReqMuPool [32]sync.Mutex
```
iurii-ssv, line 163: The size of this `attestationReqMuPool` is pretty arbitrary (the more requests we send the higher this value should be), using just 1 mutex for all the requests is probably not enough, 32 seems reasonable here

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+			return
+		}
+
+		// Assign committeeIndex passed to GetAttestationData call, the rest of attestation data stays
+		// unchanged.
+		// Note, we cannot return result object directly here modifying its Index value because it
+		// would be unsynchronised concurrent write (since it's cached for other concurrent readers
+		// to access). Hence, we return shallow copy here. We don't need to return deep copy because
+		// the callers of GetAttestationData will only use this data to read it (they won't update it).
+		result = &phase0.AttestationData{
```
nkryuchkov, line 54: I think having a function/method receiving `*phase0.AttestationData` and `committeeIndex` would be easier to read.

It would be used something like this:
```go
return insertCommitteeIndex(cachedResult), nil
```
```go
return insertCommitteeIndex(resp.Data), nil
```

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+		return cachedResult, spec.DataVersionPhase0, nil
+	}
+
+	// Have to make beacon node request and cache the result.
 	attDataReqStart := time.Now()
-	resp, err := gc.client.AttestationData(gc.ctx, &api.AttestationDataOpts{
-		Slot:           slot,
-		CommitteeIndex: committeeIndex,
-	})
+	result, err = func() (*phase0.AttestationData, error) {
```
nkryuchkov, line 71: Can we extract this into another method?

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+		if resp == nil {
+			return nil, fmt.Errorf("attestation data response is nil")
+		}
+
+		gc.attestationDataCache.Set(slot, resp.Data)
+		gc.recentAttestationSlot.Store(uint64(slot))
+
+		return resp.Data, nil
+	}()
+	metricsAttesterDataRequest.Observe(time.Since(attDataReqStart).Seconds())
```
nkryuchkov, line 99: What do you think about moving this inside the function that gets attestation data?

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+
+	// Have to make beacon node request and cache the result.
 	attDataReqStart := time.Now()
-	resp, err := gc.client.AttestationData(gc.ctx, &api.AttestationDataOpts{
-		Slot:           slot,
-		CommitteeIndex: committeeIndex,
-	})
+	result, err = func() (*phase0.AttestationData, error) {
+		// Requests with the same slot number must lock on the same mutex.
+		reqMu := &gc.attestationReqMuPool[int64(slot)%int64(len(gc.attestationReqMuPool))]
```
nkryuchkov, line 73: The slot will unlikely ever reach `int64` max value but can we use `uint64` to have the same size as `phase0.Slot`?

beacon/goclient/goclient.go
```diff
@@ -141,19 +142,40 @@ var _ NodeClientProvider = (*GoClient)(nil)
...
+	// attestationDataCache stores attestation data from Beacon node for a bunch of recently made
+	// requests (by slot number). It allows for requesting attestation data once per slot from
+	// Beacon node as well as always having/observing the same consistent data in any given slot
+	// (eg, Beacon node might return different merkle root for the same slot for different requests,
+	// and this can lead to undesirable effects for GoClient users).
+	// attestationDataCache is used concurrently, hence thread-safe map.
+	// Note, we cache attestation data by slot (and not slot+committee_index) because it's the same
+	// data across all 64 Ethereum committees assigned for each slot.
+	attestationDataCache *hashmap.Map[phase0.Slot, *phase0.AttestationData]
```
nkryuchkov, line 172: We usually use the `ttlcache` package for caches that require pruning old values, perhaps we could consider using it here as well to avoid having `pruneStaleAttestationDataRunner`

beacon/goclient/attest.go
nkryuchkov: Yes, 2 should be enough, we usually use 2 epochs for caches like this

beacon/goclient/attest.go
```diff
@@ -27,22 +27,106 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+		gc.attestationDataCache.Range(func(slot phase0.Slot, data *phase0.AttestationData) bool {
+			if uint64(slot) < (gc.recentAttestationSlot.Load() - slotRetainCnt) {
+				gc.attestationDataCache.Delete(slot)
+			}
+			return true
+		})
+	}
 
-	return resp.Data, spec.DataVersionPhase0, nil
+	ticker := time.NewTicker(10 * time.Minute)
```
nkryuchkov, line 121: Can we use 2 epoch duration here? IMO it would look nicer to clean up everything older than 2 epochs each 2 epochs

beacon/goclient/goclient.go
nkryuchkov: 32 looks great, but theoretically, we might want to run a custom eth network in kurtosis in the future for tests/simulations, so it would be great to avoid hardcoding to current network constants for the new code. What do think about using a slice and initializing its length to `SlotsPerEpoch()`?

beacon/goclient/attest.go
nkryuchkov: @iurii-ssv 

> as far as I can tell (looking at the usage in code), we don't really need **committeeIndex** to even exist as parameter for this func, removing it completely though might be not that simple because `spec` repo is also some definitions with it.

`(*AttesterRunner).executeDuty` and `(*GoClient).SubmitAggregateSelectionProof` (from `(*AggregatorRunner).ProcessPreConsensus`) seem to call `GetAttestationData` with a real committee index, which is taken from duty, so we cannot remove it right now. But yeah, we don't always need it, so probably the optional pattern would suit better here, however, we'd need to change spec for this, which is not simple, so I think passing 0 where we don't need it and leaving a comment is fine.

nkryuchkov (commented):
LGTM, but I'd suggest minor improvements

beacon/goclient/attest.go
iurii-ssv: > (*AttesterRunner).executeDuty and (*GoClient).SubmitAggregateSelectionProof (from (*AggregatorRunner).ProcessPreConsensus) seem to call GetAttestationData with a real committee index, which is taken from duty, so we cannot remove it right now.

Yeah, I think it doesn't `committeeIndex` part or the result returned, only `attData.HashTreeRoot()`, 

but regardless, removing `committeeIndex` param is probably not worth it.

beacon/goclient/attest.go
iurii-ssv: We could, but since `GetAttestationData` is quite short and simple (especially due to other improvements we've now applied) it would probably be harder to read if we do it,

I guess there is some balance to strive for when it comes to:
- having short functions
- having too many of them (having to jump around a lot when reading code)

so here I think it's a good example of "keeping it all in one place" being easier to read.

beacon/goclient/attest.go
iurii-ssv: Yes, it's easy to do, changed.

beacon/goclient/attest.go
iurii-ssv: Updated.

beacon/goclient/goclient.go
iurii-ssv: Changed.

beacon/goclient/attest.go
iurii-ssv: Changed to 2.

beacon/goclient/goclient.go
iurii-ssv: Done, unit-tests don't complain about it :)

beacon/goclient/attest.go
iurii-ssv: Moved like you suggested.

beacon/goclient/attest.go
nkryuchkov: @iurii-ssv It's used in `HashTreeRoot`, so changing it would change the hashing algorithm and make the branch incompatible with other branches. There may be operators running different versions

But yeah, if not this, we could get rid of it. Maybe we'll do it sometime in the future

beacon/goclient/attest.go
nkryuchkov: okay, now this method is smaller, so I don't mind keeping this as is

beacon/goclient/goclient.go
nkryuchkov: Yeah, I agree, let's make them uniform, and if we start seeing `ttlcache` on pprof, then we'll consider optimizing them all at once. I think almost all cases of its usage are similar: Cache something for some time like 2 epochs

y0sher (approved):
Code looks good! great job!

@moshe-blox wdyt about this? end of the day we might be missing if there's newer data.. but is it worth it to request every time if we already waited the time appropriate to have correct data?

beacon/goclient/attest.go
```diff
@@ -27,22 +28,73 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
+			return nil, fmt.Errorf("attestation data response is nil")
+		}
+
+		gc.attestationDataCache.Set(slot, resp.Data, ttlcache.DefaultTTL)
+
+		return resp.Data, nil
+	}()
+	if err != nil {
+		return nil, DataVersionNil, err
 	}
```
moshe-blox, line 80: this should work and seems to address the various edgecases 👌 

with that said, i think this could be massively simplified down to just one struct field and one function call, using https://pkg.go.dev/golang.org/x/sync/singleflight

while it's not ideal that singleflight isn't using generics, it should still be much easier to reason about the code  than the advanced implementation we have currently (imo)

iurii-ssv: I was looking for something like this ( turns out not hard enough :) ), thanks for bringing it up!

Although I was trying to find a package that handles both aspects: **inflight** + **caching**. Still, https://pkg.go.dev/tailscale.com/util/singleflight does simplify this code a bit (and unit tests don't complain) - see https://github.com/ssvlabs/ssv/pull/1798/commits/fcaa83d2825c1cc2fc7bd8d196d82b65e705b249 for how it's different compared to the code I had before.

From performance perspective, I don't think it matters much for this particular case, but if we were to push it to the limits - I think each of these 2 approaches will have it's preferred type of workload while doing worse in some circumstances.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ssvlabs/ssv
 
-go 1.22
+go 1.22.0
+
+toolchain go1.23.1
```
iurii-ssv, line 5: This change is the result of running `go mod tidy` after adding `tailscale.com/util/singleflight` (note also, I'm not using the latest version of this package because it will require to bump the `go 1.22` line up to `go 1.23.1`),

I'm new to this `toolchain` concept, from what I've gathered it's used as "recommended" version (it will be downloaded and used in case you are trying to build with lower version).

iurii-ssv:
Tested on stage, seems to be working fine.

nkryuchkov (approved):
LGTM

nkryuchkov:
CI passes for the last commit: https://github.com/ssvlabs/ssv/actions?query=branch%3Atest%2Fbeacon-client-attestations-cache

go.mod
iurii-ssv: Removed `toolchain` directive, see https://github.com/ssvlabs/ssv/pull/1840#discussion_r1834748874 for details.

beacon/goclient/attest.go
moshe-blox: @iurii-ssv

+1

the readability gain is worth the performance cost here IMO

(don't think performance cost is noticeable below many thousands of calls per second)

moshe-blox (approved):
LGTM 🔥 

i've made some changes, we should test this again before merge

iurii-ssv:
This PR have been running on **stage** for several hours now without obvious issues (as far as I can tell).

Issue #1643, opened 2024-08-21
Re-enable Go RSA after Alan fork


olegshmuelov:
the finding from https://github.com/ssvlabs/ssv/issues/1611

iurii-ssv:
> Previously we used a `cgo` library called `jemalloc` which reduced the mem usage of our database a lot. At some point we introduced cryptograpgy cgo lib for RSA (openssl) since it was much more efficient than the golang stdlib RSA lib.
> 
> We want to consider -
> 
> - Switching back to golang std RSA since it might be improved in newer golang versions (we updated to go1.22 but we might update to go1.23 if there's a significant improvement). this will allow us to bring back `jemalloc` and improve database memory even more.
> 
> - make `openssl` and `jemalloc` work together as we know previously they were clashing and not usable together. 

See also: https://github.com/ssvlabs/ssv/pull/1358

iurii-ssv:
I've [added benchmarks](https://github.com/ssvlabs/ssv/pull/1799) to compare RSA implementations (`golang 1.22/1.23` vs `openssl`).

To sum up, it looks like `openssl` is about ~2 times faster (and producing about ~2 times less memory allocations), there is no significant difference between `golang 1.22.*` - `golang 1.23.*` (just tiny improvements from what I've gathered).

Here is the full data I got running:
<details><summary>on Mac ARM laptop (native golang1.23.1)</summary>
<pre>
goos: darwin
goarch: arm64
pkg: github.com/ssvlabs/ssv/operator/keys
BenchmarkSignRSA
BenchmarkSignRSA-12       	    1370	    869123 ns/op	     896 B/op	       5 allocs/op
BenchmarkSignRSA-12       	    1383	    932841 ns/op	     896 B/op	       5 allocs/op
BenchmarkSignRSA-12       	    1374	    875742 ns/op	     896 B/op	       5 allocs/op
BenchmarkEncryptRSA
BenchmarkEncryptRSA-12    	   48724	     24098 ns/op	    1104 B/op	       7 allocs/op
BenchmarkEncryptRSA-12    	   50269	     24009 ns/op	    1104 B/op	       7 allocs/op
BenchmarkEncryptRSA-12    	   50450	     23981 ns/op	    1104 B/op	       7 allocs/op
BenchmarkVerifyRSA
BenchmarkVerifyRSA-12     	   50811	     23724 ns/op	    1104 B/op	       7 allocs/op
[...]

PR #1799, opened 2024-10-18
rsa: benchmark go vs openssl implementations

iurii-ssv:
This PR adds simple `make benchmark` command to `Makefile` in order to be able to run benchmark for a certain directory in this repo (you need to specify it replacing `TARGET_DIR_PATH` value accordingly).

Need for https://github.com/ssvlabs/ssv/issues/1643

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,6 @@ ifeq ($(COVERAGE),true)
 endif
 UNFORMATTED=$(shell gofmt -l .)
 
-#Lint
 .PHONY: lint-prepare
 lint-prepare:
 	@echo "Preparing Linter"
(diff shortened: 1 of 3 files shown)


iurii-ssv:
I also see similar (not 100% same though) benchmarks comparing `golang` and `openssl` in these files:
```
ssv/protocol/v2/types/signature_benchmark_test.go
ssv/protocol/v2/types/signature_benchmark_linux_test.go
```
these ^ seem to be quite generic, while the ones I've added in this PR focus specifically on 3 funcs we actually use (`keys.SignRSA`, `keys.EncryptRSA`, `keys.VerifyRSA`),

@nkryuchkov (git says you wrote those) do we need these older ones (if so we probably want to move them to `ssv/operator/keys` package to have it all in one place ?) or should I maybe just delete these 2 files ?

nkryuchkov:
> @nkryuchkov (git says you wrote those) do we need these older ones (if so we probably want to move them to `ssv/operator/keys` package to have it all in one place ?) or should I maybe just delete these 2 files ?

@iurii-ssv I don't have a strong opinion on this but would prefer to have them all in a single file as we may change the wrappers' implementation

iurii-ssv:
> @iurii-ssv I don't have a strong opinion on this but would prefer to have them all in a single file as we may change the wrappers' implementation

I've put non-linux benchmarks into single file, but linux-specific ones probably need to be in a separate file (so that they are only compiled on linux only via `//go:build linux`).

iurii-ssv:
CI [pipelines](https://github.com/ssvlabs/ssv/commits/ci/rsa-benchmark-go-vs-openssl-impls) passing.

PR #1800, opened 2024-10-18
openssl: upgrade bindings package

iurii-ssv:
This PR switches OpenSSL package we use for RSA from https://github.com/microsoft/go-crypto-openssl (now deprecated) to https://github.com/golang-fips/openssl,

according to benchmarks I ran, it doesn't bring any performance improvements though.

We can also close this PR for now, and we can get back to it later on. See https://github.com/ssvlabs/ssv/pull/1348 for additional historical context.

Reminder to self, before merging will need to:
- [ ] test/check (by deploying to stage) there aren't any panic/fatal/segmentation/fault/cgo errors related to this change

### DIFF
--- a/operator/keys/rsa_linux.go
+++ b/operator/keys/rsa_linux.go
@@ -7,10 +7,12 @@ import (
 	"crypto/rsa"
 	"crypto/sha256"
 	"math/big"
+	"os"
+	"runtime"
 	"sync"
 
-	"github.com/microsoft/go-crypto-openssl/openssl"
-	"github.com/microsoft/go-crypto-openssl/openssl/bbig/bridge"
+	"github.com/golang-fips/openssl/v2"
+	"github.com/golang-fips/openssl/v2/bbig"
 )
 
 type privateKey struct {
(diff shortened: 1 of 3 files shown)


operator/keys/rsa_linux.go
```diff
@@ -26,30 +28,62 @@ type publicKey struct {
...
 	// TODO: fallback to stdlib when openssl is not available
-	if err := openssl.Init(); err != nil {
+	version := getVersion()
+	if err := openssl.Init(version); err != nil {
 		panic(err)
 	}
 }
 
+// getVersion returns the highest available OpenSSL version.
+func getVersion() string {
```
nkryuchkov, line 39: Can you please add a comment that this code is taken from https://github.com/golang-fips/openssl/blob/v2/openssl_test.go?

iurii-ssv: Yup, the reason is https://github.com/golang-fips/openssl doesn't really provide any examples/docs on how this should be done, so I found this approach to work for me (in Docker),

added a note on this.

y0sher (approved):
this looks good to me we should probably check it again after https://github.com/ssvlabs/ssv/pull/1819/

PR #1803, opened 2024-10-18
network/discovery: rate limit skipped peers log

nkryuchkov:
@zktaiga reported that this log makes up to 40% of stage logs


network/discovery/metrics.go
```diff
@@ -27,6 +27,10 @@ var (
 		Name: "ssv:network:discovery:iterations",
 		Help: "Counts the number of times a node was iterated using the mixed iterator",
 	}, []string{"fork"})
+	metricSkippedPeers = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "ssv:network:discovery:skipped_peers",
+		Help: "Counts skipped peers",
+	}, []string{"enr", "peer_id", "err"})
```
zktaiga, line 33: Having peer ID or an ENR as a label is problematic because of its high cardinality (Prometheus struggles with this). We could use something such as `ssv:network:discovery:skipped_peers_total` for now if a total count would help. If not, we need some other solution.

zktaiga (changes requested):

network/discovery/metrics.go
nkryuchkov: @zktaiga reverted the change and added the log rate limit

y0sher (approved):
this should reduce logs but also not miss unique logs. good job

zktaiga (approved):

moshe-blox:
this looks good, but the map will never clean and can grow infinitely

instead of implementing cleanup, i suggest just using https://github.com/jellydator/ttlcache instead (we already import it)

y0sher (changes requested):
clean the cache, maybe use some existing ttl cache like @moshe-blox suggested

nkryuchkov:
@y0sher @moshe-blox I heard from @zktaiga we decided not to proceed with this solution, so can I close this?

moshe-blox (approved):
LGTM 🔥

iurii-ssv (commented):
Left a comment about TTL cache cleanup go-routine.

Something else I've noticed, `Bootstrap` func gets called with `tasks.Retry`. Given that `discover` is blocking and only returns when `dvs.ctx` is canceled - hence we probably don't want/can't retry anyway - so maybe we should remove these retries while at it.

iurii-ssv (approved):

anatolie-ssv (approved):

y0sher (commented):
We have decided to not use a cache for logs for now and go on with a simpler solution which is already in place for now :pray:. 

https://github.com/ssvlabs/ssv/commit/c0edeb6a2bd7cf7f58d6a3f7b8c1dd2f27a4af2e#diff-e6f6d3fa5b532909384dd304ee202074aed77d32acb079cdf344a525a8488cf3L137-R158

Issue #1802, opened 2024-10-18
network/records: panic due to insufficient subnets length


nkryuchkov:
```
INFO    P2PNetwork      Verified handshake nodeinfo     {"selfPeer": "16Uiu2HAmMmWFy5nnGguFdoaRZVBMDgvdUzh4MXDtYEjGCiQ8rH7w", "peer_id": "12D3KooWJqhpvtikxfRQSCSJiBXjKfZeUkpXe2x8yinF26Ytc88G", "peer_id": "12D3KooWJqhpvtikxfRQSCSJiBXjKfZeUkpXe2x8yinF26Ytc88G", "metadata": {"NodeVersion":"RSSV v0.1.0","ExecutionNode":"","ConsensusNode":"","Subnets":"1000000000004000010000"}, "networkID": "0x00000502"}
panic: runtime error: index out of range [104] with length 88

goroutine 37011028 [running]:
github.com/ssvlabs/ssv/network/records.SharedSubnets(...)
        /go/src/github.com/ssvlabs/ssv/network/records/subnets.go:112
github.com/ssvlabs/ssv/network/peers/connections.(*connHandler).sharesEnoughSubnets(0xc00d0d1dc0, 0xc00f255580, {0x2283860?, 0xc022483e60?})
        /go/src/github.com/ssvlabs/ssv/network/peers/connections/conn_handler.go:226 +0x534
github.com/ssvlabs/ssv/network/peers/connections.(*connHandler).Handle.func4(0xc00f255580, {0x2289d88, 0xc00d030c00}, {0x2283860, 0xc022483e60})
        /go/src/github.com/ssvlabs/ssv/network/peers/connections/conn_handler.go:138 +0x78e
github.com/ssvlabs/ssv/network/peers/connections.(*connHandler).Handle.func6.1()
        /go/src/github.com/ssvlabs/ssv/network/peers/connections/conn_handler.go:171 +0xc3
created by github.com/ssvlabs/ssv/network/peers/connections.(*connHandler).Handle.func6 in goroutine 37010913
        /go/src/github.com/ssvlabs/ssv/network/peers/connections/conn_handler.go:169 +0x150
make: *** [Makefile:119: start-node] Error 2
```

PR #1804, opened 2024-10-18
network/records: fix panic due to insufficient subnets length

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/1802

### DIFF
--- a/network/records/subnets.go
+++ b/network/records/subnets.go
@@ -105,7 +105,14 @@ func SharedSubnets(a, b []byte, maxLen int) []int {
 	if len(a) == 0 || len(b) == 0 {
 		return shared
 	}
-	for subnet, aval := range a {
+
+	// Fix https://github.com/ssvlabs/ssv/issues/1802
+	limit := len(a)
+	if len(b) < limit {
+		limit = len(b)
+	}
+
+	for subnet, aval := range a[:limit] {
 		if aval == 0 {
 			continue
 		}
--- a/network/records/subnets.go
+++ b/network/records/subnets.go
@@ -105,14 +105,7 @@ func SharedSubnets(a, b []byte, maxLen int) []int {
 	if len(a) == 0 || len(b) == 0 {
 		return shared
 	}
-
-	// Fix https://github.com/ssvlabs/ssv/issues/1802
-	limit := len(a)
-	if len(b) < limit {
-		limit = len(b)
-	}
-
-	for subnet, aval := range a[:limit] {
+	for subnet, aval := range a {
 		if aval == 0 {
 			continue
 		}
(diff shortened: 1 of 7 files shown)


network/records/subnets.go
```diff
@@ -105,7 +105,14 @@ func SharedSubnets(a, b []byte, maxLen int) []int {
...
 	}
-	for subnet, aval := range a {
+
+	// Fix https://github.com/ssvlabs/ssv/issues/1802
+	limit := len(a)
+	if len(b) < limit {
+		limit = len(b)
+	}
+
+	for subnet, aval := range a[:limit] {
```
moshe-blox, line 115: i think we should require that both arrays are of len == 128 (i think we have a constant for 128)

otherwise, what does an array of 64 bytes mean? are the rest of the 64 subnets 0 or 1? i think if its not specified then its an invalid subnets array

moshe-blox (changes requested):
well done 🔥 

left a minor change request

network/records/subnets.go
```diff
@@ -105,7 +105,14 @@ func SharedSubnets(a, b []byte, maxLen int) []int {
 	if len(a) == 0 || len(b) == 0 {
 		return shared
 	}
-	for subnet, aval := range a {
+
+	// Fix https://github.com/ssvlabs/ssv/issues/1802
+	limit := len(a)
+	if len(b) < limit {
+		limit = len(b)
+	}
```
iurii-ssv, line 113: Can just do in 1 line, I think:
```go
limit := min(len(a), len(b))
```

network/records/subnets.go
```diff
@@ -105,7 +105,14 @@ func SharedSubnets(a, b []byte, maxLen int) []int {
 	if len(a) == 0 || len(b) == 0 {
 		return shared
 	}
```
iurii-ssv, line 107: Perhaps a bit clearer would be to `return nil` here, and define `var shared []int` down below (closer to the actual place of usage).

iurii-ssv (approved):

network/records/subnets.go
nkryuchkov: @iurii-ssv I agree, thanks, but we've decided not to touch this function in this PR and find the root cause of the issue. However, we are going to refactor it in another PR

network/records/subnets.go
nkryuchkov: Yes, that's what I think too, let's refactor this in another PR. I've reverted the initial fix so I'm not going to touch this function in this PR at all

y0sher (approved):

moshe-blox (approved):

network/records/subnets.go
nkryuchkov: @moshe-blox refactored in https://github.com/ssvlabs/ssv/pull/1813, please review it

network/records/subnets.go
nkryuchkov: @iurii-ssv refactored in https://github.com/ssvlabs/ssv/pull/1813, please review it

Issue #1712, opened 2024-09-08
Feat: control-flow, start listening to P2P messages only after contract and metadata sync


y0sher:
### Description

Currently we have a solution in place where we ignore instead of reject messages before we know all the duties, because we can't know if the duty sent is actually exist before we synced on contract events and metadata. This is not ideal.

### Solution

Have a better control and dependency flow of starting the node, which means starting to listen to P2P messages only after we are synced from nodes to know the current state of the network.

#### Challenges 
Some P2P flows are currently infused in validator controller (subscribing to subnets..)

PR #1805, opened 2024-10-19
fix(p2p): start listening to P2P messages only after contract and metadata sync

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/1712


operator/metadata/updater_test.go
```diff
@@ -0,0 +1,138 @@
+package metadata
```
nkryuchkov, line 1: Unit tests for `RetrieveInitialMetadata` and `Stream` are still to be implemented

operator/metadata/updater.go
```diff
@@ -0,0 +1,255 @@
...
+		u.logger.Info("could not find validators")
+		return nil, nil
+	}
+
+	var hasMetadata bool
+	allPubKeys := make([]spectypes.ValidatorPK, 0, len(shares))
+	for _, share := range shares {
+		allPubKeys = append(allPubKeys, share.ValidatorPubKey)
+		if !share.Liquidated && share.HasBeaconMetadata() {
+			hasMetadata = true // TODO: the behavior is preserved; should it be true if all shares have metadata?
```
nkryuchkov, line 86: This TODO and a few other ones still need to be resolved

nkryuchkov: resolved

operator/metadata/updater_test.go
nkryuchkov: implemented

operator/validator/controller.go
```diff
@@ -543,29 +533,6 @@ func (c *controller) StartValidators() {
...
-
-	// Fetch metadata now if there is none. Otherwise, UpdateValidatorsMetadataLoop will handle it.
-	var hasMetadata bool
-	for _, share := range shares {
-		if !share.Liquidated && share.HasBeaconMetadata() {
-			hasMetadata = true
-			break
-		}
-	}
-	if !hasMetadata {
```
nkryuchkov, line 555: I'm not sure why but the original logic used to fetch and update if none of the shares have metadata (the `!share.Liquidated` part of the condition is useless because `shares` contains non-liquidated shares)

This is changed to fetching all public keys if at least one share metadata is missing, however, probably we want to modify this logic

iurii-ssv (commented):
Looks pretty good! Suggested a bunch of improvements (mostly small ones).

operator/metadata/updater.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+//go:generate mockgen -package=metadata -destination=./mocks.go -source=./updater.go
+
+const (
+	defaultUpdateInterval    = 12 * time.Minute
+	defaultStreamInterval    = 2 * time.Second
+	defaultUpdateSendTimeout = 30 * time.Second
+	batchSize                = 512
+	streamChanSize           = 1024
```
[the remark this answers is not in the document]
iurii-ssv: `streamChanSize           = 1024` also seems quite a lot, shouldn't it be more like `16` or something. 

I'm also not sure why this channel is buffered if we implement the logic for dropping stream events ("updates"):
- the receiving side "comes back to life at some point" and starts going through all the old updates (piled up in this channel) while we keep dropping newer updates meanwhile

seems like we can skip older updates altogether and just go for the newest one - non-buffered channel should probably do it, while also taking care of a sort-of push-back by blocking (preventing) the sender from going for the next batch until we are ready to take care of the current one (or time out)

operator/metadata/fetcher.go
```diff
@@ -0,0 +1,70 @@
...
+	start := time.Now()
+	// TODO: Refactor beacon.BeaconNode to support passing context.
+	validatorsIndexMap, err := mf.beaconNode.GetValidatorData(blsPubKeys)
+	if err != nil {
+		mf.logger.Error("failed to fetch initial validators metadata",
+			zap.Int("shares", len(pubKeys)),
+			fields.Took(time.Since(start)),
+			zap.Error(err),
+		)
+		return nil, fmt.Errorf("get validator data from beacon node: %w", err)
```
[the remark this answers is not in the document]
nkryuchkov: Yeah, I agree, I wanted to remove the log above but I wanted to keep logging how long it took, so I'll think about this once again

operator/metadata/fetcher.go
```diff
@@ -0,0 +1,70 @@
...
+			Balance:         v.Balance,
+			Status:          v.Status,
+			Index:           v.Index,
+			ActivationEpoch: v.Validator.ActivationEpoch,
+		}
+		results[spectypes.ValidatorPK(v.Validator.PublicKey)] = meta
+	}
+
+	mf.logger.Debug("⏱️ fetched validators metadata",
+		zap.Duration("elapsed", time.Since(start)),
```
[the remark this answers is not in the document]
nkryuchkov: It's copied from the existing code, but yeah, perhaps we can change it while we touch it

operator/metadata/updater.go
```diff
@@ -0,0 +1,260 @@
...
+
+	u.logger.Debug("🆕 updated validators metadata in storage",
+		fields.Took(time.Since(updateStart)),
+		zap.Int("count", len(metadata)),
+		zap.Int("shares", len(pubKeys)),
+	)
+	return metadata, nil
+}
+
+type Update struct {
```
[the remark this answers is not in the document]
nkryuchkov: Initially, I was thinking about naming it `MetadataUpdate`, but then it starts with the package name. And the package route used to be `validator/metadata`, but then I moved it. I'll think about this once again, maybe I'll move this or rename to something like `ValidatorUpdate`

operator/metadata/updater.go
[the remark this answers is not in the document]
y0sher: I agree with @iurii-ssv here but we need to make sure we are not making the validator controller get stuck

operator/metadata/updater.go
```diff
@@ -0,0 +1,260 @@
...
+func (u *Updater) sharesForUpdate() []*ssvtypes.SSVShare {
+	var existingShares, newShares []*ssvtypes.SSVShare
+	u.shareStorage.Range(nil, func(share *ssvtypes.SSVShare) bool {
+		if share.Liquidated {
+			return true
+		}
+		if share.BeaconMetadata == nil && share.MetadataLastUpdated().IsZero() {
+			newShares = append(newShares, share)
+		} else if time.Since(share.MetadataLastUpdated()) > u.updateInterval {
+			existingShares = append(existingShares, share)
```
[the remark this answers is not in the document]
nkryuchkov: `staleShares` is fine, but I'm not sure I agree with the latter because shares become stale after the update interval, and if we rename it, it won't be readable in other places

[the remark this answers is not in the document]
iurii-ssv: Maybe `shareMetadataStaleAfter` is not the best name (just an example),

what I'm saying is `updateInterval` is too generic of a name that could mean many things - hence would be nice to make it more concrete. What it really represents is "the duration (span of time) after which share metadata is considered stale" - `shareMetadataStillFresh` / `shareMetadataStillFreshDuration`

this one is pretty hard to describe with just name though :)

operator/metadata/updater.go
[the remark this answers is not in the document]
iurii-ssv: This pretty much ties with the other suggestion from below: https://github.com/ssvlabs/ssv/pull/1805#discussion_r1827875337

I'm not sure what the best combination of **package name + struct(s) name(s)** would be (ideally, it would accommodate future developments too - if we want to add a new syncer other than ValidatorMetadataUpdater) - it just didn't feel like the one I saw during that last review was the optimal one

operator/metadata/updater.go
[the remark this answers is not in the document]
nkryuchkov: I made the channel non-buffered, so we no longer need `streamChanSize`. Regarding `batchSize` it's used in a couple methods, and I think it's fine to have package-related consts on a package level

operator/metadata/updater.go
```diff
@@ -0,0 +1,260 @@
...
+
+type Updater struct {
+	logger            *zap.Logger
+	shareStorage      shareStorage
+	beaconNetwork     beacon.BeaconNetwork
+	fetcher           fetcher
+	updateInterval    time.Duration
+	streamInterval    time.Duration
+	updateSendTimeout time.Duration
+}
```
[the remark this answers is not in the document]
nkryuchkov: ValidatorMetadata is not so easy to move because storage uses it, and the ValidatorSyncer/Updater uses storage, so there's a circular dependency. I agree we'd better move it but probably later as it needs further refactoring, I don't want to increase the size of this PR as it's already quite big

network/network.go
y0sher, line 1: this is not really related, next time please do on a separate PR to make reviewing easier.

network/p2p/metrics.go
y0sher, line 1: this is not really related, next time please do on a separate PR to make reviewing easier.

network/p2p/p2p.go
y0sher, line 1: this is not really related, next time please do on a separate PR to make reviewing easier.

operator/validator/controller.go
y0sher: As the comment says, initial fetch was only if we didn't have any metadata. because it relayed on the periodic metadata loop. In this PR we WANT to achieve an initial metadata sync to all relevant validators (before p2p starts)

operator/validator/controller.go
```diff
@@ -979,104 +923,53 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
+func (c *controller) handleMetadataUpdate(ctx context.Context, update metadata.ValidatorUpdate) error {
+	startedValidators := 0
+	if c.operatorDataStore.GetOperatorID() != 0 {
+		startedValidators = c.startValidatorsForMetadata(ctx, update.Validators)
 	}
 
-	// Refresh duties if there are any new active validators.
-	afterUpdate := c.AllActiveIndices(c.beacon.GetBeaconNetwork().EstimatedCurrentEpoch(), false)
-	if c.recentlyStartedValidators > 0 || hasNewValidators(beforeUpdate, afterUpdate) {
+	if startedValidators > 0 || hasNewValidators(update.IndicesBefore, update.IndicesAfter) {
```
y0sher, line 943: why do we need both these conditions? isn't 1 enough?

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,285 @@
...
+					zap.Error(err),
+				)
+				if slept := u.sleep(ctx, u.streamInterval); !slept {
+					return
+				}
+				continue
+			}
+
+			if len(update.Validators) == 0 {
+				continue
```
y0sher, line 159: wouldn't this be a busy loop that keeps requesting non stop when there are 0 validators? I guess we need some backoff..

network/network.go
nkryuchkov: I agree, moved to another PR: https://github.com/ssvlabs/ssv/pull/1968

operator/validator/metadata/syncer.go
nkryuchkov: I added a sleep

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,288 @@
...
+		u.logger.Info("could not find non-liquidated validator shares on initial metadata retrieval")
+		return nil, nil
+	}
+
+	needToSync := false
+	allPubKeys := make([]spectypes.ValidatorPK, 0, len(shares))
+	for _, share := range shares {
+		allPubKeys = append(allPubKeys, share.ValidatorPubKey)
+		if !share.HasBeaconMetadata() {
+			needToSync = true
```
y0sher, line 93: you can break the loop here

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,288 @@
...
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+		return false
+	case <-timer.C:
+		return true
+	}
+}
```
y0sher, line 288: why don't you use wrap ctx with deadline/timeout instead of passing in duration?

y0sher (changes requested):
a few changes requested.
been waiting for this change for a while. good job

operator/validator/metadata/syncer.go
nkryuchkov: @y0sher This is two loops combined in one: to get all pubkeys and to check if need to sync, so I cannot break it because I need all pubkeys. Do you think having two loops and breaking the second one would be more readable?

operator/validator/metadata/syncer.go
nkryuchkov: @y0sher If the context is canceled, I break the loop, and if the timer fires I continue the loop, so if I wrap it to a context, then I won't know if I still need to continue iterating. Without this logic, I'd likely get rid of this method

operator/validator/controller.go
nkryuchkov: @y0sher Well, probably we should just check if we started any validators, this also eliminates some extra logic. Implemented, please review

operator/validator/controller.go
```diff
@@ -986,107 +915,51 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
+
+	c.logger.Debug("started validators after metadata update",
+		fields.Count(startedValidators),
+	)
+
 	return nil
 }
 
+func (c *controller) reportIndicesChange(ctx context.Context, timeout time.Duration) bool {
+	select {
```
y0sher, line 953: you can do
```go
cctx, cancel := context.WithTimeout(timeout)
defer cancel()
```
and then you only need `Done` and get get rid of `time.After`

operator/validator/metadata/syncer.go
y0sher: nah now I understand its ok

operator/validator/controller.go
nkryuchkov: well spotted, thanks. Changed

cli/operator/node.go
```diff
@@ -339,6 +348,10 @@ var StartNodeCmd = &cobra.Command{
 			nodeProber.AddNode("event syncer", eventSyncer)
 		}
 
+		if _, err := validatorSyncer.SyncOnStartup(cmd.Context()); err != nil {
+			logger.Fatal("failed to update metadata on startup", zap.Error(err))
+		}
```
iurii-ssv, line 353: maybe `failed to update` -> `failed to sync`

operator/node.go
```diff
@@ -103,10 +103,10 @@ func New(logger *zap.Logger, opts Options, slotTickerProvider slotticker.Provide
 	return node
 }
 
-// Start starts to stream duties and run IBFT instances
 func (n *Node) Start(logger *zap.Logger) error {
-	logger.Named(logging.NameOperator)
+	logger = logger.Named(logging.NameOperator)
 
+	// Start starts to stream duties and run IBFT instances
```
iurii-ssv, line 109: Shouldn't this comment be right above `Start` func name ?

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+
+			if len(validators) == 0 {
+				if slept := u.sleep(ctx, u.streamInterval); !slept {
+					return
+				}
+				continue
+			}
+
+			timer := time.NewTimer(u.updateSendTimeout)
```
iurii-ssv, line 162: Can maybe use `time.After` to simplify it ?

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,260 @@
...
+
+	for _, share := range shares {
+		share.SetMetadataLastUpdated(time.Now())
+	}
+	return shares
+}
+
+func (u *ValidatorSyncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
```
iurii-ssv, line 252: maybe also can use `time.After` for simplicity

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,260 @@
...
+
+		for {
+			validators, done, err := u.prepareUpdate(ctx)
+			if err != nil {
+				u.logger.Warn("failed to prepare validators metadata",
+					zap.Error(err),
+				)
+				if slept := u.sleep(ctx, u.streamInterval); !slept {
+					return
+				}
```
iurii-ssv, line 151: Wouldn't it be simpler to sleep at the start of every `for` iteration (except for the very 1st one that can be handled by additional `Iteration1st` bool variable) ? so that we don't need to have this duplicated in 3 places

iurii-ssv (commented):
Minor suggestions

operator/validator/metadata/syncer.go
nkryuchkov: Before Go 1.23, the garbage collector doesn't handle unreferenced timers, and we still use 1.22, so `time.After` will leak. I left a comment to replace it when we update the Go version

operator/validator/metadata/syncer.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/1805#discussion_r1911141842

operator/validator/metadata/syncer.go
nkryuchkov: I think we cannot replace the case inside `if done { }`, so we can remove just one duplication. Probably an if and one sleep would be a bit more complex to understand than two sleeps

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,262 @@
...
+//go:generate mockgen -package=metadata -destination=./mocks.go -source=./syncer.go
+
+const (
+	defaultSyncInterval      = 12 * time.Minute
+	defaultStreamInterval    = 2 * time.Second
+	defaultUpdateSendTimeout = 30 * time.Second
+	batchSize                = 512
+)
+
+type ValidatorSyncer struct {
```
moshe-blox, line 27: since the package is already `metadata` this could simply be `Syncer`

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,262 @@
...
+		zap.Int("metadata_count", len(metadata)),
+		zap.Int("shares_cnt", len(pubKeys)),
+	)
+
+	return metadata, nil
+}
+
+type ValidatorUpdate struct {
+	Validators ValidatorMap
+}
```
moshe-blox, line 135: can be removed?

operator/validator/metadata/fetcher.go
moshe-blox, line 1: this can be a `fetch` method on Syncer (or a standalone func)

the struct doesn't seem to help with anything (but adds verbosity)

this can still be in a separate file (fetch.go)

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,262 @@
...
+		allPubKeys = append(allPubKeys, share.ValidatorPubKey)
+		if !share.HasBeaconMetadata() {
+			needToSync = true
+		}
+	}
+
+	if !needToSync {
+		// No need to fetch metadata if all shares have it. It's going to be updated by Stream method afterwards.
+		return nil, nil
+	}
```
moshe-blox, line 99: fixed comment
```suggestion
	// If we already have fetched before: skip syncing and hand it over to Stream.
	needToSync := false
	allPubKeys := make([]spectypes.ValidatorPK, 0, len(shares))
	for _, share := range shares {
		allPubKeys = append(allPubKeys, share.ValidatorPubKey)
		if !share.HasBeaconMetadata() {
			needToSync = true
		}
	}
	if !needToSync {
		return nil, nil
	}
```

operator/validator/controller.go
```diff
@@ -986,107 +915,52 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
...
-			zap.Uint64("started_validators", c.recentlyStartedValidators),
+			zap.Int("started_validators", startedValidators),
 		)
-		select {
-		case c.indicesChange <- struct{}{}:
-		case <-time.After(2 * c.beacon.GetBeaconNetwork().SlotDurationSec()):
+		// Refresh duties if there are any new active validators.
+		if !c.reportIndicesChange(ctx, 2*c.beacon.GetBeaconNetwork().SlotDurationSec()) {
 			c.logger.Warn("timed out while notifying DutyScheduler of new validators")
 		}
```
moshe-blox, line 942: is there a reason we're not checking the indices like before?

iirc this is the reliable check for new indices and the started check is likely legacy from before

moshe-blox:
very nice to see the metadata code tested finally 💪

operator/validator/controller.go
nkryuchkov: @moshe-blox I saw this comment https://github.com/ssvlabs/ssv/pull/1805#discussion_r1896912761 and thought that as it happens after starting validators then the most reliable way would be to check how many validators actually started. And if I remove `hasNewValidators`, then the code with checking indices becomes unused

operator/validator/controller.go
```diff
@@ -452,21 +444,11 @@ func (c *controller) StartValidators() {
...
-	var pubKeysToFetch [][]byte
 	for _, share := range shares {
 		if c.operatorDataStore.GetOperatorID() != 0 && share.BelongsToOperator(c.operatorDataStore.GetOperatorID()) {
 			ownShares = append(ownShares, share)
 		}
-		networkcommons.SetCommitteeSubnet(intBuf, share.CommitteeID())
-		subnet := intBuf.Uint64()
-		if mySubnets[subnet] != 0 {
-			pubKeysToFetch = append(pubKeysToFetch, share.ValidatorPubKey[:])
-		}
```
y0sher, line 469: I can't find where this part is re-added?

operator/validator/controller.go
moshe-blox: i see

well, to answer whether either 1 is enough or not, research is needed

it's best to not remove unless we understand the implications

if i had to guess the former check is more reliable

i'd keep both in any case until we're sure

operator/validator/controller.go
nkryuchkov: added in https://github.com/ssvlabs/ssv/pull/1805/commits/8a742345cfe887311403c65429f03c5b0cadc45a

nkryuchkov: and https://github.com/ssvlabs/ssv/pull/1805/commits/db78db1b8f54e864b28e72b1e7848108bf99cbe3

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,366 @@
...
+	// to prevent blocking startup after first sync.
+	needToSync := false
+	pubKeysToFetch := make([]spectypes.ValidatorPK, 0, len(shares))
+	for _, share := range shares {
+		networkcommons.SetCommitteeSubnet(subnetsBuf, share.CommitteeID())
+		subnet := subnetsBuf.Uint64()
+		if ownSubnets[subnet] == 0 {
+			continue
+		}
+		pubKeysToFetch = append(pubKeysToFetch, share.ValidatorPubKey)
```
y0sher, line 107: what do you think of filtering the shares by the subnets? I'm not sure we need to go over the shares twice..

nkryuchkov: Yes, we could. We'd need to iterate once again to check if any of them doesn't have a metadata, although we'd be able to `break` the loop in this case

y0sher: sounds better to me, in current code you have to go over all the shares in the db twice.

y0sher: we can break it to a small task to the future though

operator/validator/metadata/syncer.go
```diff
@@ -0,0 +1,366 @@
...
+
+	myValidators := s.validatorStore.SelfValidators()
+	mySubnets := make(records.Subnets, networkcommons.SubnetsCount)
+	for _, v := range myValidators {
+		networkcommons.SetCommitteeSubnet(localBuf, v.CommitteeID())
+		mySubnets[localBuf.Uint64()] = 1
+	}
+
+	return mySubnets
+}
```
y0sher, line 366: this needs to also include FixedSubnets, you might want to pass the config var to here as well..

y0sher (changes requested):
last small changes

operator/validator/metadata/syncer.go
nkryuchkov: changed, but actually we still need to iterate all shares once again because we need a public key list, not a share list

operator/validator/metadata/syncer.go
nkryuchkov: done

operator/validator/metadata/syncer.go
nkryuchkov: @iurii-ssv Indeed this wasn't clear from that comment, I fixed it. It sleeps to wait for some data, otherwise it would likely receive empty data if it tried to get it immediately

iurii-ssv: Thanks for clarifications,

> I think we cannot replace the case inside if done { }, so we can remove just one duplication. Probably an if and one sleep would be a bit more complex to understand than two sleeps

could maybe define a `func` that "sleeps on return, with `defer`" and it keeps fetching & sending batches until it fails to do so

iurii-ssv (approved):

operator/validator/metadata/syncer.go
nkryuchkov: @iurii-ssv do you mean something like extracting the body of the for loop to a helper function?

moshe-blox (approved):
LGTM 👌 

let's not forget to test this properly

PR #1808, opened 2024-10-21
fix: change custom domain type to genesis custom domain type +1 on fork.

y0sher:
### Description

We decided to not merge all custom domains to alan domain on fork, but to make them increase domain type value +1 and stay on a different domain. 

Genesis Domain Type -> Alan Domain Type
Genesis custom domain type -> Alan custom domain type (genesis +1)

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -609,16 +609,16 @@ func setupSSVNetwork(logger *zap.Logger) (networkconfig.NetworkConfig, error) {
 		if len(byts) != 4 {
 			return networkconfig.NetworkConfig{}, errors.New("custom domain type must be 4 bytes")
 		}
+		// Assign domain type as-is for Genesis.
 		networkConfig.GenesisDomainType = spectypes.DomainType(byts)
-		alanDomainUint := binary.LittleEndian.Uint32(byts) + 1
-		alanbyts := make([]byte, 4)
-		binary.LittleEndian.PutUint32(alanbyts, alanDomainUint)
-		networkConfig.AlanDomainType = spectypes.DomainType(alanbyts)
-		if networkConfig.PastAlanFork() {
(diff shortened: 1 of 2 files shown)


cli/operator/node.go
```diff
@@ -608,12 +609,16 @@ func setupSSVNetwork(logger *zap.Logger) (networkconfig.NetworkConfig, error) {
...
 		}
+		// Assign domain type as-is for Genesis.
 		networkConfig.GenesisDomainType = spectypes.DomainType(byts)
-		if networkConfig.PastAlanFork() {
-			logger.Info("custom domain type is ineffective now after Alan fork", fields.Domain(networkConfig.GenesisDomainType))
-		} else {
-			logger.Info("running with custom domain type", fields.Domain(networkConfig.GenesisDomainType))
-		}
+		// Assign domain type incremented by 1 for Alan.
+		alanDomainType := binary.LittleEndian.Uint32(byts) + 1
```
nkryuchkov, line 615: Why do you prefer this solution over having both `CustomGenesisDomainType` and `CustomAlanDomainType` in config?

moshe-blox: @nkryuchkov custom domain type is a temporary feature we released to reduce load on mainnet, and it will be removed as soon as the Alan fork happens.

plus, there are only 2 custom domains deployed right now on mainnet and we're aware of them so we know the +1 can't cause conflicts into the fork :)

moshe-blox (approved):

cli/operator/node.go
nkryuchkov: @y0sher @moshe-blox alright, can we add a comment for this in the description of the `CustomDomainType` config option?

moshe-blox: @nkryuchkov added comment in help msg and also changed to BigEndian to increment the least significant byte (the right-most) like we did manually in `mainnet.go`

nkryuchkov (approved):

PR #1809, opened 2024-10-21
Concurrent slashing protection tests

### LINKED ISSUE #1672: Add parallel slashing protection test
Now that the same validator may run in multiple goroutines (CommitteeRunner) in parallel, we should write a comprehensive test to make sure there are no race conditions and slashing protection is thread-safe.

olegshmuelov:
close #1672

### DIFF
--- a/ekm/signer_key_manager_test.go
+++ b/ekm/signer_key_manager_test.go
@@ -849,9 +849,12 @@ func TestConcurrentSlashingProtectionAttData(t *testing.T) {
 	// Count errors and successes.
 	var slashableErrors, successCount int
 	for err := range errChan {
-		if err != nil && err.Error() == "slashable attestation (HighestAttestationVote), not signing" {
+		if err != nil {
+			if err.Error() != "slashable attestation (HighestAttestationVote), not signing" {
+				require.Fail(t, "unexpected error: %v", err)
+			}
 			slashableErrors++
-		} else if err == nil {
+		} else {
 			successCount++
 		}
 	}
@@ -904,9 +907,12 @@ func TestConcurrentSlashingProtectionBeaconBlock(t *testing.T) {
 	// Count errors and successes.
 	var slashableErrors, successCount int
 	for err := range errChan {
-		if err != nil && err.Error() == "slashable proposal (HighestProposalVote), not signing" {
+		if err != nil {
+			if err.Error() != "slashable proposal (HighestProposalVote), not signing" {
+				require.Fail(t, "unexpected error: %v", err)
+			}
 			slashableErrors++
-		} else if err == nil {
+		} else {
 			successCount++
 		}
 	}
--- a/ekm/signer_key_manager_test.go
+++ b/ekm/signer_key_manager_test.go
@@ -993,7 +993,7 @@ func TestConcurrentSlashingProtectionWithMultipleKeysAttData(t *testing.T) {
 
 					// Perform BLS verification with public key and computed signing root.
 					if !sig.VerifyByte(validator.pk, root[:]) {
-						err = fmt.Errorf("BLS verification failed for signature %x", sigBytes)
+						require.Truef(t, sig.VerifyByte(validator.pk, root[:]), "BLS verification failed for signature %x", sigBytes)
 					}
 					result.signs++
 				}
@@ -1075,7 +1075,7 @@ func TestConcurrentSlashingProtectionWithMultipleKeysBeaconBlock(t *testing.T) {
 
 					// Perform BLS verification with public key and computed signing root.
 					if !sig.VerifyByte(validator.pk, root[:]) {
-						err = fmt.Errorf("BLS verification failed for signature %x", sigBytes)
+						require.Truef(t, sig.VerifyByte(validator.pk, root[:]), "BLS verification failed for signature %x", sigBytes)
 					}
 					result.signs++
 				}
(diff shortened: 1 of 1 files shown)


ekm/signer_key_manager_test.go
```diff
@@ -788,3 +790,127 @@ func TestEkmListAccounts(t *testing.T) {
...
+	close(errChan)
+
+	// Count errors and successes.
+	var slashableErrors, successCount int
+	for err := range errChan {
+		if err != nil && err.Error() == "slashable attestation (HighestAttestationVote), not signing" {
+			slashableErrors++
+		} else if err == nil {
+			successCount++
+		}
```
iurii-ssv, line 856: Shouldn't we check here for the absence of errors other than `"slashable attestation (HighestAttestationVote), not signing"` (or leave a comment explaining why not) ?  Same for another test.

ekm/signer_key_manager_test.go
```diff
@@ -788,3 +790,127 @@ func TestEkmListAccounts(t *testing.T) {
...
+			blockContents.Block,
+			phase0.Domain{},
+			sk1.GetPublicKey().Serialize(),
+			spectypes.DomainProposer,
+		)
+		errChan <- err
+	}
+
+	// Set up concurrency.
+	const goroutineCount = 100
```
iurii-ssv, line 891: Can probably set to 1000 (or even 10000) to get the most out of this test. Same for another test.

iurii-ssv (commented):
Minor suggestions.

ekm/signer_key_manager_test.go
olegshmuelov: The purpose of this test is not to fail fast, but to ensure that under high contention, slashing protection works correctly without causing deadlocks or race conditions. We’re verifying that only one signing succeeds and the rest fail with the expected slashing error

iurii-ssv: I mean, if we encounter any error other than `slashable attestation (HighestAttestationVote), not signing` here - there is no point running this test further (doesn't matter if it deadlocks or races, "functionality doesn't function" at all as we want it to).

iurii-ssv (approved):
LGTM

ekm/signer_key_manager_test.go
olegshmuelov: I see your point! I’ve adjusted the test to fail fast on unexpected error
https://github.com/ssvlabs/ssv/pull/1809/commits/d5c4d31d0d1d6799cd4db0a067cc9024567420e3

y0sher (approved):
good job!

ekm/signer_key_manager_test.go
```diff
@@ -788,3 +790,133 @@ func TestEkmListAccounts(t *testing.T) {
...
+
+	// Define function to concurrently attempt signing.
+	signAttestation := func(wg *sync.WaitGroup, errChan chan error) {
+		defer wg.Done()
+		_, _, err := km.(*ethKeyManagerSigner).SignBeaconObject(
+			attestationData,
+			phase0.Domain{},
+			sk1.GetPublicKey().Serialize(),
+			spectypes.DomainAttester,
+		)
```
moshe-blox, line 831: we can potentially add a check for the sig here

to get a sig to hardcode, we can run this test once without parallelism just to print the sigs we know are good

ekm/signer_key_manager_test.go
```diff
@@ -788,3 +790,133 @@ func TestEkmListAccounts(t *testing.T) {
...
 }
+
+func TestConcurrentSlashingProtectionAttData(t *testing.T) {
+	require.NoError(t, bls.Init(bls.BLS12_381))
+
+	km := testKeyManager(t, nil)
+
+	sk1 := &bls.SecretKey{}
+	require.NoError(t, sk1.SetHexString(sk1Str))
+	require.NoError(t, km.AddShare(sk1))
```
moshe-blox, line 801: we should add another key here, or even make a slice of keys and run X goroutines for each

because EKM locks per key, then this would test multiple locks which greatly increases the coverage

moshe-blox (changes requested):
this looks very good!

added only 2 feature requests to increase coverage

ekm/signer_key_manager_test.go
```diff
@@ -788,3 +790,133 @@ func TestEkmListAccounts(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 2, len(accounts))
 }
+
+func TestConcurrentSlashingProtectionAttData(t *testing.T) {
+	require.NoError(t, bls.Init(bls.BLS12_381))
+
+	km := testKeyManager(t, nil)
```
moshe-blox, line 797: does this use Badger or in-memory storage?

we should make sure to use Badger because this is the risk that SSV adds to the EKM mix

ekm/signer_key_manager_test.go
olegshmuelov: To achieve this, we’d need to adjust the tests to use a mock network to obtain a static slot for signing (or use some MaxUint64) value, which would ensure consistent signature values. This change would require updating all related tests to use the new mock network. Do we still want to proceed with this approach in the current PR?

olegshmuelov: added sig verification instead

ekm/signer_key_manager_test.go
olegshmuelov:
```go
km := testKeyManager(t, nil)
```
using
```go
func getBaseStorage(logger *zap.Logger) (basedb.Database, error) {
    return kv.NewInMemory(logger, basedb.Options{})
}
```

moshe-blox (approved):
LGTM 🔥

Issue #1561, opened 2024-08-06
is `convert.RunnerRole` really needed?


moshe-blox:
It seems like a wrong thing to do, and we use it in lots of places, such as QBFT storage.

For example regarding storage, why can't we store QBFT instances using the RunnerRole? in post-fork we perform consensus per RunnerRole not per BeaconRole, so that seems like the correct thing to use.

y0sher:
@moshe-blox, in exporter, we want to save consensus results (qbft storage..) in the 'old' format, so `convert.RunnerRole` basically enables using `RunnerRole` but also saving Attester and Sync committee duties. Since exporter doesn't share anything about committee consensus now and only about post-consensus.

moshe-blox:
@y0sher i see, but still i think the QBFT storage should use the spectypes role, and only in exporter we should see usage of the convert function (and not in other packages)

PR #1810, opened 2024-10-21
fix: remove convert runner role

anatolie-ssv:
Closes:  #1561


anatolie-ssv:
this will be placed on top of the `remove save qbft instances` PR

PR #1813, opened 2024-10-22
refactor(network/records): change subnets type to [16]byte

nkryuchkov:
The current handling of records is very error-prone and causes bugs like https://github.com/ssvlabs/ssv/issues/1802. The code always expects subnets to be 128 byte long, so we can use `[128]byte` type instead of `[]byte`. If we use each bit for each subnet, we can even use `[16]byte`

### DIFF
--- a/api/handlers/node.go
+++ b/api/handlers/node.go
@@ -11,6 +11,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 
 	"github.com/ssvlabs/ssv/api"
+	"github.com/ssvlabs/ssv/network/commons"
 	networkpeers "github.com/ssvlabs/ssv/network/peers"
 	"github.com/ssvlabs/ssv/nodeprobe"
 )
@@ -170,10 +171,15 @@ func (h *Node) Health(w http.ResponseWriter, r *http.Request) error {
 func (h *Node) peers(peers []peer.ID) []peerJSON {
 	resp := make([]peerJSON, len(peers))
 	for i, id := range peers {
+		subnets, ok := h.PeersIndex.GetPeerSubnets(id)
+		if !ok {
+			subnets = commons.ZeroSubnets
+		}
+
 		resp[i] = peerJSON{
 			ID:            id,
 			Connectedness: h.Network.Connectedness(id).String(),
-			Subnets:       h.PeersIndex.GetPeerSubnets(id).String(),
+			Subnets:       subnets.String(),
 		}
 
 		for _, addr := range h.Network.Peerstore().Addrs(id) {
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -97,8 +97,11 @@ type p2pNetwork struct {
 
 	backoffConnector *libp2pdiscbackoff.BackoffConnector
 
-	fixedSubnets  []byte
-	activeSubnets []byte
+	// persistentSubnets holds subnets on node startup,
+	// these subnets should not be unsubscribed from even if all validators associated with them are removed
+	persistentSubnets commons.Subnets
+	// currentSubnets holds current subnets which depend on current active validators and committees
+	currentSubnets commons.Subnets
 
 	libConnManager connmgrcore.ConnManager
 
@@ -135,7 +138,6 @@ func New(
 		msgValidator:            cfg.MessageValidator,
 		state:                   stateClosed,
 		activeCommittees:        hashmap.New[string, validatorStatus](),
-		activeSubnets:           make([]byte, commons.SubnetsCount),
 		nodeStorage:             cfg.NodeStorage,
 		operatorPKHashToPKCache: hashmap.New[string, []byte](),
 		operatorSigner:          cfg.OperatorSigner,
@@ -320,7 +322,7 @@ func (n *p2pNetwork) peersTrimming(logger *zap.Logger) func() {
 			maximumIrrelevantPeersToDisconnect,
 			n.host.Network(),
 			connectedPeers,
-			n.activeSubnets,
+			n.currentSubnets,
 		)
 		if disconnectedCnt > 0 {
 			// we can accept more peer connections now, no need to trim
@@ -495,7 +497,7 @@ func (n *p2pNetwork) UpdateSubnets(logger *zap.Logger) {
 	// there is a pending PR to replace this: https://github.com/ssvlabs/ssv/pull/990
 	logger = logger.Named(logging.NameP2PNetwork)
 	ticker := time.NewTicker(time.Second)
-	registeredSubnets := make([]byte, commons.SubnetsCount)
+	registeredSubnets := commons.Subnets{}
 	defer ticker.Stop()
 
 	// Run immediately and then every second.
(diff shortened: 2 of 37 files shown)


network/records/subnets.go
```diff
@@ -74,16 +64,21 @@ func (s Subnets) FromString(subnetsStr string) (Subnets, error) {
...
+			return Subnets{}, err
 		}
 		data = append(data, maskData2...)
 		data = append(data, maskData1...)
 	}
-	return data, nil
+
+	if len(data) != commons.Subnets() {
+		return Subnets{}, fmt.Errorf("invalid subnets length: %d", len(data))
+	}
```
iurii-ssv, line 79: Would be nice to also include the **expected** length in error message.

network/records/subnets.go
```diff
@@ -21,41 +22,30 @@ const (
...
 // Subnets holds all the subscribed subnets of a specific node
-type Subnets []byte
-
-// Clone clones the independent byte slice
-func (s Subnets) Clone() Subnets {
-	cp := make([]byte, len(s))
-	copy(cp, s)
-	return cp
-}
+type Subnets [commons.SubnetsCount]byte
```
iurii-ssv, line 48: Perhaps clarify what `byte` values will be expected (from what I've gathered we have `commons.SubnetsCount=128` subnets, and we denote with 1 and 0 whether we participate in a certain subnet or not).

network/p2p/p2p.go
```diff
@@ -95,8 +95,8 @@ type p2pNetwork struct {
 
 	backoffConnector *libp2pdiscbackoff.BackoffConnector
 
-	fixedSubnets  []byte
-	activeSubnets []byte
+	fixedSubnets  records.Subnets
+	activeSubnets records.Subnets
```
iurii-ssv, line 99: Maybe add comments clarifying what these are (in particular `fixedSubnets`) while we are at it.

network/peers/subnets.go
```diff
@@ -29,10 +29,12 @@ func (si *subnetsIndex) UpdatePeerSubnets(id peer.ID, s records.Subnets) bool {
...
 
 	existing, ok := si.peerSubnets[id]
+	var diff map[int]byte
 	if !ok {
-		existing = make([]byte, 0)
+		diff = s.ToMap()
+	} else {
+		diff = existing.DiffSubnets(s)
 	}
-	diff := records.DiffSubnets(existing, s)
```
iurii-ssv, line 35: For better readability, this can be simplified to something like this:
```go
diff := s.ToMap()
if ok {
    diff = existing.DiffSubnets(s)
}
```
unless we are doing it **a lot** (which would create garbage for gc).

network/peers/subnets_test.go
```diff
@@ -35,39 +36,37 @@ func TestSubnetsIndex(t *testing.T) {
...
-		if sub%2 == 0 {
-			mysubnets[sub] = byte(0)
+	mysubnets := records.Subnets{}
+	for i := 0; i < commons.Subnets(); i++ {
+		if i%2 == 0 {
+			mysubnets[i] = byte(0)
 		} else {
-			mysubnets[sub] = byte(1)
+			mysubnets[i] = byte(1)
 		}
```
iurii-ssv, line 70: Could also ditch `else` part in favor of:
```go
mysubnets[i] = byte(1)
if i%2 == 0 {
    mysubnets[i] = byte(0)
}
```

network/peers/conn_manager.go
```diff
@@ -129,7 +129,7 @@ func (c connManager) getBestPeers(n int, mySubnets records.Subnets, allPeers []p
 		peerLogs = append(peerLogs, peerLog{
 			Peer:          pid,
 			Score:         score,
-			SharedSubnets: len(records.SharedSubnets(peerSubnets, mySubnets, len(mySubnets))),
+			SharedSubnets: len(peerSubnets.SharedSubnets(mySubnets, len(mySubnets))),
```
iurii-ssv, line 132: Feels redundant passing `len(mySubnets)` here, should/can be `0` ?

network/peers/conn_manager.go
```diff
@@ -231,8 +231,11 @@ func (c connManager) DisconnectFromBadPeers(logger *zap.Logger, net libp2pnetwor
 func (c connManager) DisconnectFromIrrelevantPeers(logger *zap.Logger, disconnectQuota int, net libp2pnetwork.Network, allPeers []peer.ID, mySubnets records.Subnets) int {
 	disconnectedPeers := 0
 	for _, peerID := range allPeers {
-		peerSubnets := c.subnetsIdx.GetPeerSubnets(peerID)
-		sharedSubnets := records.SharedSubnets(mySubnets, peerSubnets, len(mySubnets))
+		var sharedSubnets []int
+		peerSubnets, ok := c.subnetsIdx.GetPeerSubnets(peerID)
+		if ok {
+			sharedSubnets = mySubnets.SharedSubnets(peerSubnets, len(mySubnets))
+		}
```
iurii-ssv, line 238: Feels redundant passing `len(mySubnets)` here, should/can be `0` ?

iurii-ssv (commented):
Mostly nits from me, definitely reads better with these changes.

network/records/subnets.go
iurii-ssv: Yeah, this is probably redundant here, just helps sometimes.

network/peers/subnets.go
iurii-ssv: Right, that's just some styling preference I guess ;)

iurii-ssv (approved):
LGTM

network/records/subnets.go
nkryuchkov: >  I'm not sure it will improve perf

Well, yeah, I'm pretty sure it's far from being a bottleneck, I just thought about optimizing it more while we are touching it. I'd say it's not just about fitting the cache line; we'd also have fewer loop cycles and faster array copies.  But I'm inclined to think it's not worth spending time at the moment

iurii-ssv: Yup, I wouldn't bother with such a change just now tbh, there seems to be plenty of other improvements to make too :)

network/p2p/p2p.go
```diff
@@ -95,8 +95,10 @@ type p2pNetwork struct {
 
 	backoffConnector *libp2pdiscbackoff.BackoffConnector
 
-	fixedSubnets  []byte
-	activeSubnets []byte
+	// initialSubnets holds subnets on node startup
+	initialSubnets records.Subnets
+	// currentSubnets holds current subnets which depend on current active validators and committees
+	currentSubnets records.Subnets
```
moshe-blox, line 101: this is not really precise because these subnets should not be unsubscribed from even if all validators associated with them are removed

if we don't like 'fixedSubnets', i'd suggest 'stickySubnets' instead, and in that case 'currentSubnets' could be renamed to 'committeeSubnets' to more correctly distinguish it

(assuming this PR will be merged after the Alan fork where there are no more validators subnets, so committee subnets would make sense :)

network/records/subnets.go
```diff
@@ -96,24 +93,34 @@ func (s Subnets) Active() int {
...
+		m[subnet] = v
+	}
+	return m
+}
+
 // SharedSubnets returns the shared subnets
-func SharedSubnets(a, b []byte, maxLen int) []int {
+func (s Subnets) SharedSubnets(other Subnets) []int {
+	return s.SharedSubnetsN(other, 0)
+}
```
moshe-blox, line 108: nice touch 👌

network/p2p/p2p.go
y0sher: another option: `persistentSubnets`

iurii-ssv: Both `stickySubnets` and `persistentSubnets` are good candidates :)

I'd also expand expand the comment description to include this part moshe mentioned:

> these subnets should not be unsubscribed from even if all validators associated with them are removed

as well as throw in couple other options: `primarySubnets` / `preferredSubnets`

nkryuchkov:
Now, after the type is changed, I don't like that we always have to copy 128 bytes, so I'm thinking about changing this to `[16]byte`

api/handlers/node.go
```diff
@@ -169,10 +170,15 @@ func (h *Node) Health(w http.ResponseWriter, r *http.Request) error {
 func (h *Node) peers(peers []peer.ID) []peerJSON {
 	resp := make([]peerJSON, len(peers))
 	for i, id := range peers {
+		subnets, ok := h.PeersIndex.GetPeerSubnets(id)
+		if !ok {
+			subnets = records.ZeroSubnets
+		}
```
y0sher, line 176: can't we just return zerosubnets from `GerPeerSubnets` without `ok`?

PR #1784, opened 2024-10-10
v2-committee: improve concurrency handling

iurii-ssv:
This PR intends to:
- reduce mutex contention by moving potentially long-lasting call (which contains network request I believe) out of section-under-lock
- fix potentially dangerous mutex usage in `HandleMessage` (due to the way it currently locks mutex, it potentially can send message into queue that will get garbage-collected)
- clean up, standardize and clarify things

See additional comments below.

PR #1817, opened 2024-10-25
aggregation-duty: clarify and clean up

iurii-ssv:
Was browsing through aggregator duty flow to better understand how it works, and observing strange error (see discussions below) in logs on stage decided it's worth try and document it some, and maybe fix the error.

Putting this PR in draft for now (it is essentially ready, just want [other functionality](https://github.com/ssvlabs/ssv/pull/1784#discussion_r1812631078) to get merged first so I can rebase and fix pipelines).

### DIFF
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -1,7 +1,6 @@
 package goclient
 
 import (
-	"encoding/binary"
 	"fmt"
 	"time"
 
@@ -17,15 +16,6 @@ func (gc *GoClient) SubmitAggregateSelectionProof(slot phase0.Slot, committeeInd
 	// to broadcast the best aggregate to the global aggregate channel.
 	// https://github.com/ethereum/consensus-specs/blob/v0.9.3/specs/validator/0_beacon-chain-validator.md#broadcast-aggregate
 	gc.waitToSlotTwoThirds(slot)
-
-	// differ from spec because we need to subscribe to subnet
-	isAggregator, err := isAggregator(committeeLength, slotSig)
-	if err != nil {
-		return nil, DataVersionNil, fmt.Errorf("failed to check if validator is an aggregator: %w", err)
-	}
-	if !isAggregator {
-		return nil, DataVersionNil, fmt.Errorf("validator is not an aggregator")
-	}
 
 	attData, _, err := gc.GetAttestationData(slot, committeeIndex)
 	if err != nil {
@@ -71,27 +61,6 @@ func (gc *GoClient) SubmitAggregateSelectionProof(slot phase0.Slot, committeeInd
 // SubmitSignedAggregateSelectionProof broadcasts a signed aggregator msg
 func (gc *GoClient) SubmitSignedAggregateSelectionProof(msg *phase0.SignedAggregateAndProof) error {
 	return gc.client.SubmitAggregateAttestations(gc.ctx, []*phase0.SignedAggregateAndProof{msg})
-}
-
-// IsAggregator returns true if the signature is from the input validator. The committee
-// count is provided as an argument rather than imported implementation from spec. Having
-// committee count as an argument allows cheaper computation at run time.
-//
-// Spec pseudocode definition:
-//
-//	def is_aggregator(state: BeaconState, slot: Slot, index: CommitteeIndex, slot_signature: BLSSignature) -> bool:
-//	 committee = get_beacon_committee(state, slot, index)
-//	 modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
-//	 return bytes_to_uint64(hash(slot_signature)[0:8]) % modulo == 0
-func isAggregator(committeeCount uint64, slotSig []byte) (bool, error) {
-	modulo := committeeCount / TargetAggregatorsPerCommittee
-	if modulo == 0 {
-		// Modulo must be at least 1.
-		modulo = 1
-	}
-
-	b := Hash(slotSig)
-	return binary.LittleEndian.Uint64(b[:8])%modulo == 0, nil
 }
 
 // waitToSlotTwoThirds waits until two-third of the slot has transpired (SECONDS_PER_SLOT * 2 / 3 seconds after the start of slot)
--- a/operator/duties/attester.go
+++ b/operator/duties/attester.go
@@ -9,10 +9,9 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	genesisspectypes "github.com/ssvlabs/ssv-spec-pre-cc/types"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
-	"go.uber.org/zap"
-
 	"github.com/ssvlabs/ssv/logging/fields"
 	"github.com/ssvlabs/ssv/operator/duties/dutystore"
+	"go.uber.org/zap"
 )
 
 type AttesterHandler struct {
@@ -175,8 +174,12 @@ func (h *AttesterHandler) processExecution(epoch phase0.Epoch, slot phase0.Slot)
 	if !h.network.PastAlanForkAtEpoch(h.network.Beacon.EstimatedEpochAtSlot(slot)) {
 		toExecute := make([]*genesisspectypes.Duty, 0, len(duties)*2)
 		for _, d := range duties {
-			if h.shouldExecute(d) {
+			if h.isFresh(d) {
 				toExecute = append(toExecute, h.toGenesisSpecDuty(d, genesisspectypes.BNRoleAttester))
+				// For every attestation duty we also have to try to perform aggregation duty even if it
+				// isn't necessarily needed - we won't know if it's needed or not until we rebuild
+				// validator signature (done during pre-consensus step) and perform some computation on
+				// it - hence scheduling it for execution here.
 				toExecute = append(toExecute, h.toGenesisSpecDuty(d, genesisspectypes.BNRoleAggregator))
 			}
 		}
@@ -187,7 +190,11 @@ func (h *AttesterHandler) processExecution(epoch phase0.Epoch, slot phase0.Slot)
 
 	toExecute := make([]*spectypes.ValidatorDuty, 0, len(duties))
 	for _, d := range duties {
-		if h.shouldExecute(d) {
+		if h.isFresh(d) {
+			// For every attestation duty we also have to try to perform aggregation duty even if it
+			// isn't necessarily needed - we won't know if it's needed or not until we rebuild
+			// validator signature (done during pre-consensus step) and perform some computation on
+			// it - hence scheduling it for execution here.
 			toExecute = append(toExecute, h.toSpecDuty(d, spectypes.BNRoleAggregator))
 		}
 	}
@@ -274,9 +281,9 @@ func (h *AttesterHandler) toGenesisSpecDuty(duty *eth2apiv1.AttesterDuty, role g
 	}
 }
 
-func (h *AttesterHandler) shouldExecute(duty *eth2apiv1.AttesterDuty) bool {
+func (h *AttesterHandler) isFresh(duty *eth2apiv1.AttesterDuty) bool {
 	currentSlot := h.network.Beacon.EstimatedCurrentSlot()
-	// execute task if slot already began and not pass 1 epoch
+	// if a whole epoch of slots has passed since the slot duty is for there is no point executing it
 	var attestationPropagationSlotRange = phase0.Slot(h.network.Beacon.SlotsPerEpoch())
 	if currentSlot >= duty.Slot && currentSlot-duty.Slot <= attestationPropagationSlotRange {
 		return true
--- a/protocol/v2/qbft/controller/controller.go
+++ b/protocol/v2/qbft/controller/controller.go
@@ -6,17 +6,15 @@ import (
 	"encoding/json"
 	"fmt"
 
-	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
-	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
-
 	"github.com/pkg/errors"
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
-	"go.uber.org/zap"
-
 	"github.com/ssvlabs/ssv/logging/fields"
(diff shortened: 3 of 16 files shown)


beacon/goclient/aggregator.go
```diff
@@ -18,15 +17,6 @@ func (gc *GoClient) SubmitAggregateSelectionProof(slot phase0.Slot, committeeInd
...
 	gc.waitToSlotTwoThirds(slot)
 
-	// differ from spec because we need to subscribe to subnet
-	isAggregator, err := isAggregator(committeeLength, slotSig)
-	if err != nil {
-		return nil, DataVersionNil, fmt.Errorf("failed to check if validator is an aggregator: %w", err)
-	}
-	if !isAggregator {
-		return nil, DataVersionNil, fmt.Errorf("validator is not an aggregator")
-	}
```
iurii-ssv, line 28: Doing this here seems to be causing this error consistently showing in logs:
```
{
  "level": "\u001b[35mDEBUG\u001b[0m",
  "time": "2024-10-25T13:17:08.000420Z",
  "msg": "❗ could not handle message",
  "pubkey": "9203935ee396795ab6775b910f94d8d38850cd2331fc4416fdfeb87cc6d4cd764a9f6b7dcddd05b6a9c76a73cd68f100",
  "beacon_role": "AGGREGATOR",
  "signer": 419,
  "slot": 2829985,
  "msg_type": "partial_signature",
  "error": "failed to submit aggregate and proof: validator is not an aggregator",
  "errorVerbose": "validator is not an aggregator\nfailed to submit aggregate and proof\ngithub.com/ssvlabs/ssv/protocol/genesis/ssv/runner.(*AggregatorRunner).ProcessPreConsensus\n\t/go/src/github.com/ssvlabs/ssv/protocol/genesis/ssv/runner/aggregator.go:104\ngithub.com/ssvlabs/ssv/protocol/genesis/ssv/validator.(*Validator).ProcessMessage\n\t/go/src/github.com/ssvlabs/ssv/protocol/genesis/ssv/validator/validator.go:151\ngithub.com/ssvlabs/ssv/protocol/genesis/ssv/validator.(*Validator).ConsumeQueue\n\t/go/src/github.com/ssvlabs/ssv/protocol/genesis/ssv/validator/msgqueue_consumer.go:154\ngithub.com/ssvlabs/ssv/protocol/genesis/ssv/validator.(*Validator).StartQueueConsumer\n\t/go/src/github.com/ssvlabs/ssv/protocol/genesis/ssv/validator/msgqueue_consumer.go:60\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1695"
}
```
so,
- moving this code out of `goclient` because it doesn't really belong here
- and handling `isAggregator==true` a bit differently (just `log.Debug` instead of erorring)

operator/duties/attester.go
```diff
@@ -175,8 +174,12 @@ func (h *AttesterHandler) processExecution(epoch phase0.Epoch, slot phase0.Slot)
...
 		toExecute := make([]*genesisspectypes.Duty, 0, len(duties)*2)
 		for _, d := range duties {
-			if h.shouldExecute(d) {
+			if h.isFresh(d) {
 				toExecute = append(toExecute, h.toGenesisSpecDuty(d, genesisspectypes.BNRoleAttester))
+				// For every attestation duty we also have to try to perform aggregation duty even if it
+				// isn't necessarily needed - we won't know if it's needed or not until we rebuild
+				// validator signature (done during pre-consensus step) and perform some computation on
+				// it - hence scheduling it for execution here.
 				toExecute = append(toExecute, h.toGenesisSpecDuty(d, genesisspectypes.BNRoleAggregator))
```
iurii-ssv, line 183: Did I get this right, or there is more to it ?

protocol/v2/ssv/runner/runner_state.go
```diff
@@ -18,25 +17,24 @@ type State struct {
 	PostConsensusContainer *ssv.PartialSigContainer
 	RunningInstance        *instance.Instance
 	DecidedValue           []byte //spectypes.Encoder
-	// CurrentDuty is the duty the node pulled locally from the beacon node, might be different from decided duty
+	// StartingDuty is the duty the node pulled locally from the beacon node, might be different
+	// from the actual duty operators have decided upon.
 	StartingDuty spectypes.Duty `json:"StartingDuty,omitempty"`
```
iurii-ssv, line 22: > might be different from decided duty

Is the only reason it can be different - because other operators might have gotten different data from their beacon nodes (and hence decided value for operator cluster is different), or there is something else ?

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -98,16 +101,32 @@ func (r *AggregatorRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *sp
...
+		logger.Debug("aggregation duty won't be needed from this validator for this slot",
+			zap.Any("signer", signer),
+			fields.Slot(duty.Slot),
+		)
+		return nil
+	}
+
 	r.metrics.PauseDutyFullFlow()
 	// get block data
-	duty = r.GetState().StartingDuty.(*spectypes.ValidatorDuty)
```
iurii-ssv, line 110: This line is essentially duplicated, looks like a typo (see couple of lines above it).

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -124,6 +143,7 @@ func (r *AggregatorRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *sp
 		DataSSZ: byts,
 	}
 
+	r.metrics.StartConsensus()
 	if err := r.BaseRunner.decide(logger, r, duty.Slot, input); err != nil {
 		return errors.Wrap(err, "can't start new duty runner instance for duty")
 	}
```
iurii-ssv, line 149: Seems `r.metrics.StartConsensus()` call have been missing here ? I see it made in other runners, doesn't look like this is different.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -98,16 +101,32 @@ func (r *AggregatorRunner) ProcessPreConsensus(logger *zap.Logger, signedMsg *sp
...
+	// signer must be same for all messages, at least 1 message must be present (this is validated
+	// prior to ProcessPreConsensus call)
+	signer := signedMsg.Messages[0].Signer
 	duty := r.GetState().StartingDuty.(*spectypes.ValidatorDuty)
 
 	logger.Debug("🧩 got partial signature quorum",
-		zap.Any("signer", signedMsg.Messages[0].Signer), // TODO: always 1?
+		zap.Any("signer", signer),
 		fields.Slot(duty.Slot),
 	)
```
iurii-ssv, line 112: Answering TODO: looks like signer should be same for all messages (this is validated earlier with `PartialSignatureMessages.Validate`).

protocol/v2/qbft/controller/controller.go
```diff
@@ -194,14 +189,17 @@ func (c *Controller) GetIdentifier() []byte {
...
-		return true, nil
+		return fmt.Errorf("qbft instance at height: %d hasn't started yet", specqbft.FirstHeight)
 	}
-
-	return msg.QBFTMessage.Height > c.Height, nil
+	if msg.QBFTMessage.Height > c.Height {
+		return fmt.Errorf("message height: %d > current qbft height: %d", msg.QBFTMessage.Height, c.Height)
+	}
+	return nil
 }
```
iurii-ssv, line 203: Just rewriting this func `isFutureMessage`->`validateMsgHeight` so I can provide a detailed error message on what exactly happened (expected vs actual).

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -102,13 +107,31 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
...
+	if err != nil {
+		return fmt.Errorf("check if validator is an aggregator: %w", err)
+	}
+	if !ok {
+		logger.Debug("aggregation duty won't be needed from this validator for this slot",
+			zap.Any("signer", signer),
+			fields.Slot(duty.Slot),
+		)
+		return nil
+	}
```
iurii-ssv, line 131: This however results in SSV node behavior diverging from spec (meaning spec-tests now are failing, and I'm not sure how we should fix it ... we probably should update spec-tests)

@MatheusFranco99 could you take a look at this PR to provide feedback on whether these changes make sense ? Or maybe there is something deeper to it I'm missing.

MatheusFranco99: Hey @iurii-ssv 
Nice work on this PR! The comments with better explanations are great
The spec test diffs are probably just some error strings, right?
The changes make sense, and I'm particularly fine with adding the check that the validator is an aggregator to the spec level, which should make it more transparent.
Regarding the reordering with the ⅔ timer, I think it's fine
@GalRogozinski what do you think?

GalRogozinski: Not sure where is the reordering with the timer @MatheusFranco99?

The check is good...
We need to sync with @y0sher on priority for making it compatible with spec

MatheusFranco99: Yes, as I understand it, it shouldn't affect

iurii-ssv: > The spec test diffs are probably just some error strings, right?

@MatheusFranco99 @GalRogozinski 

I found a way to fix the way we mock some stuff for `AggregatorRunner` in **spec-tests** so that they are passing now, so this PR is essentially ready for review/merge,

I'm not sure if you need to adjust any docs in for spec repo so that it's aligned with this change though

iurii-ssv:
Force-pushing to make sure every commit in this PR is gpg-signed (otherwise we cannot merge it)

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -529,3 +547,51 @@ func constructVersionedSignedAggregateAndProof(aggregateAndProof spec.VersionedA
...
+		modulo = 1
+	}
+
+	b := hashSha256(slotSig)
+	return binary.LittleEndian.Uint64(b[:8])%modulo == 0
+}
+
+var sha256Pool = sync.Pool{New: func() interface{} {
+	return sha256.New()
+}}
```
y0sher: I think it's probably better if this wasn't "static" so we would create it inside the aggregator runner since it's using it. but I'm not stubborn on this change since this was like this just in another package before.

y0sher (dismissed):
I like this PR and the improved code quality :) good job.

protocol/v2/ssv/runner/aggregator.go
iurii-ssv: Yeah (I didn't look much into this, just moved this code over), indeed it seems like we could limit the scope of vars/funcs involved - applied in https://github.com/ssvlabs/ssv/pull/1817/commits/13d7d1a096ca8e4acc2a1330763603240971841a

~~Also note, it seems the previous implementation suffered from a potential concurrency issue (even though it relied on a Pool that's thread-safe by itself it was still possible for multiple go-routines to operate on the same `hash.Hash` instance concurrently and incorrectly due to the need to call `Hash.Reset`) - in that latest commit I'm fixing it with an explicit/dedicated mutex (I guess we either have to use mutex or create new `hash.Hash` instance every time, and mutex in this case seems very reasonable due to low/non-existent contention).~~

^ Never mind, that "concurrency issue" is a non-issue actually due to `sync.Pool` providing unique object references (no 2 go-routines can share the same object in our case), so I [reverted the `sync.Mutex` back to `sync.Pool`](https://github.com/ssvlabs/ssv/pull/1817/commits/5a88900c259028802065a00fd9a8de2a0332c2fa) as it seems like a better fit for this (but also simplified the type-check in there since there is no need for it due to all objects in the pool being of type `hash.Hash`).

y0sher (approved):
lgtm

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -530,3 +560,55 @@ func constructVersionedSignedAggregateAndProof(aggregateAndProof spec.VersionedA
...
+	h := newHasher()
+	return func(targetAggregatorsPerCommittee uint64, committeeCount uint64, slotSig []byte) bool {
+		modulo := committeeCount / targetAggregatorsPerCommittee
+		if modulo == 0 {
+			// Modulo must be at least 1.
+			modulo = 1
+		}
+
+		b := h.hashSha256(slotSig)
+		return binary.LittleEndian.Uint64(b[:8])%modulo == 0
```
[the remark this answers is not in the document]
iurii-ssv: Yes that's what we are doing here.

protocol/v2/ssv/runner/aggregator_test.go
```diff
@@ -0,0 +1,41 @@
...
+	const committeeCount = 10
+
+	slotSig := []byte(randStringBytes(64))
+
+	isAggFn := isAggregatorFn()
+	sampleResult := isAggFn(targetAggregatorsPerCommittee, committeeCount, slotSig)
+
+	const goRoutines = 1000
+	results := make(chan bool)
+	for i := 0; i < goRoutines; i++ {
```
oleg-ssvlabs: nit:
since index isn't utilized, this for loop could be simplified:

`for range goRoutines { }`

protocol/v2/ssv/runner/aggregator_test.go
```diff
@@ -0,0 +1,41 @@
...
+		go func() {
+			result := isAggFn(targetAggregatorsPerCommittee, committeeCount, slotSig)
+			results <- result
+		}()
+	}
+	for i := 0; i < goRoutines; i++ {
```
oleg-ssvlabs: same as above

oleg-ssvlabs (approved):
lgtm.  

Really good idea to move the "is aggregator" check from the `goclient`, the runner seems to be a more appropriate place.  

Just a nit comment, optional to fix.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -52,7 +68,7 @@ func NewAggregatorRunner(
 	operatorSigner ssvtypes.OperatorSigner,
 	valCheck specqbft.ProposedValueCheckF,
 	highestDecidedSlot phase0.Slot,
-) (Runner, error) {
+) (*AggregatorRunner, error) {
```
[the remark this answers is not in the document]
iurii-ssv: I think this package is only used in ssv-repo internally, so I wouldn't limit ourselves to APIs we've defined prior (but not necessarily knew what's best)

PR #1819, opened 2024-10-26
node: startup cleanup plus enable jemalloc

iurii-ssv:
This PR contains two different features (I bundled these together so I could test both extensively on stage in one fell swoop):
- while I've been trying to run SSV node in Docker locally I've encountered a bunch of minor issues (mostly with the way config parsing is handled), this PR addresses all of these
- enable **jemalloc** (it works fine with **openssl** on for go/jemalloc/badgerdb version in this PR, as per my testing done on stage)

_Before merging:_
- [ ] test on **stage**


cli/bootnode/boot_node.go
```diff
@@ -47,12 +44,11 @@ var StartBootNodeCmd = &cobra.Command{
...
 			},
 		)
-
 		if err != nil {
-			log.Fatal(err)
+			log.Fatalf("could not create logger: %v", err)
 		}
-
 		logger := zap.L()
+		defer logging.CapturePanic(logger)
```
iurii-ssv, line 51: For `bootnode` this seems have been missing.

cli/config/config.go
```diff
@@ -1,21 +1,26 @@
...
 
 // Args expose available global args for cli command
 type Args struct {
-	ConfigPath      string
+	// ConfigPath is a path to main configuration file.
+	ConfigPath string
+	// ShareConfigPath is an additional config file (path) that (if present) will overwrite
+	// configuration supplied from config file at ConfigPath.
 	ShareConfigPath string
 }
```
iurii-ssv, line 17: Perhaps we could add another comment line explaining why we might need 2 config files (not just 1).

cli/config/config.go
```diff
@@ -33,5 +38,29 @@ func ProcessArgs(cfg interface{}, a *Args, cmd *cobra.Command) {
...
+		}
+	}
+
+	if a.ConfigPath == "" && a.ShareConfigPath == "" {
+		// When no config file is provided fall back to env vars + default annotations defined on cfg struct.
+		err := cleanenv.ReadEnv(cfg)
+		if err != nil {
+			return fmt.Errorf("could not set up config based on environment variables and struct defaults: %w", err)
+		}
+	}
```
iurii-ssv, line 63: This makes sure `cfg` gets initialized with at least some defaults **when no config file is provided** (currently it seems `cfg` can only work with configuration file though because not all the required values are (or can be)  specified directly on the config struct), 

so this gets rid of a `panic` like:
```
goroutine 1 [running]:
go.uber.org/zap/zapcore.multiCore.Enabled(...)
	/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/tee.go:68
go.uber.org/zap.(*Logger).check(0x40001af000, 0xff, {0x1b4c044, 0xf})
	/go/pkg/mod/go.uber.org/zap@v1.27.0/logger.go:331 +0x60
go.uber.org/zap.(*Logger).Debug(0x1b4779a?, {0x1b4c044?, 0x185dfc0?}, {0x4000041300, 0x4, 0x4})
	/go/pkg/mod/go.uber.org/zap@v1.27.0/logger.go:238 +0x40
github.com/ssvlabs/ssv/logging.SetGlobalLogger.func1()
	/go/src/github.com/ssvlabs/ssv/logging/global.go:37 +0x2b4
github.com/ssvlabs/ssv/logging.SetGlobalLogger({0x0?, 0x40e4ec?}, {0x0, 0x0}, {0x0, 0x0}, 0x400007ca20)
	/go/src/github.com/ssvlabs/ssv/logging/global.go:100 +0x8d0
github.com/ssvlabs/ssv/cli/operator.setupGlobal()
	/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:466 +0x18c
github.com/ssvlabs/ssv/cli/operator.init.func2(0x2e4ee60, {0x1b3c80e?, 0x4?, 0x1b3c7c6?})
	/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:109 +0xa4
github.com/spf13/cobra.(*Command).execute(0x2e4ee60, {0x40002e9180, 0x1, 0x1})
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:944 +0x63c
github.com/spf13/cobra.(*Command).ExecuteC(0x2e4da40)
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x320
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
github.com/ssvlabs/ssv/cli.Execute({0x1b41a3a?, 0x5?}, {0x400059ea00?, 0x2?})
	/go/src/github.com/ssvlabs/ssv/cli/cli.go:27 +0x88
main.main()
	/go/src/github.com/ssvlabs/ssv/cmd/ssvnode/main.go:20 +0xb0
make: *** [Makefile:119: start-node] Error 2
```
in favor of easily readable error message being logged:
```
2024/10/18 12:34:32 could not create logger: could not set up config based on environment variables: field "Addr" is required but the value is not provided
```

go.mod
```diff
@@ -242,5 +241,3 @@ require (
 )
 
 replace github.com/google/flatbuffers => github.com/google/flatbuffers v1.11.0
-
-replace github.com/dgraph-io/ristretto => github.com/dgraph-io/ristretto v0.1.1-0.20211108053508-297c39e6640f
```
iurii-ssv, line 246: Not sure what this line is about, 

I've upgraded `badgerdb` to the latest release in this PR - hence we must remove it for code to compile (or at least adjust somehow).

nodeprobe/nodeprobe.go
```diff
@@ -11,6 +11,10 @@ import (
 )
 
 const (
+	EvenSyncerNode      = "Ethereum event syncer"
```
nkryuchkov, line 14:
```suggestion
	EventSyncerNode      = "Ethereum event syncer"
```

Dockerfile
```diff
@@ -38,13 +46,15 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
-  -tags="blst_enabled" \
+  -tags="blst_enabled,jemalloc,allocator" \
   -ldflags "-X main.Commit=$COMMIT -X main.Version=$VERSION -linkmode external -extldflags \"-static -lm\"" \
   ./cmd/ssvnode
 
 #
 # STEP 3: Prepare image to run the binary
 #
+# IMPORTANT: before upgrading to go 1.23 or higher make sure the following issue has been resolved:
+# https://github.com/golang/go/issues/69978
```
nkryuchkov, line 57: It was resolved a few hours ago, so now we can delete this comment

Makefile
```diff
@@ -13,14 +13,12 @@ ifndef BUILD_PATH
...
-NODE_COMMAND=--config=${CONFIG_PATH}
+NODE_COMMAND_ARGS=--config=${CONFIG_PATH}
 ifneq ($(SHARE_CONFIG),)
-  NODE_COMMAND+= --share-config=${SHARE_CONFIG}
+  NODE_COMMAND_ARGS+= --share-config=${SHARE_CONFIG}
 endif
 
-# Bootnode command.
-BOOTNODE_COMMAND=--config=${CONFIG_PATH}
+BOOTNODE_COMMAND_ARGS_ARGS=--config=${CONFIG_PATH}
```
nkryuchkov, line 21: The variable name is a bit hard to read IMO

cli/config/config.go
```diff
@@ -1,21 +1,26 @@
...
 
-// GlobalConfig expose available global config for cli command
-type GlobalConfig struct {
-	LogLevel       string `yaml:"LogLevel" env:"LOG_LEVEL" env-default:"info" env-description:"Defines logger's log level'"`
-	LogFormat      string `yaml:"LogFormat" env:"LOG_FORMAT" env-default:"console" env-description:"Defines logger's encoding, valid values are 'json' (default) and 'console''"`
-	LogLevelFormat string `yaml:"LogLevelFormat" env:"LOG_LEVEL_FORMAT" env-default:"capitalColor" env-description:"Defines logger's level format, valid values are 'capitalColor' (default), 'capital' or 'lowercase''"`
+// Global expose available global config for cli command
+type Global struct {
+	LogLevel       string `yaml:"LogLevel" env:"LOG_LEVEL" env-default:"info" env-description:"Defines logger's log level"`
+	LogFormat      string `yaml:"LogFormat" env:"LOG_FORMAT" env-default:"console" env-description:"Defines logger's encoding, valid values are 'json' and 'console'(default)"`
```
nkryuchkov, line 22:
```suggestion
	LogFormat      string `yaml:"LogFormat" env:"LOG_FORMAT" env-default:"console" env-description:"Defines logger's encoding, valid values are 'json' and 'console' (default)"`
```

nodeprobe/nodeprobe.go
```diff
@@ -109,7 +110,7 @@ func (p *Prober) probe(ctx context.Context) {
 	p.healthy.Store(healthy.Load())
 
 	if !p.healthy.Load() {
-		p.logger.Error("not all nodes are healthy")
+		p.logger.Error("not all Ethereum nodes are healthy")
```
nkryuchkov, line 113: While I agree that "Ethereum nodes" is clearer, we also add the event syncer to the list of nodes, and it doesn't belong to eth nodes

nkryuchkov (commented):
Great job

Makefile
iurii-ssv: Whoa, that's just a typo, nice that you spotted it!

nodeprobe/nodeprobe.go
iurii-ssv: applied

nodeprobe/nodeprobe.go
iurii-ssv: adjusted log line to list all 3:
```go
		p.logger.Error(fmt.Sprintf(
			"not all nodes (%s, %s, %s), are healthy",
			EventSyncerNode,
			ConsensusClientNode,
			ExecutionClientNode,
		))
```

storage/kv/badger.go
```diff
@@ -91,6 +92,11 @@ func createDB(logger *zap.Logger, options basedb.Options, inMemory bool) (*Badge
 		go badgerDB.periodicallyCollectGarbage(logger, options.GCInterval)
 	}
 
+	out := z.CallocNoRef(1, "jemalloc check")
+	defer z.Free(out)
+	jemallocEnabled := len(out) > 0
+	logger.Debug("jemalloc allocator will be used", zap.Bool("jemalloc_enabled", jemallocEnabled))
```
iurii-ssv: This is pretty much the best way [I found](https://github.com/dgraph-io/badger/blob/7f657f862cbde20503d33a546cdfc928e8d833cd/badger/main.go#L47-L50) to check if `jemalloc` is actually used or not,

note, we print **true** both for `-tags="blst_enabled,jemalloc"` and `-tags="blst_enabled,jemalloc,allocator"`, and hence it reports on the presence of jemalloc-only (nothing about `allocator`), 

`allocator` is a "different strategy" to approach memory management (from what I understand, it serves as a kind of cache between go application and jemalloc buffering large chunks of memory - hence it will make less cgo calls but will be holding onto large unused memory chunks increasing overall memory consumption):
```go
// Allocator amortizes the cost of small allocations by allocating memory in
// bigger chunks.  Internally it uses z.Calloc to allocate memory. Once
// allocated, the memory is not moved, so it is safe to use the allocated bytes
// to unsafe cast them to Go struct pointers. Maintaining a freelist is slow.
// Instead, Allocator only allocates memory, with the idea that finally we
// would just release the entire Allocator.
type Allocator struct {
	sync.Mutex
	compIdx uint64 // Stores bufIdx in 32 MSBs and posIdx in 32 LSBs.
	buffers [][]byte
	Ref     uint64
	Tag     string
}
```
It's hard to tell without doing any tests (or me not having enough context on how exactly we use Badger), but I think if we want to reduce memory footprint as much as possible - we better go with `-tags="blst_enabled,jemalloc"`.

iurii-ssv: > It's hard to tell without doing any tests (or me not having enough context on how exactly we use Badger), but I think if we want to reduce memory footprint as much as possible - we better go with -tags="blst_enabled,jemalloc"

Having done some testing on stage I would say there isn't a noticeable difference for the workloads we do (for operator managing ~500 validators), see https://github.com/ssvlabs/ssv/issues/1643#issuecomment-2447336686 for more details. So I've added `allocator` tag as well to match how it originally was configured back when we first started using `jemalloc`.

Makefile
```diff
@@ -111,28 +109,23 @@ build:
...
 .PHONY: docker
 docker:
 	@echo "node ${NODES_ID}"
-	@docker rm -f ssv_node && docker build -t ssv_node . && docker run -d --env-file .env --restart unless-stopped --name=ssv_node -p 13000:13000 -p 12000:12000/udp -it ssv_node make BUILD_PATH=/go/bin/ssvnode  start-node && docker logs ssv_node --follow
-
-.PHONY: docker-image
-docker-image:
-	@echo "node ${NODES_ID}"
-	@sudo docker rm -f ssv_node && docker run -d --env-file .env --restart unless-stopped --name=ssv_node -p 13000:13000 -p 12000:12000/udp 'ssvlabs/ssv-node:latest' make BUILD_PATH=/go/bin/ssvnode start-node
+	@docker rm -f ssv_node && docker build -t ssv_node . && docker run --env-file .env --name=ssv_node -p 16000:16000 -p 13001:13001 -p 12001:12001/udp -v ./data:/data -it ssv_node make BUILD_PATH=/go/bin/ssvnode start-node && docker logs ssv_node --follow
```
iurii-ssv, line 128: I feel like these docker commands are outdated - so I'm replacing these with a new one that I found to be useful for testing locally.

cli/operator/node.go
```diff
@@ -721,68 +701,57 @@ func setupEventHandling(
...
 	}
 
-	// load & parse local events yaml if exists, otherwise sync from contract
-	if len(cfg.LocalEventsPath) != 0 {
-		localEvents, err := localevents.Load(cfg.LocalEventsPath)
-		if err != nil {
-			logger.Fatal("failed to load local events", zap.Error(err))
-		}
+	// Sync historical registry events from Ethereum smart contract.
+	logger.Debug("syncing historical registry events", zap.Uint64("fromBlock", fromBlock.Uint64()))
```
iurii-ssv, line 705: This code is unreachable because `setupEventHandling` is only ever called when `len(cfg.LocalEventsPath) == 0`. Which kind of suggests we are never calling `EventHandler.HandleLocalEvents` ... which is kind of weird ?

I also don't see `cfg.LocalEventsPath` being used anywhere to actually read the contents of the file, not sure if such usage was removed in the past or just hasn't been added yet, @nkryuchkov   could you elaborate ? I'd rather clean it up while we are at it (by maybe removing `cfg.LocalEventsPath` altogether if we don't need it anymore).

nkryuchkov: We always need to call `setupEventHandling` because it handles both on-chain and local events. `cfg.LocalEventsPath` is used by `localevents.Load(cfg.LocalEventsPath)`

cli/operator/node.go
```diff
@@ -736,68 +716,57 @@ func setupEventHandling(
 		fromBlock = new(big.Int).SetUint64(fromBlock.Uint64() + 1)
 	}
 
-	// load & parse local events yaml if exists, otherwise sync from contract
-	if len(cfg.LocalEventsPath) != 0 {
-		localEvents, err := localevents.Load(cfg.LocalEventsPath)
```
nkryuchkov, line 741: https://github.com/ssvlabs/ssv/pull/1819/files#r1826980844 `cfg.LocalEventsPath` is used here

cli/operator/node.go
```diff
@@ -350,39 +361,39 @@ var StartNodeCmd = &cobra.Command{
...
+		}
 
 		nodeProber.Start(cmd.Context())
 		nodeProber.Wait()
-		logger.Info("ethereum node(s) are healthy")
 
-		metricsReporter.SSVNodeHealthy()
+		logger.Info("Ethereum nodes are healthy")
 
-		eventSyncer := setupEventHandling(
```
nkryuchkov, line 371: `setupEventHandling` is always called here and contains the logic for both on-chain and local events

cli/operator/node.go
```diff
@@ -736,68 +716,57 @@ func setupEventHandling(
...
-		if err != nil {
-			logger.Fatal("failed to load local events", zap.Error(err))
-		}
+	// Sync historical registry events from Ethereum smart contract.
+	logger.Debug("syncing historical registry events", zap.Uint64("fromBlock", fromBlock.Uint64()))
 
-		if err := eventHandler.HandleLocalEvents(localEvents); err != nil {
-			logger.Fatal("error occurred while running event data handler", zap.Error(err))
-		}
-	} else {
```
nkryuchkov, line 749: https://github.com/ssvlabs/ssv/pull/1819/files#r1826980844 logic for on-chain events

cli/operator/node.go
iurii-ssv: > cfg.LocalEventsPath is used by localevents.Load(cfg.LocalEventsPath)

But like I mentioned above, we never enter the branch that executes `localevents.Load(cfg.LocalEventsPath)`.

nkryuchkov: @iurii-ssv in your PR we never do because you call `setupEventHandling` only `if !usingLocalEvents`, but on stage we always call `setupEventHandling`, which calls `localevents.Load(cfg.LocalEventsPath)` `if len(cfg.LocalEventsPath) != 0`

iurii-ssv: Oh, you are right ... I see what's going on. On **stage** the following happens:
```go
		eventSyncer := setupEventHandling(
			cmd.Context(),
			logger,
			executionClient,
			validatorCtrl,
			metricsReporter,
			networkConfig,
			nodeStorage,
			operatorDataStore,
			operatorPrivKey,
			keyManager,
		)
		if len(cfg.LocalEventsPath) == 0 {
		}
```
And `setupEventHandling` returning `eventSyncer` baited me into thinking "there is no need to even call `setupEventHandling` if we aren't gonna do `nodeProber.AddNode`" - and I merged these 2 under `if len(cfg.LocalEventsPath) == 0`.

Let me revert that part, and see how I can separate those things from each other (to make it clear & explicit).

iurii-ssv: Okay, reworked this a bit, relevant commit is here https://github.com/ssvlabs/ssv/pull/1819/commits/9900a0bc9a925ec2abfce7cd5ff376f369fd61e4 and PR is ready (but I'll redeploy to stage to test it again 1 more time).

What's not 100% clear to me though is "what those local events are", I thought of it as some kind of cache - but that's unlikely to be correct because we don't "subscribe" for events that come after the last even in `cfg.LocalEventsPath` file, could you clarify (and I'll add a comment about it somewhere) ?

cli/operator/node.go
```diff
@@ -344,46 +358,80 @@ var StartNodeCmd = &cobra.Command{
...
-		nodeProber.Wait()
-		logger.Info("ethereum node(s) are healthy")
-
-		metricsReporter.SSVNodeHealthy()
-
-		eventSyncer := setupEventHandling(
-			cmd.Context(),
-			logger,
-			executionClient,
+		eventFilterer, err := executionClient.Filterer()
```
nkryuchkov, line 379: Can we try to extract some lines to a function/method/entity to avoid increasing the size of this function? It's quite large and IMO we need to gradually decrease its size

cli/operator/node.go
```diff
@@ -679,135 +696,85 @@ func setupConsensusClient(
...
 }
 
-func setupEventHandling(
+func startMetricsHandler(ctx context.Context, logger *zap.Logger, db basedb.Database, metricsReporter metricsreporter.MetricsReporter, port int, enableProf bool) {
+	logger = logger.Named(logging.NameMetricsHandler)
+	// init and start HTTP handler
+	metricsHandler := metrics.NewMetricsHandler(ctx, db, metricsReporter, enableProf, operatorNode.(metrics.HealthChecker))
+	addr := fmt.Sprintf(":%d", port)
+	if err := metricsHandler.Start(logger, http.NewServeMux(), addr); err != nil {
+		logger.Panic("failed to serve metrics", zap.Error(err))
```
nkryuchkov, line 705: Why do you prefer `Panic` over `Fatal` here?

nodeprobe/nodeprobe.go
```diff
@@ -109,7 +110,12 @@ func (p *Prober) probe(ctx context.Context) {
 	p.healthy.Store(healthy.Load())
 
 	if !p.healthy.Load() {
-		p.logger.Error("not all nodes are healthy")
+		p.logger.Error(fmt.Sprintf(
```
nkryuchkov, line 113: probably zap fields would be a bit more readable, but I don't mind leaving this as is

nkryuchkov (approved):
LGTM, but I'd prefer not to increase the size of code initializing the node

cli/operator/node.go
iurii-ssv: I agree that we want function size to be smaller (not larger) pretty much all the time, but this is not the only/best thing to optimize for - the more important thing is that general code structure stays predictable (easy to understand and navigate),

so in that sense grouping related code together rather than splitting it apart (possibly mixing it with unrelated things - aka mixing different abstraction layers) could be a better choice - and [above is a good example](https://github.com/ssvlabs/ssv/pull/1819#discussion_r1827037473) of why we might want to "untangle" code sometimes

and now that it is untangled - we can find a better way to refactor `StartNodeCmd.Run` - which I'm trying to do in https://github.com/ssvlabs/ssv/pull/1843 (better do it in a separate PR cause this one is getting larger and larger already, plus it's also better to do after we merge https://github.com/ssvlabs/ssv/pull/1820 cause it removes some code from there too)

cli/operator/node.go
iurii-ssv: Honestly - no idea, I guess it's just an old code,

and we have 4 `logger.Panic` usages in the codebase in total too, changed to `Fatal ` everywhere.

nodeprobe/nodeprobe.go
iurii-ssv: Changed to
```go
		p.logger.Error(
			"not all nodes are healthy",
			zap.Strings("nodes", []string{EventSyncerNode, ConsensusClientNode, ExecutionClientNode}),
		)
```

y0sher:
@moshe-blox @liorrutenberg 
IMO this looks good, and for now it runs on exporter for a while and we are not seeing any issue. I'm in for merging it in the node. but if you want more investigation or clarification on what happened last time we need to dedicate time for it and maybe run previous versions to reproduce and dig in.

y0sher (approved):
LGTM, I'm vote for merging this :)

y0sher (dismissed):
lgtm

iurii-ssv:
I've rebased onto latest **stage** branch but couldn't deploy a cluster to stage env to test it out (to make sure nothing broke after rebase) because Discord-bot seems to be "broken" at the moment,

so we probably should do that final test-deploy (**before** merging) - I'll try it later again and let you know

cli/operator/node.go
nkryuchkov: @iurii-ssv Sorry, I didn't see the notification and saw your question just now. This is likely not relevant anymore but local events are used for testing, we just pass the list of events instead of getting them from the node

iurii-ssv:
Finally managed to deploy this to `stage` and it seems to be running fine.

iurii-ssv:
We are hesitant to merge this PR because both **jemalloc** and **openssl** are dynamically-loaded shared libraries, and as it currently stands Golang doesn't really support running it like that for a single OS process - https://github.com/golang/go/issues/65050#issuecomment-2586175729

Recently I've looked a bit more into **jemalloc** - it seems when we install **jemalloc** package it creates both dynamic and static library (and the you'd need to choose which one to use like this: https://github.com/jemalloc/jemalloc/wiki/Getting-Started)

and then here `ristretto` sets both types of linking here https://github.com/hypermodeinc/ristretto/blob/main/z/calloc_jemalloc.go#L11

but there are 2 unfortunate aspects with this:
- by **default** linker prefers dynamic-linking if both dynamic and static libraries are found
- it seems we cannot override those hardcoded linker settings (eg. to only specify static linking library path) - https://github.com/hypermodeinc/ristretto/issues/383
- I assume jemalloc-package we install has both static & dynamic libraries in it - so that user can choose whichever linking mode he prefers by setting linker-flags

so if we want to make sure **jemalloc** is getting linked **statically** (and not dynamically) we might need to additionally remove any dynamic libraries from jemalloc-package we install

iurii-ssv:
Closing this one since it's super stale by now

PR #1820, opened 2024-10-26
alan: cleanup after fork

nkryuchkov:


iurii-ssv (approved):
Glad to see we can remove 20k+ lines of code :)

It's hard to tell though if something was accidentally broken with such a change or not (and I'd also rather remove it all at once than do it gradually piece by piece) - as long as test pipelines are passing and it's working fine on stage - LGTM!

cli/operator/node.go
```diff
@@ -614,33 +568,22 @@ func setupSSVNetwork(logger *zap.Logger) (networkconfig.NetworkConfig, error) {
...
 		}
 
-		// Assign domain type as-is for Genesis.
-		networkConfig.GenesisDomainType = spectypes.DomainType(byts)
-
-		// Assign domain type incremented by 1 for Alan.
-		alanDomainType := binary.BigEndian.Uint32(byts) + 1
-		binary.BigEndian.PutUint32(networkConfig.AlanDomainType[:], alanDomainType)
+		domainType := binary.BigEndian.Uint32(byts)
+		binary.BigEndian.PutUint32(networkConfig.DomainType[:], domainType)
```
y0sher, line 572: since we previously used alan custom domain as custom domain +1 , I think we need to put more thought into this code. At least lets try to leave it behaving as it was before. 
If you use this code and restart your node your going to be on a different domain than the other nodes in your custom domain that still didn't update. So it seems like we're going to need to keep increasing the custom domain+1 until we sunset the custom domains. 
It's a little bit tricky and for sure not pretty, but we're counting on the fact the operators won't update the custom domain themselves so we must do this increase to prevent a "fork".

PR #1823, opened 2024-10-28
feat: bump ekm version

y0sher:
### Description

We were using a [EKM](https://github.com/ssvlabs/eth2-key-manager/) version that is based on an open PR (with a fix to a deadlock we found during testing alan).

This PR updates `ssv` to use a fixed tag after the PR was merged and a tag was released in EKM.


moshe-blox (approved):
👌

PR #1833, opened 2024-10-30
fix: (exporter) update participants atomically

nkryuchkov:
Currently updating participants is not done atomically because we don't use badgerdb transactions, so there's a possibility of data races and missing some participants. 

The PR makes updating participants an atomical operation


ibft/storage/store.go
```diff
@@ -203,16 +221,49 @@ func (i *ibftStorage) GetParticipants(identifier convert.MessageID, slot phase0.
...
+
+	result := make([]spectypes.OperatorID, 0, len(seen))
+	for operatorID := range seen {
+		result = append(result, operatorID)
+	}
+
+	slices.Sort(result)
+
+	return result
+}
```
moshe-blox, line 255: i suggest to use slices.Concat and slices.Compact instead, so this func can be just 2-3 lines

https://pkg.go.dev/slices#Compact

ibft/storage/store.go
```diff
@@ -149,22 +150,35 @@ func (i *ibftStorage) CleanAllInstances(logger *zap.Logger, msgID []byte) error
...
+	if err := i.saveParticipants(txn, identifier, slot, mergedParticipants); err != nil {
+		return false, fmt.Errorf("save participants: %w", err)
+	}
+
+	if err := txn.Commit(); err != nil {
+		return false, fmt.Errorf("commit transaction: %w", err)
+	}
+
+	return true, nil
 }
```
moshe-blox, line 182: im not sure what would happen if this func is called in parallel, is `i.db.Begin()` equivalent to a mutex lock in this case?

moshe-blox (changes requested):
good catch, left 2 comments

ibft/storage/store.go
iurii-ssv: Yeah, it seems BadgerDB transaction is not doing the kind of locking we typically expect from `transaction` as concept,

from what I see in their docs/code it locks only inside of `txn.Commit()` for a short period of time to check for conflicting writes (and returns error if there are any), I think they do it to get the most performance and provide optionality (because synchronous API everybody is used to isn't hard to build on top of that).

There is also some usage of transactions in `ssv/eth/eventhandler` package, but from cursory overview of the code it seems to only do sequential updates - so nothing to worry about.

nkryuchkov: I thought it was sufficient without checking implementation details but actually I just looked at the implementation of `Commit` and now I think it's not enough, we need a mutex instead

moshe-blox: @iurii-ssv thanks for checking this
@nkryuchkov looks good now with mutex

moshe-blox:
@nkryuchkov i think we should keep both the tx and the lock

eventually the interface of this store should be updated to allow passing tx, like most other store methods do, this is to allow callers to demand their desired modifications to be done atomically (such as during event syncing and migrations)

nkryuchkov:
@moshe-blox I don't mind, done

moshe-blox (approved):

olegshmuelov (approved):
LGTM

y0sher (approved):
great catch!

Issue #1575, opened 2024-08-07
SSZ Share Storage


moshe-blox:
We should consider doing away with `gob` because it's slow and way too coupled with Go internals, so confidence any time the `Share` structure changes is low.

This would require a database migration, which, if we didn't want to do anyway, switching to SSZ is wouldn't be worth the effort.

Issue #1473, opened 2024-07-16
Unify access to ValidatorIndex across the codebase


y0sher:
### Rationale 

Validator index is used in many place to identify validators, it is fetched from the beacon node as validator metadata and save in the share. [Latest changes](https://github.com/ssvlabs/ssv-spec/blob/alan/types/share.go) to `Share` in spec's `alan` version introduced a "native" ValidatorIndex property to the share, leaving our codebase with different ways to access it (ValidatorIndex/BeaconMetadata.Index). 

It's best to have one place with populate and access this information.

### Required change

Make all metadata update code to save the index in to `share`'s `ValidatorIndex` field, and access only using it. `BeaconMetadata.Index` can be removed after that.

PR #1837, opened 2024-11-03
shares-storage: replace GOB encoding with SSZ, deduplicate validator Index field

iurii-ssv:
~~Rebased onto https://github.com/ssvlabs/ssv/pull/1778 (until it's merged - [this diff](https://github.com/ssvlabs/ssv/compare/fix/lockless-share-storage-2...shares-storage-replace-gob-encoding-with-ssz?expand=1) is more representative of the contents of this PR)~~ now rebased onto latest commit from **stage**

This PR additionally closes https://github.com/ssvlabs/ssv/issues/1473 (see [this comment](https://github.com/ssvlabs/ssv/pull/1837#issuecomment-2476248145) for why below), but primarily

this PR closes https://github.com/ssvlabs/ssv/issues/1575 (targeting both `v2` and `genesis` code - hence it should work regardless of whether it's deployed before/after Alan fork - this is because both `v2` and `genesis` use the same code from package `storage`, as far as I can tell, and we don't use `gob` anywhere else)

we could, but I'm assuming we don't want to re-sync everything from Ethereum smart contract just so we can re-populate database with shares in `ssz` format,

The question is - do we want to support a **proper full-blown** rollback back to `gob`,

just so we have some basic "safety" for production (and once we are confident there were no issues with migration - maybe couple of weeks later - we can get rid of `gob` completely and even erase the old shares key-range with a separate easy-to-apply DB migration); but also when I'm gonna roll this PR out to stage - if somebody redeploys something else to the same operator/cluster **after** this change has been running there for a while (and **before** his branch includes my change) - whole cluster will just break for him (stage branch too, until this PR is merged),

**hence the approach I'm taking** in this PR (for now) covers most of rollback scenarios (but not all):
- DB migration creates a copy of old shares key-range (if successful - applies only once), this is kind of necessary waste in terms of space until we delete old `shares` prefix range
- and then, newer code uses just `ssz` while older code uses only `gob` (and works with old shares data that are still present in DB)

the issue with rolling back to `gob` this PR doesn't address is that new shares might have been added during the time we were running this new `ssz` code version - hence for proper rollback to `gob` we'd need to copy those shares over to the old key-range too; and vise versa can also happen if node versions run are `gob`(current)->`ssz`(migration successful)->`gob`(rollback)->(added new shares to DB)->`ssz` - we won't see those newly added shares because they'll end up in old gob-related key-range (while migration intended to move them over if successful - applies only once, and has been previously run to success)

it's unlikely we'll need to "rollback with a delay" once everything has been proven to work on stage (and actually needing 100% rollback-supporting approach in practice) ... but you never know, WDYT ?

_Before merging:_
- [x] test on **stage** (can we add new validator to make sure this use-case works fine ?)
- [x] (assuming we aren't doing [100% rollback-supporting approach](https://github.com/ssvlabs/ssv/pull/1837#issuecomment-2454052940) - if we do, this will be done at the same time the next step below "delete old `shares` prefix range" is done, after this code has been running in prod for a while - both can be added to migration_5) once we are done with testing on stage add a commit to enable commented out `completed(txn)` code in this PR
- [x] ~(after we are 100% `ssz` is working and we are not going back to `gob`) delete old `shares` prefix range in a separate migration (**migration_6** or higher)~ since share data isn't super precious we decided to delete it right away as a part of  **migration_5**


iurii-ssv:
I probably should have asked sooner - why we want SSZ for encoding shares in the first place, it doesn't feel like the simplest option we can go for (see [one](https://github.com/ferranbt/fastssz/issues/180), [two](https://github.com/ferranbt/fastssz/issues/179), [three](https://github.com/ferranbt/fastssz/issues/164#issuecomment-2453405319)) - why not just go with `json` ? It doesn't look like we pass Share(s) around in p2p messages (hence compressing it as much as possible is probably not a concern ... and if it really does consume a lot of DB storage as `json` perhaps there is a simpler way to compress it than using `ssz`).

Anyway, this was my 1st time using `ssz` - just saying might not be the best option if we are to use it more / somewhere else (for future reference, here is a list of all kinds of alternatives I guess - https://github.com/alecthomas/go_serialization_benchmarks).

Another relevant question, I see other entities in `storage` package are using JSON encoding, could you clarify why those didn't use `gob` and whether we'll want to migrate those to `ssz` as well ?

iurii-ssv:
> the issue with rolling back to `gob` this PR doesn't address is that new shares might have been added during the time we were running this new `ssz` code version - hence for proper rollback to `gob` we'd need to copy those shares over to the old key-range too; and vise versa can also happen if node versions run are `gob`(current)->`ssz`(migration successful)->`gob`(rollback)->(added new shares to DB)->`ssz` - we won't see those newly added shares because they'll end up in old gob-related key-range (while migration intended to move them over if successful - applies only once, and has been previously run to success)

Just for future reference (or in case we'll need it), the way to properly solve it would be:

gob `shares` prefix rage will be the source of truth (because this is the only way older code works), new `szz` supporting code needs to
- drop and re-populate `shares_ssz` prefix range every time from `shares` prefix range
- update both `shares` and `shares_ssz` at the same time (same txn) for every update there is to a Share
- in DB migration, only do `complete(txn)` after we are actually done with it and getting rid of `gob` completely (removing a chance to get back to it "properly") - that will finalize migration 5; might as well update DB migration 5 to delete `shares` prefix range too in the same commit

protocol/v2/types/ssvshare.go
```diff
@@ -42,14 +56,39 @@ func (s *SSVShare) BelongsToOperator(operatorID spectypes.OperatorID) bool {
 	})
 }
 
-// HasBeaconMetadata checks whether the BeaconMetadata field is not nil.
+// HasBeaconMetadata checks whether SSVShare has been enriched with respective Beacon metadata.
 func (s *SSVShare) HasBeaconMetadata() bool {
-	return s != nil && s.BeaconMetadata != nil
+	return s != nil && s.Status != eth2apiv1.ValidatorStateUnknown
 }
```
iurii-ssv, line 62: Now that we've gotten rid of `s.BeaconMetadata` I believe `s.Status != eth2apiv1.ValidatorStateUnknown` check here is sufficient to determine that we've got/fetched metadata from beacon node.

eth/eventhandler/handlers.go
```diff
@@ -489,17 +488,13 @@ func (eh *EventHandler) handleValidatorExited(txn basedb.Txn, event *contract.Co
 		return nil, &MalformedEventError{Err: ErrShareBelongsToDifferentOwner}
 	}
 
-	if share.BeaconMetadata == nil {
-		return nil, nil
-	}
-
```
y0sher, line 495: we do need to check whether this share has metadata somehow

y0sher: otherwise validator index below will be 0

migrations/migration_5_gob.go
y0sher, line 1: please consult with @olegshmuelov here, we added a few stuff to the share recently.

protocol/v2/types/ssvshare.go
```diff
@@ -14,20 +14,34 @@ import (
...
+	// Balance is validator (this share belongs to) balance.
+	Balance phase0.Gwei
+	// Status is validator (this share belongs to) state.
+	Status eth2apiv1.ValidatorState
+	// ActivationEpoch is validator (this share belongs to) epoch it activates at.
+	ActivationEpoch phase0.Epoch
+	// OwnerAddress is validator (this share belongs to) owner address.
+	OwnerAddress common.Address
+	// Liquidated is validator (this share belongs to) liquidation status (true or false).
+	Liquidated bool
```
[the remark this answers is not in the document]
y0sher: would love you thoughts though - @iurii-ssv @moshe-blox @MatusKysel @nkryuchkov @olegshmuelov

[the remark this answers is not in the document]
iurii-ssv: I guess collecting those blockchain-based fields into a separate struct makes sense, maybe we can do something like this then (calling it `Metadata` or similar seems a bit too abstract to me, hence `ShareOnChainData`) ?

```go
// SSVShare is a spectypes.Share with extra data that fully describes SSV validator share.
type SSVShare struct {
	spectypes.Share
	ShareOnChainData

	// lastUpdated is used to keep track of share last update time. Note, we don't
	// store this field in DB - it just serves as a helper-indicator for when we might want
	// to update SSVShare data so it doesn't get super stale.
	lastUpdated time.Time

	// committeeID is a cached value for committee ID so we don't recompute it every time.
	committeeID atomic.Pointer[spectypes.CommitteeID]
}

// ShareOnChainData is share-related data pulled from blockchain.
type ShareOnChainData struct {
	// Balance is validator (this share belongs to) balance.
	Balance phase0.Gwei
	// Status is validator (this share belongs to) state.
	Status eth2apiv1.ValidatorState
	// ActivationEpoch is validator (this share belongs to) epoch it activates at.
	ActivationEpoch phase0.Epoch
	// OwnerAddress is validator (this share belongs to) owner address.
	OwnerAddress common.Address
	// Liquidated is validator (this share belongs to) liquidation status (true or false).
	Liquidated bool
}
```

protocol/v2/types/ssvshare.go
iurii-ssv: Also renamed `HasBeaconMetadata` -> `HasOnChainData` now to clarify what Beacon metadata actually means

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
olegshmuelov: `OwnerAddress` reflects the cluster owner, not beacon chain data.
Also the `Liquidated` parameter is not related to beacon chain data but rather reflects the ssv.network cluster status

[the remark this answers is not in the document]
iurii-ssv: > Additionally, we already have the ValidatorMetadata struct for data fetched from the beacon

Right, the problem with using this `ValidatorMetadata` struct as is would be that 
- `Index` duplicates `ValidatorIndex` (means it's a bit confusing to keep both up to date and preferably use `ValidatorIndex` to make it uniform across code-base) - which is what https://github.com/ssvlabs/ssv/issues/1473 is about
- and it also creates an undesirable dependency of "core domain entity" on implementation (meaning if we swap `protocol/v2` package for `protocol/v3` in the future `SSVShare` struct usages will also need to be changed accordingly across entire code-base)

> `OwnerAddress` reflects the cluster owner, not beacon chain data.
Also the `Liquidated` parameter is not related to beacon chain data but rather reflects the ssv.network cluster status

Aren't we pulling/initializing the values of `OwnerAddress` and `Liquidated` based on what we get from Beacon node (when we receive on-chain events) ? Which is why I think it makes sense to group it under `ShareOnChainData` ... because I don't see how it's different from, say, `ActivationEpoch`

btw, `OwnerAddress` is very similar to `FeeRecipientAddress` (which is defined on `spectypes.Share` itself) ... not sure what we should do about that discrepancy, WDYT ?

migrations/migration_5_gob.go
iurii-ssv: @olegshmuelov could you link relevant PR(s) for me to go through (to get familiar with those changes so we can discuss this further) ?

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
olegshmuelov: Regarding `ValidatorIndex`, perhaps we should keep our struct (`ValidatorMetadata`) as is and not rely on `spectypes.Share`, but instead use our own `Share` struct without `ValidatorIndex` in it.
Alternatively, we could populate `spectypes.Share` parameters in the `SSVShare` struct, excluding `ValidatorIndex`.
We should align on this decision as a team.

For `OwnerAddress` and `Liquidated`, these are based on contract interactions with the Execution Layer (EL), specifically from the ssv.network contract address, not the Consensus Layer (CL). In contrast, `ValidatorMetadata` is data fetched from the CL (beacon node).

As for `OwnerAddress` and `FeeRecipientAddress`, they serve different purposes:
- `OwnerAddress` represents the cluster creator’s owner address and acts as a fallback fee recipient if `FeeRecipientAddress` is not set.

btw, we can remove `Balance` from `ValidatorMetadata` as it’s not used at all.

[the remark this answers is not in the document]
nkryuchkov: > would love you thoughts though - @iurii-ssv @moshe-blox @MatusKysel @nkryuchkov @olegshmuelov

I agree that separating those entities is a good idea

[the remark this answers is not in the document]
iurii-ssv: > For OwnerAddress and Liquidated, these are based on contract interactions with the Execution Layer (EL), specifically from the ssv.network contract address, not the Consensus Layer (CL). In contrast, ValidatorMetadata is data fetched from the CL (beacon node).

To sum up, there are 3 data types/sources `SSVShare` struct combines:
- miscellaneous own data (internal to operator)
- data pulled from Consensus Layer (CL)
- data pulled from Execution Layer (EL)

so, I guess we want to combine **both** CL and EL into a single struct just so it's a logical subgroup inside of `SSVShare` (calling it something like `ShareOnChainData` to make it clear) ? Otherwise why group CL-related fields together but not EL-related fields ?

Personally I think re-using `ValidatorMetadata` in any way is quite misleading/unclear because 
- it only covers CL-data part
- the name itself doesn't tell us much
- like I mentioned above, it creates unnecessary dependency on implementation (maybe not a big deal - but you never know when it'll come to haunt us)

@y0sher would be nice to hear what you had in mind by "keep the data that is collected from BN as metadata in a separate child struct (like before)"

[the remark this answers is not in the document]
iurii-ssv: > I think it'll work better to put all the data the changes into a struct container. and if some of it is changed from different source (like metadata vs el events)

Grouping the "data that changes" does make sense, but the **definition** of what "change" means is a bit slippery, namely we update a bunch of things based on the data we pull from Beacon node:
```go
			share.ValidatorIndex = metadata.Index
			share.Balance = metadata.Balance
			share.Status = metadata.Status
			share.ActivationEpoch = metadata.ActivationEpoch
```

and I think we expect for `ValidatorIndex` and `ActivationEpoch` to never change, while `Balance` and `Status` probably change over time (and I think `FeeRecipientAddress` can also be overwritten with whatever new value EL event supplies).

So I'm struggling to group these fields in any meaningful way (and this is why I thought un-grouping makes most sense, at least until we can get rid of `spectypes.Share`)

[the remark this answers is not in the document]
y0sher: Since now with `ValidatorIndex` embedded into the specshare, it makes more sense to flatten the struct since there are not a lot more meaningful members that change left in metadata internal struct. Unless we want to group `ValidatorIndex` and then we need to break from the spec share type. 
So for now we decided to leave it flat until we do a more comprehensive thinking about the share struct and metadata and their best placing.

y0sher (approved):
Interesting change. I agree we need to clarify next time what is the rationale behind such change and it'll help define the scope and make it easier to do decisions.

registry/storage/shares.go
```diff
@@ -65,64 +67,62 @@ type sharesStorage struct {
...
-	OperatorID            spectypes.OperatorID
-	ValidatorPubKey       []byte             `ssz-size:"48"`
-	SharePubKey           []byte             `ssz-size:"48"`
+// storageShare represents a Share stored in DB. SSZ encodings generator has some limitations
+// in terms of what types it supports, hence we define a bunch of own types here to satisfy it,
+// see more on this here: https://github.com/ferranbt/fastssz/issues/179#issuecomment-2454371820
+type storageShare struct {
+	ValidatorIndex        uint64
+	ValidatorPubKey       []byte             `ssz-max:"48"`
+	SharePubKey           []byte             `ssz-max:"48"`
```
y0sher, line 78: can you explain why its needed? the size is actually fixed usually and not a max..

y0sher (approved):
looks good! one comment

protocol/v2/types/ssvshare.go
```diff
@@ -14,20 +14,34 @@ import (
...
 )
 
-// SSVShare is a combination of spectypes.Share and its Metadata.
+// SSVShare is a spectypes.Share with extra data that fully describes SSV validator share.
 type SSVShare struct {
 	spectypes.Share
-	Metadata
 
+	// Balance is validator (this share belongs to) balance.
+	Balance phase0.Gwei
```
olegshmuelov, line 29: We don’t need the Balance field in SSVShare as it’s not used in any logic or functionality.

registry/storage/shares.go
moshe-blox: we should follow the spec here, and this seems to be in line

migrations/migration_5_gob.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("decode storageShareGOB: %w", err)
+	}
+	s.Quorum, s.PartialQuorum = types.ComputeQuorumAndPartialQuorum(uint64(len(s.Committee)))
+	return nil
+}
+
+// Share represents a storage share.
+// The better name of the struct is storageShareGOB,
+// but we keep the name Share to avoid conflicts with gob encoding.
+type Share struct {
```
moshe-blox, line 39: if possible its better to not export on `migrations` package

protocol/v2/types/ssvshare.go
```diff
@@ -110,6 +149,14 @@ func (s *SSVShare) Quorum() uint64 {
 	return q
 }
 
+func (s *SSVShare) LastUpdated() time.Time {
+	return s.lastUpdated
+}
+
+func (s *SSVShare) SetLastUpdated(t time.Time) {
+	s.lastUpdated = t
+}
```
moshe-blox, line 158: this is still specifically for metadata, so should probably be named BeaconMetadataLastUpdated to prevent confusion

its not a fundamental method which is used a lot, its just an optimization for the metadata fetcher to not have to create a new map for all 50k+ shares to track last update times

moshe-blox:
great work 💪

i think we should have a unit test for the migration or perhaps at least on the conversion funcs, to make sure we're not losing/mutating any field by accident during migration

for testing the migration, i have an idea:
1. duplicate old `registry/storage` package into `register/storage/gob` or a subfolder in `migrations`
2. create the old gob `sharesStorage` with an in-memory database
3. save some slice of shares
5. `Migrations{migration_5}.Run()`
6. create the new ssz `sharesStorage` with the same database
7. load all shares and compare to the originally saved shares

unless we expect non-test-related code to change, this can be done in parallel as we progress with merging and QA'ing this PR

however we never tested migrations before, so pls lmk if you have another idea

other than that, i would've preferred separation like before, except for `Liquidated` which as you've noted is unrelated:
```go
type SSVShare struct {
    spectypes.Share // Contract data, with the unfortunate exception of ValidatorIndex however it's only set once
    Liquidated bool // Contract data, but the spec is missing it and its unrelated to Beacon
    BeaconMetadata  // Beacon data only
}
```
i think it's a bit less confusing to work with given the metadata and contract are unrelated and update in parallel, plus eventually we'd likely store metadata in separate database collection (to optimize updates & prevent race conditions)

however i suspect it would be a significant refactor, and i'm not strongly convicted on this (would love to hear counter-arguments), so perhaps this should be postponed to a future PR (maybe together with the metadata database separation)

registry/storage/shares.go
y0sher: current spec seems to use `ssz-size` - 
https://github.com/ssvlabs/ssv-spec/blob/main/types/share.go#L8C1-L9C54

iurii-ssv: > can you explain why its needed? the size is actually fixed usually and not a max..

I think I saw various tests failing with exact `size` (as opposed to `max`) - so I thought maybe it is meant to support pubkeys of different format (with size lower than 48 as well),

eg:
```
=== RUN   TestSubmitProposal
2025-01-20T10:36:10.106079Z	DEBUG	logger is ready	{"level": "debug", "encoder": "capital", "format": "console", "file_options": null}
2025-01-20T10:36:10.106420Z	DEBUG	logger is ready	{"level": "debug", "encoder": "capital", "format": "console", "file_options": null}
    controller_test.go:157: 
        	Error Trace:	/Users/iurii/work/ssv/operator/fee_recipient/controller_test.go:157
        	            				/Users/iurii/work/ssv/operator/fee_recipient/controller_test.go:45
        	            	failed to serialize share: marshal ssz: storageShare.SharePubKey (bytes array does not have the correct length): expected 48 and 0 found
        	Test:       	TestSubmitProposal
--- FAIL: TestSubmitProposal (0.00s)
```

but if you are saying the size always must be 48 I guess we simply need to adjust every test to have those pubkeys initialized

iurii-ssv:
> i would've preferred separation like before, except for `Liquidated` which as you've noted is unrelated:
> 
> ```go
> type SSVShare struct {
>     spectypes.Share // Contract data, with the unfortunate exception of ValidatorIndex however it's only set once
>     Liquidated bool // Contract data, but the spec is missing it and its unrelated to Beacon
>     BeaconMetadata  // Beacon data only
> }
> ```
> 
> i think it's a bit less confusing to work with given the metadata and contract are unrelated and update in parallel, plus eventually we'd likely store metadata in separate database collection (to optimize updates & prevent race conditions)
> 
> however i suspect it would be a significant refactor, and i'm not strongly convicted on this (would love to hear counter-arguments), so perhaps this should be postponed to a future PR (maybe together with the metadata database separation)

I don't really see any obvious way to group `Share` fields that would be strongly preferable over all other ones we can go with, the one you've suggested is OK (but I don't think it 100% accurately reflects what **actually** happens - the way it's currently implemented, and I haven't changed it - `ValidatorIndex` gets updated every time `UpdateValidatorsMetadata` is called, meaning even though it shouldn't change - it might, unless we implement additional measures to prevent it),

and I'm not sure grouping by "where data comes from" (especially with `ValidatorIndex` coming from Beacon metadata) is better than grouping by "whether data is mutable or not" (I would prefer this one),

which is why to me it seems the most "straightforward" way to go about it is to flatten `Share` struct as much as possible, possibly even removing the dependency on `spectypes.Share` (essentially leaving it up to code reader to find out on his own how each of those fields changes by simply following the code, without "giving any hints by grouping" because they are inaccurate),

at least for now, if we really want to group `Share` fields somehow - it would probably be better to separate it in PR on top of this one (since it's pretty large as it is)

protocol/v2/types/ssvshare.go
iurii-ssv: Changed to `BeaconMetadataLastUpdated` in https://github.com/ssvlabs/ssv/pull/1837/commits/04d5c63de94d91db81c758a04622434e56d0d9f7 - I thought making it "generic" could be useful in the future - but I guess we can change it later if necessary.

migrations/migration_5_gob.go
iurii-ssv: Originally the idea was to copy whatever structures we have from "the original code" and adjust as necessary (keeping it as close to originals as possible so it's easy to drop-in replace it),

but now since all adjustments are made we can make it package-private - done in https://github.com/ssvlabs/ssv/pull/1837/commits/04d5c63de94d91db81c758a04622434e56d0d9f7

protocol/v2/types/ssvshare.go
iurii-ssv: Removed in https://github.com/ssvlabs/ssv/pull/1837/commits/04d5c63de94d91db81c758a04622434e56d0d9f7

migrations/migration_5_gob.go
iurii-ssv: @moshe-blox I guess we cannot change this after all since gob complains it cannot decode data fetched from DB (the error goes away once I revert back to using public fields https://github.com/ssvlabs/ssv/pull/1837/commits/0219b21e8a193815d471df48dc63df242a984893):
```
{
  "level": "\u001b[31mFATAL\u001b[0m",
  "time": "2025-01-20T14:35:08.440521Z",
  "msg": "could not setup db",
  "error": "failed to run migrations: migration \"migration_5_change_share_format_from_gob_to_ssz\" failed: GetAll: decode gob share: decode storageShareGOB: gob: type mismatch: no fields matched compiling decoder for storageShareGOB",
  "errorVerbose": "GetAll: decode gob share: decode storageShareGOB: gob: type mismatch: no fields matched compiling decoder for storageShareGOB\nmigration \"migration_5_change_share_format_from_gob_to_ssz\" failed\ngithub.com/ssvlabs/ssv/migrations.Migrations.Run\n\t/go/src/github.com/ssvlabs/ssv/migrations/migrations.go:98\ngithub.com/ssvlabs/ssv/migrations.Run\n\t/go/src/github.com/ssvlabs/ssv/migrations/migrations.go:33\ngithub.com/ssvlabs/ssv/cli/operator.setupDB\n\t/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:515\ngithub.com/ssvlabs/ssv/cli/operator.init.func2\n\t/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:136\ngithub.com/spf13/cobra.(*Command).execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117\ngithub.com/spf13/cobra.(*Command).Execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041\ngithub.com/ssvlabs/ssv/cli.Execute\n\t/go/src/github.com/ssvlabs/ssv/cli/cli.go:27\nmain.main\n\t/go/src/github.com/ssvlabs/ssv/cmd/ssvnode/main.go:20\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1695\nfailed to run migrations\ngithub.com/ssvlabs/ssv/cli/operator.setupDB\n\t/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:517\ngithub.com/ssvlabs/ssv/cli/operator.init.func2\n\t/go/src/github.com/ssvlabs/ssv/cli/operator/node.go:136\ngithub.com/spf13/cobra.(*Command).execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117\ngithub.com/spf13/cobra.(*Command).Execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041\ngithub.com/ssvlabs/ssv/cli.Execute\n\t/go/src/github.com/ssvlabs/ssv/cli/cli.go:27\nmain.main\n\t/go/src/github.com/ssvlabs/ssv/cmd/ssvnode/main.go:20\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1695"
}
```

message/validation/validation.go
```diff
@@ -232,38 +231,38 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
 		}, nil
 	}
 
-	validator, exists := mv.validatorStore.Validator(msgID.GetDutyExecutorID())
+	share, exists := mv.validatorStore.Validator(msgID.GetDutyExecutorID())
```
oleg-ssvlabs, line 234: maybe rename [this](https://github.com/ssvlabs/ssv/blob/0219b21e8a193815d471df48dc63df242a984893/registry/storage/validatorstore.go#L18) method to `Share(...) (...)`, instead of `Validator(...)(...)`?

message/validation/validation.go
```diff
@@ -232,38 +231,38 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
...
 	}
 
 	// Rule: If validator is liquidated
-	if validator.Liquidated {
+	if share.Liquidated {
 		return CommitteeInfo{}, ErrValidatorLiquidated
 	}
 
-	if validator.BeaconMetadata == nil {
+	if !share.HasBeaconMetadata() {
```
oleg-ssvlabs, line 246: nice! I saw you went through the entire codebase and replaced all nil checks with this method call :) 🥇

iurii-ssv: yep, that's a bit more explicit :)

message/validation/validation.go
iurii-ssv: I would be all for it, but there are a bunch of obstacles to make it happen:
- this isn't the only place where similar renaming makes sense (just as an [example](https://github.com/ssvlabs/ssv/blob/6cee7d14f45da8255ad1c8a79e2af7a052b6ddff/operator/duties/scheduler.go#L68-L68))
- this PR is already pretty large and this kind of refactoring is better done in a separate one (on top of it)
- **validator** and **share** are sort of interchangeable entities all across code-base, so I doubt that we'll be able to clearly separate the two - so might not even worth trying

protocol/v2/types/ssvshare.go
```diff
@@ -48,14 +60,39 @@ func (s *SSVShare) BelongsToOperator(operatorID spectypes.OperatorID) bool {
...
+func (s *SSVShare) Activated() bool {
+	return s.Status.HasActivated() || s.Status.IsActive() || s.Status.IsAttesting()
+}
+
+// IsActive returns true if the validator is currently active. Cant be other state
+func (s *SSVShare) IsActive() bool {
+	return s.Status == eth2apiv1.ValidatorStateActiveOngoing
+}
+
+// Exiting returns true if the validator is existing or exited
```
oleg-ssvlabs, line 88: probably typo here `is existing` -> `is exiting`

protocol/v2/types/ssvshare.go
```diff
@@ -48,14 +60,39 @@ func (s *SSVShare) BelongsToOperator(operatorID spectypes.OperatorID) bool {
...
+func (s *SSVShare) Exiting() bool {
+	return s.Status.IsExited() || s.Status.HasExited()
+}
+
+// Slashed returns true if the validator is existing or exited due to slashing
```
oleg-ssvlabs, line 93: probably typo here `is existing` -> `is exiting`

protocol/v2/types/ssvshare.go
```diff
@@ -110,6 +147,14 @@ func (s *SSVShare) Quorum() uint64 {
 	return q
 }
 
+func (s *SSVShare) BeaconMetadataLastUpdated() time.Time {
+	return s.beaconMetadataLastUpdated
+}
+
+func (s *SSVShare) SetBeaconMetadataLastUpdated(t time.Time) {
```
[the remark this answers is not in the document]
iurii-ssv: Good question, I don't see any particular reason to use getter/setter here indeed, changing it in https://github.com/ssvlabs/ssv/pull/1837/commits/bce5cd75589a214d663646e158f4c0c6426f3cdd

moshe-blox (approved):

Issue #1578, opened 2024-08-07
Flaky p2p tests


nkryuchkov:
for commit 122441979f58b86d950fd755ced57fb51c23fedf:

#### Case 1
```
        	Error Trace:	/Users/bloxmac/GolandProjects/ssv/network/p2p/p2p_validation_test.go:270
        	Error:      	Not equal: 
        	            	expected: 3
        	            	actual  : 0
        	Test:       	TestP2pNetwork_MessageValidation
        	Messages:   	node 0
```

#### Case 2
```
--- FAIL: TestP2pNetwork_MessageValidation (13.69s)
    p2p_validation_test.go:283: 
        	Error Trace:	/home/runner/work/ssv/ssv/network/p2p/p2p_validation_test.go:283
        	Error:      	invalid order
        	Test:       	TestP2pNetwork_MessageValidation
        	Messages:   	node 0, peers [{1 0} {2 -4839.971681614018} {3 -6759.6601793682175}]
FAIL
```

### `TestP2pNetwork_SubscribeBroadcast`:
```
    p2p_test.go:132: 
        	Error Trace:	/Users/bloxmac/GolandProjects/ssv/network/p2p/p2p_test.go:132
        	Error:      	"0" is not greater than or equal to "2"
        	Test:       	TestP2pNetwork_SubscribeBroadcast
        	Messages:   	router 0
    p2p_test.go:132: 
        	Error Trace:	/Users/bloxmac/GolandProjects/ssv/network/p2p/p2p_test.go:132
        	Error:      	"0" is not greater than or equal to "2"
        	Test:       	TestP2pNetwork_SubscribeBroadcast
        	Messages:   	router 3
```

nkryuchkov:
`TestP2pNetwork_MessageValidation` case 1 should have been closed by https://github.com/ssvlabs/ssv/pull/1838

nkryuchkov:
It hasn't been addressed for quite a while, so I'd keep it closed until we see it again. Also, some cases might have been fixed by PRs that referenced this issue

PR #1838, opened 2024-11-04
tests: p2p validation - use correct mutex to read roleBroadcasts changes

iurii-ssv:
I've been observing flaky test running `make docker-unit-test` (eg. https://github.com/ssvlabs/ssv/actions/runs/11482239405/job/31954803590, it also seems to be the Case 1 described in https://github.com/ssvlabs/ssv/issues/1578):
```
--- FAIL: TestP2pNetwork_MessageValidation (11.50s)
    p2p_validation_test.go:279:
        	Error Trace:	/go/src/github.com/ssvlabs/ssv/network/p2p/p2p_validation_test.go:279
        	Error:      	Not equal:
        	            	expected: 3
        	            	actual  : 2
        	Test:       	TestP2pNetwork_MessageValidation
        	Messages:   	node 3
FAIL
coverage: 12.7% of statements
FAIL	github.com/ssvlabs/ssv/network/p2p	15.897s
```
This occasional failure seems to be due to the way `sync.Map` is being iterated with `Range` method (namely there is no way to get a consistent view for the whole map). In this PR I've simply changed the way access to this map is handled (instead of per-element update/read - map as a whole gets updated at once) and I no longer see this test failing (after running it locally a bunch of times with `make docker-unit-test`).

### DIFF
--- a/network/p2p/p2p_validation_test.go
+++ b/network/p2p/p2p_validation_test.go
@@ -25,8 +25,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"golang.org/x/exp/slices"
 
-	"github.com/ssvlabs/ssv/utils/hashmap"
-
 	"github.com/ssvlabs/ssv/message/validation"
 	beaconprotocol "github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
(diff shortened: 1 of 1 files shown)


moshe-blox (approved):
LGTM, let's hope it won't fail again 😅

PR #1839, opened 2024-11-05
peers: fix subnets index removal logic

nkryuchkov:
The logic is incorrect for `i != 0`, the peer is not removed.

This change is copied from https://github.com/ssvlabs/ssv/pull/1813 as it seems more important

### DIFF
--- a/network/peers/subnets_test.go
+++ b/network/peers/subnets_test.go
@@ -140,9 +140,7 @@ func TestUpdatePeerSubnets_Removal(t *testing.T) {
 	}
 
 	getSubnet := func(t *testing.T, subnetHex string) records.Subnets {
-		if len(subnetHex) != 32 {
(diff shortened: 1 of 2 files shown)


network/peers/subnets_test.go
```diff
@@ -120,3 +122,115 @@ func TestSubnetScore(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdatePeerSubnets_Removal(t *testing.T) {
```
nkryuchkov, line 126: This is likely not the final version of the tests, I'll be working on them on https://github.com/ssvlabs/ssv/pull/1813. I just wanted this PR to have some PoC of the fix

iurii-ssv (approved):
LGTM as long as tests cover edge cases (the code is quite involved and hence deserves thorough testing I think)

moshe-blox (approved):
good catch 💪

PR #1840, opened 2024-11-06
message-validation: graceful validation locks management

iurii-ssv:
I've been browsing message validation code and noticed there is this ever-growing map (as is mentioned in TODO-comments also) - decided to fix it while I have the relevant context (as well as document our assumptions on this).


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ssvlabs/ssv
 
-go 1.22
+go 1.23.1
```
iurii-ssv, line 3: This is pretty much a requirement to use `tailscale.com v1.76.3`

iurii-ssv: I've added [a commit on top](https://github.com/ssvlabs/ssv/pull/1840/commits/5d14e31ff5a3f8d7621a09b3fae17af5106a5e30) to drop it down to `tailscale.com v1.72.0` so we can keep using `go 1.22` like this (as I've done in https://github.com/ssvlabs/ssv/pull/1798#discussion_r1809543763):

```
go 1.22.0

toolchain go1.23.1
```

nkryuchkov (approved):
LGTM

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ssvlabs/ssv
 
-go 1.22
+go 1.22.0
+
+toolchain go1.23.1
```
nkryuchkov, line 5: Is there any reason why we need to have 1.23.1 patch version here and not 1.23.3 or 1.23.0?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ssvlabs/ssv
 
-go 1.22
+go 1.22.0
```
nkryuchkov, line 3: Why do we need to specify patch version here?

iurii-ssv: This is pretty much the result of me running `go mod tidy` - I have no idea why it added patch version, my guess would be it just makes it a bit more explicit what `go 1.22` actually means (it means the **min** version of golang needs to be 1.22.0)

go.mod
iurii-ssv: Not really, I don't think there is a specific reason.

It looks like it's just `go mod` behavior to add `toolchain` (if there is none, but `go 1.22` is specified and not `go 1.22.0`) with version that matches your local golang version (which is 1.23.1 for me),

this kind of behavior is a bit weird to me too, turns out I can just do `go 1.22.0` and omit `toolchain go1.23.1` completely (`go mod tidy` doesn't add it anymore when patch version is specified) - so I've done that to simplify things - https://github.com/ssvlabs/ssv/pull/1840/commits/de9e2f2fb52df6bc185ae207181182c0b6d55126

nkryuchkov (approved):

message/validation/validation.go
```diff
@@ -74,6 +85,9 @@ func New(
 		opt(mv)
 	}
 
+	// Start automatic expired item deletion for validationLockCache.
+	go mv.validationLockCache.Start()
```
oleg-ssvlabs, line 89: I read this method description and it made me wonder - if `.Start()` is not explicitly called, does it mean that expired items in the cache will never be cleaned up?

oleg-ssvlabs (approved):
> perhaps we want to create a helper type that combines `ttlcache.Cache` and `singleflight.Group`

I think it's a great idea. Could be implemented either in this or a separate PR

message/validation/validation.go
iurii-ssv: Yes, you got that right, it's also a bit weird to me that it's not automatically called in **constructor** (but I haven't looked into why that is)

PR #1845, opened 2024-11-08
fix: remove 'convert' package

anatolie-ssv:
Replace 'convert.RunnerRole' with BeaconRole from spectypes.

Closes: #1810


ibft/storage/store.go
```diff
@@ -138,12 +137,12 @@ func (i *ibftStorage) getParticipants(txn basedb.ReadWriter, identifier convert.
...
 }
 
-func (i *ibftStorage) saveParticipants(txn basedb.ReadWriter, identifier convert.MessageID, slot phase0.Slot, operators []spectypes.OperatorID) error {
+func (i *participantStorage) saveParticipants(txn basedb.ReadWriter, role spectypes.BeaconRole, pk spectypes.ValidatorPK, slot phase0.Slot, operators []spectypes.OperatorID) error {
 	bytes, err := encodeOperators(operators)
 	if err != nil {
 		return fmt.Errorf("encode operators: %w", err)
 	}
-	if err := i.save(txn, bytes, participantsKey, identifier[:], uInt64ToByteSlice(uint64(slot))); err != nil {
+	if err := i.save(txn, bytes, participantsKey, pk[:], uInt64ToByteSlice(uint64(role)), uInt64ToByteSlice(uint64(slot))); err != nil {
```
anatolie-ssv, line 145: @moshe-blox we use `pk[:]` in combination with `role` as key here, instead of the previous `MessageID` based identifier, wdyt?

moshe-blox: @anatolie-ssv this looks good, as long as we don't care to preserve the same keys as before (which is likely the case now)

api/handlers/exporter.go
```diff
@@ -59,35 +57,34 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
 				continue
 			}
 
+			var participantsList []qbftstorage.Participation
+
+			for _, p := range participantsRange {
+				participantsList = append(participantsList, qbftstorage.Participation{
+					ParticipantsRangeEntry: p,
+
+					DomainType: e.DomainType,
```
moshe-blox, line 82: domain type can change according to the fork

it would be better to pass the NetworkConfig to this handler rather than the domain type, so it can call the DomainType method directly

api/handlers/exporter.go
```diff
@@ -59,35 +57,34 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
 			if err != nil {
 				return api.Error(fmt.Errorf("error getting participants: %w", err))
 			}
 
-			if len(participantsList) == 0 {
+			if len(participantsRange) == 0 {
 				continue
 			}
 
+			var participantsList []qbftstorage.Participation
```
moshe-blox, line 76:
```suggestion
			participantsList := make([]qbftstorage.Participation, 0, len(participantsRange))
```
to save some memory allocations

protocol/v2/qbft/storage/participant_store.go
```diff
@@ -7,9 +7,21 @@ import (
...
+
+	DomainType spectypes.DomainType
+	Role       spectypes.BeaconRole
+	PK         spectypes.ValidatorPK
+}
+
+func (p *Participation) MsgID() spectypes.MessageID {
+	// nolint: gosec
+	return spectypes.NewMsgID(p.DomainType, p.PK[:], spectypes.RunnerRole(p.Role))
+}
```
moshe-blox, line 23: i forgot the new MessageID has a different ordering than the old one (i think role and pubkey switched places)

in order to behave backwards-compatibility for exporter users, i think that instead, we should copy the content of the NewMessageID/NewMsgID function from the genesis spec into this func here, but return the bytes array directly rather than the MessageID type

you can find the correct genesis spec repo & commit by following this import in go.mod:

```
github.com/ssvlabs/ssv-spec-pre-cc v0.0.0-20240725052506-c48532da6a63
```

(pre-cc just means pre-committee-consensus which i know is a weird name but going to be deleted soon)

plus, i think we should rename MsgID method here to legacyMsgID just to make it clear its for API backwards-compatibility and not for other usage (a comment would also help)

exporter/api/query_handlers.go
```diff
@@ -99,3 +97,17 @@ func HandleParticipantsQuery(logger *zap.Logger, qbftStorage *storage.QBFTStores
 	}
 	nm.Msg = res
 }
+
+func toParticipations(domain spectypes.DomainType, role spectypes.BeaconRole, pk spectypes.ValidatorPK, ee []qbftstorage.ParticipantsRangeEntry) (out []qbftstorage.Participation) {
```
moshe-blox, line 101: i'd add a make() here as well to reduce allocs

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -161,53 +147,51 @@ func (ncv *CommitteeObserver) ProcessMessage(msg *queue.SSVMessage) error {
 			}
 
 			logger.Info("✅ saved participants",
-				zap.String("converted_role", beaconRole.ToBeaconRole()),
+				zap.String("role", beaconRole.String()),
 				zap.Uint64("validator_index", uint64(key.ValidatorIndex)),
 				fields.Validator(validator.ValidatorPubKey[:]),
 				zap.String("signers", strings.Join(operatorIDs, ", ")),
-				zap.String("msg_id", hex.EncodeToString(msgID[:])),
+				// zap.String("msg_id", hex.EncodeToString(msgID[:])), TODO replace with?
```
moshe-blox, line 154: empty line :D

role & pubkey are enough

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -161,53 +147,51 @@ func (ncv *CommitteeObserver) ProcessMessage(msg *queue.SSVMessage) error {
...
-
-// nonCommitteeInstanceContainerCapacity returns the capacity of InstanceContainer for non-committee validators
-func nonCommitteeInstanceContainerCapacity(fullNode bool) int {
-	if fullNode {
-		// Helps full nodes reduce
-		return 2
-	}
-	return 1
+	// nolint: gosec
+	return []spectypes.BeaconRole{spectypes.BeaconRole(msg.MsgID.GetRoleType())}
```
moshe-blox, line 194: this would no longer work because BeaconRole and RunnerRole are incompatible

a switch should be done instead to map RunnerRole to BeaconRole, where the only special case would be the RoleCommittee which is already implemented above

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -161,53 +147,65 @@ func (ncv *CommitteeObserver) ProcessMessage(msg *queue.SSVMessage) error {
...
-}
 
-// nonCommitteeInstanceContainerCapacity returns the capacity of InstanceContainer for non-committee validators
-func nonCommitteeInstanceContainerCapacity(fullNode bool) int {
-	if fullNode {
-		// Helps full nodes reduce
-		return 2
+	switch msg.MsgID.GetRoleType() {
+	case spectypes.RoleCommittee:
+		// taken care above
```
moshe-blox, line 195: we can just move the code above here instead to reduce duplication/confusion

protocol/v2/qbft/spectest/msg_processing_type.go
```diff
@@ -34,7 +33,7 @@ func RunMsgProcessing(t *testing.T, test *spectests.MsgProcessingSpecTest) {
 	ks := spectestingutils.KeySetForCommitteeMember(test.Pre.State.CommitteeMember)
 	signer := spectestingutils.NewOperatorSigner(ks, 1)
 	pre := instance.NewInstance(
-		qbfttesting.TestingConfig(logger, ks, casts.RunnerRoleToConvertRole(msgId.GetRoleType())), // #nosec G104
+		qbfttesting.TestingConfig(logger, ks, spectypes.BeaconRole(msgId.GetRoleType())), // nolint: gosec
```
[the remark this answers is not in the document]
anatolie-ssv: it's strange because it doesn't complain when using `nolint` but does so when I reverted back to using `#nosec`:
https://github.com/ssvlabs/ssv/actions/runs/11877412486/job/33096853887?pr=1845#step:6:27
```
Error: protocol/v2/qbft/spectest/msg_processing_type.go:36:61: G115: integer overflow conversion int32 -> uint64 (gosec)
		qbfttesting.TestingConfig(logger, ks, spectypes.BeaconRole(msgId.GetRoleType())), // #nosec G104
```

exporter/api/query_handlers.go
```diff
@@ -66,31 +65,30 @@ func HandleParticipantsQuery(logger *zap.Logger, qbftStorage *storage.QBFTStores
...
+		res.Data = []string{"internal error - role storage doesn't exist", role.String()}
 		nm.Msg = res
 		return
 	}
 
-	msgID := convert.NewMsgID(domain, pkRaw, runnerRole)
 	from := phase0.Slot(nm.Msg.Filter.From)
 	to := phase0.Slot(nm.Msg.Filter.To)
-	participantsList, err := roleStorage.GetParticipantsInRange(msgID, from, to)
+	participantsList, err := roleStorage.GetParticipantsInRange(role, spectypes.ValidatorPK(pkRaw), from, to)
```
moshe-blox, line 85: need length validation above in pkRaw otherwise this conversion may cause issues

anatolie-ssv: added the check

ibft/storage/store.go
```diff
@@ -38,46 +41,137 @@ func init() {
...
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("collect keys of stale slots: %w", err)
+	}
+
+	for _, id := range keySet {
+		if err := i.db.Using(tx).Delete(append(prefix, id...), nil); err != nil {
+			return fmt.Errorf("remove slot: %w", err)
```
anatolie-ssv, line 109: couldn't use `DropPrefix` here because it doesn't support transactions; and 4 goroutines dropping prefix concomitantly caused database errors.

api/handlers/exporter.go
```diff
@@ -6,18 +6,17 @@ import (
...
 	ibftstorage "github.com/ssvlabs/ssv/ibft/storage"
-	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
-	"github.com/ssvlabs/ssv/utils/casts"
 )
 
 type Exporter struct {
-	DomainType spectypes.DomainType
-	QBFTStores *ibftstorage.QBFTStores
+	DomainType        spectypes.DomainType
+	NetworkConfig     networkconfig.NetworkConfig
```
y0sher, line 18: since `DomainType` is included in `NetworkConfig`, its probably not needed as a separate member of the struct.

ibft/storage/store.go
```diff
@@ -13,15 +15,16 @@ import (
...
+	"github.com/ssvlabs/ssv/operator/slotticker"
 	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
 const (
 	highestInstanceKey = "highest_instance"
 	instanceKey        = "instance"
-	participantsKey    = "participants"
+	participantsKey    = "pt"
```
y0sher, line 27: this change isn't backward compatible, it means we have to do one of the following:
1. need to do migration 
2. resync the db
3.  search by the old key if the search is from older slot

or we just leave the old key as is.

ibft/storage/store.go
```diff
@@ -38,46 +41,151 @@ func init() {
...
-func New(db basedb.Database, prefix string) qbftstorage.QBFTStore {
-	return &ibftStorage{
-		prefix: []byte(prefix),
-		db:     db,
+// New create new participant store
+func New(db basedb.Database, prefix spectypes.BeaconRole) qbftstorage.ParticipantStore {
+	role := byte(prefix & 0xff)
+	return &participantStorage{
+		prefix:    []byte{role},
+		oldPrefix: prefix.String(),
```
y0sher, line 58: the change in prefix has the same meaning as I mentioned above for the database key change.

ibft/storage/store.go
```diff
@@ -38,46 +41,151 @@ func init() {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.Next():
+			threashold := ticker.Slot() - phase0.Slot(retain) - 1 // #nosec G115
+			count, err := i.removeSlotAt(threashold)
+			if err != nil {
+				logger.Error("remove slot at", zap.String("store", i.ID()), fields.Slot(threashold))
+			}
```
y0sher, line 87: I don't think that makes any sense, don't we want to keep all history in exporter? even if we want to gc or remove old slots it must be really far in the past. it probably makes more sense to come up with some kinds of compaction format than removing the data. 

@moshe-blox @MatusKysel @liorrutenberg 
do you guys think it makes sense to say exporter supports storing data up until X slots in the past?

ibft/storage/store.go
anatolie-ssv: yes, there's a migration I just haven't committed it because I wanted to finish testing of the main functionalities without dropping old keys, will add it at a later point.

cli/operator/node.go
```diff
@@ -784,3 +788,30 @@ func startMetricsHandler(logger *zap.Logger, db basedb.Database, port int, enabl
 		logger.Panic("failed to serve metrics", zap.Error(err))
 	}
 }
+
+func initSlotCleanup(ctx context.Context, logger *zap.Logger, stores *ibftstorage.ParticipantStores, slotTickerProvider slotticker.Provider, retain int) {
+	var wg sync.WaitGroup
+
+	ticker := slotTickerProvider()
+	<-ticker.Next()
```
y0sher, line 796: this should be easier to get using `networkConfig.Beacon.EstimatedCurrentSlot` also non blocking. I don't see a point waiting a slot to calculate the threshold.

cli/operator/node.go
```diff
@@ -784,3 +789,28 @@ func startMetricsHandler(logger *zap.Logger, db basedb.Database, port int, enabl
 		logger.Panic("failed to serve metrics", zap.Error(err))
 	}
 }
+
+func initSlotCleanup(ctx context.Context, logger *zap.Logger, stores *ibftstorage.ParticipantStores, slotTickerProvider slotticker.Provider, slot phase0.Slot, retain int) {
```
y0sher, line 793: retain can probably be uint64 since we never have a negative value

cli/operator/node.go
```diff
@@ -784,3 +789,28 @@ func startMetricsHandler(logger *zap.Logger, db basedb.Database, port int, enabl
...
+	var wg sync.WaitGroup
+
+	// #nosec G115
+	threashold := slot - phase0.Slot(retain)
```
y0sher, line 797: if retain was uint64 or even phase.Slot we wouldn't need conversion and #nosec

cli/operator/node.go
```diff
@@ -268,28 +270,31 @@ var StartNodeCmd = &cobra.Command{
...
 		storageMap := ibftstorage.NewStores()
 
 		for _, storageRole := range storageRoles {
-			storageMap.Add(storageRole, ibftstorage.New(cfg.SSVOptions.ValidatorOptions.DB, storageRole.String()))
+			s := ibftstorage.New(cfg.SSVOptions.ValidatorOptions.DB, storageRole)
+			storageMap.Add(storageRole, s)
 		}
 
+		threshold := cfg.SSVOptions.Network.Beacon.EstimatedCurrentSlot()
+		initSlotCleanup(cmd.Context(), logger, storageMap, slotTickerProvider, threshold, cfg.DBOptions.RetainedSlotCount)
```
moshe-blox, line 296: lets make sure to do this only if Exporter is true

protocol/v2/qbft/storage/participant_store.go
```diff
@@ -0,0 +1,66 @@
...
+	GetAllParticipantsInRange(from, to phase0.Slot) ([]ParticipantsRangeEntry, error)
+
+	// GetParticipantsInRange returns participants in quorum for the given slot range and validator public key.
+	GetParticipantsInRange(pk spectypes.ValidatorPK, from, to phase0.Slot) ([]ParticipantsRangeEntry, error)
+
+	// GetParticipants returns participants in quorum for the given slot.
+	GetParticipants(pk spectypes.ValidatorPK, slot phase0.Slot) ([]spectypes.OperatorID, error)
+
+	// InitialSlotGC performs an initial cleanup (blocking) of slots bellow the retained threshold
+	InitialSlotGC(ctx context.Context, logger *zap.Logger, threashold phase0.Slot)
```
moshe-blox, line 62: threshold*

moshe-blox, line 62: i suggest to call it pruning
```suggestion
	Prune(ctx context.Context, logger *zap.Logger, below phase0.Slot)
```

protocol/v2/qbft/storage/participant_store.go
```diff
@@ -0,0 +1,66 @@
...
+	Signers []spectypes.OperatorID
+}
+
+// ParticipantStore is the store used by QBFT components
+type ParticipantStore interface {
+	// CleanAllInstances removes all records in old format.
+	CleanAllInstances() error
+
+	// UpdateParticipants updates participants in quorum.
+	UpdateParticipants(pk spectypes.ValidatorPK, slot phase0.Slot, newParticipants []spectypes.OperatorID) (bool, error)
```
moshe-blox, line 50:
```suggestion
	SaveParticipants(pk spectypes.ValidatorPK, slot phase0.Slot, newParticipants []spectypes.OperatorID) (bool, error)
```
just to align with our sharesStorage which uses "save"

ibft/storage/store.go
```diff
@@ -156,16 +287,14 @@ func mergeParticipants(existingParticipants, newParticipants []spectypes.Operato
...
+	return i.db.Using(txn).Set(prefix, pk, value)
 }
 
-func (i *ibftStorage) get(txn basedb.ReadWriter, id string, pk []byte, keyParams ...[]byte) ([]byte, bool, error) {
-	prefix := append(i.prefix, pk...)
-	key := i.key(id, keyParams...)
-	obj, found, err := i.db.Using(txn).Get(prefix, key)
+func (i *participantStorage) get(txn basedb.ReadWriter, pk, slot []byte) ([]byte, bool, error) {
+	prefix := i.makePrefix(slot)
+	obj, found, err := i.db.Using(txn).Get(prefix, pk)
```
moshe-blox, line 297: if found is false on error then we ignore the error

since error is more significant than existance so we should check it first imo

cli/operator/node.go
anatolie-ssv: fixed

storage/basedb/storage.go
```diff
@@ -7,10 +7,11 @@ import (
...
 type Options struct {
-	Ctx        context.Context
-	Path       string        `yaml:"Path" env:"DB_PATH" env-default:"./data/db" env-description:"Path for storage"`
-	Reporting  bool          `yaml:"Reporting" env:"DB_REPORTING" env-default:"false" env-description:"Flag to run on-off db size reporting"`
-	GCInterval time.Duration `yaml:"GCInterval" env:"DB_GC_INTERVAL" env-default:"6m" env-description:"Interval between garbage collection cycles. Set to 0 to disable."`
+	Ctx               context.Context
+	Path              string        `yaml:"Path" env:"DB_PATH" env-default:"./data/db" env-description:"Path for storage"`
+	Reporting         bool          `yaml:"Reporting" env:"DB_REPORTING" env-default:"false" env-description:"Flag to run on-off db size reporting"`
+	GCInterval        time.Duration `yaml:"GCInterval" env:"DB_GC_INTERVAL" env-default:"6m" env-description:"Interval between garbage collection cycles. Set to 0 to disable."`
+	RetainedSlotCount uint64        `yaml:"RetainedSlotCount" env:"DB_RETAINED_SLOT_COUNT" env-default:"300" env-description:"How many slots we want to keep back"`
```
moshe-blox, line 14: sorry i missed this one before, the rest of the PR looks great

i suggest to move this under Exporter option and ideally call it `ExporterRetainSlots`, so that they're together in the YAML and the relation is unmistakeable:
```
Exporter: true
ExporterRetainSlots: 300
```

btw i suggest we should bump the default to 7 days (50,400 slots) to help clients (Explorer) resync history in case they had a bug discovered, wdyt? @anatolie-ssv

anatolie-ssv: Updated the name, location and its value!

moshe-blox (approved):
LGTM 🔥

PR #1852, opened 2024-11-13
feat(metrics): implement OTeL metrics across all packages

oleg-ssvlabs:
**Background**  
The observability team is currently working on a metrics refactor to address various challenges with the current implementation. As part of this refactor, _Prometheus SDKs_ are being replaced with _OTeL SDKs_, which are industry-leading in observability and provide an abstraction layer for various observability "backends". Additionally, _OTeL SDKs_ will be used to implement _Traces_. This aspect should be taken into consideration.

**Implementation**  

1. Central observability configurator was added (new root pkg `observability`). It looks simple, although it wasn't simple to decide what should "configurator" include mainly because OTeL supports two way of provider configuration - _SDKs_ and _environment variables_ and we want SSV Node users to be able to configure observability via native OTeL env vars. Some research showed that the vision of OTeL is "SDK first" meaning there are many features that aren't possible to configure via environment variables, yet possible via SDKs.However, the long-term aim is not to wrap OTeL environment variables into custom environment variables. Just FYI.

2. Multiple components were instrumented, including the `Execution Client`, `Event Syncer`, `Duty Scheduler`, `P2P`,  `Runners`. 

3. `Conventions.md` which describes OTeL metric related conventions (mainly naming). Will continuously be updated. Conventions (are described in the document) are based on OTeL original conventions. Although, in some cases OTeL's ambiguity was tightened.


y0sher (approved):

y0sher (approved):

y0sher (changes requested):
very minor comments. Looks good! I'm excited about this :)

eth/eventhandler/event_handler.go
```diff
@@ -198,12 +202,12 @@ func (eh *EventHandler) processEvent(txn basedb.Txn, event ethtypes.Log) (Task,
 			eh.logger.Warn("could not parse event",
 				fields.EventName(abiEvent.Name),
 				zap.Error(err))
-			eh.metrics.EventProcessingFailed(abiEvent.Name)
+			eventsProcessFailureCounter.Add(ctx, 1, metric.WithAttributes(eventNameAttribute(abiEvent.Name)))
```
y0sher, line 205: Maybe creating a function like:
`RecordFailure(ctx, abiEvent.Name)`
since it repeats itself and we want to make sure all changes will be implemented in all locations.

observability/CONVENTIONS.md
y0sher, line 1: probably should throw in a word about cardinality. I think link is not enough, just a simple example of what not to do.

operator/duties/scheduler.go
```diff
@@ -383,11 +371,12 @@ func (s *Scheduler) ExecuteGenesisDuties(logger *zap.Logger, duties []*genesissp
 		if slotDelay >= 100*time.Millisecond {
 			logger.Debug("⚠️ late duty execution", zap.Int64("slot_delay", slotDelay.Milliseconds()))
 		}
-		slotDelayHistogram.Observe(float64(slotDelay.Milliseconds()))
+		slotDelayHistogram.Record(context.TODO(), slotDelay.Seconds())
```
y0sher, line 374: we could probably use use the context passed in `SchedulerOptions` (need to save it in the `Scheduler` struct).

eth/eventhandler/event_handler.go
oleg-ssvlabs: good point. There are quite few places where this metric is being recorded. I added a wrapper function and updated the usages.

operator/duties/scheduler.go
oleg-ssvlabs: fixed. Added a `ctx` field to the Scheduler object that is passed through ctor as part of `SchedulerOptions` and switched to use it instead of `context.TODO()`.

operator/duties/scheduler.go
```diff
@@ -158,6 +146,7 @@ func NewScheduler(opts *SchedulerOptions) *Scheduler {
 		ticker:   opts.SlotTickerProvider(),
 		reorg:    make(chan ReorgEvent),
 		waitCond: sync.NewCond(&sync.Mutex{}),
+		ctx:      opts.Ctx,
```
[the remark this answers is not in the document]
oleg-ssvlabs: I agree on skipping the genesis code instrumentation mainly because there is no reason to do that at this point. 
I removed `Scheduler` struct level context and instrumentation of `ExecuteGenesisDuties` method

observability/metric.go
```diff
@@ -0,0 +1,18 @@
+package observability
+
+import "go.uber.org/zap"
+
+var (
+	SecondsHistogramBuckets = []float64{0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10}
+)
+
+func GetMetric[T any](metricName string, create func(metricName string) (T, error)) T {
```
moshe-blox, line 9: convention in Go is to avoid Get prefixes, maybe `NewMetric`? (just to avoid a non-standard prefix like Setup or Create, name doesn't have to be 100% "correct")

eth/executionclient/observability.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
+	for _, status := range []executionClientStatus{statusReady, statusSyncing, statusFailure} {
+		clientStatusGauge.Record(ctx, 0,
+			metric.WithAttributes(executionClientAddrAttribute(nodeAddr)),
+			metric.WithAttributes(executionClientStatusAttribute(status)),
+		)
+	}
+}
```
moshe-blox, line 98: can just be one function with a `status bool` parameter IMO, to reduce boilerplate

oleg-ssvlabs: do you mean merge these two functions and remove `nodeAddr` attribute from the metric?

observability/metric.go
oleg-ssvlabs: good point. `New` prefix would be a better fit here since the function is responsible for instantiation of new metric object

eth/executionclient/observability.go
moshe-blox: oh sorry, i guess i didnt understand the purpose of the reset

is it really how to express enums with otel?

can't we instead just encode the status in the value? (so 0 = "syncing", 1 = "failure", etc..)

oleg-ssvlabs: This isn’t specifically OTeL-specific. It is the same pattern in Prometheus. The problem definition is: “In the metric, I want to indicate a value that isn’t a number.” Observability frameworks use number as _metric values_. One option is to do what you suggested - use indicative enums, another option is to represent it as a label with a boolean value (1/0). 
Both options have pros and cons. I cannot 100% say I fully evaluated both. We had discussion about this with @zktaiga . 
What we know for sure - the problem remains and we need to be consistent with the pattern. Probably less important which option we pick, more important to stick with the same pattern everywhere.

observability/metric.go
```diff
@@ -0,0 +1,15 @@
+package observability
+
+import "go.uber.org/zap"
+
+var (
+	SecondsHistogramBuckets = []float64{0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10}
```
zktaiga, line 6: I'm not trying to nitpick, just curious: when measuring delay/latency, can any value ever be placed on the `0` bucket?

zktaiga (approved):
Great work! It's now really straightforward to add/maintain existing metrics on top of the agnostic nature of OTel. :rocket:

observability/metric.go
zktaiga: It can be kept and we can revisit if we think it's impeding visualization.

beacon/goclient/proposer.go
```diff
@@ -166,7 +170,14 @@ func (gc *GoClient) SubmitBlindedBeaconBlock(block *api.VersionedBlindedProposal
 		Proposal: signedBlock,
 	}
 
-	return gc.client.SubmitBlindedProposal(gc.ctx, opts)
+	start := time.Now()
+	if err := gc.client.SubmitBlindedProposal(gc.ctx, opts); err != nil {
+		return err
+	}
+
+	recordRequestDuration(gc.ctx, "SubmitBlindedProposal", gc.client.Address(), http.MethodPost, time.Since(start))
```
y0sher, line 178: This comment works a lot of different API call metrics. What if the API call err'ed , some times its useful to know its err'ed and the duration it took the client to return error.

oleg-ssvlabs: The potential issue with recording this metric when an error is returned is that the `SubmitBlindedProposal` method has additional responsibilities beyond the HTTP call, such as input validation, which happens before the HTTP call and may result in an error. The purpose of this histogram is to record _response times for HTTP calls_.
This is a trade-off: by not recording errors that may result from simple input validation, we avoid diluting the histogram. However, this means missing out on network errors and unsuccessful status codes (since the library wraps everything that isn’t a 2XX status code into an error). Of course we can decide to record durations always, even if there is an error returned, that's just the pattern followed across entire codebase. Valid point though.

y0sher: well, what you mentioned also works for successful calls. if you measuring `SubmitBlindedProposal` as a whole you always adding the input checks time inside. maybe we need to measure in lower level if we want only http call for real. Also wouldn't it be easy to not corrupt the histogram data by adding an error label so in dashboard you can look only on successful calls? I guess we could also add reason label and only look at errors that can back from the API as well.

I think this is a direction we want to go in, of course you can draw action items out of it and save it for next versions. @zktaiga @oleg-ssvlabs @moshe-blox .

message/validation/errors.go
```diff
@@ -136,21 +137,19 @@ func (mv *messageValidator) handleValidationError(peerID peer.ID, decodedMessage
...
+		recordIgnoredMessage(ctx, loggerFields.Role, valErr.Text())
 		return pubsub.ValidationIgnore
 	}
 
 	if !valErr.Silent() {
 		logger.Debug("rejecting invalid message", zap.Error(valErr))
 	}
 
-	mv.metrics.MessageRejected(valErr.Text(), loggerFields.Role, loggerFields.Consensus.Round)
+	recordRejectedMessage(ctx, loggerFields.Role, valErr.Text())
```
y0sher, line 148: note this might be high cardinality because validation erros

network/discovery/observability.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type rejectionReason string
+
+const (
+	reachedLimitReason    rejectionReason = "reachedLimit"
+	noSharedSubnetsReason rejectionReason = "noSharedSubnets"
+	zeroSubnetsReason     rejectionReason = "zeroSubnets"
+)
+
```
y0sher, line 26: names here are kinda confusing, because we just rejection in message validation and also connections I guess. this sounds like we are rejecting connections or something, but its a much more lightweight action here, we are filtering/skipping peers that we found in discovery and aren't fitting for us. we are not connected or rejecting connections with us. 

maybe a better terminology will be skipping or filter?

network/p2p/observability.go
```diff
@@ -0,0 +1,112 @@
...
+		meter.Int64Gauge(
+			metricName("peers.connected"),
+			metric.WithUnit("{peer}"),
+			metric.WithDescription("number of connected peers")))
+
+	connectionsGauge = observability.NewMetric(
+		meter.Int64Gauge(
+			metricName("connections.active"),
+			metric.WithUnit("{connection}"),
+			metric.WithDescription("number of active connections")))
```
y0sher, line 36: I don't get the difference between these two metrics?
shouldn't it be the same?

oleg-ssvlabs: Sure.
If to look into implementation - `peer` and `connection` are two distinctly different entities, although they relate to each other, and their relationships are one to many, meaning one `peer` has one or more `connections`. `Connection` also has a `direction` attribute - either `inbound` or `outbound`. 
[Here](https://github.com/ssvlabs/ssv/blob/otel-metrics/network/p2p/observability.go#L63) it is visible how connections can be fetched via providing the Peer ID, and it is a collection of items. We separate them on our dashboard. Frankly, in most of the cases, it's 1:1, although since API tells us it's a collection, I assume it is possible it to be more than connection per peer. 
Wdyt?

network/topics/tracer.go
```diff
@@ -24,15 +24,9 @@ func newTracer(logger *zap.Logger) pubsub.EventTracer {
...
 func (pst *psTracer) Trace(evt *ps_pb.TraceEvent) {
-	pst.report(evt)
 	pst.log(pst.logger, evt)
 }
 
-// report reports metric
-func (pst *psTracer) report(evt *ps_pb.TraceEvent) {
-	metricPubsubTrace.WithLabelValues(evt.GetType().String()).Inc()
-}
-
```
y0sher, line 35: just mentioning this was actually pretty useful when we were debugging libp2p pubsub issues. so worth leaving it. I guess its only emitted when we turn on tracer so no harm in leaving?

oleg-ssvlabs: We do have very similar metrics that are also recorded in the P2P Topics domain - `messages in` and `messages out`, although we do not have `event type` as dimension. It was removed because it was high cardinality metric, but I understand it was not enabled by default. There is still a [log written](https://github.com/ssvlabs/ssv/blob/otel-metrics/network/topics/tracer.go#L27), maybe that would be enough for troubleshooting? Currently all the metrics are low cardinality hence we do not have "turn off/on" feature for any of the metrics. I see that as a separate feature which we can include in the next release, so we can come up with the list of metrics that are optional and disabled by default. Wdyt?

operator/duties/scheduler.go
```diff
@@ -403,9 +389,10 @@ func (s *Scheduler) ExecuteCommitteeDuties(ctx context.Context, logger *zap.Logg
 		if slotDelay >= 100*time.Millisecond {
 			logger.Debug("⚠️ late duty execution", zap.Int64("slot_delay", slotDelay.Milliseconds()))
 		}
-		slotDelayHistogram.Observe(float64(slotDelay.Milliseconds()))
+		slotDelayHistogram.Record(ctx, slotDelay.Seconds())
 		go func() {
 			s.waitOneThirdOrValidBlock(duty.Slot)
+			recordDutyExecuted(ctx, spectypes.BNRoleSyncCommittee)
```
y0sher, line 395: this is NOT sync committee, its `Committee` - a special duty that has both attestation and sync committee duties included for the whole slot/committee.

operator/node.go
```diff
@@ -201,15 +187,14 @@ func (n *operatorNode) startWSServer(logger *zap.Logger) error {
...
-func (n *operatorNode) reportOperators(logger *zap.Logger) {
+func (n *Node) reportOperators(logger *zap.Logger) {
 	operators, err := n.storage.ListOperators(nil, 0, 1000) // TODO more than 1000?
 	if err != nil {
 		logger.Warn("failed to get all operators for reporting", zap.Error(err))
 		return
 	}
 	logger.Debug("reporting operators", zap.Int("count", len(operators)))
 	for i := range operators {
-		n.metrics.OperatorPublicKey(operators[i].ID, operators[i].PublicKey)
```
y0sher, line 212: not vouching for adding it back just mentioning this might look like high cardinality in our stage environment because we have many operators but for a single operator this metric only has one value.

operator/validator/controller.go
```diff
@@ -1050,6 +1041,57 @@ func (c *controller) fetchAndUpdateValidatorsMetadata(logger *zap.Logger, pks []
...
 
+func (c *controller) ReportValidatorStatuses(ctx context.Context) {
+	ticker := time.NewTicker(time.Second * 30)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			start := time.Now()
+			validatorsPerStatus := make(map[validatorStatus]uint32)
```
y0sher, line 1052: we should definitely check the resource impact of this loop, we probably can get all these metrics inside the metadata collection loop without running this extra loop.

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -278,8 +278,6 @@ func (ncv *CommitteeObserver) resolveDuplicateSignature(container *ssv.PartialSi
 
 // copied from BaseRunner
 func (ncv *CommitteeObserver) verifyBeaconPartialSignature(signer uint64, signature spectypes.Signature, root phase0.Root, share *ssvtypes.SSVShare) error {
-	ssvtypes.MetricsSignaturesVerifications.WithLabelValues().Inc()
```
y0sher, line 281: for future versions this metric might be interesting when looking into the performance of our node.

protocol/v2/ssv/validator/validator.go
```diff
@@ -81,7 +77,7 @@ func NewValidator(pctx context.Context, cancel func(), options Options) *Validat
 		role := dutyRunner.GetBaseRunner().RunnerRoleType
 
 		v.Queues[role] = queueContainer{
-			Q: queue.WithMetrics(queue.New(options.QueueSize), options.Metrics),
+			Q: queue.New(options.QueueSize),
```
y0sher, line 80: why no metrics here? in committee you added the metrics for the queue.

y0sher (changes requested):
Amazing work! this is exciting. minor comments :)

network/p2p/observability.go
y0sher: It seems to be possible in terms of networking and libp2p, but AFAIK we limit it on our end.
So peers.connected will always be 1:1 with connections just without the direction directive. maybe we can consolidate them. @zktaiga @moshe-blox wdyt?

operator/duties/scheduler.go
oleg-ssvlabs: good catch!
This metric has `beacon role` as an attribute. I do not think there is Committee role in beacon, since it's more of a SSV term (from what I understand). We also often use `runner role` attributes, which has "committee". Wdyt if we use `runner role` instead of `beacon role` here? I assume scheduler at this point does not know whether it is attestation or sync committee duty

protocol/v2/ssv/runner/proposer.go
```diff
@@ -266,10 +271,9 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
 		start := time.Now()
 
 		logger = logger.With(
-			fields.BeaconDataTime(r.metrics.GetBeaconDataTime()),
```
moshe-blox, line 269: why was BeaconDataTime removed? i realize maybe because its tracked in a different way now in metrics, but this doesnt have to affect the log

operator/duties/scheduler.go
oleg-ssvlabs: There is another option (I think `beacon role` is slightly more valuable attribute(considering our audience) than `runner role`, since runner is our internal term) is to iterate over duties that `spectypes.CommitteeDuty` holds and read `Type` (beacon role) property. That's how we can calculate Attestation and Sync Committee duties and record the metric. What do you think would be better @y0sher ?

y0sher: I think the option to use runner role is probably better. since this metric is about execution, it will be incorrect to say we executed attester duties. we don't execute attester duties anymore, we have our internal committee duty for that. we only submit the attester duties when we finish the committee execution successfully.

oleg-ssvlabs: @y0sher fixed -> changed attribute from `ssv.beacon.role` to `ssv.runner.role`

network/discovery/observability.go
oleg-ssvlabs: Following the naming in the code, I see that mostly word "skip" is used. I will refactor the metric

network/discovery/dv5_service.go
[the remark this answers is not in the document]
oleg-ssvlabs: Added `domainTypeMismatchReason` and metric recording

network/discovery/observability.go
oleg-ssvlabs: Renamed the metric from `ssv.p2p.discovery.peers.rejected` to `ssv.p2p.discovery.peers.skipped`, also renamed attribute from `ssv.p2p.discovery.rejection_reason` to `ssv.p2p.discovery.skip_reason`

network/topics/tracer.go
oleg-ssvlabs: Let's put it in the backlog so we do not extend the scope of this PR/release if you do not mind

operator/validator/controller.go
oleg-ssvlabs: I see what you mean.

The loop in the `UpdateValidatorMetaDataLoop` method iterates over shares via Share Storage, which seems to provide the information we need (_validators with statuses_). This would allow us to avoid iterating over roughly 60k items (validators). Currently, it takes about 7-15 ms, but the concern is not so much the execution time, since it runs in a separate goroutine every 30 seconds, but rather the resource usage.

It seems like a trade-off between reducing resource usage and maintaining cleaner code.

Is it possible to fetch operator's validators and skip iterating over full list of validators? That probably would be the most optimal solution. 

Wdyt @moshe-blox @y0sher ?

moshe-blox: @oleg-ssvlabs yeah, we made `ValidatorStore` just for that :)

method is called literally OperatorValidators also 🤣 

i think you can already access it from `controller`

operator/validator/controller.go
```diff
@@ -1050,6 +1041,57 @@ func (c *controller) fetchAndUpdateValidatorsMetadata(logger *zap.Logger, pks []
...
+			for _, share := range c.validatorStore.Validators() {
```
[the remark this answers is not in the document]
oleg-ssvlabs: fixed now. Switched to use `OperatorValidators` method instead

operator/validator/controller.go
oleg-ssvlabs: fixed

operator/validator/task_executor.go
```diff
@@ -24,7 +24,7 @@ func (c *controller) taskLogger(taskName string, fields ...zap.Field) *zap.Logge
 func (c *controller) StopValidator(pubKey spectypes.ValidatorPK) error {
 	logger := c.taskLogger("StopValidator", fields.PubKey(pubKey[:]))
 
-	c.metrics.ValidatorRemoved(pubKey[:])
+	validatorsRemovedCounter.Add(c.ctx, 1)
```
moshe-blox, line 27: this isn't necessarily removed, but rather stopped (it could also be because of contract liquidation)

network/p2p/metrics.go
moshe-blox, line 1: we're relying on the logs here to debug issues from our operators, as they don't send us links to their Grafana, but rather just log files

- "connected peers status" - needed
- "topic peers status" - needed
- "topic peers distribution" - can be dropped unless @y0sher or @iurii-ssv think it's very valuable
- "peer identity" - maybe not needed because this data already appears in handshake logs IIRC

moshe-blox (changes requested):

network/discovery/forking_dv5_listener.go
```diff
@@ -95,9 +95,5 @@ type annotatedIterator struct {
 }
 
 func (i *annotatedIterator) Next() bool {
-	if !i.Iterator.Next() {
-		return false
-	}
-	metricIterations.WithLabelValues(i.fork).Inc()
-	return true
+	return i.Iterator.Next()
 }
```
moshe-blox, line 99: stage branch previously had removed dual discovery, but it has since been reverted after we've encountered issues without it

so we still need a metric that's broken down to pre-fork and post-fork discoveries at the moment

the other metrics for discovery accepted/rejected can remain as they are, as they're at a higher level than this

the metric here was just measuring the performance of each iterator in order to compare pre-fork vs post-fork discoveries

moshe-blox, line 99: stage branch previously had removed dual discovery, but it has since been reverted after we've encountered issues without it

so we still need a metric that's broken down to pre-fork and post-fork discoveries at the moment

the other metrics for discovery accepted/rejected can remain as they are, as they're at a higher level than this

the metric here was just measuring the performance of each iterator in order to compare pre-fork vs post-fork discoveries

network/p2p/metrics.go
y0sher: I agree on all. 
I think `topic peers distribution` can be useful even though we are now using mostly logs. it could be used as a valuable alert for an operator to know he has a "dead" subnet.

protocol/v2/ssv/validator/validator.go
y0sher: @moshe-blox not talking about anything specific, just wondered why the change

oleg-ssvlabs: @y0sher that metric was not used on any of the dashboards, also it was disabled in production and it was high cardinality (msgID as a label) hence it was removed.

eth/eventhandler/event_handler.go
```diff
@@ -116,8 +117,11 @@ func (eh *EventHandler) HandleBlockEventsStream(logs <-chan executionclient.Bloc
 		if err != nil {
 			return 0, fmt.Errorf("failed to process block events: %w", err)
 		}
-
 		lastProcessedBlock = blockLogs.BlockNumber
+
+		if lastProcessedBlock <= math.MaxInt64 {
+			lastProcessedBlockGauge.Record(ctx, int64(lastProcessedBlock))
+		}
```
iurii-ssv, line 124: It seems converting `uint64` -> `int64` is a common thing we do multiple times, perhaps we should have a helper function for this instead to simplify away all these additional `if`s

protocol/v2/ssv/runner/proposer.go
oleg-ssvlabs: As we discussed - `GetBeaconDataTime` was never recorded, hence it was returning 0

operator/validator/controller.go
```diff
@@ -1050,6 +1041,55 @@ func (c *controller) fetchAndUpdateValidatorsMetadata(logger *zap.Logger, pks []
...
+			for status, count := range validatorsPerStatus {
+				c.logger.
+					With(zap.String("status", string(status))).
+					With(zap.Uint32("count", count)).
+					With(zap.Duration("elapsed_time", time.Since(start))).
+					Info("recording validator status")
+				recordValidatorStatus(ctx, count, status)
+			}
+		case <-ctx.Done():
+			c.logger.Info("stopping to report validator statuses. Context cancelled")
```
iurii-ssv, line 1086: `stopping to report` -> `stopped reporting`

network/p2p/metrics.go
oleg-ssvlabs: - "connected peers status" - is it a new metric we want, or something we had before?
- "topic peers status" - is it a new metric we want, or something we had before?
- "topic peers distribution" - we have this metric [implemented](https://github.com/ssvlabs/ssv/pull/1852/files/8f22f99c9b3708c3f8f91e3d5fd4f1bdeb9cffee#diff-f86917a1e23b7036ed91c90f0c80bfa35c2efdb5c7300621b080cdea37bb35acR38)
-  "peer identity" - we have it implemented [here](https://github.com/ssvlabs/ssv/pull/1852/files/8f22f99c9b3708c3f8f91e3d5fd4f1bdeb9cffee#diff-f86917a1e23b7036ed91c90f0c80bfa35c2efdb5c7300621b080cdea37bb35acR44), although there is only single attribute set - `ssv.node.version`

iurii-ssv (commented):
Minor suggestions

cli/operator/node.go
```diff
@@ -113,9 +111,19 @@ var StartNodeCmd = &cobra.Command{
...
+			observability.WithMetrics())
+		if err != nil {
+			logger.Fatal("could not initialize observability configuration", zap.Error(err))
+		}
+
+		defer func() {
+			if err = observabilityShutdown(cmd.Context()); err != nil {
+				logger.Error("could not shutdown observability object", zap.Error(err))
+			}
+		}()
```
iurii-ssv, line 126: Can `observabilityShutdown` be `nil` ? I suspect we might get "nil-pointer dereference" here (since we don't return on non-nil error above)

beacon/goclient/attest.go
```diff
@@ -25,6 +27,8 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
 		return nil, fmt.Errorf("attester duties response is nil")
 	}
 
+	recordRequestDuration(gc.ctx, "AttesterDuties", gc.client.Address(), http.MethodPost, time.Since(start))
```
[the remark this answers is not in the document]
oleg-ssvlabs: It's a good suggestion, although it cannot be applied everywhere, since in some of the places the method/function has more logic implemented and we do not want to capture these, since the metric suppose to indicate the beacon request durations only, hence recording it at the exit would not work.
It was tricky question (we have discussion in the comments) whether to record metric if there is an error from the lib that wraps beacon calls, since errors aren't always associated with the network calls, so that's another reason `defer`s were not used

eth/eventhandler/event_handler.go
oleg-ssvlabs: I agree.
[Wdyt](https://github.com/ssvlabs/ssv/pull/1955)?

operator/validator/controller.go
oleg-ssvlabs: fixed

beacon/goclient/attest.go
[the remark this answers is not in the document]
iurii-ssv: Yeah, I understand it's probably not gonna work for every case we have, but I think it's a good default-solution to use when it applies.

> It was tricky question (we have discussion in the comments) whether to record metric if there is an error from the lib that wraps beacon calls, since errors aren't always associated with the network calls, so that's another reason defers were not used

we can handle `error` in `defer` too however we want if we define `err error` prior to that `defer` and just check `err != nil` in there - it's a bit "ugly" but still better in most cases IMO because there is that clear separation of business logic and measuring/metrics logic

cli/operator/node.go
oleg-ssvlabs: That’s a good catch. There is a [conditional assignment](https://github.com/ssvlabs/ssv/blob/otel-metrics/observability/observability.go#L48) of the shutdown function variable, and the default value is `nil`.

I’m not a big fan of named return values, as they can lead to scenarios like this. However, in this case, it does serve as documentation. An alternative could be to declare a new type to address the "documentation issue", but that might be overkill.
I will fix the issue by assigning a default value. Wdyt?

eth/eventhandler/event_handler.go
iurii-ssv: Yes that's probably the best way to do it.

cli/operator/node.go
iurii-ssv: > I’m not a big fan of named return values, as they can lead to scenarios like this. However, in this case, it does serve as documentation.

Agree with all that,

can you elaborate why we even need `once.Do` in `Initialize` ? I think it overcomplicates things

oleg-ssvlabs: > can you elaborate why we even need once.Do in Initialize (I don't see how it could run 2+ times) ? 

It's a guard from the undesired consequences if the client mistakenly calls `Initialize` function more than one time. There are many ways of implementing it, like declaring package level variable (`isInitialized` bool) and ignore the calls if it was already initialized or declaring `init()` function which runs when package is loaded and only once (although would not work for us, since there are objects in return) and such. 
It isn't mandatory, but I think it isn't bad that some functionality acts defensively, especially functionality that is exposed to other packages.

iurii-ssv: Personally, I don't think it makes much sense in this particular case to "try defend" like that - it would be a bug to call `Initialize` 2+ times just same as if we were to call many other functions of `.Start` / `.Init` / ... kind (and we don't wrap those with `once.Do`)

iurii-ssv: > I see. I do not have strong opinion here. Considering we do not follow this pattern in other packages, do you think it would be better to remove it?

Yeah, I would remove `once.Do` both - to keep it simple, and to be consistent with the other code we have

oleg-ssvlabs: removed

network/p2p/metrics.go
oleg-ssvlabs:
```
topic peers distribution
topic peers status
connected peers status
```

Log messages were added

network/discovery/forking_dv5_listener.go
oleg-ssvlabs: I introduced new metrics based on your comment: `ssv.p2p.discovery.iterations` (counter) with attribute `ssv.fork`

beacon/goclient/attest.go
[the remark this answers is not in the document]
oleg-ssvlabs: We took a slightly different approach: the function that records the metric is now called immediately after the HTTP call to the beacon. This function now also takes an error as an argument. Its implementation has gained some additional responsibilities, such as checking if the error is an API error and, if so, adding the error status code as a metric attribute.

y0sher (approved):
Good job. @moshe-blox we need to do spec align.

network/p2p/metrics.go
moshe-blox: @andrew-blox let's please not forget to check these logs still appear as they used to 🙏

moshe-blox (approved):
LGTM, well done! 🔥

moshe-blox (approved):

PR #1308, opened 2024-02-06
feat(network/config): pass custom ssv config & get beacon config from node & remove hardcoded network values

nkryuchkov:
Closes #1687, #1865 

Changes:
- split `networkconfig.NetworkConfig` into `networkconfig.SSV` and `networkconfig.Beacon`
- pass custom SSV network config (`networkconfig.SSV`) in yaml
- get beacon config (`networkconfig.Beacon`) from beacon node
- script for custom SSV (`networkconfig.SSV`) config generation
- refactor code around the `networkconfig` package
- get rid of `BeaconNetwork` interface as it's hard to work with
- use network values from config instead of hardcoded ones as a preparation for local testnet

PR #1859, opened 2024-11-19
fix: `CommitteeObserver` - post conesnsus mem leak

### LINKED ISSUE #1857: exporter: memory leak in `CommitteeObserver`.postConsensusContainer
### Description 

It seems that `postConsensusContainer` map is never cleaned and always keeps growing causing a memory leak in exporter.

*NOTE*: it also looks like we are saving all the post consensus sigs under validator index without any notion of the slot?

### Possible solutions

Change the mapping to be per-slot and remove on slots on newer slots arrival.

y0sher:
### Description

Closes #1857

Remove 12 old slots after each message processing.

*NOTE*: maybe we need tracking of latest actual slot in order to ignore old messages? what if we already removed but got new msg?


protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -250,6 +257,13 @@ func (ncv *CommitteeObserver) processMessage(
 			}
 		}
 	}
+
+	// Remove older slots container
+	thresholdSlot := currentSlot - 12 // won't get new messages after 12 slots
+	if _, exists := ncv.postConsensusContainer[thresholdSlot]; exists {
```
nkryuchkov, line 263: The linter complains we don't need this check (delete won't delete what doesn't exist)

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -250,6 +257,13 @@ func (ncv *CommitteeObserver) processMessage(
...
+		delete(ncv.postConsensusContainer, thresholdSlot)
```
nkryuchkov, line 264: Do I understand correctly that it's possible that if there are no messages in a slot, there is a gap and older messages won't be removed? If so, perhaps it would be great to leave a comment that it's very unlikely to happen, so it shouldn't be an issue

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -250,6 +257,13 @@ func (ncv *CommitteeObserver) processMessage(
 			}
 		}
 	}
+
+	// Remove older slots container
+	thresholdSlot := currentSlot - 12 // won't get new messages after 12 slots
```
nkryuchkov, line 262: Can you please comment on why 12?

nkryuchkov (approved):
Well spotted! I'd just add more comments. Approving but please fix linter

protocol/v2/ssv/validator/non_committee_validator.go
moshe-blox: we allow attestations i think up to 32 slots late

protocol/v2/ssv/validator/non_committee_validator.go
moshe-blox: yeah this could cause unbounded growth because for every skipped slot X the map will keep slot X-12 forever

i think we could try something like this instead:

```go
const containerSize phase0.Slot = 34
// ...
postConsensusContainer [containerSize]map[phase0.ValidatorIndex]*ssv.PartialSigContainer

// Access:
index := msg.Slot%containerSize
postConsensusContainer[index] = ...
```

protocol/v2/ssv/validator/non_committee_validator.go
nkryuchkov: @moshe-blox yeah, actually it's 34 (32 + lateSlotAllowance). I just thought more about it and it's not clear for me why this would be 12. If we change it to 34, let's get it from network config as we are preparing for the local testnet. Also, we can export `lateSlotAllowance` from message validation (although we'll probably need to think about a more convenient way to use as this is not the only place where we'd need 34)

protocol/v2/ssv/validator/non_committee_validator.go
moshe-blox: @nkryuchkov sure, this pattern could be coded to something like:

```go
type SlotContainer[T any] struct {
  capacity int
  items []T
}

Get(phase0.Slot)
Set(phase0.Slot, T)
```

however now that i think of it, isn't there a risk of a early or late messages overwriting the current working set? (~last 34 slots)

nkryuchkov: @moshe-blox good point, it seems the size should include allowance in the past + future. By the way, we already use 32/34 somewhere, so we need to fix it.

Also, I think we need to bind it to message validation somehow because if we change it, it will be hard to find all places where it's used

protocol/v2/ssv/validator/non_committee_validator.go
y0sher: We've decided to implement a naive solution for now where we just loop on the map (which should always be 34 len max) and remove any slot that is => current slot - 32 .
deployed it to watch how it behaves since it's done every message, though the map size should be pretty small so it might not have a huge impact.

protocol/v2/ssv/validator/non_committee_validator.go
y0sher: see this https://github.com/ssvlabs/ssv/pull/1859#discussion_r1850799907
of course we could implement something more elegant but for now we want to see the impact of just removing from the map :)

iurii-ssv (approved):
Could alternatively maybe use `ttlcache.Cache` (seems like we use it a lot whenever we need to keep map from growing).

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/ssvlabs/ssv/utils/casts"
 )
 
+const postConsensusContainerCapacity = 34
```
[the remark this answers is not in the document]
y0sher: https://github.com/ssvlabs/ssv/issues/1865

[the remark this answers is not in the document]
nkryuchkov: @y0sher #1865 is actually already done in https://github.com/ssvlabs/ssv/pull/1308, therefore I think we should already stop hardcoding them in the new code

moshe-blox (approved):

nkryuchkov (approved):

protocol/v2/ssv/validator/non_committee_validator.go
nkryuchkov: it doesn't appear in pprof so IMO we can leave it as is for now, but I left a comment for the future

y0sher:
@MatusKysel @anatolie-ssv These changes were solving some issue in old exporter. Lets make sure we don't have the same issue with exporter v2.

PR #1860, opened 2024-11-19
slotticker: improve and clarify

iurii-ssv:
This PR:
- updates and clarifies `SlotTicker` docs (eg. calls out certain assumptions such as the need to call `Next` method periodically to "keep ticking" - which is different from how one would expect Golang ticker to operate, for example)
- strives to fix or clarify certain edge-cases (see open discussions below)

_Before merging:_
- [x] test on **stage**


operator/duties/scheduler.go
```diff
@@ -279,13 +277,13 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
-				s.headSlot = slot
-				s.waitCond.Broadcast()
-				s.waitCond.L.Unlock()
 			}
+
+			// Lock the mutex before broadcasting
+			s.waitCond.L.Lock()
+			s.headSlot = slot
+			s.waitCond.Broadcast()
+			s.waitCond.L.Unlock()
```
iurii-ssv, line 286: Seems like we should advance `headSlot` regardless of whether or not **we actually had to wait** to 1/3 here (that is - if tick happened past 1/3 for whatever reason we still want to advance it, right) ?

operator/slotticker/slotticker.go
```diff
@@ -36,34 +46,38 @@ func New(logger *zap.Logger, cfg Config) *slotTicker {
...
 
 	return &slotTicker{
 		logger:       logger,
 		timer:        timerProvider(initialDelay),
 		slotDuration: cfg.SlotDuration,
 		genesisTime:  cfg.GenesisTime,
-		slot:         0,
+		slot:         initialSlot,
 	}
 }
```
iurii-ssv, line 73: Seems like here we should start at some slot that corresponds to `cfg.GenesisTime` instead of at slot 0 ?

operator/duties/scheduler.go
nkryuchkov: @iurii-ssv The only idea that comes to my mind as to why this could have been done like that is to make sure the slot is not decreased

nkryuchkov (approved):

operator/duties/scheduler.go
```diff
@@ -300,7 +300,11 @@ func (s *Scheduler) HandleHeadEvent(logger *zap.Logger) func(event *eth2apiv1.Ev
...
 
 		data := event.Data.(*eth2apiv1.HeadEvent)
-		if data.Slot != s.network.Beacon.EstimatedCurrentSlot() {
+
+		// we are interested only in events for the current slot, but to account for wall-clock
+		// differences the next slot after "what we think is the current one" can also be valid
+		if data.Slot != s.network.Beacon.EstimatedCurrentSlot() &&
+			data.Slot != s.network.Beacon.EstimatedCurrentSlot()+1 {
 			return
 		}
```
iurii-ssv, line 309: I think this also needs a slight adjustment (like I suggest in this PR) because "current slot" is based on our own "local current time" which might slightly lag behind "true current time" (there is no such thing as true time in distributed systems really, but what I mean is somebody sends us an event slightly earlier than we expect it and considers this OK to do because "his clock is running early").

Not sure how likely it is to happen in practice though - still better safe than sorry.

operator/duties/scheduler.go
iurii-ssv: > @iurii-ssv The only idea that comes to my mind as to why this could have been done like that is to make sure the slot is not decreased

Not sure if that's even possible for "slot decreasing" to happen in the current code version (or how that time-based `if` would even help with that), but I'm adding [another change here](https://github.com/ssvlabs/ssv/pull/1860/files#r1850813217) that actually does make it a likely problem to happen,

and hence we probably want to have explicit protection against "slot decreasing" (which I've now added in this PR)

operator/slotticker/slotticker.go
```diff
@@ -77,17 +91,16 @@ func (s *slotTicker) Next() <-chan time.Time {
 	if nextSlot <= s.slot {
 		// We've already ticked for this slot, so we need to wait for the next one.
 		nextSlot = s.slot + 1
-		s.logger.Debug("double tick", zap.Uint64("slot", uint64(s.slot)))
+		s.logger.Debug("slotTicker: double tick", zap.Uint64("slot", uint64(s.slot)))
```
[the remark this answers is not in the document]
iurii-ssv: Yeah, that would be a better way to do it, applied in https://github.com/ssvlabs/ssv/pull/1860/commits/b4d70ac2ca24f605f28843f28968bf288d8eb9fc

operator/slotticker/slotticker.go
iurii-ssv: > I'm guessing this didn't cause any issues because we always called `Next` before calling `Slot` on brand-new `slotTicker`

note btw, this ^ also means we never actually "tick" (handle duties) for the very first `initialSlot` - but rather wait until the next slot that comes after it,

at the worst case we essentially skip 1 slot during start, not a big deal probably - but maybe now we can address this, wdyt ?

oleg-ssvlabs (approved):

operator/slotticker/slotticker.go
```diff
@@ -12,22 +12,37 @@ import (
...
-	Slot() phase0.Slot
+	// NextTick returns a channel that will relay 1 tick signalling that "freshest" slot has started.
+	// It advances slot number SlotTicker keeps track of (potentially jumping several slots ahead)
+	// and returns a channel that will signal once the time corresponding to that "freshest" slot
+	// comes.
+	NextTick() <-chan time.Time
+	// NextSlot returns the next slot number SlotTicker will tick on (corresponding to the tick
+	// that happens upon NextTick signal).
+	NextSlot() phase0.Slot
 }
```
iurii-ssv, line 27: Now that I better understand how `SlotTicker` works I think we maybe can change it to something like:
```go
// SlotTicker to advance forward (to keep ticking) to newer slots.
type SlotTicker interface {
	// NextTick returns a channel that will relay 1 tick signalling that "freshest" slot has started.
	// It advances slot number SlotTicker keeps track of (potentially jumping several slots ahead)
	// and returns a channel that will signal once the time corresponding to that "freshest" slot
	// comes.
	NextTick() <-chan phase0.Slot
}
```

operator/slotticker/slotticker.go
```diff
@@ -75,20 +91,23 @@ func (s *slotTicker) Next() <-chan time.Time {
...
 
-// Slot returns the current slot number.
-// Note: Like the Next function, this method is also not thread-safe.
-// It should be called in a serialized manner after calling Next.
-func (s *slotTicker) Slot() phase0.Slot {
-	return s.slot
+// NextSlot implements SlotTicker.NextSlot.
+// Note, this method is not thread-safe.
+func (s *slotTicker) NextSlot() phase0.Slot {
```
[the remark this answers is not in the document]
iurii-ssv: Right, I guess I was trying to match `NextTick` emphasizing that it needs to be called to advance slots (and that `NextSlot` if called before the tick happens will actually return next slot - the only problem is this isn't thread safe thing to do, but it technically can be done)

I've renamed it to just `Slot` for now but I think it's best if we can merge `NextTick` and `NextSlot` to be represented by a [single method that doesn't leave any room for interpretation](https://github.com/ssvlabs/ssv/pull/1860#discussion_r2020138141)

operator/slotticker/slotticker.go
iurii-ssv: I guess I'd need to try it out to see if it 100% fits,

the only "potential difficulty" I can think of would be: we'd need to be able to drop older slots replacing these by the freshest/latest one on the channel `NextTick` returns (in case the receiver didn't have a chance to consume those older slots) - we probably need to spin a go-routine for it in `slotTicker` that would "loop trying to send on the channel, updating slot value as time goes by and only succeeding once there is a recipient waiting for the value on that channel"

but still it seems this implementation caveat is worth simpler interface

iurii-ssv: We've discussed it a bit further with Matus, to sum up:
- there was a requirement to not use extra go-routines for the sake of saving resources (but based on the actual usage in SSV code AFAICT we don't spawn thousands of `slotTicker` - so it probably doesn't matter, unless there is a plan to spawn thousands in the future)
- being core component used is many places there is a slight chance of regression if we decide to go for this kind of refactoring

@moshe-blox @y0sher let me know if you want me to try implementing this

operator/slotticker/slotticker.go
```diff
@@ -13,8 +13,16 @@ import (
...
+// SlotTicker provides a way to keep track of Ethereum slots as they change over time.
+// Note, the caller is RESPONSIBLE for calling NextTick method periodically in order for
+// SlotTicker to advance forward (to keep ticking) to newer slots.
 type SlotTicker interface {
-	Next() <-chan time.Time
+	// NextTick returns a channel that will relay 1 tick signalling that "freshest" slot has started.
+	// It advances slot number SlotTicker keeps track of (potentially jumping several slots ahead)
+	// and returns a channel that will signal once the time corresponding to that "freshest" slot
+	// comes.
+	NextTick() <-chan time.Time
```
y0sher, line 24: nit pick but IMO `Next` was a better name. ticking is already mentioned in the iface name and file.

operator/duties/scheduler.go
```diff
@@ -293,13 +293,15 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
 
-				// Lock the mutex before broadcasting
-				s.waitCond.L.Lock()
+			s.waitCond.L.Lock()
+			// we only want to increase s.headSlot (and never decrease it)
+			if slot > s.headSlot {
 				s.headSlot = slot
-				s.waitCond.Broadcast()
-				s.waitCond.L.Unlock()
 			}
```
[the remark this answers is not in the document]
iurii-ssv: I guess you are right - there is no reason to `Broadcast` when we haven't updated `s.headSlot` (if that's what you mean), 

I've updated that behavior now https://github.com/ssvlabs/ssv/pull/1860/commits/46354997c1770eb8a6bcb0bfaa3a0d963299fb55

operator/slotticker/slotticker.go
iurii-ssv: Renamed back to `Next`, but what do you think of [simplifying SlotTicker API](https://github.com/ssvlabs/ssv/pull/1860#discussion_r2020138141) ?

operator/slotticker/slotticker.go
y0sher: I think we can start thinking on a better implementation on a future PR, getting some insights from other implementations. I agree looping and defining a timeout to read a slot from the channel might also work..

operator/slotticker/slotticker.go
y0sher: I think we can create a different task and PR to handle this specifically in the future.

operator/duties/scheduler.go
y0sher: I'm not sure what you asked here, we never get a slot from someone else in the network in duty scheduler, it only cares about ours and the BN clock.

operator/slotticker/slotticker.go
y0sher: Honestly I like the "goroutine-less" design. but I think we can consider something else, if we want to change then it should be something well tested and benchmarked to show the change has an advantage over current design.

operator/duties/scheduler.go
```diff
@@ -287,19 +292,21 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
-				s.waitCond.L.Lock()
+			s.waitCond.L.Lock()
+			// we only want to increase s.headSlot (and never decrease it) and notify the go-routines
+			// waiting for it to happen
+			if slot > s.headSlot {
 				s.headSlot = slot
 				s.waitCond.Broadcast()
-				s.waitCond.L.Unlock()
 			}
+			s.waitCond.L.Unlock()
```
y0sher, line 309: since this looks exactly like the code in  `HandleHeadEvent` I think it makes sense to create a function for it.

operator/slotticker/slotticker.go
iurii-ssv: Yep, https://github.com/ssvlabs/ssv/issues/2200

operator/duties/scheduler.go
iurii-ssv: https://github.com/ssvlabs/ssv/pull/1860/commits/950a614c4514bbd1e664d0e999f2c4a27b7a3969

operator/duties/scheduler.go
iurii-ssv: Never mind, essentially I "was worried" that event might be delayed somewhat such that by the time we receive it `event.Slot` will be equal to `s.network.Beacon.EstimatedCurrentSlot()+1` (or even `s.network.Beacon.EstimatedCurrentSlot()+N`)

and so we would skip such an event - is it fine to just skip an event ? Often events are meant to be applied one-after-another without any skips, but from what I can tell - here it seems fine to skip since all we care about it spotting a reorg. So I'm reverting this part of the change as unnecessary (and it's incomplete solution anyway).

y0sher (approved):

PR #1867, opened 2024-11-21
exporter: save participants' indices in committee instead of operator IDs

nkryuchkov:
Pyroscope shows that updating participants allocates a lot of memory and grows quickly. The PR changes the format of saved participants from the list of operator IDs converted to slice (8 byte each, 32-104 bytes total) to `uint16` (2 bytes) bitmask of indices without memory allocations

image

### DIFF
--- a/api/handlers/exporter.go
+++ b/api/handlers/exporter.go
@@ -12,12 +12,14 @@ import (
 	"github.com/ssvlabs/ssv/exporter/convert"
 	ibftstorage "github.com/ssvlabs/ssv/ibft/storage"
 	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
+	registrystorage "github.com/ssvlabs/ssv/registry/storage"
 	"github.com/ssvlabs/ssv/utils/casts"
 )
 
 type Exporter struct {
 	DomainType spectypes.DomainType
 	QBFTStores *ibftstorage.QBFTStores
+	Shares     registrystorage.Shares
 }
 
 type ParticipantResponse struct {
@@ -79,7 +81,17 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
 			from := phase0.Slot(request.From)
 			to := phase0.Slot(request.To)
 
-			participantsList, err := qbftStore.GetParticipantsInRange(msgID, from, to)
+			share, ok := e.Shares.Get(nil, pubKey)
+			if !ok {
+				return api.Error(fmt.Errorf("share not found"))
+			}
+
+			operatorIDs := make([]spectypes.OperatorID, 0, len(share.Committee))
+			for _, sm := range share.Committee {
+				operatorIDs = append(operatorIDs, sm.Signer)
+			}
+
+			participantsList, err := qbftStore.GetParticipantsInRange(msgID, from, to, operatorIDs)
 			if err != nil {
 				return api.Error(fmt.Errorf("error getting participants: %w", err))
 			}
--- a/eth/eventhandler/validation.go
+++ b/eth/eventhandler/validation.go
@@ -12,12 +12,12 @@ import (
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
-const maxOperators = 13
+const maxCommitteeSize = 13 // TODO: define in ssv network config
 
 func (eh *EventHandler) validateOperators(txn basedb.Txn, operators []uint64) error {
 	operatorCount := len(operators)
 
-	if operatorCount > maxOperators {
+	if operatorCount > maxCommitteeSize {
 		return fmt.Errorf("too many operators (%d)", operatorCount)
 	}
 
--- a/ibft/storage/store.go
+++ b/ibft/storage/store.go
@@ -3,7 +3,6 @@ package storage
 import (
 	"encoding/binary"
 	"fmt"
-	"slices"
 	"sync"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
@@ -13,6 +12,7 @@ import (
 	"go.uber.org/zap"
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
+
 	"github.com/ssvlabs/ssv/exporter/convert"
 	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
 	"github.com/ssvlabs/ssv/storage/basedb"
@@ -70,24 +70,24 @@ func (i *ibftStorage) CleanAllInstances(msgID []byte) error {
 	return nil
 }
 
-func (i *ibftStorage) UpdateParticipants(identifier convert.MessageID, slot phase0.Slot, newParticipants []spectypes.OperatorID) (updated bool, err error) {
+func (i *ibftStorage) UpdateParticipants(identifier convert.MessageID, slot phase0.Slot, newParticipants qbftstorage.Quorum) (updated bool, err error) {
 	i.participantsMu.Lock()
 	defer i.participantsMu.Unlock()
 
 	txn := i.db.Begin()
 	defer txn.Discard()
 
-	existingParticipants, err := i.getParticipants(txn, identifier, slot)
+	existing, err := i.getParticipantsBitMask(txn, identifier, slot)
 	if err != nil {
 		return false, fmt.Errorf("get participants %w", err)
 	}
 
-	mergedParticipants := mergeParticipants(existingParticipants, newParticipants)
-	if slices.Equal(mergedParticipants, existingParticipants) {
+	merged := mergeParticipants(existing, newParticipants.ToBitMask())
+	if merged == existing {
 		return false, nil
 	}
 
-	if err := i.saveParticipants(txn, identifier, slot, mergedParticipants); err != nil {
+	if err := i.saveParticipantsBitMask(txn, identifier, slot, merged); err != nil {
 		return false, fmt.Errorf("save participants: %w", err)
 	}
 
@@ -98,11 +98,16 @@ func (i *ibftStorage) UpdateParticipants(identifier convert.MessageID, slot phas
 	return true, nil
 }
 
-func (i *ibftStorage) GetParticipantsInRange(identifier convert.MessageID, from, to phase0.Slot) ([]qbftstorage.ParticipantsRangeEntry, error) {
+func (i *ibftStorage) GetParticipantsInRange(
+	identifier convert.MessageID,
+	from,
+	to phase0.Slot,
+	committee []spectypes.OperatorID,
+) ([]qbftstorage.ParticipantsRangeEntry, error) {
 	participantsRange := make([]qbftstorage.ParticipantsRangeEntry, 0)
 
 	for slot := from; slot <= to; slot++ {
-		participants, err := i.GetParticipants(identifier, slot)
+		participants, err := i.GetParticipants(identifier, slot, committee)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get participants: %w", err)
 		}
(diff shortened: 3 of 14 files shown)


protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -2,14 +2,18 @@ package qbftstorage
...
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
+
 	"github.com/ssvlabs/ssv/exporter/convert"
 )
 
+const maxCommitteeSize = 13 // TODO: define in ssv network config
+
 // StoredInstance contains instance state alongside with a decided message (aggregated commits).
 type StoredInstance struct {
```
iurii-ssv, line 18: Not exactly relevant to this PR, but I thought we no longer use this struct after https://github.com/ssvlabs/ssv/pull/1767 (except for in related tests, but these should also be adjusted accordingly then) ? Perhaps can remove while at it.

protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -32,17 +36,63 @@ type ParticipantsRangeEntry struct {
...
+			i++
+		} else { // A[i] > B[j]
+			j++
+		}
+	}
+
+	return bitmask
+}
+
+type OperatorsBitMask uint16
```
iurii-ssv, line 67: Would be nice to describe the format in details (what bits correspond to what), in other words
- what this bit mask is
- how it's meant to be constructed

from what I've gathered, it's meant to "encode" only operators that have signed, if so - perhaps `SignersBitMask` would be a more descriptive name

protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -32,17 +36,63 @@ type ParticipantsRangeEntry struct {
 	Identifier convert.MessageID
 }
 
+type Quorum struct {
+	Signers     []spectypes.OperatorID
+	Committee   []spectypes.OperatorID
+	ValidatorPK spectypes.ValidatorPK // optional payload
+}
+
+func (q *Quorum) ToBitMask() OperatorsBitMask {
```
iurii-ssv, line 45: Perhaps `ToBitMask` -> `ToSignersBitMask` would be a bit more informative

protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -32,17 +36,63 @@ type ParticipantsRangeEntry struct {
 	Identifier convert.MessageID
 }
 
+type Quorum struct {
+	Signers     []spectypes.OperatorID
+	Committee   []spectypes.OperatorID
+	ValidatorPK spectypes.ValidatorPK // optional payload
+}
```
iurii-ssv, line 43: It doesn't seem like `qbftstorage` package is the best place to define `Quorum` struct in, wouldn't `v2/validator` be a better place for it ?

For example, `quorum.ValidatorPK` is never even accessed/needed inside of `qbftstorage` package, this kind of implies `Quorum` abstraction isn't the best fit for what we are trying to do here, and instead feels like going with a dumb "bitmask" + decoder/encoder for it is just simpler way to do it (might not be as elegant, but probably is more straightforward to debug and deal with).

iurii-ssv (approved):
Interesting way to go about DB migration! Left couple of suggestions.

protocol/v2/qbft/storage/ibft_store.go
nkryuchkov: It's used in `TestHandleDecidedQuery` but honestly I didn't check if we need it. Let's leave this for another PR as this one is a hotfix and we need to merge it ASAP

protocol/v2/qbft/storage/ibft_store.go
nkryuchkov: I agree that the structure is not the best, but moving this would cause cyclic import because the `QBFTStore` interface uses this type. I don't like the current type relations between these packages but it will probably be easier to resolve them later. 

I don't like the existence of `ValidatorPK` inside `Quorum` either, I just didn't want to create a new type for it, but I likely should because we already have `validatorIndexAndRoot`. Done

ibft/storage/store.go
```diff
@@ -70,25 +72,37 @@ func (i *ibftStorage) CleanAllInstances(msgID []byte) error {
...
+		}
+
+		quorumAdapter := qbftstorage.Quorum{
+			Signers:   existingList,
+			Committee: newParticipants.Committee,
+		}
+
+		existing = quorumAdapter.ToSignersBitMask()
+	} else if err != nil {
+		return false, fmt.Errorf("get participants bitmask: %w", err)
```
y0sher, line 96: It's worth to mention is not backward compatible, once you update with this in your code, you cannot use your db with the old code. And it's not straight forward to get this because we're not doing any migration.
It might be smart to do something similar to what we did with the alan release and just fail if the version is older. @moshe-blox wdyt?

protocol/v2/qbft/storage/ibft_store.go
y0sher: @nkryuchkov @Tom-ssvlabs lets make sure we have ticket -

> removing unused instance storage structs (`protocol/v2/qbft/storage/ibft_store.go`)

protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -32,17 +36,68 @@ type ParticipantsRangeEntry struct {
...
 }
 
+type Quorum struct {
+	Signers   []spectypes.OperatorID
+	Committee []spectypes.OperatorID
+}
+
+func (q *Quorum) ToSignersBitMask() SignersBitMask {
+	if len(q.Committee) > maxCommitteeSize || len(q.Signers) > maxCommitteeSize || len(q.Signers) > len(q.Committee) {
+		panic(fmt.Sprintf("invalid signers/quorum size: %d/%d", len(q.Committee), len(q.Signers)))
```
y0sher, line 46: we should probably return error instead of panicking, especially in an API triggered code.

protocol/v2/qbft/storage/ibft_store.go
Tom-ssvlabs: https://blox-ensemble.monday.com/boards/7946005181/pulses/7957409842

protocol/v2/qbft/storage/ibft_store.go
```diff
@@ -32,17 +36,68 @@ type ParticipantsRangeEntry struct {
...
+// As the maximal supported operator count is 13, it needs to be at least 13 bits long.
+// The starting bit is the least significant bit, so it represents the first operator in the committee.
+//
+// Example:
+// If committee is [1,2,3,4] and SignersBitMask is 0b0000_0000_0000_1101, it means quorum of [1,3,4].
+type SignersBitMask uint16
+
+func (obm SignersBitMask) Signers(committee []spectypes.OperatorID) []spectypes.OperatorID {
+	if len(committee) > maxCommitteeSize {
+		panic(fmt.Sprintf("invalid committee size: %d", len(committee)))
```
y0sher, line 76: ditto

y0sher (changes requested):
small changes

api/handlers/exporter.go
```diff
@@ -79,7 +81,17 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
-			participantsList, err := qbftStore.GetParticipantsInRange(msgID, from, to)
+			share, ok := e.Shares.Get(nil, pubKey)
+			if !ok {
+				return api.Error(fmt.Errorf("share not found"))
+			}
+
+			operatorIDs := make([]spectypes.OperatorID, 0, len(share.Committee))
+			for _, sm := range share.Committee {
+				operatorIDs = append(operatorIDs, sm.Signer)
+			}
```
moshe-blox, line 92: this would be wrong if the validator has moved between committees or missing if it was removed

moshe-blox (changes requested):

api/handlers/exporter.go
moshe-blox: there's probably a similar problem today when searching by operator or committee, however AFAIK everyone is still searching by pubkey/index like before the fork

nkryuchkov: @moshe-blox added storing committees by slot intervals, please review

protocol/v2/qbft/storage/ibft_store.go
nkryuchkov: @y0sher but this code should be unreachable and means a bug in our code:
- committee length cannot be more than 13
- signers length cannot be more than 13
- signers length cannot exceed committee length

ibft/storage/store.go
nkryuchkov: I'm not sure that failing if the version is older fits here because the new format is compatible with the old format. The one that's incompatible is the old code with the new format but we aren't going to change the old code. I left a comment

eth/eventhandler/validation.go
```diff
@@ -12,12 +12,12 @@ import (
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
-const maxOperators = 13
+const maxCommitteeSize = 13 // TODO: define in ssv network config
```
MatusKysel: fix this todo

eth/eventhandler/validation.go
```diff
@@ -12,12 +12,12 @@ import (
...
 
 func (eh *EventHandler) validateOperators(txn basedb.Txn, operators []uint64) error {
 	operatorCount := len(operators)
 
-	if operatorCount > maxOperators {
+	if operatorCount > maxCommitteeSize {
```
oleg-ssvlabs: Shouldn’t there be a better validation for the committee size?

Something like:

`n >= 4 && n <= 13 && n%3 == 1`

eth/eventhandler/validation.go
oleg-ssvlabs: I agree, we could either declare a new constant in `networkconfig` package or (following my comment below) a function

ibft/storage/store.go
```diff
@@ -12,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
```
oleg-ssvlabs: We could use the stdlib `maps` package instead of the experimental one.

There are some differences, but you can replace the usage like this:

`slices.Collect(maps.Keys(map))`

protocol/v2/qbft/storage/participant_store.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/ssvlabs/ssv/operator/slotticker"
 )
 
+const maxCommitteeSize = 13 // TODO: define in ssv network config
```
oleg-ssvlabs: same here. It would be good to have in a single place

protocol/v2/qbft/storage/quorum.go
```diff
@@ -0,0 +1,69 @@
...
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+)
+
+type Quorum struct {
+	Signers   []spectypes.OperatorID
+	Committee []spectypes.OperatorID
+}
+
+func NewQuorum(signers, committee []spectypes.OperatorID) (Quorum, error) {
+	if len(committee) > maxCommitteeSize || len(signers) > maxCommitteeSize || len(signers) > len(committee) {
```
oleg-ssvlabs: maybe we could wrap this validation in some function, because we repeat it [here](https://github.com/ssvlabs/ssv/pull/1867/files#diff-53337af20dd7f261486671a25341d42b8718571d6b1c42b0da9fe8f95bd69aafR26)

oleg-ssvlabs (commented):
great job!
Left few comments

protocol/v2/qbft/storage/quorum.go
nkryuchkov: Actually, I think we don't need it there, just here. Removed the second check.

eth/eventhandler/validation.go
nkryuchkov: We do it in `ssvtypes.ValidCommitteeSize`

protocol/v2/qbft/storage/quorum.go
nkryuchkov: I changed my mind and added the validation method as you requested

nkryuchkov:
@greptileai please review

nkryuchkov:
@greptileai what could increase the confidence score?

PR #1869, opened 2024-11-22
feat(metrics): OTeL - instrument Consensus Client

oleg-ssvlabs:
**Metrics added**
- `ssv.cl.attestation_data_request.duration` (`ssv.beacon.role` attribute(global))
- `ssv.cl.sync.status` (`ssv.cl.addr`, `ssv.cl.sync.status` attributes)
- `ssv.cl.sync.distance` (`ssv.cl.addr` attribute) **(new!)**

**Metrics removed**
- `ssv_beacon_status`
- `ssv_beacon_data_request_duration_seconds` (`role` label)

**Questions and opinions**
- `ssv.cl.syncing` namespace name. Although, it is aligned with this [PR](https://github.com/ssvlabs/ssv/pull/1852). The alternative could be `ssv.cl.sync` (`ssv.cl.sync.status`).
- `ssv.beacon.role` attribute name. Alternatives: `ssv.consensus.role`, `ssv.validator.role`. What does sound more correct to you? Keep in mind that OTeL metric attributes also have a concept of _namespaces_, so in this case `ssv.consensus`, `ssv.beacon`, `ssv.validator` - are namespaces, which means they may contain more attributes.

NOTE: temporarily target `otel-metrics` branch until it is merged to `stage`

### DIFF
--- a/beacon/goclient/observability.go
+++ b/beacon/goclient/observability.go
@@ -1,0 +1,85 @@
+package goclient
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv/observability"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type beaconNodeStatus string
+
+const (
+	observabilityName      = "github.com/ssvlabs/ssv/beacon/goclient"
+	observabilityNamespace = "ssv.cl"
+
+	statusUnknown beaconNodeStatus = "unknown"
+	statusSyncing beaconNodeStatus = "syncing"
+	statusSynced  beaconNodeStatus = "synced"
+)
+
+var (
+	meter = otel.Meter(observabilityName)
+
+	attestationDataRequestHistogram = observability.NewMetric(
+		meter.Float64Histogram(
+			metricName("attestation_data_request.duration"),
+			metric.WithUnit("s"),
+			metric.WithDescription("beacon data request duration in seconds"),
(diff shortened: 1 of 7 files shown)


beacon/goclient/observability.go
```diff
@@ -0,0 +1,76 @@
...
+
+type beaconNodeStatus string
+
+const (
+	observabilityName      = "github.com/ssvlabs/ssv/beacon/goclient"
+	observabilityNamespace = "ssv.cl"
+
+	statusUnknown beaconNodeStatus = "unknown"
+	statusSyncing beaconNodeStatus = "syncing"
+	statusOK      beaconNodeStatus = "ok"
```
zktaiga, line 21: I'd change this to `synced` since it's really the only thing we know.

zktaiga (approved):
> ssv.cl.syncing namespace name. Although, it is aligned with this https://github.com/ssvlabs/ssv/pull/1852. The alternative could be ssv.cl.sync (ssv.cl.sync.status).

`ssv.cl.sync.status` sounds more correct to me; it can describe multiple states, whereas syncing is implying the state in its name.
> ssv.beacon.role attribute name. Alternatives: ssv.consensus.role, ssv.validator.role. What does sound more correct to you? Keep in mind that OTeL metric attributes also have a concept of namespaces, so in this case ssv.consensus, ssv.beacon, ssv.validator - are namespaces, which means they may contain more attributes.

`ssv.beacon.role` sounds more correct to me as it adheres to [the spec](https://github.com/ssvlabs/ssv-spec/blob/a2e45b9fb40651e951110f13eb77f749aeddf2f4/types/beacon_types.go#L56-L75).

beacon/goclient/observability.go
oleg-ssvlabs: fixed

PR #1870, opened 2024-11-22
feat(metrics): OTeL - instrument P2P Topics component

oleg-ssvlabs:


zktaiga (approved):
> Metric/Namespace naming - ssv.p2p.messages. I used the plural noun messages for metric names because these are cumulative (only increasing) counters.

LGTM.
> The attribute inherits this name, but I think it would be more accurate to use singular forms for attributes, like ssv.p2p.messages.type. Wdyt?

Agree!

PR #1876, opened 2024-11-26
feat(metrics): OTeL - instrument P2P Message Validation component

oleg-ssvlabs:
**Metrics added**
- `ssv.p2p.message.validations.accepted` (`ssv.runner.role` attribute)
- `ssv.p2p.message.validations.ignored` (`ssv.runner.role`, `ssv.p2p.message.validation.discard_reason` attributes)
- `ssv.p2p.message.validations.rejected` (`ssv.runner.role`, `ssv.p2p.message.validation.discard_reason` attributes)
- `ssv.p2p.message.validations.duration` 
- `ssv.p2p.message.validations`

**Metrics removed**
- `ssv:p2p:pubsub:msg:val:active` (`topic` label)
- `ssv_message_validation_consensus_type` (`type`, `signers` label)
- `ssv_message_validation` (`status`, `reason`, `role`, `round` labels)
- `ssv_message_validation_ssv_type` (`type`)
- `ssv_messages_received_from_peer`(`peer_id`)
- `ssv_messages_received_total` (-)
- `ssv_message_size` (-)
- `ssv_message_validation_duration_seconds` (-)

**Questions and Opinions**
- **New metrics structure alternative**: The alternative structure would be having a single metric `ssv.p2p.message.validations` with `accepted`, `ignored`, and `rejected` set as a status label. I decided to declare 3 separate metrics instead, mainly because in most cases, we use this approach in other metrics. Additionally, depending on the status, there is a different set of labels (e.g., `ssv.p2p.message.validation.discard_reason` for failed validations).

- **Attributes namespaces**: This question has a broader scope than the given PR.  
  Quote: _"Note: the fact that an entity is located within another entity is typically not a sufficient reason for forming nested namespaces. The purpose of a namespace is to avoid name clashes, not to indicate entity hierarchies"_ ([source](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/)).  
  In our case, for attribute names, we clearly use namespaces to indicate a hierarchy of entities. However, isn't this the easiest way to achieve descriptive names while also avoiding naming collisions? It's simple to follow such convention.

- **Metric removals**: **Some** metrics were removed not because they are low-value or poor-quality metrics but because they do not belong to the _message validation_ domain. Metrics like:
  - `ssv_message_size` (should be implemented within P2P Topics since that's where these messages are routed from for validation ([PR](https://github.com/ssvlabs/ssv/pull/1870))).
  - `ssv_messages_received_total` (should be implemented within P2P Topics).
  - `ssv_message_validation_consensus_type` (should rather be an attribute on the `ssv_messages_received_total` metric).
  - `ssv_message_validation_ssv_type` (should rather be an attribute on the `ssv_messages_received_total` metric).


zktaiga (commented):
- Do we assume that the old `ssv_messages_received_total` is now split as a sum of three different metrics being `accepted + ignored + rejected`? If so, I think it could be worth having a `total`, it'd make it easier to plot on a graph and calculate percentages.

- I think it would be worth keeping duration in some form. It could help spot bottlenecks (or even attacks), as well as let us see if over a long period of time, network size has any impact on the time it takes to validate messages. I guess this could maybe apply to size as well. WDYT?

>New metrics structure alternative: The alternative structure would be having a single metric ssv.p2p.message.validations with accepted, ignored, and rejected set as a status label. I decided to declare 3 separate metrics instead, mainly because in most cases, we use this approach in other metrics. Additionally, depending on the status, there is a different set of labels (e.g., ssv.p2p.message.validation.discard_reason for failed validations).

LGTM as three different metrics (plus it's a common industry pattern).

> In our case, for attribute names, we clearly use namespaces to indicate a hierarchy of entities. However, isn't this the easiest way to achieve descriptive names while also avoiding naming collisions? It's simple to follow such convention.

I think it's fine, it's good for discoverability at the visualization layer. It is possible to search for metrics with only an attribute name. Having business-logic baked into the attributes (over something really vague) means we will be able to see which metrics have the same named attribute across different systems. On the other hand, if we only had `role` or `type`, we'd find metrics which are not even part of SSV which IMO is highly misleading.

> Metric removals: Some metrics were removed not because they are low-value or poor-quality metrics but because they do not belong to the message validation domain. 

`ssv_message_size`: Comment above about the size can be ignored then :smile: 

`ssv_messages_received_total`: Maybe my comment above still applies? Yes, maybe received messages, _**now**_, are always equal to the amount of messages validated but it would feel a bit weird to have a split of `accepted` `ignored` `rejected` but comparing against a metric that resides in another subsystem. The reason this is problematic is because arithmetic in Prometheus is based on labels which puts you in this weird position (imagine each metric having a different `subsystem` label): `ssv_p2p_message_validations_accepted / ssv_p2ptopics_messages_received_total * 100 ignoring(subsystem)`

`ssv_message_validation_consensus_type`: :+1: 

`ssv_message_validation_ssv_type`: :+1:

oleg-ssvlabs:
> * Do we assume that the old `ssv_messages_received_total` is now split as a sum of three different metrics being `accepted + ignored + rejected`? If so, I think it could be worth having a `total`, it'd make it easier to plot on a graph and calculate percentages.

I see your point and I agree. I added `ssv.p2p.message.validations` metric which indicates the total number of validations performed, so when it needs to calculate the ratio it would be possible to do `ssv.p2p.message.validations.accepted / ssv.p2p.message.validations`, instead of `ssv.p2p.message.validations.accepted / (ssv.p2p.message.validations.accepted + ssv.p2p.message.validations.rejected + ssv.p2p.message.validations.ignored)`

> * I think it would be worth keeping duration in some form. It could help spot bottlenecks (or even attacks), as well as let us see if over a long period of time, network size has any impact on the time it takes to validate messages. I guess this could maybe apply to size as well. WDYT?

The validation of incoming messages is a very simple process, primarily consisting of decoding and validating the message size. As a result, there are no potential outliers, and the process takes less than a millisecond. Do you still think this metric is necessary? If the validation were more complex, for example, dependent on the input type or similar factors, I would certainly agree to keep it

PR #1881, opened 2024-11-26
feat(metrics): OTeL - instrument P2P component

oleg-ssvlabs:
**Metrics added**
-  `ssv.p2p.peers.connected`
- `ssv.p2p.peers.per_topic` (`ssv.p2p.topic.name` attribute)
- `ssv.p2p.peers.per_version` (`ssv.node.version` attribute)

**Metrics removed**
- `ssv_p2p_all_connected_peers`
- `ssv_p2p_connected_peers` (`pubKey` label)
- `ssv:network:peers_identity` (`pubKey`, `operatorID`, `v`, `pid`, `type` labels)
- `ssv:network:router:in` (`mt` label)


zktaiga (changes requested):
As previously discussed, I think this change needs to answer two questions:
- What is the version distribution for the peers we are connected to?
  - This will require a reset.
  - No need to keep pid in memory.
- What is the version distribution for all peers we have ever observed?
  - Does not need a reset, will also help us calculate rates.
  - We need to find some way to not double-count, so it probably will not involve iterating over connected peers, but rather increasing a counter when we connect to a peer, because we want to account for peers that we connect to even for a short period of time.

zktaiga (approved):

PR #1883, opened 2024-11-26
Exporter race fix

olegshmuelov:


operator/validator/controller.go
```diff
@@ -448,57 +449,74 @@ func (c *controller) handleRouterMessages() {
...
 }
 
-func (c *controller) handleNonCommitteeMessages(msg *queue.SSVMessage, ncv *committeeObserver) error {
+func (c *controller) getCommitteeObserver(ssvMsg *queue.SSVMessage) *committeeObserver {
 	c.committeesObserversMutex.Lock()
 	defer c.committeesObserversMutex.Unlock()
 
-	if msg.MsgType == spectypes.SSVConsensusMsgType {
+	// Check if the observer already exists
+	existingObserver := c.committeesObservers.Get(ssvMsg.GetID())
```
y0sher, line 476: isn't `ttlcache` already thread safe? why we need the mutex here then?

olegshmuelov: Without the mutex, there’s a risk of race conditions where multiple goroutines might attempt to create and set the same observer simultaneously.

y0sher: this can be also solved by the cache's `GetOrSet` func, but involves creating the observer each time and sometimes for nothing. though it will be GC'd quickly if not used, there is no goroutine in the creation.

olegshmuelov: You're right that `GetOrSet` avoids race conditions without a mutex, but it can lead to unnecessary observer creation, even if quickly garbage collected. The current mutex ensures observers are only created when needed, minimizing overhead for non-trivial operations. If the cost of unused creation is negligible, switching to `GetOrSet` could simplify the code. Let me know if we’d like to explore this direction further.

operator/validator/controller.go
```diff
@@ -365,57 +365,74 @@ func (c *controller) handleRouterMessages() {
...
 
-var nonCommitteeValidatorTTLs = map[spectypes.RunnerRole]int{
-	spectypes.RoleCommittee:  64,
-	spectypes.RoleProposer:   4,
-	spectypes.RoleAggregator: 4,
-	//spectypes.BNRoleSyncCommittee:             4,
+var committeeObserverValidatorTTLs = map[spectypes.RunnerRole]int{
+	spectypes.RoleCommittee:                 64,
+	spectypes.RoleProposer:                  4,
+	spectypes.RoleAggregator:                4,
```
iurii-ssv, line 371: Maybe clarify that's denominated in **slots** (and not seconds or something)

operator/validator/controller.go
```diff
@@ -365,57 +365,74 @@ func (c *controller) handleRouterMessages() {
...
+		c.calculateObserverTTL(ssvMsg.MsgID.GetRoleType()),
+	)
+
+	return newObserver
+}
+
+func (c *controller) calculateObserverTTL(roleType spectypes.RunnerRole) time.Duration {
+	ttlSlots := committeeObserverValidatorTTLs[roleType]
+	return time.Duration(ttlSlots) * c.beacon.GetBeaconNetwork().SlotDurationSec()
+}
```
iurii-ssv, line 428: I think it's probably better to "panic" than "default to 0" in case there is a new type introduced and we forgot to add it to `committeeObserverValidatorTTLs` map (so we can discover this issue early)

iurii-ssv (approved):
LGTM (with minor comments)

PR #1884, opened 2024-11-27
feat(metrics): OTeL - instrument Validator component

oleg-ssvlabs:


operator/validator/observability.go
```diff
@@ -0,0 +1,76 @@
...
+		statusSlashed,
+		statusExiting,
+		statusNotActivated,
+		statusPending,
+		statusNoIndex,
+		statusRemoved,
+		statusUnknown} {
+		validatorStatusGauge.Record(ctx, 0,
+			metric.WithAttributes(validatorStatusAddrAttribute(status)),
+		)
```
zktaiga, line 74: In_ OTel, since we don't have some equivalent for `Reset()`, do you think recording `0` is the most adequate to signal that we don't have a value for a given metric, vs something such as `-1`? 

Since all of these statuses theoretically cover all possible states for a validator, I'm wondering if it's worth tracking and actually using the gauges to increase and decrease rather than resetting and counting every time, is it possible to verify that the final behavior is the same from a visualization perspective? e.g. you can see validators transitioning from one state to another.

zktaiga (approved):

zktaiga (approved):

nkryuchkov:
Hi @oleg-ssvlabs @zktaiga, this PR breaks the spec alignment script. Could you please sync with @moshe-blox on how to fix it and create a PR with the fix? Thanks

oleg-ssvlabs:
> Hi @oleg-ssvlabs @zktaiga, this PR breaks the spec alignment script. Could you please sync with @moshe-blox on how to fix it and create a PR with the fix? Thanks

@nkryuchkov absolutely. We are aware of this, which is why this PR was merged into a feature branch. That feature branch will be carefully reviewed before being merged into either stage or main, so the spec deviations will be resolved. Here is the [PR](https://github.com/ssvlabs/ssv/pull/1852) for the feature branch we are currently merging into.

nkryuchkov:
> > Hi @oleg-ssvlabs @zktaiga, this PR breaks the spec alignment script. Could you please sync with @moshe-blox on how to fix it and create a PR with the fix? Thanks
> 
> @nkryuchkov absolutely. We are aware of this, which is why this PR was merged into a feature branch. That feature branch will be carefully reviewed before being merged into either stage or main, so the spec deviations will be resolved. Here is the [PR](https://github.com/ssvlabs/ssv/pull/1852) for the feature branch we are currently merging into.

great, thanks for the explanation

Issue #1827, opened 2024-10-29
Missed proposal duty in first slot of the epoch (Rare Case)


olegshmuelov:
In rare cases, proposal duty messages are received before the duty starts.
These messages are rejected with no duty for this epoch since the proposal duties store is not yet populated at that time.

iurii-ssv:
Somewhat related: https://github.com/ssvlabs/ssv/issues/1825

PR #1886, opened 2024-11-27
fix(MessageValidation): first slot proposals [stage]

olegshmuelov:
close https://github.com/ssvlabs/ssv/issues/1827

**Solution:**
Don't ignore first slot proposal pre-consensus messages (randao) as long as the ProposerHandler has not fetched the duties.

The impact of this change from an attack perspective is minor, as the attacker must be part of the committee (i.e., possess a private key)

### DIFF
--- a/message/validation/validation_test.go
+++ b/message/validation/validation_test.go
@@ -14,6 +14,7 @@ import (
 
 	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/herumi/bls-eth-go-binary/bls"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	pspb "github.com/libp2p/go-libp2p-pubsub/pb"
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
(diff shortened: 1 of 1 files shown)


olegshmuelov:
test TBD

message/validation/partial_validation.go
```diff
@@ -155,7 +155,7 @@ func (mv *messageValidator) validatePartialSigMessagesByDutyLogic(
 		}
 	}
 
-	if err := mv.validateBeaconDuty(signedSSVMessage.SSVMessage.GetID().GetRoleType(), messageSlot, committeeInfo.indices); err != nil {
+	if err := mv.validateBeaconDuty(role, messageSlot, committeeInfo.indices, &partialSignatureMessages.Type); err != nil {
```
y0sher, line 158: following the above comment
```suggestion
	if err := mv.validateBeaconDuty(role, messageSlot, committeeInfo.indices, &partialSignatureMessages.Type == spectypes.RandaoPartialSig); err != nil {
```

message/validation/common_checks.go
```diff
@@ -117,10 +117,20 @@ func (mv *messageValidator) validateBeaconDuty(
 	role spectypes.RunnerRole,
 	slot phase0.Slot,
 	indices []phase0.ValidatorIndex,
+	partialMessageType *spectypes.PartialSigMsgType,
```
y0sher, line 120: I feel this could be just replaced with a boolean? and do the check outside. since you only use it to see if its a randao message. this could just be `randao bool`

y0sher (changes requested):
just a simplification. and we need to ad some test in msg validation code I guess.

nkryuchkov (approved):

Issue #1885, opened 2024-11-27
On validator registration, committee drops entire slot


Roy-blox:
**Describe the bug**
While registering new validators we have a mechanism to not attest on registration epoch + 1 to avoid slashing attacks, in post fork we see that the registration slot drops completely instead of only dropping the newly added validators.

**To Reproduce**
1. Have a post fork cluster with some validators on it
2. Bulk register some validators

**Expected behavior**
Only new validators will not attest on registration epoch while the rest attest as expected

**Actual behavior**
All validators that were supposed to attest on registration slots failed.

**Node (please complete the following information):**
 - Operator Version:  v2.0.0 - Post-fork

PR #1888, opened 2024-11-27
fix(CommitteeRunner): Skip slashable duty [stage]

olegshmuelov:

### DIFF
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -235,7 +235,7 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 			partialMsg, err := cr.BaseRunner.signBeaconObject(cr, duty, attestationData, duty.DutySlot(),
 				spectypes.DomainAttester)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


protocol/v2/ssv/runner/committee.go
```diff
@@ -230,12 +231,13 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 		}
 		switch duty.Type {
 		case spectypes.BNRoleAttester:
-			validDuties++
 			attestationData := constructAttestationData(beaconVote, duty)
 			partialMsg, err := cr.BaseRunner.signBeaconObject(cr, duty, attestationData, duty.DutySlot(),
 				spectypes.DomainAttester)
 			if err != nil {
-				return errors.Wrap(err, "failed signing attestation data")
+				logger.Error("failed signing attestation data for duty",
```
y0sher, line 238: if we know this can happen maybe its a warning? we don't want to freak out operators..

y0sher: I guess more opinions will be appriciated @moshe-blox @MatusKysel @andrew-blox

PR #1890, opened 2024-11-28
hotfix(CommitteeRunner): Skip slashable duty [main]

olegshmuelov:

### DIFF
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -233,7 +233,7 @@ func (cr *CommitteeRunner) ProcessConsensus(logger *zap.Logger, msg *spectypes.S
 			partialMsg, err := cr.BaseRunner.signBeaconObject(cr, duty, attestationData, duty.DutySlot(),
 				spectypes.DomainAttester)
 			if err != nil {
-				logger.Error("failed signing attestation data for duty",
+				logger.Warn("failed signing attestation data for duty, skipping",
 					fields.Validator(duty.PubKey[:]), fields.BeaconRole(duty.Type), zap.Error(err))
 				continue // Skip this duty but continue with others
 			}
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -229,12 +229,13 @@ func (cr *CommitteeRunner) ProcessConsensus(logger *zap.Logger, msg *spectypes.S
 		}
 		switch duty.Type {
 		case spectypes.BNRoleAttester:
-			validDuties++
 			attestationData := constructAttestationData(beaconVote, duty)
 			partialMsg, err := cr.BaseRunner.signBeaconObject(cr, duty, attestationData, duty.DutySlot(),
 				spectypes.DomainAttester)
 			if err != nil {
-				return errors.Wrap(err, "failed signing attestation data")
+				logger.Error("failed signing attestation data for duty",
+					fields.Validator(duty.PubKey[:]), fields.BeaconRole(duty.Type), zap.Error(err))
+				continue // Skip this duty but continue with others
 			}
 			postConsensusMsg.Messages = append(postConsensusMsg.Messages, partialMsg)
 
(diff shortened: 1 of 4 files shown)


protocol/v2/ssv/runner/committee.go
```diff
@@ -250,15 +251,18 @@ func (cr *CommitteeRunner) ProcessConsensus(logger *zap.Logger, msg *spectypes.S
...
 			validDuties++
+		case spectypes.BNRoleSyncCommittee:
 			blockRoot := beaconVote.BlockRoot
 			partialMsg, err := cr.BaseRunner.signBeaconObject(cr, duty, spectypes.SSZBytes(blockRoot[:]), duty.DutySlot(),
 				spectypes.DomainSyncCommittee)
 			if err != nil {
-				return errors.Wrap(err, "failed signing sync committee message")
+				logger.Warn("failed signing sync committee message for duty, skipping",
+					fields.Validator(duty.PubKey[:]), fields.BeaconRole(duty.Type), zap.Error(err))
+				continue // Skip this duty but continue with others
```
moshe-blox, line 262: this is not needed because SC isn't slashable

protocol/v2/ssv/value_check.go
```diff
@@ -68,11 +73,22 @@ func BeaconVoteValueCheckF(
...
+
+		// If all validators are slashable, return an error
+		if len(slashableValidatorPubKeys) == len(sharePublicKeys) {
+			return fmt.Errorf("all validators are slashable, attestation cannot proceed")
+		}
+
+		//If at least one validator is not slashable, return nil
 		return nil
 	}
 }
```
moshe-blox, line 94: @GalRogozinski wdyt about this?
 
i think one of the issues with our solution is that we will proceed to try to do the duty for the slashable validator (and later fail of course) which seems inefficient (which i hope is the only problem)

@liorrutenberg suggested that we should instead just do our own check before the ValueCheck to remove the slashable validators from the list and also from the Runner's duties (so it won't try to do them), however i think that's a bit complex to pull of because ValueCheck happens again with every round change right? (and not only on duty start which is easy for us to filter out slashables there)

protocol/v2/ssv/runner/committee.go
olegshmuelov: I'm aware that SC duties aren't slashable. However, the intention here is to ensure that a single error, regardless of whether it's slashable or not, does not impact the entire committee's duties for the slot. This approach aligns with the goal of isolating errors to specific validators without punishing the entire committee.
lmk if you think we should approach this differently

moshe-blox: if that looks good for @GalRogozinski i have no issues

protocol/v2/ssv/value_check.go
GalRogozinski: Hey,
I think you shouldn't make new validators slashable and that's it.
If you don't want to run new validators in the beginning add a new flag for them.
When we did the spec we were aware of this issue. It was decided to look into it later after Alan fork. I created a Monday task but it never managed to get enough priority.

Slashing should be looked out in an orderly way by the spec team and not with a hasty hotfix

moshe-blox: @GalRogozinski new validators is not the only reason this can happen

we run committee runners for each slot in parallel, so some old runner at a late round might have suddenly one slashable validator because that validator already signed a newer duty in a newer runner

in that case we'd be failing the entire slot for all other validators as well

regardless of the fact that we should indeed just avoid starting validators in the epoch where they registered (which we'll do in a future release), this is still a problem

GalRogozinski: @moshe-blox 
This was known... the reason I agreed to have it done this way is because I was afraid that there may be subtle attack-vectors on the other way. I don't remember the details now. I am also not sure if we should worry about them now.

Slashing is the most crucial part of the system, we shouldn't just touch it this way.

How often this happens w.o new validators? Is it such a huge issue that we can't wait with this?

GalRogozinski (changes requested):
Changes in slashing shouldn't be taken care as hotfix by engineering. I don't know if the current fix is good or not.. but it should be looked at as a dedicated spec task... btw, I had task for this opened on Monday for a long time

protocol/v2/ssv/value_check.go
liorrutenberg: @moshe-blox @GalRogozinski 
New validators and their duty execution are not related to consensus.
If the node decided not to start a duty on registration epoch this should be handled there and not to start consensus for it and check it there

protocol/v2/ssv/runner/committee.go
y0sher: I agree with Oleg here, all slashable or not is actually abstract in this code, so we just want to ensure a single failed signing action doesn't stop all the runner.

protocol/v2/ssv/value_check.go
y0sher: >  @GalRogozinski @olegshmuelov  wdyt about this?
>  
> i think one of the issues with our solution is that we will proceed to try to do the duty for the slashable validator (and later fail of course) which seems inefficient (which i hope is the only problem)
> 
> @liorrutenberg suggested that we should instead just do our own check before the ValueCheck to remove the slashable validators from the list and also from the Runner's duties (so it won't try to do them), however i think that's a bit complex to pull of because ValueCheck happens again with every round change right? (and not only on duty start which is easy for us to filter out slashables there)

I guess we could return the slashable ones and drop them but that changes the whole flow so not suitable for a Hotfix..

y0sher:
> Changes in slashing shouldn't be taken care as hotfix by engineering. I don't know if the current fix is good or not.. but it should be looked at as a dedicated spec task... btw, I had task for this opened on Monday for a long time

What are your suggesting? Leave it out of the hotfix version? I'm not sure we can afford waiting for spec with this bug

PR #1891, opened 2024-11-28
fix(queue): change msg types priorities [main]

MatusKysel:


protocol/v2/ssv/queue/message_prioritizer.go
```diff
@@ -96,7 +96,7 @@ func (p *committeePrioritizer) Prior(a, b *SSVMessage) bool {
 		return scoreA > scoreB
 	}
 
-	scoreA, scoreB = scoreCommitteeConsensusType(a), scoreCommitteeConsensusType(b)
+	scoreA, scoreB = scoreConsensusType(a), scoreConsensusType(b)
```
y0sher, line 99: can you please elaborate on this fix? AFAIK `scoreConsensusType` does scoring on genesis message, I'm surprised how this can work with post-fork messages.

y0sher (changes requested):

protocol/v2/ssv/queue/message_prioritizer.go
y0sher: Oh sorry my ide led me to the genesis version of the func, but now I see we also have post-fork version.
So generally the `scoreCommitteeConsensusType` makes sense, because if there are already commits going on, we want to see the quorum and pass onto post-consensus stage and not wait for proposals, also same about roundchange, if there's a roundchange going on we want to process it asap and processing a prepare/propose/commit might just slow it down. Proposal do seem higher than prepae here, so I'm still not sure how this helps, because the diff between proposal and prepare is still just 1.

y0sher (approved):

moshe-blox (approved):

anatolie-ssv (approved):

PR #1892, opened 2024-11-28
hotfix(MessageValidation): first slot proposals [main]

olegshmuelov:
close https://github.com/ssvlabs/ssv/issues/1827

Solution:
Don't ignore first slot proposal pre-consensus messages (randao) as long as the ProposerHandler has not fetched the duties.

The impact of this change from an attack perspective is minor, as the attacker must be part of the committee (i.e., possess a private key)

duplicate https://github.com/ssvlabs/ssv/pull/1886

### DIFF
--- a/operator/duties/dutystore/duties.go
+++ b/operator/duties/dutystore/duties.go
@@ -96,3 +96,11 @@ func (d *Duties[D]) ResetEpoch(epoch phase0.Epoch) {
 
 	delete(d.m, epoch)
 }
+
+func (d *Duties[D]) IsEpochSet(epoch phase0.Epoch) bool {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	_, exists := d.m[epoch]
+	return exists
+}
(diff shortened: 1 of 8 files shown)


y0sher (approved):
good job!

nkryuchkov (approved):

operator/duties/dutystore/duties.go
```diff
@@ -96,3 +96,11 @@ func (d *Duties[D]) ResetEpoch(epoch phase0.Epoch) {
...
 	delete(d.m, epoch)
 }
+
+func (d *Duties[D]) IsEpochSet(epoch phase0.Epoch) bool {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	_, exists := d.m[epoch]
+	return exists
+}
```
moshe-blox, line 106: this is risky because there are conditions where this will never be set, such as previous epoch, when failing to fetch duties or when (wrongly) detecting 0 active validators in the network

moshe-blox: maybe we could ignore the previous epoch issue if we delete epoch-2 instead of epoch-1 in the handlers (should be ok since msg validation should stop msgs for those epochs for 'too late')

and just ensure epoch is always set, even if to empty, after handler processed slot (despite failures)

wdyt?

moshe-blox (approved):

operator/duties/dutystore/duties.go
moshe-blox: solved

PR #1896, opened 2024-11-29
p2p: dead/solo subnets investigation

iurii-ssv:


network/p2p/p2p.go
```diff
@@ -357,9 +357,9 @@ func (n *p2pNetwork) PeerProtection(allPeers []peer.ID, mySubnets records.Subnet
 				aShares := len(records.SharedSubnets(n.activeSubnets, n.idx.GetPeerSubnets(a), 0))
 				bShared := len(records.SharedSubnets(n.activeSubnets, n.idx.GetPeerSubnets(b), 0))
 				if aShares < bShared {
-					return -1
-				} else if aShares > bShared {
 					return 1
+				} else if aShares > bShared {
```
nkryuchkov, line 361: while we are touching this, let's rename `aShares` to `aShared`

moshe-blox:
@iurii-ssv @y0sher 

Summarizing our convo:

- This is a good direction we'd like to explore
- Being too selective in outgoing/incoming connections is an attack vector, because its easy to fake subnets in discovery, so this change likely requires us to validate subnets claims by peers (@y0sher)
- In order to not stray too far off what we know from Ethereum clients (which dont seem to filter much if at all), we should consider being less harsh in filtering incoming connections, and instead focus only on outgoing and proper balancing to cut off the excess

iurii-ssv:
> focus only on being selective in outgoing conns and rely on proper balancing to cut off the excess (unless anyone can point to why very selective incoming isn't really a risk)

Relevant note, I think at the bare minimum we probably want to have 2 separate `maxPeers` limit per each of connection-direction (incoming/outgoing), as opposed to a single shared one - because what I observed in my testing is we would get to `maxPeers` **just from incoming** connections very fast (and keep getting more and more of them), and hence we won't be able to open much of new outgoing connections (discovery suggest). Even though we trim ~2 connections per 30s or so, it seems new/pending incoming connections quite often take those 2 slots for themselves.

iurii-ssv:
Closing since https://github.com/ssvlabs/ssv/pull/1949 supersedes this one.

PR #1898, opened 2024-12-01
fix: first slot proposals with `MinParticipationEpoch`

moshe-blox:


y0sher (approved):

olegshmuelov:
There might be a potential issue with this approach in certain edge cases. For example:

1. If a validator (e.g., Validator X) is registered at the very first slot of an epoch.
2. Immediately after the registration, there is a restart (e.g., due to a node crash or manual restart).
3. Validator X has a duty at slot n (later in the same epoch) as part of a committee with other validators.
In this scenario, the restart would potentially result in the failure of all committee slots because the state needed to recognize Validator X’s registration might not persist correctly across the restart. This could lead to missed duties for the entire committee

It might be worth considering additional safeguards to handle state persistence during restarts or improving recovery logic to avoid this failure scenario.

PR #1902, opened 2024-12-03
fix: `MinParticipationEpoch` protection [main]

olegshmuelov:

### DIFF
--- a/protocol/v2/types/ssvshare.go
+++ b/protocol/v2/types/ssvshare.go
@@ -29,6 +29,12 @@ type SSVShare struct {
 	Metadata
 
 	committeeID atomic.Pointer[spectypes.CommitteeID]
+
+	// minParticipationEpoch is the epoch at which the validator can start participating.
+	// This is set on registration and on every reactivation.
+	//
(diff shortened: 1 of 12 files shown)


moshe-blox (approved):

protocol/v2/types/ssvshare.go
```diff
@@ -56,6 +62,14 @@ func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
...
 }
 
+func (s *SSVShare) SetMinParticipationEpoch(epoch phase0.Epoch) {
+	s.minParticipationEpoch = epoch
+}
+
+func (s *SSVShare) MinParticipationEpoch() phase0.Epoch {
+	return s.minParticipationEpoch
+}
+
```
y0sher, line 72: This could potentially be a data race, but because other fields in ssvshare look like they have the same issue, and this is really rarely written to, we can leave it like that for now.

@Tom-ssvlabs please open a ticket: `SSVShare thread safety`

y0sher (approved):
Good job! one comment to the future.
also please just fix UT.

Issue #1911, opened 2024-12-04
bug: missed first slot attester duties after indices change at last slot of epoch


olegshmuelov:
When an indices change (e.g., validator registration, reactivation, or removal) occurs in the last slot of an epoch, validators scheduled to attest in the first slot of the next epoch fail to perform their duties.

**Steps to Reproduce:**
1. Trigger an indices change (e.g., validator registration, reactivation, or removal) for a cluster with existing validators at the last slot of an epoch.
2. Ensure the cluster has validators scheduled to attest in the first slot of the next epoch.
3. Observe that the attestation duties for the first slot of the next epoch are not executed.

**Expected Behavior:**
Validators scheduled to attest in the first slot of the next epoch should execute their duties regardless of the timing of the indices change.

**Actual Behavior:**
Validators miss their attestation duties in the first slot of the next epoch if the indices change occurs in the last slot of the preceding epoch.

PR #1912, opened 2024-12-04
fix(DutyScheduler): Handle missed first epoch slot duties after last-slot indices change

olegshmuelov:
Fixes an issue where indices changes (registration, reactivation, or removal) at the last slot of an epoch caused validators to miss their attestation duties for the first slot of the next epoch.

close https://github.com/ssvlabs/ssv/issues/1911


moshe-blox (approved):
LGTM, well tested 👌

y0sher (approved):

PR #1922, opened 2024-12-05
feat(exporter): allow fetch by slot only

anatolie-ssv:


anatolie-ssv:
these changes have been moved to `fix/remove-convert-role-2`

PR #1919, opened 2024-12-05
feat(p2p): limit inbound connections

y0sher:
### Description

This code calculates the amount of inbound connections we have on every new connection and makes sure we're not passing the limit of 80% of our max peers in inbound connections. if we are at the limit we just drop the peer.

PR #1917, opened 2024-12-04
p2p: dead/solo subnets remedy

iurii-ssv:
Porting changes originally made in https://github.com/ssvlabs/ssv/tree/feat/min-subnets-balancer on top of `stage` branch, and further exploring what other improvements (related to dead/solo subnets) can be made.

PR #1923, opened 2024-12-05
feat(p2p): grow MaxPeers with # of subnets

nkryuchkov:
We have noticed public operators with a lot of subnets having hard time to fill peers into all their subnets, having connectivity issues at the end making them miss duties

Since some of them may have configured max peers to the current default value (60), changing the default config value is not enough. Therefore, we need some workaround like this PR

However, we'll need to come up with a more elegant solution later

This PR is meant to be an alternative (simpler) temporary solution in case we decide not to proceed with https://github.com/ssvlabs/ssv/pull/1917 and https://github.com/ssvlabs/ssv/pull/1919 right now

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -166,6 +166,11 @@ var StartNodeCmd = &cobra.Command{
 
 		// Increase MaxPeers if the operator is subscribed to many subnets.
 		// TODO: use OperatorCommittees when it's fixed.
+		const (
+			baseMaxPeers   = 60
+			maxPeersLimit  = 150
+			peersPerSubnet = 3
+		)
 		start := time.Now()
 		myValidators := nodeStorage.ValidatorStore().OperatorValidators(operatorData.ID)
 		mySubnets := make(records.Subnets, networkcommons.SubnetsCount)
@@ -177,9 +182,6 @@ var StartNodeCmd = &cobra.Command{
 				myActiveSubnets++
 			}
 		}
-		const baseMaxPeers = 60
-		const maxPeersLimit = 150
-		const peersPerSubnet = 3
 		idealMaxPeers := min(baseMaxPeers+peersPerSubnet*myActiveSubnets, maxPeersLimit)
 		if cfg.P2pNetworkConfig.MaxPeers < idealMaxPeers {
 			logger.Warn("increasing MaxPeers to match the operator's subscribed subnets",
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -163,11 +163,16 @@ var StartNodeCmd = &cobra.Command{
 		operatorDataStore := operatordatastore.New(operatorData)
 
 		operatorCommittees := nodeStorage.ValidatorStore().OperatorCommittees(operatorData.ID)
+		// Currently, OperatorCommittees may return several committees with the same committee ID, so we need to filter the unique ones.
+		uniqueCommittees := make(map[[32]byte]struct{})
+		for _, oc := range operatorCommittees {
+			uniqueCommittees[oc.ID] = struct{}{}
+		}
 		// If operator has more than 4 committees, it needs more peers, so we need to override the value from config if it's too low.
 		// MaxPeers is used only in p2p, so the lines below must be executed before calling setupP2P function.
 		const committeeThresholdForPeerIncrease = 4
 		const minRequiredPeers = 150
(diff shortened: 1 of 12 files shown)


y0sher:
Adding here this data @nkryuchkov collected about the committees per operator distribution in mainnet network, this should us as a reference for deciding on the parameters of this change:
```
% of operators with >0 committees is 100.00
% of operators with >1 committees is 19.83
% of operators with >2 committees is 11.33
% of operators with >3 committees is 9.17
% of operators with >4 committees is 8.17
% of operators with >5 committees is 7.33
% of operators with >6 committees is 6.83
% of operators with >7 committees is 6.83
% of operators with >8 committees is 6.50
% of operators with >9 committees is 6.33
% of operators with >10 committees is 6.17
```

moshe-blox:
added new log for topic peers distribution:
```json
{"topic peers distribution","min":2,"median":5,"max":7,"dead_subnets":0,"unhealthy_subnets":1}
```

y0sher (approved):
lgtm. adding proof it works:
image

cli/operator/node.go
```diff
@@ -338,6 +338,35 @@ var StartNodeCmd = &cobra.Command{
...
+		idealMaxPeers := min(baseMaxPeers+peersPerSubnet*myActiveSubnets, maxPeersLimit)
+		if cfg.P2pNetworkConfig.MaxPeers < idealMaxPeers {
+			logger.Warn("increasing MaxPeers to match the operator's subscribed subnets",
+				zap.Int("old_max_peers", cfg.P2pNetworkConfig.MaxPeers),
+				zap.Int("new_max_peers", idealMaxPeers),
+				zap.Int("subscribed_subnets", myActiveSubnets),
+				zap.Duration("took", time.Since(start)),
+			)
+			cfg.P2pNetworkConfig.MaxPeers = idealMaxPeers
+		}
```
iurii-ssv, line 368: Is there a need to even have `cfg.P2pNetworkConfig.MaxPeers` value now if we are gonna overwrite it in most cases anyway ? Perhaps we want to remove it to keep it simple (otherwise later on we/somebody might wrongly think the `MaxPeers` value he sets in config file is the value that's gonna be used - which is not gonna be the case).

cli/operator/node.go
```diff
@@ -338,6 +338,35 @@ var StartNodeCmd = &cobra.Command{
...
+		mySubnets := make(records.Subnets, networkcommons.SubnetsCount)
+		myActiveSubnets := 0
+		for _, v := range myValidators {
+			subnet := networkcommons.CommitteeSubnet(v.CommitteeID())
+			if mySubnets[subnet] == 0 {
+				mySubnets[subnet] = 1
+				myActiveSubnets++
+			}
+		}
+		idealMaxPeers := min(baseMaxPeers+peersPerSubnet*myActiveSubnets, maxPeersLimit)
```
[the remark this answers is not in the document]
moshe-blox: i tested with only 16 subnets and 60 peers is not enough, so instead we keep the minimum of 60 + add more for every subnet

cli/operator/node.go
```diff
@@ -338,6 +338,35 @@ var StartNodeCmd = &cobra.Command{
 			nodeProber.AddNode("event syncer", eventSyncer)
 		}
 
+		// Increase MaxPeers if the operator is subscribed to many subnets.
+		// TODO: use OperatorCommittees when it's fixed.
+		const (
+			baseMaxPeers   = 60
+			maxPeersLimit  = 150
+			peersPerSubnet = 3
```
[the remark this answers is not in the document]
moshe-blox: done

cli/operator/node.go
moshe-blox: this is a temporary solution and we wouldnt want to educate everyone about a new configuration only to take it down very soon after

moshe-blox: although we could potentially rename it to TargetPeers and just leave it as is even if we end up reverting this PR

cli/operator/node.go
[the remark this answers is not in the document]
iurii-ssv: > i tested with only 16 subnets and 60 peers is not enough (but 80-90 is), so instead we keep the minimum of 60 + add more for every subnet

ah, I see

cli/operator/node.go
moshe-blox: @iurii-ssv i made some changes to make it more clear that MaxPeers is currently not the max

iurii-ssv (approved):

moshe-blox (approved):

PR #1927, opened 2024-12-06
fix(exporter): decode json payload via buffer

anatolie-ssv:
Reading input stream for requests with large numbers of pubkeys can take more time than the database access, using a buffered helps speed up these cases. #exporter

### DIFF
--- a/api/bind.go
+++ b/api/bind.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -92,7 +93,8 @@ func Bind(r *http.Request, dest interface{}) error {
 	}
 
 	if r.Header.Get("Content-Type") == "application/json" {
-		decoder := json.NewDecoder(r.Body)
+		reader := bufio.NewReader(r.Body)
+		decoder := json.NewDecoder(reader)
 		if err := decoder.Decode(dest); err != nil {
 			return err
 		}


anatolie-ssv:
> do you have number? how much this saves time?

Yes, from seconds down to sub-second:

Screenshot 2024-12-09 at 11 18 16

api/bind.go
```diff
@@ -92,7 +93,8 @@ func Bind(r *http.Request, dest interface{}) error {
 	}
 
 	if r.Header.Get("Content-Type") == "application/json" {
-		decoder := json.NewDecoder(r.Body)
+		reader := bufio.NewReader(r.Body)
+		decoder := json.NewDecoder(reader)
```
moshe-blox, line 97: @anatolie-ssv one of the downsides of using buffers is that we're allocating new buffers on every request.

WDYT about using sync.Pool  to reuse the created bufio instances?

anatolie-ssv: added!

api/bind.go
```diff
@@ -16,6 +18,10 @@ type Binder interface {
 
 var errInvalidType = errors.New("invalid type")
 
+var readerPool = sync.Pool{
+	New: func() any { return bufio.NewReader(nil) },
+}
+
 func Bind(r *http.Request, dest interface{}) error {
```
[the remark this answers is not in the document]
anatolie-ssv: Yes, but the [docs](https://github.com/golang/go/blob/1218566fe5c90e2bd39693566453828e6fbb4c12/src/net/http/request.go#L183) for the stdlib's `Request` interface say:
```
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

api/bind.go
```diff
@@ -92,7 +98,14 @@ func Bind(r *http.Request, dest interface{}) error {
 	}
 
 	if r.Header.Get("Content-Type") == "application/json" {
-		decoder := json.NewDecoder(r.Body)
+		reader := readerPool.Get().(*bufio.Reader)
+		defer func() {
+			reader.Reset(nil)
```
moshe-blox, line 103: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/bufio/bufio.go;l=71

i think this beats the purpose of the pool because it looks like Reset(nil) allocates a new buffer

is it maybe impossible to reuse a bufio.Reader safely?

y0sher (approved):
lgtm. I still think its recommended to `Close()` `r.Body` , unless it happens somewhere else?

Issue #1783, opened 2024-10-09
logs: Promote API calls errors to higher severirty logs (ERROR)


y0sher:
### Description 

Errors in interacting with external APIs like CL,EL nodes are something we want to present to the user right away when they happen. 

### Task

Make sure we print Error to the console whenever a call to EL/CL failed. Never silence them with Debug.

This might a bit harder said than done, some errors are swallowed by other errors, for example a processing of a message in consensus can trigger a call to CL which will result in an error, hence there was an error processing the message (which usually goes to debug).

Olowojaye:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #1939, opened 2024-12-15
chore: improve consensus/execution layer clients logging and error handling

olegshmuelov:
close https://github.com/ssvlabs/ssv/issues/1783

### DIFF
--- a/beacon/goclient/proposer.go
+++ b/beacon/goclient/proposer.go
@@ -33,9 +33,16 @@ func (gc *GoClient) ProposerDuties(ctx context.Context, epoch phase0.Epoch, vali
 		Indices: validatorIndices,
 	})
 	if err != nil {
+		gc.log.Error(clResponseErrMsg,
+			zap.String("api", "ProposerDuties"),
+			zap.Error(err),
+		)
 		return nil, fmt.Errorf("failed to obtain proposer duties: %w", err)
 	}
 	if resp == nil {
+		gc.log.Error(clNilResponseErrMsg,
+			zap.String("api", "ProposerDuties"),
+		)
 		return nil, fmt.Errorf("proposer duties response is nil")
 	}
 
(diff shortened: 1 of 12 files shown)


beacon/goclient/proposer.go
```diff
@@ -58,18 +65,29 @@ func (gc *GoClient) GetBeaconBlock(slot phase0.Slot, graffitiBytes, randao []byt
...
+		gc.log.Error(clNilResponseDataErrMsg,
+			zap.String("api", "Proposal"),
+		)
 		return nil, DataVersionNil, fmt.Errorf("proposal data is nil")
 	}
 
 	metricsProposerDataRequest.Observe(time.Since(reqStart).Seconds())
 	beaconBlock := proposalResp.Data
 
+	// TODO: consider adding logger.Error here as well
```
olegshmuelov, line 90: we have to decide if we want to add err logging here as well

protocol/v2/ssv/runner/committee.go
```diff
@@ -445,6 +446,7 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 	if len(attestations) > 0 {
 		submissionStart := time.Now()
 		if err := cr.beacon.SubmitAttestations(attestations); err != nil {
+			// TODO: consider removing the logger.Error here
```
olegshmuelov, line 449: duplicate err. we should consider removing it

y0sher: since this is in the context of the runner I think we can leave it even though its a duplicate.

y0sher (approved):
gj

protocol/v2/ssv/runner/committee.go
moshe-blox: yeah because you wont see the Beacon logs when filtering for some duty

PR #1945, opened 2024-12-17
feat: do not crush unsynced slots/blocks if withing given limits

anatolie-ssv:
Introduces two configuration parameters to allow specifying how many slots/blocks we're allowed to be behind the chain without halting.

### DIFF
--- a/beacon/goclient/goclient.go
+++ b/beacon/goclient/goclient.go
@@ -142,13 +142,14 @@ var _ NodeClientProvider = (*GoClient)(nil)
 
 // GoClient implementing Beacon struct
 type GoClient struct {
-	log         *zap.Logger
-	ctx         context.Context
-	network     beaconprotocol.Network
-	client      Client
-	nodeVersion string
-	nodeClient  NodeClient
-	gasLimit    uint64
+	log                *zap.Logger
+	ctx                context.Context
+	network            beaconprotocol.Network
+	client             Client
+	nodeVersion        string
+	nodeClient         NodeClient
+	gasLimit           uint64
+	allowUnsyncedSlots uint64
 
 	operatorDataStore operatordatastore.OperatorDataStore
 
@@ -200,13 +201,14 @@ func New(
 	}
 
 	client := &GoClient{
-		log:               logger,
-		ctx:               opt.Context,
-		network:           opt.Network,
-		client:            httpClient.(*eth2clienthttp.Service),
-		gasLimit:          opt.GasLimit,
-		operatorDataStore: operatorDataStore,
-		registrationCache: map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration{},
+		log:                logger,
+		ctx:                opt.Context,
+		network:            opt.Network,
+		client:             httpClient.(*eth2clienthttp.Service),
+		gasLimit:           opt.GasLimit,
+		allowUnsyncedSlots: opt.AllowUnsyncedSlots,
+		operatorDataStore:  operatorDataStore,
+		registrationCache:  map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration{},
 		attestationDataCache: ttlcache.New(
 			// we only fetch attestation data during the slot of the relevant duty (and never later),
 			// hence caching it for 2 slots is sufficient
@@ -265,7 +267,7 @@ func (gc *GoClient) Healthy(ctx context.Context) error {
 
 	// TODO: also check if syncState.ElOffline when github.com/attestantio/go-eth2-client supports it
 	metricsBeaconNodeStatus.Set(float64(statusSyncing))
-	if syncState.IsSyncing {
+	if syncState.IsSyncing && syncState.SyncDistance < phase0.Slot(gc.allowUnsyncedSlots) {
 		return fmt.Errorf("syncing")
 	}
 	if syncState.IsOptimistic {
--- a/eth/executionclient/config.go
+++ b/eth/executionclient/config.go
@@ -8,6 +8,7 @@ import (
 
 // ExecutionOptions contains config configurations related to Ethereum execution client.
 type ExecutionOptions struct {
-	Addr              string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket address"`
-	ConnectionTimeout time.Duration `yaml:"ETH1ConnectionTimeout" env:"ETH_1_CONNECTION_TIMEOUT" env-default:"10s" env-description:"Execution client connection timeout"`
+	Addr                string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket address"`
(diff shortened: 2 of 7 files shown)


eth/executionclient/execution_client.go
```diff
@@ -230,15 +248,23 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
...
-		return fmt.Errorf("syncing")
+
+		current, err := ec.blockNumberFn(ctx)
+		if err != nil {
+			return errSyncing
+		}
+		if sp.CurrentBlock < current-ec.allowUnsyncedBlocks {
+			return errSyncing
+		}
+		return nil
```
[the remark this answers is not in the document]
anatolie-ssv: No, you're right, I missed the part where syncing is also returning highest block. Pushed a fix commit.

beacon/goclient/goclient_test.go
```diff
@@ -20,6 +22,54 @@ import (
...
+			return r, nil
+		}
+
+		client.allowUnsyncedSlots = 2
+
+		err = client.Healthy(ctx)
+		require.ErrorIs(t, err, errSyncing)
+	})
+
+	t.Run("sync distance larger than allowed", func(t *testing.T) {
```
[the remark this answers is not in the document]
anatolie-ssv: fixed

beacon/goclient/goclient.go
```diff
@@ -150,6 +153,9 @@ type GoClient struct {
 	nodeClient  NodeClient
 	gasLimit    uint64
 
+	allowUnsyncedSlots phase0.Slot
```
y0sher, line 156: Can I propose the name `syncDistance` or `syncTolerance`? @anatolie-ssv @moshe-blox @MatusKysel

y0sher (approved):
looks good. minor naming comments.

eth/executionclient/config.go
```diff
@@ -10,4 +10,5 @@ import (
 type ExecutionOptions struct {
 	Addr              string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket address"`
 	ConnectionTimeout time.Duration `yaml:"ETH1ConnectionTimeout" env:"ETH_1_CONNECTION_TIMEOUT" env-default:"10s" env-description:"Execution client connection timeout"`
+	SyncTolerance     uint64        `yaml:"ETH1SyncTolerance" env:"ETH_1_SYNC_TOLERANCE" env-required:"false" env-default:"2" env-description:"The numbers of out-of-sync blocks we can tolerate"`
```
anatolie-ssv, line 13: @MatusKysel please have a look at the namings - if the follow convention and/or are proper (defaults, required)

eth/executionclient/execution_client.go
```diff
@@ -250,7 +262,9 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 	if sp != nil {
 		ec.logger.Error("Execution client is not synced")
 		ec.metrics.ExecutionClientSyncing()
-		return fmt.Errorf("syncing")
+		if sp.CurrentBlock < sp.HighestBlock-ec.syncTolerance {
```
moshe-blox, line 265: in local testnets, this could result in overflow: `sp.HighestBlock-ec.syncTolerance`

it's better to do something like:
```
sp.HighestBlock > ec.syncTolerance && sp.CurrentBlock < sp.HighestBlock-ec.syncTolerance
```

moshe-blox (changes requested):
This looks very good.

I would like to note that there's a potential issue with EL clients in cases where they're completely unsynced and aren't aware of the highest block number (as it isn't based on the clock, unlike slots).

For that reason, I suggest to add a second tolerance mechanism for maximum time since last Syncing=False

This doesn't need to be a config parameter, but rather a hardcoded value set to 1 or 2 minutes because we should really not get there and if we do we'd like to crash in any case, regardless of the user's intention, otherwise we risk slashing deleted validators who assume we've already seen their deletion contract event and stopped attesting.

moshe-blox:
@MatusKysel i believe we're already checking this today just without a threshold, we crash immediately if its unsynced: https://github.com/ssvlabs/ssv/blob/a364c2c08bb455108074e2a6d151ef830fb28c54/eth/executionclient/execution_client.go#L250-L254

eth/executionclient/execution_client.go
```diff
@@ -262,9 +266,22 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
...
+		blockOutOfLimit := sp.CurrentBlock < sp.HighestBlock-ec.syncDistanceTolerance
+
+		// block out of sync distance tolerance
+		blockOutOfSDT := hblock && blockOutOfLimit
+
+		// maximum time we can tolerate since last Syncing=False
+		timeOutOfSDT := ec.syncLastSuccess.Before(time.Now().Add(-syncTimeTolerance))
+
+		// in both cases we crash
+		if blockOutOfSDT || timeOutOfSDT {
```
moshe-blox: This could be simplified for readability, and also adding information to errors:
```suggestion
	syncDistance := max(sp.HighestBlock, sp.CurrentBlock) - sp.CurrentBlock
	if syncDistance > ec.syncDistanceTolerance {
		return fmt.Errorf("sync distance exceeds tolerance (%d): %w", syncDistance, errSyncing)
	}
	
	if ec.syncLastSuccess.Before(time.Now().Add(-syncTimeTolerance)) {
		return fmt.Errorf("not synced for too long (%s): %w", time.Since(ec.syncLastSuccess), errSyncing)
	}
```

moshe-blox: @anatolie-ssv @MatusKysel readability is subjective, but the reason i like this is because it more naturally reflects how we think:

- "crash if sync distance > X slots"
- "crash if not synced for over X time"

anatolie-ssv: integrated these changes and also inverted the `syncLastSuccess => syncLastUnsuccessful` to simplify the happy path.

moshe-blox (approved):
LGTM 🔥

eth/executionclient/execution_client.go
```diff
@@ -250,7 +262,12 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 	if sp != nil {
 		ec.logger.Error("Execution client is not synced")
 		ec.metrics.ExecutionClientSyncing()
-		return fmt.Errorf("syncing")
+		syncDistance := max(sp.HighestBlock, sp.CurrentBlock) - sp.CurrentBlock
+
+		// block out of sync distance tolerance
+		if syncDistance > ec.syncDistanceTolerance {
+			return fmt.Errorf("sync distance exceeds tolerance (%d): %w", syncDistance, errSyncing)
+		}
```
olegshmuelov, line 270: The log `ec.logger.Error("Execution client is not synced")`
should be moved inside the `if syncDistance > ec.syncDistanceTolerance` block
to avoid unnecessary logging when the sync distance is within tolerance.

olegshmuelov: https://github.com/ssvlabs/ssv/issues/1960

anatolie-ssv: > #1960

there's a follow up PR in which this will be fixed.

PR #1949, opened 2024-12-18
feat(p2p): scored discovery

iurii-ssv:
Replaces https://github.com/ssvlabs/ssv/pull/1917 additionally improving a bunch of p2p-handling code, with the main contributing factor being "selective discovery"

_Before merging:_
- [x] remove/resolve TODOs used for testing
- [x] (when testing) make sure this PR doesn't introduce memory-leaks or super-high CPU consumption type of issues
- [x] do we want to set `MinConnectivitySubnets` to higher-than-0 value (say 3) ? There isn't really a good way to tell how it affects network connectivity (but it's almost certainly gonna improve it, not worsen it); we **removed this feature out of this PR** for now (for simplicity)


network/discovery/dv5_service.go
```diff
@@ -157,7 +180,7 @@ func (dvs *DiscV5Service) Bootstrap(logger *zap.Logger, handler HandleNewPeer) e
 			return
 		}
 		handler(e)
-	}, defaultDiscoveryInterval) // , dvs.forkVersionFilter) //, dvs.badNodeFilter)
+	}, defaultDiscoveryInterval, dvs.ssvNodeFilter(logger), dvs.sharedSubnetsFilter(1), dvs.badNodeFilter(logger), dvs.alreadyConnectedFilter(), dvs.recentlyTrimmedFilter())
```
iurii-ssv, line 183: using `ssvNodeFilter` here results in `TestDiscV5Service_Bootstrap` failing (I guess because we don't appropriately mark `NewTestingNode` as "ssv"),

not sure why we need this kind of filtering (should I try to adjust this failing test, or get rid of it) @y0sher can you elaborate ?

network/discovery/dv5_service.go
```diff
@@ -56,6 +57,14 @@ type DiscV5Service struct {
 	dv5Listener Listener
 	bootnodes   []*enode.Node
 
+	// withExtraFiltering specifies whether we want to additionally filter discovered peers,
+	// since operator node and boot node are using the same discovery code (`DiscV5Service`) we
+	// want to do this extra filtering for operator node only
+	withExtraFiltering bool
+	// topicsCtrl must be non-nil when withExtraFiltering is set to true since it will be used
+	// to help us decide which peers need to be filtered out
+	topicsCtrl topics.Controller
```
iurii-ssv, line 66: I haven't looked into how boot-node uses discovery, but we are forced to handle this somehow (which `withExtraFiltering` flag is for).

network/peers/conn_manager.go
```diff
@@ -2,34 +2,37 @@ package peers
...
 
 type PeerScore float64
 
+var (
+	TrimmedRecently = ttlcache.New(ttlcache.WithTTL[peer.ID, struct{}](30 * time.Minute))
+)
+
+func init() {
+	go TrimmedRecently.Start() // start cleanup go-routine
+}
```
iurii-ssv, line 28: I know global vars aren't best way to do it, but I think we should leave it global for now and refactor it later (once this code has been running in production for a while and we've proven this solution really does help). Since we might want/have to swap it out for something else.

network/discovery/dv5_service.go
```diff
@@ -196,6 +219,36 @@ func (dvs *DiscV5Service) checkPeer(logger *zap.Logger, e PeerEvent) error {
...
+		subscribedTopics := dvs.topicsCtrl.Topics()
+		for _, topic := range subscribedTopics {
+			topicPeers, err := dvs.topicsCtrl.Peers(topic)
+			if err != nil {
+				return errors.Wrap(err, "could not get subscribed topic peers")
+			}
+
+			if len(topicPeers) >= 2 {
+				continue // this topic has enough peers
+			}
```
iurii-ssv, line 233: This means we are looking (in discovery) for peers who help with either **dead** or **solo** subnets (without actually prioritizing the dead ones, and also this kind of filtering is maybe too aggressive overall since in my testing I see that we aren't keeping at `MaxPeers` limit for long - and rather falling 10-20% below it and keeping at that level)

this isn't optimal way to do it, but in my testing on prod I found it to be working reasonably good (with 60 and 90`MaxPeers`) - so we **probably want to improve this, just maybe not in the upcoming release**

network/p2p/p2p_setup.go
```diff
@@ -48,6 +47,9 @@ const (
 	connectTimeout = time.Minute
 	// connectorQueueSize is the buffer size of the channel used by the connector
 	connectorQueueSize = 256
+	// inboundLimitRatio is the ratio of inbound connections to the total connections
+	// we allow (both inbound and outbound).
+	inboundLimitRatio = float64(0.75)
```
iurii-ssv, line 52: I'm testing now if `0.75` is a good fit, but so far
- I found the value of `0.5` to be working well
- and the value of `0.8` seems to be noticeably worse (eg. making it 30 minutes instead of 15 to get rid of dead subnets)

network/discovery/dv5_service.go
y0sher: Well, the SSV filter isn't even necessary after alan because we added protocol ID. should we just remove it cc: @moshe-blox ?

network/p2p/p2p_setup.go
iurii-ssv: I guess `0.75` is too high still, even with MaxPeers = 90 sometimes it can take up to 90 minutes to get rid of dead subnets (or even longer, if we are unlucky enough),

it's probably better to set `inboundLimitRatio` = 0.6 or even lower

iurii-ssv:
Another thing I've realized is happening - because we only trim once we've reached MaxPeers limit, at some point we might stop trimming completely - for example, we've got to max amount of the incoming connections we allow via 0.5 ratio (say, 30) + we've discovered some outgoing peers but not enough to reach MaxPeers limit,

in other words, we've discovered everything we could (and we keep discovering, just at a very slow rate - due to quite an aggressive filter there isn't many connections we are interested in) but that also means we stopped re-cycling(trimming) our incoming connections periodically because we don't trim incoming/outgoing connections separately

I've added a commit to address that - https://github.com/ssvlabs/ssv/pull/1949/commits/d230c4adeb77622f46a0a761bcc77c37da69aabf - it's pretty "hacky/ad-hoc" way do the job - ideally (to simplify this) I think we probably want to have separate limits per incoming/outgoing connections and somewhat separate trimming logic, WDYT ?

network/discovery/dv5_service.go
y0sher: or we can just mark it as `ssv` in `NewTestingNode`

network/discovery/dv5_service.go
```diff
@@ -143,11 +165,12 @@ func (dvs *DiscV5Service) Bootstrap(logger *zap.Logger, handler HandleNewPeer) e
 	const logFrequency = 10
 	var skippedPeers uint64 = 0
 
-	dvs.discover(dvs.ctx, func(e PeerEvent) {
+	dvs.discover(dvs.ctx, logger, func(e PeerEvent) {
```
y0sher, line 168: can we just set the logger in the struct instead of passing it down?

network/p2p/p2p.go
```diff
@@ -256,73 +255,174 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
 
 // startDiscovery starts the required services
 // it will try to bootstrap discovery service, and inject a connect function.
 // the connect function checks if we can connect to the given peer and if so passing it to the backoff connector.
 func (n *p2pNetwork) startDiscovery(logger *zap.Logger, connector chan peer.AddrInfo) {
-	discoveredPeers := make(chan peer.AddrInfo, connectorQueueSize)
-	go func() {
-		ctx, cancel := context.WithCancel(n.ctx)
-		defer cancel()
-		n.backoffConnector.Connect(ctx, discoveredPeers)
```
y0sher, line 321: I don't understand how the connector works without this?

iurii-ssv: Oh yeah, I should have left a comment for this change here, 

it's just dead code - what we really use is the argument we get in this function `connector chan peer.AddrInfo`

network/discovery/dv5_service.go
```diff
@@ -56,6 +57,14 @@ type DiscV5Service struct {
 	dv5Listener Listener
 	bootnodes   []*enode.Node
 
+	// withExtraFiltering specifies whether we want to additionally filter discovered peers,
+	// since operator node and boot node are using the same discovery code (`DiscV5Service`) we
+	// want to do this extra filtering for operator node only
+	withExtraFiltering bool
```
moshe-blox, line 63: bootnode doesn't pull out peers from discovery to filter them so this is not needed

iurii-ssv: Can you clarify why it needs `newDiscV5Service` service then ?

network/discovery/dv5_service.go
iurii-ssv: @moshe-blox could you comment which one ^ we want to go with ?

network/discovery/dv5_service.go
y0sher: @iurii-ssv 
Bootnode is still a regular node in terms of discovery, it has a routing table which collects other peers it has discovery interactions with, it does "pull" peers from this table to supply other peers that query it for network peers it knows. But it doesn't go through our filtering, our filtering is only used for pulling peers out of the routing table to be used for protocol connections (TCP).

network/p2p/p2p_setup.go
iurii-ssv: I'd go with 0.5 if there aren't any reasons not to

network/discovery/dv5_service.go
iurii-ssv: Okay, removed `withExtraFiltering` then, and so we are using this func now 
```go
func (n *p2pNetwork) peerScore
```
both:
- to rank "proposed" peers
- to rank peers when trimming (peer "protection")

network/discovery/dv5_service.go
iurii-ssv: yes sure, I guess that'd make it a bit cleaner, applied in https://github.com/ssvlabs/ssv/pull/1949/commits/88a752cc604452e125dbe8e8a5d86e7ae30cf2af

network/discovery/dv5_service.go
iurii-ssv: > or we can just mark it as ssv in NewTestingNode

done in https://github.com/ssvlabs/ssv/pull/1949/commits/88a752cc604452e125dbe8e8a5d86e7ae30cf2af, test is passing now

beacon/goclient/goclient.go
```diff
@@ -19,14 +19,13 @@ import (
...
-	"go.uber.org/zap"
-	"tailscale.com/util/singleflight"
-
 	"github.com/ssvlabs/ssv/logging/fields"
 	operatordatastore "github.com/ssvlabs/ssv/operator/datastore"
 	"github.com/ssvlabs/ssv/operator/slotticker"
 	beaconprotocol "github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 	"github.com/ssvlabs/ssv/utils/casts"
+	"go.uber.org/zap"
+	"tailscale.com/util/singleflight"
```
y0sher, line 28: previous location looks better

network/discovery/dv5_bootnode.go
```diff
@@ -3,12 +3,11 @@ package discovery
 import (
 	"context"
 
-	"go.uber.org/zap"
-
 	"github.com/ssvlabs/ssv/logging"
 	"github.com/ssvlabs/ssv/logging/fields"
 	"github.com/ssvlabs/ssv/networkconfig"
 	"github.com/ssvlabs/ssv/utils"
+	"go.uber.org/zap"
```
y0sher, line 10: previous import place looked better

beacon/goclient/goclient.go
iurii-ssv: I think we need a `make format` command that would take care of all the imports, it's a bit of time-waste to look out for this manually

Can we do a separate task/PR for this to fix it project-wide ?

cc @moshe-blox @nkryuchkov @oleg-ssvlabs @MatusKysel

network/discovery/dv5_bootnode.go
iurii-ssv: lets keep this discussion to https://github.com/ssvlabs/ssv/pull/1949#discussion_r1932163881

beacon/goclient/goclient.go
nkryuchkov: @iurii-ssv I agree, we often change imports according to different rules and it causes merge conflicts

network/p2p/p2p.go
```diff
@@ -249,72 +260,457 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+		// leaves some vacant slots for the next iteration - on the next iteration better
+		// peers might show up (so we don't want to "spend" all of these vacant slots at once)
+		peersToProposeCnt := max(vacantOutboundSlotCnt/2, 1)
+		// also limit how many peers we want to propose with respect to "peer synergy" (this
+		// value can't be too high for performance reasons, 12 seems like a good middle-ground)
+		const peersToProposeMaxWithSynergy = 12
+		peersToProposeCnt = min(peersToProposeCnt, peersToProposeMaxWithSynergy)
+		// peersToProposePoolCnt is a size of candidate-peers pool we'll be choosing exactly
+		// peersToProposeCnt peers from
+		peersToProposePoolCnt := 2 * peersToProposeCnt
```
y0sher, line 352: why not choose from the whole pool?

iurii-ssv: The size of `peersByPriority` can be large, 

from what I understand there isn't a much better solution to this problem than brute-force (which is what I'm doing here) because any combination might be **best** - we don't know until we compare it against the best we've got so far

so from my testing I figured 24 is the highest reasonable value:
```
		// also limit how many peers we want to propose with respect to "peer synergy" (this
		// value can't be too high for performance reasons, 12 seems like a good middle-ground)
		const peersToProposeMaxWithSynergy = 12
		peersToProposeCnt = min(peersToProposeCnt, peersToProposeMaxWithSynergy)
```

network/p2p/p2p.go
```diff
@@ -249,72 +260,457 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
 	)
 
-	go n.startDiscovery(logger, connector)
+	connectorProposals := make(chan peer.AddrInfo, connectorQueueSize)
+	go n.startDiscovery(logger, connectorProposals)
+	go func() {
+		// keep discovered peers in the pool so we can choose the best ones
+		for proposal := range connectorProposals {
+			if peers.DiscoveredPeersPool.Has(proposal.ID) {
+				// this log line is commented out as it is too spammy
```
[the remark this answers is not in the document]
iurii-ssv: Indeed we can, done - https://github.com/ssvlabs/ssv/pull/1949/commits/82d4805b9515c8f77b57247b69daab20bf9d3115

network/p2p/p2p.go
```diff
@@ -249,72 +260,449 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+	// the best ones (instead of the ones we've discovered first, for example), additionally when
+	// choosing a batch of peers to propose we evaluate "peer synergy" between them - which is a way
+	// to pick peers such that they don't have too many overlapping subnets and instead cover a
+	// wider range of dead/solo subnets for us
+	async.Interval(n.ctx, 15*time.Second, func() {
+		// give discovery some time to find the best peers right after node start, the exact
+		// best time to wait is arrived at experimentally
+		if time.Since(p2pStartTime) < 1*time.Minute {
+			return
+		}
```
iurii-ssv, line 291: Tried **5m** too but it doesn't really help (vs **1m**):

image

I think to fully get rid of dead subnets fast we need to set **MinConnectivitySubnets** to **> 0**

beacon/goclient/goclient.go
iurii-ssv: I've put up a PR to address this - https://github.com/ssvlabs/ssv/pull/2045

network/commons/common_test.go
zippoxer, line 1: why did we remove this test? is there a replacement somewhere?

network/commons/subnets.go
```diff
@@ -0,0 +1,185 @@
...
+	subnet := uint64(0)
+	for _, val := range s {
+		subnetsVec.SetBitAt(subnet, val > uint8(0))
+		subnet++
+	}
+	return hex.EncodeToString(subnetsVec.Bytes())
+}
+
+// FromString parses a given subnet string
+func (s Subnets) FromString(subnetsStr string) (Subnets, error) {
```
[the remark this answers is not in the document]
y0sher: I agree, `s` isn't even used so no reason for this to be a method.

network/commons/subnets.go
```diff
@@ -0,0 +1,185 @@
...
+		maxLen = len(a)
+	}
+	if len(a) == 0 || len(b) == 0 {
+		return shared
+	}
+	for subnet, aval := range a {
+		if aval == 0 {
+			continue
+		}
+		if b[subnet] == 0 {
```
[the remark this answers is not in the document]
iurii-ssv: It's unfortunate that we use **slices** and not **arrays** here, but looks like it works so far

I was looking into if we can change it - but that affects lots and lots of code ... so if we want to go ahead and refactor this it's probably better to do as a separate change on top later on

network/commons/subnets.go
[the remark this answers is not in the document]
iurii-ssv: yep, resolved in https://github.com/ssvlabs/ssv/pull/1949/commits/22e736e76b7c061b020c9b2278c5989d608a75ce

network/p2p/p2p_setup.go
```diff
@@ -89,18 +92,22 @@ func (n *p2pNetwork) initCfg() error {
...
 		n.fixedSubnets = subnets
 	} else {
-		n.fixedSubnets = make(records.Subnets, p2pcommons.Subnets())
+		n.fixedSubnets = make(p2pcommons.Subnets, p2pcommons.SubnetsCount)
 	}
 	if n.cfg.MaxPeers <= 0 {
 		n.cfg.MaxPeers = minPeersBuffer
 	}
+
+	// TODO - override config value for testing
```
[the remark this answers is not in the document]
iurii-ssv: right, cleaned up in https://github.com/ssvlabs/ssv/pull/1949/commits/4be96eda0906d820a8d906016ceaf749b9674148

network/commons/subnets.go
```diff
@@ -0,0 +1,185 @@
...
+
+	topicPrefix = "ssv.v2"
+)
+
+// BigIntSubnetsCount is the big.Int representation of SubnetsCount
+var bigIntSubnetsCount *big.Int
+
+func init() {
+	bigIntSubnetsCount = new(big.Int).SetUint64(SubnetsCount)
+}
```
nkryuchkov, line 31:
```suggestion
// BigIntSubnetsCount is the big.Int representation of SubnetsCount
var bigIntSubnetsCount = new(big.Int).SetUint64(SubnetsCount)
```

network/commons/subnets.go
```diff
@@ -0,0 +1,185 @@
...
+
+// GetTopicFullName returns the topic full name, including prefix
+func GetTopicFullName(baseName string) string {
+	return fmt.Sprintf("%s.%s", topicPrefix, baseName)
+}
+
+// GetTopicBaseName return the base topic name of the topic, w/o ssv prefix
+func GetTopicBaseName(topicName string) string {
+	return strings.Replace(topicName, fmt.Sprintf("%s.", topicPrefix), "", 1)
```
nkryuchkov, line 52:
```suggestion
	return strings.TrimPrefix(topicName, topicPrefix+".")
```
Do you think this might be more readable or do you prefer your version?

network/commons/subnets.go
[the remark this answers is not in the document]
nkryuchkov: we have a PR for changing the type https://github.com/ssvlabs/ssv/pull/1813

network/commons/subnets.go
iurii-ssv: I think it's maybe Yosher's sketch that I took "as is" ;) 

I prefer the one with `TrimPrefix`, applied

network/p2p/p2p.go
```diff
@@ -249,72 +259,405 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+			peers.DiscoveredPeersPool.Set(proposal.ID, discoveredPeer)
+
+			n.interfaceLogger.Debug(
+				"discovery proposed peer, adding it to the pool",
+				zap.String("peer_id", string(proposal.ID)),
+			)
+		}
+	}()
+	// start go-routine to choose the best peer(s) from the pool of discovered peers to propose
+	// these as outbound connections, all discovered peers are kept in a pooled so we can choose
```
y0sher, line 280: typo
```suggestion
	// these as outbound connections, all discovered peers are kept in a pool so we can choose
```

network/p2p/p2p.go
```diff
@@ -249,72 +259,405 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+					// remove him from DiscoveredPeersPool (since if we do - discovery might suggest this
+					// peer again essentially resetting this peer's retry attempts counter to 0)
+
+					// this log line is commented out as it is too spammy
+					//n.interfaceLogger.Debug(
+					//	"Not gonna try to connect discovered peer: ran out of retries",
+					//	zap.String("peer_id", string(key)),
+					//)
+					return true
+				}
```
y0sher, line 402: should we ever restart the tries ? It might be a helpful peer who just denied us because he reached the peer limit. unfortunately we can't know the reason why the peer denied us..

iurii-ssv: I've added a clarifying comment:

> // ... eventually
					// (after some time passes) this peer will automatically get removed from DiscoveredPeersPool
					// so it can be discovered again (effectively resetting peer's retry attempts counter to 0)
					
does it make sense ?

network/peers/conn_manager.go
```diff
@@ -2,38 +2,44 @@ package peers
...
+	// ConnectRetries keeps track of how many times we tried to connect to this peer.
+	ConnectRetries int
+}
+
+var (
+	// DiscoveredPeersPool keeps track of recently discovered peers so we can rank them and choose
+	// the best candidates to connect to.
+	DiscoveredPeersPool = ttl.New[peer.ID, DiscoveredPeer](15*time.Minute, 5*time.Minute)
+	TrimmedRecently     = ttl.New[peer.ID, struct{}](30*time.Minute, 5*time.Minute)
+)
```
y0sher, line 31: I think this should be part of a struct and not "static"

operator/validator/controller.go
```diff
@@ -467,6 +468,10 @@ func (c *controller) StartValidators(ctx context.Context) {
 
 		// Start validators.
 		c.startValidators(inited, committees)
+
+		if err := c.network.SubscribeFillerSubnets(c.logger); err != nil {
+			c.logger.Error("failed to subscribe to filler subnets", zap.Error(err))
+		}
```
y0sher, line 474: if we're not enabling this feature, we might want to drop it out of this PR.

y0sher (changes requested):
small changes :)

operator/validator/controller.go
iurii-ssv: `MinConnectivitySubnets` currently set to 0 (so it's no-op effectively),
- I thought we might want to deploy
- and then maybe bump up to 3 to see what effect it has

but if we are not planning to add this at all I can remove it ofc

network/peers/conn_manager.go
iurii-ssv: Done here - https://github.com/ssvlabs/ssv/pull/1949/commits/e4ae99e9a4a033e99ee4cb5cd0763598e43712d4

iurii-ssv: and some related fixes on top - https://github.com/ssvlabs/ssv/pull/1949/commits/5d3304eb0785f1ff524c424f0b2e8276fb8b4e59

network/p2p/p2p.go
y0sher: Yeah I see we leave it to the ttl logic. lgtm

y0sher: Maybe in the future we could asses how much "important" that peer is and try again faster if tits worth it.

iurii-ssv: > Maybe in the future we could asses how much "important" that peer is and try again faster if tits worth it.

yeah, it just complicates implementation a bit

operator/validator/controller.go
y0sher: Since `MinConnectivitySubnets` is a hard-coded const, we're not going to be able to "bump" it without releasing another version anyway. So there's no point in leaving that it.

network/p2p/p2p.go
```diff
@@ -475,3 +825,67 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
+			if elem == peerID {
+				continue
+			}
+			result = append(result, elem)
+		}
+		return result
+	}
+
+	topic := strconv.Itoa(subnet)
+	subnetPeers, err := n.topicsCtrl.Peers(topic)
```
y0sher, line 861: since this function (`score`) is going to be called a few times in each iteration. I think it makes sense to call `n.topicsCtrl.Peers` one time and pass it down. It's not going to change and it uses a channel (essentially a lock)

operator/validator/controller.go
iurii-ssv: True, I mean - **if we do plan** to test it out eventually - there is no harm in shipping this code as well ?

**If we don't plan** to test it out - then it should be removed ofc,

I just don't think it's worth bothering moving it our in a separate PR ?

network/p2p/p2p.go
iurii-ssv: I think it's called with different `subnet` every time:
```go
	sharedSubnets := commons.SharedSubnets(n.activeSubnets, peerSubnets, 0)
	for _, subnet := range sharedSubnets {
		result += n.score(peerID, subnet)
	}
```
meaning `n.topicsCtrl.Peers` returns different result(s) every time ?

y0sher: ahh I see now.

operator/validator/controller.go
iurii-ssv: Okay, removing then - https://github.com/ssvlabs/ssv/pull/1949/commits/8d677191a04dc07e5afac990d4e56df92948b204

network/p2p/p2p.go
```diff
@@ -475,3 +825,67 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
+			zap.String("topic", topic),
+			zap.Error(fmt.Errorf("could not get topic peers: %w", err)),
+		)
+		return 0.0
+	}
+	subnetPeersExcluding := len(filterOutPeer(peerID, subnetPeers))
+
+	const targetPeersPerSubnet = 3
+	return score(targetPeersPerSubnet, subnetPeersExcluding)
+}
```
zippoxer, line 874: is this not equivalent?
```suggestion
// score assesses how valuable the contribution of peerID to specified subnet is by calculating
// how valuable this peer would have been if we didn't have him, but then connected with.
func (n *p2pNetwork) score(peerID peer.ID, subnet int) float64 {
	topic := strconv.Itoa(subnet)
	subnetPeers, err := n.topicsCtrl.Peers(topic)
	if err != nil {
		n.interfaceLogger.Debug(
			"cannot score peer with respect to this subnet, assuming zero contribution",
			zap.String("topic", topic),
			zap.Error(fmt.Errorf("could not get topic peers: %w", err)),
		)
		return 0.0
	}
	subnetPeersExcluding := 0
    for _, p := range subnetPeers {
        if p != peerID {
            subnetPeersExcluding++
        }
    }

	const targetPeersPerSubnet = 3
	return score(targetPeersPerSubnet, subnetPeersExcluding)
}
```

network/p2p/p2p.go
```diff
@@ -249,72 +273,407 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+			n.interfaceLogger.Debug(
+				"discovery proposed peer, adding it to the pool",
+				zap.String("peer_id", string(proposal.ID)),
+			)
+		}
+	}()
+	// start go-routine to choose the best peer(s) from the pool of discovered peers to propose
+	// these as outbound connections, all discovered peers are kept in a pool so we can choose
+	// from that same pool across different runs of this interval-func
+	async.Interval(n.ctx, 15*time.Second, func() {
```
zippoxer, line 296: i believe its time to extract this 200 lines func into its own method 😅

zippoxer: @y0sher @iurii-ssv this is an improvement, but i still think this is just hard to read compared to what it does

zippoxer: there are many comments and they're very long

comments should only be added where it's hard or impossible to infer what's happening from the code, or for visual separation (and in the latter case it should be a name/title, not an explanation)

iurii-ssv: > comments should only be added where it's hard or impossible to infer what's happening from the code, or for visual separation of steps (and in the latter case it should be a name/title, not an explanation)

I try to leave only comments that have "useful" info in them, I think it's better to have **more** comments than **almost no** comments

but I agree it's hard to write concise and complete comments, do you have specific suggestions on what to remove/rephrase ?

moshe-blox: 2. i might be wrong, but the scoring mechanism looks more complex than it has to be

@y0sher @iurii-ssv  curious to hear arguments why it should be more complex than what we had in the simulation?

network/p2p/p2p.go
iurii-ssv: I think it is equivalent, but it's less readable IMO because `filterOutPeer` hides away this chunk of code that you'd need to read trough to figure out what's going on (maybe it's a matter of preference, but doesn't seem all that important either way):
```
subnetPeersExcluding := 0
    for _, p := range subnetPeers {
        if p != peerID {
            subnetPeersExcluding++
        }
    }
```

network/p2p/p2p.go
moshe-blox: @iurii-ssv 
> I think it's better to have more comments than almost no comments

i agree:
```
concise comments > verbose comments > no comments
```

i'm not saying the comments are bad, it'd be bad if we had no comments at all!

i'm just trying to push us from level 2 to 3 🫡

iurii-ssv: > but the scoring mechanism looks more complex than it has to be

The func closest to what we had in simulation is this one - https://github.com/ssvlabs/ssv/blob/4895ae1272b7e2962884170a67523a7ef94f7ca5/network/p2p/p2p.go#L845-L845

but it doesn't map in straightforward way onto this solution we've decided to go with (it only works with our own peers):

image

and so it's only used when trimming currently connected peers, 

while `scoreSubnetSum` is the most straightforward implementation of the algo we want for discovery

> none of the comments here are needed because they're obvious from reading the variable names

yeah, maybe it's a bit too much, removed redundant comments here -  https://github.com/ssvlabs/ssv/pull/1949/commits/81c8268ab0083ddc254b2a1111219537a6010f10

y0sher: I agree splitting out of the `Start` function could make both the `Start` function and the scoring function easier to read.

y0sher: Also on the comments part, I partially agree with @moshe-blox , I think it might be worth to "sum up" some comments so instead of comment every line we can "prepare the reader" to what the upcoming code will do, and trust the the variable names will talk for themselves. unless of course the code is a bit more complex so it needs further explaining.

iurii-ssv: > however in this case we're under the `p2p.Start` method which does many things, not one, so it'd make sense to extract the connection-making task to its own method, such as the tasks below it already are:
> 
> https://github.com/ssvlabs/ssv/blob/4895ae1272b7e2962884170a67523a7ef94f7ca5/network/p2p/p2p.go#L475-L482
> 
> so we're not splitting this logic in any sense

I wasn't thinking about `Start` func itself - it's a kind of refactoring I would usually prefer to do on top, but if you think it simplifies (not complicates) review - maybe we can move discovery-related parts into separate func - let me see what I can do about that

network/p2p/p2p.go
```diff
@@ -249,72 +267,412 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+				zap.String("topic", tpc),
+				zap.Error(err),
+			)
+			continue
+		}
+		for _, p := range peerz {
+			myPeersSet[p] = struct{}{}
+		}
+	}
+
```
moshe-blox, line 609: can we perhaps use `n.topicsCtrl.Peers("")` instead of this loop?

iurii-ssv: It seems we should be able to, changed in https://github.com/ssvlabs/ssv/pull/1949/commits/bd189c12ba94f2c532f6dafefc6f3b27f277874c - lets see if it works in the next deploy

network/p2p/p2p.go
```diff
@@ -249,72 +267,402 @@ func (n *p2pNetwork) Start(logger *zap.Logger) error {
...
+				if s[i] == 1 {
+					soloSubnetCnt++
+				}
+				if s[i] == 2 {
+					duoSubnetCnt++
+				}
+			}
+			return float64(maxPossibleScore - (deadSubnetPriority*deadSubnetCnt + soloSubnetPriority*soloSubnetCnt + duoSubnetPriority*duoSubnetCnt))
+		}
+
```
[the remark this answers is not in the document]
iurii-ssv: Yeah, what you suggest will make it a bit more readable, so I've changed the `scoreSubnetPeers` func to be a method of `SubnetSum`:
```
// scoreSubnetVector estimates the score of subnet-vector v (each byte in v is 0 or 1) based on 
// how many dead/solo/duo subnets it contributes to the current subnet-sum (higher score is better)
func (s *SubnetSum) scoreSubnetVector(v commons.Subnets) (score float64) {
	const (
		duoSubnetPriority  = 1
		soloSubnetPriority = 3
		deadSubnetPriority = 9
	)

	result := 0.0
	for i := 0; i < commons.SubnetsCount; i++ {
		if v[i] != 1 {
			continue
		}
		if s[i] == 0 {
			result += deadSubnetPriority
		}
		if s[i] == 1 {
			result += soloSubnetPriority
		}
		if s[i] == 2 {
			result += duoSubnetPriority
		}
	}
	return result
}
```
and also refactored the way we work with `SubnetSum` a bit - https://github.com/ssvlabs/ssv/pull/1949/commits/690ee3e459d45fd2400343d5f16c13a242236545

let me know if that's in line with what you had in mind :)

y0sher (approved):

PR #1955, opened 2024-12-19
Wrap uint64 metric value recording

oleg-ssvlabs:


observability/metric.go
```diff
@@ -1,15 +1,33 @@
...
 }
+
+func RecordUint64Value(ctx context.Context,
+	value uint64,
+	recordF func(ctx context.Context, value int64, options ...metric.RecordOption),
+	options ...metric.RecordOption) {
+	if value > math.MaxInt64 {
+		logger.Warn("value exceeds int64 range, metric not recorded")
+		return
+	}
```
iurii-ssv, line 30: Probably better to log it as `Error` not `Warn` since we shouldn't ever see this happen.

Also, I might be tripping but I think in https://github.com/ssvlabs/ssv/pull/1852 I saw `uint32` -> `int32` conversion as well, but you'd probably know better

iurii-ssv (approved):
LGTM

observability/metric.go
oleg-ssvlabs: changed log severity to Error. Added unit test

PR #1956, opened 2024-12-19
feat(metrics): Implement ssv.p2p.discovery.iterations counter with 'ssv.fork' attribute

oleg-ssvlabs:


network/discovery/forking_dv5_listener.go
```diff
@@ -95,5 +98,9 @@ type annotatedIterator struct {
 }
 
 func (i *annotatedIterator) Next() bool {
+	if i.Iterator.Next() {
+		peerDiscoveryIterationsCounter.Add(
+			context.TODO(), 1, metric.WithAttributes(attribute.String("ssv.fork", i.fork)))
+	}
 	return i.Iterator.Next()
```
moshe-blox, line 105: we shouldn't call Next twice, it's best to reuse the same code we had before

PR #1958, opened 2024-12-20
feat: (blossom) prepare to 1000 validators hardfork

MatusKysel:
The PR prepares the codebase for the blossom hard fork, which increases the max validator count to 1000: https://github.com/ssvlabs/SIPs/pull/53

Changes:
- `message/validation`: adjust message sizes
- `networkconfig`: set blossom epoch

### DIFF
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -43,7 +43,9 @@ const (
 	maxPartialSignatureMsgsSizeOld     = partialSigMsgTypeSize + slotSize + maxSizePartialSignatureMessagesOld*maxSizePartialSignatureMessage + 4
 	maxEncodedPartialSignatureSizeOld  = maxPartialSignatureMsgsSizeOld + maxPartialSignatureMsgsSizeOld/encodingOverheadDivisor
 	///
-	maxSizePartialSignatureMessages = 1512
+	maxAttestations                 = 1000
+	maxSyncCommittees               = 512
+	maxSizePartialSignatureMessages = maxAttestations + maxSyncCommittees
 	partialSigMsgTypeSize           = 8 // uint64
 	maxPartialSignatureMsgsSize     = partialSigMsgTypeSize + slotSize + maxSizePartialSignatureMessages*maxSizePartialSignatureMessage + 4
 	maxEncodedPartialSignatureSize  = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -36,12 +36,12 @@ const (
 )
 
 const (
-	partialSignatureSize           = 96
-	partialSignatureMsgSize        = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
-	maxPartialSignatureMessages    = 1000
-	partialSigMsgTypeSize          = 8 // uint64
-	maxPartialSignatureMsgsSize    = partialSigMsgTypeSize + slotSize + maxPartialSignatureMessages*partialSignatureMsgSize
-	maxEncodedPartialSignatureSize = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor + 4
+	partialSignatureSize            = 96
+	maxSizePartialSignatureMessage  = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	maxSizePartialSignatureMessages = 1512
+	partialSigMsgTypeSize           = 8 // uint64
+	maxPartialSignatureMsgsSize     = partialSigMsgTypeSize + slotSize + maxSizePartialSignatureMessages*maxSizePartialSignatureMessage + 4
+	maxEncodedPartialSignatureSize  = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor
 )
 
 const (
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -36,8 +36,13 @@ const (
 )
 
 const (
-	partialSignatureSize            = 96
-	maxSizePartialSignatureMessage  = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	partialSignatureSize           = 96
+	maxSizePartialSignatureMessage = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	/// Pre fork values TODO remove after Blossom
+	maxSizePartialSignatureMessagesOld = 1000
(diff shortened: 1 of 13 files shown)


message/validation/const.go
```diff
@@ -37,11 +37,16 @@ const (
...
-	partialSigMsgTypeSize          = 8 // uint64
-	maxPartialSignatureMsgsSize    = partialSigMsgTypeSize + slotSize + maxPartialSignatureMessages*partialSignatureMsgSize
-	maxEncodedPartialSignatureSize = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor + 4
+	maxSizePartialSignatureMessage = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	/// Pre fork values TODO remove after Blossom
+	maxSizePartialSignatureMessagesOld = 1000
+	maxPartialSignatureMsgsSizeOld     = partialSigMsgTypeSize + slotSize + maxSizePartialSignatureMessagesOld*maxSizePartialSignatureMessage + 4
+	maxEncodedPartialSignatureSizeOld  = maxPartialSignatureMsgsSizeOld + maxPartialSignatureMsgsSizeOld/encodingOverheadDivisor
+	///
+	maxSizePartialSignatureMessages = 1512
```
y0sher, line 46: can @MatheusFranco99 @GalRogozinski please approve these numbers.

message/validation/partial_validation.go
```diff
@@ -23,10 +23,15 @@ func (mv *messageValidator) validatePartialSignatureMessage(
 ) {
 	ssvMessage := signedSSVMessage.SSVMessage
 
-	if len(ssvMessage.Data) > maxEncodedPartialSignatureSize {
+	localMaxEncodedPartialSigSize := maxEncodedPartialSignatureSize
+	if mv.netCfg.Beacon.EpochStartTime(mv.netCfg.BlossomEpoch).Before(receivedAt) {
```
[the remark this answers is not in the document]
y0sher: it looks ok to me. though up until now we relayed on slot numbers and not on time.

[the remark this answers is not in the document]
nkryuchkov: @MatusKysel I think it's fine, I don't know how we could do this differently

message/validation/const.go
MatheusFranco99: Approved. Max possible is 1000 attestations + 512 sync committee = 1512

nkryuchkov, line 46:
```suggestion
	maxAttestations = 1000
	maxSyncCommittees = 512
	maxSizePartialSignatureMessages = maxAttestations + maxSyncCommittees
```

y0sher: @nkryuchkov can you take care of this and also fix the UT please 🙏

nkryuchkov (approved):

iurii-ssv (approved):

kchojn (approved):

message/validation/const.go
```diff
@@ -37,11 +37,18 @@ const (
 
 const (
 	partialSignatureSize           = 96
-	partialSignatureMsgSize        = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
-	maxPartialSignatureMessages    = 1000
-	partialSigMsgTypeSize          = 8 // uint64
-	maxPartialSignatureMsgsSize    = partialSigMsgTypeSize + slotSize + maxPartialSignatureMessages*partialSignatureMsgSize
-	maxEncodedPartialSignatureSize = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor + 4
+	maxSizePartialSignatureMessage = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	/// Pre fork values TODO remove after Blossom
```
nkryuchkov, line 41:
```suggestion
	// Pre-fork values TODO remove after Blossom
```

message/validation/const.go
```diff
@@ -37,11 +37,18 @@ const (
...
+	maxSizePartialSignatureMessagesOld = 1000
```
nkryuchkov, line 42: IMO `PreFork` suffix would be better. It's might be confusing what oldness is meant

message/validation/const.go
```diff
@@ -37,11 +37,18 @@ const (
...
+	maxPartialSignatureMsgsSizeOld     = partialSigMsgTypeSize + slotSize + maxSizePartialSignatureMessagesOld*maxSizePartialSignatureMessage + 4
+	maxEncodedPartialSignatureSizeOld  = maxPartialSignatureMsgsSizeOld + maxPartialSignatureMsgsSizeOld/encodingOverheadDivisor
+	///
```
nkryuchkov, line 45:
```suggestion
	// Post-fork values
```

message/validation/const.go
```diff
@@ -37,11 +37,18 @@ const (
...
+	maxAttestations                 = 1000
+	maxSyncCommittees               = syncCommitteeSize
+	maxSizePartialSignatureMessages = maxAttestations + maxSyncCommittees
+	partialSigMsgTypeSize           = 8 // uint64
```
nkryuchkov, line 49: It's used by both pre-fork and post-fork values, so let's move it above both

moshe-blox:
hey @MatheusFranco99 

can we please have an explanation here for why we don't need this change?

and do you think when should we? (at which validator count)

MatheusFranco99:
Hey @moshe-blox 

> can we please have an explanation here for why we don't need this change?

The released implementation has some caps on the number of committee duties, specifically 1512 considering 1000 validators per operator, i.e. 1000 attestation + 512 sync committees.
However, even though the contract may increase the number of validators per operator to 2000, it's very unlikely that this cap value will ever be reached.
But, even though it doesn't demand a fork on its own, if another change creates a fork, then this should be included.

> and do you think when should we? (at which validator count)

To know the number that would start to become problematic, I would need to revisit the formulas. Do you want me to check it? I mean, are we planning to increase more than 2k?

MatheusFranco99:
@MatusKysel @moshe-blox 
> Please try to re-visit with 2k and 3k, if the chances hitting it increase by that much

It seems that we are fine. I've improved the estimations and the probability that a committee duty has more than 1512 beacon duties is:

| Validators | Optimistic Estimation | Conservative Estimation |
| --- | --- | --- |
| 1000 |  $10^{-13,000}$ | $10^{-5,600}$ |
| 2000 |  $10^{-6,000}$ | $10^{-2,600}$ |
| 3000 |  $10^{-4,000}$ | $10^{-1,600}$ |

The expected number of slots such that the condition fail is $p^{-1}$. E.g. for $p = 10^{-1600}$, it would be $10^{1600}$ slots.

PR #1959, opened 2024-12-20
v2.2.0 fix(ec): enable time based out of sync tolerance

anatolie-ssv:
Introducing a 1 minute tolerance window that allows for the node to be out of sync.


eth/executionclient/execution_client.go
```diff
@@ -49,12 +49,15 @@ type ExecutionClient struct {
...
 	syncDistanceTolerance uint64
 	syncProgressFn        func(context.Context) (*ethereum.SyncProgress, error)
+	lastHealthy           time.Time
 
 	// variables
 	client *ethclient.Client
 	closed chan struct{}
 }
 
+const syncTimeTolerance = 1 * time.Minute
```
moshe-blox, line 59: should probably rename to something link unhealthyDurationTolerance

moshe-blox:
we should probably do a similar thing with the Beacon node

beacon/goclient/goclient.go
```diff
@@ -244,36 +258,51 @@ func (gc *GoClient) Healthy(ctx context.Context) error {
...
+		// within the distance but out of time
+		if unhealthyDuration > unhealthyDurationTolerance {
+			return fmt.Errorf("not synced for too long (%d): %w", unhealthyDuration, errSyncing)
+		}
 	}
+
 	if syncState.IsOptimistic {
 		gc.log.Error("Consensus client is in optimistic mode")
 		return fmt.Errorf("optimistic")
 	}
```
anatolie-ssv, line 302: not sure how to treat this `optimistic` case @moshe-blox

PR #1964, opened 2024-12-23
v2.2.0 feat(cl/el): support multiple nodes

nkryuchkov:
- [x] support multiple consensus nodes
- [x] support multiple execution nodes
- [x] adjust metrics


beacon/goclient/goclient.go
```diff
@@ -156,27 +156,72 @@ func New(
...
+			fields.Address(httpClient.Address()),
+			zap.String("client", string(ParseNodeClient(nodeVersionResp.Data))),
+			zap.String("version", nodeVersionResp.Data),
+		)
+
+		consensusClients = append(consensusClients, httpClient)
+		consensusClientsAsServices = append(consensusClientsAsServices, httpClient)
+	}
+
+	err := assertSameSpec(opt.Context, consensusClients...)
```
y0sher, line 198: we should add some comments about this and the above, the conditions to make multiple bn work together (dialable, same spec, version, etc..)

beacon/goclient/proposer.go
```diff
@@ -187,17 +189,85 @@ func (gc *GoClient) SubmitBlindedBeaconBlock(block *api.VersionedBlindedProposal
...
+	// Although we got a blinded block from one node and that node has to submit it,
+	// other nodes might know this payload too and have a chance to submit it successfully.
+	//
+	// So we do the following:
+	//
+	// Submit the blinded proposal to all clients concurrently.
+	// If any client succeeds, cancel the remaining submissions.
+	// Wait for all submissions to finish or timeout after 1 minute.
+	//
+	// TODO: Make sure this the above is correct. Should we submit only to the node that returned the block?
```
y0sher, line 219: we should test this code thoroughly. 
is it correct that other nodes might know the payload for a blinded block though?

y0sher: Trying to submit only to the node that provided the block is going to require more state management. if its not an issue to sent to all of them this solution is perferred for now.

cmd/tcpproxy/tcpproxy.go
y0sher, line 1: can you explain more about this file?

eth/executionclient/config.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // ExecutionOptions contains config configurations related to Ethereum execution client.
 type ExecutionOptions struct {
-	Addr                  string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket address"`
+	Addr                  string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket address. Supports multiple comma-separated addresses"`
```
y0sher, line 11: maybe just have `[]string` ? AFAIK viper knows how to parse this.

eth/executionclient/execution_client.go
```diff
@@ -31,76 +32,164 @@ var (
...
+	closed      chan struct{}
+	closingOnce sync.Once
 }
 
 // New creates a new instance of ExecutionClient.
 func New(ctx context.Context, nodeAddr string, contractAddr ethcommon.Address, opts ...Option) (*ExecutionClient, error) {
+	addrList := strings.Split(nodeAddr, ";")
+	if len(addrList) == 0 {
+		return nil, fmt.Errorf("no node address provided")
+	}
```
y0sher, line 66: as I said above, I think it makes more sense to pass a slice. and to the parsing on a higher level.

eth/executionclient/execution_client.go
```diff
@@ -31,76 +32,164 @@ var (
...
+		return nil, fmt.Errorf("no execution clients could be initialized")
+	}
+
+	same, err := client.assertSameChainIDs(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect to execution client: %w", err)
+		return nil, fmt.Errorf("assert same chain IDs: %w", err)
+	}
+	if !same {
+		return nil, fmt.Errorf("execution clients' chain IDs are not same")
```
y0sher, line 103: maybe give some more info, want, have?

eth/executionclient/execution_client.go
```diff
@@ -31,76 +32,164 @@ var (
...
 	client.syncProgressFn = client.syncProgress
 
 	return client, nil
 }
 
-func (ec *ExecutionClient) syncProgress(ctx context.Context) (*ethereum.SyncProgress, error) {
-	return ec.client.SyncProgress(ctx)
+// assertSameChainIDs checks if all healthy clients have the same chain ID.
+// It sets firstChainID to the chain ID of the first healthy client encountered.
+func (ec *ExecutionClient) assertSameChainIDs(ctx context.Context) (bool, error) {
```
y0sher, line 113: since this returns err, it's recommended to propagate the error message here then print on a higher level. doing both can result in a lot of redundancy in the logs.

eth/executionclient/execution_client.go
```diff
@@ -31,76 +32,164 @@ var (
...
+	return sp, nil
 }
 
 // Close shuts down ExecutionClient.
 func (ec *ExecutionClient) Close() error {
-	close(ec.closed)
-	ec.client.Close()
+	ec.closingOnce.Do(func() {
+		close(ec.closed)
+		ec.clientsMu.RLock()
```
y0sher, line 169: IMO this seems like too much concurrency primitives to do a simple thing. maybe we can just use context?

eth/executionclient/execution_client.go
```diff
@@ -31,76 +32,164 @@ var (
...
+	})
 	return nil
 }
 
 // FetchHistoricalLogs retrieves historical logs emitted by the contract starting from fromBlock.
 func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64) (logs <-chan BlockLogs, errors <-chan error, err error) {
-	currentBlock, err := ec.client.BlockNumber(ctx)
+	res, err := ec.doCall(ctx, func(ctx context.Context, mc *ManagedClient) (any, error) {
+		return mc.getClient().BlockNumber(ctx)
+	})
```
y0sher, line 184: why not wrap it up with identical interface?

eth/executionclient/execution_client.go
```diff
@@ -345,63 +457,73 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
...
+type callFunc func(ctx context.Context, mc *ManagedClient) (any, error)
 
-	logger.Info("reconnected to execution client", zap.Duration("took", time.Since(start)))
-}
+// doCall carries out a call on the healthy clients until one succeeds.
+func (ec *ExecutionClient) doCall(ctx context.Context, call callFunc) (any, error) {
+	ec.clientsMu.RLock()
+	clients := make([]*ManagedClient, len(ec.clients))
+	copy(clients, ec.clients)
+	ec.clientsMu.RUnlock()
```
y0sher, line 502: is the `ec.client` changing? why do you need to protect it? how does this copy help if its all pointers?

y0sher: also why not just use `getHealthyClient` here?

eth/executionclient/managed_client.go
```diff
@@ -0,0 +1,192 @@
...
+// disconnect closes the current client connection.
+func (mc *ManagedClient) disconnect() {
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
+
+	if mc.client != nil {
+		if mc.client.Client() != nil {
+			mc.client.Close()
+		}
+		mc.client = nil
```
y0sher, line 75: why set it to nil ? isn't setting healthy enough?

eth/executionclient/managed_client.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	mc.wg.Add(1)
+	go mc.monitorHealth()
+
+	return mc, nil
+}
+
+// connect establishes a connection to the Ethereum client.
+func (mc *ManagedClient) connect(ctx context.Context) error {
```
y0sher, line 51: if this happens only once at startup maybe the mutex protection (for the client) is not needed

protocol/v2/blockchain/beacon/client.go
```diff
@@ -63,7 +63,7 @@ type BeaconNode interface {
 type Options struct {
 	Context        context.Context
 	Network        Network
-	BeaconNodeAddr string `yaml:"BeaconNodeAddr" env:"BEACON_NODE_ADDR" env-required:"true"`
+	BeaconNodeAddr string `yaml:"BeaconNodeAddr" env:"BEACON_NODE_ADDR" env-required:"true" env-description:"Beacon node address. Supports multiple comma-separated addresses'"`
```
y0sher, line 66: the config parsing should now how to parse it to []string directly.

y0sher (changes requested):
good job :)
some questions and comments.

moshe-blox:
I suggest instead of modifying ExecutionClient implementation to be a multiclient, we should declare an interface for the Ethereum client with only the methods we call, and a standalone Multiclient implementation that implements this interface.

This would mean that our ExecutionClient will not have to be changed at all (at least theoretically).

By the way, we should also think how to handle cases where subscriptions fail. AFAIK SubscribeNewHead is called on one of the Eth clients. What if that client dies? Would we cancel the subscription and call it again on the other client?

cmd/tcpproxy/tcpproxy.go
nkryuchkov: This is just for testing. Setting up a real node looked more complex to me. so I used this TCP proxy to redirect all data to a beacon node and be able to shut it down to simulate outage. Although it doesn't simulate a case when the node is up but goes out of sync

eth/executionclient/config.go
nkryuchkov: Then it won't be backward compatible with the current values everyone uses, so it would require changing configs. Also, I think a significant amount of operators will continue using just one node, so a slice parameter might be less convenient

eth/executionclient/execution_client.go
nkryuchkov: @y0sher There are logs for each mismatch (`Chain ID mismatch`). Since there may be several clients, there may be several want's, so I thought having returning some generic error and printing a log for each mismatch would be better

eth/executionclient/config.go
nkryuchkov:
```bash
% ./bin/ssvnode start-node 
2025/01/09 18:15:53 could not create logger could not read config: config file parsing error: yaml: unmarshal errors:
  line 14: cannot unmarshal !!str `ws://pr...` into []string
```

eth/executionclient/managed_client.go
nkryuchkov: in `attemptReconnect` too

eth/executionclient/managed_client.go
nkryuchkov: I just changed the implementation to use multi-client, so I'm resolving this

eth/executionclient/execution_client.go
nkryuchkov: I just changed the implementation to use multi-client, so I'm resolving this

eth/executionclient/managed_client.go
nkryuchkov: I just changed the implementation to use multi-client, so I'm resolving this

eth/executionclient/execution_client.go
nkryuchkov: fixed

eth/executionclient/execution_client.go
nkryuchkov: done

eth/executionclient/execution_client.go
nkryuchkov: can you please explain what you mean?

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,404 @@
...
+	go func() {
+		defer close(logs)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ec.closed:
+				return
+			default:
+				f := func(client SingleClientProvider) (any, error) {
```
y0sher, line 227: this can be defined outside of the loop. even though compiler/runtime takes care of it anyway.

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,404 @@
...
+				if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) {
+					// Closed gracefully.
+					return
+				}
+				if err != nil {
+					ec.logger.Fatal("failed to stream registry events", zap.Error(err))
+					return // Terminate the loop after logging Fatal
+				}
+				// On success, terminate the loop
+				return
```
y0sher, line 252: honestly I don't know if this is simpler or clearer, just shorter.. let me know what you think
```suggestion
				if err !=nil && !(errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled)) {
					ec.logger.Fatal("failed to stream registry events", zap.Error(err))
				}
				return
```

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,404 @@
...
+				}
+
+				_, err := ec.call(ec.setMethod(ctx, "StreamLogs"), f)
+				if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) {
+					// Closed gracefully.
+					return
+				}
+				if err != nil {
+					ec.logger.Fatal("failed to stream registry events", zap.Error(err))
+					return // Terminate the loop after logging Fatal
```
y0sher, line 249: anyway no need to return after fatal since it stops the program.

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+	}
+
+	return multiErr
+}
+
+func (ec *MultiClient) call(ctx context.Context, f func(client SingleClientProvider) (any, error)) (any, error) {
+	if len(ec.clients) == 1 {
+		return f(ec.clients[0])
+	}
```
y0sher, line 348: I guess this should never happen since we check before initializing. we can leave it anyway..

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,404 @@
...
+
+	for i := 0; i < len(ec.clients); i++ {
+		ec.currClientMu.Lock()
+		currentIdx := ec.currClientIdx
+		client := ec.clients[currentIdx]
```
y0sher, line 353: is `clients` ever changing ? I don't think there's a reason to lock it.

y0sher (changes requested):
I like new version of execution multi better :) left a few comments.

eth/executionclient/multi_client.go
nkryuchkov: Only `currClientIdx` changes. Indeed, this line can be moved outside the lock

eth/executionclient/multi_client.go
nkryuchkov: It needs to be a closure because `fromBlock` is updated inside

eth/executionclient/multi_client.go
nkryuchkov: I don't think we should call `Fatal` on client closure or context cancellation because they should indicate "successful closure", e.g. graceful shutdown if the pod is restarted. There won't be probably any difference in actual behavior but I'd prefer to keep process exit codes clean

eth/executionclient/multi_client.go
nkryuchkov: I added a comment

eth/executionclient/multi_client.go
nkryuchkov: Sorry, I didn't read your suggestion well enough, so my previous message is wrong. Yes, we can do what you suggest

nkryuchkov: Changed

y0sher (approved):
lgtm!

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,399 @@
...
+	return logs
+}
+
+// Healthy returns if execution client is currently healthy: responds to requests and not in the syncing state.
+func (ec *MultiClient) Healthy(ctx context.Context) error {
+	f := func(client SingleClientProvider) (any, error) {
+		return nil, client.Healthy(ctx)
+	}
+
+	if _, err := ec.call(ec.setMethod(ctx, "Healthy"), f); err != nil {
```
moshe-blox, line 261: IIRC it's a good standard in Go to have this as a func (rather than method) called `withMethod` and the key as an empty struct type

example: https://mainfunction.org/articles/golang-context-with-value-best-practices/

(we don't need the FromContext method)

moshe-blox: if withMethod conflicts with option funcs in this package (With<Option>) then we could maybe just call it `contextWithMethod` to be explicit

nkryuchkov: Yes, I agree, it doesn't have to be a method. Why don't we need the `FromContext` method? It makes extracting value easier

moshe-blox:
Potential problem with the execution multiclient is that if one client is synced but the other isn't then we might select the lower synced client in certain method calls such as when fetching logs or when checking for healthines.

What do you think? How should we solve this? @nkryuchkov @y0sher

nkryuchkov:
@moshe-blox I agree, IIRC there used to be some handling for this in the initial implementation but I didn't port it to the multi-client. Please review the recent commits. Perhaps we also need to check cached healthiness on each iteration in `streamLogsToChan`/`FetchHistoricalLogs` to switch to another client if the current one is out of sync

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+				zap.String("address", addr))
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// FetchHistoricalLogs retrieves historical logs emitted by the contract starting from fromBlock.
+func (ec *MultiClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64) (<-chan BlockLogs, <-chan error, error) {
```
y0sher, line 127: Just fetching from a single client and then starting from scratch if is failed might be a smarter solution here. Since we sync history only on startup and its not a heavy long process, we should go with a simpler solution.

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+func (ec *MultiClient) useNextClient(ctx context.Context, currentIdx int, client SingleClientProvider) {
+	ec.currClientMu.Lock()
+	idx := ec.currClientIdx
+	// The index might have already changed if a parallel request failed.
+	if idx == currentIdx {
+		ec.currClientIdx = (ec.currClientIdx + 1) % len(ec.clients)
+	}
+	ec.currClientMu.Unlock()
+
+	go client.reconnect(ctx)
```
moshe-blox, line 461: this used to happen only in streaming, so we should either do the same, or rely on ethclient's built-in reconnection (requires properly testing it)

eth/executionclient/defaults.go
```diff
@@ -8,8 +8,12 @@ const (
 	DefaultConnectionTimeout           = 10 * time.Second
 	DefaultReconnectionInitialInterval = 1 * time.Second
 	DefaultReconnectionMaxInterval     = 64 * time.Second
+	DefaultHealthInvalidationInterval  = 10 * time.Second // TODO: decide on this value, for now choosing a bit less than block interval
```
moshe-blox, line 11: this should be double the interval of NodeProber (i think 1 slot) to let it finish its own requests and avoid doing two in parallel

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-ec.closed:
+				return
+			default:
+				// Update healthiness of all nodes and make sure at least one of them is available.
+				if err := ec.Healthy(ctx); err != nil {
+					ec.logger.Fatal("no healthy clients", zap.Error(err))
+				}
```
moshe-blox, line 251: this can be removed as it happens inside `call`

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+		}(i, client)
+	}
+	wg.Wait()
+
+	if !atLeastOneHealthy.Load() {
+		return errs
+	}
+
+	return nil
+}
```
moshe-blox, line 317: isn't it the same as just using `call` as before which should also return no error only if at least one is healthy?

nkryuchkov: @moshe-blox if all clients are healthy, `call` would check only one client. The current `Healthy` implementation calls `Healthy` on all clients to update their state

eth/executionclient/multi_client.go
nkryuchkov: yes, but we can quickly check here if there are any healthy clients to avoid proceeding if all clients are down

eth/executionclient/defaults.go
nkryuchkov: `NodeProber`'s interval is 24 seconds, so I thought it needed to be done more often. There should not be parallel requests because the healthiness is cached for a short time

eth/executionclient/multi_client.go
moshe-blox: i see, that makes sense, thanks 👍

can you pls try using https://pkg.go.dev/github.com/sourcegraph/conc/pool#ErrorPool for this? i imagine it can simplify the code here quite a bit

it should handle concurrency and join the errors for you

plus you can just check that the err returned by `Wait` isn't nil instead of having `atLeastOneHealth`

nkryuchkov: @moshe-blox great idea, as I understand it's similar to https://pkg.go.dev/golang.org/x/sync/errgroup, right? Do you have any objections against using `errgroup` instead?

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+			continue
+		}
+
+		ec.logger.Debug("call succeeded, returning value",
+			zap.String("method", methodFromContext(ctx)),
+			zap.String("addr", ec.nodeAddrs[currentIdx]))
+		return v, nil
+	}
+
+	return nil, fmt.Errorf("all clients returned an error")
```
moshe-blox, line 449: can we pls return the joined errors? this is what the callers will see and it needs to be descriptive

fmt.Errorf("all clients failed: %w", ..)

nkryuchkov: done

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,476 @@
...
+		if _, err := ec.call(contextWithMethod(ctx, "FetchHistoricalLogs [process logs]"), processLogsFunc); err != nil {
+			errCh <- err
+		}
+	}()
+
+	return logsCh, errCh, nil
+}
+
+// StreamLogs subscribes to events emitted by the contract.
+func (ec *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan BlockLogs {
```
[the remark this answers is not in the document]
nkryuchkov: done

eth/executionclient/execution_client.go
```diff
@@ -249,11 +288,33 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
...
+	} else {
+		// Reset the healthy channel if it was closed.
+		select {
+		case <-ec.healthy:
+		default:
+			ec.healthy = make(chan struct{})
+		}
+	}
+	return errors.Join(ErrUnhealthy, err)
+}
```
moshe-blox, line 310: i think if we add lock & defer unlock at the beginning of the func, it should solve the parallel request issue

the first parallel caller with lock and update time, the other caller will wait for lock and see that time updated and return immediately

eth/executionclient/multi_client.go
nkryuchkov: Actually, I'm not sure if it will look better because we need to return no error if at least one client is healthy, so it would require parsing the combined error

eth/executionclient/execution_client.go
nkryuchkov: done

y0sher (approved):

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+		}
+	}()
+
+	return logs
+}
+
+// Healthy returns if execution client is currently healthy: responds to requests and not in the syncing state.
+func (mc *MultiClient) Healthy(ctx context.Context) error {
+	atLeastOne := atomic.Bool{}
```
nkryuchkov, line 204: @moshe-blox @y0sher I think `atLeastOne` is not clear enough, let's rename it to `atLeastOneHealthy`?

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+			atLeastOne.Store(true)
+			return nil
+		})
+	}
+	err := p.Wait()
+	if atLeastOne.Load() {
+		return nil
+	}
+	return fmt.Errorf("no healthy clients: %w", err)
```
nkryuchkov, line 223: @moshe-blox @y0sher probably would be a bit more readable
```suggestion
	if atLeastOne.Load() {
		return nil
	}
	return fmt.Errorf("no healthy clients: %w", p.Wait())
```

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,365 @@
...
+	logBatchSize                uint64
+	syncDistanceTolerance       uint64
+
+	contractAddress ethcommon.Address
+	chainID         *big.Int
+	closed          chan struct{}
+
+	nodeAddrs          []string
+	clients            []SingleClientProvider
+	currentClientIndex atomic.Int64
```
nkryuchkov, line 42: really not a big deal, but perhaps `Uint64` would describe an index better

beacon/goclient/proposer.go
```diff
@@ -187,17 +189,85 @@ func (gc *GoClient) SubmitBlindedBeaconBlock(block *api.VersionedBlindedProposal
...
+			}
+
+			if result.err != nil {
+				gc.log.Debug("Consensus client returned an error while submitting blinded proposal. As at least one node must submit successfully, it's expected that some nodes may fail to submit.",
+					zap.String("api", "SubmitBlindedProposal"),
+					zap.String("client_addr", result.address),
+					zap.Error(result.err),
+				)
+			} else {
+				cancel()
```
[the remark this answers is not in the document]
moshe-blox: actually that wouldnt be enough because theres a `defer cancel()` lol

so we must wait for all

eth/executionclient/multi_client.go
y0sher: but you must wait before you load the int to know that all goroutines are done.

eth/executionclient/multi_client.go
y0sher: since indexing in go is an `int` by default which is int64 , I thought it'll be an easier cast than uint.

eth/executionclient/multi_client.go
nkryuchkov: yes, you're right, maybe something like this to follow the happy path?

```suggestion
	if err := p.Wait(); err != nil && !atLeastOne.Load() {
			return fmt.Errorf("no healthy clients: %w", err)
	}
	return nil
```

eth/executionclient/multi_client.go
```diff
@@ -0,0 +1,365 @@
...
+				zap.String("next_addr", mc.nodeAddrs[nextClientIndex]),
+				zap.Error(err))
+
+			allErrs = errors.Join(allErrs, err)
+			mc.currentClientIndex.Store(int64(nextClientIndex)) // Advance.
+			continue
+		}
+
+		// Update currentClientIndex to the successful client.
+		mc.currentClientIndex.Store(int64(clientIndex))
```
nkryuchkov, line 346: Probably, there could be a case in which a node fails on some of the requests (e.g. some API subset is disabled), so the clients would be switched back and forth. But I'm not sure we need to do something with it now

beacon/goclient/proposer.go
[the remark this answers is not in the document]
nkryuchkov: fixed, please review

eth/executionclient/multi_client.go
moshe-blox: @nkryuchkov your first comment here is very readable IMO

```go
	err := p.Wait()
	if healthyClients.Load() > 0 {
		return nil
	}
	return fmt.Errorf("no healthy clients: %w", err)
```
i pushed this to avoid "atLeastOne" naming which i cant think of a good name for lol
(also did this in SubmitBlindedBeaconBlock)

WDYT?

beacon/goclient/goclient.go
```diff
@@ -190,43 +234,116 @@ func New(
...
+	}
+
+	for _, param := range paramsToCheck {
+		if a[param] != b[param] {
+			return fmt.Errorf("param %s mismatch, got %v and %v", param, a[param], b[param])
+		}
+	}
+
+	return nil
 }
```
moshe-blox, line 323: i dont know the benefit of checking all of these

checking just `GENESIS_FORK_VERSION` should be sufficient, as it's supposed to be the "network ID"

moshe-blox: i see that this variable is in genesis data, so essentially sameSpec isnt necessary

nkryuchkov: I agree, done

eth/executionclient/multi_client.go
nkryuchkov: @moshe-blox LGTM

moshe-blox (approved):

PR #1967, opened 2024-12-26
fix(storage/shares): use transaction when saving metadata

anatolie-ssv:


moshe-blox:
great idea :)

moshe-blox:
just noticed SetMany is using a BadgerDB "batch", do you think there's any advantage in using a transaction for only 1 batch?

anatolie-ssv:
> just noticed SetMany is using a BadgerDB "batch", do you think there's any advantage in using a transaction for only 1 batch?

yes I just noticed too that "batch" writes use a transaction under the hood, so likely this approach is redundant...

PR #1968, opened 2024-12-28
all: avoid passing logger to methods

nkryuchkov:
as per title


y0sher (approved):
gj, we should do this code-wide :)

nkryuchkov:
@y0sher I did it code-wide, although I might have missed something. Anyway, we'll be able to fix leftovers if we find them. Please review it again

operator/duties/scheduler_mock.go
oleg-ssvlabs, line 1: Did you update this file manually or it was regenerated with `mockgen`?

oleg-ssvlabs (approved):
Awesome, these are certainly great changes!  
An alternative approach could be having a separate `logger` package that encapsulates Zap and is loaded as a package (rather than being injected through a constructor or function/method). 
Something like this:

```Go
package logger

import "go.uber.org/zap"

var logger = zap.L()

func Info(msg string, fields ...zap.Field) {
	logger.Info(msg, fields...)
}
```
And the usage:

```Go
package main

import "github.com/ssvlabs/utils/logger"

func main() {
	logger.Info("Hello, world!")
}
```

However, if we want to have named loggers at the object level or loggers with additional properties applicable to all log records, we would likely still need to inject it through a constructor.

anatolie-ssv (approved):

ayaz-ssvlabs (approved):

operator/duties/scheduler_mock.go
nkryuchkov: @oleg-ssvlabs with `mockgen`

iurii-ssv (approved):
Good stuff!

iurii-ssv (approved):

operator/duties/scheduler.go
```diff
@@ -404,10 +405,10 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, logger *zap.Logger, dutie
 }
 
 // ExecuteCommitteeDuties tries to execute the given committee duties
-func (s *Scheduler) ExecuteCommitteeDuties(ctx context.Context, logger *zap.Logger, duties committeeDutiesMap) {
+func (s *Scheduler) ExecuteCommitteeDuties(ctx context.Context, duties committeeDutiesMap) {
 	for _, committee := range duties {
 		duty := committee.duty
-		logger := s.loggerWithCommitteeDutyContext(logger, committee)
+		logger := s.loggerWithCommitteeDutyContext(committee) // TODO: extract this in dutyExecutor (validator controller), don't pass logger to ExecuteCommitteeDuty
```
olegshmuelov, line 411: I agree with the TODO here, passing a `logger` into `ExecuteCommitteeDuty` isn’t ideal, especially considering how and where it’s used.

**Just to add:** `ExecuteCommitteeDuty` is implemented in the validator controller (i.e., a different package). This logger is then passed deeper into the `Committee`, where it’s still labeled from the `DutyScheduler`, making logs misleading in terms of origin and responsibility.

Example:
```go
logger := s.loggerWithCommitteeDutyContext(committee) // From scheduler
s.dutyExecutor.ExecuteCommitteeDuty(ctx, logger, committee.id, duty)
```

Eventually ends up in:
```go
func (c *Committee) OnExecuteDuty(ctx context.Context, logger *zap.Logger, msg *types.EventMsg) error {
         ...
	if err := c.StartDuty(ctx, logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee duty: %w", err)
	}

	if err := c.StartConsumeQueue(logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee consume queue: %w", err)
	}
         ...
}
```
So now we’ve got logs coming from inside `Committee`, but they're still labeled as if they came from the `DutyScheduler`.

Ideally, the `controller` / `Committee` (probably they already do) should build and manage their own logger internally using the proper context (e.g., `controller.logger.With(...)` or via a stored logger field).

We should create a GitHub issue to track and properly refactor this, so that the logger is owned and managed within the controller layer, not passed from upstream logic.

olegshmuelov (approved):
I agree with the TODO here, passing a `logger` into `ExecuteCommitteeDuty` isn’t ideal, especially considering how and where it’s used.

**Just to add:** `ExecuteCommitteeDuty` is implemented in the validator controller (i.e., a different package). This logger is then passed deeper into the `Committee`, where it’s still labeled from the `DutyScheduler`, making logs misleading in terms of origin and responsibility.

Example:
```go
logger := s.loggerWithCommitteeDutyContext(committee) // From scheduler
s.dutyExecutor.ExecuteCommitteeDuty(ctx, logger, committee.id, duty)
```

Eventually ends up in:
```go
func (c *Committee) OnExecuteDuty(ctx context.Context, logger *zap.Logger, msg *types.EventMsg) error {
         ...
	if err := c.StartDuty(ctx, logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee duty: %w", err)
	}

	if err := c.StartConsumeQueue(logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee consume queue: %w", err)
	}
         ...
}
```
So now we’ve got logs coming from inside `Committee`, but they're still labeled as if they came from the `DutyScheduler`.

Ideally, the `controller` / `Committee` (probably they already do) should build and manage their own logger internally using the proper context (e.g., `controller.logger.With(...)` or via a stored logger field).

We should create a GitHub issue to track and properly refactor this, so that the logger is owned and managed within the controller layer, not passed from upstream logic.

operator/duties/scheduler.go
olegshmuelov: > We should create a GitHub issue to track and properly refactor this, so that the logger is owned and managed within the controller layer, not passed from upstream logic.

https://github.com/ssvlabs/ssv/issues/2100

y0sher:
Please @nkryuchkov resolve conflicts, I like this change a lot, but we need to make sure we are not losing any context we had previously by not passing some logger into functions. (I see we left it as is in dutySchuedler like @olegshmuelov mentioned to not drop context)

nkryuchkov:
> Please @nkryuchkov resolve conflicts, I like this change a lot, but we need to make sure we are not losing any context we had previously by not passing some logger into functions. (I see we left it as is in dutySchuedler like @olegshmuelov mentioned to not drop context)

resolved

PR #1969, opened 2024-12-29
refactor: optimize subnet calculations using big.Int for reduced memory usage

y0sher:
### Description

I observed high memory usage and a lot of allocations stemming from the `CommitteeSubnet` code used to calculate the subnets for which validators we should pull metadata for (Changed in #1787)

### Changes

- Initialized a global `SubnetsCount` `big.Int` representation to reduce allocations.
- Created `SetCommitteeSubnet` which receives a `big.Int` instance and uses it instead of allocating new ints.
- Used the new function in all the metadata collection loop calls to CommitteeSubnet.

### DIFF
--- a/network/commons/common.go
+++ b/network/commons/common.go
@@ -25,6 +25,13 @@ const (
 	topicPrefix = "ssv.v2"
 )
 
+// BigIntSubnetsCount is the big.Int representation of SubnetsCount
+var bigIntSubnetsCount *big.Int
+
+func init() {
+	bigIntSubnetsCount = new(big.Int).SetUint64(SubnetsCount)
+}
+
 const (
 	signatureSize    = 256
 	signatureOffset  = 0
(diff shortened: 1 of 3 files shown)


Roy-blox (approved):

network/commons/common.go
```diff
@@ -57,10 +64,16 @@ func GetTopicBaseName(topicName string) string {
...
-	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), new(big.Int).SetUint64(SubnetsCount))
+	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), bigIntSubnetsCount)
 	return subnet.Uint64()
 }
 
+// SetCommitteeSubnet returns the subnet for the given committee, it doesn't allocate memory but uses the passed in big.Int
+func SetCommitteeSubnet(bigInst *big.Int, cid spectypes.CommitteeID) uint64 {
+	bigInst.SetBytes(cid[:])
+	return bigInst.Mod(bigInst, bigIntSubnetsCount).Uint64()
+}
```
moshe-blox, line 75: i think it would make more sense if it just set the value and not returned it

the caller can just call Uint64() and it would emphasize the big.Int is being modified

that's less convenient, but it's an optimized func so it's OK imo

operator/validator/controller.go
```diff
@@ -508,15 +511,22 @@ func (c *controller) StartValidators() {
 	}
 }
 
-func (c *controller) selfSubnets() records.Subnets {
+// selfSubnets calculates the operator's subnets by adding up the fixed subnets and the active committees
+// it recvs big int to avoid unnecessary conversions, if intCid is nil it will allocate a new big int
+func (c *controller) selfSubnets(intCid *big.Int) records.Subnets {
```
moshe-blox, line 516: in this context its just a buffer and has no usage outside this func, so to avoid confusion i'd call it just "buffer" and clarify it's for re-using memory (don't have to go into much details)

moshe-blox: taking inspiration from https://pkg.go.dev/bytes#Buffer

which has a similar pattern to what you did where if you dont pass `buf` it will allocate for you

moshe-blox (approved):
LGTM 🔥

PR #1973, opened 2025-01-05
spec alignment (v1.0.2): custom GasLimit , 1000 validators

y0sher:
### Description

This PR updates and aligns to spec version of v1.0.2

- It updates the test to support spec change of allowing bigger message size to support 1000 validators. (This change won't take effect until contract updates, so don't try)

- Following the efforts of some Ethereum community members to increase the Gas Limit, we enable this feature in our SSV operator node. 

####  Gas Limit
The SSV Operator node only uses gas limit when registering to MEV relay. It will be used for external blocks built using the consensus client with MEV. The execution node limits are set on the nodes themselves and will be used on local blocks.  (check out https://pumpthegas.org/)

Setting it using : 
```yaml
ValidatorOptions:
    GasLimit: 35000000
```
or set the env var `GAS_LIMIT` to set the Gas Limit used for validator registrations. you should be able to check the validator registration gas fee in the relay using this endpoint:
`<relay-url>/relay/v1/data/validator_registration?pubkey=<validator pk>`


moshe-blox (approved):

protocol/v2/blockchain/beacon/client.go
```diff
@@ -52,6 +52,7 @@ type signer interface {
 // BeaconNode interface for all beacon duty calls
 type BeaconNode interface {
 	specssv.BeaconNode // spec beacon interface
+	GasLimit() uint64
```
y0sher, line 55: this is not the prettiest way to expose `GasLimit`, but its already passed to the client. exposing it directly to the running will require changing more critical path code like the runner creation and options.

liorrutenberg:
@y0sher @moshe-blox approved but doesnt pass spec-test/alignment?

liorrutenberg (changes requested):
fails on tests

y0sher:
Since the scope of the PR extended, I moved it here - https://github.com/ssvlabs/ssv/pull/1973

PR #1978, opened 2025-01-09
feat(eventsyncing): health calls check last seen block distance

anatolie-ssv:
Enhance `event syncer` component to verify the distance between `lastProceessedBlock` time against a staleness threshold during `Health` checks.


operator/storage/storage.go
```diff
@@ -33,6 +34,8 @@ type Storage interface {
 
 	SaveLastProcessedBlock(rw basedb.ReadWriter, offset *big.Int) error
 	GetLastProcessedBlock(r basedb.Reader) (*big.Int, bool, error)
+	SaveHighestSeenBlock(rw basedb.ReadWriter, offset *big.Int) error
+	GetHighestSeenBlock(r basedb.Reader) (*big.Int, bool, error)
```
moshe-blox, line 38: why do we need a new field? can't we use the last processed block?

eth/eventsyncer/event_syncer.go
```diff
@@ -80,6 +85,29 @@ func (es *EventSyncer) Healthy(ctx context.Context) error {
...
+		return fmt.Errorf("failed to read last processed block: %w", err)
+	}
+	if !found || highestSeenBlock == nil || highestSeenBlock.Uint64() == 0 {
+		return fmt.Errorf("last seen block is not set")
+	}
+
+	return es.blockBelowThreashold(ctx, highestSeenBlock)
+}
+
+func (es *EventSyncer) blockBelowThreashold(ctx context.Context, block *big.Int) error {
```
moshe-blox, line 100:
```suggestion
func (es *EventSyncer) blockBelowThreshold(ctx context.Context, block *big.Int) error {
```

eth/eventsyncer/event_syncer.go
```diff
@@ -80,6 +85,29 @@ func (es *EventSyncer) Healthy(ctx context.Context) error {
...
+	header, err := es.executionClient.HeaderByNumber(ctx, block)
+	if err != nil {
+		return fmt.Errorf("failed to get header for block %d: %w", block, err)
+	}
+
+	// #nosec G115
+	if header.Time != 0 && header.Time < uint64(time.Now().Add(-es.stalenessThreshold).Unix()) {
```
moshe-blox, line 107: if header.Time is 0 something is wrong we should treat it as block is too old imo (unless we find out its a legit response and handle it differently)

y0sher:
Please add some description :pray:

eth/eventsyncer/event_syncer.go
```diff
@@ -110,6 +130,13 @@ func (es *EventSyncer) SyncHistory(ctx context.Context, fromBlock uint64) (lastP
 		return 0, fmt.Errorf("event replay: lastProcessedBlock (%d) is lower than fromBlock (%d)", lastProcessedBlock, fromBlock)
 	}
 
+	// Check if the block is too old.
+	// #nosec G115
+	b := big.NewInt(int64(es.lastProcessedBlock))
```
y0sher, line 135: I get the impression that this is safer, and you don't need to convert uint to int.
```suggestion
	b := new(big.Int).SetUint64(es.lastProcessedBlock)
```

eth/executionclient/logs.go
```diff
@@ -9,6 +9,7 @@ import (
 // BlockLogs holds a block's number and it's logs.
 type BlockLogs struct {
 	BlockNumber uint64
+	LastSeen    uint64
```
y0sher, line 12: where is this used?

y0sher (approved):

eth/eventsyncer/event_syncer.go
```diff
@@ -80,6 +85,21 @@ func (es *EventSyncer) Healthy(ctx context.Context) error {
...
+func (es *EventSyncer) blockBelowThreshold(ctx context.Context, block *big.Int) error {
+	header, err := es.executionClient.HeaderByNumber(ctx, block)
+	if err != nil {
+		return fmt.Errorf("failed to get header for block %d: %w", block, err)
+	}
+
+	// #nosec G115
+	if header.Time > 0 && header.Time < uint64(time.Now().Add(-es.stalenessThreshold).Unix()) {
+		return fmt.Errorf("block %d is too old", block)
+	}
```
moshe-blox, line 101:
```suggestion
	if header.Time < uint64(time.Now().Add(-es.stalenessThreshold).Unix()) {
		return fmt.Errorf("block %d is too old (timestamp %d)", block, header.Time)
	}
```
maybe this could work to not ignore time == 0 (which is a failure)

eth/eventsyncer/event_syncer.go
```diff
@@ -80,6 +85,21 @@ func (es *EventSyncer) Healthy(ctx context.Context) error {
...
+}
+
+func (es *EventSyncer) blockBelowThreshold(ctx context.Context, block *big.Int) error {
+	header, err := es.executionClient.HeaderByNumber(ctx, block)
+	if err != nil {
+		return fmt.Errorf("failed to get header for block %d: %w", block, err)
+	}
+
+	// #nosec G115
+	if header.Time > 0 && header.Time < uint64(time.Now().Add(-es.stalenessThreshold).Unix()) {
```
moshe-blox, line 99:
```suggestion
	if time.Since(time.Unix(header.Time, 0)) < es.stalenessThreshold) {
```

moshe-blox: > we cant ignore zero time because it may also happen because of an Eth client bug (we've actually witnessed a similar bug before with another field)
> do we create the blocks in the simulator? if so we can probably control the time (ideally)
> if we have no choice, we could set stalenessThreshold to MaxUint64 in only the failing tests

moshe-blox (approved):
LGTM, well done! 💪
 
@anatolie-ssv  Let's make sure to sync @andrew-blox and @Roy-blox on what to look for when testing this, for example: syncing a cluster from scratch.

PR #1979, opened 2025-01-13
main - spec alignment (v1.0.2): custom GasLimit , 1000 validators #1973 

y0sher:
### Description

This PR updates and aligns to spec version of v1.0.2

- It updates the test to support spec change of allowing bigger message size to support 1000 validators. (This change won't take effect until contract updates, so don't try)

- Following the efforts of some Ethereum community members to increase the Gas Limit, we enable this feature in our SSV operator node. 

####  Gas Limit
The SSV Operator node only uses gas limit when registering to MEV relay. It will be used for external blocks built using the consensus client with MEV. The execution node limits are set on the nodes themselves and will be used on local blocks.  (check out https://pumpthegas.org/)

Setting it using : 
```yaml
ValidatorOptions:
    GasLimit: 35000000
```
or set the env var `GAS_LIMIT` to set the Gas Limit used for validator registrations. you should be able to check the validator registration gas fee in the relay using this endpoint:
`<relay-url>/relay/v1/data/validator_registration?pubkey=<validator pk>`

#### ⚠️  **Caution** ⚠️  
This update enables changing the gas limit for all the MEV proposals of the validators of the operator. ALL operators of the committee MUST set the same gas limit. Otherwise **validator registrations will fail, changes to fee recipient might not take effect.**


moshe-blox (approved):
LGTM 🔥

PR #1982, opened 2025-01-16
fix(exporter): missing identifier value

anatolie-ssv:
Log line refers to a previously removed field causing a `panic`. Fixed by passing the missing value.

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -379,10 +379,8 @@ func (c *controller) handleWorkerMessages(msg network.DecodedSSVMessage) error {
 			SyncCommRoots:     c.syncCommRoots,
 			DomainCache:       c.domainCache,
 		}
-		msgID := ssvMsg.GetID()
-		identifier := msgID[:]
 		ncv = &committeeObserver{
-			CommitteeObserver: validator.NewCommitteeObserver(identifier, committeeObserverOptions),
+			CommitteeObserver: validator.NewCommitteeObserver(ssvMsg.GetID(), committeeObserverOptions),
 		}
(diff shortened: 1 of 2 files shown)


y0sher (approved):

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -32,7 +32,7 @@ type CommitteeObserver struct {
 	Storage                *storage.ParticipantStores
 	beaconNetwork          beacon.BeaconNetwork
 	networkConfig          networkconfig.NetworkConfig
-	qbftController         *qbftcontroller.Controller
+	identifier             []byte
```
nkryuchkov, line 35: can we store the `MessageID` type?

operator/validator/controller.go
```diff
@@ -379,8 +379,10 @@ func (c *controller) handleWorkerMessages(msg network.DecodedSSVMessage) error {
 			SyncCommRoots:     c.syncCommRoots,
 			DomainCache:       c.domainCache,
 		}
+		msgID := ssvMsg.GetID()
+		identifier := msgID[:]
 		ncv = &committeeObserver{
-			CommitteeObserver: validator.NewCommitteeObserver(committeeObserverOptions),
+			CommitteeObserver: validator.NewCommitteeObserver(identifier, committeeObserverOptions),
```
nkryuchkov, line 385: if we store it as `MessageID` inside the committee observer (see my other comment), we can just pass `ssvMsg.GetID()`

protocol/v2/ssv/validator/non_committee_validator.go
anatolie-ssv: yep, fixed

moshe-blox (approved):

PR #1983, opened 2025-01-16
fix(eventsyncer): remove check for block below threshold in the sync history method

anatolie-ssv:
temporarily remove the check while a proper solution is implemented


iurii-ssv (approved):
LGTM, and "sync from scratch" seems to be working now on **stage** with that change

moshe-blox (approved):

PR #1986, opened 2025-01-17
fix(eventsyncer): retry if historical sync finished an on old block

anatolie-ssv:
During historical sync (from scratch) we can end up on a block that - by the time the syncing is done - is too old.

This PR enables the historical sync to be retried until it finishes on a block that is within the tolerance window.


y0sher:
@anatolie-ssv , @nkryuchkov 
this PR might clash with #1964 , please lets coordinate on this.
Basically, we shouldn't retry for ever, since it might be stuck while we have another healthy client.

anatolie-ssv:
> @anatolie-ssv , @nkryuchkov this PR might clash with #1964 , please lets coordinate on this. Basically, we shouldn't retry for ever, since it might be stuck while we have another healthy client.

but wouldn't a stuck loop break in this check?
```go
if lastProcessedBlock == prevProcessedBlock {
	break
}
```
but I'll reach out to Nikita and talk it over!

moshe-blox:
This is close, a few notes:

- When block stops advancing, we're breaking and returning success, skipping the staleness check (false positive)
- If the EL is syncing from genesis and now at a really old block, we'll keep looping for hours/days until it's done, however maybe we prefer to crash and let the user know their block is too old right away?

I committed a draft to address this, please let me know if it looks OK.

moshe-blox (approved):

eth/eventsyncer/event_syncer.go
```diff
@@ -105,36 +105,55 @@ func (es *EventSyncer) blockBelowThreshold(ctx context.Context, block *big.Int)
...
+		if lastProcessedBlock == 0 {
+			return 0, fmt.Errorf("handle historical block events: lastProcessedBlock is 0")
+		}
+		if lastProcessedBlock < fromBlock {
+			// Event replay: this should never happen!
+			return 0, fmt.Errorf("event replay: lastProcessedBlock (%d) is lower than fromBlock (%d)", lastProcessedBlock, fromBlock)
+		}
+
+		err = es.blockBelowThreshold(ctx, new(big.Int).SetUint64(lastProcessedBlock))
+		if err == nil {
```
nkryuchkov, line 137: should we log it (maybe `DEBUG`) if it's not `nil`?

nkryuchkov (approved):

eth/eventsyncer/event_syncer.go
anatolie-ssv: this is taken care of by the `INFO` logger a few lines below.

anatolie-ssv:
> This is close, a few notes:
> 
>     * When block stops advancing (the `==` case), we're breaking and returning success, skipping the staleness check (false positive)
> 
>     * Given there's no max retries: if the EL is syncing from genesis and now at a really old block, we'll keep looping for hours/days until it's done, however maybe we prefer to crash and let the user know their block is too old right away?
> 
> 
> I committed a draft to address this, please let me know if it looks OK.

Yes, it does make sense, the only change I would suggest is to perform the 'stuck check'
```
Not advancing, so can't sync any further
``` 
a bit earlier just before checking for `block too old`

moshe-blox (approved):

moshe-blox:
LGTM 🔥

PR #1991, opened 2025-01-20
Linter config

### LINKED ISSUE #1990: Go Linter: outdated configuration file
Currently, the `.golangci.yaml` file is significantly outdated and results in [multiple warnings](<https://github.com/ssvlabs/ssv/actions/runs/12865864305/job/35867262617#step:6:9>). This can lead to ignored linter rules and degraded code quality. 

**Suggestion:** Review the broken configuration and align it with the latest linter version.

oleg-ssvlabs:
Fixes #1990 

Not all of the deprecated configurations were replaced, so I had to remove them. 
In general, I would simplify this file further, following these rules:
- Set configuration values only if they differ from the default. 
- Configure settings only enabled linters, no need to set configurations for linters that are disabled (only 10 are enabled)
- Avoid comments, as they may eventually become misleading. 

Although, for now, it seems all warnings were fixed


.github/workflows/lint.yml
```diff
@@ -4,24 +4,19 @@ on:
   push:
     branches:
       - "**"
-#  pull_request:
-#    branches:
-#      - "**"
-
-  workflow_dispatch:
```
oleg-ssvlabs, line 11: Let me know if we want this workflow to be dispatchable

anatolie-ssv (approved):

y0sher (approved):

iurii-ssv (approved):

PR #1992, opened 2025-01-21
fix(pubsub): increase subscription buffer size to 256 and update go-libp2p-pubsub

olegshmuelov:
1. Increase Subscription Buffer Size to **256**:
   - This is a follow-up to [PR #1989](https://github.com/ssvlabs/ssv/pull/1989), where the subscription buffer size was increased to 128. That change significantly reduced message drops and improved performance.
   - However, a small number of `"Can't deliver message to subscription"` logs persisted during high traffic bursts. Increasing the buffer size to **256** provides additional capacity to handle message bursts more effectively and further reduces the likelihood of drops.
2. Update `go-libp2p-pubsub` to **v0.12.0**:


y0sher (approved):

moshe-blox (approved):
Great performance boost 💪

PR #1996, opened 2025-01-21
executionclient: add a comment about SubscribeNewHead choice

nkryuchkov:
Without the comment, it's not clear why we didn't choose `SubscribeFilterLogs`

### DIFF
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -319,6 +319,10 @@ func (ec *ExecutionClient) isClosed() bool {
 func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- BlockLogs, fromBlock uint64) (lastBlock uint64, err error) {
 	heads := make(chan *ethtypes.Header)
 
+	// Generally, execution client can stream logs using SubscribeFilterLogs, but we chose to use SubscribeNewHead + FilterLogs.
+	// We must receive all events as they determine the state of the ssv network, so a discrepancy can result in slashing.
+	// Therefore, we decided to implement more atomic behaviour, where we can revert the tx if there was an error in processing all the events of a block.
+	// So we can restart from this block once everything is good. Doing this based on the event stream was a bit harder.
 	sub, err := ec.client.SubscribeNewHead(ctx, heads)
 	if err != nil {
 		ec.logger.Error(elResponseErrMsg,
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -320,9 +320,20 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
 	heads := make(chan *ethtypes.Header)
 
 	// Generally, execution client can stream logs using SubscribeFilterLogs, but we chose to use SubscribeNewHead + FilterLogs.
+	//
 	// We must receive all events as they determine the state of the ssv network, so a discrepancy can result in slashing.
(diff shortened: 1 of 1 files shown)


eth/executionclient/execution_client.go
```diff
@@ -319,6 +319,10 @@ func (ec *ExecutionClient) isClosed() bool {
 func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- BlockLogs, fromBlock uint64) (lastBlock uint64, err error) {
 	heads := make(chan *ethtypes.Header)
 
+	// Generally, execution client can stream logs using SubscribeFilterLogs, but we chose to use SubscribeNewHead + FilterLogs.
+	// We must receive all events as they determine the state of the ssv network, so a discrepancy can result in slashing.
+	// Therefore, we decided to implement more atomic behaviour, where we can revert the tx if there was an error in processing all the events of a block.
+	// So we can restart from this block once everything is good. Doing this based on the event stream was a bit harder.
```
moshe-blox, line 325: atomicity refers to how we save to the database, which could be done with streaming also

the specific bugs we had with streaming were because you can't specify the `fromBlock`:
- you first sync history from genesis to block 100, but then stream sometimes starts late at 102 (missed 101)
- you inevitably miss blocks during any stream connection interruptions (such as EL restarts)

nkryuchkov: updated, please review

oleg-ssvlabs: > atomicity refers to how we save an entire block to the database in one tx, which could be done with streaming also
> 
> the specific bugs we had with streaming were because of missing blocks:
> 
> * you first sync history from genesis to block 100, but then stream sometimes starts late at 102 (missed 101)
> * you inevitably miss blocks during any stream connection interruptions (such as EL restarts)
> 
> which is because you can't specify the `fromBlock` with streaming, it always starts at the highest

it's a very valid point. 
It might be solvable in this way:
- Subscribe to log streams
- Check if the block is latest
- If block is not latest, backfill _until the block received from the stream_ is reached. While doing that, block the streaming go channel
- Once backfilled, process the log received from the go channel. Resume stream listener.

Although, there are pitfalls, like - how long the channel could be blocked? Aren't event logs are missed while it's blocked? It's probably not a buffered channel and even if it is buffered, what happens if the buffer cap is reached? Also, with Streams, it seems, there is no way to know what is the `last processed block `. The last processed will be last block where event was published, but if the event was published 100 blocks back? All things considered, I think current implementation is just simpler and more robust.
If you understand what I mean :)

oleg-ssvlabs (approved):

eth/executionclient/execution_client.go
moshe-blox: @oleg-ssvlabs i understand, but what would be the point of that, if you have to anyway fetch blocks individually to fill streaming gaps, then why not just do only this to keep the system simper and more predictable? (rather than alternate between 2 mechanisms)

oleg-ssvlabs: Fetching blocks would only be needed for backfilling. Let's say, application was down for some time and missed some of the blocks. During the start up it would backfill that gap and once it is done, it would subscribe to the stream. During the entire runtime of the application, it would listen to the stream (which potentially can be days/weeks/months) instead of polling for blocks, which is more efficient from the network traffic perspective(though not sure how Subscribe works under the hood? Maybe it just polls?).
Again, it's just an alternative, it is not better. It has its trade offs. It would be a very strong option if Subscribe could take `startFromBlock` parameter

y0sher (approved):

PR #2000, opened 2025-01-22
hotfix(el/cl): allow multi clients to start if at least one node is up

y0sher:
https://github.com/ssvlabs/ssv/pull/1964 was merged but it doesn't allow nodes to be unreachable on start. The idea was that when we have issues with nodes, it's always the out-of-sync issue and the node is still reachable, and we want to catch cases when we provide a bad node URL. However, it makes testing more complicated as we cannot shut down the node, we have to make it desynced instead. Also, there may be cases when a node is down due to some issue like infra issues, so this PR allows SSV node to start if at least one node is up


eth/executionclient/multi_client.go
```diff
@@ -203,16 +234,35 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan
...
 				mc.logger.Warn("client is not healthy",
 					zap.String("addr", mc.nodeAddrs[i]),
 					zap.Error(err))
+
+				mc.clientsMu[i].Unlock()
 				return err
 			}
 			healthyClients.Add(1)
+
+			mc.clientsMu[i].Unlock()
```
y0sher, line 265: it doesn't really makes sense to do goroutine for each if we look everything..

y0sher: Oh its a mutex per client.

eth/executionclient/multi_client.go
```diff
@@ -203,16 +234,35 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan
 func (mc *MultiClient) Healthy(ctx context.Context) error {
 	healthyClients := atomic.Int32{}
 	p := pool.New().WithErrors().WithContext(ctx)
-	for i, client := range mc.clients {
+
+	for i := range mc.clients {
+		i := i
 		p.Go(func(ctx context.Context) error {
-			err := client.Healthy(ctx)
+			mc.clientsMu[i].Lock()
```
y0sher, line 241: why not defer Unlock here?

eth/executionclient/multi_client.go
```diff
@@ -203,16 +234,35 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan
...
+
+			if mc.clients[i] == nil {
+				if err := mc.connect(ctx, i); err != nil {
```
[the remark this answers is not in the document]
nkryuchkov: We do it in two places:

1.
```go
			if mc.clients[i] == nil {
				if err := mc.connect(ctx, i); err != nil {
					mc.logger.Warn("failed to connect to client",
						zap.String("addr", mc.nodeAddrs[i]),
						zap.Error(err))

					mc.clientsMu[i].Unlock()
					return err
				}
			}
```

2. 
```go
		if client == nil {
			if err := mc.connect(ctx, i); err != nil {
				mc.logger.Warn("failed to connect to client",
					zap.String("addr", mc.nodeAddrs[i]),
					zap.Error(err))

				allErrs = errors.Join(allErrs, err)
				mc.currentClientIndex.Store(int64(nextClientIndex)) // Advance.
				mc.clientsMu[i].Unlock()
				continue
			}
		}
```

So the repetitive parts are the `nil` check and the log. Do you want me to move them inside the `connect` and rename them to something like `connectIfDisconnected`?

eth/executionclient/multi_client.go
nkryuchkov: well spotted, thanks. Fixed

nkryuchkov (approved):

nkryuchkov:
Spec Alignment failure is supposed to be fixed by https://github.com/ssvlabs/ssv/pull/2003

beacon/goclient/goclient.go
```diff
@@ -241,55 +203,143 @@ func New(
...
+				defer gc.genesisMu.Unlock()
+
+				gc.log.Fatal("client returned unexpected genesis",
+					zap.String("address", s.Address()),
+					zap.Any("client_genesis", genesis.Data),
+					zap.Any("expected_genesis", gc.genesis),
+					zap.Error(err),
+				)
+				return // Tests may override Fatal's behavior
+			}
```
y0sher, line 294: this might crash the program after a while that it's running if the second client won't be actuveat start but become active later. We should instead just stop using this client.
This is much easier to achieve if we just check the genesis comparing to a local stored value (e.g: saving genesis fork version in networkconfig )

eth/executionclient/multi_client.go
```diff
@@ -63,66 +74,92 @@ func NewMulti(ctx context.Context, nodeAddrs []string, contractAddr ethcommon.Ad
...
+			multiClient.logger.Error("failed to connect to node",
+				zap.String("address", nodeAddrs[clientIndex]),
+				zap.Error(err))
+
+			multiErr = errors.Join(multiErr, err)
+			continue
 		}
 
-		multiClient.clients = append(multiClient.clients, singleClient)
+		multiClient.connectedCount.Add(1)
```
y0sher, line 88: I don't undestand why it needs to be atomic, never used anywhere else, only in close but only store.

y0sher: changed by me

eth/executionclient/multi_client.go
```diff
@@ -285,30 +340,54 @@ func (mc *MultiClient) Close() error {
...
+		return f(mc.clients[0]) // no need for mutex because one client is always non-nil
 	}
 
 	// Iterate over the clients in round-robin fashion,
 	// starting from the most likely healthy client (currentClientIndex).
 	var startingIndex = int(mc.currentClientIndex.Load())
 	var allErrs error
-	for i := 0; i < len(mc.clients); i++ {
+	for i := range mc.clients {
+		mc.clientsMu[i].Lock()
```
iurii-ssv, line 372: hmm, honestly the way we are using `mc.clientsMu` is quite scary (and all over the place)

here for example we are locking `mc.clientsMu[i].Lock()` and then we might lock that same mutex in `mc.connect(ctx, i)` before unlocking it first - by doing this:
```
	mc.clientsMu[clientIndex].Lock()
	mc.clients[clientIndex] = singleClient
	mc.clientsMu[clientIndex].Unlock()
```

unlocking without `defer`(s) is also a foot-gun - maybe wrap the necessary functionality into helper-function so it resolves that ?

**btw, is there any reason** to hold `mc.clientsMu[i]` mutex beyond the point of when client is initialized (`mc.clients` item is updated) ? We are holding it to the very end (calling `client.Healthy` and `v, err := f(client)` and such) - might not be a big deal, but that creates a bigger-than-necessary area to catch a dead-lock

iurii-ssv (commented):
Left some suggestions

beacon/goclient/goclient.go
```diff
@@ -241,55 +203,143 @@ func New(
...
 	)
 	if err != nil {
-		logger.Error("Consensus http client initialization failed",
+		gc.log.Error("Consensus http client initialization failed",
 			zap.String("address", addr),
 			zap.Error(err),
 		)
 
-		return nil, fmt.Errorf("create http client: %w", err)
+		return fmt.Errorf("create http client: %w", err)
```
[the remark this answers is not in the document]
y0sher: The main difference and issue is that when logging with zap and adding fields its easy to search them by the label value. we'll need to squeeze everything to the `fmt.Errorf`, it'll still be searchable, but not by label.

beacon/goclient/goclient.go
nkryuchkov: Clients should never have the same geneses because it would mean they use different ETH networks, so IMO we shouldn't continue in this case. If it were just a log, the multi-client would just silently log it and switch to the next client.

beacon/goclient/goclient.go
```diff
@@ -241,55 +203,143 @@ func New(
...
+				gc.genesisMu.Lock()
+				defer gc.genesisMu.Unlock()
+
+				gc.log.Fatal("client returned unexpected genesis",
+					zap.String("address", s.Address()),
+					zap.Any("client_genesis", genesis.Data),
+					zap.Any("expected_genesis", gc.genesis),
+					zap.Error(err),
+				)
+				return // Tests may override Fatal's behavior
```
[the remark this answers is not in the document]
nkryuchkov: IMO yes, because it means the client is misconfigured (using a different ETH network)

eth/executionclient/multi_client.go
```diff
@@ -63,66 +73,94 @@ func NewMulti(ctx context.Context, nodeAddrs []string, contractAddr ethcommon.Ad
...
+		return nil, fmt.Errorf("no available clients: %w", multiErr)
+	}
+
+	return multiClient, nil
+}
+
+func (mc *MultiClient) connect(ctx context.Context, clientIndex int) error {
+	// The underlying client may call Fatal on unsuccessful reconnection attempt.
+	// Therefore, we need to override Fatal's behavior to avoid crashing.
+	logger := mc.logger.WithOptions(zap.WithFatalHook(zapcore.WriteThenNoop))
```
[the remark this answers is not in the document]
nkryuchkov: I mean the original execution client implementation: https://github.com/ssvlabs/ssv/blob/ff4fa98192d2ba4e8759f8058ffd21f275e83bfb/eth/executionclient/execution_client.go#L266

I'll make the comment clearer

beacon/goclient/goclient.go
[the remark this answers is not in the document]
nkryuchkov: I agree that we could think more about our logging approach, but I think this package is currently using an approach similar to other packages. If we decide to improve logging (e.g. use custom error types with fields), we need to do it project-wide

beacon/goclient/goclient.go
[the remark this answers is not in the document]
iurii-ssv: I guess you are right - failing silently isn't great either ...

I just worry that ETH client might return different/invalid genesis due to something other than our own misconfiguration (bug, some circumstances we didn't cover with tests we've done, etc.) - resulting in complete SSV node downtime

the error message `client returned unexpected genesis` might also be a bit more explanatory, informing operator that he needs to check ETH-client configuration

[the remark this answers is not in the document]
nkryuchkov: > I just worry that ETH client might return different/invalid genesis due to something other than our own misconfiguration (bug, some circumstances we didn't cover with tests we've done, etc.)

Perhaps we could increase our confidence by checking `GenesisForkVersion` only, it should be enough. It's [parsed in client](https://github.com/attestantio/go-eth2-client/blob/v0.21.7/api/v1/genesis.go#L91), so its byte representation should be equal.

> resulting in complete SSV node downtime

Of course, we need to QA it with all clients, but even if something happens (although I don't foresee any cases why it could), then if the first available node is up, it won't cause any downtime because the code saves the genesis of the first available node and asserts against it. So it shouldn't be worse than the single client logic (the current logic on stage)

> the error message client returned unexpected genesis could be a bit more detailed, informing operator that he needs to check ETH-client configuration

It logs `err` which contains details on what field assertion failed: https://github.com/ssvlabs/ssv/blob/e1cc2e3cc9c81946f9c4fb808176bced2e4e74ca/beacon/goclient/goclient.go#L342-L356

eth/executionclient/multi_client.go
```diff
@@ -285,30 +340,54 @@ func (mc *MultiClient) Close() error {
...
 	}
 
 	return multiErr
 }
 
 func (mc *MultiClient) call(ctx context.Context, f func(client SingleClientProvider) (any, error)) (any, error) {
 	if len(mc.clients) == 1 {
-		return f(mc.clients[0])
+		return f(mc.clients[0]) // no need for mutex because one client is always non-nil
 	}
```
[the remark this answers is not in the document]
nkryuchkov: Yes, it works that way. IMO, this check doesn't add much code and it's very easy to read. Thinking about how the code below would handle one client is much harder. Generally, we never use the multi-client with only one client except for tests (I think we need to use the previous implementation with only one client because it's been used for a long time without any issues)

beacon/goclient/goclient.go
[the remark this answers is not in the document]
iurii-ssv: > It logs err which contains details on what field assertion failed:

I mean, from operator standpoint - he won't know what `genesis time mismatch` (or similar) means, ideally he'd need "clear instructions" on how to fix it - which is to hint him to re-check his ETH-client configuration

eth/executionclient/multi_client.go
```diff
@@ -34,22 +35,31 @@ type MultiClient struct {
...
 
 	contractAddress ethcommon.Address
+	chainIDMu       sync.Mutex
 	chainID         *big.Int
 	closed          chan struct{}
 
 	nodeAddrs          []string
-	clients            []SingleClientProvider
+	clientsMu          []sync.Mutex           // each client has own mutex
+	clients            []SingleClientProvider // nil if not connected
```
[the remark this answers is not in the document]
nkryuchkov: > would be nice to clarify that "client itself doesn't need mutex held locked when its methods are called, these mutexs are only necessary to initialize clients in lazy thread-safe manner" right ?

yes, definitely

> it's not obvious/clear to me why we can't just initialize clients slice with respective clients in NewMulti and by-pass that whole thread-safe-lazy-initialization problem altogether (no mutexs needed); would be nice to clarify that in a describing comment

Because if a node is down on node start, we won't be able to create a client for it. The beacon client library supports delayed start, but the execution client library doesn't.

> another option, it seems there isn't really a need to enforce 100% atomicity for clients initialization - meaning, it's fine if we create 2+ clients for the same address and send some requests from not-the-very-same SingleClientProvider interface instance - thus instead of mutexs we could simply use sync.Map (perhaps our own wrapper hashmap)

Initially, I thought about using `atomic.Pointer` but the issue with it (and `sync.Map` as well) is that there are places where we need to lock for a few lines, e.g. when we reconnect to the client: https://github.com/ssvlabs/ssv/blob/e1cc2e3cc9c81946f9c4fb808176bced2e4e74ca/eth/executionclient/multi_client.go#L248-L272

With `atomic/sync.Map`, we'd be able to check for `nil` atomically, but then another call might have reconnected and made `mc.clients[i]` non-nil

> yet another option, or if there is a need for that 100% atomic initialization (so that every request we send is done though the very same SingleClientProvider interface instance) we could use sync.Map with LoadOrStore (but also use Load to smart-check if we want to spawn a new client or not, so we don't spawn clients every time we want to make a request)

But still, it's not 100% atomic because after we use `Load` to smart-check, the value may change before we call `LoadOrStore`. It's a rare case though, but mutex looks more reliable and not harder to read to me

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
iurii-ssv: I think it does handle the case of 1-client correctly (if it doesn't that would indicate there might be some other issue with that code, plus it's something that's easy to test)

I guess a related discussion would be - https://github.com/ssvlabs/ssv/pull/1308#discussion_r1930463059 (where we also "fall back" to older code rather than using the new implementation)

I understand that "fall back to previous code" might be slightly safer to do short-term, but this approach keeps accruing tech-debt over time

eth/executionclient/multi_client.go
nkryuchkov: Well spotted, indeed `connect` locked locked mutex. Fixed

nkryuchkov: > btw, is there any reason to hold mc.clientsMu[i] mutex beyond the point of when client is initialized (mc.clients item is updated) ? We are holding it to the very end (calling client.Healthy and v, err := f(client) and such) - might not be a big deal, but that creates a bigger-than-necessary area to catch a dead-lock

I agree, fixed

y0sher: This comment seems to me most important, if we can simplify the mutex usage pattern across this package it'll be great. Currently it's risky.

nkryuchkov: > unlocking without defer(s) is also a foot-gun - maybe wrap the necessary functionality into helper-function so it resolves that ?

here it's inconvenient because of `allErrs` but I reduced the scope of the mutex

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
iurii-ssv: > we need to lock for a few lines

to lock for a few lines, or alternatively "have everything set up" (created new client) and just swap it out at the very last moment

> But still, it's not 100% atomic because after we use Load to smart-check, the value may change before we call LoadOrStore

if it changed (initialized in parallel) we'll get it loaded and use the loaded value, if not - we'll store our own, but in the end the very same **struct** reference will be used to make all the calls (which is what I mean by 100% atomic in that context)

smart-check is just so we don't need to do "have everything set up" step for every client call/interaction (but only some times)

but anyway, that ^ is not that much better than using mutexs (like we are doing already), it's just still not clear to me why we **have** to have atomicity even with "reconnect" scenarios - just let it create duplicate clients (eventually only 1 of them will be left for future usage(s)), sure it's a bit inefficient but besides minor inefficiency isn't that gonna work ?

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
nkryuchkov: In this case, I think it's not a tech debt because the `ExecutionClient` implementation remains. I don't like using multi-client just for one client because it also changes logging, and adds some overhead. Similarly, we didn't use the beacon client's multi-client implementation just for one client. The first implementation didn't have `MultiClient`, in that case, I would prefer using just one client for everything. But we decided to keep the original version the least modified possible, so in this case, keeping using the old implementation just for one client looks good to me.

Maybe we need to allow `MultiClient` to be used only if 2+ clients are provided though

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
nkryuchkov: > but anyway, that ^ is not that much better than using mutexs (like we are doing already), it's just still not clear to me why we have to have atomicity even with "reconnect" scenarios - just let it create duplicate clients (eventually only 1 of them will be left for future usage(s)), sure it's a bit inefficient but besides minor inefficiency isn't that gonna work (initiating requests to the same ETH node using different client instances) ?

There may be a resource leak if we create duplicate clients, and I don't see that it gives significant benefits. I agree that mutexes are dangerous and must be reviewed very well (and your review proved it) but it seems to me like the only suitable solution. Now we have just 3 places where we use them, so it's not so bad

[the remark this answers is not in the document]
nkryuchkov: I created `getClient` helper so there are two mutex usages now: in `getClient` and in `Close`

eth/executionclient/multi_client.go
```diff
@@ -285,10 +350,17 @@ func (mc *MultiClient) Close() error {
...
 	var multiErr error
-	for i, client := range mc.clients {
-		if err := client.Close(); err != nil {
-			mc.logger.Debug("Failed to close client", zap.String("address", mc.nodeAddrs[i]), zap.Error(err))
-			multiErr = errors.Join(multiErr, err)
+	for i := range mc.clients {
+		mc.clientsMu[i].Lock()
+		client := mc.clients[i]
+		mc.clients[i] = nil
+		mc.clientsMu[i].Unlock()
```
iurii-ssv, line 357: Not sure why we are doing `mc.clients[i] = nil` here - if it's a way to tell GC that he can go ahead and collect this I think it's smart enough to figure this one out on its own

iurii-ssv (commented):
Final comments/clarifications

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
iurii-ssv: > There may be a resource leak if we create duplicate clients

I think unused references (and all they reference) will be garbage-collected eventually, memory leak can happen if we "accidentally" keep referencing those duplicate clients (which there is no need to do)

> I agree that mutexes are dangerous and must be reviewed very well (and your review proved it) but it seems to me like the only suitable solution. Now we have just 3 places where we use them, so it's not so bad

Looks much simpler now 👍 

If we were to re-implement it from scratch - I would say "creating duplicate clients" is a simpler/dumber solution I'd prefer. But since everything is pretty much ready with mutexes it LGTM

eth/executionclient/multi_client.go
```diff
@@ -34,22 +35,31 @@ type MultiClient struct {
 	syncDistanceTolerance       uint64
 
 	contractAddress ethcommon.Address
+	chainIDMu       sync.Mutex
```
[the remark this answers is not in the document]
nkryuchkov: yeah, I did the same for the genesis fork version, so we can do it here too. Fixed

eth/executionclient/multi_client.go
nkryuchkov: Because `nil` means that the client is disconnected in this package semantically. But I don't mind removing it, done

eth/executionclient/multi_client.go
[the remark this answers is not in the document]
nkryuchkov: > I think unused references (and all they reference) will be garbage-collected eventually, memory leak can happen if we "accidentally" keep referencing those duplicate clients (which there is no need to do)

I think, connecting to a client may create quite a lot under the hood like goroutines which might be terminated only on calling the `Close` method for example, so GC would still hold the memory. Perhaps everything would indeed be garbage collected, but still, IMO we would need to run pprof to make sure nothing leaks because otherwise, it would be hard to catch, so I still prefer a non-leak-prone solution

iurii-ssv (approved):

eth/executionclient/execution_client.go
```diff
@@ -426,9 +404,6 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
 		case <-ec.closed:
 			return fromBlock, ErrClosed
 
-		case <-ec.healthyCh:
-			return fromBlock, ErrUnhealthy
```
nkryuchkov, line 430: This is not working as expected, I saw it returning `ErrUnhealthy` when the client was healthy. This channel looks tricky and dangerous to work with (see another comment about closing channel), so I'm removing it
FYI @moshe-blox

eth/executionclient/execution_client.go
```diff
@@ -294,22 +287,7 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 		return nil
 	}
 
-	ec.healthyChMu.Lock()
-	defer ec.healthyChMu.Unlock()
-
-	if err := ec.healthy(ctx); err != nil {
-		close(ec.healthyCh)
```
nkryuchkov, line 301: This triggers closing of closed channel which panics, I'm removing it as it's dangerous to use
FYI @moshe-blox

eth/executionclient/multi_client.go
```diff
@@ -278,36 +339,61 @@ func (mc *MultiClient) Filterer() (*contract.ContractFilterer, error) {
...
 	return multiErr
 }
 
-func (mc *MultiClient) call(ctx context.Context, f func(client SingleClientProvider) (any, error)) (any, error) {
+// call calls f for all clients until it succeeds.
+// If forever is false, it tries all clients only once and if no client is available then it returns an error.
+// If forever is true, it iterates clients forever.
+// It's used in StreamLogs because it's called once per the node lifetime,
+// and it's possible that clients go up and down several times, therefore there's no limit.
+func (mc *MultiClient) call(ctx context.Context, f func(client SingleClientProvider) (any, error), forever bool) (any, error) {
```
iurii-ssv, line 370: Would probably read simpler with a separate method `callRetryingForever`, but maybe that's more of a preference thing.

eth/executionclient/multi_client.go
```diff
@@ -278,36 +339,61 @@ func (mc *MultiClient) Filterer() (*contract.ContractFilterer, error) {
...
 		}
 	}
 
 	return multiErr
 }
 
-func (mc *MultiClient) call(ctx context.Context, f func(client SingleClientProvider) (any, error)) (any, error) {
+// call calls f for all clients until it succeeds.
+// If forever is false, it tries all clients only once and if no client is available then it returns an error.
+// If forever is true, it iterates clients forever.
```
iurii-ssv, line 367: Well, it seems now this is not entirely true for `len(mc.clients) == 1` (which is why I sort of called it a tech debt)

eth/executionclient/multi_client.go
```diff
@@ -278,36 +339,61 @@ func (mc *MultiClient) Filterer() (*contract.ContractFilterer, error) {
...
 	// starting from the most likely healthy client (currentClientIndex).
-	var startingIndex = int(mc.currentClientIndex.Load())
+	startingIndex := int(mc.currentClientIndex.Load())
 	var allErrs error
-	for i := 0; i < len(mc.clients); i++ {
+	limit := len(mc.clients)
+	if forever {
+		limit = math.MaxInt32
+	}
+	for i := 0; i < limit; i++ {
```
iurii-ssv, line 383: It seems when "spinning forever" here - under certain circumstances (when clients error immediately)
- we can potentially exhaust `math.MaxInt32` (so I'd use 64 bit instead)
- we can generate a lot of spamming logs

so we'd probably want to introduce some kind of delay (when "spinning forever") after we've done 1 round over all the clients (after `i` has increased by `len(mc.clients)`)

eth/executionclient/execution_client.go
iurii-ssv: +1

Just briefly looking at this channel - it's not entirely clear how it is supposed to be used (what it's for, etc.)

so if we aren't gonna remove it entirely we need to at least re-think it's purpose/usage

eth/executionclient/multi_client.go
nkryuchkov: Yes, it can potentially exhaust but I thought it's hardly reachable and we need to review the logic if so, so I thought this would be a bit simpler to understand. I agree with spamming logs, but I think we need to switch clients ASAP without any delays.

I rewrote this using another approach

eth/executionclient/multi_client.go
nkryuchkov: Fixed the comment

eth/executionclient/multi_client.go
nkryuchkov: I tried extracting the method internals and have two methods but I don't like how it looks. Also, I don't like copying the method and having to maintain both. I decided to pass `maxTries` instead

eth/executionclient/multi_client.go
iurii-ssv: > I agree with spamming logs, but I think we need to switch clients ASAP without any delays.

I was thinking of something like 100ms-1000ms delay, but maybe for time constraint duty (like block production) that's still an additional burden ...

we can get back to it if it (log-spamming) proves to be an actual issue in prod

eth/executionclient/multi_client.go
iurii-ssv: It kind of beg the question now - why behavior for 1 client is different (my guess would be because it's just "old behavior we want to preserve"), maybe worth leaving a comment about it

iurii-ssv (approved):
LGTM (with minor comments)

eth/executionclient/multi_client.go
```diff
@@ -429,6 +430,7 @@ func (mc *MultiClient) call(ctx context.Context, f func(client SingleClientProvi
 			continue
 		}
 
+		logger.Debug("call succeeded")
```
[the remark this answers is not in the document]
nkryuchkov: yes, it was useful while QA was in progress

PR #2001, opened 2025-01-22
feat(share exit-epoch): Add new field "exit-epoch" to Share data structure and database

oleg-ssvlabs:
This PR adds a new field to the Validator’s Share data structures (both domain and database models): `ExitEpoch`. This field is part of the Validator’s metadata and is retrieved from the beacon API call (`/eth/v1/beacon/states/{state_id}/validators` endpoint). It is required for implementing changes related to Sync Committee duties, which must also be performed for Validators in the Exit state.

Following the Beacon spec (i.e., the value returned from the Beacon API), the `exit_epoch` field equals `2**64 - 1` (max `uint64`) if the validator is in "active" mode or essentially not exiting. During the migration, the `ExitEpoch` field will be set to `0`, meaning the migration itself will not attempt to synchronize it with the correct value from the Beacon state. This is because, after the migration is completed, a separate job will run during the node startup to synchronize all Validator Shares with the Beacon state. At that point, the `ExitEpoch` field in storage will be updated to match the Beacon state.


registry/storage/shares_encoding.go
nkryuchkov, line 1: Do we need any kind of DB migration because the share format is changed?

oleg-ssvlabs: Currently this PR is waiting for #2002 to be merged. We will try to include the new field in `migration 5`. Although, we still need to confirm whether we need any kind of changes in the db. Additional field is a backwards compatible kind of changes, probably depends how serializer behaves and whether we want the existing shares to default to something else than 0. So, at this point I do not have the precise answer

iurii-ssv: > Although, we still need to confirm whether we need any kind of changes in the db.

jfyi, I believe for SSZ serialized it's not backward-compatible (at least based on the test I've written in #2002)

y0sher:
Is this still WIP? I don't see where it changes the way we do sync committee..

operator/validator/controller_test.go
```diff
@@ -651,7 +657,7 @@ func TestGetValidatorStats(t *testing.T) {
 					Committee: operators,
 				},
 				Status:          3, // ValidatorStatePendingInitialized
-				ActivationEpoch: passedEpoch,
+				ActivationEpoch: activationEpoch,
```
y0sher, line 660: why here you didn't add `exitEpoch`?

oleg-ssvlabs:
> Is this still WIP? I don't see where it changes the way we do sync committee..

Yes, it is still in WIP (converted to Draft). Currently testing the change

operator/validator/controller_test.go
oleg-ssvlabs: I would say this is quite a broad question. In this particular scenario, there are many other fields that exist in `SSVShare` types, yet they aren't set. This might mean that other fields are not important for this specific test, though I am not entirely sure if that's the case here. 

That said, I think it's a good pattern to set only the fields relevant to the function under test. If the function interacts with or mutates fields A and B, there’s no need to set other fields during the Arrange step. This approach makes it explicit: fields A and B are set because they directly impact the test outcome.

Hence, I was not confident if I need to modify all the tests here. I will into these

operator/validator/controller_test.go
```diff
@@ -651,7 +657,7 @@ func TestGetValidatorStats(t *testing.T) {
 					Committee: operators,
 				},
 				Status:          3, // ValidatorStatePendingInitialized
```
oleg-ssvlabs, line 659: This comment is misleading. Status 3 is equal to `ValidatorStateActiveOngoing`. Status `ValidatorStatePendingInitialized` fails the test

operator/validator/controller_test.go
```diff
@@ -115,34 +116,36 @@ func TestSetupValidatorsExporter(t *testing.T) {
...
 				ValidatorPubKey: spectypes.ValidatorPK(secretKey.GetPublicKey().Serialize()),
 			},
-			Status:          3, // ValidatorStatePendingInitialized
-			ActivationEpoch: passedEpoch,
-			Liquidated:      false,
+			Status:                    eth2apiv1.ValidatorStateActiveOngoing,
+			ActivationEpoch:           activationEpoch,
+			ExitEpoch:                 exitEpoch,
+			Liquidated:                false,
+			BeaconMetadataLastUpdated: time.Now(),
```
oleg-ssvlabs, line 123: I added this field mainly because these are declared as `sharesWithMetadata`, and when metadata is set/updated the `BeaconMetadataLastUpdated` is set. Although, it does not have any affect on the test itself

registry/storage/shares_test.go
```diff
@@ -164,29 +208,6 @@ func TestSharesStorage(t *testing.T) {
 
 		require.Equal(t, 2, len(existingValidators))
 	})
-
-	require.NoError(t, storage.Shares.Delete(nil, validatorShare.ValidatorPubKey[:]))
```
oleg-ssvlabs, line 168: I removed `Delete` and `Drop` tests because they are already tested in other [places](https://github.com/ssvlabs/ssv/blob/stage/registry/storage/shares_test.go#L192)

registry/storage/shares_test.go
```diff
@@ -599,16 +626,15 @@ func fakeParticipatingShare(index phase0.ValidatorIndex, pk spectypes.ValidatorP
...
+		Status:          v1.ValidatorStateActiveOngoing,
 		ActivationEpoch: 4,
+		ExitEpoch:       math.MaxUint64,
 		OwnerAddress:    common.HexToAddress("0xFeedB14D8b2C76FdF808C29818b06b830E8C2c0e"),
 		Liquidated:      false,
 	}
 }
 
 func generateMaxPossibleShare() (*Share, error) {
-	threshold.Init()
```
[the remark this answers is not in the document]
oleg-ssvlabs: the initialization happens [here](https://github.com/ssvlabs/ssv/blob/e4389ae14254eec1f758045191b3e0058e88588d/registry/storage/shares_test.go#L37), so this felt redundant, also it did not impact the test results. If you think this line of code should be there, I can add it back

iurii-ssv (approved):
LGMT

@oleg @y0sher @moshe one thing worth noting is that IF we are gonna make this change a part of `migration_5` (rather than creating a dedicated `migration_6`) it will break the nodes we've already deployed `migration_5` to (that's mostly **stage** I think, but could be we've deployed it somewhere else eg. to some of our **prod** nodes) - meaning **once this changes is deployed DB needs to be nuked** so it can sync from smart-contracts properly

operator/validator/controller_test.go
```diff
@@ -115,34 +116,36 @@ func TestSetupValidatorsExporter(t *testing.T) {
 				Committee:       operators,
 				ValidatorPubKey: spectypes.ValidatorPK(secretKey.GetPublicKey().Serialize()),
 			},
-			Status:          3, // ValidatorStatePendingInitialized
-			ActivationEpoch: passedEpoch,
-			Liquidated:      false,
+			Status:                    eth2apiv1.ValidatorStateActiveOngoing,
```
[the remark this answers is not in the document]
oleg-ssvlabs: [fixed](https://github.com/ssvlabs/ssv/pull/2001/commits/fba0f3ae95fcac741621ba743520f001c8bfbbed)

migrations/migration_6_model.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Decode decodes Share using ssz.
+func (s *migration_6_OldStorageShare) Decode(data []byte) error {
+	if len(data) > types.MaxAllowedShareSize {
+		return fmt.Errorf("share size is too big, got %v, max allowed %v", len(data), types.MaxAllowedShareSize)
+	}
+	if err := s.UnmarshalSSZ(data); err != nil {
+		return fmt.Errorf("decode Share: %w", err)
+	}
+	s.Quorum, s.PartialQuorum = types.ComputeQuorumAndPartialQuorum(uint64(len(s.Committee)))
```
oleg-ssvlabs, line 51: Due to the bug in SSZ encoding generator, I assume this was the workaround. [Issue](https://github.com/ferranbt/fastssz/issues/188).
Newly generated encoder does not require these fields to be set manually.
You may ask - maybe during the **encoding** and saving into the storage Share does not have these fields set, hence these needs to be set during **Decoding** (fetching from the storage). It seems to be [not the case](https://github.com/ssvlabs/ssv/blob/stage/registry/storage/shares.go#L273)

iurii-ssv: Not sure if this behavior was just as buggy with **gob** - probably not, so it's unlikely to be a work-around but rather just by luck switching to **ssz** didn't result in any bugs,

but also I would think the unit-tests we have in `package storage` should cover things like this ? Maybe we should review those

migrations/migration_6_share_exit_epoch.go
```diff
@@ -0,0 +1,109 @@
...
+			Graffiti:            share.Graffiti,
+			ValidatorIndex:      phase0.ValidatorIndex(share.ValidatorIndex),
+		},
+		OwnerAddress:    share.OwnerAddress,
+		Liquidated:      share.Liquidated,
+		Status:          v1.ValidatorState(share.Status), // nolint: gosec //G115: integer overflow conversion uint64 -> int
+		ActivationEpoch: phase0.Epoch(share.ActivationEpoch),
+	}
+
+	return storage.FromSSVShare(domainShare)
```
iurii-ssv, line 108: Wouldn't it be simpler for `mapShare` to map `migration_6_OldStorageShare ` fields directly into `storage.Share` fields (instead of additionally using `mapShare`) ?

Otherwise we'd probably want to clarify (leaving a comment) **why** we are doing it like that

migrations/migration_6_share_exit_epoch.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/ssvlabs/ssv/registry/storage"
+	"github.com/ssvlabs/ssv/storage/basedb"
+	"go.uber.org/zap"
+)
+
+const migration_6_Name = "migration_6_share_exit_epoch"
+
+var migration_6_share_exit_epoch = Migration{
+	Name: migration_6_Name,
+	Run: func(ctx context.Context, l *zap.Logger, opt Options, key []byte, completed CompletedFunc) error {
```
iurii-ssv, line 21: Here is a thought, because BadgerDB doesn't support (as far as I know) general-purpose **atomic** transactions migrations like this one are non-atomic - meaning it is possible this migration will update some DB state but not complete fully

thus I think it's better to design migration(s) to be idempotent when possible (such that we can repeat it multiple times regardless of intermittent DB failures)

for **migration_6** that would mean:
- we'd want to create "new" Share(s) under **separate** prefix (instead of overriding existing shares - since once you've overridden a Share you won't be able to decode it with old SSZ decoder)
- and we delete "old" Share(s) (stored under old prefix) once ALL new shares have been successfully written to DB
- and then we **complete** migration

note, we still might end up in a state where we've deleted old shares but crashed before completing **migration_6** - it still should work fine (idempotent) on the next run

wdyt ?

iurii-ssv (commented):
Minor suggestions

migrations/migration_6_share_exit_epoch.go
oleg-ssvlabs: It's a good point. I actually had a comment for that, which I later removed probably by mistake.
I will add it back

oleg-ssvlabs: Added

migrations/migration_6_share_exit_epoch.go
oleg-ssvlabs: I agree with your point.  
I would say it is certainly concerning.  

BadgerDB's `SetMany` method isn't atomic, as it iterates over all items in the array and updates the database for each item, with each update potentially failing.  
Another operation is the database migration completion function, which updates the migration state in the database by setting its status to "completed" so it does not run again - this update may also fail. That would mean all Shares were migrated, but on the next node startup, the migration would attempt to run again, leading to failure because the old encoder wouldn't be able to decode the migrated Shares.  

I like the idea of using prefixes since it seems we cannot make the entire migration atomic anyway. I just wonder which prefixes to use - perhaps introducing some naming convention?

migrations/migration_6_share_exit_epoch.go
iurii-ssv: > We cannot directly map migration_6_OldStorageShare to storage.Share because storage.Share
> contains unexported types ('storageOperator'). Instead, we map to SSVShare first
> and use storage.FromSSVShare() which has access to set those unexported types.
		
Hmm, wouldn't it be simpler to just export all the fields ? Maybe it's not ideal - but results into simpler solution overall

although either approach is so-so ...

oleg-ssvlabs: > Hmm, wouldn't it be simpler to just export all the fields ?

I was thinking the same way, although I was wondering if there is a specific reason why that type is not exported.  
One reason could be to prohibit the use of struct literals and enforce instantiation of the `Share` through a constructor. Since a constructor doesn't seem to exist, maybe it should always be instantiated from the domain model using `storage.FromSSVShare`. So I went that route in this implementation

migrations/migration_6_share_exit_epoch.go
oleg-ssvlabs: Implemented a new Share storage prefix in [this](https://github.com/ssvlabs/ssv/pull/2001/commits/51be6e70d350950b3278f63ec119225e8944c565) commit.  

One thing that was questionable in the migration implementation: what happens if the Share migration succeeds, the old prefix is dropped, but the completion function (which marks the migration as "done") fails? This would cause the migration to run again on the next node startup.  

I had these questions:
- Would it work to iterate (GET) over a prefix that does not exist?
- Would it work to drop a prefix that does not exist?

I implemented a [test](https://github.com/ssvlabs/ssv/pull/2001/commits/51be6e70d350950b3278f63ec119225e8944c565#diff-22c5aababb6acc1e117caab1075dac572ae285c39d559628017680da2ac0d865R66) for that, and it does seem to work (does not return an error).

iurii-ssv: As far as I know prefix(es) are just a handy tool to access BadgerDB (which is mostly a key-value store)

migrations/migration_6_share_exit_epoch.go
iurii-ssv: I don't really have a strong opinion here, but just having some fields/structs package-private and not others - eg. `Share struct` itself - might be a bit confusing (hence I thought making it all public would make more sense)

iurii-ssv (approved):
LGTM

migrations/migration_6_share_exit_epoch.go
moshe-blox: > Here is a thought, because BadgerDB doesn't support (as far as I know) general-purpose atomic transactions migrations like this one are non-atomic - meaning it is possible this migration will update some DB state but not complete fully
@iurii-ssv 

AFAIK Badger supports ACID and it can be made atomic by creating a transaction

moshe-blox: @oleg-ssvlabs we've also added SetMany to our transaction abstraction (ReadWriter interface) so it can be atomic

moshe-blox: idk whats the performance profile of a transaction with 60k~ writes, so we should probably test it on a non-operator mainnet node to confirm its ok

iurii-ssv: I think the issue with Badger transactions is that they are limited in size (from what I've read about it - they don't use write-ahead log a typical/traditional DB would use, and hence they have to apply all the changed directly onto their structures in memory first) - https://github.com/hypermodeinc/badger/issues/1325

or is that outdated ?

moshe-blox: oh right, so we'll have to call Set instead of SetMany

this shouldn't be a problem to store even 100k share writes in memory (lets say even if each write weighs 1kb thats only 100mb and this all runs before the node so lots of free memory)

but of course we have to test to know

oleg-ssvlabs: following the discussion, @moshe-blox, do you think I need to make changes in the migration logic?

eth/eventhandler/event_handler_test.go
```diff
@@ -748,6 +749,7 @@ func TestHandleBlockEventsStream(t *testing.T) {
 			require.NotNil(t, valShare)
 			valShare.ValidatorIndex = 1
 			valShare.ActivationEpoch = 0
+			valShare.ExitEpoch = math.MaxUint64
```
olegshmuelov, line 752: Instead of using `ExitEpoch = math.MaxUint64` throughout the code, we can use `goclient.FarFutureEpoch`, which is already defined as `phase0.Epoch = math.MaxUint64`.
This makes the code more readable and aligns with existing constants.

oleg-ssvlabs: Good point. I replaced all usages of `math.MaxUint64` with `goclient.FarFutureEpoch` in all unit tests.

migrations/migration_6_share_exit_epoch.go
y0sher: @iurii-ssv @oleg-ssvlabs maybe we could use [WriteBatch](https://github.com/hypermodeinc/badger/blob/ef28ef36b5923f12ffe3a1702bdfa6b479db6637/batch.go#L37-L42) like they recommend?

registry/storage/shares.go
```diff
@@ -21,7 +21,7 @@ import (
 // sharesPrefix specifies the prefix used for storing Share(s) in DB.
 // Note, previously gob-encoded Share(s) were stored with `shares/` prefix, this has been
 // changed in migration_5_change_share_format_from_gob_to_ssz.
-var sharesPrefix = []byte("shares_ssz/")
+var sharesPrefix = []byte("shares_ssz_v2/")
```
moshe-blox, line 24: we should probably go towards a standard for collection names, something like `<collection>-<version>` should be enough

so in this case just `shares-2` should work (assuming it was just `shares` before the SSZ upgrade)

registry/storage/shares.go
```diff
@@ -113,7 +118,6 @@ func (s *Share) Decode(data []byte) error {
 	if err := s.UnmarshalSSZ(data); err != nil {
 		return fmt.Errorf("decode Share: %w", err)
 	}
-	s.Quorum, s.PartialQuorum = types.ComputeQuorumAndPartialQuorum(uint64(len(s.Committee)))
```
moshe-blox, line 116: whats the reason for removing this?

migrations/migration_6_share_exit_epoch.go
oleg-ssvlabs: @y0sher `WriteBatch` is already used by [`SetMany`](https://github.com/ssvlabs/ssv/blob/main/storage/kv/badger.go#L123).  
Even though batch writes using the `Set()` method may not look atomic, I think it does **try** to be atomic. From what I understand, the transaction(s) are committed when the `Flush` method is called — meaning `Set()` does not immediately commit.  
It seems atomicity holds as long as all writes fit into a single transaction. According to the discussion [here](https://github.com/hypermodeinc/badger/issues/1325), the maximum transaction size is 15% of the table size.  
So if we're trying to modify all records at once, it won't be atomic — Badger will split the writes into roughly 100 / 15 ≈ 6 transactions.

iurii-ssv: I think, like Oleg describes, we can't really use `WriteBatch` or `SetMany` to get atomicity for this scenario - and that why I think making migration idempotent is our best option for cases like this ...

registry/storage/shares.go
oleg-ssvlabs: It is related to [this](https://github.com/ferranbt/fastssz/issues/188) issue with fast SSZ.  
I believe this line was originally added as an implicit fix due to the encoder not being able to handle these fields -hence they were recalculated manually.  
[This line](https://github.com/ssvlabs/ssv/pull/2001/files#diff-0a812846b4ce22ff86e4618d79215d56303d5b88889db6f7478811bda11ab357R41) confirms that these fields were not part of the encoder. After I changed [this](https://github.com/ssvlabs/ssv/pull/2001/files#diff-5a2be62f4b213364f822f71cf79a6405eb2bb1dd344687bc26466ea33f31b554L83) and regenerated the encoder, the previously missing fields were correctly included. It's a weird problem. The GitHub issue I created (mentioned above) has minimal reproducible example. 

Therefore, I believe this line has become redundant.

migrations/migration_6_share_exit_epoch.go
y0sher: It seems like this is a specific issue for shares migration since this is our biggest table. So we might want to use some workaround to actually make this "atomic" maybe by protecting the migration with some in db lock, and making the last "Set" atomic with removing this lock.

registry/storage/shares.go
y0sher: I think `shares_v2` is good enough.

migrations/migration_6_share_exit_epoch.go
oleg-ssvlabs: @y0sher With the new shares table prefix, I don’t think we really need to aim for atomicity. In case of a mid-migration failure, the node will crash. Upon restart, it will attempt to re-run the migration, and with the current setup, re-running a partially completed migration shouldn’t be an issue. So I don’t really see much value in trying to migrate all shares atomically

registry/storage/shares.go
oleg-ssvlabs: renamed to `shares_v2`. `shares_v3` would work as well

iurii-ssv (approved):

moshe-blox (approved):

moshe-blox (approved):
lgtm :fire:

y0sher (approved):

PR #2002, opened 2025-01-22
migrations: finalize migration_5 (share encoding GOB -> SSZ)

iurii-ssv:
This PR finalizes the work started in https://github.com/ssvlabs/ssv/pull/1837 essentially "completing" **migration_5**

it doesn't change **migration_5** in a sense that it is still equivalent and backward-compatible with it's previous version, rather it just adds some sanity-checks.


migrations/migration_5_share_gob_to_ssz.go
```diff
@@ -47,23 +70,129 @@ var migration_5_change_share_format_from_gob_to_ssz = Migration{
...
+				// from completing, so we don't return error here
+				return nil
+			}
+			if !matchGOBvsSSZ(shareGOB, shareSSZ) {
+				return fmt.Errorf(
+					"GOB share doesn't match corresponding SSZ share, GOB: %s, SSZ: %s",
+					litter.Sdump(shareGOB),
+					litter.Sdump(shareSSZ),
+				)
+			}
```
iurii-ssv, line 106: as per @moshe-blox recommendation I found `litter` to be super useful for debugging, what do you think of using it like that (for printing detailed error message) ? 

Alternatively we could use `logger` to print the contents of `shareGOB` and `shareSSZ`, but since we also have to return `error` that means we essentially have to duplicate error message (and it will be logged twice)

migrations/migration_5_share_gob_to_ssz.go
```diff
@@ -47,23 +70,129 @@ var migration_5_change_share_format_from_gob_to_ssz = Migration{
...
+			sID := shareID(shareSSZ.ValidatorPubKey)
+			shareGOB, ok := sharesGOB[sID]
+			if !ok {
+				// this shouldn't really happen & we should probably return error if it does, but
+				// on stage since we already have some SSV nodes that migrated to SSZ format and
+				// potentially added new validators (new SSZ shares) erroring would prevent migration
+				// from completing, so we don't return error here
+				return nil
+			}
```
iurii-ssv, line 99: I guess this is an unfortunate side-effect of deploying **migration_5** code to **stage** and not completing it later (in this PR)

alternatively, we could "fix that" by merging this PR -> deploying this to all stage nodes -> creating PR that removes this constraint -> merging that new PR & deploying that to prod eventually (but that would mean we will never be able to test this such a change on **stage**)

registry/storage/shares.go
```diff
@@ -69,10 +69,13 @@ type sharesStorage struct {
...
-// in terms of what types it supports, hence we define a bunch of own types here to satisfy it,
-// see more on this here: https://github.com/ferranbt/fastssz/issues/179#issuecomment-2454371820
-type storageShare struct {
+// Share represents a Share stored in DB. It's either our own share (share belongs to our
+// operator) in which case it has non-empty SharePubKey, or a "generic" representation of some
+// validator (who's shares are managed by other operators).
+// Note, SSZ encodings generator has some limitations in terms of what types it supports, hence
+// we define a bunch of own types here to satisfy it, see more on this here:
+// https://github.com/ferranbt/fastssz/issues/179#issuecomment-2454371820
+type Share struct {
```
oleg-ssvlabs, line 78: Awesome!
It's great you renamed it. `storage.Share` implies it's a storage entity. Would be good if linter could complain about such things :)

registry/storage/shares.go
```diff
@@ -249,17 +252,17 @@ func (s *sharesStorage) Save(rw basedb.ReadWriter, shares ...*types.SSVShare) er
...
 		if err != nil {
 			return basedb.Obj{}, fmt.Errorf("failed to serialize share: %w", err)
 		}
-		return basedb.Obj{Key: s.storageKey(share.ValidatorPubKey[:]), Value: value}, nil
+		return basedb.Obj{Key: SharesDBKey(share.ValidatorPubKey[:]), Value: value}, nil
 	})
 }
 
-func specShareToStorageShare(share *types.SSVShare) *storageShare {
+func FromSpecShare(share *types.SSVShare) *Share {
```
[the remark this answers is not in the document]
oleg-ssvlabs: I agree. Spec share is just type embedded into the domain Share

migrations/migration_5_share_gob_to_ssz.go
```diff
@@ -47,23 +70,129 @@ var migration_5_change_share_format_from_gob_to_ssz = Migration{
 			return fmt.Errorf("GetAll: %w", err)
 		}
 
-		if err := opt.Db.SetMany(storagePrefix, len(sets), func(i int) (basedb.Obj, error) {
-			return sets[i], nil
+		sharesGOBTotal = len(sharesGOB)
+		if sharesGOBTotal == 0 {
+			return nil // we won't be creating any SSZ shares
```
oleg-ssvlabs, line 75: I would probably write a log message here, so it's not to silent about it

migrations/migration_5_share_gob_to_ssz.go
```diff
@@ -47,23 +70,129 @@ var migration_5_change_share_format_from_gob_to_ssz = Migration{
...
-		//if err := opt.Db.DropPrefix(append(storagePrefix, sharesPrefixGOB...)); err != nil {
-		//	return fmt.Errorf("DropPrefix: %w", err)
-		//}
-		//
-		//// This makes sure migration applies only once.
-		//if err := completed(opt.Db); err != nil {
-		//	return fmt.Errorf("complete transaction: %w", err)
-		//}
+		sharesSSZTotal := 0
+		if err := opt.Db.GetAll(storage.SharesDBPrefix(opstorage.OperatorStoragePrefix), func(i int, obj basedb.Obj) error {
```
oleg-ssvlabs, line 85: From what I understood, this step if a verification check to confirm the migration was successful and all Gob shares became SSZ. Do I understand it right?

migrations/migration_5_share_gob_to_ssz.go
oleg-ssvlabs: yeah, I think it is better to return an error here (in the next PR as you suggest). Temporarily I would also add some log message, so on stage we could see when this scenario happening

migrations/migration_5_share_gob_to_ssz.go
oleg-ssvlabs: It seems to have pretty good [score](https://snyk.io/advisor/golang/github.com/sanity-io/litter), although, would it be enough to just use the logger? I assume the challenge is that zap logger does not log pointer objects with its full properties if to use something like `zap.Any(shareGOB)` ?

migrations/migration_5_share_gob_to_ssz.go
iurii-ssv: Yes, correct, it's just sanity-checks to make sure migration did exactly what we want it to do

registry/storage/shares.go
iurii-ssv: Yeah, previously I think we couldn't rename it because GOB-encoding also somehow incorporated **struct** name - but now we certainly can.

migrations/migration_5_share_gob_to_ssz.go
iurii-ssv: > Temporarily I would also add some log message, so on stage we could see when this scenario happening

I don't think there is a need for log message here since it doesn't really tell us anything new (I confirmed already that returning error on **stage** results in exactly what my comment describes) - so logging error here is simply non-actionable, right (hence why even bother logging) ?

migrations/migration_5_share_gob_to_ssz.go
iurii-ssv: This will be logged in final `migration completed` statement as `gob_shares_total` == `0`, so there isn't really much usefulness to logging it here too,

but in general I agree that errors (or edge-cases) shouldn't be silent.

migrations/migration_5_share_gob_to_ssz_test.go
```diff
@@ -0,0 +1,83 @@
...
+		shareGOBHex, err := hex.DecodeString(shareGOBHexString)
+		assert.NoError(t, err)
+		shareGOBGot := &storageShareGOB{}
+		err = shareGOBGot.Decode(shareGOBHex)
+		assert.NoError(t, err)
```
oleg-ssvlabs, line 20: do we have any conventions regarding when to use `assert` versus `require` ? :) they are quite similar in its behavior. Frankly, I used to use `require` (considering AAA) in Arrange step and `assert` in Assert step :) From what I saw in the tests we mainly use only `require` everywhere. It does not really matter much, although maybe at some point we could align on that

migrations/migration_5_share_gob_to_ssz.go
iurii-ssv: > I assume the challenge is that zap logger does not log pointer objects with its full properties if to use something like zap.Any(shareGOB) ?

I think `zap` logger should produce full object printout with `zap.Any` (although it will probably struggle if struct contains **interfaces** - https://github.com/ssvlabs/ssv/pull/1981),

the main reason I don't want to use logger here is that we'll essentially be logging this `GOB share doesn't match corresponding SSZ share, GOB: ...` error twice:
- here like you suggest
- and the caller will also log error we return here (and we have to return error to signal migration failed)

operator/storage/storage.go
```diff
@@ -18,7 +18,7 @@ import (
 var HashedPrivateKey = "hashed-private-key"
```
oleg-ssvlabs, line 18: could be constant

oleg-ssvlabs: hmm, it cannot be constant some methods mutate it. Probably not the best thing to mutate package level variables during the runtime. 
Anyway, unrelated to this PR

migrations/migration_5_share_gob_to_ssz_test.go
iurii-ssv: I don't think there is a strict set of guidelines we use, 

but I myself try to use `require` whenever I can just to "fail fast" - since it reduces noise in output and in 99.9% cases the error you are looking for will be the first one

for all other cases where seeing just the very first error isn't enough I would use `assert`,

here I just mistakenly used `assert` - fixed in https://github.com/ssvlabs/ssv/pull/2002/commits/ba4b807e8ac76e3bea3d3dfcabc8f2b6f99ebb6f now

migrations/migration_5_share_gob_to_ssz.go
oleg-ssvlabs: I see. One question:
For the migrated nodes on Stage, it looks like it wouldn't run into this scenario anyway, because migration 5 was already applied and it seems like if migration was already applied, it [won't be applied again](https://github.com/ssvlabs/ssv/blob/a8258fa62ee8b5f374cf04c2bd77fd7e816cd836/migrations/migrations.go#L82).

migrations/migration_5_share_gob_to_ssz_test.go
oleg-ssvlabs: thank you for explaining. No, it's good, I do not have any preferences nor think it has any importance of making it conventional, especially considering they behave differently. So probably, depending on the case we can use either `require` or `assert`

migrations/migration_5_share_gob_to_ssz.go
iurii-ssv: > For the migrated nodes on Stage, it looks like it wouldn't run into this scenario anyway, because migration 5 was already applied and it seems like if migration was already applied, it [won't be applied again](https://github.com/ssvlabs/ssv/blob/a8258fa62ee8b5f374cf04c2bd77fd7e816cd836/migrations/migrations.go#L82).

your reasoning is correct except that **migration_5** haven't been **completed** for some nodes on stage even though it has been **applied** (because we haven't marked it as **completed** until this PR is merged - https://github.com/ssvlabs/ssv/blob/3e76188c8954fe690b675ee227fd1a507e428656/migrations/migration_5_share_gob_to_ssz.go#L56-L65),

normally you would mark transaction as **completed** in the same PR where it was originally created (not in a separate PR like I'm doing here),

but for this particular **migration_5** we decided to do it this way so we can merge that original PR (it also had other features in it we wanted to have on **stage**) without blocking QA team from testing it for example,

but as I outline in that comment, that results into more "unfortunate inconveniences we have to deal with"

oleg-ssvlabs: understand, thank you for explaining!

registry/storage/shares.go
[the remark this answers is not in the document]
y0sher: I think it makes sense to use `SSVShare` since that's the current name of the struct. It's not the prettiest but it's better than having two names.

y0sher (approved):

PR #2004, opened 2025-01-24
fix(flaky test) - fix inconsistent results in unit test

oleg-ssvlabs:
Given unit test was sometimes failing with a message like:

`Error:      	"49.996792ms" is not greater than "50ms"`

The fix does not make the test entirely consistent, instead, it adds a 10% margin. This means the test will pass if the sync committee duty execution time is higher than `45ms` (previously `50ms`). (Slot duration is mocked to be [150ms](https://github.com/ssvlabs/ssv/blob/main/operator/duties/scheduler_test.go#L116) instead of the default 12 seconds).

**PS:** I am not sure what exactly this test verifies, but it seems the sync committee duty execution time should be **higher than 50ms**. I ran the test over 1000 times with this change, and there were no failures anymore.


iurii-ssv (approved):
This does result in lots of pipeline failures, it would be nice to get to the root cause ofc - but a temporary fix is also something we can use 👍

nkryuchkov (approved):

PR #2010, opened 2025-02-03
eth/executionclient: rate limit multi client Healthy call

nkryuchkov:
No limit overloads the node

### DIFF
--- a/eth/executionclient/multi_client.go
+++ b/eth/executionclient/multi_client.go
@@ -279,7 +279,7 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan 
 
 // Healthy returns if execution client is currently healthy: responds to requests and not in the syncing state.
 func (mc *MultiClient) Healthy(ctx context.Context) error {
-	if time.Since(time.Unix(mc.lastHealthy.Load(), 0)) <= mc.healthInvalidationInterval {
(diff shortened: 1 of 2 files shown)


eth/executionclient/multi_client.go
```diff
@@ -71,6 +71,7 @@ type MultiClient struct {
 	clientsMu          []sync.Mutex           // clientsMu allow for lazy initialization of each client in `clients` slice in thread-safe manner (atomically)
 	clients            []SingleClientProvider // nil if not connected
 	currentClientIndex atomic.Int64
+	lastHealthy        atomic.Int64
```
iurii-ssv, line 74: maybe `lastHealthy` -> `lastHealthyTimestamp` (or `lastHealthyTsUnix`)

iurii-ssv (approved):
LGTM

y0sher (approved):

eth/executionclient/multi_client.go
nkryuchkov: IMO both are fine,`Timestamp` and `TsUnix` indicate what kind of data is stored, but when time is stored as an int, it's common that it's a timestamp

iurii-ssv: I mean, it could mean **slot number** for all I know (without looking up how it's used)

PR #2018, opened 2025-02-04
feat(attestation data) - Weighted attestation data

oleg-ssvlabs:


beacon/goclient/attest.go
```diff
@@ -101,9 +106,302 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+	gc.log.
+		With(
+			zap.Duration("elapsed", time.Since(started)),
+			zap.Int("responded", responded),
+			zap.Int("errored", errored),
+			zap.Int("timed_out", timedOut),
+		).Debug("results")
+
+	if bestAttestationData == nil {
+		return nil, errors.New("no attestations received")
```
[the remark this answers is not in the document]
oleg-ssvlabs: We use `errors.New()` a lot in the codebase (`456 results`). 
I believe `fmt.Errorf` makes sense to use when there is a need to instantiate formatted error, not just a plain string. Or at least that's how I always used it :)
I do not think we are consisted in that matter

beacon/goclient/attest.go
```diff
@@ -101,9 +106,302 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
+	client Client,
+	respCh chan *attestationDataResponse,
+	errCh chan *attestationDataError,
+	opts *api.AttestationDataOpts,
+) {
+	addr := client.Address()
+	attDataReqStart := time.Now()
+
+	logger := gc.log.With(zap.String("client_addr", addr), fields.Slot(opts.Slot), fields.CommitteeIndex(opts.CommitteeIndex))
```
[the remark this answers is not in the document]
oleg-ssvlabs: fixed

beacon/goclient/attest.go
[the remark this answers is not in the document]
oleg-ssvlabs: changed to `fmt.Errorf`

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		bestClient          string
+	)
+
+	// Loop 1: prior to soft timeout.
+	for responded+errored+timedOut+softTimedOut != requests {
+		select {
+		case resp := <-respCh:
+			responded++
+			gc.log.
+				With(
```
iurii-ssv, line 145: probably better to have `gc.log.With(` on 1 line (not 2) ? It will also improve formatting of the fields supplied to `With`

I guess applies to multiple places in this PR

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+						zap.Int("responded", responded),
+						zap.Int("errored", errored),
+						zap.Int("timed_out", timedOut),
+					).Debug("soft timeout reached with responses")
+			} else {
+				gc.log.
+					With(
+						zap.Duration("elapsed", time.Since(started)),
+						zap.Int("errored", errored),
+					).Error("soft timeout reached with no responses")
```
iurii-ssv, line 185: `soft timeout reached with no responses` -> `oft timeout reached with no successful responses`

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+			}
+			// Set the number of requests that have soft timed out.
+			softTimedOut = requests - responded - errored - timedOut
```
iurii-ssv, line 188: I'm confused on how this `softTimedOut` is calculated ... why do we subtract out `timedOut` value ?

I would think `softTimedOut` is == `timedOut`

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
 	return data, spec.DataVersionPhase0, nil
 }
 
+func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.AttestationData, error) {
+	// We have two timeouts: a soft timeout and a hard timeout.
+	// At the soft timeout, we return if we have any responses so far.
+	// At the hard timeout, we return unconditionally.
+	// The soft timeout is half the duration of the hard timeout.
+	ctx, cancel := context.WithTimeout(gc.ctx, gc.commonTimeout)
+	softCtx, softCancel := context.WithTimeout(ctx, gc.commonTimeout/2)
```
iurii-ssv, line 117: I think it's sort of customary to call `defer cancel()` just in case so we don't leak resources

it's not an issue here but if in the future we add some return statement before we actually cancel context ... that could grow into a leak I think

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+			// Anyone not responded by now is considered errored.
+			timedOut = requests - responded - errored
+			gc.log.
+				With(
+					zap.Duration("elapsed", time.Since(started)),
+					zap.Int("responded", responded),
+					zap.Int("errored", errored),
+					zap.Int("timed_out", timedOut),
+				).Error("hard timeout reached")
+		}
```
iurii-ssv, line 232: I guess it'd be better to have similar handling for hard timeout(s) as we have for soft timeouts, that is differentiate complete failure logging it like 
```
				gc.log.
					With(
						zap.Duration("elapsed", time.Since(started)),
						zap.Int("errored", errored),
					).Error("hard timeout reached with no responses")
```

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+	logger := gc.log.With(
+		zap.Duration("elapsed", time.Since(started)),
+		zap.Int("responded", responded),
+		zap.Int("errored", errored),
+		zap.Int("timed_out", timedOut),
+		zap.Bool("with_weighted_attestation_data", true),
+	)
+	if bestAttestationData == nil {
+		logger.Error("no attestations received")
+		return nil, fmt.Errorf("no attestations received")
```
iurii-ssv, line 245: I think there is no need to log this error here (we are duplicating error essentially), just let the caller handle it - it will probably log it on some upper layer if it is important

you might say we are loosing the context of `log. With` ... but I would suggest to add those details (if they are important) to the error message with `fmt.Errorf` instead

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		case resp := <-respCh:
+			responded++
+			gc.log.
+				With(
+					zap.Duration("elapsed", time.Since(started)),
+					zap.String("client_addr", resp.clientAddr),
+					zap.Int("responded", responded),
+					zap.Int("errored", errored),
+					zap.Int("timed_out", timedOut),
+				).Debug("response received")
```
iurii-ssv, line 205: I suspect even under `Debug` this might generate too much logs,

I would comment this line out (so we can uncomment it back for debugging purposes) unless we absolutely need it - we will log the final result at the end of this `func` eventually (probably not need more details than that)

I guess logging errors is the only thing we need

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+	if attestationData == nil {
+		logger.Error(clNilResponseDataErrMsg)
+		errCh <- &attestationDataError{
+			clientAddr: addr,
+			err:        fmt.Errorf("attestation data nil"),
+		}
+		return
+	}
+
+	logger.Debug("scoring attestation data")
```
iurii-ssv, line 338: This log line is probably just gonna spam

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+	score := gc.scoreAttestationData(ctx, addr, attestationData)
```
iurii-ssv, line 339: could inline `gc.scoreAttestationData(ctx, addr, attestationData)`

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+}
+
+func (gc *GoClient) blockRootToSlot(ctx context.Context, root phase0.Root) (phase0.Slot, error) {
+	logger := gc.log.With(fields.BlockRoot(root))
+	cacheResult := gc.blockRootToSlotCache.Get(root)
+	if cacheResult != nil {
+		cachedSlot := cacheResult.Value()
+		logger.
+			With(fields.Slot(cachedSlot), zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
+			Debug("obtained slot from cache")
```
iurii-ssv, line 388: this log line also spammy

beacon/goclient/goclient.go
```diff
@@ -126,6 +127,7 @@ type GoClient struct {
 	registrationLastSlot phase0.Slot
 	registrationCache    map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration
 
+	blockRootToSlotCache *ttlcache.Cache[phase0.Root, phase0.Slot]
```
iurii-ssv, line 130: I guess the way we are using `blockRootToSlotCache` in this PR is non-atomic (specifically in `blockRootToSlot` func), meaning it's possible to issue multiple requests so callers might even receive different results (results from their own requests might be different from the result that got written to cache last)

see this below for how `singleflight` package can be used to resolve that:
```
	// attestationReqInflight helps prevent duplicate attestation data requests
	// from running in parallel.
	attestationReqInflight singleflight.Group[phase0.Slot, *phase0.AttestationData]
	// attestationDataCache helps reuse recently fetched attestation data.
	// AttestationData is cached by slot only, because Beacon nodes should return the same
	// data regardless of the requested committeeIndex.
	attestationDataCache *ttlcache.Cache[phase0.Slot, *phase0.AttestationData]
```

beacon/goclient/goclient.go
```diff
@@ -163,14 +167,20 @@ func New(
...
 			// we only fetch attestation data during the slot of the relevant duty (and never later),
 			// hence caching it for 2 slots is sufficient
 			ttlcache.WithTTL[phase0.Slot, *phase0.AttestationData](2 * opt.Network.SlotDurationSec()),
 		),
-		commonTimeout: commonTimeout,
-		longTimeout:   longTimeout,
+		// `blockRootToSlotCache` is used for attestation data scoring. When multiple Consensus clients are used,
+		// the cache helps reduce the number of Consensus Client calls by `n-1`, where `n` is the number of Consensus clients
+		// that successfully fetched attestation data and proceeded to the scoring phase. Capacity is rather an arbitrary number,
+		// intended for cases where some objects within the application may need to fetch attestation data for more than one slot.
```
iurii-ssv, line 179: I think it makes more sense to have this comment where `blockRootToSlotCache` is **defined** (not where it's **initialized**) to match how it's done for pretty much every other field on this struct (or any other struct in SSV for that matter)

I guess the part about "value 5 being arbitrary" belongs here, but the rest we should move I think

btw, you probably want to make comment line(s) shorter so it fits on a screen without having to scroll sideways, there isn't exact rule to follow but I'd suggest to limit line lengths (both for comments and code) to ~100 symbols when possible

logging/fields/fields.go
```diff
@@ -400,6 +401,10 @@ func CommitteeID(val spectypes.CommitteeID) zap.Field {
 	return zap.String(FieldCommitteeID, hex.EncodeToString(val[:]))
 }
 
+func CommitteeIndex(val phase0.CommitteeIndex) zap.Field {
+	return zap.Uint64(FieldCommitteeIndex, uint64(val))
+}
```
iurii-ssv, line 406: This added code doesn't seem to be used ?

And by the way you placed it near `CommitteeID` func above - but the term "committee" here has different meaning (it's Ethereum committee which has nothing to do with our SSV committee) - you probably want to call it `EthereumCommitteeIndex` if you intend to use it

protocol/v2/blockchain/beacon/client.go
```diff
@@ -61,10 +61,11 @@ type BeaconNode interface {
...
-	Network               Network
-	BeaconNodeAddr        string `yaml:"BeaconNodeAddr" env:"BEACON_NODE_ADDR" env-required:"true" env-description:"Beacon node address. Supports multiple semicolon separated addresses. ex: http://localhost:5052;http://localhost:5053"`
-	SyncDistanceTolerance uint64 `yaml:"SyncDistanceTolerance" env:"BEACON_SYNC_DISTANCE_TOLERANCE" env-default:"4" env-description:"The number of out-of-sync slots we can tolerate"`
+	Context                     context.Context
+	Network                     Network
+	BeaconNodeAddr              string `yaml:"BeaconNodeAddr" env:"BEACON_NODE_ADDR" env-required:"true" env-description:"Beacon node address. Supports multiple semicolon separated addresses. ex: http://localhost:5052;http://localhost:5053"`
+	SyncDistanceTolerance       uint64 `yaml:"SyncDistanceTolerance" env:"BEACON_SYNC_DISTANCE_TOLERANCE" env-default:"4" env-description:"The number of out-of-sync slots we can tolerate"`
+	WithWeightedAttestationData bool   `yaml:"WithWeightedAttestationData" env:"WITH_WEIGHTED_ATTESTATION_DATA" env-default:"false" env-description:"Enables Attestation Data scoring from multiple Beacon nodes"`
```
iurii-ssv, line 68: `Enables Attestation Data scoring from multiple Beacon nodes` -> `Enables Attestation Data fetching & scoring using multiple Beacon nodes simultaneously (as opposed to fetching Attestation Data from just one Beacon node)`

iurii-ssv (commented):
Bunch of suggestions, mostly minor

appreciate all the details/comments you've wrote down 👍

beacon/goclient/attest_test.go
```diff
@@ -330,3 +234,353 @@ func TestGoClient_GetAttestationData(t *testing.T) {
...
+		require.NoError(t, err)
+
+		startTime := time.Now()
+		client.GetAttestationData(phase0.Slot(100), phase0.CommitteeIndex(100))
+
+		softTimeout := client.commonTimeout / 2
+		require.Less(t, time.Since(startTime), softTimeout)
+	})
+
+	t.Run("multiple beacon clients: awaits for soft timeout when one the servers slow responder", func(t *testing.T) {
```
anatolie-ssv, line 338: *when one _of_ the servers is _a_ slow responser?

beacon/goclient/attest.go
```diff
@@ -101,9 +104,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+
+	logger.Debug("slot was not found in cache. Fetching from Consensus Client")
+	blockResponse, err := gc.multiClient.BeaconBlockHeader(ctx, &api.BeaconBlockHeaderOpts{
+		Block: root.String(),
+	})
+	if err != nil {
+		return 0, fmt.Errorf("failed to obtain block header: %w", err)
+	}
+
+	fetchedSlot := blockResponse.Data.Header.Message.Slot
```
anatolie-ssv, line 400: maybe I'm wrong but this looks like a scary deep reach down, are we guaranteed to receive the entire hierarchy?

anatolie-ssv (approved):

beacon/goclient/attest.go
```diff
@@ -60,38 +74,27 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		)
+		if gc.withWeightedAttestationData {
+			attestationData, err = gc.weightedAttestationData(slot)
+			if err != nil {
+				return attestationData, err
+			}
+		} else {
+			attestationData, err = gc.simpleAttestationData(slot)
+			if err != nil {
+				return attestationData, err
```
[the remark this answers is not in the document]
oleg-ssvlabs: I agree. Fixed

beacon/goclient/attest.go
oleg-ssvlabs: fixed in multiple places

beacon/goclient/attest.go
oleg-ssvlabs: ✅

beacon/goclient/goclient.go
iurii-ssv: But also we'd want to fix `attestationDataCache` too first

https://github.com/ssvlabs/ssv/pull/2029

beacon/goclient/attest.go
oleg-ssvlabs: Agree. Changed to defer both hard and soft cancels

beacon/goclient/attest.go
oleg-ssvlabs: I agree. This statement is very confusing.
Imagine scenario: 
Requests Total = 10
Responded = 5
Errored = 2

**If there is response**, then [timedOut](https://github.com/ssvlabs/ssv/blob/9227e09e034369e0dc1d22675ae0655b3f1ce920/beacon/goclient/attest.go#L172) will be assigned to be equal to `10 - 5 - 2 = 3`. This assignment ensures the code does not enter the loop that listens to Hard Timeout. It makes sense.

[Then](https://github.com/ssvlabs/ssv/blob/9227e09e034369e0dc1d22675ae0655b3f1ce920/beacon/goclient/attest.go#L188), 
`softTimedOut = 10(total) - 5(responded) - 2(errored) - 3(timedOut) = 0. `

The aim of this should be to exit Soft Timeout loop. This is the condition:

`for responded+errored+timedOut+softTimedOut != requests`

It would not make any difference, since softTimedOut was already equal to `0` and it's the only place where the value is assigned. However, it would make a difference if there are no responses. `timedOut` would be equal to `0`:

`softTimedOut = 10(total) - 5(responded) - 2(errored) - 0(timedOut) = 3`

It would allow exiting the Soft Timeout loop. In general, the condition where `ctx.Done()` is handled means: "regardless of whether there are responses or not, we want to exit the Soft Timeout loop. Additionally, if there is at least one response, we want **not** to enter the Hard Timeout loop."

The code would be 100% easier to read with early returns, though. 
Not sure if it makes :)

oleg-ssvlabs: Wdyt about [this](https://github.com/ssvlabs/ssv/pull/2018/commits/8154a26edc728b31c6e73058c56b2198632547c6) commit?

beacon/goclient/attest.go
```diff
@@ -150,7 +149,6 @@ func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.Attestati
 				zap.String("client_addr", resp.clientAddr),
 				zap.Int("responded", responded),
 				zap.Int("errored", errored),
-				zap.Int("timed_out", timedOut),
```
oleg-ssvlabs: I do not see the reason of this field here. `timedOut` will always be zero here, because it is assigned when soft timeout is reached and after that it immediately exits the loop 😒

beacon/goclient/attest.go
```diff
@@ -165,75 +163,63 @@ func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.Attestati
 				zap.String("client_addr", err.clientAddr),
 				zap.Int("responded", responded),
 				zap.Int("errored", errored),
-				zap.Int("timed_out", timedOut),
```
oleg-ssvlabs: same here

beacon/goclient/attest.go
oleg-ssvlabs: I'll ask another question. Wouldn't it be much easier to simply listen to two contexts in a single `for - select` loop?

beacon/goclient/attest.go
oleg-ssvlabs: There are many logs in the file that are fairly spammy and are repeated for every outgoing call to Consensus Client. However, they do provide value when it comes to understanding the flow. This particular log line is recorded once per outgoing call to Consensus Node

beacon/goclient/goclient.go
oleg-ssvlabs: agree. Fixed

beacon/goclient/attest.go
oleg-ssvlabs: good point. Considering that objects in the hierarchy are pointers, there is a chance for panic. I will add nil checks

beacon/goclient/attest.go
iurii-ssv: Okay, maybe worth reviewing log-spam separately from what this PR is about

beacon/goclient/attest.go
iurii-ssv: > I'll ask another question. Wouldn't it be much easier to simply listen to two contexts in a single for - select loop?

that's probably even better if it's simple to rewrite it like that

beacon/goclient/attest.go
oleg-ssvlabs: fixed. Added [response validation](https://github.com/ssvlabs/ssv/pull/2018/files#)

protocol/v2/blockchain/beacon/client.go
oleg-ssvlabs: fixed

logging/fields/fields.go
oleg-ssvlabs: yes, it is not currently in use. It was in use at some point in my feature branch, then I was slightly hesitant about removing it since it would not really harm to have it there. 
I will remove it for now.

beacon/goclient/goclient.go
oleg-ssvlabs: Yes, you are right.
It is not that cache does not serve the purpose, I would say rather - it could serve its purpose better. Meaning, there is no guarantee it will be utilized, especially in cases when multiple Consensus Clients have very similar response times, then there is a high chance of multiple goroutines entering the [blockRootToSlot()](https://github.com/ssvlabs/ssv/pull/2018/files#diff-b6b7fb7789644004990ef2a0a1be46d921fcd4b6bf46869b67ff758a2aa4239eR364) simultaneously and performing Consensus Node calls and writing into cache. As you say, it would not harm or cause any sort of misbehavior, but the cache would not be utilized. 
I will look into it. I think both `mutex` and `singleflight` could be used here

iurii-ssv: > I think both mutex and singleflight could be used here

jfyi, when implementing attestation code around https://github.com/ssvlabs/ssv/pull/2029 just **singleflight** was enough (it was kind of the goal to simplify away 2 cache-checks and also get rid of mutex)

beacon/goclient/attest.go
oleg-ssvlabs: > > I'll ask another question. Wouldn't it be much easier to simply listen to two contexts in a single for - select loop?
> 
> that's probably even better if it's simple to rewrite it like that

I've tried to refactor using single for-select loop. Still ended up having fair amount of conditions that were not super simple to understand, although the code was slightly shorter. I think it is not worth it to do this refactor. Current implementation is fairly readable. Wdyt @iurii-ssv ?

iurii-ssv: Whichever you feel is a simpler solution I'm fine with (it looks better now btw) :)

beacon/goclient/goclient.go
oleg-ssvlabs: > > I think both mutex and singleflight could be used here
> 
> jfyi, when implementing attestation code around #2029 just **singleflight** was enough (it was kind of the goal to simplify away 2 cache-checks and also get rid of mutex)

I used `singleflight` in the implementation

logging/fields/fields.go
oleg-ssvlabs: removed

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+			errored++
+			gc.log.With(
+				zap.Duration("elapsed", time.Since(started)),
+				zap.String("client_addr", err.clientAddr),
+				zap.Int("responded", responded),
+				zap.Int("errored", errored),
+				zap.Error(err.err),
+			).Error("error received")
+		case <-softCtx.Done():
+			softTimedOut = numberOfRequests - (responded + errored)
```
iurii-ssv, line 168: maybe `responded` -> `succeeded` so it's clear response means success

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+				}
+			case err := <-errCh:
+				errored++
+				gc.log.With(
+					zap.Duration("elapsed", time.Since(started)),
+					zap.String("client_addr", err.clientAddr),
+					zap.Int("responded", responded),
+					zap.Int("errored", errored),
+					zap.Error(err.err),
+				).Error("error received")
```
iurii-ssv, line 203: perhaps `error received` -> `received error fetching attestation data` (and similar for other log lines you've added) so it's clear what this error is about

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		gc.log.Error(clResponseErrMsg,
+			zap.String("api", "AttestationData"),
+			zap.Error(err),
+		)
+		return nil, fmt.Errorf("failed to get attestation data: %w", err)
+	}
+	if resp == nil {
+		gc.log.Error(clNilResponseErrMsg,
+			zap.String("api", "AttestationData"),
+		)
```
iurii-ssv, line 259: could be on 1 line for ease of reading (I guess applies to some other lines in this file too)

beacon/goclient/goclient.go
```diff
@@ -125,18 +128,28 @@ type GoClient struct {
...
+	// that successfully fetched attestation data and proceeded to the scoring phase. Capacity is rather an arbitrary number,
+	// intended for cases where some objects within the application may need to fetch attestation data for more than one slot.
+	blockRootToSlotCache *ttlcache.Cache[phase0.Root, phase0.Slot]
 	// attestationReqInflight helps prevent duplicate attestation data requests
 	// from running in parallel.
 	attestationReqInflight singleflight.Group[phase0.Slot, *phase0.AttestationData]
 
+	// blockRootToSlotReqInflight helps prevent duplicate BeaconBlockHeader requests
+	// from running in parallel.
+	blockRootToSlotReqInflight singleflight.Group[phase0.Root, phase0.Slot]
```
iurii-ssv, line 142: I would group related fields together:
```
attestationReqInflight singleflight.Group[phase0.Slot, *phase0.AttestationData]
attestationDataCache *ttlcache.Cache[phase0.Slot, *phase0.AttestationData]

blockRootToSlotReqInflight singleflight.Group[phase0.Root, phase0.Slot]
blockRootToSlotCache *ttlcache.Cache[phase0.Root, phase0.Slot]
```

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		cacheResult := gc.blockRootToSlotCache.Get(root)
+		if cacheResult != nil {
+			cachedSlot := cacheResult.Value()
+			logger.
+				With(fields.Slot(cachedSlot), zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
+				Debug("obtained slot from cache")
+			return cachedSlot, nil
+		}
+
+		logger.Debug("slot was not found in cache. Fetching from Consensus Client")
```
iurii-ssv, line 377: maybe worth adding **root** to this log line

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		blockResponse, err := gc.multiClient.BeaconBlockHeader(ctx, &api.BeaconBlockHeaderOpts{
+			Block: root.String(),
+		})
+		if err != nil {
+			return 0, fmt.Errorf("failed to obtain block header: %w", err)
```
iurii-ssv, line 382: probably worth adding **root**: `fmt.Errorf("failed to obtain block header: %w", err)` -> `fmt.Errorf("failed to obtain block header for root %s: %w", root, err)`

iurii-ssv (commented):
LGTM, 

just left some nit-comments

iurii-ssv (approved):

beacon/goclient/attest.go
oleg-ssvlabs: there is actually [root](https://github.com/ssvlabs/ssv/pull/2018/files/14e61378fa3d8d955b8f5d066f2977fc6110e381#diff-b6b7fb7789644004990ef2a0a1be46d921fcd4b6bf46869b67ff758a2aa4239eR365) in this log

beacon/goclient/attest.go
oleg-ssvlabs: good point. Fixed

beacon/goclient/attest.go
oleg-ssvlabs: Fixed

beacon/goclient/attest.go
oleg-ssvlabs: Fixed, agree

beacon/goclient/goclient.go
oleg-ssvlabs: Fixed

beacon/goclient/attest.go
iurii-ssv: > and the logger is already [enriched](https://github.com/ssvlabs/ssv/pull/2018/files/14e61378fa3d8d955b8f5d066f2977fc6110e381#diff-b6b7fb7789644004990ef2a0a1be46d921fcd4b6bf46869b67ff758a2aa4239eR337) with the block root

oh didn't see that, nah it's fine then

beacon/goclient/attest_test.go
```diff
@@ -501,6 +522,7 @@ func createBeaconServer(t *testing.T, options beaconServerResponseOptions) (*htt
 
 		//this endpoint is not called for simple attestation data
 		if strings.HasPrefix(r.URL.Path, "/eth/v1/beacon/headers") {
+			time.Sleep(options.BeaconHeadersResponseDuration)
```
iurii-ssv, line 525: maybe leave a clarifying comment for why this sleep is here (and why that amount) - simulate some sort of delay I guess

oleg-ssvlabs: Added few [comments](https://github.com/ssvlabs/ssv/pull/2018/commits/1d45a94e8963861501ef1da3ca37c41db6cc4872)

iurii-ssv (approved):

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+
+		logger.Debug("slot was not found in cache. Fetching from Consensus Client")
+
+		timeoutCtx, cancel := context.WithTimeout(ctx, gc.weightedAttestationDataSoftTimeout/2)
+		defer cancel()
+
+		blockResponse, err := gc.multiClient.BeaconBlockHeader(timeoutCtx, &api.BeaconBlockHeaderOpts{
+			Block: root.String(),
+		})
+		if err != nil {
```
moshe-blox, line 381: requesting the Beacon header adds an unnecessary delay for attestations since most of times we already know it via DutyScheduler's subscription:

https://github.com/ssvlabs/ssv/blob/1d45a94e8963861501ef1da3ca37c41db6cc4872/operator/duties/scheduler.go#L164-L166

perhaps we should "standardize" Beacon head subscriptions so that we always subscribe in goClient on startup and allow consumers to hook in to our subscription (probably only DutyScheduler)

wdyt?

moshe-blox: few notes:
- the cases where we don't know the Beacon blocks ahead of the time of the attestation are when the block wasn't received yet by our Beacon node (in which case there's no point in querying it), or maybe if there is a problem with the subscription (although i expect go-eth2-client to handle subscription reconnections already)
- i'm not sure what's the difference between "head" and "block" event subscriptions, that would be interesting to check

y0sher: yeah I agree here, I think that vouch using this cache across more components..

beacon/goclient/attest.go
```diff
@@ -101,9 +103,309 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
...
+		zap.String("client_addr", bestClient),
+		zap.Float64("score", bestScore)).
+		Debug("successfully fetched attestation data")
+
+	return bestAttestationData, nil
+}
+
+func shouldWaitForAttestationDataResponse(responded, errored, timedOut, requestsTotal int) bool {
+	return responded+errored+timedOut != requestsTotal
+}
```
moshe-blox, line 239: this code is a bit hard to reason about and verify it's correct

for a future PR, i'd look at how this is implemented in beacon-kit because i think we can come up with something more straightforward than both:

https://github.com/ssvlabs/beacon-kit/blob/f56a4c98bc15583ba401980d02a9e527473f3328/multi/client.go#L97-L176

beacon/goclient/attest.go
oleg-ssvlabs: > requesting the Beacon header adds an unnecessary delay for attestations since most of times we already know it via DutyScheduler's subscription:
> 
> https://github.com/ssvlabs/ssv/blob/1d45a94e8963861501ef1da3ca37c41db6cc4872/operator/duties/scheduler.go#L164-L166
> 
> in order to avoid subscription twice (once here and once in scheduler), perhaps we should "standardize" Beacon head event subscriptions so that we always subscribe in goClient on startup and allow consumers to hook in to our subscription (probably only DutyScheduler)
> 
> wdyt?

I think it makes sense, but in this implementation, it isn’t really a subscription. It’s the HTTP call where Block Root is provided as input, and the Slot is expected as output. Additionally, we are making a strong effort to minimize extra calls to the Beacon node by using `singleflight` and caching.
In general it introduces some overhead, but it’s fairly minor based on the metrics I collected from 200+ samples -the median remains almost the same, while the average is slightly higher. However, I wouldn’t rely too heavily on these metrics, as they were gathered from rather unstable Beacon nodes in our staging environment.
When it comes to the Scheduler implementation - yes, I think it's a good idea. It seems we could create a `hashmap<Root, Slot>` with certain capacity that would serve us as a cache. Then it could be utilized by Weighted Attestation Data. Although, we still might need to build some defensive measures, for scenarios like - what if the item is not in the cache? So the issue in the Central Cache does not cascade and cause failures in other components.
In general, I like the idea of a centrally located cache that can provide information to other components

oleg-ssvlabs: Also, I agree with Yosher. I think it would be good to release this version (weighted attestation data is disabled by default), so we can enable it for a number of nodes and observe the behavior in mainnet. Then we can further optimize in the next release

beacon/goclient/attest.go
oleg-ssvlabs: There are a good number of conditions covered by unit tests, although they could be improved. The main challenge with the tests is that we do not mock the HTTP client from `go-eth2-client`, meaning we are effectively testing the library itself, which makes the arrange step in tests quite complex. Mocking this dependency would require a larger refactor to allow dependency injection through the ctor.
I aimed to keep changes to the implementation minimal because, from what I understand, the code has been proven to work as expected. However, some bugs were found, [PRs](https://github.com/attestantio/vouch/pull/294) were created(one [issue](https://github.com/attestantio/vouch/issues/297) as well), and they have been acknowledged by the Vouch maintainers.

Side question - is there a reason `beacon-kit` isn't used by SSV Node?

beacon/goclient/attest.go
moshe-blox: If anything should provide root->slot mapping it's the Beacon client. Scheduler still needs to be notified of new blocks, so it's likely going to have to hook into the Beacon client's existing stream

> we are making a strong effort to minimize extra calls to the Beacon node by using singleflight and caching.
This is good, however the issue isn't load, but latency.

For us it's not an issue, however some operators have their Beacon nodes on a different datacenter, and some Beacon clients respond slowly at times.

If we remove this call, it should half the delay to start attesting, with the only call remaining being `GetAttestationData`.

Though, I agree this could be done in the next PR :)

beacon/goclient/attest.go
```diff
@@ -54,46 +69,339 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot) (
...
+		if cacheResult != nil {
+			cachedSlot := cacheResult.Value()
+			logger.
+				With(zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
+				Debug("obtained slot from cache")
+			return cachedSlot, nil
+		}
+
+		logger.Debug("slot was not found in cache. Fetching from Consensus Client")
+
```
moshe-blox, line 368: to keep our logging consistent (lower case & prefer commas rather than periods)
```suggestion
		logger.Debug("slot was not found in cache, fetching from Consensus Client")

```

beacon/goclient/events.go
```diff
@@ -0,0 +1,88 @@
...
+type event interface {
+	*apiv1.HeadEvent
+}
+
+type subscriber[T event] struct {
+	Identifier string
+	Channel    chan T
+}
+
+func (gc *GoClient) SubscribeToHeadEvents(ctx context.Context, subscriberIdentifier string) (<-chan *apiv1.HeadEvent, error) {
```
moshe-blox, line 20: it's better to let callers provide the channel so they can control their buffer size

beacon/goclient/goclient.go
```diff
@@ -215,9 +250,44 @@ func New(
...
 	go client.attestationDataCache.Start()
 
+	if err := client.launchEventListener(opt.Context); err != nil {
+		return nil, errors.Wrap(err, "failed to launch head event listener")
+	}
+
 	return client, nil
 }
 
+func (gc *GoClient) launchEventListener(ctx context.Context) error {
```
moshe-blox, line 260: launchBlockRootListener should be a less confusing name, bc this name sounds like its generic for events and  should be in `events.go` (but its not)

beacon/goclient/events.go
```diff
@@ -0,0 +1,88 @@
...
+
+	logger.Info("adding 'head' event subscriber")
+
+	if len(gc.headEventSubscribers) == 0 {
+		logger.Info("launching event listener")
+		if err := gc.startEventListener(ctx); err != nil {
+			const errMsg = "failed to start event listener"
+			logger.Error(errMsg, zap.Error(err))
+			return nil, errors.Wrap(err, errMsg)
+		}
```
y0sher, line 33: I don't understand why we need to do this "lazy" starting of the event listener. we're always going to subscribe so its better to just start it on the ctor.

operator/duties/scheduler.go
```diff
@@ -209,6 +207,35 @@ func (s *Scheduler) Start(ctx context.Context, logger *zap.Logger) error {
...
+			select {
+			case <-ctx.Done():
+				return
+			case headEvent := <-headEventChan:
+				logger.
+					With(fields.Slot(headEvent.Slot)).
+					With(fields.BlockRoot(headEvent.Block)).
+					Info("received head event. Processing...")
+
+				headEventHandler(headEvent)
```
olegshmuelov, line 231: Potential nil dereference: 
If `headEventChan` closes unexpectedly or an invalid event is received, `headEvent` could be nil, leading to a panic.

oleg-ssvlabs: Good point. Fixed in [this commit](https://github.com/ssvlabs/ssv/pull/2018/commits/964a3e4ace27fc16b2e657e5a80192822fc08e8f)

beacon/goclient/events.go
y0sher: I do see what you mean. But even without the blockrootcache, there's DutyScheduler listening on these events. It seems like there's no foreseen situation where we can remove all the subscriptions, so I don't see a lot of advantage to preparing for such situation before we can even imagine why would it happen. I'm not saying it doesn't work as it is, it just adds complexity and tangles the code a bit too much IMO for no good reason.

we can use some more input here - @MatusKysel @moshe-blox

moshe-blox: @oleg-ssvlabs if we decide to drop weighted attestation data, we'd still need to subscribe to head events for DutyScheduler

and if we decide to drop that as well, we can just empty supportedTopics slice😄

should be a simpler change than modifying the code, so perhaps the condition should be if `len(supportedTopics) > 0` then start listening

beacon/goclient/events.go
```diff
@@ -14,39 +15,35 @@ type event interface {
...
-			return nil, errors.Wrap(err, errMsg)
-		}
-	}
-
 	headEventSubscriber := subscriber[*apiv1.HeadEvent]{
 		Identifier: subscriberIdentifier,
-		Channel:    make(chan *apiv1.HeadEvent, 32),
+		Channel:    ch,
 	}
 	gc.headEventSubscribers = append(gc.headEventSubscribers, headEventSubscriber)
```
oleg-ssvlabs, line 32: I think we need to lock this slice modification, because event publisher iterates over this slice in a separate goroutine

oleg-ssvlabs: Also we need to add a check `if len(gc.supportedTopics) == 0`, then return error, otherwise subscribers will be added, but no events will be published, which will confuse the caller

beacon/goclient/goclient.go
```diff
@@ -101,13 +103,24 @@ type MultiClient interface {
...
 }
 
+type operatorDataStore interface {
+	AwaitOperatorID() spectypes.OperatorID
+}
+
+type EventTopic string
+
+const (
+	HeadEventTopic EventTopic = "head"
```
moshe-blox, line 121: to follow Go enum standards:
```suggestion
	TopicHeadEvent EventTopic = "head"
```

oleg-ssvlabs: pushed the change. Do not commit the suggestion because all references needs to be updated as well

PR #2022, opened 2025-02-04
fix: (exporter) memory leak in partial sig container

y0sher:
### Description
I already fixed it in #1859, but we haven't merged it so we kept running with the leak.

### DIFF
--- a/protocol/v2/ssv/validator/non_committee_validator.go
+++ b/protocol/v2/ssv/validator/non_committee_validator.go
@@ -3,6 +3,7 @@ package validator
 import (
 	"encoding/hex"
 	"fmt"
+	"github.com/ssvlabs/ssv/message/validation"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -3,6 +3,7 @@ package validator
 import (
 	"encoding/hex"
 	"fmt"
+	"github.com/ssvlabs/ssv/message/validation"
```
nkryuchkov, line 6: IMO it should be in the third block, but it's no big deal

anatolie-ssv (approved):

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -59,19 +61,22 @@ type CommitteeObserverOptions struct {
...
+		beaconNetwork:     opts.NetworkConfig.Beacon,
+		networkConfig:     opts.NetworkConfig,
+		ValidatorStore:    opts.ValidatorStore,
+		newDecidedHandler: opts.NewDecidedHandler,
+		attesterRoots:     opts.AttesterRoots,
+		syncCommRoots:     opts.SyncCommRoots,
+		domainCache:       opts.DomainCache,
 	}
+
+	co.postConsensusContainer = make(map[phase0.Slot]map[phase0.ValidatorIndex]*ssv.PartialSigContainer, co.postConsensusContainerCapacity())
```
nkryuchkov, line 77: why not inside of `&CommitteeObserver{ ... }?

nkryuchkov (approved):

protocol/v2/ssv/validator/non_committee_validator.go
y0sher: because you need to call `co.postConsensusContainerCapacity()`

nkryuchkov (approved):

PR #2025, opened 2025-02-05
fix(Exporter): set fixed subnets to all subnets if exporter is set

y0sher:
### Description

We added a separate metadata syncer component in #1712 , We copied over the logic for `FixedSubnets`, but we called p2p to get it, but since `p2p` we only setup later, the value was empty.

### Solution

This PR parses the fixed subnets separately for the syncer, and sets it to all subnets for an Exporter.


y0sher:
https://github.com/ssvlabs/ssv/pull/2026

PR #2026, opened 2025-02-05
 fix(Exporter): set fixed subnets to all subnets if exporter is set #2025 

y0sher:
### Description

We added a separate metadata syncer component in #1712 , We copied over the logic for `FixedSubnets`, but we called p2p to get it, but since `p2p` we only setup later, the value was empty.

### Solution

This PR parses the fixed subnets separately for the syncer, and sets it to all subnets for an Exporter.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -331,13 +331,24 @@ var StartNodeCmd = &cobra.Command{
 		cfg.SSVOptions.ValidatorOptions.ValidatorStore = nodeStorage.ValidatorStore()
 		cfg.SSVOptions.ValidatorOptions.OperatorSigner = types.NewSsvOperatorSigner(operatorPrivKey, operatorDataStore.GetOperatorID)
 
+		fixedSubnets, err := records.Subnets{}.FromString(cfg.P2pNetworkConfig.Subnets)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


nkryuchkov (approved):

cli/operator/node.go
```diff
@@ -331,13 +331,24 @@ var StartNodeCmd = &cobra.Command{
...
+		fixedSubnets, err := records.Subnets{}.FromString(cfg.P2pNetworkConfig.Subnets)
+		if err != nil {
+			logger.Fatal("failed to parse fixed subnets", zap.Error(err))
+		}
+		if cfg.SSVOptions.ValidatorOptions.Exporter {
+			fixedSubnets, err = records.Subnets{}.FromString(records.AllSubnets)
+			if err != nil {
+				logger.Fatal("failed to parse fixed subnets", zap.Error(err))
+			}
+		}
```
iurii-ssv, line 343: I guess just a clarification question to make sure this is an expected behavior - overriding it this way for exporter means exporter **always** subscribes to ALL subnets (essentially ignoring `cfg.P2pNetworkConfig.Subnets` value)

iurii-ssv (approved):

Issue #1925, opened 2024-12-05
beacon-client: TestGoClient_GetAttestationData race on CI


iurii-ssv:
Spotted a race on CI: https://github.com/ssvlabs/ssv/actions/runs/12178026089/job/33967103053
```
--- FAIL: TestGoClient_GetAttestationData/concurrency:_race_conditions_and_deadlocks (0.17s)
        attest_test.go:287: 
            	Error Trace:	/home/runner/work/ssv/ssv/beacon/goclient/attest_test.go:287
            	            				/home/runner/go/pkg/mod/github.com/sourcegraph/conc@v0.3.0/pool/pool.go:154
            	            				/home/runner/go/pkg/mod/github.com/sourcegraph/conc@v0.3.0/panics/panics.go:23
            	            				/home/runner/go/pkg/mod/github.com/sourcegraph/conc@v0.3.0/waitgroup.go:32
            	            				/opt/hostedtoolcache/go/1.22.9/x64/src/runtime/asm_amd64.s:1695
            	Error:      	Not equal: 
            	            	expected: [98 62 193 32 40 215 119 213 90 158 0 124 139 205 63 182 38 40 120 242 223 208 13 29 129 3 116 184 246 61 247 179]
            	            	actual  : [22 98 163 210 136 176 51 132 54 215 64 131 180 206 104 144 138 14 206 6 97 170 35 106 205 149 200 164 195 246 232 252]
            	            	
            	            	Diff:
            	            	--- Expected
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,4 +1,4 @@
            	            	 (phase0.Root) (len=32) {
            	            	- 00000000  62 3e c1 20 28 d7 77 d5  5a 9e 00 7c 8b cd 3f b6  |b>. (.w.Z..|..?.|
            	            	- 00000010  26 28 78 f2 df d0 0d 1d  81 03 74 b8 f6 3d f7 b3  |&(x.......t..=..|
            	            	+ 00000000  16 62 a3 d2 88 b0 33 84  36 d7 40 83 b4 ce 68 90  |.b....3.6.@...h.|
            	            	+ 00000010  8a 0e ce 06 61 aa 23 6a  cd 95 c8 a4 c3 f6 e8 fc  |....a.#j........|
            	            	 }
            	Test:       	TestGoClient_GetAttestationData/concurrency:_race_conditions_and_deadlocks
        attest_test.go:287: 
[...]

PR #2029, opened 2025-02-07
Fix duplicate API calls in GetAttestationData due to race between cache check and singleflight registration

MatusKysel:
Description:
I discovered a race condition in the GetAttestationData method. The current implementation performs an early cache check outside of the singleflight mechanism. In a high-concurrency scenario, this can lead to the following issue:

1. Request R1 enters the singleflight closure and starts the API call.
2. Request R2 performs the cache check before R1 has stored the result and therefore sees no cached value.
3. R2 then enters the singleflight mechanism (or, in some cases, misses the already in-flight registration) and triggers its own API call.
4. Meanwhile, R1 finishes, caches its result, and returns.
5. R2’s separate API call then runs, causing two API calls for the same slot even though the cache should have prevented duplicates.

To address this, the fix moves the cache check inside the singleflight closure so that all calls go through the singleflight mechanism. This way, when multiple goroutines call GetAttestationData concurrently, they all participate in the same singleflight call. Inside the closure, we perform a “double-check” of the cache before actually calling the API. If a concurrent request has already fetched and cached the result, the closure returns the cached value immediately, and no duplicate API call is made.

resolves https://github.com/ssvlabs/ssv/issues/1925

### DIFF
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -48,18 +48,18 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
 	spec.DataVersion,
 	error,
 ) {
-	// Check cache.
-	cachedResult := gc.attestationDataCache.Get(slot)
-	if cachedResult != nil {
-		data, err := withCommitteeIndex(cachedResult.Value(), committeeIndex)
-		if err != nil {
-			return nil, DataVersionNil, fmt.Errorf("failed to set committee index: %w", err)
-		}
-		return data, spec.DataVersionPhase0, nil
-	}
-
 	// Have to make beacon node request and cache the result.
 	result, err, _ := gc.attestationReqInflight.Do(slot, func() (*phase0.AttestationData, error) {
+		// Check cache.
+		cachedResult := gc.attestationDataCache.Get(slot)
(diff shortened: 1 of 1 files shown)


beacon/goclient/attest.go
```diff
@@ -60,6 +60,16 @@ func (gc *GoClient) GetAttestationData(slot phase0.Slot, committeeIndex phase0.C
 
 	// Have to make beacon node request and cache the result.
 	result, err, _ := gc.attestationReqInflight.Do(slot, func() (*phase0.AttestationData, error) {
+		// Check cache again
```
iurii-ssv, line 63: I think **singleflight** takes care of this in it's `Do` method:

image

this addition would be necessary if we were to use plain-simple mutex, but I don't think here it does anything for us

iurii-ssv: oh hmm, you are actually right, there is that small window where we checked in cache but haven't entered **singleflight** just yet ...

yeah I think that's how I originally wanted to do it - otherwise there is no reason to use **singleflight** even (could just use a simple mutex)

iurii-ssv (approved):

anatolie-ssv (approved):

y0sher (approved):

PR #2030, opened 2025-02-07
refactor(ValidatorRegistration): simplify & optimize submission

iurii-ssv:
This PR clarifies validator-registration flow and also aims to simplify & improve the way validator-registrations are sent to Beacon node, namely we want to:
- submit each validator-registration once per epoch (like Lighthouse or Prysm)
- submit newly produced registrations next slot (they waited 10 epochs, so lets not make them wait another 1)
- to reduce BN load - don't submit everything at the same time (also keep in mind that  multiple operators in cluster can be connected to the same Beacon node - that might create bottleneck scenario if not accounted for)


y0sher (approved):
lgtm. Although as you mentioned I don't think this is a real issue, since the cache is by pubkey, we are overriding the pubkey each time and the cache doesn't grow. There are edge cases where a validator is removed and it might stay in the cache forever unused.

moshe-blox:
@iurii-ssv it may in theory grow perpetually, however its bounded by the amount of registered active validators, so should never be a problem 

whether we can drop the registrations or not depends on how often we should be submitting registrations

i recall that other validator clients submit every epoch, and if thats the case then we should probably stick to it

however i did notice that we seem to submit all registrations every slot? if so that does seem like a bit of an overkill 😄

we should probably be submitting for every validator only once per epoch

iurii-ssv:
I don't really know how validator registrations are supposed to work in full, but cleaning up cache helps with "submitting too often / submitting too much data" (since we only submit when cache is not empty)

submitting every epoch is another way to reduce the amount of requests/data sent I guess, but
- maybe cleaning up cache is good enough ?
-  maybe we want to submit frequently so that the delay of submission is as low as possible (but maybe that doesn't affect anything really)

@moshe-blox ^

moshe-blox:
> I don't really know how validator registrations are supposed to work in full
i think that's once per epoch, you can verify it by reading other validator clients such as Lighthouse

if we remove from the cache, we'll only submit once per 10 epochs, which isn't ideal because if the above is true

what we can do maybe is a hot cache & cold cache system, where we move submitted registrations to the cold cache which is submitted only once per epoch (if `slot%32==0`), where as the hot cache is submitted every slot to keep the small delay

beacon/goclient/goclient.go
```diff
@@ -122,9 +122,14 @@ type GoClient struct {
...
 
-	registrationMu       sync.Mutex
+	// registrationMu synchronises access to registrations
+	registrationMu sync.Mutex
+	// registrations is a set of validator-registrations (their latest versions) to be sent to
+	// Beacon node to ensure various entities in Ethereum network, such as Relays, are aware of
+	// participating validators
+	registrations map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration
+	// registrationLastSlot helps to decide when registrations should be sent to Beacon node
 	registrationLastSlot phase0.Slot
```
iurii-ssv, line 132: Note `registrationLastSlot` does not need to be protected by `registrationMu` - so we can simplify the code around `registrationMu` usage a bit (which I did)

beacon/goclient/proposer.go
```diff
@@ -372,51 +371,44 @@ func (gc *GoClient) registrationSubmitter(slotTickerProvider slotticker.Provider
...
+			oneEpochPassed := slotsSinceLastRegistration >= phase0.Slot(slotsPerEpoch)
+			operatorSubmissionSlotModulo := operatorID % slotsPerEpoch
+			operatorSubmissionSlot := uint64(currentSlot)%slotsPerEpoch == operatorSubmissionSlotModulo
+
+			twoEpochsAndOperatorDelayPassed := uint64(slotsSinceLastRegistration) >= slotsPerEpoch*2+operatorSubmissionSlotModulo
+
+			if oneEpochPassed && operatorSubmissionSlot || twoEpochsAndOperatorDelayPassed {
+				gc.submitRegistrationsFromCache(currentSlot)
+				gc.registrationLastSlot = currentSlot
+			}
```
iurii-ssv, line 387: @moshe-blox I think this logic we have here (I've just moved it out of `submitRegistrationsFromCache ` to simplify it) roughly does similar thing to what you describe in https://github.com/ssvlabs/ssv/pull/2030#issuecomment-2740027286 (to reduce submission frequency)

it's maybe not perfect but if it works I think we should leave it as is - since improvements would be marginal

btw, it seems to me `twoEpochsAndOperatorDelayPassed` here is a dead code since whenever `twoEpochsAndOperatorDelayPassed` is **true** then `oneEpochPassed && operatorSubmissionSlot` will also be **true** (since `oneEpochPassed && operatorSubmissionSlot` must turn into **true** by the end of 2nd epoch)

beacon/goclient/goclient.go
iurii-ssv: `registrationLastSlot` can be completely removed (with the latest solution)

beacon/goclient/proposer.go
iurii-ssv: We discussed it further with Moshe and decided to simplify this a bit

operator/duties/validatorregistration.go
```diff
@@ -45,11 +48,11 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
 			slot := h.ticker.Slot()
 			next = h.ticker.Next()
 			epoch := h.network.Beacon.EstimatedEpochAtSlot(slot)
-			shares := h.validatorProvider.SelfParticipatingValidators(epoch + phase0.Epoch(validatorRegistrationEpochInterval))
+			shares := h.validatorProvider.SelfParticipatingValidators(epoch + phase0.Epoch(frequencyEpochs))
 
 			var vrs []ValidatorRegistration
 			for _, share := range shares {
-				if uint64(share.ValidatorIndex)%registrationSlotInterval != uint64(slot)%registrationSlotInterval {
+				if !h.suitableRegistrationSubmissionSlot(share.ValidatorPubKey, slot) {
```
iurii-ssv, line 55: I'm using "validator pubkey" here now (instead of "validator index") just to match how we do it in `GoClient` (where we don't have "validator index")

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...370d","2169fe3af9e88ab9","a85dccf18844fc79","2bf91d3c3920c5c8","f4ce01b385c68a2","4366899a2cb05197","6f0aecc3342b13c0","874b67f800dd74d6","587785bbbcdc2016","d7397265deb360a6","d9fabd130a8ecdfb","707b55fec1513c90","e51b0346b0d4b612","417e005d18c1f7d4","306964e5ede31618","8c7ca4bb2b8abe1e","b878d79c5774d12d","15599c2b2bd60293","a7891090c37daba1","4192f3dd9bdddea8","ba5b31e5a1adb8e2","dadf83703db9234","3fd72034ab5b908","e939b25394581c4","67d24278154582d6","bc6a77eea2a1ba4b","41fa2c6b8e3aed38","1b6e3c30093fcd6e","98377d7e6f0eea0a","267886c6b07733d7","280250b5f6148515","1b79a8c6288d49d8","47c23de3a5c71c7e","de0f0cd6b40ee150","e91f5b3d583f23af","2036a5ff6aed3717","5ed363e465b9c98b","3ed6b5d40fb4a3b0","e7219eac943d306f","186382b7e69faced","943be3ce709a99d3","3a8800d72c312f75","b9bf74de5674c15d","cb9ad84a103b6499","acdcac18a6e34419", "ee323f91ee6f543c","be9935e5c2bedaef","60ac8eb415e6c748","119e95426dc4affb","e918003e25abf4cd","44c0efaa4b189434","de6ef5f1aa07b38d","b71a695e14c7e892","c9b7984a5d2604b5","72e9cd10959558de","82f692ac2f5817f3","3e0d3d7ade790448","52595c2a99ba6031","ea6ad19cdc29b230","695cba9190654139","4418aa6010632480","e27b9c1ea1c64357","fe62097e8814c106","1cfd3a3660a62879"]
```
iurii-ssv, line 1: Diff:
```
index 1cfd3a3660a62879..2222222
@@ -1,39 +1,34 @@
 	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
 	if err != nil {
 		return errors.Wrap(err, "failed processing validator registration message")
 	}
 	if !quorum {
 		return nil
 	}
 	root := roots[0]
 	if err != nil {
 		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee,
 			r.GetShare().ValidatorIndex)
 		return errors.Wrap(err, "got pre-consensus quorum but it has invalid signatures")
 	}
 	copy(specSig[:], fullSig)
-	share := r.GetShare()
-	if share == nil {
-		return errors.New("no share to get validator public key")
-	}
 	registration, err := r.calculateValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
 	if err != nil {
 		return errors.Wrap(err, "could not calculate validator registration")
 	}
 			Message:	registration,
 			Signature:	specSig,
 		},
 	}
-	err = r.beacon.SubmitValidatorRegistration(signedRegistration)
-	if err != nil {
 		return errors.Wrap(err, "could not submit validator registration")
 	}
 	r.GetState().Finished = true
 	return nil
 }
\ No newline at end of file
```

beacon/goclient/goclient.go
```diff
@@ -134,9 +134,12 @@ type GoClient struct {
...
 
-	registrationMu       sync.Mutex
-	registrationLastSlot phase0.Slot
-	registrationCache    map[phase0.BLSPubKey]*api.VersionedSignedValidatorRegistration
+	// registrationMu synchronises access to registrations
+	registrationMu sync.Mutex
+	// registrations is a set of validator-registrations (their latest versions) to be sent to
+	// Beacon node to ensure various entities in Ethereum network, such as Relays, are aware of
+	// participating validators
+	registrations map[phase0.BLSPubKey]*validatorRegistration
```
olegshmuelov, line 142: Would it make sense to eventually expire old validator registrations from the registrations map (e.g., if a validator is exited/slashed or removed)?

beacon/goclient/validator.go
```diff
@@ -36,3 +48,83 @@ func (gc *GoClient) GetValidatorData(validatorPubKeys []phase0.BLSPubKey) (map[p
...
+				slotInEpoch := uint64(currentSlot) % gc.network.SlotsPerEpoch()
+				validatorHash := sha256.Sum256(validatorPk[:8])
+				validatorDescriptor := binary.LittleEndian.Uint64(validatorHash[:])
+				shouldSubmit := validatorDescriptor%gc.network.SlotsPerEpoch() == slotInEpoch
+
+				if r.new || shouldSubmit {
+					r.new = false
+					registrations = append(registrations, r.VersionedSignedValidatorRegistration)
+				}
+			}
```
olegshmuelov, line 110: With ~1000 registrations max, the tradeoff between copying `maps.Values()` and filtering inside the lock seems minor. Was this mainly to reduce lock time and avoid blocking `SubmitValidatorRegistration()`?

setting `r.new = false` after unlocking could lead to a race — if the registration gets replaced before that line runs, it might cause unnecessary resubmissions.

beacon/goclient/validator.go
```diff
@@ -36,3 +48,83 @@ func (gc *GoClient) GetValidatorData(validatorPubKeys []phase0.BLSPubKey) (map[p
...
+			registrations := make([]*api.VersionedSignedValidatorRegistration, 0)
+			for _, r := range allRegistrations {
+				validatorPk, err := r.PubKey()
+				if err != nil {
+					gc.log.Error("Failed to get validator pubkey", zap.Error(err), fields.Slot(currentSlot))
+					continue
+				}
+
+				// Distribute the registrations evenly across the epoch based on the pubkeys.
+				slotInEpoch := uint64(currentSlot) % gc.network.SlotsPerEpoch()
```
olegshmuelov, line 101: Minor: `slotInEpoch := uint64(currentSlot) % gc.network.SlotsPerEpoch()` could be moved outside the loop since it doesn't change per validator

beacon/goclient/goclient.go
iurii-ssv: It would make sense to do that, but I don't think it would justify the added complexity (since "validator is exited/slashed or removed" events are rare and node-restarts solve it eventually)

beacon/goclient/validator.go
iurii-ssv: > setting r.new = false after unlocking could lead to a race — if the registration gets replaced before that line runs, it might cause unnecessary resubmissions

Right, not sure if in **stage** branch all racy behavior is avoided - but here in this PR we certainly have races like that - still I think these are non-harmful (sending registration couple times extra isn't that bad)

while on the upside it simplifies mutex usage quite a bit

> With ~1000 registrations max, the tradeoff between copying maps.Values() and filtering inside the lock seems minor. Was this mainly to reduce lock time and avoid blocking SubmitValidatorRegistration()?

I think there just is no need to hold this mutex locked while filtering,

unless you want to make operations involving `r.new` atomic ... which IMO isn't worth the added complexity (it's also not super obvious to lock this mutex to achieve that - comment would help somewhat with that ... but long comments aren't ideal either)

beacon/goclient/goclient.go
iurii-ssv: Hmm, that's an interesting scenario you are describing,

I'm not sure how realistic/bad it actually is (maybe @moshe-blox or @y0sher could chime in on it) but from what I understand:
- both Y and Z recipients "belong" to the "same user" (the owner of Validator X) - meaning he should be able to get those funds even though they might be sent to old address due to all these circumstances
- eventually (after a restart) operator 1 will stop sending out this old info, and so it will resolve

thus it doesn't seem too bad ? wdyt

olegshmuelov:
This PR improves and clarifies how validator-registrations are submitted to the Beacon Node. The design separates the process into:
1. Duty Scheduling (`ValidatorRegistrationHandler`)
   - Every slot, the handler loads participating shares for epoch + `frequencyEpochs` (currently +10).
   - It executes validator-registration duties only if:
   ```go
   uint64(share.ValidatorIndex) % registrationSlots == uint64(slot) % registrationSlots
   ```
2. Duty Execution (`ProcessPreConsensus`)  → Submission
   - Once reach pre consensus quorum, a duty sends a `VersionedSignedValidatorRegistration` to `GoClient.SubmitValidatorRegistration()`.
   
3. Registration Submission (`registrationSubmitter`)
   - Runs every slot, checks all registrations.
   - Submits:
      - Fresh (.new) registrations immediately
      - All registrations once per epoch, distributed deterministically by pubkey hash
   - Batches submissions in groups of 500 to avoid BN overload.
   
   
**Known Issues**:
Delayed submission for new added validators or fee recipient updates.
There’s a timing edge case where a new validator registration or fee recipient update might not take effect in time for a proposal. Here’s how it can happen:
1. A validator is added to the SSV network, or its fee recipient is updated
This creates a new validator registration that needs to be submitted to the Beacon Node.
2. The duty scheduler attempts to assign a BNRoleValidatorRegistration duty
   - But duties are only scheduled if:
   ```go
   validatorIndex % registrationSlots == slot % registrationSlots
   ```
   - If the validator’s index doesn’t match the current slot (especially near the end of the `frequencyEpochs` window), the duty might be skipped for now.
3. In the meantime, a block proposal duty is triggered for that validator
4. The proposal is built and submitted with the old or fallback fee recipient (e.g., the owner address)
   since the Beacon Node hasn’t received the updated validator registration yet.


another **spec** side issue:
https://github.com/ssvlabs/ssv-spec/issues/504

olegshmuelov:
> **Known Issues**: Delayed submission for new added validators or fee recipient updates. There’s a timing edge case where a new validator registration or fee recipient update might not take effect in time for a proposal. Here’s how it can happen:
> 
> 1. A validator is added to the SSV network, or its fee recipient is updated
>    This creates a new validator registration that needs to be submitted to the Beacon Node.
> 2. The duty scheduler attempts to assign a BNRoleValidatorRegistration duty
>    
>    * But duties are only scheduled if:
>    
>    ```go
>    validatorIndex % registrationSlots == slot % registrationSlots
>    ```
>    * If the validator’s index doesn’t match the current slot (especially near the end of the `frequencyEpochs` window), the duty might be skipped for now.
> 3. In the meantime, a block proposal duty is triggered for that validator
> 4. The proposal is built and submitted with the old or fallback fee recipient (e.g., the owner address)
>    since the Beacon Node hasn’t received the updated validator registration yet.

If we aim to further improve and harden the validator registration flow, we should consider addressing at least one of the known issues above

beacon/goclient/goclient.go
olegshmuelov: > * both Y and Z recipients "belong" to the "same user" (the owner of Validator X) - meaning he should be able to get those funds even though they might be sent to old address due to all these circumstances

The assumption that "Y and Z belong to the same user" is risky.
In a permissionless system like Ethereum, the protocol can't rely on that being true.

olegshmuelov: > Also, that seems to be a problem for **stage** version as well, right ?

If registrations aren’t removed from the cache after validator removals, then yes this issue likely exists in both stage and production.

beacon/goclient/validator.go
iurii-ssv: Oh wait you are right, I meant for `new` to be `atomic.Bool`

I thought I've defined it like that already but looks like I forgot - changed it now https://github.com/ssvlabs/ssv/pull/2030/commits/80f8d9448d21d6903e78492b6376c399c07a36cd so the data race should no longer be an issue

olegshmuelov: Thanks for the detailed explanation! Appreciate you thinking it through.
Just worth noting - this is still a data race under the Go memory model, and relying on intuition like “only one goroutine accesses it after init” isn’t safe in concurrent code. Even benign races are better avoided, especially in infra-level systems.
That said, fine to leave it as-is if we agree the impact is negligible.

beacon/goclient/goclient.go
iurii-ssv: I pretty much agree with all your input, just not sure where exactly it fits on our priority list - so I'll create an issue to document the problem for now (so we don't loose it) https://github.com/ssvlabs/ssv/issues/2105 but it doesn't have to a part of this PR

beacon/goclient/validator.go
y0sher: Even though there shouldn't be significant resources issue because of this being inside the loop, I still thinks in terms of readability it makes more sense to happen once out of the loop. It's being inside the loop might make you think it needs to be calculated per each registration.

moshe-blox: +1

IMO wherever we define `currentSlot` is where we should define `slotInEpoch`

beacon/goclient/goclient.go
iurii-ssv: @olegshmuelov I've looked a bit more into this and I think the following:

the scenario you've described might happen shortly after "fee recipient is updated to Z" while cluster 2,3,4,5 hasn't had the time to submit validator-registration with new fee recipient BUT as soon as cluster 2,3,4,5 does submit it I think it should take precedence due to fresher timestamp on that validator-registration (compared to older validator-registration operator 1 might be re-submitting over and over again)

so it **seems to result in correct behavior as is**, I can't think of anything we can change to improve it - "expire old validator registrations from the registrations map" doesn't seem to change anything really from what I understand)

olegshmuelov: Thanks for digging into this!

It makes sense if the BN or relay consistently prioritizes the timestamp field — in that case, newer registrations should take precedence. But since this behavior isn’t enforced by the spec and can vary across clients and relays, we can’t safely rely on it.

For example, Lighthouse allows overriding the registration timestamp via `--builder-registration-timestamp-override`, which can make older registrations appear newer. Similarly, Teku supports setting a static `registration_overrides.timestamp`, meaning operators can pin timestamps manually. So rather than relying on how each operator configures their node, we should avoid creating a situation where this turns into a subtle bug.

Also worth noting, if a validator is removed from SSV and migrated to a different staking provider, we don’t want to keep ghost-submitting stale registrations from the old cluster indefinitely. Cleaning up outdated entries proactively would help avoid that.

olegshmuelov: Agreed on the timestamp it's a VC-side override
Also +1 on using `validatorStore.SelfParticipatingValidators(...)`
We could also delete any non-participating validators from the cache - if a validator isn’t in the participating set anymore, we can safely delete it from `gc.registrations` to prevent unnecessary memory growth or ghost submissions.

y0sher: I agree eventually we need to clean the cache, It's currently not very bad since like @olegshmuelov said it existed in the code before. `SelfParticipatingValidators` could be a good idea. though this rings to me like we just need to manage validators in a central component and provide to all the code a way to track changes. #2060 also touches similar places. So maybe we could live with this inside until we provide some wider solution.

iurii-ssv: > i think a trivial and performant solution would be to iterate validatorStore.SelfParticipatingValidators(epoch + phase0.Epoch(frequencyEpochs)) (like DutyScheduler does) instead of the cache, and then only submit for those validators

\+

> We could also delete any non-participating validators from the cache - if a validator isn’t in the participating set anymore, we can safely delete it from gc.registrations to prevent unnecessary memory growth or ghost submissions.

= we essentially need to 
- remove non-participating validators from the cache (keeping only those returned by `validatorStore.SelfParticipatingValidators(epoch + phase0.Epoch(frequencyEpochs)`)
- and "submit registrations for all validators in the cache"

I'll can sketch this out ^

But a bit more pressing issue I think is what I describe here - https://github.com/ssvlabs/ssv-spec/issues/504#issuecomment-2780862223 - @moshe-blox @y0sher @olegshmuelov wdyt of solution I mentioned in there ?

beacon/goclient/validator.go
iurii-ssv: I'll address this in https://github.com/ssvlabs/ssv/pull/2130

beacon/goclient/goclient.go
iurii-ssv: I've added a [commit](https://github.com/ssvlabs/ssv/pull/2130/commits/70ace1119394f26ebd135a7bdee12d322ab50b46) to https://github.com/ssvlabs/ssv/pull/2130 that should address this ^ (essentially applying [Moshe's suggestion](https://github.com/ssvlabs/ssv/pull/2030#discussion_r2030101100) which should take care of "ghost submissions" but not doing any pruning - since it's a bit harder to implement while memory growth issue it solves is probably negligible ... so probably not worth bothering with ?)

PR #2033, opened 2025-02-12
chore(metrics): measure save participant store

anatolie-ssv:


ibft/storage/metrics.go
oleg-ssvlabs, line 1: We name files that contain metrics/traces implementation `observability.go`

ibft/storage/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+
+const (
+	observabilityName      = "github.com/ssvlabs/ssv/iqbft/storage"
+	observabilityNamespace = "ssv.storage"
+)
+
+var (
+	meter = otel.Meter(observabilityName)
+
+	saveParticipantsDurationHistogram = observability.NewMetric(
```
oleg-ssvlabs, line 22: I would not keep **operation**(`save`) as part of the metric name. I think it is better to set it as attribute and make metric itself more generic. [OTeL example](https://opentelemetry.io/docs/specs/semconv/database/database-metrics/#metric-dbclientoperationduration)
So
`ssv.storage.operation.duration`
instead of 
`ssv.storage.save.duration`

then you can add attribute `db.operation.name` equal to `save` (`semconv.DBOperationName(...)` exists at `semconv "go.opentelemetry.io/otel/semconv/v1.26.0"` pkg)

ibft/storage/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+		meter.Float64Histogram(
+			metricName("save.duration"),
+			metric.WithUnit("ms"),
```
oleg-ssvlabs, line 25: We use `seconds` for all durations. So, `metric.WithUnit("s")`.
In Go `duration.Seconds` is type of `float64`, so we aren't losing any precision

ibft/storage/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+			metric.WithDescription("participants save duration"),
+			metric.WithExplicitBucketBoundaries(observability.SecondsHistogramBuckets...)))
+)
+
+func metricName(name string) string {
+	return fmt.Sprintf("%s.%s", observabilityNamespace, name)
+}
+
+type metrics struct {
```
oleg-ssvlabs, line 34: We were usually not building any wrappers around OTeL and kept it simple, like [this](https://github.com/ssvlabs/ssv/blob/main/beacon/goclient/observability.go#L57). So start - end were implemented directly in the code. I am not saying this is the best way of doing things, it's just the way we have it

oleg-ssvlabs (changes requested):

ibft/storage/store.go
```diff
@@ -163,6 +163,9 @@ func (i *participantStorage) CleanAllInstances() error {
 }
 
 func (i *participantStorage) SaveParticipants(pk spectypes.ValidatorPK, slot phase0.Slot, newParticipants []spectypes.OperatorID) (updated bool, err error) {
+	start := time.Now()
+	defer recordRequestDuration(i.ID(), time.Since(start), updated)
```
oleg-ssvlabs, line 167: https://github.com/ssvlabs/ssv/actions/runs/13283332483/job/37086350562?pr=2033#step:6:28

ibft/storage/observability.go
```diff
@@ -0,0 +1,43 @@
...
+func metricName(name string) string {
+	return fmt.Sprintf("%s.%s", observabilityNamespace, name)
+}
+
+func recordRequestDuration(name string, duration time.Duration, updated bool) {
+	saveParticipantsDurationHistogram.Record(
+		context.Background(),
+		duration.Seconds(),
+		metric.WithAttributes(
+			attribute.String("store", name),
```
oleg-ssvlabs, line 40: I would this attribute here:
`semconv.DBOperationName(operationName)` (you need to import `semconv "go.opentelemetry.io/otel/semconv/v1.26.0"` package)

`operationName` as string would need to be added to `recordRequestDuration` as function parameter. 
When it comes to other attributes - they need to have namespaces. Attribute does not belong to metric. Attributes live in its own "universe". They are searchable.
so 
`updated` -> `ssv.storage.operation.updated`(or something similar)
`store` -> `ssv.storage.name` (or something similar)
so attribute must be part of some meaningful hierarchy of objects

oleg-ssvlabs (approved):
🔥

nkryuchkov (changes requested):
LGTM, except typo in the path

nkryuchkov (approved):

PR #2034, opened 2025-02-12
fix(exporter): memory leak and lock contentions

y0sher:
Includes 
- #1874 
- #1883 
- remove some unnecessary mutexes.


operator/validator/controller.go
```diff
@@ -350,57 +347,69 @@ func (c *controller) handleRouterMessages() {
...
+	existingObserver, ok := c.committeesObservers.Get(msgID)
+	if ok {
+		c.committeesObserversMutex.RUnlock()
+		return existingObserver
+	}
+
+	c.committeesObserversMutex.RUnlock()
+
 	c.committeesObserversMutex.Lock()
 	defer c.committeesObserversMutex.Unlock()
```
moshe-blox, line 382: i think it'a potential race condition because another goroutine might also see non-existance in-between our RUnlock -> Lock and then overwrite us later

if we want maximum optimization, we could re-check existance after Lock and exit if it already exists

however if you don't think it's a big deal for performance then we can just do this:
```suggestion
	c.committeesObserversMutex.Lock()
	defer c.committeesObserversMutex.Unlock()

	// Check if the observer already exists
	existingObserver, ok := c.committeesObservers.Get(msgID)
	if ok {
		return existingObserver
	}
```

operator/validator/controller.go
```diff
@@ -175,9 +176,9 @@ type controller struct {
 	historySyncBatchSize int
 	messageValidator     validation.MessageValidator
 
-	// nonCommittees is a cache of initialized committeeObserver instances
-	committeesObservers      *ttlcache.Cache[spectypes.MessageID, *committeeObserver]
-	committeesObserversMutex sync.Mutex
+	// committeeObservers is a cache of initialized committeeObserver instances
+	committeesObservers      *hashmap.Map[spectypes.MessageID, *committeeObserver] // todo: need to evict?
```
moshe-blox, line 180: we don't have to, it's an optimization to stay lean by evicting:
- removed validators
- proposal duties

if you don't delete them they accumulate over time, however we can try without it and revert if needed

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -205,29 +209,41 @@ func (ncv *CommitteeObserver) getBeaconRoles(msg *queue.SSVMessage, root phase0.
...
 }
 
+// nonCommitteeInstanceContainerCapacity returns the capacity of InstanceContainer for non-committee validators
+func nonCommitteeInstanceContainerCapacity(fullNode bool) int {
+	if fullNode {
+		// Helps full nodes reduce
+		return 2
+	}
+	return 1
+}
```
moshe-blox, line 222: is it unused?

operator/validator/controller.go
y0sher: I guess this optimization is not necessary, I tried it before the last one that worked. I'll revert.

operator/validator/controller.go
y0sher: I agree we should eventually evict but I felt ttl cache might not be the best here. maybe we could somehow remove it on actual events in the future.

message/validation/consensus_state.go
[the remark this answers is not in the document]
nkryuchkov: @moshe-blox I think you're right, perhaps we could remove them

message/validation/consensus_state.go
```diff
@@ -4,36 +4,29 @@ import (
...
 
-	return cs.state[signer]
+	return cs.operators[idx]
 }
 
 type OperatorState struct {
-	mu              sync.RWMutex
-	state           []*SignerState // the slice index is slot % storedSlotCount
+	mu              sync.Mutex
+	signers         []*SignerState // the slice index is slot % storedSlotCount
```
nkryuchkov, line 29: it's operator/signer state indexed by slot, so I'm not sure if this name is better

nkryuchkov (approved):

PR #2035, opened 2025-02-13
committee: improve concurrency handling

iurii-ssv:
Copy of https://github.com/ssvlabs/ssv/pull/1784 to replace it.


moshe-blox:
the changes in `committee.go` look good, i pushed a couple of renames

regarding the error override mechanism: curious to know why do you think it's worth the added complexity and maintenance overhead? if we'd like to see more context, can't we just add a log?

iurii-ssv:
> regarding the error override mechanism: curious to know why do you think it's worth the added complexity and maintenance overhead? if we'd like to see more context, can't we just add a log?

Do you mean we could just check if there is an error or not without checking error message ? 

If yes, I think it's worth to match error-messages exactly in spec tests because there are many different error scenarios - when we change/refactor some code that all of a sudden starts to return **different** error "there is a good chance such a refactoring could have done something we didn't intend"

so it's just an additional safety for us to make sure core functionality still works as previously (if it becomes too much friction to maintain this error mapping we could ofc get rid of it - but I'd feel much safer with it in place)

iurii-ssv:
> this error message is now saying "consensus", however this should be redundant as the parent who logs this error should attach the msg type already:
> 
> https://github.com/ssvlabs/ssv/blob/e9d57254dafb5c05669df9795370d86a3239e47d/protocol/v2/ssv/validator/committee.go#L224-L226

Right 👍 updated here - https://github.com/ssvlabs/ssv/pull/2035/commits/fcbc42d7611b33c5a709ff9019b4e00a460efd85

iurii-ssv:
Removed that error-override mechanism (used in spectests) to keep in simple

zippoxer (approved):

zippoxer (approved):

moshe-blox (approved):

PR #2036, opened 2025-02-13
feat: Doppelganger

olegshmuelov:
Doppelganger Protection (DG) in ssv.network prevents validators from signing if they are already active elsewhere, reducing the risk of slashing. It leverages validator liveness tracking and post-consensus quorum to ensure safe operation in a distributed staking setup.


doppelganger/design.md
```diff
@@ -0,0 +1,15 @@
+# Doppelganger Protection - Design Overview
+
+This document provides an overview of the Doppelganger Protection mechanism in `ssv.network`.
+
+🔗 **Full Documentation**: [Doppelganger Protection Design on HackMD](https://hackmd.io/@TjjCyxe9Qxm0hOn3HZ_YIA/HygAf4M5yg)
```
y0sher, line 5: lets include the full doc here. no need to link to outside of the repo where there's no version tracking

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+//go:generate mockgen -package=doppelganger -destination=./doppelganger_mock.go -source=./doppelganger.go
+
+// DefaultRemainingDetectionEpochs represents the initial number of epochs
+// a validator must pass without liveness detection before being considered safe to sign.
+const DefaultRemainingDetectionEpochs uint64 = 2
+
+// PermanentlyUnsafe is a special flag value used to mark a validator as permanently unsafe for signing.
+// It indicates that the validator was detected as live on another node and should not be trusted for signing.
+const PermanentlyUnsafe = ^uint64(0)
+
```
y0sher, line 28: since we have a low range of values (2/1/0) and the permanently unsafe. I think we should switch to using `int8` and `PermanentlyUnsafe` could be `-1`. this will make the data structure a bit lighter.

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+type DoppelgangerProvider interface {
```
y0sher, line 29: general note that I think fits most of this file. there's no need to put `Doppleganger` at the beginning of each struct/iface. The package name is `doppleganger` so if only call it `Provider` for example so it will be called like - `doppleganger.Provider`. In short, no need to repeat ourselves, its better to keep things short for naming.

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+	validatorProvider  ValidatorProvider
+	slotTickerProvider slotticker.Provider
+	logger             *zap.Logger
+	doppelgangerState  map[phase0.ValidatorIndex]*doppelgangerState
+
+	startEpoch phase0.Epoch
+}
+
+// NewDoppelgangerHandler initializes a new instance of the Doppelgänger protection.
+func NewDoppelgangerHandler(opts *DoppelgangerOptions) DoppelgangerProvider {
```
y0sher, line 66: return concrete types not interfaces. there's no benefit to returning the interface.
other components can use the interface for dep injection if they want.

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+		currentValidatorSet[idx] = struct{}{}
+	}
+
+	// Add new validators with DefaultRemainingDetectionEpochs
+	for _, idx := range validatorIndices {
+		if _, exists := ds.doppelgangerState[idx]; !exists {
+			ds.doppelgangerState[idx] = &doppelgangerState{
+				remainingEpochs: DefaultRemainingDetectionEpochs,
+			}
+			ds.logger.Debug("Added validator to Doppelganger state", fields.ValidatorIndex(idx))
```
y0sher, line 124: we can probably do one log with a list to not spam our logs..

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (ds *doppelgangerHandler) updateDoppelgangerState(validatorIndices []phase0.ValidatorIndex) {
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	// Create a set for current validators
+	currentValidatorSet := make(map[phase0.ValidatorIndex]struct{}, len(validatorIndices))
+	for _, idx := range validatorIndices {
+		currentValidatorSet[idx] = struct{}{}
+	}
```
y0sher, line 116: since this is only used in line 129 is best to move it above the place where its used. even if just for readability.

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+	if _, exists := ds.doppelgangerState[validatorIndex]; !exists {
+		ds.logger.Warn("Validator not found in Doppelganger state", fields.ValidatorIndex(validatorIndex))
+		return
+	}
+
+	delete(ds.doppelgangerState, validatorIndex)
+	ds.logger.Debug("Removed validator from Doppelganger state", fields.ValidatorIndex(validatorIndex))
+}
+
+func (ds *doppelgangerHandler) StartMonitoring(ctx context.Context) {
```
y0sher, line 150: IMO `Start` makes more sense as its a convention.

y0sher: Even `Start(ctx) error` and return an error when ctx is done like `http` server module is doing for example

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+	ticker := ds.slotTickerProvider()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.Next():
+			currentSlot := ticker.Slot()
+			currentEpoch := ds.network.Beacon.EstimatedEpochAtSlot(currentSlot)
+			slotsPerEpoch := ds.network.Beacon.SlotsPerEpoch()
```
y0sher, line 163: can this potentially change? if not lets define it once at the top of the func

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+				firstRun = false
+			}
+
+			ds.checkLiveness(ctx, currentEpoch-1)
+		}
+	}
+}
+
+func (ds *doppelgangerHandler) checkLiveness(ctx context.Context, epoch phase0.Epoch) {
+	ds.mu.Lock()
```
y0sher, line 188: this could be `RLock`

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+			state.remainingEpochs = PermanentlyUnsafe // Mark as permanently unsafe
+			continue
+		}
+
+		// If the validator was previously marked as permanently unsafe (detected as live elsewhere),
+		// but is now considered inactive, we reset the detection period to be safer.
+		// Since we just checked for liveness now, we reduce the default detection period by 1
+		// to ensure it gets checked once again in the next epoch before being marked safe.
+		if state.remainingEpochs == PermanentlyUnsafe {
+			state.remainingEpochs = DefaultRemainingDetectionEpochs - 1
```
y0sher, line 241: That makes me think **Permanently**Unsafe might not be good name. wdyt about just `Unsafe` or something like `PendingChecks`

doppelganger/doppelganger_state.go
```diff
@@ -0,0 +1,28 @@
+package doppelganger
+
+// doppelgangerState tracks the validator's state in Doppelganger Protection.
+type doppelgangerState struct {
```
y0sher, line 4: Like I commented in `doppleganger.go`, I think this should just be `state` , also the filename.

doppelganger/doppelganger_state.go
```diff
@@ -0,0 +1,28 @@
...
+	remainingEpochs uint64 // The number of epochs that must be checked before it's considered safe.
```
y0sher, line 5: Like I commented on the const states, I think this could be just `int8` since only options are 1/2 and invalid (imo -1)

protocol/v2/ssv/runner/committee.go
```diff
@@ -237,6 +241,12 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 		switch validatorDuty.Type {
 		case spectypes.BNRoleAttester:
 			validDuties++
+
+			if cr.doppelgangerHandler.ValidatorStatus(validatorDuty.ValidatorIndex) != doppelganger.SigningEnabled {
+				logger.Warn("Doppelganger check in progress, signing not permitted", fields.ValidatorIndex(validatorDuty.ValidatorIndex))
```
y0sher, line 246: what about a case where there's only one valid duty but it's a doppleganger duty. might make us send an empty post-consensus message.

protocol/v2/ssv/runner/committee.go
```diff
@@ -410,6 +420,11 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 
 			vlogger.Debug("🧩 reconstructed partial signatures committee",
 				zap.Uint64s("signers", getPostConsensusCommitteeSigners(cr.BaseRunner.State, root)))
+
+			if cr.doppelgangerHandler.ValidatorStatus(validator) == doppelganger.SigningDisabled {
+				cr.doppelgangerHandler.MarkAsSafe(validator)
+			}
+
```
y0sher, line 427: not sure about it, but should we do it as well if the only duty that we have quorum on is sync committte? since we send sync committes even when we're in doppleganger state.

so we might want to move this downwards into the place where we know the role is attestaions.

@moshe-blox @MatusKysel @olegshmuelov wdyt?

protocol/v2/ssv/runner/proposer.go
```diff
@@ -189,9 +193,15 @@ func (r *ProposerRunner) ProcessConsensus(ctx context.Context, logger *zap.Logge
 		}
 	}
 
+	duty := r.BaseRunner.State.StartingDuty.(*spectypes.ValidatorDuty)
+	if r.doppelgangerHandler.ValidatorStatus(duty.ValidatorIndex) == doppelganger.SigningDisabled {
+		logger.Warn("Doppelganger check in progress, signing not permitted", fields.ValidatorIndex(duty.ValidatorIndex))
+		return nil
+	}
```
y0sher, line 200: why is this needed if you do it in `executeDuty` already?

y0sher (changes requested):
Good job! this looks good, mostly code-style comments and some questions.

doppelganger/doppelganger.go
olegshmuelov: Good point, but I’d prefer to keep it as uint64. The memory savings with int8 are negligible here, and switching might introduce unnecessary type conversions.

doppelganger/doppelganger.go
olegshmuelov: Keeping the return type as an interface ensures that future extensions or modifications interact with the intended abstraction. If we return the struct, we’d have to make `doppelgangerHandler` public, which increases the risk that someone might bypass the interface and directly use the struct, potentially leading to unintended dependencies or misuse of internal logic.

doppelganger/doppelganger.go
olegshmuelov: I see the point, but I think keeping it where it is makes sense. It ensures we first gather the full set before making any modifications, keeping the logic structured. Moving it down wouldn’t add much value in readability or performance.

doppelganger/doppelganger.go
olegshmuelov: **PermanentlyUnsafe** is meant to indicate that a validator was detected **as live** elsewhere and should not be trusted for signing. It’s a distinct state, not just pending further checks. `Unsafe` might be misleading since a validator can be unsafe only for a certain period (remaining epochs), whereas **PermanentlyUnsafe** clearly marks a validator as permanently unsafe to use. `PendingChecks` also wouldn’t fit, as this status isn't about awaiting verification—it specifically signals that the validator is live elsewhere and should not be used. Given this, I think **PermanentlyUnsafe** best represents its purpose.

doppelganger/doppelganger_state.go
olegshmuelov: https://github.com/ssvlabs/ssv/pull/2036#discussion_r1969601922

doppelganger/doppelganger.go
moshe-blox: for provider i agree we should keep it doppelganger.go because thats the main entry point & logic

for the rest of the files (state/mock) there's no reason to prefix all modern editors allow you to search for "doppelganger/state" and also show the parent directory in the results

image

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+type DoppelgangerProvider interface {
+	ValidatorStatus(validatorIndex phase0.ValidatorIndex) DoppelgangerStatus
+	StartMonitoring(ctx context.Context)
+	MarkAsSafe(validatorIndex phase0.ValidatorIndex)
+	RemoveValidatorState(validatorIndex phase0.ValidatorIndex)
+}
+
+// ValidatorProvider represents a provider of validator information.
+type ValidatorProvider interface {
+	SelfParticipatingValidators(epoch phase0.Epoch) []*types.SSVShare
```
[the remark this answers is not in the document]
y0sher: I guess its ok if this comment is in inside ValidatorStore and not here. And SelfParticipating means the validators our operator is managing that are active. (Participating = active, not liquidated)

protocol/v2/ssv/runner/committee.go
```diff
@@ -237,6 +241,12 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 		switch validatorDuty.Type {
 		case spectypes.BNRoleAttester:
 			validDuties++
+
+			if cr.doppelgangerHandler.ValidatorStatus(validatorDuty.ValidatorIndex) != doppelganger.SigningEnabled {
+				logger.Warn("Doppelganger check in progress, signing not permitted", fields.ValidatorIndex(validatorDuty.ValidatorIndex))
+				continue
+			}
```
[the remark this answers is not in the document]
y0sher: No, because sync committee isn't slashable hence doppleganger shouldn't affect doing sync committee duties.

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+
+	firstRun := true
+	ticker := ds.slotTickerProvider()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.Next():
+			currentSlot := ticker.Slot()
```
moshe-blox, line 161: slot ticker can skip slots forward or backward due to system clock, system sleep or high load

we should make sure to keep validators disabled under these scenarios

let's focus on slot skipping forward or triggering the same slot twice, as these are the most likely fault conditions

the simplest solution i can think of is to keep a `previousEpoch` above the loop and if it didn't go up exactly by 1 from the last iteration: we should reset all doppelganger states back to default (2 remaining epochs)

(checking `previousEpoch` should be done after the last epoch trigger)

wdyt?

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+			if (!firstRun && uint64(currentSlot)%slotsPerEpoch != slotsPerEpoch-1) || ds.startEpoch == currentEpoch {
+				continue
+			}
+
+			if firstRun {
+				ds.startEpoch = currentEpoch
+				firstRun = false
+			}
+
+			ds.checkLiveness(ctx, currentEpoch-1)
```
moshe-blox, line 182: IMO we should set a deadline for this ctx until the end of the slot with -200ms or more for some safety margin

protocol/v2/ssv/runner/committee.go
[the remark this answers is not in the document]
iurii-ssv: Would be nice to leave a clarifying comment in code.

doppelganger/doppelganger.go
olegshmuelov: **PermanentlyUnsafe** is meant to indicate that a validator was detected as **live** elsewhere and should not be trusted for signing. It’s a distinct state, not just pending further checks. `Unsafe` might be misleading since a validator can be unsafe only for a certain period (remaining epochs), whereas **PermanentlyUnsafe** clearly marks a validator as permanently unsafe to use. `PendingChecks` also wouldn’t fit, as this status isn't about awaiting verification—it specifically signals that the validator is live elsewhere and should not be used. Given this, I think **PermanentlyUnsafe** best represents its purpose.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -189,9 +193,15 @@ func (r *ProposerRunner) ProcessConsensus(ctx context.Context, logger *zap.Logge
 		}
 	}
 
+	duty := r.BaseRunner.State.StartingDuty.(*spectypes.ValidatorDuty)
+	if r.doppelgangerHandler.ValidatorStatus(duty.ValidatorIndex) == doppelganger.SigningDisabled {
```
moshe-blox, line 197: it's safer to check `!= SigningEnabled` instead because there's a 3rd variant (unknown) and also in case other variants are added to the enum in the future

although since there's only really 2 possible options (can sign / can't sign), i'd switch the state or at least the method to a boolean to prevent such mistakes by callers

protocol/v2/ssv/runner/committee.go
olegshmuelov: good catch!

beacon/goclient/validatorliveness.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func (gc *GoClient) ValidatorLiveness(ctx context.Context, epoch phase0.Epoch, validatorIndices []phase0.ValidatorIndex) ([]*apiv1.ValidatorLiveness, error) {
+	resp, err := gc.multiClient.ValidatorLiveness(ctx, &api.ValidatorLivenessOpts{
+		Epoch:   epoch,
+		Indices: validatorIndices,
+	})
+	if err != nil {
+		gc.log.Error(clResponseErrMsg,
+			zap.String("api", "ValidatorLiveness"),
```
oleg-ssvlabs, line 20: it would be nicer and spare some space, if the logger was enriched with the `api` field and reused.
Like this:

```Go
logger := gc.log.With("api", "ValidatorLiveness")
```

doppelganger/doppelganger.go
oleg-ssvlabs: I agree with @y0sher. It’s a common Go convention to avoid using package name prefixes in all exported types and primitives. `doppleganger.Provider` looks much better

protocol/v2/ssv/runner/committee.go
olegshmuelov: good catch!
https://github.com/ssvlabs/ssv/pull/2036/commits/7d978a71f61dcd7cd917e8b3d0d86578ff926181

doppelganger/doppelganger.go
oleg-ssvlabs: Do we actually need this interface?
There are two patterns: 
- package Foo exposes interface. Package Bar, that has dependency on package Foo, **imports the Foo interface** to use it for dependency injection.
- package Foo does not expose interface. Package Bar, that has dependency on package Foo, declares the Foo interface with only methods it uses, instead of importing the full `Foo` interface from the `Foo` package. 

In our case we have a mix of both patterns. Doppelganger package exports DoppelgangerProvider, yet the package that has Doppelganger as dependency(`eventhandler`) [declares](https://github.com/ssvlabs/ssv/pull/2036/files#diff-630aac148609b829f915b937c0dd7e59ed4ebf24c5f700677e16120996f12237R61) the Doppelganger interface instead of importing.

So that begs the question - why do we need this interface? For generating mocks only? If the mocks are generated, they presumably will be used by the packages that depend on `Doppelganger` package(`eventhandler`), but isn't it that the purpose of declaring the [interface](https://github.com/ssvlabs/ssv/pull/2036/files#diff-630aac148609b829f915b937c0dd7e59ed4ebf24c5f700677e16120996f12237R61) in this way aims to actually completely remove cross package references which often lead to circular dependencies?

doppelganger/doppelganger.go
oleg-ssvlabs: I agree with @y0sher . Closely related to the comment I wrote above. With Go implicit interfaces, there is no need for ctors to return interfaces. It's a pattern for other OOP languages with explicit interfaces

oleg-ssvlabs: also, since it's within `doppelganger` package, it could be renamed to `NewHandler`

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+type doppelgangerHandler struct {
+	mu                 sync.RWMutex
+	network            networkconfig.NetworkConfig
+	beaconNode         BeaconNode
+	validatorProvider  ValidatorProvider
+	slotTickerProvider slotticker.Provider
+	logger             *zap.Logger
+	doppelgangerState  map[phase0.ValidatorIndex]*doppelgangerState
```
oleg-ssvlabs, line 60: I see we use `mutex` for making this map thread-safe, potentially we could use [our own](https://github.com/ssvlabs/ssv/blob/main/utils/hashmap/hashmap.go) generic thread safe hashmap data structure instead

oleg-ssvlabs, line 60: If this is not yet implemented, it would be beneficial to ensure that the map does not grow indefinitely, possibly by setting a capacity limit or implementing a clearing mechanism

protocol/v2/ssv/runner/proposer.go
olegshmuelov: Good point! `executeDuty` is part of the pre-consensus phase, where we prevent signing the partial RANDAO signature if the validator is DG-unsafe. However, if we still reach ProcessConsensus, this check acts as an additional safeguard to ensure we don't sign the block in case the validator's status changed in the meantime.
Does this make sense?

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/ssvlabs/ssv/networkconfig"
+	"github.com/ssvlabs/ssv/operator/slotticker"
+	"github.com/ssvlabs/ssv/protocol/v2/types"
+)
+
+//go:generate mockgen -package=doppelganger -destination=./doppelganger_mock.go -source=./doppelganger.go
+
+// DefaultRemainingDetectionEpochs represents the initial number of epochs
+// a validator must pass without liveness detection before being considered safe to sign.
+const DefaultRemainingDetectionEpochs uint64 = 2
```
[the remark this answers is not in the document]
olegshmuelov: `Initial` is more precise since this value starts at 2 and decreases over time. `Default` might suggest it remains constant, which isn’t the case here. We use `DefaultRemainingDetectionEpochs` because it sets the starting point before the value dynamically decreases.

doppelganger/doppelganger.go
olegshmuelov: Maybe we can track `previousSlot` instead of `previousEpoch` (or both). If `currentSlot` jumps forward more than expected (e.g., more than 1 slot at a time), we could temporarily pause doppelganger checks until normal slot progression has resumed.

doppelganger/doppelganger.go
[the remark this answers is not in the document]
iurii-ssv: > Default might suggest it remains constant, which isn’t the case here

Right, actually it seems `DefaultRemainingDetectionEpochs ` -> `InitialRemainingDetectionEpochs` would make most sense ?

doppelganger/doppelganger.go
iurii-ssv: Right, I also noticed we use `mutex` mostly just to synchronize memory and not actually "block actions from running concurrently" (aka executing atomically) ?

If that's the case `hashmap` would indeed be slightly simpler way to do it

doppelganger/doppelganger.go
y0sher: https://duncanleung.com/go-idiom-accept-interfaces-return-types/

y0sher: btw you don't have to make `dopplegangerHandler` public

doppelganger/doppelganger.go
y0sher: I don't understand. this can be changed to safe, hence it's not permanent.

doppelganger/doppelganger.go
y0sher: if we decide to keep it uint64, I think it makes more sense it'll be using the `Epoch`, then we could use eth's `FarFutureEpoch` . I just don't like saving max int in mem many times.

doppelganger/doppelganger.go
iurii-ssv: > Could you clarify how we can return a concrete type without making the struct public? Or did you mean something else?

You could assign it to variable and compiler will figure out the type I think (without you having to name/reference that type explicitly),

or you could define an interface on the "using" side such that struct implements it and assign it to that interface (eg. if you need to keep it as a member-field on some other struct)

but I'd say it's fine to export and use **struct** itself, you mostly need interface(s) in cases when:
- you want to have multiple implementations to choose from
- or you want to mock it out for testing

doppelganger/doppelganger.go
iurii-ssv: > it specifically signals that the validator is live elsewhere and should not be used.

It's live somewhere else until it's no longer, right ? So validator in Ethereum is never really permanent (you can deregister it at any point)

**Permanent** is just a strong word that misleads in this case - **Unsafe** seems to describe the actual behavior more precisely but maybe there is a better term to use even ...

doppelganger/doppelganger.go
y0sher: It's allowed in go to return not exported types. it's actually a common practice where you want to allow only one way of instantiating the struct (so structName{} isn't available outside of the package and you must use an exported constructor func like New)

protocol/v2/ssv/runner/proposer.go
y0sher: Yeah if we're able to reach it even if we didn't execute the duty? As if there was pre-consensus from the rest of the committee on the randao?

doppelganger/doppelganger.go
olegshmuelov: > if we decide to keep it uint64, I think it makes more sense it'll be using the `Epoch`, then we could use eth's `FarFutureEpoch` . I just don't like saving max int in mem many times.

sound reasonable. I will use `FarFutureEpoch`

doppelganger/doppelganger.go
olegshmuelov: I've redefined some logic and removed this flag. Now using detectedAsLive() and markAsLive() to handle this case explicitly

doppelganger/doppelganger.go
olegshmuelov: hashmap is more useful for cases where entire values (not just internal fields) need atomic replacements. Since we are modifying struct fields rather than replacing the entire object, a mutex-protected map is the more efficient and safer choice for our use case.

doppelganger/doppelganger.go
olegshmuelov: Keeping the collection step at the beginning ensures that all necessary data is gathered upfront before modifications. While we don’t use it immediately when adding validators, having the full set available before any state changes keeps the logic structured and avoids mixing data collection with state updates. If you feel strongly about moving it, I'm happy to discuss further.

doppelganger/doppelganger.go
olegshmuelov: The map does not grow indefinitely because `updateDoppelgangerState` ensures that validators no longer present in the retrieved set are removed. This ensures that `validatorsState` map remains manageable, eliminating the need for extra limits or periodic cleanup.

doppelganger/state.go
```diff
@@ -0,0 +1,37 @@
...
+// requiresFurtherChecks returns true if the validator is *not* safe to sign yet.
+func (ds *doppelgangerState) requiresFurtherChecks() bool {
+	return ds.remainingEpochs > 0
+}
+
+// decreaseRemainingEpochs decreases remaining epochs.
+func (ds *doppelgangerState) decreaseRemainingEpochs() {
+	if ds.remainingEpochs > 0 {
+		ds.remainingEpochs--
+	}
```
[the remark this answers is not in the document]
olegshmuelov: A panic seems too aggressive. Instead, we can log a warning if `remainingEpochs` is already 0, avoiding crashes while still catching unexpected behavior:
```go
func (ds *doppelgangerState) decreaseRemainingEpochs() {
    if ds.remainingEpochs == 0 {
       logger.Warn("Attempted to decrease remaining epochs at 0")
       return
    }
    ds.remainingEpochs--
}
```
wdyt/?>/

doppelganger/doppelganger.go
y0sher: Even though this is not a critical decision, I would love some input from more team members just for the sake of the technical discussion or code conventions that we want to have in our project.
@iurii-ssv @MatusKysel @moshe-blox @oleg-ssvlabs @anatolie-ssv @kchojn @julienh-ssv

doppelganger/state.go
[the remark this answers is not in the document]
y0sher: I think Error makes more sense since this is critical and means we have a bug

doppelganger/design.md
y0sher, line 1: just a suggestion if you rename this file to README.md github will automatically present when you look at the folder.

doppelganger/doppelganger.go
iurii-ssv: I think "moving definitions to the place where they are used" (or "limiting the scope of usage as much is possible") is best for code readability/understandability in 95+% of the cases,

there are exception where it makes sense to group a bunch of things together for example - which would violate that convention

kchojn: Variables should be declared as locally as possible, imo the set should be created right before the validator removal loop

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,325 @@
...
+
+			// Detect if an unexpected gap in epochs occurred (e.g., due to system sleep or clock drift).
+			// If we detect a skipped epoch, we reset all doppelganger states to avoid unsafe signing.
+			if previousEpoch > 0 && currentEpoch != previousEpoch+1 {
+				h.logger.Warn("Epoch skipped unexpectedly, resetting all Doppelganger states",
+					zap.Uint64("previous_epoch", uint64(previousEpoch)),
+					zap.Uint64("current_epoch", uint64(currentEpoch)),
+				)
+
+				h.resetDoppelgangerStates()
```
y0sher, line 217: I'll add a small comment that will say this means our operator will probably skip at least a few slots (assuming it sees quorum next slot or something)

doppelganger/state.go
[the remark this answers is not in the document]
iurii-ssv: > Instead, we can log a warning

Logging as `Error` would be better I guess, the downside is - it's hard to notice unless you are looking for it

I suggested `panic` because it's really not that different from hitting some nil-pointer, and if we hit it - it's likely to happen during testing

doppelganger/doppelganger.go
oleg-ssvlabs: conventions aside, why do we iterate over `validatorIndices` collection two times? Potentially we could do all the things within one iteration and reduce time complexity from _O(2n)_ to _O(n)_ where _n_ is the number of items in `validatorIndices` collection.

olegshmuelov: > conventions aside, why do we iterate over `validatorIndices` collection two times([here](https://github.com/ssvlabs/ssv/pull/2036/files/59d5b30c5cd62653c9b9d3dca835c17d8d9a4f1e#diff-7519ed53af5a13cb16224c296695552722e2e918bda8685d9fd0d0371dc849beR114) and [here](https://github.com/ssvlabs/ssv/pull/2036/files/59d5b30c5cd62653c9b9d3dca835c17d8d9a4f1e#diff-7519ed53af5a13cb16224c296695552722e2e918bda8685d9fd0d0371dc849beR119))? Potentially we could do all the things within one iteration and reduce time complexity from _O(2n)_ to _O(n)_ where _n_ is the number of items in `validatorIndices` collection.

good catch :)

doppelganger/doppelganger.go
```diff
@@ -0,0 +1,265 @@
...
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	state, exists := ds.doppelgangerState[validatorIndex]
+	if !exists {
+		ds.logger.Warn("Validator not found in Doppelganger state", fields.ValidatorIndex(validatorIndex))
+		return
+	}
+
+	ds.logger.Debug("mark validator as doppelganger safe", fields.ValidatorIndex(validatorIndex))
```
anatolie-ssv: nitpicking: it seems there's a mix of capitalized/non-capitalized log messages, perhaps using one style would be easier "for the eye"

y0sher: A while ago I'm pretty sure we decided on having all logs lower case.

moshe-blox (approved):

moshe-blox (approved):

PR #2049, opened 2025-02-27
beacon-client: fix concurrency issues in tests

iurii-ssv:
I've noticed there are couple of issues:
- `t *testing.T` is not meant to be used concurrently:
```go
// T is a type passed to Test functions to manage test state and support formatted test logs.
//
// FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as
// the Parallel method, must be called only from the goroutine running the
// Test function.
//
// The other reporting methods, such as the variations of Log and Error,
type T struct {
```
- and there is also this error (see https://stackoverflow.com/questions/57828645/how-to-handle-superfluous-response-writeheader-call-in-order-to-return-500 for more details):
```
2025/02/27 10:15:16 http: superfluous response.WriteHeader call from github.com/ssvlabs/ssv/beacon/goclient.mockServer.func1 (goclient_test.go:264)
```

this PR should address both

### DIFF
--- a/beacon/goclient/tests/shared.go
+++ b/beacon/goclient/tests/shared.go
@@ -2,15 +2,14 @@ package tests
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"os"
-	"testing"
 	"time"
 
 	"github.com/ssvlabs/ssv-spec/types"
 	"github.com/ssvlabs/ssv/operator/slotticker"
-	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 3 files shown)


nkryuchkov (approved):

oleg-ssvlabs:
It's a really good change.  

The issue is that we merged some changes affecting `goclient_test.go`, so it conflicts.  
[This](https://github.com/ssvlabs/ssv/pull/2018/files#diff-7085246f03cf44eb61d6adc80a991ade6124c871b24e04af0711d89256430ca0) file was added because `MockServers` were created in multiple tests, so having a shared instance across tests seemed like a good idea.  

As you can see, this shared `MockServer` still has the same issue. So maybe you could rebase `stage` into your branch and fix the issue there?

beacon/goclient/tests/shared.go
```diff
@@ -24,34 +23,34 @@ func (m MockDataStore) AwaitOperatorID() types.OperatorID {
...
+			if err != nil {
+				panic(fmt.Sprintf("onRequestFn returned error: %v", err))
+			}
 		}
 
-		require.NoError(t, err)
-
 		w.Header().Set("Content-Type", "application/json")
 		if _, err := w.Write(resp); err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
```
oleg-ssvlabs, line 53: this was quite weird condition. 
The 500 codes should be explicitly mocked by the test, instead of doing this :)

oleg-ssvlabs (approved):

y0sher (approved):

beacon/goclient/tests/shared.go
```diff
@@ -24,34 +23,34 @@ func (m MockDataStore) AwaitOperatorID() types.OperatorID {
...
 type requestCallback = func(r *http.Request, resp json.RawMessage) (json.RawMessage, error)
 
-func MockServer(t *testing.T, onRequestFn requestCallback) *httptest.Server {
+func MockServer(onRequestFn requestCallback) *httptest.Server {
 	var mockResponses map[string]json.RawMessage
 	f, err := os.Open("./tests/mock-beacon-responses.json")
-	require.NoError(t, err)
-	require.NoError(t, json.NewDecoder(f).Decode(&mockResponses))
+	if err != nil {
+		panic(fmt.Sprintf("os.Open returned error: %v", err))
```
moshe-blox, line 30: why panics in test? what was wrong with require.NoError?

oleg-ssvlabs: `require.NoError` is dependent on `t *testing.T`, which caused concurrency issues

iurii-ssv: Yeah, **panic** isn't ideal but `t *testing.T` cannot be used concurrently - so I don't really see any viable way to use `require.NoError` in cases like that (we can try to propagate error to the original go-routine that's running current test so it can call `require.NoError` ... but that's not always possible and not worth bothering with I think)

moshe-blox: @oleg-ssvlabs makes sense, thx

PR #2007, opened 2025-01-29
feat: support of electra fork

MatusKysel:
https://hackmd.io/p2v3-l1GTjmk3nVikKxzcQ?view

PR #2051, opened 2025-02-28
Electra sign aggregation

nkryuchkov:
Since electra fork, `AggregateAndProof` can be of two types: `*phase0.AggregateAndProof` and `*electra.AggregateAndProof`. #2007 doesn't add an ability to sign `*electra.AggregateAndProof`, so the ssv-nodes cannot sign aggregate and proof and have `could not sign beacon object: could not cast obj to AggregateAndProof` in their logs

Depends on https://github.com/ssvlabs/eth2-key-manager/pull/112


go.mod
```diff
@@ -252,3 +252,5 @@ require (
 replace github.com/google/flatbuffers => github.com/google/flatbuffers v1.11.0
 
 replace github.com/dgraph-io/ristretto => github.com/dgraph-io/ristretto v0.1.1-0.20211108053508-297c39e6640f
+
+replace github.com/ssvlabs/eth2-key-manager => github.com/ssvlabs/eth2-key-manager v1.5.1-0.20250228021238-46519df1ff6e // https://github.com/ssvlabs/eth2-key-manager/pull/112
```
anatolie-ssv, line 256: minor: is there much value in keeping this comment here? the commit message will be pointing back to this PR number anyway and we can save this reference in the description.

anatolie-ssv (approved):

kchojn (approved):

go.mod
nkryuchkov: @anatolie-ssv I just don't like leaving any `replace` without comment because it's unclear why it's needed. A link allows one to follow it and check if it is merged and `replace` can be removed. But now, since the other PR is merged, we can remove it, of course

Issue #1855, opened 2024-11-18
p2p: races (discovered with CI unit-test) 


iurii-ssv:
Noticed CI pipeline failing - https://github.com/ssvlabs/ssv/actions/runs/11890850478/job/33130362086 - there are some races related to p2p & networking (upon cursory overview the root cause seems to be in our implementation itself, and not the test code, but obviously requires further investigation),

the relevant part describing those races:
```
==================
WARNING: DATA RACE
Read at 0x00c01772aa18 by goroutine 2247:
  github.com/ssvlabs/ssv/network/p2p.(*p2pNetwork).connectionsAtLimit()
      /home/runner/work/ssv/ssv/network/p2p/p2p_setup.go:330 +0x9e
  github.com/ssvlabs/ssv/network/p2p.(*p2pNetwork).connectionsAtLimit-fm()
      <autogenerated>:1 +0x1f
  github.com/ssvlabs/ssv/network/peers/connections.(*connGater).InterceptAccept()
      /home/runner/work/ssv/ssv/network/peers/connections/conn_gater.go:85 +0x1d9
  github.com/libp2p/go-libp2p/p2p/net/upgrader.(*listener).handleIncoming()
      /home/runner/go/pkg/mod/github.com/libp2p/go-libp2p@v0.36.3/p2p/net/upgrader/listener.go:88 +0xbf7
  github.com/libp2p/go-libp2p/p2p/net/upgrader.(*upgrader).UpgradeListener.gowrap1()
      /home/runner/go/pkg/mod/github.com/libp2p/go-libp2p@v0.36.3/p2p/net/upgrader/upgrader.go:119 +0x33

Previous write at 0x00c01772aa18 by goroutine 1897:
  github.com/ssvlabs/ssv/network/p2p.(*p2pNetwork).setupPeerServices()
      /home/runner/work/ssv/ssv/network/p2p/p2p_setup.go:190 +0x10a4
  github.com/ssvlabs/ssv/network/p2p.(*p2pNetwork).SetupServices()
      /home/runner/work/ssv/ssv/network/p2p/p2p_setup.go:155 +0xc4
  github.com/ssvlabs/ssv/network/p2p.(*p2pNetwork).Setup()
      /home/runner/work/ssv/ssv/network/p2p/p2p_setup.go:74 +0x45e
  github.com/ssvlabs/ssv/network/p2p.(*LocalNet).NewTestP2pNetwork()
[...]

Issue #1732, opened 2024-09-11
Data race in TestP2pNetwork_MessageValidation


AKorpusenko:
- https://github.com/ssvlabs/ssv/actions/runs/10813573125/job/29998518929
- https://github.com/ssvlabs/ssv/actions/runs/10831727491/job/30054372420

nkryuchkov:
https://github.com/ssvlabs/ssv/actions/runs/11111346936/job/30871052939

nkryuchkov:
https://github.com/ssvlabs/ssv/actions/runs/11466922731/job/31916750156?pr=1813

PR #2053, opened 2025-02-28
fix race condition on p2p setup

MatusKysel:
We are passing peerIndex - `idx` to the connection gater before we initialise it, this was causing race as `idx` variable could be read from connGater and also set in parallel with this flag it should work as expected 

fixes https://github.com/ssvlabs/ssv/issues/1855 & https://github.com/ssvlabs/ssv/issues/1732

### DIFF
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -75,7 +75,7 @@ type p2pNetwork struct {
 	host         host.Host
 	streamCtrl   streams.StreamController
 	idx          peers.Index
-	isIdxSet     int32
+	isIdxSet     atomic.Bool
 	disc         discovery.Service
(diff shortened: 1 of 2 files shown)


network/p2p/p2p.go
```diff
@@ -75,6 +75,7 @@ type p2pNetwork struct {
 	host         host.Host
 	streamCtrl   streams.StreamController
 	idx          peers.Index
+	isIdxSet     int32
```
nkryuchkov, line 78: Can we use `atomic.Bool`?

nkryuchkov (approved):

oleg-ssvlabs (approved):

PR #2055, opened 2025-02-28
Fix: Add a command line flag to allow signing slashable attestations for Holesky

JKincorperated:
Node operators need to delete their slashing DB (on Holesky), this is currently impossible in the SSV client. This adds a config entry that can disable attestation slashing protections. It also adds a rather large warning against doing this.

See: https://github.com/ethereum/pm/issues/1337

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -94,7 +94,9 @@ type config struct {
 	SSVAPIPort                   int                              `yaml:"SSVAPIPort" env:"SSV_API_PORT" env-description:"Port to listen on for the SSV API."`
 	LocalEventsPath              string                           `yaml:"LocalEventsPath" env:"EVENTS_PATH" env-description:"path to local events"`
 	EnableDoppelgangerProtection bool                             `yaml:"EnableDoppelgangerProtection" env:"ENABLE_DOPPELGANGER_PROTECTION" env-description:"Flag to enable Doppelganger protection for validators."`
-	AllowSigningSlashable        bool                             `yaml:"AllowSigningSlashableAttestations" env:"ALLOW_SIGNING_SLASHABLE_ATTESTATIONS" env-description:"allow the node to sign slashable attestations"`
+
+	// This may be needed in cases where finality has been lost due to failure of a large portion of the network such as the Holesky Pectra fork.
+	AllowSigningSlashable bool `yaml:"AllowSigningSlashableAttestations" env:"ALLOW_SIGNING_SLASHABLE_ATTESTATIONS" env-description:"allow the node to sign slashable attestations"`
 }
 
 var cfg config
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -94,6 +94,7 @@ type config struct {
 	SSVAPIPort                   int                              `yaml:"SSVAPIPort" env:"SSV_API_PORT" env-description:"Port to listen on for the SSV API."`
 	LocalEventsPath              string                           `yaml:"LocalEventsPath" env:"EVENTS_PATH" env-description:"path to local events"`
 	EnableDoppelgangerProtection bool                             `yaml:"EnableDoppelgangerProtection" env:"ENABLE_DOPPELGANGER_PROTECTION" env-description:"Flag to enable Doppelganger protection for validators."`
+	AllowSigningSlashable        bool                             `yaml:"AllowSigningSlashableAttestations" env:"CAN_SIGN_SLASHABLE_ATTESTATIONS" env-description:"allow the node to sign slashable attestations"`
 }
 
 var cfg config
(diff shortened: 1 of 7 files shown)


cli/operator/node.go
```diff
@@ -94,6 +94,7 @@ type config struct {
 	SSVAPIPort                   int                              `yaml:"SSVAPIPort" env:"SSV_API_PORT" env-description:"Port to listen on for the SSV API."`
 	LocalEventsPath              string                           `yaml:"LocalEventsPath" env:"EVENTS_PATH" env-description:"path to local events"`
 	EnableDoppelgangerProtection bool                             `yaml:"EnableDoppelgangerProtection" env:"ENABLE_DOPPELGANGER_PROTECTION" env-description:"Flag to enable Doppelganger protection for validators."`
+	AllowSigningSlashable        bool                             `yaml:"AllowSigningSlashableAttestations" env:"CAN_SIGN_SLASHABLE_ATTESTATIONS" env-description:"allow the node to sign slashable attestations"`
```
iurii-ssv, line 97: I think **env** name should match **yaml** name (either allow or can in both)

cli/operator/node.go
```diff
@@ -188,7 +189,14 @@ var StartNodeCmd = &cobra.Command{
...
 
-		keyManager, err := ekm.NewETHKeyManagerSigner(logger, db, networkConfig, ekmHashedKey)
+		if cfg.AllowSigningSlashable {
+			logger.Warn("    --------------------- ! WARNING ! ---------------------")
+			logger.Warn("This node is configured to allow signing slashable attesations.")
+			logger.Warn("          DO NOT RUN THIS IN A PRODUCTION ENVIRONMENT")
+			logger.Warn("    --------------------- ! WARNING ! ---------------------")
+		}
+
+		keyManager, err := ekm.NewETHKeyManagerSigner(logger, db, networkConfig, ekmHashedKey, cfg.AllowSigningSlashable)
```
iurii-ssv, line 199: Maybe we can check if we are running on **mainnet** or not and terminate process here if we are indeed on mainnet

ekm/eth_key_manager_signer.go
```diff
@@ -246,6 +248,9 @@ func (km *ethKeyManagerSigner) IsAttestationSlashable(pk spectypes.ShareValidato
 		if err != nil {
 			return err
 		}
+		if km.canSignSlashable {
+			return nil
+		}
```
iurii-ssv, line 253: This could be the very first code-snippet in `IsAttestationSlashable` func

iurii-ssv (commented):
Could you clarify why exactly we need to support this feature as opposed to manually just deleting and re-syncing SSV node DB (I haven't dug too deep into what happened with Holesky) ?

The code could use a comment explaining the reasoning for defining such a configuration setting.

JKincorperated:
> Could you clarify why exactly we need to support this feature as opposed to manually just deleting and re-syncing SSV node DB (I haven't dug too deep into what happened with Holesky) ?

I found that deleteing the DB still prevented validators being slashed (even though it resynced the contract)

iurii-ssv:
> I found that deleteing the DB still prevented validators being slashed (even though it resynced the contract)

That doesn't really answer my question which is (to phrase it differently) - why would we ever want to disable slashing protection ?

JKincorperated:
> That doesn't really answer my question which is (to phrase it differently) - why would we ever want to disable slashing protection ?

Specifically on Holesky during the Pectra fork, several execution clients were misconfigured meaning that a large portion of the network attested to the wrong chain. Slashing protection prevented nodes from attesting to the right chain, preventing finality, so the AllCoreDevs team decided to [attempt mass-slashings](https://github.com/ethereum/pm/blob/master/Pectra/holesky-postmortem.md) at slot `3737760`, the only way to do this was to disable slashing protection.

nkryuchkov:
Hi @JKincorperated, thank you for the contribution! I saw you opened the PR before slot `3737760` but unfortunately, it wasn't reviewed before the event. Now, when it's in the past and the Holesky network seems to get up, as I understand, the change isn't necessary anymore, right?

JKincorperated:
> Hi @JKincorperated, thank you for the contribution! I saw you opened the PR before slot `3737760` but unfortunately, it wasn't reviewed before the event. Now, when it's in the past and the Holesky network seems to get up, as I understand, the change isn't necessary anymore, right?

The Holesky network, while now finalizing, is still unstable (only ~4% over 66%) increasing overall participating validators would improve network health. Also, many SSV nodes are not attesting at all to avoid slashing, which will eventually cause thousands of "dead" validators to be registered to operators.

JKincorperated:
I'll close this now, EthPandaOps is canning Holesky for Hoodi, so this flag would not be applicable anymore.

PR #2061, opened 2025-03-05
feat(sync-committee) - Execute SyncCommittee duties for Validators in Exit state

oleg-ssvlabs:
**Description**  
This PR continues the work from **#2001**.  

According to the Beacon spec, validators are assigned to perform sync committee duties **one period** (**256 epochs**) in advance.  

This PR modifies the filtering conditions that determine which validators are eligible for sync committee duties. It now allows validators that have exited (**ValidatorStateExitedUnslashed, ValidatorStateExitedSlashed, ValidatorStateWithdrawalPossible, ValidatorStateWithdrawalDone**) if their **Exit Epoch** falls within the **current or previous period**

Duty fetching and execution flow(components affected by this PR):
image

Duty execution flow:
image


olegshmuelov (changes requested):
left a comment

olegshmuelov (commented):
**Potential issue:** exited validators not included in committee duties
In the `buildCommitteeDuties` function, we use:
```go
```
which ultimately calls:
```go
func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
    return !s.Liquidated && s.IsAttesting(epoch)
}
```
However, `IsParticipating(epoch)` only checks for active attesting validators, meaning exited validators who are still eligible for Sync Committee duties will never be included in the committee duty map.

To properly build Sync Committee duties in committee map, we must base it on `IsSyncCommitteeEligible()`, not just active attestation participation.

should be something like that
```go
func (c *validatorStore) SelfEligibleValidators(epoch phase0.Epoch) []*types.SSVShare {
    if c.operatorID == nil {
        return nil
    }
    validators := c.OperatorValidators(c.operatorID())
    var eligible []*types.SSVShare
    for _, validator := range validators {
        if validator.IsParticipating(epoch) || validator.IsSyncCommitteeEligible(epoch) {
            eligible = append(eligible, validator)
        }
    }
    return eligible
}
```

Then, update `buildCommitteeDuties` to correctly include Sync Committee-eligible validators:
```go
vs := h.validatorProvider.SelfEligibleValidators(epoch)
```

protocol/v2/types/ssvshare.go
```diff
@@ -101,6 +101,23 @@ func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
...
+		return true
+	}
+
+	if s.Status.IsExited() || s.Status == eth2apiv1.ValidatorStateWithdrawalPossible || s.Status == eth2apiv1.ValidatorStateActiveSlashed {
+		// if validator exited within (currentPeriod - 1) period, then it is eligible
+		// because Sync committees are assigned EPOCHS_PER_SYNC_COMMITTEE_PERIOD(256) in advance
+		epochsPerSyncCommitteePeriod := phase0.Epoch(256)
+		if epoch-s.ExitEpoch <= epochsPerSyncCommitteePeriod {
+			return true
+		}
```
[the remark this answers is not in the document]
oleg-ssvlabs: I see what you mean.  
The way it works now is that [first](https://github.com/ssvlabs/ssv/pull/2061/files#diff-812236a27621c8a478adb368a8ae554c4f52edc6beb70341811119aeae9ed644R192) epoch of the period is passed to this method.  
Although, the method's semantics could certainly be better - either rename its `epoch` parameter to `firstEpochOfPeriod` or modify the implementation so it can calculate eligibility for any given epoch, rather than requiring the first epoch of the period. 
I like your proposed fix. I will update the implementation.

It seems`EstimatedSyncCommitteePeriodAtEpoch` is not accessible within `SSVShare` object, but it doesn't look like we need it here. Do I understand it right?

protocol/v2/types/ssvshare.go
```diff
@@ -101,6 +101,23 @@ func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
...
 
+func (s *SSVShare) IsSyncCommitteeEligible(epoch phase0.Epoch) bool {
+	if s.IsParticipating(epoch) {
+		return true
+	}
+
+	if s.Status.IsExited() || s.Status == eth2apiv1.ValidatorStateWithdrawalPossible || s.Status == eth2apiv1.ValidatorStateActiveSlashed {
+		// if validator exited within Current Period OR Current Period - 1, then it is eligible
+		// because Sync committees are assigned EPOCHS_PER_SYNC_COMMITTEE_PERIOD(256) in advance
+		epochsPerSyncCommitteePeriod := phase0.Epoch(256)
```
olegshmuelov, line 112: Using `epochsPerSyncCommitteePeriod := phase0.Epoch(256)` has potential issues:

1. Future Local Testnet – If we introduce a local testnet with a different sync committee period, this hardcoding could lead to incorrect behavior.  
2. Prevents Mocking in Tests – Hardcoding `256` makes it harder to mock different period lengths for testing.  
3. Not Future-Proof – Ethereum specs may evolve, so keeping this configurable avoids unnecessary refactoring.

You can use the approach suggested [here](https://github.com/ssvlabs/ssv/pull/2061#discussion_r1989115747) or [here (preferable)](https://github.com/ssvlabs/ssv/pull/2061#discussion_r1989145191).

oleg-ssvlabs: Yes, it certainly isn’t the best way to do that.  
I think the main issue is that we don’t have a central place in the codebase where we keep these constants and calculations.  
Currently, they are part of the [Network](https://github.com/ssvlabs/ssv/blob/main/protocol/v2/blockchain/beacon/network.go#L132) object.  
It would be better if these methods were stateless functions (without a receiver), so they could be easily accessed from anywhere in the codebase, without Network object instantiation if possible
Thank you for suggestions, I will look into it

oleg-ssvlabs: Updated

operator/duties/sync_committee.go
```diff
@@ -180,26 +181,29 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
+		return s.IsSyncCommitteeEligible(firstEpochOfPeriod)
+	})
+
+	if len(eligibleIndices) == 0 {
 		h.logger.Debug("no active validators for period", fields.Epoch(currentEpoch), zap.Uint64("period", period))
 		return nil
 	}
 
-	inCommitteeIndices := indicesFromShares(h.validatorProvider.SelfParticipatingValidators(firstEpoch))
+	inCommitteeIndices := indicesFromShares(h.validatorProvider.SelfParticipatingValidators(firstEpochOfPeriod))
```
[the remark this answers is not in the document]
oleg-ssvlabs: Good point. I will look into that

operator/duties/sync_committee.go
```diff
@@ -180,37 +181,40 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
+	eligibleIndices := h.validatorController.FilterIndices(waitForInitial, func(s *types.SSVShare) bool {
+		return s.IsSyncCommitteeEligible(currentEpoch, h.network.Beacon.EstimatedSyncCommitteePeriodAtEpoch)
+	})
+
+	if len(eligibleIndices) == 0 {
 		h.logger.Debug("no active validators for period", fields.Epoch(currentEpoch), zap.Uint64("period", period))
 		return nil
 	}
 
-	inCommitteeIndices := indicesFromShares(h.validatorProvider.SelfParticipatingValidators(firstEpoch))
```
oleg-ssvlabs, line 196: The current implementation has slightly better performance compared to the previous one.  
Previously, three iterations were performed in total:  
1. [First iteration](https://github.com/ssvlabs/ssv/blob/main/registry/storage/validatorstore.go#L195) over all operator validators.  
2. [Second iteration](https://github.com/ssvlabs/ssv/blob/main/operator/duties/scheduler.go#L439) over participating operator validators.  
3. [Third iteration](https://github.com/ssvlabs/ssv/blob/main/operator/duties/sync_committee.go#L198) over participating operator validators.  

The new implementation now performs only a [single iteration](https://github.com/ssvlabs/ssv/pull/2061/files#diff-812236a27621c8a478adb368a8ae554c4f52edc6beb70341811119aeae9ed644R207) over all operator validators.

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
oleg-ssvlabs: > Alternatively we could also pass EstimatedSyncCommitteePeriodAtEpoch func directly to IsSyncCommitteeEligible, allowing the function to determine the exit period dynamically

Implemented this option

operator/duties/sync_committee.go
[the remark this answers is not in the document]
oleg-ssvlabs: I implemented it in a slightly different way.  
There is a method [`SelfValidators`](https://github.com/ssvlabs/ssv/pull/2061/files#diff-812236a27621c8a478adb368a8ae554c4f52edc6beb70341811119aeae9ed644R205) that allows fetching all Operator's validators without any filters.  

I believe filters aren't really needed here because `eligibleIndices` is fetched with all Sync Committee filters already applied, and `eligibleIndices` is a subset of the operator's validators.  

Wdyt? Does that make sense?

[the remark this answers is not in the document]
olegshmuelov: Yeah, looks good.
We still have to apply this change to original [buildCommitteeDuties](https://github.com/ssvlabs/ssv/pull/2061#pullrequestreview-2673592819) issue as well

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
olegshmuelov: I saw that you're passing the epoch and calculating the period each time.  
Instead, we can calculate it once and pass the period directly, as suggested earlier:
```go
func (s *SSVShare) IsSyncCommitteeEligible(currentPeriod uint64, periodFromEpoch func(epoch phase0.Epoch) uint64) bool {
```

iurii-ssv (commented):
Minor comments

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
oleg-ssvlabs: > I saw that you're passing the epoch and calculating the period each time. Instead, we can calculate it once and pass the period directly, as suggested earlier:
> 
> ```go
> func (s *SSVShare) IsSyncCommitteeEligible(currentPeriod uint64, periodFromEpoch func(epoch phase0.Epoch) uint64) bool {
> ```

The epoch is used in [this](https://github.com/ssvlabs/ssv/blob/post-exit-sync-committee-duty/protocol/v2/types/ssvshare.go#L109) condition, so it is needed

operator/duties/sync_committee.go
```diff
@@ -180,37 +181,40 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
 func (h *SyncCommitteeHandler) fetchAndProcessDuties(ctx context.Context, period uint64, waitForInitial bool) error {
 	start := time.Now()
-	firstEpoch := h.network.Beacon.FirstEpochOfSyncPeriod(period)
+	firstEpochOfPeriod := h.network.Beacon.FirstEpochOfSyncPeriod(period)
 	currentEpoch := h.network.Beacon.EstimatedCurrentEpoch()
-	if firstEpoch < currentEpoch {
-		firstEpoch = currentEpoch
+	if firstEpochOfPeriod < currentEpoch {
+		firstEpochOfPeriod = currentEpoch
 	}
```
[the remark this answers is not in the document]
oleg-ssvlabs: Good point.
I made some minor refactor to clarify how target epoch is set.

operator/duties/sync_committee.go
```diff
@@ -180,37 +181,40 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
 	lastEpoch := h.network.Beacon.FirstEpochOfSyncPeriod(period+1) - 1
 
-	allActiveIndices := h.validatorController.AllActiveIndices(firstEpoch, waitForInitial)
-	if len(allActiveIndices) == 0 {
+	eligibleIndices := h.validatorController.FilterIndices(waitForInitial, func(s *types.SSVShare) bool {
+		return s.IsSyncCommitteeEligible(currentEpoch, h.network.Beacon.EstimatedSyncCommitteePeriodAtEpoch)
+	})
```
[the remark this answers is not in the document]
oleg-ssvlabs: It's a good question

I think it does make a difference, considering that the period may be a [future period](https://github.com/ssvlabs/ssv/blob/post-exit-sync-committee-duty/operator/duties/sync_committee.go#L157). If `currentEpoch` is passed and sync committees are requested for a future period, it sets validators that exited in `requestedPeriod - 2` period as eligible because we passed the current epoch instead of the first epoch of the next period(should be eligible if exited within `requestedPeriod -1`).  

There is one condition that may seem slightly incorrect - checking if a validator is [active](https://github.com/ssvlabs/ssv/blob/post-exit-sync-committee-duty/operator/duties/sync_committee.go#L157) while providing a future epoch. While it may not be semantically ideal, it still works because it results in the check:  

```go
s.ActivationEpoch <= epoch // where epoch will be a future epoch
```

So I will update and pass `firstEpochOfPeriod` to `IsSyncCommitteeEligible` method instead of `currentPeriod`.
Wdyt?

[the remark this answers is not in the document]
iurii-ssv: Tbh I lack some context to be able to reason about all the nuances of this code ... maybe @olegshmuelov could advise which epoch should be used here ?

My argument for using `firstEpochOfPeriod` (which [should actually be called `targetEpoch`](https://github.com/ssvlabs/ssv/pull/2061#discussion_r1999072464)) is based on purely looking at `fetchAndProcessDuties` func in isolation - it accepts `period` so we should probably use the epoch derived from that period ... I'm not sure what role `currentEpoch` plays except for the following guess

> my understanding is it's a sort of fallback in case period supplied is too old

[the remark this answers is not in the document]
oleg-ssvlabs: The condition is:
- If period passed is current period - target epoch should be current epoch
- If period passed is future period - target epoch should be first epoch of the future epoch

I made some refactor

operator/duties/committee.go
```diff
@@ -85,7 +85,7 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
 func (h *CommitteeHandler) buildCommitteeDuties(attDuties []*eth2apiv1.AttesterDuty, syncDuties []*eth2apiv1.SyncCommitteeDuty, epoch phase0.Epoch, slot phase0.Slot) committeeDutiesMap {
 	// NOTE: Instead of getting validators using duties one by one, we are getting all validators for the slot at once.
 	// This approach reduces contention and improves performance, as multiple individual calls would be slower.
-	vs := h.validatorProvider.SelfParticipatingValidators(epoch)
+	vs := h.validatorProvider.SelfValidators()
```
iurii-ssv, line 88: That was a bug I guess ?

operator/duties/sync_committee.go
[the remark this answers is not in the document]
oleg-ssvlabs: updated to call `SelfValidators` method instead of `SelfParticipatingValidators` in `buildCommitteeDuties` method.

operator/duties/committee.go
moshe-blox: @oleg-ssvlabs wouldn't this start attester duties for liquidated and non-participating validators?

if our approach is to keep attestation flow untouched, then why not add `IsParticipating` to the attDuties loop and `IsSyncCommitteeEligible` to the syncDuties loop, instead of moving filtering in `CommitteeRunner`?

oleg-ssvlabs: Yes, you are right @moshe-blox.

I reviewed the entire fetching and execution flow (attached diagram). It seems we apply the same filters during both fetching and execution. I assume the reason for that is that execution may take place with a delay, hence the validators may be in a different state than they were during the fetch.

I made some refactor(please review. It is slightly more verbose. I was aiming to make it more clear) and updated the condition to check if the validator exists in `SelfValidators` for Sync Committee duties and in `SelfParticipatingValidators` for attestation duties. Also I removed [this](https://github.com/ssvlabs/ssv/pull/2061/files#diff-51ce5aff6d9f47886f80a933d8dbbbd1c3bf135fdd6c7e38f78ba0c52c842b25L116) nil check, because it looks to me that it cannot be `nil`

operator/duties/sync_committee.go
```diff
@@ -178,46 +179,54 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
 	h.dutiesExecutor.ExecuteDuties(ctx, h.logger, toExecute)
 }
 
+// period here might be next period
```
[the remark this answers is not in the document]
oleg-ssvlabs: Added a comment

olegshmuelov:
Quick note: once a PR is opened, it's best to avoid force-pushing unless absolutely necessary.
It makes it harder to track changes incrementally and can cause sync issues for reviewers (e.g., with git pull requiring conflict resolution or rebase configs).

If needed, you can squash commits right before merging — that keeps the review history clear while still keeping the final history clean. 🙏

protocol/v2/types/ssvshare.go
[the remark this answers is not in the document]
olegshmuelov: Makes sense to keep `epoch` for `IsAttesting`, but we can still avoid recalculating the period by passing both `epoch` and the precomputed `currentPeriod`:
```go
func (s *SSVShare) IsSyncCommitteeEligible(epoch phase0.Epoch, currentPeriod uint64, periodFromEpoch func(phase0.Epoch) uint64) bool
```

That way we don’t recompute `periodFromEpoch(epoch)` internally and the logic stays clearer. What do you think?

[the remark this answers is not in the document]
olegshmuelov: Makes sense to keep epoch for `IsAttesting`, but we can still avoid recalculating the period by passing both epoch and the precomputed `currentPeriod`:
```go
func (s *SSVShare) IsSyncCommitteeEligible(epoch phase0.Epoch, currentPeriod uint64, periodFromEpoch func(phase0.Epoch) uint64) bool
```
That way we don’t recompute `periodFromEpoch(epoch)` internally and the logic stays clearer. What do you think?

operator/duties/committee.go
```diff
@@ -85,59 +84,81 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
-	vs := h.validatorProvider.SelfParticipatingValidators(epoch)
-	validatorCommitteeMap := make(validatorCommitteeDutyMap)
-	committeeMap := make(committeeDutiesMap)
-	for _, v := range vs {
-		validatorCommitteeMap[v.ValidatorIndex] = &committeeDuty{
-			id:          v.CommitteeID(),
-			operatorIDs: v.OperatorIDs(),
-		}
-	}
+	selfValidators := h.validatorProvider.SelfValidators()
```
y0sher, line 87: wouldn't this bring also liquidated validators? we want exited but not but liquidated

y0sher: you could just use `SelfParticipatingValidators`.

y0sher, line 87: we do want to keep doing it for exited validators but I don't think we want for liquidated validators.

operator/duties/sync_committee.go
```diff
@@ -178,46 +179,56 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
-	lastEpoch := h.network.Beacon.FirstEpochOfSyncPeriod(period+1) - 1
 
-	allActiveIndices := h.validatorController.AllActiveIndices(firstEpoch, waitForInitial)
-	if len(allActiveIndices) == 0 {
+	eligibleIndices := h.validatorController.FilterIndices(waitForInitial, func(s *types.SSVShare) bool {
+		return s.IsSyncCommitteeEligible(targetEpoch, h.network.Beacon.EstimatedSyncCommitteePeriodAtEpoch)
+	})
+
+	if len(eligibleIndices) == 0 {
 		h.logger.Debug("no active validators for period", fields.Epoch(currentEpoch), zap.Uint64("period", period))
```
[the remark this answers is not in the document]
oleg-ssvlabs: updated

operator/duties/committee.go
oleg-ssvlabs: It's a good point.  
Looking into the past implementation, it seems the intention was for the Fetcher (the one that saves duties to the store) and the Executor (the one that fetches from the store and executes them) to apply the exact same filters.  
So I updated the code so that `committee.go` applies the same filters for both attestation and sync committee duties

oleg-ssvlabs:
> This PR would be a good place to fix #2132 as well — it’s in the same code path. We should fetch sync committee duties only for own subnet validators (for message validation) and submit subscriptions only for operator validators.

It's good point, although I would really appreciate if we could implement the fix for #2132 in a separate PR in order not to further extend the scope of this PR, so it can be more easily tested and reviewed

protocol/v2/types/ssvshare.go
```diff
@@ -101,6 +101,26 @@ func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
...
+	}
+
+	if s.IsAttesting(epoch) {
+		return true
+	}
+
+	if s.Status.IsExited() || s.Status == eth2apiv1.ValidatorStateWithdrawalPossible || s.Status == eth2apiv1.ValidatorStateActiveSlashed {
+		// if validator exited within Current Period OR Current Period - 1, then it is eligible
+		// because Sync committees are assigned EPOCHS_PER_SYNC_COMMITTEE_PERIOD in advance
+		if periodFromEpoch(epoch)-periodFromEpoch(s.ExitEpoch) <= 1 {
```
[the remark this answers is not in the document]
oleg-ssvlabs: appreciate. Fixed

oleg-ssvlabs:
> Quick note: once a PR is opened, it's best to avoid force-pushing unless absolutely necessary. It makes it harder to track changes incrementally and can cause sync issues for reviewers (e.g., with git pull requiring conflict resolution or rebase configs).
> 
> If needed, you can squash commits right before merging — that keeps the review history clear while still keeping the final history clean. 🙏

Sure. I won't force push anymore. Just an old habit :)

operator/duties/sync_committee.go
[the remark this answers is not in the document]
oleg-ssvlabs: I added some comments [here](https://github.com/ssvlabs/ssv/pull/2061/files#diff-812236a27621c8a478adb368a8ae554c4f52edc6beb70341811119aeae9ed644R182)

operator/duties/sync_committee.go
```diff
@@ -178,46 +179,56 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
 	currentEpoch := h.network.Beacon.EstimatedCurrentEpoch()
-	if firstEpoch < currentEpoch {
-		firstEpoch = currentEpoch
+	firstEpochOfPeriod := h.network.Beacon.FirstEpochOfSyncPeriod(period)
+
+	targetEpoch := currentEpoch
+	isFuturePeriod := firstEpochOfPeriod > currentEpoch
+	if isFuturePeriod {
+		targetEpoch = firstEpochOfPeriod
 	}
```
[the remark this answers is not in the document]
oleg-ssvlabs: [Updated](https://github.com/ssvlabs/ssv/pull/2061/commits/b267cfdec38cdd0eafc150f1f80ab9bba7fae74d) @olegshmuelov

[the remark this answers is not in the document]
olegshmuelov: Thanks! Looks good

operator/duties/attester.go
```diff
@@ -182,14 +183,21 @@ func (h *AttesterHandler) processExecution(ctx context.Context, epoch phase0.Epo
 
 func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase0.Epoch, slot phase0.Slot) error {
 	start := time.Now()
-	indices := indicesFromShares(h.validatorProvider.SelfParticipatingValidators(epoch))
 
-	if len(indices) == 0 {
+	var selfAttestingShares []*types.SSVShare
```
iurii-ssv, line 187: nit: maybe better to call it `targetShares` (and similar for `selfAttestingIndices` I guess) since `selfAttestingShares` is not super precise (it would kind of imply it "includes" liquidated ones) and shorter names are often easier to read

Similar could also be applied to the code around `isAttestingFilter`

operator/duties/attester_test.go
```diff
@@ -34,28 +34,34 @@ func setupAttesterDutiesMock(
...
 
-		return shares
+			for index := range uniqueIndices {
+				attestingShare := &types.SSVShare{
+					Share: spectypes.Share{
+						ValidatorIndex: index,
+					},
+					ActivationEpoch: epoch,
+					Liquidated:      false,
+					//this particular status is needed so that ActivationEpoch can be taken into consideration when checking the IsAttesting() condition.
```
iurii-ssv, line 53: `//this` -> `// this`

operator/duties/committee.go
```diff
@@ -85,59 +84,76 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
-	for _, d := range syncDuties {
-		if h.shouldExecuteSync(d, slot) {
-			specDuty := h.toSpecSyncDuty(d, slot, spectypes.BNRoleSyncCommittee)
-			h.appendBeaconDuty(validatorCommitteeMap, committeeMap, specDuty)
+		if h.shouldExecuteAtt(duty) {
+			share, found := h.validatorProvider.Validator(duty.PubKey[:])
+
+			if found && share.IsAttesting(epoch) && !share.Liquidated {
+				addToCommitteeMap(resultCommitteeMap, committee, h.toSpecAttDuty(duty, spectypes.BNRoleAttester))
+			}
```
iurii-ssv, line 114: nit: maybe this part could be moved under `shouldExecuteAtt` func to simplify `buildCommitteeDuties` ?
```
share, found := h.validatorProvider.Validator(duty.PubKey[:])

			if found && share.IsAttesting(epoch) && !share.Liquidated {
```

and similar applies to `shouldExecuteSync` I guess

operator/duties/committee.go
```diff
@@ -85,59 +84,78 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
+		}
 
-func (h *CommitteeHandler) appendBeaconDuty(vc validatorCommitteeDutyMap, c committeeDutiesMap, beaconDuty *spectypes.ValidatorDuty) {
-	if beaconDuty == nil {
-		h.logger.Error("received nil beaconDuty")
-		return
-	}
+		if h.shouldExecuteSync(duty, slot) {
+			share, found := h.validatorProvider.Validator(duty.PubKey[:])
+			if !found || !share.IsSyncCommitteeEligible(epoch, h.network.Beacon.EstimatedSyncCommitteePeriodAtEpoch) {
```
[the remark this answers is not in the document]
oleg-ssvlabs: After the latest code review, we decided to include the `IsSyncCommitteeEligible` check within the `IsParticipating` call, since we identified that the majority of callers actually want to see sync committee eligible validators in the list for their use cases. Fetching and executing _attestations_ and _proposals_ conditions were changed not to include sync committee eligible validators so `IsParticipating`/`SelfParticipatingValidators` is not used there anymore (changed to `!Liquidated` + `IsAttesting()` as it was before)

operator/duties/committee.go
```diff
@@ -85,59 +84,76 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
+	validatorCommittees := map[phase0.ValidatorIndex]committeeDuty{}
+
+	for _, validatorShare := range selfValidators {
+		committeeDuty := committeeDuty{
+			id:          validatorShare.CommitteeID(),
+			operatorIDs: validatorShare.OperatorIDs(),
 		}
+
+		validatorCommittees[validatorShare.ValidatorIndex] = committeeDuty
 	}
```
iurii-ssv, line 98: It's probably a matter of opinion but I'll mention it in case you want to apply this suggestion,

I think it makes sense to group code like this "more tightly" for easier readability, eg. this whole block would read better as (especially when it's a part of large function where we have multiple blocks like this one):
```go
	validatorCommittees := map[phase0.ValidatorIndex]committeeDuty{}
	for _, validatorShare := range selfValidators {
		committeeDuty := committeeDuty{
			id:          validatorShare.CommitteeID(),
			operatorIDs: validatorShare.OperatorIDs(),
		}
		validatorCommittees[validatorShare.ValidatorIndex] = committeeDuty
	}
```

that is to say sometimes using "newline" sparingly results in better high-level structure (at the cost of making child-blocks of code a bit harder to read) - but high-level structure is read much more often hence it makes sense to optimize for that

operator/duties/sync_committee_test.go
```diff
@@ -691,3 +589,88 @@ func TestScheduler_SyncCommittee_Early_Block(t *testing.T) {
 	cancel()
 	require.NoError(t, schedulerPool.Wait())
 }
+
+func eligibleShares() []*ssvtypes.SSVShare {
+	var eligibleShares []*ssvtypes.SSVShare
```
iurii-ssv, line 594: Could just call `eligibleShares` as `result` to keep it "clean and short"

operator/duties/sync_committee_test.go
```diff
@@ -691,3 +589,88 @@ func TestScheduler_SyncCommittee_Early_Block(t *testing.T) {
...
+
+	participationShares := []*ssvtypes.SSVShare{
+		{
+			//share that is participating
```
iurii-ssv, line 598: nit: typically a spacebar is used when writing comments that are meant to be read (`// share that is participating`)

registry/storage/validatorstore.go
```diff
@@ -191,13 +196,14 @@ func (c *validatorStore) SelfParticipatingValidators(epoch phase0.Epoch) []*type
 	if c.operatorID == nil {
 		return nil
 	}
-	validators := c.OperatorValidators(c.operatorID())
+	shares := c.OperatorValidators(c.operatorID())
```
iurii-ssv, line 199: Might as well rename `OperatorValidators` -> `OperatorShares` I guess

tool.mod
```diff
@@ -16,9 +16,9 @@ require (
 	github.com/Abirdcfly/dupword v0.1.3 // indirect
 	github.com/Antonboom/errname v1.1.0 // indirect
 	github.com/Antonboom/nilnil v1.1.0 // indirect
-	github.com/Antonboom/testifylint v1.6.0 // indirect
+	github.com/Antonboom/testifylint v1.6.1 // indirect
```
iurii-ssv, line 19: Unless `tool.mod` needs to be updated for this PR to function it's probably better to update dependancies in a separate RP ?

iurii-ssv (approved):
LGTM, mostly nits from me

operator/duties/attester_test.go
oleg-ssvlabs: fixed

operator/duties/attester.go
oleg-ssvlabs: fixed - renamed to "eligibleShares". Also, changed the filter name in `proposer.go`

tool.mod
oleg-ssvlabs: reverted the commit

registry/storage/validatorstore.go
oleg-ssvlabs: It seems we’re using “Validator” and “Share” interchangeably. There are several public methods that return a Share or a collection of Shares, but are named “Validator(s)". So I am not sure about this change :(

operator/duties/sync_committee_test.go
oleg-ssvlabs: fixed

operator/duties/sync_committee_test.go
oleg-ssvlabs: fixed

registry/storage/validatorstore.go
iurii-ssv: Okay, no need to worry about it in this PR

operator/duties/committee.go
oleg-ssvlabs: fixed

operator/duties/committee.go
oleg-ssvlabs: yes, it's a good point. I was thinking about it as well.
Pushed the change. Please review

protocol/v2/types/ssvshare.go
```diff
@@ -97,8 +99,24 @@ func (s *SSVShare) Slashed() bool {
 	return s.Status == eth2apiv1.ValidatorStateExitedSlashed || s.Status == eth2apiv1.ValidatorStateActiveSlashed
 }
 
-func (s *SSVShare) IsParticipating(epoch phase0.Epoch) bool {
-	return !s.Liquidated && s.IsAttesting(epoch)
+func (s *SSVShare) IsParticipating(cfg networkconfig.NetworkConfig, epoch phase0.Epoch) bool {
+	return !s.Liquidated && s.IsSyncCommitteeEligible(cfg, epoch)
+}
```
olegshmuelov, line 104: Issue with `SelfParticipatingValidators` in Validator Registration (Minor)

In the current implementation of `validatorStore.SelfParticipatingValidators`, validators are filtered using `share.IsParticipating()`, which internally checks for sync committee eligibility (`IsSyncCommitteeEligible`).

However, for validator registration duties, we don't want to register validators that have exited (even if they're technically eligible for sync committees). Relying on `IsSyncCommitteeEligible` might incorrectly include exited-but-eligible validators.

This issue is minor and likely doesn't significantly impact current functionality.

olegshmuelov (approved):
LGTM :ok_hand:

protocol/v2/types/ssvshare.go
moshe-blox: @olegshmuelov good point

@olegshmuelov @oleg-ssvlabs wdyt of this clarification and fix?
https://github.com/ssvlabs/ssv/pull/2061/commits/5c236283b5a8431b47eb164fba066d8903d6cb4c

does the *any* vs *all* distinction make it clearer?

operator/duties/committee.go
```diff
@@ -97,39 +97,31 @@ func (h *CommitteeHandler) buildCommitteeDuties(attDuties []*eth2apiv1.AttesterD
...
 	committeeDutyMap committeeDutiesMap,
-	committee committeeDuty,
+	validatorCommittees map[phase0.ValidatorIndex]committeeDuty,
 	specDuty *spectypes.ValidatorDuty,
 ) {
+	committee, ok := validatorCommittees[specDuty.ValidatorIndex]
+	if !ok {
+		h.logger.Error("failed to find committee for validator", zap.Uint64("validator_index", uint64(specDuty.ValidatorIndex)))
+		return
+	}
```
y0sher, line 123: It's a bit confusing this function can fail, it should've probably been in `shouldExecute..` , but it's also good to keep the should functions around ethereum eligibility and not internal issues.

operator/duties/committee.go
```diff
@@ -85,59 +85,55 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
 			id:          committee.id,
 			operatorIDs: committee.operatorIDs,
 			duty: &spectypes.CommitteeDuty{
-				Slot:            beaconDuty.Slot,
-				ValidatorDuties: make([]*spectypes.ValidatorDuty, 0),
+				Slot:            specDuty.Slot,
+				ValidatorDuties: []*spectypes.ValidatorDuty{},
 			},
 		}
-		c[committee.id] = cd
```
olegshmuelov, line 137: missing

operator/duties/sync_committee.go
```diff
@@ -178,46 +179,52 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
...
-	}
-	lastEpoch := h.network.Beacon.FirstEpochOfSyncPeriod(period+1) - 1
 
-	allActiveIndices := h.validatorController.AllActiveIndices(firstEpoch, waitForInitial)
-	if len(allActiveIndices) == 0 {
-		h.logger.Debug("no active validators for period", fields.Epoch(currentEpoch), zap.Uint64("period", period))
-		return nil
+	if period > h.network.Beacon.EstimatedSyncCommitteePeriodAtEpoch(epoch) {
+		epoch = h.network.Beacon.FirstEpochOfSyncPeriod(period)
 	}
```
moshe-blox, line 191: why this change?

i think it was correct to use firstEpoch rather than currentEpoch for the participation checks

wdyt @olegshmuelov?

olegshmuelov: https://github.com/ssvlabs/ssv/pull/2061#discussion_r2011906861

oleg-ssvlabs: @moshe-blox there is a comment on top of this function explaining which epochs are selected

operator/duties/committee.go
olegshmuelov: please see
https://github.com/ssvlabs/ssv/pull/2061#discussion_r2046568780

operator/duties/committee.go
```diff
@@ -85,59 +85,66 @@ func (h *CommitteeHandler) processExecution(ctx context.Context, period uint64,
...
 	}
 
-	cd.duty.ValidatorDuties = append(c[committee.id].duty.ValidatorDuties, beaconDuty)
+	cd.duty.ValidatorDuties = append(cd.duty.ValidatorDuties, specDuty)
+
+	committeeDutyMap[committee.id] = cd
```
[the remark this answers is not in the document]
oleg-ssvlabs: Good point!
updated

y0sher (approved):
good job 💯

protocol/v2/types/ssvshare.go
oleg-ssvlabs: > @olegshmuelov good point
> 
> @olegshmuelov @oleg-ssvlabs wdyt of this clarification and fix? [5c23628](https://github.com/ssvlabs/ssv/commit/5c236283b5a8431b47eb164fba066d8903d6cb4c)
> 
> does the _any_ vs _all_ distinction make it clearer?

yes, I see what you mean.
We could think about renaming then as well. For example:
`IsParticipating` -> `IsEligibleForAnyDuty`
`IsParticipatingAndAttesting` -> `IsEligibleForAllDuties`

However, I don't have a strong opinion on the naming.  
I think it's good enough as it is now, though.

moshe-blox (commented):
lgtm 👍

PR #2016, opened 2025-02-03
chore(future): use go 1.24 tool

anatolie-ssv:
Aligning our tool usage with the ~~upcoming~~ (now released) `go1.24` features.
- [X] update relevant libraries
- [X] update source file to mock via `go tool`
- [X] update go version to 1.24 in github workflow and Dockerfiles

PR #2071, opened 2025-03-10
chore: migrate from golang.org/x/exp/{maps,slices} to stdlib (go 1.24)

kchojn:
> [!IMPORTANT]
> This pr depends on: https://github.com/ssvlabs/ssv/pull/2016. 
>
> Some slices functionality requires `go` minimum 1.23 (current 1.22), so this solution allows us to take advantage of the functions and modules that `go` offers, instead of writing our own additional functions

Deps:

- [x] golang.org/x/exp/slog
- [x] golang.org/x/exp/slices
- [x] golang.org/x/exp/maps- new `maps` returns an iterator (insted of a slice of values), thats why we need to *collect* it somehow. More here: https://stackoverflow.com/a/71635953/7771104


.github/workflows/lint.yml
```diff
@@ -20,8 +20,8 @@ jobs:
         with:
           go-version: "1.22.x"
 
-      - name: Run lint-prepare
-        run: make lint-prepare
+      - name: Run tool
```
nkryuchkov, line 23: Should it be `Run tools`? It looks a bit inconsistent. Or maybe `Install tools`?

tool.mod
```diff
@@ -0,0 +1,273 @@
+module github.com/anatollupacescu/llm
```
nkryuchkov, line 1: typo (likely from https://github.com/ssvlabs/ssv/pull/2016)

go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	go.uber.org/mock v0.5.0
 	go.uber.org/multierr v1.11.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/exp v0.0.0-20240808152545-0cdaa3abc0fa
+	golang.org/x/exp v0.0.0-20240808152545-0cdaa3abc0fa // indirect
```
nkryuchkov, line 52: You deleted usage of `golang.org/x/exp`, so it's not a direct dependency anymore. Indirect dependencies should go in the next `require` block

nkryuchkov (commented):
LGTM, except the redundant dependency in go.mod

go.mod
kchojn: strange,
`go mod tidy` didn't want to move it, so I had to move it manually, then use `go mod tidy`. Fixed

nkryuchkov (approved):

beacon/goclient/validator.go
```diff
@@ -86,7 +87,7 @@ func (gc *GoClient) registrationSubmitter(slotTickerProvider slotticker.Provider
 
 			// Select registrations to submit.
 			gc.registrationMu.Lock()
-			allRegistrations := maps.Values(gc.registrations)
+			allRegistrations := slices.Collect(maps.Values(gc.registrations))
```
moshe-blox, line 90: i think it's possible to iterate without Collect

moshe-blox (approved):
lgtm 👌

Issue #1828, opened 2024-10-29
logs: duty_trace_id


iurii-ssv:
Having spent some time by now browsing through logs, it got me thinking about the following.

_The problem:_

Duty is uniquely defined by `slot`, `duty-type` and `validator`

when troubleshooting the most important question we typically want to answer is what happened to a particular duty execution (why it failed, what took too long, ...). The current "reliable" approach seems to be to search logs by `slot` and then manually trim the result  down. We can filter additionally by `duty-type` and `validator` (which can be represented either by index or by pubkey) but that becomes less "reliable" because with the current implementation different fields are added to `logger` in different places - it's hard to guarantee that every log line related to a particular duty contains all 3 ( `slot`+`duty-type`+`validator(index || pubkey)` ). Plus it's easy to make a mistake when composing such filter(s). And we might also find false-positive unrelated lines this way.

For committee duty, for example, this 3-filter pattern doesn't seem to work because we don't attach `validator` identifiers (index or pubkey) to `logger` (we only log committee validator(s) identifiers in a few select places). We can try to add it ofc, but implementing a single point of management for this instead of further spreading it out across whole code base (as well as do it at the earliest possible moment - which is the duty start / p2p message arrival - so we "mark" all the logs we can) seems like better long-term solution.

_Suggested solution (general idea):_
- `duty-trace-id` is a unique duty identifier constructed as `slot`+`duty-type`+`validator(index)` (it's pretty similar to `duty_id` which is just a strictly "weaker" version of `duty-trace-id`, see more on it below)
- `duty-trace-id` comes from 2 different sources: 1) locally started duty processing flow 2) messages received from peers **as soon as msg is successfully parsed** (but **before** msg **validation** kicks in)
- `duty-trace-id` needs to be added/logged for every duty-related log (can be done by adding it to `logger` object via `logger.With`) but only for duty execution (we probably don't want to add `duty-trace-id` to duty fetching flow, but we can)
- (**optional**, see alternative below) `duty-trace-id` needs to be added to every p2p duty-related message (so it can be immediately parsed and logged on the receiving side), it is is extra data to pass around in p2p network (although it's just 1 small field - so shouldn't make much difference) which is undesirable

_Suggested solution (details + concerns):_
[...]

PR #2076, opened 2025-03-10
feat(traces): OTeL Traces implementation(duty flow)

oleg-ssvlabs:
**Description**

   - There are three statuses for spans: `Ok`, `Error`, and `Unset`.  Ideally, all spans should have their status explicitly set to either `Ok` or `Error`—`Unset` is not expected.
   - The namespaces for _metrics_, _traces_, and _attributes_. Namespaces should be consistent across all observability "primitives".  There is a chance we have some inconsistencies
   Example:  `ssv.validator.duty` vs. `ssv.duty`. If `Duty` belongs to `Validator`, use `ssv.validator.duty` everywhere.
   - [OpenTelemetry Specification](<https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status>) explains why `OK` status is used without a message. ("Description MUST be IGNORED for StatusCode Ok & Unset values."). Even if the message is set for OK statuses, it will be ignored by OTeL and not displayed in Grafana (yeah, SDKs could have been better here)
   - Some libraries needs to be updated for proper context propagation, especially for methods that perform I/O (e.g., HTTP calls).  Example: `p2p.Broadcast()`.
   - **Traces Configuration**. OTeL supports configuration via both environment variables and SDKs, with SDK configurations taking precedence. To ensure users can fully leverage native OTeL environment variables, SDK usage is kept to a minimum, preventing overrides of manually set environment variables.  Currently, only one configuration flag has been introduced for traces: **`TRACES_ENABLED`**. If set to `false`(**default**), a [noop](https://github.com/ssvlabs/ssv/pull/2076/files#diff-48521b63e66aa332a01d574fb455c0ba112c671dc6b239c594f6815e53477a79R22) trace provider will be used. Otherwise, tracing will be enabled, and all configurations will be read from environment variables. If no environment variables are set, default values will be applied according to the OTeL specification.  
      
**Uncertainties**: 
   - Are both logs and events needed for recording of the same type of event? Should it be either-or? [Example](https://github.com/ssvlabs/ssv/pull/2076/files#diff-3f9aada2a6ec9bd4a5e60f6011bdc8d2bae689ed69c8b4584e817587cb7d9cc4R126).

   - Duty flows are separated into `Committee` versus everything else (method/function level). This is reflected in different _span names_.  Should we keep separating them on a span level, or should we use the same names with attributes that help differentiate duties (like `ssv.runner.role,` etc)
     Span name examples:  
     - `ssv.validator.execute_committee_duty`  
     - `ssv.validator.execute_duty`  
     - `ssv.validator.start_committee_duty`  
     - `ssv.validator.start_duty`

   - Some enums in the libraries used by SSV Node lack a `String()` method, which complicates logging and tracing.  
   Example: `types.PartialSigMsgType` (`ssv-spec` lib). Something that should potentially be implemented by these libraries (we own source code)

   - Should these [enums](<https://github.com/ssvlabs/ssv/blob/main/protocol/v2/message/msg.go#L12>) be moved to the `spec types` package instead?


protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -28,24 +31,35 @@ type queueContainer struct {
...
 			msgID := msg.MsgID.String()
-			logger.Warn("❗ dropping message because the queue is full",
+			const errMsg = "❗ dropping message because the queue is full"
+			logger.Warn(errMsg,
 				zap.String("msg_type", message.MsgTypeToString(msg.MsgType)),
 				zap.String("msg_id", msgID))
+			span.SetStatus(codes.Error, errMsg)
 		}
-		// logger.Debug("📬 queue: pushed message", fields.MessageID(msg.MsgID), fields.MessageType(msg.MsgType))
+		span.SetStatus(codes.Ok, "")
```
iurii-ssv, line 58: The way this code executes - status **Ok** overrides **Error** here (which isn't what we want)

protocol/v2/ssv/validator/validator.go
```diff
@@ -92,14 +95,26 @@ func NewValidator(pctx context.Context, cancel func(), options Options) *Validat
 
 // StartDuty starts a duty for the validator
 func (v *Validator) StartDuty(ctx context.Context, logger *zap.Logger, duty spectypes.Duty) error {
+	ctx, span := tracer.Start(ctx,
+		fmt.Sprintf("%s.start_duty", observabilityNamespace),
```
iurii-ssv, line 99: I think the pattern of `fmt.Sprintf("%s.start_duty", observabilityNamespace),` repeats all over the place - maybe it's worth simplifying to via some helper **func** that will take the part that's different every time as a parameter (here for example it's `start_duty`) and outputs the final string value we want to have here

protocol/v2/ssv/validator/committee.go
```diff
@@ -193,67 +226,121 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
 // ProcessMessage processes Network Message of all types
 func (c *Committee) ProcessMessage(ctx context.Context, logger *zap.Logger, msg *queue.SSVMessage) error {
+	msgType := msg.GetType()
+	ctx, span := tracer.Start(ctx, fmt.Sprintf("%s.process_committee_message", observabilityNamespace),
+		trace.WithAttributes(
+			observability.ValidatorMsgIDAttribute(msg.GetID()),
+			observability.ValidatorMsgTypeAttribute(msgType),
+			observability.RunnerRoleAttribute(msg.GetID().GetRoleType()),
+		),
+		trace.WithLinks(trace.LinkFromContext(msg.Context)))
```
iurii-ssv, line 236: Could you clarify the difference between `ctx` passed into `ProcessMessage` vs `msg.Context` and why here we are using `msg.Context` (and not `ctx`),

tbh it's pretty confusing ... isn't there a better way to address this ?

And also why we use `trace.WithLinks` here but not other places ?

protocol/v2/ssv/queue/messages.go
```diff
@@ -18,6 +19,7 @@ var (
 
 // DecodedSSVMessage is a bundle of SSVMessage and it's decoding.
 type SSVMessage struct {
+	Context          context.Context
```
iurii-ssv, line 22: It would be nice to clarify why this `Context` is needed here, see relevant question here too - TODO

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -40,43 +56,65 @@ func (c *Committee) HandleMessage(_ context.Context, logger *zap.Logger, msg *qu
...
 	if r == nil {
-		return fmt.Errorf("no runner found for message's slot")
+		err := fmt.Errorf("no runner found for slot %d", duty.Slot)
+		span.SetStatus(codes.Error, err.Error())
+		return err
 	}
 
 	// required to stop the queue consumer when timeout message is received by handler
-	queueCtx, cancelF := context.WithDeadline(c.ctx, time.Unix(c.BeaconNetwork.EstimatedTimeAtSlot(duty.Slot+runnerExpirySlots), 0))
+	queueCtx, cancelF := context.WithDeadline(ctx, time.Unix(c.BeaconNetwork.EstimatedTimeAtSlot(duty.Slot+runnerExpirySlots), 0))
```
iurii-ssv, line 107: This change probably means we can remove `ctx    context.Context` member-field on `Committee struct` ?

We can probably apply similar for `Validator struct` if we pass down `ctx` on similar methods accordinggy.

protocol/v2/ssv/runner/committee.go
```diff
@@ -351,24 +445,31 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
 
-	logger.Debug("🧩 got partial signatures",
-		zap.Bool("quorum", quorum),
+	const eventMsg = "🧩 got partial signatures"
+	span.AddEvent(eventMsg)
+	logger.Debug(eventMsg,
+		zap.Bool("quorum", hasQuorum),
 		fields.Slot(cr.BaseRunner.State.StartingDuty.DutySlot()),
 		zap.Uint64("signer", signedMsg.Messages[0].Signer),
 		zap.Int("sigs", len(roots)),
```
iurii-ssv, line 454: I think you've copied this attribute in a couple of places, from what I can tell to **avoid confusion** it should actually be called "roots" (and not "sigs")

because while it's technically correct that "each root = separate signature" it the context of QBFT a signature **typically** means partial signature (of which you need a quorum to consider some message/datavalid/complete)

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ssvlabs/ssv
 
-go 1.22.6
+go 1.24.0
+
+toolchain go1.24.1
```
iurii-ssv, line 5: In general I think 

- we should bump **golang** version is a separate PR (not bundle together with a large PR where we could miss or forget it happened), 
- I'm not sure if it's a good idea to go from 1.22 straight to 1.24 because it's often the case earlier patches have some issues in them so perhaps the latest 1.23 patch would be better option in terms of stability ?
- and also it would be nice to clarify why exactly this is required for this PR

perhaps this is what you meant to clarify by (it just doesn't mention **go 1.24** though):
> Some libraries needs to be updated for proper context propagation, especially for methods that perform I/O (e.g., HTTP calls). Example: p2p.Broadcast().

iurii-ssv (commented):
> Are both logs and events needed for recording of the same type of event? Should it be either-or? [Example](https://github.com/ssvlabs/ssv/pull/2076/files#diff-3f9aada2a6ec9bd4a5e60f6011bdc8d2bae689ed69c8b4584e817587cb7d9cc4R126).

Is it possible to hide away the calls to the logger behind some sort of "Tracer" abstraction (such that we only use Events and "Tracer" simply logs every Event through our Zap logger - so we don't need to call both logger and tracer) ? That would be best in terms of how it makes our code look & read - but it could be quite an overhead to implement.

I guess it depends on how serious we are about tracing
- if we want to instrument much more code similar to how we are doing it here - I'd say duplicating logging/tracing calls isn't ideal
- if this PR represents the bulk of tracing-related work (and we aren't planning to use it for many other components) - it's probably fine as it currently is

^ cc @y0sher @moshe-blox

> Should these [enums](https://github.com/ssvlabs/ssv/blob/main/protocol/v2/message/msg.go#L12) be moved to the spec types package instead?

I would say yes - moving these to **ssv-spec** would probably make more sense IMO

go.mod
kchojn: fyi, it seems to me that some work has begun to make upgrade to 1.24 (one of my prs is based on this pr https://github.com/ssvlabs/ssv/pull/2016)

observability/config.go
```diff
@@ -1,5 +1,16 @@
...
+	}
+
+	metricsConfig struct {
+		enabled bool
+	}
+
+	Config struct {
+		traces  tracesConfig
+		metrics metricsConfig
+	}
```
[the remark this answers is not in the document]
oleg-ssvlabs: It's a good point. However, the current implementation makes it easier to evolve these configs further.  
Previously, we had multiple traces and metrics-related configs, so I decided to separate them.  

I don’t have a strong opinion on this. It would look slightly cleaner now if we applied your suggestion, but we might need to revert to this structure once more configurations are introduced.

protocol/v2/ssv/validator/committee.go
oleg-ssvlabs: It's a good question.  
This method is a [handler](https://github.com/ssvlabs/ssv/blob/stage/protocol/v2/ssv/validator/committee_queue.go#L76) for messages pushed to our internal queue.  

Looking into the OTeL [documentation](https://opentelemetry.io/docs/concepts/signals/traces/#span-links), it seems that links should be used for asynchronous operations (like message queues).  
Spans are bound by an ID propagated through the Go context. In our case, we need to bind the message producer and consumer.  

The Go context passed as a `ProcessMessage` method parameter cannot be used for that, as it does not contain the trace ID from the producer. Instead, the context from the message (`*queue.SSVMessage`) should be used.
So, from what I understood, this is the way to do that

oleg-ssvlabs: > And also why we use trace.WithLinks here but not other places ?

Looking into the code I've instrumented - this is the only place where the process is asynchronous.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -247,12 +334,15 @@ func (r *AggregatorRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 			logger.Error("❌ could not submit to Beacon chain reconstructed contribution and proof",
 				fields.SubmissionTime(time.Since(start)),
 				zap.Error(err))
-			return errors.Wrap(err, "could not submit to Beacon chain reconstructed signed aggregate")
+			err := errors.Wrap(err, "could not submit to Beacon chain reconstructed signed aggregate")
```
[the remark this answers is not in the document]
oleg-ssvlabs: fixed

protocol/v2/ssv/validator/msgqueue_consumer.go
oleg-ssvlabs: good catch. Fixed. Recording the event instead

protocol/v2/ssv/queue/messages.go
oleg-ssvlabs: I am not 100% confident about adding a comment in this particular case.  
The issue is that if the context usage changes (e.g., if it gets used for more cases in the future), the comment might not be updated and could become misleading.  
I think it's fairly easy to "search for usages" here

protocol/v2/ssv/validator/committee_queue.go
oleg-ssvlabs: I will look into the usage of this context actually. 
I think the struct level context helps to propagate cancellations, so I might switch back to the previous context usage

go.mod
oleg-ssvlabs: > In general I think
> 
> * we should bump **golang** version is a separate PR (not bundle together with a large PR where we could miss or forget it happened),
> * I'm not sure if it's a good idea to go from 1.22 straight to 1.24 because it's often the case earlier patches have some issues in them so perhaps the latest 1.23 patch would be better option in terms of stability ?
> * and also it would be nice to clarify why exactly this is required for this PR
> 
> perhaps this is what you meant to clarify by (it just doesn't mention **go 1.24** though):
> 
> > Some libraries needs to be updated for proper context propagation, especially for methods that perform I/O (e.g., HTTP calls). Example: p2p.Broadcast().

Agree with your comment. I updated the PR description to explain why the Go version was bumped.  
Currently, there is a PR that updates Go to the latest version. I believe it is worth prioritizing.

observability/attributes.go
```diff
@@ -1,37 +1,162 @@
...
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/ssvlabs/ssv-spec/qbft"
 	"github.com/ssvlabs/ssv-spec/types"
+	"github.com/ssvlabs/ssv/protocol/v2/message"
+	protocolTypes "github.com/ssvlabs/ssv/protocol/v2/types"
 )
 
 const (
 	RunnerRoleAttrKey = "ssv.runner.role"
```
oleg-ssvlabs, line 21: We need to export this attribute key, because it is used [here](https://github.com/ssvlabs/ssv/blob/main/protocol/v2/qbft/instance/observability.go#L46) where the runner role is type of string (I would say edge case)

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Do you mean **_log_** attribute or **_OTeL_** trace/metric attribute?  
I named it [signatures](https://github.com/ssvlabs/ssv/pull/2076/files#diff-c615c5d14e307fe437330a4144eaa9c319e9f30cf23aef1c724d8712057b1868R435).  
I'm not really keen on changing anything related to logs because we have a variety of tools that work directly with log messages, and modifying them may break these tools.
You are right about `ssv.validator.signature` attribute. It is used in two places. I will create a shared attribute

oleg-ssvlabs: Implemented [these](https://github.com/ssvlabs/ssv/pull/2076/files#diff-c615c5d14e307fe437330a4144eaa9c319e9f30cf23aef1c724d8712057b1868R433) two attributes in the observability package, so they can be reused

protocol/v2/ssv/validator/validator.go
oleg-ssvlabs: I understand what you mean.  
I was thinking about that for a while. Then I thought, maybe it is too simple and does not "deserve" a wrapper.  

However, we have [these](https://github.com/ssvlabs/ssv/blob/main/network/topics/observability.go#L34) wrappers in every `observability.go` file for setting the metric name, which can be used for both traces and metrics since their names follow the same pattern.  

Of course, it should be refactored. I will look into it.

oleg-ssvlabs: Made some [refactor](https://github.com/ssvlabs/ssv/pull/2076/commits/9c0761991141ea133a58817b275e8925a74b9d24).  
From what I understand, in OTeL's terminology, there is a definition of an "instrument," and the instrument can be a log, metric, or trace.

protocol/v2/ssv/validator/committee.go
```diff
@@ -193,67 +226,121 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
+	defer span.End()
+
+	if msg.SignedSSVMessage != nil {
```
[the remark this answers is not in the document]
iurii-ssv: Okay I've created a separate issue here - https://github.com/ssvlabs/ssv/issues/2083 - in case we want to address this separately from this PR (so we don't loose it)

protocol/v2/ssv/runner/committee.go
iurii-ssv: > Do you mean log attribute or OTeL trace/metric attribute?

I mean both I guess - it's called `sigs` for logs (and you've copied it as `signatures` I guess)

> I'm not really keen on changing anything related to logs because we have a variety of tools that work directly with log messages, and modifying them may break these tools.

it seems like a low-level implementation detail that tools shouldn't really depend on (and break every time we want to improve naming)
- I understand if you "do grep by some specific log-pattern" changing it will break it for you
- but I don't think this should prevent us from improving the way our logs are phrased / structured to improve/clarify them

protocol/v2/ssv/validator/committee.go
iurii-ssv: I see, that makes sense,

although browsing through the code I don't think we initialize `msg.Context` everywhere we need to (I assume that would be every `TryPush` call-site, including in `onTimeout`)

protocol/v2/ssv/queue/messages.go
iurii-ssv: Makes sense, I agree with the concept you are describing in general - but here it's a sort of unique/specific use-case we are after (meaning we don't want to be using `SSVMessage.Context` randomly whenever we want)

oleg-ssvlabs: Yeah, I see what you mean. 
I've added a comment and renamed to `TraceContext`

protocol/v2/ssv/validator/committee.go
oleg-ssvlabs: It's a valid concern.
I think it's a problem that SSVMessage context can be `nil` sometimes (when the message is pushed from either [here](https://github.com/ssvlabs/ssv/blob/feat/traces/protocol/v2/ssv/validator/timer.go#L114) and [here](https://github.com/ssvlabs/ssv/blob/feat/traces/protocol/v2/ssv/validator/timer.go#L50)). I think it should be fixed. However, I am not sure these components should be instrumented with traces. They can be, but this PR is focuses on mostly very critical components, so I was not instrumenting the entire codebase. Do you think these two methods should be instrumented?

oleg-ssvlabs: Added Go context to the above mentioned functions

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: You are right. I think it’s not sustainable to build tools that rely on log messages by directly matching log messages or labels.  
I’ve been trying to avoid modifying logs, but if you think it’s necessary, I’ll update it.  

Metrics, Traces, and Logs should ideally complement each other.  
I don’t think it’s a good idea to use traces to record the same events that are already logged.  
For now, this approach works, but once we become more confident in using traces for troubleshooting various scenarios, we should probably remove these logs, or at least decide where to draw the line.

oleg-ssvlabs: Does [this](https://github.com/ssvlabs/ssv/pull/2076/commits/04a5c7afbcee124d9d34fbe2c2e481ba155df933) make sense? 
Still not sure about [this](https://github.com/ssvlabs/ssv/pull/2076#discussion_r1991610642)

protocol/v2/ssv/validator/committee.go
iurii-ssv: > Do you think these two methods should be instrumented?

I would instrument for completeness

protocol/v2/ssv/runner/committee.go
iurii-ssv: > For now, this approach works, but once we become more confident in using traces for troubleshooting various scenarios, we should probably remove these logs, or at least decide where to draw the line.

Deduplicating is best ofc, I've left some thoughts on it in a comment above - https://github.com/ssvlabs/ssv/pull/2076#pullrequestreview-2671883766

> Does [this](https://github.com/ssvlabs/ssv/pull/2076/commits/04a5c7afbcee124d9d34fbe2c2e481ba155df933) make sense?

I would call it `roots` or `roots_cnt` (not `root_len`), but that's minor detail I guess

oleg-ssvlabs: > I would call it roots or roots_cnt (not root_len), but that's minor detail I guess
renamed to `roots`

protocol/v2/ssv/runner/committee.go
```diff
@@ -537,20 +649,27 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
 				cr.GetBeaconNode().GetBeaconNetwork().EstimatedEpochAtSlot(cr.GetBaseRunner().State.StartingDuty.DutySlot()),
 				spectypes.BNRoleSyncCommittee)
 		}
-
-		logger.Info("✅ successfully submitted sync committee",
+		const eventMsg = "✅ successfully submitted sync committee"
+		span.AddEvent(eventMsg, trace.WithAttributes(
+			observability.BeaconSlotAttribute(cr.BaseRunner.State.StartingDuty.DutySlot()),
+			observability.DutyRoundAttribute(cr.BaseRunner.State.RunningInstance.State.Round),
+			attribute.String("ssv.validator.duty.block_root", hex.EncodeToString(syncCommitteeMessages[0].BeaconBlockRoot[:])),
```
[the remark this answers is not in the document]
iurii-ssv: > Maybe this should be ssv.beacon.block.root instead?

Probably yes

observability/attributes.go
iurii-ssv: I guess we'll be able to refactor this away as per https://github.com/ssvlabs/ssv/pull/2098#discussion_r2016796682

logging/fields/fields.go
```diff
@@ -365,8 +365,8 @@ func FeeRecipient(pubKey []byte) zap.Field {
 	return zap.Stringer(FieldFeeRecipient, stringer.HexStringer{Val: pubKey})
 }
 
-func FormatDutyID(epoch phase0.Epoch, slot phase0.Slot, role string, index phase0.ValidatorIndex) string {
-	return fmt.Sprintf("%v-e%v-s%v-v%v", role, epoch, slot, index)
+func FormatDutyID(epoch phase0.Epoch, slot phase0.Slot, beaconRole spectypes.BeaconRole, index phase0.ValidatorIndex) string {
```
oleg-ssvlabs, line 368: I made this change because I noticed that sometimes `BeaconRole` is passed here, and other times `RunnerRole` so I decided to restrict it to `BeaconRole`

message/validation/logger_fields.go
```diff
@@ -98,7 +101,7 @@ func (mv *messageValidator) addDutyIDField(lf *LoggerFields) {
 		// get the validator index from the msgid
 		v, ok := mv.validatorStore.Validator(lf.DutyExecutorID)
 		if ok {
-			lf.DutyID = fields.FormatDutyID(mv.netCfg.Beacon.EstimatedEpochAtSlot(lf.Slot), lf.Slot, lf.Role.String(), v.ValidatorIndex)
+			lf.DutyID = fmt.Sprintf("%v-e%v-s%v-v%v", lf.Role.String(), mv.netCfg.Beacon.EstimatedEpochAtSlot(lf.Slot), lf.Slot, v.ValidatorIndex)
```
oleg-ssvlabs, line 104: related to [this](https://github.com/ssvlabs/ssv/pull/2076/files#r2089176541) comment. This is the only place where `RunnerRole` is used, so I decided to just hardcode the formatting. Also, it is only used within this package

protocol/v2/ssv/validator/validator.go
```diff
@@ -209,3 +295,10 @@ func (v *Validator) withDutyID(logger *zap.Logger, role spectypes.RunnerRole) *z
 
 	return logger
 }
+
+func (v *Validator) fetchTraceContext(ctx context.Context, msgID spectypes.MessageID) (traceCtx context.Context, dutyID string) {
+	if dutyID, ok := v.dutyIDs.Get(msgID.GetRoleType()); ok {
```
oleg-ssvlabs, line 300: Sometimes this map does not contain DutyIDs we are looking for there. I am not sure if that's by design, or it's a bug

protocol/v2/ssv/runner/sync_committee_aggregator.go
```diff
@@ -236,19 +300,36 @@ func (r *SyncCommitteeAggregatorRunner) ProcessConsensus(ctx context.Context, lo
...
+		span.SetStatus(codes.Error, err.Error())
+		return err
 	}
 
-	if !quorum {
+	if !hasQuorum {
+		span.AddEvent("no quorum")
+		span.SetStatus(codes.Ok, "")
 		return nil
 	}
```
iurii-ssv, line 334: I see sometimes `span.AddEvent` is used to supply "additional context", for example:
```
span.AddEvent("no quorum")
span.SetStatus(codes.Ok, "")
```
while other times we add this "additional context" directly as a 2nd argument to `span.SetStatus` call:
```
err := errors.Wrap(err, "could not create consensus data")
span.SetStatus(codes.Error, err.Error())
```
is there any meaningful difference between these 2 variants ? If not it's probably preferable to keep it consistent and always use the 2nd option to avoid that extra line of code (`span.AddEvent ...`)

protocol/v2/ssv/validator/validator.go
iurii-ssv: Briefly looking at the code I suspect this happens only in [this place](https://github.com/ssvlabs/ssv/blob/6df8e5e85b673ce25acb32fabd80b0a3eb9a6e18/protocol/v2/ssv/validator/msgqueue_consumer.go#L35-L35) because Operator can receive a message from another Operator **before** it has started "his own duty" ... in which case there won't be an entry in that map (which gets initialized [here](https://github.com/ssvlabs/ssv/blob/6df8e5e85b673ce25acb32fabd80b0a3eb9a6e18/protocol/v2/ssv/validator/validator.go#L124-L124) during duty start).

This maybe means we might want to use `fields.FormatDutyID` directly instead of relying on `dutyIDs` map ? Why do we even have `dutyIDs` (except for maybe to "cache the computation of duty ID") ? I guess it also indirectly maps RunnerRole->BeaconRole which ties into this https://github.com/ssvlabs/ssv/pull/2076#discussion_r2089181757 edge-case ... that would maybe suggest we need to create a deterministic RunnerRole->BeaconRole mapping (static one, in **ssv-spec** repo maybe) to use instead of `dutyIDs` map ...

I guess it relates to how "duty ID" is constructed and treated in our codebase as described in https://github.com/ssvlabs/ssv/issues/1828 - although it's probably out of scope of this PR

iurii-ssv (approved):
Going over these changes again I didn't spot any issues/bugs, just left couple of comments we can discuss & develop further later on

message/validation/logger_fields.go
iurii-ssv: There are some notes in https://github.com/ssvlabs/ssv/pull/2076#discussion_r2090933133 on how we can potentially address this "properly"

protocol/v2/ssv/validator/validator.go
oleg-ssvlabs: I see. It makes sense.
I am not confident it is possible to collect all needed variables to construct `dutyID` in places like [here](https://github.com/ssvlabs/ssv/pull/2076/files#diff-a8a2ec99c58b542c7d1a45dbdca1e59ddec5c5ca5278d35540bbdd4a95855cf0R147) or [here](https://github.com/ssvlabs/ssv/pull/2076/files#diff-797a0f5a07b46bdd9098b79186baa1648692712befe47ae75096e67a8056cc89R35). From what I can see there is a switch case on the message type and only after type assertion variables like `slot` are present which are needed at the beginning of the function to start the trace. 
Regarding the purpose of this map, I believe it’s used solely to enrich the logger (and now traces) with the `DutyID`.  This likely means that in some places where it’s needed, constructing the `DutyID` directly isn't possible due to missing variables, as I mentioned above. 
Wdyt? I am not a big fan of this map if that's the only purpose it serves, although we need to have logger/tracer enriched by DutyID

protocol/v2/ssv/runner/sync_committee_aggregator.go
oleg-ssvlabs: That's a good question.

`span.SetStatus(codes.Ok, "")`-  this is a somewhat confusing API in the OpenTelemetry SDKs.  
The message passed as the second parameter is always ignored and never shows up when traces are viewed.  
Because of that, I often add an event beforehand (especially when it's not at the end of the method, but rather when an early return happens due to some condition).

On the other hand, for an `Error` status:  
`span.SetStatus(codes.Error, err.Error())` -  the error message passed here **is not ignored** by OpenTelemetry and will appear in the backend.  
(I believe the data goes through multiple transformations before it ends up in storage). So in this case, adding an explicit event isn't really necessary. That's at least the pattern I follow across entire codebase (for OK and Error statuses)

message/validation/logger_fields.go
oleg-ssvlabs: > There are some notes in [#2076 (comment)](https://github.com/ssvlabs/ssv/pull/2076#discussion_r2090933133) on how we can potentially address this "properly"

Do you mean specifically adding a mapper to `ssv-spec` repo to handle this case?

iurii-ssv: > Do you mean specifically adding a mapper to ssv-spec repo to handle this case?

Yep, if we can map `RunnerRole` onto `BeaconRole` I think we can re-use `FormatDutyID`

protocol/v2/ssv/validator/validator.go
iurii-ssv: > I am not confident it is possible to collect all needed variables to construct dutyID in places like [here](https://github.com/ssvlabs/ssv/pull/2076/files#diff-a8a2ec99c58b542c7d1a45dbdca1e59ddec5c5ca5278d35540bbdd4a95855cf0R147) or [here](https://github.com/ssvlabs/ssv/pull/2076/files#diff-797a0f5a07b46bdd9098b79186baa1648692712befe47ae75096e67a8056cc89R35). From what I can see there is a switch case on the message type and only after type assertion variables like slot are present which are needed at the beginning of the function to start the trace.

Right, I assumed every duty-related p2p message should contain all the data to identify corresponding duty (which I think is `slot`+`duty-type`+`validator(index || pubkey)` but I don't know for sure, it's a bit hard to tell whether it's true or not without deep-diving into the related code.

observability/TRACES.md
```diff
@@ -0,0 +1,17 @@
+# Traces
+
+SSV Node implements the OpenTelemetry specification for traces.  
+It supports all standard OTel environment variables, as well as one SSV Node–specific environment variable to enable or disable traces.
+
+When traces are enabled, the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` ([docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_traces_endpoint)) environment variable **must** be set.  
+Depending on the selected protocol (`http` or `grpc`), you may also need to set `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` ([docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_traces_protocol)).
```
iurii-ssv, line 7: I wonder whether we could/should have a check on startup that would prevent SSV node from starting if in SSV config we have **traces enabled** while `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` env variable (and maybe `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`) is not defined ?

oleg-ssvlabs: I thought about implementing validations, but once I started expanding on the idea, it quickly became too complex - so I decided not to proceed with it.

The challenge is that there are **many** OTeL environment variables, and in many cases, multiple variables need to be configured in relation to one another.  
For example: if `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is `gRPC` endpoint, then `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` must also be set to `grpc`, etc.

Implementing good validation on our side would require to do a lot of research.Also, if the OTel project changes how these environment variables work, our validations could become outdated or even incorrect.

So, in my opinion, it's better to delegate that responsibility to the operator(check the OTeL docs).  
If something is misconfigured, OTeL will emit warnings in the logs, which should make it easy to identify and fix the issue.
All in all, this doc gives some hints about what's needed to be configured. If operator needs to do something more complex, they can always find into in OTeL documentation.

message/validation/logger_fields.go
```diff
@@ -66,6 +68,7 @@ func (mv *messageValidator) buildLoggerFields(decodedMessage *queue.SSVMessage)
 
 	descriptor.DutyExecutorID = decodedMessage.SSVMessage.GetID().GetDutyExecutorID()
 	descriptor.Role = decodedMessage.SSVMessage.GetID().GetRoleType()
+	decodedMessage.SSVMessage.GetID().GetRoleType()
```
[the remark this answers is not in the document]
oleg-ssvlabs: That's something that shouldn't be there - nice catch! Fixed

protocol/v2/qbft/controller/observability.go
```diff
@@ -0,0 +1,14 @@
+package controller
+
+import (
+	"go.opentelemetry.io/otel"
+)
+
+const (
+	observabilityName      = "github.com/ssvlabs/ssv/protocol/v2/qbft/controller"
+	observabilityNamespace = "ssv.validator"
```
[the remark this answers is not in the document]
oleg-ssvlabs: slightly adjusted the trace names

observability/TRACES.md
oleg-ssvlabs: I added an example configuration using environment variables, and also moved `TRACES.md` to the `/docs` directory. Since `LOGS.md` is already located there, I think it's a fitting place for the traces documentation as well

operator/duties/attester.go
```diff
@@ -193,13 +236,19 @@ func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 
 	eligibleIndices := indicesFromShares(eligibleShares)
 	if len(eligibleIndices) == 0 {
-		h.logger.Debug("no active validators for epoch", fields.Epoch(epoch))
+		const eventMsg = "no active validators for epoch"
+		h.logger.Debug(eventMsg, fields.Epoch(epoch))
+		span.AddEvent(eventMsg)
+		span.SetStatus(codes.Ok, "")
```
moshe-blox, line 242: why is posting OK necessary? if the span is done without error, does it not imply its OK?

this could shave off quite a bit of repetition in the code

protocol/v2/qbft/controller/controller.go
```diff
@@ -55,24 +60,38 @@ func NewController(
...
+	ctx, span := tracer.Start(ctx,
+		observability.InstrumentName(observabilityNamespace, "qbft.controller.start"),
+		trace.WithAttributes(observability.BeaconSlotAttribute(phase0.Slot(height))))
+	defer span.End()
 
 	if err := c.GetConfig().GetValueCheckF()(value); err != nil {
-		return errors.Wrap(err, "value invalid")
+		err = errors.Wrap(err, "value invalid")
+		span.SetStatus(codes.Error, err.Error())
+		return err
```
moshe-blox: since this is such a common pattern, wdyt of adding this func in observability package:
```go
func Errorf(span Span, f string, args ...any) error {
    err := fmt.Errorf(f, args...)
    span.SetStatus(codes.Error, err.Error)
    return err
}
```

so these 3 lines can be replaced with 1:
```suggestion
return observability.Errorf(span, "value invalid: %w", err)
```

this does introduce a bit of "magic" so not sure i'd go for it, but just thought its an interesting trade-off to consider

protocol/v2/qbft/instance/instance.go
```diff
@@ -83,30 +94,41 @@ func (i *Instance) Start(ctx context.Context, logger *zap.Logger, value []byte,
...
+
 				if err := i.Broadcast(logger, proposal); err != nil {
 					logger.Warn("❌ failed to broadcast proposal", zap.Error(err))
+					span.SetStatus(codes.Error, err.Error())
 				}
+
+				span.SetStatus(codes.Ok, "")
 			}
 		}
+		span.SetStatus(codes.Ok, "")
```
moshe-blox: is there any effect from setting this twice?

operator/duties/attester.go
oleg-ssvlabs: The default status is "unset".

There are three possible statuses: "unset", "ok", and "error".

It’s better to explicitly set the status to "ok", as "unset" can be seen as "unknown"

protocol/v2/qbft/controller/controller.go
oleg-ssvlabs: I see what you mean.

It looks clean and less verbose, but from other perspective - all instrumented methods would return something that is being instantiated from `observability` package, which may look odd. 

Would be good to also hear other opinions on this.

protocol/v2/qbft/instance/instance.go
oleg-ssvlabs: good catch. Fixed

protocol/v2/qbft/controller/controller.go
y0sher: I agree this is repeated a lot and would look better encapsulated. you could create a local function in each package instead of calling `observability` though I think if `observability` is imported in the package anyway it's not a big deal.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -90,9 +93,19 @@ func (r *ProposerRunner) HasRunningDuty() bool {
...
 
 func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Logger, signedMsg *spectypes.PartialSignatureMessages) error {
-	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(ctx, r, signedMsg)
+	ctx, span := tracer.Start(ctx,
+		observability.InstrumentName(observabilityNamespace, "runner.process_pre_consensus"),
+		trace.WithAttributes(
+			observability.BeaconSlotAttribute(signedMsg.Slot),
+			observability.ValidatorPartialSigMsgTypeAttribute(signedMsg.Type),
+		))
+	defer span.End()
```
y0sher, line 119: where do we note that this is a proposer runner trace event?

y0sher: I think there should be some attribute with the runner role added to every trace in this file

protocol/v2/qbft/controller/controller.go
iurii-ssv: Couple of suggestions on this from me:
- indeed it would be nice to have a wrapper for this code to compress it into 1-liner, but it would probably be a bit more consistent to define it in the same package where `tracer = otel.Tracer(observabilityName)` is defined (not in `observability` package)
- for consistency I would additionally define a complementary method that simply wraps `span.SetStatus(codes.Ok, "")`
- (optionally) we can even push this 1 step further and create a wrapper to abstract away `span` and `tracer` altogether such that there will be a `Start` method that returns "our own `span` object", and our own implementation of `defer span.End()` could maybe **panic** in case we entirely forgot to call corresponding `Success() / Fail()` method; But I'm not sure if that's worth the added effort & risk to go that far :0

oleg-ssvlabs: Pushed the change based on what @moshe-blox wrote. (`observability.Errorf()/Error()` implementation)

@iurii-ssv good suggestions, but I didn't really want to create any abstractions over OTeL native packages. They are fairly simple, our abstractions would most likely end up being very similar from usability perspective. Some small util functions - fine, to avoid repetitiveness. 
Maybe further on, when Traces evolve based on the feedback from people who will use them while debugging, we may add more abstractions. It's not a final version of the code though

protocol/v2/ssv/runner/proposer.go
oleg-ssvlabs: This particular span is part of a trace that already includes the runner role attribute.

We don’t have a specific rule about duplicating attributes across all spans within a single trace, but I don’t think it’s necessary. It’s easy to determine the runner role by looking at the root or other spans within the trace

y0sher:
> > Are both logs and events needed for recording of the same type of event? Should it be either-or? [Example](https://github.com/ssvlabs/ssv/pull/2076/files#diff-3f9aada2a6ec9bd4a5e60f6011bdc8d2bae689ed69c8b4584e817587cb7d9cc4R126).
> 
> Is it possible to hide away the calls to the logger behind some sort of "Tracer" abstraction (such that we only use Events and "Tracer" simply logs every Event through our Zap logger - so we don't need to call both logger and tracer) ? That would be best in terms of how it makes our code look & read - but it could be quite an overhead to implement.
> 
> I guess it depends on how serious we are about tracing
> 
>     * if we want to instrument much more code similar to how we are doing it here - I'd say duplicating logging/tracing calls isn't ideal
> 
>     * if this PR represents the bulk of tracing-related work (and we aren't planning to use it for many other components) - it's probably fine as it currently is
> 
> 
> ^ cc @y0sher @moshe-blox 

tbh I only partially agree, it sounds good to wrap repeated code around but looking through the code I see there are some corner cases that won't be included in this generic func so easily so it's better to leave the explicit original implementation

protocol/v2/ssv/validator/validator.go
oleg-ssvlabs: I created a separate ticket for this issue.

PR #2078, opened 2025-03-11
feat: add metric for tracking round changes

kchojn:


protocol/v2/qbft/instance/prepare.go
```diff
@@ -100,7 +100,7 @@ func getRoundChangeJustification(state *specqbft.State, prepareMsgContainer *spe
 	return ret, nil
 }
 
-// validSignedPrepareForHeightRoundAndRoot known in dafny spec as validSignedPrepareForHeightRoundAndDigest
+// validSignedPrepareForHeightRoundAndRootIgnoreSignature known in dafny spec as validSignedPrepareForHeightRoundAndDigest
```
nkryuchkov, line 103: @kchojn What do you think about changing the function name instead (`validSignedPrepareForHeightRoundAndRoot`)? I checked the spec and `validSignedPrepareForHeightRoundAndDigest` doesn't verify the signature, so IMO the `IgnoreSignature` suffix is kind of redundant

```dfy
    predicate validSignedPrepareForHeightRoundAndDigest(
        sPayload: SignedPrepare, 
        height:nat, 
        round:nat, 
        digest:Hash,
        validators: seq<Address>
        )
    {
        var uPayload := sPayload.unsignedPayload;
        && uPayload.height == height
        && uPayload.round == round
        && uPayload.digest == digest
        && recoverSignedPrepareAuthor(sPayload) in validators
    }
```

protocol/v2/qbft/instance/observability.go
```diff
@@ -23,6 +24,16 @@ const (
...
 )
 
+// roundChangeReason represents the reason for a round change in the QBFT protocol
+type roundChangeReason string
+
+const (
+	timeoutReason       roundChangeReason = "timeout"
+	partialQuorumReason roundChangeReason = "partial-quorum"
+	receivedReason      roundChangeReason = "received"
+	justifiedReason     roundChangeReason = "justified"
```
nkryuchkov, line 34: IMO starting with `reason` would be more idiomatic (e.g. https://gobyexample.com/enums). If you decide to change it, perhaps we could also change the `stage` type

```suggestion
	reasonTimeout            roundChangeReason = "timeout"
	reasonPartialQuorum roundChangeReason = "partial-quorum"
	reasonReceived          roundChangeReason = "received"
	reasonJustified           roundChangeReason = "justified"
```

protocol/v2/qbft/instance/round_change.go
```diff
@@ -44,6 +44,8 @@ func (i *Instance) uponRoundChange(
 		zap.Uint64("msg_round", uint64(msg.QBFTMessage.Round)),
 	)
 
+	i.metrics.RecordRoundChange(ctx, msg.QBFTMessage.Round, receivedReason)
```
nkryuchkov, line 47: Do we need to record it? We do it once again below depending on the if branch:

```go
i.metrics.RecordRoundChange(ctx, msg.QBFTMessage.Round, justifiedReason)
```
```go
i.metrics.RecordRoundChange(ctx, newRound, partialQuorumReason)
```

I think we don't need `receivedReason`, do we?

protocol/v2/qbft/instance/round_change.go
```diff
@@ -120,6 +124,9 @@ func (i *Instance) uponChangeRoundPartialQuorum(ctx context.Context, logger *zap
 	if err != nil {
 		return errors.Wrap(err, "failed to hash instance start value")
 	}
+
+	i.metrics.RecordRoundChange(ctx, newRound, partialQuorumReason)
```
nkryuchkov, line 128: What do you think about moving this to the `uponRoundChange` method for consistency? Now, when I read it, from first sight, I feel like it's recorded only for `justifiedReason` until I check the contents of `uponChangeRoundPartialQuorum`

protocol/v2/qbft/instance/observability.go
kchojn: yes, this is more consistent with go, honestly, I was used to using reason as a suffix, but I changed it, thanks!
`stage` is fixed as well

protocol/v2/qbft/instance/round_change.go
kchojn: you are right, `reveivedReason`does seem redundant since im already recording more specific reasons, fixed

nkryuchkov (approved):

observability/attributes.go
```diff
@@ -32,6 +32,10 @@ func DutyRoundAttribute(round qbft.Round) attribute.KeyValue {
 	}
 }
 
+func RoundChangeReasonAttribute(reason string) attribute.KeyValue {
+	return attribute.String("ssv.validator.round.change.reason", reason)
```
oleg-ssvlabs, line 36: We aren't 100% strict with this type of convention because it is slightly vague, but the dot `.` separator is used for separating _namespaces_, and the underscore `_` is used for separating multi-word property names (in the structure `{namespace}.{property}`).  

[Example](https://github.com/ssvlabs/ssv/blob/main/network/p2p/observability.go#L43).  

I know it's not always easy to determine what qualifies as a namespace versus just a long property name, but I think `ssv.validator.round.change` doesn't necessarily need to be a namespace unless we anticipate having other properties apart from `reason`.  

Another point, is that `round` [belongs](https://github.com/ssvlabs/ssv/pull/2078/files#diff-3c40455d22033afac7bc4a7270754723c838819c1e33c5aadae8433395476e37R30) to a `duty` in the hierarchy.

So maybe `ssv.validator.duty.round.change_reason`?

protocol/v2/qbft/instance/metrics.go
```diff
@@ -24,35 +24,24 @@ func (m *metrics) StartStage() {
 	m.stageStart = time.Now()
 }
 
-func (m *metrics) EndStageProposal(ctx context.Context, round qbft.Round) {
+func (m *metrics) EndStage(ctx context.Context, round qbft.Round, s stage) {
```
oleg-ssvlabs, line 27: Nice, I think it is much better to pass the stage name as an argument

protocol/v2/qbft/instance/observability.go
```diff
@@ -32,6 +42,12 @@ var (
 			metric.WithUnit("s"),
 			metric.WithDescription("validator stage(proposal, prepare, commit) duration"),
 			metric.WithExplicitBucketBoundaries(observability.SecondsHistogramBuckets...)))
+
+	roundChangesCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("round.changes"),
```
oleg-ssvlabs, line 48: To align with the attribute name, maybe `metricName("duty.round_changes")`?  

Or, considering that we already have metrics named like  
`ssv.event_syncer.handler.events_processed`,  
perhaps `metricName("duty.rounds_changed")` could be a good option too.

oleg-ssvlabs (commented):
Good job. I left few comments

One question. Have you tried to deploy and visualize this metric in Grafana?

kchojn:
> Good job. I left few comments
> 
> One question. Have you tried to deploy and visualize this metric in Grafana?

Thanks, will do it on Monday, I will probably need someone's help (1st time)

observability/attributes.go
kchojn: Yes, it was problematic, in fact, it always is. My first thought, was your suggestion, so I 100% agree, fixed!

protocol/v2/qbft/instance/observability.go
kchojn: analyzing the rest of the `metricName`s, I changed to `metricName(“duty.rounds_changed”)`, thanks, should be fine now

zktaiga (approved):
This will be very helpful to track, great work! We should also modify the dashboard that we distribute to include a panel that showcases this new metric.

kchojn:
Metrics have been tested and a panel has been added (Experiments/Round change), lgtm

oleg-ssvlabs (approved):
Great job. Looks neat!

protocol/v2/qbft/instance/prepare.go
```diff
@@ -102,7 +102,7 @@ func getRoundChangeJustification(state *specqbft.State, prepareMsgContainer *spe
 
 // validSignedPrepareForHeightRoundAndRoot known in dafny spec as validSignedPrepareForHeightRoundAndDigest
 // https://entethalliance.github.io/client-spec/qbft_spec.html#dfn-qbftspecification
-func validSignedPrepareForHeightRoundAndRootIgnoreSignature(
+func validSignedPrepareForHeightRoundAndRoot(
```
y0sher, line 105: can you please explain this change? `validSignedPrepareForHeightRoundAndRootIgnoreSignature` was taken from the spec https://github.com/ssvlabs/ssv-spec/blob/main/qbft/prepare.go#L82

kchojn: @y0she it was based on this comment: https://github.com/ssvlabs/ssv/pull/2078#discussion_r1990188268

lmk if we should update the specs, or if I should restore the original func name

y0sher: I think in general, In most cases, which should follow the specs. And wait for them to update even if something is "wrong" unless it's necessary to function. Also I can't see how is this part of the scope of this PR.

protocol/v2/qbft/instance/metrics.go
```diff
@@ -24,35 +24,24 @@ func (m *metrics) StartStage() {
...
 		metric.WithAttributes(
-			stageAttribute(proposalStage),
+			stageAttribute(s),
 			roleAttribute(m.role),
 			observability.DutyRoundAttribute(round)))
 	m.stageStart = time.Now()
 }
 
-func (m *metrics) EndStagePrepare(ctx context.Context, round qbft.Round) {
-	validatorStageDurationHistogram.Record(
```
moshe-blox, line 39: is there a good reason to delete this metric?

kchojn: @moshe-blox its not removed; i replaced the duplicated code with this function which takes stage as an argument

`EndStagePrepare` -> `EndStage(*, stage)`

```go
func (m *metrics) EndStage(ctx context.Context, round qbft.Round, s stage) {
	validatorStageDurationHistogram.Record(
		ctx,
		time.Since(m.stageStart).Seconds(),
		metric.WithAttributes(
			stageAttribute(s),
			roleAttribute(m.role),
			observability.DutyRoundAttribute(round)))
	m.stageStart = time.Now()
}
```

moshe-blox (approved):

Issue #1924, opened 2024-12-05
Flaky TestScheduler_Committee_Early_Block_Attester_Only


nkryuchkov:
https://github.com/ssvlabs/ssv/actions/runs/12181517247/job/33978348024

```
--- FAIL: TestScheduler_Committee_Early_Block_Attester_Only (0.41s)
    scheduler.go:389: Unexpected call to *duties.MockDutyExecutor.ExecuteDuty([context.Background.WithCancel.WithCancel 0xc000575000 0xc0003e7280]) at /home/runner/work/ssv/ssv/operator/duties/scheduler.go:389 because: there are no expected calls of the method "ExecuteDuty" for that receiver
```

PR #2079, opened 2025-03-11
fix: flaky `TestScheduler_Committee_Early_Block_Attester_Only` (#1924)

olegshmuelov:
fix https://github.com/ssvlabs/ssv/issues/1924

Fixes a race condition in `TestScheduler_Committee_Early_Block_Attester_Only` where `ExecuteDuty` was sometimes called before mock expectations were set. The issue was resolved by ensuring `setExecuteDutyFuncs` is executed before ticking the slot.

### DIFF
--- a/operator/duties/committee_test.go
+++ b/operator/duties/committee_test.go
@@ -1007,11 +1007,12 @@ func TestScheduler_Committee_Early_Block_Attester_Only(t *testing.T) {
 	// STEP 3: wait for attester duties to be executed faster than 1/3 of the slot duration
 	startTime := time.Now()
 	currentSlot.Set(phase0.Slot(2))
-	ticker.Send(currentSlot.Get())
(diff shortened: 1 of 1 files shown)


nkryuchkov (approved):

oleg-ssvlabs (approved):

iurii-ssv (commented):
I actually have been looking into this a while ago (just didn't come around to push the change),

@olegshmuelov could you take a look at this diff I came up with back then and see if we want to borrow something from it - https://github.com/iurii-ssv/ssv/commit/956dadf0c4fe533a80f07e92d1983d359cb2d66e ? I think there are a bunch of similar tests that also needs the same kind of fix.

iurii-ssv (approved):

y0sher (approved):

PR #2084, opened 2025-03-13
feat: reject messages from removed operators

kchojn:
- we don't want to stop the whole committee because of a single removed operator.
- we do want to encourage validators to move to full committees without removed operators.
- We don't want to accept messages from removed operators.

Hence the way we want to approach this PR is to check to remove the operator data and check the operator data store in msg validation then ignore their messages.

This pr is changing its approach to removed operators and could probably close the old pr: https://github.com/ssvlabs/ssv/pull/1796

## Changes

Here's a changelog based on the provided diff:

```markdown
### Added
- Added `OperatorsExist` interface for message validation
- Added `ValidatorStore` interface for message validation
- Added validation for removed operators in message processing pipeline
- Added helper methods in message validation:
  - `validateSignerNotZero`
  - `validateSignerUnique`
  - `validateSignerExists`
- Added error `ErrRemovedOperator`/ `ErrOperatorValidation` for handling removed operator cases
- Added (message/validation/errors.go) `Is/Unwrap` methods, for better integration with `go`

### Changed
- Improved `handleOperatorRemoved` to actually delete operator data from storage
- Refactored message validator to filter out removed operators from signatures
- Split `WithSelfAccept` option into separate `WithSelfAccept` and `WithSelfPID` options
- Updated validator tests to use the new interfaces and options
- Enhanced operator ID validation in signed messages
- Fixed `buildOperatorKey` to use more efficient string concatenation

### Fixed
- Fixed `OperatorRemoved` event handling in event handler tests

### CODE REVIEW

File: message/validation/signed_ssv_message.go [hunk 1]
```diff
@@ -156,3 +161,25 @@ func (mv *messageValidator) belongsToCommittee(operatorIDs []spectypes.OperatorI
 
 	return nil
 }
+
+// validateOperator checks if the operator exists and is not removed
+func (mv *messageValidator) validateOperator(operatorID spectypes.OperatorID) error {
```
y0sher, line 166: can we UT this function?, ideally we can merge this with codecov happy.

eth/eventhandler/handlers.go
```diff
@@ -120,8 +120,14 @@ func (eh *EventHandler) handleOperatorRemoved(txn basedb.Txn, event *contract.Co
 		fields.Owner(od.OwnerAddress),
 	)
 
-	// This function is currently no-op and it will do nothing. Operator removed event is not used in the current implementation.
+	// Delete the operator data from storage to mark it as removed
+	// This will cause message validation to reject messages from this operator
+	if err := eh.nodeStorage.DeleteOperatorData(txn, event.OperatorId); err != nil {
+		logger.Error("failed to delete operator data", zap.Error(err))
+		return fmt.Errorf("could not delete operator data: %w", err)
```
y0sher, line 127: I don't really like that we log AND return the err because it will probably be logged above as well. The only issue that if we don't log here the above log won't contain the added fields (pk, owner, etc..). So I guess we can leave this.

message/validation/signed_ssv_message.go
```diff
@@ -67,6 +67,11 @@ func (mv *messageValidator) validateSignedSSVMessage(signedSSVMessage *spectypes
...
 		}
 		prevSigner = signer
+
+		// Rule: Reject messages from removed operators
+		if err := mv.validateOperator(signer); err != nil {
+			return err
+		}
 	}
 
 	// Rule: Len(Signers) must be equal to Len(Signatures)
```
y0sher, line 77: lets make sure we don't assign `OperatorIDs` as the whole committee if an operator was removed. (if we do this rule will fail)

cli/operator/node.go
```diff
@@ -265,7 +265,7 @@ var StartNodeCmd = &cobra.Command{
 
 		messageValidator := validation.New(
 			networkConfig,
-			nodeStorage.ValidatorStore(),
+			nodeStorage,
```
y0sher, line 268: To be honest, because only `GetOperatorData` is used, I would've passed in both `ValidatorStore` and a more narrow interface (defined in `validation`) with only `GetOperatorData`. 
But that's maybe style, I would love some other team members to chime in on this nitpick :)
 
@iurii-ssv @oleg-ssvlabs @MatusKysel @nkryuchkov @olegshmuelov

y0sher (changes requested):
Looks good generally! some minor comments :)
Also I'd love @nkryuchkov to review since its msg validation code.

eth/eventhandler/handlers.go
kchojn: good point, fixed!

MatheusFranco99:
The msg validation code change looks good and it's in the appropriate place 👍

iurii-ssv (commented):
Some suggestions

message/validation/errors.go
```diff
@@ -117,6 +117,8 @@ var (
 	ErrInvalidPartialSignatureTypeCount        = Error{text: "sent more partial signature messages of a certain type than allowed", reject: true}
 	ErrTooManyPartialSignatureMessages         = Error{text: "too many partial signature messages", reject: true}
 	ErrEncodeOperators                         = Error{text: "encode operators", reject: true}
+	ErrRemovedOperator                         = Error{text: "operator is removed", reject: true}
+	ErrOperatorValidation                      = Error{text: "failed to validate operator data", reject: true}
```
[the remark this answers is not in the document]
MatheusFranco99: Agree
Also, the operator may re-join and the rejection would be dependent on the local view of the net, so better to ignore the msg indeed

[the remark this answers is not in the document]
nkryuchkov: @MatheusFranco99 do you mean `ErrRemovedOperator` should not be rejected either?

message/validation/signed_ssv_message.go
```diff
@@ -156,3 +161,25 @@ func (mv *messageValidator) belongsToCommittee(operatorIDs []spectypes.OperatorI
...
 	return nil
 }
+
+// validateOperator checks if the operator exists and is not removed
+func (mv *messageValidator) validateOperator(operatorID spectypes.OperatorID) error {
+	if mv.nodeStorage == nil {
+		return nil
+	}
+
+	operatorData, found, err := mv.nodeStorage.GetOperatorData(nil, operatorID)
```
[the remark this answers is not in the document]
kchojn: I overlooked this method, it seems to be better, no need to retrieve and unmarshal all the data, thanks!

message/validation/errors.go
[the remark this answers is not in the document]
iurii-ssv: That's an interesting question on how we can handle intermittent DB errors ... does this whole preliminary validation even make sense then ?

I guess it's a broader question on how we do retries, I'm not sure how it currently works but theoretically/ideally I think we want either of two approaches:
1) soft-reject a message in case on intermittent error such that the sender (or libp2p in this case) will retry it a couple more times until we hard-reject it
2) we accept or straight away hard-reject a message every time (without any options to soft-reject it) and handle temporary failures such as DB connection issues in our internal logic (perhaps in duty runners we simply re-schedule a duty to re-run later if we weren't able to fetch some data from DB)

cli/operator/node.go
oleg-ssvlabs: > the main question is whether we want to define small interfaces on the caller side

I really like this pattern. There are a lot of benefits to using it, and I think it is very Go idiomatic. It reduces cross-package references and makes it very clear which specific functions the caller depends on. When the interface is slim, there is also really no need to use frameworks for mocks. It is fairly easy to mock it manually. 
I think Go stdlib has a lot of cases when the package itself exposes the interface, but I think stdlib is slightly different use case.

oleg-ssvlabs: So I would define interface like this:
```Go
type validatorStore interface {
	Validator(pubKey []byte) (*types.SSVShare, bool)
	Committee(id spectypes.CommitteeID) (*storage.Committee, bool)
}
```
in `validation` package. However if we start following this pattern, then we would also need to define an interface [here](https://github.com/ssvlabs/ssv/blob/main/message/validation/validation.go#L40) as well.  
Also, I do not think there is really a reason for [returning the interface](https://github.com/ssvlabs/ssv/blob/main/message/validation/validation.go#L58) in the constructor. In this case it could be a pointer

message/validation/errors.go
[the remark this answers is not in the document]
nkryuchkov: IMO nodes just need to ignore anything on internal error, because the inability to process a message is their fault, so others should not be punished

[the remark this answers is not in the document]
iurii-ssv: > IMO nodes just need to ignore anything on internal error, because the inability to process a message is their fault, so others should not be punished

Skipping a validation step doesn't make much sense to me (that would be "optional validation" - which is a weird concept to have lol) - if that's what "ignore" means in this context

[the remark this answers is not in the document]
iurii-ssv: > is this DB problem a real hurdle? I mean, does it appear often?

I don't think it's gonna happen often (but still it returns `error` for a reason :) ) - it's just interesting this seems to be the first time such a temporary-error issue showed up among validation rules afaict

Mostly I'm trying to think long-term here and figure out how we want to go about "temporary internal errors" with respect to validation, and unless I'm missing something it seems there are 2 approaches:
- we design validation rules so that they are 100% executable (no chance of temporary failure)
- we implement some sort of retry-mechanism for validation rule(s) themselves  (and only accept messages for further processing once validation rule succeeds) - this way our internal logic doesn't need to perform additional checks and can just rely on message having been validated prior, this simplifies internal logic; We could do retries later on in our internal logic instead ofc (as opposed to validation logic) - but that probably means we don't need "validation rule that might or might not work"  in the first place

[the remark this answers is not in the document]
iurii-ssv: > IMO nodes just need to ignore anything on internal error,

Okay, digging a bit more into this - I guess @nkryuchkov meant that we should return `ValidationIgnore` (and not `ValidationReject`) and not "skipping validation and also returning `ValidationAccept`"

It's not entirely clear to me how bad it is to just drop a message like this (with `ValidationIgnore`) ... I'm guessing once we ignore a message we won't receive it again from gossipsub, right @MatheusFranco99 ? Meaning we'll be dropping some messages due to that - which would negatively impact duty execution overall (especially if we accumulate more of similar validation rules over time)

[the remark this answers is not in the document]
MatheusFranco99: Yes, that's precisely it (returning `ValidationIgnore`).

Yes, we won't receive the message again.

It will impact if we have a wrong state that states an operator as "removed" while it's actually "active". However, such a drop shouldn't compromise the message propagation overall, except if the whole network is facing network delays, but still, in this case, we have Gossip messages that will do the job of delivering the messages.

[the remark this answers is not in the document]
iurii-ssv: Okay, it makes sense to me now, so it probably should be:
```
ErrOperatorValidation                      = Error{text: "failed to validate operator data"}
```
and retry mechanism ("soft-reject" kind of thing) is probably an overkill for this specific use-case - but might be useful to consider at some point if we end up adding more and more validation rules with the possibility of temporary failures

message/validation/signed_ssv_message.go
kchojn:
```
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:28:     validateSignedSSVMessage                        100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:108:    validateSSVMessage                              100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:152:    validRole                                       100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:167:    belongsToCommittee                              100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:182:    validateSignerNotZero                           100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:190:    validateSignerUnique                            100.0%
github.com/ssvlabs/ssv/message/validation/signed_ssv_message.go:198:    validateSignerExists                            100.0%
```

message/validation/signed_ssv_message.go
kchojn: i filtered out removed operators, let me know if it looks good now, these rules are also covered in tests

cli/operator/node.go
kchojn: Based on conversations, I created 2 small interfaces in the `validation` package
```go
// Operators defines the minimal interface needed for validation
type Operators interface {
	OperatorsExist(r basedb.Reader, ids []spectypes.OperatorID) (bool, error)
}

// ValidatorStore defines the minimal interface needed for validation
type ValidatorStore interface {
	Validator(pubKey []byte) (*ssvtypes.SSVShare, bool)
	Committee(id spectypes.CommitteeID) (*registrystorage.Committee, bool)
}
```

it's more straightforward to test now. I updated the dependent files as well.

message/validation/errors.go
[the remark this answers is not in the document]
kchojn: Thank you guys, very helpful conversation, fixed!

message/validation/validation.go
```diff
@@ -19,12 +19,24 @@ import (
...
+// Operators defines the minimal interface needed for validation
+type Operators interface {
+	OperatorsExist(r basedb.Reader, ids []spectypes.OperatorID) (bool, error)
+}
+
+// ValidatorStore defines the minimal interface needed for validation
+type ValidatorStore interface {
+	Validator(pubKey []byte) (*ssvtypes.SSVShare, bool)
+	Committee(id spectypes.CommitteeID) (*registrystorage.Committee, bool)
+}
```
iurii-ssv, line 38: Is there a need for these interfaces to be public ?

kchojn: no, mb, its used only in this package, fixed

message/validation/signed_ssv_message.go
```diff
@@ -54,27 +55,46 @@ func (mv *messageValidator) validateSignedSSVMessage(signedSSVMessage *spectypes
...
+		if err := mv.validateSignerUnique(signer, prevSigner); err != nil {
+			return err
 		}
+
+		// Rule: Signer must exist (not removed)
+		if err := mv.validateSignerExists(signer); err != nil {
+			// If operator is removed, skip it and continue with next operator
+			if errors.Is(err, ErrRemovedOperator) {
+				continue
+			}
```
iurii-ssv, line 86: I'm not sure why we don't just return `err` in case of `ErrRemovedOperator` here ... @kchojn could you clarify ?

kchojn: @iurii-ssv , so the issue is: when operators are removed, they are still included in committee duties, which causes validation errors when messages are processed? i mean we should filter out removed operators when building committee duties? so its gonna be like 2 layer approach to protect?
I'm trying to find the best solution, after searching the code, this seems to be the best place (operator/duties/committee.go) ,but I am not 100% sure

iurii-ssv: > when operators are removed, they are still included in committee duties, which causes validation errors when messages are processed?

Probably - that's something we need to check, and if it is indeed the case

> i mean we should filter out removed operators when building committee duties?

I think yes - we probably want to filter out those removed operators from the message before we send a message. If that's reasonably easy to do. And when validating (2nd layer) - we could return error now instead of just filtering removed operator out (since we would expect for messages to not have removed operators attached)

> I'm trying to find the best solution, after searching the code, this seems to be the best place (operator/duties/committee.go) ,but I am not 100% sure

I think we need to look where/how `SignedSSVMessage.OperatorIDs` is initialized to see if we ever can add removed operator to it (`CommitteeHandler` takes care of just 2 duty types, but there are other duty types that send `SignedSSVMessage` messages). From what I see searching by `OperatorIDs:` we never assign more than 1 operator to `SignedSSVMessage.OperatorIDs` ... but that's probably wrong (why it is slice then?)

so, looking some more I see sometimes we can initialize `SignedSSVMessage.OperatorIDs` like [this](https://github.com/ssvlabs/ssv/blob/aedc6368df1c156c6455ef7d4157fa8d2bb51178/protocol/v2/qbft/instance/commit.go#L72-L81) ...

and so it's pretty hard to tell whether or not we can add a removed operator to `SignedSSVMessage.OperatorIDs` doing it like that (and likely we do - since operator IDs is something we probably "cache in memory" multiple times at multiple places in memory)

thus your original solution is probably the correct one (to just ignore operator if he is removed on the message-receiving side), hence we probably want to revert https://github.com/ssvlabs/ssv/pull/2084/commits/7de957a24338b223ccbab34ce150bfc910dea645 and that should be it

but also, renaming PR accordingly would make more sense as well `reject messages` -> `prune removed operators` - and this would imply we probably want to do this "pruning" not as a part of validation (not in `validation` package) but maybe soon after validation is done - gotta check the code to find the best place for it

iurii-ssv: > we probably want to do this "pruning" not as a part of validation (not in validation package) but maybe soon **after** validation is done

or maybe even **before** depending if other validation rules use removed  operator (his signature) or not

for example if we validate there is a certain number of signatures somewhere - we probably want to not count removed operator signature in there (hence we should prune it before doing validation)

y0sher (commented):
looks good. I think spec should also be involved in how we do msg validation @MatheusFranco99

kchojn:
> looks good. I think spec should also be involved in how we do msg validation @MatheusFranco99

thanks, @MatheusFranco99 this is the part that should be confirmed: https://github.com/ssvlabs/ssv/pull/2084#discussion_r2002887653 -> (refers to this conv: https://github.com/ssvlabs/ssv/pull/2084#discussion_r2001393171)

eth/eventhandler/handlers.go
```diff
@@ -120,7 +120,11 @@ func (eh *EventHandler) handleOperatorRemoved(txn basedb.Txn, event *contract.Co
 		fields.Owner(od.OwnerAddress),
 	)
 
-	// This function is currently no-op and it will do nothing. Operator removed event is not used in the current implementation.
+	// Delete the operator data from storage to mark it as removed
+	// This will cause message validation to reject messages from this operator
+	if err := eh.nodeStorage.DeleteOperatorData(txn, event.OperatorId); err != nil {
+		return fmt.Errorf("could not delete operator data: %w", err)
+	}
```
olegshmuelov, line 127: This comment might be a bit misleading, we're not marking the operator as removed, we're fully deleting it.
Maybe something like:
```go
// Permanently remove operator data to prevent further message validation.
```

message/validation/signed_ssv_message.go
iurii-ssv: Okay, just to be clear, current implementation in this PR doesn't do what **MatheusFranco99** describes (it filters out operator/signature, not ignores message as a whole)

@MatheusFranco99 it is however tricky to implement it the way you describe because (like I'm explaining above) we can't really tell for sure whether or not we attach removed operators to messages operators send or not (once they've been removed):
- it's hard to track down every place where `SignedSSVMessage.OperatorIDs` are initialized (and bundled in the message we eventually receive)
- and even if we can find all those places it might be hard to "stop adding remove operator to signed ssv message" immediately after it has been removed since it might be cached in some in-memory structures
- and it's also pretty hard to test all this

so I'm not sure it's worth taking the risk of the solution @MatheusFranco99 suggests (because we might end up in a situation where we suddenly start ignoring lots of otherwise good messages simply because some operator has been removed, and it's probably gonna cause the whole cluster to stop operating) - @MatheusFranco99 what if we just drop/ignore operators (and their signatures) from the message if we don't recognize them (effectively treating any unknown operators a message references as if it isn't there) ?

cc @y0sher

MatheusFranco99: If a certain operator is in `SignedSSVMessage.OperatorIDs`, then one of the two follows:
- if `len(SignedSSVMessage.OperatorIDs) == 1`, then this operator created the message and signed it. If the sig is correct, it's definitely the one sending it (unless its key was stolen, but we don't need to cover that).
- if len > 1, then it's a `decided` message, which is an aggregation of `commit` messages. Again, the operator once sent such `commit` msg. `decided`'s are sent right after the `commit` phase, so the operator would need to participate in the commit phase, and remove itself before the `decided` msgs. Either way, `decided` messages are not necessary for the protocol. They are an extra help for slow nodes.

Regarding not ignoring a message from an unknown operator, it seems odd to me to let a msg be propagated if you don't know who the operator is. It looks like a gateway for an attacker to flood you with bogus messages and get away with it.

iurii-ssv: Okay, clarification helps, in that case my concern in irrelevant

@kchojn we should be ignoring whole message then (not filter out unknown operators)

iurii-ssv (approved):
LGTM

message/validation/options.go
```diff
@@ -5,20 +5,26 @@ import (
...
 		mv.selfAccept = selfAccept
 	}
 }
+
+// WithSelfPID sets the node's own peer ID.
+func WithSelfPID(selfPID peer.ID) Option {
+	return func(mv *messageValidator) {
+		mv.selfPID = selfPID
+	}
+}
```
moshe-blox, line 30: those are never used in isolation, so there's no point to confuse the caller as to what WithSelfPID means you want to enable it :) (its disabled by default)

if you wish you can remove the boolean case, because anytime you want to call WithSelfAccept 
```suggestion
// WithSelfAccept blindly accepts messages sent from self. Useful for testing.
func WithSelfAccept(selfPID peer.ID) Option {
	return func(mv *messageValidator) {
		mv.selfPID = selfPID
		mv.selfAccept = true
	}
}
```

registry/storage/operators.go
```diff
@@ -225,5 +227,5 @@ func (s *operatorsStorage) DropOperators() error {
 
 // buildOperatorKey builds operator key using operatorsPrefix & index, e.g. "operators/1"
 func buildOperatorKey(id spectypes.OperatorID) []byte {
-	return bytes.Join([][]byte{operatorsPrefix, []byte(strconv.FormatUint(id, 10))}, []byte("/"))
+	return append(operatorsPrefix, []byte("/"+strconv.FormatUint(id, 10))...)
 }
```
moshe-blox, line 231: what was the issue with `bytes.Join`? joining bytes slices is kind of the purpose of it :)

also with `append` it becomes really ugly when you have to join more than 2 slices

message/validation/options.go
y0sher: I agree. In general I think we shouldn't do things that are not related to the PRs scope. If you've encountered a change you think will benefit the code base, break it into a separate PR.

registry/storage/operators.go
y0sher: I agree. In general I think we shouldn't do things that are not related to the PRs scope. If you've encountered a change you think will benefit the code base, break it into a separate PR.

moshe-blox (approved):

moshe-blox (dismissed):
looks good, great job

left one comment

registry/storage/operators.go
moshe-blox: @kchojn @y0sher i missed it before, but this breaks the key (slash was after now its before the id)

kchojn: @moshe-blox `bytes.Join` is much slower than append, `go` still has open performance issues with bytes.join;
The change was added at the beginning of this pr (when I also edited this file). Anyway, I make a revert of the change, you are right, its not part of this pr

message/validation/options.go
kchojn: i get it,  gonna revert it!

kchojn:
@MatusKysel its fixed, codecov included new `mocks`, so I added to the ignored section; the rest of the changes are tested

iurii-ssv (approved):

message/validation/validation_test.go
```diff
@@ -1032,9 +1041,99 @@ func Test_ValidateSSVMessage(t *testing.T) {
 		require.ErrorIs(t, err, ErrDuplicatedSigner)
 	})
 
+	// Get error when receiving a message with an operator that does not exist and has not been removed
+	t.Run("operator exists and not removed", func(t *testing.T) {
+		validator := New(netCfg, validatorStore, operators, dutyStore, signatureVerifier, phase0.Epoch(0)).(*messageValidator)
+
+		err := validator.validateSignerIsKnown(1)
```
oleg-ssvlabs, line 1048: It would be good to hear other opinions on this.
What do you think about writing unit tests for non-exported methods/functions? @y0sher @moshe-blox @iurii-ssv @MatusKysel @nkryuchkov
Personally, I only test exported ones. To me at least, internal implementations are important, but mainly in terms of how they affect the behavior of exported functionality

oleg-ssvlabs (commented):
Good job.
Left one comment that is rather a question

oleg-ssvlabs (approved):

registry/storage/operators.go
moshe-blox: @kchojn i understand, but it's probably a difference on the order of nanoseconds and it's outside a hot path, so we're not too stressed to optimize it :)

nkryuchkov (approved):
LGTM

y0sher (approved):
only spec alignment left

y0sher (approved):

PR #2085, opened 2025-03-17
fix(weighted-attest-data): use muticlient event listener when Weighted Attest Data feature is off

oleg-ssvlabs:


beacon/goclient/events.go
```diff
@@ -58,16 +59,28 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
 	}
 
-	for _, client := range gc.clients {
-		if err := client.Events(ctx, strTopics, gc.eventHandler); err != nil {
-			gc.log.Error(clResponseErrMsg, zap.String("api", "Events"), zap.Error(err))
+	isMultiClientListener := !gc.withWeightedAttestationData
+	logger := gc.log.
+		With(zap.Int("clients_len", len(gc.clients))).
+		With(zap.String("topics", strings.Join(strTopics, ", "))).
+		With(zap.Bool("is_multi_client_listener", isMultiClientListener))
```
moshe-blox, line 66: i dont think there's a reason to log this because we already log the clients on creation, the topics are static at compile time, and weighted param should be logged separately before (if not already?)
```suggestion
	logger := gc.log.With(
		zap.Int("clients_len", len(gc.clients)),
		zap.String("topics", strings.Join(strTopics, ", ")),
		zap.Bool("is_multi_client_listener", isMultiClientListener))
```

beacon/goclient/events.go
```diff
@@ -58,16 +59,28 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
+
+	logger.Info("subscribing to events")
+	if isMultiClientListener {
```
moshe-blox, line 84: i think that instead of flipping the boolean and introducing a new name, maybe we should keep using `gc.withWeightedAttestationData` in the if condition but add a short comment explaining why we switch?

moshe-blox: see you just added a comment :)

y0sher: I agree it is simpler to just do it 
```go
if gc.withWeightedAttestationData {
    for _, client := range gc.clients { ... }
} else {
    if err := gc.multiClient.Events(ctx, strTopics, gc.eventHandler);  //...
}
```

oleg-ssvlabs: alright, refactored

y0sher (approved):

moshe-blox (approved):

nkryuchkov (approved):

PR #2087, opened 2025-03-17
chore(ai): add `.cursorrules`

moshe-blox:


nkryuchkov (commented):
I've used Cursor just a little bit in comparison to ChatGPT, but I noticed that ChatGPT avoids using the testify library by default. Should we add the rule to use testify where possible? 

@moshe-blox Do you know if Cursor's models use it if not explicitly asked?

y0sher (approved):

iurii-ssv (approved):

PR #2090, opened 2025-03-17
feat(metrics): add `Doppelganger` observability metrics

olegshmuelov:
- Added metrics to track validators in DG state and unsafe validators.
- ~Used `Int64UpDownCounter` for accurate real-time tracking.~
- Replaced with `Int64Gauge` for simpler and more consistent slot-based recording. see [comment](https://github.com/ssvlabs/ssv/pull/2090#discussion_r2023163634)


doppelganger/doppelganger.go
```diff
@@ -145,11 +146,15 @@ func (h *handler) updateDoppelgangerState(validatorIndices []phase0.ValidatorInd
 	}
 
 	if len(addedValidators) > 0 {
-		h.logger.Debug("Added validators to Doppelganger state", zap.Uint64s("validator_indices", addedValidators))
+		validatorsStateCounter.Add(ctx, int64(len(addedValidators)))
+		unsafeValidatorsCounter.Add(ctx, int64(len(addedValidators)))
```
y0sher, line 150: @zktaiga @olegshmuelov 
Can't this be one metric ValidatorStateCounter and with a label "unsafe": true/false ?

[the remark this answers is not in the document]
olegshmuelov: We use `Int64UpDownCounter` to track changes incrementally. A gauge would require recalculating the total each time, which isn’t needed since we already track additions/removals.

olegshmuelov: Great idea!
We can consolidate this into a single metric with an attribute unsafe: true/false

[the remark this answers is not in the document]
oleg-ssvlabs: @olegshmuelov, do you mean that `len(addedValidators)` does not fully represent the state of the `validatorsStateCounter`/`unsafeValidatorsCounter` metric?

Gauges are snapshots of values at a given time and are not additive, whereas UpDownCounters are.

It’s an interesting situation because I don’t think we’ve ever used UpDownCounters, and they don’t exist in Prometheus. I’m curious how OTeL transforms them into Prometheus readable metrics.

olegshmuelov: However, maintaining the correct counts while transitioning between unsafe: true and unsafe: false states would add extra complexity. I think it's clearer to have one counter for all validators in state and another specifically for the unsafe count

oleg-ssvlabs: @olegshmuelov it's a great idea regarding the label. Please use `.count` suffix in the metric name, since we have that as [convention](https://github.com/ssvlabs/ssv/blob/main/observability/CONVENTIONS.md?plain=1#L20)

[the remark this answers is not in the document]
olegshmuelov: > @olegshmuelov, do you mean that `len(addedValidators)` does not fully represent the state of the `validatorsStateCounter`/`unsafeValidatorsCounter` metric?
> 
> Gauges are snapshots of values at a given time and are not additive, whereas UpDownCounters are.
> 
> It’s an interesting situation because I don’t think we’ve ever used UpDownCounters, and they don’t exist in Prometheus. I’m curious how OTeL transforms them into Prometheus readable metrics.

The key reason we use `Int64UpDownCounter` instead of a gauge is that we are tracking incremental changes rather than reporting an absolute snapshot. Since we already maintain the state of validators internally, we can accurately track additions and removals without needing to recalculate the total every time, which is what a gauge would require.

If I understand correctly, OTel typically maps UpDownCounter to a standard counter in Prometheus with a delta aggregation. This means Prometheus would see only the net changes over time rather than the total state at any given moment.

oleg-ssvlabs (approved):
Looks awesome

y0sher (changes requested):
@zktaiga I think it makes sense to include the dashboard update in this PR before merging

y0sher (approved):
I think this is ready to merge apart from @zktaiga 's part. We can split the dashboard to a separate PR .

doppelganger/doppelganger.go
```diff
@@ -292,20 +306,30 @@ func (h *handler) processLivenessData(epoch phase0.Epoch, livenessData []*eth2ap
 				fields.ValidatorIndex(response.Index),
 				zap.Uint64("remaining_epochs", uint64(state.remainingEpochs)))
 		} else {
+			validatorsStateCounter.Add(ctx, -1, metric.WithAttributes(unsafeAttribute(true)))
+			validatorsStateCounter.Add(ctx, 1, metric.WithAttributes(unsafeAttribute(false)))
```
moshe-blox, line 310: i think there could be a race condition case where the validator was unsafe to begin with, but by the time the liveness request finished we've already observed quorum for it and incremented the metric, only to then increment it again here

wdyt?

moshe-blox: if that's true, i wouldn't rush to solve this

i think we can come up with something which fundamentally prevents this and similar issues

y0sher: Maybe something simpler like just updating all the states in every slot..

olegshmuelov: Good catch, I switched to `Gauge` and now update all states once per slot to avoid this race.

doppelganger/observability.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/ssvlabs/ssv/observability"
+)
+
+const (
+	observabilityName      = "github.com/ssvlabs/ssv/operator/doppelganger"
```
oleg-ssvlabs, line 14: The `observabilityName` variable typically reflects the full package path, including the Go module. I believe it should be `github.com/ssvlabs/ssv/doppelganger` instead.

doppelganger/observability.go
```diff
@@ -0,0 +1,37 @@
...
+		),
+	)
+)
+
+func metricName(name string) string {
+	return fmt.Sprintf("%s.%s", observabilityNamespace, name)
+}
+
+func unsafeAttribute(isUnsafe bool) attribute.KeyValue {
+	const attrName = observabilityNamespace + ".unsafe"
```
oleg-ssvlabs, line 35: I’m not sure the attribute name is descriptive enough. From what I understand, it relates to the validator object. So maybe something like `ssv.doppelganger.validator.unsafe`? Or, if this attribute is more generic and not strictly tied to the doppelganger feature, it could be just `ssv.validator.unsafe` or `ssv.validator.is_unsafe`

doppelganger/observability.go
olegshmuelov: You're right, updated it to `github.com/ssvlabs/ssv/doppelganger` to match the actual package path.

doppelganger/observability.go
olegshmuelov: Good point
updated to `ssv.doppelganger.validators.unsafe` for clarity and consistency with the metric name.

y0sher (approved):
looks and much simpler than previous version

doppelganger/doppelganger.go
```diff
@@ -309,6 +315,20 @@ func (h *handler) resetDoppelgangerStates() {
...
 
+func (h *handler) recordValidatorStates(ctx context.Context) {
+	var safe, unsafe uint64
+	for _, state := range h.validatorsState {
+		if state.safe() {
+			safe++
+		} else {
+			unsafe++
+		}
+	}
```
moshe-blox, line 326: aren't we missing a lock here around the loop?

y0sher: good catch! I think you're right, I was sure we are locking for each slot, but it's inside the functions.

olegshmuelov: fixed

y0sher (approved):

doppelganger/doppelganger.go
```diff
@@ -309,6 +315,23 @@ func (h *handler) resetDoppelgangerStates() {
...
+	defer h.mu.RUnlock()
+
+	var safe, unsafe uint64
+	for _, state := range h.validatorsState {
+		if state.safe() {
+			safe++
+		} else {
+			unsafe++
+		}
+	}
```
moshe-blox, line 329: i suggest this to free the lock sooner for potential signing requests:
```suggestion
	var safe, unsafe uint64
	func() {
		h.mu.RLock()
		defer h.mu.RUnlock()

		for _, state := range h.validatorsState {
			if state.safe() {
				safe++
			} else {
				unsafe++
			}
		}
	}()
```

moshe-blox (approved):
great edits, looks much simpler now 👌

PR #2093, opened 2025-03-18
fix(exporter): remove global lock for committee observer

anatolie-ssv:


protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -214,7 +220,7 @@ type validatorIndexAndRoot struct {
 	Root           phase0.Root
 }
 
-func (ncv *CommitteeObserver) processMessage(
+func (ncv *CommitteeObserver) verifySigAndgetQuorums(
```
iurii-ssv, line 223: `verifySigAndgetQuorums` -> `verifySigAndGetQuorums`

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -29,6 +30,8 @@ import (
 )
 
 type CommitteeObserver struct {
+	sync.Mutex
```
iurii-ssv, line 33: Could you explain why exactly we need this lock here (preferably in a comment above it) ?

PR #2094, opened 2025-03-18
chore: increase test coverage 

kchojn:


y0sher (approved):
lgtm

api/server/server.go
```diff
@@ -14,15 +14,19 @@ import (
...
+// Server represents the HTTP API server for SSV.
 type Server struct {
 	logger *zap.Logger
 	addr   string
 
 	node       *handlers.Node
 	validators *handlers.Validators
 	exporter   *handlers.Exporter
+
+	httpServer *http.Server
```
iurii-ssv, line 26: I'm not sure why you need to store `httpServer` as a dedicated field on `Server` struct (maybe for testing?) - I would try and see if we can get rid of it to keep it simple

api/server/server.go
```diff
@@ -39,7 +43,24 @@ func New(
 	}
 }
 
+// Run starts the server and blocks until it's shut down.
 func (s *Server) Run() error {
+	router := s.setupRouter()
```
iurii-ssv, line 48: Given that `Run` is small enough I would inline the contents of `setupRouter` here because jumping from func to func add some cognitive overhead as well ... 

That is to say, I think a good rule to follow (regarding func size):
- if func is too small consider inlining it (unless it's called multiple times)
- if func doesn't fit on ~ 1 screen consider breaking it up

api/handling.go
```diff
@@ -23,7 +23,7 @@ func Handler(h HandlerFunc) http.HandlerFunc {
 			// wrapped inside another error should error, not render.
 			switch e := err.(type) {
 			case render.Renderer:
-				if err := render.Render(w, r, e); err != nil {
+				if err := e.Render(w, r); err != nil {
 					http.Error(w, err.Error(), http.StatusInternalServerError)
 				}
 			default:
```
iurii-ssv, line 29: I'm not sure this comment above is relevant anymore:
```
//nolint:errorlint
			// errors.As would be incorrect here since a renderer.Renderer
			// wrapped inside another error should error, not render.
```
perhaps we can get rid of it and simplify this whole block of code to this (tests and linter don't seem to complain, but I haven't exactly verified whether or not tests cover this part of code) ?
```go
var errRenderer render.Renderer
if errors.As(err, &errRenderer) {
  if err := errRenderer.Render(w, r); err != nil {
    http.Error(w, err.Error(), http.StatusInternalServerError)
  }
  return
}
if err := render.Render(w, r, Error(err)); err != nil {
  http.Error(w, err.Error(), http.StatusInternalServerError)
}
```

iurii-ssv (commented):
Good job! Just minor suggestions from me.

> QUESTION: Is this correct behavior?
> * https://github.com/ssvlabs/ssv/blob/f2c54fb4e7d77aee10572670c4cccec30cfcfa70/storage/kv/txn.go#L57

I would rather return `return basedb.Obj{}, false, err` instead - but it probably doesn't matter since we must check and handle error anyway (disregarding other params as "undefined")

> FIXED: Someone please confirm, but this fix addresses the issue where an empty JSON string ("") is being unmarshalled to nil instead of an empty Hex{} slice.
> * https://github.com/ssvlabs/ssv/blob/7131dfde551e3389c97692990e563d1dd3dd2248/api/types.go#L27-L31

It's likely not an issue because `Hex` type doesn't have any methods relying on it not to be `nil` ... but I agree initializing it to empty slice is safer in general. That said, I think we should do this in `Bind` method (not `UnmarshalJSON`).

While at it, let's maybe change receiver to be a **pointer** here `func (h Hex) MarshalJSON()` -> `func (h *Hex) MarshalJSON()` for consistency ?

api/server/server.go
kchojn: yes, it makes testing easier, but it also gives us a lot of options, e.g.: if we would like to implement graceful shutdown, common practice in go:
- https://github.com/Shopify/toxiproxy/blob/6b13594aaf323c1bd7dcc1474705ef5cd7e039c6/api.go#L37
- https://github.com/devtron-labs/devtron/blob/22ea7618abd42d8cdc0ebf54e427a5d87e46587b/App.go#L56

previous implementation was untestable

api/server/server.go
kchojn: I'll revert it back, but I prefer more solutions where responsibility flows down to specific functions (more SRP). So the setup of the whole router should remain a separate function - to keep `Run()` focused only on server lifecycle.
A separate function makes it easy to extend functionality in the future + is more testable imo

api/handling.go
kchojn: I checked and confirmed, its no longer relevant, refactored!

kchojn:
> Good job! Just minor suggestions from me.
> 
> > QUESTION: Is this correct behavior?
> > 
> > * https://github.com/ssvlabs/ssv/blob/f2c54fb4e7d77aee10572670c4cccec30cfcfa70/storage/kv/txn.go#L57
> 
> I would rather return `return basedb.Obj{}, false, err` instead - but it probably doesn't matter since we must check and handle error anyway (disregarding other params as "undefined")
> 
> > FIXED: Someone please confirm, but this fix addresses the issue where an empty JSON string ("") is being unmarshalled to nil instead of an empty Hex{} slice.
> > 
> > * https://github.com/ssvlabs/ssv/blob/7131dfde551e3389c97692990e563d1dd3dd2248/api/types.go#L27-L31
> 
> It's likely not an issue because `Hex` type doesn't have any methods relying on it not to be `nil` ... but I agree initializing it to empty slice is safer in general. That said, I think we **should** do this in `Bind` method (not `UnmarshalJSON`).
> 
> While at it, let's maybe change receiver to be a **pointer** here `func (h Hex) MarshalJSON()` -> `func (h *Hex) MarshalJSON()` for consistency ?

--- 

> It's likely not an issue because Hex type doesn't have any methods relying on it not to be nil ... but I agree initializing it to empty slice is safer in general. That said, I think we should do this in Bind method (not UnmarshalJSON).

Good point, fixed!

> While at it, let's maybe change receiver to be a **pointer** here `func (h Hex) MarshalJSON()` -> `func (h *Hex) MarshalJSON()` for consistency ?

We keep the receiver as a value since Marshal doesntt modify the object. Thus, consistency here favors clarity of intent over matching pointer based Unmarshal method (https://pkg.go.dev/encoding/json)

iurii-ssv:
> We keep the receiver as a value since Marshal doesntt modify the object. Thus, consistency here favors clarity of intent over matching pointer based Unmarshal method (https://pkg.go.dev/encoding/json)

I agree it makes sense to "pass-by-value" when we don't intend to modify the object,

but my IDE complains about different receiver style ... and for a good reason - having both pointer and value receivers could be error-prone in general - so I would say "pass-by-value" doesn't apply as much to **struct-methods** such as this one (most methods on structs often are/should be "pointer-receivers")

api/server/server_test.go
```diff
@@ -0,0 +1,379 @@
...
+		ReadTimeout:       12 * time.Second,
+		WriteTimeout:      12 * time.Second,
+	}
+
+	require.NotNil(t, srv.httpServer)
+	require.Equal(t, ":8080", srv.httpServer.Addr)
+	require.Equal(t, 10*time.Second, srv.httpServer.ReadHeaderTimeout)
+	require.Equal(t, 12*time.Second, srv.httpServer.ReadTimeout)
+	require.Equal(t, 12*time.Second, srv.httpServer.WriteTimeout)
+	require.Equal(t, handler, srv.httpServer.Handler)
```
iurii-ssv, line 181: It seems these assertions check the code you wrote **in the test itself** - so they are kind of pointless ?

api/server/server.go
iurii-ssv: > A separate function makes it easy to extend functionality in the future + is more testable imo

Totally agree, if you need to move it in a separate function to be able to test it - go for it

> I prefer more solutions where responsibility flows down to specific functions (more SRP)

Also agree with "code needs to be structured such that abstraction layers aren't mixed together" (as per clean code book) - but I think that mostly applies to more complex code because each abstraction has some added cognitive cost

plus it's often too hard to organize your code to follow that rule specifically because often there isn't a clear separation between different abstractions (or different layers) - and different people might "classify" abstractions slightly differently

all it is to say we probably don't want to dwell on it too much ^ but rather loosely follow this rule as a nice guiding principle

api/server/server_test.go
kchojn: yep, I refactored the server 2-3 times, and they were from the previous test with the ListendAndServer func replaced. The test is useless, deleted, thx

api/server/server.go
kchojn: This is true; that is why I wasnt tied to my previous implementation. Marking as resolved, thanks for the additional point of view, always happy to read valuable observations

kchojn:
> > We keep the receiver as a value since Marshal doesntt modify the object. Thus, consistency here favors clarity of intent over matching pointer based Unmarshal method (https://pkg.go.dev/encoding/json)
> 
> I agree it makes sense to "pass-by-value" when we don't intend to modify the object,
> 
> but my IDE complains about different receiver style ... and for a good reason - having both pointer and value receivers could be error-prone in general - so I would say "pass-by-value" doesn't apply as much to **struct-methods** such as this one (most methods on structs often are/should be "pointer-receivers")

@iurii-ssv i will leave it as it is (original), there is an additional type that uses `Hex`, changing to pointer will also cause me to have to add Marshal/Unmarshal for it (it doesn't have custom methods now)

iurii-ssv (approved):

liorrutenberg:
@kchojn - notice the lint is failing, please fix it

kchojn:
> @kchojn - notice the lint is failing, please fix it

I've rebased my branch onto origin/stage, should be fixed now!

iurii-ssv:
Btw @kchojn pls use `git merge ...` instead of `git rebase ... ` when you can - it keeps diff reasonably small for reviewers to verify the changes are good

kchojn:
> Btw @kchojn pls use `git merge ...` instead of `git rebase ... ` when you can - it keeps diff reasonably small for reviewers to verify the changes are good

thanks @iurii-ssv , I have reversed `rebase`, from now on I will use `merge`

PR #2098, opened 2025-03-24
chore: increase test coverage 2

kchojn:


observability/observability.go
```diff
@@ -3,40 +3,42 @@ package observability
...
+	resources, err := deps.ResourceMerge(resource.Default(), resource.NewWithAttributes(
 		semconv.SchemaURL,
 		semconv.ServiceName(appName),
 		semconv.ServiceVersion(appVersion),
 	))
+
 	if err != nil {
-		err = errors.Join(errors.New("failed to instantiate observability resources"), err)
-		return shutdown, err
+		return shutdown, errors.Join(errors.New("failed to instantiate observability resources"), err)
```
iurii-ssv, line 31: Can't we just use `fmt.Errorf` (instead of `errors.Join`) here and in similar places below?

observability/observability.go
```diff
@@ -3,40 +3,42 @@ package observability
...
 	}
 
-	if config.metricsEnabled {
+	if cfg.metricsEnabled {
 		var promExporter *prometheus.Exporter
-		promExporter, err = prometheus.New()
+		promExporter, err = deps.PrometheusNew()
+
```
iurii-ssv, line 37: We probably don't need a newline between `promExporter, err = deps.PrometheusNew()` and the line where `err` is handled

utils/threadsafe/bytes_test.go
```diff
@@ -0,0 +1,71 @@
+package threadsafe
```
iurii-ssv, line 1: It doesn't seem like we are using `package threadsafe` anywhere ? Unless it's something imported as go-module in other projects (and most of it can be replaced by `atomic` package anyway).

@moshe-blox @y0sher should we remove it ?

observability/attributes_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+// TestRunnerRoleAttribute_Creates verifies RunnerRoleAttribute creates correct key-value attribute.
+func TestRunnerRoleAttribute_Creates(t *testing.T) {
+	t.Parallel()
+
+	role := types.RunnerRole(1)
+
+	attr := RunnerRoleAttribute(role)
+
+	require.Equal(t, RunnerRoleAttrKey, string(attr.Key))
```
iurii-ssv, line 34: Maybe we don't need `RunnerRoleAttrKey` as a public constant (eg. we don't have one for `ssv.beacon.role` ... and that bugs me a bit)

namely it seems the only reason we have it is so we can use it in this func:
```go
func roleAttribute(role string) attribute.KeyValue
```

but I think this func itself ^ can go and be replaced by `observability.RunnerRoleAttribute` ?

iurii-ssv (commented):
Mostly LGTM (left minor comments)

utils/threadsafe/bytes_test.go
kchojn: I confrim, We dont use it anywhere, lmk and I'll remove the unnecessary code

observability/attributes_test.go
iurii-ssv: Lets ask @oleg-ssvlabs as the original designer of these - I haven't looked deep into this myself, just think we can make these functions follow same kind of convention

oleg-ssvlabs: > Lets ask @oleg-ssvlabs (as the original designer of these) - I haven't looked deep into this myself, just think we can make these functions follow same kind of convention - what do you think of it Oleg ?

The attribute key was exposed because we had one place where the runner role was a type of string and, it seems, could not be converted to `types.RunnerRole` (ssv-spec types package), basically what @kchojn describes above. 
As an option we could do something like this if that works:

```Go
var name spectypes.RunnerRole
if len(identifier) == 56 {
	name = spectypes.MessageID(identifier).GetRoleType()
} else {
	name = spectypes.RoleUnknown
} 
```
Then we wouldn't need to expose this attribute key

protocol/v2/qbft/instance/instance.go
```diff
@@ -39,11 +38,12 @@ func NewInstance(
...
 ) *Instance {
-	var name string
+	var name spectypes.RunnerRole
 	if len(identifier) == 56 {
-		name = spectypes.MessageID(identifier).GetRoleType().String()
+		name = spectypes.MessageID(identifier).GetRoleType()
 	} else {
-		name = base64.StdEncoding.EncodeToString(identifier)
+		// base64.StdEncoding.EncodeToString(identifier)
+		name = spectypes.RoleUnknown
```
kchojn, line 46: @y0sher this was implemented by you, so lmk if it looks (I use a specific type, instead of string)

utils/cliflag/cliflag.go
```diff
@@ -6,29 +6,30 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const requiredSuffix = " (required)"
```
iurii-ssv, line 9: Could probably be a local const (to keep the usage scope as small as possible) ?

protocol/v2/qbft/instance/instance.go
iurii-ssv: Since `name` is for metrics only I guess it's OK to do it this way (otherwise we'd probably want to error or log this)

iurii-ssv (approved):
LGTM

oleg-ssvlabs (approved):
Good job @kchojn 
I am just not 100% sure about `attributes_test.go`. I mean it's good to have them, but probably not really necessary. 
Otherwise tests are very clean and easy to read

nkryuchkov (approved):
great job

PR #2101, opened 2025-03-25
feat: Add AMD64/ARM64 docker support 

kchojn:
This PR adds multi arch Docker build capabilities (`Dockerfile.multiarch`), allowing us to build and run the node on different platforms (`amd64, arm64, arm`) from a single Dockerfile.

Key differences from the standard `Dockerfile`:
- Added cross-compilation toolchains for arm64 and arm architectures
- Implemented dynamic compiler selection based on target architecture
- Used Docker's BuildKit platform variables (BUILDPLATFORM, TARGETPLATFORM, TARGETARCH)


tools/cross-compiler/README.md
```diff
@@ -0,0 +1,82 @@
...
+1. Create a Docker BuildKit builder
+2. Build the SSV node for both amd64 and arm64 architectures
+3. Push the resulting multi-architecture image
+
+### Environment Variables
+
+The build process requires the following environment variables:
+
+- `IMAGE_NAME`: The full name and tag for the Docker image to be built and pushed
+- `SOURCE_COMMIT`: Git commit to be used as the image label
```
iurii-ssv, line 42: If these vars are only relevant to `Building Multi-Architecture Docker Images` section maybe we can put it in there (no need for a dedicated section) ?

tools/cross-compiler/README.md
```diff
@@ -0,0 +1,82 @@
...
+    - `build.sh` - Builds the SSV node for a specific target architecture
+
+## Usage
+
+### Building Multi-Architecture Docker Images
+
+To build multi-architecture Docker images, run:
+
+```bash
+IMAGE_NAME=your-registry/your-image:tag SOURCE_COMMIT=$(git rev-parse HEAD) ./hooks/build
```
iurii-ssv, line 28: maybe this should be a command in Makefile instead ?

kchojn: Gonna keep this comment, Im waiting for Vaclav to confirm the solution, then will add it!

hooks/build
```diff
@@ -1,10 +1,53 @@
...
 export DOCKER_BUILDKIT=1
 
-# Extract the Git tag from the SOURCE_COMMIT
+# Platforms to build for
+PLATFORMS="linux/amd64,linux/arm64"
+
+# Check if SOURCE_COMMIT is set
+if [ -z "${SOURCE_COMMIT}" ]; then
+  echo "SOURCE_COMMIT not set. Unable to determine version."
+fi
```
olegshmuelov, line 13: If `SOURCE_COMMIT` is required for versioning, consider exiting early instead of just printing a warning:
```bash
if [ -z "${SOURCE_COMMIT}" ]; then
  echo "SOURCE_COMMIT not set. Aborting."
  exit 1
fi
```
Or fallback to something like `v0.0.0-dev` if local builds without it are expected.

Dockerfile
```diff
@@ -1,49 +1,50 @@
...
+# Install dependencies for cross-compilation
+RUN /tmp/scripts/install_deps.sh
+
+# Create compiler wrappers to handle cross-compilation flags
+RUN /tmp/scripts/create_wrappers.sh
 
 WORKDIR /go/src/github.com/ssvlabs/ssv/
-COPY go.mod .
-COPY go.sum .
+COPY go.mod go.sum ./
```
y0sher, line 17: wouldn't `./` include `go.sum` and `go.mod`?

kchojn: optimization pattern - it copies only the dependency files first, runs the mod download, and then copies the rest of the code later. This uses layer caching, so if deps remain the same, we don't need to re download them every time

Dockerfile
```diff
@@ -1,49 +1,50 @@
...
+  /tmp/scripts/build.sh "${TARGETARCH}" /go/bin/ssvnode
+
+# Show the built binary for verification
+RUN ls -la /go/bin/
 
 #
-# STEP 3: Prepare image to run the binary
+# STEP 3: Prepare minimal image to run the binary
 #
 FROM golang:1.22 AS runner
```
oleg-ssvlabs, line 49: It is better to use `scratch` images as they are slimmer (if possible) 
There’s a pretty good Dockerfile [here](https://github.com/ssvlabs/ssv-pulse/blob/main/build/Dockerfile) — you might be able to find some inspiration there.

Dockerfile
```diff
@@ -1,69 +1,73 @@
...
+    COMMIT=$(git rev-parse HEAD) && \
+    VERSION=$(git describe --tags $(git rev-list --tags --max-count=1) --always) && \
+    CGO_ENABLED=1 GOOS=linux GOARCH=${TARGETARCH} go install -tags="blst_enabled" \
+      -ldflags "-X main.Commit=$COMMIT -X main.Version=$VERSION -linkmode external -extldflags \"-static -lm\"" \
+      ./cmd/ssvnode
+
+#
+# STEP 3: Prepare runtime dependencies
+#
+FROM debian:stable-slim AS deps
```
vaclav-ssvlabs, line 44: Let's set to tag version like `1.2.3`, having tags like `stable` and `latest` are dangerous, as when there is an OS version update whole build can crash without us changing anything

Dockerfile
```diff
@@ -1,69 +1,73 @@
...
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+      dnsutils ca-certificates && \
+    rm -rf /var/lib/apt/lists/*
 
 #
-# STEP 3: Prepare image to run the binary
+# STEP 4: Final minimal image
 #
-FROM golang:1.22 AS runner
+FROM scratch AS runner
```
vaclav-ssvlabs, line 54: Why are you changing the base? I am afraid that  some commands won't work anymore - for example, we recommend starting via `make start-node`, which runs this command - https://github.com/ssvlabs/ssv/blob/main/Makefile#L7

The `dig` command won't be in `scratch` base image, and it will fail. And also some other users, including us are often using `exec` into the container and `scratch` doesn't have any shell, so it won't be possible. This has many other caveats and issues. Changing the base image with something that is already heavily in use is not easy.


Let's not change the base OS of the image to something else.

Dockerfile.multiarch
```diff
@@ -0,0 +1,94 @@
...
+# Global build arguments
+#
+ARG BUILDPLATFORM
+ARG TARGETPLATFORM
+ARG TARGETARCH
+
+#
+# STEP 1: Get cross-compilation tools from tonistiigi/xx
+#
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:latest AS xx
```
vaclav-ssvlabs, line 11: Using a third-party unknown source of Docker image can be very dangerous, especially when you use the latest tag

As this is Docker image for a validator client that includes running with private keys, if the guy sees this and includes some backdoor in the base Docker image  and releases it we could be releasing an image with the backdoor  and potentionaly go backgrupt.

If there is no official image from the distros, then maybe we should rather just  use the base image and install if it's possible

Looking at this guy's Docker Hub, which has only 3 stars and the github repo ,I don't think its safe to use it.

kchojn: Thanks Vaclav, I dropped all external deps, from now on, the multiarch Dockerfile is simple

vaclav-ssvlabs (approved):

moshe-blox:
@kchojn was this tested?

if not, can we please create a temporary Docker Hub repo hooked to this GitHub repo and branch and test that it builds & pushes multiple images and we can run them successfully?

y0sher (approved):

iurii-ssv (approved):
LGTM if it's working

hooks/push
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo "Push is taken care of by buildx, during the build phase"
```
y0sher, line 3: do we need this file? maybe without this file it will push automatically? (and not use `--push`)

kchojn: I took this from other repos. Actually without this 
, it did not work, only overwriting the push helped. Buildx during hooks behaves a slightly differently

moshe-blox: sounds good then, thanks

PR #2106, opened 2025-03-27
perf: weighted attestation data enhancements

oleg-ssvlabs:
This PR addresses several edge cases that may occur during the weighted attestation data flow, particularly those related to scoring.

- The singleflight lock on block root lookups was removed, allowing all clients to retrieve the slot by block root in parallel. This change increases the likelihood of successfully fetching the slot, which is critical for scoring. A vulnerability in the previous implementation was that if one client timed out, other clients might not get a chance to perform the same lookup due to the soft timeout being reached.

- Retry logic was introduced for fetching the slot by block root. This addition is important, as the Block event often arrives just a few milliseconds after the initial scoring attempt, which previously led to slot being not included in the score.

- The timeout for the `BeaconBlockHeader` call was reduced to 1/4 of the soft timeout. The goal is to prevent a single `BlockHeader` call from consuming the entire timeout budget. 

- The system now listens to **Block** events instead of **Head** events when the weighted feature is enabled. This aligns the implementation more closely with Vouch’s behavior.

Other changes:
- Parallel submissions of: Sync Committee Subscriptions, Blinded Proposals, Proposal Preparations (some of them are only when new flag `withParallelSubmissions` is enabled)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ UNFORMATTED=$(shell gofmt -l .)
 .PHONY: lint-prepare
 lint-prepare:
 	@echo "Preparing Linter"
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.64.8
 
 .PHONY: lint
 lint:
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -352,7 +352,7 @@ func (gc *GoClient) scoreAttestationData(ctx context.Context,
 	defer ticker.Stop()
 
 	var (
-		retries uint8
+		retries uint32
 		start   = time.Now()
 	)
 
@@ -378,14 +378,14 @@ func (gc *GoClient) scoreAttestationData(ctx context.Context,
 		select {
 		case <-ctx.Done():
 			logger.
-				With(zap.Uint8("retry_count", retries)).
+				With(zap.Uint32("try", retries)).
 				With(zap.Duration("total_elapsed", time.Since(start))).
 				Error("timeout for obtaining slot for block root was reached. Returning base score")
 			return score
 		case <-ticker.C:
 			retries++
 			logger.
-				With(zap.Uint8("retry_count", retries)).
+				With(zap.Uint32("try", retries)).
 				With(zap.Duration("total_elapsed", time.Since(start))).
 				Warn("retrying to obtain slot for block root")
 		}
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -432,10 +432,7 @@ func (gc *GoClient) blockRootToSlot(ctx context.Context, client Client, root pha
 }
 
 func isBlockHeaderResponseValid(response *api.Response[*eth2apiv1.BeaconBlockHeader]) bool {
-	if response == nil || response.Data == nil || response.Data.Header == nil || response.Data.Header.Message == nil {
-		return false
-	}
-	return true
+	return response != nil && response.Data != nil && response.Data.Header != nil && response.Data.Header.Message != nil
 }
 
 func weightedAttestationDataRequestIDField(id uuid.UUID) zap.Field {
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -320,6 +320,8 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
 		return
 	}
 
+	logger = logger.With(fields.BlockRoot(attestationData.BeaconBlockRoot))
+
 	logger.Debug("scoring attestation data")
 	score := gc.scoreAttestationData(ctx, client, attestationData, logger)
 
@@ -337,10 +339,6 @@ func (gc *GoClient) scoreAttestationData(ctx context.Context,
 	attestationData *phase0.AttestationData,
 	logger *zap.Logger,
 ) float64 {
-	logger = logger.With(
-		fields.BlockRoot(attestationData.BeaconBlockRoot),
-		zap.Uint64("attestation_data_slot", uint64(attestationData.Slot)))
-
 	// Initial score is based on height of source and target epochs.
 	score := float64(attestationData.Source.Epoch + attestationData.Target.Epoch)
 	logger.
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -143,7 +143,7 @@ func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.Attestati
 				zap.Int("errored", errored),
 			).Debug("response received")
 
-			if resp.score > bestScore {
+			if bestAttestationData == nil || resp.score > bestScore {
 				if bestAttestationData != nil {
 					logger.Info("updating best attestation data because of higher score",
 						zap.String("client_addr", resp.clientAddr),
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sync/atomic"
 	"time"
 
 	"github.com/attestantio/go-eth2-client/api"
@@ -12,6 +13,7 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/google/uuid"
 	"github.com/jellydator/ttlcache/v3"
+	"github.com/sourcegraph/conc/pool"
 	"github.com/ssvlabs/ssv/logging/fields"
 	"go.uber.org/zap"
 )
@@ -437,7 +439,53 @@ func weightedAttestationDataRequestIDField(id uuid.UUID) zap.Field {
 	return zap.String("weighted_data_request_id", id.String())
 }
 
-// SubmitAttestations implements Beacon interface
+// multiClientSubmit is a generic function that submits data to multiple clients concurrently.
+// If any client succeeds, the remaining submissions are cancelled.
+// Returns nil if at least one client successfully submitted the data.
+func (gc *GoClient) multiClientSubmit(
+	operationName string,
+	submitFunc func(ctx context.Context, client Client) error,
+) error {
+	logger := gc.log.With(zap.String("api", operationName))
+
+	submissions := atomic.Int32{}
+	p := pool.New().WithErrors().WithContext(gc.ctx).WithMaxGoroutines(len(gc.clients))
+	for _, client := range gc.clients {
+		client := client
+		p.Go(func(ctx context.Context) error {
+			clientAddress := client.Address()
+			logger := logger.With(zap.String("client_addr", clientAddress))
+
+			start := time.Now()
+			err := submitFunc(ctx, client)
+			recordRequestDuration(ctx, operationName, clientAddress, http.MethodPost, time.Since(start), err)
+			if err != nil {
+				logger.Debug("a client failed to submit",
+					zap.Error(err))
+				return fmt.Errorf("client %s failed to submit %s: %w", clientAddress, operationName, err)
+			}
+
+			logger.Debug("a client submitted successfully")
+
+			submissions.Add(1)
+			return nil
+		})
+	}
+	err := p.Wait()
+	if submissions.Load() > 0 {
+		// At least one client has submitted successfully,
+		// so we can return without error.
+		return nil
+	}
+	if err != nil {
+		logger.Error("all clients failed to submit",
+			zap.Error(err))
+		return fmt.Errorf("failed to submit %s", operationName)
+	}
+	return nil
+}
+
+// SubmitAttestations implements Beacon interface and sends attestations to the first client that succeeds
 func (gc *GoClient) SubmitAttestations(attestations []*spec.VersionedAttestation) error {
 	clientAddress := gc.multiClient.Address()
 	logger := gc.log.With(
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -267,8 +267,8 @@ func (gc *GoClient) simpleAttestationData(slot phase0.Slot) (*phase0.Attestation
 	logger.With(
 		zap.Duration("elapsed", time.Since(attDataReqStart)),
 		zap.Bool("with_weighted_attestation_data", false),
-		zap.String("client_addr", gc.multiClient.Address),
-		fields.BlockRoot(resp.Data.BeaconBlockRoot)
+		zap.String("client_addr", gc.multiClient.Address()),
+		fields.BlockRoot(resp.Data.BeaconBlockRoot),
 	).Debug("successfully fetched attestation data")
 
 	return resp.Data, nil
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -353,7 +353,7 @@ func (gc *GoClient) scoreAttestationData(ctx context.Context,
 	score := float64(attestationData.Source.Epoch + attestationData.Target.Epoch)
 	logger.
 		With(zap.Float64("base_score", score)).
-		Info("base score was set. Fetching slot for block root")
+		Debug("base score was set. Fetching slot for block root")
 
 	ctx, cancel := context.WithTimeout(ctx, gc.weightedAttestationDataSoftTimeout/2)
 	defer cancel()
@@ -447,8 +447,7 @@ func weightedAttestationDataRequestIDField(id uuid.UUID) zap.Field {
 	return zap.String("weighted_data_request_id", id.String())
 }
 
-// multiClientSubmit is a generic function that submits data to multiple clients concurrently.
-// If any client succeeds, the remaining submissions are cancelled.
+// multiClientSubmit is a generic function that submits data to multiple beacon clients concurrently.
 // Returns nil if at least one client successfully submitted the data.
 func (gc *GoClient) multiClientSubmit(
 	operationName string,
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -487,13 +487,20 @@ func (gc *GoClient) multiClientSubmit(
 
 // SubmitAttestations implements Beacon interface and sends attestations to the first client that succeeds
 func (gc *GoClient) SubmitAttestations(attestations []*spec.VersionedAttestation) error {
+	opts := &api.SubmitAttestationsOpts{Attestations: attestations}
+	if gc.withParallelSubmissions {
+		return gc.multiClientSubmit("SubmitAttestations", func(ctx context.Context, client Client) error {
+			return client.SubmitAttestations(gc.ctx, opts)
+		})
+	}
+
 	clientAddress := gc.multiClient.Address()
 	logger := gc.log.With(
 		zap.String("api", "SubmitAttestations"),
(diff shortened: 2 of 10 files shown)


beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+		Debug("block root to slot cache updated from the BeaconBlockHeader call")
+
+	return slot, nil
+}
+
+func isBlockHeaderResponseValid(response *api.Response[*eth2apiv1.BeaconBlockHeader]) bool {
+	if response == nil || response.Data == nil || response.Data.Header == nil || response.Data.Header.Message == nil {
+		return false
+	}
+	return true
```
moshe-blox, line 438:
```suggestion
	return response != nil && response.Data != nil && response.Data.Header != nil && response.Data.Header.Message != nil
```

beacon/goclient/attest.go
```diff
@@ -143,7 +143,14 @@ func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.Attestati
 				zap.Int("errored", errored),
 			).Debug("response received")
 
-			if bestAttestationData == nil || resp.score > bestScore {
+			if resp.score > bestScore {
```
moshe-blox, line 146: what if score is 0?

moshe-blox, line 146: what if score is 0?

beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+		cachedSlot := cacheResult.Value()
+		logger.
+			With(zap.Uint64("cached_slot", uint64(cachedSlot))).
+			With(zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
+			Debug("obtained slot from cache")
+		return cachedSlot, nil
+	}
 
-		return 0, nil
+	logger.Debug("slot was not found in cache. Fetching from the client")
```
moshe-blox, line 408: aligning style to rest of the code:
```suggestion
	logger.Debug("slot was not found in cache, fetching from the client")
```

beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+				With(zap.Uint8("retry_count", retries)).
+				With(zap.Duration("total_elapsed", time.Since(start))).
+				Warn("retrying to obtain slot for block root")
 		}
+	}
+}
 
-		logger.Debug("slot was not found in cache, returning: '0'")
+func (gc *GoClient) blockRootToSlot(ctx context.Context, client Client, root phase0.Root, logger *zap.Logger) (phase0.Slot, error) {
+	cacheResult := gc.blockRootToSlotCache.Get(root)
```
moshe-blox, line 398: every log in the flow should have the block root
```suggestion
	logger := logger.With(fields.BlockRoot(root))
	cacheResult := gc.blockRootToSlotCache.Get(root)
```

beacon/goclient/attest.go
```diff
@@ -312,7 +321,7 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
 	}
 
 	logger.Debug("scoring attestation data")
-	score := gc.scoreAttestationData(ctx, attestationData, logger)
+	score := gc.scoreAttestationData(ctx, client, attestationData, logger)
```
moshe-blox, line 324: every log in the flow should show the block root as soon as we have it, let's add: `logger := logger.With(fields.BlockRoot(root))`

beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+			With(zap.Uint64("cached_slot", uint64(cachedSlot))).
+			With(zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
+			Debug("obtained slot from cache")
+		return cachedSlot, nil
+	}
 
-		return 0, nil
+	logger.Debug("slot was not found in cache. Fetching from the client")
+
+	timeoutContext, cancel := context.WithTimeout(ctx, gc.weightedAttestationDataSoftTimeout/4)
```
moshe-blox, line 410: i know it's been reduced significantly, but IMO lets go for `/5` for 1 second to leave 1 second for consensus to finish before change round

beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+			With(zap.Error(err)).
+			Warn("failed to obtain slot for block root")
+		select {
+		case <-ctx.Done():
 			logger.
-				With(zap.Uint64("cached_slot", uint64(cachedSlot))).
-				With(zap.Int("cache_len", gc.blockRootToSlotCache.Len())).
-				Debug("obtained slot from cache")
-			return cachedSlot, nil
+				With(zap.Uint8("retry_count", retries)).
```
moshe-blox, line 383: this can be called just `try` (as in try number) because it starts at 0

beacon/goclient/attest.go
```diff
@@ -324,53 +333,109 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
-		zap.Uint64("target_epoch", uint64(attestationData.Target.Epoch)),
-		zap.Float64("score", score),
-	).Debug("scored attestation data")
+	ticker := time.NewTicker(time.Millisecond * 100)
+	defer ticker.Stop()
 
-	return score
-}
+	var (
+		retries uint8
```
moshe-blox, line 357: This micro-optimization adds negligible value while introducing the risk of uint8 overflow—probably not worth the mental overhead.

moshe-blox:
@oleg-ssvlabs i made these changes so we can try them overnight, please see https://github.com/ssvlabs/ssv/pull/2108

beacon/goclient/attest.go
oleg-ssvlabs: Updated.  
Also removed the `attestation_data_slot` field from the logger because it’s always equal to the slot we request attestation data for, and that field is already present in the logger.

beacon/goclient/attest.go
oleg-ssvlabs: I see what you mean.
Changed to `uint32`

beacon/goclient/attest.go
oleg-ssvlabs: fixed

beacon/goclient/attest.go
oleg-ssvlabs: The soft timeout is equal to [common / 2](https://github.com/ssvlabs/ssv/blob/main/beacon/goclient/goclient.go#L217) (2.5 sec), meaning the timeout set here will be equal to `625ms` (not `1.25s` as you assumed)

beacon/goclient/attest.go
oleg-ssvlabs: reverted to the previous implementation to mainly remove tight coupling with the scoring feature

beacon/goclient/attest.go
moshe-blox: sry, my bad

for now we could at least reduce 2.5s to 2s to make it less bad when 1 bad BN times out very often

beacon/goclient/events.go
```diff
@@ -105,56 +105,60 @@ func (gc *GoClient) eventHandler(e *apiv1.Event) {
...
 				Debug("event slot is lower or equal than last processed slot")
-			gc.lastProcessedHeadEventSlotLock.Unlock()
+			gc.lastProcessedEventSlotLock.Unlock()
 			return
 		}
 
-		gc.lastProcessedHeadEventSlot = headEventData.Slot
-		gc.lastProcessedHeadEventSlotLock.Unlock()
+		gc.lastProcessedEventSlot = eventData.Slot
+		gc.lastProcessedEventSlotLock.Unlock()
```
moshe-blox, line 154: this should be done for head events also, otherwise we're emitting duplicates to listeners such as DutyScheduler

since there are no listeners to block events yet, then i think this potentially can be moved to head events rather than duplicated, otherwise it requires a new lastProcessedEventSlot field

Makefile
```diff
@@ -32,7 +32,7 @@ UNFORMATTED=$(shell gofmt -l .)
 .PHONY: lint-prepare
 lint-prepare:
 	@echo "Preparing Linter"
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.64.8
```
y0sher, line 35: is this the new or old version? because current stage is using latest version..

beacon/goclient/attest.go
```diff
@@ -144,6 +146,13 @@ func (gc *GoClient) weightedAttestationData(slot phase0.Slot) (*phase0.Attestati
 			).Debug("response received")
 
 			if bestAttestationData == nil || resp.score > bestScore {
+				if bestAttestationData != nil {
+					logger.Info("updating best attestation data because of higher score",
```
y0sher, line 150: I guess this should be `Debug`

Makefile
oleg-ssvlabs: It's latest V1. 
Stage uses V2. V2 migration would introduce a lot of changes. I don't know what is the relation between `rc/v2.3.0` and `stage` branch, but I do not think I can rebase `stage` into this branch and pull all the changes, since `stage` branch contains changes that should end up in `2.3.1` and not in `2.3.0`

beacon/goclient/attest.go
oleg-ssvlabs: fixed

beacon/goclient/events.go
oleg-ssvlabs: fixed. Please review the changes

beacon/goclient/attest.go
oleg-ssvlabs: Should the Soft/Hard timeouts for the Weighted still depend on DefaultCommonTimeout?
Currently, the Hard timeout is equal to DefaultCommonTimeout, and the Soft timeout is DefaultCommonTimeout / 2.

Would it make more sense to hardcode the Hard timeout for Weighted to 4 seconds, and set the Soft timeout to half of that (i.e., 2 seconds)?

Makefile
y0sher: so if v2.3.0 doesn't have the update to v2 we can leave it like this.

beacon/goclient/attest.go
oleg-ssvlabs: The block root is actually [there](https://github.com/ssvlabs/ssv/pull/2106/files/436a7cfa7a5845b9286f5e42140c1d20d311d344#diff-b6b7fb7789644004990ef2a0a1be46d921fcd4b6bf46869b67ff758a2aa4239eR153)

beacon/goclient/proposer.go
```diff
@@ -364,21 +297,9 @@ func (gc *GoClient) SubmitBeaconBlock(block *api.VersionedProposal, sig phase0.B
...
-	if err != nil {
-		logger.Error(clResponseErrMsg, zap.Error(err))
-		return err
-	}
-
-	logger.Debug("consensus client submitted beacon block")
-	return nil
+	return gc.multiClientSubmit("SubmitProposal", func(ctx context.Context, client Client) error {
+		return client.SubmitProposal(gc.ctx, opts)
+	})
```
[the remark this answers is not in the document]
moshe-blox: proposals are extremely rare, why do you think it should be a flag?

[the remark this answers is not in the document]
y0sher: Not a separate flag but under the same flag as attestations and sync committee. And I'm not sure it should be I just think in terms of UX that you enable a flag to send attestations and sync committee you don't know that proposals are sent parallel by default.

tbh it's just confusing in my opinion that the default is different for proposals and other duties. But it does makes sense because what you said about the rarity of proposals.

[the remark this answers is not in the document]
moshe-blox: the purpose of that flag isnt to give the user make advanced technical decisions like Vouch, but rather allow him to tell us he can "take on the extra load" for more perf (which is smthn most can understand)

exactly how its best to do that with the small details is for us to decide :)

our help msg for the flag needs to specify thay this is only for atts & sync committeee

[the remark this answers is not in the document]
moshe-blox: if we some day figure out there is benefit in giving more control here then we can split this flag

beacon/goclient/proposer.go
```diff
@@ -230,74 +228,9 @@ func (gc *GoClient) SubmitBlindedBeaconBlock(block *api.VersionedBlindedProposal
...
-		return nil
-	}
-	if err != nil {
-		logger.Error("no consensus clients have been able to submit blinded proposal. See adjacent logs for error details.")
-		return fmt.Errorf("no consensus clients have been able to submit blinded proposal")
-	}
-	return nil
+	return gc.multiClientSubmit("SubmitBlindedProposal", func(ctx context.Context, client Client) error {
+		return client.SubmitBlindedProposal(ctx, opts)
+	})
```
olegshmuelov, line 233: Not a regression, but worth revisiting:
`SubmitBlindedProposal` submits to all beacon clients, which was also true before this PR.
However, blinded blocks must be submitted to the node that produced them.

beacon/goclient/attest.go
```diff
@@ -324,68 +345,170 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
 	logger *zap.Logger,
 ) float64 {
-	logger = logger.With(
-		fields.BlockRoot(attestationData.BeaconBlockRoot),
-		zap.Uint64("attestation_data_slot", uint64(attestationData.Slot)))
 	// Initial score is based on height of source and target epochs.
 	score := float64(attestationData.Source.Epoch + attestationData.Target.Epoch)
+	logger.
+		With(zap.Float64("base_score", score)).
+		Info("base score was set. Fetching slot for block root")
```
olegshmuelov, line 356: This `.Info()` log for base score might be a bit noisy since it runs per client per attestation
maybe drop it to `.Debug()`?

beacon/goclient/attest.go
```diff
@@ -324,68 +345,170 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
+	return response != nil && response.Data != nil && response.Data.Header != nil && response.Data.Header.Message != nil
 }
 
 func weightedAttestationDataRequestIDField(id uuid.UUID) zap.Field {
 	return zap.String("weighted_data_request_id", id.String())
 }
 
-// SubmitAttestations implements Beacon interface
+// multiClientSubmit is a generic function that submits data to multiple clients concurrently.
+// If any client succeeds, the remaining submissions are cancelled.
```
olegshmuelov, line 451: This comment says remaining submissions are cancelled after the first success, but the code doesn't cancel the context, so all clients still run. Either cancel the context on success or update the comment to reflect current behavior

beacon/goclient/attest_test.go
```diff
@@ -317,6 +317,23 @@ func TestGoClient_GetAttestationData_Weighted(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("single beacon client: populates boolRootToSlot cache", func(t *testing.T) {
```
olegshmuelov, line 320: typo: `boolRootToSlot` -> `blockRootToSlot`

beacon/goclient/attest.go
```diff
@@ -324,68 +345,170 @@ func (gc *GoClient) fetchWeightedAttestationData(ctx context.Context,
...
-func (gc *GoClient) blockRootToSlot(root phase0.Root, logger *zap.Logger) (phase0.Slot, error) {
-	slot, err, _ := gc.blockRootToSlotReqInflight.Do(root, func() (phase0.Slot, error) {
-		cacheResult := gc.blockRootToSlotCache.Get(root)
-		if cacheResult != nil {
-			cachedSlot := cacheResult.Value()
+	for {
+		slot, err := gc.blockRootToSlot(ctx, client, attestationData.BeaconBlockRoot, logger)
+		if err == nil {
+			// Increase score based on the nearness of the head slot.
+			score += float64(1) / float64(1+attestationData.Slot-slot)
```
olegshmuelov, line 373: There's a small risk of division by zero here:
`score += 1 / (1 + attestationData.Slot - slot)`
`If slot == attestationData.Slot + 1`, the denominator becomes 0 and will panic.

beacon/goclient/attest.go
oleg-ssvlabs: @y0sher do you think we should rather change the implementation?

beacon/goclient/attest.go
moshe-blox: > It isn’t really possible for slot to be equal to attestationData.Slot + 1.
even if its true, still worth to assert it (return err if slot is higher than attestationData.Slot) rather than panic, in case theres a bug elsewhere that caused this to happen

olegshmuelov: Since `blockRootToSlot(...)` isn't scoped strictly to the attestation data endpoint (e.g. it can pull from cache or from a delayed `Block` event), there's still a theoretical risk of getting `slot > attestationData.Slot` due to clock drift, or stale cache.
It could also happen during fork reorgs if the block root maps to a later slot on the new chain if I understand correctly.

oleg-ssvlabs: I agree, it is better to guard ourselves from this. 
I wonder what we should do in case we run into this scenario. Should we skip calculating the score (keep only "base" score) or if block root points to the slot in the future hardcode it to `1`? `1` suppose to be a maximum in this `float64(1) / float64(1+attestationData.Slot-slot)` calculation, it shouldn't be negative nor, of course, panic.
Wdyt @olegshmuelov @moshe-blox ?

olegshmuelov: One small suggestion, if we ever see this log in production, we'll probably want to reproduce it. Maybe worth adding more fields to the log, like `attestationData.Slot`, the returned `slot`, and the `BeaconBlockRoot`, to help debug where the drift is coming from.
Also there is a tiny typo in the log message: "score was not be updated" -> "score was not updated".

y0sher (approved):

beacon/goclient/attest.go
moshe-blox: @oleg-ssvlabs i think it makes sense that hard timeout matches the timeout in the simple flow (DefaultCommonTimeout) to give as much time as it does (in the worst case scenario)

if you dont like converting to floats, maybe you can just divide it by 3, i'm perfectly fine with a 1.666s soft timeout if you are :)

in the future, i think the soft timeout should be configurable rather than relative, as it has an entirely different purpose

for example: as an operator, i might be willing to wait 5s for any attestation at all, but if i already have one, i want to be greedy and maybe not even wait more than 500ms after it (to have a timely attestation with high chance for 0 inclusion distance)

for the next release, we're going to test different soft timeout schemes to maximize performance, but that isnt necessary now

moshe-blox: for example: we can technically have even a 3s soft timeout if the block event arrived 3 seconds earlier than the waiting period of 4 seconds past the slot, and still be as fast as most validator clients that always wait 4 seconds to start attesting

moshe-blox (approved):
looks good, great work @oleg-ssvlabs 🔥

PR #2110, opened 2025-04-01
fix(exporter): merge experimental improvements

anatolie-ssv:
https://blox-ensemble.monday.com/boards/7946005181/pulses/8582493361

deployed and running:
https://grafana.ops.ssvlabsinternal.com/goto/OahUC50HR?orgId=1

### DIFF
--- a/protocol/v2/ssv/validator/non_committee_validator.go
+++ b/protocol/v2/ssv/validator/non_committee_validator.go
@@ -415,9 +415,9 @@ func (ncv *CommitteeObserver) saveSyncCommRoots(
 	return nil
 }
 
-func (o *CommitteeObserver) postConsensusContainerCapacity() int {
+func (ncv *CommitteeObserver) postConsensusContainerCapacity() int {
 	// #nosec G115 -- slots per epoch must be low epoch not to cause overflow
-	return int(o.beaconConfig.GetSlotsPerEpoch()) + int(validation.LateSlotAllowance)
+	return int(ncv.beaconConfig.GetSlotsPerEpoch()) + int(validation.LateSlotAllowance)
 }
 
 func constructAttestationData(vote *spectypes.BeaconVote, slot phase0.Slot, committeeIndex phase0.CommitteeIndex) *phase0.AttestationData {


operator/validator/controller.go
```diff
@@ -177,8 +179,8 @@ type controller struct {
 	historySyncBatchSize int
 	messageValidator     validation.MessageValidator
 
-	// nonCommittees is a cache of initialized committeeObserver instances
-	committeesObservers      *ttlcache.Cache[spectypes.MessageID, *committeeObserver]
+	// committeeObservers is a cache of initialized committeeObserver instances
+	committeesObservers      *hashmap.Map[spectypes.MessageID, *committeeObserver] // todo: need to evict?
```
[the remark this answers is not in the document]
anatolie-ssv: yeah I noticed that too but the thing is that this PR is all the experimental changes that were in this branch, I just merged it into stage, maybe @y0sher knows what the intention here was?

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -35,10 +36,13 @@ type CommitteeObserver struct {
 	networkConfig     networkconfig.NetworkConfig
 	ValidatorStore    registrystorage.ValidatorStore
 	newDecidedHandler qbftcontroller.NewDecidedHandler
+	rootsMtx          *sync.RWMutex
```
iurii-ssv, line 39: Could group related fields together:
```go
rootsMtx          *sync.RWMutex
attesterRoots     *ttlcache.Cache[phase0.Root, struct{}]
syncCommRoots     *ttlcache.Cache[phase0.Root, struct{}]
```

also, there is no need to use pointer for this definition `*sync.Mutex` (and for `pccMtx`) since we are calling `CommitteeObserver` methods by pointer

Dockerfile
```diff
@@ -38,7 +46,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
   COMMIT=$(git rev-parse HEAD) && \
   VERSION=$(git describe --tags $(git rev-list --tags --max-count=1) --always) && \
   CGO_ENABLED=1 GOOS=linux go install \
-  -tags="blst_enabled" \
+  -tags="blst_enabled,jemalloc,allocator" \
```
iurii-ssv, line 49: From what I recall - last time we've discussed jemalloc we were hesitant to enable it due to potential "undefined behavior" that could result when multiple libraries are dynamically lined into Golang binary (in our case that would be OpenSSL + jemalloc)

I've summarized our research on this [here](https://github.com/ssvlabs/ssv/pull/1819#issuecomment-2700085830) - but didn't find a straightforward solution

@moshe-blox @y0sher ^

nkryuchkov:
this PR might supersede https://github.com/ssvlabs/ssv/pull/1874

nkryuchkov:
@anatolie-ssv Can you please add a list of PRs this PR is built of? This would significantly help us understand where some particular change is from if we have any questions regarding some lines

anatolie-ssv:
> @anatolie-ssv Can you please add a list of PRs this PR is built of? This would significantly help us understand where some particular change is from if we have any questions regarding some lines

this PR is the result of merging `v2.2.1-unstable.2` tag into the stage branch as requested here:

https://blox-ensemble.monday.com/boards/7946005181/pulses/8582493361

it is still pending removal of jemalloc and including pebbledb;

I have no information on what PRs have been used to create the tag.

nkryuchkov:
Includes:
- https://github.com/ssvlabs/ssv/pull/1874
- https://github.com/ssvlabs/ssv/pull/1883
- https://github.com/ssvlabs/ssv/pull/2034

message/validation/quorum.go
```diff
@@ -0,0 +1,58 @@
+package validation
+
+import (
+	"fmt"
+	"math/bits"
+
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+)
+
+// TODO: Take all of these from https://github.com/ssvlabs/ssv/pull/1867 once it's merged.
```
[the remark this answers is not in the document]
nkryuchkov: yes, this PR (#2110) doesn't include #1867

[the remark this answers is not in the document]
nkryuchkov: @MatusKysel sure, this TODO is just to avoid code duplication if #1867 gets merged. However, I'm not sure if we'll merge it or not, so maybe we'll need to remove this comment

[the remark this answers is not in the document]
nkryuchkov: @MatusKysel @anatolie-ssv Actually, as we are about to merge this, let's delete this comment and I'll just merge changes into #1867 and use `Quorum` from this PR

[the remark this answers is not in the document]
nkryuchkov: @MatusKysel @anatolie-ssv removed, please pull the latest changes

ibft/storage/store.go
```diff
@@ -256,15 +297,23 @@ func (i *participantStorage) getParticipants(txn basedb.ReadWriter, pk spectypes
 	}
 
 	operators := decodeOperators(val)
+
+	// Update cache.
+	i.cacheMu.Lock()
+	if i.cachedSlot == slot {
+		i.cachedParticipants[pk] = operators
+	}
+	i.cacheMu.Unlock()
```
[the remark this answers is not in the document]
anatolie-ssv: @moshe-blox have you considered the corner case @iurii-ssv is mentioning here when you implemented the cache over participants feature?

operator/validator/controller.go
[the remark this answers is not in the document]
y0sher: there is not much reason to clean this cache as it should only happen for removed/liquidated validators which isn't a frequent event, and even now it will be as big as 1500 items without cleaning. So IMO we should do it in the future.

ibft/storage/store.go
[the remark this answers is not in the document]
y0sher: @iurii-ssv I agree a simpler design might save us a lot of headache, but we could see in our code implementations like Shares registry that are also doing similar patterns with a memory cache and storage.
@anatolie-ssv is this going to change a lot in exporter v2? if so I'm ok with leaving this less optimal code until then since it's an exporter only functionality anyway.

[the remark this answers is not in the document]
anatolie-ssv: no, in fact in exporter v2 participant storage is no longer use and instead there's a 'duty trace storage' component

ibft/storage/store.go
```diff
@@ -28,22 +29,52 @@ const (
...
+		cachedParticipants: make(map[spectypes.ValidatorPK][]spectypes.OperatorID),
 	}
+
+	// Persist in-memory participants to DB once every slot.
+	slotTicker := slotTickerProvider()
+	go func() {
+		for range slotTicker.Next() {
+			slot := slotTicker.Slot()
+			// Flush previous slot participants.
+			st.cacheMu.Lock()
```
iurii-ssv, line 61: This may be a matter of preference, but what do you think of defining a helper-func here for the sake of calling `defer st.cacheMu.UnLock()` ? The main goal of doing it like that would be to use mutex with `defer` to make sure mutex is unlocked even if later on we add some `return` statement in there (forgetting about unlocking mutex accordingly):
```go
			slot := slotTicker.Slot()

			// Flush previous slot participants.
			flushPrevSlotData := func() {
				st.cacheMu.Lock()
				defer st.cacheMu.Unlock()

				start := time.Now()
				for pk, participants := range st.cachedParticipants {
					if err := st.saveParticipants(pk, st.cachedSlot, participants); err != nil {
						logger.Error("failed to save participants", fields.Validator(pk[:]), zap.Error(err))
					}
				}
				logger.Debug("saved slot participants", fields.Slot(st.cachedSlot), fields.Took(time.Since(start)))

				// Reset cache for new slot.
				st.cachedParticipants = make(map[spectypes.ValidatorPK][]spectypes.OperatorID)
				st.cachedSlot = slot
			}
			flushPrevSlotData()
```

ibft/storage/store_test.go
```diff
@@ -34,8 +35,15 @@ func TestRemoveSlot(t *testing.T) {
 
 	role := spectypes.BNRoleAttester
 
+	slotTickerProvider := func() slotticker.SlotTicker {
+		return slotticker.New(zap.NewNop(), slotticker.Config{
+			SlotDuration: 60 * time.Second, //
```
iurii-ssv, line 40: ` //` is probably unfinished comment (to clarify why it's at 60s?) or maybe we can just remove it

operator/validator/controller.go
```diff
@@ -360,20 +353,19 @@ func (c *controller) handleRouterMessages() {
...
-	//spectypes.BNRoleSyncCommittee:             4,
-	spectypes.RoleSyncCommitteeContribution: 4,
-}
+// TODO: ecivt committeeObserver of dead committees by time or event
+//var committeeObserverValidatorTTLs = map[spectypes.RunnerRole]int{
+//	spectypes.RoleCommittee:                 64,
+//	spectypes.RoleProposer:                  4,
+//	spectypes.RoleAggregator:                4,
+//	spectypes.RoleSyncCommitteeContribution: 4,
+//}
```
iurii-ssv, line 362: It seems this TODO is related to this comment https://github.com/ssvlabs/ssv/pull/2110#discussion_r2032536572 ?

Maybe instead of leaving TODO in the code we could create a Github issue describing/linking relevant parts together to revisit in the future ?

Note, I'm not against TODOs in the code per se, it's just that Github issues are easier to track and enrich with context while also keeping the code itself clean at the same time.

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -37,14 +37,18 @@ type CommitteeObserver struct {
...
-	domainCache       *DomainCache
+
+	rootsMtx      sync.RWMutex
+	attesterRoots *ttlcache.Cache[phase0.Root, struct{}]
+	syncCommRoots *ttlcache.Cache[phase0.Root, struct{}]
 
 	// cache to identify and skip duplicate computations of attester/sync committee roots
 	beaconVoteRoots *ttlcache.Cache[BeaconVoteCacheKey, struct{}]
 
+	domainCache *DomainCache
```
iurii-ssv, line 48: Could you clarify 
- why do we need "cache for that" ?
- and I don't see any "Set" operations on it (only "Get" ones), why that is ?

protocol/v2/ssv/validator/non_committee_validator.go
```diff
@@ -37,14 +37,18 @@ type CommitteeObserver struct {
...
+
 	// TODO: consider using round-robin container as []map[phase0.ValidatorIndex]*ssv.PartialSigContainer similar to what is used in OperatorState
+	pccMtx                 sync.Mutex
 	postConsensusContainer map[phase0.Slot]map[phase0.ValidatorIndex]*ssv.PartialSigContainer
```
iurii-ssv, line 52: It seems we process messages sequentially (synchronizing go-routines when we pull next message out of message queue/channel) ? In that case `pccMtx` seems redundant if it tries to do the same thing - serialize message processing ...

I guess the same applies to `rootsMtx` and even `committeesObserversMutex sync.Mutex` added in this PR ?

So while I generally agree that it's better not to "make assumptions" (such as messages being serialized already) I think the added complexity of having to maintain all these mutexes is probably not worth it (unless we are explicitly planning to get rid of serialized message processing and parallelize it), WDYT ?

iurii-ssv (commented):
Some comments/questions

ibft/storage/store.go
anatolie-ssv: unlocking in defer now

protocol/v2/ssv/validator/non_committee_validator.go
iurii-ssv: Ah I see, 

consider calling it `DomainProvider` (caching being an implementation detail `CommitteeObserver` doesn't need to know about) and maybe `Fetch` method call instead of `Get`

ibft/storage/store.go
```diff
@@ -28,22 +29,52 @@ const (
...
 // instanceType is what separates different iBFT eth2 duty types (attestation, proposal and aggregation)
 type participantStorage struct {
-	logger         *zap.Logger
-	prefix         []byte
-	oldPrefix      string // kept back for cleanup
-	db             basedb.Database
-	participantsMu sync.Mutex
+	logger    *zap.Logger
+	prefix    []byte
+	oldPrefix string // kept back for cleanup
```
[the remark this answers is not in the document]
anatolie-ssv: removed

protocol/v2/ssv/validator/non_committee_validator.go
anatolie-ssv: renamed

protocol/v2/ssv/validator/non_committee_validator.go
iurii-ssv: Oh wow, I didn't notice that `Worker` itself has a bunch of workers of its own ... then it means we had a concurrency bug previously and this PR would fix that 👍

iurii-ssv (approved):
LGTM, the only thing that's sort of bothering me is the way `cachedParticipants` can be accessed concurrently - as mentioned in https://github.com/ssvlabs/ssv/pull/2110#discussion_r2033071192

ibft/storage/store.go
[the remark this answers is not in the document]
iurii-ssv: I guess implementing full-blown in-memory cache (to keep data in-memory for **every** slot we have) is too much of a memory overhead (and it's not super clear what would be the best way to update corresponding DB data on disk - because dumping it all every slot or so it too much of a data as well) ... 

but, @anatolie-ssv maybe we can address the potential concurrency bugs by simply separating 2 cases from one another ?
- whenever `i.cachedSlot == slot` for any `participantStorage` **func** call we only use cache (never query DB), also locking `cacheMu` for the entire **func** call
- whenever `i.cachedSlot != slot` for any `participantStorage` **func** call we only use DB
- and we dump `cachedSlot` data periodically to DB, like you are currently doing already

The only edge-case I see with that approach ^ would be that we might have "saved" some data for "future" slots directly to DB (because `cachedSlot` was at lower number) ... and to account for that we could read/check DB for the next slot when advancing `cachedSlot` number (so we initialize `cachedParticipants` to the actual data we have in DB rather than empty map once we are done with the dump for the current slot).

Btw, is **caching just the "last slot" sufficient** ? Shouldn't we also cache the "slot before the last one" to process messages (I didn't check if that's really happening or not, but maybe worth adding logs to see if that's  something we could use as well) ?

operator/validator/controller.go
anatolie-ssv: @y0sher should this be moved to a github issue? we could also use a quick design chat as I'm not sure I understand `evict... by time or event` implies

ibft/storage/store.go
oleg-ssvlabs, line 1: From what I understand, the changes in the `store` are mostly about introducing cache for quick access to the participant list for a given slot.  
I’m not sure I fully understand the implementation.

This is the diagram so it’s easier to follow the implementation in the PR (please let me know if I missed some pieces):  
image

Just to summarize:
- Participants in the store are **only** updated (before this PR) when the public method `SaveParticipants` is called. The cache is also **always** updated after participants are saved in the DB.
- Participants are fetched through `GetParticipants`. If items aren’t found in the cache for a given slot, they are fetched from the DB and the cache is updated.

Questions:
- `saveParticipantsJob`: I don’t think I fully understand the purpose of it. Why dump from the cache to the DB? Isn’t there literally **no chance** the cache contains items the DB doesn’t? The DB is the only source of truth - the cache should be based on the DB, not the other way around.
- `saveParticipantsJob`: is there a reason to clear the cache on every new slot? We could merge (as I suggested below) slot and participants into a single object and just update it each time `SaveParticipants` is called.

So, at least from how I understand the implementation now, I don’t think the "participants job" is really needed.  
Also, `GetParticipants` should always find items in the cache - there’s no chance they aren’t there, because the only place the participants DB is updated is via `SaveParticipants`, which always sets the cache, although and don't think it's a bad idea that `GetParticipants` does what it does today - checks if items in cache, if not - populates from DB. To make it even simpler, `SaveParticipants` could just update `only` DB and cache can be updated only when items are fetched (first Get will be slower, before cache is populated for the first time)

ibft/storage/store.go
```diff
@@ -9,40 +9,73 @@ import (
...
-	db             basedb.Database
-	participantsMu sync.Mutex
+	logger *zap.Logger
+
+	prefix []byte
+	db     basedb.Database
+
+	// Participants cache for the current slot. Flushed to DB once every slot.
+	cachedParticipants map[spectypes.ValidatorPK][]spectypes.OperatorID
+	cachedSlot         phase0.Slot
```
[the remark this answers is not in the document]
iurii-ssv: +1 on combining things in that way in general, but (just an opinion) I maybe wouldn't go for it if it would cause lots of painful refactoring (eg. rewriting a bunch of unit tests and such)

ibft/storage/store.go
iurii-ssv: I guess it also depends on what kind of load we are optimizing for ? It might be that exporter is write-heavy (many save/store operations) meaning we use the cache primarily to "batch writes" ... but I also would appreciate if that was documented a little bit more extensively in code-comments

anatolie-ssv: @y0sher there are valid questions about the cache implementation that I can't properly answer, can you address them please?

protocol/v2/ssv/validator/domain_cache.go
```diff
@@ -21,20 +21,20 @@ type domainCacheKey struct {
 }
 
 // NewDomainCache must be Start()-ed the same way as ttlcache.
-func NewDomainCache(beaconNode beacon.BeaconNode, ttl time.Duration) *DomainCache {
-	return &DomainCache{
+func NewDomainCache(beaconNode beacon.BeaconNode, ttl time.Duration) *DomainProvider {
```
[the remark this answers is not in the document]
anatolie-ssv: renamed

PR #2112, opened 2025-04-01
chore(dutytracer): deploy version with pebbledb (2)

anatolie-ssv:
This PR merges the `pebble` branch into `feat/exporter-v2`


moshe-blox:
@MatusKysel @anatolie-ssv 
we need two implementations for the db, so for example:
- `storage/badger`
- `storage/pebble`
and a CLI argument called `DB_ENGINE` to switch between "pebble" and "badger"

moshe-blox:
@anatolie-ssv @MatusKysel can we please enable SyncWrites in the Badger implementation?

this makes it slower but guarantees data is persisted on every call, which is very important for slashing protection

this means that we'll keep Badger as our operator engine so it can be slow but more secure, and recommend Pebble for exporters only

cli/operator/node.go
```diff
@@ -138,7 +138,7 @@ var StartNodeCmd = &cobra.Command{
 		}
 		cfg.DBOptions.Ctx = cmd.Context()
 		var db basedb.Database
-		if cfg.DBOptions.Type == "pebble" || cfg.DBOptions.Type == "pebbledb" {
+		if cfg.DBOptions.Engine == "pebble" || cfg.DBOptions.Engine == "pebbledb" {
```
moshe-blox, line 141: IMO if we've settled on "pebble" and "badger" then lets enforce those to keep it simple

the CLI field can be configured default to have default of "badger", so you can switch on the engine and error if its not either of our two options, without forcing existing users to explicitly specify "badger"

anatolie-ssv: fixed

cli/operator/node.go
```diff
@@ -603,24 +608,46 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
 	// to reclaim any space that may have been freed up.
-	// Close & reopen the database to trigger any unknown internal
-	// startup/shutdown procedures that the storage engine may have.
 	start := time.Now()
-	if err := reopenDb(); err != nil {
-		return nil, err
+	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
+	defer cancel()
+	if err := db.FullGC(ctx); err != nil {
+		return nil, errors.Wrap(err, "failed to collect garbage")
```
nkryuchkov, line 615: let's not use http://github.com/pkg/errors for new code as it's been discontinued

cli/operator/node.go
```diff
@@ -603,24 +608,46 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
+	}
+
+	logger.Debug("post-migrations garbage collection completed", fields.Duration(start))
+
+	return db, nil
+}
+
+func setupPebbleDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.PebbleDB, error) {
```
nkryuchkov, line 623: code of `setupBadgerDB` and `setupPebbleDB` mostly matches, can we extract anything in a separate function to reduce code duplication?

storage/kv/gc.go
```diff
@@ -31,25 +31,21 @@ func (b *BadgerDB) periodicallyCollectGarbage(logger *zap.Logger, interval time.
...
-	b.gcMutex.Lock()
-	defer b.gcMutex.Unlock()
-
-	err := b.db.RunValueLogGC(0.5)
-	if errors.Is(err, badger.ErrNoRewrite) {
-		// No garbage to collect.
-		return nil
-	}
-	return err
+	return b.gc(ctx, 0.7)
```
nkryuchkov, line 34: why exactly 0.7?

storage/kv/pebble.go
```diff
@@ -0,0 +1,257 @@
...
+
+// PebbleDB struct
+type PebbleDB struct {
+	logger *zap.Logger
+
+	db *pebble.DB
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	//wg     sync.WaitGroup
```
nkryuchkov, line 23: do we still need it?

storage/kv/pebble.go
```diff
@@ -0,0 +1,257 @@
...
+
+		// Call the iterator callback.
+		if err := iterator(obj); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func makePrefixIter(dbOrBatch pebble.Reader, prefix []byte) (*pebble.Iterator, error) {
```
nkryuchkov, line 118: Can you please add some comments on why we need the incremental logic in this function? Also, I think `keyUpperBound` needs unit tests

storage/kv/pebble.go
```diff
@@ -0,0 +1,257 @@
...
+// TODO: implement gc?
+
+// QuickGC runs a short garbage collection cycle to reclaim some unused disk space.
+func (pdb *PebbleDB) QuickGC(ctx context.Context) error {
+	return nil
+}
+
+// FullGC runs a long garbage collection cycle to reclaim (ideally) all unused disk space.
+func (pdb *PebbleDB) FullGC(ctx context.Context) error {
+	//pdb.logger.Info("Starting FullGC")
```
nkryuchkov, line 249: Can you please clean up these comments? If we need them, let's put a TODO to understand why we need them and why they are commented out. If we don't need them, let's remove them

storage/kv/txn.go
```diff
@@ -3,6 +3,8 @@ package kv
 import (
 	"errors"
 
+	"github.com/cockroachdb/pebble"
+
```
nkryuchkov, line 7: could you please remove this blank line to make imports sortable?

storage/kv/pebble.go
nkryuchkov, line 1: Could you please add tests for this file? Other files in this package are tested, so IMO we shouldn't add any code without tests here

anatolie-ssv: please find the tests for this PR here: https://github.com/ssvlabs/ssv/pull/2113

cli/operator/node.go
moshe-blox: @nkryuchkov we can maybe exact to applyMigrations

storage/kv/pebble.go
anatolie-ssv: this feature has been moved to this PR: https://github.com/ssvlabs/ssv/pull/2150 sorry for the confusion

PR #2126, opened 2025-04-04
message/validation: index state by peer ID

nkryuchkov:
Depends on https://github.com/ssvlabs/ssv/pull/1874


kchojn (approved):

iurii-ssv (approved):
Could maybe:
- expand PR name to `message/validation: index state by peer ID + message ID` for clarify
- add a short PR description to clarify why this is needed (is it to make sure we prevent collisions for messages from different peers, or maybe something else is the reason for this change ?)

Issue #2105, opened 2025-03-26
validator changing cluster & fee-recipient (both at the same time) could be problematic


iurii-ssv:
The issue is described here - https://github.com/ssvlabs/ssv/pull/2030#discussion_r2012519394

iurii-ssv:
Has been addressed by the PRs linked above.

Issue #1361, opened 2024-04-01
First Relay Validator Registration


olegshmuelov:
**Describe the bug**
There is an issue in the validator registration process that could lead to registration delays. Specifically, the current logic, which uses the validator's BeaconMetadata.Index in conjunction with registrationSlotInterval for scheduling registration duties, has the unintended potential to postpone a validator’s registration by up to 10 epochs. This delay stems from the way these values are calculated and compared to determine a validator's eligibility for registration at any given slot.

**Expected behavior**
Register validator to relay in first epoch the node starts

**Additional context**
Thisi s a regression bug
see [commit](https://discord.com/channels/@me/950674438955884574/1221809356715196506) that remove the regression logic
the [PR](https://github.com/bloxapp/ssv/pull/1179) the logic was removed
I addition write tests to validator registration scheduler, covering all scenarios

PR #2130, opened 2025-04-07
validator-registration: duty enhancements

iurii-ssv:
This PR is addressing several edge-cases that are hard to reproduce due to being dependent on timing of certain events in the SSV cluster):
- making sure `FeeRecepientAddress` for some Validator will be updated even without SSV node restart (meaning previously `FeeRecepientAddress` update [wasn't recognized by SSV node until SSV node restart](https://github.com/ssvlabs/ssv-spec/issues/504#issuecomment-2780862223)); Note, from what we've observed in practice fee-recipient does eventually update to the latest value - but it's still a [data-race](https://github.com/ssvlabs/ssv-spec/issues/504#issuecomment-2786836545) this PR aims to address properly
- making sure there isn't a large delay (that previously could potentially be 10 epochs long) between `FeeRecepientAddress` update and validator-registration request being sent to Beacon node, the max expected delay now is ~1 epoch
- making sure SSV node won't submit validator-registration requests to Beacon nodes for validators that are no longer served by SSV cluster (previously we still would be sending such requests until SSV node restart)

See prior relevant discussions for additional context:
- https://github.com/ssvlabs/ssv-spec/issues/504#issuecomment-2780862223
- https://github.com/ssvlabs/ssv/pull/2030#discussion_r2030101100

I've also found another relevant issue this PR still doesn't address - https://github.com/ssvlabs/ssv/issues/1361:
- this PR does address the aspect related to `FeeRecepientAddress` updates (so that validator-registration delay is 1 epoch at most, not 10 epochs) like I've mentioned above
- I'm not 100% sure the stated desired behavior "Register validator to relay in first epoch the node starts" is super useful/necessary to have since it's unlikely to have much of an effect, and there are other nodes in the cluster that could handle validator registration while 1 SSV node is restarting

So unless you have another interpretation ^ I would say this PR closes https://github.com/ssvlabs/ssv/issues/1361 as well. Also closes https://github.com/ssvlabs/ssv/issues/2105

_Before merging:_
- [x] resolve TODOs
- [x] test on **stage** (with added per-validator debug-logs, as in this [diff](https://github.com/ssvlabs/ssv/compare/stage...testing/validator-registrations-missing?expand=1))
- [x] make sure the change doesn't result in accidental SSV network fork


operator/validators/validators_map.go
```diff
@@ -18,18 +18,20 @@ type committeeIterator func(validator *validator.Committee) bool
...
 type ValidatorsMap struct {
-	ctx        context.Context
-	vlock      sync.RWMutex
-	mlock      sync.RWMutex
+	ctx context.Context
+
+	vLock      sync.RWMutex
 	validators map[spectypes.ValidatorPK]*validator.Validator
+
+	cLock      sync.RWMutex
```
nkryuchkov, line 26: What do you think about `committeeMu`? It won't look large. It seems we have already forgotten to change it once because the one-letter prefix is not searchable.

iurii-ssv: I don't mind slightly longer names if that helps, updated

protocol/v2/ssv/runner/committee.go
```diff
@@ -343,7 +343,6 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 	}
 	logger = logger.With(fields.Slot(signedMsg.Slot))
 
-	// TODO: (Alan) revert?
```
iurii-ssv, line 346: Not sure what this was about, @Y0sher any ideas ? We have 2 of these comments in total - I'm removing both in this PR while cleaning up in the areas nearby.

operator/duties/scheduler.go
```diff
@@ -397,7 +397,6 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, logger *zap.Logger, dutie
 			if duty.Type == spectypes.BNRoleAttester || duty.Type == spectypes.BNRoleSyncCommittee {
 				s.waitOneThirdOrValidBlock(duty.Slot)
 			}
-			recordDutyExecuted(ctx, duty.RunnerRole())
 			s.dutyExecutor.ExecuteDuty(ctx, logger, duty)
```
iurii-ssv, line 400: Feels like we should record this inside of `ExecuteDuty` and `ExecuteCommitteeDuty` since scheduler isn't the only one who can trigger duty execution (for example, in this PR I'm triggering validator-registration duty based on Ethereum event)

operator/validator/controller.go
```diff
@@ -184,9 +184,8 @@ type controller struct {
 	syncCommRoots            *ttlcache.Cache[phase0.Root, struct{}]
 	domainCache              *validator.DomainCache
 
-	recentlyStartedValidators uint64
-	indicesChange             chan struct{}
-	validatorExitCh           chan duties.ExitDescriptor
+	indicesChange   chan struct{}
+	validatorExitCh chan duties.ExitDescriptor
```
iurii-ssv, line 188: `recentlyStartedValidators` is unused, so I'm removing it

operator/duties/validator_registration.go
```diff
@@ -65,17 +58,13 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
...
 				}})
 
-				vrs = append(vrs, ValidatorRegistration{
-					ValidatorIndex: share.ValidatorIndex,
-					FeeRecipient:   hex.EncodeToString(share.FeeRecipientAddress[:]),
-				})
+				h.logger.Debug("validator registration duty sent",
+					zap.Uint64("slot", uint64(slot)),
+					zap.Uint64("validator_index", uint64(share.ValidatorIndex)),
+					zap.String("validator_pubkey", pk.String()))
```
iurii-ssv, line 66: Reading `share.FeeRecipientAddress` value here is technically a data race (since we aren't doing explicit synchronisation with writes happening when handling Ethereum event) - see more on this [here](https://github.com/ssvlabs/ssv-spec/issues/504#issuecomment-2786836545)

so I'm removing it from this log line since it doesn't seem super important (if it is - we could query database to fetch it like I'm now doing in `calculateValidatorRegistration`)

Also getting rid of `ValidatorRegistration struct` that's used only for logging here to simplify this code a bit

beacon/goclient/validator.go
```diff
@@ -84,22 +84,42 @@ func (gc *GoClient) registrationSubmitter(slotTickerProvider slotticker.Provider
...
+					continue
+				}
+				targetRegs[pk] = r
+			}
+			// 2. find & add newly created validator registrations
+			for pk, r := range gc.registrations {
+				if r.new {
+					targetRegs[pk] = r
+				}
+			}
```
iurii-ssv, line 111: It's a bit wasteful to recalculate 1 and 2 every slot just so we can "distribute submissions throughout an epoch" - we could ofc "cache" this calculation but maybe it's not worth complicating this code for, WDYT ?

beacon/goclient/events_test.go
```diff
@@ -73,12 +73,16 @@ func TestSubscribeToHeadEvents(t *testing.T) {
...
-		BeaconNodeAddr: serverURL,
-		Context:        context.Background(),
-		Network:        beacon.NewNetwork(types.MainNetwork),
-	},
-		tests.MockSlotTickerProvider)
+	server, err := New(
+		zap.NewNop(),
+		Options{
+			BeaconNodeAddr: serverURL,
+			Context:        context.Background(),
```
oleg-ssvlabs, line 80: we could use `t.Context()` here (new in Go 1.24)

iurii-ssv: mm, changed

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...5c2bedaef","60ac8eb415e6c748","119e95426dc4affb","e918003e25abf4cd","44c0efaa4b189434","de6ef5f1aa07b38d","b71a695e14c7e892","c9b7984a5d2604b5","72e9cd10959558de","82f692ac2f5817f3","3e0d3d7ade790448","52595c2a99ba6031","ea6ad19cdc29b230","695cba9190654139","4418aa6010632480","e27b9c1ea1c64357","fe62097e8814c106","1cfd3a3660a62879","ddf404c2905d3b26","8ac6baea1b755066","2bc023c7e062f24b","8c43f241aeac6d75","832fc9a9a2b71d00","34e8b7999dfb14b5","ccc48c575f7ab4eb","9de0e5d104caf802","e46f98bfca35d4a0","db882c2be6efd7d4","6577bf381d78bae2","f66dbe09241b5a4","48884cab13bd3903","fd6796ea6d131c3","dfde38eed6f3305d","9b895440c1743c4e","783942f921da9985","e0c97cefd6558c77","75c75fa4503282","680487ae23f269a1","9b70726386e38170","ef7b63d9848dc185","3fe59bc388b9ef3","dd93140e21e6f919","143c7821cd65bcac","70d29c9e7a4ef797","b26376133e68b3db","31a354962a9bcfa0","50e62cbde34516ee","9273478f6feccd62","eddf20a67e3fddef","5a44c960cc4178dd","5499865be87b7e00","446545b04b35fba","8887c18020ddb307","3adda38e5fadcfb","9f105a95e82012a4","50cea598241b0bc9","6396e2866ef1b30","1c0da45940bfa76f","22b9026cb0221ce0","2011029b33aad319","3517ccdcc876d88b","6034651b79ebf41d","b0c3777f72f72671","767d6797a04294ea"]
```
iurii-ssv, line 1: Diff:
```
diff --git a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner b/ssv/validator_registration.go@NewValidatorRegistrationRunner
index 3517ccdcc876d88b..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@NewValidatorRegistrationRunner
@@ -1,31 +1,25 @@
 func NewValidatorRegistrationRunner(
-	domainType DomainType,
 	beaconNetwork BeaconNetwork,
 	share map[phase0.ValidatorIndex]*Share,
-	validatorOwnerAddress common.Address,
-	beacon beacon.BeaconNode,
+	beacon BeaconNode,
 	network Network,
-	signer ekm.BeaconSigner,
-	operatorSigner ssvtypes.OperatorSigner,
-	recipientsStorage recipientsStorage,
+	signer BeaconSigner,
+	operatorSigner *OperatorSigner,
 	gasLimit uint64,
 ) (Runner, error) {
 	if len(share) != 1 {
 	}
 	return &ValidatorRegistrationRunner{
 		BaseRunner: &BaseRunner{
 			RunnerRoleType:	RoleValidatorRegistration,
-			DomainType:	domainType,
 			BeaconNetwork:	beaconNetwork,
 			Share:		share,
 		},
-		beacon:			beacon,
-		network:		network,
-		signer:			signer,
-		operatorSigner:		operatorSigner,
-		recipientsStorage:	recipientsStorage,
-		validatorOwnerAddress:	validatorOwnerAddress,
+		beacon:		beacon,
+		network:	network,
+		signer:		signer,
+		operatorSigner:	operatorSigner,
 		gasLimit:	gasLimit,
 	}, nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner b/ssv/validator_registration.go@ValidatorRegistrationRunner
index 6034651b79ebf41d..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner
@@ -1,11 +1,9 @@
 ValidatorRegistrationRunner struct {
 	BaseRunner	*BaseRunner
-	beacon			beacon.BeaconNode
-	network			Network
-	signer			ekm.BeaconSigner
-	operatorSigner		ssvtypes.OperatorSigner
-	valCheck		ProposedValueCheckF
-	recipientsStorage	recipientsStorage
-	validatorOwnerAddress	common.Address
+	beacon		BeaconNode
+	network		Network
+	signer		BeaconSigner
+	operatorSigner	*OperatorSigner
+	valCheck	ProposedValueCheckF
 	gasLimit	uint64
 }
\ No newline at end of file
index b0c3777f72f72671..2222222
@@ -1,35 +1,34 @@
 	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
 	if err != nil {
 		return errors.Wrap(err, "failed processing validator registration message")
 	}
 	if !quorum {
 		return nil
 	}
 	root := roots[0]
 	if err != nil {
 		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee,
 			r.GetShare().ValidatorIndex)
 		return errors.Wrap(err, "got pre-consensus quorum but it has invalid signatures")
 	}
 	copy(specSig[:], fullSig)
 	registration, err := r.calculateValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
 	if err != nil {
 		return errors.Wrap(err, "could not calculate validator registration")
 	}
 			Message:	registration,
 			Signature:	specSig,
 		},
 	}
-	err = r.beacon.SubmitValidatorRegistration(signedRegistration)
-	if err != nil {
 		return errors.Wrap(err, "could not submit validator registration")
 	}
 	r.GetState().Finished = true
 	return nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
index 767d6797a04294ea..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
@@ -1,18 +1,15 @@
 func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0.Slot) (*v1.ValidatorRegistration, error) {
+	share := r.GetShare()
+	if share == nil {
+		return nil, errors.New("no share to get validator public key")
+	}
 	pk := phase0.BLSPubKey{}
-	copy(pk[:], r.GetShare().ValidatorPubKey[:])
+	copy(pk[:], share.ValidatorPubKey[:])
 	epoch := r.BaseRunner.BeaconNetwork.EstimatedEpochAtSlot(slot)
-	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
-	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
-	}
-	if !found {
-		return nil, fmt.Errorf("get recipient data from storage: not found")
-	}
 	return &v1.ValidatorRegistration{
-		FeeRecipient:	rData.FeeRecipient,
+		FeeRecipient:	share.FeeRecipientAddress,
 		GasLimit:	r.gasLimit,
 		Timestamp:	r.BaseRunner.BeaconNetwork.EpochStartTime(epoch),
 		Pubkey:		pk,
 	}, nil
 }
\ No newline at end of file
```

beacon/goclient/validator.go
```diff
@@ -84,22 +84,42 @@ func (gc *GoClient) registrationSubmitter(slotTickerProvider slotticker.Provider
 			return
 		case <-ticker.Next():
 			currentSlot := ticker.Slot()
+			currentEpoch := gc.network.EstimatedCurrentEpoch()
```
y0sher, line 87: I think it's better to use `EstimatedEpochAtSlot(currentSlot)` here.

operator/validator/task_executor.go
```diff
@@ -82,8 +83,24 @@ func (c *controller) UpdateFeeRecipient(owner, recipient common.Address) error {
 	c.validatorsMap.ForEachValidator(func(v *validator.Validator) bool {
 		if v.Share.OwnerAddress == owner {
 			v.Share.FeeRecipientAddress = recipient
-
 			logger.Debug("updated recipient address")
+
+			// Kick off validator registration duty to notify various Ethereum actors (e.g. Relays)
+			// about fee recipient change as soon as possible.
+			slot := c.networkConfig.Beacon.EstimatedCurrentSlot()
```
y0sher, line 90: IMO this is problematic, we can't gurantee that all 4 operators will have the same estimated current slot when executing this code. Take a look at how we do validator exits, we use the block time from the event's block to scheduele and be able to agree on which slot to execute the duty. 
In your solution if there's a diff in the processing time of events or the clock we are going to fail the duty because some operators will try executing the duty on a different slot.

protocol/v2/ssv/runner/committee.go
y0sher: honetly can't remember exactly but from looking at the code it seems like it was already addressed.

y0sher (changes requested):
code looks good, one issue with the logic.

operator/validator/task_executor.go
zippoxer: @y0sher +1

the slot of the duty must be derived from the event time rather than OS time, because operators often receive at different times (especially during EL hiccups)

eth/eventhandler/event_handler.go
```diff
@@ -398,7 +399,12 @@ func (eh *EventHandler) processEvent(ctx context.Context, txn basedb.Txn, event
 			return nil, nil
 		}
 
-		task := NewUpdateFeeRecipientTask(eh.taskExecutor, feeRecipientAddressUpdatedEvent.Owner, feeRecipientAddressUpdatedEvent.RecipientAddress)
+		task := NewUpdateFeeRecipientTask(
+			eh.taskExecutor,
+			feeRecipientAddressUpdatedEvent.Owner,
+			feeRecipientAddressUpdatedEvent.RecipientAddress,
+			phase0.Slot(feeRecipientAddressUpdatedEvent.Raw.BlockNumber),
```
y0sher, line 406: not sure this works, I'm pretty sure block number increments only when a block is actually added to the chain. Empty (“missed”) slots don’t create a block, so the block number does not increase for that slot.

Also previously, we weren't triggering a duty here (in NewUpdateFeeRecipientTask implementation) so the slot or block number wasn't needed. So far the way we have triggered duties based on tx block number time + some delta to determine the slot to start the duty (please see validator exit tx and how it's being translated to a duty).

iurii-ssv: Ah actually you are right about block number being it's own thing - so we'd need a similar translation as [validator-exit is doing](https://github.com/ssvlabs/ssv/blob/16aae3d1847c10b7d4422cba63c5800291e98ff3/operator/duties/voluntary_exit.go#L127-L150) from **block_number** -> to **slot_number**

I'll push a change to address that part soon

y0sher: yes just note that validator exit duty is also being scheduled to a future slot - https://github.com/ssvlabs/ssv/blob/16aae3d1847c10b7d4422cba63c5800291e98ff3/operator/duties/voluntary_exit.go#L74C1-L74C56

iurii-ssv: For validator-registration I don't really see a reason to "schedule it for the future slot", the slot that corresponds to `block_number` (the block denoting fee-recipient change) seems like the correct one to use ?

operator/validator/task_executor.go
iurii-ssv: This discussion is continued in https://github.com/ssvlabs/ssv/pull/2130#discussion_r2084677694

operator/duties/validator_registration.go
```diff
@@ -0,0 +1,133 @@
...
+func (h *ValidatorRegistrationHandler) blockSlot(ctx context.Context, blockNumber uint64) (phase0.Slot, error) {
+	block, err := h.executionClient.BlockByNumber(ctx, new(big.Int).SetUint64(blockNumber))
+	if err != nil {
+		return 0, fmt.Errorf("request block %d from execution client: %w", blockNumber, err)
+	}
+
+	blockSlot := h.network.Beacon.EstimatedSlotAtTime(int64(block.Time())) // #nosec G115
+
+	return blockSlot, nil
+}
```
iurii-ssv, line 133: I don't think we need to "cache" execution client calls since cache re-use here is probably rare (significantly more rare than for validator-exit-duty)

moshe-blox: @iurii-ssv is this called per-event? what if there are 1000 fee recipient events?

eth/eventhandler/event_handler.go
moshe-blox: @iurii-ssv that slot already happened, so nothing will get executed

we add a delay to give everyone enough time to receive the event from their EL, which can often happen a few slots later

IMO we should copy the same behavior we have in VoluntaryExit, to stay consistent across the codebase (same problem -> same solution)

moshe-blox: we should probably add a short comment on this constant to make it clear:

https://github.com/ssvlabs/ssv/blob/16aae3d1847c10b7d4422cba63c5800291e98ff3/operator/duties/voluntary_exit.go#L74

iurii-ssv: Okay I see now - the reason we don't want for **slot** to be in the past is so that corresponding duty will "be considered valid"

added a commit to address that - https://github.com/ssvlabs/ssv/pull/2130/commits/9f5cfdc7c2515f37b439988908358403bd7003e1

operator/duties/validator_registration.go
iurii-ssv: Yes that's correct, is that likely however to encounter multiple recipient events in the same block in practice ?

Maybe several events could happen occasionally in 1 block but 100 seems unlikely ... anyway, I guess we don't want to make any assumptions on this and can easily add a cache (exactly as in voluntary-exit duty) - which is just several additional lines of code https://github.com/ssvlabs/ssv/pull/2130/commits/d81ca4bdf3d4d61361f18fba942100cb9848e3f7

we do however duplicate this "block_number->slot_number mapping, with caching" functionality across 2 handlers now (validator-registration & voluntary-exit):
- keeping these duplicate implementations allows us not to worry about concurrent access to cached values
- so if we want to create some sort of helper-entity to unify it we'd need thread-safe map

personally I wouldn't bother deduping this just yet, but if you think we should - let me know

operator/duties/voluntary_exit.go
```diff
@@ -134,14 +139,16 @@ func (h *VoluntaryExitHandler) blockSlot(ctx context.Context, blockNumber uint64
...
 
 	h.blockSlots[blockNumber] = blockSlot
+
+	// Clean up older cached values since they are not relevant anymore.
 	for k, v := range h.blockSlots {
-		if v < blockSlot && blockSlot-v >= voluntaryExitSlotsToPostpone {
+		if blockSlot >= v+10 {
 			delete(h.blockSlots, k)
 		}
 	}
```
iurii-ssv, line 154: I've changed this condition here to simplify it as much as we can:
- `v < blockSlot` is always true when `blockSlot-v >= voluntaryExitSlotsToPostpone` is true (so only `blockSlot-v >= voluntaryExitSlotsToPostpone` is meaninful here)
- `voluntaryExitSlotsToPostpone` seems irrelevant here - so I'd rather use some "dumb constant" instead to keep it simple

iurii-ssv: I'm not sure it has "slotsToPostpone" meaning though (or at least it's not easy to see if/why/how it does)

so instead I'm sort of changing this meaning to something like "recently queried block numbers" because that's all we really need from it, and put a number 10 to it to make it somewhat future-proof (previous one was 4)

I guess we can make it a constant to name it as such - https://github.com/ssvlabs/ssv/pull/2130/commits/4d20f4d3f95b1f4152a22432ed1635b1b9215723 - although here it doesn't seem super useful to me given I've left a comment explaining this block of code:
```
// Clean up older cached values since they are not relevant anymore.
```

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -246,18 +257,21 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
 }
 
 func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0.Slot) (*v1.ValidatorRegistration, error) {
-	share := r.GetShare()
-	if share == nil {
-		return nil, errors.New("no share to get validator public key")
-	}
-
```
iurii-ssv, line 260: Runner can't be created with `nil` share (we check for that in constructor and error if that's the case), so checks like this one are redundant - hence I'm removing these.

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -246,18 +257,21 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
+	if err != nil {
+		return nil, fmt.Errorf("get recipient data from storage: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("get recipient data from storage: not found")
+	}
+
 	return &v1.ValidatorRegistration{
-		FeeRecipient: share.FeeRecipientAddress,
+		FeeRecipient: rData.FeeRecipient,
```
iurii-ssv: Runners shouldn't access `share.FeeRecipientAddress` directly since it can be updated concurrently - and unfortunately we don't have a mechanism to sync such concurrent updates across various go-routines - hence querying storage is the only "safe" way to get up-to-date `FeeRecipient` value

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -317,3 +331,153 @@ func (r *ValidatorRegistrationRunner) GetRoot() ([32]byte, error) {
...
+	beacon         beacon.BeaconNode
+	validatorStore validatorStore
+
+	// registrationMu synchronises access to registrations
+	registrationMu sync.Mutex
+	// registrations is a set of validator-registrations (their latest versions) to be sent to
+	// Beacon node to ensure various entities in Ethereum network, such as Relays, are aware of
+	// participating validators
+	registrations map[phase0.BLSPubKey]*validatorRegistration
+}
```
iurii-ssv: `VRSubmitter` contains functionality previously residing with `GoClient`, I had to create a separate service for it so there isn't a cyclic dependency during struct-initialization (recently we've re-arranged code such that `storage` needs `GoClient`, but with this PR `VRSubmitter` needs `storage` hence `VRSubmitter` can't be a part of `GoClient`),

regardless, `GoClient` contains a lot of functionality, cutting off a piece of it into separate service is good anyways

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -22,8 +22,8 @@ import (
 // MessageHandler process the msg. return error if exist
 type MessageHandler func(ctx context.Context, logger *zap.Logger, msg *queue.SSVMessage) error
 
-// queueContainer wraps a queue with its corresponding state
-type queueContainer struct {
+// QueueContainer wraps a queue with its corresponding state
+type QueueContainer struct {
 	Q          queue.Queue
 	queueState *queue.State
 }
```
iurii-ssv, line 29: Exporting `QueueContainer` to make it easily usable in tests

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -319,3 +332,153 @@ func (r *ValidatorRegistrationRunner) GetRoot() ([32]byte, error) {
...
+					continue
+				}
+				targetRegs[pk] = r
+			}
+			// 2. find and add newly created validator registrations
+			for pk, r := range s.registrations {
+				if r.new {
+					targetRegs[pk] = r
+				}
+			}
```
olegshmuelov, line 442: The registrations map in VRSubmitter never removes entries for validators that are no longer participating. This could lead to unbounded memory growth.

Consider implementing cleanup logic:
```go
// In the start() method, after processing registrations:
// Clean up registrations for non-participating validators once per epoch (or every ~10 epochs to reduce overhead)
if currentSlot % s.beacon.GetBeaconNetwork().SlotsPerEpoch() == 0 { // Run cleanup once per epoch
    participatingPubKeys := make(map[phase0.BLSPubKey]bool)
        pk := phase0.BLSPubKey{}
        copy(pk[:], share.ValidatorPubKey[:])
        participatingPubKeys[pk] = true
    }

    for pk := range s.registrations {
        if !participatingPubKeys[pk] {
            delete(s.registrations, pk)
        }
    }
}

File: operator/validator/task_executor.go [hunk 19]
```diff
@@ -73,16 +73,33 @@ func (c *controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+			}
+
+			go func() {
+				select {
+				case c.validatorRegistrationCh <- regDesc:
+					logger.Debug("added validator registration task to pipeline")
+				case <-time.After(2 * c.networkConfig.GetSlotDuration()):
+					logger.Error("failed to schedule validator registration duty!")
+				}
+			}()
```
olegshmuelov: Consider adding context cancellation:

```go
go func() {
    select {
    case c.validatorRegistrationCh <- regDesc:
        logger.Debug("added validator registration task to pipeline")
    case <-time.After(2 * c.networkConfig.GetSlotDuration()):
        logger.Error("failed to schedule validator registration duty!")
    case <-c.ctx.Done():
        logger.Debug("context cancelled while scheduling validator registration")
    }
}()
```

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -319,3 +332,153 @@ func (r *ValidatorRegistrationRunner) GetRoot() ([32]byte, error) {
...
+
+			currentSlot := ticker.Slot()
+			currentEpoch := config.EstimatedEpochAtSlot(currentSlot)
+			slotInEpoch := uint64(currentSlot) % config.SlotsPerEpoch
+
+			// Select registrations to submit.
+			targetRegs := make(map[phase0.BLSPubKey]*validatorRegistration, 0)
+			s.registrationMu.Lock()
+			// 1. find and add validators participating in the 10th epoch from now
+			shares := s.validatorStore.SelfParticipatingValidators(currentEpoch + 10)
```
olegshmuelov: This uses `SelfParticipatingValidators` which includes exited validators that are still eligible for sync committee duties. We should use validators that are actually attesting (proposing in our context):

```go
for _, share := range h.validatorProvider.SelfValidators() {
	if share.IsParticipatingAndAttesting(epoch) {

	}
}

File: protocol/v2/ssv/runner/validator_registration.go [hunk 21]
```diff
@@ -248,18 +258,21 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
+	copy(pk[:], r.GetShare().ValidatorPubKey[:])
 
 	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
 
+	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
+	if err != nil {
+		return nil, fmt.Errorf("get recipient data from storage: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("get recipient data from storage: not found")
```
olegshmuelov: Add the owner address to the error message for better debugging:

```go
if !found {
    return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
}
```

operator/duties/validator_registration.go
```diff
@@ -0,0 +1,157 @@
...
+		case regDescriptor, ok := <-h.validatorRegCh:
+			if !ok {
+				return
+			}
+
+			// Calculate duty slot in a deterministic manner to ensure every Operator will have the same
+			// slot value for this duty. Additionally, add validatorRegistrationSlotsToPostpone slots on
+			// top to ensure the duty is scheduled with a slot number never in the past since several slots
+			// might have passed by the time we are processing this event here.
+			const validatorRegistrationSlotsToPostpone = phase0.Slot(4)
```
olegshmuelov: Consider making these magic numbers configurable fields in the handler for better testability:

```go
const (
    // slotsToPostpone ensures duty is scheduled with future slot
    slotsToPostpone = phase0.Slot(4)
    
    // recentlyQueriedBlocks controls cache cleanup threshold
    recentlyQueriedBlocks = 10
)

type ValidatorRegistrationHandler struct {
    baseHandler
    validatorRegCh <-chan RegistrationDescriptor
    blockSlots     map[uint64]phase0.Slot
    
    // Configurable for testing
    slotsToPostpone      phase0.Slot
    recentlyQueriedBlocks int
}

// In NewValidatorRegistrationHandler:
return &ValidatorRegistrationHandler{
    validatorRegCh:       validatorRegistrationCh,
    blockSlots:           map[uint64]phase0.Slot{},
    slotsToPostpone:      slotsToPostpone,
    recentlyQueriedBlocks: recentlyQueriedBlocks,
}
```

This allows tests to use smaller values for faster execution and edge case testing.

olegshmuelov (changes requested):
Great work on this PR!

 - Just to note that during rollout, there will be temporary consensus failures for immediate registrations until 3f+1 operators update. Registrations will still succeed at the regular deterministic slot, so no validator impact - just some log noise during transition.

 - test coverage currently at 37.88%, we need to add tests for the new registration handler, fee recipient updates, and concurrent access scenarios.

protocol/v2/ssv/runner/validator_registration.go
iurii-ssv: True, however I'm not sure if we want to add this clean-up logic here simply because of how small `registrations` is:
- total max amount of validators SSV node processes is on the order of 1000 - 10 000
- `registrations` values are pointers (if it was large struct then it would make things worse)

@y0sher let me know if you want this clean-up logic added or not

protocol/v2/ssv/runner/validator_registration.go
iurii-ssv: Applied in https://github.com/ssvlabs/ssv/pull/2130/commits/a633f2aba09e58b8e88f7cd0508b517c8d783825

iurii-ssv:
> Just to note that during rollout, there will be temporary consensus failures for immediate registrations until 3f+1 operators update. Registrations will still succeed at the regular deterministic slot, so no validator impact - just some log noise during transition.

Not sure why this PR could result in temporary consensus failures (I assume we are talking about validator-registration duty here), @olegshmuelov could you clarify ? The duty scheduler itself (`HandleDuties` method) hasn't been changed in any way - just moved from one file to another.

What's been changed is "the way we send already signed validator-registrations out into Ethereum network".

Maybe what you mean is the newly added code in `UpdateFeeRecipient` that generates "new" duty that Operators that are running previous code version won't be aware of ?

olegshmuelov:
> Not sure why this PR could result in temporary consensus failures (I assume we are talking about validator-registration duty here), @olegshmuelov could you clarify ? The duty scheduler itself (`HandleDuties` method) hasn't been changed in any way - just moved from one file to another.
> 
> What's been changed is "the way we send already signed validator-registrations out into Ethereum network".
> 
> Maybe what you mean is the newly added code in `UpdateFeeRecipient` that generates "new" duty that Operators that are running previous code version won't be aware of ? But that's probably fine since it's a strict enhancement over what we currently have.

@iurii-ssv The issue is the new immediate duty scheduling path:

  Flow: `FeeRecipientAddressUpdated` event -> `UpdateFeeRecipientTask` -> `validatorRegistrationCh` -> Schedule duty at blockSlot + 4

  During rollout:
  - OLD operators: Only schedule at deterministic slots
  - NEW operators: Also schedule immediately at `blockSlot` + 4

  Example:
  Block 1000: Fee recipient updated
  Slot 1004: NEW operators try consensus → FAILS if <3f+1 updated
  Slot 4000 (deterministic): ALL operators try consensus → SUCCEEDS

  No validator impact - registrations still succeed at deterministic slots. Just temporary consensus failures until 3f+1 operators update if there is a fee recipient update event during this time.

olegshmuelov:
btw, while analyzing this I found 2 more issues:

**Issue 1: Non-deterministic Timing**

  The current blockSlot + 4 is problematic because:
  - Each operator might process the event at different times
  - Network delays, sync issues, or restarts could cause operators to calculate different duty slots
  - This breaks consensus even among NEW operators!

  Solution: Use the event's block number directly:
  ```go
  // Instead of current non-deterministic approach:
  dutySlot := blockSlot + 4  // BAD: depends on when operator processes event

  // Use deterministic approach:
  dutySlot := eventBlockSlot + someBuffer  // GOOD: same for all operators
```
  **Issue 2: ValidatorAdded Fee Recipients**

  Currently, new validators don't get immediate fee recipient registration. They have to wait for the deterministic slot, which could be up to 10 epochs!

  Solution: Add the same immediate registration logic to `handleValidatorAdded`:
  ```go
  // In handleValidatorAdded, after saving the share:
  if share != nil && eh.operatorDataStore.IsOperatorShare(share) {
      // Send to validatorRegistrationCh for immediate registration
      task := NewUpdateFeeRecipientTask(...)
  }
  ```

**Comment by iurii-ssv** [comment]
> @iurii-ssv The issue is the new immediate duty scheduling path:

Yes got it, that's a good thing to point out. We are on the same page about that, and it's not an issue (at least not an issue we can do something about).

> Issue 1: Non-deterministic Timing

@olegshmuelov I'm not sure I understand this issue, namely the way I implemented it currently mirrors exactly what we are doing for validator-exit duty (and I think 100% matches what you suggest) which is:
- every Operator receives an event (about fee-recipient update), they might receive this event at different times but the block-number in that event will be the same for all operators (regardless of whether event is received sooner/later, the event data is the same)
- every Operator takes that block number and fetches the actual Ethereum block (via `executionClient.BlockByNumber`), and they also should get the same block-data in response (even though they receive results at different times from their respective Ethereum nodes); unless there has been a fork in Ethereum network itself and depending on which fork their Ethereum client follows it might get a different block ... but that's probably not something we want to address here
- and in that block data every Operator has the same `block.Time()` value they use to derive `eventBlockSlot` that's consistent/same across ALL Operators
- and we add 4 slots on top just to make sure we don't "schedule duty for the slot that has already passed" (and if we do so occasionally that's unfortunate ... but not a huge deal)

> Issue 2: ValidatorAdded Fee Recipients

@olegshmuelov I'm actually lacking some context on how validator-add flow works with respect to fee-recipient, from what I see in the code:
- when `ValidatorAdded` event happens we don't even initialize fee-recipient to anything
- and yet, in order to compose & sign & broadcast validator-registration fee-recipient needs to be set
- and I don't see any default fee-recipient values being set for newly added validators ... so I'm not sure if/how validator-registrations even work for newly added-validators (without somebody triggering `FeeRecipientAddressUpdated` event externally)

so I'd need to know more about it if we want to add some handling for this in this PR

File: operator/duties/validator_registration.go (hunk 22 above)
Comment (iurii-ssv) [reply operator/duties/validator_registration.go]: I've added some tests for validator-registration duty, and those indeed require us to re-organize some of these constants similar to how you suggested (without the need for it I don't think it's worth changing it like that just because the code with concrete values in place is simpler to reason about, as opposed to highly-parametrized code).

I'll see what else I can add in terms of tests, but for now resolving this discussion to "keep the more important ones in focus

**Comment by olegshmuelov** [comment]
Re: Issue 1 (Non-deterministic Timing) - my bad, I missed that you're already using the event's block time to derive the slot. The implementation is good\! 👍

**Comment by olegshmuelov** [comment]
> 
> > Issue 1: Non-deterministic Timing
> 
> @olegshmuelov I'm not sure I understand this issue, namely the way I implemented it currently mirrors exactly what we are doing for validator-exit duty (and I think 100% matches what you suggest) which is:
> 
> * every Operator receives an event (about fee-recipient update), they might receive this event at different times but the block-number in that event will be the same for all operators (regardless of whether event is received sooner/later, the event data is the same)
> * every Operator takes that block number and fetches the actual Ethereum block (via `executionClient.BlockByNumber`), and they also should get the same block-data in response (even though they receive results at different times from their respective Ethereum nodes); unless there has been a fork in Ethereum network itself and depending on which fork their Ethereum client follows it might get a different block ... but that's probably not something we want to address here
> * and in that block data every Operator has the same `block.Time()` value they use to derive `eventBlockSlot` that's consistent/same across ALL Operators
> * and we add 4 slots on top just to make sure we don't "schedule duty for the slot that has already passed" (and if we do so occasionally that's unfortunate ... but not a huge deal)

 my bad, I missed that you're already using the event's block time to derive the slot. The implementation is good!

File: protocol/v2/ssv/runner/validator_registration.go [hunk 23]
```diff
@@ -248,18 +258,21 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
 
 	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
 
+	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
+	if err != nil {
+		return nil, fmt.Errorf("get recipient data from storage: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
+	}
```
olegshmuelov, line 278: @iurii-ssv I think we might have an issue here

  The change to use `recipientsStorage.GetRecipientData()` instead of the share's fee recipient breaks the default behavior for new validators.

  **The bug:**
  When a validator is added via `ValidatorAdded` event but no `FeeRecipientAddressUpdated` event is emitted, the validator registration will fail with "recipient data not found".

  **Previous behavior (working):**
  - New validators default to using the owner address as fee recipient
  - Validator registrations happen at the next deterministic slot (could be up to 10 epochs)

  **Current behavior (broken):**
  - `GetRecipientData()` returns not found error
  - Validator registration fails
  - These validators will never submit registrations

  This breaks a critical path where validators added without explicit fee recipient updates (which defaults to owner address) won't be able to register at all.

 We should address this together with [Issue 2](https://github.com/ssvlabs/ssv/pull/2130#issuecomment-3057551605) - perhaps redesign how we handle fee recipients to cover all cases
 
  Happy to discuss the approach if you'd like.

iurii-ssv: Regarding the bug you are describing here, I think we need to modify `setShareFeeRecipient` so it also updates `FeeRecipient` in `recipientsStorage` as well  - https://github.com/ssvlabs/ssv/pull/2130/commits/ca9c6a9367833f604953dc20e2642e2baca82b6f

As for "Issue 2" - I think it's better to handle in a separate PR (after we merge this one) since:
- this PR is getting large already as it is, and it's probably better to ship small/incremental upgrades rather than trying to fix everything right away
- I think the fix for "Issue 2" would look similar to how `FeeRecipientUpdate` event is extended in this PR, with maybe additional waiting for validator to finish its initialization so that things like `FeeRecipient` have meaningful values we can work with (that is to say - I don't think any major rewrites of what's implemented in https://github.com/ssvlabs/ssv/pull/2130 will be needed)

iurii-ssv:
Force-pushing to make sure every commit in this PR is gpg-signed (otherwise we cannot merge it).

protocol/v2/ssv/runner/validator_registration.go
y0sher: I agree with both of you. This could get worse in a scenario where an operator changes a lot of operators and registrations and getting filled. though not a very high likely scenario so it's not rushing to solve it now unless we'd see it has a noticeable effect.
So for now I'm resolving this but lets create an issue about this cache not cleaning.

protocol/v2/ssv/runner/validator_registration.go
y0sher: If this is still correct - that new validators that just been added won't have fee recipient address, and even more if previously we were setting it as the owner address. we need to address it make sure if we fail getting it from storage we just set the owner address or we set it in storage right away after validator added event.

protocol/v2/ssv/runner/validator_registration.go
iurii-ssv: Created https://github.com/ssvlabs/ssv/issues/2435 for now - so we don't lose this discussion.

protocol/v2/ssv/runner/validator_registration.go
iurii-ssv: Actually it looks like [the last commit I've added](https://github.com/ssvlabs/ssv/commit/ca9c6a9367833f604953dc20e2642e2baca82b6f) (sorry, I had to squash commits to make them signed, but the code is still in this PR's diff) addresses both:
- https://github.com/ssvlabs/ssv/pull/2130#discussion_r2203357316
- "Issue 2" - `ValidatorAdded` works exactly as it is in **stage** branch, but when new validator is added we also fetch & set metadata for that validator which results into `onShareStart` call (where we call `setShareFeeRecipient` which is now updated to `SaveRecipientData` to owner address, if needed)

@olegshmuelov could you confirm if that's correct (or am I missing something) ?

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...ba39a","c89772b6c110dc14","780c3281d914a06a","2153cdb8ff8acb14","24c638d05acc6258","37cdbf3fe0d2c0be","debb273257b832ac","b9f863429c5bd028","622dd7837dafba8e","bcacf2c847388455","2b6e432e076cde11","666f04b7d5e29eb","bec26aa6892bbca1","b006e20b3cb4f8bc","b6012c4075285778","a74d021f7b35fdb2","ae46fff3a49f773d","42214aa8ab8e4ab9","2ecfd0e26e23264c","af3b5d02752960e7","5e26dc96bd3c8e75","b7982786a04cc064","a6688aa594abdde6","fd1de2bf3d1bb8ea","b10c72c36fa90666","1c4a85421b52e46a","5ab7b7a9107c354f","f88e0383750294a6","6fc7c41b42b4e3e5","f28f3a41bc8c741d","b49179c36c111cca","5df9f55a706e5291","4994abbc55e58278","2c02276fa72f61b9","937ce36b0949796a","3f6f707d5ecf3903","11a340273beaf077","2630b7d3f7463316","33dd6bfa496bc889","a9685eff5edb9118","2fd0165ccc59ddf","de0733313f796fc9","a5458fa616bc6d6d","73244bf5a8b49e23","42052e9213a5780d","26672cc9a6ae0af0","b7ca13196a824fc5","801f7d8a8151c73e","20b4765e446c4803","25c4b276cfa24926","cebe87e9824d213e","be65f009ecd0d4b8","4d1a0519b660e5f","ad8332e881c48041","fbcd6b75d8baa5c1","787162ea04b0c4ed","c97a6f1c933535b","b5b9b0a45f6df79e","e9569ea2dc0f1406","5a3c840d9fc7bec9","7cf54a264ab481f3","cbbb16ca9342dc4c","46636a126e9d32ad","c90ffdb18233f0ea"]
```
iurii-ssv, line 1:
```
diff --git a/protocol/v2/ssv/validator/committee.go@Committee b/ssv/committee.go@Committee
index c97a6f1c933535b..2222222
--- a/protocol/v2/ssv/validator/committee.go@Committee
+++ b/ssv/committee.go@Committee
@@ -1,13 +1,6 @@
 Committee struct {
-	logger	*zap.Logger
-	ctx	context.Context
-	cancel	context.CancelFunc
-	beaconConfig	*networkconfig.Beacon
-	mtx	sync.RWMutex
-	Queues	map[phase0.Slot]QueueContainer
-	Runners	map[phase0.Slot]*runner.CommitteeRunner
-	Shares	map[phase0.ValidatorIndex]*Share
-	CommitteeMember	*CommitteeMember
-	dutyGuard	*CommitteeDutyGuard
-	CreateRunnerFn	CommitteeRunnerFunc
+	Runners		map[spec.Slot]*CommitteeRunner
+	CommitteeMember	CommitteeMember
+	CreateRunnerFn	CreateRunnerFn
+	Share		map[spec.ValidatorIndex]*Share
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/validator/committee.go@NewCommittee b/ssv/committee.go@NewCommittee
index b5b9b0a45f6df79e..2222222
--- a/protocol/v2/ssv/validator/committee.go@NewCommittee
+++ b/ssv/committee.go@NewCommittee
@@ -1,24 +1,13 @@
 func NewCommittee(
-	cancel context.CancelFunc,
-	committeeMember *CommitteeMember,
-	createRunnerFn CommitteeRunnerFunc,
-	shares map[phase0.ValidatorIndex]*Share,
-	dutyGuard *CommitteeDutyGuard,
+	committeeMember CommitteeMember,
+	share map[spec.ValidatorIndex]*Share,
+	createRunnerFn CreateRunnerFn,
 ) *Committee {
-	if shares == nil {
-		shares = make(map[phase0.ValidatorIndex]*Share)
-	}
-	return &Committee{
-		logger:			logger,
-		beaconConfig:		beaconConfig,
-		ctx:			ctx,
-		cancel:			cancel,
-		Queues:			make(map[phase0.Slot]QueueContainer),
-		Runners:		make(map[phase0.Slot]*runner.CommitteeRunner),
-		Shares:			shares,
+	c := &Committee{
+		Runners:		make(map[spec.Slot]*CommitteeRunner),
 		CommitteeMember:	committeeMember,
 		CreateRunnerFn:		createRunnerFn,
-		dutyGuard:		dutyGuard,
+		Share:			share,
 	}
+	return c
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/validator/validator.go@NewValidator b/ssv/validator.go@NewValidator
index e9569ea2dc0f1406..2222222
--- a/protocol/v2/ssv/validator/validator.go@NewValidator
+++ b/ssv/validator.go@NewValidator
@@ -1,31 +1,19 @@
-func NewValidator(pctx context.Context, cancel func(), options *Options) *Validator {
-	v := &Validator{
-		mtx:			&sync.RWMutex{},
-		ctx:			pctx,
-		cancel:			cancel,
-		NetworkConfig:		options.NetworkConfig,
-		DutyRunners:		options.DutyRunners,
-		Network:		options.Network,
-		Operator:		options.Operator,
-		Share:			options.SSVShare,
-		Signer:			options.Signer,
-		OperatorSigner:		options.OperatorSigner,
-		Queues:			make(map[RunnerRole]QueueContainer),
-		state:			uint32(NotStarted),
-		dutyIDs:		hashmap.New[RunnerRole, string](),
-		messageValidator:	options.MessageValidator,
+func NewValidator(
+	network Network,
+	beacon BeaconNode,
+	committeeMember *CommitteeMember,
+	share *Share,
+	signer BeaconSigner,
+	operatorSigner *OperatorSigner,
+	runners map[RunnerRole]Runner,
+) *Validator {
+	return &Validator{
+		DutyRunners:		runners,
+		Network:		network,
+		Beacon:			beacon,
+		Share:			share,
+		CommitteeMember:	committeeMember,
+		Signer:			signer,
+		OperatorSigner:		operatorSigner,
 	}
-	for _, dutyRunner := range options.DutyRunners {
-		dutyRunner.GetBaseRunner().TimeoutF = v.onTimeout
-		role := dutyRunner.GetBaseRunner().RunnerRoleType
-		v.Queues[role] = QueueContainer{
-			Q:	queue.New(options.QueueSize),
-			queueState: &queue.State{
-				HasRunningInstance:	false,
-				Height:			0,
-				Slot:			0,
-			},
-		}
-	}
-	return v
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner b/ssv/validator_registration.go@NewValidatorRegistrationRunner
index 5a3c840d9fc7bec9..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@NewValidatorRegistrationRunner
@@ -1,31 +1,25 @@
 func NewValidatorRegistrationRunner(
-	networkConfig *networkconfig.Network,
+	beaconNetwork BeaconNetwork,
 	share map[phase0.ValidatorIndex]*Share,
-	validatorOwnerAddress common.Address,
-	beacon beacon.BeaconNode,
+	beacon BeaconNode,
 	network Network,
-	signer ekm.BeaconSigner,
-	operatorSigner ssvtypes.OperatorSigner,
-	recipientsStorage recipientsStorage,
-	ValidatorRegistrationSubmitter ValidatorRegistrationSubmitter,
+	signer BeaconSigner,
+	operatorSigner *OperatorSigner,
 	gasLimit uint64,
 ) (Runner, error) {
 	if len(share) != 1 {
 	}
 	return &ValidatorRegistrationRunner{
 		BaseRunner: &BaseRunner{
 			RunnerRoleType:	RoleValidatorRegistration,
-			NetworkConfig:	networkConfig,
+			BeaconNetwork:	beaconNetwork,
 			Share:		share,
 		},
-		beacon:				beacon,
-		network:			network,
-		signer:				signer,
-		operatorSigner:			operatorSigner,
-		recipientsStorage:		recipientsStorage,
-		validatorRegistrationSubmitter:	ValidatorRegistrationSubmitter,
-		validatorOwnerAddress:		validatorOwnerAddress,
+		beacon:		beacon,
+		network:	network,
+		signer:		signer,
+		operatorSigner:	operatorSigner,
 		gasLimit:	gasLimit,
 	}, nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/validator/validator.go@Validator b/ssv/validator.go@Validator
index 7cf54a264ab481f3..2222222
--- a/protocol/v2/ssv/validator/validator.go@Validator
+++ b/ssv/validator.go@Validator
@@ -1,16 +1,9 @@
 Validator struct {
-	mtx	*sync.RWMutex
-	ctx	context.Context
-	cancel	context.CancelFunc
-	NetworkConfig	*networkconfig.Network
-	DutyRunners	runner.ValidatorDutyRunners
+	DutyRunners	DutyRunners
 	Network		Network
-	Operator	*CommitteeMember
-	Share		*ssvtypes.SSVShare
-	Signer		ekm.BeaconSigner
-	OperatorSigner	ssvtypes.OperatorSigner
-	Queues	map[RunnerRole]QueueContainer
-	dutyIDs	*hashmap.Map[RunnerRole, string]
-	state	uint32
-	messageValidator	validation.MessageValidator
+	Beacon		BeaconNode
+	CommitteeMember	*CommitteeMember
+	Share		*Share
+	Signer		BeaconSigner
+	OperatorSigner	*OperatorSigner
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner b/ssv/validator_registration.go@ValidatorRegistrationRunner
index cbbb16ca9342dc4c..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner
@@ -1,12 +1,9 @@
 ValidatorRegistrationRunner struct {
 	BaseRunner	*BaseRunner
-	beacon				beacon.BeaconNode
-	network				Network
-	signer				ekm.BeaconSigner
-	operatorSigner			ssvtypes.OperatorSigner
-	valCheck			ProposedValueCheckF
-	recipientsStorage		recipientsStorage
-	validatorRegistrationSubmitter	ValidatorRegistrationSubmitter
-	validatorOwnerAddress		common.Address
+	beacon		BeaconNode
+	network		Network
+	signer		BeaconSigner
+	operatorSigner	*OperatorSigner
+	valCheck	ProposedValueCheckF
 	gasLimit	uint64
 }
\ No newline at end of file
index 46636a126e9d32ad..2222222
@@ -1,48 +1,34 @@
-	ctx, span := tracer.Start(
-		observability.InstrumentName(observabilityNamespace, "runner.process_pre_consensus"),
-		trace.WithAttributes(
-			observability.BeaconSlotAttribute(signedMsg.Slot),
-			observability.ValidatorPartialSigMsgTypeAttribute(signedMsg.Type),
-			observability.ValidatorPublicKeyAttribute(phase0.BLSPubKey(r.GetShare().ValidatorPubKey)),
-		))
-	defer span.End()
-	hasQuorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
+	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
 	if err != nil {
-		return traces.Errorf(span, "failed processing validator registration message: %w", err)
+		return errors.Wrap(err, "failed processing validator registration message")
 	}
-	if !hasQuorum {
-		span.AddEvent("no quorum")
+	if !quorum {
 		return nil
 	}
 	root := roots[0]
-	span.AddEvent("reconstructing beacon signature", trace.WithAttributes(observability.BeaconBlockRootAttribute(root)))
 	if err != nil {
-		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
-		return traces.Errorf(span, "got pre-consensus quorum but it has invalid signatures: %w", err)
+		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee,
+			r.GetShare().ValidatorIndex)
+		return errors.Wrap(err, "got pre-consensus quorum but it has invalid signatures")
 	}
 	copy(specSig[:], fullSig)
 	registration, err := r.calculateValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
 	if err != nil {
-		return traces.Errorf(span, "could not calculate validator registration: %w", err)
+		return errors.Wrap(err, "could not calculate validator registration")
 	}
 			Message:	registration,
 			Signature:	specSig,
 		},
 	}
-	span.AddEvent("submitting validator registration")
-	if err := r.validatorRegistrationSubmitter.Enqueue(signedRegistration); err != nil {
-		return traces.Errorf(span, "could not submit validator registration: %w", err)
+		return errors.Wrap(err, "could not submit validator registration")
 	}
-	const eventMsg = "validator registration submitted successfully"
-	span.AddEvent(eventMsg)
 	r.GetState().Finished = true
 	return nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
index c90ffdb18233f0ea..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
@@ -1,26 +1,15 @@
 func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0.Slot) (*v1.ValidatorRegistration, error) {
-	pk := phase0.BLSPubKey{}
-	copy(pk[:], r.GetShare().ValidatorPubKey[:])
-	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
-	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
-	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
-	}
-	if !found {
-		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
-	}
-	gasLimit := r.gasLimit
-	if gasLimit == 0 {
-		defaultGasLimit := DefaultGasLimit
-		if !r.BaseRunner.NetworkConfig.GasLimit36Fork() {
-			defaultGasLimit = DefaultGasLimitOld
-		}
-		gasLimit = defaultGasLimit
+	share := r.GetShare()
+	if share == nil {
+		return nil, errors.New("no share to get validator public key")
 	}
+	pk := phase0.BLSPubKey{}
+	copy(pk[:], share.ValidatorPubKey[:])
+	epoch := r.BaseRunner.BeaconNetwork.EstimatedEpochAtSlot(slot)
 	return &v1.ValidatorRegistration{
-		FeeRecipient:	rData.FeeRecipient,
-		GasLimit:	gasLimit,
-		Timestamp:	r.BaseRunner.NetworkConfig.EpochStartTime(epoch),
+		FeeRecipient:	share.FeeRecipientAddress,
+		GasLimit:	r.gasLimit,
+		Timestamp:	r.BaseRunner.BeaconNetwork.EpochStartTime(epoch),
 		Pubkey:		pk,
 	}, nil
 }
\ No newline at end of file
```

PR #2133, opened 2025-04-08
fix(network/discovery): Correct `HostDNS` usage in discovery service

kchojn:


network/discovery/enode.go
```diff
@@ -33,26 +33,37 @@ func createLocalNode(privKey *ecdsa.PrivateKey, storagePath string, ipAddr net.I
...
-		}
-		localNode.SetFallbackIP(hostIP)
-		localNode.SetStaticIP(hostIP)
-	}
+	// Try DNS first
 	if len(hostDNS) > 0 {
 		ips, err := net.LookupIP(hostDNS)
 		if err != nil {
-			return errors.Wrap(err, "could not resolve host address")
+			return errors.Wrapf(err, "could not resolve host DNS: %s", hostDNS)
```
nkryuchkov, line 44: Can we avoid using `https://github.com/pkg/errors` in the new code as it's been deprecated for a while? Let's use `fmt.Errorf` instead

network/discovery/service_test.go
```diff
@@ -326,3 +324,100 @@ func TestDiscV5ServiceListenerType(t *testing.T) {
...
+		{
+			name:           "both empty",
+			hostAddress:    "",
+			hostDNS:        "",
+			expectedResult: "",
+			expectError:    false,
+		},
+	}
+
+	for i, tc := range testCases {
```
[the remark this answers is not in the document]
kchojn: bruh, they fixed it:
> for loops to make these variables have per-iteration scope instead of per-loop scope

network/discovery/enode.go
```diff
@@ -33,26 +33,37 @@ func createLocalNode(privKey *ecdsa.PrivateKey, storagePath string, ipAddr net.I
...
-		}
-		localNode.SetFallbackIP(hostIP)
-		localNode.SetStaticIP(hostIP)
-	}
+	// Try DNS first
 	if len(hostDNS) > 0 {
 		ips, err := net.LookupIP(hostDNS)
 		if err != nil {
-			return errors.Wrap(err, "could not resolve host address")
+			return fmt.Errorf("could not resolve host DNS: %s, %w", hostDNS, err)
```
oleg-ssvlabs, line 44: It seems we started to follow this pattern for returning errors across the codebase,  
so maybe you could also replace other usages of `errors.Wrap` in this file,  
so we can remove [this](https://github.com/ssvlabs/ssv/pull/2133/files#diff-c7b9a729d36756fa0f3d17cf48d89939649dfe09cf92fef3e6aef11ec40a81a8R13) reference completely.

There is an [errors package](https://pkg.go.dev/errors) in the stdlib, but I don’t think we favor it - isn’t that right, @moshe-blox @y0sher?

kchojn: I have replaced `pkg/errors` with `fmt.Errorf`

moshe-blox (approved):
looks good, but we'll need to remember to QA it with a real DNS to see that it actually works @Roy-blox @andrew-blox

network/discovery/enode.go
```diff
@@ -33,38 +32,49 @@ func createLocalNode(privKey *ecdsa.PrivateKey, storagePath string, ipAddr net.I
...
 			firstIP := ips[0]
+			localNode.SetStaticIP(firstIP)
 			localNode.SetFallbackIP(firstIP)
+
+			return nil
 		}
 	}
+
+	// Use HostAddress as fallback
+	if len(hostAddr) > 0 {
```
nkryuchkov, line 56: Should we handle the case when both `hostDNS` and `hostAddr` are empty?

nkryuchkov (approved):

network/discovery/enode.go
kchojn: no need to update these conditions. The node will handle it automatically, its also tested here: https://github.com/ssvlabs/ssv/blob/0bc5eeff0daef49ae053eba54c6822b8b71a1d11/network/discovery/service_test.go#L370

iurii-ssv (approved):

PR #2138, opened 2025-04-10
fix (logs): corrent total consensus time calculation

y0sher:


oleg-ssvlabs (approved):
I've noticed that we're not consistent in how durations are logged. Found two different approaches so far: 
- [First](https://github.com/ssvlabs/ssv/blob/main/logging/fields/fields.go#L284)
- [Second](https://github.com/ssvlabs/ssv/blob/main/logging/fields/fields.go#L292)

I think it would be better to use `zap.Duration` everywhere instead of implicitly converting durations to seconds.

**Examples:**
- 5 seconds with `zap.Duration()`: `"5s"`
- 5 seconds with formatted string: `"5.00000"`

I think option 1 looks better. We could either implement this change in this PR or create an issue. Wdyt @y0sher ?

iurii-ssv (approved):

moshe-blox (approved):
lgtm 🔥

moshe-blox:
@oleg-ssvlabs the reason we convert durations to seconds is because its really hard to compare durations across logs when they're in different units of time (us/ms/s)

nkryuchkov (approved):

PR #2139, opened 2025-04-10
eth/executionclient: multi client metrics

nkryuchkov:
Add metrics to track execution client switching and healthiness in the EL multi-client implementation

- Client switch counter
- Method call metrics (count/errors/duration)
- Gauge for healthy client count
- Init success/failure tracking

### DIFF
--- a/eth/executionclient/observability.go
+++ b/eth/executionclient/observability.go
@@ -52,6 +52,45 @@ var (
 			metricName("sync.last_processed_block"),
 			metric.WithUnit("{block_number}"),
 			metric.WithDescription("last processed block by execution client")))
+
+	// MultiClient metrics
+	clientSwitchCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("client.switch"),
+			metric.WithDescription("number of times the execution client has been switched")))
+
+	multiClientMethodCallsCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("multi_client.method_calls"),
+			metric.WithDescription("number of method calls to the multi client")))
+
+	multiClientMethodErrorsCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("multi_client.method_errors"),
+			metric.WithDescription("number of method call errors in the multi client")))
+
+	multiClientMethodDurationHistogram = observability.NewMetric(
+		meter.Float64Histogram(
+			metricName("multi_client.method_duration"),
+			metric.WithUnit("s"),
+			metric.WithDescription("multi client method call duration in seconds"),
+			metric.WithExplicitBucketBoundaries(observability.SecondsHistogramBuckets...)))
+
+	healthyClientsGauge = observability.NewMetric(
+		meter.Int64Gauge(
+			metricName("multi_client.healthy_clients"),
+			metric.WithUnit("{clients}"),
+			metric.WithDescription("number of healthy clients in the multi client")))
+
+	clientInitCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("client.init"),
+			metric.WithDescription("number of times a client was initialized")))
+
+	clientInitErrorCounter = observability.NewMetric(
+		meter.Int64Counter(
+			metricName("client.init_error"),
+			metric.WithDescription("number of times a client initialization failed")))
 )
 
 func metricName(name string) string {
@@ -85,5 +124,50 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
 			metric.WithAttributes(semconv.ServerAddress(nodeAddr)),
 			metric.WithAttributes(executionClientStatusAttribute(status)),
 		)
+	}
+}
+
+// recordClientSwitch records a client switch event
+func recordClientSwitch(ctx context.Context, fromAddr, toAddr string) {
+	clientSwitchCounter.Add(ctx, 1,
+		metric.WithAttributes(
+			attribute.String("from_addr", fromAddr),
+			attribute.String("to_addr", toAddr),
+		),
+	)
+}
+
+// recordMultiClientMethodCall records metrics for multi client method calls
+func recordMultiClientMethodCall(ctx context.Context, method string, nodeAddr string, duration time.Duration, err error) {
+	attrs := []attribute.KeyValue{
+		attribute.String("method", method),
+		semconv.ServerAddress(nodeAddr),
+	}
+
+	multiClientMethodCallsCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+	multiClientMethodDurationHistogram.Record(ctx, duration.Seconds(), metric.WithAttributes(attrs...))
+
+	if err != nil {
+		errorAttrs := append(attrs, attribute.String("error", err.Error()))
+		multiClientMethodErrorsCounter.Add(ctx, 1, metric.WithAttributes(errorAttrs...))
+	}
+}
+
+// recordHealthyClientsCount records the number of healthy clients
+func recordHealthyClientsCount(ctx context.Context, count int64) {
+	healthyClientsGauge.Record(ctx, count)
+}
+
+// recordClientInit records metrics for client initialization
+func recordClientInit(ctx context.Context, nodeAddr string, success bool, err error) {
+	attrs := []attribute.KeyValue{
+		semconv.ServerAddress(nodeAddr),
+	}
+
+	clientInitCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+
+	if !success {
+		errorAttrs := append(attrs, attribute.String("error", err.Error()))
+		clientInitErrorCounter.Add(ctx, 1, metric.WithAttributes(errorAttrs...))
 	}
 }
--- a/eth/executionclient/multi_client.go
+++ b/eth/executionclient/multi_client.go
@@ -160,6 +160,7 @@ func (mc *MultiClient) connect(ctx context.Context, clientIndex int) error {
 		WithSyncDistanceTolerance(mc.syncDistanceTolerance),
 	)
 	if err != nil {
+		recordClientInit(ctx, mc.nodeAddrs[clientIndex], false, err)
 		return fmt.Errorf("create single client: %w", err)
 	}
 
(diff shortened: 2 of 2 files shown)


eth/executionclient/observability.go
oleg-ssvlabs, line 1: I’m not sure if there’s a clean way to do this.
What do you think about determining whether it’s a multi-client based on a metric attribute (label) rather than including it in the metric name? I know this might be tricky, since it seems some metrics may exist only for multi-client setups and not for regular ones. Wdyt @nkryuchkov ?

eth/executionclient/observability.go
```diff
@@ -87,3 +126,48 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+// recordClientInit records metrics for client initialization
+func recordClientInit(ctx context.Context, nodeAddr string, success bool, err error) {
+	attrs := []attribute.KeyValue{
+		semconv.ServerAddress(nodeAddr),
+	}
+
+	clientInitCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+
+	if !success {
+		errorAttrs := append(attrs, attribute.String("error", err.Error()))
```
oleg-ssvlabs, line 170: I think this attribute may have very high cardinality, so I’d prefer not to include it

y0sher: @nkryuchkov @oleg-ssvlabs to be honest I'm not sure what is the best approach, error messages should be of some limited cardinality. but it does makes sense to include the error in the metric, one option is to maybe use an enum of error type and make the relations in the dashboard.
https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-metrics

eth/executionclient/observability.go
```diff
@@ -87,3 +126,48 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+
+	multiClientMethodCallsCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+	multiClientMethodDurationHistogram.Record(ctx, duration.Seconds(), metric.WithAttributes(attrs...))
+
+	if err != nil {
+		errorAttrs := append(attrs, attribute.String("error", err.Error()))
+		multiClientMethodErrorsCounter.Add(ctx, 1, metric.WithAttributes(errorAttrs...))
+	}
+}
+
```
zktaiga, line 155: I think what would make multiclient metrics truly useful would be to have the client name/address as an attribute. In its current state, we can't easily compare client performance.

eth/executionclient/observability.go
```diff
@@ -87,3 +126,48 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+// recordHealthyClientsCount records the number of healthy clients
+func recordHealthyClientsCount(ctx context.Context, count int64) {
+	healthyClientsGauge.Record(ctx, count)
+}
```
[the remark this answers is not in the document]
y0sher: I agree it's best to track all clients and just mention which is healthy at which time.

eth/executionclient/observability.go
y0sher: yes @nkryuchkov this was basically the essence of this task, to be able to compare two clients across one or multiple nodes.

nkryuchkov: @zktaiga @y0sher I'm not sure I understand what you mean by having address as an attribute, it's in `nodeAddr` and it is an attribute. Actually, there's a mistake in the attribute key, `ServerAddress` should be `ClientAddress`, is that what you mean?

eth/executionclient/observability.go
nkryuchkov: IMO we can just get rid of it because the only possible errors are various connection issues. Removed

eth/executionclient/observability.go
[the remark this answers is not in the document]
nkryuchkov: I added a total count. What should the mapping of which one is healthy look like?

eth/executionclient/observability.go
nkryuchkov: The metrics that belong to the multi-client are method-related or healthy count/total count, so they can't be done for a single client

y0sher (approved):

eth/executionclient/observability.go
y0sher: yup @zktaiga it seems that the address is already as an attribute `semconv.ServerAddress(nodeAddr),`

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+			attribute.String("to_addr", toAddr),
+		),
+	)
+}
+
+// recordMultiClientMethodCall records metrics for multi client method calls
+func recordMultiClientMethodCall(ctx context.Context, method string, nodeAddr string, duration time.Duration, err error) {
+	attrs := []attribute.KeyValue{
+		attribute.String("method", method),
+		semconv.ServerAddress(nodeAddr),
```
zktaiga, line 150: I'd change this attribute to `client_address`.

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+
+// recordAllClientsCount records the number of clients
+func recordAllClientsCount(ctx context.Context, count int64) {
+	allClientsGauge.Record(ctx, count)
+}
+
+// recordClientInit records metrics for client initialization
+func recordClientInit(ctx context.Context, nodeAddr string, success bool, err error) {
+	attrs := []attribute.KeyValue{
+		semconv.ServerAddress(nodeAddr),
```
zktaiga, line 174: Same here.

zktaiga (approved):

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+	healthyClientsGauge.Record(ctx, count)
+}
+
+// recordAllClientsCount records the number of clients
+func recordAllClientsCount(ctx context.Context, count int64) {
+	allClientsGauge.Record(ctx, count)
+}
+
+// recordClientInit records metrics for client initialization
+func recordClientInit(ctx context.Context, nodeAddr string, success bool, err error) {
```
oleg-ssvlabs, line 172: You can remove the `err` parameter from this function because it is not used anymore

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+	attrs := []attribute.KeyValue{
+		semconv.ServerAddress(nodeAddr),
+	}
+
+	clientInitCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+
+	if !success {
+		clientInitErrorCounter.Add(ctx, 1)
```
oleg-ssvlabs, line 180: You could also add a server address attribute here

eth/executionclient/observability.go
```diff
@@ -52,6 +52,51 @@ var (
...
+			metric.WithUnit("{clients}"),
+			metric.WithDescription("number of healthy clients in the multi client")))
+
+	allClientsGauge = observability.NewMetric(
+		meter.Int64Gauge(
+			metricName("multi_client.all_clients"),
+			metric.WithUnit("{clients}"),
+			metric.WithDescription("number of clients in the multi client")))
+
+	clientInitCounter = observability.NewMetric(
```
oleg-ssvlabs, line 91: Both `clientInitCounter` and `clientInitErrorCounter` are recorded in one place [here](https://github.com/ssvlabs/ssv/pull/2139/files#diff-f4616f8348b655889155b26d0977e95aa5f8834b9bd2f3659691bcf9aef959f8R172).  
What if, instead of having two separate metrics, we added an attribute to indicate whether the client initialization succeeded or failed?

The attribute could be a boolean, such as `ssv.el.init.status` or something similar (with namespace)

eth/executionclient/observability.go
oleg-ssvlabs: However I think we could consider merging two metrics into one as I mentioned [here](https://github.com/ssvlabs/ssv/pull/2139/files#r2057803385)

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+			attribute.String("from_addr", fromAddr),
+			attribute.String("to_addr", toAddr),
+		),
+	)
+}
+
+// recordMultiClientMethodCall records metrics for multi client method calls
+func recordMultiClientMethodCall(ctx context.Context, method string, nodeAddr string, duration time.Duration, err error) {
+	attrs := []attribute.KeyValue{
+		attribute.String("method", method),
```
oleg-ssvlabs, line 149: I would add a namespace to this attribute — something like `ssv.el.method`.

We usually add namespaces to attributes as well, since it's possible to search through them independently.  
So it's not really that the attribute "belongs to a metric" — they live on their own, and having a namespace provides better context.

eth/executionclient/observability.go
```diff
@@ -87,3 +132,51 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
 		)
 	}
 }
+
+// recordClientSwitch records a client switch event
+func recordClientSwitch(ctx context.Context, fromAddr, toAddr string) {
+	clientSwitchCounter.Add(ctx, 1,
+		metric.WithAttributes(
+			attribute.String("from_addr", fromAddr),
```
oleg-ssvlabs, line 140: same as [here](https://github.com/ssvlabs/ssv/pull/2139/files#r2057836695)

eth/executionclient/observability.go
```diff
@@ -87,3 +127,63 @@ func resetExecutionClientStatusGauge(ctx context.Context, nodeAddr string) {
...
+// recordAllClientsCount records the number of clients
+func recordAllClientsCount(ctx context.Context, count int64) {
+	allClientsGauge.Record(ctx, count)
+}
+
+// recordClientInitStatus records metrics for client initialization
+func recordClientInitStatus(ctx context.Context, nodeAddr string, success bool) {
+	attrs := []attribute.KeyValue{
+		semconv.ServerAddress(nodeAddr),
+		attribute.Key("success").Bool(success),
```
oleg-ssvlabs, line 185: I would add some namespace to this attribute as well.
Also, you could use `attribute.Bool("key", value)`

oleg-ssvlabs (approved):
🔥

PR #2142, opened 2025-04-11
feat(ssvsigner): more tests

kchojn:


ssvsigner/keystore/file_test.go
```diff
@@ -18,45 +18,91 @@ const (
 	testPubKeyBase64 = "base64EncodedPublicKey"
 )
 
-func TestDecryptKeystoreWithInvalidData(t *testing.T) {
-	encryptedJSONData := []byte(`{"version":4,"pubKey":"` + testPubKeyBase64 + `","crypto":{"kdf":"scrypt","checksum":{"function":"sha256","params":{"dklen":32,"salt":"base64EncodedSalt"},"message":"base64EncodedMessage"},"cipher":{"function":"aes-128-ctr","params":{"iv":"base64EncodedIV"},"message":"base64EncodedEncryptedMessage"},"kdfparams":{"n":262144,"r":8,"p":1,"salt":"base64EncodedSalt"}}}`)
-	_, err := DecryptKeystore(encryptedJSONData, testPassword)
-	require.Error(t, err)
+func TestMain(m *testing.M) {
+	require.NoError(&testing.T{}, bls.Init(bls.BLS12_381))
```
nkryuchkov, line 22: I think using `&testing.T` is not reliable and may theoretically panic or lead to unexpected results.

I wrote a simple test using both and the test using `&testing.T{}` passes although it should fail:

```go
func Test_testing_T(t *testing.T) {
	t.Run("1", func(t *testing.T) {
		assert.NoError(&testing.T{}, fmt.Errorf("test"))
	})
	t.Run("2", func(t *testing.T) {
		assert.NoError(t, fmt.Errorf("test"))
	})
}
```

```
=== RUN   Test_testing_T
=== RUN   Test_testing_T/1
--- PASS: Test_testing_T/1 (0.00s)
=== RUN   Test_testing_T/2
    file_test.go:33: 
        	Error Trace:	/Users/bloxmac/GolandProjects/ssv/ssvsigner/keystore/file_test.go:33
        	            	test
        	Test:       	Test_testing_T/2
--- FAIL: Test_testing_T/2 (0.00s)

--- FAIL: Test_testing_T (0.00s)
```

Let's do something like this:
```suggestion
	if err := bls.Init(bls.BLS12_381); err != nil {
		t.Fatal(err)
	}
```

nkryuchkov (changes requested):
Great work, just one request

ssvsigner/keystore/file_test.go
kchojn: fixed, i updated the testmain function -  this approach guarantees that initialization errors are detected and handled properly

ssvsigner/ekm/local_key_manager_test.go
```diff
@@ -6,14 +6,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ssvlabs/ssv-spec/types/testingutils"
```
nkryuchkov, line 9: not sorted

nkryuchkov (approved):

PR #2143, opened 2025-04-11
OperatorData: change PublicKey type to string

### LINKED ISSUE #2119: OperatorData: change `PublicKey` type to `string`
We currently store it as a base64 string in a `[]byte` type, so we almost always have to convert it to `string`. Also, `string` would be more correct semantically

nkryuchkov:
Closes #2119 

The public key is base64 string, so it's semantically better to change it to string. 

It's stored in DB as JSON where the public key is encoded as []byte, so the PR implements custom MarshalJSON/UnmarshalJSON to keep the same format for compatibility


iurii-ssv (approved):
For sure making **base64** encoding a string (instead of **[]byte**) - like this PR does - is a better way to go 👍

nkryuchkov:
@y0sher no because of custom `(o OperatorData) MarshalJSON()` and `(o OperatorData) UnmarshalJSON()` methods which keep the same format. But yes, we need to test it

PR #2149, opened 2025-04-15
fix:automatic retry of validator registration slot error

2019jack:
### Problem Background

During the validator registration process, the current system is prone to "invalid partial sig slot" errors due to node clock synchronization issues, which are manifested as follows:
- The slot of the partial signature message does not match the currently processed slot
- The validator registration failed, resulting in the relay server not recognizing the validator
- Ultimately, the validator cannot successfully generate blocks in the specified slot

Through log analysis, we observed that the validator 0x85aaf09e... broadcasted the registered partial signature in slot 11487919, but because the slot in the message (11487599) did not match the currently processed slot, the verification failed, which in turn affected the subsequent block generation of slot 11488351.

### Solution

Implement an automatic retry mechanism. When the verification registration fails due to slot mismatch, the registration process is automatically retried in the next slot. Specifically include:

1. Add `scheduleRetry` method in `ValidatorRegistrationRunner` to capture and handle slot mismatch errors

2. Add `Reset` method in `BaseRunner` to support state reset for retry

3. Use dynamic waiting time strategy to optimize retry waiting time according to the remaining time of the slot

### Implementation details

1. Modify `ProcessPreConsensus` method to detect "invalid partial sig slot" error and trigger retry

2. Implement `scheduleRetry` method to create a new registration task for the next slot and execute it

3. Add `Reset` method to clear the current Runner state

4. Use the smart waiting strategy `math.Min(100*time.Millisecond, remainingTime/10)`

### Test suggestions

1. Simulate the scenario of clock asynchrony between nodes to verify whether the retry mechanism works properly

2. Check the log to confirm whether the retry is correctly triggered when the "invalid partial sig slot" error occurs

3. Verify whether the system performance is affected under high load
4. Test the system's recovery ability under extreme situations (such as multiple consecutive failures)

### Expected results

1. Reduce the failure rate of validator registration due to timing problems
2. Increase the probability of normal block generation by validators
3. Reduce the need for manual intervention
4. The system handles clock synchronization differences between nodes more stably

Ensure that validators can complete block propagation within the specified 1 second and improve the overall efficiency of the network.

### DIFF
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -338,6 +338,17 @@ func (b *BaseRunner) hasRunningDuty() bool {
 	return !b.State.Finished
 }
 
+// Reset clears the current runner state to prepare for a retry
+func (b *BaseRunner) Reset() {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	
+	if b.State != nil {
+		b.State.Finished = true
+		b.State = nil
+	}
+}
+
 func (b *BaseRunner) ShouldProcessDuty(duty spectypes.Duty) error {
 	if b.QBFTController.Height >= specqbft.Height(duty.DutySlot()) && b.QBFTController.Height != 0 {
 		return errors.Errorf("duty for slot %d already passed. Current height is %d", duty.DutySlot(),
--- a/protocol/v2/ssv/runner/validator_registration.go
+++ b/protocol/v2/ssv/runner/validator_registration.go
@@ -19,6 +19,9 @@ import (
 	"github.com/ssvlabs/ssv/logging/fields"
 	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
+	"strings"
+	"time"
+	"math"
 )
 
 type ValidatorRegistrationRunner struct {
@@ -75,6 +78,12 @@ func (r *ValidatorRegistrationRunner) HasRunningDuty() bool {
 func (r *ValidatorRegistrationRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Logger, signedMsg *spectypes.PartialSignatureMessages) error {
 	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
 	if err != nil {
+		logger.Error("pre-consensus msg processing failed", zap.Error(err))
+		
+		if strings.Contains(err.Error(), "invalid partial sig slot") {
+			r.scheduleRetry(ctx, logger)
+		}
+		
 		return errors.Wrap(err, "failed processing validator registration message")
 	}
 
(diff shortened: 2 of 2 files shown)


iurii-ssv:
Hi @2019jack thanks for looking into it, while the issues you describe require some considerations and extra handling - let me add some context so we are on the same page on what actually happens vs what we are doing to address it,

every Operator fetches duties it needs to perform (Validator Registration for example) independently from other Operators based on his own scheduler (based on his local clock), 

every Operator indeed runs on it's own machine with it's own local clock - I guess our basic assumption is that clock differences between different Operators shouldn't diverge too much (ideally it should be on the order of 100ms), but even if the clock difference is something like couple of seconds SSV cluster still should work fine because 1 Ethereum slot lasts 12 seconds

this is because what happens is:
- once slot N comes around (from some Operator's clock perspective) Operator enqueues new duty for execution
- Operator's duty-runner pops duties off that queue and tries to execute them sequentially one after another (keeping track of the current/processed slot per duty-type all received p2p messages are verified against) - this ensures all operators are roughly on the same page as long as their clocks are reasonably in-sync
- in parallel Operator receives messages from other Operators, ideally duty-runner should have already picked up a duty (for some duty-type and slot N) before receiving corresponding messages for that duty-type and slot N from other Operators ... but that might not always be the case - if a message from another Operator arrives earlier (due to local clock differences) I think our own Operator will drop/reject it with "invalid partial sig slot"

I think this issue isn't specific to Validator Registration duty but rather affects all duty types, and probably applies to both **PreConsensus** and **Consensus** QBFT phases (as I've noted here - https://github.com/ssvlabs/ssv/issues/1827#issuecomment-2444180331)

From what I can tell the solution outlined in this PR wouldn't 100% suite us because:
- rather than retrying on the next slot duty-runner should wait a bit on the current slot and retry processing the message (because like I explained above the difference can't be multiple slots, it shouldn't be more than couple of seconds at most)
- we shouldn't "block" duty-runner for long because [it processes his queue sequentially](https://github.com/ssvlabs/ssv/blob/0102d22fdda93da42166b37e06778b58275a3b75/protocol/v2/ssv/validator/validator.go#L121-L121) (that includes messages from other Operators as well as duties he schedules on his own) - if we do "block" naively we could be delaying duty-runner for longer than necessary (we should sort of skip the message processing re-enqueueing it or something for duty-runner to pick up again later)
- I don't think we should `Reset` duty-runner (because if we do we might be "breaking current duty processing from previous slot that's still in progress" I think), and creating new duty in the duty-runner(be it `BaseRuner` or any specific `Runner`) itself sort of "on-demand" is also probably not the simplest way to do it (the simplest way would probably be re-enqueueing the message for duty-runner to pick up again later, like I mentioned above)

cc @olegshmuelov @y0sher @moshe-blox 

@2019jack as for the slot values you've mentioned in PR description I suspect 
- you either have a very large clock-drift (on the order of minutes if not hours) - which means you should look into addressing that instead
- or maybe you experience some different issue - which we can only debug if you can provide various logs, probably involving multiple operators

PR #2150, opened 2025-04-15
feat(db): add pebbledb support to 'improvements' branch

anatolie-ssv:
Put pebbledb support on top of `v2.2.1-unstable.2` changeset

deployed to stage exporter-2: https://grafana.ops.ssvlabsinternal.com/goto/VxWc2GJHR?orgId=1

deployed to:
- ssv-node-45-46(badger): https://grafana.ops.ssvlabsinternal.com/goto/lXbJkM1Ng?orgId=1
- ssv-node-47-48(pebble): https://grafana.ops.ssvlabsinternal.com/goto/gqS-kM1HR?orgId=1


cli/operator/node.go
```diff
@@ -585,6 +597,39 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
+	// If migrations were applied, we run a full garbage collection cycle
+	// to reclaim any space that may have been freed up.
+	start := time.Now()
+	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
+	defer cancel()
+	if err := db.FullGC(ctx); err != nil {
+		return nil, errors.Wrap(err, "failed to collect garbage")
+	}
+
+	logger.Debug("post-migrations garbage collection completed", fields.Duration(start))
```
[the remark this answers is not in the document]
anatolie-ssv: well - copilot - you're wrong here...

moshe-blox:
is it possible to separate the implementations to two different packages?

- rename storage/kv to storage/badger
- move pebble code to storage/pebble

this is a good practice in Go: one package for the interfaces, and one package for each implementation

also this should leave storage/kv package untouched (just renamed) which is easier to review

config/config.example.yaml
```diff
@@ -8,6 +8,8 @@ global:
 db:
   # Path to a persistent directory to store the node's database.
   Path: ./data/db
+  # Type is the database type, currently only badger and pebble are supported
+  Engine: badger
```
moshe-blox, line 12: lets pls hide this from examples & docs until we test it properly with operator nodes (in the future)

for now its for exporter only and we dont want operators to play with it yet :)

y0sher: shouldn't we make pebble work only with exporter for now then?

cli/operator/node.go
```diff
@@ -585,6 +597,39 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
+		return nil, errors.Wrap(err, "failed to run migrations")
+	}
+	if applied == 0 {
+		return db, nil
+	}
+
+	// If migrations were applied, we run a full garbage collection cycle
+	// to reclaim any space that may have been freed up.
+	start := time.Now()
+	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
```
oleg-ssvlabs, line 622: I think it is better to use `cfg.DBOptions.Ctx` here instead of `context.Background()` or, as I mention [here](https://github.com/ssvlabs/ssv/pull/2150/files#r2047019006), just use the context passed as parameter

oleg-ssvlabs: Also, is 6 minutes reasonable timeout? It may block the execution(start up) flow of the node for up to 6 minutes, it sounds like a lot. Have you tested how long does it take on average to complete this task?

cli/operator/node.go
```diff
@@ -585,6 +597,39 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
...
+	defer cancel()
+	if err := db.FullGC(ctx); err != nil {
+		return nil, errors.Wrap(err, "failed to collect garbage")
```
oleg-ssvlabs, line 625: better to use `fmt.Errorf()` to follow our conventional way of returning wrapped errors

cli/operator/node.go
oleg-ssvlabs, line 628: I'm thinking it might be a good idea to also log the duration in cases where `db.FullGC()` returns an error, so we can see how long it took before it failed.

anatolie-ssv: added

cli/operator/node.go
anatolie-ssv: added

cli/operator/node.go
anatolie-ssv: added

cli/operator/node.go
```diff
@@ -585,6 +597,39 @@ func setupDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.Badger
 	return db, nil
 }
 
+func setupPebbleDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*kv.PebbleDB, error) {
```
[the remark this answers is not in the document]
anatolie-ssv: it seems there's some more logic around `cfg.DBOptions.Ctx` and I'm currently not sure it's a valid substitute.

config/config.example.yaml
anatolie-ssv: removed

cli/operator/node.go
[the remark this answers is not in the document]
y0sher: we are using `cfg.DBOptions.Ctx` in the badger DB function, I agree passing down context is better, but I think we should keep the 2 funcs similar or even converge some of the code to the same function if we can.

storage/pebble/pebble.go
```diff
@@ -0,0 +1,258 @@
+package kv
```
oleg-ssvlabs, line 1: should the package name match the directory name (pebble)?

storage/pebble/pebble.go
```diff
@@ -0,0 +1,258 @@
...
+
+// PebbleDB struct
+type PebbleDB struct {
+	logger *zap.Logger
+
+	db *pebble.DB
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	//wg     sync.WaitGroup
```
oleg-ssvlabs, line 24: maybe we could remove the commented out code since it isn't used?

moshe-blox:
image
thats great, thanks @anatolie-ssv  :)

moshe-blox:
@anatolie-ssv no GC required with Pebble?

storage/pebble/pebble.go
```diff
@@ -0,0 +1,258 @@
...
+	if errors.Is(err, pebble.ErrNotFound) {
+		return basedb.Obj{}, false, nil
+	}
+	if err != nil {
+		return basedb.Obj{}, false, err
+	}
+	// PebbleDB returned slice is valid until closer.Close() is called
+	// hence we copy
+	//
+	defer func() { _ = closer.Close() }()
```
oleg-ssvlabs, line 62: maybe it is better to handle the error here? We could at least log, so it doesn't go unnoticed

cli/operator/node.go
```diff
@@ -552,8 +565,41 @@ func setupGlobal() (*zap.Logger, error) {
...
+		return nil, errors.Wrap(err, "failed to collect garbage")
+	}
+
+	logger.Debug("post-migrations garbage collection completed", fields.Duration(start))
+
+	return db, nil
+}
+
+func setupPebbleDB(logger *zap.Logger, eth2Network beaconprotocol.Network) (*pebble.PebbleDB, error) {
+	db, err := pebble.NewPebbleDB(cfg.DBOptions.Ctx, logger, cfg.DBOptions.Path, &cpebble.Options{})
```
moshe-blox, line 602: the risk with engine support is that when the user switches from badger to pebble it might overwrite the badgerdb files and corrupt it

lets think of a way to prevent the user from shooting himself in the foot

i think the simplest might be to append "pebble" to the db path with filepath.Join?

anatolie-ssv:
> @anatolie-ssv no GC required with Pebble?

No, from what I read:
```PebbleDB automatically manages compaction in the background. When certain conditions are met (e.g., the number of sstables at a given level exceeds a threshold, or the disk space usage grows)```

and it doesn't expose anything except a `.Compact` method that takes ranges of keys, and should only be used when targeting specific performance degradation.

File: storage/pebble/pebble.go (hunk 8 above)
Comment (oleg-ssvlabs) [reply storage/pebble/pebble.go:62]: [here](https://github.com/ssvlabs/ssv/pull/2150/files#diff-dd2d0ee1b9bfc18f7d28ac64476087cb4f77532d304c857c8e9db8d0471f178dR100) the call is not deferred, so could we not defer it here as well so we can properly process the error (return the error to the caller)

File: storage/pebble/pebble.go (hunk 7 above)
Comment (anatolie-ssv) [reply storage/pebble/pebble.go]: fixed

File: config/config.example.yaml (hunk 2 above)
Comment (y0sher) [reply config/config.example.yaml]: MO this is not fully result, if we really never want to see a case where someone 'accidentally' uses pebble for an operator node, we should restrict it and not just not mention it.
wdyt? @anatolie-ssv @moshe-blox @MatusKysel

File: cli/operator/node.go (hunk 9 above)
Comment (anatolie-ssv) [reply cli/operator/node.go:603]: added `-pebble` suffix to the folder name

File: config/config.example.yaml (hunk 2 above)
Comment (iurii-ssv) [reply config/config.example.yaml]: > MO this is not fully resolved, if we really never want to see a case where someone 'accidentally' uses pebble for an operator node, we should restrict it and not just not mention it.

just as counter-argument:
- and typically using "undocumented" feature is considered "at your own risk" (and it's gonna be an intentional usage because it's impossible to set **pebble** here instead of **badger** "by accident")
- does anything prevent operator from forking the code-base and removing any constraining check we might add if they really want to try it out ?

File: storage/pebble/pebble.go [hunk 10]
```diff
@@ -0,0 +1,283 @@
...
+var _ basedb.Database = &PebbleDB{}
+
+// PebbleDB struct
+type PebbleDB struct {
+	logger *zap.Logger
+
+	db *pebble.DB
+
+	ctx    context.Context
+	cancel context.CancelFunc
```
iurii-ssv, line 23: Looks like this `ctx` (and `cancel`) isn't used ?

storage/pebble/pebble.go
```diff
@@ -0,0 +1,283 @@
...
+	iter, err := makePrefixIter(pdb.db, prefix)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err := iter.Close(); err != nil {
+			pdb.logger.Error("close iterator on get all", zap.Error(err))
+		}
+	}()
```
iurii-ssv, line 149: Why not propagate `err` instead of logging it (like you are doing in `FullGC`) ? We probably should handle it the same way for all cases

storage/pebble/pebble.go
```diff
@@ -0,0 +1,283 @@
...
+	return batch.Commit(pebble.Sync)
+}
+
+func (pdb *PebbleDB) Update(fn func(basedb.Txn) error) error {
+	batch := pdb.db.NewIndexedBatch()
+	txn := newPebbleTxn(batch, pdb)
+	if err := fn(txn); err != nil {
+		return err
+	}
+	return batch.Commit(pebble.Sync)
```
iurii-ssv, line 247: This looks a bit unnecessarily hard to read to me, can't we refactor it into something like this (by rearranging some code a bit) ?
```go
func (pdb *PebbleDB) SetMany(prefix []byte, n int, next func(int) (basedb.Obj, error)) error {
	txn := pdb.NewTxn()
	if err := txn.SetMany(prefix, n, next); err != nil {
		return err
	}
	return txn.Commit()
}
```

I guess this applies to other methods as well.

storage/pebble/txn.go
```diff
@@ -0,0 +1,127 @@
...
+				continue
+			}
+			return err
+		}
+
+		valCopy := make([]byte, len(value))
+		copy(valCopy, value)
+		if err := closer.Close(); err != nil {
+			return err
+		}
```
iurii-ssv, line 79: It seems here in `GetMany` we propagate error up the stack from `closer.Close()` while in `Get` we skip such an error ... shouldn't we handle both cases the same way (probably propagate errors) ?

storage/pebble/txn.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (t *pebbleTxn) GetAll(prefix []byte, fn func(int, basedb.Obj) error) error {
+	iter, err := makePrefixIter(t.batch, prefix)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = iter.Close() }()
+	i := 0
+	// SeekPrefixGE starts prefix iteration mode.
```
iurii-ssv, line 99: I think this comment:
- either should have more info in it (on what it's about)
- or shouldn't be here since it is an implementation detail this code doesn't care about

note, there are 2 places with this comment

storage/pebble/txn.go
```diff
@@ -0,0 +1,127 @@
...
+	for iter.First(); iter.Valid(); iter.Next() {
+		v, err := iter.ValueAndErr()
+		if err != nil {
+			continue
+		}
+		i++ // TODO: should we index including failed keys?
```
iurii-ssv, line 105: On this TODO, I think we should stop iteration as soon as we encounter 1st `error` and return that `error` (because from what I understand iterator returning an error would mean some internal DB issues or data corruption ... it doesn't make sense to just skip it) ?

storage/pebble/txn_test.go
```diff
@@ -0,0 +1,331 @@
...
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/storage/basedb"
+)
+
+// Pebble DB txn requires its own set of tests because it has a different behaviour:
+// - it doesn't support concurrent transactions
+// - it doesn't support committing a transaction after it's been discarded
```
iurii-ssv, line 18: > it doesn't support committing a transaction after it's been discarded

I'm not sure the concept of Discarded transaction needs a mention, could you clarify why you want it documented ?

> it doesn't support concurrent transactions

Could you expand what this means - I thought any DB allows for concurrent usage and the way it handles it under the hood (be it serializing execution or handling some parts in parallel) is kind of irrelevant to the caller ?

storage/pebble/pebble.go
```diff
@@ -0,0 +1,283 @@
...
+	return iter.Error()
+}
+
+func (pdb *PebbleDB) Begin() basedb.Txn {
+	txn := pdb.db.NewIndexedBatch()
+	return newPebbleTxn(txn, pdb)
+}
+
+func (pdb *PebbleDB) BeginRead() basedb.ReadTxn {
+	txn := pdb.db.NewIndexedBatch() // TODO: Read txn
```
iurii-ssv, line 178: What's this TODO about ?

iurii-ssv (commented):
Some suggestions

storage/pebble/pebble.go
```diff
@@ -0,0 +1,283 @@
...
+	"context"
+	"errors"
+
+	"github.com/cockroachdb/pebble"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/storage/basedb"
+)
+
+// TODO: reconsider prefixes?
```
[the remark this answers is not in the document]
anatolie-ssv: this TODO (and all the following comments and TODOs) are from the original PR from @y0sher and I'm afraid I don't know their context; I left them intact so we could tackle them when Yosher gets around to review this PR; @y0sher please, what did you mean by 'reconsider prefixes' here? is is safe to be removed or does it require any action?

storage/pebble/pebble.go
anatolie-ssv: in the other implementation - BadgerDB - _close_ method does not return error hence here I decided to align the functionality to follow same semantics; also, in PebbleDB `Close()` returns the same "accumulated errors" that the `iter.Close()` returns at the end of the function, so the caller receives it either way; the only issue with logging it here is that it would likely be double logged: here, and on the caller side (most likely)

storage/pebble/pebble.go
anatolie-ssv: removed

storage/pebble/txn_test.go
anatolie-ssv: this comment was relevant only when pebble and badger were sharing same package; I forgot to remove the comment after splitting the packages;

storage/pebble/txn.go
moshe-blox: @iurii-ssv @anatolie-ssv definitely, we should never ignore DB errors especially given it's responsible for recording slashing data
we should return them and let the caller decide what to do (in most cases it should crash the node)

storage/pebble/txn.go
```diff
@@ -0,0 +1,127 @@
...
+	return nil
+}
+
+func (t *pebbleTxn) GetAll(prefix []byte, fn func(int, basedb.Obj) error) error {
+	iter, err := makePrefixIter(t.batch, prefix)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = iter.Close() }()
```
moshe-blox, line 69: i think we should return an error if this failed to close (it might do more than just closed in some cases)

storage/pebble/pebble.go
anatolie-ssv: it was before the read only transaction were implemented, removed now

storage/pebble/txn.go
anatolie-ssv: the errors returned from the `.Close()` method are the same as the errors returned from the `iter.Error()` at the end of the function, so we're not ignoring them.

moshe-blox:
@CodiumAI-Agent /review

moshe-blox:
@CodiumAI-Agent /describe

moshe-blox:
@CodiumAI-Agent /review

moshe-blox:
@CodiumAI-Agent /improve

storage/pebble/txn.go
anatolie-ssv: the only reason I defer the call to `Close()` is because - if a panic happens in the callback function - the defer statement will be executed regardless, but if we return `Close()` instead of `iter.Error()` that would lead to a leak.

iurii-ssv: We could keep using `defer` if we want and handle error from `Close` in there ... 

but I wouldn't do that - it's a bit bulky and IMO unnecessary since `panic` in callback function would mean we have a bug in the code that needs to be addressed, I imagine not calling `Close` in this case is of smaller importance

anatolie-ssv: I think the conclusion of this [article](https://dev.ribic.ba/defer-mutex-unlocks/) got stuck with me too much and that's why I went by the route of deferring `Close` statements here.

iurii-ssv: Yeah `defer` is a very nice thing to have/use - I don't mind it, 

although here it seems to me `panic` is unlikely to happen without us noticing it during testing

PR #2153, opened 2025-04-16
networkconfig: get beacon config from beacon node

nkryuchkov:

Based on https://github.com/ssvlabs/ssv/pull/2145

### DIFF
--- a/beacon/goclient/goclient.go
+++ b/beacon/goclient/goclient.go
@@ -2,6 +2,7 @@ package goclient
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"strings"
@@ -123,11 +124,15 @@ const (
 
 // GoClient implementing Beacon struct
 type GoClient struct {
-	log          *zap.Logger
-	ctx          context.Context
-	beaconConfig networkconfig.BeaconConfig
-	clients      []Client
-	multiClient  MultiClient
+	log *zap.Logger
+	ctx context.Context
+
+	beaconConfigMu   sync.Mutex
+	beaconConfig     *networkconfig.BeaconConfig
+	beaconConfigInit chan struct{}
+
+	clients     []Client
+	multiClient MultiClient
 	specssv.VersionCalls
 
 	syncDistanceTolerance phase0.Slot
@@ -173,7 +178,6 @@ type GoClient struct {
 	lastProcessedHeadEventSlotLock sync.Mutex
 	lastProcessedHeadEventSlot     phase0.Slot
 
-	genesisForkVersion phase0.Version
 	ForkLock           sync.RWMutex
 	ForkEpochElectra   phase0.Epoch
 	ForkEpochDeneb     phase0.Epoch
@@ -189,9 +193,7 @@ func New(
 	operatorDataStore operatorDataStore,
 	slotTickerProvider slotticker.Provider,
 ) (*GoClient, error) {
-	logger.Info("consensus client: connecting",
-		fields.Address(opt.BeaconNodeAddr),
-		fields.Network(opt.BeaconConfig.GetBeaconName()))
+	logger.Info("consensus client: connecting", fields.Address(opt.BeaconNodeAddr))
 
 	commonTimeout := opt.CommonTimeout
 	if commonTimeout == 0 {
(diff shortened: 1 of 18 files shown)


protocol/v2/blockchain/beacon/client.go
```diff
@@ -70,7 +68,6 @@ type BeaconNode interface {
 // Options for controller struct creation
 type Options struct {
 	Context                     context.Context
```
iurii-ssv, line 70: Can we get rid of this `Context                     context.Context` and pass `ctx` directly in any func that needs it (it's error-prone to have context stored as field, and it's harder to follow when it should be used as opposed to if we should use another `ctx`) ?

beacon/goclient/goclient.go
```diff
@@ -123,18 +118,20 @@ const (
...
-	ctx          context.Context
-	beaconConfig networkconfig.BeaconConfig
-	clients      []Client
-	multiClient  MultiClient
+	log *zap.Logger
+	ctx context.Context
+
+	beaconConfigMu   sync.Mutex
+	beaconConfig     *networkconfig.BeaconConfig
+	beaconConfigInit chan struct{}
```
iurii-ssv, line 126: Currently we only use `beaconConfigMu` to write the value of `beaconConfig` (so we can initialize it exactly once and do it correctly)

ideally we also want to use `beaconConfigMu` whenever we read the value of `beaconConfig` - if there is a go-routine that was started before `goclient.New` call it won't be aware of config being initialized (meaning it's a data race)
- in practice we probably initialize `goclient.New` before any other go-routine that reads it started
- but it's an "implicit" assumption that might unknowingly be violated in the future

I'm not sure it's ^ something we want to address, just calling it out so we are aware

utils/boot_node/node.go
```diff
@@ -107,8 +107,8 @@ func (n *bootNode) Start(ctx context.Context, logger *zap.Logger) error {
 	node := listener.LocalNode().Node()
 	logger.Info("Running",
 		zap.String("node", node.String()),
-		zap.String("network", n.network.Name),
-		fields.ProtocolID(n.network.DiscoveryProtocolID),
+		zap.Stringer("config", n.ssvConfig),
```
iurii-ssv, line 110: nit: probably better to use `zap.Stringer` for both cases (node, config) or `zap.String` for both - to keep it consistent

iurii-ssv (approved):
LGTM, left couple of suggestions

protocol/v2/blockchain/beacon/client.go
nkryuchkov: Yeah, I agree, I don't like storing it in the structure either. However, I think it will increase the scope of the PR and delay its merging. I'll consider doing it in another PR

nkryuchkov: I checked the amount of work on this, and we'll likely need to change signatures for the majority of the beacon client options, so I created an issue:

https://github.com/ssvlabs/ssv/issues/2161

By the way, I noticed we use the pattern of storing a context in a structure quite often. So we might need to do similar change in other packages

beacon/goclient/goclient.go
nkryuchkov: implemented

kchojn (approved):

beacon/goclient/goclient.go
```diff
@@ -257,6 +283,13 @@ func New(
 	return client, nil
 }
 
+// getBeaconConfig provides thread-safe access to the beacon configuration
+func (gc *GoClient) getBeaconConfig() *networkconfig.BeaconConfig {
```
oleg-ssvlabs, line 287: could be just `beaconConfig()` - I think that's a bit more idiomatic when it comes to naming getters in Go.

oleg-ssvlabs: Also, I think there are still places where `gc.beaconConfig` is accessed directly, bypassing the getter. 
Like [here](https://github.com/ssvlabs/ssv/blob/networkconfig-from-beacon-node/beacon/goclient/aggregator.go#L37) and [here](https://github.com/ssvlabs/ssv/blob/networkconfig-from-beacon-node/beacon/goclient/signing.go#L102)

nkryuchkov: I agree about the `beaconConfig` name, but there's already a struct field named like that. Regarding those places, well spotted! I missed them after merging

networkconfig/ssv.go
```diff
@@ -1,13 +1,34 @@
...
+	"encoding/json"
+	"fmt"
 	"math/big"
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
 //go:generate go tool -modfile=../tool.mod mockgen -package=networkconfig -destination=./ssv_mock.go -source=./ssv.go
 
+var SupportedSSVConfigs = map[string]SSVConfig{
```
oleg-ssvlabs, line 13: I think it would be better not to export this type so that the object remains immutable.

kchojn (approved):

protocol/v2/blockchain/beacon/client.go
nkryuchkov: @iurii-ssv https://github.com/ssvlabs/ssv/pull/2188

PR #2162, opened 2025-04-18
fix(eth/executionclient): flaky TestFetchLogsInBatches when startBlock > endBlockRetry

kchojn:
- **Early bad‑input check**: 
    - moved the `startBlock > endBlock` validation *outside* the goroutine
    - this prevents unnecessary goroutine creation for invalid inputs
    - guarantees immediate and consistent error response
    - eliminates race conditions between error reporting and channel closing
- **Channel naming**: 
    - renamed `logs ➔ logCh` and `errors ➔ errCh` to follow Go best practices (short, describes payload, Ch suffix). 
    - eliminates variable name collision with the standard library `errors` package
- **Comment tweak**:
    - updated function docs

### DIFF
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -145,19 +145,22 @@ func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock ui
 	return
 }
 
-// Calls FilterLogs multiple times and batches results to avoid fetching enormous amount of events
+// Calls FilterLogs multiple times and batches results to avoid fetching an enormous number of events.
 func (ec *ExecutionClient) fetchLogsInBatches(ctx context.Context, startBlock, endBlock uint64) (<-chan BlockLogs, <-chan error) {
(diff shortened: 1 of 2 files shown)


iurii-ssv (approved):
LGTM,

also I've noticed a TODO comment we could address in this PR while at it:
```
		// TODO: (Alan) should it really be here?
		if err := <-fetchError; err != nil {
			return 0, fmt.Errorf("error occurred while fetching historical logs: %w", err)
		}
```

I think we should read both `fetchLogs` and `fetchError` channels in `HandleBlockEventsStream` so we:
- can terminate early if `ctx` is canceled (we'll need to replace `for blockLogs := range logs` with a `select` statement)
- don't rely on `fetchLogs` to be closed for us to even be able to check if there are any errors

kchojn:
thanks @iurii-ssv, i have created an issue based on your comment: https://github.com/ssvlabs/ssv/issues/2176. The scope is a bit different imo, so it will be easier to track the changes

PR #2156, opened 2025-04-17
networkconfig: CLI for custom SSV config generation

nkryuchkov:
Part of PRs that will replace https://github.com/ssvlabs/ssv/pull/1308

Based on https://github.com/ssvlabs/ssv/pull/2153

PR #2164, opened 2025-04-19
networkconfig: make hardcoded values configurable

nkryuchkov:

Depends on https://github.com/ssvlabs/ssv/pull/2156

Required as a preparation for the local testnet

### DIFF
--- a/beacon/goclient/spec.go
+++ b/beacon/goclient/spec.go
@@ -14,8 +14,14 @@ import (
 )
 
 const (
-	DefaultSlotDuration  = 12 * time.Second
-	DefaultSlotsPerEpoch = phase0.Slot(32)
+	DefaultSlotDuration                         = 12 * time.Second
+	DefaultSlotsPerEpoch                        = phase0.Slot(32)
+	DefaultEpochsPerSyncCommitteePeriod         = phase0.Epoch(256)
+	DefaultSyncCommitteeSize                    = uint64(512)
+	DefaultSyncCommitteeSubnetCount             = uint64(4)
+	DefaultTargetAggregatorsPerSyncSubcommittee = uint64(16)
+	DefaultTargetAggregatorsPerCommittee        = uint64(16)
+	DefaultIntervalsPerSlot                     = uint64(3)
 )
 
 // BeaconConfig must be called if GoClient is initialized (gc.beaconConfig is set)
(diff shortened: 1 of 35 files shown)


beacon/goclient/spec.go
```diff
@@ -74,6 +80,66 @@ func (gc *GoClient) fetchBeaconConfig(client *eth2clienthttp.Service) (networkco
 		}
 	}
 
+	epochsPerSyncCommitteePeriod := DefaultEpochsPerSyncCommitteePeriod
+	if epochsPerSyncCommitteePeriodRaw, ok := specResponse.Data["EPOCHS_PER_SYNC_COMMITTEE_PERIOD"]; ok {
+		if epochsPerSyncCommitteePeriodDecoded, ok := epochsPerSyncCommitteePeriodRaw.(uint64); ok {
+			epochsPerSyncCommitteePeriod = phase0.Epoch(epochsPerSyncCommitteePeriodDecoded)
+		} else {
+			gc.log.Warn("epochs per sync committee not known by chain, using default value",
```
iurii-ssv, line 88: I'm not sure if "not known by chain" phrasing is best, maybe something like "isn't set by Beacon node" (or similar) would better describe what it actually entails ?

beacon/goclient/sync_committee_contribution.go
```diff
@@ -23,7 +23,8 @@ func (gc *GoClient) IsSyncCommitteeAggregator(proof []byte) (bool, error) {
 	hash := sha256.Sum256(proof)
 
 	// Keep the signature if it's an aggregator.
-	modulo := SyncCommitteeSize / SyncCommitteeSubnetCount / TargetAggregatorsPerSyncSubcommittee
+	cfg := gc.BeaconConfig()
+	modulo := cfg.SyncCommitteeSize / cfg.SyncCommitteeSubnetCount / cfg.TargetAggregatorsPerSyncSubcommittee
```
iurii-ssv, line 27: I guess the order of `/` operation evaluation is left-to-right, but it's probably better to add brackets `()` explicitly just in case for computations like this

beacon/goclient/types.go
```diff
@@ -7,12 +7,6 @@ import (
...
 var (
-	SyncCommitteeSize                    uint64 = 512
-	SyncCommitteeSubnetCount             uint64 = 4
-	TargetAggregatorsPerSyncSubcommittee uint64 = 16
-	EpochsPerSyncCommitteePeriod         uint64 = 256
-	TargetAggregatorsPerCommittee        uint64 = 16
 	// FarFutureEpoch is the null representation of an epoch.
-	FarFutureEpoch   phase0.Epoch = math.MaxUint64
-	IntervalsPerSlot uint64       = 3
+	FarFutureEpoch phase0.Epoch = math.MaxUint64
```
iurii-ssv, line 11: If it is indeed a "is the null representation" why not just call it just that `UndefinedEpoch` (or `NullEpoch`) ?

Or maybe we want to have both `UndefinedEpoch` and `FarFutureEpoch` defined and use whichever is more appropriate in a particular situation.

networkconfig/beacon.go
```diff
@@ -141,6 +151,20 @@ func (b BeaconConfig) EstimatedTimeAtSlot(slot phase0.Slot) time.Time {
...
+func (b BeaconConfig) IntervalDuration() time.Duration {
+	if b.IntervalsPerSlot > math.MaxInt64 {
+		panic("intervals per slot out of range")
+	}
+	return b.SlotDuration / time.Duration(b.IntervalsPerSlot) // #nosec G115: intervals per slot cannot exceed math.MaxInt64
+}
+
+func (b BeaconConfig) EpochDuration() time.Duration {
+	if b.SlotsPerEpoch > math.MaxInt64 {
+		panic("slot out of range")
```
iurii-ssv, line 163: `slot out of range` -> `slots per epoch out of range`

networkconfig/test-network.go
```diff
@@ -25,5 +25,6 @@ var TestNetwork = NetworkConfig{
 		Bootnodes: []string{
 			"enr:-Li4QFIQzamdvTxGJhvcXG_DFmCeyggSffDnllY5DiU47pd_K_1MRnSaJimWtfKJ-MD46jUX9TwgW5Jqe0t4pH41RYWGAYuFnlyth2F0dG5ldHOIAAAAAAAAAACEZXRoMpD1pf1CAAAAAP__________gmlkgnY0gmlwhCLdu_SJc2VjcDI1NmsxoQN4v-N9zFYwEqzGPBBX37q24QPFvAVUtokIo1fblIsmTIN0Y3CCE4uDdWRwgg-j",
 		},
+		TotalEthereumValidators: 10, // just some random number
```
iurii-ssv, line 28: Maybe this value (for testnet) should be high enough so we never accidentally reach the message-limits derived from it while testing something with local testnet (unless ofc we are testing those message-limits specifically)

otherwise we might encounter some hard-to-debug failures just because of that ?

iurii-ssv (approved):
LGTM, left minor suggestions

beacon/goclient/types.go
nkryuchkov: Well, honestly, I don't like `FarFutureEpoch` being `math.MaxUint64` either, but IMO it should be out of scope of this PR

networkconfig/test-network.go
nkryuchkov: good idea

beacon/goclient/sync_committee_contribution.go
MatheusFranco99: Hmm indeed before we had a division first and now the multiplication is happening first.
There may be edge cases in which they are not equivalent due to rounding.

The equivalent would be
```go
modulo := (SyncCommitteeSize / SyncCommitteeSubnetCount) / TargetAggregatorsPerSyncSubcommittee
```
right?

Due to the [Ethereum spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#aggregation-selection):

```py
def is_sync_committee_aggregator(signature: BLSSignature) -> bool:
    modulo = max(1, SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT // TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE)
    return bytes_to_uint64(hash(signature)[0:8]) % modulo == 0
```

I would either:
- use it as it currently is and add a reference to the Ethereum spec (explaining why this bad-looking / /)
- or use the equivalent form

beacon/goclient/types.go
iurii-ssv: Yeah we can get back to it later - https://github.com/ssvlabs/ssv/issues/2173

beacon/goclient/sync_committee_contribution.go
iurii-ssv: > I rewrote it to the version below because I think it's more readable. What do you think?

I like the version you offered more as well, but since this is indeed a behavior-change I think it's best to use this (+ leave a clarifying comment that thhese integer divisions MUST be done this way because the behavior MUST match Ethereum spec - like **MatheusFranco99** mentioned):
```go
modulo := (SyncCommitteeSize / SyncCommitteeSubnetCount) / TargetAggregatorsPerSyncSubcommittee
```

nkryuchkov: Reverted. I think it must be left as it was, similarly to the spec:
```
modulo := cfg.SyncCommitteeSize / cfg.SyncCommitteeSubnetCount / cfg.TargetAggregatorsPerSyncSubcommittee
```

beacon/goclient/types.go
nkryuchkov: I've been thinking more about this, and now I think it indeed should be `FarFutureEpoch = math.MaxUint64` because it's the value eth nodes are using when the epoch is not defined (see https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/fork.md, `FULU_FORK_EPOCH`). So IMO we shouldn't change anything

iurii-ssv: Thanks for looking into it, lets keep this discussion to https://github.com/ssvlabs/ssv/issues/2173

iurii-ssv: Thanks for looking into it, lets keep this discussion to https://github.com/ssvlabs/ssv/issues/2173

oleg-ssvlabs (commented):
great job!

beacon/goclient/spec.go
```diff
@@ -63,23 +69,90 @@ func (gc *GoClient) fetchBeaconConfig(client *eth2clienthttp.Service) (networkco
...
 			slotsPerEpoch = phase0.Slot(slotsPerEpochDecoded)
 		} else {
-			gc.log.Warn("slots per epoch not known by chain, using default value",
+			gc.log.Warn("slots per epoch wasn't found in beacon node response, using default value",
 				zap.Any("value", slotsPerEpoch))
 		}
 	}
 
+	epochsPerSyncCommitteePeriod := DefaultEpochsPerSyncCommitteePeriod
+	if epochsPerSyncCommitteePeriodRaw, ok := specResponse["EPOCHS_PER_SYNC_COMMITTEE_PERIOD"]; ok {
```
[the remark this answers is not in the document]
nkryuchkov: It tries to convert all numbers to uint64: https://github.com/attestantio/go-eth2-client/blob/v0.21.7/http/spec.go#L138

beacon/goclient/spec.go
```diff
@@ -63,23 +69,90 @@ func (gc *GoClient) fetchBeaconConfig(client *eth2clienthttp.Service) (networkco
...
+		if targetAggregatorsPerSyncSubcommitteeDecoded, ok := targetAggregatorsPerSyncSubcommitteeRaw.(uint64); ok {
+			targetAggregatorsPerSyncSubcommittee = targetAggregatorsPerSyncSubcommitteeDecoded
+		} else {
+			gc.log.Warn("target aggregators per sync subcommittee wasn't found in beacon node response, using default value",
+				zap.Any("value", targetAggregatorsPerSyncSubcommittee))
+		}
+	}
+
+	intervalsPerSlot := DefaultIntervalsPerSlot
+	if intervalsPerSlotRaw, ok := specResponse["INTERVALS_PER_SLOT"]; ok {
```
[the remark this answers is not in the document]
nkryuchkov: Yes, some nodes lack it. In this case, `DefaultIntervalsPerSlot` will be used

beacon/goclient/sync_committee_contribution.go
```diff
@@ -33,7 +36,7 @@ func (gc *GoClient) IsSyncCommitteeAggregator(proof []byte) (bool, error) {
 
 // SyncCommitteeSubnetID returns sync committee subnet ID from subcommittee index
 func (gc *GoClient) SyncCommitteeSubnetID(index phase0.CommitteeIndex) (uint64, error) {
```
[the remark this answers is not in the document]
nkryuchkov: I think https://github.com/ssvlabs/ssv/pull/2188 is a better place for it because it already changes signatures. I implemented it there

Issue #2066, opened 2025-03-08
`computeVoluntaryExitDomain` uses `CAPELLA_FORK_VERSION`


nkryuchkov:
**Describe the bug**
Do we need to use the latest fork version instead?

PR #1338, opened 2024-03-11
fix: voluntary exit signature in post-deneb networks

y0sher:
### Issue:
Following the Deneb fork on the Ethereum Holesky network, the domain in signatures of voluntary exit messages was fixed to the Capella fork and epoch. This change was not accounted for in our codebase, leading to a critical bug where voluntary exits failed on networks post-Deneb.

### Solution:
This PR introduces a fix that aligns our handling of voluntary exit signatures with the updated domain requirements set forth by the Deneb fork. It ensures voluntary exits are correctly processed in networks operating under the post-Deneb specifications.

PR #2166, opened 2025-04-21
beacon: clarify & optimize voluntary exit domain handling

iurii-ssv:
Closes https://github.com/ssvlabs/ssv/issues/2066, see relevant PR - https://github.com/ssvlabs/ssv/pull/1338

Answering the question from https://github.com/ssvlabs/ssv/issues/2066 - it seems https://eips.ethereum.org/EIPS/eip-7044 is saying we always should use `CAPELLA_FORK_VERSION` for voluntary exit domain (at least until there is a next EIP that invalidates this), so in this PR:
- I'm aiming to clarify that
- plus, as per TODO you'll see in the diff, it seems `computeVoluntaryExitDomain` calculates the same result making Beacon request every time - so we can calculate it once and cache the result (for simplicity I'm not using mutex and instead I'm using atomic pointer - hence we **might** calculate it couple times ... but that doesn't matter in this case)


beacon/goclient/signing.go
```diff
@@ -96,7 +107,8 @@ func (gc *GoClient) computeVoluntaryExitDomain(ctx context.Context) (phase0.Doma
 
 func (gc *GoClient) DomainData(epoch phase0.Epoch, domain phase0.DomainType) (phase0.Domain, error) {
 	switch domain {
-	case spectypes.DomainApplicationBuilder: // no domain for DomainApplicationBuilder. need to create.  https://github.com/bloxapp/ethereum2-validator/blob/v2-main/signing/keyvault/signer.go#L62
+	case spectypes.DomainApplicationBuilder:
+		// DomainApplicationBuilder is constructed based on SSV network configuration
```
iurii-ssv, line 111: This link doesn't seem relevant anymore (I couldn't quite make sense of it) ? Can I remove it @y0sher or do you have anything to add here ?

beacon/goclient/signing.go
```diff
@@ -41,9 +55,6 @@ func (gc *GoClient) computeVoluntaryExitDomain(ctx context.Context) (phase0.Doma
 		return phase0.Domain{}, fmt.Errorf("spec response data is nil")
 	}
 
-	// TODO: consider storing fork version and genesis validators root in goClient
-	//		instead of fetching it every time
-
```
nkryuchkov, line 46: I think it would be great to have a short descriptive comment right above getting `CAPELLA_FORK_VERSION` that would explain why we need the Capella version. IMO, the one in `DomainData` has some kind of distance from the confusing place and doesn't quickly explain why Capella. Maybe something like `EIP-7044 requires CAPELLA_FORK_VERSION for DomainVoluntaryExit: https://eips.ethereum.org/EIPS/eip-7044`?

iurii-ssv: Added

nkryuchkov (approved):

oleg-ssvlabs (approved):

beacon/goclient/signing.go
```diff
@@ -17,6 +17,20 @@ import (
...
+	if value != nil {
+		return *value, nil
+	}
+
+	v, err := gc.computeVoluntaryExitDomain(ctx)
+	if err != nil {
+		return phase0.Domain{}, fmt.Errorf("compute voluntary exit domain: %w", err)
+	}
+	gc.voluntaryExitDomainCached.Store(&v)
+	return v, nil
```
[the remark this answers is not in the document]
iurii-ssv: True,

> for simplicity I'm not using mutex and instead I'm using atomic pointer - hence we might calculate it couple times ... but that doesn't matter in this case

beacon/goclient/signing.go
nkryuchkov: @iurii-ssv I think this comment is about this line below:
```go
GenesisValidatorsRoot: phase0.Root{},
```

Perhaps, we need to keep it commented

PR #2174, opened 2025-04-24
executionclient: fix log streaming logic

nkryuchkov:
Fix bug with increasing `fromBlock` when we shouldn't increase it. `streamLogsToChan` currently returns the last block it fetched even if it errored. The new logic returns the next block to process

### DIFF
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -439,7 +439,7 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logCh chan<- Bl
 			logStream, fetchErrors := ec.fetchLogsInBatches(ctx, fromBlock, toBlock)
 			for block := range logStream {
 				logCh <- block
-				fromBlock = max(fromBlock, block.BlockNumber+1)
+				fromBlock = block.BlockNumber + 1
 			}
 			if err := <-fetchErrors; err != nil {
 				// If we get an error while fetching, we return the last block we fetched.
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -229,9 +229,7 @@ func (ec *ExecutionClient) fetchLogsInBatches(ctx context.Context, startBlock, e
 }
 
 // StreamLogs subscribes to events emitted by the contract.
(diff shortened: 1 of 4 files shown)


eth/executionclient/execution_client.go
```diff
@@ -41,7 +41,7 @@ type Provider interface {
 type SingleClientProvider interface {
 	Provider
 	SyncProgress(ctx context.Context) (*ethereum.SyncProgress, error)
-	streamLogsToChan(ctx context.Context, logs chan<- BlockLogs, fromBlock uint64) (lastBlock uint64, err error)
+	streamLogsToChan(ctx context.Context, logs chan<- BlockLogs, fromBlock uint64) (nextBlockToProcess uint64, err error)
```
kchojn, line 44: @nkryuchkov, maybe it's a good time to update the channel name to be more in line with the `go`? This is a broad topic with plenty of room for flexibility, so please consider this just as a suggestion!
https://groups.google.com/g/golang-nuts/c/NnZ-CuQAiT4

oleg-ssvlabs (approved):

eth/executionclient/execution_client.go
```diff
@@ -433,15 +435,17 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
...
 				continue
 			}
+
 			logStream, fetchErrors := ec.fetchLogsInBatches(ctx, fromBlock, toBlock)
 			for block := range logStream {
-				logs <- block
-				lastBlock = block.BlockNumber
+				logCh <- block
+				fromBlock = max(fromBlock, block.BlockNumber+1)
 			}
```
iurii-ssv, line 443: Just calling it out since this looks a bit strange to me - namely why do we need `max` here ? Shouldn't we fix `fetchLogsInBatches` itself so it doesn't return blocks with height lower than `fromBlock` ?

nkryuchkov: It done to make sure blocks never decrease as an extra assertion but probably we can remove it and add an this assertion to `fetchLogsInBatches`

iurii-ssv (approved):

PR #2178, opened 2025-04-26
beacon,networkconfig: move forks from beacon to networkconfig

nkryuchkov:

Depends on https://github.com/ssvlabs/ssv/pull/2164


beacon/goclient/goclient.go
```diff
@@ -372,38 +359,20 @@ func (gc *GoClient) singleClientHooks() *eth2clienthttp.Hooks {
...
-
-			spec, err := specForClient(ctx, logger, s)
-			if err != nil {
-				logger.Error(clResponseErrMsg,
-					zap.String("api", "Spec"),
 					zap.Error(err),
+					zap.Any("current_forks", currentConfig.Forks),
+					zap.Any("got_forks", beaconConfig.Forks),
+					//zap.Stringer("client_config", beaconConfig),
+					//zap.Stringer("expected_config", currentConfig),
```
iurii-ssv, line 366: Since we terminate after logging that as `Fatal` I think we might want to print `client_config` and `expected_config` (meaning we don't need to worry about how large the log line is since it's printed 1 time) ?

iurii-ssv (approved):

networkconfig/beacon.go
```diff
@@ -46,9 +50,10 @@ type BeaconConfig struct {
 	TargetAggregatorsPerSyncSubcommittee uint64
 	TargetAggregatorsPerCommittee        uint64
 	IntervalsPerSlot                     uint64
-	ForkVersion                          phase0.Version
+	GenesisForkVersion                   phase0.Version
 	GenesisTime                          time.Time
 	GenesisValidatorsRoot                phase0.Root
+	Forks                                map[spec.DataVersion]phase0.Fork // TODO: 1) support updating electra; 2) lock
 }
```
iurii-ssv, line 57: What about this TODO, will it be resolved later on (in a separate PR) ?

nkryuchkov: I left a TODO and then realized that the PR will be merged only after the fork, so I didn't create a PR for that. However, I think, it might be useful as a preparation for the Fulu fork, but it will add significant complexity to the network config code. I still haven't decided about it but I tend to think we'll need another PR because it's been already implemented in the current code

networkconfig/holesky.go
oleg-ssvlabs, line 1: I wonder.
Is it possible not to even have these static configurations for various networks?
Is it possible to fetch all data, including the network name? So, no statically hardcoded configuration at all

nkryuchkov: Yes, I deleted all hardcoded beacon configs (except test configs) in https://github.com/ssvlabs/ssv/pull/2183. Perhaps, I should have done it before the current PR, I'll think what I can do here

networkconfig/mainnet.go
```diff
@@ -22,9 +23,47 @@ var Mainnet = NetworkConfig{
...
+				CurrentVersion:  phase0.Version{0x01, 0x00, 0x00, 0x00}, // ALTAIR_FORK_VERSION
+			},
+			// Bellatrix (Merge) @ epoch 144896
+			spec.DataVersionBellatrix: {
+				Epoch:           phase0.Epoch(144896),
+				PreviousVersion: phase0.Version{0x01, 0x00, 0x00, 0x00},
+				CurrentVersion:  phase0.Version{0x02, 0x00, 0x00, 0x00}, // BELLATRIX_FORK_VERSION
+			},
+			// Capella @ epoch 194048
+			spec.DataVersionCapella: {
```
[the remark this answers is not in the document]
nkryuchkov: It's gotten dynamically from BNs, so these lines are actually unused. I have added them for consistency. Perhaps I should have done it before the current PR, let me think what I can do to improve clarity

networkconfig/beacon.go
```diff
@@ -181,6 +186,81 @@ func (b BeaconConfig) GetSyncCommitteeSize() uint64 {
...
+
+			version := versions[i-1]
+			fork := b.Forks[version]
+			return version, &fork
+		}
+	}
+
+	version := versions[len(versions)-1]
+	fork := b.Forks[version]
+	return version, &fork
```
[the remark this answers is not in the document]
nkryuchkov: It's unexpected to happen and means a bug, so IMO panic is better here

beacon/goclient/aggregator.go
```diff
@@ -34,7 +34,9 @@ func (gc *GoClient) SubmitAggregateSelectionProof(slot phase0.Slot, committeeInd
 	if err != nil {
 		return nil, DataVersionNil, fmt.Errorf("failed to get attestation data: %w", err)
 	}
-	if gc.DataVersion(gc.getBeaconConfig().EstimatedEpochAtSlot(attData.Slot)) < spec.DataVersionElectra {
+
+	dataVersion, _ := gc.beaconConfig.ForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(attData.Slot))
```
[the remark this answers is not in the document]
nkryuchkov: 2nd argument is not an error

networkconfig/beacon.go
nkryuchkov: removed the comment

networkconfig/mainnet.go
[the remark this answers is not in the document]
nkryuchkov: Some configs, including mainnet, are still used in tests as of this PR, so the configs still need to be filled with correct values, although they are taken from the beacon node on SSV node start. So I thought a cleaner solution would be to fill all values and do a cleanup PR in the end (https://github.com/ssvlabs/ssv/pull/2183)

oleg-ssvlabs (approved):

PR #2183, opened 2025-04-29
networkconfig: delete hardcoded beacon configs

nkryuchkov:
Part of PRs that will replace https://github.com/ssvlabs/ssv/pull/1308

Depends on https://github.com/ssvlabs/ssv/pull/2178

PR #2184, opened 2025-04-29
feat(eth): ethereum sync finality support

kchojn:
TODO:
- set up the correct fork name (comments, vars, names, etc)


eth/executionclient/options.go
```diff
@@ -14,114 +14,102 @@ type OptionMulti func(client *MultiClient)
...
-		s.reconnectionMaxInterval = interval
-	}
-}
-
-// WithReconnectionMaxIntervalMulti sets max reconnection interval.
-func WithReconnectionMaxIntervalMulti(interval time.Duration) OptionMulti {
-	return func(s *MultiClient) {
-		s.reconnectionMaxInterval = interval
+	return func(c *MultiClient) {
+		c.connectionTimeout = timeout
```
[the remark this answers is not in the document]
kchojn: lol man 💀

eth/executionclient/constants.go
```diff
@@ -0,0 +1,10 @@
+package executionclient
+
+const (
+	SlotsPerEpoch    = 32
```
y0sher, line 4: this is read from the beacon config and might be different on different networks.. (devnet or some testnet). usually its 32 but it's best to just read all this info from the beacon node. (we have PRs for that and it will probably be merged before this PR.

kchojn: Great, what about `ForkEpoch`? Do we have a chain config or something that can be used here? To be honest, I would love to have something centralized, like `ChainConfig` in geth so that we can maintain everything (forks, sips, checks, e.g., IsFinalityFork(), etc.) in one place

y0sher: we have networkconfig package and a struct that initialized in the main endpoint of the program, it should have all chain related config.

nkryuchkov: @kchojn yes, the list of forks with their epoch will be there too. You can check what it will look like in https://github.com/ssvlabs/ssv/pull/2183

nkryuchkov: Oh, you actually meant SSV fork, not ETH fork, right? We'll need to create everything SSV-related inside the SSV config

kchojn: @nkryuchkov  Yes, SSV fork. I think i will create something like ssv_fork +  will move logic in there with forks (alan and upcoming)

networkconfig/config.go
```diff
@@ -31,6 +29,7 @@ type NetworkConfig struct {
 	Name string
 	BeaconConfig
 	SSVConfig
+	ForksConfig
```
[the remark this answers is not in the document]
kchojn: Yeah, answered in a different comment. Gonna put everything in `SSVConfig`

eth/executionclient/constants.go
nkryuchkov: Sounds good

networkconfig/config.go
```diff
@@ -43,7 +42,7 @@ func (n NetworkConfig) String() string {
 }
 
 func (n NetworkConfig) NetworkName() string {
-	return fmt.Sprintf("%s:%s", n.Name, forkName)
+	return fmt.Sprintf("%s:%s", n.Name, "alan")
```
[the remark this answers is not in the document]
kchojn:
```go
func (n NetworkConfig) NetworkName() string {
	activeFork := n.Forks.ActiveFork(0) // Alan

	return fmt.Sprintf("%s:%s", n.Name, strings.ToLower(activeFork.Name))
}
```

i added something like this. Should it always be `alan`? tbh I don't currently know how to approach this, and it has previously got me thinking

networkconfig/hoodi.go
```diff
@@ -23,4 +23,7 @@ var Hoodi = NetworkConfig{
 			"enr:-Ja4QIKlyNFuFtTOnVoavqwmpgSJXfhSmhpdSDOUhf5-FBr7bBxQRvG6VrpUvlkr8MtpNNuMAkM33AseduSaOhd9IeWGAZWjRbnvgmlkgnY0gmlwhCNVVTCJc2VjcDI1NmsxoQNTTyiJPoZh502xOZpHSHAfR-94NaXLvi5J4CNHMh2tjoNzc3YBg3RjcIITioN1ZHCCD6I",
 		},
 	},
+	ForksConfig: ForksConfig{
+		FinalityConsensusEpoch: 120, // TODO: use a different name when we have a better one, value as well
```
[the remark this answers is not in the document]
kchojn: changed to -> `MaxEpoch = phase0.Epoch(math.MaxUint64)`

iurii-ssv (approved):
LGTM

networkconfig/config.go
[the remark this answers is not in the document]
nkryuchkov: It's currently `alan` but on the next SSV fork it will be different

eth/eventsyncer/event_syncer.go
```diff
@@ -81,27 +82,62 @@ func (es *EventSyncer) Healthy(ctx context.Context) error {
...
+		}
+
+		// #nosec G115
+		blockTime := time.Unix(int64(finalizedHeader.Time), 0)
+		staleness := time.Since(blockTime)
+
+		if staleness > es.finalizedStalenessThreshold {
+			return fmt.Errorf("finalized block %d is too old (age: %s)",
+				finalizedHeader.Number.Uint64(), staleness.Round(time.Second))
+		}
```
moshe-blox: it doesnt make much sense to check the finalized block directly from the EL because it doesnt mean we've processed it

also an absolute threshold for block age post-finality syncing is wrong because Ethereum itself can just not finalize for hours/days, which is an extreme scenario but i'm not sure we'd like SSV nodes to just crash in this scenario

i'd rewrite this func to instead measure our latency against the latest Ethereum finalized block: 
```go
		header, _ := es.executionClient.HeaderByNumber(ctx, block)

		blockTime := time.Unix(int64(header.Time), 0)
        latestBlockTime := time.Now()
        if finalizedFork {
            // Get finalized block and replace latestBlockTime with its time.
        }
		staleness := latestBlockTime.Sub(blockTime)

		if staleness > es.stalenessThreshold {
			return fmt.Errorf("block %d is too old (age: %s)",
				block.Uint64(), staleness.Round(time.Second))
		}
```

@y0sher wdyt?

eth/executionclient/config.go
```diff
@@ -12,3 +16,34 @@ type Options struct {
 	ConnectionTimeout     time.Duration `yaml:"ETH1ConnectionTimeout" env:"ETH_1_CONNECTION_TIMEOUT" env-default:"10s" env-description:"Timeout for execution client connections"`
 	SyncDistanceTolerance uint64        `yaml:"ETH1SyncDistanceTolerance" env:"ETH_1_SYNC_DISTANCE_TOLERANCE" env-default:"5" env-description:"Maximum number of blocks behind head considered in-sync"`
 }
+
+type Config struct {
+	SlotsPerEpoch          uint64       // Slots per epoch
```
moshe-blox: dont we have this in our NetworkConfig?

eth/executionclient/config.go
```diff
@@ -12,3 +16,34 @@ type Options struct {
 	ConnectionTimeout     time.Duration `yaml:"ETH1ConnectionTimeout" env:"ETH_1_CONNECTION_TIMEOUT" env-default:"10s" env-description:"Timeout for execution client connections"`
 	SyncDistanceTolerance uint64        `yaml:"ETH1SyncDistanceTolerance" env:"ETH_1_SYNC_DISTANCE_TOLERANCE" env-default:"5" env-description:"Maximum number of blocks behind head considered in-sync"`
 }
+
+type Config struct {
```
moshe-blox: why do we need both Options and Config?

why not just use Options which we already had?

eth/executionclient/execution_client.go
```diff
@@ -129,22 +132,55 @@ func (ec *ExecutionClient) Close() error {
...
-		ec.logger.Error(elResponseErrMsg,
-			zap.String("method", "eth_blockNumber"),
-			zap.Error(err))
-		return nil, nil, fmt.Errorf("failed to get current block: %w", err)
-	}
-	if currentBlock < ec.followDistance {
-		return nil, nil, ErrNothingToSync
+	var toBlock uint64
+
+	if ec.isPostForkState.Load() {
```
moshe-blox: why do we need a variable to track this? why not just call the func that checks this based on the slot/epoch, so we can avoid managing any state?

eth/executionclient/execution_client.go
```diff
@@ -408,52 +485,150 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logs chan<- Blo
...
+	// Only make this RPC call if we don't know our fork state yet
+	currentBlock, err := ec.client.BlockNumber(ctx)
+	if err != nil {
+		ec.logger.Error(elResponseErrMsg,
+			zap.String("method", "eth_blockNumber"),
+			zap.Error(err))
+		return false
+	}
+
+	currentEpoch := phase0.Epoch(currentBlock / ec.config.SlotsPerEpoch)
```
moshe-blox: this isnt how to derive epoch from EL blocks, have to derive the CL slot first

eth/executionclient/config.go
y0sher: yeah I think we should keep using network config instead of this.

eth/executionclient/config.go
```diff
@@ -12,3 +16,34 @@ type Options struct {
...
+	SlotsPerEpoch          uint64       // Slots per epoch
+	FinalityConsensusEpoch phase0.Epoch // Epoch at which finality fork activates
```
y0sher: the fork epoch will be specified in `NetworkConfig`, it will be part of a bigger fork and not just finality fork

eth/executionclient/config.go
kchojn: as far as i rememebr, the networkconfig was pretty big, that's why I separated this stuff.
I will pass NetworkConfig

eth/executionclient/config.go
kchojn: it wasnt part of the env/yaml configuration. However, the config will be removed

eth/executionclient/config.go
kchojn: Ik, but I don't know much detail about the upcoming fork, so I added custom naming/things to process further.
Anyway,I mentioned it in different comment - gonna remove the config

eth/eventsyncer/event_syncer.go
y0sher: @moshe-blox I agree we should have a fallback solution to when we don't have finality, the latency function is basically similar to what we have today no? compare latest known block time to the current system time? 
`	if header.Time < uint64(time.Now().Add(-es.stalenessThreshold).Unix()) { `

eth/executionclient/execution_client.go
kchojn: it should be fixed there: https://github.com/ssvlabs/ssv/blob/503cc9c7fab2aeb557bd3d9ddcdd8c6454967b93/eth/executionclient/execution_client.go#L666

eth/eventsyncer/event_syncer.go
kchojn: this is my proposition with better documentation:
```go
// Healthy determines if the EventSyncer is functioning correctly. It returns nil if healthy.
// It checks if the syncer has processed new blocks, hasn't been "stuck" without progress for too long
// (this check is skipped if in finalized mode and already caught up), and if its last processed block
// is sufficiently close to the relevant chain head (current time or EL finalized block time).
func (es *EventSyncer) Healthy(ctx context.Context) error {
	lastProcessedBlock, found, err := es.nodeStorage.GetLastProcessedBlock(nil)
	if err != nil {
		return fmt.Errorf("failed to read last processed block: %w", err)
	}
	if !found || lastProcessedBlock == nil || lastProcessedBlock.Uint64() == 0 {
		return fmt.Errorf("last processed block is not set")
	}

	lastBlockNum := lastProcessedBlock.Uint64()
	if es.lastProcessedBlock != lastBlockNum {
		es.lastProcessedBlock = lastBlockNum
		es.lastProcessedBlockChange = time.Now()
		return nil
	}

	// Check if we're making progress (only relevant for follow-distance approach)
	if !es.executionClient.IsFinalizedFork(ctx) {
		if time.Since(es.lastProcessedBlockChange) > es.stalenessThreshold {
			return fmt.Errorf("syncing is stuck at block %d", lastBlockNum)
		}
	}

	// Check if our current position is too far behind
	return es.blockBelowThreshold(ctx, lastProcessedBlock)
}

// blockBelowThreshold checks if the given block is acceptably recent. Returns nil if fresh.
// Pre-finality, it compares the block's time against `time.Now()` using `es.stalenessThreshold`.
// Post-finality, it compares against the EL's finalized block time. If the given block is newer
// than the EL's finalized block, it's considered fresh (waiting for EL finality).
func (es *EventSyncer) blockBelowThreshold(ctx context.Context, block *big.Int) error {
	header, err := es.executionClient.HeaderByNumber(ctx, block)
	if err != nil {
		// Ensure block number is used in error message if that's the intent for %d
		return fmt.Errorf("failed to get header for block %d: %w", block.Uint64(), err)
	}

	var referenceTime time.Time
	if es.executionClient.IsFinalizedFork(ctx) {
		// Post-fork: Compare against finalized block time
		finalizedHeader, err := es.executionClient.HeaderByNumber(ctx, big.NewInt(rpc.FinalizedBlockNumber.Int64()))
		if err != nil {
			return fmt.Errorf("failed to get finalized block header: %w", err)
		}

		// If we're ahead of finalized, we're not behind - just waiting
		if finalizedHeader.Number.Uint64() < block.Uint64() {
			return nil
		}

		// #nosec G115
		referenceTime = time.Unix(int64(finalizedHeader.Time), 0)
	} else {
		// Pre-fork: Compare against current time
		referenceTime = time.Now()
	}

	// Check if block is older than threshold from reference time
	// #nosec G115
	blockTime := time.Unix(int64(header.Time), 0)
	if blockTime.Before(referenceTime.Add(-es.stalenessThreshold)) {
		return fmt.Errorf("block %d is too old (age: %s behind reference)",
			block.Uint64(), referenceTime.Sub(blockTime).Round(time.Second))
	}

	return nil
}
```
lmk what do you think

kchojn:
> @kchojn now that #2307 was merged, I think we need to address [#2307 (comment)](https://github.com/ssvlabs/ssv/pull/2307#discussion_r2166425601) here as well

thanks @iurii-ssv

nkryuchkov:
superseded by https://github.com/ssvlabs/ssv/pull/2695

Issue #2161, opened 2025-04-17
Delete Context from beacon client Options


nkryuchkov:
https://github.com/ssvlabs/ssv/pull/2153#discussion_r2048586997

PR #2188, opened 2025-04-29
beacon: pass context to methods

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/2161

Currently, the beacon methods aren't cancellable because they use the common beacon client context, which matches the ssv-node lifetime


iurii-ssv (approved):
LGTM

oleg-ssvlabs (commented):
great job!
Left few minor comments related to the usage of `t.Context()` in tests

oleg-ssvlabs (approved):

protocol/v2/blockchain/beacon/client.go
```diff
@@ -49,7 +120,17 @@ type signer interface {
 
 // BeaconNode interface for all beacon duty calls
 type BeaconNode interface {
-	specssv.BeaconNode // spec beacon interface
```
oleg-ssvlabs, line 52: Is it OK not to use spec definitions?

nkryuchkov: IMO, yes. We decided to diverge from spec where we need it to keep spec looking like spec without any overhead

protocol/v2/ssv/runner/runner.go
```diff
@@ -44,9 +44,9 @@ type Runner interface {
 	// ProcessPostConsensus processes all post-consensus msgs, returns error if can't process
 	ProcessPostConsensus(ctx context.Context, logger *zap.Logger, signedMsg *spectypes.PartialSignatureMessages) error
 	// expectedPreConsensusRootsAndDomain an INTERNAL function, returns the expected pre-consensus roots to sign
-	expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error)
+	expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType)
```
iurii-ssv, line 47: Removing `error` from this **func** means we'll lose "error cause" (since we aren't logging it in `expectedPreConsensusRootsAndDomain` itself) ... although I do agree runner(s) code is more complex than it could/should be I'm not sure this particular change is positive for us, any specific reason to do it ?

y0sher (approved):
gj

protocol/v2/ssv/runner/runner.go
nkryuchkov: but it never actually returns an error

y0sher: yup, I don't see how this is part of this pr .. lets revert this

y0sher: if it never actually returns an error I think we can go with this change.

nkryuchkov: I checked it again, and it does for other runners. I did some automatic refactoring, which wasn't correct. @iurii-ssv thanks for spotting this

nkryuchkov: fixed

iurii-ssv (approved):
Nice!

Issue #1789, opened 2024-10-11
message/validation: unexpected 'too many duties per epoch'


nkryuchkov:
we have many such errors on stage nodes
```
{"level":"\u001b[35mDEBUG\u001b[0m","time":"2024-10-11T18:47:48.016262Z","msg":"ignoring invalid message","duty_executor_id":"0000000000000000000000000000000086a7f656e5353ab6e594842f5203ebc0de276e1ed91f263af058f5ea9978a539","role":"COMMITTEE_RUNNER","ssv_message_type":"consensus","slot":2730838,"duty_id":"COMMITTEE-283_285_287_289-e85338-s2730838","round":5,"qbft_message_type":"round_change","peer_id":"16Uiu2HAmVLuJYsxWGdMRFLPL678aXLqTkXRa1irhnWzAAaQevHvi","error":"too many duties per epoch, got 33 (role COMMITTEE_RUNNER), want less than 32"}
{"level":"\u001b[35mDEBUG\u001b[0m","time":"2024-10-11T18:47:48.014139Z","msg":"ignoring invalid message","duty_executor_id":"00000000000000000000000000000000c0bdc3d221e8bdc11d63fd6a91ac4e37fa690d8cf947a92fc39ba5a84ddd82a5","role":"COMMITTEE_RUNNER","ssv_message_type":"consensus","slot":2730838,"duty_id":"COMMITTEE-271_287_311_318-e85338-s2730838","round":5,"qbft_message_type":"round_change","peer_id":"16Uiu2HAmNTwEvEqrgVGyjyDuxEoPsPMgZXt5DzupPhaTnMeEFLYg","error":"too many duties per epoch, got 33 (role COMMITTEE_RUNNER), want less than 32"}
[...]

PR #2190, opened 2025-04-30
message/validation: fix previous epoch duty count calculation

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/1789


message/validation/consensus_state.go
```diff
@@ -68,10 +68,12 @@ func (os *OperatorState) Set(slot phase0.Slot, epoch phase0.Epoch, state *Signer
...
 		os.maxEpoch = epoch
-		os.prevEpochDuties = os.lastEpochDuties
-		os.lastEpochDuties = 1
+		os.prevEpochDuties = os.currEpochDuties
+		os.currEpochDuties = 1
+	} else if epoch == os.maxEpoch {
+		os.currEpochDuties++
 	} else {
-		os.lastEpochDuties++
+		os.prevEpochDuties++
```
diegomrsantos, line 76: Is this to cover a late message for the previous epoch?

iurii-ssv (approved):
Nice find, while browsing through related code I've noticed there was a somewhat relevant change in that area - https://github.com/ssvlabs/ssv/pull/1758 - that I couldn't quite make sense of due to a **TODO** in https://github.com/ssvlabs/ssv/pull/1758 tests:
```
// TODO: this should fail, see https://github.com/ssvlabs/ssv/pull/1758
```

@nkryuchkov @moshe-blox maybe you could clarify why that **TODO** exists and if we could address it in this PR as well ?

and also there is another out-of-date comment we probably want to fix:
```
// the first count starts at 0 and we use an inclusive comparison (>=).
```

kchojn (approved):

nkryuchkov:
> Nice find, while browsing through related code I've noticed there was a somewhat relevant change in that area - #1758 - that I couldn't quite make sense of due to a **TODO** in #1758 tests:
> 
> ```
> // TODO: this should fail, see https://github.com/ssvlabs/ssv/pull/1758
> ```
> 
> @nkryuchkov @moshe-blox maybe you could clarify why that **TODO** exists and if we could address it in this PR as well ?
> 
> and also there is another out-of-date comment we probably want to fix:
> 
> ```
> // the first count starts at 0 and we use an inclusive comparison (>=).
> ```

@iurii-ssv well spotted! I think https://github.com/ssvlabs/ssv/pull/1758 is a wrong attempt to fix the issue this PR fixes. The comparison should be inclusive because the state hasn't been updated yet. When we receive the first message that belongs to a duty, the count is 0 because updating happens after validation. When we receive the last possible one (e.g., 32th), the counter is 31, and then after validation passes, the counter increases to 32. So if we receive a message and the counter is already at limit (32), it means we are already beyond the limit.

@moshe-blox @y0sher FYI: I'm going to revert https://github.com/ssvlabs/ssv/pull/1758

nkryuchkov:
@iurii-ssv @MatusKysel @kchojn please review the latest change

kchojn (approved):

iurii-ssv (approved):

diegomrsantos:
Relevent to you guys https://github.com/sigp/anchor/pull/327

PR #2191, opened 2025-04-30
fix(Dockerfile.multiarch): Cross-compilation wasn't producing correct architecture binaries

kchojn:
> [!IMPORTANT]  
> This PR should be merged before the next release.
> It solves: https://github.com/ssvlabs/ssv/issues/2182

```md
# Cross-Platform Build Fix

## What We Fixed

### Issue: Cross-compilation wasn't producing correct architecture binaries
When building for ARM64 on an AMD64 host (or vice versa), the resulting binary had the host architecture rather than the target architecture.

## Changes Made

1. **Added Target Architecture Development Libraries**
   - Used `dpkg --add-architecture` to enable multi-arch support
   - Installed `libc6-dev:<target-arch>` packages to provide proper static libraries
   - Ensured availability of architecture-specific `libm.a` and `libmvec.a`
   
2. **Ensured Proper Cross-Compiler Selection**
   - Maintained explicit compiler selection based on `TARGETARCH`
   - Mapped each architecture to its specific cross-compiler:
     - AMD64: `x86_64-linux-gnu-gcc`
     - ARM64: `aarch64-linux-gnu-gcc`
     - ARM: `arm-linux-gnueabihf-gcc`

3. **Fixed Build Environment Variables**
   - Passed correct `CC` variable to Go build process

## Results

- Binary verification confirms correct ELF machine type:
  - ARM64 binaries show machine type `0xB7` (AArch64)
  - AMD64 binaries show machine type `0x3E` (x86-64)
- Docker containers now run on their target platforms without "Exec format error"
- Multi-architecture builds produce compatible binaries for each platform
```

Tests (macOs)
```bash
# ================================================
# VERIFICATION SCRIPT FOR ARM64 BUILD
# ================================================

# Create temporary container to extract binary
docker create --name temp_arm64_container ssvlabs/ssv-node:arm64-test
docker cp temp_arm64_container:/go/bin/ssvnode ./ssvnode_arm64
docker rm temp_arm64_container

# Basic file examination
echo "=== ARM64 Binary Basic Info ==="
file ./ssvnode_arm64

# Check ELF header with hexdump
echo "=== ARM64 Binary Hexdump ==="
hexdump -n 64 -C ./ssvnode_arm64

# Detailed ELF metadata check
echo "=== ARM64 Binary ELF Header ==="
docker run --rm -v $(pwd):/work alpine sh -c "apk add --no-cache binutils && readelf -h /work/ssvnode_arm64"

# Check Docker image metadata
[...]

Issue #2192, opened 2025-05-01
Consider reorganizing `Dockerfile` structure and build configuration


kchojn:
During my research on Go compilation and Docker best practices, I noticed our Docker configuration has several unusual aspects compared to other blockchain projects:
- mixed configuration:  we're using Debian-based images, but have comments and configuration related to Alpine (like `GODEBUG="netdns=go"`)
- static linking complexity: our current approach using `CGO_ENABLED=1` with `-linkmode external ... static` creates significant cross-compilaton challenges
- image size concerns: our final image is quite large, we could consider apline based images for smaller footprint (better for static linking due to musl libc), debial slim variants like bullseye-slim (2 separate Dockers?), distroless images for runtime
- build tags consistency: we could standarizing our approach to build tags like `blst_enabled` and `netgo` when appropriate 

Our future goals:
- consistent base images across build and runtime stages
- simplified cross-compilation support
- reduced image size
- clear documentation of intentional decisions, like `CGO` usage (blst needs it) 
- images with a specific digest

Hope this makes sense.

PR #2193, opened 2025-05-04
stage-main arm64 hotfix

y0sher:
### Description

Hotfix applied in #2191


kchojn (approved):

oleg-ssvlabs:
I got this warning while building the Docker image on Mac:

`WARN: RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 94)`

Docker Engine version: `28.1.1`.

It’s nothing critical, but maybe we could remove the `$TARGETPLATFORM` from the `FROM` statement, as the warning suggests.

Dockerfile.multiarch
```diff
@@ -36,47 +49,64 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
+          -linkmode external \
+          -extldflags '-static -lm'" \
+        -o /go/bin/ssvnode ./cmd/ssvnode
 
 #
-# STEP 3: Prepare image to run the binary
+# STEP 3: Runner image
 #
-FROM golang:1.24 AS runner
+FROM --platform=$TARGETPLATFORM golang:1.24 AS runner
```
oleg-ssvlabs, line 94: I think it’s better to use digests (SHA256 hashes) when specifying Docker image tags, like `golang:1.24@sha256:XXXX`

iurii-ssv (approved):

Dockerfile.multiarch
kchojn: I see that @MatusKysel pushed the changes, but imo it should also be part of this issue: https://github.com/ssvlabs/ssv/issues/2192
Our original `Dockerfile` still uses img without a specific tag

y0sher:
> I got this warning while building the Docker image on Mac:
> 
> `WARN: RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 94)`
> 
> Docker Engine version: `28.1.1`.
> 
> It’s nothing critical, but maybe we could remove the `$TARGETPLATFORM` from the `FROM` statement, as the warning suggests.
I ran it with --platform=linux/arm64 and saw no error

oleg-ssvlabs (approved):

PR #2196, opened 2025-05-07
proposer-duty: configurable MEV delay

iurii-ssv:
This PR adds a configuration setting for Operators to be able to configure a delay for  block-proposer duty that potentially allows for extracting better MEV (the later the Ethereum block is built the potentially more valuable it is).

_Before merging:_
- [x] test on **stage**


y0sher (changes requested):
Code looks good but I think we should add some example value in the help message

iurii-ssv:
> Code looks good but I think we should add some example value in the help message

Right, I wasn't sure what would be the best place to put up a detailed description/example ... I think it's best to link to an external document that explains everything - like I've added in https://github.com/ssvlabs/ssv/pull/2196/commits/ee2c93537a519e63b9312a56e2acab51806046ad now

protocol/v2/ssv/runner/proposer.go
```diff
@@ -55,11 +62,21 @@ func NewProposerRunner(
...
 	}
 
+	// Validate MEVDelay value, for details on how this value should be chosen see https://github.com/ssvlabs/ssv/blob/main/docs/MEV_CONSIDERATIONS.md#how-to-choose-mevdelay-value
+	const mevBoostRelayTimeout = 200 * time.Millisecond
+	const qbftMaxExpectedTime = 350 * time.Millisecond
+	const miscellaneousTime = 150 * time.Millisecond
+	const maxReasonableMEVDelay = roundtimer.QuickTimeout - mevBoostRelayTimeout - qbftMaxExpectedTime - miscellaneousTime
+	if mevDelay > maxReasonableMEVDelay {
+		return nil, fmt.Errorf("chosen MEVDelay value is too high: %s, max reasonable allowed value: %s", mevDelay, maxReasonableMEVDelay)
+	}
```
iurii-ssv, line 78: This would give us `maxReasonableMEVDelay` of `1.3s`, it does assume however that `mevBoostRelayTimeout` was manually adjusted down to `~200ms` (from it's default value of `~1s`)

without Operator adjust his `mevBoostRelayTimeout` the `maxReasonableMEVDelay` value should be more like `500ms` (like I explain in https://github.com/ssvlabs/ssv/blob/ddd37f369aa2b88872da0e62258ca6c04c4d7294/docs/MEV_CONSIDERATIONS.md#how-to-choose-mevdelay-value) ... but we cannot know whether Operator adjusted it or not, so we have to be "permissive" about it

protocol/v2/ssv/runner/sync_committee_aggregator.go
```diff
@@ -130,6 +131,7 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
 
 	if len(selectionProofs) == 0 {
 		r.GetState().Finished = true
+		r.measurements.EndDutyFlow()
 		return nil
 	}
```
iurii-ssv, line 136: I think we are "terminating" duty here and hence should also "end our measurements" in order to measure properly (even though, as it is right now, we don't make use of these measurements in this case - in the future we might need to measure/log this correctly - so I'm fixing it now while at it)

docs/MEV_CONSIDERATIONS.md
```diff
@@ -0,0 +1,73 @@
...
+```
+~100ms + MEVDelay + ~1000ms + ~350ms + ~1000ms + ~150ms < ~4000ms
+```
+Note:
+- these numbers from above are just rough estimates observed during our testing, use these "estimates" 
+  at your own risk
+- `MEVBoostRelayTimeout` value is set by Operator and can be adjusted in conjunction with `MEVDelay`,
+  for example, 2s can be allocated between `MEVBoostRelayTimeout` and `MEVDelay` as 1s & 1s or 0.5s & 1.5s
+  which would result in different MEV outcomes depending on the exact Relays Operator has configured his
+  MEV-boost with
```
y0sher, line 73: I think this or a shorter version of this about what are the best values we know so far and what you probably shouldn't exceed (ethereum's limit to publish a block. should be on top of this document, we aim to specify a pretty conservative value. 

Then below all the details about how we came up with this value and how can you optimize to something more suitable for you.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -247,108 +283,107 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
...
-		specSig := phase0.BLSSignature{}
-		copy(specSig[:], sig)
+	// only 1 root, verified by expectedPostConsensusRootsAndDomain
+	root := roots[0]
 
-		r.measurements.EndPostConsensus()
-		recordPostConsensusDuration(ctx, r.measurements.PostConsensusTime(), spectypes.RoleProposer)
+	sig, err := r.GetState().ReconstructBeaconSig(r.GetState().PostConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
+	if err != nil {
+		// If the reconstructed signature verification failed, fall back to verifying each partial signature
```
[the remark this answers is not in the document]
iurii-ssv: Yes, I was looking at this code while also adding/extending proposer duty logs and realized that in a couple of duties there is a `for _, root := range roots` construction that's not necessary because there can only be 1 root as per the comment:

```
// only 1 root, verified by expectedPostConsensusRootsAndDomain
root := roots[0]
```

Given the change is small I bundled it with this PR, but can move to a separate PR if you prefer that

docs/MEV_CONSIDERATIONS.md
iurii-ssv: Okay, reversed the notes order now and somewhat rephrased "our recommendations" - https://github.com/ssvlabs/ssv/pull/2196/commits/e27213129d9b4e332779c37d8a9a4cab47e09a2e - let me know if it makes sense or if there is anything else

protocol/v2/ssv/runner/proposer.go
```diff
@@ -55,11 +62,28 @@ func NewProposerRunner(
...
+	const blockSubmissionTime = 1000 * time.Millisecond
+	const maxMEVDelay = 4*time.Second - randaoTime - mevBoostRelayTimeout - qbftTime - blockSubmissionTime - miscellaneousTime
+	// Also, other SSV nodes in our cluster might not even have MEVDelay configured, meaning they will start
+	// QBFT sooner and timeout round 1 sooner. To prevent that, we'll need to cap maxMEVDelay accordingly
+	// (so it does not exceed qbftConstrainingTime).
+	const qbftConstrainingTime = roundtimer.QuickTimeout - qbftTime
+	const maxReasonableMEVDelay = min(maxMEVDelay, qbftConstrainingTime)
+	if mevDelay > maxReasonableMEVDelay {
+		return nil, fmt.Errorf("chosen MEVDelay value is too high: %s, max reasonable allowed value: %s", mevDelay, maxReasonableMEVDelay)
+	}
```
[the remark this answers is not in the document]
moshe-blox: i like the effort to calculate the ideal time 👍

with that said, i'd keep it outside the code, as you already did in the document

perhaps we can explain the reasoning in a comment here also

however the code should have a simple final constant such as `2200*time.Millisecond` for example

protocol/v2/ssv/runner/proposer.go
```diff
@@ -110,19 +136,28 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
+		fields.PreConsensusTime(r.measurements.PreConsensusTime()),
+	)
+
+	// Wait out MEV delay if any is configured.
+	time.Sleep(r.mevDelay)
 
+	// Fetch the block our operator will propose if it is a Leader (note, even if our operator
+	// isn't leading the 1st QBFT round it might become a Leader in case of round change - hence
+	// we are always fetching Ethereum block here just in case we need to propose it).
 	start := time.Now()
```
moshe-blox, line 161: this is true for all duties ATM, so i dont think it's worth to explain it here in the runners, because then you'd have to copy/paste this elsewhere as well

since we plan to change this in the upcoming fork, perhaps this can be left as-is for now :)

protocol/v2/ssv/runner/proposer.go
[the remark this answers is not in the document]
iurii-ssv: Okay, lets move the whole rationale behind `maxReasonableMEVDelay` to `.md` doc to keep it in one place (namely round-change aspect wasn't mentioned there before), this code is simplified to this now:
```go
	// Validate mevDelay value, for details on how this value should be chosen see:
	// https://github.com/ssvlabs/ssv/blob/main/docs/MEV_CONSIDERATIONS.md#choosing-mevdelay-value
	const maxReasonableMEVDelay = 1650 * time.Millisecond
	if mevDelay > maxReasonableMEVDelay {
		return nil, fmt.Errorf("chosen MEVDelay value is too high: %s, max reasonable allowed value: %s", mevDelay, maxReasonableMEVDelay)
	}
```

protocol/v2/ssv/runner/proposer.go
```diff
@@ -139,6 +164,7 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
 	logger.Info("🧊 got beacon block proposal",
 		zap.String("block_hash", blockSummary.Hash.String()),
 		zap.Bool("blinded", blockSummary.Blinded),
+		zap.Duration("mev_delay", r.mevDelay),
 		zap.Duration("took", time.Since(start)),
```
olegshmuelov, line 168: To improve clarity in logs related to block fetching, consider replacing:
```go
zap.Duration("took", time.Since(start)),
```
with:
```go
zap.Duration("block_fetch_time", time.Since(start) - r.mevDelay),
zap.Duration("total_time", time.Since(start)),
```

docs/MEV_CONSIDERATIONS.md
```diff
@@ -0,0 +1,80 @@
+## Choosing `MEVDelay` value
+
+As per our own estimates the max reasonable value of `MEVDelay` for Ethereum mainnet is around ~1.65s, 
+although we recommend starting with something like 300ms gradually increasing it up - the higher 
+`MEVDelay` value is the higher the chance of missing Ethereum block proposal will be.
+
+As per the notes in other sections of this document `MEVDelay` depends on a number of things, to find
```
y0sher, line 7: I think the first section should be a getting started section where we give an example of the CLI flag with the recommended value.

iurii-ssv: I guess something like this - https://github.com/ssvlabs/ssv/blob/f98225b1577b231249c24a7f100cee6df5033ed5/docs/MEV_CONSIDERATIONS.md#getting-started-with-mev-configuration ?

protocol/v2/ssv/runner/proposer.go
```diff
@@ -109,19 +125,28 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
-	logger.Debug("🧩 reconstructed partial RANDAO signatures",
+	logger.Debug(
+		"🧩 reconstructed partial RANDAO signatures",
 		zap.Uint64s("signers", getPreConsensusSigners(r.GetState(), root)),
-		fields.PreConsensusTime(r.measurements.PreConsensusTime()))
+		fields.PreConsensusTime(r.measurements.PreConsensusTime()),
+	)
+
+	// Wait out MEV delay if any is configured.
+	time.Sleep(r.mevDelay)
```
moshe-blox, line 145: we should sleep until r.mevDelay past the duty slot time, rather than always r.mevDelay exactly

thats because we might have started the duty a bit late (happens sometimes under load) and we don't really want to delay more than requested in that case

protocol/v2/ssv/runner/proposer.go
```diff
@@ -109,19 +125,34 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
 		zap.Uint64s("signers", getPreConsensusSigners(r.GetState(), root)),
-		fields.PreConsensusTime(r.measurements.PreConsensusTime()))
+		fields.PreConsensusTime(r.measurements.PreConsensusTime()),
+	)
 
+	// Wait out Proposer delay if any is configured, instead of sleeping exactly r.proposerDelay
+	// duration take slot start time as the sleep starting point and adjust the amount to sleep
+	// accordingly - this is done to make sure block-proposal is still made on time even if
+	// Proposer duty started a bit late (eg. due to node being under load).
+	slotStartTime := r.BaseRunner.NetworkConfig.GetSlotStartTime(duty.Slot)
```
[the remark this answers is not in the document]
iurii-ssv: We could do that, but I feel like the resulting **func** size is not quite deserving of a separate **func** ... so I wouldn't worry about it

protocol/v2/ssv/runner/proposer.go
```diff
@@ -55,11 +61,19 @@ func NewProposerRunner(
...
 		return nil, errors.New("must have one share")
 	}
 
+	// Validate proposerDelay value, for details on how this value should be chosen see:
+	// https://github.com/ssvlabs/ssv/blob/main/docs/MEV_CONSIDERATIONS.md#getting-started-with-mev-configuration
+	const maxReasonableProposerDelay = 1650 * time.Millisecond
+	if proposerDelay > maxReasonableProposerDelay {
+		return nil, fmt.Errorf("chosen ProposerDelay value is too high: %s, max reasonable allowed value: %s", proposerDelay, maxReasonableProposerDelay)
+	}
+
```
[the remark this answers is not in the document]
iurii-ssv: Hmm, actually out of 3 options I can think of:
- terminate
- default to 0 when `proposerDelay > maxReasonableProposerDelay`
- default to `maxReasonableProposerDelay` when `proposerDelay > maxReasonableProposerDelay`

the last one seems like most attractive, cc @y0sher @moshe-blox what do you guys think of it ?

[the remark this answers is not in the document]
iurii-ssv: > I like the last one 👍

Okay, applied for now https://github.com/ssvlabs/ssv/pull/2196/commits/a88bfa5eef156e5129d383f1bfa005ca1e32b082

protocol/v2/ssv/runner/aggregator.go
y0sher, line 1: to be honest, I like the edits here but I'd rather this PR to only touch the proposer runner flow code..

protocol/v2/ssv/runner/proposer.go
```diff
@@ -109,19 +126,34 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
+		"🧩 reconstructed partial RANDAO signatures",
 		zap.Uint64s("signers", getPreConsensusSigners(r.GetState(), root)),
-		fields.PreConsensusTime(r.measurements.PreConsensusTime()))
+		fields.PreConsensusTime(r.measurements.PreConsensusTime()),
+	)
 
+	// Wait out Proposer delay if any is configured, instead of sleeping exactly r.proposerDelay
+	// duration take slot start time as the sleep starting point and adjust the amount to sleep
+	// accordingly - this is done to make sure block-proposal is still made on time even if
+	// Proposer duty started a bit late (eg. due to node being under load).
```
y0sher: wdyt about this summary - 
```suggestion
// Sleep the remaining proposerDelay since slot start, ensuring on-time proposals even if duty began late.
```

protocol/v2/ssv/runner/proposer.go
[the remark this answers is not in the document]
y0sher: it's just easier to review a PR where every change is related. Trying to figure out if it has an impact and review something that is unrelated might delay the PR being merged without good reason. IMO we should separate this stuff. Also if this introduces a bug it won't be straight forward to find where it happened since it's hidden in this PR.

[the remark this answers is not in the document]
iurii-ssv: Yes, it's true, let me separate out the unrelated functionality & put up another PR for it (shouldn't be too hard)

protocol/v2/ssv/runner/aggregator.go
iurii-ssv: https://github.com/ssvlabs/ssv/pull/2196#discussion_r2124446513

protocol/v2/ssv/runner/proposer.go
iurii-ssv: Sounds good, applied

iurii-ssv:
Putting this PR in Draft for now until https://github.com/ssvlabs/ssv/pull/2196#discussion_r2124446513 is resolved

PR #2199, opened 2025-05-12
test(protocol/v2/ssv/validator/committee_queue_test.go): verify the functionality of handling messages in a committee queue

kchojn:


protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1834 @@
...
+//
+// POSSIBLE ISSUE:
+// This test suite has identified a potential issue with the queue's handling of messages under load and saturation conditions.
+//
+//  1. Message Dropping: When a queue fills up, new messages get dropped completely, even
+//     high-priority ones. This happens because the queue uses a non-blocking approach when
+//     receiving new messages. If there's no space, messages are discarded rather than waiting
+//     for space to become available. This means important messages could be lost during periods
+//     of heavy load or when messages get stuck in the queue due to filtering rules.
+func TestQueueLoadAndSaturationScenarios(t *testing.T) {
```
iurii-ssv, line 1488: I think it's better to create Github issue so we can improve this in the future and not forget about it (you could reference this comment from Github issue so it directly leads to this code & description you've left)

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1834 @@
...
+			Queues:        make(map[phase0.Slot]queueContainer),
+			Runners:       make(map[phase0.Slot]*runner.CommitteeRunner),
+			BeaconNetwork: mainBeaconNetwork,
+		}
+
+		slot := phase0.Slot(456)
+		queueCapacity := 5
+		currentRound := specqbft.Round(1)
+
+		runnerMutex := &sync.RWMutex{}
```
iurii-ssv, line 1680: I don't see what `runnerMutex` is doing for us given that we never "read/write" anything related to Runner (or its state) concurrently:
- we always initialize Runner (and its state)
- and only then we spawn a go-routine that interacts with Runner concurrently (but `go ... ()` statement takes care of synchronization for us - like I mentioned in another comment)

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1834 @@
...
+	Loop:
+		for {
+			select {
+			case <-handlerCalled:
+			case <-timeout:
+				break Loop
+			case <-ctx2.Done():
+				break Loop
+			}
+		}
```
iurii-ssv, line 1810: Just an opinion, it's probably easier to read something like this instead of loops with stuff like `break Loop`, wdyt?
```
		for done := false; !done; {
			select {
			case <-handlerCalled:
			case <-timeout:
				done = true
			case <-ctx2.Done():
				done = true
			}
		}
```

iurii-ssv (commented):
Code reads nicely (especially with all the comments),

Left minor some suggestions, mostly nits (feel free to ignore/skip those), but also sharing some thoughts on concurrency patterns (maybe even too much - but that would help us get on the same page sooner)

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1834 @@
...
+	}
+	decoded, err := queue.DecodeSignedSSVMessage(signed)
+	require.NoError(t, err)
+	return decoded
+}
+
+// safeConsumeQueue wraps ConsumeQueue execution in a goroutine with mutex protection.
+func safeConsumeQueue(t *testing.T, ctx context.Context, committee *Committee, q queueContainer,
+	logger *zap.Logger, handler func(context.Context, *zap.Logger, *queue.SSVMessage) error,
+	committeeRunner *runner.CommitteeRunner, mutex *sync.RWMutex) {
```
iurii-ssv, line 81: Do we need this mutex param for something ? Similar to other places in this PR I don't think this mutex does anything for us ...

Also, I guess "safe" in func name is also related to it, I'm just not seeing what it does for us (it's not like we re-use that mutex across multiple go-routines) - in case we don't need a mutex after all maybe we should inline this func instead

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1834 @@
...
+
+	t.Helper()
+
+	go func() {
+		mutex.RLock()
+		defer mutex.RUnlock()
+		err := committee.ConsumeQueue(ctx, q, logger, handler, committeeRunner)
+		require.NoError(t, err)
+	}()
+}
```
iurii-ssv, line 91: We cannot really use `t *testing.T` from another go-routine as is stated in their docs:
```go
// T is a type passed to Test functions to manage test state and support formatted test logs.
//
// FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as
// the Parallel method, must be called only from the goroutine running the
// Test function.
//
// The other reporting methods, such as the variations of Log and Error,
type T struct {
```

This is the only place in this PR affected by it.

kchojn:
Thanks @iurii-ssv for a valuable review! All these tricks with mutex, come from the fact that tests work differently locally and differently during ci/cd. I need to understand more about the github environment. Anyway, I like your comments and will fix the issues.

protocol/v2/ssv/validator/committee_queue_test.go
kchojn: yes, this is also mentioned in the code as well .
```go
// This test suite highlights a known behavior where the queue drops messages when full,
// even high-priority ones, due to its non-blocking intake.
// This behavior and potential improvements are tracked
// in GitHub issue #1680 (https://github.com/ssvlabs/ssv/issues/1680).
```

iurii-ssv (changes requested):
Almost there!

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -0,0 +1,1870 @@
...
+	// The `testMsgDrop` (Prepare message) should not be found.
+	// All `queueCapacity` messages found should be Proposal messages from Step 1.
+	foundDroppedMessage := false
+	proposalMessageCount := 0
+	for i := 0; i < queueCapacity; i++ {
+		popCtx, popCancel := context.WithTimeout(t.Context(), 200*time.Millisecond)
+		// Use FilterAny since we are just checking the contents, not a live consumption scenario.
+		// The prioritizer does not matter here as we drain the queue completely.
+		msg := qContainer.Q.Pop(popCtx, queue.NewCommitteeQueuePrioritizer(qContainer.queueState), queue.FilterAny)
+		popCancel()
```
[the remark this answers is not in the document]
kchojn: I have moved it in possible places. I left explicit `popCacnel()` call inside loops to avoid resource leaks (these are also simple cases)

iurii-ssv (approved):

iurii-ssv (approved):

oleg-ssvlabs (approved):

y0sher (approved):

PR #2226, opened 2025-05-27
message/validation: remove redundant slot checks

nkryuchkov:
As pointed out in https://github.com/sigp/anchor/pull/336/files#r2108843801, the slot check is already done in `Get` (old name) / `GetSignerState` (new name), so the outer check is redundant

Also, `GetSignerState`/`SetSignerState` seem like better names


kchojn (approved):
lgtm

oleg-ssvlabs (approved):

PR #2227, opened 2025-05-27
[SSV-1, SSV-13] ssvsigner: Fix Potential Race Condition in `RemoteKeyManager` Between Signing Checks and `BumpSlashingProtection()`

nkryuchkov:
> Synchronize `BumpSlashingProtection(pubKey)` with the signing locks used by `SignBeaconObject()`. For
example, have `BumpSlashingProtection(pubKey)` acquire `signLocks[pubKey, "attestation"]` and
`signLocks[pubKey, "proposal"]` before modifying the database. This ensures no interleaving can violate the intended check-then-update atomicity


ssvsigner/ekm/remote_key_manager.go
```diff
@@ -118,6 +118,14 @@ func (km *RemoteKeyManager) AddShare(
...
+	attLock.Lock()
+	defer attLock.Unlock()
+
+	propLock := km.lock(pubKey, lockProposal)
+	propLock.Lock()
+	defer propLock.Unlock()
+
 	if err := km.BumpSlashingProtection(pubKey); err != nil {
 		return fmt.Errorf("could not bump slashing protection: %w", err)
 	}
```
iurii-ssv, line 131: I'm not sure I follow what exactly locks managed by `km.lock` protect (or maybe what operations they ensure are atomic), namely I don't really understand why 
- here in `AddShare` method we are keeping `km.signerClient.AddValidators` method call outside of the lock-section, while in `SignBeaconObject` method we hold these locks also while `km.signerClient.Sign` is called; at the very least it looks inconsistent and confusing IMO
- I also see [another place where `eh.keyManager.BumpSlashingProtection` is called](https://github.com/ssvlabs/ssv/blob/649ed7b58541195b5b51a268423316de539b2f9c/eth/eventhandler/handlers.go#L429-L429) **without** any of these locks being acquired ... if the intent behind the locking is to permit only 1 `BumpSlashingProtection` at a time we probably should address that as well ?

iurii-ssv: Hey @nkryuchkov could you elaborate on that ^ ?

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -117,6 +117,14 @@ func (km *RemoteKeyManager) AddShare(
...
 
+	attLock := km.lock(pubKey, lockAttestation)
+	attLock.Lock()
+	defer attLock.Unlock()
+
+	propLock := km.lock(pubKey, lockProposal)
+	propLock.Lock()
+	defer propLock.Unlock()
+
```
moshe-blox: if these locks are needed for BumpSlashingProtection why not lock there instead?

moshe-blox: unless the callers might have already locked in the other code paths?

ssvsigner/ekm/remote_key_manager.go
oleg-ssvlabs: I have a very similar question.

I looked a bit into the code and it seems that the locks are designed not to mutually exclude access to a specific object, but rather to mutually exclude access to all objects related to a specific Validator + Operation. This is because very different resources are being locked with the same mutex lock.

I am not sure about the pattern, so it would be good to hear some explanation.

nkryuchkov: The lock design is copied from EKM. We need the locks to prevent updating the slashing protection DB on new events and checking slashability at the same time. Well spotted about locking `Sign` call and not locking `BumpSlashingProtection` in event syncer — fixed.

ssvsigner/ekm/remote_key_manager.go
nkryuchkov: moved them inside `BumpSlashingProtection`

iurii-ssv (approved):

oleg-ssvlabs (approved):

nkryuchkov:
supersedes https://github.com/ssvlabs/ssv/pull/2257

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -523,3 +548,15 @@ func (km *RemoteKeyManager) lock(sharePubkey phase0.BLSPubKey, operation string)
 	km.signLocks[key] = &sync.RWMutex{}
 	return km.signLocks[key]
 }
+
+func (km *RemoteKeyManager) BumpSlashingProtection(pubKey phase0.BLSPubKey) error {
```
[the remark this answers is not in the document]
nkryuchkov: The order is not important

kchojn (approved):

olegshmuelov (approved):
LGTM

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -503,12 +518,22 @@ func (km *RemoteKeyManager) GetOperatorID() spectypes.OperatorID {
...
+type lockOperation string
+
+const (
+	lockAttestation                    lockOperation = "attestation"
+	lockProposal                       lockOperation = "proposal"
+	lockSyncCommittee                  lockOperation = "sync_committee"
+	lockSyncCommitteeSelectionData     lockOperation = "sync_committee_selection_data"
+	lockSyncCommitteeSelectionAndProof lockOperation = "sync_committee_selection_and_proof"
+)
+
```
y0sher: this could be an "enum"-style integer const..

iurii-ssv (approved):

y0sher (approved):

PR #2230, opened 2025-05-27
[SSV-5] ssvsigner: High CPU Usage and Potential DoS in the Add Validator Flow Due to Unbounded Share Processing

nkryuchkov:
> Implement a strict limit on the number of shares processed in a single `POST /v1/validators` request (e.g., 5–
10 shares). Clients adding many shares must batch their requests, reducing maximum processing time and CPU load per request.

### DIFF
--- a/ssvsigner/server.go
+++ b/ssvsigner/server.go
@@ -39,6 +39,10 @@ const (
 	pathOperatorSign     = "/v1/operator/sign"     // TODO: /api/v1/ssv/sign ?
 )
 
+const (
+	addShareLimit = 10
+)
+
 type Server struct {
 	logger          *zap.Logger
 	operatorPrivKey keys.OperatorPrivateKey
@@ -162,6 +166,13 @@ func (s *Server) handleAddValidator(ctx *fasthttp.RequestCtx) {
 	}
 
 	logger = logger.With(zap.Int("req_count", len(req.ShareKeys)))
+
+	if len(req.ShareKeys) > addShareLimit {
+		logger.Warn("requested too many shares to be added")
+		s.writeJSONErr(ctx, logger, fasthttp.StatusBadRequest,
+			fmt.Errorf("requested too many shares to be added: %d", len(req.ShareKeys)))
+		return
+	}
 
 	var importKeystoreReq web3signer.ImportKeystoreRequest
 	for i, share := range req.ShareKeys {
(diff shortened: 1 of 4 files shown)


iurii-ssv (dismissed):
Asking just in case: do we need to implement some sort of self-limiting on SSV node side as well (to make sure we don't end up trying to add more than 10 validators in a perpetual loop or something, hitting this newly added error all the time) ?

iurii-ssv (dismissed):
LGTM (see also 1 comment above)

kchojn (dismissed):

oleg-ssvlabs (dismissed):

nkryuchkov:
> Asking just in case: do we need to implement some sort of self-limiting on SSV node side as well (to make sure we don't end up trying to add more than 10 validators in a perpetual loop or something, hitting this newly added error all the time) ?

The client that's used by ssv-node doesn't allow exceeding the limit. Also, the node calls it only for 1 validator at the moment

ssvsigner/server.go
```diff
@@ -39,6 +39,10 @@ const (
 	pathOperatorSign     = "/v1/operator/sign"     // TODO: /api/v1/ssv/sign ?
 )
 
+const (
+	addShareLimit = 10
```
kchojn, line 44: btw, where did the number come from? can we document it somehow?

kchojn (approved):

oleg-ssvlabs (approved):

iurii-ssv (approved):

ssvsigner/server.go
```diff
@@ -39,6 +39,11 @@ const (
 	pathOperatorSign     = "/v1/operator/sign"     // TODO: /api/v1/ssv/sign ?
 )
 
+const (
+	// Processing one share takes ~0.5-0.8s, so 10 shares seem a reasonable limit.
+	addShareLimit = 10
+)
```
iurii-ssv, line 45: Do we process shares sequentially though ? If not then maybe we could multiply that number by "average amount of threads a typical machine has" (which is maybe 4 at least these days) ?

Or actually maybe we could just mention this limit targets 1-thread machines - which is also a sort of standard way to do it.

iurii-ssv (approved):

ssvsigner/server.go
nkryuchkov: Well, it might depend on the remote signer implementation/configuration and the machine it uses, so while it's likely 4 or so, I'd rather assume it might be just 1 and keep the comment more abstract.

PR #2233, opened 2025-05-28
go.mod: update go-bitfield dependency

iurii-ssv:
Maybe replaces https://github.com/ssvlabs/ssv/pull/2223

It seems https://github.com/ssvlabs/ssv/pull/2223 is not sufficient to successfully update **go-bitfield** dependency - as you can see the build/pipelines in https://github.com/ssvlabs/ssv/pull/2223 fail and it seems some manual adjustments are needed to make it actually work correctly (which this PR implements essentially).


iurii-ssv:
It seems we'd also need to upgrade `go-eth2-client` to undergo the same dependency update since here: https://github.com/ssvlabs/ssv/blob/f00244cc74497fae7533f53aaf19d21854abc3d7/ssvsigner/ekm/local_key_manager_test.go#L264-L264

we need to pass in the `bitfield.Bitvector128` (so we still would need to import the older package).

Closing for now (can re-open once we need it) since keeping this rebased is also a chore.

PR #2236, opened 2025-05-28
[SSV-6] ssvsigner: Flawed Error Caching in checkCachePrivkey() Leads to Suppressed Errors and Potential Panic

nkryuchkov:
> The `once.Do()` ensures that this function only runs once. However, the `err` variable is within this
`checkCachePrivkey()` scope only. In other words, a local `err` variable is captured by the closure passed to `sync.Once.Do()`.
If `rsaPrivateKeyToOpenSSL()` fails during the first invocation, `err` is set accordingly. But on subsequent calls,
`sync.Once.Do()` does not re-run the closure and a new local `err` is initialized to `nil`, causing `checkCachePrivkey` to return
`(cachedPrivKey=nil, err=nil)`. This results in `SignRSA()` being called with a `nil` OpenSSL key handle, likely leading to a
panic.

### DIFF
--- a/ssvsigner/keys/rsa_linux.go
+++ b/ssvsigner/keys/rsa_linux.go
@@ -14,15 +14,17 @@ import (
 )
 
 type privateKey struct {
-	privKey       *rsa.PrivateKey
-	cachedPrivKey *openssl.PrivateKeyRSA
-	once          sync.Once
+	privKey        *rsa.PrivateKey
+	cachedPrivKey  *openssl.PrivateKeyRSA
+	opensslInitErr error
+	once           sync.Once
 }
 
 type publicKey struct {
-	pubKey       *rsa.PublicKey
-	cachedPubkey *openssl.PublicKeyRSA
-	once         sync.Once
+	pubKey         *rsa.PublicKey
+	cachedPubkey   *openssl.PublicKeyRSA
+	opensslInitErr error
+	once           sync.Once
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


iurii-ssv (approved):

kchojn (approved):

ssvsigner/keys/rsa_linux.go
```diff
@@ -54,11 +56,10 @@ func rsaPublicKeyToOpenSSL(pub *rsa.PublicKey) (*openssl.PublicKeyRSA, error) {
...
-	var err error
 	priv.once.Do(func() {
-		priv.cachedPrivKey, err = rsaPrivateKeyToOpenSSL(priv.privKey)
+		priv.cachedPrivKey, priv.opensslInitErr = rsaPrivateKeyToOpenSSL(priv.privKey)
 	})
-	return priv.cachedPrivKey, err
+	return priv.cachedPrivKey, priv.opensslInitErr
 }
 
 func SignRSA(priv *privateKey, data []byte) ([]byte, error) {
```
oleg-ssvlabs, line 65: is it possible that different private would be passed here as parameter?

Is it an operator private key?

oleg-ssvlabs (approved):

ssvsigner/keys/rsa_linux.go
nkryuchkov: Well, it's a wrapper for an RSA private key, but it's called from `(*privateKey).Sign`, which is known to callers as `OperatorPrivateKey`

oleg-ssvlabs: ah I see now. It looked like regardless what key is passed, the same cached key would have been returned, but now I see that `sync.Once` is actually on a key level.

y0sher (approved):

PR #2241, opened 2025-05-28
[S3] ssvsigner: Fragile Error String Comparison for Wallet-Not-Found

nkryuchkov:
> The `OpenWallet()` function checks for a missing wallet by comparing `err.Error() != "could not find wallet"`. Relying on exact error strings is brittle and may break if wrapped or modified.


iurii-ssv (dismissed):

ssvsigner/ekm/signer_storage.go
```diff
@@ -40,6 +40,10 @@ const (
 	highestProposalPrefix = prefix + "highest_prop-"
 )
 
+var (
+	ErrWalletNotFound = errors.New("could not find wallet")
```
oleg-ssvlabs: this error is used within the same package, so it could be private

kchojn:
> > While we're introducing sentinel errors, we should be consistent imo. Currently we have 3 different error handling patterns here:
> > 
> > * sentinel error
> > * wrapped error
> > * plain error
> > 
> > ```go
> > 	if !found {
> > 		return nil, ErrWalletNotFound
> > 	}
> > 	if err != nil {
> > 		return nil, fmt.Errorf("open wallet: %w", err)
> > 	}
> > 	if len(obj.Value) == 0 {
> > 		return nil, errors.New("failed to open wallet")
> > 	}
> > ```
> 
> @kchojn I think changing it just here won't give us much benefit, we need to do it once for the whole package, but given this is an audit PR, I'd rather keep its scope lower

@nkryuchkov makes sense. would be good to have this standardized in the future

kchojn (approved):

oleg-ssvlabs (approved):

y0sher (approved):

iurii-ssv (approved):

PR #2246, opened 2025-05-30
ssvsigner: handle request error after remote signer processed keys

nkryuchkov:
If a request fails (e.g. network issue, load balancer issue), the keys might still be added by the signer.
On restart, the node would crash due to duplicate or not found keys. To avoid that, we allow `StatusDuplicated`/`StatusNotFound`. Allowing the events with those statuses just from the first block would be enough but then the implementation becomes tricky, and anyway any failed block is first after restart.

As a more robust solution, we might implement a more complex solution, e.g. with request IDs and ability to track request completion by request ID

### DIFF
--- a/ssvsigner/ekm/remote_key_manager_test.go
+++ b/ssvsigner/ekm/remote_key_manager_test.go
@@ -73,7 +73,7 @@ func (s *RemoteKeyManagerTestSuite) TestRemoteKeyManagerWithMockedOperatorKey() 
 	s.client.On("AddValidators", mock.Anything, ssvsigner.ShareKeys{
 		PubKey:           pubKey,
 		EncryptedPrivKey: encShare,
-	}).Return(nil)
+	}).Return([]web3signer.Status{web3signer.StatusImported}, nil)
 
 	err := rm.AddShare(s.T().Context(), encShare, pubKey)
 
@@ -100,7 +100,7 @@ func (s *RemoteKeyManagerTestSuite) TestRemoveShareWithMockedOperatorKey() {
 	mockSlashingProtector.On("RemoveHighestAttestation", pubKey).Return(nil)
 	mockSlashingProtector.On("RemoveHighestProposal", pubKey).Return(nil)
 
-	s.client.On("RemoveValidators", mock.Anything, []phase0.BLSPubKey{pubKey}).Return(nil)
+	s.client.On("RemoveValidators", mock.Anything, []phase0.BLSPubKey{pubKey}).Return([]web3signer.Status{web3signer.StatusDeleted}, nil)
 
 	err := rm.RemoveShare(s.T().Context(), pubKey)
 


ssvsigner/client.go
```diff
@@ -137,7 +138,19 @@ func (c *Client) AddValidators(ctx context.Context, shares ...ShareKeys) (err er
...
+		allowedStatuses := []web3signer.Status{
+			web3signer.StatusImported,
+
+			// A failed request does not guarantee that the keys were not added.
+			// It's possible that the ssv-signer successfully added the keys,
+			// but a network error occurred before a response could be received.
+			// Or, if ssv-signer is behind a load balancer, the load balancer may return an error.
+			// In such cases, the node would crash and, upon restarting, encounter a duplicate key error.
+			// To handle this gracefully, we allow returning a duplicate key error without treating it as a failure.
+			web3signer.StatusDuplicated,
```
iurii-ssv: Although I like the context provided in this comment - it does make certain assumptions on how `AddValidators` method is used/called (eg. that it's SSV node calling it, and crashing). 

Maybe it's better to propagate an error that represents `web3signer.StatusDuplicated` status to the caller to let the caller decide how it wants to treat it, this way we could also log it as `Error` (or `Warn`) + the comment placement would be more appropriate.

I guess applies to `RemoveValidators` as well

iurii-ssv (approved):
LGTM, but left a minor suggestion

oleg-ssvlabs (approved):

ssvsigner/client.go
```diff
@@ -174,8 +187,20 @@ func (c *Client) RemoveValidators(ctx context.Context, pubKeys ...phase0.BLSPubK
...
+
+			// A failed request does not guarantee that the keys were not deleted.
+			// It's possible that the ssv-signer successfully deleted the keys,
+			// but a network error occurred before a response could be received.
+			// Or, if ssv-signer is behind a load balancer, the load balancer may return an error.
+			// In such cases, the node would crash and, upon restarting, encounter a not found key error.
+			// To handle this gracefully, we allow returning a not found key error without treating it as a failure.
+			web3signer.StatusNotFound,
+		}
+
```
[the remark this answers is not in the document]
y0sher: seems correct

ssvsigner/client.go
```diff
@@ -137,7 +138,19 @@ func (c *Client) AddValidators(ctx context.Context, shares ...ShareKeys) (err er
...
+			// A failed request does not guarantee that the keys were not added.
+			// It's possible that the ssv-signer successfully added the keys,
+			// but a network error occurred before a response could be received.
+			// Or, if ssv-signer is behind a load balancer, the load balancer may return an error.
+			// In such cases, the node would crash and, upon restarting, encounter a duplicate key error.
+			// To handle this gracefully, we allow returning a duplicate key error without treating it as a failure.
+			web3signer.StatusDuplicated,
+		}
+
```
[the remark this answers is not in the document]
y0sher: lets do it

ssvsigner/client.go
```diff
@@ -166,20 +165,19 @@ func (c *Client) RemoveValidators(ctx context.Context, pubKeys ...phase0.BLSPubK
...
+	statuses = make([]web3signer.Status, 0, len(resp.Data))
 	for _, data := range resp.Data {
-		if data.Status != web3signer.StatusDeleted {
-			return fmt.Errorf("received status %s", data.Status)
-		}
+		statuses = append(statuses, data.Status)
 	}
 
-	return nil
+	return nil, nil
```
olegshmuelov: This function should return the collected statuses, not nil.

Change line 180 from:
```go
return nil, nil
```
to:
```go
return statuses, nil
```
Also tests use blank identifier `_` to ignore the statuses return value, failing to validate the core functionality of this PR.

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -119,10 +141,31 @@ func (km *RemoteKeyManager) AddShare(
...
+		switch status {
+		case web3signer.StatusDeleted:
+			continue
+
+		case web3signer.StatusDuplicated:
+			// A failed request does not guarantee that the keys were not deleted.
+			// It's possible that the ssv-signer successfully deleted the keys,
+			// but a network error occurred before a response could be received.
+			// Or, if ssv-signer is behind a load balancer, the load balancer may return an error.
+			// In such cases, the node would crash and, upon restarting, encounter a not found key error.
```
olegshmuelov, line 157: The status handling logic seems inconsistent. For `RemoveValidators`, we should expect `StatusNotFound` when trying to delete a non-existent key, not `StatusDuplicated`.

Consider changing:
```go
case web3signer.StatusDuplicated:
    // Comment mentions "not found key error" but checks for duplicated
```

to:
```go
case web3signer.StatusNotFound:
    // A failed request does not guarantee that the keys were not deleted...
```
The current logic doesn't align with the comment or the expected Web3Signer behavior.

olegshmuelov (changes requested):

ssvsigner/client_test.go
```diff
@@ -257,7 +264,7 @@ func (s *SSVSignerClientSuite) TestRemoveValidators() {
 				writeJSONResponse(w, tc.expectedStatusCode, tc.expectedResponse)
 			})
 
-			err := s.client.RemoveValidators(t.Context(), tc.pubKeys...)
+			_, err := s.client.RemoveValidators(t.Context(), tc.pubKeys...)
```
olegshmuelov: should be something like that
```go
statuses, err := s.client.RemoveValidators(t.Context(), tc.pubKeys...)

if !tc.expectError {
    assert.Equal(t, tc.expectStatuses, statuses)
}
s.assertErrorResult(err, tc.expectError, t)
```

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -104,10 +105,31 @@ func (km *RemoteKeyManager) AddShare(
...
-	if err := km.signerClient.AddValidators(ctx, shareKeys); err != nil {
+	statuses, err := km.signerClient.AddValidators(ctx, shareKeys)
+	if err != nil {
 		return fmt.Errorf("add validator: %w", err)
 	}
 
+	for _, status := range statuses {
+		switch status {
+		case web3signer.StatusImported:
+			continue
```
y0sher: I think we should cover this part in tests

y0sher: also how is this differs than next case when there's no "continue"?

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -119,10 +141,31 @@ func (km *RemoteKeyManager) AddShare(
...
+
+		case web3signer.StatusNotFound:
+			// A failed request does not guarantee that the keys were not deleted.
+			// It's possible that the ssv-signer successfully deleted the keys,
+			// but a network error occurred before a response could be received.
+			// Or, if ssv-signer is behind a load balancer, the load balancer may return an error.
+			// In such cases, the node would crash and, upon restarting, encounter a not found key error.
+			// To handle this gracefully, we allow returning a not found key error without treating it as a failure.
+			km.logger.Warn("Attempted to delete non-existing share from the remote signer. " +
```
y0sher: similar comments as above

ssvsigner/ekm/remote_key_manager.go
nkryuchkov: @y0sher I think `continue` looks clearer than nothing, but I can remove it

y0sher: I don't mind lets just do continue in both or nothing in both.

olegshmuelov (approved):

PR #2204, opened 2025-05-14
fix(eth/executionclient): halve log events batch size on error

kchojn:

PR #2247, opened 2025-05-31
eth/executionclient: halve batch size on websocket read limit

nkryuchkov:
Increasing `DefaultHistoricalLogsBatchSize` from 200 to 1000 in https://github.com/ssvlabs/ssv/pull/2204 caused syncing from a new database to fail due to the `websocket: read limit exceeded` error.

The PR adds the mentioned error to the list of errors when subdivision applies 

@kchojn Do you think it's worth reducing the `DefaultHistoricalLogsBatchSize` to something in the middle?

### DIFF
--- a/eth/executionclient/errors.go
+++ b/eth/executionclient/errors.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/gorilla/websocket"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


kchojn (commented):
In my opinion, we should restore the previous limit (`200`), if this one was tested and reliable and caused less chance of errors.
This solution seems to be just a workaround for the underlying issue. The proper fix would be to make the ws read limit configurable on the client side: https://github.com/ethereum/go-ethereum/pull/27801

e.g.:
```go
rpcClient, err := rpc.DialOptions(ctx, ec.nodeAddr, 
    rpc.WithWebsocketMessageSizeLimit(10 * 1024 * 1024), 
)

ec.client = ethclient.NewClient(rpcClient)
```

let me know what do you think

nkryuchkov:
> In my opinion, we should restore the previous limit (`200`), if this one was tested and reliable and caused less chance of errors.
> 
> This solution seems to be just a workaround for the underlying issue. The proper fix would be to make the ws read limit configurable on the client side: https://github.com/ethereum/go-ethereum/pull/27801
> 
> 
> 
> e.g.:
> 
> ```go
> 
> rpcClient, err := rpc.DialOptions(ctx, ec.nodeAddr, 
> 
>     rpc.WithWebsocketMessageSizeLimit(10 * 1024 * 1024), 
> 
> )
> 
> 
> 
> ec.client = ethclient.NewClient(rpcClient)
> 
> ```
> 
> 
> 
> let me know what do you think 

@kchojn  sounds good, let's just change the value then and close this PR

PR #2259, opened 2025-06-04
linter: fix & improve formatting

iurii-ssv:
I've noticed that our import-formatting (and linting) feature we've added couple of month ago doesn't fully 100% as I expected, for example if you `git checkout stage` and run `make format` you'll see some files in there require formatting.

Turns out in order for **golangci** linter to catch it we need to explicitly enable `goimports` (and `gofmt` for additional `-s` simplify option) formatter.

This PR is:
- fixing imports for certain files that silently bypassed linter
- making sure the linter will notice mis-formatted imports
- performing some relevant cleanup in the **Makefile**


.golangci.yaml
```diff
@@ -214,11 +214,12 @@ issues:
...
   fix: false
 formatters:
+  enable:
+    - gofmt
+    - goimports
   settings:
     gofmt:
       simplify: true
-    gofumpt:
-      extra-rules: false
```
iurii-ssv, line 221: I'm not including/enabling `gofumpt` because while some suggestions it makes are useful - others are a bit questionable,

the way it was configured - it was disabled anyway, we could consider adding it in the future though.

moshe-blox (approved):
lgtm

PR #2263, opened 2025-06-04
compilation: fix due to skipped pipeline

iurii-ssv:
Apparently the aggregate PR that was supposed to sync **stage** branch with changes from **main** -https://github.com/ssvlabs/ssv/pull/2255 - didn't have its own pipeline running automatically, so that pipeline was triggered only when https://github.com/ssvlabs/ssv/pull/2255 got merged into **stage** branch, and it broke.

The root cause is that https://github.com/ssvlabs/ssv/pull/2255 contains https://github.com/ssvlabs/ssv/pull/2199 that's targeting **main** branch and hence it doesn't compile with some code in **stage** branch.

This PR fixes **stage** branch pipelines by adapting the changes in https://github.com/ssvlabs/ssv/pull/2199 to work with **stage** code version.

### DIFF
--- a/protocol/v2/ssv/validator/committee_queue_test.go
+++ b/protocol/v2/ssv/validator/committee_queue_test.go
@@ -15,9 +15,10 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
 	"go.uber.org/zap"
 
-	qbfttests "github.com/ssvlabs/ssv/integration/qbft/tests"
+	"github.com/ssvlabs/ssv/networkconfig"
 
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -362,10 +375,9 @@ func TestFilterNoProposalAccepted(t *testing.T) {
 	defer cancel()
 
 	committee := &Committee{
-		ctx:           ctx,
-		Queues:        make(map[phase0.Slot]queueContainer),
-		Runners:       make(map[phase0.Slot]*runner.CommitteeRunner),
-		BeaconNetwork: qbfttests.NewTestingBeaconNodeWrapped().GetBeaconNetwork(),
```
oleg-ssvlabs, line 368: I assume for this test, and some other below, it wasn't necessary `BeaconNetwork` to be set? Meaning the test scenario does not actually interact with it. Do I understand it right?

oleg-ssvlabs (approved):

protocol/v2/ssv/validator/committee_queue_test.go
iurii-ssv: Yes I think it doesn't touch it - it would **panic** otherwise,

it's actually `beaconConfig` now in **stage** branch (`BeaconNetwork` is the reason tests didn't compile because it's **main** branch equivalent)

The very 1st test in this file does perform a call to `beaconConfig`, I think the rest of the tests were just copy-pasting it "by habit"

oleg-ssvlabs: Good you removed it

y0sher: if this wasn't true, I can't see how the test would've passed without it.

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -166,15 +167,18 @@ func setupMessageCollection(capacity int) (chan *queue.SSVMessage, func(context.
...
+		ctx:          ctx,
+		Queues:       make(map[phase0.Slot]queueContainer),
+		Runners:      make(map[phase0.Slot]*runner.CommitteeRunner),
+		beaconConfig: networkconfig.NewMockBeacon(ctrl),
```
nkryuchkov, line 181: Can we use `networkconfig.TestNetwork.BeaconConfig` to keep it simpler? The mock beacon config is currently used where we need some irregular logic, like making `GetSlotStartTime` always return `time.Now()`, and I think we'll need to get rid of it eventually because such config usage is far from the real scenario, so the `TestNetwork` config should be enough

y0sher: @nkryuchkov I guess we can create a GH issue to set this in the future.

nkryuchkov: @y0sher https://github.com/ssvlabs/ssv/issues/2264

iurii-ssv: Let's try to do it in 2.3.4

oleg-ssvlabs: actually already created a task for this

PR #2266, opened 2025-06-04
[SSV-2] ssvsigner: SSV Signer Exposes Signing Operations over Unencrypted HTTP or TLS

nkryuchkov:
> Modify the SSV signer to enforce TLS for all connections by removing the HTTP fallback option. Make mutual TLS
mandatory for the signing endpoints by requiring both server and client authentication certificates. The server should not start if
proper mutual TLS is not configured with appropriate client certificates. Update the configuration validation to reject insecure settings
and ensure the implementation matches the documentation.
> 
> Furthermore, ensure that the certificate authority used to issue certificates cannot issue arbitrary certificates. It is also possible to use
a self-signed certificate by SSV.


ssvsigner/cmd/ssv-signer/ssv-signer.go
```diff
@@ -173,21 +181,14 @@ func setupWeb3SignerClient(endpoint string, timeout time.Duration, tlsConfig tls
 func startServer(logger *zap.Logger, listenAddr string, operatorKey keys.OperatorPrivateKey, web3SignerClient *web3signer.Web3Signer, tlsConfig tls.Config) error {
 	logger.Info("starting ssv-signer server",
 		zap.String("addr", listenAddr),
-		zap.Bool("tls_enabled", tlsConfig.ServerKeystoreFile != ""),
+		zap.Bool("tls_enabled", true),
```
kchojn: I think we should always choose to check some conditions, as we did before, instead of writing a hardcoded value like `true`.

ssvsigner/tls/tls.go
```diff
@@ -245,27 +201,23 @@ func createClientTLSConfig(certificate tls.Certificate, trustedFingerprints map[
...
 		Certificates: []tls.Certificate{certificate},
-	}
-
-	// Configure client authentication based on fingerprints
-	if len(trustedFingerprints) > 0 {
-		// Require client certificates but verify them using our custom verification
-		tlsConfig.ClientAuth = tls.RequireAnyClientCert
-		tlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		ClientAuth:   tls.RequireAnyClientCert,
+		// Removed ClientCAs
```
kchojn: is this comment needed?

kchojn (commented):
looks good, minor comments

ssvsigner/cmd/ssv-signer/ssv-signer.go
nkryuchkov: I just wanted to keep this logged to keep the same fields, but actually it's redundant

kchojn (dismissed):

oleg-ssvlabs (commented):
There are places in the code that still look like TLS is optional, like [here](https://github.com/ssvlabs/ssv/blob/stage/ssvsigner/server.go#L121), maybe change some conditions there as well?

ssvsigner/cmd/ssv-signer/ssv-signer.go
```diff
@@ -28,9 +28,9 @@ type CLI struct {
...
 	// Server TLS configuration (for incoming connections to SSV Signer)
-	KeystoreFile         string `env:"KEYSTORE_FILE" env-description:"Path to PKCS12 keystore file for server TLS connections"`
-	KeystorePasswordFile string `env:"KEYSTORE_PASSWORD_FILE" env-description:"Path to file containing the password for server keystore file"`
-	KnownClientsFile     string `env:"KNOWN_CLIENTS_FILE" env-description:"Path to known clients file for authenticating clients"`
+	KeystoreFile         string `env:"KEYSTORE_FILE" required:"" env-description:"Path to PKCS12 keystore file for server TLS connections"`
+	KeystorePasswordFile string `env:"KEYSTORE_PASSWORD_FILE" required:"" env-description:"Path to file containing the password for server keystore file"`
+	KnownClientsFile     string `env:"KNOWN_CLIENTS_FILE" required:"" env-description:"Path to known clients file for authenticating clients"`
 
 	// Client TLS configuration (for connecting to Web3Signer)
 	Web3SignerKeystoreFile         string `env:"WEB3SIGNER_KEYSTORE_FILE" env-description:"Path to PKCS12 keystore file for TLS connection to Web3Signer"`
```
oleg-ssvlabs: Aren't Client certificates required as well for mutual TLS(and mTLS seems to be a requirement following the description)?

nkryuchkov:
> There are places in the code that still look like TLS is optional, like [here](https://github.com/ssvlabs/ssv/blob/stage/ssvsigner/server.go#L121), maybe change some conditions there as well?

Well spotted. I'm actually considering allowing disabling TLS with a specific flag, which might be very useful for testing, e.g., for a local testnet, so maybe I'll revert non-TLS code back

ssvsigner/cmd/ssv-signer/ssv-signer.go
nkryuchkov: We have only ssv-signer server and web3signer client options here. The ssv-signer client options are defined in `SSVSignerConfig` in ssv-node

y0sher (commented):
I know this requirement came from the audit, but as I see it web3signer doesn't require TLS and it's disabled by default. It makes sense since these kind of measures can also be done on the infrastructure level. I'm not sure if we will follow this suggestion. Would love more opinions though - @iurii-ssv @nkryuchkov @kchojn @moshe-blox @liorrutenberg

kchojn:
A few thoughts from me:
- web3signer docs recommend mTLS for production, they have it in their best practices (https://docs.web3signer.consensys.io/get-started/key-best-practices#enable-tls-authentication-between-validator-client-and-web3signer)
- I think infra-side TLS level is not enough, and only app-level mTLS can fully protect against internal threats (compromised pod?)
- to balance it somehow, we can default TLS to enabled in production, but also add `--allow-insecure-http` flag (for example behind SSV_SIGNER_UNSAFE=1 env var) for local development/testing etc. I think its a common pattern. Would be good to add some `log.Warn` when `insecure` flag is enabled

iurii-ssv:
> Would love more opinions though

I think the simplest & secure way to go about it is something similar to what @kchojn suggests:

> to balance it somehow, we can default TLS to enabled in production, but also add --allow-insecure-http flag (for example behind SSV_SIGNER_UNSAFE=1 env var) for local development/testing etc. I think its a common pattern. Would be good to add some log.Warn when insecure flag is enabled

But maybe implemented as follows:
- for production (we can check if we are running with **mainnet**) make mTLS a MUST
- for any other environment mTLS will be optional (and applied only if configured)

this will ensure production is always secure + also allow for easy testing setups

ssvsigner/cmd/ssv-signer/ssv-signer.go
```diff
@@ -105,16 +114,26 @@ func run(logger *zap.Logger, cli CLI) error {
...
+		if cli.KeystoreFile == "" {
+			return fmt.Errorf("server TLS keystore file is required")
+		}
+		if cli.KeystorePasswordFile == "" {
+			return fmt.Errorf("server TLS keystore password file is required")
+		}
+		if cli.KnownClientsFile == "" {
+			return fmt.Errorf("known clients file is required for client authentication")
+		}
+	}
```
iurii-ssv: I think the way we currently handle TLS is that it is possible for `cli.AllowInsecureHTTP == true` while TLS is actually **enabled** (being used),

this might be a bit confusing/unexpected for people running ssv-signer, so unless there is a reason to do it like that for the case of `cli.AllowInsecureHTTP == true` I would instead:
- either verify that both `ClientKeystoreFile` and `ClientServerCertFile` are empty to ensure no TLS is enabled
- or implement `AllowInsecureHTTP` flag such that it explicitly disables TLS (and rename it to something like `DisableTLS` to reflect that)

option 2 seems like the most appropriate to me, wdyt ?

iurii-ssv (commented):
Left 1 comment

ssvsigner/cmd/ssv-signer/ssv-signer.go
nkryuchkov: @iurii-ssv but what if we choose the second option, and a user passes some TLS-related flags? I think it might be confusing. I'd rather do an inverted check that no TLS args are passed if insecure mode is on. It will also keep the code simple

nkryuchkov: Actually after more thoughts I think we need to allow providing just some options but not all of them, it might be useful for testing

iurii-ssv: Yeah it's a bit awkward to handle these options

ssvsigner/cmd/ssv-signer/ssv-signer.go
```diff
@@ -105,16 +114,56 @@ func run(logger *zap.Logger, cli CLI) error {
...
+			}
+		}
+
+		if filesFound == len(allFiles) {
+			return fmt.Errorf("insecure mode is enabled, but all TLS options are provided, please disable the insecure mode")
+		}
+	} else {
+		if cli.KeystoreFile == "" {
+			return fmt.Errorf("server TLS keystore file is required")
+		}
```
y0sher, line 148: I don't get it, shouldn't it be ok to not provide any keystore files if `AllowInsecureHTTP` is true?

iurii-ssv: As the code written, I think it is OK to not provide any keystore files if `AllowInsecureHTTP` is true, 

could you expand on what's looking wrong here @y0sher ?

y0sher: Oh you are correct! I misread the ending brace location, thanks.

y0sher (approved):

PR #2268, opened 2025-06-05
validator-startup: simplify & address edge-cases

iurii-ssv:
This PR clarifies & simplifies validator-startup functionality to address 2 things I've noticed while working on the code in this area:
- the code in the **stage** branch seems to be "subscribing to 1 random subnet only when there is at least 1 non-liquidated share in DB" - is that the correct behavior ? I don't understand how that would be different from the case when we have some shares in DB, but they aren't our own (`ownShares` is empty) - in that case we will subscribe to 1 random subnet. So in this PR I unify this behavior to be the same for both cases - which is "subscribe to 1 random subnet when there are no non-liquidated own shares in DB"
- additionally, whenever SSV node couldn't **initialize** or **start** at least 1 validator (when it has non-liquidated own shares in DB) we should crash SSV node because it won't really work properly in that case anyway (and corresponding error will be easy to find in case of such a node-crush)


operator/validator/controller_test.go
```diff
@@ -191,28 +190,6 @@ func TestSetupValidatorsExporter(t *testing.T) {
 			network.EXPECT().ActiveSubnets().Return(subnets).AnyTimes()
 			network.EXPECT().FixedSubnets().Return(commons.Subnets{}).AnyTimes()
 
-			if tc.shareStorageListResponse == nil {
```
nkryuchkov, line 195: Please delete `shareStorageListResponse` from the struct

nkryuchkov: Actually, other fields aren't used either. I think, we can remove table test structure here

iurii-ssv: Done, reduced this test to the necessary bare minimum

operator/validator/controller.go
```diff
@@ -436,43 +436,59 @@ func (c *controller) handleNonCommitteeMessages(
...
+	// Initialize validators.
+	validators, committees, err := init()
+	if err != nil {
+		return fmt.Errorf("init validators: %w", err)
+	}
+	if len(validators) == 0 {
+		// If no validators were initialized - we're not subscribed to any subnets,
+		// we have to subscribe to 1 random subnet to participate in the network.
+		if err := c.network.SubscribeRandoms(1); err != nil {
+			c.logger.Error("failed to subscribe to random subnets", zap.Error(err))
```
olegshmuelov: Could you clarify why this error is only logged rather than returned?

operator/validator/controller.go
```diff
@@ -436,43 +436,59 @@ func (c *controller) handleNonCommitteeMessages(
...
+		}
+		return nil
```
olegshmuelov, line 486: Consider adding a log statement before returning nil to make this scenario more visible:
```go
c.logger.Info("successfully subscribed to random subnet, no validators to start")
return nil
```
This helps operators understand that the node is running without validators but is still participating in the network.

operator/validator/controller.go
```diff
@@ -436,43 +436,59 @@ func (c *controller) handleNonCommitteeMessages(
...
+
+	// Start validators.
+	started := c.startValidators(validators, committees)
+	if started == 0 {
+		return fmt.Errorf("none of %d validators were successfully started", len(validators))
 	}
+	return nil
 }
 
 // setupValidators setup and starts validators from the given shares.
```
olegshmuelov: not related to your changes

The comment is misleading. This function only sets up validators, it doesn't start them.
```go
// setupValidators sets up validators from the given shares.
```

operator/validator/controller.go
iurii-ssv: That's an interesting question, I simply "ported" the behavior we have in **stage** branch - but upon another look it does seem like we should return an error here (meaning, crash the node) ...

@MatusKysel any objections if we return error instead of logging here ?

operator/validator/controller.go
```diff
@@ -436,43 +436,59 @@ func (c *controller) handleNonCommitteeMessages(
...
+		defer close(c.committeeValidatorSetup)
+
+		// Load non-liquidated shares that belong to our own Operator.
+		ownShares := c.sharesStorage.List(
+			nil,
+			registrystorage.ByNotLiquidated(),
+			registrystorage.ByOperatorID(c.operatorDataStore.GetOperatorID()),
+		)
+		if len(ownShares) == 0 {
+			c.logger.Info("no own non-liquidated validator shares found in DB")
```
y0sher: If we are writing a info log, I think it's appropriate to at least add something simpler and more "user-oriented" like - "no validators to start" and then specify the reason like you did.

operator/validator/controller.go
y0sher: @olegshmuelov @iurii-ssv 
I think we should return an error and crash the node. without subscribing any subnet we won't connect to anyone.

iurii-ssv: applied

operator/validator/controller.go
iurii-ssv: adjusted log-lines as needed - https://github.com/ssvlabs/ssv/commit/711da58553a072a7f377c3d58a9e9b6554584173

olegshmuelov (dismissed):
LGTM

operator/validator/controller.go
```diff
@@ -436,49 +436,66 @@ func (c *controller) handleNonCommitteeMessages(
...
-		c.startValidators(inited, committees)
+		return validators, committees, nil
 	}
+
+	// Initialize validators.
+	validators, committees, err := init()
+	if err != nil {
+		return fmt.Errorf("init validators: %w", err)
+	}
+	if len(validators) == 0 {
```
oleg-ssvlabs, line 479: From what I understand, this condition can only be `true` in one scenario – when there are no “own shares” in the database, right?

oleg-ssvlabs (approved):
good job.
The implementation now seems much cleaner and easier to follow.
I left one question

operator/validator/controller.go
iurii-ssv: I think there is more to it than that, meaning there are various other scenarios when we couldn't initialize validators here (eg. if there is no validator metadata at this moment - see `skipping validator until it becomes active` case)

this PR is not changing any of that behavior though (except for what's mentioned in the PR description)

nkryuchkov (approved):

olegshmuelov (approved):

y0sher (approved):
lgtm

PR #2270, opened 2025-06-06
[SSV-11] ssvsigner: Incomplete Database State Due to Early Error Returns in Multi-Step Operations

nkryuchkov:
> **Description**: Functions in the codebase that perform multiple related database operations in sequence can leave the system in an inconsistent state if one operation fails. When operations return immediately after encountering an error, previously completed steps are not rolled back.
> 
> The following function can result in incomplete database state during early errors:
> 1. `LocalKeyManager.AddShare`
> 2. `LocalKeyManager.RemoveShare`
> 3. `RemoteKeyManager.AddShare`
> 4. `RemoteKeyManager.RemoveShare`
> 5. `SlashingProtector.BumpSlashingProtection`
>
> **Recommendation**: Consider implementing transactions or compensating actions for multi-step operations. Either wrap related operations in a database transaction so they can be rolled back together, or implement cleanup/recovery logic to handle partial failures. For non-database operations, consider implementing a pattern where changes are prepared but only committed after all steps succeed.


eth/eventhandler/handlers.go
oleg-ssvlabs, line 1: unrelated to this PR:

I wonder if the Share should be removed from the Key Manager on Validator Exited event? Somewhere [here](https://github.com/ssvlabs/ssv/pull/2270/files#diff-237800d70d85b7c40aac9c9692b20fbad16a1ff8cc54d2461a700e769b9954c1R487) in the handler.

ssvsigner/ekm/local_key_manager.go
```diff
@@ -263,16 +270,21 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
 		return ShareDecryptionError{Err: errors.New("share private key does not match public key")}
 	}
 
+	if err := km.slashingProtector.BumpSlashingProtectionTxn(txn, phase0.BLSPubKey(sharePrivKey.GetPublicKey().Serialize())); err != nil {
```
oleg-ssvlabs, line 298: From what I understand this line is not related to issue described in this PR?

Previously the slashing protection was not bumped when Share was added - and it wasn't correct. Do I understand it right?

oleg-ssvlabs:
For other reviewing this PR, I added a diagram that illustrates the communication of involved components and how Transaction object is passed:

image

ssvsigner/ekm/local_key_manager.go
```diff
@@ -282,38 +294,41 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
 // RemoveShare removes the share from the local wallet, clears the associated
 // slashing-protection records (highest attestation/proposal) for the given
 // public key, and returns an error on any storage issue.
-func (km *LocalKeyManager) RemoveShare(_ context.Context, pubKey phase0.BLSPubKey) error {
+func (km *LocalKeyManager) RemoveShare(_ context.Context, txn basedb.Txn, pubKey phase0.BLSPubKey) error {
 	km.walletLock.Lock()
 	defer km.walletLock.Unlock()
 
-	pubKeyHex := hex.EncodeToString(pubKey[:]) // pubKey.String() would add the "0x" prefix so we cannot use it
+	if err := km.slashingProtector.RemoveHighestAttestationTxn(txn, pubKey); err != nil {
```
oleg-ssvlabs, line 326: Unrelated to this PR, but I think this type embedding (`LocalKeyManager` IS `SlashingProtector`) feels a little bit redundant. I think SlashingProtector could be just injected through ctor? Wdyt?

ssvsigner/ekm/local_key_manager.go
```diff
@@ -282,38 +294,41 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
...
+		return fmt.Errorf("could not remove highest attestation: %w", err)
+	}
+	if err := km.slashingProtector.RemoveHighestProposalTxn(txn, pubKey); err != nil {
+		return fmt.Errorf("could not remove highest proposal: %w", err)
+	}
 
+	pubKeyHex := hex.EncodeToString(pubKey[:]) // pubKey.String() would add the "0x" prefix so we cannot use it
 	acc, err := km.wallet.AccountByPublicKey(pubKeyHex)
```
oleg-ssvlabs, line 334: May I ask what exactly means when `acc` is nil? 
From what I understand `wallet` is in-memory storage of Accounts. What is an account?

oleg-ssvlabs (commented):
Great job!

I left few comments.

Additional question I have is - from what I understand both Node Store and Signer Store is a single database instance, otherwise we wouldn't be able to propagate a transaction across two different database instances. Do I understand it right?

ssvsigner/ekm/local_key_manager.go
iurii-ssv: Yeah +1 on this, type-embedding is typically needed when you want to expose some public stuff on the type being embedded - which isn't really the case here ? 

If that's easy to refactor I think we can do it in this PR while at it.

ssvsigner/ekm/local_key_manager.go
```diff
@@ -243,9 +245,14 @@ func (km *LocalKeyManager) signBeaconObject(
 
 // AddShare decrypts the provided share private key (encryptedSharePrivKey)
 // using the operatorDecrypter, verifies that it matches sharePubKey, and
-// saves it to the local wallet. It also calls BumpSlashingProtection to
+// saves it to the local wallet. It also calls bumps slashing protection to
```
iurii-ssv: `calls bumps` -> `bumps`

ssvsigner/ekm/key_manager.go
```diff
@@ -28,21 +30,21 @@ func (e ShareDecryptionError) Unwrap() error {
 // and slashingProtector (for slashing checks and updates).
 type KeyManager interface {
 	BeaconSigner
-	slashingProtector
+	BumpSlashingProtectionTxn(txn basedb.Txn, pubKey phase0.BLSPubKey) error
```
iurii-ssv: Maybe `BumpSlashingProtectionTxn` -> `BumpSlashingProtection` - don't think other methods on this **interface** have `Txn` suffix + there is no `BumpSlashingProtection` to differentiate from ?

ssvsigner/ekm/local_key_manager.go
iurii-ssv: > What is an account?

Account is probably just some **struct** with ways to represent & access private/public keys

> From what I understand wallet is in-memory storage of Accounts.

I think it's not in-memory actually, we pass in db-based storage [here](https://github.com/ssvlabs/ssv/blob/fe9d41cfcab8644899d075964d4c6c38d66c08e4/ssvsigner/ekm/local_key_manager.go#L77-L77) - hence it will store some data on disk,

I think the problem is that if transaction is eventually rolled back the `Wallet` will not know about it since it doesn't support transactions (it doesn't expose any ways for it's methods to be called within a transaction, and transaction is an implementation details of storage - while `Wallet` can support different storage types)

This all means that it's possible for `acc != nil` if transaction was rolled back previously (and SSV node crashed but restarted), and so we need to handle this separately on our own since maybe `wallet.AddValidatorAccount` isn't idempotent ... but I'm guessing at this point :)

ssvsigner/ekm/local_key_manager.go
```diff
@@ -282,38 +294,41 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
...
 	return nil
 }
 
-func (km *LocalKeyManager) saveShare(privKey *bls.SecretKey) error {
+func (km *LocalKeyManager) saveAccount(privKey *bls.SecretKey) error {
 	key, err := core.NewHDKeyFromPrivateKey(privKey.Serialize(), "")
 	if err != nil {
-		return fmt.Errorf("could not generate HDKey: %w", err)
+		return fmt.Errorf("generate HDKey: %w", err)
 	}
```
iurii-ssv: I wonder if we need to return "custom error message" instead of wrapping `err` that might contain private-key material in it (similar to what we did in another PR recently) ?

ssvsigner/ekm/local_key_manager.go
```diff
@@ -282,38 +294,41 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
...
 	account := wallets.NewValidatorAccount("", key, nil, "", nil)
 	if err := km.wallet.AddValidatorAccount(account); err != nil {
-		return fmt.Errorf("could not save new account: %w", err)
+		return fmt.Errorf("add to wallet: %w", err)
```
iurii-ssv: Maybe expand: `add to wallet` -> `add validator account to wallet`

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -113,12 +114,18 @@ func (km *RemoteKeyManager) AddShare(
...
-		return fmt.Errorf("could not bump slashing protection: %w", err)
+	// If txn gets rolled back after share is saved,
+	// there will be some inconsistency between syncer state and remote signer.
+	// However, syncer crashes node on an error and restarts the sync process from the failing block,
+	// so it will attempt to save the same share again, which won't be an issue
+	// because AddValidators doesn't fail if the same share exists.
+	// TODO: The comment assumes that https://github.com/ssvlabs/ssv/pull/2246 has already been ported but it hasn't yet
+	if err := km.signerClient.AddValidators(ctx, shareKeys); err != nil {
+		return fmt.Errorf("add validator: %w", err)
 	}
```
iurii-ssv: I think we should remove this TODO and just "not merge this PR until #2246 is merged" ? Applies to other places too.

ssvsigner/ekm/signer_storage.go
```diff
@@ -46,14 +46,25 @@ type Storage interface {
...
+	RetrieveHighestAttestationTxn(r basedb.Reader, pubKey []byte) (*phase0.AttestationData, bool, error)
+	SaveHighestProposalTxn(rw basedb.ReadWriter, pubKey []byte, slot phase0.Slot) error
+	RetrieveHighestProposalTxn(r basedb.Reader, pubKey []byte) (phase0.Slot, bool, error)
+	RemoveHighestAttestationTxn(rw basedb.ReadWriter, pubKey []byte) error
+	RemoveHighestProposalTxn(rw basedb.ReadWriter, pubKey []byte) error
 	ListAccountsTxn(r basedb.Reader) ([]core.ValidatorAccount, error)
 	SaveAccountTxn(rw basedb.ReadWriter, account core.ValidatorAccount) error
-
-	BeaconNetwork() beacon.BeaconNetwork
 }
```
iurii-ssv, line 72: Similar to another comment, I don't think we need to mention `Txn` suffix in there since there is no other alternative ?

And also, it's not even true that `Reader` or `ReadWriter` is an actual DB transaction - might be something else.

iurii-ssv (commented):
Some minor suggestions

nkryuchkov:
> @nkryuchkov Could we make this PR simpler by also editing the EKM lib itself instead of moving all the code into this repo?

We aren't moving a lot of code currently, could you please give some examples? I thought about changing EKM first, but it doesn't know about badgerdb, so I don't want to make it depend on the DB

nkryuchkov:
> Additional question I have is - from what I understand both Node Store and Signer Store is a single database instance, otherwise we wouldn't be able to propagate a transaction across two different database instances. Do I understand it right?

@olegshmuelov Yes, we use the same instance created in `setupDB` for all storages. Theoretically, we can misuse the components and pass different DBs, but I don't have any quick solution for this in my mind

ssvsigner/ekm/remote_key_manager.go
nkryuchkov: Please don't resolve this conversation, let's keep it unresolved until https://github.com/ssvlabs/ssv/pull/2246 is merged

ssvsigner/ekm/local_key_manager.go
nkryuchkov: I extracted it from `if acc == nil {` because it's reverted on failed tx.

Previously, we didn't have any tx, so we just checked if it was added by checking if the account exists. Now we still need to do it only for saving share, but as bumping uses tx, we can extract it

ssvsigner/ekm/local_key_manager.go
nkryuchkov: `slashingProtector`'s methods are widely used by other code and tests, but not all of them, however, I agree it's better not to embed it — fixed

ssvsigner/ekm/local_key_manager.go
nkryuchkov: @iurii-ssv yes, you're right. I'd pass txn to the wallet methods, however, it would require modifying EKM and making it rely on badgerDB, although it is currently not even a dependency. But not passing a txn should also work as we can check if account exists

ssvsigner/ekm/key_manager.go
nkryuchkov: It was used by the embedding, but after we got rid of it, yes, we can do it

ssvsigner/ekm/remote_key_manager.go
```diff
@@ -113,12 +114,18 @@ func (km *RemoteKeyManager) AddShare(
...
```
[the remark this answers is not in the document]
nkryuchkov: it's intended: https://github.com/ssvlabs/ssv/pull/2239

ssvsigner/ekm/signer_storage.go
nkryuchkov: Some of them are duplicates of EKM's interfaces that use txn, the rest have `Txn` suffixes to be consistent with the others

`Reader` and `ReadWriter` are done similarly to underlying DB methods to keep interfaces narrow

iurii-ssv: Okay, I see 👍

iurii-ssv (dismissed):

y0sher (approved):

kchojn (approved):

eth/eventhandler/handlers.go
nkryuchkov: https://github.com/ssvlabs/ssv/issues/2309

ssvsigner/ekm/mock.go
```diff
@@ -268,17 +268,17 @@ func (m *MockSlashingProtector) UpdateHighestProposal(pubKey phase0.BLSPubKey, s
...
 }
 
-func (m *MockSlashingProtector) RemoveHighestAttestation(pubKey phase0.BLSPubKey) error {
+func (m *MockSlashingProtector) RemoveHighestAttestationTxn(txn basedb.Txn, pubKey phase0.BLSPubKey) error {
 	args := m.Called(pubKey)
 	return args.Error(0)
 }
 
-func (m *MockSlashingProtector) RemoveHighestProposal(pubKey phase0.BLSPubKey) error {
+func (m *MockSlashingProtector) RemoveHighestProposalTxn(txn basedb.Txn, pubKey phase0.BLSPubKey) error {
```
olegshmuelov: The mock methods are missing the `txn` parameter in the `m.Called()` invocation. This currently doesn't cause test failures because the test expectations are also missing the `txn` parameter, but this creates a fragile situation where future correct tests will fail unexpectedly.

**Mock fix needed:**
```go
func (m *MockSlashingProtector) RemoveHighestAttestationTxn(txn basedb.Txn, pubKey phase0.BLSPubKey) error {
	args := m.Called(txn, pubKey)
	return args.Error(0)
}

func (m *MockSlashingProtector) RemoveHighestProposalTxn(txn basedb.Txn, pubKey phase0.BLSPubKey) error {
	args := m.Called(txn, pubKey)
	return args.Error(0)
}
```

**Test expectations also need fixing** (in `remote_key_manager_test.go`):
```go
// Instead of:
mockSlashingProtector.On("RemoveHighestAttestationTxn", pubKey).Return(nil)

// Should be:  
mockSlashingProtector.On("RemoveHighestAttestationTxn", txn, pubKey).Return(nil)
```

iurii-ssv (approved):

olegshmuelov (changes requested):

iurii-ssv (approved):

PR #2274, opened 2025-06-09
feat(network/discovery/enode.go): implement `HostDNS` to `HostAddress` failover mechanism

kchojn:
Add failover logic: if DNS resolution fails, fallback to HostAddress.
Only return DNS error if both DNS and HostAddress fail.


network/discovery/enode.go
```diff
@@ -51,9 +51,11 @@ func addAddresses(localNode *enode.LocalNode, hostAddr, hostDNS string) error {
 
 			return nil
 		}
+
+		return fmt.Errorf("no IPs found for host DNS: %s", hostDNS)
```
nkryuchkov: What do you think about reverting the condition to follow the happy path?

network/discovery/service.go
```diff
@@ -47,6 +48,14 @@ type Options struct {
 	TrimmedRecently     *ttl.Map[peer.ID, struct{}]
 }
 
+// Validate checks if the options are valid.
+func (o *Options) Validate() error {
+	if len(o.HostDNS) > 0 && len(o.HostAddress) > 0 {
+		return fmt.Errorf("HostDNS and HostAddress are mutually exclusive, please provide only one")
```
nkryuchkov: Can we reword the error as Go errors [should not be capitalized according to convention](https://go.dev/wiki/CodeReviewComments#error-strings) (although I agree `HostDNS` can't be lowercase, but this can trigger the linter)

y0sher (approved):

PR #2278, opened 2025-06-10
[SSV-15] ssvisgner: Re-Adding a Share After Removal Can Cause Slashing 

kchojn:
**Description**: Currently, when a share is removed from the protocol, its highest attestation is also removed from the local database,
and in case a local signer setup is being used, the local database is the only safeguard against double signing.

The `RemoveShare` method in `local_key_manager.go` will remove the highest proposal and the highest atttestation for the share
pub key from the local database:

```go
	if acc != nil {
		if err := km.RemoveHighestAttestation(pubKey); err != nil {
			return fmt.Errorf("could not remove highest attestation: %w", err)
		}
		if err := km.RemoveHighestProposal(pubKey); err != nil {
			return fmt.Errorf("could not remove highest proposal: %w", err)
		}
		if err := km.wallet.DeleteAccountByPublicKey(pubKeyHex); err != nil {
			return fmt.Errorf("could not delete share: %w", err)
		}
	}
```

If the same share is re-added again, the database will have no history of its signing history. This is not an issue if a remote signer is
used, because web3signer implements its own slashing protection. But if the same beacon object is requested to be signed, due to a
system glitch, for instance, then it is possible for a slashing event to happen.

**Recommendation**: Consider keeping the share history in the database, especially if the local setup is used


ssvsigner/ekm/local_key_manager.go
```diff
@@ -267,10 +267,25 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
...
+		if att, found, _ := km.RetrieveHighestAttestation(pubKey); found && att != nil {
+			hasHistory = true
+		}
+		if slot, found, _ := km.RetrieveHighestProposal(pubKey); found && slot > 0 {
+			hasHistory = true
+		}
+
+		// Only bump slashing protection if there's no existing history
+		// This preserves protection for re-added shares
+		if !hasHistory {
```
oleg-ssvlabs: I am not sure I fully understand this.

Before this change, when the share was added, the slashing protection was bumped (meaning it was recorded that this validator last time attested and proposed on current slot and epoch).

With this change, only validators that have no history (previously did not exist and were not removed) will have their slashing protection bumped.

Question: shouldn't we still bump it for both new and re-added validators?

iurii-ssv: > Question: shouldn't we still bump it for both new and re-added validators?

I think we "should still bump it for both new and re-added validators" just to be on the safe side of things just because we can do that (even though in practice it's unlikely to ever matter)

the only problematic case where not bumping for re-added validator would result into some issue would be if for some reason in DB we have some "stale" data (maybe Operator had older & newer backups, and recovered with the old one - that's unlikely, but possible)

iurii-ssv (commented):
Like [Oleg mentioned above](https://github.com/ssvlabs/ssv/pull/2278#discussion_r2140333495) I think we should bump slashing protection regardless of whether we have it in DB

ssvsigner/ekm/local_key_manager.go
kchojn: sounds good, fixed

oleg-ssvlabs (dismissed):
I am not 100% sure how the issue of re-adding the share after removal can cause the slashing, mainly because slashing protection is added as soon as share is added(and also newly added Validators start only the next epoch, so there is a delay), but probably there is still some possibility it to happen. Also, keeping the historical duty data wouldn't harm after removal the share anyway.

ssvsigner/ekm/local_key_manager.go
```diff
@@ -280,7 +282,7 @@ func (km *LocalKeyManager) AddShare(_ context.Context, encryptedPrivKey []byte,
 }
 
 // RemoveShare removes the share from the local wallet, clears the associated
-// slashing-protection records (highest attestation/proposal) for the given
+// slashing-protection records (attestation and proposal) for the given
 // public key, and returns an error on any storage issue.
 func (km *LocalKeyManager) RemoveShare(_ context.Context, pubKey phase0.BLSPubKey) error {
```
iurii-ssv: I guess this comment needs to be re-written such that we now aren't "clearing the associated slashing-protection records" ?

iurii-ssv (approved):
Left minor comment

ssvsigner/ekm/local_key_manager.go
kchojn: > I guess this comment needs to be re-written such that we now aren't "clearing the associated slashing-protection records" ?

Thanks, i missed it, fixed

iurii-ssv (dismissed):

moshe-blox:
@olegshmuelov good call

this likely becomes a bigger task that can't make it into the audit, however i think there's a simple way to do it:
- on validator removal:
  - copy slashing data from EKM to `slashing-data-archive/<validatorPubkey>` in our DB
- on validator registration:
  - populate slashing data with `max(currentEpoch, storedEpoch)` if they have a record

olegshmuelov:
> this likely becomes a bigger task that can't make it into the audit, however i think there's a simple way to do it with only 2 small changes:
> 
> * on validator removal:
>   
>   * copy slashing data from EKM to `slashing-data-archive/<validatorPubkey>` in our DB
> * on validator registration:
>   
>   * instead of `currentEpoch`, populate slashing data based on `max(currentEpoch, storedEpoch)` for source/target epochs, and likewise for proposals (with slots)
>   
>   wdyt?

I think this workaround adds complexity for an extremely unlikely scenario that doesn't justify the effort:

**Clarifying the unlikely scenario**: For slashing to occur, n*f+1 operators from the removed validator's committee need to have time drift issues AND the same validator needs to be re-registered with those same n*f+1 operators in the new committee. This is extremely unlikely (though maybe I'm mistaken about the slashing scenario?).

**Issues with the proposed approach:**

1. **Technical correction**: If we go with this solution, on validator registration we should populate SP with `max(currentEpoch, highestAttestationData)` and `max(currentSlot, highestProposerData)` - not `max(currentEpoch, storedEpoch)`. The bump SP will always be above the storedEpoch except for time drift issues - but maybe this is what you mean.
Note that when we register a validator, we already have minimal slashing protection by bumping the validator slashing protection with the current registration epoch.

2. **Doesn't solve web3signer**: This approach only addresses LocalKeyManager, but RemoteKeyManager has the same issue - both the local SP database and web3signer's built-in SP database would need
   similar workarounds, multiplying the complexity.

3. **Added complexity**: New archive system, cleanup logic, more test coverage

moshe-blox:
@olegshmuelov 

- time drift isn't required, a faulty majority hard fork as happened on Holesky can cause this issue, and if it happens, it's likely to happen with most operators (majority fork after all)
- yes, this doesn't solve it in web3signer, but given we check local SP before remote, it should block the slashable signatures
- cleanup logic isn't required

olegshmuelov:
Thanks for the clarifications @moshe-blox:
  1. **Good point about Holesky hard fork** - that's a more realistic scenario than time drift. Still relatively rare but more concerning.
  2. **Local SP blocking remote** - Yes, that helps, though it still leaves web3signer in an inconsistent state.
  3. **Cleanup** - You're right, maybe not immediately needed. I was thinking about long-term storage growth from archived validators.
  4. **Complexity** - Fair point about implementation vs test complexity.

Given this discussion, I think we have two reasonable paths:
  **Option A**: Implement @moshe-blox's workaround with proper documentation about limitations
  **Option B**: @kchojn's suggestion - revert and document the architectural limitation

Both options have merit. If we prefers Option A for audit compliance, let's at least ensure we document this as a temporary workaround with clear TODOs for the proper solution.

  @kchojn @moshe-blox WDYT?

ssvsigner/ekm/signer_storage.go
```diff
@@ -38,6 +39,11 @@ const (
 	accountsPath          = "accounts_%s"
 	highestAttPrefix      = prefix + "highest_att-"
 	highestProposalPrefix = prefix + "highest_prop-"
+
+	// slashingArchiveAttPrefix is the prefix for archived slashing protection data.
+	// It stores slashing protection data (highest attestation and proposal)
+	// keyed by validator public key. This is a temporary solution for audit finding SSV-15
+	slashingArchiveAttPrefix = prefix + "archive_att-"
```
iurii-ssv: Since this is responsible for both `attestations` and `proposals`, the `Att`/`_att` suffix is a bit misleading here, we probably should rename this a bit to account for that,

maybe `slashingArchiveAttPrefix` -> `slashingArchivePrefix` and `archive_att` -> `archive_sp`

eth/eventhandler/handlers.go
```diff
@@ -436,6 +450,13 @@ func (eh *EventHandler) handleClusterReactivated(txn basedb.Txn, event *contract
 			return nil, fmt.Errorf("could not bump slashing protection: %w", err)
 		}
 
+		// Apply any archived slashing protection data for this validator.
+		// TODO(SSV-15): This ensures slashing protection continuity across validator reactivation cycles
+		// as part of the temporary solution for audit finding SSV-15.
+		if err := eh.keyManager.ApplyArchivedSlashingProtection(share.ValidatorPubKey[:], phase0.BLSPubKey(share.SharePubKey)); err != nil {
+			eh.logger.Warn("could not apply archived slashing protection data during reactivation", zap.Error(err))
+		}
```
iurii-ssv: I think we need to restore archived values **before** we call `eh.keyManager.BumpSlashingProtection` (otherwise I'm not sure what restoring archived values will do for us) ?

Also, I think we should probably return error instead of logging it as Warn - since we don't want to risk slashing (why have this protection otherwise if it's "best effort only")

ssvsigner/ekm/slashing_protector_archive.go
```diff
@@ -0,0 +1,40 @@
...
+	// TODO(SSV-15): Remove this method once proper architectural solution is implemented.
+	ArchiveSlashingProtection(validatorPubKey []byte, sharePubKey []byte) error
+
+	// ApplyArchivedSlashingProtection applies archived slashing protection data for a validator.
+	// This method is part of the temporary solution for audit finding SSV-15.
+	// It should be called when adding a share to ensure slashing protection continuity
+	// across validator re-registration cycles. It uses maximum value logic to prevent regression.
+	//
+	// TODO(SSV-15): Remove this method once proper architectural solution is implemented.
+	ApplyArchivedSlashingProtection(validatorPubKey []byte, sharePubKey phase0.BLSPubKey) error
```
iurii-ssv: Why not make these 2 methods to accept txn ? Seems like we already have all the necessary code in place for that ...

ssvsigner/ekm/slashing_protector.go
```diff
@@ -115,19 +131,122 @@ func (sp *SlashingProtector) UpdateHighestProposal(pubKey phase0.BLSPubKey, slot
...
+	currentSlot := sp.signerStore.BeaconNetwork().EstimatedCurrentSlot()
+	currentEpoch := sp.signerStore.BeaconNetwork().EstimatedEpochAtSlot(currentSlot)
+
+	// Apply archived attestation data using max(current, archived) logic
+	if archive.HighestAttestation != nil {
+		// Get current slashing protection data for the share
+		currentAtt, foundCurrent, err := sp.signerStore.RetrieveHighestAttestationTxn(txn, sharePubKey[:])
+		if err != nil {
+			return fmt.Errorf("could not retrieve current highest attestation: %w", err)
+		}
```
iurii-ssv: Do we need to fetch "current slashing protection data" here, can't we just always use "max of current epoch and archived values" instead to keep it simple ?

I guess also applies to proposals below

ssvsigner/ekm/signer_storage.go
```diff
@@ -472,3 +493,124 @@ func (s *storage) decrypt(nonceCipherText []byte) ([]byte, error) {
...
+	}
+	if foundProp {
+		archive.HighestProposal = highestProp
+	}
+
+	// Check if we already have archived data and merge if needed
+	existing, found, err := s.getArchivedSlashingProtection(rw, validatorPubKey)
+	if err != nil {
+		return fmt.Errorf("could not check existing archive: %w", err)
+	}
```
iurii-ssv, line 557: Can't we just always overwrite the existing data in the storage in `ArchiveSlashingProtectionTxn` **func** ? Correct me if I'm wrong, but I think we can do that because slashing protection epochs are only supposed to grow and never decrease (meaning, the current value we are about to save to DB can't be lower than whatever was saved in the past)

Or maybe that's supposed to prevent signing in case of hard-forks that "revert current epoch to some past value" - we probably should leave a comment to clarify that (if that's the targeted use-case here) ?

eth/eventhandler/handlers.go
```diff
@@ -253,6 +253,13 @@ func (eh *EventHandler) handleShareCreation(
 			return nil, fmt.Errorf("could not add share encrypted key: %w", err)
 		}
 
+		// Apply any archived slashing protection data for this validator.
+		// TODO(SSV-15): This ensures slashing protection continuity across validator re-registration cycles
+		// as part of the temporary solution for audit finding SSV-15.
+		if err := eh.keyManager.ApplyArchivedSlashingProtection(share.ValidatorPubKey[:], phase0.BLSPubKey(share.SharePubKey)); err != nil {
+			eh.logger.Warn("could not apply archived slashing protection data", zap.Error(err))
+		}
```
iurii-ssv: Similar to https://github.com/ssvlabs/ssv/pull/2278#discussion_r2166400923 we probably should 
- do that before we add share
- and also abort (not just log Warn) here in case of error

ssvsigner/ekm/slashing_protector.go
kchojn: i think its needed and we always should use logic like `max(current_slashing_protection, archived_slashing_protection, current_time)`, never just `max(archived_slashing_protection, current_time)`, so we can be sure we neber regress protection lvl.
E.g.: 
- node synced to epoch 150, slashing protection = 150
- node crashes/restarts etc  
- current time = 100 (node was offline)
- archived data = 50

- with simplified approach: max(50, 100) = 100 
- with current approach: max(150, 50, 100) = 150

let me know if it makes sense

iurii-ssv: This comment of mine is similar to https://github.com/ssvlabs/ssv/pull/2278#discussion_r2166405913

I see what you are saying, I might be wrong (or maybe we just don't want make certain assumptions) but I think:
- the "current epoch" we use for slashing protection - `EstimatedCurrentSlot` - does not depend on Ethereum node sync state in any way - it's deterministically derived from current time itself (but I guess some sort of hard-fork can change what epoch-number the current time maps to ... I'm just not 100% sure how we'd want to handle a case like that)
- SSV node is probably waiting for Ethereum node(s) to sync before proceeding to perform any duties (where signing happens)

maybe @olegshmuelov @nkryuchkov @y0sher @moshe-blox could take a look at this

PR #2279, opened 2025-06-10
[SSV-20] ssvsigner: Server-Side Request Forgery via Web3signer_endpoint Configuration

kchojn:
**Description**: The `Web3SignerEndpoint` configuration parameter, sourced from CLI arguments or environment variables, is
validated by `validateConfig` using `url.ParseRequestURI `. This validation is insufficient as `url.ParseRequestURI` allows
various URL schemes and does not restrict target hosts (e.g., `localhost` , internal IPs, potentially other schemes if not filtered by
the HTTP client). The validated endpoint is then used by `setupWeb3SignerClient` to instantiate a `web3signer.Web3Signer`
client, which makes HTTPS requests. An attacker who can control the `WEB3SIGNER_ENDPOINT` value could specify an endpoint
pointing to internal network services (e.g., `http://localhost:xxxx , http://internal-service.local/api `). This could
allow the attacker to scan internal networks, access sensitive internal endpoints, or interact with internal services through the ssv-
signer application, as the underlying HTTP client ( `github.com/carlmjohnson/requests` wrapping `net/http` ) may permit such
connections.
Because the system is assumed to be communicating over mTLS with the client, the impact and likelihood is limited.

### DIFF
--- a/ssvsigner/cmd/internal/validation/url.go
+++ b/ssvsigner/cmd/internal/validation/url.go
@@ -1,0 +1,50 @@
+package validation
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+)
+
+// ValidateWeb3SignerEndpoint validates that the endpoint is safe from SSRF attacks.
+func ValidateWeb3SignerEndpoint(endpoint string) error {
+	u, err := url.ParseRequestURI(endpoint)
+	if err != nil {
+		return fmt.Errorf("invalid url format: %w", err)
+	}
+
+	// Only allow http/https schemes
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return fmt.Errorf("invalid url scheme %q: only http/https allowed", u.Scheme)
+	}
+
+	hostname := u.Hostname()
+	if hostname == "" {
+		return fmt.Errorf("missing hostname in url")
+	}
+
+	// Check for localhost/loopback
+	if strings.EqualFold(hostname, "localhost") ||
+		strings.HasPrefix(hostname, "127.") ||
+		hostname == "::1" {
+		return fmt.Errorf("localhost/loopback addresses are not allowed")
+	}
+
+	// Parse IP if it's an IP address
+	if ip := net.ParseIP(hostname); ip != nil {
+		if ip.IsUnspecified() {
+			return fmt.Errorf("invalid ip address type: %s", ip)
+		}
+
+		if ip.IsMulticast() {
+			return fmt.Errorf("invalid ip address type: %s", ip)
+		}
+
+		if ip.IsPrivate() || ip.IsLoopback() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast() {
+			return fmt.Errorf("private/local ip addresses are not allowed: %s", ip)
+		}
+	}
+
+	return nil
+}
--- a/ssvsigner/cmd/internal/validation/url.go
+++ b/ssvsigner/cmd/internal/validation/url.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"strings"
 )
 
 // ValidateWeb3SignerEndpoint validates that the endpoint is safe from SSRF attacks.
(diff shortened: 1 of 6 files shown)


ssvsigner/cmd/internal/validation/url.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Parse IP if it's an IP address
+	if ip := net.ParseIP(hostname); ip != nil {
+		if ip.IsUnspecified() {
+			return fmt.Errorf("invalid ip address type: %s", ip)
+		}
+
+		if ip.IsMulticast() {
+			return fmt.Errorf("invalid ip address type: %s", ip)
+		}
```
iurii-ssv: These error messages are the same even though the root cause is slightly different, perhaps we should add the root cause in the message as well, eg. `fmt.Errorf("invalid ip address type (unspecified): %s", ip)`

iurii-ssv (approved):
I guess the change does address the stated problem "best we can", 

the way I understand it however - I'm not 100% sure it's a "real" problem that warrants a solution (given that `Web3Signer struct` has a limited set of hard-coded methods it's unlikely to provide an attacker with any meaningful way to exploit it)

ssvsigner/cmd/internal/validation/url.go
```diff
@@ -0,0 +1,50 @@
...
+	if hostname == "" {
+		return fmt.Errorf("missing hostname in url")
+	}
+
+	// Check for localhost/loopback
+	if strings.EqualFold(hostname, "localhost") ||
+		strings.HasPrefix(hostname, "127.") ||
+		hostname == "::1" {
+		return fmt.Errorf("localhost/loopback addresses are not allowed")
+	}
```
iurii-ssv: So, that check (and some of the ones below) means we are not gonna allow for **Web3Signer** to run on the same host with **ssv-signer** ?

That seems quite restrictive/arbitrary requirement ... is it really worth it ?

iurii-ssv (dismissed):
LGTM as long as we are fine with the restrictions (not being able to run Web3Signer on the same host as ssv-signer) this PR imposes

oleg-ssvlabs (dismissed):

ssvsigner/cmd/internal/validation/url.go
```diff
@@ -0,0 +1,43 @@
...
+			return fmt.Errorf("invalid ip address type (unspecified): %s", ip)
+		}
+
+		if ip.IsMulticast() {
+			return fmt.Errorf("invalid ip address type (multicast): %s", ip)
+		}
+
+		// Only block non-loopback private IPs
+		if !ip.IsLoopback() && (ip.IsPrivate() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast()) {
+			return fmt.Errorf("private/local ip addresses are not allowed: %s", ip)
```
oleg-ssvlabs: is this error message correct?
`ip.IsLoopback() ` means local ip address (127.X.X.X), so this validation really checks only private network IP addresses excluding local.

Another question: does it makes sense to do that? I mean - allow local addresses, but disallow private network addresses? If there is another application runs on the same host which exposes endpoints with some sensitive data, this validation would allow to target it.

kchojn: the error was incorrect (i refactored slighty behavior and forgot to update it)
Anyways, I assume that usually web3signer will be run on another machine.
If that's the case, should we block localhost/loopback addresses as well for complete SSRF protection? 
We have three options:
- keep current behavior (allow localhost/block private)
- block everything local/private
- add a config flag like `--allow-localhost`

@y0sher wdyt?

y0sher: To be honest I'm not sure why not to put ssv-signer and web3signer on the same machine? for sure the node should be separated from both, but requiring 3 different setups for a remote-signer node seems a bit overkill to me. would love more opinions @nkryuchkov @moshe-blox  @iurii-ssv

kchojn: Im not against putting ssv-signer and web3signer on the same machine, I just gave the possibilities. But indeed, I would like to see more opinions

ssvsigner/cmd/internal/validation/url.go
```diff
@@ -0,0 +1,43 @@
...
+		if ip.IsUnspecified() {
+			return fmt.Errorf("invalid ip address type (unspecified): %s", ip)
+		}
+
+		if ip.IsMulticast() {
+			return fmt.Errorf("invalid ip address type (multicast): %s", ip)
+		}
+
+		// Only block non-loopback private IPs
+		if !ip.IsLoopback() && (ip.IsPrivate() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast()) {
```
oleg-ssvlabs: I think this condition is a little bit confusing.

Isn't this would give the same result:
`if ip.IsPrivate() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast()) { }`

I think these checks are mutually exclusive. I don't think, for example, IP can be both `loopback` and `private`.

ssvsigner/cmd/internal/validation/url.go
iurii-ssv: > To be honest I'm not sure why not to put ssv-signer and web3signer on the same machine? for sure the node should be separated from both, but requiring 3 different setups for a remote-signer node seems a bit overkill to me.

I think running all 3 as different processes on the same machine is also a valid setup (perhaps not as secure, but probably better than using local signer still),

I don't see how the issue described in the report can be meaningfully exploited (and it's also classified as "informational" in the report)

y0sher (dismissed):

oleg-ssvlabs (dismissed):

iurii-ssv (dismissed):

iurii-ssv (approved):

ssvsigner/cmd/ssv-signer/ssv-signer_test.go
```diff
@@ -13,16 +13,39 @@ import (
...
+		{
+			name:     "private IP blocked",
+			endpoint: "http://192.168.1.1:9000",
+			wantErr:  "invalid WEB3SIGNER_ENDPOINT: private/internal ip addresses are not allowed",
+		},
 	}
 
-	err := run(logger, cli)
-	require.ErrorContains(t, err, "invalid WEB3SIGNER_ENDPOINT format")
+	for _, tt := range tests {
```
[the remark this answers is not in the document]
y0sher: I don't think this is needed in recent go versions

PR #2262, opened 2025-06-04
fasttrack: proposer-duty: configurable proposer delay

iurii-ssv:
This PR implements https://github.com/ssvlabs/ssv/pull/2260 for **main** branch (so we can release proposer-delay feature sooner).

PR #2283, opened 2025-06-10
backport: (stage <- main) proposer-duty: configurable proposer delay

iurii-ssv:
This PR adds features (latest commits) from https://github.com/ssvlabs/ssv/pull/2262 so we can have the same functionality in **stage** and **main** branches.


iurii-ssv:
> @iurii-ssv isn't it better to merge whole main back? just asking as we did v2.3.2 release this week

@MatusKysel just for some context, 

typically it would be simpler to merge stuff we've originally merged into **main** back to **stage** as a whole batch (with `git merge`), but for this particular feature I don't think it's the right way to do it because:
- we had originally merged "most of it" into **stage** branch
- then we merged a PR that back-ports that ^ change into **main** (+ also added some minor things this PR contains)
- and so we've got slightly different functionality for ProposerDelay in **stage** and **main** now, it seems it's just simpler to merge this PR so the code is 100% the same (to the extent it can be, I mean) BEFORE we proceed with merging stuff from **main** to **stage**

because there are quite a bit of things in such a merge process it would be easy to miss something or forget which is "the latest state of things" - so I think if some piece of functionality already made it into both **stage** & **main** branches it's probably easier to align them first as much as possible

oleg-ssvlabs (approved):

PR #2285, opened 2025-06-11
[SSV-19] ssvsigner: Local Database Is a Single Point of Failure

kchojn:
**Description**: When the protocol operates with a local key manager, a local slashing database is used. This database functions as the
main protection against slashing events. In the case that the database goes down or gets erased for any reason, if there is no proper
backup, the slashing history will be erased. Continuing the operation after this point can cause slashing events.

**Recommendation**: Consider outlining that in user-facing documentation. Also consider various backup options to recover the
database if it goes down or get erased.

### Actions:
For now, we only update the docs

### DIFF
--- a/ssvsigner/README.md
+++ b/ssvsigner/README.md
@@ -386,11 +386,10 @@ database, cleaning keys in Web3Signer is required to ensure all shares are prope
 2. **Key Protection**: Store operator keys securely restricting access to them.
 3. **Access Control**: Limit access to the SSV-Signer and Web3Signer endpoints to only the necessary services.
 4. **Slashing Protection Database Backup**: The slashing protection database is critical to prevent slashing. Its loss or
-   corruption erases the slashing history, which can lead to double-signing and severe penalties if operation
-   continues. It is **crucial** for operators to implement a robust backup and recovery strategy for their databases
-   (PostgreSQL for Web3Signer, or the node's local database for local signing setups). Failure to maintain database
-   backups can lead to significant financial loss. Operators are responsible for their own database management and
-   protection.
+   corruption can lead to double-signing and severe penalties if operation continues. It is **crucial** for operators
+   to implement a robust backup and recovery strategy for their databases (PostgreSQL for Web3Signer, or the node's
+   local database for local signing setups). Failure to maintain database backups can lead to significant financial
+   loss. Operators are responsible for their own database management and protection.
 
 ## Performance Considerations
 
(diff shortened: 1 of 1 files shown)


oleg-ssvlabs (dismissed):

ssvsigner/README.md
```diff
@@ -385,6 +385,12 @@ database, cleaning keys in Web3Signer is required to ensure all shares are prope
 1. **Network Security**: Ensure communication between all components occurs over secure networks.
 2. **Key Protection**: Store operator keys securely restricting access to them.
 3. **Access Control**: Limit access to the SSV-Signer and Web3Signer endpoints to only the necessary services.
+4. **Slashing Protection Database Backup**: The slashing protection database is critical to prevent slashing. Its loss or
+   corruption erases the slashing history, which can lead to double-signing and severe penalties if operation
+   continues. It is **crucial** for operators to implement a robust backup and recovery strategy for their databases
+   (PostgreSQL for Web3Signer, or the node's local database for local signing setups). Failure to maintain database
+   backups can lead to significant financial loss. Operators are responsible for their own database management and
+   protection.
```
iurii-ssv: Maybe trim down this part (cause it's sort-of self repeating): `Its loss or
   corruption erases the slashing history, which can lead to double-signing and severe penalties if operation
   continues.` -> `Its loss or
   corruption can lead to double-signing and severe penalties if operation
   continues.`

nkryuchkov (dismissed):

iurii-ssv (commented):
TBH I'm not sure it's the correct approach to "force operators to backup DB" (and it's not even clear how to backup & restore parts related to slashing-protection specifically), 

in fact, sometimes Operator might need to delete his DB to bypass some bug and re-initialize everything from scratch - so that SSV node will re-populate the data in DB as necessary (including initializing slashing-protection values as needed, I think)

So, while there is a potential to get some "extra-safety" with such backups - I don't think it's really doable in practice ... @nkryuchkov @y0sher WDYT ?

---

But if we do want to make it Operator's responsibility:
- it seems to apply not only to remote-signer but also to local-signing setups, right ? If so, we maybe want to place this warning somewhere under `ssv/docs` (not in the`ssvsigner` folder)
- we probably need to provide some instructions on how to backup & recover specific parts related to slashing-protection

y0sher:
I think it also makes sense to ask @RaekwonIII to include in the published docs out of this repo.

oleg-ssvlabs (approved):

PR #1391, opened 2024-05-12
Alan Fixes

moshe-blox:

PR #2290, opened 2025-06-15
feat(eth/executionclinet): http fallback + adaptive batcher

kchojn:
TODO:

- [ ] clean up
- [ ] http_fallback mock + tests

### DIFF
--- a/eth/executionclient/adaptive_batch.go
+++ b/eth/executionclient/adaptive_batch.go
@@ -1,0 +1,116 @@
+package executionclient
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+const (
+	MinBatchSize     = 200
+	MaxBatchSize     = 2000
+	DefaultBatchSize = 500
+
+	batchIncreaseRatio = 1.5
+	batchDecreaseRatio = 0.7
+	successThreshold   = 5
+	latencyTarget      = 500 * time.Millisecond
+)
+
+// AdaptiveBatcher manages dynamic batch sizing.
+type AdaptiveBatcher struct {
+	currentSize    atomic.Uint64
+	successCount   atomic.Uint32
+	lastAdjustTime atomic.Int64
+	minSize        uint64
+	maxSize        uint64
+}
+
+// NewAdaptiveBatcher creates a new adaptive batcher.
+func NewAdaptiveBatcher(initial, min, max uint64) *AdaptiveBatcher {
+	if min == 0 {
+		min = MinBatchSize
+	}
+
+	if max == 0 {
+		max = MaxBatchSize
+	}
+
+	if initial == 0 {
+		initial = DefaultBatchSize
+	}
+
+	if initial < min {
+		initial = min
+	}
+
+	if initial > max {
+		initial = max
+	}
+
+	ab := &AdaptiveBatcher{
+		minSize: min,
+		maxSize: max,
+	}
+
+	ab.currentSize.Store(initial)
+	ab.lastAdjustTime.Store(time.Now().Unix())
+
+	return ab
+}
+
+// GetSize returns current batch size.
+func (ab *AdaptiveBatcher) GetSize() uint64 {
+	return ab.currentSize.Load()
+}
+
+// RecordSuccess records a successful batch operation.
+func (ab *AdaptiveBatcher) RecordSuccess(duration time.Duration) {
+	count := ab.successCount.Add(1)
+
+	if count >= successThreshold && duration < latencyTarget {
+		ab.tryIncrease()
+		ab.successCount.Store(0)
+	}
+}
+
+// RecordFailure records a failed batch operation.
+func (ab *AdaptiveBatcher) RecordFailure() {
+	ab.tryDecrease()
+	ab.successCount.Store(0)
+}
+
+// Reset resets the batcher to the default size.
+func (ab *AdaptiveBatcher) Reset() {
+	ab.currentSize.Store(DefaultBatchSize)
+	ab.successCount.Store(0)
+	ab.lastAdjustTime.Store(time.Now().Unix())
+}
+
+// tryIncrease attempts to increase batch size.
+func (ab *AdaptiveBatcher) tryIncrease() {
+	current := ab.currentSize.Load()
+	newSize := uint64(float64(current) * batchIncreaseRatio)
+
+	if newSize > ab.maxSize {
+		newSize = ab.maxSize
+	}
+	if newSize > current {
+		ab.currentSize.Store(newSize)
+		ab.lastAdjustTime.Store(time.Now().Unix())
+	}
+}
+
+// tryDecrease attempts to decrease batch size.
+func (ab *AdaptiveBatcher) tryDecrease() {
+	current := ab.currentSize.Load()
+	newSize := uint64(float64(current) * batchDecreaseRatio)
+
+	if newSize < ab.minSize {
+		newSize = ab.minSize
+	}
+
+	if newSize < current {
+		ab.currentSize.Store(newSize)
+		ab.lastAdjustTime.Store(time.Now().Unix())
+	}
+}
--- a/eth/executionclient/config.go
+++ b/eth/executionclient/config.go
@@ -11,4 +11,11 @@ type Options struct {
 	Addr                  string        `yaml:"ETH1Addr" env:"ETH_1_ADDR" env-required:"true" env-description:"Execution client WebSocket URL(s). Multiple clients are supported via semicolon-separated URLs (e.g. 'ws://localhost:8546;ws://localhost:8547')"`
 	ConnectionTimeout     time.Duration `yaml:"ETH1ConnectionTimeout" env:"ETH_1_CONNECTION_TIMEOUT" env-default:"10s" env-description:"Timeout for execution client connections"`
 	SyncDistanceTolerance uint64        `yaml:"ETH1SyncDistanceTolerance" env:"ETH_1_SYNC_DISTANCE_TOLERANCE" env-default:"5" env-description:"Maximum number of blocks behind head considered in-sync"`
+
+	// HTTP fallback configuration
+	HTTPFallbackAddr string `yaml:"ETH1HTTPFallbackAddr" env:"ETH_1_HTTP_FALLBACK_ADDR" env-description:"HTTP endpoint for fallback when WebSocket limits are exceeded. If empty, will attempt to convert WebSocket URL to HTTP"`
+
+	// Adaptive batching
+	MinBatchSize uint64 `yaml:"ETH1MinBatchSize" env:"ETH_1_MIN_BATCH_SIZE" env-default:"200" env-description:"Minimum batch size for log fetching"`
+	MaxBatchSize uint64 `yaml:"ETH1MaxBatchSize" env:"ETH_1_MAX_BATCH_SIZE" env-default:"2000" env-description:"Maximum batch size for log fetching"`
 }
--- a/eth/executionclient/defaults.go
+++ b/eth/executionclient/defaults.go
@@ -13,7 +13,5 @@ const (
 	// TODO ALAN: revert
 	DefaultHistoricalLogsBatchSize = 200
 	defaultLogBuf                  = 8 * 1024
-	maxReconnectionAttempts        = 5000
-	reconnectionBackoffFactor      = 2
 	healthCheckInterval            = 30 * time.Second
 )
--- a/eth/executionclient/errors.go
+++ b/eth/executionclient/errors.go
@@ -3,6 +3,7 @@ package executionclient
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/rpc"
 )
@@ -34,4 +35,9 @@ func isRPCQueryLimitError(err error) bool {
 	}
 
 	return false
+}
+
+// isWSReadLimitError checks if the given error is related to exceeding the read limit in a WebSocket connection.
+func isWSReadLimitError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "websocket: read limit exceeded")
 }
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -53,7 +53,8 @@ type ExecutionClient struct {
 	contractAddress ethcommon.Address
 
 	// optional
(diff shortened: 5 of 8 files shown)


eth/executionclient/defaults.go
```diff
@@ -11,9 +11,15 @@ const (
...
+	// Adaptive batcher constants
+	DefaultBatchSize    = 500
+	DefaultMinBatchSize = 200
+	DefaultMaxBatchSize = 2000
+
+	DefaultIncreaseRatio = 150 // 150% = 1.5x (50% increase)
+	DefaultDecreaseRatio = 70  // 70% = 0.7x (30% decrease)
+
+	DefaultHighLogsThreshold = 1000 // reduce batch when more than 1000 logs returned
 )
```
iurii-ssv: I think that `// TODO ALAN: revert` can also be removed now as it doesn't seem relevant anymore (originally added in https://github.com/ssvlabs/ssv/pull/1391/files#diff-e8d4fb87f6b7393d4b6dbe2fbe51c1739da2f99713ab1b6147eabc5a1b673de9R12-R14)

eth/executionclient/errors.go
```diff
@@ -41,3 +42,8 @@ func isRPCQueryLimitError(err error) bool {
 func isInterruptedError(err error) bool {
 	return errors.Is(err, ErrClosed) || errors.Is(err, rpc.ErrClientQuit) || errors.Is(err, context.Canceled)
 }
+
+// isWSReadLimitError checks if the given error is related to exceeding the read limit in a WebSocket connection.
+func isWSReadLimitError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "websocket: read limit exceeded")
+}
```
iurii-ssv: Can we use `errors.Is` with `ErrReadLimit` instead ?

Looks like it is a public (package-exported) error type:
```go
// ErrReadLimit is returned when reading a message that is larger than the
// read limit set for the connection.
var ErrReadLimit = errors.New("websocket: read limit exceeded")
```

eth/executionclient/multi_client.go
```diff
@@ -61,9 +61,14 @@ type MultiClient struct {
...
 	reconnectionMaxInterval     time.Duration
 	healthInvalidationInterval  time.Duration
-	logBatchSize                uint64
 	syncDistanceTolerance       uint64
 
+	// HTTP fallback configuration
+	httpFallbackAddr string
+
+	// adaptive batching
+	batcher *AdaptiveBatcher
```
iurii-ssv: It seems on `MultiClient` we don't actually need `AdaptiveBatcher` itself but rather only need `BatcherConfig` ? I think it's better to simplify that

eth/executionclient/fallback_client.go
```diff
@@ -0,0 +1,160 @@
...
+// FetchLogs retrieves logs for a specific block using HTTP.
+func (f *FallbackClient) FetchLogs(ctx context.Context, contractAddr ethcommon.Address, blockNum uint64) ([]ethtypes.Log, error) {
+	f.clientMu.RLock()
+	client := f.client
+	connected := f.connected.Load()
+	f.clientMu.RUnlock()
+
+	if !connected || client == nil {
+		return nil, fmt.Errorf("http fallback not connected")
+	}
```
iurii-ssv: I think we can simplify `!connected || client == nil` to just `!connected`

Probably applies to couple other places as well

eth/executionclient/fallback_client.go
```diff
@@ -0,0 +1,160 @@
...
+
+//go:generate go tool -modfile=../../tool.mod mockgen -package=executionclient -destination=./fallback_client_mock.go -source=./fallback_client.go
+
+// FallbackClientInterface defines the interface for fallback client operations.
+type FallbackClientInterface interface {
+	Connect(ctx context.Context, timeout time.Duration) error
+	FetchLogs(ctx context.Context, contractAddr ethcommon.Address, blockNum uint64) ([]ethtypes.Log, error)
+	FetchLogsViaReceipts(ctx context.Context, contractAddr ethcommon.Address, blockNum uint64) ([]ethtypes.Log, error)
+	Close() error
+}
```
iurii-ssv: Would be nice to expand `fallback client operations` part to explain why we need it

eth/executionclient/execution_client.go
```diff
@@ -300,6 +335,84 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
 	return nil, err
 }
 
+// handleSingleBlockOverflow attempts to retrieve logs from a single block that exceeds the standard read limits.
+// It first tries using an HTTP fallback if configured, and as a final measure, fetches logs via transaction receipts.
+// Returns the logs found or an error in case of failure.
+func (ec *ExecutionClient) handleSingleBlockOverflow(ctx context.Context, blockNum uint64) ([]ethtypes.Log, error) {
```
iurii-ssv: > // handleSingleBlockOverflow attempts to retrieve logs from a single block that exceeds the standard read limits.

Would be nice to clarify (or maybe provide some link) what are the circumstances when this might happen

eth/executionclient/execution_client.go
```diff
@@ -300,6 +335,84 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
...
+	ec.logger.Warn("single block exceeds read limit, attempting fallback",
+		fields.BlockNumber(blockNum))
```
iurii-ssv: `attempting fallback` -> `gonna attempt to fetch logs via fallback methods`

But it's also probably better let the caller log it IMO (since `handleSingleBlockOverflow` shouldn't make unnecessary assumptions about who/how calls it, and maybe even be named something like `fetchLogsForBlock` instead)

eth/executionclient/execution_client.go
```diff
@@ -300,6 +335,84 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
...
+
+				var logs []ethtypes.Log
+				for _, tx := range block.Transactions() {
+					receipt, err := ec.client.TransactionReceipt(ctx, tx.Hash())
+					if err != nil {
+						ec.logger.Debug("failed to fetch receipt",
+							fields.TxHash(tx.Hash()),
+							zap.Error(err))
+						continue
+					}
```
iurii-ssv: If we "continue" here we might miss some logs, not sure if that's critical or not - but it might be ... seems like we should just abort and return `error` instead (not just log it, certainly not as `Debug`) ?

eth/executionclient/execution_client.go
```diff
@@ -53,21 +53,25 @@ type ExecutionClient struct {
...
 
 	syncDistanceTolerance uint64
 	syncProgressFn        func(context.Context) (*ethereum.SyncProgress, error)
 
+	// adaptive batching
+	batcher *AdaptiveBatcher
+
 	// variables
 	client         *ethclient.Client
+	httpFallback   FallbackClientInterface
```
iurii-ssv: While `httpFallbackAddr` seems optional, I think `httpFallback` will always be configured ? There is no need to check for `httpFallback != nil` in a bunch of places then

eth/executionclient/config.go
```diff
@@ -10,5 +10,24 @@ import (
...
+	SyncDistanceTolerance uint64        `yaml:"ETH1SyncDistanceTolerance" env:"ETH_1_SYNC_DISTANCE_TOLERANCE" env-default:"5" env-description:"Maximum number of blocks that the execution client can lag behind the current head"`
+	HTTPFallbackAddr      string        `yaml:"ETH1HTTPFallbackAddr" env:"ETH_1_HTTP_FALLBACK_ADDR" env-description:"HTTP fallback address for the execution client"`
+
+	// Adaptive batch configuration
+	BatchSize         uint64 `yaml:"ETH1BatchSize" env:"ETH1BatchSize" env-default:"500" env-description:"Initial batch size for log fetching"`
+	MinBatchSize      uint64 `yaml:"ETH1MinBatchSize" env:"ETH1MinBatchSize" env-default:"200" env-description:"Minimum batch size for log fetching"`
+	MaxBatchSize      uint64 `yaml:"ETH1MaxBatchSize" env:"ETH1MaxBatchSize" env-default:"2000" env-description:"Maximum batch size for log fetching"`
+	HighLogsThreshold uint32 `yaml:"ETH1HighLogsThreshold" env:"ETH1HighLogsThreshold" env-default:"1000" env-description:"Threshold for reducing batch size when too many logs are returned"`
```
iurii-ssv: Do we need to expose these as configuration-settings to Operators ? I'm not sure I understand when Operator will need to tweak these (eg. why ever change `MinBatchSize`).

eth/executionclient/adaptive_batch.go
```diff
@@ -0,0 +1,201 @@
...
+		ab.increase()
+	} else if logsCount > int(ab.config.HighLogsThreshold) {
+		ab.decrease()
+	}
+}
+
+// RecordFailure records a failed operation and decreases batch size.
+func (ab *AdaptiveBatcher) RecordFailure() {
+	ab.decrease()
+}
```
iurii-ssv: I'm not sure `RecordResult` and `RecordFailure` provide a good/descriptive API for what's actually happening, maybe it would be more "transparent" to export `Increase` and `Decrease` ?

eth/executionclient/adaptive_batch.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+// AdaptiveBatcher dynamically adjusts batch sizes based on operation performance.
+// Thread-safe for concurrent use.
+type AdaptiveBatcher struct {
+	config BatcherConfig
+
+	size           atomic.Uint64
+	lastAdjustTime atomic.Int64
```
iurii-ssv: It seems `lastAdjustTime atomic.Int64` is never read, so do we even need it ?

eth/executionclient/execution_client.go
```diff
@@ -171,17 +190,22 @@ func (ec *ExecutionClient) fetchLogsInBatches(ctx context.Context, startBlock, e
 
 			results, err := ec.subdivideLogFetch(ctx, query)
 			if err != nil {
+				ec.batcher.RecordFailure()
 				errCh <- err
 				return
 			}
```
iurii-ssv: I think we also call `RecordFailure ` in `subdivideLogFetch` itself ... so it seems wrong to also call it here (also, without even checking error-type - because maybe it has nothing to do with batch-size)

Plus it looks like we abort log-fetching anyway - so what's the point of lowering batch-size in that case

eth/executionclient/execution_client.go
```diff
@@ -171,17 +190,22 @@ func (ec *ExecutionClient) fetchLogsInBatches(ctx context.Context, startBlock, e
...
+
 			ec.logger.Info("fetched registry events",
 				fields.FromBlock(fromBlock),
 				fields.ToBlock(toBlock),
 				zap.Uint64("target_block", endBlock),
 				zap.String("progress", fmt.Sprintf("%.2f%%", float64(toBlock-startBlock+1)/float64(endBlock-startBlock+1)*100)),
 				zap.Int("events", len(results)),
-				fields.Took(time.Since(start)),
+				fields.Took(duration),
 			)
```
iurii-ssv: I think it's cleaner to inline `time.Since(start)` because it's not super-important thing here compared to the rest of the logic (so it doesn't make much sense to allocate 2 extra code-lines to it)

but more importantly, I think inlining `time.Since(start)` is how we do it in most other places - so it's better to keep the same style for things like this (or change it everywhere ... but, again, I don't think it's worth it)

eth/executionclient/execution_client.go
```diff
@@ -300,6 +335,84 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
...
+			fn: func() ([]ethtypes.Log, error) {
+				if ec.httpFallback == nil {
+					return nil, fmt.Errorf("no http fallback configured")
+				}
+				if err := ec.httpFallback.Connect(ctx, ec.connectionTimeout); err != nil {
+					return nil, err
+				}
+				return ec.httpFallback.FetchLogs(ctx, ec.contractAddress, blockNum)
+			},
+		},
```
iurii-ssv: We are connecting here, but not in `http_receipts`, `ws_receipts` below, I guess this is because we assume `http_filter` will be called 1st (so the connection will be established already) ?

I don't think we should be making assumptions like this - it's better to connect explicitly somewhere else. Plus in case we couldn't connect in `http_filter` (later on in `http_receipts`, `ws_receipts` we shouldn't even be trying to call other methods on `httpFallback`).

Maybe we could even make `Connect` private (an implementation detail) such that it will be called every time any other method on `httpFallback` is called (it will be no-op except for the 1st time, essentially providing us with lazy-connect implementation)

eth/executionclient/adaptive_batch.go
```diff
@@ -0,0 +1,201 @@
...
+	InitialSize uint64
+	MinSize     uint64
+	MaxSize     uint64
+
+	// Scaling ratios (as percentages)
+	IncreaseRatio uint64 // e.g., 150 for 50% increase
+	DecreaseRatio uint64 // e.g., 70 for 30% decrease
+
+	// Log-based thresholds for adaptation
+	HighLogsThreshold uint32
```
iurii-ssv: I'm not sure this describes it best, could you maybe describe in that comment why this is needed ?

eth/executionclient/fallback_client.go
```diff
@@ -0,0 +1,160 @@
...
+// Close closes the HTTP client connection.
+func (f *FallbackClient) Close() error {
+	f.clientMu.Lock()
+	defer f.clientMu.Unlock()
+
+	if f.client != nil {
+		f.client.Close()
+		f.client = nil
+		f.connected.Store(false)
+	}
```
iurii-ssv: It's probably "cleaner" to check `f.connected.Load()` instead of `f.client != nil`,

Or we could check `f.client` vs `nil` everywhere instead and just get rid of `f.connected` entirely ?

iurii-ssv (commented):
1st pass, left some suggestions,

also, not sure if `FallbackClient` name is the best in terms of describing what this client is, maybe something like `LogFetcher` (or `LogClient`) would better reflect what's going on

kchojn:
@greptileai pls review it burh

PR #2292, opened 2025-06-16
db: iterator should propagate error to the caller (not log it)

iurii-ssv:
I think DB should let the caller decide how to proceed in case we couldn't iterate over ALL objects (in methods like "get all"), otherwise we might silently skip over some item(s) - which can be pretty bad in case we need to migrate a collection of objects, for example.

Originally spotted in https://github.com/ssvlabs/ssv/pull/2194#discussion_r2097661335


kchojn (approved):

nkryuchkov (dismissed):

oleg-ssvlabs (dismissed):
lgtm!

PR #2294, opened 2025-06-16
feat(e2e): add `SSV-Signer` slashing protection tests

olegshmuelov:
## Summary

  Add comprehensive E2E slashing protection tests for `SSV-Signer` with triple-layer protection validation.

  ### What's Added

  - **Complete E2E Test Suite**: New test infrastructure for `SSV-Signer` slashing protection
  - **Triple Protection Testing**: Validates `LocalKeyManager` + `RemoteKeyManager` + `Web3Signer` protection layers
  - **Comprehensive Coverage**: Attestation and block proposal slashing tests with container restart scenarios
  - **Production-Grade Setup**: Uses `Web3Signer 25.4.1` with `PostgreSQL` and `mainnet` beacon config
  - **Container Management**: Docker-based test environment with `testcontainers`

  ### Test Coverage

  - ✅ Attestation slashing protection (double vote, surrounding vote)
  - ✅ Block proposal slashing protection (double proposal, lower slot)
  - ✅ Concurrent signing safety
  - ✅ Container restart persistence
  - ✅ Triple protection validation

  ## Test plan

  - [ ] Run `go test ./ssvsigner/e2e/signing/` to verify all tests pass


ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
+# SSV-Signer E2E Tests
+
+End-to-end testing suite for SSV-Signer slashing protection functionality.
+
+## Overview
+
+Tests the complete signing flow with triple-layer slashing protection:
+**SSV Node** → **LocalKeyManager** + **RemoteKeyManager** → **SSV-Signer** → **Web3Signer** → **PostgreSQL**
```
iurii-ssv: I'm not sure what "triple-layer" protection refers to, maybe this should be clarified or phrased differently ?

There is a mention below, but I'm still not 100% understand what it refers to (and it's probably better to define it sooner anyway):

> **Triple Protection**: Tests LocalKeyManager, RemoteKeyManager, and Web3Signer slashing protection

ssvsigner/e2e/testdata/migrations/V00002__removeUniqueConstraints.sql
```diff
@@ -0,0 +1,21 @@
+DO
```
iurii-ssv, line 1: Perhaps `*.sql` file names should use the same **camel_case** naming convention ? 

Or are these migrations just copy-pasted from somewhere (as opposed us writing these from scratch) ? If so, I think README.md in ssvsigner/e2e should mention that and have a link included so we can update it in the future if needed (maybe we should also mention that if we need to bump Web3Signer version used we also might have to add some migrations if new ones were added for that newer Web3Signer version)

ssvsigner/e2e/testenv/containers.go
```diff
@@ -0,0 +1,235 @@
...
+
+	_ "github.com/lib/pq"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/network"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// startContainers starts all Docker containers in the correct order
+func (env *TestEnvironment) startContainers() error {
+	var err error
```
iurii-ssv: Nit: there probably is no need to define `err` here

ssvsigner/e2e/suite.go
```diff
@@ -0,0 +1,239 @@
...
+	switch domainType {
+	case spectypes.DomainAttester:
+		errMsg = "slashable attestation"
+
+	case spectypes.DomainProposer:
+		errMsg = "slashable proposal"
+
+	default:
+		return
+	}
```
iurii-ssv: It's probably better to `panic(fmt.Sprintf("unsupported domain type: %s", domainType))` here,

otherwise we might "think test works" while, in fact, it doesn't even check anything

ssvsigner/e2e/testenv/environment.go
```diff
@@ -0,0 +1,403 @@
...
+	ssvSignerContainer  testcontainers.Container
+
+	// Essential clients
+	postgresDB       *sql.DB
+	ssvSignerClient  *ssvsigner.Client
+	web3SignerClient *web3signer.Web3Signer
+	localKeyManager  *ekm.LocalKeyManager
+	remoteKeyManager *ekm.RemoteKeyManager
+	localDB          basedb.Database
+	remoteDB         basedb.Database
```
iurii-ssv: Nit: `Essential clients` -> `Essential components`

Also, I would rename `postgresDB` -> `web3SignerPostgresDB` and other postgres mentions as well to clarify what exactly it represents.

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Run with longer timeout for slow systems
+go test ./signing/ -timeout=300s
+```
+
+## Key Features
+
+- **Triple Protection**: Tests LocalKeyManager, RemoteKeyManager, and Web3Signer slashing protection
+- **Real Network Config**: Uses mainnet beacon parameters and fork schedule
+- **Container Isolation**: Fresh environment per test with persistent volumes
```
iurii-ssv: Might want to expand this a bit to explain how exactly persistent volumes are utilized/help in these tests

ssvsigner/e2e/testenv/environment.go
```diff
@@ -0,0 +1,403 @@
...
+		env.postgresDB.Close()
+	}
+
+	if env.localDB != nil {
+		env.localDB.Close()
+	}
+
+	if env.remoteDB != nil {
+		env.remoteDB.Close()
+	}
```
iurii-ssv: I see a bunch of nil-checks like this for components that I assume **must** be initialized (or tests just won't start), so I'm not sure whether/why we need these - can maybe remove ?

I guess it applies to a bunch of different fields (eg. `localKeyManagerPath`)

ssvsigner/web3signer/block_data.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/attestantio/go-eth2-client/spec"
+	"github.com/attestantio/go-eth2-client/spec/capella"
+	"github.com/attestantio/go-eth2-client/spec/deneb"
+	"github.com/attestantio/go-eth2-client/spec/electra"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
+
+// ConvertBlockToBeaconBlockData converts various block types to Web3Signer BeaconBlockData format
+func ConvertBlockToBeaconBlockData(obj interface{}) (*BeaconBlockData, error) {
+	var ret *BeaconBlockData
```
iurii-ssv: Nit: it's probably simpler do just inline `return` statements under each `case` clause instead of defining `var ret *BeaconBlockData` here

the general idea being "the sooner we **return** from a func the simpler it is to reason about"

ssvsigner/e2e/signing/attestation_test.go
```diff
@@ -0,0 +1,601 @@
...
+				validLocalRoot = localRoots[i]
+			} else {
+				// All successful signatures should be identical
+				s.Require().Equal(validLocalSig, localSignatures[i])
+				s.Require().Equal(validLocalRoot, localRoots[i])
+			}
+		}
+	}
+
+	s.Require().Equal(successCount, 1)
```
iurii-ssv: > // Verify that all successful operations produced identical results

Maybe it's more precise to state it like `// Verify that only 1 operation was successful` ? We might also want to check that error/erorr-message) is as expected (duplicate attempt to sign or something)

That also means we probably don't need to have that `if validLocalSig == nil` above.

Btw, I think this applies to 2 places in this code.

ssvsigner/e2e/signing/proposer_test.go
```diff
@@ -0,0 +1,539 @@
...
+				validLocalRoot = localRoots[i]
+			} else {
+				// All successful signatures should be identical
+				s.Require().Equal(validLocalSig, localSignatures[i])
+				s.Require().Equal(validLocalRoot, localRoots[i])
+			}
+		}
+	}
+
+	s.Require().Equal(successCount, 1)
```
iurii-ssv: Nit: typically expected result is the 1st argument, while actual result is 2nd argument (but it doesn't really matter I guess, jfyi)

ssvsigner/e2e/signing/attestation_test.go
```diff
@@ -0,0 +1,601 @@
...
+
+	validatorKeyPair := s.RequireAddValidator(ctx)
+
+	s.T().Logf("🔀 Phase 1: Testing concurrent attestation signing")
+	testEpoch := testCurrentEpoch + 10
+	testSlot := s.GetEnv().GetMockBeacon().GetEpochFirstSlot(testEpoch) + 5
+	attestationData := common.NewTestAttestationData(testEpoch-1, testEpoch, testSlot)
+	attestationData.BeaconBlockRoot = phase0.Root{0x42} // Distinguish this concurrent test attestation
+
+	s.T().Logf("🔀 Phase 2: Testing LocalKeyManager concurrent protection")
```
iurii-ssv: I'm not sure we need a separate `Phase 1` here since it just does some basic setup without doing any actual signing ? Just to keep things consistent.

ssvsigner/e2e/signing/proposer_test.go
```diff
@@ -0,0 +1,539 @@
...
+
+	s.Require().Empty(web3SignerErrors)
+
+	// Verify all successful signatures are identical (deterministic signing)
+	for _, web3sig := range web3SignerSignatures {
+		s.Require().Equal(validLocalSig, web3sig)
+	}
+
+	// Web3Signer allow same slot/parent_root
+	s.Require().Equal(len(web3SignerSignatures), numGoroutines)
```
iurii-ssv: `Web3Signer allow same` -> `Web3Signer allows same`

and similar checks for local/remote signers above (eg. `s.Require().Equal(remoteSuccessCount, 1)`) might use similar comment, emphasizing they allow only successful 1 signature by design

ssvsigner/e2e/suite.go
```diff
@@ -0,0 +1,239 @@
...
+
+	err = remoteKeyManager.AddShare(ctx, validatorKeyPair.EncryptedShare, validatorKeyPair.BLSPubKey)
+	s.Require().NoError(err, "Failed to add share to remote key manager")
+
+	err = localKeyManager.AddShare(ctx, validatorKeyPair.EncryptedShare, validatorKeyPair.BLSPubKey)
+	s.Require().NoError(err, "Failed to add share to local key manager")
+
+	validators, err := ssvSignerClient.ListValidators(ctx)
+	s.Require().NoError(err, "Failed to list validators")
+	s.Require().True(slices.Contains(validators, validatorKeyPair.BLSPubKey))
```
iurii-ssv: I would leave a comment here clarifying what role `ssvSignerClient. ListValidators` plays here, something like `Verifying SSV-Signer operational status` (similar to how it's done in tests)

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+go test ./signing/ -run TestBlockSlashing
+
+# Run with longer timeout for slow systems
+go test ./signing/ -timeout=300s
+```
+
+## Key Features
+
+- **Triple Protection**: Tests LocalKeyManager, RemoteKeyManager, and Web3Signer slashing protection
+- **Real Network Config**: Uses mainnet beacon parameters and fork schedule
```
iurii-ssv: > - **Real Network Config**: Uses mainnet beacon parameters and fork schedule

Are **mainnet** parameters utilized in some specific way ? I'm not sure why it deserves a separate mention (if it does, maybe we want expand on it, if it doesn't - maybe just not mention it at all)

I guess also applies to this from above:

> - **Mainnet beacon configuration** for realistic testing

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Tests the complete signing flow with triple-layer slashing protection:
+**SSV Node** → **LocalKeyManager** + **RemoteKeyManager** → **SSV-Signer** → **Web3Signer** → **PostgreSQL**
+
+## Test Environment
+
+- **Web3Signer 25.4.1** with PostgreSQL slashing protection database
+- **Mainnet beacon configuration** for realistic testing
+- **Docker containers** managed via testcontainers-go
+- **Shared test suite** for easy test reuse across signing domains
```
iurii-ssv: > - **Shared test suite** for easy test reuse across signing domains

This is an implementation detail that's not really relevant to anybody who just wants to run these tests, I would omit it

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+# Run all signing tests
+go test ./signing/
+
+# Run specific domain tests
+go test ./signing/ -run TestAttestationSlashing
+go test ./signing/ -run TestBlockSlashing
+
+# Run with longer timeout for slow systems
+go test ./signing/ -timeout=300s
+```
```
iurii-ssv: Running `go test ./signing/` seems to fail for me with the following error:
```
--- FAIL: TestAttestationSlashing (73.34s)
    --- FAIL: TestAttestationSlashing/TestAttestationDoubleVote (48.67s)
        suite.go:36:
            	Error Trace:	/Users/iurii/work/ssv/ssvsigner/e2e/suite.go:36
            	            				/Users/iurii/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:192
            	            	failed to start containers: failed to start SSV-Signer: failed to start ssv-signer: create container: Error response from daemon: pull access denied for ssv-signer-test, repository does not exist or may require 'docker login'
            	Test:       	TestAttestationSlashing/TestAttestationDoubleVote
            	Messages:   	Failed to start test environment
```

Do I need to pre-build a docker image based on `ssv-signer-test.dockerfile` file ? If so that probably should be documented here, and it maybe even worth having a Makefile in `ssvsigner/e2e` (or even `ssvsigner`) that defines a single command (eg. `make run`) that could do all the necessary setup for me

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+- **Real Network Config**: Uses mainnet beacon parameters and fork schedule
+- **Container Isolation**: Fresh environment per test with persistent volumes
+- **Comprehensive Coverage**: Slashing protection, concurrency, and restart scenarios
+- **Easy Extension**: Shared test suite enables adding new signing domains
+
+## Requirements
+
+- Docker (for testcontainers)
+- Go 1.21+
+- ~2GB available memory for containers
```
iurii-ssv: I think requirements section belongs under `Test Environment` or `Running Tests` (maybe even all 3 should be combined into 1 section) just because if I want to run these tests I need to know right away that I must have Docker installed

ssvsigner/e2e/README.md
iurii-ssv: Also, `ssv-signer-test.dockerfile` file is very similar to `ssvsigner/Dockerfile` - I'm not sure why we need to have a separate one ... this probably deserves a mention in the docs (or maybe we could merge them somehow instead) ?

ssvsigner/e2e/suite.go
```diff
@@ -0,0 +1,239 @@
...
+	resp, err := web3SignerClient.Sign(ctx, pubKey, req)
+	if err != nil {
+		return nil, phase0.Root{}, err
+	}
+
+	return resp.Signature[:], root, nil
+}
+
+// RequireAddValidator sets up a validator in all key managers and verifies it's properly registered
+func (s *E2ETestSuite) RequireAddValidator(ctx context.Context) *common.ValidatorKeyPair {
```
iurii-ssv: maybe `RequireAddValidator`-> `AddValidator` 

because it's kind of self-explanatory that if this setup-step fails the test also would fail (this is what require essentially means)

iurii-ssv (commented):
Reads very nicely, great job!

Left mostly minor suggestions (some nits may be a matter of preference, feel free to skip).

ssvsigner/e2e/README.md
olegshmuelov: Thanks for the thorough review @iurii-ssv! I've addressed the Docker setup issues:
  - Removed the redundant test Dockerfile and now using the main `ssvsigner/Dockerfile` (best practice)
  - Added a minimal Makefile at `ssvsigner/` level with `make test-e2e` target
  - Tests now automatically build the required Docker image before running
  - Added `-count=1` to ensure E2E tests always run fresh (no caching)
  - Moved requirements section to be more prominent in the README

  The setup is now much simpler - just run `make test-e2e` from the ssvsigner directory and everything works. Let me know if you have any other concerns!

iurii-ssv: Seems to be running nicely for me now, cool!

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,84 @@
...
+If you prefer to run tests manually:
+
+```bash
+# Build Docker image (from repository root)
+docker build -f ssvsigner/Dockerfile -t ssv-signer:latest .
+
+# Run tests (from ssvsigner/e2e directory)
+go test ./signing/                                    # All tests
+go test ./signing/ -run TestAttestationSlashing      # Attestation tests only
+go test ./signing/ -run TestBlockSlashing            # Proposer tests only
```
iurii-ssv: Just noticed there is an extra unnecessary space here, feel free to borrow from this: https://github.com/ssvlabs/ssv/compare/e2e/ssv-signer...iurii-suggestions/e2e/ssv-signer?expand=1

ssvsigner/e2e/README.md
olegshmuelov: I kept this in Key Features because the E2E suite is specifically designed for extensibility. Currently we only test attestation and proposer signing
  domains, but the shared suite architecture makes it easy to add other signing domains (sync committee, voluntary exits, etc.) in the future. This extensibility is a key architectural decision
  worth highlighting for future contributors.

ssvsigner/e2e/testdata/migrations/V00002__removeUniqueConstraints.sql
olegshmuelov: addressed here https://github.com/ssvlabs/ssv/pull/2294/commits/2332b42261e17ae101480b13720a663fb4cbca0e

ssvsigner/e2e/testenv/environment.go
olegshmuelov: Regarding the nil checks in cleanup - I kept them because setup can fail partway through (like container issues), leaving some DBs uninitialized. Better to have robust cleanup that never panics
  than to assume everything was set up successfully.

ssvsigner/e2e/README.md
olegshmuelov: addressed here https://github.com/ssvlabs/ssv/pull/2294/commits/2332b42261e17ae101480b13720a663fb4cbca0e

ssvsigner/e2e/README.md
olegshmuelov: addressed here https://github.com/ssvlabs/ssv/pull/2294/commits/2332b42261e17ae101480b13720a663fb4cbca0e

ssvsigner/e2e/README.md
olegshmuelov: addressed here https://github.com/ssvlabs/ssv/pull/2294/commits/2332b42261e17ae101480b13720a663fb4cbca0e

ssvsigner/e2e/README.md
olegshmuelov: addressed here https://github.com/ssvlabs/ssv/pull/2294/commits/2332b42261e17ae101480b13720a663fb4cbca0e

ssvsigner/e2e/testenv/containers.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/suite.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/testenv/environment.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/web3signer/block_data.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/signing/attestation_test.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/signing/proposer_test.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/signing/attestation_test.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/signing/proposer_test.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/suite.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

ssvsigner/e2e/suite.go
olegshmuelov: addressed here 9a81eeddb537f46e0872eaf25dc950af3efa579a

nkryuchkov (dismissed):
Impressive work!

ssvsigner/e2e/testenv/environment.go
iurii-ssv: Agree that explicit error-message is better than some rogue **panic**, however in the case of "partway fail" wouldn't we spot such a failure sooner (eg. by getting an error from `startContainers` **func** or similar) ?

I guess what I'm trying to say is that if these nil-checks never actually happen then maybe it's not worth complicating the code for.

ssvsigner/e2e/README.md
iurii-ssv: The new phrasing is more appropriate 👍 

> **Easy Extension**: Shared test suite enables adding new signing domains

iurii-ssv (approved):

ssvsigner/e2e/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Test Structure
+
+```
+e2e/
+├── suite.go                    # Shared E2ETestSuite
+├── common/                     # Utilities (keys, beacon objects, mocks)
+├── signing/                    # Test implementations
+│   ├── attestation_test.go     # Attestation slashing tests
+│   └── proposer_test.go        # Block proposal slashing tests
```
y0sher: should we add other duties? sync committee, etc..

olegshmuelov: This PR: E2E tests for SSV-Signer slashing protection (attester + proposer).
Other signing domains (sync committee, etc.) are non-slashable but now easy to add with the established infrastructure.
Current scope covers all slashable operations.

ssvsigner/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: docker-build
+docker-build:
+	cd .. && docker build -f ssvsigner/Dockerfile -t ssv-signer:latest .
+
+.PHONY: test-e2e
+test-e2e: docker-build
+	cd e2e && go test -count=1 -v ./signing/ -timeout=300s
```
oleg-ssvlabs: During the initial launch on my machine, the test timed out. Also, looking into the pipeline execution time [here](https://github.com/ssvlabs/ssv/actions/runs/15765375399/job/44440637973?pr=2294), it took around 5 minutes.

I think it’s better to increase the timeout (maybe double it), so the tests do not occasionally fail.

oleg-ssvlabs (changes requested):
The implementation looks awesome - very clean and easy to read.

One question:

Have you considered also testing the web layer of the SSV Signer?

So far, the attestation and proposer tests mostly cover the remote, local, and web3 signers. We could potentially extend this to also test the HTTP handlers for all SSV Signer endpoints. That would make it even more E2E 🙂 Wdyt?

oleg-ssvlabs (dismissed):

iurii-ssv (dismissed):

ssvsigner/e2e/signing/attestation_test.go
y0sher, line 1: wdyt about adding a test that checks how the remote/local signer works with parallel signing?

olegshmuelov: c9390970737af7842d1e74e7d7f0ade952664a0c

ssvsigner/e2e/signing/attestation_test.go
```diff
@@ -0,0 +1,615 @@
...
+
+		consistentSignatures++
+	}
+
+	s.T().Logf("✅ Signature consistency validated: %d/%d validators have identical signatures across all key managers",
+		consistentSignatures, numValidators)
+
+	s.Require().Equal(numValidators, consistentSignatures,
+		"Expected all %d validators to have consistent signatures across key managers, but only %d were consistent",
+		numValidators, consistentSignatures)
```
iurii-ssv: Perhaps it's better to swap these lines around ? So that we won't print "... success ..." in case `Require()` fails

ssvsigner/e2e/signing/attestation_test.go
```diff
@@ -0,0 +1,615 @@
...
+	testSlot := s.GetEnv().GetMockBeacon().GetEpochFirstSlot(testEpoch) + 5
+	domain, err := s.CalculateDomain(spectypes.DomainAttester, testEpoch)
+	s.Require().NoError(err)
+
+	attestations := make([]*phase0.AttestationData, numValidators)
+	for i := 0; i < numValidators; i++ {
+		attestations[i] = common.NewTestAttestationData(testEpoch-1, testEpoch, testSlot)
+		attestations[i].Index = phase0.CommitteeIndex(i % 64)
+		attestations[i].BeaconBlockRoot = phase0.Root{byte(i % 256)}
+	}
```
iurii-ssv, line 509: Nit: It would probably be simpler overall if we used a separate `numAttestations` (so it's not equal to `numValidators`), and maybe we could produce `numValidators x numAttestations` signatures in total

iurii-ssv (dismissed):
Left minor suggestions

ssvsigner/e2e/signing/attestation_test.go
olegshmuelov: The current 1:1 mapping is intentional - this test specifically verifies that parallel signing works correctly. Adding multiple attestations per validator would increase complexity without adding value for this particular test case.

iurii-ssv (approved):

iurii-ssv (dismissed):

ssvsigner/e2e/suite.go
```diff
@@ -0,0 +1,311 @@
...
+// E2ETestSuite provides common test functionality for all signing domains
+type E2ETestSuite struct {
+	suite.Suite
+	env *testenv.TestEnvironment
+	ctx context.Context
+}
+
+// SetupTest initializes a fresh test environment for each test
+func (s *E2ETestSuite) SetupTest() {
+	s.ctx = context.Background()
```
oleg-ssvlabs: nit: you could potentially use `s.T().Context()` here

ssvsigner/e2e/suite.go
```diff
@@ -0,0 +1,311 @@
...
+	err = remoteKeyManager.AddShare(ctx, nil, validatorKeyPair.EncryptedShare, validatorKeyPair.BLSPubKey)
+	s.Require().NoError(err, "Failed to add share to remote key manager")
+
+	err = localKeyManager.AddShare(ctx, nil, validatorKeyPair.EncryptedShare, validatorKeyPair.BLSPubKey)
+	s.Require().NoError(err, "Failed to add share to local key manager")
+
+	// Verify SSV-Signer operational status and that validator was successfully added
+	validators, err := ssvSignerClient.ListValidators(ctx)
+	s.Require().NoError(err, "Failed to list validators")
+	s.Require().True(slices.Contains(validators, validatorKeyPair.BLSPubKey))
```
oleg-ssvlabs: nit - you could use `Contains` method like this:

`s.Require().Contains(validators, validatorKeyPair.BLSPubKey)`

ssvsigner/e2e/testenv/migrations.go
```diff
@@ -0,0 +1,221 @@
...
+		if c >= '0' && c <= '9' {
+			rank = rank*10 + int(c-'0')
+		}
+	}
+
+	return rank
+}
+
+// getMigrationStatus returns the current migration status
+func (env *TestEnvironment) GetMigrationStatus() ([]string, error) {
```
oleg-ssvlabs: I don't think this method is used. Is it needed?

ssvsigner/e2e/testenv/migrations.go
oleg-ssvlabs, line 1: May I ask what these migrations are?

If Web3Signer depends on Postgres, I assume it will apply all necessary migrations on a startup?

olegshmuelov: right,
the migrations are for the web3signer slashing protection DB
please sse
https://github.com/ssvlabs/ssv/blob/7111e3233e3a8e3ab5abca2a40c8dbe911ffa556/ssvsigner/README.md?plain=1#L36

iurii-ssv (approved):

ssvsigner/e2e/testenv/migrations.go
oleg-ssvlabs: thank you, I see now.

I thought `web3signer` itself should take care of its database and run the migrations during the startup. It's slightly weird they require to run DB migrations manually before launching the `web3signer` itself. 

Thank you for explaining.

oleg-ssvlabs (approved):

moshe-blox (commented):
LGTM

well done @olegshmuelov

PR #2277, opened 2025-06-09
ssvsigner: fix unit tests

nkryuchkov:
ssv-signer unit tests have been broken for a while because we don't have lint/test CI for it

PR #2295, opened 2025-06-16
ssvsigner: fix linter & UT, add CI checks [main]

nkryuchkov:
Port CI setup from https://github.com/ssvlabs/ssv/pull/2277 as well as some linter/test fixes.

Also, some tests in `ssvsigner/cmd/ssv-signer` are broken on `main` branch


.github/workflows/ssvsigner-lint.yml
```diff
@@ -0,0 +1,27 @@
+name: SSV-Signer lint
+
+on:
+  push:
+    branches:
+      - "**"
```
MatusKysel: https://github.com/ssvlabs/ssv/commit/d87a2620dd18cb5ab68c02835fc7cb749a007cd0 align it to this, so we are running it on PR from external forks

iurii-ssv: This also applies to all other pipelines (not just linter), right ?

ssvsigner/tls/tls.go
```diff
@@ -345,7 +345,7 @@ func verifyClientCertificate(rawCerts [][]byte, trustedFingerprints map[string]s
 // loadPasswordFromFile loads a password from a file.
 // The password is expected to be the first line of the file.
 func loadPasswordFromFile(filePath string) (string, error) {
-	data, err := os.ReadFile(filePath)
+	data, err := os.ReadFile(filePath) // #nosec G304
```
olegshmuelov: Since these are configuration file paths (not user input during runtime), the security risk is lower, but it's still good practice to clean the paths.
```go
filepath.Clean(filePath)
```
should be applied this pattern to all 4

Makefile
```diff
@@ -39,11 +40,28 @@ lint:
...
 	@go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./...
 
+.PHONY: ssvsigner-test
+ssvsigner-test:
+	@echo "Running ssv-signer unit tests"
+	@cd ssvsigner && go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 ./... && cd ..
+
+.PHONY: unit-test-with-signer
+unit-test-with-signer: unit-test ssvsigner-test
+	@echo "Running all tests including ssv-signer"
```
olegshmuelov: IMO this target is redundant. Instead, modify the `unit-test` target to include ssvsigner tests:
 ```makefile
  .PHONY: unit-test
  unit-test:
        @echo "Running unit tests"
        @go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 `go list ./... | grep -ve "spectest\|integration\|ssv/scripts/"`
        @echo "Running ssv-signer unit tests"
        @$(MAKE) ssvsigner-test
```
Then remove this target entirely.

File: Makefile [hunk 4]
```diff
@@ -39,11 +40,28 @@ lint:
...
+
```
iurii-ssv: Maybe instead of this command we should extend existing `make unit-test` to always run ssv-signer tests ? I don't see why not, plus this will allow for re-using `make docker-unit-tests` as well,

and we could do similar for **linter** (I'm not sure why we'd not want to always run both `make lint` and `make ssvsigner-lint` as a single command)

I guess if we want to have separate Github pipelines for **ssv** and **ssv-signer** code we could arrange for that as well, I just think it'd be better if `make lint` and `make docker-unit-test` could execute all the checks (for **ssv** and **ssv-signer** code) when running these locally

iurii-ssv (approved):
LGTM but maybe we don't need to merge this into **main** branch separately, and can rather wait for this change to arrive with the rest of changes from **stage** branch ?

Makefile
```diff
@@ -39,11 +40,28 @@ lint:
...
 .PHONY: full-test
 full-test:
 	@echo "Running all tests"
 	@go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./...
 
+.PHONY: ssvsigner-test
+ssvsigner-test:
+	@echo "Running ssv-signer unit tests"
+	@cd ssvsigner && go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 ./... && cd ..
```
olegshmuelov: The `&& cd ..` is unnecessary, see previous comment

Makefile
```diff
@@ -30,6 +30,7 @@ UNFORMATTED=$(shell gofmt -l .)
 
 GET_TOOL=go get -modfile=tool.mod -tool
 RUN_TOOL=go tool -modfile=tool.mod
+SSVSIGNER_RUN_TOOL=go tool -modfile=../tool.mod
```
olegshmuelov, line 33: This should be `go run` instead of `go tool`:

```makefile
  SSVSIGNER_RUN_TOOL=go run -modfile=../tool.mod
```
go tool is for built-in Go tools, while go run is needed to execute packages from tool.mod like golangci-lint.

Makefile
```diff
@@ -39,11 +40,28 @@ lint:
 		exit 1; \
 	fi
 
+.PHONY: ssvsigner-lint
+ssvsigner-lint:
+	cd ssvsigner && $(SSVSIGNER_RUN_TOOL) golangci-lint -c ../.golangci.yaml run -v ./... && cd ..
+	@if [ ! -z "${UNFORMATTED}" ]; then \
+		echo "Some files requires formatting, please run 'go fmt ./...'"; \
+		exit 1; \
+	fi
```
olegshmuelov: A few suggestions to improve integration
```makefile
  .PHONY: lint
  lint:
        $(RUN_TOOL) golangci-lint run -v ./...
        @$(MAKE) ssvsigner-lint

  .PHONY: ssvsigner-lint
  ssvsigner-lint:
        cd ssvsigner && GOWORK=off $(SSVSIGNER_RUN_TOOL) github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -c ../.golangci.yaml -v ./...
```

  Changes:
  1. Added @$(MAKE) ssvsigner-lint to lint target so both run together
  2. Added GOWORK=off (required when using -modfile)
  3. Used full package path for golangci-lint with go run
  4. Removed unnecessary && cd ..
  5. Removed the UNFORMATTED check (it's never set)

  This way make lint runs everything, and the separate workflow files can be deleted!

Makefile
```diff
@@ -39,11 +40,28 @@ lint:
...
+
 .PHONY: full-test
```
olegshmuelov, line 44: not sure if used, but
we have to add ssvsigner tests to full-test as well:
```makefile
  full-test:
        @echo "Running all tests"
        @go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./...
        @cd ssvsigner && go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./...
```

olegshmuelov (changes requested):
Once ssvsigner tests are integrated, we need to upload both coverage files:
  ```yaml
  - name: Upload coverage reports to Codecov
    uses: codecov/codecov-action@v5
    with:
      files: ./coverage.out,./ssvsigner/coverage.out
    env:
      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   ```

Makefile
olegshmuelov: agree
pls see
https://github.com/ssvlabs/ssv/pull/2295#discussion_r2151729433
https://github.com/ssvlabs/ssv/pull/2295#discussion_r2151800760

ssvsigner/keys/keys_test.go
```diff
@@ -104,12 +103,12 @@ func TestSign_Error(t *testing.T) {
...
 		PublicKey:   rsa.PublicKey{},
 		D:           nil,
-		Primes:      nil,
+		Primes:      []*big.Int{nil, nil},
 		Precomputed: rsa.PrecomputedValues{},
 	}}
 
 	_, err := privKey.Sign([]byte("test"))
-	require.ErrorContains(t, err, "missing public modulus")
+	require.Error(t, err) // We use a different implementation for linux, so the error text may be different.
```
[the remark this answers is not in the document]
nkryuchkov: there are no other possible errors

.github/workflows/ssvsigner-lint.yml
nkryuchkov: I added this to the lint pipeline and it caused CI to run twice in https://github.com/ssvlabs/ssv/pull/2277. Anyway, since I've deleted `ssvsigner-lint.yml`, I'm removing it from this PR

olegshmuelov (approved):
Looks great!

Issue #2234, opened 2025-05-28
TestStreamLogs: flaky test


iurii-ssv:
```
Execution client returned an error	{"operation": "SubscribeNewHead", "error": "client is closed"}
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).streamLogsToChan
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:413
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).StreamLogs.func1
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:251
2025-05-26T13:20:06.326Z	ERROR	execution_client	executionclient/execution_client.go:272	failed to stream registry events, reconnecting	{"error": "subscribe heads: client is closed"}
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).StreamLogs.func1
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:272
2025-05-26T13:20:07.326Z	INFO	execution_client	executionclient/execution_client.go:487	reconnecting	{"address": "ws://127.0.0.1:36821"}
2025-05-26T13:20:07.327Z	ERROR	execution_client	executionclient/execution_client.go:468	Execution client returned an error	{"operation": "DialContext", "error": "dial tcp 127.0.0.1:36821: connect: connection refused"}
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).connect
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:468
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).reconnect.func1
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:488
github.com/ssvlabs/ssv/utils/tasks.ExecWithInterval
	/home/runner/work/ssv/ssv/utils/tasks/exec_interval.go:18
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).reconnect
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:486
github.com/ssvlabs/ssv/eth/executionclient.(*ExecutionClient).StreamLogs.func1
	/home/runner/work/ssv/ssv/eth/executionclient/execution_client.go:273
[...]

PR #2296, opened 2025-06-17
execution-client: fix flaky test

iurii-ssv:
It seems the issue originates from us not handling `rpc.ErrClientQuit` error as special case (similar to how we would handle `ErrClosed` error), because of that we never hit [this return statement](https://github.com/ssvlabs/ssv/blob/a67abd0a2cafc4897234311c8d29de8b7ade331d/eth/executionclient/execution_client.go#L319-L322) and simply timeout (while trying to re-connect).

Closes https://github.com/ssvlabs/ssv/issues/2234

### DIFF
--- a/eth/executionclient/multi_client_test.go
+++ b/eth/executionclient/multi_client_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 	"go.uber.org/zap"
@@ -345,23 +346,16 @@ func TestMultiClient_StreamLogs(t *testing.T) {
 
 	logsCh := mc.StreamLogs(ctx, 200)
 
+	var receivedLogs []BlockLogs
 	var wg sync.WaitGroup
 	wg.Add(2) // Expecting two logs: 200, 201
-
-	var receivedLogs []BlockLogs
-	var mu sync.Mutex
-
 	go func() {
 		for blk := range logsCh {
-			mu.Lock()
 			receivedLogs = append(receivedLogs, blk)
-			mu.Unlock()
 			wg.Done()
 		}
 	}()
-
 	wg.Wait()
-
 	require.Len(t, receivedLogs, 2, "expected to receive two logs")
 	require.Equal(t, uint64(200), receivedLogs[0].BlockNumber)
 	require.Equal(t, uint64(201), receivedLogs[1].BlockNumber)
@@ -412,23 +406,16 @@ func TestMultiClient_StreamLogs_Success(t *testing.T) {
 
 	logsCh := mc.StreamLogs(ctx, 200)
 
+	var receivedLogs []BlockLogs
 	var wg sync.WaitGroup
 	wg.Add(2) // Expecting two logs: 200, 201
-
-	var receivedLogs []BlockLogs
-	var mu sync.Mutex
-
 	go func() {
 		for blk := range logsCh {
-			mu.Lock()
 			receivedLogs = append(receivedLogs, blk)
-			mu.Unlock()
 			wg.Done()
 		}
 	}()
-
 	wg.Wait()
-
 	require.Len(t, receivedLogs, 2, "expected to receive two logs")
 	require.Equal(t, uint64(200), receivedLogs[0].BlockNumber)
 	require.Equal(t, uint64(201), receivedLogs[1].BlockNumber)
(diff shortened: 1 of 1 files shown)


eth/executionclient/execution_client.go
```diff
@@ -315,7 +316,7 @@ func (ec *ExecutionClient) StreamLogs(ctx context.Context, fromBlock uint64) <-c
 				return
 			default:
 				nextBlockToProcess, err := ec.streamLogsToChan(ctx, logs, fromBlock)
-				if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) {
+				if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) || errors.Is(err, rpc.ErrClientQuit) {
```
nkryuchkov: Can we extract the check into a function for readability?

eth/executionclient/execution_client.go
```diff
@@ -544,31 +545,34 @@ func (ec *ExecutionClient) connect(ctx context.Context) error {
...
 	tasks.ExecWithInterval(func(lastTick time.Duration) (stop bool, cont bool) {
 		logger.Info("reconnecting")
 		if err := ec.connect(ctx); err != nil {
 			if ec.isClosed() {
 				return true, false
 			}
-			// continue until reaching to limit, and then panic as Ethereum execution client connection is required
 			if lastTick >= ec.reconnectionMaxInterval {
-				logger.Panic("failed to reconnect", zap.Error(err))
+				logger.Panic("Failed to reconnect to Ethereum execution client", zap.Error(err))
```
nkryuchkov: Do we need to mention the Ethereum execution client? The `ExecutionClient`'s logger should be named

nkryuchkov (approved):

eth/executionclient/multi_client.go
```diff
@@ -260,7 +261,7 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan
 				// fromBlock's value in the outer scope is updated here, so this function needs to be a closure
 				f := func(client SingleClientProvider) (any, error) {
 					nextBlockToProcess, err := client.streamLogsToChan(ctx, logs, fromBlock)
-					if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) {
+					if errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled) || errors.Is(err, rpc.ErrClientQuit) {
```
olegshmuelov: While we're fixing error handling here, we also need to update line 273 below:
```diff
- if err != nil && !errors.Is(err, ErrClosed) && !errors.Is(err, context.Canceled) {
+ if err != nil && !errors.Is(err, ErrClosed) && !errors.Is(err, context.Canceled) && !errors.Is(err, rpc.ErrClientQuit) {
```
IIUC without this change, `MultiClient.StreamLogs()` could still call `Fatal()` on `rpc.ErrClientQuit`, causing the same flaky behavior this PR aims to fix.

eth/executionclient/execution_client.go
```diff
@@ -544,31 +545,34 @@ func (ec *ExecutionClient) connect(ctx context.Context) error {
...
 			} else {
-				logger.Warn("could not reconnect, still trying", zap.Error(err))
+				logger.Warn("Could not reconnect to Ethereum execution client, still trying", zap.Error(err))
```
olegshmuelov: nit: error messages should be lowercase per project conventions

eth/executionclient/execution_client.go
```diff
@@ -544,31 +545,34 @@ func (ec *ExecutionClient) connect(ctx context.Context) error {
 	return nil
 }
 
-// reconnect tries to reconnect multiple times with an exponent interval.
-// It panics when reconnecting limit is reached.
-// It must not be called twice in parallel.
+// reconnect tries to reconnect multiple times with an exponent interval. It panics when
+// reconnect-limit is reached since SSV node can't operate without stable connection to
+// Ethereum execution client.
+// It must not be called concurrently.
```
olegshmuelov: Minor comment improvement:
```go
// reconnect tries to reconnect multiple times with an exponential interval.
// It panics when reconnection limit is reached since SSV node can't operate
// without stable connection to Ethereum execution client.
```

eth/executionclient/execution_client.go
olegshmuelov: Consider adding unit tests to verify `rpc.ErrClientQuit` is properly handled:
1. Test that `ExecutionClient.StreamLogs()` returns gracefully on `rpc.ErrClientQuit`
2. Test that `MultiClient.call()` doesn't switch clients on `rpc.ErrClientQuit`
3. Test that `MultiClient.StreamLogs()` doesn't Fatal on `rpc.ErrClientQuit`

This would prevent regression and document expected behavior.

eth/executionclient/execution_client.go
```diff
@@ -544,31 +545,34 @@ func (ec *ExecutionClient) connect(ctx context.Context) error {
...
 func (ec *ExecutionClient) reconnect(ctx context.Context) {
 	logger := ec.logger.With(fields.Address(ec.nodeAddr))
 
 	start := time.Now()
+	success := false
```
olegshmuelov: The `success` variable seems redundant since we're already logging inside the loop. Consider simplifying:
```diff
- success := false
  tasks.ExecWithInterval(func(lastTick time.Duration) (stop bool, cont bool) {
      logger.Info("reconnecting")
      if err := ec.connect(ctx); err != nil {
          // error handling...
          return false, false
      }
-     success = true
+     logger.Info("reconnected to execution client", zap.Duration("took", time.Since(start)))
      return true, false
  }, ...)
- if success {
-     logger.Info("reconnected to execution client", zap.Duration("took", time.Since(start)))
- }
```
This avoids the extra variable and ensures we always log successful reconnection.

olegshmuelov (changes requested):

eth/executionclient/execution_client.go
iurii-ssv: Yes, that would make sense I guess (also because this check is duplicated multiple times), done

eth/executionclient/execution_client.go
iurii-ssv: Right, I keep forgetting `logger` has that feature : ) Reverted now

eth/executionclient/multi_client.go
iurii-ssv: Nice catch, applied

eth/executionclient/execution_client.go
iurii-ssv: Oh, right, I saw quite a bit of logger-messages start with capital letter - so I kinda assumed that's how it's supposed to be,

changed to lowercase now

eth/executionclient/execution_client.go
iurii-ssv: applied

eth/executionclient/execution_client.go
iurii-ssv: applied

eth/executionclient/execution_client.go
iurii-ssv: > 1. Test that ExecutionClient.StreamLogs() returns gracefully on `rpc.ErrClientQuit`

This one ^ is hard to test for unfortunately because it would require for us to "mock out" the execution `client` (which is contrary to how existing tests are set up - they are using the "real" client). 

So it think we should just let existing tests to handle this scenario probabilistically (meaning occasionally the code will execute such that `rpc.ErrClientQuit` is encountered - which is when https://github.com/ssvlabs/ssv/issues/2234 manifests), we just need to keep an eye out for that type of failure to never happen ever again.

> 2. Test that `MultiClient.call()` doesn't switch clients on `rpc.ErrClientQuit`
> 3. Test that `MultiClient.StreamLogs()` doesn't Fatal on `rpc.ErrClientQuit`

Yeah, adding these check makes sense since `MultiClient` tests are written such that we can easily mock out `streamLogsToChan` calls, added now in https://github.com/ssvlabs/ssv/pull/2296/commits/6fd6c57d014c4026005452bb6ba88584fb884ee2

eth/executionclient/multi_client_test.go
```diff
@@ -345,23 +346,16 @@ func TestMultiClient_StreamLogs(t *testing.T) {
...
 	go func() {
 		for blk := range logsCh {
-			mu.Lock()
 			receivedLogs = append(receivedLogs, blk)
-			mu.Unlock()
 			wg.Done()
 		}
 	}()
-
 	wg.Wait()
```
iurii-ssv, line 358: While browsing through these tests I've noticed this mutex isn't necessary since we are using `sync.WaitGroup` to wait for go-routine to finish (and we are not even using this mutex properly here anyway since we don't read-lock it when querying `receivedLogs` variable).

I think we can even get rid of `sync.WaitGroup` (and additional go-routine) by simply ranging over **logs channel**, but maybe that kind of refactoring is not worth bothering with - if you want to see it done though, let me know.

olegshmuelov (approved):

PR #2300, opened 2025-06-18
validator: attempt to fix flaky test TestHandleNonCommitteeMessages

iurii-ssv:
This PR aims to address the `TestHandleNonCommitteeMessages` flaky failure (see failing Github pipeline https://github.com/ssvlabs/ssv/actions/runs/15730946442/job/44331699107):
```
2025-06-18T11:03:47.556136Z	DEBUG	TestHandleNonCommitteeMessages	router message handler stopped
panic: time out!

goroutine 142 [running]:
github.com/ssvlabs/ssv/operator/validator.TestHandleNonCommitteeMessages.func2()
	/home/runner/work/ssv/ssv/operator/validator/controller_test.go:307 +0x38
created by github.com/ssvlabs/ssv/operator/validator.TestHandleNonCommitteeMessages in goroutine 139
	/home/runner/work/ssv/ssv/operator/validator/controller_test.go:305 +0xec7
FAIL	github.com/ssvlabs/ssv/operator/validator	6.608s
# github.com/ssvlabs/ssv/operator/validator/metadata.test
/usr/bin/ld: warning: bint64.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
```

I'm not entirely sure what the root-cause of this failure is, there isn't enough info in the logs to tell 
- if message just didn't manage to arrive in under 4s time (which seems strange anyway, but maybe possible due to some infra-related slow-downs)
- or whether the message was lost entirely
- or yet **another unlikely possibility** would be that `t.Context()` is somehow canceled somewhere at the start of this test (although we don't do it ourselves, so that would me a bug in Golang testing library itself)

hence I'm updating the way we check for arrived messages such that we'll log what messages have actually arrived (and how long it took)

### DIFF
--- a/operator/fee_recipient/controller_test.go
+++ b/operator/fee_recipient/controller_test.go
@@ -160,7 +160,7 @@ func populateStorage(t *testing.T, storage registrystorage.Shares, operatorData 
 		require.NoError(t, storage.Save(nil, createShare(i, operatorData.ID)))
 	}
 
-	// add none committee share
+	// add non-committee share
 	require.NoError(t, storage.Save(nil, createShare(2000, spectypes.OperatorID(1))))
 
 	all := storage.List(nil, registrystorage.ByOperatorID(operatorData.ID), registrystorage.ByNotLiquidated())
--- a/operator/validator/controller_test.go
+++ b/operator/validator/controller_test.go
@@ -8,7 +8,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"sync"
 	"testing"
 	"time"
 
@@ -23,6 +22,9 @@ import (
 
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
+
+	"github.com/ssvlabs/ssv/ssvsigner/ekm"
+	"github.com/ssvlabs/ssv/ssvsigner/keys"
 
 	"github.com/ssvlabs/ssv/beacon/goclient"
 	ibftstorage "github.com/ssvlabs/ssv/ibft/storage"
@@ -44,8 +46,6 @@ import (
 	"github.com/ssvlabs/ssv/protocol/v2/types"
 	registrystorage "github.com/ssvlabs/ssv/registry/storage"
 	registrystoragemocks "github.com/ssvlabs/ssv/registry/storage/mocks"
-	"github.com/ssvlabs/ssv/ssvsigner/ekm"
-	"github.com/ssvlabs/ssv/ssvsigner/keys"
 	kv "github.com/ssvlabs/ssv/storage/badger"
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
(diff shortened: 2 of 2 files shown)


operator/validator/controller_test.go
```diff
@@ -245,70 +245,76 @@ func TestHandleNonCommitteeMessages(t *testing.T) {
...
 
-	// Only exporter handles non committee messages
+	// Only exporter handles non-committee messages
 	ctr.validatorOptions.Exporter = true
 
 	go ctr.handleRouterMessages()
 
-	var wg sync.WaitGroup
-
+	receivedMsgs := make(chan network.DecodedSSVMessage, 1)
```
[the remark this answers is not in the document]
iurii-ssv: There is really no need to have it even buffered at 1, since "blocking/unblocking" behavior of handler is something that's nice to test as well

I'm just using buffer of 1 so there is some buffering potentially happening

operator/validator/controller_test.go
```diff
@@ -245,70 +245,76 @@ func TestHandleNonCommitteeMessages(t *testing.T) {
...
+	timeout := time.After(30 * time.Second)
+	for {
+		select {
+		case msg := <-receivedMsgs:
+			logger.Debug("received message", zap.Any("msg", msg))
+			receivedMsgsCnt++
+			if receivedMsgsCnt == 3 {
+				// Wait a bit and check in case there are unexpected messages still to come.
+				time.Sleep(100 * time.Millisecond)
+				require.Equal(t, 0, len(receivedMsgs), "unexpectedly received more than 3 messages")
```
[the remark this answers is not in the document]
iurii-ssv: I think for this particular use-case both `len` and `select` result in the same behavior (both are probabilistic) with `len` being shorter/easier-to-read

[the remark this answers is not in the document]
iurii-ssv: Sleep is just shorter/cleaner version to write this down IMO - see also https://github.com/ssvlabs/ssv/pull/2300#discussion_r2156793870

operator/validator/controller_test.go
```diff
@@ -245,70 +245,76 @@ func TestHandleNonCommitteeMessages(t *testing.T) {
...
 	})
 
-	go func() {
-		time.Sleep(time.Second * 4)
-		panic("time out!")
-	}()
-
-	wg.Wait()
+	receivedMsgsCnt := 0
+	timeout := time.After(30 * time.Second)
```
olegshmuelov, line 302: The timeout was increased from 4s to 30s. This makes the test much slower to fail and may mask underlying performance issues. Consider a more reasonable timeout

operator/validator/controller_test.go
```diff
@@ -245,70 +245,76 @@ func TestHandleNonCommitteeMessages(t *testing.T) {
...
+			MsgType: spectypes.SSVConsensusMsgType, // this message will be processed
 			MsgID:   identifier,
 			Data:    generateDecidedMessage(t, identifier),
 		},
 	})
-
 	ctr.messageRouter.Route(context.TODO(), &queue.SSVMessage{
 		SSVMessage: &spectypes.SSVMessage{
-			MsgType: spectypes.SSVConsensusMsgType,
+			MsgType: spectypes.SSVConsensusMsgType, // this message will be processed
```
olegshmuelov: nit: consider using `t.Context()`

operator/fee_recipient/controller_test.go
```diff
@@ -160,7 +160,7 @@ func populateStorage(t *testing.T, storage registrystorage.Shares, operatorData
 		require.NoError(t, storage.Save(nil, createShare(i, operatorData.ID)))
 	}
 
-	// add none committee share
```
olegshmuelov, line 163: :)

operator/validator/controller_test.go
iurii-ssv: I just wanted to make absolutely sure the failure isn't due to some 4s delay in message processing (I assume it may be on the order of seconds due to CI allocating too little resources to the pipeline running the test, although unlikely - still it's a possibility),

while the failure itself ideally should never happen (test shouldn't really fail often if ever) - so I don't think it's a significant burden to have long timeouts in such cases,

the active waiting however, on the contrary, I've set to be 100ms (this will happen every run, so naturally it should be short)

operator/validator/controller_test.go
iurii-ssv: applied 👍

y0sher (approved):

Issue #1866, opened 2024-11-21
consider using sync.Pool for heavy allocations


nkryuchkov:
**Is your feature request related to a problem? Please describe.**
Some parts of code like message validation or non-committee observer allocate a lot of memory

**Describe the solution you'd like**
Analyze if we could use `sync.Pool` to reuse the same memory and reduce the amount of allocations. Implement it if so.

**Additional context**
Use Pyroscope to find where we need to apply it first

nkryuchkov:
https://github.com/ssvlabs/ssv/pull/2302 introduces `sync.Pool` to the code base. However, we likely have a lot of other places we need to profile and apply the optimization

vyzo:
A comment here about using sync pools in general: it is not always the right thing to do, as there are adverse effects (notably: memory fragmentation) from holding on to old pages for long, or even for ever. Yes, i know i is a weak reference internally, but that is not the issue.

Also, pool get is not necessary faster than new allocations -- in fact, if a bump allocator is used, it will always be faster than anything involving a mutex. Allocator details change from version to version.

Having said all that, pools are useful if there is measurable gc pressure at a hotspot.  It makes most sense for large pools that have little fragmentation. A pool of just a few small objects makes little sense, and itcan degrade performance in ways that is hard to measure.

vyzo:
also, the compiler keeps getting better. If escape analysis can prove a pointer stays local in the current stack, it can use stack allocation and thats as fast as it can get.

PR #2302, opened 2025-06-18
network: change topic ID calculation algorithm

nkryuchkov:
image

### DIFF
--- a/network/commons/subnets.go
+++ b/network/commons/subnets.go
@@ -2,6 +2,8 @@ package commons
 
 import (
 	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"math"
@@ -41,6 +43,10 @@ func CommitteeTopicID(cid spectypes.CommitteeID) []string {
 	return []string{fmt.Sprintf("%d", CommitteeSubnet(cid))}
 }
 
+func CommitteeTopicIDPostFork(committee []spectypes.OperatorID) []string {
+	return []string{fmt.Sprintf("%d", CommitteeSubnetPostFork(committee))}
+}
+
 // GetTopicFullName returns the topic full name, including prefix
 func GetTopicFullName(baseName string) string {
 	return fmt.Sprintf("%s.%s", topicPrefix, baseName)
@@ -54,6 +60,28 @@ func GetTopicBaseName(topicName string) string {
 // CommitteeSubnet returns the subnet for the given committee
 func CommitteeSubnet(cid spectypes.CommitteeID) uint64 {
 	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), bigIntSubnetsCount)
+	return subnet.Uint64()
+}
+
+func CommitteeSubnetPostFork(committee []spectypes.OperatorID) uint64 {
+	if len(committee) < 4 {
+		panic(fmt.Sprintf("committee is too short: %v", committee))
+	}
+
+	var lowestHash *big.Int
+
+	for _, v := range committee {
+		operatorBytes := make([]byte, 8)
+		binary.LittleEndian.PutUint64(operatorBytes[:], v)
+		operatorHash := sha256.Sum256(operatorBytes)
+		operatorHashNum := new(big.Int).SetBytes(operatorHash[:])
+
+		if lowestHash == nil || operatorHashNum.Cmp(lowestHash) == -1 {
+			lowestHash = operatorHashNum
+		}
+	}
+
+	subnet := new(big.Int).Mod(lowestHash, bigIntSubnetsCount)
 	return subnet.Uint64()
 }
 


network/commons/subnets.go
```diff
@@ -37,8 +39,12 @@ func SubnetTopicID(subnet uint64) string {
 	return fmt.Sprintf("%d", subnet)
 }
 
-func CommitteeTopicID(cid spectypes.CommitteeID) []string {
-	return []string{fmt.Sprintf("%d", CommitteeSubnet(cid))}
+func CommitteeTopicIDAlan(cid spectypes.CommitteeID) []string {
+	return []string{fmt.Sprintf("%d", CommitteeSubnetAlan(cid))}
+}
```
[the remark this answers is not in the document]
nkryuchkov: I think specifying fork name it belongs to is better

networkconfig/mainnet.go
```diff
@@ -30,4 +31,14 @@ var MainnetSSV = &SSVConfig{
 	},
 	TotalEthereumValidators: 1064860, // active_validators from https://mainnet.beaconcha.in/index/data on Apr 18, 2025
```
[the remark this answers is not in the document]
nkryuchkov: irrelevant to the PR

networkconfig/mainnet.go
```diff
@@ -30,4 +31,14 @@ var MainnetSSV = &SSVConfig{
...
 	GasLimit36Epoch:         385150,  // Aug-09-2025 06:40:23 AM UTC
+	Forks: []SSVFork{
+		{
+			Name:  AlanFork,
+			Epoch: 0, // We don't need to set the real Alan fork epoch because pre-Alan ssv-nodes aren't supported anymore.
+		},
+		{
+			Name:  NetworkTopologyFork,
+			Epoch: math.MaxUint64, // TODO: set correct epoch
```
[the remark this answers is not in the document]
nkryuchkov: It will likely be set after merging, we haven't decided on the epoch

nkryuchkov:
@greptileai please review

networkconfig/ssv_fork.go
```diff
@@ -0,0 +1,28 @@
+package networkconfig
+
+import (
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
+
+const (
+	AlanFork            ForkName = iota
+	NetworkTopologyFork          // TODO: rename
```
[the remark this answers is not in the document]
nkryuchkov: Alan is the fork name

nkryuchkov:
@greptileai what needs to be done to increase confidence?

nkryuchkov:
@greptileai please review the latest changes

registry/storage/validatorstore.go
```diff
@@ -54,6 +54,8 @@ type Committee struct {
 	Operators  []spectypes.OperatorID
 	Validators []*types.SSVShare
 	Indices    []phase0.ValidatorIndex
+	Subnet     uint64
+	SubnetAlan uint64
```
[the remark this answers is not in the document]
nkryuchkov: Alan is known to be pre-fork

network/p2p/p2p.go
```diff
@@ -70,6 +70,12 @@ type HostProvider interface {
 	Host() host.Host
 }
 
+type validatorStatusAndSubnet struct {
+	status     validatorStatus
+	subnet     uint64
+	subnetAlan uint64
```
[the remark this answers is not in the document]
nkryuchkov: Alan is known to be pre-fork

nkryuchkov:
@greptileai please review the latest changes

nkryuchkov:
@greptileai review the latest changes

nkryuchkov:
@greptileai review it again

nkryuchkov:
@greptileai please review the latest changes

diegomrsantos:
@nkryuchkov could you please link the SIP/Spec that this PR implements?

diegomrsantos:
> @nkryuchkov could you please link the SIP/Spec that this PR implements?

I believe that's the SIP https://github.com/ssvlabs/SIPs/pull/43?

diegomrsantos:
One small suggestion around NetworkTopologyFork: in Ethereum, Altair introduced a short gossip transition where nodes subscribe to both pre- and post-fork topics, share the duplicate cache, and only drop the old topics a bit after the fork.

In this PR we already do the “before fork” half (dual subscription pre-fork), but then cut straight to post-fork only at the fork epoch. Would it be worth keeping subscriptions to the pre-fork topology topics for a very short window after the fork (e.g. 1 epoch / a few slots) and sharing the dup cache between old/new topics, purely to absorb in-flight messages and slight timing skew?

The fork epoch would still define the only “correct” topology for duties — this would just be a cheap robustness tweak in the gossip layer. If you think the extra complexity isn’t worth it here, happy to drop the idea; I mostly wanted to surface the trade-off explicitly.

nkryuchkov:
@diegomrsantos 

> > @nkryuchkov could you please link the SIP/Spec that this PR implements?
> 
> I believe that's the SIP [ssvlabs/SIPs#43](https://github.com/ssvlabs/SIPs/pull/43)?

Yes, this one. I added it to the PR description.


> One small suggestion around NetworkTopologyFork: in Ethereum, Altair introduced a short gossip transition where nodes subscribe to both pre- and post-fork topics, share the duplicate cache, and only drop the old topics a bit after the fork.
> 
> In this PR we already do the “before fork” half (dual subscription pre-fork), but then cut straight to post-fork only at the fork epoch. Would it be worth keeping subscriptions to the pre-fork topology topics for a very short window after the fork (e.g. 1 epoch / a few slots) and sharing the dup cache between old/new topics, purely to absorb in-flight messages and slight timing skew?
> 
> The fork epoch would still define the only “correct” topology for duties — this would just be a cheap robustness tweak in the gossip layer. If you think the extra complexity isn’t worth it here, happy to drop the idea; I mostly wanted to surface the trade-off explicitly.

I think it might be a good idea. It shouldn't add a lot of complexity in comparison to the whole fork logic

network/commons/subnets.go
diegomrsantos: @nkryuchkov do we need to change the "v2" prefix here?

diegomrsantos:
> @diegomrsantos
> 
> > > @nkryuchkov could you please link the SIP/Spec that this PR implements?
> > 
> > 
> > I believe that's the SIP [ssvlabs/SIPs#43](https://github.com/ssvlabs/SIPs/pull/43)?
> 
> Yes, this one. I added it to the PR description.
> 
> > One small suggestion around NetworkTopologyFork: in Ethereum, Altair introduced a short gossip transition where nodes subscribe to both pre- and post-fork topics, share the duplicate cache, and only drop the old topics a bit after the fork.
> > In this PR we already do the “before fork” half (dual subscription pre-fork), but then cut straight to post-fork only at the fork epoch. Would it be worth keeping subscriptions to the pre-fork topology topics for a very short window after the fork (e.g. 1 epoch / a few slots) and sharing the dup cache between old/new topics, purely to absorb in-flight messages and slight timing skew?
> > The fork epoch would still define the only “correct” topology for duties — this would just be a cheap robustness tweak in the gossip layer. If you think the extra complexity isn’t worth it here, happy to drop the idea; I mostly wanted to surface the trade-off explicitly.
> 
> I think it might be a good idea. It shouldn't add a lot of complexity in comparison to the whole fork logic

Thinking more about the duplicate cache sharing, it's not worth the complexity. Honest nodes won't publish in both pre- and post-fork topics. 

But regarding the fork transition, I have a proposal. I posted it on Discord, but it might be better to discuss it here.

For 1) I'd like to talk about the current dual-subscription in #2302 and how it would look with a new digest approach:

**Current (#2302) dual-subscription flow**
- Topics stay as `ssv.v2.<subnet>`, but pre- vs post-fork subnets are computed differently:
  - Pre-fork ("Alan") subnet/topic: CommitteeTopicIDAlan() → subnet = CommitteeSubnetAlan(committeeID % 128).
  - Post-fork ("network topology") subnet/topic: CommitteeTopicID() → subnet = CommitteeSubnet(operatorIDs, min-hash % 128).
- Before the fork epoch: we SUBSCRIBE to BOTH sets of topics (old + new) so we're ready, but we only PUBLISH on the old topics. Message validation rejects post-fork topics while still pre-fork.
- After the fork epoch: we PUBLISH on the new topics, and the validator rejects old-topic messages. The dual-subscribe is hardcoded for this specific fork; no digest or negotiation in handshake.

**How it would work with a digest versioning scheme**
- One network tag: replace domaintype + network_id with a single typed [u8;4] network_tag used in ENR and handshake (reject on mismatch).
- Gossip digest: a short tag (e.g., first 4 bytes of SHA256) over (network_tag + gossip_schema_version + message_format_version).
- Topics become `ssv.<digest>.<subnet>`. The subnet derivation logic can change per schema version; the digest keeps topics disjoint.
- Dual-subscribe becomes generic: config says `current_digest` + optional `next_digest`. We subscribe to both; validator gates by digest. No Alan-specific branches.
- Handshake advertises supported digest(s); peers pick an overlap or disconnect, avoiding mixed meshes. Discovery still filters on network_tag; you can expose the current digest in ENR for visibility (like eth forkid), but don't hard-filter there if you need overlap.
- Grace period: keep both digests before the fork epoch and for some time after (like Ethereum post-Altair), then drop the old digest.

Why bother: standard, repeatable dual-subscription for any future gossip-incompatible change, one source of truth for network identity, and no bespoke Alan vs network-topology code next time we tweak topics/schema/encoding.

Please let me know what you guys think.

vyzo:
You need to be careful with the validation actions in order to avoid messing with peer scores.

Before the transition, if you receive messages in the new topics, the validator should IGNORE, not REJECT.
This is important as ignoring does not curry behavioral penalties, while REJECT does. It is also semantically clear for the spec, as your node does not know whether other peers have actually transitioned before.

Aftet the transition, the logic shifts, accept on new topics but IGNORE in the old topics (again, not REJECT). In fact, since we are going to ignore old messages anyway, we dont even need a cooldown period and arguably can immediately leave the old topics. 

Is there a tangible benefit for staying in the old topics after the transition?
If so, the cooldown period should not be long and leave after that anyway.

message/validation/validation.go
```diff
@@ -227,7 +227,15 @@ func (mv *messageValidator) committeeChecks(signedSSVMessage *spectypes.SignedSS
 	}
 
 	// Rule: Check if message was sent in the correct topic
-	messageTopics := commons.CommitteeTopicID(committeeInfo.committeeID)
+	var messageTopics []string
+
+	// Unlike the logic in p2p, where we subscribe the post-fork subnets before fork to be ready at the fork,
+	// we don't expect post-fork messages to be received before the fork.
```
vyzo: yes, but.  You need to warm up the post fork mesh, so it is better to subscribe before.

network/commons/subnets.go
```diff
@@ -51,16 +56,53 @@ func GetTopicBaseName(topicName string) string {
 	return strings.TrimPrefix(topicName, topicPrefix+".")
 }
 
-// CommitteeSubnet returns the subnet for the given committee
-func CommitteeSubnet(cid spectypes.CommitteeID) uint64 {
-	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), bigIntSubnetsCount)
-	return subnet.Uint64()
+var bigIntPool = sync.Pool{
```
vyzo: is this such a hotspot to need a pool?

network/p2p/p2p_pubsub.go
```diff
@@ -54,13 +55,27 @@ func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedS
 	var topics []string
 
 	if msg.SSVMessage.MsgID.GetRoleType() == spectypes.RoleCommittee {
-		topics = commons.CommitteeTopicID(spectypes.CommitteeID(msg.SSVMessage.MsgID.GetDutyExecutorID()[16:]))
+		// Unlike the logic in p2p, where we subscribe the post-fork subnets before fork to be ready at the fork,
```
vyzo: Again, i think this is mistaken.  We should subscribe to post fork topics a little before to warm up the mesh.

vyzo (commented):
first pass, so just a couple comments for now.

vyzo:
A couple comments on the proposed digest versioning scheme:

I don't think we should do it. 

It is a little too clever scheme that obscures topic names for marginal gains.

This is an infrequent transition that requires care and operational oversight.
Having opaque topic identifiers does not help, in fact it hinders debugging. 

Also, 4 bytes... it is too short, you are likely to get conflicts from birthday paradox eventually, and it won't be easy to fix (or even detect) the issue.

network/commons/subnets.go
vyzo: In fact, i don't think this should be done as part of this pr.

It is an orthogonal change that gets sneaked in with something unrelated, it should go in its own pr (possibly with other pool introductions) and get measured for effect.

vyzo: More generally, if bigints are used heavily and seem to be hotspots, we should make a specialized pool for this and use it everywhere.

Many little pools are a lot worse than one big pool memory wise.

diegomrsantos:
> Is there a tangible benefit for staying in the old topics after the transition?

The only benefit might be receiving late but useful messages (not duplicates) that were published to the pre-fork topics. 

> If so, the cooldown period should not be long and leave after that anyway.

I agree. They might be useful for a few seconds, but probably won't matter after that.

@vyzo what do you think?

vyzo:
alright, but make sure we ignore the messages, not reject in the pubsub validator.

diegomrsantos:
> A couple comments on the proposed digest versioning scheme:
> 
> I don't think we should do it.
> 
> It is a little too clever scheme that obscures topic names for marginal gains.
> 
> This is an infrequent transition that requires care and operational oversight. Having opaque topic identifiers does not help, in fact it hinders debugging.
> 
> Also, 4 bytes... it is too short, you are likely to get conflicts from birthday paradox eventually, and it won't be easy to fix (or even detect) the issue.

Makes sense, I agree that fully opaque topic IDs are not great for ops/debugging.

What I’m proposing is closer to Ethereum’s pattern: keep topics readable, but add a short fork-id segment purely for domain separation. In Ethereum the fork digest sits in the topic path (`/eth2/<fork_digest>/<name>/<encoding>`), so different forks naturally end up in different meshes.  ￼

For SSV (where the topic is basically “the subnet”), the equivalent could be something like `ssv.<fork_id>.<subnet>`. That keeps topic strings human-readable, but avoids mixed-version meshes and reduces how much we have to rely on careful IGNORE vs REJECT behavior during transitions (since old nodes simply won’t be on the new fork topics).  ￼

That said, for this specific change (same message formats, just a remap of committee→subnet), I agree it’s not strictly necessary to introduce a fork-id or bump the prefix, dual-subscription + fork-epoch gating is likely enough. My point is more about having a clean, repeatable mechanism for future gossip-incompatible changes without bespoke fork branches each time.  ￼

Re: 4 bytes,  ETH uses 4 bytes and even notes that 4 bytes is sufficient in practice. If we ever feel uneasy, we can choose a longer fork-id (e.g. 8 bytes) or pair it with a readable label; it’s just a protocol design choice / topic string convention.

vyzo:
ok, these are reasonable arguments.

Still:
- if we adopt something like this, we should keep something semantic in the name.
- i don't see a compelling reason to do it _now_.  I think we should keep the pr to the minimal changes needed and have a more complete discussion for the topic names; a SIP is more appropriate than introducing this change here.

diegomrsantos:
The sequence diagram in https://github.com/ssvlabs/ssv/pull/2302#issuecomment-3612576298 is helpful and already captures the intended transition state machine:

- **pre-fork**: dual-subscribe (old+new) to warm the new mesh, but **publish via old**
- **post-fork**: **publish via new** and (quickly) be **new-only**

What I’d like to propose is: let’s extract this into a short, *normative* Network Topology Fork Transition section in SIP-43, so go-ssv + Anchor can’t accidentally diverge as the PR evolves.

### Routing Policy (single source of truth)
Define one routing policy (driven by epoch/slot) that answers:
- `subscribe_schemas(t)` → {old} / {old+new} / {new}
- `publish_schema(t)` → old / new
- `accept_schemas(t)` → which side we actually *process*

`FORK_EPOCH` should be a **placeholder** resolved via network config (same as current code paths where fork epoch is configured/disabled), not something the spec hardcodes.

### Transition rules (consistent with the diagram + our current direction)
- **Warmup window (pre-fork)**: subscribe old+new, publish old, accept old
  - if we receive messages on the *new* topics before the fork: validator should **IGNORE, not REJECT** (to avoid peer-score penalties while the network is in mixed rollout / timing skew).
- **At fork**: publish new.
- **Post-fork**: drop old topics **immediately or within a few slots max**.
  - If we do keep an overlap of a few slots, I think it should be “subscribe+ACCEPT both” during that overlap (not subscribe and then IGNORE/REJECT).

nkryuchkov:
@MatheusFranco99 can we please define the Network Topology Fork Transition section in https://github.com/ssvlabs/SIPs/pull/43? It looks worth defining in the SIP

@diegomrsantos @vyzo good points on the fork transition, thanks. Let's define the approach in the SIP, and I'll adjust the implementation once everyone agrees on it.

network/commons/subnets.go
nkryuchkov: @vyzo 

> is this such a hotspot to need a pool?

Yes, this is quite a hot spot. We even added `SetCommitteeSubnet` that writes the result to the passed `*big.Int` and it indeed showed a significant reduction of allocations (it was a while ago, and I don't have exact numbers unfortunately). 

The new algorithm is more complex and allocation-heavy, so it definitely needs performance optimizations. The `sync.Pool` solution seemed to me the best tradeoff while I was implementing this. 

Alternatively, we could create a struct holding the allocated `big.Int`s, but we would have to make it thread-safe and make sure we don't have mutex contention issues. Also, this would inflate the PR.

Pre-allocating `big.Int`s before the loops calling `CommitteeSubnet` and passing those variables would probably be the fastest solution with the least amount of allocations, but it's super ugly. Passing the output variable to `SetCommitteeSubnet` already doesn't look great, to be honest.


> In fact, i don't think this should be done as part of this pr.
> 
> It is an orthogonal change that gets sneaked in with something unrelated, it should go in its own pr (possibly with other pool introductions) and get measured for effect.

If you mean the new implementation, I think the optimization is an important part of it, given that we already had allocation issues with the simple Alan algorithm. If we remove it, the performance won't be acceptable. Although, I'm not ruling out that there might be a better optimization.

We could probably leave the Alan implementation in its original state, but I don't like having to pass output variables  just for Alan, given that all places where this is called have been changed anyway.



> More generally, if bigints are used heavily and seem to be hotspots, we should make a specialized pool for this and use it everywhere.
> 
> Many little pools are a lot worse than one big pool memory wise.

This is the only place we have faced issues with, so the only pool we have is just for this. We have [a ticket to consider using `sync.Pool` in other places too](https://github.com/ssvlabs/ssv/issues/1866). It needs some work on spotting such places. Of course, if we find them, we'll need to think about managing the pools, but I'm not sure we'll find anything worth optimizing.

vyzo: Ok, but i still think it should be a separate pr, it is completely orthogonal to the intent of this pr.

for preallocation, the ugliness could be hidden behind a facade. But this probably an overkill, we can just use a pool.

vyzo: but still, please cherry pick for a separate pr and we can also try some preallocation approaches and see if there is any worthwhile improvement. I suspect the pool will be good enough.

nkryuchkov: > Ok, but i still think it should be a separate pr, it is completely orthogonal to the intent of this pr.
> 
> for preallocation, the ugliness could be hidden behind a facade. But this probably an overkill, we can just use a pool.

Why is it completely orthogonal? The intent of this PR is a production-ready implementation of the algorithm that meets our performance requirements. If the pool does that, it must be a part of the PR. If something does it better, we'll remove the pool.

I'll try to benchmark it in tests, the local testnet, and on stage nodes to get some numbers. I see I added subnet caching, so maybe it will let us get rid of the subnet calculation optimizations.

vyzo: see pr title; this is a change for the topic ID calculation algorithm.

The bigint allocation change is a performance optimization from a separate issue.

We need some discipline in change management. You are fixing a (real, obseved) issue that should be reviewed seperately. When we look at the (squashed) commit history months from now, we need to know what is what at a glance.

nkryuchkov: Why can't the performance optimization be a part of the algorithm? We can't even merge a PR if it causes performance issues. It will roll out to stage nodes and worsen the performance. I don't see any problem in seeing the optimized solution in the squashed commit, and the optimization isn't complex to understand.

vyzo: it is not causing a new issue, it is a known one.

Also, it is not that we are going to ship them seperately. Furthermore, this is not optimizing the algorithm... it is completely irrelevant to the algorithm.

Again, it is fixing a separate issue with implementation performance. It is completely irrelevant to correctness.

vyzo: i am a little bit of a maniac when it comes to code and pr flow cleanliness. I am sorry about that, but it does make things better in the long run.

nkryuchkov: @vyzo Can you please explain what flow you're suggesting? I'm not sure I'm following. 

I share your point of view on the code/PR cleanliness. But my opinion is that a PR must be atomic (unless there are important reasons not to do so), also in terms of correctness and performance.


> Also, it is not that we are going to ship them seperately. Furthermore, this is not optimizing the algorithm... it is completely irrelevant to the algorithm.

I would agree if we discussed spec. But for the implementation, the algorithm includes meeting all the requirements, like performance, metrics, logs, tests, and so on. And I don't understand why we need to extract some of them or all of them into a different PR. So I would appreciate it if you could explain that in a more detailed way

> Again, it is fixing a separate issue with implementation performance. It is completely irrelevant to correctness.

Well, the issue is that the algorithm we are migrating to is more complex and allocation-heavy, so we need to apply some optimizations, which makes them inseparable. If we didn't migrate, we wouldn't have to add those optimizations

vyzo: well ok, we are not going to argue for ever; do what you think is best.

however, if you do it in a separate pr, you can explore potential faster approaches and see if it makes sense. Maybe not a big deal for this, but it could inform future decisions without information pollution in this pr.

nkryuchkov: @vyzo sure, I won't be experimenting here. Thanks

diegomrsantos: I agree with making PRs as small and focused as possible. It's easier to review and spot mistakes. IMO, extracting the bigint/pool perf work makes sense only if this PR is still within acceptable alloc/CPU budgets on its own; if not, then the perf work is part of making this change shippable.

If we do split it, we should make the dependency explicit and enforce sequencing:

1. Open a perf PR stacked on this branch (base = network-topology-topic-id) so the diff is purely perf.
2. Keep this PR Draft (or add a do-not-merge label) and put “Blocked by #<perfPR>” in the description.
3. After the perf PR is approved + green, flip this PR to Ready and merge it.
4. Immediately retarget the perf PR to stage and get it merged as well.

GitHub can’t atomically merge two PRs, but this makes it very unlikely we accidentally ship only one side.

MatheusFranco99:
@nkryuchkov Thanks! This has been solved in the SIP in this [commit](https://github.com/ssvlabs/SIPs/pull/43/commits/e8b2bbb7bafc1c758b703178a12ea1f8a46b23ce)

nkryuchkov:
@greptileai please review the latest changes and make sure they correspond this specification:

## Spec Change
For each committee, we assign its topic in the following way:

```javascript
CONSTANTS:
NUM_TOPICS = 128
MAX_UINT256 = (2^256) - 1

FUNCTION AssignCommitteeToTopic(committee: list[uint64]):
	min_int = MAX_UINT256
	
	FOR each operator_id IN committee:
		# 1. Serialize the uint64 Operator ID to 8 bytes (Little-Endian)
		input_bytes = serialize_uint64_little_endian(operator_id)
		
		# 2. Compute the SHA-256 hash, resulting in 32 bytes
		hash_bytes = SHA256(input_bytes)
		
		# 3. Convert to uint256 (Big-Endian)
		current_int = uint256_from_bytes_big_endian(hash_bytes)
		
		# 4. Get minimum
		IF current_int < min_int:
			min_int = current_int
	
	# 5. Modulo for topic ID
	RETURN min_int MOD NUM_TOPICS
```

## Fork Transition

To safely transition through the fork, we define the transition policy below. Let:
- $old$: denotes the set of topics the operator should be subscribed to before the fork (according to its committees),
- $new$: denotes the set of topics the operator should subscribe to after the fork.

### Before the Fork

The operator is subscribed to the $old$ topics and publishes to them.

`PRIOR_WINDOW` epochs before the fork, it subscribes to all $new \cup old$.
During this window, for the new topics ($new \setminus old$):
- the operator warms up its mesh for the new topics by setting up GRAFT connections,
- no message is expected until the fork as the topics are brand-new.
Still, due to clocks discrepancies, any incoming non-rejectable message (according to the new fork message validation rules) can be accepted and propagated, while rejectable ones can be ignored to avoid score degradation.

We set `PRIOR_WINDOW = 1` epochs to allow
enough time for the operator to set up the new topics before the fork,
and avoid a long overloading period (due to the extra messages being processed in the new topic).

### At and After the Fork

At the fork epoch, the operator immediately unsubscribes from the old topics
(more precisely, $old \setminus new$).
It keeps its subscription to the $new$ topics and only publishes to them.

During the unsubscription period, any message in the unsubscribed topics is **accepted**.

> [!NOTE]
> Note that this behaviour oversees the right last messages sent before the fork that weren't yet delivered,
and could be useful to finish a duty.
> While the more robust approach would be to keep listening on those topics for an extra tiny window,
> this adds extra complexity, extends the performance degradation (as more redundant messages are processed),
> and possibly incurs a negligible gain as duties lasting more than 4 seconds are also less likely to be included either way
> (remember that the last initiated duties will be aggregator and sync committee duties that start 4 seconds before the fork).

nkryuchkov:
@greptileai please review the latest changes and make sure they correspond this specification:

## Topic Names

Currently, prior to this change, topics were named by `ssv.v2.<subnet>` for `<subnet>` from 0 to 127.

On the Boole fork, which introduces this change, the new topics will be named by `/ssv/<ethereum_network_name>/boole/<subnet>`,
where `<subnet>` varies from 0 to 127 as usual, and `<ethereum_network_name>` is the name of the Ethereum network in lower-case, i.e. `mainnet`, `sepolia`, `holesky`, or `hoodi`.
For example, topic 0 for mainnet would be `/ssv/mainnet/boole/0`.

vyzo (dismissed):
looks good generally, but i might have missed some subtlety.

network/topics/scoring.go
```diff
@@ -215,18 +216,31 @@ func topicScoreParams(logger *zap.Logger, cfg *PubSubConfig, committeesProvider
...
+			if topic == booleTopic {
+				topicCommittees = append(topicCommittees, committee)
+			}
+		case netCfg.BooleForkInPriorWindow(currentEpoch):
+			if topic == alanTopic {
+				topicCommittees = append(topicCommittees, committee)
+				continue
+			}
+			if committee.SubnetAlan != committee.Subnet && topic == booleTopic {
+				topicCommittees = append(topicCommittees, committee)
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai We need to add one committee only once. So if it's found for Alan, we don't need to try to find it for Boole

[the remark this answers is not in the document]
nkryuchkov: @greptileai applied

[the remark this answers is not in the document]
nkryuchkov: @greptileai I have given it more thought and decided to revert this change because we have `continue` to ensure we don't enter the Boole branch if the Alan branch matches. Also, both `topic == alanTopic` and `topic == booleTopic` check full topics which don't intersect

nkryuchkov:
@greptileai please review this once again

nkryuchkov:
@greptileai another review please

cli/operator/node.go
```diff
@@ -623,12 +624,18 @@ var StartNodeCmd = &cobra.Command{
...
+				if subnet := v.CommitteeSubnet(); !mySubnets.IsSet(subnet) {
 					mySubnets.Set(subnet)
 					myActiveSubnets++
 				}
+				if !booleFork {
+					if alanSubnet := v.CommitteeSubnetAlan(); !mySubnets.IsSet(alanSubnet) {
+						mySubnets.Set(alanSubnet)
+						myActiveSubnets++
+					}
+				}
```
y0sher: shouldn't we also count in both when in the preparation window?

nkryuchkov: We use this for max peers calculation and do it only once at the start. So post-fork, we use only the post-fork amount. Pre-fork, we use pre-fork+post-fork, so we will have both in the preparation window.

Perhaps we could make this value dynamic, but I'm afraid the logic will become very complex and error-prone and nevertheless won't give any benefits

network/commons/subnets.go
```diff
@@ -37,37 +42,87 @@ func SubnetTopicID(subnet uint64) string {
...
-func GetTopicFullName(baseName string) string {
-	return fmt.Sprintf("%s.%s", topicPrefix, baseName)
+// TopicFullName returns the topic full name, including prefix.
+func TopicFullName(networkName, baseName string) string {
+	return fmt.Sprintf("%s/%s/%s/%s", topicRoot, networkName, booleTopicFork, baseName)
 }
 
-// GetTopicBaseName return the base topic name of the topic, w/o ssv prefix
+// GetTopicBaseName return the base topic name of the topic, w/o ssv prefix.
 func GetTopicBaseName(topicName string) string {
```
iurii-ssv: Maybe `GetTopicBaseName` -> `TopicShortName`, plus we probably want to write some unit-tests for this func (I don't think current tests cover it) ?

network/commons/subnets.go
```diff
@@ -37,37 +42,87 @@ func SubnetTopicID(subnet uint64) string {
...
+func Topics(networkName string) []string {
+	topics := make([]string, 0, SubnetsCount*2)
 	for i := uint64(0); i < SubnetsCount; i++ {
-		topics[i] = GetTopicFullName(SubnetTopicID(i))
+		baseName := SubnetTopicID(i)
+		topics = append(topics, AlanTopicFullName(baseName))
+		topics = append(topics, TopicFullName(networkName, baseName))
 	}
 	return topics
 }
```
iurii-ssv, line 103: I would leave a comment here to clarify that we can remove the Alan topics from this func once we are past the Boole fork

network/p2p/p2p_pubsub.go
```diff
@@ -132,16 +145,47 @@ func (n *p2pNetwork) SubscribeRandoms(numSubnets int) error {
 // SubscribedSubnets returns the subnets the node is subscribed to, consisting of the fixed subnets
 // and the active committees/validators.
 func (n *p2pNetwork) SubscribedSubnets() commons.Subnets {
-	// Compute the new subnets according to the active committees/validators.
-	updatedSubnets := n.persistentSubnets
+	alanSubnets, booleSubnets := n.subscribedSubnetsWithFormats()
+	return unionSubnets(alanSubnets, booleSubnets)
+}
+
+func (n *p2pNetwork) subscribedSubnetsWithFormats() (commons.Subnets, commons.Subnets) {
```
iurii-ssv: What does the `WithFormats` refer to ? Seems like `subscribedSubnetsForCurrentEpoch` would be more appropriate name for it.

Would leave a comment that we can get rid of Alan subnets after we are past Boole fork.

network/p2p/p2p.go
```diff
@@ -461,20 +467,22 @@ func (n *p2pNetwork) isReady() bool {
 	return atomic.LoadInt32(&n.state) == stateReady
 }
 
-// UpdateSubnets will update the registered subnets according to active validators
-// NOTE: it won't subscribe to the subnets (use subscribeToFixedSubnets for that)
+// UpdateSubnets will update the registered subnets according to active validators.
 func (n *p2pNetwork) UpdateSubnets() {
```
iurii-ssv, line 473: This `NOTE` here says we already call `subscribeToFixedSubnets` to **subscribe** to those subnets ... so now we subscribe twice ? We probably want to resolve the discrepancy ... was there a reason to call `subscribeToFixedSubnets` elsewhere (in `p2pNetwork.Start` it is I think) ?

> in `p2pNetwork.Start` it is I think

I suspect the idea is that we want to block and wait for `subscribeToFixedSubnets` to complete before exiting from `p2pNetwork.Start` func ... so maybe we can't just simply move this code from there (without also ensuring it executes fully, before we can proceed with the rest of startup)

network/p2p/p2p.go
```diff
@@ -515,17 +531,47 @@ func (n *p2pNetwork) UpdateSubnets() {
...
+		if addedBooleSubnets.HasActive() {
+			for _, addedSubnet := range addedBooleSubnets.SubnetList() {
+				if err := n.subscribeSubnet(addedSubnet, true); err != nil {
+					n.logger.Debug("could not subscribe to subnet", zap.Uint64("subnet", addedSubnet), zap.Error(err))
+					errs = errors.Join(errs, err)
+				} else {
+					n.logger.Debug("subscribed to subnet", zap.Uint64("subnet", addedSubnet))
+				}
+			}
+		}
```
iurii-ssv, line 569: Probably want to mention in these logs that's Boole subnet we are subscribing to here,

and the same for removing/unsubscribing below

network/p2p/p2p.go
```diff
@@ -515,17 +531,47 @@ func (n *p2pNetwork) UpdateSubnets() {
...
+		if addedAlanSubnets.HasActive() {
+			for _, addedSubnet := range addedAlanSubnets.SubnetList() {
+				if err := n.subscribeSubnet(addedSubnet, false); err != nil {
+					n.logger.Debug("could not subscribe to subnet", zap.Uint64("subnet", addedSubnet), zap.Error(err))
+					errs = errors.Join(errs, err)
+				} else {
+					n.logger.Debug("subscribed to subnet", zap.Uint64("subnet", addedSubnet))
+				}
+			}
+		}
```
iurii-ssv, line 552: Probably want to mention in these logs that's Alan subnet we are subscribing to here

and the same for removing/unsubscribing below

network/p2p/p2p.go
```diff
@@ -515,17 +531,47 @@ func (n *p2pNetwork) UpdateSubnets() {
...
 			}
-
-			// Unsubscribe from the removed subnets.
-			for _, removedSubnet := range removedSubnets {
-				if err := n.unsubscribeSubnet(removedSubnet); err != nil {
+		}
+		if removedAlanSubnets.HasActive() {
+			for _, removedSubnet := range removedAlanSubnets.SubnetList() {
+				if err := n.unsubscribeSubnet(removedSubnet, false); err != nil {
+					n.logger.Debug("could not unsubscribe from subnet", zap.Uint64("subnet", removedSubnet), zap.Error(err))
```
iurii-ssv: Kinda redundant to log the `err` here (+ doing it under `Debug`) while we also log the aggregate `errs` below in `updated subnets` line

network/topics/controller.go
```diff
@@ -158,18 +156,12 @@ func (ctrl *topicsCtrl) Peers(name string) ([]peer.ID, error) {
 
 // Topics lists all topics this node is subscribed to
 func (ctrl *topicsCtrl) Topics() []string {
-	topics := ctrl.ps.GetTopics()
-	for i, tp := range topics {
-		topics[i] = commons.GetTopicBaseName(tp)
-	}
-	return topics
+	return ctrl.ps.GetTopics()
 }
```
iurii-ssv, line 160: This is a behavior change I believe, was this not working correctly previously ?

I see that you now prefer using the full topic name in many places, just wanted to make sure every such usage has been adjusted accordingly.

protocol/v2/types/ssvshare.go
```diff
@@ -52,6 +53,10 @@ type SSVShare struct {
 
 	// committeeID is a cached value for committee ID so we don't recompute it every time.
 	committeeID atomic.Pointer[spectypes.CommitteeID]
+	// committeeSubnet is a cached value for committee subnet so we don't recompute it every time.
+	committeeSubnet atomic.Pointer[uint64]
+	// committeeSubnetAlan is a cached value for committee subnet in Alan fork so we don't recompute it every time.
+	committeeSubnetAlan atomic.Pointer[uint64]
```
iurii-ssv: Why not use `atomic.Uint64` instead of `atomic.Pointer` ?

message/validation/validation.go
```diff
@@ -227,12 +229,25 @@ func (mv *messageValidator) committeeChecks(signedSSVMessage *spectypes.SignedSS
...
 	// Rule: Check if message was sent in the correct topic
-	messageTopics := commons.CommitteeTopicID(committeeInfo.committeeID)
-	topicBaseName := commons.GetTopicBaseName(topic)
-	if !slices.Contains(messageTopics, topicBaseName) {
+	var messageTopicNames []string
+	currentEpoch := mv.netCfg.EstimatedCurrentEpoch()
+	alanTopic := commons.SubnetTopicID(committeeInfo.subnetAlan)
+	booleTopic := commons.SubnetTopicID(committeeInfo.subnet)
+	alanTopicFullName := commons.AlanTopicFullName(alanTopic)
+	booleTopicFullName := commons.TopicFullName(mv.netCfg.Beacon.Name, booleTopic)
```
iurii-ssv: Maybe rename `commons.SubnetTopicID` -> `commons.SubnetTopicShortName` to be explicit ? And same for Alan

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...715461a","133e2cb50ecb8f8c","cca5f9fec3dcebe2","fc1b1557f77c7cf6","f071d8bb5b0cf786","944df94b87c2cd5","a85ae171d39e0a84","c0be33f50490fb3b","53500d2b52a17288","1820d9255597f7fb","c28c83b918bdc2b7","e996089159ec055c","abc9b4b9071c1d44","e416688694148819","6a006d8d371ed3d6","70d6ef8386733e6c","26350cbaf72af2fa","cf67fb1860d7a4e7","82164728c372257","ae5af77b8fa64ce1","ba115ceafa50370e","8d4aa3e756b26ed1","86cb4feef121de3a","650eff932e5bdd86","3a2dfb376466ca1b","78a91c2f37a8db23","b0160030aa9b3f5c","a41ce874e5a13274","33029957e12824a3","4b4de4a8d490276","40985ff0314c8b4c","900c00ca8d80ee7e","b115174ebee5cf28","de191916a1f95911","7f808ce21eca9166","5550d972d005f131","70374c35c02f841d","ec1593e2a3a93cb7","8f58e8703de43eea","c911960bd5da13c5","4a4660eab34adedf","ffb8aef048156e37","49f3fc4d95d36c3a","7353a6fef084d04d","a2c96e61414bc60a","3f30544352da816b","eb07b24430298f1a","a9eafe539df0931","89bac13747ec1329","795d22e114b3eb39","aae0c202d4386353","3926bebbc3f746d2","36bad94ccdfac307","689f1fca82b3b35f","fcaf8651b6530ec8","8b9c32081daabe01","bb621df5b3e6af55","e21813fa5d9b6684","242c7ae3a0976683","d6b3274fa95bfb9d","81196984a532ba75","96d240cebcbf31bb","f541803278b608","fc3404ff2f75d203"]
```
iurii-ssv: I think we no longer using the spec-alignment pipeline, so there is no need to change it ?

networkconfig/network.go
```diff
@@ -30,3 +33,34 @@ func (n Network) StorageName() string {
...
+	}
+	return booleEpoch - boolePriorWindowEpochs
+}
+
+func (n Network) BooleForkInPriorWindow(epoch phase0.Epoch) bool {
+	if n.BooleForkAtEpoch(epoch) {
+		return false
+	}
+	return epoch >= n.BoolePriorWindowStartEpoch()
+}
```
iurii-ssv: We should probably cover these methods with some basic unit-tests ?

iurii-ssv (commented):
I might have missed it, but I don't think we unsubscribe from old topics anywhere in this PR **once we've passed** the Boole fork ?

Also:
- let me cross-link a discussion of considerations about ["early arrivals" on the new topic](https://github.com/ssvlabs/SIPs/pull/43#discussion_r2712662418) ... seems like we currently allow for "early arrivals" and fully process them (not just validate & propagate the message while skipping processing), BUT we are using the pre-Boole-fork rule-set when validating these early arrivals ?
- as for the "late arrivals" for the messages on the old topic, we are currently ignoring them with `ErrIncorrectTopic` (as per `Rule: Check if message was sent in the correct topic`) ... but maybe we'd wanna add [a "grace period" to process some of them](https://github.com/ssvlabs/SIPs/pull/43#discussion_r2712483250)

network/p2p/p2p.go
```diff
@@ -461,20 +467,22 @@ func (n *p2pNetwork) isReady() bool {
...
-// UpdateSubnets will update the registered subnets according to active validators
-// NOTE: it won't subscribe to the subnets (use subscribeToFixedSubnets for that)
+// UpdateSubnets will update the registered subnets according to active validators.
 func (n *p2pNetwork) UpdateSubnets() {
 	// TODO: this is a temporary fix to update subnets when validators are added/removed,
 	// there is a pending PR to replace this: https://github.com/ssvlabs/ssv/pull/990
 	ticker := time.NewTicker(time.Second)
 	registeredSubnets := commons.Subnets{}
+	registeredAlanSubnets := commons.Subnets{}
+	registeredBooleSubnets := commons.Subnets{}
```
iurii-ssv: Maybe we should add `prev` prefix for all these `registered` vars ? It would make the intent clearer I think

network/p2p/p2p.go
```diff
@@ -461,20 +467,22 @@ func (n *p2pNetwork) isReady() bool {
...
 	defer ticker.Stop()
 
 	// Run immediately and then every second.
 	for ; true; <-ticker.C {
 		start := time.Now()
 
-		updatedSubnets := n.SubscribedSubnets()
+		alanSubnets, booleSubnets := n.subscribedSubnetsWithFormats()
+		updatedSubnets := unionSubnets(alanSubnets, booleSubnets)
 		n.currentSubnets = updatedSubnets
```
iurii-ssv: I think renaming `updatedSubnets`->`latestSubnets` would clarify this code a bit

network/p2p/p2p.go
iurii-ssv: > This NOTE here says we already call subscribeToFixedSubnets to subscribe to those subnets ... so now we subscribe twice ? We probably want to resolve the discrepancy ... was there a reason to call subscribeToFixedSubnets elsewhere (in p2pNetwork.Start it is I think) ?

Okay, never mind, I think I figured it out. The `UpdateSubnets` does a very specific part of the job (I guess it is handling the subscriptions during the fork transition period) ... **I think we want to leave a more detailed description-comment for what `UpdateSubnets` is supposed to do**.

> I suspect the idea is that we want to block and wait for `subscribeToFixedSubnets` to complete before exiting from `p2pNetwork.Start` func ... so maybe we can't just simply move this code from there (without also ensuring it executes fully, before we can proceed with the rest of startup)

The `subscribeToFixedSubnets` in `p2pNetwork.Start` func hasn't "been moved" (it's still there), so we are potentially subscribing twice ... but the code seems to be able to handle it fine (making 2nd call a no-op).

network/commons/subnets.go
nkryuchkov: `BaseName` looks more correct to me, because `ShortName` kind of implies 1:1 mapping. However, converting from a base name to a full name requires knowing the fork for it. Perhaps, something like `TopicNumberFromFullName` would be most precise here, but then we likely need to change the return type. I'll think how we can improve this

network/commons/subnets.go
nkryuchkov: We actually can avoid appending Alan topics if we start in the Boole fork

protocol/v2/types/ssvshare.go
nkryuchkov: To make it optional/nullable

scripts/spec-alignment/differ.config.yaml
nkryuchkov: Yeah, it was changed when we used it. I'll remove it

network/topics/controller.go
nkryuchkov: A topic's full name derives from an SSV fork and an Ethereum network now, so I preferred using full names in `topicsCtrl`, rather than adding extra logic for handling base names. Actually, `topicsCtrl` used a lot of conversions between base and full names, and it would be easier to use full names in it even just for the Alan code

iurii-ssv: Yeah I agree it's better to use full topic name, unless we are printing it out in logs or something and make it short for brevity,

I just wanted to make sure this is not a breaking change

network/p2p/p2p.go
nkryuchkov: It used to be in the code, and I tried to avoid changing what we don't have to change in this PR. It appears in the change diff because we have two loops for each fork now. I'll probably try to check if we can join the loops

message/validation/validation.go
nkryuchkov: I tried to avoid changing what we don't have to change in this PR, but anyway, there are many changes related to topic IDs, so maybe it's worth changing. However, I think if we do that, we could also try to make the base/short name just a number because it doesn't have to be a string, but then it will inflate the PR. So the current state looked like a good tradeoff to me. I will think more about this, though

network/p2p/p2p_pubsub.go
nkryuchkov: It seems I forgot to adjust the name after making changes to this method. `subscribedSubnetsForCurrentEpoch` looks indeed better

network/p2p/p2p.go
nkryuchkov: I wanted them to match `registeredSubnets` and tried to avoid extra changes. Maybe it is worth doing, though

network/p2p/p2p.go
nkryuchkov: Maybe it should be `currentSubnets` then? It would also match `n.currentSubnets`

nkryuchkov:
@iurii-ssv 

> * BUT we are using the pre-Boole-fork rule-set when validating these early arrivals (seems like [we should be using post-Boole-fork rule-set instead](https://github.com/ssvlabs/SIPs/pull/43#discussion_r2712783799)) ?

Well, the only rule change in this PR is topics, and we allow both topics on special epochs. Checks in other PRs should check the message slot to determine which fork it is for

> * as for the "late arrivals" for the messages on the old topic, we are currently ignoring them with `ErrIncorrectTopic` (as per `Rule: Check if message was sent in the correct topic`) ... but maybe we'd wanna add [a "grace period" to process some of them](https://github.com/ssvlabs/SIPs/pull/43#discussion_r2712483250) ... seems like SIP suggests to keep processing old messages for exactly 1 additional slot after the `ForkEpoch` (and ofc we'd need to use pre-Boole-fork validation rules to validate those messages)

Yeah, it's been recently added and hasn't been applied yet

> * we'll need to re-open this PR against `[boole-fork](https://github.com/ssvlabs/ssv/tree/boole-fork)` branch I think ?

I'll just change the base branch

network/p2p/p2p.go
iurii-ssv: Yeah, sure

network/p2p/p2p.go
iurii-ssv: Right, but since we are changing quite a bit of code in this **func** I think we should refactor it as well

message/validation/validation.go
iurii-ssv: I'm mostly confused by the `TopicID` part (there are other funcs like `CommitteeTopicID` we should rename I think)  - either `base` or `short` would better reflect what it actually is

network/commons/subnets.go
iurii-ssv: Okay, `BaseName` is also fine, just wanted:
- it to be consistent everywhere - see https://github.com/ssvlabs/ssv/pull/2302#discussion_r2716327655
- to remove the `Get` prefix so the name follows the same convention as `func TopicFullName` (there is no Get prefix there)

network/commons/subnets.go
```diff
@@ -37,37 +42,87 @@ func SubnetTopicID(subnet uint64) string {
...
+	return topicName
 }
 
-// CommitteeSubnet returns the subnet for the given committee
-func CommitteeSubnet(cid spectypes.CommitteeID) uint64 {
-	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), bigIntSubnetsCount)
-	return subnet.Uint64()
+var bigIntPool = sync.Pool{
+	New: func() any { return new(big.Int) },
 }
```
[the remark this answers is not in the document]
iurii-ssv: Yeah, that could backfire I guess since `new(big.Int)` is supposed to represent `0` I think ? And with our usage of `sync.Pool` it might not

iurii-ssv:
> @iurii-ssv
> 
> > * BUT we are using the pre-Boole-fork rule-set when validating these early arrivals (seems like [we should be using post-Boole-fork rule-set instead](https://github.com/ssvlabs/SIPs/pull/43#discussion_r2712783799)) ?
> 
> Well, the only rule change in this PR is topics, and we allow both topics on special epochs. Checks in other PRs should check the message slot to determine which fork it is for

Yeah, the important part is to use the **message slot**, not the **current slot** when validating the message - and that should result in us applying the correct rules both for "early arrivals on the **new** topic" and "late arrivals on the **old** topic".

Which I think https://github.com/ssvlabs/ssv/pull/2503 is already doing it like that, so it should be fine.

network/commons/subnets.go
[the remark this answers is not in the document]
nkryuchkov: @iurii-ssv zeroing it there adds redundant allocations. And we overwrite it after pulling from the pool. I agree it's a bit error-prone, though. I will probably remove the pool because I added caching

nkryuchkov:
> the important part is to use the **message slot**, not the **current slot** when validating the message - and that should result in us applying the correct rules both for "early arrivals on the **new** topic" and "late arrivals on the **old** topic".

Well, yes, and the only place we don't do that is the topic check, because we have special rules in spec for that. But probably checking the message slots for topics can allow us to drop the special cases

network/commons/subnets.go
nkryuchkov: I decided to remove the base topic name. It's just a number converted to a string. With the changes in this PR, it only makes things more complex to understand. It doesn't add any simplicity.

@iurii-ssv please review https://github.com/ssvlabs/ssv/pull/2302/commits/8384871011673b2e8da1dfd3f3773d65d3c22e2f

message/validation/validation.go
nkryuchkov: removed, please check https://github.com/ssvlabs/ssv/pull/2302/#discussion_r2716452559

network/commons/subnets.go
iurii-ssv: Yeah we need to clean this up 👍

network/commons/subnets.go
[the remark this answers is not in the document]
nkryuchkov: removed

iurii-ssv (commented):
LGTM, seems the only thing left to implement is [`SUBSEQUENT_WINDOW` recently added to SIP](https://github.com/ssvlabs/SIPs/pull/43/changes)

nkryuchkov:
> seems the only thing left to implement is [`SUBSEQUENT_WINDOW` recently added to SIP](https://github.com/ssvlabs/SIPs/pull/43/changes)

Yeah, we need to reach a consensus on it in the SIP

nkryuchkov:
@greptileai please review the latest changes

network/p2p/p2p_pubsub.go
```diff
@@ -51,23 +51,33 @@ func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedS
...
+		committeeID := spectypes.CommitteeID(msg.SSVMessage.MsgID.GetDutyExecutorID()[16:])
+		if n.cfg.NetworkConfig.BooleFork() {
+			val, exists := n.nodeStorage.ValidatorStore().Committee(committeeID)
+			if !exists {
+				return fmt.Errorf("could not find share for validator %s", hex.EncodeToString(msg.SSVMessage.MsgID.GetDutyExecutorID()))
+			}
+			topic = val.BooleSubnet.BooleTopic(n.cfg.NetworkConfig.Beacon.Name)
+		} else {
+			topic = commons.AlanCommitteeSubnet(committeeID).AlanTopic()
+		}
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai It's intended. On fork, we stop broadcasting on old topics. Then we have one slot after fork to receive all messages from old topics

networkconfig/network.go
```diff
@@ -30,3 +34,40 @@ func (n Network) StorageName() string {
...
+func (n Network) inBoolePriorWindow(slot phase0.Slot) bool {
+	if n.BooleForkAtSlot(slot) {
+		return false
+	}
+	epoch := n.EstimatedEpochAtSlot(slot)
+	booleEpoch := n.SSV.Forks.Boole
+	if booleEpoch <= boolePriorWindowEpochs {
+		return true
+	}
+	return epoch >= booleEpoch-boolePriorWindowEpochs
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai fixed, please check it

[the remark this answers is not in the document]
nkryuchkov: @greptileai fixed, please check it again

nkryuchkov:

network/commons/subnets.go
```diff
@@ -18,56 +22,82 @@ const (
...
+// Topics returns the available Alan and Boole topics for the given network.
+func Topics(netCfg *networkconfig.Network) []string {
+	topics := make([]string, 0, SubnetsCount*2)
+	for subnet := Subnet(0); subnet < Subnet(SubnetsCount); subnet++ {
+		if !netCfg.BooleFork() {
+			topics = append(topics, subnet.AlanTopic())
+		}
+		topics = append(topics, subnet.BooleTopic(netCfg.Beacon.Name))
 	}
 	return topics
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai It's a one-time startup calculation, so I think we need both pre-fork and only Boole post-fork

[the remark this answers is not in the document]
nkryuchkov: @greptileai `math.MaxUint64` is technically a valid fork epoch, why do we need a special case for it? Why isn't the current version good enough

networkconfig/network.go
```diff
@@ -23,7 +23,7 @@ func (n Network) String() string {
 
 const alanForkName = "alan"
 const boolePriorWindowEpochs = phase0.Epoch(1)
-const booleUnsubscriptionWindowSlots = phase0.Slot(1)
+const booleSubsequentWindowSlots = phase0.Slot(1)
```
iurii-ssv: `boolePriorWindowEpochs` and `booleSubsequentWindowSlots` could use brief comments explaining "what these are for"

networkconfig/network.go
```diff
@@ -30,3 +35,58 @@ func (n Network) StorageName() string {
...
+}
+
+func (n Network) inBoolePriorWindow(slot phase0.Slot) bool {
+	if n.BooleForkAtSlot(slot) {
+		return false
+	}
+	booleEpoch := n.SSV.Forks.Boole
+	if booleEpoch == 0 {
+		return false
+	}
```
iurii-ssv: This block of code is covered by `if n.BooleForkAtSlot(slot)` check above already, so I would delete it:
```go
	booleEpoch := n.SSV.Forks.Boole
	if booleEpoch == 0 {
		return false
	}
```
Actually, the entire `inBoolePriorWindow` func looks way more complex than it should be ... can't we just do this ? `TestBooleForkInPriorWindow` doesn't complain
```
	if n.BooleForkAtSlot(slot) {
		return false
	}

	priorWindowStartEpoch := phase0.Epoch(0)
	if boolePriorWindowEpochs <= n.SSV.Forks.Boole {
		priorWindowStartEpoch = n.SSV.Forks.Boole - boolePriorWindowEpochs
	}

	return n.EstimatedEpochAtSlot(slot) >= priorWindowStartEpoch
```

networkconfig/network.go
```diff
@@ -30,3 +35,58 @@ func (n Network) StorageName() string {
...
+	if n.SSV.Forks.Boole > maxEpoch {
+		return false
+	}
+	start := n.FirstSlotAtEpoch(n.SSV.Forks.Boole)
+	if start > phase0.Slot(math.MaxUint64)-booleSubsequentWindowSlots {
+		return false
+	}
+	end := start + booleSubsequentWindowSlots
+	return slot >= start && slot < end
+}
```
iurii-ssv, line 87: Can't we simplify `inBooleSubsequentWindow` down to:
```go
	// Check if the Boole fork is well in the past (signified by 0 value) and if so - the "subsequent window" is
	// no longer even relevant (hence return false).
	if n.SSV.Forks.Boole == 0 {
		return false
	}

	if !n.BooleForkAtSlot(slot) {
		return false
	}

	return slot < (n.FirstSlotAtEpoch(n.SSV.Forks.Boole) + booleSubsequentWindowSlots)
```

networkconfig/network.go
nkryuchkov: I wanted to make it explicit that if `n.SSV.Forks.Boole` or `boolePriorWindowEpochs` is 0, then there is no prior epoch. Perhaps it's not necessary, though

networkconfig/network.go
nkryuchkov: If `n.SSV.Forks.Boole == phase0.Epoch(math.MaxUint64)`, and if we remove this check, then we will still try to calculate the subsequent window, which becomes out of bounds, so I think we need it.

Regarding the `maxEpoch` check, we have to work with slots, not epochs here, so any value beyond `phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)` (maybe except `math.MaxUint64` if we have a special case for it) will overflow in `FirstSlotAtEpoch` when multiplying by `b.SlotsPerEpoch`

`booleSubsequentWindowSlots` is IMO good for clarity, especially given that we have a lot of guards here, so it allows avoiding calculating what would happen in this case

nkryuchkov: I agree that the function became somewhat complex, but it clearly lists the guards. If we had a complex one-liner, it would require thinking about whether we missed any edge cases. Although we have unit tests. Well, I'll think if we can make this better

iurii-ssv: > If `n.SSV.Forks.Boole == phase0.Epoch(math.MaxUint64)`, and if we remove this check, then we will still try to calculate the subsequent window, which becomes out of bounds, so I think we need it.
> 
> Regarding the `maxEpoch` check, we have to work with slots, not epochs here, so any value beyond `phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)` (maybe except `math.MaxUint64` if we have a special case for it) will overflow in `FirstSlotAtEpoch` when multiplying by `b.SlotsPerEpoch`

The unit-tests don't cover any of this though, if we are worried about those overflows we should probably add corresponding tests for them.

Besides tests, I guess my main "complain" about all those checks is that it's hard to figure out why each of those is needed ... so I would maybe add a documenting comment for each of those that are not obvious (eg. the comment I sketched for the `n.SSV.Forks.Boole == 0` check).

nkryuchkov: > The unit-tests don't cover any of this though, if we are worried about those overflows we should probably add corresponding tests for them.

Yeah, I'm already adding more cases. I'll see what we can remove afterwards

networkconfig/network.go
iurii-ssv: > I wanted to make it explicit that if `n.SSV.Forks.Boole` or `boolePriorWindowEpochs` is 0, then there is no prior epoch. Perhaps it's not necessary, though

It's just not super clear why we need to treat these as "edge-cases" if they simply fit within the "general formula",

it's fine to handle those explicitly - but then I would expect to have a clarifying comment for every such handling (so you don't need to guess why it is handled separately)

cli/operator/node.go
```diff
@@ -623,12 +624,18 @@ var StartNodeCmd = &cobra.Command{
...
 			mySubnets := networkcommons.Subnets{}
 			myActiveSubnets := 0
+			booleFork := networkConfig.BooleForkAtSlot(networkConfig.EstimatedCurrentSlot())
 			for _, v := range myValidators {
-				subnet := networkcommons.CommitteeSubnet(v.CommitteeID())
-				if !mySubnets.IsSet(subnet) {
+				if subnet := v.BooleCommitteeSubnet(); !mySubnets.IsSet(subnet) {
 					mySubnets.Set(subnet)
 					myActiveSubnets++
 				}
```
y0sher, line 632: The SIP suggests starting to listen to the Boole subnet when we are in the prior window. Here we always subscribe to them right away at node start, is this intentional?

nkryuchkov: @y0sher This code calculates `myActiveSubnets`, used to calculate `idealMaxPeers`. It's set only once at the start, and it's quite complex to make it dynamic. So, I decided to use `Alan ∩ Boole` subnets as ideal max peers for starting pre-fork and Boole subnets for starting post-fork

network/p2p/p2p_pubsub.go
```diff
@@ -51,23 +51,33 @@ func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedS
 		return fmt.Errorf("could not encode signed ssv message: %w", err)
 	}
 
-	var topics []string
-
+	var topic string
 	if msg.SSVMessage.MsgID.GetRoleType() == spectypes.RoleCommittee {
-		topics = commons.CommitteeTopicID(spectypes.CommitteeID(msg.SSVMessage.MsgID.GetDutyExecutorID()[16:]))
+		committeeID := spectypes.CommitteeID(msg.SSVMessage.MsgID.GetDutyExecutorID()[16:])
+		if n.cfg.NetworkConfig.BooleFork() {
```
olegshmuelov: SIP-43 says "topic derived from its slot/fork."

Currently:
- Broadcast uses BooleFork() (current time)
- Validation uses message's Height (message slot)

Is this intentional? Edge case: if a pre-Boole duty completes just after fork,
broadcast picks Boole topic but validation expects Alan based on message slot.

Unlikely in practice (duties complete within slot), but wanted to confirm the
intended behavior.

cc @MatheusFranco99 @GalRogozinski

MatheusFranco99: Agree, I think the msg's slot can be used for deciding the topic for broadcasting

nkryuchkov: @olegshmuelov you're right, good suggestion. Applied, please check https://github.com/ssvlabs/ssv/pull/2302/commits/4e7ec51db058696fedfca48dfbf59be2d8f52761

PR #2304, opened 2025-06-19
netowkrconfig: use pointer receivers (to avoid lots of copies)

iurii-ssv:
I've noticed that `NetworkConfig` **struct** is quite large in size (and might grow further still) while at the same time it uses value-method-receivers,

while I haven't done any measuring, from what I know about Go this results in lots of unnecessary copying every time such methods are called (as opposed to when they are called by pointer-receiver),

so, unless there is a specific reason not to - we probably better off using pointer-receiver for structs with a bunch of fields (maybe 5+) for perf/memory reasons

_Before merging:_
- [x] test on **stage**

### DIFF
--- a/networkconfig/ssv.go
+++ b/networkconfig/ssv.go
@@ -137,6 +137,6 @@ func (s *SSVConfig) GetDomainType() spectypes.DomainType {
 	return s.DomainType
 }
 
-func (s SSVConfig) GetGasLimit36Epoch() phase0.Epoch {
+func (s *SSVConfig) GetGasLimit36Epoch() phase0.Epoch {
 	return s.GasLimit36Epoch
 }


iurii-ssv:
> Also, a potential drawback of these changes is the possibility of unintended mutations through the pointer, which wouldn’t be an issue if each object received its own copy of the configuration.
> Not sure how others feel about it

True, however I'd say it's unlikely to result in bugs in practice since all struct-fields are typed (so not only you'd need to assign something to it "by accident" but also you'd need to supply an appropriate type for such a change to be "not super noticeable")

nkryuchkov (approved):

y0sher:
@iurii-ssv this is failing UT and has conflicts

iurii-ssv:
> @iurii-ssv this is failing UT and has conflicts

Yeah, I mentioned above that I wanted to get some positive feedback on this before committing to finishing this PR off (so we can actually merge it), pipelines are fixed now @y0sher

nkryuchkov (approved):

oleg-ssvlabs (approved):

PR #2307, opened 2025-06-19
gas-limit: default to 36

iurii-ssv:
Depends on:
- (no longer needed) https://github.com/ssvlabs/ssv-spec/pull/553 (maybe actually we'd want to upgrade `ssv-spec` lib version later on due to https://github.com/ssvlabs/ssv/pull/2307#discussion_r2164306063 and leave a **TODO** for it in this PR instead)
- https://github.com/ssvlabs/ssv/pull/2308
- (no longer needed) https://github.com/ssvlabs/ssv/pull/2304 so we can change `GetGasLimit36Epoch` to be a pointer-receiver

_Before merging:_
- [x] (will do this in https://github.com/ssvlabs/ssv/pull/2304) change `GetGasLimit36Epoch` to be a pointer-receiver (once https://github.com/ssvlabs/ssv/pull/2304 is merged)
- [x] resolve TODOs
- [x] test on **stage**

### DIFF
--- a/protocol/v2/ssv/testing/runner.go
+++ b/protocol/v2/ssv/testing/runner.go
@@ -182,7 +182,7 @@ var ConstructBaseRunner = func(
 			net,
 			km,
 			opSigner,
-			validator.DefaultGasLimit,
+			validator.DefaultGasLimitOld,
 		)
 	case spectypes.RoleVoluntaryExit:
 		r, err = runner.NewVoluntaryExitRunner(
@@ -434,7 +434,7 @@ var ConstructBaseRunnerWithShareMap = func(
 			net,
 			km,
 			opSigner,
-			validator.DefaultGasLimit,
+			validator.DefaultGasLimitOld,
 		)
 	case spectypes.RoleVoluntaryExit:
 		r, err = runner.NewVoluntaryExitRunner(


networkconfig/beacon.go
```diff
@@ -50,10 +51,13 @@ type BeaconConfig struct {
 	TargetAggregatorsPerSyncSubcommittee uint64
 	TargetAggregatorsPerCommittee        uint64
 	IntervalsPerSlot                     uint64
-	GenesisForkVersion                   phase0.Version
-	GenesisTime                          time.Time
-	GenesisValidatorsRoot                phase0.Root
-	Forks                                map[spec.DataVersion]phase0.Fork
+	// GasLimit36Epoch is an epoch when to upgrade from default gas limit value of 30_000_000
+	// to 36_000_000.
+	GasLimit36Epoch       phase0.Epoch
```
y0sher: Does it have to be part of BeaconConfig? I see it more as an SSV thing. I'd rather it'd be part of the SSV config.

protocol/v2/ssv/validator/opts.go
```diff
@@ -42,12 +49,75 @@ type Options struct {
...
+		result.QueueSize = max(result.QueueSize, historySyncBatchSize*2)
+	}
+
+	if result.GasLimit == 0 {
+		// TODO: we probably want to define DefaultGasLimit36 const in spectypes package next to DefaultGasLimit,
+		// the question is whether 36_000_000 should override the current value of TODO (which is 30_000_000) or
+		// we should keep DefaultGasLimit defined as 30_000_000 while introducing new DefaultGasLimit36 = 36_000_000
+		// const instead.
+		// TODO: Note also, spec-tests rely on this value being set to 30_000_000 (for the runner under test),
+		// hence we'll need to address that as well for spec-tests to pass
```
y0sher: we could probably change spec/spec-tests for this case if needed and it could make it easier.

networkconfig/beacon.go
iurii-ssv: Yep, I'll move it there.

networkconfig/mainnet.go
```diff
@@ -29,4 +29,6 @@ var MainnetSSV = SSVConfig{
 		"enr:-Li4QH7FwJcL8gJj0zHAITXqghMkG-A5bfWh2-3Q7vosy9D1BS8HZk-1ITuhK_rfzG3v_UtBDI6uNJZWpdcWfrQFCxKGAYnQ1DRCh2F0dG5ldHOIAAAAAAAAAACEZXRoMpD1pf1CAAAAAP__________gmlkgnY0gmlwhBLb3g2Jc2VjcDI1NmsxoQKeSDcZWSaY9FC723E9yYX1Li18bswhLNlxBZdLfgOKp4N0Y3CCE4mDdWRwgg-h",
 	},
 	TotalEthereumValidators: 1064860, // active_validators from https://mainnet.beaconcha.in/index/data on Apr 18, 2025
+	// TODO - set proper value for mainnet
+	GasLimit36Epoch: 0,
```
iurii-ssv: I guess we should merge this PR until we have a definitive value for it (for when the switch to gas-limit-36 will happen), @y0sher could you advise on this ?

Also, I assume all other networks will have it set at 0 (so the change will take place immediately) ?

networkconfig/beacon.go
iurii-ssv: Done, the only TODO left in that regard is WHEN (what epoch) this value should be for **mainnet** - https://github.com/ssvlabs/ssv/pull/2307/files#r2163611539

networkconfig/mainnet.go
y0sher: yeah with stage we can do 0 and QA team will change this value to test. then we set a final value before we release (about 2 week in the future)

protocol/v2/ssv/validator/opts.go
```diff
@@ -89,8 +89,20 @@ func NewCommonOptions(
...
+		// - we've fixed spec-tests (in ssv-spec repo)
+		// - https://github.com/ssvlabs/ssv-spec/pull/553 is merged
+		// we can switch to using spectypes.DefaultGasLimit value instead of this hardcoded constant
+		defaultGasLimit := uint64(36_000_000)
+		//defaultGasLimit := uint64(spectypes.DefaultGasLimit)
+		if result.NetworkConfig.EstimatedCurrentEpoch() < result.NetworkConfig.GetGasLimit36Epoch() {
+			const oldDefaultGasLimit = uint64(30_000_000)
+			defaultGasLimit = oldDefaultGasLimit
+		}
+		result.GasLimit = defaultGasLimit
```
iurii-ssv: It seems something is `ssv-spec` repo itself has broken recently, when I try to upgrade from `v1.1.3` to https://github.com/ssvlabs/ssv-spec/pull/553 spec tests start failing in strange ways (meaning it doesn't seems anything to do with https://github.com/ssvlabs/ssv-spec/pull/553),

leaving this **TODO** here until we can resolve it (it probably shouldn't block this PR)

protocol/v2/ssv/validator/opts.go
iurii-ssv: Seems like they were failing due to `ssv-spec` lib update itself, rather than anything else - we'd need to figure this out (but maybe not until we finish & merge this PR) - see more on this https://github.com/ssvlabs/ssv/pull/2307#discussion_r2164306063

protocol/v2/ssv/validator/opts.go
y0sher: lets use the spec tag instead your pr for now and not relay on their const, it's best to not relay on them and let them update the spec on their own pace.

iurii-ssv: Okay, resolved in https://github.com/ssvlabs/ssv/pull/2307/commits/fa6bfec5838fa29fc5f23a08f32247e4046edd1f then

networkconfig/mainnet.go
iurii-ssv: I've set **mainnet** value to `math.MaxUint64` - https://github.com/ssvlabs/ssv/pull/2307/commits/7ec9d1947598cee5cb5cf7e6c31b0b995e301945 - just to make absolutely sure we don't accidentally ship this change forgetting to update it

y0sher (approved):

networkconfig/ssv.go
```diff
@@ -34,15 +35,21 @@ func GetSSVConfigByName(name string) (SSVConfig, error) {
...
+	RegistrySyncOffset   *big.Int
+	RegistryContractAddr ethcommon.Address
+	Bootnodes            []string
+	DiscoveryProtocolID  [6]byte
+	// TotalEthereumValidators value needs to be maintained — consider getting it from external API
+	// with default or per-network value(s) as fallback
+	TotalEthereumValidators int
+	// GasLimit36Epoch is an epoch when to upgrade from default gas limit value of 30_000_000
+	// to 36_000_000.
+	GasLimit36Epoch phase0.Epoch
```
nkryuchkov, line 52: I'm not sure if it's worth doing in this PR, but what do you think about using the `Forks SSVForkConfig` field from https://github.com/ssvlabs/ssv/pull/2184 and adding this fork there?

nkryuchkov (approved):

networkconfig/ssv.go
iurii-ssv: Yeah it seems like a fitting place for it, however I think we'll merge this PR before we merge https://github.com/ssvlabs/ssv/pull/2184

so I think it'll be up to @kchojn to move it there in https://github.com/ssvlabs/ssv/pull/2184

PR #2323, opened 2025-06-24
fix(observability) - fix wrong logger initialization order, add more logs

oleg-ssvlabs:


observability/observability.go
```diff
@@ -25,11 +25,13 @@ var (
 	defaultTracerProvider = trace_noop.NewTracerProvider()
 )
 
-func Initialize(ctx context.Context, appName, appVersion string, options ...Option) (shutdown func(context.Context) error, err error) {
+func Initialize(ctx context.Context, appName, appVersion string, l *zap.Logger, options ...Option) (shutdown func(context.Context) error, err error) {
```
nkryuchkov, line 28: Can we pass logger as optional?

nkryuchkov (dismissed):

observability/observability.go
oleg-ssvlabs: Do you mean adding new Option `WithLogger()` somewhere [here](https://github.com/ssvlabs/ssv/blob/stage/observability/option.go#L5)?

nkryuchkov: @oleg-ssvlabs yes, something similar

observability/logger.go
```diff
@@ -2,4 +2,12 @@ package observability
 
 import "go.uber.org/zap"
 
-var logger = zap.L().Named("Observability")
+var logger *zap.Logger
+
+func initLogger(l *zap.Logger) *zap.Logger {
```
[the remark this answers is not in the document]
oleg-ssvlabs: it's a good point, but It think it is slightly unnecessary for such "low scope" variable as observability logger

observability/observability.go
oleg-ssvlabs: This package is mainly responsible for configuring observability for the entire application - currently Metrics and Traces.  
It would make sense to introduce a `WithLogger()` option if it were meant to configure the logger for the entire application, not just the _internal_ logger used within the observability package.

We actually have a task planned for the 2.3.5 release where we aim to move the central logger configurator under the `observability` package - then we will add a `WithLogger()` option. @nkryuchkov

nkryuchkov: okay, sounds good to me; resolving

nkryuchkov (approved):

PR #2328, opened 2025-06-25
feat(e2e): add `SSV-Signer` slashing protection tests [main]

olegshmuelov:
Ports E2E slashing protection tests from stage branch (PR #2294).

closes https://github.com/ssvlabs/ssv/issues/2336

### DIFF
--- a/ssvsigner/web3signer/web3signer_test.go
+++ b/ssvsigner/web3signer/web3signer_test.go
@@ -2,7 +2,6 @@ package web3signer
 
 import (
 	"bytes"
-	"context"
 	"crypto/tls"
 	"encoding/hex"
 	"encoding/json"


ssvsigner/cmd/internal/validation/url.go
```diff
@@ -71,20 +71,46 @@ func init() {
...
 //   - Special-use addresses (e.g., http://0.0.0.0, multicast, broadcast)
 //   - Non-HTTP(S) schemes (e.g., file://, ftp://)
 //
+// testTrustedNetworks: comma-separated list of CIDR ranges to allow for testing ONLY.
+// This parameter bypasses SSRF protection for specified networks and should NEVER be used in production.
+// Example: "172.17.0.2/32,10.0.0.5/32" (prefer specific IPs over broad ranges)
+//
 // Returns an error if the endpoint is invalid or potentially unsafe, nil otherwise.
-func ValidateWeb3SignerEndpoint(endpoint string) error {
+func ValidateWeb3SignerEndpoint(endpoint string, testTrustedNetworks string) error {
```
y0sher: I'm not sure if this is the best approach, I understand we need to accept private IPs in e2e testing, we could've just used the `allow-insecure` flag but it will mean we are not testing the TLS impl at all.

Makefile
```diff
@@ -61,7 +61,7 @@ unit-test:
 .PHONY: ssvsigner-test
 ssvsigner-test:
 	@echo "Running ssv-signer unit tests"
-	@cd ssvsigner && go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 ./...
+	@cd ssvsigner && go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 `go list ./... | grep -ve "ssvsigner/e2e"`
```
iurii-ssv, line 64: Wouldn't this result in `make ssvsigner-test` running e2e tests only (no unit tests) ? Applies for https://github.com/ssvlabs/ssv/pull/2294 I guess

ssvsigner/cmd/internal/validation/url.go
iurii-ssv: Perhaps we could upgrade `ALLOW_INSECURE_HTTP` behavior such that with `ALLOW_INSECURE_HTTP = true` it would ALSO allow for full TLS configuration (instead of `return fmt.Errorf("insecure mode is enabled, but all TLS options are provided, please disable the insecure mode")` like it currently does) **but** will simply omit some IP address checks that are causing issues for tests in `validation.ValidateWeb3SignerEndpoint` ?

If not, I think it's probably better to do something along the lines of what **olegshmuelov** is doing here since it's probably the best way we could test ssvsigner with TLS.

Also, I don't think we've added this to https://github.com/ssvlabs/ssv/pull/2294 - reminding just in case

iurii-ssv (commented):
Just 1 question (+ we need to decide on https://github.com/ssvlabs/ssv/pull/2328#discussion_r2166782167 as well), overall looks great

ssvsigner/cmd/internal/validation/url.go
kchojn: Please check this: https://github.com/ssvlabs/ssv/pull/2333
its already merged into main

Makefile
olegshmuelov: The design is intentional and works as follows:

  - `make ssvsigner-test` runs unit tests for the ssvsigner package only (uses `grep -ve "ssvsigner/e2e"` to exclude E2E tests)
  - `make unit-test` runs all project unit tests AND then calls `make ssvsigner-test`
  - E2E tests are excluded because they require the built binary and special setup

  To run the ssvsigner E2E tests, you can use `make test-e2e` which properly builds the binary first and runs the E2E suite. There's also a dedicated GitHub workflow that runs these E2E tests in CI/CD.

iurii-ssv: Ah I see, didn't notice it actually **excludes** with `grep`, thanks for clarifying, makes sense

iurii-ssv (approved):
LGTM once https://github.com/ssvlabs/ssv/pull/2328#discussion_r2166782167 is addressed

iurii-ssv (approved):

ssvsigner/web3signer/web3signer_test.go
```diff
@@ -406,6 +406,53 @@ func TestSign(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			_, web3Signer := setupTestServer(t, tt.setupServer)
+
+			err := web3Signer.UpCheck(context.Background())
```
kchojn:
```suggestion
			err := web3Signer.UpCheck(t.Context())
```

kchojn (approved):
cool

kchojn (approved):

iurii-ssv (approved):

iurii-ssv:
@olegshmuelov not sure if it's already applied somewhere, but maybe we should add "closes https://github.com/ssvlabs/ssv/issues/2336" to PR description for auto-close as well

olegshmuelov:
Closing this PR in favor of continuing work on the original stage PR: https://github.com/ssvlabs/ssv/pull/2294

PR #2329, opened 2025-06-25
ssvsigner: fix parsing arguments

nkryuchkov:
`kong.Parse` was deleted in https://github.com/ssvlabs/ssv/pull/2279, so the config is currently not parsed

### DIFF
--- a/ssvsigner/cmd/purge-keys/purge-keys.go
+++ b/ssvsigner/cmd/purge-keys/purge-keys.go
@@ -36,12 +36,7 @@ type CLI struct {
 func main() {
 	var cli CLI
 
-	kong.Must(&cli,
-		kong.Name("purge-keys"),
-		kong.UsageOnError(),
-	)
-
-	_ = kong.Parse(&cli)
+	_ = kong.Parse(&cli, kong.Name("purge-keys"), kong.UsageOnError())
 
 	log, err := logger.SetupProductionLogger()
 	if err != nil {
--- a/ssvsigner/cmd/ssv-signer/ssv-signer.go
+++ b/ssvsigner/cmd/ssv-signer/ssv-signer.go
@@ -46,12 +46,7 @@ type CLI struct {
 func main() {
 	var cli CLI
 
-	kong.Must(&cli,
-		kong.Name("ssv-signer"),
-		kong.UsageOnError(),
-	)
-
-	_ = kong.Parse(&cli)
+	_ = kong.Parse(&cli, kong.Name("ssv-signer"), kong.UsageOnError())
 
 	log, err := logger.SetupLogger(cli.LogLevel, cli.LogFormat)
 	if err != nil {


y0sher (approved):

ssvsigner/cmd/purge-keys/purge-keys.go
```diff
@@ -35,10 +36,7 @@ type CLI struct {
 func main() {
 	var cli CLI
 
-	kong.Must(&cli,
-		kong.Name("purge-keys"),
-		kong.UsageOnError(),
-	)
+	_ = kong.Parse(&cli, kong.Name("purge-keys"), kong.UsageOnError())
```
olegshmuelov, line 39: you can just
```go
kong.Parse(&cli, kong.Name("purge-keys"), kong.UsageOnError())
```

iurii-ssv: TBH I think it's a bit better to do `_ =` (and advocated for by lots of Golang people) since if not - it's not clear whether it's ignored intentionally or we just missed/forgot about the return value

nkryuchkov: I prefer assigning variables always when a method returns something, even if we don't need it, to show explicitly that the variable is ignored

olegshmuelov: `kong.Parse` only returns a single `*kong.Context`, no error.
If we need the context we should bind it:
```go
ctx := kong.Parse(&cli, kong.Name("purge-keys"), kong.UsageOnError())
```
If we don’t use the context at all, the bare call is already explicit enough:
```go
kong.Parse(&cli, kong.Name("purge-keys"), kong.UsageOnError())
```
Assigning to _ doesn’t buy us anything in this case, because there’s nothing else to ignore.

olegshmuelov: Writing `_ = kong.Parse(...)` adds a token but zero clarity—the reader still has to know the function’s signature to see what was ignored.

Linters won’t complain about the bare call, since no error is being dropped.

nkryuchkov: @olegshmuelov assigning to _ hints us at the method's signature and that it returns something which might be useful if the code gets extended

I'd still prefer keeping `_ =`. Indeed, the benefit is kind of vague and barely exists, but IMO it isn't worse than removing it

iurii-ssv: > Linters won’t complain about the bare call, since no error is being dropped.

I guess with linters being advanced enough these days it's fine to omit things like `_ =` ... but also I think Goland IDE complains about this by default when you omit it. It probably doesn't matter much anyway :)

nkryuchkov: IMO, we can come to the conclusion that the difference is not very important here and it's rather a personal preference, so let's keep it as is for now. I'm resolving it. If someone has any objections, please unresolve it

PR #2335, opened 2025-06-26
fix(exporter): set gitflow branch to stage

julienh-ssv:
Change branch so that benchmark is only run on stage commits.

### DIFF
--- a/.github/workflows/dutytracer-bench.yml
+++ b/.github/workflows/dutytracer-bench.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - feat/exporter-v2
+      - stage
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/dutytracer-bench.yml
```diff
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - feat/exporter-v2
+      - stage
```
vaclav-ssvlabs, line 8: @MatusKysel 

```
on:
  pull_request:
    branches:
      - main
```

MatusKysel: @julienh-ssv please adjust it so it's running only for PRs to main

PR #2337, opened 2025-06-26
beacon/goclient: remove `nodeSyncingFn`

nkryuchkov:
We are currently using `nodeSyncingFn` only for mocks in tests. The PR implements syncing mocks by replacing data returned by the mock server.

Required for https://github.com/ssvlabs/ssv/pull/2339


beacon/goclient/goclient_test.go
```diff
@@ -19,50 +19,64 @@ import (
...
-		client.syncDistanceTolerance = 2
+			return rawData, nil
+		}
 
-		err = client.Healthy(t.Context())
-		require.ErrorIs(t, err, errSyncing)
+		return mockResponses[r.URL.Path], nil
 	})
+	c, err := mockClient(t.Context(), mockServer.URL, commonTimeout, longTimeout)
+	require.NoError(t, err)
```
iurii-ssv: Nit: not exactly the target of this PR, but maybe while at it we could rename `mockClient` to `newClient`  since it's actually a real client under-the-hood. It could also return `GoClient` to further simplify things, I think.

Plus, maybe we could also leverage "options" instead of `client.syncDistanceTolerance = syncDistanceTolerance` to actually test they are applied correctly (the more real production code we can re-use in tests the more valuable these tests will be)

beacon/goclient/goclient_test.go
```diff
@@ -19,50 +19,64 @@ import (
...
 
-		client.syncDistanceTolerance = 3
+	client.syncDistanceTolerance = syncDistanceTolerance
 
-		err = client.Healthy(t.Context())
-		require.NoError(t, err)
-	})
+	replaceSyncing.Store(true)
+
+	return client.Healthy(t.Context())
```
iurii-ssv: Would be nice to leave clarifying comments for:
- what `client.Healthy` calls (1st and 2nd) are testing/expecting
- and why we are using `replaceSyncing` to change the behavior

Splitting this into two separate tests would maybe make even more sense btw

iurii-ssv (dismissed):
Minor suggestions

beacon/goclient/goclient_test.go
nkryuchkov: I removed the second call, IMO it's redundant after removing `syncDistanceTolerance` assignment

beacon/goclient/goclient_test.go
nkryuchkov: I deleted it, it's not really that useful

beacon/goclient/goclient_test.go
nkryuchkov: added a comment about `replaceSyncing`

PR #2339, opened 2025-06-26
beacon/goclient: check if at least one client is healthy

nkryuchkov:
Depends on https://github.com/ssvlabs/ssv/pull/2337

Checks if at least one client is healthy instead of relying on `go-eth2-client` multi-client's `NodeSyncing` call, which seems to check only the current active client on exceeded deadline: https://github.com/attestantio/go-eth2-client/blob/master/multi/client.go#L181


olegshmuelov (commented):
GJ! To ensure the robustness of this new functionality, it would be helpful to add test coverage for the multi-client scenarios:

1. Multiple clients where all are healthy
2. Multiple clients where some are healthy and others are not
3. Multiple clients where all are unhealthy
4. Behavior when health checks timeout
5. Concurrent health checking across multiple clients

The existing tests only cover single-client scenarios, so adding these would give us confidence in the multi-client behavior.

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,89 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig networkco
...
+	if !hasHealthy.Load() {
+		return errs
+	}
+
+	return nil
+}
+
+// checkNodeHealthiness checks client's healthiness but checking if it's synced.
+// Errors are logged and metrics are recorded accordingly, so the caller doesn't need to do it.
+func (gc *GoClient) checkNodeHealthiness(ctx context.Context, client Client) error {
```
iurii-ssv: Maybe simply `checkNodeHealthiness` -> `checkNodeHealth`

and `but checking if it's synced.` -> `by checking if it's synced.` ?

Also, this part is kind of self-explanatory - I don't think we need a comment for it unless it's mentioning **why** we are doing it this way (but even then, in this case I don't think it's something that needs a comment)

> // Errors are logged and metrics are recorded accordingly, so the caller doesn't need to do it.

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,89 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig networkco
...
+	defer cancel()
+
+	for _, client := range gc.clients {
+		wg.Add(1)
+		go func(client Client) {
+			defer wg.Done()
+
+			err := gc.checkNodeHealthiness(ctx, client)
+
+			if err != nil {
```
iurii-ssv: No need for newline in between, I guess

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,89 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig networkco
...
+				errsMu.Lock()
+				errs = errors.Join(errs, err)
+				errsMu.Unlock()
+			} else {
+				if hasHealthy.CompareAndSwap(false, true) {
+					cancel() // one healthy node is enough
+				}
+			}
```
iurii-ssv: Nit: on error it would be simpler to just `return` right away (instead of using `else` here)

I think it's a sort of convention in Golang to keep the **main/success** code-path consistently aligned (so it's easy to read) while returning as soon as possible on **side/error** code-paths

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,89 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig networkco
...
 
-func (gc *GoClient) nodeSyncing(ctx context.Context, opts *api.NodeSyncingOpts) (*api.Response[*apiv1.SyncState], error) {
-	return gc.multiClient.NodeSyncing(ctx, opts)
-}
-
 var errSyncing = errors.New("syncing")
 
-// Healthy returns if beacon node is currently healthy: responds to requests, not in the syncing state, not optimistic
+// Healthy returns if the consensus client (for single client) or at least one of consensus clients (for multi client)
+// is currently healthy: responds to requests, not in the syncing state, not optimistic
```
iurii-ssv: I'm not sure I understand what this part means, maybe we could re-phrase it a bit:

> is currently healthy: responds to requests, not in the syncing state

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,89 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig networkco
...
+		}(client)
+	}
+
+	wg.Wait()
+
+	if !hasHealthy.Load() {
+		return errs
+	}
+
+	return nil
```
iurii-ssv: Maybe not worth rewriting this, but just note:

while `cancel()` should help with exiting-on-success - it's not the best to rely on it functioning properly, if the underlying code doesn't respect context-cancel we might still end up waiting (on `wg.Wait()`) for the slowest call

iurii-ssv (commented):
LGTM, left minor suggestions

beacon/goclient/goclient.go
nkryuchkov: Yes, it used to be outside of the goroutine so I didn't pay attention to it

beacon/goclient/goclient.go
nkryuchkov: I'd keep the optimistic part because it's a part of the logic and we have a link below with explanation what it means

nkryuchkov:
> GJ! To ensure the robustness of this new functionality, it would be helpful to add test coverage for the multi-client scenarios:
> 
> 1. Multiple clients where all are healthy
> 2. Multiple clients where some are healthy and others are not
> 3. Multiple clients where all are unhealthy
> 4. Behavior when health checks timeout
> 5. Concurrent health checking across multiple clients
> 
> The existing tests only cover single-client scenarios, so adding these would give us confidence in the multi-client behavior.

done

beacon/goclient/goclient_test.go
```diff
@@ -17,54 +22,352 @@ import (
...
+					SyncDistance: phase0.Slot(4),
+					IsSyncing:    true,
+				},
+			},
+		}
+
+		err := runHealthyTest(t, syncResponseList, 2, true)
+		require.ErrorIs(t, err, errSyncing)
+	})
+}
```
kchojn: i think it would be better to refactor this to testCases, something like this:
```go
testCases := []struct {
		name                  string
		syncResponseList      []syncResponse
		concurrentHealthCheck bool
		wantErr               error
	}{
```
less code duplication + more readable structure

kchojn (changes requested):

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,96 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig *networkc
...
+	errCh := make(chan error, len(gc.clients))
+
+	for _, client := range gc.clients {
+		go func() {
+			if err := gc.checkNodeHealth(ctx, client); err != nil {
+				select {
+				case errCh <- err:
+				default:
+				}
+				return
```
iurii-ssv: Since we have a buffered channel we could just send `errCh <- err` (without `select`), plus we probably don't want to drop errors in any case

beacon/goclient/goclient.go
```diff
@@ -416,53 +416,96 @@ func (gc *GoClient) applyBeaconConfig(nodeAddress string, beaconConfig *networkc
...
+	// Wait only for the result: either one success, or all errors
+	var errs error
+	for i := 0; i < len(gc.clients); i++ {
+		select {
+		case <-ctx.Done():
+			// If canceled, and we have a healthy client, return no error
+			if hasHealthy.Load() {
+				return nil
+			}
+			// Otherwise, collect errors and keep going
```
iurii-ssv: Nit: since `ctx` can be canceled due to parent-context as well - this part `If canceled, and we have a healthy client, return no error` isn't 100% correct

maybe rephrase to: `If we have a healthy client - return with no error. Otherwise, collect errors and keep going.`

iurii-ssv (approved):
Looks good! Just couple more nits from me.

beacon/goclient/goclient.go
iurii-ssv: On the first read I didn't quite get what "is currently healthy" refers to, something like this would help a bit:
```go
// Beacon node is healthy if: responds to requests, not in the syncing state, not optimistic
```

kchojn (dismissed):

iurii-ssv (approved):

iurii-ssv (dismissed):

kchojn (approved):

PR #2340, opened 2025-06-27
Run ProcessConsensus validator duty signing in separate goroutines

oleg-ssvlabs:


protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,116 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
 	version, _ := cr.BaseRunner.NetworkConfig.ForkAtEpoch(epoch)
 
 	span.SetAttributes(
 		observability.BeaconSlotAttribute(duty.DutySlot()),
 		observability.BeaconEpochAttribute(epoch),
 		observability.BeaconVersionAttribute(version),
+		observability.DutyCountAttribute(len(duty.(*spectypes.CommitteeDuty).ValidatorDuties)),
+	)
+
+	ctx, cancel := context.WithCancel(ctx)
```
oleg-ssvlabs, line 279: If an error occurs in one of the goroutines, the context will be canceled, which will [prevent](https://github.com/ssvlabs/ssv/pull/2340/files#diff-c615c5d14e307fe437330a4144eaa9c319e9f30cf23aef1c724d8712057b1868R292) further iteration over the collection of validator duties.

protocol/v2/ssv/runner/committee.go
```diff
@@ -243,11 +246,11 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 		return nil
 	}
 
+	span.AddEvent("instance is decided")
```
iurii-ssv, line 249: Do we need `checking if instance is decided` and `instance is decided` events ?

If we do, I think it's better to do it inside of `baseConsensusMsgProcessing` func (probably the simplest would be to check the returned value with `defer`) so that:
- it will apply to all the runners we have (since all runners are based on `BaseRunner`)
- `ProcessConsensus` is less cluttered with that helper-stuff

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,117 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-	blockedAttesterDuties := 0
-
 	epoch := cr.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(duty.DutySlot())
 	version, _ := cr.BaseRunner.NetworkConfig.ForkAtEpoch(epoch)
 
 	span.SetAttributes(
 		observability.BeaconSlotAttribute(duty.DutySlot()),
 		observability.BeaconEpochAttribute(epoch),
 		observability.BeaconVersionAttribute(version),
+		observability.DutyCountAttribute(len(duty.(*spectypes.CommitteeDuty).ValidatorDuties)),
```
iurii-ssv: Nit: since this type-cast is duplicated here it might be better to just create a separate variable of type `spectypes.CommitteeDuty` and just use it

similar to how we are doing it in `StartNewDuty`: https://github.com/ssvlabs/ssv/blob/0aff0461f8d5d434d65058afe028e8fde9400590/protocol/v2/ssv/runner/committee.go#L103-L106

protocol/v2/ssv/runner/committee.go
```diff
@@ -410,6 +425,66 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+		zap.String("pub_key", hex.EncodeToString(validatorDuty.PubKey[:])),
+		zap.Any("attestation_data", attestationData),
+		zap.String("attestation_data_root", hex.EncodeToString(attDataRoot[:])),
+		zap.String("signing_root", hex.EncodeToString(partialMsg.SigningRoot[:])),
+		zap.String("signature", hex.EncodeToString(partialMsg.PartialSignature[:])),
+	)
+
+	span.SetStatus(codes.Ok, "")
+
+	return isBlocked, partialMsg, nil
```
iurii-ssv: Nit: it's simpler to return `false` instead of `isBlocked` whenever possible (applies to 3 return statements in total I think)

protocol/v2/ssv/runner/committee.go
```diff
@@ -410,6 +425,66 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+	// Sync committee duties are not slashable, so they are always allowed.
+	if !cr.doppelgangerHandler.CanSign(validatorDuty.ValidatorIndex) {
+		const eventMsg = "Signing not permitted due to Doppelganger protection"
+		span.AddEvent(eventMsg)
+		logger.Warn(eventMsg, fields.ValidatorIndex(validatorDuty.ValidatorIndex))
+		isBlocked = true
+
+		span.SetStatus(codes.Ok, "")
+		return
+	}
```
iurii-ssv: Nit: I think it's easier to read and understand the code without "naked return statements" - meaning here we could just `return true, nil, nil` (mostly because this code will be read multiple times in the future, and every time the reader will have to "fill this in" by himself)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,117 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-			// Sync committee duties are not slashable, so they are always allowed.
-			if !cr.doppelgangerHandler.CanSign(validatorDuty.ValidatorIndex) {
-				const eventMsg = "Signing not permitted due to Doppelganger protection"
-				span.AddEvent(eventMsg, attr)
-				logger.Warn(eventMsg, fields.ValidatorIndex(validatorDuty.ValidatorIndex))
-				blockedAttesterDuties++
-				continue
+			if ctx.Err() != nil {
+				return
 			}
```
iurii-ssv: Maybe we should move this `if ctx.Err() != nil` check outside of `go ...` and check it even before we do `wg.Add(1)` breaking out of this loop entirely (so we don't spawn go-routines that will just exit straight away)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,117 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+				}
+
+				lock.Lock()
+				postConsensusMsg.Messages = append(postConsensusMsg.Messages, partialSigMsg)
+				lock.Unlock()
+
+			default:
+				errCh <- fmt.Errorf("invalid duty type: %s", validatorDuty.Type)
+				cancel()
 			}
```
iurii-ssv: I would add a `return` statement here (under `default` clause) for consistency, this might also save us from a potential bug if there is some code added later (in under this whole `switch` code-block)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,117 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+		wg.Wait()
+		close(doneCh)
+	}()
+
+	select {
+	case err := <-errCh:
+		return observability.Error(span, err)
+	case <-doneCh:
+		span.AddEvent("finished signing validator duties")
 	}
```
iurii-ssv: I think it is technically possible for `case <-doneCh` clause to execute even if we encountered some error (in some go-routine), meaning we would just drop/skip such an error and consider duty to be done successfully when in fact it wasn't successful ... this is because `select` doesn't guarantee the ordering for its `case` statements (I think it's randomized) ...

TBH I think it would be cleaner to get rid of `sync.WaitGroup` entirely and simply use 2 channels (`errCh`, `resultCh`) counting the number of received values on both (should sum up to `len(duty.(*spectypes.CommitteeDuty).ValidatorDuties)` in total) and erroring if necessary. This might also allow us to get rid of `lock  sync.Mutex` as well - we could just send & read the results sequentially on `resultCh`, wdyt ?

iurii-ssv (commented):
Good stuff, left some suggestions

protocol/v2/ssv/runner/committee.go
y0sher: I guess traces could tell it more accurately, but maybe this is the reason why we can't get big gains from parallelization? eventually doppleganger will lock for each of these validators and all the goroutines will basically wait on the same lock. Maybe we can try to implement checking a list of validators while locking one time in doppleganger and then referring to the results here? unless traces show clearly that his isn't an issue.

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: The `baseConsensusMsgProcessing` is not orchestrated with traces so far, probably mainly because we did not see a strong need for that (I assume you mean moving these events to `baseConsensusMsgProcessing` method). 

Sorry, I may not fully understand your suggestion. The helper-stuff you mean two events: "instance is not decided" and "instance is decided". I wouldn't say it's really cluttering the flow, not sure it's worth moving them to `baseConsensusMsgProcessing` method. Currently, they also serve as a breakpoints, so in the trace it will possible to see how long did it take to execute something in between events, so that's pretty useful.

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: good point. Fixed 2706920

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: agree, d543bb81

protocol/v2/ssv/runner/committee.go
iurii-ssv: I guess my main concern is that we are adding it here for `CommitteeRunner` but **not other runners**,

if we want to apply this to all the runners we have I think it would be better to do that in `baseConsensusMsgProcessing` func because every runner calls it ?

I'm not sure however whether we can fetch the relevant `span` from `ctx` (because we need that `span` in `baseConsensusMsgProcessing`, but only `ctx` is passed as an argument in there) ... if not - then we'll need to pass `span` as an explicit argument)

Anyway, I don't have a strong opinion on what's the best way to implement it, just thought it is worth pointing out

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: agree, eda7e9c

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: agree, 01acb81e7e92877dc5081631

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Added 8a208ebca281b9b2b6e

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I don't really think it's a bottleneck, because from what I can see, `CanSign` [uses read locks](https://github.com/ssvlabs/ssv/blob/stage/doppelganger/doppelganger.go#L92), which can be acquired by multiple goroutines at the same time.

Also, a 15% improvement is quite decent - it's a reduction in the total duty execution time, where Consensus is just one of the phases. I could probably also check how much faster the `ProcessConsensus` method became specifically.

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I see your point.

I think both options are viable and you are right, technically it is possible for the `done` channel to be closed after error published to the error channel and select statement catching the `done` channel closure first.

wdyt about this implementation 0385a9a638a51194e85525bdb32533ea00dd0955 ?

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I see what you mean.

At this point, I just added identical span events to the other runners. I know it's not very DRY, but we still need to instrument the flow in the runners themselves. Even if we instrument `baseConsensusMsgProcessing`, we would still need to start another span there. 

I don't think passing the span around is the idiomatic way to do it, at least based on what I know.

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +258,121 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+		select {
+		case err := <-errCh:
+			return observability.Error(span, err)
+		case signature := <-signaturesCh:
+			postConsensusMsg.Messages = append(postConsensusMsg.Messages, signature)
+			if dutiesProcessed.Load() == uint32(len(committeeDuty.ValidatorDuties)) { //nolint:gosec
+				break listen
+			}
 		}
 	}
```
iurii-ssv: Nit: maybe it's a matter of preference, but breakpoints (like that `listen` here) are a bit harder to read/follow vs for example explicit iteration variable:
```go
	for done := false; !done; {
		select {
		case err := <-errCh:
			return observability.Error(span, err)
		case signature := <-signaturesCh:
			postConsensusMsg.Messages = append(postConsensusMsg.Messages, signature)
			if dutiesProcessed.Load() == uint32(len(committeeDuty.ValidatorDuties)) { //nolint:gosec
				done = true
				continue
			}
		}
	}
```
Anyway, it's just how I see it, not a strong opinion, feel free to ignore

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +258,121 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-				spectypes.DomainSyncCommittee,
-			)
-			if err != nil {
-				return observability.Errorf(span, "failed signing sync committee message: %w", err)
+				signaturesCh <- partialSigMsg
+				dutiesProcessed.Add(1)
+			default:
+				errCh <- fmt.Errorf("invalid duty type: %s", validatorDuty.Type)
+				cancel()
+				return
```
iurii-ssv: It might be cleaner if the reader of `errCh` would call `cancel` in case an `error` shows up (instead of these go-routines calling `cancel` in multiple places)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +258,121 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
 
-		span.AddEvent("validating duty", attr)
-		if err := cr.DutyGuard.ValidDuty(validatorDuty.Type, spectypes.ValidatorPK(validatorDuty.PubKey), validatorDuty.DutySlot()); err != nil {
-			const eventMsg = "duty is no longer valid"
-			span.AddEvent(eventMsg, attr)
-			logger.Warn(eventMsg, fields.Validator(validatorDuty.PubKey[:]), fields.BeaconRole(validatorDuty.Type), zap.Error(err))
-			continue
+	var (
+		errCh        = make(chan error, 1)
+		signaturesCh = make(chan *spectypes.PartialSignatureMessage, len(committeeDuty.ValidatorDuties))
```
iurii-ssv: I think `errCh` should also be buffered with `len(committeeDuty.ValidatorDuties)` since multiple errors **might** be sent on it while "the listener" will only read the very 1st one

Currently in case of multiple errors we'll end up with some "hanging go-routines that are trying to send on the channel nobody reads anymore" I think (I doubt garbage-collector could automatically resolve it and terminate those go-routines, but I'm not 100% sure)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +258,121 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+listen:
+	for {
+		select {
+		case err := <-errCh:
+			return observability.Error(span, err)
+		case signature := <-signaturesCh:
+			postConsensusMsg.Messages = append(postConsensusMsg.Messages, signature)
+			if dutiesProcessed.Load() == uint32(len(committeeDuty.ValidatorDuties)) { //nolint:gosec
+				break listen
+			}
```
iurii-ssv: I don't think we need to define a separate `dutiesProcessed` variable (I think we define it just to check that condition here),

we could simply replace `dutiesProcessed.Load() == uint32(len(committeeDuty.ValidatorDuties))` -> `len(postConsensusMsg.Messages) == len(committeeDuty.ValidatorDuties)` ?

protocol/v2/ssv/runner/committee.go
iurii-ssv: Yes, that's roughly what I had in mind - I've went over the current code version and left some minor comments

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I had to introduce that variable because of early returns like [these](https://github.com/ssvlabs/ssv/commit/0385a9a638a51194e85525bdb32533ea00dd0955#r161268474)

iurii-ssv: Ah, I see, makes sense

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: good point, fixed 34707f0e37794922176e9a3c902e4a89c0b780e1

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: fixed here 34707f0e37794922176e9a3c902e4a89c0b780e1

Although, I kept the `signaturesCh` unbuffered, since having it buffered caused some occasional issues. These were related to the `done` channel being closed before all messages were consumed from the channel. 

Now, with a capacity of 0, it will behave in a more blocking manner, which should help prevent such issues.

iurii-ssv: Looks good now 👍 

Maybe we should leave a note for why we are defining `errCh        = make(chan error, len(committeeDuty.ValidatorDuties))` as buffered:
```go
// errCh is buffered because the receiver is only interested in the very 1st error on this channel and will not 
// bother reading all other errors sent on this channel - buffering allows senders to unblock and terminate 
// regardless of what the receiver does 
errCh        = make(chan error, len(committeeDuty.ValidatorDuties))
```

iurii-ssv (approved):
LGTM

iurii-ssv (approved):

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,123 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-			// Doppelganger protection applies only to attester duties since they are slashable.
-			// Sync committee duties are not slashable, so they are always allowed.
-			if !cr.doppelgangerHandler.CanSign(validatorDuty.ValidatorIndex) {
-				const eventMsg = "Signing not permitted due to Doppelganger protection"
-				span.AddEvent(eventMsg, attr)
-				logger.Warn(eventMsg, fields.ValidatorIndex(validatorDuty.ValidatorIndex))
-				blockedAttesterDuties++
-				continue
+		wg.Add(1)
+		go func(ctx context.Context, validatorDuty *spectypes.ValidatorDuty) {
```
MatusKysel: isn't it better to use some worker pool here ? as this could spawn a lot of go routines for reach duty

oleg-ssvlabs: I actually gave it some thought.

I’ve noticed there are two outgoing HTTP calls during the processing of each duty. One call is to get the domain data from the beacon node - this isn’t particularly problematic since `go-eth2` has locking implemented there. However, it’s still a bit concerning, if in a future version those locks are removed, it might lead to issues. The second HTTP call happens during signing, specifically when the key manager is set to be **remote**.

While the number of goroutines is not a problem - they are lightweight, and spawning thousands of them shouldn’t be an issue - making thousands of parallel HTTP requests can easily overwhelm the HTTP connections.

Based on this, I decided to implement a worker pool, with the number of workers currently set to `30`. It still delivers a performance improvement(499.5ms -> 440ms in entire duty execution duration), and gives us the ability to control parallelism

Commit - 4bfe9a0c2e97f9e69bac4f58aaece5db455232bb

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,141 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-				ctx,
-				cr,
-				validatorDuty,
-				spectypes.SSZBytes(blockRoot[:]),
-				validatorDuty.DutySlot(),
-				spectypes.DomainSyncCommittee,
-			)
-			if err != nil {
-				return observability.Errorf(span, "failed signing sync committee message: %w", err)
+				switch validatorDuty.Type {
```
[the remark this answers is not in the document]
oleg-ssvlabs: It’s a good suggestion, and I actually even attempted it. However, I ended up not doing it because, in the end, it didn’t reduce the number of lines in the case block.

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,141 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-
-			// Doppelganger protection applies only to attester duties since they are slashable.
-			// Sync committee duties are not slashable, so they are always allowed.
-			if !cr.doppelgangerHandler.CanSign(validatorDuty.ValidatorIndex) {
-				const eventMsg = "Signing not permitted due to Doppelganger protection"
-				span.AddEvent(eventMsg, attr)
-				logger.Warn(eventMsg, fields.ValidatorIndex(validatorDuty.ValidatorIndex))
-				blockedAttesterDuties++
-				continue
+	span.AddEvent("signing validator duties")
```
iurii-ssv: Nit: could maybe move this `span.AddEvent("signing validator duties")` line up a bit (right before `var (` section) - it looks kinda random to have it here

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,141 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-			span.AddEvent(eventMsg, attr)
+		go func(ctx context.Context) {
+			defer wg.Done()
 
-		case spectypes.BNRoleSyncCommittee:
-			totalSyncCommitteeDuties++
+			for validatorDuty := range dutiesCh {
+				if ctx.Err() != nil {
+					return
+				}
```
iurii-ssv, line 320: Checking context-cancel here feels a bit redundant - I would remove it for simplicity because it doesn't add much (it will catch very very small amount of cases, if any)

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,141 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-			logger.Debug(eventMsg,
-				zap.Uint64("validator_index", uint64(validatorDuty.ValidatorIndex)),
-				zap.String("pub_key", hex.EncodeToString(validatorDuty.PubKey[:])),
-				zap.Any("attestation_data", attestationData),
-				zap.String("attestation_data_root", hex.EncodeToString(attDataRoot[:])),
-				zap.String("signing_root", hex.EncodeToString(partialMsg.SigningRoot[:])),
-				zap.String("signature", hex.EncodeToString(partialMsg.PartialSignature[:])),
-			)
-			span.AddEvent(eventMsg, attr)
+		go func(ctx context.Context) {
```
iurii-ssv: Do we need to pass `ctx` as a func-argument ? I'm not sure why it would be different from other variables this func references "as closure"

iurii-ssv (dismissed):

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: yes, I agree. 
I think it's [this](https://github.com/ssvlabs/ssv/blob/41afd3e78abd820e5d7121108d363ca794303372/protocol/v2/ssv/runner/committee.go#L303) condition will primarily be catching the cancellation errors, because channel with workers will be blocking. Good point

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I agree, I think it's redundant here.

fixed 1e17bbffd785565d2242037a5da4117d5542365e

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: fixed 1e17bbffd785565d2242037a5da4117d5542365e

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: fixed 1e17bbffd785565d2242037a5da4117d5542365e (moved before cancellation context declaration)

iurii-ssv (dismissed):

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,138 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
-				attestationData,
-				validatorDuty.DutySlot(),
-				spectypes.DomainAttester,
-			)
-			if err != nil {
-				return observability.Errorf(span, "failed signing attestation data: %w", err)
+	go func() {
+		for _, duty := range committeeDuty.ValidatorDuties {
+			if ctx.Err() != nil {
+				break
```
nkryuchkov: `dutiesCh` isn't closed on context error, so I think `for validatorDuty := range dutiesCh {` would block in this case, waiting for the next value in the channel
```suggestion
	go func() {
		defer close(dutiesCh)
		for _, duty := range committeeDuty.ValidatorDuties {
			if ctx.Err() != nil {
				break
```

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,138 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+		close(dutiesCh)
+	}()
+
+	for range workerCount {
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			for validatorDuty := range dutiesCh {
```
nkryuchkov, line 317: The body looks quite heavy, so perhaps we could check for a context error here to avoid redundant actions?

protocol/v2/ssv/runner/committee.go
```diff
@@ -256,105 +259,138 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
...
+				if err := cr.DutyGuard.ValidDuty(validatorDuty.Type, spectypes.ValidatorPK(validatorDuty.PubKey), validatorDuty.DutySlot()); err != nil {
+					const eventMsg = "duty is no longer valid"
+					span.AddEvent(eventMsg, trace.WithAttributes(
+						observability.ValidatorIndexAttribute(validatorDuty.ValidatorIndex),
+						observability.ValidatorPublicKeyAttribute(validatorDuty.PubKey),
+						observability.BeaconRoleAttribute(validatorDuty.Type),
+					))
+					logger.Warn(eventMsg, fields.Validator(validatorDuty.PubKey[:]), fields.BeaconRole(validatorDuty.Type), zap.Error(err))
+					return
```
nkryuchkov: As I understand, we used to `continue` to the next duty, but now we are interrupting the execution. Can you please briefly explain this change? A comment would likely be great for readability, as the function is quite difficult to follow.

This also applies to similar cases in this function

nkryuchkov (commented):
Great optimization! It would be great to have some tests covering the goroutines logic

nkryuchkov:
Also, can we try to split `ProcessConsensus` in another PR? It's quite hard to follow, but splitting it here would make reviewing more difficult. Although I'm not sure if it could be split nicely

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: The channel is closed [here](https://github.com/ssvlabs/ssv/pull/2340/files#diff-c615c5d14e307fe437330a4144eaa9c319e9f30cf23aef1c724d8712057b1868R308) once all iterations are done. So on context error it will break from the loop and proceed to closing the channel. So it won't really make any behavioral difference? Wdyt?

I will defer the close call, I think it is more idiomatic though

oleg-ssvlabs: fixed 5576ba247a102e21963e7c19fc953820fcd66234

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: I agree there is a chance for the message to end up in the channel even after the context is cancelled.

fixed 5576ba247a102e21963e7c19fc953820fcd66234

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: this is a good catch!

Returning here looks like would kill some workers in such situations. 

Fixed 5576ba247a102e21963e7c19fc953820fcd66234 please review

oleg-ssvlabs:
> Also, can we try to split `ProcessConsensus` in another PR? It's quite hard to follow, but splitting it here would make reviewing more difficult. Although I'm not sure if it could be split nicely

I understand what you mean. I think the changes are quite cohesive, so it would be hard to split. I already split this from #2399

protocol/v2/ssv/runner/committee.go
nkryuchkov: @oleg-ssvlabs ah, I'm sorry, I somehow misread the `break` as `return`

nkryuchkov:
> > Also, can we try to split `ProcessConsensus` in another PR? It's quite hard to follow, but splitting it here would make reviewing more difficult. Although I'm not sure if it could be split nicely
> 
> I understand what you mean. I think the changes are quite cohesive, so it would be hard to split. I already split this from #2399

ok, let's keep it as is

> > Great optimization! It would be great to have some tests covering the goroutines logic
> 
> ty. Spec tests were really good when it comes to covering this method, including concurrency. There were plenty of race conditions that were caught by them. So I think, in general, the coverage feels good in this case

I didn't think that they could cover it while reviewing. Alright, then they are likely sufficient

Issue #2189, opened 2025-04-30
message/validation: duplicated round change messages


nkryuchkov:
Observed in logs:

```
  |   | 2025-04-29T22:51:08.291051Z COMMITTEE_RUNNER ssv-node-exporter-2 message is duplicated, got round change, having proposal, prepare, commit, round change |  
-- | -- | -- | --
  |   | 2025-04-29T22:51:08.288829Z COMMITTEE_RUNNER ssv-node-exporter message is duplicated, got round change, having proposal, prepare, commit, round change
```

nkryuchkov:
The log hasn't been observed in the logs of the last 2 days. Keeping this closed

PR #2341, opened 2025-06-27
message/validation: don't treat same round RC with increased RCJ as duplicated

nkryuchkov:
Closes https://github.com/ssvlabs/ssv/issues/2189

### DIFF
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -65,25 +65,33 @@ func (e Error) Is(target error) bool {
 	return e.text == t.text
 }
 
+// ignored errors
 var (
-	ErrWrongDomain                             = Error{text: "wrong domain"}
-	ErrNoShareMetadata                         = Error{text: "share has no metadata"}
-	ErrUnknownValidator                        = Error{text: "unknown validator"}
-	ErrValidatorLiquidated                     = Error{text: "validator is liquidated"}
-	ErrValidatorNotAttesting                   = Error{text: "validator is not attesting"}
-	ErrEarlySlotMessage                        = Error{text: "message was sent before slot starts"}
-	ErrLateSlotMessage                         = Error{text: "current time is above duty's start +34(committee and aggregator) or +3(else) slots"}
-	ErrSlotAlreadyAdvanced                     = Error{text: "signer has already advanced to a later slot"}
-	ErrRoundAlreadyAdvanced                    = Error{text: "signer has already advanced to a later round"}
-	ErrDecidedWithSameSigners                  = Error{text: "decided with same number of signers"}
-	ErrPubSubDataTooBig                        = Error{text: "pub-sub message data too big"}
-	ErrIncorrectTopic                          = Error{text: "incorrect topic"}
-	ErrNonExistentCommitteeID                  = Error{text: "committee ID doesn't exist"}
-	ErrRoundTooHigh                            = Error{text: "round is too high for this role"}
-	ErrValidatorIndexMismatch                  = Error{text: "partial signature validator index not found"}
-	ErrTooManyDutiesPerEpoch                   = Error{text: "too many duties per epoch"}
-	ErrNoDuty                                  = Error{text: "no duty for this epoch"}
-	ErrEstimatedRoundNotInAllowedSpread        = Error{text: "message round is too far from estimated"}
+	ErrWrongDomain                      = Error{text: "wrong domain"}
+	ErrNoShareMetadata                  = Error{text: "share has no metadata"}
+	ErrUnknownValidator                 = Error{text: "unknown validator"}
+	ErrValidatorLiquidated              = Error{text: "validator is liquidated"}
+	ErrValidatorNotAttesting            = Error{text: "validator is not attesting"}
+	ErrEarlySlotMessage                 = Error{text: "message was sent before slot starts"}
+	ErrLateSlotMessage                  = Error{text: "current time is above duty's start +34(committee and aggregator) or +3(else) slots"}
+	ErrSlotAlreadyAdvanced              = Error{text: "signer has already advanced to a later slot"}
+	ErrRoundAlreadyAdvanced             = Error{text: "signer has already advanced to a later round"}
+	ErrDecidedWithSameSigners           = Error{text: "decided with same number of signers"}
+	ErrPubSubDataTooBig                 = Error{text: "pub-sub message data too big"}
+	ErrIncorrectTopic                   = Error{text: "incorrect topic"}
+	ErrNonExistentCommitteeID           = Error{text: "committee ID doesn't exist"}
+	ErrRoundTooHigh                     = Error{text: "round is too high for this role"}
+	ErrValidatorIndexMismatch           = Error{text: "partial signature validator index not found"}
+	ErrTooManyDutiesPerEpoch            = Error{text: "too many duties per epoch"}
+	ErrNoDuty                           = Error{text: "no duty for this epoch"}
+	ErrEstimatedRoundNotInAllowedSpread = Error{text: "message round is too far from estimated"}
+	ErrUnknownOperator                  = Error{text: "operator is unknown"}
+	ErrOperatorValidation               = Error{text: "failed to validate operator data"}
+	ErrRCLowerJustificationCount        = Error{text: "lower round change justification count"}
+)
+
+// rejected errors
+var (
 	ErrEmptyData                               = Error{text: "empty data", reject: true}
 	ErrMismatchedIdentifier                    = Error{text: "identifier mismatch", reject: true}
 	ErrSignatureVerification                   = Error{text: "signature verification", reject: true}
@@ -129,8 +137,6 @@ var (
 	ErrDuplicatedMessage                       = Error{text: "message is duplicated", reject: true}
 	ErrInvalidPartialSignatureTypeCount        = Error{text: "sent more partial signature messages of a certain type than allowed", reject: true}
 	ErrTooManyPartialSignatureMessages         = Error{text: "too many partial signature messages", reject: true}
-	ErrUnknownOperator                         = Error{text: "operator is unknown"}
-	ErrOperatorValidation                      = Error{text: "failed to validate operator data"}
 )
 
 func (mv *messageValidator) handleValidationError(ctx context.Context, peerID peer.ID, decodedMessage *queue.SSVMessage, err error) pubsub.ValidationResult {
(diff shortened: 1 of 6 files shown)


message/validation/seen_msg_types.go
```diff
@@ -72,18 +72,15 @@ func (c *SeenMsgTypes) ValidateConsensusMessage(signedSSVMessage *spectypes.Sign
 			return err
 		}
 	case specqbft.CommitMsgType:
-		if len(signedSSVMessage.OperatorIDs) == 1 {
```
nkryuchkov, line 75: moved outside of the method, we just don't need to call it otherwise

message/validation/consensus_validation.go
```diff
@@ -221,10 +221,33 @@ func (mv *messageValidator) validateQBFTLogic(
...
+					err.got = len(consensusMessage.RoundChangeJustification)
+					err.want = fmt.Sprintf(">%d", signerState.MaxRCJ)
 					return err
 				}
+
+				// Rule: Peer must send only 1 proposal, 1 prepare, 1 commit per round,
+				// as well as 1 round-change per each RCJ length (not checking round changes with higher RCJ length).
+				var needDuplicatedMsgCheck bool
+				switch consensusMessage.MsgType {
+				case specqbft.CommitMsgType:
```
nkryuchkov, line 240: Moved from `ValidateConsensusMessage`

message/validation/consensus_validation.go
```diff
@@ -353,6 +376,12 @@ func (mv *messageValidator) processSignerState(
 		signerState.SeenSigners[quorum.ToBitMask()] = struct{}{}
 	}
 
+	if consensusMessage.MsgType == specqbft.RoundChangeMsgType {
+		// #nosec G115 maxRCJ never exceeds max operators count
+		// maxRCJ always increases, otherwise the message is not accepted making this code unreachable
+		signerState.MaxRCJ = uint8(len(consensusMessage.RoundChangeJustification))
+	}
```
iurii-ssv: > // maxRCJ always increases, otherwise the message is not accepted making this code unreachable

Maybe it's better to move this comment on top of `MaxRCJ` field itself (+ also add some basic description of what it is)

message/validation/consensus_validation.go
```diff
@@ -221,10 +221,33 @@ func (mv *messageValidator) validateQBFTLogic(
 					}
 				}
 
-				// Rule: Peer must send only 1 proposal, 1 prepare, 1 commit, and 1 round-change per round
-				if err := signerState.SeenMsgTypes.ValidateConsensusMessage(signedSSVMessage, consensusMessage); err != nil {
+				// #nosec G115 maxRCJ never exceeds max operators count
+				rcjLen := uint8(len(consensusMessage.RoundChangeJustification))
```
iurii-ssv, line 225: Maybe expand `max operators count` -> `max operators in the cluster` to be clear what "operator count" is referred to (could be interpreted as "max operators in SSV network" for example)

message/validation/consensus_validation.go
```diff
@@ -221,10 +221,33 @@ func (mv *messageValidator) validateQBFTLogic(
...
+				case specqbft.RoundChangeMsgType:
+					needDuplicatedMsgCheck = rcjLen == signerState.MaxRCJ
+				default:
+					needDuplicatedMsgCheck = true
+				}
+				if needDuplicatedMsgCheck {
+					if err := signerState.SeenMsgTypes.ValidateConsensusMessage(consensusMessage); err != nil {
+						return err
+					}
+				}
```
iurii-ssv: I'm lacking some context on RCJs but this change does seem to target https://github.com/ssvlabs/ssv/issues/2189 in proper way,

but I don't really understand:
- why we treat commit messages differently from proposal/prepare
- what `needDuplicatedMsgCheck = len(signedSSVMessage.OperatorIDs) == 1` has to do with duplicate messages

Also, maybe we should rename `ValidateConsensusMessage` -> `EnsureUnique` (or similar) for this code to be a bit more consistent ?

iurii-ssv (commented):
Could you clarify couple of things for me (as per the comment from above) ?

message/validation/consensus_validation.go
nkryuchkov: This check is needed only for commit messages, but not decided messages because there may be a few decided. The logic for them used to be inside `ValidateConsensusMessage`, but adding the round change logic seemed too much for me, so I moved all the logic about whether we need validation out of the `ValidateConsensusMessage`, hence, commits are also in the change diff.

I'm not sure about renaming because it should be read as `SeenMsgTypes.ValidateConsensusMessage` (there's also `SeenMsgTypes.ValidatePartialSignatureMessage`, so we have different checks on both message types for seen messages. Also, uniqueness is the current logic, IIRC we used to check the message count for decideds here, so I'd keep it a bit abstract without exposing the logic in the name

message/validation/errors.go
```diff
@@ -131,6 +131,7 @@ var (
 	ErrTooManyPartialSignatureMessages         = Error{text: "too many partial signature messages", reject: true}
 	ErrUnknownOperator                         = Error{text: "operator is unknown"}
 	ErrOperatorValidation                      = Error{text: "failed to validate operator data"}
+	ErrRCShorterJustifications                 = Error{text: "round change has less justifications than previous ones"}
```
nkryuchkov: @MatheusFranco99 can you please review this check and share your opinion if we actually need it?

message/validation/consensus_validation.go
iurii-ssv: I see, the way this PR groups certain code together makes sense,

the reason why I suggested renaming is - the way `ValidateConsensusMessage` is called (it's called when `needDuplicatedMsgCheck`) makes it impossible/error-prone to change `ValidateConsensusMessage` implementation without potentially introducing bugs (to add some more checks in there without affecting that caller) ... meaning the caller and `ValidateConsensusMessage` are super-tightly coupled, and keeping the name generic "misleadingly hides that coupling"

looking at it some more I think it would be best to rename:
- `ValidateConsensusMessage` -> `ValidateMessageLimits`
- `needDuplicatedMsgCheck` -> `needMsgLimitCheck` (because, if possible, the caller shouldn't make assumptions about what kind of checks `ValidateConsensusMessage / ValidateMessageLimits` performs)

renaming like this ^ will help to break that coupling (so that it would be hard to add checks to  `ValidateMessageLimits` that are inappropriate/breaking for this func's **name**, and hence for the existing callers)

iurii-ssv: > This check is needed only for commit messages, but not decided messages because there may be a few decided.

Perhaps we could add a comment that mentions this distinction ?

nkryuchkov: @iurii-ssv I agree. Fixed

message/validation/errors.go
MatheusFranco99: Why are we allowing more than one round change message per round per operator?
I don't remember it being possible from the [IBFT paper](https://arxiv.org/abs/2002.03613), nor from our spec 🤔

message/validation/consensus_validation.go
```diff
@@ -221,10 +221,36 @@ func (mv *messageValidator) validateQBFTLogic(
...
-				// Rule: Peer must send only 1 proposal, 1 prepare, 1 commit, and 1 round-change per round
-				if err := signerState.SeenMsgTypes.ValidateConsensusMessage(signedSSVMessage, consensusMessage); err != nil {
+				// #nosec G115 maxRCJ never exceeds max operators count
+				rcjLen := uint8(len(consensusMessage.RoundChangeJustification))
+
+				// Rule: Round change justification length should not decrease
+				if rcjLen < signerState.MaxRCJ {
+					err := ErrRCShorterJustifications
+					err.got = len(consensusMessage.RoundChangeJustification)
+					err.want = fmt.Sprintf(">%d", signerState.MaxRCJ)
```
olegshmuelov, line 231: The error message format is incorrect. The check is for >= but the message says >.

Fix:
```go
err.want = fmt.Sprintf(">=%d", signerState.MaxRCJ)
```

The validation allows equal RCJ length, so the error should reflect that.

message/validation/signer_state.go
```diff
@@ -22,6 +22,9 @@ type SignerState struct {
 	// 10 (f=3): C(10,7)+C(10,8)+C(10,9)+C(10,10)=176
 	// 13 (f=4): C(13,9)+C(13,10)+C(13,11)+C(13,12)+C(13,13)=1093
 	SeenSigners map[SignersBitMask]struct{}
+	// MaxRCJ stores max seen round change justification length.
+	// It always increases, making same round messages with shorter RCJ ignored.
+	MaxRCJ uint8
```
olegshmuelov, line 27: Consider adding a `Reset()` method comment explaining why `MaxRCJ` is intentionally NOT reset. This prevents future confusion or "fixes" that could introduce security vulnerabilities.

Suggested update:
```go
// Reset resets the state's round, message counts, and proposal data to the given values.
// It also updates the start time to the current time.
// Note: MaxRCJ is intentionally NOT reset to maintain security across rounds within the same slot.
// This prevents accepting round change messages with weaker justifications in later rounds.
func (s *SignerState) Reset(slot phase0.Slot, round specqbft.Round) {

File: message/validation/validation_test.go [hunk 9]
```diff
@@ -1438,6 +1438,47 @@ func Test_ValidateSSVMessage(t *testing.T) {
...
+		_, err = validator.handleSignedSSVMessage(signedSSVMessage, topicID, peerID, receivedAt)
+		require.NoError(t, err)
+
+		signedSSVMessageShorterRCJ := generateSignedMessage(ks, committeeIdentifier, slot, func(message *specqbft.Message) {
+			message.MsgType = specqbft.RoundChangeMsgType
+		})
+		signedSSVMessage.FullData = nil
+
+		_, err = validator.handleSignedSSVMessage(signedSSVMessageShorterRCJ, topicID, peerID, receivedAt)
+		require.ErrorIs(t, err, ErrRCShorterJustifications)
```
olegshmuelov, line 1481: Consider adding a test for `MaxRCJ` persistence across round advancement to ensure
security is maintained when moving from round N to round N+1 within the same slot.

olegshmuelov (changes requested):
LGTM with minor fixes

nkryuchkov:
I'm closing this because the referenced issue hasn't been observed for a while. I will reopen the PR if we see it again

PR #2358, opened 2025-07-01
protocol: majority fork protection

nkryuchkov:
https://github.com/ssvlabs/SIPs/pull/69
https://github.com/ssvlabs/ssv-spec/issues/555

### DIFF
--- a/protocol/v2/qbft/instance/proposal.go
+++ b/protocol/v2/qbft/instance/proposal.go
@@ -76,7 +76,7 @@ func (i *Instance) isValidProposal(msg *specqbft.ProcessingMessage) error {
 		return errors.New("signer not in committee")
 	}
 
-	if !msg.SignedMessage.MatchedSigners([]spectypes.OperatorID{i.proposer(msg.QBFTMessage.Round)}) {
+	if !msg.SignedMessage.MatchedSigners([]spectypes.OperatorID{i.ProposerForRound(msg.QBFTMessage.Round)}) {
 		return errors.New("proposal leader invalid")
 	}
 
@@ -212,7 +212,11 @@ func (i *Instance) isProposalJustification(
 	return nil
 }
 
-func (i *Instance) proposer(round specqbft.Round) spectypes.OperatorID {
+func (i *Instance) Proposer() spectypes.OperatorID {
+	return i.ProposerForRound(i.State.Round)
+}
+
+func (i *Instance) ProposerForRound(round specqbft.Round) spectypes.OperatorID {
 	// TODO - https://github.com/ConsenSys/qbft-formal-spec-and-verification/blob/29ae5a44551466453a84d4d17b9e083ecf189d97/dafny/spec/L1/node_auxiliary_functions.dfy#L304-L323
 	return i.config.GetProposerF()(i.State, round)
 }
--- a/go.mod
+++ b/go.mod
@@ -152,7 +152,7 @@ require (
 	github.com/hashicorp/golang-lru v0.5.5-0.20210104140557-80c98217689d // indirect
 	github.com/holiman/billy v0.0.0-20250707135307-f2f9b9aae7db // indirect
 	github.com/holiman/bloomfilter/v2 v2.0.3 // indirect
-	github.com/holiman/uint256 v1.3.2 // indirect
+	github.com/holiman/uint256 v1.3.2
 	github.com/huandu/go-clone v1.6.0 // indirect
 	github.com/huin/goupnp v1.3.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
--- a/ibft/storage/store.go
+++ b/ibft/storage/store.go
@@ -58,7 +58,7 @@ func (i *participantStorage) Prune(ctx context.Context, threshold phase0.Slot) {
 	i.logger.Info("removed stale slot entries", zap.String("store", i.ID()), fields.Slot(threshold), zap.Int("count", count), zap.Duration("took", time.Since(start)))
 }
 
-// PruneContinuously on every tick looks up and removes the slots that fall below the retain threshold
+// PruneContinously on every tick looks up and removes the slots that fall below the retain threshold
 func (i *participantStorage) PruneContinously(ctx context.Context, slotTickerProvider slotticker.Provider, retain phase0.Slot) {
 	ticker := slotTickerProvider()
 	i.logger.Info("start stale slot cleanup loop", zap.String("store", i.ID()))
--- a/operator/dutytracer/store.go
+++ b/operator/dutytracer/store.go
@@ -193,7 +193,7 @@ func (c *Collector) GetCommitteeDuty(slot phase0.Slot, committeeID spectypes.Com
 	return clone, nil
 }
 
-// hasSignersForRole checks if the duty has signers for the given role
+// hasSignersForRoles checks if the duty has signers for the given role
 // since we don't store a boolean flag to separate duties by their role in the db
 // we rely on the fact that during collection we separate the signers in their
 // corresponding fields (Attester and SyncCommittee) based on the role
--- a/protocol/v2/blockchain/beacon/blind/blind.go
+++ b/protocol/v2/blockchain/beacon/blind/blind.go
@@ -1,0 +1,291 @@
+package blind
+
+import (
+	"fmt"
+
+	"github.com/attestantio/go-eth2-client/api"
+	apiv1capella "github.com/attestantio/go-eth2-client/api/v1/capella"
+	apiv1deneb "github.com/attestantio/go-eth2-client/api/v1/deneb"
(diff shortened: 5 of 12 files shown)


nkryuchkov:
@greptileai please review

nkryuchkov:
@greptile review please

nkryuchkov:
@greptileai is there anything that can be done to increase the confidence score?

operator/validator/controller.go
```diff
@@ -1095,11 +1095,11 @@ func SetupRunners(
 		spectypes.RoleVoluntaryExit,
 	}
 
-	buildController := func(role spectypes.RunnerRole, valueCheckF specqbft.ProposedValueCheckF) *qbftcontroller.Controller {
+	buildController := func(role spectypes.RunnerRole, valueChecker ssv.ValueChecker) *qbftcontroller.Controller {
 		config := &qbft.Config{
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
-			ValueCheckF:  nil, // is set per role type
+			ValueChecker: valueChecker, // is set per role type
```
iurii-ssv: Nit: maybe we should move `// is set per role type` comment to be on top of `ValueChecker` field, expanding it a bit (to define what this is)

protocol/v2/qbft/instance/proposal.go
```diff
@@ -137,8 +138,15 @@ func (i *Instance) isProposalJustification(
...
+	valueChecker := i.config.GetValueChecker()
+	if voteChecker, ok := valueChecker.(ssv.VoteChecker); ok && i.spData != nil {
+		if err := voteChecker.CheckValueWithSP(fullData, i.spData); err != nil {
+			return errors.Wrap(err, "proposal fullData invalid")
+		}
+	} else {
+		if err := valueChecker.CheckValue(fullData); err != nil {
+			return errors.Wrap(err, "proposal fullData invalid")
+		}
 	}
```
iurii-ssv, line 142: This looks like a "work-around kind of way to go about it", I think what we want/should try to do instead is:
- merge `CheckValue` and `CheckValueWithSP` methods into single `CheckValue` method - the idea being the concrete implementation of this interface defines "what kind of check needs to be done" by storing all the necessary data on the implementation struct itself - we need to store the `spData *ssvtypes.SlashingProtectionData` on `voteChecker` (while other types of checkers won't need to store it)
- instead of fetching value-checker via `i.config.GetValueChecker()` we'll need to spawn it on per-qbft-instance basis - shortly before we create the `Instance` object itself (probably need to pass it with the `Instance.Start` call) so that here we could just use it like `i.valueChecker.CheckValue(fullData)` every time regardless of what the underlying implementation is
- from what I see in the code, looks like we'll need to instantiate `ValueChecker` in `executeDuty` of every runner and pass it down into `BaseRunner.decide` (so it can pass it down into the actual `Instance`) ... seems like we can get rid of `GetValChecker()` func on every runner and just store the instantiated `valueChecker` on runner-struct (if we even need to)
- and so we won't need to pass `spData *ssvtypes.SlashingProtectionData` in many func (for funcs that don't really need to know about it)
- we can probably also get rid of `VoteChecker` interface (just `voteChecker struct` will be sufficient)

protocol/v2/ssv/value_check.go
```diff
@@ -2,165 +2,232 @@ package ssv
...
-	beaconConfig *networkconfig.Beacon,
-	expectedType spectypes.BeaconRole,
-	validatorPK spectypes.ValidatorPK,
-	validatorIndex phase0.ValidatorIndex,
-) error {
-	if beaconConfig.EstimatedEpochAtSlot(duty.Slot) > beaconConfig.EstimatedCurrentEpoch()+1 {
-		return errors.New("duty epoch is into far future")
+type ValueChecker interface {
+	CheckValue(value []byte) error
+}
```
iurii-ssv, line 20: Wouldn't `ssv/protocol/v2/qbft` be a better place for `ValueChecker` to reside in ... it seems like a qbft-specific thing, wdyt ?

As for concrete implementations, those `qbft` / `Instance` shouldn't ideally be aware of, hence it's probably better to define those structs under `ssv/protocol/v2/ssv` (or `ssv/protocol/v2/ssv/qbft`)

protocol/v2/ssv/runner/runner.go
```diff
@@ -443,3 +451,7 @@ func (b *BaseRunner) ShouldProcessNonBeaconDuty(duty spectypes.Duty) error {
 func (b *BaseRunner) OnTimeoutQBFT(ctx context.Context, logger *zap.Logger, msg ssvtypes.EventMsg) error {
 	return b.QBFTController.OnTimeout(ctx, logger, msg)
 }
+
+type nopValueChecker struct{}
+
+func (nopValueChecker) CheckValue(value []byte) error { return nil }
```
iurii-ssv: I think we should define a `nopValueChecker` next to the rest of value-checkers for completeness sake

protocol/v2/types/spdata.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
+
+// SlashingProtectionData holds duty source and target epochs.
+// It's used for majority fork protection and only for spectypes.BeaconVote.
+type SlashingProtectionData struct {
+	SourceEpoch phase0.Epoch
+	TargetEpoch phase0.Epoch
+}
```
iurii-ssv: Perhaps `SlashingProtectionData` -> `SlashingProtectionBeaconVote` would be more descriptive,

but if we apply https://github.com/ssvlabs/ssv/pull/2358/files#r2379043751 I think this struct just becomes a part of `voteChecker struct` (it is an implementation detail of the checker after all, not ideal to have it hanging around in public scope)

protocol/v2/ssv/value_check.go
```diff
@@ -2,165 +2,232 @@ package ssv
...
+	} else if blockData, _, err := cd.GetBlockData(); err == nil {
+		block = blockData
+	} else {
 		return errors.New("no block data")
 	}
+
+	slot, err := block.Slot()
+	if err != nil {
+		return errors.Wrap(err, "failed to get slot from block data")
+	}
```
iurii-ssv: Love the `var block interface` thingy 👍 

- the `errors.Wrap` we could replace with `fmt.Errorf` here and in 1 other place in this file, for consistency
- but also we seem to be dropping the original error(s) ? We should probably use `errors.Join` to preserve those and add the accumulated error to the `failed to get slot from block data` eventually

protocol/v2/ssv/runner/committee.go
```diff
@@ -267,7 +268,7 @@ func (cr *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 	defer span.End()
 
 	span.AddEvent("checking if instance is decided")
-	decided, decidedValue, err := cr.BaseRunner.baseConsensusMsgProcessing(ctx, logger, cr.GetValCheckF(), msg, &spectypes.BeaconVote{})
+	decided, decidedValue, err := cr.BaseRunner.baseConsensusMsgProcessing(ctx, logger, cr.GetValChecker().CheckValue, msg, &spectypes.BeaconVote{})
```
iurii-ssv: So here we are using `CheckValue` instead of `CheckValueWithSP` ... I'm not sure if there is some intent behind it (if yes it needs to be documented) - we already should have the `spData` to use with `CheckValueWithSP` by this point (it's calculated in `executeDuty`)

so the current behavior seems to be:
- we **reject** the `value` if we noticed it doesn't pass our `CheckValueWithSP` during qbft-proposal stage
- we **accept** the `value` if it's "decided" value as long as `CheckValue` is satisfied (`CheckValueWithSP` is unnecessary)

iurii-ssv (commented):
I think [we need to adjust the way we work with (pass around) `ValueChecker`](https://github.com/ssvlabs/ssv/pull/2358/files#r2379043751) to avoid the tight coupling between components involved - so we can work with this code later on in a simpler manner

protocol/v2/ssv/value_check.go
nkryuchkov: It's also used in `protocol/v2/ssv/{runner,testing}`, so I'm not sure moving it will improve something

protocol/v2/ssv/runner/runner.go
nkryuchkov: deleted

protocol/v2/types/spdata.go
nkryuchkov: deleted

protocol/v2/ssv/value_check.go
iurii-ssv: Yeah I see now they are also used in `runner` package ... so the current placement is a better fit indeed.

iurii-ssv (approved):

Issue #2160, opened 2025-04-17
Simplify Beacon interface


iurii-ssv:
We could simplify `Beacon interface` to make its methods a bit more consistent with each other (and shorter as well) by removing unnecessary `Get` prefix, see https://github.com/ssvlabs/ssv/pull/2145#discussion_r2048697688 for details

nkryuchkov:
Depends on https://github.com/ssvlabs/ssv/pull/2145

iurii-ssv:
Couple of notes:
- it's called `BeaconNetwork interface` now
- I think we can rename it to `Network interface` since it's in `package beacon`

PR #2363, opened 2025-07-02
networkconfig: simplify interfaces (Beacon, SSV)

iurii-ssv:
This PR:
- moves `NetworkName()` into independent **func** (as opposed to being a **method** on `Network interface`) + also renaming it to `NetworkType()`, this is because `Beacon interface` also has a method with the same name which makes it a bit confusing/error-prone to use
- ~~simplifies `Beacon interface` to get rid of unnecessary `Get` prefixes (that makes the caller-code much cleaner/shorter)~~
- ~~(as a side-effect of conforming to `Beacon interface`) it forces us to get rid of "pointless getters" on `BeaconConfig struct` and make all the fields on `BeaconConfig struct` fields package-private so they can be accessed via publicly-exposed methods only~~
- ~~I was considering applying similar adjustments to `SSV interface` (so SSV-config struct/interface will 100% mirror the way it's organized for Beacon-config) ... but there such an approach of making `SSVConfig struct` package-private would result in lots of minor code-changes (replacing direct **struct-field** references with **func** calls in every external package) - so I found a simpler approach of just wrapping `SSVConfig struct` with `ssvConfigAdaptor struct` to just work (get rid of `Get` prefixes) without the need for huge refactor; Beacon-config struct/interface simply doesn't require such a wrapper/adaptor to fit within existing code - but we potentially can apply the same trick for it as well just to keep things 100% consistent~~
- ~~additionally, having both `NetworkConfig struct` and `Network interface` definitions is confusing (there is no clear differentiation for when each of these should be used), so I'm deleting `NetworkConfig struct` in this PR to simplify things (deleting `Network interface` would be another option, but it's harder to do due to various tests needing to mock it out)~~

Some/most parts of this PR ^ might be replaced by https://github.com/ssvlabs/ssv/pull/2392 (as currently scratched-out)

Closes https://github.com/ssvlabs/ssv/issues/2160


networkconfig/beacon.go
```diff
@@ -31,53 +32,80 @@ type Beacon interface {
...
+		genesisValidatorsRoot:                genesisValidatorsRoot,
+		forks:                                forkData,
 	}
+}
 
-	return string(marshaled)
+// String implements Stringer interface.
+func (b *BeaconConfig) String() string {
+	return litter.Options{HidePrivateFields: false}.Sdump(b)
 }
```
iurii-ssv, line 94: Due to the way JSON-encoding works in Golang only exported fields will be shown in the marshal-output, so this func needs to be updated (we can't automatically json-encode this struct anymore), and I think using `litter` is the simplest way to do it since we don't really need an actual JSON here but simply some sort of struct-dump (for logging purposes)

iurii-ssv:
Closing in favor of https://github.com/ssvlabs/ssv/pull/2392 since it does exactly the same sort of refactoring, just using `structs` instead of `interfaces` for configuration entities,

`struct`-based approach is a bit simpler overall - so I think we'd prefer it over what I have implemented in this PR,

the reason I didn't go for `struct`-based approach myself is I wasn't sure how many adjustments unit-tests will need to make it work, but since **nkryuchkov** gave it a try and succeeded I think it's best to go with it.

PR #2366, opened 2025-07-03
[SSV-2] ssvsigner: error on empty fingerprints

nkryuchkov:
The PR checks if the loaded file has any fingerprints

### DIFF
--- a/ssvsigner/tls/tls.go
+++ b/ssvsigner/tls/tls.go
@@ -113,6 +113,10 @@ func (c *Config) LoadServerTLSConfig() (*tls.Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("load known clients: %w", err)
 		}
+
+		if len(trustedFingerprints) == 0 {
+			return nil, fmt.Errorf("no client fingerprints found; mutual TLS required")
+		}
 	}
 
 	return createServerTLSConfig(certificate, trustedFingerprints)


kchojn (dismissed):
i guess it should be a different PR, but imo we could extend our tls tests
```go
ok      github.com/ssvlabs/ssv/ssvsigner/tls    0.930s  coverage: 40.0% of statements
```
Coverage has dropped significantly since the init implementation of TLS (we have added a lot of new checks)

nkryuchkov:
> i guess it should be a different PR, but imo we could extend our tls tests
> 
> ```go
> ok      github.com/ssvlabs/ssv/ssvsigner/tls    0.930s  coverage: 40.0% of statements
> ```
> 
> Coverage has dropped significantly since the init implementation of TLS (we have added a lot of new checks)

I agree. We also need to review them as some tests are supposed to follow the happy path but they expect an error, e.g. `TestServerTLSConfigWithCompleteConfig`

ssvsigner/tls/tls.go
```diff
@@ -113,6 +113,10 @@ func (c *Config) LoadServerTLSConfig() (*tls.Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("load known clients: %w", err)
 		}
+
+		if len(trustedFingerprints) == 0 {
+			return nil, fmt.Errorf("no client fingerprints found; mutual TLS required")
+		}
```
iurii-ssv, line 119: I might be missing the whole picture here, but I think this added code still maybe contradicts a bit the general code structure and this comment from above ?

> // For Case 3: Load client fingerprints if provided

- it says "if provided", but it seems it MUST be provided ?
- what if `ServerKnownClientsFile` is empty and we don't even drop down into this `if` clause ? Shouldn't we also return error if `c.ServerKnownClientsFile == ""` then ?

iurii-ssv (commented):
Left one question

ssvsigner/tls/tls.go
oleg-ssvlabs: not sure if that is exactly what you mean, but `createServerTLSConfig` looks like require `trustedFingerprints` parameter, but it is actually optional, because of [this](https://github.com/ssvlabs/ssv/blob/ssvsigner-audit-ssv-2-fix-fingerprints/ssvsigner/tls/tls.go#L273) condition.

I think the code could be refactored to make it a little bit more clear, that `trustedFingerprints` is actually optional.

nkryuchkov: @iurii-ssv 

> it says "if provided", but it seems it MUST be provided ?

Fixed

> what if ServerKnownClientsFile is empty and we don't even drop down into this if clause ? Shouldn't we also return error if c.ServerKnownClientsFile == "" then ?

We also allow to partially use TLS if the insecure flag is provide, so I think it's ok. Otherwise, the validation would not allow `ServerKnownClientsFile` to be empty

PR #2374, opened 2025-07-03
fix(syncer) - Metadata syncer bug fixes/improvements

oleg-ssvlabs:
**Description**:

The Syncer is responsible for synchronizing (i.e., aligning with the Beacon network) all validator metadata handled by the node. This includes all validators from all committees the node is part of.

On mainnet, the Exporter node (handling 120k+ validators) is experiencing issues syncing all validators within the provided refresh (or “sync”) interval, which defaults to 12 minutes. In practice, this means that all 120k validators’ metadata should be refreshed every 12 minutes. However, looking at actual metrics, it takes longer to sync all validators instead of 12 minutes.

Few bugs were identified while looking into implementation:
- The cause of the unnecessary syncing delays was a [bug](https://github.com/ssvlabs/ssv/pull/2374/files#diff-f25917176b5804f2dbb61733a57e69315dcf01c75908cae4986ef04582387012R219) in the condition logic: the Syncer would sleep for 2 seconds each time it detected that metadata didn’t need to be updated. However, in the vast majority of cases, metadata is already up to date with the network state - so this sleep caused significant unnecessary delays. This condition was changed.
After this fix, the average batch processing duration dropped from **1.15 seconds** to **246 milliseconds.**
- And [this line](https://github.com/ssvlabs/ssv/pull/2374/files#diff-5a2be62f4b213364f822f71cf79a6405eb2bb1dd344687bc26466ea33f31b554R461) turned out to be the main culprit.  
Shares whose metadata was already in sync with the Beacon (which is the absolute majority) did **not** have their `BeaconMetadataLastUpdated` field updated, which caused the Syncer to keep picking the same shares over and over again.

However, this does not fully solve **potential** upcoming problems specific to the Exporter.

The **sync interval** (i.e., how often shares should be refreshed) is a configurable value, which defaults to **12 minutes**. The number of validators is dynamic and tends to grow over time.

For example, if syncing a **batch of 512 shares** takes 2 seconds (imagine a slower Beacon Node), then within the default 12-minute sync interval, it's possible to sync nearly **180k shares**.

However, if someone reduces the sync interval from 12 to, say, 5 minutes(because it's a config!), the Exporter will no longer be able to sync all shares in time and it will also be very hard to spot this issue(we need better observability and alerts).

Therefore, the batch size should be **dynamic** (adjusted at runtime), based on:
- the configured sync interval,
- the number of validators,
- and the average duration of batch execution.

### DIFF
--- a/eth/executionclient/multi_client.go
+++ b/eth/executionclient/multi_client.go
@@ -269,12 +269,18 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan 
 				}
 
 				_, err := mc.call(contextWithMethod(ctx, "StreamLogs"), f, 0)
-				if err != nil && !isInterruptedError(err) {
-					// NOTE: There are unit tests that trigger Fatal and override its behavior.
-					// Therefore, the code must call `return` afterward.
-					mc.logger.Fatal("failed to stream registry events", zap.Error(err))
+				if err == nil {
+					return
 				}
-				return
+
+				if isInterruptedError(err) {
+					mc.logger.Debug("stream logs stopped", zap.Error(err))
+					return
+				}
+
+				// NOTE: There are unit tests that trigger Fatal and override its behavior.
+				// Therefore, the code must call `return` afterward.
+				mc.logger.Fatal("failed to stream logs", zap.Error(err))
 			}
 		}
 	}()
--- a/network/p2p/p2p_setup.go
+++ b/network/p2p/p2p_setup.go
@@ -380,7 +380,16 @@ func (n *p2pNetwork) inboundLimit() int {
 	return int(float64(n.cfg.MaxPeers) * inboundLimitRatio)
 }
 
+// connectionStats returns the number of inbound and outbound connections.
+// It safely handles the case where the host is not yet initialized, which can
+// occur during network setup when the connection gater is active but libp2p.New()
+// hasn't completed yet. In this case, it returns (0, 0) since no connections
+// exist before the host is fully initialized.
 func (n *p2pNetwork) connectionStats() (inbound, outbound int) {
+	if n.host == nil {
+		return 0, 0
+	}
+
 	return connectionStats(n.host)
 }
 


kchojn (approved):

operator/validator/metadata/syncer.go
```diff
@@ -85,7 +98,7 @@ func WithSyncInterval(interval time.Duration) Option {
 	}
 }
 
-func (s *Syncer) SyncOnStartup(ctx context.Context) (beacon.ValidatorMetadataMap, error) {
+func (s *Syncer) SyncAll(ctx context.Context) (beacon.ValidatorMetadataMap, error) {
```
y0sher, line 104: Don't you want to change the func body as well to sync all shares even if they already exist and not just these that don't have any metadata? this will skip status updates for example.

operator/validator/metadata/syncer.go
```diff
@@ -307,14 +320,10 @@ func (s *Syncer) nextBatchFromDB(_ context.Context, subnetsBuf *big.Int) beacon.
...
 func (s *Syncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
-	// TODO: use time.After when Go is updated to 1.23
-	timer := time.NewTimer(d)
-	defer timer.Stop()
-
 	select {
 	case <-ctx.Done():
 		return false
-	case <-timer.C:
+	case <-time.After(d):
```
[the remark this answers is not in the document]
nkryuchkov: https://go.dev/wiki/Go123Timer

operator/validator/metadata/syncer.go
oleg-ssvlabs: This func doesn’t quite work that way from what I can see.

What it actually does is **sync all shares if _at least one_ share lacks metadata**.

Do you think it should still sync all, even if all shares already have beacon metadata?

operator/validator/metadata/syncer.go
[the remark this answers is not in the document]
oleg-ssvlabs: > Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

So, Copilot's suggestion doesn't seem to be relevant

operator/validator/metadata/syncer.go
y0sher: yes because metadata might be outdated.

oleg-ssvlabs: fixed f1efdd915b7f70d4be764ebdbecf9b4f39ddc927

operator/validator/metadata/syncer.go
```diff
@@ -134,6 +137,10 @@ func (s *Syncer) Sync(ctx context.Context, pubKeys []spectypes.ValidatorPK) (bea
 		zap.Int("received_count", len(metadata)),
 	)
 
+	if len(metadata) == 0 {
```
oleg-ssvlabs: because of [this](https://github.com/ssvlabs/ssv/pull/2374/files#diff-f25917176b5804f2dbb61733a57e69315dcf01c75908cae4986ef04582387012R170)

operator/validator/metadata/syncer.go
```diff
@@ -160,12 +167,21 @@ func (s *Syncer) Fetch(ctx context.Context, pubKeys []spectypes.ValidatorPK) (be
...
+	// and validator's metadata will be updated.
 	validatorsIndexMap, err := s.beaconNode.GetValidatorData(ctx, blsPubKeys)
 	if err != nil {
 		return nil, fmt.Errorf("get validator data from beacon node: %w", err)
 	}
 
-	results := make(beacon.ValidatorMetadataMap, len(validatorsIndexMap))
+	results := make(beacon.ValidatorMetadataMap, len(pubKeys))
+
+	for _, key := range pubKeys {
```
oleg-ssvlabs, line 179: This change is necessary for the case where `GetValidatorData` does return data for some public keys, but not for others (it happens in both Hoodi and Mainnet).
In such cases, we still want to include those keys (for which no data is returned) in the list, along with an empty Metadata object - since Beacon doesn’t provide any data for them.

This change ensures that metadata is marked as updated for such keys and they will only be retried by Syncer once the sync interval has passed.

operator/validator/metadata/syncer.go
nkryuchkov: This might be a better place for that comment because it would answer why we need this check here. If we follow the rule that the comment should answer the question 'why?', it won't help with anything inside `fetchMetadata`

operator/validator/metadata/syncer.go
nkryuchkov: I think we need this comment in the code

nkryuchkov (approved):
Great job!

operator/validator/metadata/syncer.go
oleg-ssvlabs: I added a comment on top of this method instead f2e794cdad242db539662e804d3bb10b130a43b1
wdyt?

registry/storage/shares.go
```diff
@@ -456,6 +458,8 @@ func (s *sharesStorage) UpdateValidatorsMetadata(data beacon.ValidatorMetadataMa
 			}
 
 			if metadata.Equals(share.BeaconMetadata()) {
+				share.BeaconMetadataLastUpdated = time.Now()
+				unchangedShares = append(unchangedShares, share)
```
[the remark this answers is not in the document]
oleg-ssvlabs: It’s a good point. However, based on the logs, saving a batch of items (`512`) currently takes around `7–8 ms`, which is a relatively low number. So, no optimization is needed on that end.

operator/validator/metadata/syncer.go
oleg-ssvlabs: I removed this condition in this [PR](https://github.com/ssvlabs/ssv/pull/2374/commits/f2e794cdad242db539662e804d3bb10b130a43b1).

Instead, I added a check for the length of `pubkeys`. If it's not empty, the `metadata` map will never be empty either, because even when metadata is not fetched, we still assign empty metadata objects. So this check any longer necessary

cli/operator/node.go
```diff
@@ -582,7 +582,7 @@ var StartNodeCmd = &cobra.Command{
 			nodeProber.AddNode("event syncer", eventSyncer)
 		}
 
-		if _, err := metadataSyncer.SyncOnStartup(cmd.Context()); err != nil {
+		if _, err := metadataSyncer.SyncAll(cmd.Context()); err != nil {
 			logger.Fatal("failed to sync metadata on startup", zap.Error(err))
```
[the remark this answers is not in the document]
oleg-ssvlabs: Syncing all shares still happens on node startup, so I believe the message continues to be relevant

operator/validator/metadata/syncer.go
iurii-ssv: Could make describe in a comment `// Sync ...` what this func is supposed to do so we have the intent documented for posterity

operator/validator/metadata/syncer.go
```diff
@@ -289,7 +300,7 @@ func (s *Syncer) nextBatchFromDB(_ context.Context, subnetsBuf *big.Int) beacon.
 			staleShares = append(staleShares, share)
 		}
 
-		return len(newShares) < batchSize
+		return len(newShares)+len(staleShares) < batchSize
```
iurii-ssv, line 309: I think the `len(newShares) < batchSize` condition here is "on purpose" to make sure we iterate/prioritize ALL new shares before we start considering updates for `staleShares` as well, should this be changed ?

cc @nkryuchkov

registry/storage/shares.go
```diff
@@ -432,6 +433,7 @@ func (s *sharesStorage) Delete(rw basedb.ReadWriter, pubKey []byte) error {
 func (s *sharesStorage) UpdateValidatorsMetadata(data beacon.ValidatorMetadataMap) (beacon.ValidatorMetadataMap, error) {
 	var (
 		changedShares   []*types.SSVShare
+		unchangedShares []*types.SSVShare
 		changedMetadata beacon.ValidatorMetadataMap
 	)
```
iurii-ssv: I'm not sure I understand why we need this `unchangedShares` at all, I think `BeaconMetadataLastUpdated` isn't stored in DB as per this comment:
```
	// BeaconMetadataLastUpdated is used to keep track of share last update time. Note, we don't
	// store this field in DB - it just serves as a helper-indicator for when we might want
	// to update SSVShare metadata we fetch from Beacon node so it doesn't get super stale.
	BeaconMetadataLastUpdated time.Time
```
so the way I understand this - we want to update `share.BeaconMetadataLastUpdated = time.Now()` in case metadata didn't change and that's it (nothing needs to be updated in DB) ?

I would also maybe expand a comment on `BeaconMetadataLastUpdated` to clarify that not only it keeps track of the actual metadata updates but also of the last time such an update was considered (but perhaps not applied since there was no change to apply).

iurii-ssv (commented):
Good stuff!

operator/validator/metadata/syncer.go
oleg-ssvlabs: It’s a good question. I’ve been thinking about it as well.

There are a few things to consider:
- ~The previous implementation didn’t properly respect the configured `batchSize`. It continued appending both new and stale shares until the newShares collection reached the batchSize, which could be misleading.~
- Syncing all shares in the Exporter (120k+) takes around 2 seconds. Processing 120k+ shares in batches of 512 is actually very fast. Of course, it takes more than 2 seconds in total mainly because it requires 120k / 512 number of HTTP requests to CL.
- Since there is only 2 seconds delay in between sync runs, most likely it will be less than 512 new shares in the db which will cause this loop to iterate 120k times and with read locks used it will lock writes for that duration. Not sure how big deal that is and how often writes are executed, but potentially a bottleneck.

All things considered (previously wrongly concluded not respecting the batch size), there is value in both keeping this change and reverting to the previous implementation.

registry/storage/shares.go
oleg-ssvlabs: good catch e0cb4d2bf7b18af1a6eaf0b15d5d1889f400547f

operator/validator/metadata/syncer.go
iurii-ssv: Yeah, I would say we are better **with** the change you've made here 👍 

just cc @nkryuchkov @MatusKysel in case they have any objections

operator/validator/metadata/syncer.go
oleg-ssvlabs: added few comments c4a3f0a80b5af93393c7959aca20c8466ef7aaf0

iurii-ssv (approved):

operator/validator/metadata/syncer.go
nkryuchkov: @oleg-ssvlabs LGTM

operator/validator/metadata/syncer.go
nkryuchkov: sounds good to me

PR #2375, opened 2025-07-03
all: cancelable sleep

nkryuchkov:
interrupt sleep on context cancelation

### DIFF
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -27,7 +27,7 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
 	// As specified in spec, an aggregator should wait until two thirds of the way through slot
 	// to broadcast the best aggregate to the global aggregate channel.
 	// https://github.com/ethereum/consensus-specs/blob/v0.9.3/specs/validator/0_beacon-chain-validator.md#broadcast-aggregate
-	gc.waitToSlotTwoThirds(slot)
+	gc.waitToSlotTwoThirds(ctx, slot)
 
 	// differ from spec because we need to subscribe to subnet
 	isAggregator := gc.isAggregator(committeeLength, slotSig)
@@ -202,7 +202,7 @@ func (gc *GoClient) isAggregator(committeeCount uint64, slotSig []byte) bool {
 }
 
 // waitToSlotTwoThirds waits until two-third of the slot has transpired (SECONDS_PER_SLOT * 2 / 3 seconds after slot start time)
-func (gc *GoClient) waitToSlotTwoThirds(slot phase0.Slot) {
+func (gc *GoClient) waitToSlotTwoThirds(ctx context.Context, slot phase0.Slot) {
 	config := gc.getBeaconConfig()
 	oneInterval := config.IntervalDuration()
 	finalTime := config.GetSlotStartTime(slot).Add(2 * oneInterval)
@@ -210,5 +210,10 @@ func (gc *GoClient) waitToSlotTwoThirds(slot phase0.Slot) {
 	if wait <= 0 {
 		return
 	}
-	time.Sleep(wait)
+
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(wait):
+	}
 }
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -20,6 +20,8 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 	"go.uber.org/zap"
 
+	"github.com/ssvlabs/ssv/ssvsigner/keys"
+
 	"github.com/ssvlabs/ssv/logging"
 	"github.com/ssvlabs/ssv/logging/fields"
 	"github.com/ssvlabs/ssv/message/validation"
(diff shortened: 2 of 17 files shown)


utils/tasks/exec_queue_test.go
```diff
@@ -13,7 +14,7 @@ func TestExecQueue(t *testing.T) {
 	var i int64
 	q := NewExecutionQueue(1 * time.Millisecond)
 
-	go q.Start()
+	go q.Start(context.Background())
```
kchojn: we dont use `t.Context()` here?

protocol/v2/qbft/spectest/msg_processing_type.go
```diff
@@ -67,7 +67,11 @@ func RunMsgProcessing(t *testing.T, test *spectests.MsgProcessingSpecTest) {
 	require.NoError(t, err)
 
 	// broadcasting is asynchronic, so need to wait a bit before checking
-	time.Sleep(time.Millisecond * 5)
+	select {
+	case <-t.Context().Done():
+		return
+	case <-time.After(5 * time.Millisecond):
```
kchojn, line 73: magic number, i would move that number to the top of the package + add some docs

network/p2p/p2p.go
```diff
@@ -228,11 +229,17 @@ func (n *p2pNetwork) Close() error {
...
-		time.Sleep(8 * time.Second)
 		ctx, cancel := context.WithCancel(n.ctx)
 		defer cancel()
+
+		// Wait for own subnets to be subscribed to and updated.
+		// TODO: wait more intelligently with a channel.
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(8 * time.Second):
```
kchojn, line 240: magic number, i would move that number (+TODO) to the top of the package + add some docs

kchojn (changes requested):

utils/tasks/exec_queue_test.go
nkryuchkov: I haven't fixed tests yet, perhaps I should have made this draft

protocol/v2/qbft/spectest/msg_processing_type.go
nkryuchkov: I'll likely encounter many similar cases in tests. I'm not sure we should do this in this PR to avoid inflating it. But I'll think about it, thanks

kchojn: ah this one is from test, nvm

network/p2p/p2p.go
nkryuchkov: Same as https://github.com/ssvlabs/ssv/pull/2375#discussion_r2185523032, resolving.

Please reopen if you have any objections

utils/tasks/exec_queue_test.go
nkryuchkov: Fixed this one, but decided not to fix tests because they might have more complex logic

beacon/goclient/aggregator.go
```diff
@@ -202,13 +202,18 @@ func (gc *GoClient) isAggregator(committeeCount uint64, slotSig []byte) bool {
...
 	finalTime := config.GetSlotStartTime(slot).Add(2 * oneInterval)
 	wait := time.Until(finalTime)
 	if wait <= 0 {
 		return
 	}
-	time.Sleep(wait)
+
+	select {
+	case <-ctx.Done():
+		return
```
oleg-ssvlabs: don't you think it is better to return an error here? 

Like this:
```Go
return ctx.Err()
```

It would be better for the caller [here](https://github.com/ssvlabs/ssv/pull/2375/files#diff-a543d17cca4f1962b20be72ccddbf8029b2c2dd51b7909bd2cb86730cdc874fcR30) to return an error instead of proceeding with the flow when a cancellation or deadline occurs

network/p2p/test_utils.go
```diff
@@ -84,12 +84,18 @@ func CreateAndStartLocalNet(pCtx context.Context, logger *zap.Logger, options Lo
...
 				}
+
 				ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
 				defer cancel()
+
 				var peers []peer.ID
 				for len(peers) < options.MinConnected && ctx.Err() == nil {
 					peers = node.(HostProvider).Host().Network().Peers()
-					time.Sleep(time.Millisecond * 100)
+					select {
```
oleg-ssvlabs, line 94: I think you can actually remove few things from here and change it to:

```Go
for len(peers) < options.MinConnected {
	peers = node.(HostProvider).Host().Network().Peers()
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-time.After(100 * time.Millisecond):
	}
}
logger.Debug("found enough peers", zap.Int("for node", i), zap.Int("nodesQuantity", options.Nodes), zap.String("found", fmt.Sprintf("%+v", peers)))

```

protocol/v2/qbft/spectest/msg_processing_type.go
```diff
@@ -67,7 +67,11 @@ func RunMsgProcessing(t *testing.T, test *spectests.MsgProcessingSpecTest) {
 	require.NoError(t, err)
 
 	// broadcasting is asynchronic, so need to wait a bit before checking
```
oleg-ssvlabs: (nit-unrelated, but while here) - asynchronic -> asynchronous

oleg-ssvlabs (commented):
good job!
left few minor comments

iurii-ssv (approved):

PR #2378, opened 2025-07-07
fix(eth/executionclient/multi_client.go): log all errors in MultiClient StreamLogs

kchojn:
Why debug lvl:
- these are expected during normal shutdown sequences
- info/warn/error would create false positives
- available when debugging shutdown issues

### DIFF
--- a/network/p2p/p2p_setup.go
+++ b/network/p2p/p2p_setup.go
@@ -380,7 +380,16 @@ func (n *p2pNetwork) inboundLimit() int {
 	return int(float64(n.cfg.MaxPeers) * inboundLimitRatio)
 }
 
+// connectionStats returns the number of inbound and outbound connections.
+// It safely handles the case where the host is not yet initialized, which can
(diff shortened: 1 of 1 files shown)


eth/executionclient/multi_client.go
```diff
@@ -269,10 +269,14 @@ func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan
...
 
 				_, err := mc.call(contextWithMethod(ctx, "StreamLogs"), f, 0)
-				if err != nil && !isInterruptedError(err) {
-					// NOTE: There are unit tests that trigger Fatal and override its behavior.
-					// Therefore, the code must call `return` afterward.
-					mc.logger.Fatal("failed to stream registry events", zap.Error(err))
+				if err != nil {
+					if isInterruptedError(err) {
+						mc.logger.Debug("stream logs stopped", zap.Error(err))
+					} else {
```
nkryuchkov: Can we invert the condition to follow the happy flow?

PR #2379, opened 2025-07-07
ssvsigner: load possible host names from whole certificate

vaclav-ssvlabs:
The previous behavior had unpredictable behavior when sometimes it check Servername sometimes it uses TLS hosts and so on.

But often with TLS the certificate has multiple valid  hosts that can be used in different scenarios and we should check all of them.

### DIFF
--- a/ssvsigner/tls/tls.go
+++ b/ssvsigner/tls/tls.go
@@ -327,7 +327,7 @@ func verifyServerCertificate(state tls.ConnectionState, trustedFingerprints map[
 		}
 	}
 	// If we reach here, the server certificate is not trusted
-	return fmt.Errorf("server certificate fingerprint for hosts %q not trusted: %#v", hosts, formatFingerprint(fingerprintHex))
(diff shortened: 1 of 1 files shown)


ssvsigner/tls/tls.go
```diff
@@ -292,26 +292,42 @@ func verifyServerCertificate(state tls.ConnectionState, trustedFingerprints map[
...
 	fingerprintHex := hex.EncodeToString(fingerprint[:])
 
-	// Get the hostname from multiple possible sources
-	host := state.ServerName
-	if host == "" && len(cert.DNSNames) > 0 {
-		host = cert.DNSNames[0]
-	} else if host == "" {
-		host = cert.Subject.CommonName
+	// Get the possible hostnames from multiple possible sources
+	var hosts []string
```
kchojn, line 296: i agree with copilot, we should create `hostSet` to avoid checking the same hostname multiple times. then we can iterate through the hostSet

kchojn (changes requested):

PR #2383, opened 2025-07-08
fix(ssvsigner) tls load hostname from all cert data

vaclav-ssvlabs:
Ensures that we check  all certificate entries for the all possible host

### DIFF
--- a/ssvsigner/tls/tls.go
+++ b/ssvsigner/tls/tls.go
@@ -296,24 +296,42 @@ func verifyServerCertificate(state tls.ConnectionState, trustedFingerprints map[
 	fingerprint := sha256.Sum256(cert.Raw)
 	fingerprintHex := hex.EncodeToString(fingerprint[:])
 
-	// Get the hostname from multiple possible sources
-	host := state.ServerName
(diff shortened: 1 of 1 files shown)


ssvsigner/tls/tls.go
```diff
@@ -296,24 +296,42 @@ func verifyServerCertificate(state tls.ConnectionState, trustedFingerprints map[
...
+	}
+	// iterate over all possible hostnames in certificate
+	for host := range hosts {
+		// Check fingerprint against our trusted list
+		if expectedFingerprint, ok := trustedFingerprints[host]; ok {
+			expectedFingerprint = normalizeFingerprint(expectedFingerprint)
+			if expectedFingerprint == fingerprintHex {
+				return nil
+			}
+			return fmt.Errorf("server certificate fingerprint mismatch for %s: expected %s, got %s",
```
nkryuchkov: Do we need to return here? Should we continue the loop instead?

nkryuchkov (dismissed):

nkryuchkov (approved):

y0sher (approved):

iurii-ssv (approved):

PR #2388, opened 2025-07-09
linter: enable useful linters

iurii-ssv:
Even thought we have defined some configuration for many linters (in `.golangci.yaml`) - most of them are still disabled by default, unless enabled explicitly (under `enabled` section). I went over those linters we have listed (but disabled by default in `.golangci.yaml` file):
- enabled those that are helpful (meaning their suggestions are reasonable IMO & existing code either already satisfies their rule-sets or can be simply adjusted as needed)
- completely removed configuration for all other linters (these are mostly super-opinionated linters that are probably creating more friction than benefit, I don't think we can easily adopt these)

Note, for `unused` linter specifically I found it reports lots of false-positives even on "conservative settings" - so I don't see it to be a viable linter to use.

_Before merging:_
- [x] test on **stage**

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1054,8 +1054,12 @@ func (c *controller) ReportValidatorStatuses(ctx context.Context) {
 			validatorsPerStatus := make(map[validatorStatus]uint32)
 
 			for _, share := range c.validatorStore.OperatorValidators(c.operatorDataStore.GetOperatorID()) {
-				if share.IsParticipating(c.networkConfig, c.networkConfig.EstimatedCurrentEpoch()) {
+				currentEpoch := c.networkConfig.EstimatedCurrentEpoch()
+				if share.IsParticipating(c.networkConfig, currentEpoch) {
 					validatorsPerStatus[statusParticipating]++
+				}
+				if share.IsAttesting(currentEpoch) {
+					validatorsPerStatus[statusAttesting]++
 				}
 				if !share.HasBeaconMetadata() {
 					validatorsPerStatus[statusNotFound]++
--- a/api/handlers/model.go
+++ b/api/handlers/model.go
@@ -106,6 +106,15 @@ func toProposalTrace(rt *model.ProposalTrace) *proposalTrace {
 	if rt == nil {
 		return nil
 	}
+
+	// Filter out zero-valued ProposalTrace objects created by SSZ encoding.
+	// These occur when rounds exist with only roundChanges (or possibly also prepares/commits) but no actual proposal.
+	// Without filtering, they appear as confusing proposals with round=0, leader=0, epoch timestamp.
+	// While 0-valued proposals are technically valid according to our data model, they show up as confusing API responses.
+	if rt.Round == 0 && rt.Signer == 0 && rt.ReceivedTime == 0 {
+		return nil
+	}
+
 	return &proposalTrace{
 		Round:                 rt.Round,
 		BeaconRoot:            rt.BeaconRoot,
--- a/operator/duties/attester.go
+++ b/operator/duties/attester.go
@@ -230,7 +230,7 @@ func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 
 	var eligibleShares []*types.SSVShare
 	for _, share := range h.validatorProvider.SelfValidators() {
-		if share.IsParticipatingAndAttesting(epoch) {
+		if share.IsAttesting(epoch) {
 			eligibleShares = append(eligibleShares, share)
 		}
 	}
--- a/operator/duties/committee.go
+++ b/operator/duties/committee.go
@@ -195,7 +195,7 @@ func (h *CommitteeHandler) toSpecSyncDuty(duty *eth2apiv1.SyncCommitteeDuty, slo
 
 func (h *CommitteeHandler) shouldExecuteAtt(duty *eth2apiv1.AttesterDuty, epoch phase0.Epoch) bool {
 	share, found := h.validatorProvider.Validator(duty.PubKey[:])
-	if !found || !share.IsParticipatingAndAttesting(epoch) {
+	if !found || !share.IsAttesting(epoch) {
 		return false
 	}
 
--- a/operator/duties/proposer.go
+++ b/operator/duties/proposer.go
@@ -189,7 +189,7 @@ func (h *ProposerHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 
 	var allEligibleIndices []phase0.ValidatorIndex
 	for _, share := range h.validatorProvider.Validators() {
-		if share.IsParticipatingAndAttesting(epoch) {
+		if share.IsAttesting(epoch) {
 			allEligibleIndices = append(allEligibleIndices, share.ValidatorIndex)
 		}
 	}
@@ -203,7 +203,7 @@ func (h *ProposerHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 
 	selfEligibleIndices := map[phase0.ValidatorIndex]struct{}{}
 	for _, share := range h.validatorProvider.SelfValidators() {
-		if share.IsParticipatingAndAttesting(epoch) {
+		if share.IsAttesting(epoch) {
 			selfEligibleIndices[share.ValidatorIndex] = struct{}{}
 		}
 	}
--- a/operator/duties/validatorregistration.go
+++ b/operator/duties/validatorregistration.go
@@ -55,7 +55,7 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
 
 			var vrs []ValidatorRegistration
 			for _, share := range shares {
-				if !share.IsParticipatingAndAttesting(epoch + phase0.Epoch(frequencyEpochs)) {
+				if !share.IsAttesting(epoch + phase0.Epoch(frequencyEpochs)) {
 					// Only attesting validators are eligible for registration duties.
 					continue
 				}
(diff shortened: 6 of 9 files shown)


operator/validator/controller.go
```diff
@@ -1151,7 +1150,6 @@ func SetupRunners(
 	options *validator.CommonOptions,
 ) (runner.ValidatorDutyRunners, error) {
 	runnersType := []spectypes.RunnerRole{
-		spectypes.RoleCommittee,
```
iurii-ssv, line 1158: Removing this since we don't set up committee-runner here - `SetupCommitteeRunners` is setting up committee-runner(s)

migrations/migration_6_share_exit_epoch_test.go
```diff
@@ -119,8 +119,8 @@ func assertMigratedShares(t *testing.T, db basedb.Database, key []byte, seededSh
 
 func seedDatabase(numOfItems int, db basedb.Database, storageKey []byte) ([]*migration_6_OldStorageShare, error) {
 	var (
-		dbShares     []basedb.Obj
-		seededShares []*migration_6_OldStorageShare
+		dbShares     = make([]basedb.Obj, 0, numOfItems)
+		seededShares = make([]*migration_6_OldStorageShare, 0, numOfItems)
```
nkryuchkov: While we are touching this, maybe just
```suggestion
		dbShares     := make([]basedb.Obj, 0, numOfItems)
		seededShares := make([]*migration_6_OldStorageShare, 0, numOfItems)
```

?

It will make the code simpler

network/discovery/iterator_test.go
```diff
@@ -43,7 +43,7 @@ func TestFairMixIterator_Next_False(t *testing.T) {
 	iterator.AddSource(preFork)
 	iterator.AddSource(postFork)
 
-	var expectedNodes []*enode.Node
+	expectedNodes := make([]*enode.Node, 0, len(preforkNodes)+len(postForkNodes))
```
nkryuchkov, line 46: Unrelated to the PR, but should we create an issue while we are touching it about whether we actually need the fork-related code here? IIRC we removed all related code.

operator/dutytracer/collector.go
```diff
@@ -519,8 +518,7 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 		tracerInFlightMessageHist.Record(ctx, time.Since(start).Seconds())
 	}()
 
-	switch msg.MsgType {
-	case spectypes.SSVConsensusMsgType:
+	if msg.MsgType == spectypes.SSVConsensusMsgType {
```
nkryuchkov, line 537: I think the switch might be a bit more readable, especially for such large cases. What do you think about adding a `default` case to switch to make linter happy with it?

operator/validator/controller.go
```diff
@@ -411,8 +411,7 @@ func (c *controller) handleNonCommitteeMessages(
 	c.committeesObserversMutex.Lock()
 	defer c.committeesObserversMutex.Unlock()
 
-	switch msg.MsgType {
-	case spectypes.SSVConsensusMsgType:
+	if msg.MsgType == spectypes.SSVConsensusMsgType {
```
nkryuchkov, line 414: I'd add the `default` case instead

protocol/v2/qbft/instance/instance.go
```diff
@@ -106,23 +106,23 @@ func (i *Instance) Start(ctx context.Context, logger *zap.Logger, value []byte,
...
-				logger.Debug(eventMsg)
-				span.AddEvent(eventMsg, trace.WithAttributes(attribute.String("root", hex.EncodeToString(r[:]))))
-
-				if err := i.Broadcast(logger, proposal); err != nil {
-					logger.Warn("❌ failed to broadcast proposal", zap.Error(err))
-					span.RecordError(err)
-				}
+			}
+
+			r, err := specqbft.HashDataRoot(i.StartValue) // @TODO (better than decoding?)
```
nkryuchkov: while we are touching it
```suggestion
			r, err := specqbft.HashDataRoot(i.StartValue) // TODO (better than decoding?)
```

nkryuchkov (approved):
LGTM, although I'd make some minor changes

network/discovery/iterator_test.go
iurii-ssv: Yeah, I guess, https://github.com/ssvlabs/ssv/issues/2398 - feel free to update the description as necessary

operator/dutytracer/collector.go
iurii-ssv: Personally, I think either way is fine, and sometimes it depends on the particular case
- it's probably better to use `switch` when there are many choices and we can combine the handling for some of them (we'd need to duplicate code to implement the same with `if`), or we want to exhaust them all
- `switch-default` adds extra code lines, so all else being equal I'd say `if` is easier to read (and reason about)

anyway, for this particular case I would leave it with `if` since `switch-default` doesn't improve it in any way (and the readability aspect for these is a bit subjective IMO)

operator/validator/controller.go
iurii-ssv: answered here - https://github.com/ssvlabs/ssv/pull/2388#discussion_r2206867475

nkryuchkov (approved):

network/discovery/logger/common.go
```diff
@@ -56,7 +56,7 @@ func FormatError(err error) map[string]any {
 	return map[string]any{
 		"kind":  reflect.TypeOf(err).String(),
 		"error": err.Error(),
-		"stack": nil, // @TODO
+		"stack": nil, // TODO
```
oleg-ssvlabs: unrelated: maybe remove this TODO? There is not description regarding what to do :)

operator/validator/controller.go
```diff
@@ -481,8 +482,8 @@ func (c *controller) setupValidators(shares []*ssvtypes.SSVShare) ([]*validator.
 	c.logger.Info("starting validators setup...", zap.Int("shares count", len(shares)))
 	var errs []error
 	var fetchMetadata [][]byte
-	var validators []*validator.Validator
-	var committees []*validator.Committee
+	validators := make([]*validator.Validator, 0, len(shares))
```
oleg-ssvlabs, line 485: I actually thought the linter would be smart enough to detect that the new slice with preallocated capacity (`validators`) will always reach the capacity set during initialization (`shares` length).  
However, since appending is conditional, it's likely that the `validators` slice won't be fully populated in this case.

Anyway, I think there are no drawbacks to setting more capacity than actually needed, the most important thing is that we don't exceed it and trigger dynamic resizing which is costly.

protocol/v2/ssv/runner/sync_committee_aggregator.go
```diff
@@ -106,6 +106,7 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
 	recordPreConsensusDuration(ctx, r.measurements.PreConsensusTime(), spectypes.RoleSyncCommitteeContribution)
 
 	// collect selection proofs and subnets
+	//nolint: prealloc
```
oleg-ssvlabs, line 109: We had a few other cases where appending was conditional and the preallocated capacity likely wouldn’t be fully used.  
Not sure if we need to disable the linter for that reason. I don’t really have a strong opinion here, maybe it depends on how likely the final length will deviate significantly from the preallocated capacity.

oleg-ssvlabs (approved):
Awesome changes - I think the codebase looks much cleaner now.  
I left a few comments, but none of them are strictly necessary to change.

network/discovery/logger/common.go
iurii-ssv: Applied, I guess the intent was to add stack-trace later on - but there is probably no need

operator/validator/controller.go
iurii-ssv: Yep, pretty much the only case where I wouldn't over-allocate like that would be
- if on average the actual slice size is quite small compared to pre-allocated value
- and at the same time pre-allocated is large

and one other case would be if it's "hard to predict" the size we need

protocol/v2/ssv/runner/sync_committee_aggregator.go
iurii-ssv: I have some doubts about this particular case - it might "meet [these rare conditions I've mentioned](https://github.com/ssvlabs/ssv/pull/2388#discussion_r2216078984)" ... so yeah - I don't have a strong opinion on it either

oleg-ssvlabs (approved):

api/handlers/exporter.go
```diff
@@ -353,7 +353,7 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
 		return api.BadRequestError(errors.New("either pubkeys or indices is required"))
 	}
 
-	var pubkeys []spectypes.ValidatorPK
+	pubkeys := make([]spectypes.ValidatorPK, 0, len(request.Indices))
```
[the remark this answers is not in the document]
iurii-ssv: Yeah, this is one place I missed, thanks, fixed!

operator/validator/controller.go
```diff
@@ -977,7 +977,7 @@ func (c *controller) startValidator(v *validator.Validator) (bool, error) {
 	started, err := c.validatorStart(v)
 	if err != nil {
 		validatorErrorsCounter.Add(c.ctx, 1)
-		return false, errors.Wrap(err, "could not start validator")
+		return false, fmt.Errorf("could not start validator")
```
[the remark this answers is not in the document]
iurii-ssv: Good catch, fixed!

nkryuchkov (approved):

oleg-ssvlabs (approved):

PR #2390, opened 2025-07-09
ssvsigner: log error from web3signer

nkryuchkov:
The error returned from web3signer is currently not logged because we use the default request validator from github.com/carlmjohnson/requests

The PR adds a custom validator that passes the returned error for logging

### DIFF
--- a/ssvsigner/web3signer/web3signer.go
+++ b/ssvsigner/web3signer/web3signer.go
@@ -51,13 +51,15 @@ type ListKeysResponse []phase0.BLSPubKey
 // ListKeys lists keys in Web3Signer using https://consensys.github.io/web3signer/web3signer-eth2.html#tag/Public-Key/operation/ETH2_LIST
 func (w3s *Web3Signer) ListKeys(ctx context.Context) (ListKeysResponse, error) {
 	var resp ListKeysResponse
+	var errResp string
 	err := requests.
 		URL(w3s.baseURL).
 		Client(w3s.httpClient).
 		Path(pathPublicKeys).
 		ToJSON(&resp).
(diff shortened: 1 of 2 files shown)


iurii-ssv (approved):

ssvsigner/web3signer/web3signer.go
```diff
@@ -141,31 +148,34 @@ func (w3s *Web3Signer) Sign(ctx context.Context, sharePubKey phase0.BLSPubKey, r
...
 	}
 
-	if se := new(requests.ResponseError); errors.As(err, &se) {
-		return HTTPResponseError{Err: err, Status: se.StatusCode}
+	if re := new(requests.ResponseError); errors.As(err, &re) {
+		return HTTPResponseError{Err: err, Status: re.StatusCode, ErrText: errResp}
 	}
 
-	return HTTPResponseError{Err: err, Status: http.StatusInternalServerError}
+	return HTTPResponseError{Err: err, Status: http.StatusInternalServerError, ErrText: errResp}
```
oleg-ssvlabs, line 165: Unrelated to this PR, but I think it's not ideal that we wrap all errors returned from Web3Signer in `500` status codes. I checked a couple of endpoints, and they do return various different codes, like `400`, `404`, `412`, etc.

Is there a way to check the response status code with the `requests` library? I quickly glanced through their docs but couldn't find it.

In the future I would add Open API spec to SSV Signer, possibly implement request validation (I know web3signer might validate it anyway), and consider implement retries with a backoff on network errors or so. 

Out of all of this, I think wrapping all status codes into `500` is still the biggest issue.

oleg-ssvlabs: I am approving, because it's not really related to this PR

Issue #2187, opened 2025-04-29
Too early aggregator messages


nkryuchkov:
We have a lot of ignored messages that are too early by a bit less than 6s or 12s. All of them happen only for `AGGREGATOR_RUNNER`. 

```
  |   | 2025-04-29T22:05:30.022090Z AGGREGATOR_RUNNER ssv-node-exporter-2 message was sent before slot starts, got early by 5.978048312s |  
  |   | 2025-04-29T22:05:30.021894Z AGGREGATOR_RUNNER ssv-node-exporter-2 message was sent before slot starts, got early by 5.978292044s |  
  |   | 2025-04-29T22:05:24.028281Z AGGREGATOR_RUNNER ssv-node-exporter message was sent before slot starts, got early by 11.971844519s |  
  |   | 2025-04-29T22:05:24.027145Z AGGREGATOR_RUNNER ssv-node-exporter message was sent before slot starts, got early by 11.973015211s |  
  |   | 2025-04-29T22:05:24.018872Z AGGREGATOR_RUNNER ssv-node-exporter-2 message was sent before slot starts, got early by 11.98137392s |  
  |   | 2025-04-29T22:05:24.018905Z AGGREGATOR_RUNNER ssv-node-exporter-2 message was sent before slot starts, got early by 11.981422695s |  
  |   | 2025-04-29T22:05:18.162860Z AGGREGATOR_RUNNER ssv-node-exporter message was sent before slot starts, got early by 5.837329461s |  
  |   | 2025-04-29T22:05:18.162810Z AGGREGATOR_RUNNER ssv-node-exporter-2 message was sent before slot starts, got early by 5.837355692s |  

```

nkryuchkov:
Example:

Nodes 1-4 started aggregator duty processing at the correct time:

```
  |   | {"level":"\u001b[34mINFO\u001b[0m","time":"2025-04-29T22:14:24.029746Z","name":"Controller.Validator","msg":"ℹ️ starting duty processing","pubkey":"abda05638733b2f63a8f6b1b55d54a94b70e2fba0281764e118ede694519cbd9e1be1ab113e92bba47b4430fea474acc","role":"AGGREGATOR_RUNNER","slot":312622,"role":"AGGREGATOR_RUNNER","duty_id":"AGGREGATOR-e9769-s312622-v1102338","height":310960} |  
[...]

PR #2393, opened 2025-07-10
fix: partial signature logging shows wrong slot and consensus fields

olegshmuelov:
## Problem
  When partial signature message validation fails (e.g., "early by 59.849432ms"), error logs showed:
  - `e0-s0` (slot 0) for `duty_id` instead of the actual slot
  - Consensus fields like `qbft_message_type: "proposal"` for non-consensus messages

  ## Solution
  - Return decoded message on validation error to preserve slot info for logging
  - Only populate consensus fields for consensus messages

  ## Changes
  - `partial_validation.go`: Return message on error (like consensus validation does)
  - `logger_fields.go`: Fixed consensus fields initialization
  - Added comprehensive tests with regression test

  Partially addresses #2187

### DIFF
--- a/message/validation/logger_fields.go
+++ b/message/validation/logger_fields.go
@@ -55,9 +55,7 @@ func (d LoggerFields) AsZapFields() []zapcore.Field {
 }
 
 func (mv *messageValidator) buildLoggerFields(decodedMessage *queue.SSVMessage) *LoggerFields {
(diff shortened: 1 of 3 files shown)


nkryuchkov (dismissed):
Great job

kchojn (dismissed):

kchojn (approved):

PR #2380, opened 2025-07-07
gas-limit-36: enable on exact epoch configured

iurii-ssv:
This is a follow-up to https://github.com/ssvlabs/ssv/pull/2307 so that gas-limit-36 feature will take an effect on exact epoch configured rather than after SSV node restart (after that epoch has passed), 

this also guarantes the switch will happen for all operators in SSV cluster at the same time as well.

PR #2395, opened 2025-07-14
Stage <- Main + resolved conflicts

nkryuchkov:
Porting the latest change from `main` to `stage`

Supersedes https://github.com/ssvlabs/ssv/pull/2352


kchojn (approved):

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -48,6 +55,7 @@ func NewValidatorRegistrationRunner(
 	signer ekm.BeaconSigner,
 	operatorSigner ssvtypes.OperatorSigner,
 	gasLimit uint64,
+	gasLimit36Epoch phase0.Epoch,
```
iurii-ssv: We don't need to pass this `gasLimit36Epoch` param here for **stage** branch, as the comment from above states:
```go
        // TODO: gasLimit36Epoch is temporarily living here on main branch, on stage branch ValidatorRegistrationRunner
	// has access to SSV-config from where it can just reference GasLimit36Epoch - so, eventually, we'll want to get
	// rid of this field here and replace it with SSV-config usage instead.
	gasLimit36Epoch phase0.Epoch
```
so I think we should remove this parameter (and also this struct-member-field and its `TODO` comment as well) and just use `r.BaseRunner.NetworkConfig.GetGasLimit36Epoch()` as was implemented in https://github.com/ssvlabs/ssv/pull/2380/files (but I guess I'll just close #2380 since we have this PR already)

iurii-ssv (dismissed):

kchojn (dismissed):

oleg-ssvlabs (dismissed):

iurii-ssv (dismissed):

kchojn (approved):

olegshmuelov (approved):

PR #2397, opened 2025-07-15
fix(db): only use pebble for exporter

y0sher:
### Description

We introduced `pebbledb` to be used for the exporter node as we saw some bad patterns on the badger db mem performance. It seems to have helped us and mem usage looks much better on exporter with pebble db.

The changed code was added when considering letting all operator nodes use `pebbledb` as their database engine, but this is really under tested and there are a lot of scenarios needed to be looked on including migrations more. 

This PR sets pebbledb to work only on exporter and regular operator nodes to keep using badger db. at least for the upcoming version.

CC: @Tom-ssvlabs @andrew-blox @Roy-blox @MatusKysel


storage/basedb/storage.go
```diff
@@ -7,8 +7,9 @@ import (
 
 // Options for creating all db type
 type Options struct {
-	Ctx        context.Context
-	Engine     string        `yaml:"Engine" env:"DB_ENGINE" env-default:"badger" env-description:"Database engine (badger or pebble)"`
+	Ctx context.Context
+	// TODO: support pebble for regular node operators after testing
+	//Engine     string        `yaml:"Engine" env:"DB_ENGINE" env-default:"badger" env-description:"Database engine (badger or pebble)"`
```
iurii-ssv, line 12: Just to make sure, this `Engine` config setting didn't make it into any release yet, correct ? Because if it did - we should just "ignore" it, not "comment-out completely"

cli/operator/node.go
```diff
@@ -270,15 +270,12 @@ var StartNodeCmd = &cobra.Command{
...
+		if cfg.ExporterOptions.Enabled {
 			logger.Info("using pebble db")
 			db, err = setupPebbleDB(logger, networkConfig.BeaconConfig, operatorPrivKey)
-		case "badger":
+		} else {
 			logger.Info("using badger db")
 			db, err = setupBadgerDB(logger, networkConfig.BeaconConfig, operatorPrivKey)
-		default:
-			err = fmt.Errorf("invalid db engine: %s", cfg.DBOptions.Engine)
 		}
```
iurii-ssv, line 279: This means we won't be able to run Exporter with BadgerDB anymore, right ? I guess it's fine, just calling it out in case that wasn't the exact intention here

cc @MatusKysel jfyi

iurii-ssv (approved):
LGTM, just couple of comments to make sure the code is written as intended

storage/basedb/storage.go
y0sher: you are correct, not in any release yet.

cli/operator/node.go
y0sher: yup your right, exporter been running on pebble for a while now and we are also considering switching the whole node to run on pebble. but I'm taking it out of this release in order to be able to test more.

PR #2399, opened 2025-07-15
feat(performance / 3k validators): Partially run ProcessPostConsensus phase in separate goroutines

oleg-ssvlabs:
During the SSV node test with the increased number of validators (**3k**), a few bottlenecks were identified. These bottlenecks increase the time required for consensus and duty execution overall. The main issues were observed in the **Process Consensus** and **Process Post Consensus** phases.

This PR attempts to optimize the **Process Post Consensus** phase by parallelizing execution using goroutines.

## Metrics(2846 validators, 150 samples) 
(`ssv.validator.execute_committee_duty` trace):

Stage(`c213e9873677bb9f1dd078553c56d8ed43f57a72`) branch:
Median Duration: `484.0 ms`
Average Duration: `546.49 ms`

#2340 branch (with consensus phase optimizations, 30 workers)
Median Duration: `422.5 ms` (`12.71%` faster than `stage`)
Average Duration: `445.17 ms` (`18.54%` faster than `stage`)

This branch:
Median Duration: `346.0 ms` (`28.51%` faster than `stage`)
Average Duration: `420.91ms` (`22.98%` faster than `stage`)

## List of changes
- Parallel execution of signature reconstructions in Post Consensus phase
- `PartialSigContainer` - added necessary locks for required memory synchronization
- ~`PartialSigContainer` - `GetSignatures` returns a copy now (it is slightly expensive though time is `O(n)` with `n` number of entries). It was required to multiple data races. Also, I think it isn't bad to implement immutability~
- Got rid of `github.com/pkg/errors` dependency in `committee.go` file
- Added/improved span events and attributes
- Refactor `getPostConsensusCommitteeSigners` function to improve performance
- Refactor all functions in `runner_state_helpers.go` to call `GetSignatures` instead of accessing the map directly

I did not implement worker pool here on purpose because the block that currently runs in separate goroutines does **not** do any I/O calls that can potentially be problematic with high concurrency (especially HTTP).

## NOTE:
~Targeting other feature branch. Once `optimizations` is merged, I will change the target branch to `stage`~


protocol/v2/ssv/runner/committee.go
```diff
@@ -681,23 +708,20 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 
 	logger = logger.With(fields.PostConsensusTime(cr.measurements.PostConsensusTime()))
 
-	// Submit multiple attestations
 	attestations := make([]*spec.VersionedAttestation, 0, len(attestationsToSubmit))
 	for _, att := range attestationsToSubmit {
 		if att != nil && att.ValidatorIndex != nil {
-			span.AddEvent("adding attestation to submit to beacon", trace.WithAttributes(
```
oleg-ssvlabs, line 691: There’s a limit to how many events a span can contain - `128`. For an operator with more than 128 validators, this event will exceed that limit.

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+	listener:
+		for {
+			select {
+			case err := <-errCh:
+				executionErr = err
+			case signatureResult, ok := <-signatureCh:
+				if !ok {
+					break listener
+				}
+				validatorObjects, exists := beaconObjects[signatureResult.validatorIndex]
```
oleg-ssvlabs, line 673: This entire block could have been moved into the function that runs in the goroutines and stream sync committee and attestation messages for example to two different channels. However, since the operations here seem cheap and fast, and moving them would require additional locking and troubleshooting of other data races (I've tried that), I decided to structure it this way.

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-				att := sszObject.(*spec.VersionedAttestation)
-				// Insert signature
-				att, err = specssv.VersionedAttestationWithSignature(att, specSig)
-				if err != nil {
-					anyErr = errors.Wrap(err, "could not insert signature in versioned attestation")
+	listener:
+		for {
+			select {
+			case err := <-errCh:
+				executionErr = err
```
oleg-ssvlabs, line 667: I’m not a big fan of this implementation - overriding the error like this (it was in the original version). I tried wrapping errors using `errors.Join()`, but that caused many tests to fail. We can consider creating a separate task for this.

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-			if role == spectypes.BNRoleSyncCommittee {
-				syncMsg := sszObject.(*altair.SyncCommitteeMessage)
-				// Insert signature
-				syncMsg.Signature = specSig
+					errCh <- fmt.Errorf("got post-consensus quorum but it has invalid signatures: %w", err)
+					return
+				}
 
-				syncCommitteeMessagesToSubmit[validator] = syncMsg
+				vlogger.Debug("🧩 reconstructed partial signatures committee", zap.Uint64s("signers", getPostConsensusCommitteeSigners(cr.BaseRunner.State, root)))
```
oleg-ssvlabs: `getPostConsensusCommitteeSigners` is an **expensive call**.  
The time complexity of it is **O(n × y)**, where `n` is the number of Duties and `y` is the number of Signatures.

I'm **not confident if I understand it correctly**, but for **3k validators** it could be:

**3,000(operator) × 3,000(signatures) = 9,000,000** iterations(worst case)?

This call is used **only for logs**.  
Potentially we could re-evaluate how valuable having `signers` in this log message is

Update:
I think number of signatures is equal to the cluster size (since Alan), so **3,000 * 12 = 35,000** iterations

protocol/v2/ssv/partial_sig_container.go
```diff
@@ -15,6 +17,7 @@ import (
 type SigningRoot string
 
 type PartialSigContainer struct {
+	lock sync.RWMutex
```
iurii-ssv: Nit: I guess we could leave a note that this `lock` is meant to make `PartialSigContainer` safe to use across go-routines

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-				continue
-			}
-			sszObject, exists := validatorObjs[root]
-			if !exists {
-				anyErr = errors.Wrap(err, "could not find beacon object for validator")
-				continue
-			}
+				lock.RLock()
+				share := cr.BaseRunner.Share[validatorIndex]
+				lock.RUnlock()
```
iurii-ssv: I don't think this `lock` has any effect, what was the intent behind it ?

I think `cr.BaseRunner.Share` map is initialized once upon runnerconstruction and is never updated - you can just read it here without any need for locks (since go-routing that's doing the reading has been spawned **after** runner-construction)

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+					span.AddEvent(eventMsg)
+					vlogger.Error(eventMsg, fields.Slot(cr.BaseRunner.State.StartingDuty.DutySlot()), zap.Error(err))
 
-			// Store objects for multiple submission
-			if role == spectypes.BNRoleSyncCommittee {
-				syncMsg := sszObject.(*altair.SyncCommitteeMessage)
-				// Insert signature
-				syncMsg.Signature = specSig
+					errCh <- fmt.Errorf("got post-consensus quorum but it has invalid signatures: %w", err)
+					return
```
iurii-ssv: Can probably re-use `eventMsg` here to dedup `"got post-consensus quorum but it has invalid signatures"`

protocol/v2/ssv/runner/committee.go
```diff
@@ -607,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+				}
 
-				syncCommitteeMessagesToSubmit[validator] = syncMsg
+				vlogger.Debug("🧩 reconstructed partial signatures committee", zap.Uint64s("signers", getPostConsensusCommitteeSigners(cr.BaseRunner.State, root)))
+				result := signatureResult{
+					validatorIndex: validatorIndex,
+				}
+				copy(result.signature[:], sig)
```
iurii-ssv: Perhaps this can be simplified to (I haven't tested it myself, don't forget to check if it actually works if you go for it):
```go
				result := signatureResult{
					validatorIndex: validatorIndex,
					signature:      (phase0.BLSSignature)(sig),
				}
```

protocol/v2/ssv/runner/committee.go
```diff
@@ -597,6 +596,19 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+		type signatureResult struct {
+			signature      phase0.BLSSignature
+			validatorIndex phase0.ValidatorIndex
+		}
+		var (
+			lock        sync.RWMutex
+			wg          sync.WaitGroup
+			errCh       = make(chan error, len(validators))
+			signatureCh = make(chan signatureResult, len(validators))
+		)
```
iurii-ssv: So we are processing 1 root at a time (but all validators sign that root in parallel) ? Seems like we could get more benefit out of this if we processed all the roots in parallel,

although maybe there is typically ~1 root on average and occasionally it's 2-3 for example - in that case it probably doesn't matter that much (I guess we could check in logs to see what's the real-world value for it often is)

protocol/v2/ssv/runner/committee.go
iurii-ssv: Yeah spec-tests even though they are nice to have constraint us quite a bit in a sense that we cannot change some error-strings

I've suggested a sort-of middle-ground approach for overriding expected-error-messages in spec-tests in the past - https://github.com/ssvlabs/ssv/pull/1784#discussion_r1818179653 - but talking it over with Moshe we decided that it's maybe gonna create more friction than provide benefit (just linking it here in case you are thinking of something similar)

protocol/v2/ssv/runner/committee.go
iurii-ssv: Regarding `getPostConsensusCommitteeSigners` specifically, I think it's more like O(3000 / 32 * 13) because those 3k validators are spread out into 32 slots of the epoch,

however:
- I don't even think that we need to log it at all, it doesn't seem like a useful info to have (or maybe I just don't see a use-case for it)
- I think we can pull it out of `for _, validator := range validators` loop and call once per `root` (but I'd rather delete it entirely)

iurii-ssv: Not sure if I fully get how `CommitteeRunner` works, but I'd like to share some notes here (and also have it for myself for future reference):
- every slot there are a bunch of Ethereum attestation-duties some validators need to perform (if we have 3000 validators, they will be all spread out across 1 epoch, for every epoch, say ~100 validators per slot)
- there is just 1 `CommitteeRunner` per operator (per slot), and it takes ALL those Ethereum attestation-duties and executes them  as a single CommitteeDuty, namely - it fetches `BeaconVote` (that depends only on that target `slot`, and hence ~it must be the same value for all validators in that committee, otherwise cluster won't be able to sign the result if there is no agreement on what to sign~ it doesn't have to be the same value for QBFT to finish with an agreement on the same value, that would be too much to ask, rather it needs to satisfy `ProposedValueCheckF` func)
- QBFT consensus ensures Operators agree on the same `BeaconVote` value, and in `ProcessConsensus` they partially-sign `AttestationData` that's derived from `BeaconVote` value "adapting" it to the validator that's signing - so we get N partial-signatures (where N is the amount of validators originally assigned by Ethereum to perform those attestation duties in the target slot); I think those partially-signed `AttestationData` objects still contain the same data though (differing only by something called `CommitteeIndex`) and hence have the same "signing-root" (which roughly is a hash of that data)
- and then we go into post-consensus where we just merge those partial signatures together to reconstruct full signature for each Ethereum validator-duty (the neat part is that post-consensus message aggregates those partial-signatures together so that there is no need to broadcast into p2p network each partial-signature of each validator separately)
- in `ProcessPostConsensus` some tricky stuff happens, namely `hasQuorum` turns `true` the very 1st time **some** signing-root got quorum (it might seem like we repeat all the work on every post-consensus message that results into `ProcessPostConsensus` call - but in reality we execute it once and "Skip if no quorum" and "Skip if already submitted" - so all in all it's fine). That signing-root is repeated a lot in `PartialSigContainer` due to `AttestationData` objects often containing the same data - hence we need to "Get unique roots to avoid repetition"

iurii-ssv:
@oleg-ssvlabs one other optimization we might consider for `CommitteeRunner` is to calculate `expectedPostConsensusRootsAndBeaconObjects` result once per `CommitteeRunner` (instead of once per `ProcessPostConsensus`) - seems like it iterates over all Ethereum duties for the target slot on every post-consensus message (just to validate post-consensus message) ... this might be contributing quite a lot to the overall duty cost (and/or accumulates over time), WDYT ?

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Fetched some stats regarding `getPostConsensusCommitteeSigners` execution time:

Based on 1000 samples:
- **Median:** 0.128 ms  
- **Average:** 2.241 ms  
- **P90:** 0.319 ms  
- **P99:** 90.530 ms  

P99 is really concerning. I occasionally saw it take around 200 ms.  
I believe this is mostly caused by locks, where a writer acquires the lock and blocks reads while updating signatures collection. 

I believe that log message is still needed - not only for QA but also for tools like Pulse, where it is [used](https://github.com/ssvlabs/ssv-pulse/blob/main/internal/analyzer/parser/consensus/service.go#L18) to calculate consensus times (it also reads the `signers` field).

While we shouldn't be blocked from refactoring due to external dependencies, in this case, simply moving the log out of the validators loop and logging it only once per root would substantially improve the situation.

It would be interesting to check the entire duty execution times and see if this change had any impact.

oleg-ssvlabs: 9374878de7328be8d7146b87623bc0712a75bd53 moved the log out of validators loop

protocol/v2/ssv/partial_sig_container.go
oleg-ssvlabs: fixed ac8b185ede1b0e115cc46dfe6bd58dd8c08e341c

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Good point. It seems like, as you say, there are no concurrent writes. Removed the lock ac8b185ede1b0e115cc46dfe6bd58dd8c08e341c

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: fixed 4a23df56b6d5aa56eeffb53d67b49ef27fbf55e2

protocol/v2/ssv/runner/committee.go
```diff
@@ -508,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 				continue
 			}
 
-			// Reconstruct signature
-			share := cr.BaseRunner.Share[validator]
-			pubKey := share.ValidatorPubKey
-			vlogger := logger.With(zap.Uint64("validator_index", uint64(validator)), zap.String("pubkey", hex.EncodeToString(pubKey[:])))
+			wg.Add(1)
+			go func(validatorIndex phase0.ValidatorIndex, root [32]byte, roots map[[32]byte]struct{}) {
```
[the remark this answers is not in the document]
oleg-ssvlabs: I see the point. However I think it's slightly unnecessary optimization. Passing the `deduplicatedRoots` map directly doesn't seem to be a problem.

protocol/v2/ssv/runner/committee.go
```diff
@@ -508,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-				continue
-			}
-			sszObject, exists := validatorObjs[root]
-			if !exists {
-				anyErr = errors.Wrap(err, "could not find beacon object for validator")
-				continue
-			}
+				lock.RLock()
+				share := cr.BaseRunner.Share[validatorIndex]
+				lock.RUnlock()
```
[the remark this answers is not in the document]
oleg-ssvlabs: the lock was removed.

protocol/v2/ssv/runner/committee.go
```diff
@@ -508,71 +619,86 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+		go func() {
+			wg.Wait()
+			close(signatureCh)
+		}()
+
+	listener:
+		for {
+			select {
+			case err := <-errCh:
+				executionErr = err
```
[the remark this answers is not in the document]
oleg-ssvlabs: it is by design - we want to continue collecting signatures even if errors occur. We will consider joining errors instead

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: it does seem to work (tested on stage) - ba76bb96ac9b09b902e7e2c5546151c6665833c7

oleg-ssvlabs:
> @oleg-ssvlabs one other optimization we might consider for `CommitteeRunner` is to calculate `expectedPostConsensusRootsAndBeaconObjects` result once per `CommitteeRunner` (instead of once per `ProcessPostConsensus`) - seems like it iterates over all Ethereum duties for the target slot on every post-consensus message (just to validate post-consensus message) ... this might be contributing quite a lot to the overall duty cost (and/or accumulates over time), WDYT ?

I might not fully understand the flow, however from what I see in traces, the `ProcessPostConsensus` method can be triggered multiple times for a single duty, and there seem to be three possible scenarios:
- **No quorum**
- **No running duty** (I assume this happens when the duty has already finished and a post-consensus message arrives afterwards/late)
- **Quorum is achieved**, and the method submits attestations or sync committees to the beacon

In scenarios 1 and 2, `expectedPostConsensusRootsAndBeaconObjects` isn't triggered. Scenario 3 seems to happen only once.

So based on this, it appears that `expectedPostConsensusRootsAndBeaconObjects` is executed only once per duty. Wdyt?

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Yes, you're correct.

From what I can see, there are either **1 or 2 block roots**:
- 1 for attester duties
- 2 for attester + sync committee duties (when validators are assigned to sync committees by the beacon)

The number of validators involved in sync committee duties is much smaller, so the performance impact would be minimal if we chose to run this in parallel for each block root instead.

protocol/v2/ssv/partial_sig_container.go
```diff
@@ -65,14 +74,26 @@ func (ps *PartialSigContainer) GetSignature(validatorIndex phase0.ValidatorIndex
...
+	if signatures == nil {
 		return nil
 	}
-	return ps.Signatures[validatorIndex][signingRootHex(signingRoot)]
+
+	signaturesCopy := make(map[spectypes.OperatorID]spectypes.Signature, len(signatures))
+	maps.Copy(signaturesCopy, signatures)
+
+	// Return a copy to avoid external mutation and avoid data races
+	return signaturesCopy
```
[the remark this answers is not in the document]
oleg-ssvlabs: I removed this code. Data races previously were occurring, because one of the callers of that function was running in separate goroutines and there were concurrent writes.

iurii-ssv:
> > @oleg-ssvlabs one other optimization we might consider for `CommitteeRunner` is to calculate `expectedPostConsensusRootsAndBeaconObjects` result once per `CommitteeRunner` (instead of once per `ProcessPostConsensus`) - seems like it iterates over all Ethereum duties for the target slot on every post-consensus message (just to validate post-consensus message) ... this might be contributing quite a lot to the overall duty cost (and/or accumulates over time), WDYT ?
> 
> I might not fully understand the flow, however from what I see in traces, the `ProcessPostConsensus` method can be triggered multiple times for a single duty, and there seem to be three possible scenarios:
> 
> * **No quorum**
> * **No running duty** (I assume this happens when the duty has already finished and a post-consensus message arrives afterwards/late)
> * **Quorum is achieved**, and the method submits attestations or sync committees to the beacon
> 
> In scenarios 1 and 2, `expectedPostConsensusRootsAndBeaconObjects` isn't triggered. Scenario 3 seems to happen only once.
> 
> So based on this, it appears that `expectedPostConsensusRootsAndBeaconObjects` is executed only once per duty. Wdyt?

I got `expectedPostConsensusRootsAndBeaconObjects` and `expectedPostConsensusRootsAndDomain` mixed up treating it as the same thing. So yes what you describe seems correct, as long as "Scenario 3" is triggered mostly ~1 time I guess it's fine to re-do that computation.

protocol/v2/ssv/runner/committee.go
iurii-ssv: Yeah, I guess if we've measured what it looks like in practice (how many different roots we get) then it's fine as it is.

protocol/v2/ssv/partial_sig_container.go
```diff
@@ -84,11 +83,7 @@ func (ps *PartialSigContainer) GetSignatures(validatorIndex phase0.ValidatorInde
 		return nil
 	}
 
-	signaturesCopy := make(map[spectypes.OperatorID]spectypes.Signature, len(signatures))
-	maps.Copy(signaturesCopy, signatures)
-
-	// Return a copy to avoid external mutation and avoid data races
-	return signaturesCopy
+	return signatures
 }
```
iurii-ssv: I think the copy returned from `GetSignatures` still **needs** to be a copy since once we return from `GetSignatures` func we release the lock ... and then `AddSignature` method can potentially update this `map` we returned concurrently, but the reader will not be locking the `lock` to properly syncrhonize with that update,

does it make sense ?

I see you mentioned [here](https://github.com/ssvlabs/ssv/pull/2399#discussion_r2213083630) that the caller of `GetSignatures` performed some writes on the result returned (and that's what was reported as race) ? Note however, it cannot/shouldn't perform reads on it either (unless we make it a copy) - that would also be racy because we return a map-reference which points to structures that can/will be updated concurrently (in `AddSignature`) - so syncrhonization is necessary for reader to see those updates correctly.

And TBH it doesn't matter really how we **currently** use `PartialSigContainer` - we should design it such that it cannot be misused (this PR must ensure it's tread-safe to use in the future).

Regarding copy-on-write co-pilot suggestion, we'd need to estimate/measure the usage to tell if it makes sense or not (perhaps we have more writes - and that would only make it worse).

protocol/v2/ssv/runner/committee.go
```diff
@@ -610,97 +621,113 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-			if role == spectypes.BNRoleSyncCommittee {
-				syncMsg := sszObject.(*altair.SyncCommitteeMessage)
-				// Insert signature
-				syncMsg.Signature = specSig
+				vlogger.Debug("🧩 reconstructed partial signature")
+
+				signatureCh <- signatureResult{
+					validatorIndex: validatorIndex,
+					signature:      (phase0.BLSSignature)(sig),
+				}
```
iurii-ssv, line 652: Could maybe add `validatorIndex` for this log-line to be more meaningful ? I would also log `sig` as well for completeness

protocol/v2/ssv/partial_sig_container.go
oleg-ssvlabs: I see what you mean.

My thoughts were that `GetSignatures` is called in multiple places, and copying the map (which may be large - I'm not sure about the size) could be **unnecessarily** expensive. So, if it's not currently causing an issue, I thought it might be better to avoid the extra cost. If a race condition does appear later, we’d likely catch it and fix it.

That said, since the lock is already implemented in `PartialSigContainer`, it’s only "semi-functioning" - because maps are passed by reference, and iterating over the map returned by `GetSignatures` can still cause race conditions in case of concurrent writes.

I’ll add it back.

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: the `vlogger` is [already](https://github.com/ssvlabs/ssv/pull/2399/files#diff-c615c5d14e307fe437330a4144eaa9c319e9f30cf23aef1c724d8712057b1868R631) enriched with validator index and pubkey

iurii-ssv (dismissed):

iurii-ssv (approved):

protocol/v2/ssv/runner/committee.go
```diff
@@ -610,97 +621,113 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
-				if err != nil {
-					anyErr = errors.Wrap(err, "could not insert signature in versioned attestation")
+				validatorObjects, exists := beaconObjects[signatureResult.validatorIndex]
+				if !exists {
+					executionErr = fmt.Errorf("could not find beacon object for validator")
 					continue
 				}
+				sszObject, exists := validatorObjects[root]
+				if !exists {
+					executionErr = fmt.Errorf("could not find beacon object for validator")
```
MatusKysel: wrong error

oleg-ssvlabs: good catch, we actually have this issue in [stage branch](https://github.com/ssvlabs/ssv/blob/stage/protocol/v2/ssv/runner/committee.go#L651) as well. 
I will fix it

oleg-ssvlabs: fixed e991e3c145627e9f3901a9aa32bba78261c51467

iurii-ssv (dismissed):

protocol/v2/ssv/partial_sig_container.go
```diff
@@ -15,6 +17,9 @@ import (
 type SigningRoot string
 
 type PartialSigContainer struct {
+	// lock protects access to the Signatures map to prevent concurrent read/write access
+	// from multiple goroutines
+	lock sync.RWMutex
```
nkryuchkov: I think renaming it to `signaturesMu` would make it self-explanatory because the comment is quite obvious except pointing what variable it guards

protocol/v2/ssv/runner/committee.go
```diff
@@ -610,97 +621,113 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+		}()
 
-			} else if role == spectypes.BNRoleAttester {
-				// Only mark as safe if this is an attester role
-				// We want to mark the validator as safe as soon as possible to minimize unnecessary delays in enabling signing.
-				// The doppelganger check is not performed for sync committee duties, so we rely on attester duties for safety confirmation.
-				cr.doppelgangerHandler.ReportQuorum(validator)
+	listener:
+		for {
+			select {
```
nkryuchkov, line 663: Probably we could also check `ctx` to allow cancellation

protocol/v2/ssv/runner/committee.go
```diff
@@ -610,97 +621,113 @@ func (cr *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap
...
+					validatorIndex: validatorIndex,
+					signature:      (phase0.BLSSignature)(sig),
+				}
+			}(validator, root, deduplicatedRoots)
+		}
 
-				syncCommitteeMessagesToSubmit[validator] = syncMsg
+		go func() {
+			wg.Wait()
+			close(signatureCh)
```
nkryuchkov, line 658: do we need to close `errCh` too?

iurii-ssv: Just my 2 cents, I don't think there is a need to `close` channel that will be garbage-collected anyway. Closing a channel is typically used to signal "the end of something", and for this use-case I think the close on `signatureCh` takes care of that.

But if we add close for `errCh` we'll also need to process `selects` on it accordingly (or we will get stuck in a loop perpetually receiving `nil` I think)

protocol/v2/ssv/partial_sig_container.go
oleg-ssvlabs: fixed a848cf06936648cbf9c1c6ea7f0c14713050e510

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: Good point. Fixed d0b8f2a79ec2ad3e6335f3cfa9ae47a9417197bf

protocol/v2/ssv/runner/committee.go
oleg-ssvlabs: That's basically what @iurii-ssv is referring to.  
The `signatureCh` is explicitly closed to signal that the channel producers have finished, and the consumer can [stop listening](https://github.com/ssvlabs/ssv/blob/post-consensus-phases/protocol/v2/ssv/runner/committee.go#L669).  
The error channel should be garbage collected once there are no producers or consumers referencing it, i.e., after the listener exits the `select` loop.

nkryuchkov (commented):
LGTM

Issue #2100, opened 2025-03-25
avoid passing scheduler logger to `ExecuteCommitteeDuty` — let controller manage context


olegshmuelov:
I agree with the TODO here, passing a `logger` into `ExecuteCommitteeDuty` isn’t ideal, especially considering how and where it’s used.

**Just to add:** `ExecuteCommitteeDuty` is implemented in the validator controller (i.e., a different package). This logger is then passed deeper into the `Committee`, where it’s still labeled from the `DutyScheduler`, making logs misleading in terms of origin and responsibility.

Example:
```go
logger := s.loggerWithCommitteeDutyContext(committee) // From scheduler
s.dutyExecutor.ExecuteCommitteeDuty(ctx, logger, committee.id, duty)
```

Eventually ends up in:
```go
func (c *Committee) OnExecuteDuty(ctx context.Context, logger *zap.Logger, msg *types.EventMsg) error {
         ...
	if err := c.StartDuty(ctx, logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee duty: %w", err)
	}

	if err := c.StartConsumeQueue(logger, executeDutyData.Duty); err != nil {
		return fmt.Errorf("could not start committee consume queue: %w", err)
	}
         ...
}
```
So now we’ve got logs coming from inside `Committee`, but they're still labeled as if they came from the `DutyScheduler`.

Ideally, the `controller` / `Committee` (probably they already do) should build and manage their own logger internally using the proper context (e.g., `controller.logger.With(...)` or via a stored logger field).

We should create a GitHub issue to track and properly refactor this, so that the logger is owned and managed within the controller layer, not passed from upstream logic.

_Originally posted by @olegshmuelov in https://github.com/ssvlabs/ssv/pull/1968#pullrequestreview-2713640158_

iurii-ssv:
Working on it (so stale-bot doesn't close)

iurii-ssv:
This has been addressed by https://github.com/ssvlabs/ssv/pull/2400

PR #2400, opened 2025-07-15
duties: correct logger usage

iurii-ssv:
The general rule/approach this PR tries to follow with respect to "who should initialize logger" and "how logger should be passed around" is:
- every service/component (eg. `scheduler`, `controller`, etc.) have their own "named logger" and they will use it whenever they need to log something; that is to say - they won't ever accept somebody-else's logger (public-facig API should't have logger as an argument - the only "exception" is when method is meant public simply because we want to unit-test it in some external package)
- withing a certain service/component it makes sense to re-use the same logger enriched with some data (eg. for `Committee` component we can create logger with some additional duty-related data once and pass it around between this component's private methods; the only caveat is that some methods that are meant to be private are exposed as public simply because of how unit-tests are written, and on top of that there exists quite a bit of inderection in this code with abstract handler `funcs` and such ... so the resulting code doesn't looks any simpler than what it was - but it should work properly now with respect to logging)

Closes https://github.com/ssvlabs/ssv/issues/2100


operator/validator/controller.go
```diff
@@ -629,45 +628,30 @@ func (c *controller) ExecuteDuty(ctx context.Context, logger *zap.Logger, duty *
 		trace.WithLinks(trace.LinkFromContext(ctx)))
 	defer span.End()
 
-	// because we're using the same duty for more than 1 duty (e.g. attest + aggregator) there is an error in bls.Deserialize func for cgo pointer to pointer.
-	// so we need to copy the pubkey val to avoid pointer
-	pk := make([]byte, 48)
-	copy(pk, duty.PubKey[:])
```
iurii-ssv, line 670: This shouldn't be a concern anymore since now we are using arrays (not slices) here - so they are always getting copied, so I'm removing this comment.

protocol/v2/ssv/validator/startup.go
```diff
@@ -25,33 +18,14 @@ func (v *Validator) Start(logger *zap.Logger) (started bool, err error) {
...
-			if s.ValidatorPubKey == v.Share.ValidatorPubKey {
-				share = s
-				break
-			}
-		}
-
-		if share == nil { // TODO: handle missing share?
-			logger.Warn("❗ share is missing", fields.Role(role))
-			continue
-		}
```
iurii-ssv, line 42: This check for `share is missing` is redundant here since every Runner checks for share not being `nil` during initialization, so I'm removing it to simplify things.

protocol/v2/ssv/validator/committee.go
```diff
@@ -55,7 +56,7 @@ func NewCommittee(
 	ctx context.Context,
 	cancel context.CancelFunc,
 	logger *zap.Logger,
-	beaconConfig networkconfig.Beacon,
+	networkConfig networkconfig.Network,
```
nkryuchkov, line 59: Why not beacon config?

nkryuchkov (approved):
awesome change

protocol/v2/ssv/validator/committee.go
```diff
@@ -33,10 +33,11 @@ type CommitteeRunnerFunc func(slot phase0.Slot, shares map[phase0.ValidatorIndex
 
 type Committee struct {
 	logger *zap.Logger
+
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	beaconConfig networkconfig.Beacon
+	networkConfig networkconfig.Network
```
oleg-ssvlabs: May I ask what was the reason for changing this?

It seems that `Beacon` is a slimmer abstraction compared to `Network`, and the `Committee` object doesn't seem to require anything beyond what `Beacon` provides.

protocol/v2/ssv/validator/duty_executer.go
oleg-ssvlabs, line 1: nit:
duty_executer.go -> duty_executor.go

protocol/v2/ssv/validator/duty_executer.go
```diff
@@ -2,16 +2,41 @@ package validator
...
 	"go.uber.org/zap"
 
 	"github.com/ssvlabs/ssv/logging/fields"
 	"github.com/ssvlabs/ssv/observability"
+	"github.com/ssvlabs/ssv/protocol/v2/message"
+	"github.com/ssvlabs/ssv/protocol/v2/ssv/queue"
 	"github.com/ssvlabs/ssv/protocol/v2/types"
 )
 
+func (v *Validator) ExecuteDuty(ctx context.Context, duty *spectypes.ValidatorDuty) error {
```
oleg-ssvlabs: It’s certainly a really good change to move this logic out of the Controller.

The only thing I’m unsure about is having methods with the same receiver spread across multiple files, and even having methods with different receivers within the same file.

I believe we already follow this pattern in the `validator` package, so refactoring it (creating separate objects like `ValidatorExecutor` as an example) would be a fairly large task.

That said, it might be something worth considering eventually - especially if these methods aren’t tightly coupled to the state of the `Validator` object.

protocol/v2/ssv/validator/duty_executer.go
```diff
@@ -77,3 +129,47 @@ func (c *Committee) OnExecuteDuty(ctx context.Context, logger *zap.Logger, msg *
 	span.SetStatus(codes.Ok, "")
 	return nil
 }
+
+// createDutyExecuteMsg returns ssvMsg with event type of execute duty
+func createDutyExecuteMsg(duty *spectypes.ValidatorDuty, pubKey phase0.BLSPubKey, domain spectypes.DomainType) (*spectypes.SSVMessage, error) {
```
oleg-ssvlabs: potentially you can merge `createDutyExecuteMsg` and `createCommitteeDutyExecuteMsg` together into a single generic function.
Like this:
```Go
func createDutyMsg[T spectypes.Duty](duty T, msgIdentifier []byte, role spectypes.RunnerRole, domain spectypes.DomainType) (*spectypes.SSVMessage, error) {
	data, err := json.Marshal(duty)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal duty data: %w", err)
	}

	return dutyDataToSSVMsg(domain, msgIdentifier, role, data)
}
```

and call them like:
```Go
createDutyMsg(duty, duty.PubKey[:], duty.RunnerRole(), v.NetworkConfig.GetDomainType())
createDutyMsg(duty, c.CommitteeMember.CommitteeID[:], spectypes.RoleCommittee, c.networkConfig.GetDomainType())
```

protocol/v2/ssv/validator/startup.go
```diff
@@ -25,33 +18,14 @@ func (v *Validator) Start(logger *zap.Logger) (started bool, err error) {
 	if !ok {
 		return false, errors.New("network does not support subscription")
```
oleg-ssvlabs: It’s unrelated, but while we’re here - should we consider removing the `github.com/pkg/errors` package and replacing it with either `fmt.Errorf()` or the standard library’s `errors` package?

protocol/v2/ssv/validator/committee.go
iurii-ssv: Yeah, in general I agree with slimmer-abstraction aspect,

but I've moved some code from `controller` to `Committee` (specifically created `Committee.ExecuteDuty` method) and it needs to do`c.networkConfig.GetDomainType()` which is available on `Network` but not on `Beacon`

protocol/v2/ssv/validator/duty_executer.go
iurii-ssv: Yeah, that's how the code was originally organized - and if we want to change it better do it in a dedicated PR.

I myself prefer to have methods with the same receiver all in 1 file.

protocol/v2/ssv/validator/duty_executer.go
iurii-ssv: That's an interesting take, however keeping these as separate `func` probably makes sense still because they just happen to have similar signatures "by accident" - doesn't mean func-signatures will stay similar in the future

and I think refactoring it this way has a bit more implications than just swapping out couple of func-calls because we have these wrapper-structs (and I assume we'll want to combine these as well):
```
type ExecuteDutyData struct {
	Duty *types.ValidatorDuty
}

type ExecuteCommitteeDutyData struct {
	Duty *types.CommitteeDuty
}
```
but these ^ are also used elsewhere, so it's not trivial to replace them with template-based versions (maybe it is, but we'd need to carefully test the code involved to make sure we didn't change encoded format, for example).

So I'm not against such a change, but I think it's better to do it in a separate PR, and frankly for the amount of effort that might be needed it might not worth bothering with - but we could create a Github issue to take a look at it later :)

protocol/v2/ssv/validator/startup.go
iurii-ssv: Yep, applied

iurii-ssv:
@greptileai could you review this PR for me pls ?

protocol/v2/ssv/validator/duty_executor.go
```diff
@@ -0,0 +1,175 @@
...
+	logger := c.logger.
+		With(fields.CommitteeID(c.CommitteeMember.CommitteeID)).
+		With(fields.DutyID(committeeDutyID)).
+		With(fields.Role(duty.RunnerRole())).
+		With(fields.CurrentSlot(c.networkConfig.EstimatedCurrentSlot())).
+		With(fields.Slot(duty.Slot)).
+		With(fields.Epoch(dutyEpoch)).
+		With(fields.StartTimeUnixMilli(c.networkConfig.GetSlotStartTime(duty.Slot)))
+
+	return c.OnExecuteDuty(ctx, logger, dec.Body.(*types.EventMsg))
```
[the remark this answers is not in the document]
iurii-ssv: Right, this code however was moves "as is" from another file and hence this PR doesn't change the original behavior. That said, I'll keep an eye on it during testing - if it does `panic` we'll spot it there.

protocol/v2/ssv/validator/duty_executer.go
oleg-ssvlabs: I see what you mean, I think I missed these wrappers. 
The generic type parameter could be less restrictive `spectypes.Duty` -> `any` and function could be called:

```Go
createDutyMsg(types.ExecuteCommitteeDutyData{Duty: duty}, c.CommitteeMember.CommitteeID[:], spectypes.RoleCommittee, c.networkConfig.GetDomainType())
```

Although I don't think it's really that valuable optimization, so no need changing the current implementation.

nkryuchkov (approved):

nkryuchkov (dismissed):

oleg-ssvlabs (dismissed):

operator/validator/controller.go
```diff
@@ -615,7 +613,7 @@ func (c *controller) GetValidator(pubKey spectypes.ValidatorPK) (*validator.Vali
 	return c.validatorsMap.GetValidator(pubKey)
 }
 
-func (c *controller) ExecuteDuty(ctx context.Context, logger *zap.Logger, duty *spectypes.ValidatorDuty) {
+func (c *controller) ExecuteDuty(ctx context.Context, duty *spectypes.ValidatorDuty) {
 	dutyID := fields.FormatDutyID(c.networkConfig.EstimatedEpochAtSlot(duty.Slot), duty.Slot, duty.Type, duty.ValidatorIndex)
```
y0sher, line 651: DutyID should be added to all logs in this func.

operator/validator/controller.go
```diff
@@ -631,45 +629,30 @@ func (c *controller) ExecuteDuty(ctx context.Context, logger *zap.Logger, duty *
...
+		c.logger.Error("could not execute validator duty", zap.Error(err))
+		span.SetStatus(codes.Error, err.Error())
+		return
 	}
 
 	span.SetStatus(codes.Ok, "")
 }
 
-func (c *controller) ExecuteCommitteeDuty(ctx context.Context, logger *zap.Logger, committeeID spectypes.CommitteeID, duty *spectypes.CommitteeDuty) {
+func (c *controller) ExecuteCommitteeDuty(ctx context.Context, committeeID spectypes.CommitteeID, duty *spectypes.CommitteeDuty) {
```
y0sher, line 694: lets reconstruct dutyID and add it to the logs, this was the purpose of passing down the logger. we want to be able to search widely with dutyid

operator/validator/controller.go
y0sher: this might be relevant to slot, epoch and potentially more fields.

operator/validator/controller.go
y0sher: this might be relevant to slot, epoch and potentially more fields.

operator/validator/controller.go
iurii-ssv: Right, I myself was conflicted on whether or not we want to bother creating a dedicated logger with DutyID (and slot/epoch, etc.) in this **func**, and I didn't for now to keep things simple. The problem is this kind of pattern solves the issue only partially - there are plenty of other places where we log duty-related stuff without DutyID (or even slot/epoch) added ... which is why we need to implement a more general-purpose/robust [mechanism for duty-trace-id](https://github.com/ssvlabs/ssv/issues/1828) to log it everywhere.

But I guess you are right, until then we should have some sort of work-around - I've added [a commit here](https://github.com/ssvlabs/ssv/pull/2400/commits/228518ade0d4379919986ab70f902d55d3f4ec0c) now to address it for now.

operator/validator/controller.go
iurii-ssv: Fixed, see also https://github.com/ssvlabs/ssv/pull/2400#discussion_r2266296287

y0sher (approved):

oleg-ssvlabs (dismissed):
lgtm

oleg-ssvlabs (approved):

PR #2404, opened 2025-07-17
feat(metrics) - New validator status - 'participating_attesting'

oleg-ssvlabs:


iurii-ssv (approved):
IMO it would be nice to rename `ParticipatingAndAttesting` -> `Attesting` in all places (because if we are attesting we are 100% participating)

operator/validator/observability.go
```diff
@@ -19,15 +19,16 @@ const (
...
-	statusNotFound      validatorStatus = "not_found"
-	statusActive        validatorStatus = "active"
-	statusSlashed       validatorStatus = "slashed"
-	statusExiting       validatorStatus = "exiting"
-	statusNotActivated  validatorStatus = "not_activated"
-	statusPending       validatorStatus = "pending"
-	statusNoIndex       validatorStatus = "no_index"
-	statusUnknown       validatorStatus = "unknown"
+	statusParticipatingAttesting validatorStatus = "participating_attesting"
+	statusParticipating          validatorStatus = "participating"
```
y0sher: we should probably add some comment about what's the diff between participating and participating_attesting , here and in other places..

zktaiga: Agree, I think it's a bit confusing in its current state.

iurii-ssv: What do you think of renaming `participating_attesting` -> `attesting` (in all places) ? I think that's roughly what it actually means (if we assume the `&& !Liquidated` part as a given)

nkryuchkov (dismissed):

operator/validator/observability.go
oleg-ssvlabs: @iurii-ssv Do you mean renaming the metric status, or also in other places (e.g., Share status methods)?

@y0sher @zktaiga It's already quite easy to navigate and see exactly how the status is set, which leads to [this line](https://github.com/ssvlabs/ssv/blob/stage/protocol/v2/types/ssvshare.go#L136) with a comment that explains the status. Adding another comment as you suggest would essentially be copy-pasting what's already written [here](https://github.com/ssvlabs/ssv/blob/stage/protocol/v2/types/ssvshare.go#L134).

I was thinking about keeping the status on the metric 1:1 aligned with the share statuses. Each status already has a comment explaining its meaning.

If you still think it is better to add a comment here, I don't have a problem with that

iurii-ssv: > @iurii-ssv Do you mean renaming the metric status, or also in other places (e.g., Share status methods)?

I think we should rename everything, there is no reason to call it `ParticipatingAndAttesting` if we can call it `Attesting` (since if you are attesting - you are for sure participating),

I think we sort-of discussed it previously and decided to not rename because "it might be confusing with the way the underlying 3rd-party library" uses these terms ... but IMO it's better to have clean & short names on our side for our sanity (regardless of how it maps onto 3rd-party concepts). So I'm bringing it up again simply because now we have this complexity creeping into our metric-names, and who knows where it will end up in in the future.

oleg-ssvlabs: Both `IsParticipatingAndAttesting` and `IsAttesting` methods are used outside of the `types` package.

`IsAttesting`, apart from tests, is only used [here](https://github.com/ssvlabs/ssv/blob/stage/message/validation/validation.go#L302). So, if we could replace it with `IsParticipatingAndAttesting`, we could deprecate the current implementation of `IsAttesting` and rename `IsParticipatingAndAttesting`.  

Probably worth looking into it.  
Wdyt @iurii-ssv @y0sher?

iurii-ssv: Well, yes I think we can simply replace it (and get rid of the existing `IsAttesting` method - so we can free up the name to do the re-naming I'm talking about) because we are checking for `if share.Liquidated` just a couple of lines prior.

oleg-ssvlabs: @iurii-ssv yes, I agree. We do not need statuses that aren't utilized anywhere in the code. 
Do you mean something like [this](https://github.com/ssvlabs/ssv/pull/2421)?

iurii-ssv: > Do you mean something like [this](https://github.com/ssvlabs/ssv/pull/2421)?

Yes 👍

oleg-ssvlabs: Fixed names now after merging above mentioned PR. `participating_attesting` -> `attesting`

PR #2405, opened 2025-07-18
WIP - Observability pkgs

oleg-ssvlabs:


cli/generate_operator_keys.go
```diff
@@ -5,24 +5,23 @@ import (
 	"os"
 	"path/filepath"
 
+	ssv_log "github.com/ssvlabs/ssv/observability/log"
```
MatusKysel: are you sure about this ? I think we don't use `_` in our project e.g. `ssvtypes`, `spectypes` ...

network/discovery/options.go
```diff
@@ -4,14 +4,14 @@ import (
 	"crypto/ecdsa"
 	"net"
 
-	"github.com/ethereum/go-ethereum/log"
+	eth_logger "github.com/ethereum/go-ethereum/log"
```
MatusKysel: same here

observability/traces/provider.go
```diff
@@ -0,0 +1,30 @@
+package traces
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/contrib/exporters/autoexport"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdk_trace "go.opentelemetry.io/otel/sdk/trace"
```
MatusKysel: also here

utils/boot_node/node.go
```diff
@@ -162,19 +161,19 @@ func (n *bootNode) createListener(ipAddr string, port uint16, privateKey *ecdsa.
 	}
 	conn, err := net.ListenUDP(networkVersion, udpAddr)
 	if err != nil {
-		log.Fatal(err)
+		n.logger.Fatal(err.Error())
 	}
 	localNode, err := n.createLocalNode(privateKey, ip, port)
 	if err != nil {
-		log.Fatal(err)
+		n.logger.Fatal(err.Error())
```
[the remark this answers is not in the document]
oleg-ssvlabs: fixed 28aa9ef6a04b8c4e5934f2050cd452deca227069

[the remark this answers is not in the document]
oleg-ssvlabs: fixed 28aa9ef6a04b8c4e5934f2050cd452deca227069

utils/boot_node/node.go
```diff
@@ -162,19 +161,19 @@ func (n *bootNode) createListener(ipAddr string, port uint16, privateKey *ecdsa.
...
 	}
 
 	listener, err := discover.ListenV5(conn, localNode, discover.Config{
 		PrivateKey:   privateKey,
 		V5ProtocolID: &n.ssvConfig.DiscoveryProtocolID,
 	})
 	if err != nil {
-		log.Fatal(err)
+		n.logger.Fatal(err.Error())
```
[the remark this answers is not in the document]
oleg-ssvlabs: fixed 28aa9ef6a04b8c4e5934f2050cd452deca227069

cli/generate_operator_keys.go
oleg-ssvlabs: You are right, fixed a12fe9e6e19b6c80f68d7912f6c5d686ce50fcc1

network/discovery/options.go
oleg-ssvlabs: fixed e6d9711e0be724376264d5cbf63e28d94aa65a72

observability/traces/provider.go
oleg-ssvlabs: fixed e6d9711e0be724376264d5cbf63e28d94aa65a72

observability/log/names.go
```diff
@@ -0,0 +1,36 @@
...
+	NameSignerStorage        = "SignerStorage"
+	NameValidator            = "Validator"
+	NameWSServer             = "WSServer"
+	NameConnHandler          = "ConnHandler"
+	NameSlotTicker           = "SlotTicker"
+	NameShareMetadataSyncer  = "ShareMetadataSyncer"
+
+	NameBadgerDBLog            = "BadgerDBLog"
+	NameBadgerDBReporting      = "BadgerDBReporting"
+	NameCreateThreshold        = "CreateThreshold"
```
iurii-ssv: I'm not sure why we have these names separated into 2 sections by a newline, if there is some reason for it (I don't see any obvious ones) we should probably leave a comment about that. Or just remove the newline.

observability/traces/provider.go
```diff
@@ -0,0 +1,30 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+func InitializeProvider(ctx context.Context, resources *resource.Resource, isEnabled bool) (trace.TracerProvider, func(context.Context) error, error) {
+	if !isEnabled {
+		return noop.NewTracerProvider(), nil, nil
+	}
```
iurii-ssv: I've noticed both `metrics.InitializeProvider` and `traces.InitializeProvider` return `nil` func here while error is also `nil`.

This can potentially lead to nil-pointer dereference, depending on how we handle it:
- looks like it does in our case
- regardless, I'd rather design it such that the caller wouldn't need to worry about it in the first place

So, I think we just need to return a no-op func (for both `metrics.InitializeProvider` and `traces.InitializeProvider`)  ?

observability/traces/trace.go
```diff
@@ -11,12 +11,21 @@ import (
...
+var logger *zap.Logger
+
 const traceIDByteLen = 16
 
-// TraceContext returns a new context with a deterministic trace ID based on the input string.
+func InitLogger(l *zap.Logger) {
+	if l == nil {
+		l = zap.NewNop()
+	}
+	logger = l
```
iurii-ssv: I think there is no reason to handle `if l == nil` case in all 3 `InitLogger` funcs we have,

this kind of patter should be used sparingly IMO only when there is no other way around it because it can lead to hard-to-debug issues, wouldn't you agree ?

observability/metrics/metric.go
```diff
@@ -8,9 +8,19 @@ import (
 	"go.uber.org/zap"
 )
 
-var SecondsHistogramBuckets = []float64{0, 0.001, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10}
+var (
+	logger                  *zap.Logger
+	SecondsHistogramBuckets = []float64{0, 0.001, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10}
+)
```
iurii-ssv: I think we should leave a comment for why we have this global logger here:
```go
// logger is defined as global var here to keep package API as simple as possible (instead of returning error we log them with this logger in some places)
```

observability/traces/trace.go
```diff
@@ -11,12 +11,21 @@ import (
 	"go.uber.org/zap"
 )
 
+var logger *zap.Logger
```
iurii-ssv: ditto: I think we should leave a comment for why we have this global logger here:
```go
// logger is defined as global var here to keep package API as simple as possible (instead of returning error we log them with this logger in some places)
```

observability/logger.go
```diff
@@ -1,13 +1,23 @@
...
 		l = zap.NewNop()
 	}
-	logger = l.Named("Observability")
+
+	logger = l.Named(log.NameObservability)
+	metrics.InitLogger(logger)
+	traces.InitLogger(logger)
+
 	return logger
 }
```
iurii-ssv: For `observability` package specifically, I don't think we need to define a `var logger *zap.Logger` - we could get rid of it if we simply add logger to `buildResources` func as a param (this is the only place where it's used).

This will keep things simpler overall, with all the global loggers we have around I'm worried it becomes way to easy to use the wrong one (or is just time-consuming to figure out which one should be used in any particular case).

observability/traces/trace.go
oleg-ssvlabs: I agree, in this case I think it is unnecessary.

I like these "defensive" patterns so the package does not allow external consumers to do something they shouldn't do. However, it depends on the scope. In this case, the consumer is very close, so I don't think it's really needed. Similar patterns are most useful when building libraries or when the interaction with the package feels complex or so, which is not this case :)

observability/log/names.go
oleg-ssvlabs: agree a947ef6845def26f3374f3b6838f45ac1de4f717

observability/traces/provider.go
oleg-ssvlabs: Good catch, fixed 6b6051c5444db611b460c1b36b04c4eec5b8d1d4

observability/traces/trace.go
oleg-ssvlabs: fixed cf610e0757644f997845f5c19b24b6cb9867fca5

observability/metrics/metric.go
oleg-ssvlabs: fixed 7f3513a7c9a702cbf36491c1a5e8ab7bebe7a734

observability/traces/trace.go
oleg-ssvlabs: fixed 7f3513a7c9a702cbf36491c1a5e8ab7bebe7a734

observability/logger.go
oleg-ssvlabs: yes, I was actually thinking about it as well. Then I thought maybe there will be more occasions where global could be needed. Anyway, for now removed it cf610e0757644f997845f5c19b24b6cb9867fca5

y0sher (approved):
lgtm

PR #2407, opened 2025-07-18
re-enable duty scheduler for exporter

MatusKysel:
This PR will replace https://github.com/ssvlabs/ssv/pull/2406 
We need to re-enable dutyscheduler that was disable - to make exporter more stable (many restarts on unavailable BNs), but we need dutyscheduler to fetch duties so we can pass msg validation

### DIFF
--- a/operator/node.go
+++ b/operator/node.go
@@ -3,9 +3,6 @@ package operator
 import (
 	"context"
 	"fmt"
-	"os"
-	"os/signal"
-	"syscall"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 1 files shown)


nkryuchkov (commented):
@MatusKysel Could you please explain the difference with https://github.com/ssvlabs/ssv/pull/2406?

operator/node.go
```diff
@@ -124,12 +121,10 @@ func (n *Node) Start() error {
 		}
 	}()
 
-	if !n.exporterOptions.Enabled {
-		// Start the duty scheduler, and a background goroutine to crash the node
-		// in case there were any errors.
-		if err := n.dutyScheduler.Start(ctx); err != nil {
-			return fmt.Errorf("failed to run duty scheduler: %w", err)
-		}
```
MatusKysel, line 132: @nkryuchkov this is the difference https://github.com/ssvlabs/ssv/pull/2406

nkryuchkov: @MatusKysel yeah, I meant, could you please write some short PR description about whether it's going to supersede https://github.com/ssvlabs/ssv/pull/2406 or has another purpose? And about why we need to call `n.dutyScheduler.Start` always?

MatusKysel: added

PR #2410, opened 2025-07-22
Fix exporter-v2: publish decided quorums to websocket

julienh-ssv:
## What this PR does:

Exporter v2 did not send any messages to the websocket. This PR adds a `DecidedListener` to exporter v2 to listen to new decided quorums. When the node starts, the websocket publisher is passed as DecidedListener to the exporter.

## Implementation notes

After several attempts, the most robust was to store the quorums on the temporary in-memory trace directly. Quorum are not persisted on disk.

## How to run on local:

  **1. Edit you config to make sure to enable exporter v2 and websockets:**

```
global:
  LogLevel: debug # enable debug to see the websocket trace
  
exporter:
  Enabled: true
  Mode: archive
  RetainSlots: 5

SSVAPIPort: 31602
WebSocketAPIPort: 31603
WithPing: true
```

**2. Compile and run:**

```
make build
go run $(pwd)/cmd/ssvnode start-node --config="$CONFIG_PATH"  | grep DecidedListener
```

**3. Start a websocket connection**

```
wscat -c ws://localhost:31603/stream
```

### DIFF
--- a/exporter/api/decided/stream.go
+++ b/exporter/api/decided/stream.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ssvlabs/ssv/exporter/api"
 	"github.com/ssvlabs/ssv/logging/fields"
+	dutytracer "github.com/ssvlabs/ssv/operator/dutytracer"
 	"github.com/ssvlabs/ssv/protocol/v2/qbft/controller"
 	qbftstorage "github.com/ssvlabs/ssv/protocol/v2/qbft/storage"
 )
@@ -30,5 +31,40 @@ func NewStreamPublisher(logger *zap.Logger, domainType spectypes.DomainType, ws 
 
 		logger.Debug("broadcast decided stream", fields.PubKey(msg.PubKey[:]), fields.Slot(msg.Slot))
 		feed.Send(api.NewParticipantsAPIMsg(domainType, msg))
+	}
+}
+
+// NewDecidedListener handles incoming newly decided messages.
+// it forward messages to websocket stream, where messages are cached (1m TTL) to avoid flooding
+func NewDecidedListener(logger *zap.Logger, domainType spectypes.DomainType, ws api.WebSocketServer) func(dutytracer.DecidedInfo) {
+	feed := ws.BroadcastFeed()
+	logger = logger.Named("DecidedListener")
+	cache := cache.New(time.Minute, 90*time.Second) // 1m TTL, 1.5m eviction to avoid flooding ws stream
+
+	return func(msg dutytracer.DecidedInfo) {
+		participation := qbftstorage.Participation{
+			ParticipantsRangeEntry: qbftstorage.ParticipantsRangeEntry{
+				PubKey:  msg.PubKey,
+				Slot:    msg.Slot,
+				Signers: msg.Signers,
+			},
+			Role:   msg.Role,
+			PubKey: msg.PubKey,
+		}
+
+		key := fmt.Sprintf("%x:%d:%d", msg.PubKey[:], msg.Slot, len(msg.Signers))
+		_, ok := cache.Get(key)
+		if ok {
+			// already sent in the last minute, skipping to avoid flooding ws stream
+			return
+		}
+		logger.Info("sending to websocket feed",
+			zap.String("validator_pk", fmt.Sprintf("%x", msg.PubKey[:])),
+			zap.String("signers", fmt.Sprintf("%v", msg.Signers)),
+			zap.Uint64("slot", uint64(msg.Slot)),
+			zap.String("role", msg.Role.String()),
+		)
+		cache.SetDefault(key, true)
+		feed.Send(api.NewParticipantsAPIMsg(domainType, participation))
 	}
 }
--- a/operator/dutytracer/collector.go
+++ b/operator/dutytracer/collector.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -27,6 +28,13 @@ import (
 	registrystorage "github.com/ssvlabs/ssv/registry/storage"
 	"github.com/ssvlabs/ssv/utils/hashmap"
 )
+
+type DecidedInfo struct {
+	PubKey  spectypes.ValidatorPK
+	Slot    phase0.Slot
+	Role    spectypes.BeaconRole
+	Signers []spectypes.OperatorID
+}
 
 type Collector struct {
 	logger *zap.Logger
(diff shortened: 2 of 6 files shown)


exporter/api/decided/stream.go
```diff
@@ -32,3 +34,45 @@ func NewStreamPublisher(logger *zap.Logger, domainType spectypes.DomainType, ws
...
+	domainType spectypes.DomainType
+	cache      *cache.Cache
+}
+
+func NewDecidedListener(logger *zap.Logger, domainType spectypes.DomainType, ws api.WebSocketServer) *DecidedListener {
+	return &DecidedListener{
+		feed:       ws.BroadcastFeed(),
+		logger:     logger,
+		domainType: domainType,
+		cache:      cache.New(time.Minute, time.Minute*3/2),
```
iurii-ssv: Would be nice to clarify the reasoning in a comment on the chosen cache expiration time

operator/dutytracer/collector.go
```diff
@@ -28,6 +29,17 @@ import (
...
+type DecidedInfo struct {
+	PubKey  spectypes.ValidatorPK
+	Slot    phase0.Slot
+	Role    spectypes.BeaconRole
+	Signers []spectypes.OperatorID
+}
+
+type DecidedListener interface {
+	OnDecided(msg DecidedInfo)
+}
```
iurii-ssv: It's better to keep interfaces package-private when possible (so that package API surface stays minimal/simple), unless somebody really needs it (which doesn't seem to be the case here)

operator/dutytracer/collector.go
```diff
@@ -800,3 +821,156 @@ func getOrCreateRound(trace *model.ConsensusTrace, rnd uint64) *model.RoundTrace
...
+	for _, partialMsg := range msg.Messages {
+		validator, exists := c.validators.ValidatorByIndex(partialMsg.ValidatorIndex)
+		if !exists {
+			c.logger.Debug("DecidedListener: validator not found by index",
+				zap.Uint64("validator_index", uint64(partialMsg.ValidatorIndex)))
+			continue
+		}
+
+		var validatorPK spectypes.ValidatorPK
+		copy(validatorPK[:], validator.ValidatorPubKey[:])
```
iurii-ssv, line 852: Nit: could just use `validator.ValidatorPubKey[:]` everywhere instead ? I don't see why we need to copy here explicitly, unless you could argue it provides for better readability (I wouldn't agree with that though)

operator/dutytracer/collector.go
```diff
@@ -800,3 +821,156 @@ func getOrCreateRound(trace *model.ConsensusTrace, rnd uint64) *model.RoundTrace
...
+	if !found || len(committee.Operators) == 0 {
+		return
+	}
+
+	threshold := uint64(len(committee.Operators))*2/3 + 1
+
+	// Track published quorums to avoid duplicates (validator -> role -> signers hash)
+	if trace.publishedQuorums == nil {
+		trace.publishedQuorums = make(map[phase0.ValidatorIndex]map[spectypes.BeaconRole]string)
+	}
```
iurii-ssv: It's better to put this comment on top of `trace.publishedQuorums` definition, it's probably out of place here ?

operator/dutytracer/collector.go
```diff
@@ -628,6 +644,9 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 			return nil
 		}
 	case spectypes.SSVPartialSignatureMsgType:
+		c.logger.Debug("DecidedListener: processing partial signature message",
+			zap.String("msg_id", msg.MsgID.String()))
+
```
iurii-ssv: I wonder if it would be better to define a `logger` with that additional context on it (the `"DecidedListener:` prefix - perhaps using `logger.Named`, and have `zap.String("msg_id", msg.MsgID.String())` on every relevant/related log-line) and pass it in every **func** that needs it as an argument instead of using `c.logger` everywhere ?

iurii-ssv (dismissed):
Looks good, just left some minor suggestions

exporter/api/decided/stream.go
julienharbulot: ok

operator/dutytracer/collector.go
julienharbulot: Actually the interface is used outside the package and since it must be implemented by caller it's good practice to allow caller to declare implementation at compile time using e.g. `var _ dutytracer.DecidedListener = (*MyImlementation)(nil)`

operator/dutytracer/collector.go
julienharbulot: yes you are right

operator/dutytracer/collector.go
julienharbulot: Yes that would be nice but it's not possible because the types won't compile, we need to convert it to `spectypes.ValidatorPK` and there are no direct type cast possible. -- On the upside, it's frequently done this way in exporter v1 so we're consistent with previous code.

operator/dutytracer/collector.go
julienharbulot: Yes that's a nice idea, quite idiomatic too.

operator/dutytracer/collector.go
iurii-ssv: Just to elaborate on how I approach interfaces (not saying this is how we should do it, but we probably want to get on the same page to keep it consistent across the code-base) cc @nkryuchkov  @MatusKysel @olegshmuelov  @y0sher @oleg-ssvlabs
- define & use `interface` only if you must (eg. for unit-tests) - don't define it just for the sake of creating it, use `struct` instead
- define public/exported `interface` only if you have multiple intended users (in multiple different packages), typically it would be a small-ish & flexible interface like `io.Writer` and not something super-concrete like `DecidedListener`; If there is only 1 user, or different users need different interface parts - it's probably better to define a package-private interface(s) on the user-side

This ^ helps to keep the amount of `interface` definitions manageable, and package-isolated (we wouldn't want to have a public `interface` for every `struct` that is also "visible/importable" from any other package, right ?)

julienharbulot: I think we’re on the same page for the most part :)

In this case, the interface is implemented by both the unit tests and the api package, and since the caller needs to implement it, making it public felt natural—especially as it enables compile-time checks at the implementation site.

That said, I’m totally happy to align with whatever convention the team lands on. If we decide to keep it private, I’m happy to adjust accordingly.

julienharbulot: I switched it to use a function callback instead—felt simpler and avoids the interface discussion for now. Should be easier to align on 👍

iurii-ssv (dismissed):
👍

operator/dutytracer/collector.go
iurii-ssv: Yeah, **func** here is a natural substitute as well.

iurii-ssv (approved):
It's better to use `git merge stage` (instead of `git rebase stage`) when possible to preserve commit history (so the reviewers don't need to go over all the changes again to ensure there is nothing unexpected in there, they will be able to simply check the latest merge-commit)

PR #2412, opened 2025-07-23
LogAnalyzer: Benchmark missed logs (SubscribeFilterQuery/Finalized)

julienh-ssv:
Simplified description:

The way we request logs from the node:

```go
query := ethereum.FilterQuery{
    FromBlock: big.NewInt(start),
    ToBlock:   big.NewInt(end),
    Addresses: []common.Address{targetContract},
}

logs, err := client.FilterLogs(ctx, query)
```

Is not guaranteed to return logs up to `ToBlock`. For some random reason, it could return no logs for block ToBlock even if that block has logs.

-----------------

Short of a better fix for now, this PR adds some benchmark to attempt to detect when we miss logs.

### DIFF
--- a/api/handlers/model.go
+++ b/api/handlers/model.go
@@ -106,6 +106,15 @@ func toProposalTrace(rt *model.ProposalTrace) *proposalTrace {
 	if rt == nil {
 		return nil
 	}
+
+	// Filter out zero-valued ProposalTrace objects created by SSZ encoding.
+	// These occur when rounds exist with only roundChanges (or possibly also prepares/commits) but no actual proposal.
+	// Without filtering, they appear as confusing proposals with round=0, leader=0, epoch timestamp.
+	// While 0-valued proposals are technically valid according to our data model, they show up as confusing API responses.
+	if rt.Round == 0 && rt.Signer == 0 && rt.ReceivedTime == 0 {
+		return nil
+	}
+
 	return &proposalTrace{
 		Round:                 rt.Round,
 		BeaconRoot:            rt.BeaconRoot,
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -238,6 +238,8 @@ func (gc *GoClient) weightedAttestationData(ctx context.Context, slot phase0.Slo
 		zap.Float64("score", bestScore)).
 		Debug("successfully fetched attestation data")
 
+	recordAttestationDataClientSelection(ctx, bestClientAddr)
+
 	return bestAttestationData, nil
 }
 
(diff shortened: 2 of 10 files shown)


eth/eventhandler/event_handler.go
```diff
@@ -146,6 +146,11 @@ func (eh *EventHandler) HandleBlockEventsStream(ctx context.Context, logs <-chan
 }
 
 func (eh *EventHandler) processBlockEvents(ctx context.Context, block executionclient.BlockLogs) ([]Task, error) {
+	if len(block.Logs) == 0 {
+		// Unexpected: blocks should generally contain logs.
+		// Emit a warning as this may signal a programming bug or missed event.
+		eh.logger.Warn("Processing block events: logs for this block are empty", fields.BlockNumber(block.BlockNumber))
```
y0sher: is this comment correct? ssv contract not necessarily have event logs in each and every block.

julienharbulot: Yes I just checked on etherscan and there's a lot less traffic than expected, I'll edit the comment and change the level to debug

eth/eventhandler/event_handler.go
```diff
@@ -146,6 +146,9 @@ func (eh *EventHandler) HandleBlockEventsStream(ctx context.Context, logs <-chan
 }
 
 func (eh *EventHandler) processBlockEvents(ctx context.Context, block executionclient.BlockLogs) ([]Task, error) {
+	if len(block.Logs) == 0 {
+		eh.logger.Debug("Processing block events: logs for this block are empty", fields.BlockNumber(block.BlockNumber))
```
kchojn: lowercase pls

kchojn (changes requested):

eth/eventhandler/event_handler.go
julienharbulot: ok

kchojn (dismissed):

eth/executionclient/execution_client.go
```diff
@@ -439,6 +441,71 @@ func (ec *ExecutionClient) isClosed() bool {
...
+
+	if inner, ok := l.missingFromBatch[blocklogs.BlockNumber]; ok {
+		for _, log := range blocklogs.Logs {
+			delete(inner, log.Index)
+		}
+		if len(inner) == 0 {
+			delete(l.missingFromBatch, blocklogs.BlockNumber)
+		}
+	}
+}
```
y0sher: we actually never saw a case where we missed just a single log. The only real case we observed is when for a certain block exec-client api returned as if we have 0 logs instead of X.

PR #2414, opened 2025-07-25
Add metric for attestation data client selection tracking

zktaiga:
Until now, we knew latency to each client, but without debug logs, we did not have a way to know which client was picked by WAD and with which frequency. This native metric will allow us to track the over-time performance of clients as scored by WAD.

image


kchojn (approved):

y0sher:
> LGTM, but maybe we could also use this metric for `simpleAttestationData` as well ?
> 
> Because we have multi-client it is possible (expected even) for different clients to be used every time (even if currently we sort of try to use the 1st URL in the list of configured CL clients I have a feeling this won't be implemented like that for long, eg. some failover improvements [I've implemented here](https://github.com/attestantio/go-eth2-client/pull/249) will violate this assumption, and it's not like these assumptions hold 100% anyway).

seem to be in simpleAttestationData as well.

y0sher:
@zktaiga I committed and forced push with signature so it will be merge-able.

PR #2418, opened 2025-08-01
migrations: fix order

nkryuchkov:
The `main` branch already contains the HKDF migrations as 7th, so it must go 7th. The current state of `stage` has it as 8th, which may cause incompatibilities with the `main` branch.

### DIFF
--- a/migrations/migration_7_derive_signer_key_with_hkdf.go
+++ b/migrations/migration_7_derive_signer_key_with_hkdf.go
@@ -9,8 +9,8 @@ import (
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
-var migration_8_derive_signer_key_with_hkdf = Migration{
-	Name: "migration_8_derive_signer_key_with_hkdf",
+var migration_7_derive_signer_key_with_hkdf = Migration{
+	Name: "migration_7_derive_signer_key_with_hkdf",
 	Run: func(ctx context.Context, logger *zap.Logger, opt Options, key []byte, completed CompletedFunc) (err error) {
 		return opt.Db.Update(func(txn basedb.Txn) error {
 			if opt.OperatorPrivKey == nil {
--- a/migrations/migration_7_derive_signer_key_with_hkdf_test.go
+++ b/migrations/migration_7_derive_signer_key_with_hkdf_test.go
@@ -22,7 +22,7 @@ import (
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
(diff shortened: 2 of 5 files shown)


iurii-ssv (approved):
Yeah we probably want to fix that 👍 

To make sure I understand what exactly happens with this change, and how the problem manifests:
- we have migration_7_A (= migration_7_derive_signer_key_with_hkdf.go) in **main** merged already
- we have migration_7_B (migration_7_populate_validator_index_mapping.go) in **stage** merged already + also migration_8 (= migration_8_derive_signer_key_with_hkdf.go) on top
- when we deploy **stage** branch onto DB previously managed by **main** branch version we cannot apply migration_8 because it is not idempotent and hence cannot be applied on top of migration_7_A (we try to decrypt DB entries with encryption key in the old format while they are already encrypted with encryption key in the new format)
- because migrations 7_A (or 8) and 7_B are independent from each other it's fine to swap the order we apply these in in **stage** branch (the number in migration name doesn't mean much since [migration application-order is defined here](https://github.com/ssvlabs/ssv/blob/4f3ae69cd23e5259d01c365b3dd4269f0c3063fa/migrations/migrations.go#L24-L34))

I'm not sure however how this will work for existing **stage**-based DBs (I don't think it will unless we delete the DB) - from their perspective we essentially define 2 new migrations:
- `migration_7_derive_signer_key_with_hkdf` which cannot be applied on top of `migration_8_derive_signer_key_with_hkdf` (similar to how `migration_8_derive_signer_key_with_hkdf` cannot be applied on top of `migration_7_derive_signer_key_with_hkdf` - which is what this PR is trying to solve)
- `migration_8_populate_validator_index_mapping` will apply fine on top of `migration_7_populate_validator_index_mapping` because it's idempotent (it will just re-build the mapping, I think)

nkryuchkov:
> > @iurii-ssv I'm not sure we need a workaround for stage, because it's not intended for use by anyone but our stage nodes. I think creating a new DB would be better than maintaining a workaround in the code
> 
> Yeah, I agree, as long as it's easy to fix by deleting DB (I'm just worried that it's too much of manual work, and if somebody deploys a branch that's not rebased on top of this PR for example he'll get failing migration as well unless he rebases)

We can automate this by creating a special branch with a flag to delete the current database under certain conditions. After merging this to stage, if someone doesn't merge it into their PR, they'll just get a failure and will know that they need to pull changes from stage, it shouldn't cause any conflicts. We'll just need to make sure everyone is aware

kchojn (approved):

PR #2419, opened 2025-08-01
duties: log more details

iurii-ssv:
This PR is meant to improve logs so that they contain more complete data, as well as make logging more consistent across different duty types, for example:
- we want more granular logging for every duty we have (to be able to split the duty up into a bunch of steps to isolate the exact  step where some problem happens, and also to get the exact time it took to execute a particular step)
- if we want to find out why a particular validator couldn't successfully attest in a particular slot - we could filter by **slot** + **validator_idx**, but we won't have any logs regarding whether or not we've submitted an attestation for that particular validator because we only log the total attestation-count submitted (which is not enough info to debug with)
- this PR also adds/removes some log-fields to/from some log lines as well as rearranges some code to simplify logging, essentially clearing the ground for https://github.com/ssvlabs/ssv/issues/1828

I've also re-arranged some spec-test files in this PR a bit to simplify the dependancies between them (not super relevant to logging, but I thought I might need that for some functionality in this PR to work ... ended up removing that functionality but keeping the re-arrangement of the file structure ... could move it to a separate PR but since it only affects tests there probably is no need).

### DIFF
--- a/operator/dutytracer/collector.go
+++ b/operator/dutytracer/collector.go
@@ -434,26 +434,6 @@ func (c *Collector) processPartialSigCommittee(receivedAt uint64, msg *spectypes
 	}
 }
 
-func (c *Collector) saveValidatorToCommitteeLink(slot phase0.Slot, msg *spectypes.PartialSignatureMessages, committeeID spectypes.CommitteeID) {
-	for _, msg := range msg.Messages {
-		slotToCommittee, found := c.validatorIndexToCommitteeLinks.Get(msg.ValidatorIndex)
-		if !found {
-			slotToCommittee, _ = c.validatorIndexToCommitteeLinks.GetOrSet(msg.ValidatorIndex, hashmap.New[phase0.Slot, spectypes.CommitteeID]())
-		}
-
-		slotToCommittee.Set(slot, committeeID)
-	}
-}
-
-func (c *Collector) saveLateValidatorToCommiteeLinks(slot phase0.Slot, msg *spectypes.PartialSignatureMessages, committeeID spectypes.CommitteeID) {
-	for _, msg := range msg.Messages {
-		if err := c.store.SaveCommitteeDutyLink(slot, msg.ValidatorIndex, committeeID); err != nil {
-			// for late messages we can log with warn
-			c.logger.Warn("save late validator to committee links", zap.Error(err), fields.Slot(slot), fields.ValidatorIndex(msg.ValidatorIndex), fields.CommitteeID(committeeID))
-		}
-	}
-}
-
 func (c *Collector) getSyncCommitteeRoot(ctx context.Context, slot phase0.Slot, in []byte) (phase0.Root, error) {
 	var beaconVote = new(spectypes.BeaconVote)
 	if err := beaconVote.Decode(in); err != nil {
@@ -745,12 +725,9 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 
 			if late {
 				_ = c.inFlightCommittee.Delete(committeeID)
-				c.saveLateValidatorToCommiteeLinks(slot, pSigMessages, committeeID)
 				return c.store.SaveCommitteeDuty(&trace.CommitteeDutyTrace)
 			}
 
-			// cache the link between validator index and committee id
-			c.saveValidatorToCommitteeLink(slot, pSigMessages, committeeID)
 			return nil
 		}
 
@@ -1105,7 +1082,6 @@ func (c *Collector) computeAndPersistScheduleForSlot(slot phase0.Slot) error {
 	// Attester indices for this slot (InCommittee only)
 	if c.duties.Attester != nil {
 		for _, d := range c.duties.Attester.CommitteeSlotDuties(epoch, slot) {
-			// d may be nil in edge cases; guard defensively
 			if d == nil {
 				continue
 			}
@@ -1113,7 +1089,7 @@ func (c *Collector) computeAndPersistScheduleForSlot(slot phase0.Slot) error {
 		}
 	}
 
-	// Proposer indices for this slot (all scheduled proposers)
+	// Proposer indices for this slot
 	if c.duties.Proposer != nil {
 		for _, idx := range c.duties.Proposer.SlotIndices(epoch, slot) {
 			schedule[idx] |= rolemask.BitProposer
@@ -1134,7 +1110,34 @@ func (c *Collector) computeAndPersistScheduleForSlot(slot phase0.Slot) error {
 	if len(schedule) == 0 {
 		return nil
 	}
-	return c.store.SaveScheduled(slot, schedule)
+
+	if err := c.store.SaveScheduled(slot, schedule); err != nil {
+		return fmt.Errorf("save scheduled: %w", err)
+	}
+
+	// Populate committee links only for validators with scheduled duties
+	committees := c.validators.ParticipatingCommittees(epoch)
+	committeeByValidator := buildValidatorToCommitteeIndex(committees)
+
+	for validatorIndex := range schedule {
+		if committeeID, found := committeeByValidator[validatorIndex]; found {
+			slotToCommittee, _ := c.validatorIndexToCommitteeLinks.GetOrSet(validatorIndex, hashmap.New[phase0.Slot, spectypes.CommitteeID]())
+			slotToCommittee.Set(slot, committeeID)
+		}
+	}
+
+	return nil
+}
+
+// buildValidatorToCommitteeIndex creates a reverse lookup map from validator index to committee ID.
+func buildValidatorToCommitteeIndex(committees []*registrystorage.Committee) map[phase0.ValidatorIndex]spectypes.CommitteeID {
+	result := make(map[phase0.ValidatorIndex]spectypes.CommitteeID)
+	for _, cmt := range committees {
+		for _, validatorIndex := range cmt.Indices {
+			result[validatorIndex] = cmt.ID
+		}
+	}
+	return result
 }
 
 // runScheduleWorker performs schedule computations and DB writes off the hot path.
--- a/operator/dutytracer/collector_schedule_test.go
+++ b/operator/dutytracer/collector_schedule_test.go
@@ -4,7 +4,9 @@ import (
 	"testing"
 
 	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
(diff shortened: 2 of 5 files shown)


y0sher (approved):
lgtm

operator/duties/scheduler.go
```diff
@@ -419,9 +424,6 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, duties []*spectypes.Valid
 		slotDelayHistogram.Record(ctx, slotDelay.Seconds())
 
 		go func(ctx context.Context) {
-			if duty.Type == spectypes.BNRoleAttester || duty.Type == spectypes.BNRoleSyncCommittee {
-				s.waitOneThirdOrValidBlock(duty.Slot)
-			}
 			recordDutyExecuted(ctx, duty.RunnerRole())
 			s.dutyExecutor.ExecuteDuty(ctx, duty)
 		}(ctx)
```
iurii-ssv: Removing this `if` here because `ExecuteDuties` manages validator-based duties while `BNRoleAttester` and `BNRoleSyncCommittee` are committee-based duties (managed by `ExecuteCommitteeDuties`), so this is a dead code effectively at this point

protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -143,23 +144,28 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
 	}
 
 	if len(selectionProofs) == 0 {
-		span.AddEvent("no selection proofs")
-		span.SetStatus(codes.Ok, "")
 		r.GetState().Finished = true
 		r.measurements.EndDutyFlow()
+		const dutyFinishedNoProofsEvent = "successfully finished duty processin (no selection proofs)"
```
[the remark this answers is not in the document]
iurii-ssv: fixed 👍

protocol/v2/ssv/runner/runner.go
```diff
@@ -165,21 +165,30 @@ func (b *BaseRunner) baseStartNewNonBeaconDuty(ctx context.Context, logger *zap.
...
+	defer span.End()
+
 	if err := b.ValidatePreConsensusMsg(ctx, runner, signedMsg); err != nil {
 		return false, nil, errors.Wrap(err, "invalid pre-consensus message")
 	}
 
 	hasQuorum, roots := b.basePartialSigMsgProcessing(signedMsg, b.State.PreConsensusContainer)
+
+	if hasQuorum {
+		const gotPostConsensusQuorumEvent = "got pre consensus quorum"
```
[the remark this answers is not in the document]
iurii-ssv: fixed

protocol/v2/qbft/instance/instance.go
```diff
@@ -74,57 +72,52 @@ func (i *Instance) ForceStop() {
...
-		_, span := tracer.Start(ctx,
-			observability.InstrumentName(observabilityNamespace, "qbft.instance.start"),
-			trace.WithAttributes(observability.BeaconSlotAttribute(phase0.Slot(height))))
-		defer span.End()
-
-		i.StartValue = value
-		i.bumpToRound(specqbft.FirstRound)
-		i.State.Height = height
-		i.metrics.Start()
-		i.config.GetTimer().TimeoutForRound(height, specqbft.FirstRound)
```
iurii-ssv, line 96: We don't need a `startOnce` thingy because we only start the `Instance` in `Controller.StartNewInstance` which checks whether or not the instance already exists before creating & starting it (which is how it actually should be done, using `sync.Once` is a hacky way to go about it)

y0sher (dismissed):
@iurii-ssv this look good but also breaks spec alignment badly, so I'm wondering if we should do similar changes in spec or just diverge from their code entirely.

operator/duties/scheduler.go
```diff
@@ -541,19 +546,33 @@ func (s *Scheduler) loggerWithCommitteeDutyContext(committeeDuty *committeeDuty)
 // advanceHeadSlot will set s.headSlot to the provided slot (but only if the provided slot is higher,
 // meaning s.headSlot value can never decrease) and notify the go-routines waiting for it to happen.
 func (s *Scheduler) advanceHeadSlot(slot phase0.Slot) {
+	s.logger.Debug("advancing head slot (maybe)")
```
nkryuchkov, line 549: is it a debug leftover?

operator/duties/scheduler.go
```diff
@@ -541,19 +546,33 @@ func (s *Scheduler) loggerWithCommitteeDutyContext(committeeDuty *committeeDuty)
...
 	s.waitCond.L.Unlock()
 }
 
-// waitOneThirdOrValidBlock waits until one-third of the slot has passed (SECONDS_PER_SLOT / 3 seconds after
-// slot start time), or for head block event that might come in even sooner than one-third of the slot passes.
-func (s *Scheduler) waitOneThirdOrValidBlock(slot phase0.Slot) {
+// waitOneThirdIntoSlotOrValidBlock waits until one-third of the slot has passed (SECONDS_PER_SLOT / 3 seconds after
+// slot start time), or for a head block event that might come in even sooner than one-third of the slot passes.
+func (s *Scheduler) waitOneThirdIntoSlotOrValidBlock(slot phase0.Slot) {
+	s.logger.Debug("waiting 1/3 into slot (maybe)")
```
nkryuchkov, line 567: ditto

protocol/v2/qbft/instance/proposal.go
```diff
@@ -51,10 +51,7 @@ func (i *Instance) uponProposal(ctx context.Context, logger *zap.Logger, msg *sp
 		return errors.Wrap(err, "could not create prepare msg")
 	}
 
-	logger.Debug("📢 got proposal, broadcasting prepare message",
-		fields.QBFTRound(i.State.Round),
-		zap.Any("proposal_signers", msg.SignedMessage.OperatorIDs),
-		zap.Any("prepare_signers", prepare.OperatorIDs))
+	logger.Debug("📢 got proposal, broadcasting prepare message", zap.Any("prepare_signers", prepare.OperatorIDs))
```
nkryuchkov, line 54: We might be missing the QBFT round here, or am I wrong?

protocol/v2/ssv/runner/measurements.go
```diff
@@ -16,102 +16,58 @@ type measurementsStore struct {
...
 }
 
-func NewMeasurementsStore() measurementsStore {
-	return measurementsStore{}
+func newMeasurementsStore() *dutyMeasurements {
+	return &dutyMeasurements{}
 }
 
-func (cm *measurementsStore) PreConsensusTime() time.Duration {
+func (cm *dutyMeasurements) PreConsensusTime() time.Duration {
```
nkryuchkov:
```suggestion
func (dm *dutyMeasurements) PreConsensusTime() time.Duration {
```

protocol/v2/ssv/runner/proposer.go
```diff
@@ -349,29 +347,33 @@ func (r *ProposerRunner) ProcessConsensus(ctx context.Context, logger *zap.Logge
...
 	}
 
+	r.measurements.StartPostConsensus()
 	span.AddEvent("broadcasting post consensus partial signature message")
 	if err := r.GetNetwork().Broadcast(msgID, msgToBroadcast); err != nil {
 		return fmt.Errorf("can't broadcast partial post consensus sig: %w", err)
 	}
+	const broadcastedPostConsensusMsgEvent = "broadcasted post-consensus partial signature message"
+	logger.Debug(broadcastedPostConsensusMsgEvent)
+	span.AddEvent(broadcastedPostConsensusMsgEvent)
```
nkryuchkov, line 356: Do you think it might be worth creating some kind of wrapper for this pattern? Of course not in this PR, but it might be a good place for this discussion

protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -590,7 +609,7 @@ func (r *SyncCommitteeAggregatorRunner) GetShare() *spectypes.Share {
 	return nil
 }
 
-func (r *SyncCommitteeAggregatorRunner) GetState() *State {
+func (r *SyncCommitteeAggregatorRunner) bState() *State {
```
nkryuchkov: what does `b` in `bState` stand for?

operator/duties/scheduler.go
iurii-ssv: Not really, it's added so we have the exact times for how long the `s.waitCond` is held logged (in case we need to debug delays)

protocol/v2/qbft/instance/proposal.go
iurii-ssv: Now some of the qbft-related fields (such as round/height) are added to the `logger` object in `logWithMessageFields` func that executes as soon as we receive & start processing the message - so that those are present for every log-line that's relevant for that message (instead of logging it selectively and forgetting to add, at times),

I've re-checked a bunch of log-lines and pushed [another relevant commit here](https://github.com/ssvlabs/ssv/pull/2419/commits/b155587d14c59bb811d673a0a5042d0f515c7dc4)

protocol/v2/ssv/runner/proposer.go
iurii-ssv: Potentially yes, but maybe later :)

protocol/v2/ssv/runner/measurements.go
iurii-ssv: Applied in https://github.com/ssvlabs/ssv/pull/2419/commits/83fdc331065af0fbc30fb8205be3af4ac3da946a

protocol/v2/ssv/runner/sync_committee_contribution.go
iurii-ssv: b = base, the reasons to change it are:
- it's shorter (and we call it a lot)
- it's better to keep it package-private to ensure we don't access it outside of the package (potentially without proper concurrent synchronization)

I guess we can just call it `state()` - renamed in https://github.com/ssvlabs/ssv/pull/2419/commits/12e6860112d18205c3348c90789d0389eace7458

operator/duties/scheduler.go
nkryuchkov: @iurii-ssv can we do it in one or two logs?

iurii-ssv: I think we need all 3 lines:
- start of the func
- (if it happened) the slot-advancement itself with all the relevant details
- end of the func (since the previous happens only "maybe" we can't omit it)

I don't see how to simplify it further without giving up the details it currently logs, you have something particular in mind ?

nkryuchkov (approved):

PR #2423, opened 2025-08-04
feat(exporter): add get all validator traces endpoint

julienh-ssv:


operator/dutytracer/store.go
```diff
@@ -56,6 +56,61 @@ func (c *Collector) GetCommitteeID(slot phase0.Slot, pubkey spectypes.ValidatorP
...
 
+func (c *Collector) GetAllValidatorDuties(role spectypes.BeaconRole, slot phase0.Slot) ([]*ValidatorDutyTrace, error) {
+	duties := []*ValidatorDutyTrace{}
+
+	// lookup in cache
+	c.validatorTraces.Range(func(pubkey spectypes.ValidatorPK, validatorSlots *hashmap.Map[phase0.Slot, *validatorDutyTrace]) bool {
+		traces, found := validatorSlots.Get(slot)
+		if found {
+			traces.Lock()
+			defer traces.Unlock()
```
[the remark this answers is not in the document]
iurii-ssv: Actually this seems to be a false-positive alarm since the loop is not a typical loop here - it executes func to completion on every iteration. So there isn't any issues with defer-unlocking here.

api/handlers/exporter.go
```diff
@@ -41,6 +41,7 @@ func NewExporter(logger *zap.Logger, participantStores *ibftstorage.ParticipantS
 
 type dutyTraceStore interface {
 	GetValidatorDuty(role spectypes.BeaconRole, slot phase0.Slot, pubkey spectypes.ValidatorPK) (*dutytracer.ValidatorDutyTrace, error)
+	GetAllValidatorDuties(role spectypes.BeaconRole, slot phase0.Slot) ([]*dutytracer.ValidatorDutyTrace, error)
```
iurii-ssv: Nit: I would rename `GetAllValidatorDuties` -> `GetValidatorDuties`:
- this is more consistent with `GetValidatorDuty` naming (and maybe the naming used in storage, I think)
- `All` part might imply we want to get everything - when in fact we still filter by role/slot

api/handlers/exporter.go
```diff
@@ -349,8 +350,19 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
...
+		return role == spectypes.BNRoleSyncCommittee || role == spectypes.BNRoleAttester
+	}
+
+	// either PubKeys or Indices are required for committee duty roles
 	if len(request.PubKeys) == 0 && len(request.Indices) == 0 {
-		return api.BadRequestError(errors.New("either pubkeys or indices is required"))
+		for _, r := range request.Roles {
+			role := spectypes.BeaconRole(r)
+			if isCommitteeDuty(role) {
+				return api.BadRequestError(errors.New("either pubkeys or indices is required for role " + role.String()))
```
iurii-ssv: Nit: I think it's better to use `fmt.Errorf` for error-formatting

api/handlers/exporter.go
```diff
@@ -383,8 +395,19 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
 		slot := phase0.Slot(s)
 		for _, r := range request.Roles {
 			role := spectypes.BeaconRole(r)
+
+			if len(pubkeys) == 0 {
+				duties, err := e.traceStore.GetAllValidatorDuties(role, slot)
+				if err != nil {
+					e.logger.Debug("error getting validator duties", zap.Error(err), fields.Slot(slot), fields.BeaconRole(role))
+					continue
+				}
```
iurii-ssv: Shouldn't we log under `Error` here (not `Debug`) ?

operator/dutytracer/store.go
```diff
@@ -56,6 +56,61 @@ func (c *Collector) GetCommitteeID(slot phase0.Slot, pubkey spectypes.ValidatorP
...
+	})
+
+	// go to disk for the older ones
+	storeDuties, err := c.getAllValidatorDutiesFromDisk(role, slot)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(duties, storeDuties...), nil
+}
```
iurii-ssv: Just to make sure I understand this correctly, 

do we store non-overlapping traces on disk and in cache ? Otherwise we'll likely end up with duplicates in the resulting slice ...

api/handlers/exporter.go
iurii-ssv: Do you know why "either PubKeys or Indices are required for committee duty roles" ? Would be nice to document the "why" as well.

iurii-ssv (dismissed):
Looks good! Left some minor suggestions

api/handlers/exporter.go
julienharbulot: ok

julienharbulot: ok

api/handlers/exporter.go
julienharbulot: Yes we can improve this, I'll use a better error check consistent with what's done when getting the CommitteeDuty a few line below this line

api/handlers/exporter.go
julienharbulot: ok

operator/dutytracer/store.go
julienharbulot: Yes they are non-overlapping

iurii-ssv (commented):
@julienh-ssv looks like unit-tests are failing now,

btw, you can close(resolve) the discussions that "you've addressed IF they don't have any further comments for me/reviewers to see" on your own - so it's easier for me to figure out where this PR is at when I get back to it for another review-round (whether you've addressed everything in some way, or simply in the middle of it, or maybe skipped/forgot some)

iurii-ssv (approved):
LGTM 👍

api/handlers/exporter.go
```diff
@@ -349,8 +350,19 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
 		return api.BadRequestError(fmt.Errorf("at least one role is required"))
 	}
 
+	// these roles map to a committee duty
+	isCommitteeDuty := func(role spectypes.BeaconRole) bool {
```
y0sher, line 354: I don't see a reason why this should be a closure func.

PR #2428, opened 2025-08-07
feat(metrics) - introduction of a new metric (number of quorums), refactor, bug fix  

oleg-ssvlabs:
### Background

This PR is related to increasing the maximum number of validators to **3,000**.  
The concern was that **aggregator** and **sync committee contribution** duties may hit a bottleneck (related to P2P queues).  
To verify whether these types of duties still run successfully, I added a new metric that tracks the number of successful **quorums** in the **pre-consensus phase** (before the `IsAggregator` / `IsSyncCommitteeAggregator` check) - recorded on epoch basis as a gauge.

---

### List of Changes

- Introduced a **new metric**: number of successful pre-consensus quorums for Aggregator and Sync Committee Contribution roles  
- Fixed a **bug** related to metric enablement  
- **Refactored** the epoch-based metric recorder  
- Implemented a **unit test**  
- Renamed file: `sync_committee_aggregator.go` → `sync_committee_contribution.go`

### DIFF
--- a/operator/dutytracer/collector_test.go
+++ b/operator/dutytracer/collector_test.go
@@ -385,7 +385,7 @@ func TestValidatorDuties(t *testing.T) {
 	validators := registrystoragemocks.NewMockValidatorStore(ctrl)
 	dutyStore := new(mockDutyTraceStore)
 
-	collector := New(logger, validators, nil, dutyStore, networkconfig.TestNetwork.BeaconConfig, nil)
+	collector := New(logger, validators, nil, dutyStore, networkconfig.TestNetwork.Beacon, nil)
 
 	var wantBeaconRoot phase0.Root
 	bnVal := [32]byte{1, 2, 3}


protocol/v2/ssv/runner/observability.go
```diff
@@ -9,25 +9,79 @@ import (
...
 )
 
+type attributeConsensusPhase string
+
 const (
 	observabilityName      = "github.com/ssvlabs/ssv/protocol/v2/ssv"
 	observabilityNamespace = "ssv.validator"
+
+	attributeConsensusPhasePreConsensus attributeConsensusPhase = "pre_consensus"
+)
```
iurii-ssv, line 26: I'm not sure why we need `attributeConsensusPhasePreConsensus` (and corresponding `attributeConsensusPhase `), it seems like we can just update the `recordSuccessfulQuorum` func to be `recordSuccessfulQuorumPreConsensus` (and later on, if necessary, add similar functions for Consensus/PostConsensus phases).

iurii-ssv (dismissed):
Good to have those metrics 👍

protocol/v2/ssv/runner/observability.go
oleg-ssvlabs: yeah, I guess it's just a different approach. I don't really have strong opinion here - more functions versus type of more constants

iurii-ssv: With "separate func" I think it's a bit more explicit and also short.

Btw, we could use similar kind of reasoning for `EpochMetricRecorder` - it currently has this map `data  map[types.BeaconRole]epochCounter` that we need to protect with mutex, perhaps we could have "hardcoded" a separate explicit counter per duty type and didn't even need a mutex at all ? Although I'm not sure how it would align with:

> // When the epoch advances, all roles from the internal data map that had recorded duties
// in the previous epoch will have their metrics flushed (recorded), not just the role passed in.
// This is necessary because not all duties are executed every epoch, so to ensure accurate
// metric reporting, all completed roles from the previous epoch must be recorded once the new epoch begins.

TBH it feels like we could have a simpler-to-understand implementation for this overall - maybe "using a scheduler that would dump whatever data we have for the past epoch for every duty" (it might be more code overall, but probably simpler to reason about and more performant). But since this is limited in scope to a very specific area I don't mind this solution either (no need to re-write it), just saying that maybe we should prefer more dumb/straightforward/concrete logic when possible.

PR #2431, opened 2025-08-08
deps: update go-eth2-client

iurii-ssv:
This PR updates `go-eth2-client` library (our own fork we use) to the latest version with the main intent to adopt the change implemented in https://github.com/ssvlabs/go-eth2-client/pull/11 

I've also went over the changes in https://github.com/ssvlabs/ssv/pull/2339 to check if https://github.com/ssvlabs/go-eth2-client/pull/11 completely removes the need for these ... unfortunately it doesn't due to the following (as is now documented in the comment for `Healthy` func):
```go
// ... multi-client implementation does not allow
// for sync-distance to be > 1 while we do want to allow for that (in case we have multiple clients and none of them
// has sync-distance of <= 1 we still want Healthy to report success so that SSV node can survive such occasional CL
// hiccups without having to restart).
```

_Before merging:_
- [x] test on **stage**

### DIFF
--- a/beacon/goclient/goclient_test.go
+++ b/beacon/goclient/goclient_test.go
@@ -3,12 +3,9 @@ package goclient
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
-	"net/http/httptest"
 	"strings"
-	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


beacon/goclient/goclient_test.go
```diff
@@ -198,30 +184,7 @@ func runHealthyTest(
...
-				return
-			}
-		}()
-	}
-
-	wg.Wait()
-
-	return errs
+	return c.Healthy(t.Context())
 }
```
iurii-ssv, line 185: @nkryuchkov I'm removing the `concurrentHealthCheck` param now because it was never used, I assume it was added during the writing of these tests just to see how `Healthy` func works in concurrent settings ? Anyway, existing tests should be covering that already and `go-eth2-client` itself should have some concurrency-related tests as well

y0sher (approved):
lgtm

PR #2432, opened 2025-08-08
chore(e2e) - decommission old E2E implementation

oleg-ssvlabs:


iurii-ssv (approved):
Nice!

y0sher (approved):

y0sher (commented):
Yes we didn't run this because it was outdated since Alan fork changes, but this e2e suite was intended to cover very specific thing like slashing protection and signing, I think maybe the code here can inspire how to check these stuff in the future to not fully discard the concept.

PR #2436, opened 2025-08-11
chore(Makefile): Remove default HostAddress

kchojn:
> [!NOTE]
> I think we should mention in the release notes that we are now enforcing this behavior and that `HostAddress` default disappears from the build

Related pr: [feat(network/discovery/enode.go): make HostDNS and HostAddress mutually exclusive](https://github.com/ssvlabs/ssv/pull/2274)

Removed the default `HOST_ADDRESS` assignment from the `Makefile`.

Previously, the `Makefile` would automatically assign a public IP to `HOST_ADDRESS` if it wasn't already set. This behavior is being removed because the application requires that either `HostAddress` or `HostDNS` is explicitly provided by the user. The automatic setting in the `Makefile` could override a user's intention to use `HostDNS` or a different IP address.

By removing this default, the configuration becomes more explicit. Users are now required to set either `HOST_ADDRESS` or `HOST_DNS` via their configuration file or environment variables, ensuring the node's network configuration is always intentional.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,6 @@ ifdef DEBUG_PORT
 	@dlv  --continue --accept-multiclient --headless --listen=:${DEBUG_PORT} --api-version=2 exec \
 	 ${BUILD_PATH} start-node -- ${NODE_COMMAND}
 else
-	@echo "Running node on address: ${HOST_ADDRESS})"
 	@${BUILD_PATH} start-node ${NODE_COMMAND}
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,6 @@ ifndef GOPATH
     export GOPATH
 endif
 
-ifndef HOST_ADDRESS
-    HOST_ADDRESS=$(shell dig @resolver4.opendns.com myip.opendns.com +short)
-    export HOST_ADDRESS
-endif
-
 ifndef BUILD_PATH
     BUILD_PATH="/go/bin/ssvnode"
     export BUILD_PATH
--- a/network/discovery/service.go
+++ b/network/discovery/service.go
@@ -49,7 +49,11 @@ type Options struct {
 }
 
 // Validate checks if the options are valid.
-func (o *Options) Validate() error {
+func (o *Options) Validate(logger *zap.Logger) error {
+	logger.Debug("validating discovery options",
+		zap.String("host_address", o.HostAddress),
+		zap.String("host_dns", o.HostDNS),
+	)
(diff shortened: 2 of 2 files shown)


oleg-ssvlabs (dismissed):

network/discovery/service.go
```diff
@@ -49,7 +49,11 @@ type Options struct {
 }
 
 // Validate checks if the options are valid.
-func (o *Options) Validate() error {
+func (o *Options) Validate(logger *zap.Logger) error {
+	logger.Debug("validating discovery options",
+		zap.String("host_address", o.HostAddress),
+		zap.String("host_dns", o.HostDNS),
+	)
```
iurii-ssv: I think it's better to log things like this at the caller site because passing logger in an argument is not ideal (it's only typically done when there is no other option to do it),

and with this particular change I don't think it makes much sense because it's somewhere in the middle of what I think we'd actually want to do:
- we should either log the entire `Options` struct (not only some parts but not others) if we want to do it for debugging and whatnot
- or we should only highlight the "problematic fields" in the `error` returned - `return fmt.Errorf("only one of HostDNS or HostAddress may be set, got HostDNS = %s,  got HostAddress = %s", o. HostDNS, o.HostAddress)`

iurii-ssv (changes requested):
Left a suggestion

network/discovery/service.go
iurii-ssv: Oh, wait, you've already applied it (looks like I've reviewed some previous commit), approved now

iurii-ssv (approved):

network/discovery/service.go
kchojn: yes, thank you for your review. I decided to use a cleaner approach, which also meets your expectations

Makefile
```diff
@@ -121,7 +116,6 @@ ifdef DEBUG_PORT
 	@dlv  --continue --accept-multiclient --headless --listen=:${DEBUG_PORT} --api-version=2 exec \
 	 ${BUILD_PATH} start-node -- ${NODE_COMMAND}
 else
-	@echo "Running node on address: ${HOST_ADDRESS})"
```
vaclav-ssvlabs, line 124: Since this log line is removed, it would be good to have a log entry from the ssv-node to give a user information about which IP is used for the p2p advertisment

MatusKysel (approved):
I am approving this but we need to discus it more it @vaclav-ssvlabs before merging it

PR #2442, opened 2025-08-12
runners: errors should have all the details

iurii-ssv:
This PR aims to add the missing details for various errors that runners/QBFT-related code emits, these are necessary to debug rare cases like the following (when it's not clear whether we have some bug in our QBFT code or if it's something expected - in which case we probably want to rework it such that it doesn't emit a "cryptic error"):
```
{
  "level": "DEBUG",
  "time": "2025-08-12T07:21:32.034520Z",
  "msg": "❗ could not handle message",
  "pubkey": "a690fd19afc6f07017c7ea12aea4071c05f9d08646fc525b2e6f5284f2bba72e4d13e1d5cbefab525a5f7f879821c2cc",
  "role": "AGGREGATOR_RUNNER",
  "msg_height": 1064157,
  "msg_round": 1,
  "consensus_msg_type": 2,
  "signers": [
    49,
    50,
    52
  ],
  "msg_type": "consensus",
  "error": "failed processing consensus message: decided wrong instance"
}
```

### DIFF
--- a/protocol/v2/ssv/runner/runner_signatures.go
+++ b/protocol/v2/ssv/runner/runner_signatures.go
@@ -26,7 +26,7 @@ func (b *BaseRunner) signBeaconObject(
 	epoch := runner.GetBaseRunner().NetworkConfig.EstimatedEpochAtSlot(slot)
 	domain, err := runner.GetBeaconNode().DomainData(ctx, epoch, signatureDomain)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not get beacon domain")
+		return nil, fmt.Errorf("could not get beacon domain: %w", err)
 	}
 	if _, ok := runner.GetBaseRunner().Share[duty.ValidatorIndex]; !ok {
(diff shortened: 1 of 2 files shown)


protocol/v2/ssv/runner/runner_signatures.go
```diff
@@ -51,45 +51,24 @@ func (b *BaseRunner) signBeaconObject(
...
-	var shareSample *spectypes.Share
-	for _, share := range b.Share {
-		shareSample = share
-		break
-	}
-	if shareSample == nil {
-		return errors.New("can not get committee because there is no share in runner")
-	}
-	committee := shareSample.Committee
+	committee := b.Share[0].Committee
```
iurii-ssv, line 71: Simplifying this because runner cannot be initialized without a `Share` (this is checked during runner-construction).

iurii-ssv:
Closing this PR and moving these changes over to https://github.com/ssvlabs/ssv/pull/2445 (since we need to resolve spec-tests either way it's better to do it in a single PR as it requires coordination with spec-team)

PR #2443, opened 2025-08-12
hotfix: fix check on missing operator data on startup

MatusKysel:


iurii-ssv (approved):
Glad to see we were able to identify the issue (cc @oleg-ssvlabs),

I wonder if we have (or plan to implement) some mechanism to:
- either let Validator know that he needs to register with +1 (or +N) operators to compensate for those that went offline
- or do this automatically for him

would be nice to also document it somewhere (or cross-link with this issue if we have it documented somewhere else)

Issue #2426, opened 2025-08-06
duty-runners: robust message replay


iurii-ssv:
The issue is described in detail [here](https://github.com/ssvlabs/ssv/pull/2149#issuecomment-2824241788), and we recently observed it in production, namely the following trimmed log-snippet shows how it can manifest for proposer-duty:
```
2025/08/06 15:25:02 INFO time_into_slot_ms=-24 {"L":"DEBUG","T":"2025-07-13T18:25:10.976Z","N":"Controller.Validator","M":"❗ could not handle message","pubkey":"81cddb0745c797fdfd530a7e22483eb58bd14b799eb74d66ec9b4cf98f47d026082d9ed3d6cb7b87e9b3acb9b6ff6633","role":"PROPOSER_RUNNER","signer":1225,"slot":12133924,"msg_type":"partial_signature","error":"failed processing randao message: invalid pre-consensus message: invalid partial sig slot","errorVerbose"...} duty_type=proposer
2025/08/06 15:25:02 INFO time_into_slot_ms=-19 {"L":"DEBUG","T":"2025-07-13T18:25:10.981Z","N":"Controller.Validator","M":"❗ could not handle message","pubkey":"81cddb0745c797fdfd530a7e22483eb58bd14b799eb74d66ec9b4cf98f47d026082d9ed3d6cb7b87e9b3acb9b6ff6633","role":"PROPOSER_RUNNER","signer":1226,"slot":12133924,"msg_type":"partial_signature","error":"failed processing randao message: invalid pre-consensus message: invalid partial sig slot","errorVerbose"...} duty_type=proposer
2025/08/06 15:25:02 INFO time_into_slot_ms=-10 {"L":"DEBUG","T":"2025-07-13T18:25:10.990Z","N":"Controller.Validator","M":"❗ could not handle message","pubkey":"81cddb0745c797fdfd530a7e22483eb58bd14b799eb74d66ec9b4cf98f47d026082d9ed3d6cb7b87e9b3acb9b6ff6633","role":"PROPOSER_RUNNER","signer":1227,"slot":12133924,"msg_type":"partial_signature","error":"failed processing randao message: invalid pre-consensus message: invalid partial sig slot","errorVerbose"...} duty_type=proposer
[...]

PR #2445, opened 2025-08-12
runners: robust message replay

iurii-ssv:
Closes https://github.com/ssvlabs/ssv/issues/2426

Having analysed the most common errors duty-runners encounter I've compiled a non-comprehensive/not-precise list of errors here to start with, split by category (it's kinda impractical to find every possible error a runner might encounter by simply looking at the code - so instead, or in addition to that, I've browsed through 30d log-history of various **stage** clusters, and our **prod** cluster to get the most commonly encountered ones):
```
# Fatal/Non-retryable errors (CL-request errors):

could not submit sync committee messages
could not submit Beacon block
could not get sync committee contribution
no clients to which to make call
could not submit blinded Beacon block
failed to submit aggregate and proof
Could not find any blocks with given slot

# Fatal/Non-retryable errors (miscellaneous errors):

// runner has no valid duties to execute (it's checking for that dynamically and is started regardless), while another operator thinks there are valid duties to execute + sent this message to us
no valid duties
// this seems to happen only to commit messages, but we need to investigate this error further
decided wrong instance
// we are done with this QBFT instance for one reason or another
instance stopped processing messages
// we are done with this QBFT instance for one reason or another
instance stopped processing timeouts
// we are done with this QBFT instance because it reached decided state
instance is already decided
// we are done with this QBFT instance because it reached decided state
consensus has already finished
// this looks like there is some issue on the message-sender side (why would he send un-justified proposal), but maybe some intricate validation logic on our side could also be the reason
proposal not justified
// we've already moved on to the next round, this message is irrelevant now
past round

# Retryable errors:

// we might not have started the duty yet, while another operator already did + sent this message to us
no running duty
// we might not have changed the round to the next one yet, while another operator already did + sent this message to us
wrong msg round
// we might not have started the QBFT instance yet, while another operator already did + sent this message to us
instance not found
// the message we've got is "from the future", it means we haven't started QBFT instance for the slot the message is targeting yet, while another operator already did + sent this message to us
future msg from height
// // the message we've got is targetting QBFT height that's different from the one our QBFT instance is targetting, not sure how/why this happens but for now we'll replay this message in case this is similar to "future message"-case
wrong msg height
// the message we've got is targetting slot that's different from the one our duty is targetting, not sure how/why this happens but for now we'll replay this message in case this is similar to "future message"-case
invalid partial sig slot
// we might not have received proposal message yet, while another operator already did and started preparing it
did not receive proposal for this round
// we might not have finished QBFT consensus phase yet, while another operator already did + sent this message to us
no decided value
```

The general idea is that we have 2 broad categories a duty-runner might encounter - "fatal errors" and "retryable errors". The names are self-explanatory, we want to replay those p2p messages that result into "retryable errors". To make it possible we need to annotate/wrap every error we are interested in so that we can check for it in the message-handler code (both for validator-based runners code and committee-based runners). 

_Before merging:_
- [x] make sure the added retries don't introduce harmful contention on the affected message-queues leading to some new delays at different duty-execution steps (eg. pre-consensus/consensus/post-consensus times for various duties shouldn't increase on average); This can also be monitored for via these log lines: `queue is full`, `queue is half-full`
- [x] **cpu**/**memory** usage shouldn't significantly grow due to this change
- [ ] (not super important) **throughput** seems to have increased with the changes in this PR by a factor of x2-x2.5 - I'm not sure what to make of it (need to check how it's measured because I don't see any direct way message replay could result into this - except for the fact that now we actually execute more duties in total since they aren't getting stuck, but x2 factor seems too much of an increase to come out of just that); ~~Could also be that I'm comparing with the older stage-branch version and this PR isn't the reason~~ unlikely, I've tested with another cluster now to see the same result
- [x] test on **stage**
- [ ] fix spec-tests (is addressed by https://github.com/ssvlabs/ssv-spec/pull/576)
- [x] resolve TODOs


protocol/v2/ssv/runner/runner.go
```diff
@@ -206,18 +207,30 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
...
 	if err != nil {
 		return false, nil, err
 	}
 
-	// we allow all consensus msgs to be processed, once the process finishes we check if there is an actual running duty
-	// do not return error if no running duty
-	if !b.hasRunningDuty() {
-		logger.Debug("no running duty")
-		return false, nil, nil
-	}
```
iurii-ssv: Since we can replay runner messages now, we don't need to have this ad-hoc handling for consensus messages (where we "apply the message-effects to the best extent we can") anymore - instead we'll return `ErrNoRunningDuty` so that the message will be replayed later (and hopefully the duty will be running by that time, if not we won't be able to progress either way)

protocol/v2/ssv/runner/error.go
```diff
@@ -0,0 +1,20 @@
...
+
+var (
+	ErrNoValidDutiesToExecute = fmt.Errorf("no valid duties to execute")
+
+	// Below is a list of retryable errors.
+
+	ErrNoRunningDuty         = fmt.Errorf("no running duty")
+	ErrInvalidPartialSigSlot = fmt.Errorf("invalid partial sig slot")
+	ErrInstanceNotFound      = fmt.Errorf("instance not found")
+	ErrFutureMsg             = fmt.Errorf("future msg")
```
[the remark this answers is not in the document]
iurii-ssv: `ErrFutureMsg` makes more sense than `ErrFuture` because runner package has a bunch of other stuff not related to messages

protocol/v2/ssv/runner/runner.go
iurii-ssv: Although this breaks a bunch of spec-tests ... so I'll revert that for now and leave a TODO-comment in the code to change this in the future (because we want to merge this PR asap)

.golangci.yaml
```diff
@@ -31,7 +31,7 @@ linters:
     - whitespace
   settings:
     errorlint:
-      errorf: true
+      errorf: false
```
iurii-ssv: Looks like this linter is classifying that kind of pattern as false-positive. I don't know how else to do what I'm trying to do here - which is to return an error of type `ErrNoProposalForRound` while also preserving the original error-message (the linter wants me to preserve the original error type, but we don't really need it - we are re-wrapping it with our own `ErrNoProposalForRound` so that the user of this code can check for it with `errors.Is` without worrying about the root-cause he shouldn't even know anything about).
```
protocol/v2/qbft/controller/controller.go:153:59: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
		return nil, fmt.Errorf("%w: %v", ErrNoProposalForRound, err)
```

protocol/v2/ssv/validator/opts.go
```diff
@@ -78,7 +74,7 @@ func NewCommonOptions(
 		NewDecidedHandler:   newDecidedHandler,
 		FullNode:            fullNode,
 		ExporterOptions:     exporterOptions,
-		QueueSize:           DefaultQueueSize,
+		QueueSize:           1000,
```
iurii-ssv, line 77: Not sure what queue size we should use exactly, but once I've added message replay the increased load on validator-queues showed up as `queue is full` warning-messages (meaning we are dropping some messages that are potentially critical to duty execution),

I think `1000` is reasonable enough because:
- this is what committee-queues use
- it doesn't show any increase in memory in Grafana by itself, although it might allow for significant memory spikes IF queues get full

Just calling it out in case you have better suggestions .

protocol/v2/qbft/controller/controller.go
```diff
@@ -149,6 +149,12 @@ func (c *Controller) UponExistingInstanceMsg(ctx context.Context, logger *zap.Lo
 	}
 
 	decided, _, decidedMsg, err := inst.ProcessMsg(ctx, logger, msg)
+	if errors.Is(err, instance.ErrNoProposalForRound) {
+		return nil, fmt.Errorf("%w: %v", ErrNoProposalForRound, err)
+	}
+	if errors.Is(err, instance.ErrWrongMsgRound) {
+		return nil, fmt.Errorf("%w: %v", ErrWrongMsgRound, err)
+	}
```
y0sher: why is this needed over just returning err?

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -111,11 +114,42 @@ func (c *Committee) ConsumeQueue(
...
+			return msgIDType(fmt.Sprintf("%d-%d-%d-%s-%s", msgSlot, msg.MsgType, psm.Type, msg.MsgID, signer))
+		}
+		return idUndefined
+	}
+	// msgRetries keeps track of how many times we've tried to handle a particular message. Since this map
+	// grows over time, we need to clean it up automatically. There is no specific TTL value to use for its
+	// entries - it just needs to be large enough to prevent unnecessary (but non-harmful) retries from happening.
+	msgRetries := ttlcache.New(
+		//ttlcache.WithTTL[msgIDType, int](10 * time.Minute),
+		ttlcache.WithTTL[msgIDType, int](100000 * time.Hour), // TODO - testing growth
```
y0sher: why not lower TTL? we don't want to retry a message from current epoch in next epoch for example, so an epoch or even a few slots could be enough ttl.

protocol/v2/qbft/controller/controller.go
iurii-ssv: We need to wrap errors that we want to make retryable (and all other errors by default are not gonna be retryable), the only other option I can think of would be to filter by error-message - but that's a poor solution because:
- error message can change and we won't know about it (so we will stop retrying those errors), and I think we'd want to update many of these error messages sooner-or-later because they lack some critical details (I'm updating some in this PR as well to make sense of them, so it's probably inevitable)
- ~~error message by itself is not always a good way to differentiate between when we want to retry and when we don't - there could be some cases when the same error is retryable or not depending on what duty/runner it comes from (for example, with error-wrapping we can explicitly make `ErrNoProposalForRound` error retryable for proposer-duty but not for aggregator-duty - this is not what we want to do with this specific error, but it illustrates the general idea)~~ I guess we could filter for "wrapped error-message" to get the same flexibility

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: The `msgRetries` map is there to prevent infinite retries (and limit it to `retryCount` const we've defined such that retries last for ~1 slot), it doesn't help in any way with "filtering out old messages" (there is no such intention for this map in particular) - so lowering TTL will result in **more** retries in general, not **less**

> we don't want to retry a message from current epoch in next epoch for example

So this ^ is a tricky part to handle when it comes to retries, as I mentioned in PR description - we need to classify errors in 2 categories "by hand" (and that's why [we need to wrap those errors that we want to make retryable](https://github.com/ssvlabs/ssv/pull/2445/#discussion_r2276052005)), those we want to retry and those we don't

errors we want to retry are explicitly listed here (all other errors result in "message drop"):
```
case (errors.Is(err, runner.ErrNoRunningDuty) || errors.Is(err, runner.ErrFuturePartialSigMsg) ||
				errors.Is(err, runner.ErrInstanceNotFound) || errors.Is(err, runner.ErrFutureConsensusMsg) ||
				errors.Is(err, runner.ErrNoProposalForRound) || errors.Is(err, runner.ErrWrongMsgRound) ||
				errors.Is(err, runner.ErrNoDecidedValue)) && msgRetryCnt < retryCount:
```

There are some interesting cases where the current error message isn't specific enough and can mean either a message-from-past (we **do not** want to retry that) or a message-from-future (we **do** want to retry that) - eg. `invalid partial sig slot`; I'm looking into those now

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -111,11 +114,41 @@ func (c *Committee) ConsumeQueue(
...
+		}
+		return idUndefined
+	}
+	// msgRetries keeps track of how many times we've tried to handle a particular message. Since this map
+	// grows over time, we need to clean it up automatically. There is no specific TTL value to use for its
+	// entries - it just needs to be large enough to prevent unnecessary (but non-harmful) retries from happening.
+	msgRetries := ttlcache.New(
+		ttlcache.WithTTL[msgIDType, int](10 * time.Minute),
+	)
+	go msgRetries.Start()
```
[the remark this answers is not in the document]
iurii-ssv: Nice catch, this applies to 2 places actually

protocol/v2/qbft/instance/error.go
```diff
@@ -0,0 +1,11 @@
+package instance
+
+import (
+	"fmt"
+)
+
+var (
+	ErrNoProposalForRound = fmt.Errorf("did not receive proposal for this round")
```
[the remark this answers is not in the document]
iurii-ssv: We need to define a separate set of errors in every package and map them as necessary so that the user(s) of `controller` package don't need to know anything about `instance` package (if they need to check for specific error-type)

it's just a clean separation of concerns, here it's easy to do so I don't see why not (error-names just "happen to match")

protocol/v2/ssv/runner/runner_signatures.go
```diff
@@ -51,34 +51,24 @@ func (b *BaseRunner) signBeaconObject(
...
-	if psigMsgs.Slot != slot {
-		return errors.New("invalid partial sig slot")
+	if psigMsgs.Slot < expectedSlot {
+		// this message is targeting a slot that's already passed - our runner has advanced to the next slot already,
+		// and we cannot process it anymore
+		return fmt.Errorf("invalid partial sig slot: %d, want: %d", psigMsgs.Slot, expectedSlot)
+	}
+
+	if psigMsgs.Slot > expectedSlot {
+		return fmt.Errorf("%w, got: %d, want: %d", ErrFuturePartialSigMsg, psigMsgs.Slot, expectedSlot)
```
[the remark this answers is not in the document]
iurii-ssv: Applied in https://github.com/ssvlabs/ssv/pull/2445/commits/a470c8c7e4f67e92a75bbd6618533a7eeec82517

protocol/v2/ssv/validator/committee_queue.go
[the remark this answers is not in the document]
iurii-ssv: Fixed in https://github.com/ssvlabs/ssv/pull/2445/commits/0eb434840daf79bcc06892df80e363e2692088d6

protocol/v2/ssv/validator/msgqueue_consumer.go
```diff
@@ -150,34 +184,68 @@ func (v *Validator) ConsumeQueue(logger *zap.Logger, msgID spectypes.MessageID,
...
+				go func() {
+					time.Sleep(retryDelay)
+					if pushed := q.Q.TryPush(msg); !pushed {
+						logger.Warn(
+							"❗ not gonna replay message because the queue is full",
+							zap.String("message_identifier", string(messageID(msg))),
+							fields.MessageType(msg.MsgType),
+						)
+					}
+				}()
```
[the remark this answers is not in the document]
iurii-ssv: Nice find, the `q` cannot change (and even if we refactor the enclosing func somehow it still wouldn't make sense for it to change) - so it's okay to use it directly, `msg` however can change ... I wish closures were more straightforward to use ... fixed in https://github.com/ssvlabs/ssv/pull/2445/commits/e9cb9586b6f442ddcb7d6da838c7234eb2d75e5d

.golangci.yaml
oleg-ssvlabs: wouldn't `fmt.Errorf("%w: %w", ErrNoProposalForRound, err)` work in this case?

https://go.dev/play/p/5035RNFbuE3

operator/validator/controller_test.go
```diff
@@ -800,7 +801,7 @@ func setupTestValidator(validatorPk spectypes.ValidatorPK, ownerAddressBytes, fe
 
 		Queues: map[spectypes.RunnerRole]validator.QueueContainer{
 			spectypes.RoleValidatorRegistration: {
-				Q: queue.New(1000),
+				Q: queue.New(zap.NewNop(), 1000),
```
oleg-ssvlabs: for tests, I think we use `log.TestLogger(t)` more often (also used in multiple places in this particular test file).

protocol/v2/qbft/controller/decided.go
```diff
@@ -70,7 +70,8 @@ func (c *Controller) ValidateDecided(msg *specqbft.ProcessingMessage) error {
 		return errors.New("not a decided msg")
 	}
 
-	if err := instance.BaseCommitValidationVerifySignature(msg, msg.QBFTMessage.Height, c.CommitteeMember.Committee); err != nil {
```
oleg-ssvlabs, line 73: may I ask why this change was needed?

protocol/v2/ssv/queue/queue.go
```diff
@@ -196,6 +197,14 @@ func (q *priorityQueue) Len() int {
 	return n
 }
 
+// maybeAlertInboxIsTooLarge logs a warning that we'd ideally never want to hit, if we do - we'd want to do
+// something about it.
+func (q *priorityQueue) maybeAlertInboxIsTooLarge() {
```
oleg-ssvlabs: not sure about the name for this method.

Maybe something like `checkInboxCapacity()`?

Maybe not in the scope of this PR - we could consider writing a metrics here as well.

protocol/v2/ssv/queue/queue.go
```diff
@@ -40,32 +42,31 @@ type Queue interface {
...
 	lastRead time.Time
 }
 
 // New returns an implementation of Queue optimized for concurrent push and sequential pop.
 // Pops aren't thread-safe, so don't call Pop from multiple goroutines.
-func New(capacity int) Queue {
+func New(logger *zap.Logger, capacity int) Queue {
 	return &priorityQueue{
-		inbox: make(chan *SSVMessage, capacity),
+		logger: logger,
```
oleg-ssvlabs: I would name the logger here, so logs are easily searchable

protocol/v2/ssv/validator/opts.go
oleg-ssvlabs: Does this value depend on the number of validators the node/committee is handling? If yes, we should take in consideration that committee validator cap will be increased to 3k.

In general, it would be good to test these changes at 3k clusters

protocol/v2/ssv/runner/runner.go
```diff
@@ -207,12 +208,30 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
 	}
 
 	decidedMsg, err := b.QBFTController.ProcessMsg(ctx, logger, msg)
+	if errors.Is(err, controller.ErrInstanceNotFound) {
```
oleg-ssvlabs: very subjective refactor suggestion:
```Go
if err != nil {
	var wrappedErr error

	switch {
	case errors.Is(err, controller.ErrInstanceNotFound):
		wrappedErr = fmt.Errorf("%w: %v", ErrInstanceNotFound, err)
	case errors.Is(err, controller.ErrFutureMsg):
		wrappedErr = fmt.Errorf("%w: %v", ErrFutureConsensusMsg, err)
	case errors.Is(err, controller.ErrNoProposalForRound):
		wrappedErr = fmt.Errorf("%w: %v", ErrNoProposalForRound, err)
	case errors.Is(err, controller.ErrWrongMsgRound):
		wrappedErr = fmt.Errorf("%w: %v", ErrWrongMsgRound, err)
	}

	if wrappedErr != nil {
		return false, nil, wrappedErr
	}
}
```

protocol/v2/qbft/controller/controller.go
oleg-ssvlabs: So, from what I understand, each package in the flow now defines a list of static errors that the root caller is interested in (for retries, for example). Therefore, these wraps are needed to propagate the errors back so the caller can perform error comparison. Do I understand it right?

protocol/v2/ssv/runner/runner_signatures.go
```diff
@@ -51,34 +51,24 @@ func (b *BaseRunner) signBeaconObject(
...
+	}
+
+	if psigMsgs.Slot > expectedSlot {
+		return fmt.Errorf("%w, message slot: %d, want slot: %d", ErrFuturePartialSigMsg, psigMsgs.Slot, expectedSlot)
 	}
 
-	// Get signer
-	msgSigner := psigMsgs.Messages[0].Signer // signer is the same in all psigMsgs.Messages and len(psigMsgs.Messages) > 0 (guaranteed by psigMsgs.Validate())
+	// Get signer, it is the same in all psigMsgs.Messages and len(psigMsgs.Messages) > 0 (guaranteed by psigMsgs.Validate()).
+	msgSigner := psigMsgs.Messages[0].Signer
```
oleg-ssvlabs, line 94: I understand the comment, but I would still check the length of the messages here, so it does not rely that `Validate` was previously called and it validates the length

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -111,11 +114,42 @@ func (c *Committee) ConsumeQueue(
...
-) error {
+) {
+	logger.Debug("📬 queue consumer is running")
+	defer logger.Debug("📪 queue consumer is closed")
+
 	state := *q.queueState
 
-	logger.Debug("📬 queue consumer is running")
-	lens := make([]int, 0, 10)
+	type msgIDType string
```
oleg-ssvlabs: Is this type declared purely to improve clarity?  

If so, I would prefer using a type alias (`type msgIDType = string`) instead of creating a new type, since no additional functionality is defined for it.

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -111,11 +114,42 @@ func (c *Committee) ConsumeQueue(
...
+	// messageID returns an ID that represents a potentially retryable message (msg.ID is the same for messages
+	// with different signers, slots, types, rounds, etc. - so we can't use just msg.ID as a unique identifier)
+	messageID := func(msg *queue.SSVMessage) msgIDType {
```
oleg-ssvlabs: It seems we repeat a lot of code in this file and in `msgqueue_consumer.go`.  

One repeated part is the "message type builder," and the other is the "retry handler."  
Since the message type builder belongs to the retry handler, these could be combined into a single type.  
It might be a bit tricky to come up with a good APIs for this type, but there is currently a significant amount of duplicated code.  

Separating this logic would also make testing easier

protocol/v2/ssv/validator/opts.go
iurii-ssv: > Does this value depend on the number of validators the node/committee is handling?

This particular change this PR is doing is - it's increasing this number for validator code (not committee code),

so we'd need to address this in a separate PR for committees, but from what I know about committee-duty, @MatusKysel correct me if I'm wrong, the number of messages there doesn't depend on the amount of validators it processes (qbft agrees on the vote that's independent from validator amount, and we send all individual validator-signatures all bundled together in 1 post-consensus message) - so there is nothing to address I don't think

> In general, it would be good to test these changes at 3k clusters

Yeah, the more testing we can do for this change - the better, I'm testing it on 2 different clusters on stage (both are close to 3k validators)

protocol/v2/qbft/controller/controller.go
iurii-ssv: > Do I understand it right?

Yeah, I guess, to put it differently, the errors are "re-defined" in each package because we want to hide away the implementation details (so that the user/caller of `controller` package doesn't have to know about qbft-`instance` package and can only deal with the errors `controller` package exposes),

sometimes there is no need for this kind of abstracting-away ... but I think it's a reasonable approach to use by-default (and it makes sense to diverge from it if there are reason to diverge, like if it was 100 different errors to re-define - maybe it wouldn't be worth it, esp. because in general this kind of wrapping would need to happen all over the cod, not in a single place like in this PR, we are just a bit "lucky" it's all in one place here)

iurii-ssv:
> One question - what was the previous behavior when the messages were not retried? Did that lead to a round change?

So, the **handler** (it's processing pre-consensus/consensus/post-consensus phases) would return an error and we'd drop this message as if it never came ... this could lead to various scenarios (and it depends on how our duty is implemented), couple of examples:
- maybe we managed to process other 3 / 4 messages just fine and were able to finish QBFT in round 1
- maybe we didn't reconstruct RANDAO (for proposer-duty) during pre-consensus phase and the duty is "just stuck now" (like literally it cannot proceed, but maybe we clean it up after some slots pass, I'm not sure)

operator/validator/controller_test.go
iurii-ssv: Yeah, I don't mind migrating to that, applied in https://github.com/ssvlabs/ssv/pull/2445/commits/8bcce63080641debf9a1578f66b59c4ec5b99671

protocol/v2/ssv/queue/queue.go
iurii-ssv: > Maybe something like checkInboxCapacity()?

The problem with `check` is that:
- it doesn't tell you anything about what it actually does (and, unless I read the impl-code, I'm inclined to assume it does a LOT - while here we just need to simply "alert")
- and when reading `check...` I'd probably expect some sort of result returned from it so I, the caller, can re-check the result and decide what to do with it

`maybeAlert...` is a bit ugly as well, not gonna lie

> Maybe not in the scope of this PR - we could consider writing a metrics here as well.

Yeah, queue size is a good candidate to be a metric - lets create [a task for it to address separately](https://github.com/ssvlabs/ssv/issues/2450)

protocol/v2/ssv/runner/runner.go
iurii-ssv: Personally I don't like the `switch` very much because it sort-of violates one of the core Golang idioms - the one that goes something like "there should be the only way to write this kind of functionality, the Go way" ... and with `switch`-statement we have at least 2 ways to do the same thing now,

and I guess [I don't like code inlined into if-statement(s)](https://github.com/ssvlabs/ssv/pull/2445#discussion_r2278572776) for the same reason,

appreciate your suggestion, but in terms of code-readability I wouldn't say one is clearly/obviously better than another - and so if that's the case I would leave the code as is,

regarding my views on `switch`-statement specifically I wrote [couple of thoughts here](https://github.com/ssvlabs/ssv/pull/2388#discussion_r2205959969) btw

protocol/v2/ssv/queue/queue.go
iurii-ssv: I was considering doing that - but then for some reason I thought it would overwrite the parent-logger name ... when in fact it just adds a suffix (as in `{parent}.suffix`), applied

protocol/v2/qbft/controller/decided.go
iurii-ssv: It's not really, I guess I was thinking this line is getting too long to "stick it under the `if` statement" - this is why I don't particularly like those over-loaded ifs and prefer to have a dedicated line for the `if err != nil {` (it helps the reader to immediately see what the if-condition is, since sometimes it would be something different)

I've reverted this now so it doesn't stand out for no reason :)

protocol/v2/ssv/runner/runner_signatures.go
iurii-ssv: If we do check the `len` here:
- we'd need to check the `len` in bunch of other similar places
- and what about checking `psigMsgs` is not `nil` ?

I think we have to make some assumptions for the code to stay sane/readable. I do however agree with you that sometimes it's better to double-check ... here though the `psigMsgs.Validate()` call happens just several lines above - so it would be a bit too defensive of a check I think

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: Yeah, I'm pretty much copy-pasting it :) And that's how the original committee-handler was implemented (by copy-pasting validator-handler) but I think eventually the 2 of these diverged enough that it doesn't make sense to deduplicate the whole consume-queue method anymore,

and that's also why sometimes trying to de-dup prematurely can get you in trouble,

As for my code I'm adding on top, I'm a bit hesitant to de-dup it just yet (and even cover this code with unit-tests for the very same reason) - lets first agree on its final form and then I can come back to it to see if we can do something about it. Just keep this tread open until then.

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: The reason to define a separate type here is so that compiler can type-check **for us** and prevent stupid bugs like doing `msgRetryItem := msgRetries.Get(msg.MsgID.String())` instead of what I really need to do - which is `msgRetryItem := msgRetries.Get(messageID(msg))`

Here it's a bit too easy to pass in the wrong `string` instead of the `msgIDType` type, so I think 1 additional line is worth it.

protocol/v2/ssv/validator/committee.go
```diff
@@ -36,9 +36,6 @@ type CommitteeRunnerFunc func(slot phase0.Slot, shares map[phase0.ValidatorIndex
 type Committee struct {
 	logger *zap.Logger
 
-	ctx    context.Context
-	cancel context.CancelFunc
```
iurii-ssv, line 40: There isn't any reason to store this `ctx` on `Committee struct` because the only place it's currently used in is `StartConsumeQueue` func, but this func accepts its own context-param - so we can re-use that instead. So I think we should simplify this while at it (and prefer `ctx` as func-argument over storing `ctx` as struct variable in general for simplicity/consistency).

protocol/v2/ssv/validator/committee_queue.go
oleg-ssvlabs: ah yes, I see the usage below. Compiler still allows passing untyped string constant, like `msgRetries.Get("random value")`, but it does not really matter. Having it strongly typed makes sense in this situation

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -169,55 +200,64 @@ func (c *Committee) ConsumeQueue(
...
-		// signer must be same for all messages, at least 1 message must be present (this is validated prior)
-		signer := psm.Messages[0].Signer
-		baseFields = []zap.Field{
-			zap.Uint64("signer", signer),
-			fields.Slot(psm.Slot),
+			const couldNotHandleMsgLogPrefix = "❗ could not handle message, "
+			switch {
+			case errors.Is(err, runner.ErrNoValidDutiesToExecute):
+				logger.Error(couldNotHandleMsgLogPrefix+"dropping message and terminating committee-runner", zap.Error(err))
+			case (errors.Is(err, runner.ErrNoRunningDuty) || errors.Is(err, runner.ErrFuturePartialSigMsg) ||
```
oleg-ssvlabs: wdyt about refactoring this? We could create a collection of retryable errors? Somewhere in [this file](https://github.com/ssvlabs/ssv/pull/2445/files#diff-88790c53467aa53cfd082d7f50f49c010478eb9299fda16fd73ccd2dd46724abR1)? 

Something like:

```Go
var retryableErrors = []error{
    runner.ErrNoRunningDuty,
    runner.ErrFuturePartialSigMsg,
    runner.ErrInstanceNotFound,
    runner.ErrFutureConsensusMsg,
    runner.ErrNoProposalForRound,
    runner.ErrWrongMsgRound,
}

func isRetryable(err error) bool {
    return slices.ContainsFunc(benignErrors, func(t error) bool {
        return errors.Is(err, t)
    })
}
```

protocol/v2/ssv/validator/validator_queue.go
```diff
@@ -170,53 +201,89 @@ func (v *Validator) ConsumeQueue(msgID spectypes.MessageID, handler MessageHandl
...
+		err = handler(ctx, msg)
+		if err != nil {
+			// We'll re-queue the message to be replayed later in case the error we got is retryable.
+			// We are aiming to cover most of the slot time (~12s), but we don't need to cover all 12s
+			// since most duties must finish well before that anyway, and will take additional time
+			// to execute as well.
+			// Retry delay should be small so we can proceed with the corresponding duty execution asap.
+			const (
+				retryDelay = 25 * time.Millisecond
+				retryCount = 40
```
oleg-ssvlabs: Do we really need to limit the number of retries?  
Wouldn't it be better to just stop when the handler returns a non-retryable error?  
Or you see it as  additional safety mechanism to prevent infinite retries in case of a bug or similar issue?

protocol/v2/ssv/validator/validator_queue.go
```diff
@@ -119,19 +120,49 @@ func (v *Validator) ConsumeQueue(msgID spectypes.MessageID, handler MessageHandl
 	}
 
 	v.logger.Debug("📬 queue consumer is running")
+	defer v.logger.Debug("📪 queue consumer is closed")
 
-	lens := make([]int, 0, 10)
+	type msgIDType string
+	// messageID returns an ID that represents a potentially retryable message (msg.ID is the same for messages
+	// with different signers, slots, types, rounds, etc. - so we can't use just msg.ID as a unique identifier)
+	messageID := func(msg *queue.SSVMessage) msgIDType {
```
oleg-ssvlabs: I would consider moving this code out and potentially even creating a new object, like `MessageRetryer`, that would also be responsible for message ID construction. This approach would make the code cleaner, reduce the size of this file, and make it easier to test. I guess it ultimately comes down to our vision for the package structure. So it's pretty subjective opinion

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: I like it very much 👍 applied in https://github.com/ssvlabs/ssv/pull/2445/commits/69bb040282f92c5c3239923833475aadbf3f451a

protocol/v2/ssv/validator/validator_queue.go
iurii-ssv: Yes, it's a basic safety-mechanism that we need at least for now (to not leak resources) - currently there are some retryable errors that might or might not succeed within 40 attempts (if we get to 40th attempt it almost 100% means that we won't be able to successfully replay the message, or if we do - it's too late anyway)

we may be able to improve our errors to the point where the separation between retryable and non-retryable is crystal-clear, but we are not there yet

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: Hmm, interesting, I guess string-constant is an exception because it does complain about `msg.MsgID.String()` or `msg.MsgID`

protocol/v2/ssv/validator/validator_queue.go
iurii-ssv: Yeah, now that this PR is approaching it's final stages I guess it's time to see if we can refactor/dedup this code a bit - https://github.com/ssvlabs/ssv/pull/2445/commits/916a68cab4cc2e1140fb4d8e5cc9d5024243a3af

We could maybe also de-duplicate retry-logic to make a single function both `Validator` and `Committee` can re-use, but I think it would be so abstract it's not worth it (also because the amount of code it will de-dup is small, and there is a chance to get a different behavior with it by accident)

protocol/v2/qbft/controller/controller.go
MatheusFranco99: Wouldn't be simpler to create an error structure with a flag to indicate whether the msg is invalid because it's bad, or if it went wrong because the state is not prepared for it?
I mean using this flag instead of strings inside the error

iurii-ssv: I might not fully see the vision:
- how do we unwrap that structure to check if the `retryable` flag is true/false ? Casting `error` to a `RetryableError` **struct** or something like that I guess
- but then, can we unwrap it if it's several packages away from the place we handle it ? As in: handler (checks the `re-tryable` flag) -> runner -> controller -> instance (error originates here) - which roughly mirrors what we currently have
- I don't think it is any better than using standard Golang wrapping + using `errors.Is` (maybe we could somehow adapt it to fit our needs, but it doesn't feel like we should be doing thing like that just to preserve current error-messages - we would like to enrich them with more details over time anyway)

iurii-ssv: To be fair, if we want to "preserve the original error-message" we can do it something like that, I think:
```go
// define a wrapper we can use with `errors.Is` - `errors.Is(err, RetryableError)`
type RetryableError struct {
    originalErr error
}

func NewRetryableError(originalErr error) *RetryableError {
    return &RetryableError{
        originalErr: originalErr,
    }
}

func (e *RetryableError) Error() string {
    return e.originalErr
}

// And then this code turns into:
...
if errors.Is(err, instance.ErrNoProposalForRound) {
  return nil, NewRetryableError(err)
}
if errors.Is(err, instance.ErrWrongMsgRound) {
  return nil, NewRetryableError(err)
}
...
```
and we might have to define this `RetryableError`-wrapper for every package involved (runner/controller/instance is probably our case).

But that's additional code that we don't need (it's not the simplest solution that comes to mind). And it doesn't help us with "splitting" `invalid partial sig slot` error in 2 different errors (in `validatePartialSigMsg` func), like I did in this PR:
- `invalid partial sig slot`
- `ErrFuturePartialSigMsg` (this is a newly-defined `error` current spec-tests don't expect)
```go
  if psigMsgs.Slot < expectedSlot {
		// this message is targeting a slot that's already passed - our runner has advanced to the next slot already,
		// and we cannot process it anymore
		return fmt.Errorf("invalid partial sig slot: %d, want: %d", psigMsgs.Slot, expectedSlot)
	}

	if psigMsgs.Slot > expectedSlot {
		return fmt.Errorf("%w, message slot: %d, want slot: %d", ErrFuturePartialSigMsg, psigMsgs.Slot, expectedSlot)
  }
```

**Update:** it does cut down on those `if`-checks though, so I decided to go for it (cause it results into more readable code) - https://github.com/ssvlabs/ssv/pull/2445/commits/abefb695e26812befc9bda81a026cd0f83c21c42
```go
	if errors.Is(err, &instance.RetryableError{}) {
		return nil, NewRetryableError(err)
	}
```

**Update 2:** I had to throw in an [additional commit on top](https://github.com/ssvlabs/ssv/pull/2445/commits/847a4fbfb16b291f15f168c8daa7dd4cf75607d4) to actually make this newly-defined `RetryableError` work - which is why I didn't initially go for this kind of solution ... still, it is a better solution long-term.

.golangci.yaml
iurii-ssv: Yeah, that's actually working nicely, but I decided to apply a different refactoring anyway that removes the need for `"%w: %w"` construction entirely (and rather wraps it with a separately-defined **struct**)

y0sher (dismissed):

protocol/v2/qbft/controller/controller.go
```diff
@@ -135,22 +130,24 @@ func (c *Controller) ProcessMsg(ctx context.Context, logger *zap.Logger, signedM
...
 	prevDecided, _ := inst.IsDecided()
 
 	// if previously decided, we don't process more messages
 	if prevDecided {
-		return nil, errors.New("not processing consensus message since instance is already decided")
+		return nil, spectypes.NewError(spectypes.SkipConsensusMessageAsInstanceIsDecidedErrorCode, "not processing consensus message since instance is already decided")
 	}
 
 	decided, _, decidedMsg, err := inst.ProcessMsg(ctx, logger, msg)
+	if errors.Is(err, &instance.RetryableError{}) {
```
nkryuchkov: Can you please explain this check? Wasn't it meant to be `errors.As`?

protocol/v2/qbft/controller/error_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestRetryableError(t *testing.T) {
+	someErr := fmt.Errorf("some error")
+	require.False(t, errors.Is(someErr, &RetryableError{}))
+
+	retryableErr := NewRetryableError(someErr)
+	require.True(t, errors.Is(retryableErr, &RetryableError{}))
+}
```
nkryuchkov, line 16:
```suggestion
func TestRetryableError(t *testing.T) {
	someErr := fmt.Errorf("some error")
	require.NotErrorIs(t, someErr, &RetryableError{})

	retryableErr := NewRetryableError(someErr)
	require.ErrorIs(t, retryableErr, &RetryableError{})
}
```

protocol/v2/qbft/instance/error.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func NewRetryableError(originalErr error) *RetryableError {
+	return &RetryableError{
+		originalErr: originalErr,
+	}
+}
+
+func (e RetryableError) Error() string {
+	return e.originalErr.Error()
```
nkryuchkov, line 33: It might be worth adding a nil check, although this is likely not expected to ever be nil

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -174,10 +170,8 @@ func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Roun
 }
 
 func (t *RoundTimer) waitForRound(round specqbft.Round, timeout <-chan time.Time) {
-	ctx, cancel := context.WithCancel(t.ctx)
-	defer cancel()
 	select {
-	case <-ctx.Done():
+	case <-t.ctx.Done():
```
nkryuchkov, line 174: While we are touching it, it might be the best time to discuss whether we need to use the context passed to `New` instead of adding a `ctx` parameter to `waitForRound`. We shouldn't change it in this PR though

protocol/v2/ssv/queue/queue.go
nkryuchkov: what about something like `warnIfInboxIsTooBig`?

protocol/v2/ssv/runner/error_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestRetryableError(t *testing.T) {
+	someErr := fmt.Errorf("some error")
+	require.False(t, errors.Is(someErr, &RetryableError{}))
+
+	wrappedErr := NewRetryableError(someErr)
+	require.True(t, errors.Is(wrappedErr, &RetryableError{}))
+
+	rErr := &RetryableError{}
+	require.False(t, errors.As(someErr, &rErr))
+	require.True(t, errors.As(wrappedErr, &rErr))
```
nkryuchkov: ditto

protocol/v2/ssv/runner/runner.go
```diff
@@ -257,19 +258,29 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
 		prevDecided, _ = b.State.RunningInstance.IsDecided()
 	}
 	if prevDecided {
-		return true, nil, errors.New("not processing consensus message since consensus has already finished")
+		return true, nil, spectypes.NewError(spectypes.SkipConsensusMessageAsConsensusHasFinishedErrorCode, "not processing consensus message since consensus has already finished")
 	}
 
 	decidedMsg, err := b.QBFTController.ProcessMsg(ctx, logger, msg)
+	if errors.Is(err, &controller.RetryableError{}) {
```
nkryuchkov: ditto

protocol/v2/ssv/runner/runner.go
```diff
@@ -257,19 +258,29 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
 		prevDecided, _ = b.State.RunningInstance.IsDecided()
 	}
 	if prevDecided {
-		return true, nil, errors.New("not processing consensus message since consensus has already finished")
+		return true, nil, spectypes.NewError(spectypes.SkipConsensusMessageAsConsensusHasFinishedErrorCode, "not processing consensus message since consensus has already finished")
```
nkryuchkov, line 260: Perhaps, we could go on with improving this and eventually use something shorter, not in this PR, though

protocol/v2/ssv/runner/runner.go
```diff
@@ -257,19 +258,29 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
...
+		return false, nil, NewRetryableError(err)
+	}
 	if err != nil {
 		return false, nil, err
 	}
 
-	// we allow all consensus msgs to be processed, once the process finishes, we check if there is
-	// an actual running duty - we consider this messaged "processed" (we might or might not get another
-	// message, hopefully we are running the duty by that time to finish the processing below)
+	// TODO: since we can replay runner messages now (implemented in https://github.com/ssvlabs/ssv/pull/2445),
```
nkryuchkov: Why do we need to mention the current PR?

protocol/v2/ssv/runner/runner_validations.go
```diff
@@ -50,38 +51,59 @@ func (b *BaseRunner) FallBackAndVerifyEachSignature(container *ssv.PartialSigCon
...
 	}
 
-	// TODO https://github.com/ssvlabs/ssv-spec/issues/142 need to fix with this issue solution instead.
-	if len(b.State.DecidedValue) == 0 {
-		return errors.New("no decided value")
+	// slotIsRelevant ensures the post-consensus message is even remotely relevant (eg. we might have already
+	// moved on to another duty that's targeting the next slot but received a post-consensus message relevant
+	// for the duty from the previous slot), this is a relaxed check that helps to filter out inappropriate
+	// messages as soon as possible, the exact slot validation occurs below.
+	slotIsRelevant := func(slot phase0.Slot) error {
```
nkryuchkov, line 62: Do we need to do these checks, given that we have message validation that should filter out invalid messages?

protocol/v2/ssv/runner/runner_validations.go
```diff
@@ -96,6 +118,24 @@ func (b *BaseRunner) ValidatePostConsensusMsg(ctx context.Context, runner Runner
...
+			// by design, hence there is no need to store slot number on decidedValue for CommitteeRunner.
+			expectedSlot := b.State.CurrentDuty.DutySlot()
+			return b.validatePartialSigMsg(psigMsgs, expectedSlot)
+		}
+	}
+	if err := validateMsg(); err != nil {
+		return err
+	}
+
+	return nil
```
nkryuchkov:
```suggestion
	return validateMsg()
```

or we can wrap the error

protocol/v2/ssv/spectest/msg_processing_type.go
```diff
@@ -79,16 +80,16 @@ func (test *MsgProcessingSpecTest) runPreTesting(ctx context.Context, logger *za
 	switch test.Runner.(type) {
 	case *runner.CommitteeRunner:
 		guard := validator.NewCommitteeDutyGuard()
-		c = baseCommitteeWithRunnerSample(ctx, logger, ketSetMap, test.Runner.(*runner.CommitteeRunner), guard)
+		c = baseCommitteeWithRunnerSample(logger, ketSetMap, test.Runner.(*runner.CommitteeRunner), guard)
```
nkryuchkov: while we are touching this
```suggestion
		c = baseCommitteeWithRunnerSample(logger, keySetMap, test.Runner.(*runner.CommitteeRunner), guard)
```

protocol/v2/ssv/validator/committee.go
```diff
@@ -236,127 +224,23 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
-		return nil
-	default:
-		return traces.Errorf(span, "unknown message type: %d", msgType)
-	}
-
-	span.SetStatus(codes.Ok, "")
-	return nil
-}
+func (c *Committee) unsafePruneExpiredRunners(logger *zap.Logger, currentSlot phase0.Slot) {
+	const runnerExpirySlots = 34
```
nkryuchkov: Can we export this value in the message validation?

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -129,7 +111,7 @@ func (c *Committee) ConsumeQueue(
 			// If no proposal was accepted for the current round, skip prepare & commit messages
 			// for the current round.
 			filter = func(m *queue.SSVMessage) bool {
-				sm, ok := m.Body.(*specqbft.Message)
+				sm, ok := m.Body.(*qbft.Message)
```
nkryuchkov: I think this needs to be `specqbft`, we use the `spec` prefix everywhere except some code that's almost fully copied from the spec

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -151,55 +133,184 @@ func (c *Committee) ConsumeQueue(
...
+
+			msgLogger = logWithMessageMetadata(msgLogger, msg).
+				With(zap.String("message_identifier", string(c.messageID(msg)))).
+				With(zap.Int("attempt", msgRetryCnt+1))
+
+			const couldNotHandleMsgLogPrefix = "could not handle message, "
+			switch {
+			case errors.Is(err, runner.ErrNoValidDutiesToExecute):
+				msgLogger.Error("❗ "+couldNotHandleMsgLogPrefix+"dropping message and terminating committee-runner", zap.Error(err))
+			case errors.Is(err, &runner.RetryableError{}) && msgRetryCnt < retryCount:
```
nkryuchkov: meant to be `errors.As`?

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -77,131 +75,292 @@ func (v *Validator) EnqueueMessage(ctx context.Context, msg *queue.SSVMessage) {
...
-		defer v.mtx.RUnlock()
-		var ok bool
-		q, ok = v.Queues[msgID.GetRoleType()]
-		if !ok {
-			return errors.New(fmt.Sprintf("queue not found for role %s", msgID.GetRoleType().String()))
+		v.logger.Debug("📬 queue consumer is running")
+		defer v.logger.Debug(
+			"📪 queue consumer is closed",
+			zap.Bool("ctx_done", ctx.Err() != nil),
+		)
```
nkryuchkov: Did you mean `ctx.Err() != nil` to be executed on the deferred call?

```suggestion
		defer func() {
			v.logger.Debug("📪 queue consumer is closed",
				zap.Bool("ctx_done", ctx.Err() != nil),
			)
		}
```

The current implementation calculates the args immediately and defers the call with the pre-calculated args

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -77,131 +75,292 @@ func (v *Validator) EnqueueMessage(ctx context.Context, msg *queue.SSVMessage) {
...
+			// Handle the message, potentially scheduling a message-replay for later.
+			err = handler(ctx, msgLogger, msg)
+			if err != nil {
+				// We'll re-queue the message to be replayed later in case the error we got is retryable.
+				// We are aiming to cover most of the slot time (~12s), but we don't need to cover all 12s
+				// since most duties must finish well before that anyway, and will take additional time
+				// to execute as well.
+				// Retry delay should be small so we can proceed with the corresponding duty execution asap.
+				const (
+					retryDelay = 25 * time.Millisecond
```
nkryuchkov: Can we bind durations to the network config just in case the [EIP-7782](https://ethereum-magicians.org/t/eip-7782-the-case-for-2x-shorter-slot-times-in-glamsterdam/24616) will be merged?

ssvsigner/go.sum
```diff
@@ -331,8 +331,10 @@ github.com/spaolacci/murmur3 v1.1.0 h1:7c1g84S4BPRrfL5Xrdp6fOJ206sU9y293DDHaoy0b
 github.com/spaolacci/murmur3 v1.1.0/go.mod h1:JwIasOWyU6f++ZhiEuf87xNszmSA2myDM2Kzu9HwQUA=
 github.com/ssvlabs/eth2-key-manager v1.5.6 h1:BMxVCsbcIlUiiO0hpePkHxzX0yhKgMkEzVSoNmSXySM=
 github.com/ssvlabs/eth2-key-manager v1.5.6/go.mod h1:tjzhmMzrc0Lzc/OMW1h9Mz8AhmKH7FQC/nFiMNJ0bd8=
-github.com/ssvlabs/ssv-spec v1.1.4 h1:xyjMNe9spQcWBxePPdGRar/9eQiq0XQ4htNBbYXYHqY=
-github.com/ssvlabs/ssv-spec v1.1.4/go.mod h1:GedhFYGHVJRYYH3nEp05Gn14tyvg6VbTbaIxrMtI7Cg=
+github.com/ssvlabs/ssv-spec v1.1.5-0.20251020132806-d592ea7e7dd6 h1:WzFTuLfjFbkwZthFispxU3fF9GUqXs4hOgxToWPbS28=
```
nkryuchkov: Let's run `go mod tidy`

.golangci.yaml
nkryuchkov: Could it be because we need `errors.Join(ErrNoProposalForRound, err)` instead of `fmt.Errorf("%w: %w", ErrNoProposalForRound, err)` here? I think the latter is semantically incorrect here, as it's usually used to build a chain of errors, where they are wrapped. Duplicating the `%w` verb includes more than one error, and I think it's the use case for `errors.Join`

go.sum
```diff
@@ -770,8 +770,10 @@ github.com/ssvlabs/eth2-key-manager v1.5.6 h1:BMxVCsbcIlUiiO0hpePkHxzX0yhKgMkEzV
 github.com/ssvlabs/eth2-key-manager v1.5.6/go.mod h1:tjzhmMzrc0Lzc/OMW1h9Mz8AhmKH7FQC/nFiMNJ0bd8=
 github.com/ssvlabs/go-eth2-client v0.6.31-0.20250922150906-26179dd60c9c h1:iNQoRbEajriawtkSFiyHsJNiXyfyTrPrnmO0NaWiNv4=
 github.com/ssvlabs/go-eth2-client v0.6.31-0.20250922150906-26179dd60c9c/go.mod h1:fvULSL9WtNskkOB4i+Yyr6BKpNHXvmpGZj9969fCrfY=
-github.com/ssvlabs/ssv-spec v1.1.4 h1:xyjMNe9spQcWBxePPdGRar/9eQiq0XQ4htNBbYXYHqY=
-github.com/ssvlabs/ssv-spec v1.1.4/go.mod h1:GedhFYGHVJRYYH3nEp05Gn14tyvg6VbTbaIxrMtI7Cg=
+github.com/ssvlabs/ssv-spec v1.1.5-0.20251020132806-d592ea7e7dd6 h1:WzFTuLfjFbkwZthFispxU3fF9GUqXs4hOgxToWPbS28=
```
nkryuchkov: ditto

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: It kind of depends on what we want to achieve, `context` is a bit of a messy thing (since it have many applications),

here, it actually seems like we are spawning these go-routines (in `TimeoutForRound`) but also terminating these once timeout happens ... so I don't really see any issues/need for additional `ctx` parameter here (there doesn't seems to be any memory leak happening, on a glance)

if anything, I would probably even want to remove the `t.ctx` entirely to simplify things because we are not re-creating `RoundTimer` instances (we spawn a bunch and they live forever anyway)

Did you have some particular idea for what we want to improve here with respect to `ctx` usage ?

protocol/v2/ssv/runner/runner.go
iurii-ssv: yeah, this is just an initial implementation that works with minimal effort

protocol/v2/ssv/runner/runner_validations.go
iurii-ssv: TBH I don't really like it either, but it allows us to drop a significant portion of non-retryable messages **sooner/faster** - instead of **having** to classify them as retryable (because some are indeed retryable) due to this code below which "cannot be easily moved without a significant refactoring, and possibly additional spec-tests adjustments":
```
	if b.State.RunningInstance == nil {
		return NewRetryableError(spectypes.WrapError(spectypes.NoRunningConsensusInstanceErrorCode, ErrInstanceNotFound))
	}

	// TODO https://github.com/ssvlabs/ssv-spec/issues/142 need to fix with this issue solution instead.
	decided, decidedValueBytes := b.State.RunningInstance.IsDecided()
	if !decided || len(b.State.DecidedValue) == 0 {
		return NewRetryableError(spectypes.WrapError(spectypes.NoDecidedValueErrorCode, ErrNoDecidedValue))
	}
```
So I feel like it's the best trade-off we can get here with what we have.

protocol/v2/ssv/validator/committee.go
iurii-ssv: It's a good idea, but it's not trivial since (hence out of scope of this PR)
- we wouldn't want to import message-validation package here (I think)
- it is dynamically defined - https://github.com/ssvlabs/ssv/blob/b5f1a490cbefea6267553322c10c6857e4afa915/message/validation/validation.go#L325-L328 so we probably need to move it somewhere else (creating a dedicated component/service for it, but then we probably want this component/service to also provide other similar values if we do that ...)

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: Right, applied here and in a bunch of other files (where relevant)

protocol/v2/ssv/validator/queue_validator.go
iurii-ssv: Right - good point, 

I used it for debugging, and we probably don't need it anymore, removed

protocol/v2/ssv/runner/runner.go
iurii-ssv: This is just an observation I wanted to re-visit, removed the TODO now

protocol/v2/qbft/instance/error.go
iurii-ssv: I think in the nil-case we would want to panic to find the issue early - so I would prefer that as "a default implementation/behavior" because it would be hard to ensure we do this kind of nil-check consistently in all the places

protocol/v2/qbft/controller/controller.go
iurii-ssv: This check works because `RetryableError` implements `Is` method, but apparently (after reading a bit more about it) it is not the most common way to implement it, and `errors.Is` is "supposed to be used mostly for checks against sentinel-values" ...

so I updated these checks to use a helper-func `IsRetryable` (this also keeps it more readable I think)

protocol/v2/ssv/validator/queue_validator.go
iurii-ssv: Yeah that would be better, applied,

just note: `retryDelay` doesn't depend on any network-related configuration while `retryCount` does

protocol/v2/ssv/validator/committee.go
nkryuchkov: @iurii-ssv what do you think about something like this?
```suggestion
	const lateSlotAllowance = 2 // from message/validation/const.go
	runnerExpirySlots := c.networkConfig.SlotsPerEpoch + lateSlotAllowance 
```

protocol/v2/qbft/roundtimer/timer.go
nkryuchkov: Well, how I understand the context is that it's something that the caller wants to control. Indeed, we have the goroutine spawn, but we could pass the caller's context, however, the caller isn't blocked, so maybe we don't need it. So perhaps let's leave it as is

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -151,55 +130,183 @@ func (c *Committee) ConsumeQueue(
...
+
+		// Handle the message, potentially scheduling a message-replay for later.
+		err = handler(ctx, msgLogger, msg)
+		if err != nil {
+			// We'll re-queue the message to be replayed later in case the error we got is retryable.
+			// We are aiming to cover most of the slot time (~12s), but we don't need to cover the
+			// full slot (all 12s) since most duties must finish well before that anyway, and will
+			// take additional time to execute as well.
+			// Retry delay should be small so we can proceed with the corresponding duty execution asap.
+			const retryDelay = 25 * time.Millisecond
```
nkryuchkov, line 155: Can we extract this into a method because it's duplicated in `protocol/v2/ssv/validator/queue_validator.go`? And maybe some code around it too

message/validation/const.go
```diff
@@ -12,7 +12,6 @@ const (
 	// clockErrorTolerance is the maximum amount of clock error we expect to see between nodes.
 	clockErrorTolerance     = time.Millisecond * 50
 	allowedRoundsInFuture   = 1
-	allowedRoundsInPast     = 2
```
nkryuchkov, line 15: I don't like unused code either, but this is used in the commented-out code in `roundBelongsToAllowedSpread`, so perhaps we also need to clean up `roundBelongsToAllowedSpread` and leave a TODO to implement that check. Back then, there were some issues with this check, IIRC, due to round durations

.golangci.yaml
iurii-ssv: Hm, I've actually changed the `%w: %w` in the recent commits in this PR to something else - so we can revert `errorf: true` now,

regarding `errors.Join` - good point, for some reason I thought `errors.Is` and `errors.As` wouldn't work with `errors.Join` ... but apparently it does

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: While those code-blocks are very similar they are still different enough to warrant separate funcs (and they might diverge even further in the future),

I'm not against de-duping it later on, if we find a way to converge this to a single component, but I don't think it's a good idea to do in this PR

message/validation/const.go
iurii-ssv: Ah okay, didn't think to look for in-comment mentions, reverted 👍

protocol/v2/ssv/validator/committee.go
iurii-ssv: Creative :) Applied

protocol/v2/ssv/queue/queue.go
iurii-ssv: Updated to `warnIfInboxIsTooBig` :)

y0sher (approved):

PR #2447, opened 2025-08-14
fix(all): remove dead code

julienh-ssv:
**What**
This PR removes unused (“dead”) code that is no longer referenced anywhere in the project.

Also adds `deadcode` check to the linter. 

As an added benefit, this change helps reduce binary size & improve compile times.

**Safety**
Safe because all removed code was unused.

### DIFF
--- a/operator/validators/testutils.go
+++ b/operator/validators/testutils.go
@@ -6,7 +6,7 @@ import (
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/validator"
 )
 
-// WithInitialState sets initial state
+// Test helper to set initial state during tests.
 func WithInitialState(vstate map[spectypes.ValidatorPK]*validator.Validator, mstate map[spectypes.CommitteeID]*validator.Committee) Option {
 	return func(vm *ValidatorsMap) {
 		vm.validators = vstate


utils/threshold/threshold.go
```diff
@@ -2,23 +2,10 @@ package threshold
...
-	curveOrder = new(big.Int)
-)
-
-// Init initializes BLS
-func Init() {
-	_ = bls.Init(bls.BLS12_381)
-	_ = bls.SetETHmode(bls.EthModeDraft07)
-
-	curveOrder, _ = curveOrder.SetString(bls.GetCurveOrder(), 10)
-}
```
MatusKysel, line 20: why this was moved ?

beacon/goclient/mocks/server.go
```diff
@@ -1,3 +1,8 @@
+//go:build testutils
+
+// This file contains helpers for tests only.
+// It will not be compiled into production binaries.
+
 package mocks
```
iurii-ssv: Interesting solution, I wonder if renaming `server.go` to `server_test.go` would achieve the same result ? If so that would be a simpler way to go about it I think

eth/executionclient/execution_client_test.go
```diff
@@ -62,6 +62,20 @@ func simTestBackend(testAddr ethcommon.Address) *simulator.Backend {
...
+		s.logBatchSize = size
+	}
+}
+
+// WithLogBatchSizeMulti sets log batch size.
+func WithLogBatchSizeMulti(size uint64) OptionMulti {
+	return func(s *MultiClient) {
+		s.logBatchSize = size
+	}
+}
```
iurii-ssv: I think it's better to keep all options (those `func With...`) in 1 place, 

I assume you've moved these here because they are only used in tests ? If that's the case I would maybe even get rid of these options and inline `s.logBatchSize = size` instead

migrations/migration_6_share_exit_epoch_test.go
```diff
@@ -117,6 +117,15 @@ func assertMigratedShares(t *testing.T, db basedb.Database, key []byte, seededSh
...
+// Encode encodes Share using ssz.
+func (s *migration_6_OldStorageShare) Encode() ([]byte, error) {
+	result, err := s.MarshalSSZ()
+	if err != nil {
+		return nil, fmt.Errorf("marshal ssz: %w", err)
+	}
+	return result, nil
+}
+
 func seedDatabase(numOfItems int, db basedb.Database, storageKey []byte) ([]*migration_6_OldStorageShare, error) {
```
iurii-ssv: I'm not sure why we'd want to move this `Encode` func here, could you clarify ? I think it's better to keep all the methods in a single file (unless the struct owning those methods is so large that we want to split it into many files grouping methods in some logical way)

network/discovery/testutils.go
```diff
@@ -27,6 +32,12 @@ type Bootnode struct {
...
 }
 
+// Close implements io.Closer
+func (b *Bootnode) Close() error {
+	b.cancel()
+	return nil
+}
+
 // NewBootnode creates a new bootnode
 func NewBootnode(pctx context.Context, logger *zap.Logger, ssvConfig *networkconfig.SSV, opts *BootnodeOptions) (*Bootnode, error) {
```
iurii-ssv: I think the general convention we follow is that **constructor(s)** (the `func New...` method) should be the very 1st method following **struct**-definition so it's easy to figure out whether or not a **struct** has a constructor even

that's why the func order was the way it was, not sure if you have any strong reason for this change ?

network/testing/keys.go
```diff
@@ -20,11 +24,20 @@ type NodeKeys struct {
...
 }
 
+// GenNetworkKey generates a new network key
+func GenNetworkKey() (*ecdsa.PrivateKey, error) {
+	privInterfaceKey, _, err := crypto.GenerateSecp256k1Key(crand.Reader)
+	if err != nil {
+		return nil, errors.WithMessage(err, "could not generate 256k1 key")
+	}
+	return commons.ECDSAPrivFromInterface(privInterfaceKey)
+}
```
iurii-ssv: same reasoning for constructor

nodeprobe/nodeprobe_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	"go.uber.org/zap"
 )
 
+func (p *Prober) Healthy(context.Context) (bool, error) {
+	return p.healthy.Load(), nil
+}
+
 func TestProber(t *testing.T) {
```
iurii-ssv: same reasoning as above, I don't think it's a good idea to spread out struct-methods like that

Makefile
```diff
@@ -107,9 +116,14 @@ docker-benchmark:
 	@docker run --rm ssv_tests make benchmark
 
 .PHONY: build
+.DEFAULT_GOAL := build # this makes `make` default to `make build`
 build:
 	CGO_ENABLED=1 go build -o ./bin/ssvnode -ldflags "-X main.Commit=`git rev-parse HEAD` -X main.Version=`git describe --tags $(git rev-list --tags --max-count=1)`" ./cmd/ssvnode/
 
+.PHONY: build-tests
+build-tests:
+	./scripts/build-tests.sh
```
iurii-ssv: I assume this is needed mostly to "measure" how long tests compilation takes ? I would leave a comment about that, 

and maybe even comment-out this make-command (and also leave a note that it's a very niche thing, but useful one - hence we wouldn't want to remove it entirely) since it's not something we really need to use on a daily basis

**Update:** on a second thought, if the only thing we really need is to measure that build times improved before/after why not do just that (and not have a separate script/make-command in the first place) ?

iurii-ssv (commented):
This looks mostly good to me, however some of the code-moves are not really make for a better code-structure IMO

Makefile
```diff
@@ -31,51 +31,60 @@ GET_TOOL=go get -modfile=tool.mod -tool
 RUN_TOOL=go tool -modfile=tool.mod
 SSVSIGNER_RUN_TOOL=go tool -modfile=../tool.mod
 
+TEST_TAGS=blst_enabled,testutils
```
iurii-ssv: I think the `make full-test` command might also need `testutils` ?

Makefile
```diff
@@ -31,51 +31,60 @@ GET_TOOL=go get -modfile=tool.mod -tool
...
 	cd ssvsigner && GOWORK=off $(SSVSIGNER_RUN_TOOL) github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -c ../.golangci.yaml -v ./...
 
+.PHONY: deadcode-lint
+deadcode-lint:
+	./scripts/deadcode.sh
+
 .PHONY: full-test
 full-test:
 	@echo "Running all tests"
 	@go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 -v ./...
```
oleg-ssvlabs: blst_enabled -> $(TEST_TAGS)

Makefile
```diff
@@ -177,6 +191,7 @@ tools:
 	$(GET_TOOL) github.com/ferranbt/fastssz/sszgen
 	$(GET_TOOL) github.com/ethereum/go-ethereum/cmd/abigen
 	$(GET_TOOL) github.com/golangci/golangci-lint/v2/cmd/golangci-lint
+	$(GET_TOOL) golang.org/x/tools/cmd/deadcode@latest
```
oleg-ssvlabs: Nit: We do not currently specify the versions of the tools. To maintain consistency, we should either specify them everywhere or not at all :)

It might be worth considering locking the versions, at least to the major + minor (if possible).

beacon/goclient/mocks/server.go
oleg-ssvlabs: it's a good point. 

I like this solution better. To my knowledge, compiler skips files with `_test` suffixes. It would be simpler

Makefile
oleg-ssvlabs: Also, could we potentially rather enable "unused" linter?

Some discussion [here](https://github.com/golangci/golangci-lint/issues/4254)

scripts/build-tests.sh
oleg-ssvlabs, line 1: could you please explain the purpose of this script?

> Compile all test packages without running them, show compilation errors

Wouldn't we achieve the same with:

```bash
go build -tags=blst_enabled,testutils ./...
```

I primarily think about maintainability of it.

eth/eventsyncer/event_syncer_test.go
```diff
@@ -50,6 +50,12 @@ var (
 	testAddr = crypto.PubkeyToAddress(testKey.PublicKey)
 )
 
+func WithStalenessThreshold(threshold time.Duration) Option {
```
oleg-ssvlabs: Similar to what @iurii-ssv mentioned [here](https://github.com/ssvlabs/ssv/pull/2447/files#r2276616691). 

Instead of moving this to a test file, I would rather set `stalenessThreshold` directly in the test, skipping the options pattern.  

The options pattern is useful for external consumers of the package. In this case, it feels redundant.

oleg-ssvlabs (changes requested):
Great work on cleaning up unused code!  

The only concern I have is that if an object `Foo` exposes a method `Work()` and it is currently unused, it does not necessarily mean it should be removed. `Foo` may expose it because it is part of its intended functionality, and in the future, that method might be utilized. This mindset is more applicable to libraries than to a single executable like our SSV Node, although one could argue that each package within a module is, in a sense, its own library.  

In other words, we should be cautious about removing all unused code, as some may be temporarily unused and still fit well within the package's scope and semantic.  

I’ve left a couple of comments that will likely require changes.  
I will submit another review later.

utils/threshold/threshold.go
julienharbulot: because it's only used by the tests not anywhere in the main code

migrations/migration_6_share_exit_epoch_test.go
julienharbulot: Removed it was unused

network/discovery/testutils.go
julienharbulot: that was a typo, fixed

network/testing/keys.go
julienharbulot: ok

Makefile
julienharbulot: removed

Makefile
julienharbulot: thanks

Makefile
julienharbulot: Fixed the version for consistency, good catch thanks 👍 

Regarding `deadcode` vs `unused`, I found that `deadcode` is much more powerful because it also detects unused public functions.

scripts/build-tests.sh
julienharbulot: removed

eth/executionclient/execution_client_test.go
julienharbulot: let's try and inline it

nodeprobe/nodeprobe_test.go
julienharbulot: ok, removed as well as it was unused

eth/eventsyncer/event_syncer_test.go
julienharbulot: okay let's inline it

beacon/goclient/mocks/server.go
julienharbulot: okay fixed

observability/log/context.go
```diff
@@ -18,8 +18,3 @@ func FromContext(ctx context.Context) *zap.Logger {
 	}
 	return logger
 }
-
-// WithContext returns a context with the logger.
-func WithContext(ctx context.Context, logger *zap.Logger) context.Context {
-	return context.WithValue(ctx, loggerKey, logger)
-}
```
iurii-ssv, line 25: That doesn't need to move

observability/log/fields/testutils.go
```diff
@@ -0,0 +1,16 @@
+package fields
```
iurii-ssv, line 1: Maybe this file should end with `_test.go`, this probably applies to every **testutils** fille in this PR ?

operator/validators/testutils.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/ssvlabs/ssv/protocol/v2/ssv/validator"
+)
+
+// WithInitialState sets initial state
+func WithInitialState(vstate map[spectypes.ValidatorPK]*validator.Validator, mstate map[spectypes.CommitteeID]*validator.Committee) Option {
+	return func(vm *ValidatorsMap) {
+		vm.validators = vstate
+		vm.committees = mstate
+	}
+}
```
iurii-ssv: That doesn't need to move

iurii-ssv (commented):
Good stuff 👍 , couple more nits to go (I mentioned above that it'd be better to split this PR in smaller ones to address separate concerns, but at this point it's easier to just merge it as is - just keep that in mind for future similar PRs)

observability/log/context.go
julienharbulot: Dead code that is not used in prod has to move to a test file or testutils, otherwise linter will fail.

observability/log/fields/testutils.go
julienharbulot: Actually it won't work: `_test.go` files cannot be imported from other modules so `testutils_test.go` won't be importable.

observability/log/context.go
iurii-ssv: I mean, similar to all the previous cases - if the `With...` option isn't used anywhere besides test-code we should probably inline whatever we need from it in the test itself (rather than having `With...` option defined in a test file)

and the same, I guess, applies for this comment - https://github.com/ssvlabs/ssv/pull/2447#discussion_r2280779920

julienharbulot: yes you are right it can be removed 👍

operator/validators/testutils.go
julienharbulot: This one is used in another package's tests (`validator` no final `s`) so we should leave it because the fields `.validators` and `.committeers` are unexported.

Screenshot 2025-08-18 at 3 58 33 pm

iurii-ssv: > yeah I see, hmm, this is a bit ugly ... I would bundle this With... option with the rest of the options that are used in prod code, but we are gonna fight the linter then ... so personally I would use //nolint: comment to "correct the linter" but if you think it's better to leave this code in testutils.go that's also fine I guess

Issue #2452, opened 2025-08-18
logs: events should have slot numbers (similar to all other message types)


iurii-ssv:
Currently [Event-type messages](https://github.com/ssvlabs/ssv/blob/90feb4ec3e91a964818b36b7433b1816d354d162/protocol/v2/message/msg.go#L14-L14) don't bear the corresponding slot number they are targeting - as is [explained in the comments here](https://github.com/ssvlabs/ssv/blob/90feb4ec3e91a964818b36b7433b1816d354d162/protocol/v2/ssv/queue/messages.go#L33-L51).

This is unfortunate because we end up with some logs (generated by those events) without a good way to link them to their corresponding duties (we cannot even reconstruct duty ID for committee-duty, for example, without slot number).

So we should implement this so that every event bears an associated slot number to resolve this.

iurii-ssv:
Resolved by https://github.com/ssvlabs/ssv/pull/2453

PR #2453, opened 2025-08-18
duties: correct logger usage (part 2)

iurii-ssv:
This is a follow-up to https://github.com/ssvlabs/ssv/pull/2400 to address the following:
- to add missing log-fields, remove duplicate ones
- to standardize/refactor the way we set up the `logger` with all the necessary context (**slot**, **dutyID**, etc.) - so it's very clear what fields are added in every case; This is to make sure there won't be a "(part 3)" necessary
- also resolves https://github.com/ssvlabs/ssv/issues/2452 - this is the simplest way to make logger work properly after https://github.com/ssvlabs/ssv/pull/2400 was merged (without this change the complexity of handling Event message-type is just too much to deal with, we have to "untangle this code sooner-or-later" in order to implement https://github.com/ssvlabs/ssv/issues/1828 and since https://github.com/ssvlabs/ssv/pull/2400 is already merged - this is the best time to do it)


protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -1549,13 +1555,13 @@ func TestQueueLoadAndSaturationScenarios(t *testing.T) {
 	mainLogger, _ := zap.NewDevelopment()
```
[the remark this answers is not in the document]
iurii-ssv: Yes, this is left unchanged intentionally because this will be updated in https://github.com/ssvlabs/ssv/pull/2445#discussion_r2278560696 (a different PR) and I don't want to update it here as well to keep merge-conflicts as few as possible

cc @oleg-ssvlabs

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,20 +230,36 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
+			observability.ValidatorMsgIDAttribute(msgID),
+			observability.RunnerRoleAttribute(msgID.GetRoleType()),
+			observability.CommitteeIDAttribute(committeeID),
+			observability.BeaconSlotAttribute(slot),
 			observability.DutyIDAttribute(dutyID),
-		))
+		),
+		trace.WithLinks(trace.LinkFromContext(msg.TraceContext)),
+	)
 	defer span.End()
```
iurii-ssv: This `trace.WithLinks(trace.LinkFromContext(msg.TraceContext)),` part is present for `(v *Validator)` but was missing here (on `(c *Committee)`)

cc @oleg-ssvlabs just to clarify it's not intentionally omitted or something like that

protocol/v2/qbft/instance/timeout.go
```diff
@@ -20,8 +20,10 @@ func (i *Instance) UponRoundTimeout(ctx context.Context, logger *zap.Logger) err
 		return traces.Errorf(span, "instance stopped processing timeouts")
 	}
 
-	newRound := i.State.Round + 1
-	logger.Debug("⌛ round timed out", fields.Round(newRound))
+	prevRound := i.State.Round
+	newRound := prevRound + 1
+
+	logger.Debug("⌛ round timed out", fields.Round(prevRound))
```
iurii-ssv: This incorrectly reports future round, for example here we are timing out `round 1` while in the log it says `"round": 2` (as if we are timing out `round 2`):
```
{
  "level": "DEBUG",
  "time": "2025-08-18T14:29:10.001290Z",
  "msg": "⌛ round timed out",
  "pubkey": "a171eed8e86355e977eb736e7c1738a81c419cc815ddb8849aea80a981fa14753fdfe53a96496f42bb625ce841ae15b6",
  "msg_type": "event",
  "msg_id": "0000311401000000a171eed8e86355e977eb736e7c1738a81c419cc815ddb8849aea80a981fa14753fdfe53a96496f42bb625ce841ae15b6",
  "role": "AGGREGATOR_RUNNER",
  "slot": 1109495,
  "duty_id": "AGGREGATOR-e34671-s1109495-v1158258",
  "round": 2
}
```

protocol/v2/message/msg.go
```diff
@@ -8,8 +8,6 @@ import (
 )
 
 const (
-	// SSVSyncMsgType extends spec msg type
-	SSVSyncMsgType spectypes.MsgType = 100
 	// SSVEventMsgType extends spec msg type
 	SSVEventMsgType spectypes.MsgType = 200
 )
```
iurii-ssv, line 13: Removing this message type since it's no longer relevant/used

protocol/v2/ssv/validator/duty_executor.go
```diff
@@ -39,23 +38,14 @@ func (v *Validator) ExecuteDuty(ctx context.Context, duty *spectypes.ValidatorDu
...
 	if err != nil {
 		return traces.Errorf(span, "failed to get execute duty data: %w", err)
 	}
-
-	span.SetAttributes(
-		observability.BeaconSlotAttribute(executeDutyData.Duty.Slot),
-		observability.RunnerRoleAttribute(executeDutyData.Duty.RunnerRole()),
-	)
-	logger = logger.With(fields.Slot(executeDutyData.Duty.DutySlot()), fields.Role(executeDutyData.Duty.RunnerRole()))
+	duty := executeDutyData.Duty
```
iurii-ssv: These fields are added already by that point (with the way I've implemented it now in this PR)

protocol/v2/ssv/validator/validator.go
```diff
@@ -117,16 +113,6 @@ func (v *Validator) StartDuty(ctx context.Context, logger *zap.Logger, duty spec
...
 
-	// Log with duty ID.
-	baseRunner := dutyRunner.GetBaseRunner()
-	v.dutyIDs.Set(spectypes.MapDutyToRunnerRole(vDuty.Type), fields.FormatDutyID(baseRunner.NetworkConfig.EstimatedEpochAtSlot(vDuty.Slot), vDuty.Slot, vDuty.Type, vDuty.ValidatorIndex))
-	logger = v.withDutyID(logger, spectypes.MapDutyToRunnerRole(vDuty.Type))
-
-	// Log with height.
-	if baseRunner.QBFTController != nil {
-		logger = logger.With(fields.Height(baseRunner.QBFTController.Height))
-	}
```
iurii-ssv, line 128: This `height` (`qbft_height`) field is already added for consensus messages in the handler itself (`Validator.ProcessMessage` func) and for any other logs it doesn't make sense to have it, so removing.

protocol/v2/ssv/validator/validator.go
```diff
@@ -48,9 +48,6 @@ type Validator struct {
 
 	Queues map[spectypes.RunnerRole]QueueContainer
 
-	// dutyIDs is a map for logging a unique ID for a given duty
-	dutyIDs *hashmap.Map[spectypes.RunnerRole, string]
```
iurii-ssv, line 52: Doing it like that ("caching" `duty_id`) introduces certain issues that are hard to deal with - we can't just use this map whenever we want to get a `duty_id` out of it (depending on where we are in the execution flow the `duty_id` might not exist yet, or it might point to some prior value that's been stored in there by a previous message ... that gets complex and brittle very fast).

So, now that https://github.com/ssvlabs/ssv/issues/2452 is resolved in this PR we can simply reconstruct the `duty_id` (in the message handler - `Validator.ProcessMessage` / `Committee.ProcessMessage` funcs - the "highest entry-point") for every message (both p2p messages, and events we generate to kick-off duties / timeout-qbfts) and simply attach it to the `logger` that will be used to log everything after that point.

protocol/v2/ssv/validator/committee.go
```diff
@@ -137,13 +129,8 @@ func (c *Committee) prepareDutyAndRunner(ctx context.Context, logger *zap.Logger
...
-	_, span := tracer.Start(ctx,
-		observability.InstrumentName(observabilityNamespace, "prepare_duty_runner"),
-		trace.WithAttributes(
-			observability.RunnerRoleAttribute(duty.RunnerRole()),
-			observability.DutyCountAttribute(len(duty.ValidatorDuties)),
-			observability.BeaconSlotAttribute(duty.Slot)),
-	)
+
+	_, span := tracer.Start(ctx, observability.InstrumentName(observabilityNamespace, "prepare_duty_runner"))
 	defer span.End()
```
iurii-ssv: @oleg-ssvlabs just to make sure I got that right, if parent-`ctx` already has these `attribute.KeyValue` added we don't need to add it here as well (for every child-`ctx`), correct ? That's why I'm removing it in a bunch of places in this PR - because I'm adding all these at the highest level possible just once (in the message handler - `Validator.ProcessMessage` / `Committee.ProcessMessage` funcs),

this is how `logger.With` works, and I assume `trace` should work similar (if it doesn't - I think we need to find a better solution than attaching all these fields manually),

~~**TODO:** I'll need to finish this for all `executeDuty` impls (haven't just yet because I want to confirm this first)~~

protocol/v2/ssv/validator/committee.go
oleg-ssvlabs: Yes, that's spot on.  

In general, we could consider removing these links (together with the message `TraceContext`), because links are typically used for connecting two separate traces - and we had that problem before switching to deterministic-duty-id-based trace IDs.  

It doesn’t cause any harm, but I’m not confident it provides real value anymore. The only reason I didn’t remove them is that it’s a common pattern to bind producer and consumer with trace links and potentially in some other cases like one message - one producer - multiple consumer, etc. 

So all in all, we can keep these links and, yes, if we keep them, it should implemented here.

protocol/v2/ssv/validator/committee.go
oleg-ssvlabs: I would say it depends.  

It affects the search patterns. TraceQL searches for matches within a single span. With this change, it would not be possible to say: *"give me a span named `prepare_duty_runner` with attribute `runner.role == committee_runner`"*. This may sound suboptimal, but in general our search patterns are different and usually look more like: *"give me all traces where there is a span (ignoring the name) that contains `runner.role == committee_runner`"*.  

**So child spans do not inherit attributes from the parent.** However, having all needed attributes(duty related) only within the root span of a trace seems to be a good pattern. But, honestly, I have some doubts.

@zktaiga wdyt about this? How this would affect the Grafana drilldown? We won't be able to see "span `Foo` had X number of failures and it typically contained `Bar` attribute". Instead we would need to navigate to the Trace of the span Foo and find Duty related attributes in the root span.

observability/attributes.go
```diff
@@ -112,8 +112,6 @@ func ValidatorMsgTypeAttribute(msgType types.MsgType) attribute.KeyValue {
 		return attribute.String(attrKey, "SSVConsensusMsgType")
 	case types.SSVPartialSignatureMsgType:
 		return attribute.String(attrKey, "SSVPartialSignatureMsgType")
-	case message.SSVSyncMsgType:
```
oleg-ssvlabs, line 115: just out of curiosity - what was this message?

protocol/v2/ssv/validator/committee.go
iurii-ssv: For now I've reverted the changes related to "removing presumably duplicated attributes" (in https://github.com/ssvlabs/ssv/pull/2453/commits/cc14694656fc8c090079f496fc2c5988910bac77) since it's probably better to do across the entire code-base at once (and in a separate PR).

But my point still stands (for future reference, out of the scope of this PR): it doesn't make sense to add **duty-wide** attributes (like `slot`, `runner_role`, etc.) for child-spans because not every child span has access to these, so some will be missing these ... it should be managed all in one place (as we do it for `logger` using `logger.With`). So I think this is something that can/should be resolved on UI/Query side.

observability/log/fields/fields.go
```diff
@@ -55,17 +48,15 @@ const (
...
 	FieldENR                 = "enr"
 	FieldEpoch               = "epoch"
-	FieldErrors              = "errors"
 	FieldEvent               = "event"
-	FieldEventID             = "event_id"
 	FieldFeeRecipient        = "fee_recipient"
 	FieldFromBlock           = "from_block"
-	FieldHeight              = "height"
+	FieldQBFTHeight          = "qbft_height"
+	FieldQBFTRound           = "qbft_round"
```
oleg-ssvlabs, line 55: [Pulse](https://github.com/ssvlabs/ssv-pulse) reads this field(analyzer), so it would be good to adapt it to this change (ideally making it possible to read both `round` and `qbft_round` fields)

protocol/v2/qbft/instance/timeout.go
```diff
@@ -20,8 +20,10 @@ func (i *Instance) UponRoundTimeout(ctx context.Context, logger *zap.Logger) err
 		return traces.Errorf(span, "instance stopped processing timeouts")
 	}
 
-	newRound := i.State.Round + 1
-	logger.Debug("⌛ round timed out", fields.Round(newRound))
+	prevRound := i.State.Round
+	newRound := prevRound + 1
+
+	logger.Debug("⌛ round timed out", fields.QBFTRound(prevRound))
```
oleg-ssvlabs, line 26: Good change, it makes more sense to specify the previous round instead of new

protocol/v2/ssv/runner/committee.go
```diff
@@ -94,11 +94,7 @@ func NewCommitteeRunner(
 }
 
 func (cr *CommitteeRunner) StartNewDuty(ctx context.Context, logger *zap.Logger, duty spectypes.Duty, quorum uint64) error {
-	ctx, span := tracer.Start(ctx,
-		observability.InstrumentName(observabilityNamespace, "runner.start_committee_duty"),
-		trace.WithAttributes(
-			observability.RunnerRoleAttribute(duty.RunnerRole()),
-			observability.BeaconSlotAttribute(duty.DutySlot())))
+	ctx, span := tracer.Start(ctx, observability.InstrumentName(observabilityNamespace, "runner.start_committee_duty"))
```
oleg-ssvlabs: so you decided to remove span attributes where you are confident the parent spans already contain them?

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,38 +227,52 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
 
-	logger := c.logger.With(zap.String("message_type", message.MsgTypeToString(msgType)))
-
-	// Validate message
+	// Validate message (+ verify SignedSSVMessage's signature)
 	if msgType != message.SSVEventMsgType {
-		span.AddEvent("validating message and signature")
 		if err := msg.SignedSSVMessage.Validate(); err != nil {
-			return traces.Errorf(span, "invalid SignedSSVMessage: %w", err)
+			return fmt.Errorf("invalid SignedSSVMessage: %w", err)
```
oleg-ssvlabs, line 236: was there a specific reason you moved where span starts? I mean this error won't be visible in the span, also, I think the caller method does not have traces implemented so these errors won't be visible in traces at all

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,38 +227,52 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
-			return traces.Errorf(span, "Message invalid: %w", err)
+			// TODO - we should improve this error message as is suggested by the commented-out code here
+			// (and also remove nolint annotation), currently we cannot do it due to spec-tests expecting
+			// this exact format we are stuck with.
+			//return fmt.Errorf("SSVMessage invalid: %w", err)
+			return fmt.Errorf("Message invalid: %w", err) //nolint:staticcheck
 		}
 	}
 
+	slot, err := msg.Slot()
```
oleg-ssvlabs, line 250: From what I understand, this is a change in behavior - returning an error if the slot couldn’t be fetched.  
Although this should not really happen, unless the unmarshalling that happens within `msg.Slot()` fails?

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,38 +227,52 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
+	logger := c.logger.
+		With(fields.MessageType(msgType)).
+		With(fields.MessageID(msgID)).
+		With(fields.RunnerRole(msgID.GetRoleType())).
+		With(fields.Slot(slot)).
+		With(fields.DutyID(dutyID))
+
+	ctx, span := tracer.Start(traces.Context(ctx, dutyID),
+		observability.InstrumentName(observabilityNamespace, "process_committee_message"),
+		trace.WithAttributes(
```
oleg-ssvlabs, line 265: I think the majority of these attributes are not really needed here, since this span will be linked to an existing trace via a deterministic trace ID. However, I believe it is still better to keep them for now, as their relevance may depend on the message type or so.

protocol/v2/ssv/validator/msgqueue_consumer.go
oleg-ssvlabs, line 1: unrelated to the PR - could we maybe rename this file so it aligns with the `committee_queue.go`? Maybe `validator_queue.go` ?

protocol/v2/ssv/queue/messages.go
```diff
@@ -32,19 +33,27 @@ func (d *SSVMessage) DecodedSSVMessage() {}
...
+	case *ssvtypes.EventMsg:
+		switch m.Type {
+		case ssvtypes.Timeout:
 			data, err := m.GetTimeoutData()
 			if err != nil {
-				return 0, ErrUnknownMessageType // TODO alan: other error
+				return 0, fmt.Errorf("get Timeout data: %w", err)
 			}
 			return phase0.Slot(data.Height), nil
+		case ssvtypes.ExecuteDuty:
```
oleg-ssvlabs, line 48: I assume this is the main change that allowed us to remove [this](https://github.com/ssvlabs/ssv/pull/2453/files#r2284894843)?  Pretty cool you figured it out

observability/attributes.go
iurii-ssv: TBH I don't know, but not it's no longer used anywhere (hence removing)

protocol/v2/ssv/runner/committee.go
iurii-ssv: No, this seems to be the only place I "forgot to revert", thanks for spotting this - should be reverted now in https://github.com/ssvlabs/ssv/pull/2453/commits/52e71f066e5a85912ef11c40811c3f9653f72595

protocol/v2/ssv/validator/committee.go
iurii-ssv: Yes, the reason is - failing spec-tests that expect `invalid SignedSSVMessage` (kind of, but not really),

the idea now is to fetch/reconstruct `slot` from the `msg` as early as possible, this is done by the following code (added in this PR), in a similar way for both `Validator` and `Commitee`:
```go
	slot, err := msg.Slot()
	if err != nil {
		return fmt.Errorf("couldn't get message slot: %w", err)
	}
```
and so the `slot` is added to both `logger` and `trace` as early as possible,

the problem is - there are some spec-tests that have invalid-messages, so when you try to fetch a slot from invalid (or nil) message **before** validating it you get `couldn't get message slot: ... ` kind of error which is different from spec-tests expectations that we cannot simply update (although I'm looking into if/how we can - we'll have to to that for https://github.com/ssvlabs/ssv/pull/2445 PR, for example). But it also makes sense to fail with `invalid SignedSSVMessage` error and not `couldn't get message slot: ...` in case of invalid/nil message - hence, failing spec-tests do make sense in this case (and we need to make them work rather than adjusting spec-tests expectations).

So what I'm really aiming at here is to have `logger` / `trace` with all the data on it (including `slot` that we might or might not be able to fetch).

Essentially it's a trade-off between:
- having incomplete `logger` / `trace` (without `slot` in it, in this case)
- or having `logger` / `trace` with all the data on it, but possibly missing some execution-parts entirely (which is less/simpler code overall)

I do agree that eventually we'd want to log/trace as much as possible and not take shortcuts like this one (we'll be able to get the best out of both soon-ish), but for now I think
- the number 1 priority is to **clean up the code involved so we can actually improve upon it in the follow-up PRs** (even though at the moment it means we have to give some things up)
- otherwise we'll just drown in complexity, TBH modifying the code in this area without breaking anything is already taking too long ... just because of how tangled it is

That is to say - I do plan to improve the way we handle these logs/traces in the near future (as a part of https://github.com/ssvlabs/ssv/issues/1828), just want to do it 1 step at a time.

> I think the caller method does not have traces implemented so these errors won't be visible in traces at all

So, I think the caller/parent should take care of it. I would say this is out of scope of this PR (or at least not super high on "priority list").

protocol/v2/ssv/validator/committee.go
iurii-ssv: True, this could potentially show up as a new kind of "error" message-processing is now prone to - which is why I wanted to implement https://github.com/ssvlabs/ssv/issues/2452 as a "separate feature" ... but eventually decided to go for it in this PR simply because of how ad-hoc and complex the other options are,

I am monitoring for the relevant error-logs on **stage** and hasn't seen any in the past ~24h ... so we should be fine - the cluster is working as expected and I have no reasons to believe this "change in behavior has any practical effect".

protocol/v2/ssv/validator/committee.go
iurii-ssv: I've added these just to be consistent with what we do for `logger`, we can clean up `trace`-related attributes in a separate/follow-up PRs later on (as we discussed already)

protocol/v2/ssv/validator/msgqueue_consumer.go
iurii-ssv: Yes that's a good idea, hope it doesn't screw up the PR diff too much - applied in https://github.com/ssvlabs/ssv/pull/2453/commits/765bf98c863b5dc3796b3aafc5598d8f1045c8a6

protocol/v2/ssv/queue/messages.go
iurii-ssv: Yeah, Events must have a `slot` on them for us to be able to reconstruct `duty_id` for every message, and like I mentioned in a [comment above](https://github.com/ssvlabs/ssv/pull/2453#discussion_r2287499967) - based on what I see in **stage** logs this does seem to work flawlessly

observability/log/fields/fields.go
iurii-ssv: I don't see it reading `height`, but I see it reads `round` indeed - so I've created a [PR here to resolve it](https://github.com/ssvlabs/ssv-pulse/pull/112)

I guess you are right - it's better to fix it now so we don't forget, however it's probably inevitable that we'll be missing/skipping over some of these cases in practice ... and will have to address it at some future point anyway for those missed/skipped ones

oleg-ssvlabs (dismissed):

oleg-ssvlabs (approved):

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,38 +227,52 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
+			return fmt.Errorf("SignedSSVMessage has an invalid signature: %w", err)
 		}
-
 		if err := c.validateMessage(msg.SignedSSVMessage.SSVMessage); err != nil {
-			return traces.Errorf(span, "Message invalid: %w", err)
+			// TODO - we should improve this error message as is suggested by the commented-out code here
+			// (and also remove nolint annotation), currently we cannot do it due to spec-tests expecting
+			// this exact format we are stuck with.
+			//return fmt.Errorf("SSVMessage invalid: %w", err)
+			return fmt.Errorf("Message invalid: %w", err) //nolint:staticcheck
```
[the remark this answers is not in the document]
iurii-ssv: Well, I did document it, so I think this is resolved ?

protocol/v2/ssv/validator/committee.go
```diff
@@ -230,38 +227,52 @@ func (c *Committee) prepareDuty(logger *zap.Logger, duty *spectypes.CommitteeDut
...
 		if err := spectypes.Verify(msg.SignedSSVMessage, c.CommitteeMember.Committee); err != nil {
-			return traces.Errorf(span, "SignedSSVMessage has an invalid signature: %w", err)
+			return fmt.Errorf("SignedSSVMessage has an invalid signature: %w", err)
 		}
-
 		if err := c.validateMessage(msg.SignedSSVMessage.SSVMessage); err != nil {
-			return traces.Errorf(span, "Message invalid: %w", err)
+			// TODO - we should improve this error message as is suggested by the commented-out code here
+			// (and also remove nolint annotation), currently we cannot do it due to spec-tests expecting
+			// this exact format we are stuck with.
```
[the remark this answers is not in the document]
iurii-ssv: This is a duplicate of the [comment from below](https://github.com/ssvlabs/ssv/pull/2453#discussion_r2287874444), and we use `TODO` (instead of `TECHDEBT`) to denote tech-debt

PR #2457, opened 2025-08-21
fix: establish single source of truth for fee recipients

olegshmuelov:
## Summary
  Establishes shares storage as the single source of truth for fee recipients by fixing the architecture where fee recipients were incorrectly persisted in shares but never used. All fee recipient
  reads now go through shares storage, which internally manages the data using recipients storage.

  ## Key Changes

  ### Architecture Fix
  - **Single Interface**: All components must read fee recipients through shares storage (Validator Store)
  - **Proper Wiring**: Recipients storage notifies shares storage of updates via `SharesUpdater` interface
  - **Runtime Enrichment**: Shares automatically fetch current fee recipients when loaded from DB

  ## Why This Matters
  Previously:
  - Fee recipients stored in shares could become stale
  - Validator controller had to manually set fee recipients
  - Multiple components could access recipients storage directly
  - Risk of using outdated fee recipient data

  Now:
  - ✅ Single source of truth enforced architecturally
  - ✅ Automatic enrichment - no manual intervention needed
  - ✅ Impossible to use stale data
  - ✅ Clean separation - recipients storage is implementation detail

### DIFF
--- a/migrations/migration_6_share_exit_epoch.go
+++ b/migrations/migration_6_share_exit_epoch.go
@@ -100,13 +100,12 @@ func mapShare(share *migration_6_OldStorageShare) *storage.Share {
 
 	domainShare := &types.SSVShare{
 		Share: spectypes.Share{
-			ValidatorPubKey:     validatorPubKey,
-			SharePubKey:         share.SharePubKey,
-			Committee:           committee,
-			DomainType:          share.DomainType,
-			FeeRecipientAddress: share.FeeRecipientAddress,
-			Graffiti:            share.Graffiti,
-			ValidatorIndex:      phase0.ValidatorIndex(share.ValidatorIndex),
+			ValidatorPubKey: validatorPubKey,
+			SharePubKey:     share.SharePubKey,
+			Committee:       committee,
+			DomainType:      share.DomainType,
+			Graffiti:        share.Graffiti,
+			ValidatorIndex:  phase0.ValidatorIndex(share.ValidatorIndex),
 		},
 		OwnerAddress:    share.OwnerAddress,
 		Liquidated:      share.Liquidated,
--- a/network/topics/controller_test.go
+++ b/network/topics/controller_test.go
@@ -12,7 +12,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/libp2p/go-libp2p"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
@@ -320,6 +322,17 @@ func committeeTopic(cidHex string) string {
 	return commons.CommitteeTopicID(spectypes.CommitteeID(cid))[0]
 }
 
+// noOpRecipientReader is a no-op implementation for tests
+type noOpRecipientReader struct{}
+
+func (n *noOpRecipientReader) GetRecipientData(_ basedb.Reader, _ common.Address) (*registrystorage.RecipientData, bool, error) {
+	return nil, false, nil
+}
+
+func (n *noOpRecipientReader) GetRecipientDataMany(_ basedb.Reader, _ []common.Address) (map[common.Address]bellatrix.ExecutionAddress, error) {
+	return make(map[common.Address]bellatrix.ExecutionAddress), nil
+}
+
 type P struct {
 	host host.Host
 	ps   *pubsub.PubSub
@@ -409,7 +422,7 @@ func newPeer(ctx context.Context, logger *zap.Logger, t *testing.T, msgValidator
 	db, err := kv.NewInMemory(logger, basedb.Options{})
 	require.NoError(t, err)
 
-	_, validatorStore, err := registrystorage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, []byte("test"))
+	_, validatorStore, err := registrystorage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, &noOpRecipientReader{}, []byte("test"))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/operator/dutytracer/eviction_test.go
+++ b/operator/dutytracer/eviction_test.go
@@ -39,7 +39,7 @@ func TestEviction(t *testing.T) {
 	}
 
 	dutyStore := store.New(db)
-	_, vstore, _ := storage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, nil)
+	_, vstore, _ := storage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, &noOpRecipientReader{}, nil)
 
 	collector := New(t.Context(), zap.NewNop(), vstore, mockDomainDataProvider{}, dutyStore, networkconfig.TestNetwork.Beacon)
 
--- a/operator/fee_recipient/controller.go
+++ b/operator/fee_recipient/controller.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
@@ -30,7 +29,6 @@ type ControllerOptions struct {
 	BeaconClient       beaconprotocol.BeaconNode
 	BeaconConfig       *networkconfig.Beacon
 	ShareStorage       storage.Shares
-	RecipientStorage   storage.Recipients
 	SlotTickerProvider slotticker.Provider
 	OperatorDataStore  operatordatastore.OperatorDataStore
 }
@@ -42,7 +40,6 @@ type recipientController struct {
 	beaconClient       beaconprotocol.BeaconNode
 	beaconConfig       *networkconfig.Beacon
 	shareStorage       storage.Shares
-	recipientStorage   storage.Recipients
 	slotTickerProvider slotticker.Provider
 	operatorDataStore  operatordatastore.OperatorDataStore
 }
@@ -54,7 +51,6 @@ func NewController(logger *zap.Logger, opts *ControllerOptions) *recipientContro
 		beaconClient:       opts.BeaconClient,
 		beaconConfig:       opts.BeaconConfig,
 		shareStorage:       opts.ShareStorage,
-		recipientStorage:   opts.RecipientStorage,
 		slotTickerProvider: opts.SlotTickerProvider,
 		operatorDataStore:  opts.OperatorDataStore,
 	}
@@ -134,30 +130,11 @@ func (rc *recipientController) submit(ctx context.Context, shares []*types.SSVSh
 }
 
 func (rc *recipientController) toProposalPreparation(shares []*types.SSVShare) (map[phase0.ValidatorIndex]bellatrix.ExecutionAddress, error) {
-	// build unique owners
-	keys := make(map[common.Address]bool)
-	var uniq []common.Address
-	for _, entry := range shares {
-		if _, value := keys[entry.OwnerAddress]; !value {
-			keys[entry.OwnerAddress] = true
-			uniq = append(uniq, entry.OwnerAddress)
-		}
-	}
-
-	// get recipients
-	rds, err := rc.recipientStorage.GetRecipientDataMany(nil, uniq)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not get recipients data")
-	}
-
-	// build proposal preparation
+	// Build proposal preparation directly from shares which already have fee recipients
 	m := make(map[phase0.ValidatorIndex]bellatrix.ExecutionAddress)
 	for _, share := range shares {
-		feeRecipient, found := rds[share.OwnerAddress]
-		if !found {
-			copy(feeRecipient[:], share.OwnerAddress.Bytes())
-		}
-		m[share.ValidatorIndex] = feeRecipient
+		// Use the fee recipient directly from the share (already enriched)
+		m[share.ValidatorIndex] = share.FeeRecipientAddress
 	}
 
 	return m, nil
--- a/operator/fee_recipient/controller_test.go
+++ b/operator/fee_recipient/controller_test.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"testing"
-	"time"
 
 	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
@@ -20,8 +18,6 @@ import (
 	"github.com/ssvlabs/ssv/networkconfig"
 	"github.com/ssvlabs/ssv/observability/log"
 	operatordatastore "github.com/ssvlabs/ssv/operator/datastore"
-	"github.com/ssvlabs/ssv/operator/slotticker"
-	"github.com/ssvlabs/ssv/operator/slotticker/mocks"
 	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 	"github.com/ssvlabs/ssv/protocol/v2/types"
 	registrystorage "github.com/ssvlabs/ssv/registry/storage"
@@ -40,109 +36,220 @@ func TestSubmitProposal(t *testing.T) {
 
 	operatorDataStore := operatordatastore.New(operatorData)
 
-	db, shareStorage, recipientStorage := createStorage(t)
-	defer db.Close()
+	db, shareStorage, recipientStorage := createStorageWithRecipients(t)
+	defer func() { require.NoError(t, db.Close()) }()
 
 	beaconConfig := networkconfig.TestNetwork.Beacon
 	populateStorage(t, shareStorage, operatorData)
 
 	frCtrl := NewController(logger, &ControllerOptions{
-		Ctx:               context.TODO(),
+		Ctx:               t.Context(),
 		BeaconConfig:      beaconConfig,
 		ShareStorage:      shareStorage,
-		RecipientStorage:  recipientStorage,
 		OperatorDataStore: operatorDataStore,
 	})
 
-	t.Run("submit first time or halfway through epoch", func(t *testing.T) {
-		numberOfRequests := 4
-		var wg sync.WaitGroup
-		wg.Add(numberOfRequests) // Set up the wait group before starting goroutines
+	t.Run("custom fee recipients from storage", func(t *testing.T) {
+		// Set custom fee recipients for some validators
+		customRecipient1 := bellatrix.ExecutionAddress{0x11, 0x22, 0x33, 0x44, 0x55}
+		customRecipient2 := bellatrix.ExecutionAddress{0xaa, 0xbb, 0xcc, 0xdd, 0xee}
+
+		owner1 := common.HexToAddress("0x0000000000000000000000000000000000000000")
+		owner2 := common.HexToAddress("0x0000000000000000000000000000000000000001")
+
+		_, err := recipientStorage.SaveRecipientData(nil, &registrystorage.RecipientData{
+			Owner:        owner1,
+			FeeRecipient: customRecipient1,
+		})
+		require.NoError(t, err)
+
+		_, err = recipientStorage.SaveRecipientData(nil, &registrystorage.RecipientData{
+			Owner:        owner2,
+			FeeRecipient: customRecipient2,
+		})
+		require.NoError(t, err)
+
+		var capturedRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress
 
 		client := beacon.NewMockBeaconNode(ctrl)
-		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
-			wg.Done()
-			return nil
-		}).Times(numberOfRequests)
-
-		ticker := mocks.NewMockSlotTicker(ctrl)
-		mockTimeChan := make(chan time.Time)
-		mockSlotChan := make(chan phase0.Slot)
-		ticker.EXPECT().Next().Return(mockTimeChan).AnyTimes()
-		ticker.EXPECT().Slot().DoAndReturn(func() phase0.Slot {
-			return <-mockSlotChan
-		}).AnyTimes()
+		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(
+			func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
+				capturedRecipients = feeRecipients
+				return nil
+			}).Times(1)
 
 		frCtrl.beaconClient = client
-		frCtrl.slotTickerProvider = func() slotticker.SlotTicker {
-			return ticker
+
+		err = frCtrl.prepareAndSubmit(t.Context())
+		require.NoError(t, err)
+
+		// Verify custom recipients are used for validators with index 0 and 1
+		require.Equal(t, customRecipient1, capturedRecipients[0])
+		require.Equal(t, customRecipient2, capturedRecipients[1])
+
+		// Verify owner address is used as default for others (e.g., index 2)
+		owner2Addr := common.HexToAddress("0x0000000000000000000000000000000000000002")
+		var expectedDefault bellatrix.ExecutionAddress
+		copy(expectedDefault[:], owner2Addr.Bytes())
+		require.Equal(t, expectedDefault, capturedRecipients[2])
+	})
+
+	t.Run("owner address as default fee recipient", func(t *testing.T) {
+		// Create fresh storage without custom recipients
+		db2, shareStorage2, _ := createStorageWithRecipients(t)
+		defer func() { require.NoError(t, db2.Close()) }()
+		populateStorage(t, shareStorage2, operatorData)
+
+		frCtrl2 := NewController(logger, &ControllerOptions{
+			Ctx:               t.Context(),
+			BeaconConfig:      beaconConfig,
+			ShareStorage:      shareStorage2,
+			OperatorDataStore: operatorDataStore,
+		})
+
+		var capturedRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress
+
+		client := beacon.NewMockBeaconNode(ctrl)
+		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(
+			func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
+				capturedRecipients = feeRecipients
+				return nil
+			}).Times(1)
+
+		frCtrl2.beaconClient = client
+
+		err := frCtrl2.prepareAndSubmit(t.Context())
+		require.NoError(t, err)
+
+		// Verify all validators use their owner address as fee recipient
+		for i := 0; i < 10; i++ { // Check first 10 validators
+			ownerAddr := common.HexToAddress(fmt.Sprintf("0x%040x", i))
+			var expectedRecipient bellatrix.ExecutionAddress
+			copy(expectedRecipient[:], ownerAddr.Bytes())
+			require.Equal(t, expectedRecipient, capturedRecipients[phase0.ValidatorIndex(i)],
+				"Validator %d should use owner address as default fee recipient", i)
+		}
+	})
+
+	t.Run("correct validator index to fee recipient mapping", func(t *testing.T) {
+		var capturedRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress
+
+		client := beacon.NewMockBeaconNode(ctrl)
+		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(
+			func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
+				capturedRecipients = feeRecipients
+				return nil
+			}).Times(1)
+
+		frCtrl.beaconClient = client
+
+		err := frCtrl.prepareAndSubmit(t.Context())
+		require.NoError(t, err)
+
+		// Verify mapping has correct number of entries (100 committee validators)
+		require.Len(t, capturedRecipients, 100)
+
+		// Verify each validator index maps to correct recipient
+		for i := 0; i < 100; i++ {
+			idx := phase0.ValidatorIndex(i)
+			_, exists := capturedRecipients[idx]
+			require.True(t, exists, "Validator index %d should be in the mapping", i)
 		}
 
-		go frCtrl.Start(t.Context())
+		// Verify non-committee validator (index 2000) is not included
+		_, exists := capturedRecipients[2000]
+		require.False(t, exists, "Non-committee validator should not be included")
+	})
 
-		slots := []phase0.Slot{
-			1,  // first time
-			2,  // should not call submit
-			20, // should not call submit
-			phase0.Slot(beaconConfig.SlotsPerEpoch / 2), // halfway through epoch
-			63, // should not call submit
+	t.Run("batch processing edge cases", func(t *testing.T) {
+		// Test with exactly 500 shares (one batch)
+		db2, shareStorage2, _ := createStorageWithRecipients(t)
+		defer func() { require.NoError(t, db2.Close()) }()
+
+		// Create exactly 500 shares
+		for i := 0; i < 500; i++ {
+			ownerAddr := common.HexToAddress(fmt.Sprintf("0x%040x", i))
+			share := &types.SSVShare{
+				Share: spectypes.Share{
+					ValidatorPubKey: spectypes.ValidatorPK([]byte(fmt.Sprintf("pk%046d", i))),
+					SharePubKey:     []byte(fmt.Sprintf("pk%046d", i)),
+					ValidatorIndex:  phase0.ValidatorIndex(i),
+					Committee:       []*spectypes.ShareMember{{Signer: operatorData.ID}},
+				},
+				Status:       eth2apiv1.ValidatorStateActiveOngoing,
+				OwnerAddress: ownerAddr,
+				Liquidated:   false,
+			}
+			require.NoError(t, shareStorage2.Save(nil, share))
 		}
 
-		for _, s := range slots {
-			mockTimeChan <- time.Now()
-			mockSlotChan <- s
-			time.Sleep(time.Millisecond * 500)
-		}
+		frCtrl2 := NewController(logger, &ControllerOptions{
+			Ctx:               t.Context(),
+			BeaconConfig:      beaconConfig,
+			ShareStorage:      shareStorage2,
+			OperatorDataStore: operatorDataStore,
+		})
 
-		wg.Wait()
+		batchSize := 0
 
-		close(mockTimeChan) // Close the channel after test
-		close(mockSlotChan)
+		client := beacon.NewMockBeaconNode(ctrl)
+		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(
+			func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
+				batchSize = len(feeRecipients)
+				return nil
+			}).Times(1)
+
+		frCtrl2.beaconClient = client
+
+		err := frCtrl2.prepareAndSubmit(t.Context())
+		require.NoError(t, err)
+
+		// Should have exactly 1 batch of 500
+		require.Equal(t, 500, batchSize)
 	})
 
 	t.Run("error handling", func(t *testing.T) {
-		var wg sync.WaitGroup
+		// Test that errors are handled gracefully
+		db3, shareStorage3, _ := createStorageWithRecipients(t)
+		defer func() { require.NoError(t, db3.Close()) }()
+		populateStorage(t, shareStorage3, operatorData)
+
+		frCtrl3 := NewController(logger, &ControllerOptions{
+			Ctx:               t.Context(),
+			BeaconConfig:      beaconConfig,
+			ShareStorage:      shareStorage3,
+			OperatorDataStore: operatorDataStore,
+		})
+
 		client := beacon.NewMockBeaconNode(ctrl)
-		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
-			wg.Done()
-			return errors.New("failed to submit")
-		}).MinTimes(2).MaxTimes(2)
+		client.EXPECT().SubmitProposalPreparation(gomock.Any(), gomock.Any()).Return(errors.New("failed to submit")).Times(1)
 
-		ticker := mocks.NewMockSlotTicker(ctrl)
-		mockTimeChan := make(chan time.Time, 1)
-		ticker.EXPECT().Next().Return(mockTimeChan).AnyTimes()
-		ticker.EXPECT().Slot().Return(phase0.Slot(100)).AnyTimes()
+		frCtrl3.beaconClient = client
 
-		frCtrl.beaconClient = client
-		frCtrl.slotTickerProvider = func() slotticker.SlotTicker {
-			return ticker
-		}
-
-		go frCtrl.Start(t.Context())
-		mockTimeChan <- time.Now()
-		wg.Add(2)
-		wg.Wait()
-		close(mockTimeChan)
+		// Should handle error gracefully without panic
+		err := frCtrl3.prepareAndSubmit(t.Context())
+		require.NoError(t, err) // prepareAndSubmit logs errors but returns nil
 	})
 }
 
-func createStorage(t *testing.T) (basedb.Database, registrystorage.Shares, registrystorage.Recipients) {
+func createStorageWithRecipients(t *testing.T) (basedb.Database, registrystorage.Shares, registrystorage.Recipients) {
 	logger := log.TestLogger(t)
 	db, err := kv.NewInMemory(logger, basedb.Options{})
 	require.NoError(t, err)
 
-	shareStorage, _, err := registrystorage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, []byte("test"))
+	recipientStorage, setSharesUpdater := registrystorage.NewRecipientsStorage(logger, db, []byte("test"))
+	shareStorage, _, err := registrystorage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, recipientStorage, []byte("test"))
 	if err != nil {
 		t.Fatal(err)
 	}
-	return db, shareStorage, registrystorage.NewRecipientsStorage(logger, db, []byte("test"))
+	setSharesUpdater(shareStorage)
+	return db, shareStorage, recipientStorage
 }
 
 func populateStorage(t *testing.T, storage registrystorage.Shares, operatorData *registrystorage.OperatorData) {
 	createShare := func(index int, operatorID spectypes.OperatorID) *types.SSVShare {
-		ownerAddr := fmt.Sprintf("%d", index)
-		ownerAddrByte := [20]byte{}
-		copy(ownerAddrByte[:], ownerAddr)
+		// Create owner address consistently with fmt.Sprintf("0x%040x", index)
+		ownerAddr := common.HexToAddress(fmt.Sprintf("0x%040x", index))
 
 		return &types.SSVShare{
 			Share: spectypes.Share{ValidatorPubKey: spectypes.ValidatorPK([]byte(fmt.Sprintf("pk%046d", index))),
@@ -151,12 +258,12 @@ func populateStorage(t *testing.T, storage registrystorage.Shares, operatorData 
 				Committee:      []*spectypes.ShareMember{{Signer: operatorID}},
 			},
 			Status:       eth2apiv1.ValidatorStateActiveOngoing,
-			OwnerAddress: common.BytesToAddress(ownerAddrByte[:]),
+			OwnerAddress: ownerAddr,
 			Liquidated:   false,
 		}
 	}
 
-	for i := 0; i < 1000; i++ {
+	for i := 0; i < 100; i++ {
 		require.NoError(t, storage.Save(nil, createShare(i, operatorData.ID)))
 	}
 
@@ -164,5 +271,5 @@ func populateStorage(t *testing.T, storage registrystorage.Shares, operatorData 
 	require.NoError(t, storage.Save(nil, createShare(2000, spectypes.OperatorID(1))))
 
 	all := storage.List(nil, registrystorage.ByOperatorID(operatorData.ID), registrystorage.ByNotLiquidated())
-	require.Equal(t, 1000, len(all))
+	require.Equal(t, 100, len(all))
 }
--- a/operator/storage/storage.go
+++ b/operator/storage/storage.go
@@ -67,19 +67,25 @@ type storage struct {
 
 // NewNodeStorage creates a new instance of Storage
 func NewNodeStorage(beaconCfg *networkconfig.Beacon, logger *zap.Logger, db basedb.Database) (Storage, error) {
+	recipientStore, setSharesUpdater := registrystorage.NewRecipientsStorage(logger, db, OperatorStoragePrefix)
+
 	stg := &storage{
 		logger:         logger,
 		db:             db,
 		operatorStore:  registrystorage.NewOperatorsStorage(logger, db, OperatorStoragePrefix),
-		recipientStore: registrystorage.NewRecipientsStorage(logger, db, OperatorStoragePrefix),
+		recipientStore: recipientStore,
 	}
 
 	var err error
 
-	stg.shareStore, stg.validatorStore, err = registrystorage.NewSharesStorage(beaconCfg, db, OperatorStoragePrefix)
+	// Create shares storage with recipients dependency
+	stg.shareStore, stg.validatorStore, err = registrystorage.NewSharesStorage(beaconCfg, db, recipientStore, OperatorStoragePrefix)
 	if err != nil {
 		return nil, err
 	}
+
+	// Wire back for updates: recipients -> shares (clean and simple!)
+	setSharesUpdater(stg.shareStore)
 
 	return stg, nil
 }
--- a/protocol/v2/ssv/runner/validator_registration.go
+++ b/protocol/v2/ssv/runner/validator_registration.go
@@ -1,19 +1,19 @@
 package runner
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"sync"
 
 	"github.com/attestantio/go-eth2-client/api"
 	v1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec"
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/cespare/xxhash/v2"
-	"github.com/ethereum/go-ethereum/common"
 	ssz "github.com/ferranbt/fastssz"
 	"github.com/pkg/errors"
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
@@ -31,8 +31,6 @@ import (
 	"github.com/ssvlabs/ssv/operator/slotticker"
 	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
-	registrystorage "github.com/ssvlabs/ssv/registry/storage"
-	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
 const (
@@ -48,9 +46,7 @@ type ValidatorRegistrationRunner struct {
 	signer                         ekm.BeaconSigner
 	operatorSigner                 ssvtypes.OperatorSigner
 	valCheck                       specqbft.ProposedValueCheckF
-	recipientsStorage              recipientsStorage
 	validatorRegistrationSubmitter ValidatorRegistrationSubmitter
-	validatorOwnerAddress          common.Address
 
 	gasLimit uint64
 }
@@ -58,12 +54,10 @@ type ValidatorRegistrationRunner struct {
 func NewValidatorRegistrationRunner(
 	networkConfig *networkconfig.Network,
 	share map[phase0.ValidatorIndex]*spectypes.Share,
-	validatorOwnerAddress common.Address,
 	beacon beacon.BeaconNode,
 	network specqbft.Network,
 	signer ekm.BeaconSigner,
 	operatorSigner ssvtypes.OperatorSigner,
-	recipientsStorage recipientsStorage,
 	ValidatorRegistrationSubmitter ValidatorRegistrationSubmitter,
 	gasLimit uint64,
 ) (Runner, error) {
@@ -82,9 +76,7 @@ func NewValidatorRegistrationRunner(
 		network:                        network,
 		signer:                         signer,
 		operatorSigner:                 operatorSigner,
-		recipientsStorage:              recipientsStorage,
 		validatorRegistrationSubmitter: ValidatorRegistrationSubmitter,
-		validatorOwnerAddress:          validatorOwnerAddress,
 
 		gasLimit: gasLimit,
 	}, nil
@@ -260,17 +252,18 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
 }
 
 func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0.Slot) (*v1.ValidatorRegistration, error) {
-	pk := phase0.BLSPubKey{}
-	copy(pk[:], r.GetShare().ValidatorPubKey[:])
-
-	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
-
-	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
-	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
+	share := r.GetShare()
+	if share == nil {
+		return nil, errors.New("no share to get validator public key")
 	}
-	if !found {
-		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
+
+	pk := phase0.BLSPubKey{}
+	copy(pk[:], share.ValidatorPubKey[:])
+
+	// Check for zero fee recipient address
+	zeroAddress := bellatrix.ExecutionAddress{}
+	if bytes.Equal(share.FeeRecipientAddress[:], zeroAddress[:]) {
+		return nil, errors.New("fee recipient address is zero, validator rewards would be lost")
 	}
 
 	// Set the default GasLimit value if it hasn't been specified already, use 36 or 30 depending
@@ -284,8 +277,9 @@ func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0
 		gasLimit = defaultGasLimit
 	}
 
+	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
 	return &v1.ValidatorRegistration{
-		FeeRecipient: rData.FeeRecipient,
+		FeeRecipient: share.FeeRecipientAddress,
 		GasLimit:     gasLimit,
 		Timestamp:    r.BaseRunner.NetworkConfig.EpochStartTime(epoch),
 		Pubkey:       pk,
@@ -344,10 +338,6 @@ func (r *ValidatorRegistrationRunner) GetRoot() ([32]byte, error) {
 	}
 	ret := sha256.Sum256(marshaledRoot)
 	return ret, nil
-}
-
-type recipientsStorage interface {
-	GetRecipientData(r basedb.Reader, owner common.Address) (*registrystorage.RecipientData, bool, error)
 }
 
 type ValidatorRegistrationSubmitter interface {
--- a/registry/storage/recipients.go
+++ b/registry/storage/recipients.go
@@ -17,6 +17,20 @@ var (
 	recipientsPrefix = []byte("recipients")
 )
 
+// SharesUpdater is an interface for updating shares when fee recipient changes
+type SharesUpdater interface {
+	UpdateFeeRecipientForOwner(owner common.Address, feeRecipient bellatrix.ExecutionAddress)
+}
+
+// SetSharesUpdaterFunc is a function type for setting the shares updater after initialization
+type SetSharesUpdaterFunc func(SharesUpdater)
+
+// RecipientReader is a read-only interface for querying recipient data
+type RecipientReader interface {
+	GetRecipientData(r basedb.Reader, owner common.Address) (*RecipientData, bool, error)
+	GetRecipientDataMany(r basedb.Reader, owners []common.Address) (map[common.Address]bellatrix.ExecutionAddress, error)
+}
+
 type Nonce uint16
 
 // RecipientData the public data of a recipient
@@ -60,8 +74,7 @@ type recipientDataJSON struct {
 
 // Recipients is the interface for managing recipients data
 type Recipients interface {
-	GetRecipientData(r basedb.Reader, owner common.Address) (*RecipientData, bool, error)
-	GetRecipientDataMany(r basedb.Reader, owners []common.Address) (map[common.Address]bellatrix.ExecutionAddress, error)
+	RecipientReader // Embed read-only interface
 	GetNextNonce(r basedb.Reader, owner common.Address) (Nonce, error)
 	BumpNonce(rw basedb.ReadWriter, owner common.Address) error
 	SaveRecipientData(rw basedb.ReadWriter, recipientData *RecipientData) (*RecipientData, error)
@@ -71,27 +84,38 @@ type Recipients interface {
 }
 
 type recipientsStorage struct {
-	logger *zap.Logger
-	db     basedb.Database
-	lock   sync.RWMutex
-	prefix []byte
+	logger        *zap.Logger
+	db            basedb.Database
+	lock          sync.RWMutex
+	prefix        []byte
+	sharesUpdater SharesUpdater // For notifying shares storage of updates
 }
 
-// NewRecipientsStorage creates a new instance of Storage
-func NewRecipientsStorage(logger *zap.Logger, db basedb.Database, prefix []byte) Recipients {
-	return &recipientsStorage{
+// NewRecipientsStorage creates a new instance of Storage and returns a setter for wiring
+func NewRecipientsStorage(logger *zap.Logger, db basedb.Database, prefix []byte) (Recipients, SetSharesUpdaterFunc) {
+	rs := &recipientsStorage{
 		logger: logger,
 		db:     db,
 		prefix: prefix,
 	}
+
+	// Return both the interface and a setter function for clean wiring
+	setter := func(updater SharesUpdater) {
+		rs.sharesUpdater = updater
+	}
+
+	return rs, setter
 }
+
+// SetSharesUpdater is no longer needed - wiring is done via the setter function returned by NewRecipientsStorage
 
 // GetRecipientsPrefix returns DB prefix
 func (s *recipientsStorage) GetRecipientsPrefix() []byte {
 	return recipientsPrefix
 }
 
-// GetRecipientData returns data of the given recipient by owner address, if not found returns owner address as a default fee recipient
+// GetRecipientData returns data of the given recipient by owner address, returns nil if not found.
+// Note: When enriching shares, if no recipient data is found, the owner address is used as default fee recipient
 func (s *recipientsStorage) GetRecipientData(r basedb.Reader, owner common.Address) (*RecipientData, bool, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
@@ -217,7 +241,15 @@ func (s *recipientsStorage) SaveRecipientData(rw basedb.ReadWriter, recipientDat
 		return nil, errors.Wrap(err, "could not marshal recipient data")
 	}
 
-	return recipientData, s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	err = s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	if err != nil {
+		return nil, err
+	}
+
+	// Notify shares storage of the fee recipient update
+	s.sharesUpdater.UpdateFeeRecipientForOwner(recipientData.Owner, recipientData.FeeRecipient)
+
+	return recipientData, nil
 }
 
 func (s *recipientsStorage) DeleteRecipientData(rw basedb.ReadWriter, owner common.Address) error {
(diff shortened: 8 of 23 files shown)


scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...750294a6","6fc7c41b42b4e3e5","f28f3a41bc8c741d","b49179c36c111cca","5df9f55a706e5291","4994abbc55e58278","2c02276fa72f61b9","937ce36b0949796a","3f6f707d5ecf3903","11a340273beaf077","2630b7d3f7463316","33dd6bfa496bc889","a9685eff5edb9118","2fd0165ccc59ddf","de0733313f796fc9","a5458fa616bc6d6d","73244bf5a8b49e23","42052e9213a5780d","26672cc9a6ae0af0","b7ca13196a824fc5","801f7d8a8151c73e","20b4765e446c4803","25c4b276cfa24926","cebe87e9824d213e","be65f009ecd0d4b8","4d1a0519b660e5f","ad8332e881c48041","fbcd6b75d8baa5c1","787162ea04b0c4ed","c97a6f1c933535b","b5b9b0a45f6df79e","e9569ea2dc0f1406","5a3c840d9fc7bec9","7cf54a264ab481f3","cbbb16ca9342dc4c","46636a126e9d32ad","c90ffdb18233f0ea","e5e2698cc827129b","b31079cc52d70820","99ebbf611546fd06","a6c82644fe7b2bf3","c110b52f3fc04964","b2ba608989ac677","373a9525ae8811a8","44a2015641f07e3","d6aafafb59a8dd6e","b7f2e4dec9742e39","bb626b923fe3850","b2315a5d4749f93","f1b684011e369981","fb2604dba9677890","b6908fa26415a154","70247f4b2e200f5b","d3ac29b25686ae94","636c42279c683d53","523ff663791a047a","cee19c2663445d45","819aa7ef21eb1f94","6e7d2035244a01b4","5a73c0f03029acba","96abe9e4636a4d4a","16f41dbe0a2ff32a","e16b89eb783f91bd","c0c8738f2d82c5bd"]
```
olegshmuelov:
```diff
diff --git a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner b/ssv/validator_registration.go@NewValidatorRegistrationRunner
index 16f41dbe0a2ff32a..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@NewValidatorRegistrationRunner
@@ -1,27 +1,25 @@
 func NewValidatorRegistrationRunner(
-	networkConfig *networkconfig.Network,
+	beaconNetwork BeaconNetwork,
 	share map[phase0.ValidatorIndex]*Share,
-	beacon beacon.BeaconNode,
+	beacon BeaconNode,
 	network Network,
-	signer ekm.BeaconSigner,
-	operatorSigner ssvtypes.OperatorSigner,
-	ValidatorRegistrationSubmitter ValidatorRegistrationSubmitter,
+	signer BeaconSigner,
+	operatorSigner *OperatorSigner,
 	gasLimit uint64,
 ) (Runner, error) {
 	if len(share) != 1 {
 		return nil, errors.New("must have one share")
 	}
 	return &ValidatorRegistrationRunner{
 		BaseRunner: &BaseRunner{
 			RunnerRoleType:	RoleValidatorRegistration,
-			NetworkConfig:	networkConfig,
+			BeaconNetwork:	beaconNetwork,
 			Share:		share,
 		},
-		beacon:				beacon,
-		network:			network,
-		signer:				signer,
-		operatorSigner:			operatorSigner,
-		validatorRegistrationSubmitter:	ValidatorRegistrationSubmitter,
+		beacon:		beacon,
+		network:	network,
+		signer:		signer,
+		operatorSigner:	operatorSigner,
 		gasLimit:	gasLimit,
 	}, nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner b/ssv/validator_registration.go@ValidatorRegistrationRunner
index e16b89eb783f91bd..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner
@@ -1,10 +1,9 @@
 ValidatorRegistrationRunner struct {
 	BaseRunner	*BaseRunner
-	beacon				beacon.BeaconNode
-	network				Network
-	signer				ekm.BeaconSigner
-	operatorSigner			ssvtypes.OperatorSigner
-	valCheck			ProposedValueCheckF
-	validatorRegistrationSubmitter	ValidatorRegistrationSubmitter
+	beacon		BeaconNode
+	network		Network
+	signer		BeaconSigner
+	operatorSigner	*OperatorSigner
+	valCheck	ProposedValueCheckF
 	gasLimit	uint64
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
index c0c8738f2d82c5bd..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.calculateValidatorRegistration
@@ -1,27 +1,15 @@
 func (r *ValidatorRegistrationRunner) calculateValidatorRegistration(slot phase0.Slot) (*v1.ValidatorRegistration, error) {
 	share := r.GetShare()
 	if share == nil {
 		return nil, errors.New("no share to get validator public key")
 	}
-	zeroAddress := bellatrix.ExecutionAddress{}
-	if bytes.Equal(share.FeeRecipientAddress[:], zeroAddress[:]) {
-		return nil, errors.New("fee recipient address is zero, validator rewards would be lost")
-	}
 	pk := phase0.BLSPubKey{}
 	copy(pk[:], share.ValidatorPubKey[:])
-	gasLimit := r.gasLimit
-	if gasLimit == 0 {
-		defaultGasLimit := DefaultGasLimit
-		if !r.BaseRunner.NetworkConfig.GasLimit36Fork() {
-			defaultGasLimit = DefaultGasLimitOld
-		}
-		gasLimit = defaultGasLimit
-	}
-	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
+	epoch := r.BaseRunner.BeaconNetwork.EstimatedEpochAtSlot(slot)
 	return &v1.ValidatorRegistration{
 		FeeRecipient:	share.FeeRecipientAddress,
-		GasLimit:	gasLimit,
-		Timestamp:	r.BaseRunner.NetworkConfig.EpochStartTime(epoch),
+		GasLimit:	r.gasLimit,
+		Timestamp:	r.BaseRunner.BeaconNetwork.EpochStartTime(epoch),
 		Pubkey:		pk,
 	}, nil
 }
\ No newline at end of file
```

registry/storage/recipients.go
```diff
@@ -217,7 +241,15 @@ func (s *recipientsStorage) SaveRecipientData(rw basedb.ReadWriter, recipientDat
...
 	}
 
-	return recipientData, s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	err = s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	if err != nil {
+		return nil, err
+	}
+
+	// Notify shares storage of the fee recipient update
+	s.sharesUpdater.UpdateFeeRecipientForOwner(recipientData.Owner, recipientData.FeeRecipient)
```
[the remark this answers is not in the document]
iurii-ssv: I think it's better to `panic` in this case to find this out quickly (rather than silently miss these updates if we forgot to wire this up correctly)

registry/storage/recipients.go
```diff
@@ -217,7 +239,15 @@ func (s *recipientsStorage) SaveRecipientData(rw basedb.ReadWriter, recipientDat
...
```
iurii-ssv: I'm not sure if this cyclic-wiring between `recipientsStorage` and `sharesStorage` is worth the complexity it brings, it's an interesting solution but from my experience cyclic-wiring make code harder to modify long-term,

it seems the only place where we update fee-recipient is `handleFeeRecipientAddressUpdated` - so maybe we could instead call "UpdateFeeRecipient" for both (`recipientsStorage` and `sharesStorage`) from there ? And if we need to do that in multiple places - we could come up with a separate service/abstraction that just turns this into a single func-call

Anyway, we can address it separately later (if we even need to), just saying that by default I wouldn't go for this kind of pattern (unless I absolutely have to)

registry/storage/shares.go
```diff
@@ -174,8 +183,72 @@ func (s *sharesStorage) loadPubkeyToIndexMappings() (map[spectypes.ValidatorPK]p
 	return m, err
 }
 
+// enrichSharesWithFeeRecipients batch-enriches shares with fee recipients for better performance
+func (s *sharesStorage) enrichSharesWithFeeRecipients(shares []*types.SSVShare) error {
```
iurii-ssv: `enrichSharesWithFeeRecipients` method is currently called from `loadFromDB` and `saveToDB`. For `loadFromDB` there is no need to worry about concurrent access to `sharesStorage.shares` (and I guess `sharesStorage.sharesByOwner` as well) since it's running prior to go-routines that are meant to access in-memory shares are started. 

But for `saveToDB` I think the situation is tricky ...
- `saveToDB` (and hence `enrichSharesWithFeeRecipients`) is called with `shares ...*types.SSVShare` that's different depending on the caller
- I assume our intent is to update fee-recipient for when it's called with `sharesStorage.shares` ? So it would seem we **shouldn't** update fee-recipient in this call-sequence `Save`->`saveToDB`->`enrichSharesWithFeeRecipients`, but we probably **should** in this call-sequence `UpdateValidatorsMetadata
`->`saveToDB`->`enrichSharesWithFeeRecipients` ?
- additionally, I think access to in-memory shares (`sharesStorage.shares`) should be protected by `sharesStorage.memoryMtx` (and not `sharesStorage.storageMtx`) - and currently it seems the  `UpdateValidatorsMetadata
`->`saveToDB`->`enrichSharesWithFeeRecipients` call-sequence doesn't lock `sharesStorage.memoryMtx` (it probably should in some way, but also such that it doesn't deadlock with `sharesStorage.storageMtx` ofc)

registry/storage/shares.go
```diff
@@ -62,13 +64,18 @@ type Shares interface {
...
 
 type sharesStorage struct {
-	db             basedb.Database
-	storagePrefix  []byte
-	shares         map[string]*types.SSVShare
-	validatorStore *validatorStore
+	db              basedb.Database
+	storagePrefix   []byte
+	shares          map[string]*types.SSVShare
+	sharesByOwner   map[common.Address][]*types.SSVShare // Index for efficient owner lookups
```
iurii-ssv: Perhaps we could group these 3 together (in a dedicated block) - so it's clear what `memoryMtx` protects ?
```go
...

    // memoryMtx allows more granular access to in-memory shares map
	// to minimize time spent in 'locked' mode when updating state
	memoryMtx sync.RWMutex
	shares          map[string]*types.SSVShare
	sharesByOwner   map[common.Address][]*types.SSVShare // Index for efficient owner lookups

...
```

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -260,19 +252,20 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
-
-	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
-
-	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
-	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
+	share := r.GetShare()
+	if share == nil {
+		return nil, errors.New("no share to get validator public key")
 	}
```
iurii-ssv: There isn't really a need to check for `if share == nil` since the runner can't / shouldn't be started with `nil` share (I think this is checked for somewhere prior).

Also, I think here we shouldn't really use `BaseRunner.Share.FeeRecipientAddress` here for validator-registrations because it's initialized once on SSV node startup and is never updated, or if it is updated indirectly (due to `BaseRunner.Share` pointing to the same object that is updated) - we are not explicitly synchronizing such an update with the code here (the update is happening in another go-routine) ... that's why `ValidatorRegistrationRunner` had `recipientsStorage recipientsStorage` on it previously - otherwise we can potentially end-up with corrupted data for fee-recipient

iurii-ssv (commented):
Great stuff! We do very much need this indeed

network/topics/controller_test.go
```diff
@@ -320,6 +322,17 @@ func committeeTopic(cidHex string) string {
 	return commons.CommitteeTopicID(spectypes.CommitteeID(cid))[0]
 }
 
+// noOpRecipientReader is a no-op implementation for tests
+type noOpRecipientReader struct{}
```
oleg-ssvlabs: Question: Do we usually name these as `mock` or `noop`?  

This object is used only in tests, potentially for mocking the behavior of `RecipientReader`. Even though it currently doesn’t do anything, I would prefer naming it `mock`, because there’s a chance its behavior will be mocked in the future to set up some test cases.

operator/fee_recipient/controller_test.go
```diff
@@ -40,109 +36,220 @@ func TestSubmitProposal(t *testing.T) {
...
-		wg.Add(numberOfRequests) // Set up the wait group before starting goroutines
+	t.Run("custom fee recipients from storage", func(t *testing.T) {
+		// Set custom fee recipients for some validators
+		customRecipient1 := bellatrix.ExecutionAddress{0x11, 0x22, 0x33, 0x44, 0x55}
+		customRecipient2 := bellatrix.ExecutionAddress{0xaa, 0xbb, 0xcc, 0xdd, 0xee}
+
+		owner1 := common.HexToAddress("0x0000000000000000000000000000000000000000")
+		owner2 := common.HexToAddress("0x0000000000000000000000000000000000000001")
+
+		_, err := recipientStorage.SaveRecipientData(nil, &registrystorage.RecipientData{
```
oleg-ssvlabs: Very minor refactor suggestion, which mainly provides the following benefits:  
- Easier to extend the test with more recipients  
- No need to repeat the `SaveRecipientData` call  
- No need to repeat the `require.NoError` check  

```Go
recipients := []*registrystorage.RecipientData{
		{
			Owner:        owner1,
			FeeRecipient: customRecipient1,
		},
		{
			Owner:        owner2,
			FeeRecipient: customRecipient2,
		},
}
for _, recipient := range recipients {
		_, err := recipientStorage.SaveRecipientData(nil, recipient)
		require.NoError(t, err)
}
```

oleg-ssvlabs: Also you could inline Owner and FeeRecipient assignment and during the assertion you can do this instead:

```Go
require.Equal(t, recipients[0]. FeeRecipient, capturedRecipients[0])
```

operator/fee_recipient/controller_test.go
```diff
@@ -40,109 +36,220 @@ func TestSubmitProposal(t *testing.T) {
...
+			func(ctx context.Context, feeRecipients map[phase0.ValidatorIndex]bellatrix.ExecutionAddress) error {
+				capturedRecipients = feeRecipients
+				return nil
+			}).Times(1)
 
 		frCtrl.beaconClient = client
-		frCtrl.slotTickerProvider = func() slotticker.SlotTicker {
-			return ticker
+
+		err = frCtrl.prepareAndSubmit(t.Context())
```
oleg-ssvlabs: I am not really sure about the idea of testing non-exported methods.  
I understand that the test setup looked a little more complex before, but to me, testing only exported methods/functions is the right approach. Otherwise, the tests do not cover the full execution flow.

operator/fee_recipient/controller_test.go
```diff
@@ -40,109 +36,220 @@ func TestSubmitProposal(t *testing.T) {
...
 
-	db, shareStorage, recipientStorage := createStorage(t)
-	defer db.Close()
+	db, shareStorage, recipientStorage := createStorageWithRecipients(t)
+	defer func() { require.NoError(t, db.Close()) }()
 
 	beaconConfig := networkconfig.TestNetwork.Beacon
 	populateStorage(t, shareStorage, operatorData)
 
 	frCtrl := NewController(logger, &ControllerOptions{
```
oleg-ssvlabs: because you almost in every test instantiate new controller, I would rather move this out of this "shared" place. The `frCtrl` is reused only in two tests, others create their own local instances.

operator/fee_recipient/controller_test.go
```diff
@@ -40,109 +36,220 @@ func TestSubmitProposal(t *testing.T) {
...
+		frCtrl3.beaconClient = client
 
-		go frCtrl.Start(t.Context())
-		mockTimeChan <- time.Now()
-		wg.Add(2)
-		wg.Wait()
-		close(mockTimeChan)
+		// Should handle error gracefully without panic
+		err := frCtrl3.prepareAndSubmit(t.Context())
+		require.NoError(t, err) // prepareAndSubmit logs errors but returns nil
```
nkryuchkov: I think while we are touching it, we need to delete `prepareAndSubmit`'s return parameter instead of writing a comment that it always returns `nil`

operator/storage/storage.go
```diff
@@ -67,20 +67,26 @@ type storage struct {
 
 // NewNodeStorage creates a new instance of Storage
 func NewNodeStorage(beaconCfg *networkconfig.Beacon, logger *zap.Logger, db basedb.Database) (Storage, error) {
+	recipientStore, setSharesUpdater := registrystorage.NewRecipientsStorage(logger, db, OperatorStoragePrefix)
```
oleg-ssvlabs: don't you think it would better to rather inject both `recipientStore`, `shareStore` and `validatorStore` via ctor?

registry/storage/recipients.go
```diff
@@ -17,6 +17,20 @@ var (
 	recipientsPrefix = []byte("recipients")
 )
 
+// SharesUpdater is an interface for updating shares when fee recipient changes
+type SharesUpdater interface {
```
nkryuchkov: Shouldn't this be `FeeRecipientUpdater`?

oleg-ssvlabs: it seems all of these types are used only within `storage` package, so there is no need to export them

registry/storage/recipients.go
```diff
@@ -71,27 +84,36 @@ type Recipients interface {
...
+// NewRecipientsStorage creates a new instance of Storage and returns a setter for wiring
+func NewRecipientsStorage(logger *zap.Logger, db basedb.Database, prefix []byte) (Recipients, SetSharesUpdaterFunc) {
+	rs := &recipientsStorage{
 		logger: logger,
 		db:     db,
 		prefix: prefix,
 	}
+
+	// Return both the interface and a setter function for clean wiring
+	setter := func(updater SharesUpdater) {
```
oleg-ssvlabs: Instead of returning this setter, wouldn't it be cleaner to just inject Share storage and call it somewhere [here](https://github.com/ssvlabs/ssv/pull/2457/files#diff-296461dc4249eae7e67edc89864352afe485bb634af6cb9ae9f0fac501c86421R248) if that is possible.

registry/storage/shares.go
```diff
@@ -174,8 +183,72 @@ func (s *sharesStorage) loadPubkeyToIndexMappings() (map[spectypes.ValidatorPK]p
 	return m, err
 }
 
+// enrichSharesWithFeeRecipients batch-enriches shares with fee recipients for better performance
+func (s *sharesStorage) enrichSharesWithFeeRecipients(shares []*types.SSVShare) error {
+	// Collect unique owners while loading shares
+	uniqueOwners := make(map[common.Address]struct{})
+	var owners []common.Address
```
oleg-ssvlabs: I would rename the `owners` variable, because this slice also contains unique items

registry/storage/shares.go
```diff
@@ -174,8 +183,72 @@ func (s *sharesStorage) loadPubkeyToIndexMappings() (map[spectypes.ValidatorPK]p
...
+	}
+
+	return nil
+}
+
+// removeFromOwnerIndex removes a share from the owner's index
+func (s *sharesStorage) removeFromOwnerIndex(owner common.Address, share *types.SSVShare) {
+	shares := s.sharesByOwner[owner]
+	for i, sh := range shares {
+		if sh == share {
```
nkryuchkov: The logic with pointer comparison and removal seems correct, but not trivial, so IMO we need some tests for this logic

registry/storage/recipients.go
```diff
@@ -17,6 +17,20 @@ var (
...
+// SharesUpdater is an interface for updating shares when fee recipient changes
+type SharesUpdater interface {
+	UpdateFeeRecipientForOwner(owner common.Address, feeRecipient bellatrix.ExecutionAddress)
+}
+
+// SetSharesUpdaterFunc is a function type for setting the shares updater after initialization
+type SetSharesUpdaterFunc func(SharesUpdater)
+
+// RecipientReader is a read-only interface for querying recipient data
+type RecipientReader interface {
```
oleg-ssvlabs: I am not sure about the semantics of this interface.  

Would it be possible to have a single method instead, like this:

```go
type RecipientReader interface {
    GetRecipientData(r basedb.Reader, owners ...common.Address) (map[common.Address]*RecipientData, error)
}
```

Most likely, you implemented it this way because of the underlying use of `db.Get()` and `db.GetMany()`.
However, I think implementation details should not affect the semantics of what is exposed (`GetRecipientData` vs `GetRecipientDataMany`).
You could potentially just check the length of owners and, based on that, decide whether to use `db.Get()` or `db.GetMany()`.

registry/storage/recipients.go
```diff
@@ -217,7 +239,15 @@ func (s *recipientsStorage) SaveRecipientData(rw basedb.ReadWriter, recipientDat
 		return nil, errors.Wrap(err, "could not marshal recipient data")
 	}
 
-	return recipientData, s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	err = s.db.Using(rw).Set(s.prefix, buildRecipientKey(recipientData.Owner), raw)
+	if err != nil {
+		return nil, err
```
oleg-ssvlabs: most of the methods return wrapped errors, so maybe worth following the same pattern here

registry/storage/shares.go
```diff
@@ -174,8 +183,72 @@ func (s *sharesStorage) loadPubkeyToIndexMappings() (map[spectypes.ValidatorPK]p
...
+			// Default to owner address if no custom recipient
+			copy(share.FeeRecipientAddress[:], share.OwnerAddress.Bytes())
+		}
+	}
+
+	return nil
+}
+
+// removeFromOwnerIndex removes a share from the owner's index
+func (s *sharesStorage) removeFromOwnerIndex(owner common.Address, share *types.SSVShare) {
```
oleg-ssvlabs: Should this method also be thread-safe?

registry/storage/shares_test.go
```diff
@@ -748,3 +766,292 @@ func (t *testStorage) Recreate(logger *zap.Logger) error {
...
+	recipientStorage, setSharesUpdater := NewRecipientsStorage(logger, db, []byte("test-recipient"))
+	sharesStorage, _, err := NewSharesStorage(networkconfig.TestNetwork.Beacon, db, recipientStorage, []byte("test"))
+	require.NoError(t, err)
+	setSharesUpdater(sharesStorage)
+
+	owner := common.HexToAddress("0x1234567890123456789012345678901234567890")
+	customFeeRecipient := bellatrix.ExecutionAddress{0x11, 0x22, 0x33}
+
+	// Create 3 shares for the same owner
+	share1 := fakeParticipatingShare(1, generateRandomPubKey(), []uint64{1, 2, 3, 4})
```
oleg-ssvlabs: Similar to what I mentioned above, it is better to utilize collections when working with multiple instances.

oleg-ssvlabs (commented):
great work!

I left couple of comments

registry/storage/recipients.go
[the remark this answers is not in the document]
olegshmuelov: the design changed, not relevant anymore

operator/storage/storage.go
olegshmuelov: It's outdated, but could be still relevant.
I understand the benefit of dependency injection here. However, there's a circular dependency consideration - `shareStore` and `validatorStore` are created together by `NewSharesStorage` and require
  `recipientStore` as a dependency.

  If we inject all stores, the caller would need to:
  1. Create `recipientStore` first
  2. Pass it to `NewSharesStorage` to create `shareStore` and `validatorStore`
  3. Then pass all of them to `NewNodeStorage`

  This pushes the initialization complexity to every caller. The current approach encapsulates this initialization logic in one place.

registry/storage/recipients.go
olegshmuelov: outdated

registry/storage/recipients.go
olegshmuelov: outdated

registry/storage/shares.go
olegshmuelov: outdated

registry/storage/shares.go
olegshmuelov: outdated

registry/storage/recipients.go
olegshmuelov: outdated

registry/storage/shares.go
olegshmuelov: outdated

network/topics/controller_test.go
olegshmuelov: outdated

registry/storage/recipients.go
olegshmuelov: outdated

registry/storage/recipients.go
olegshmuelov: outdated

registry/storage/shares.go
olegshmuelov: outdated

operator/fee_recipient/controller_test.go
olegshmuelov: 5be8391acc0fc298606286fe59f672bc76ddcdd7

operator/fee_recipient/controller_test.go
olegshmuelov: c7cc0dfe2285a77c32cc452e739b2110d226446e

registry/storage/recipients.go
olegshmuelov: 6bf73b7d1feb498a708097c36d7ae8560918cb9e

registry/storage/shares_test.go
olegshmuelov: 4e36071c2d4eb16b6725cba88d69c67f7b2ebc6c

eth/eventhandler/handlers.go
```diff
@@ -460,18 +461,16 @@ func (eh *EventHandler) handleFeeRecipientAddressUpdated(txn basedb.Txn, event *
...
-		recipientData = &registrystorage.RecipientData{
-			Owner: event.Owner,
-		}
+	// Reject zero fee recipient address
+	if feeRecipient == (bellatrix.ExecutionAddress{}) {
+		return false, fmt.Errorf("invalid zero fee recipient address")
 	}
-	copy(recipientData.FeeRecipient[:], event.RecipientAddress.Bytes())
-	r, err := eh.nodeStorage.SaveRecipientData(txn, recipientData)
+
```
olegshmuelov: Should we enforce this restriction?

  Arguments for keeping it:
  - Protects users from accidentally burning rewards due to misconfiguration
  - Zero address is likely an error rather than intentional
  - Safer default behavior

  Arguments for removing it:
  - User autonomy - if someone wants to burn rewards, why stop them?
  - Some users might intentionally want to burn rewards (e.g., for tax reasons, ideological reasons, or testing)
  - Ethereum protocol itself allows zero fee recipient

  What do you think? Should we be paternalistic and protect users from potential mistakes, or respect user freedom even if it means they might burn rewards?

operator/fee_recipient/controller_test.go
iurii-ssv: > but to me, testing only exported methods/functions is the right approach

To chime in on this, all else being equal I would agree that testing public methods/API is better because it gives us higher coverage and usage-examples,

but there is a reason Golang permits to place tests in the same package with the code (rather than `_test` package only - which allows for testing public methods only) - in practice it's often simpler to test some involved private func that has many outcomes ... while public funcs that use it are not really adding any complexity on top that's worth testing for (so it wouldn't be worth to be forced to test that "complex private piece of code" through publicly-exposed means only),

I didn't really look at how this applies to this code though, it's probably better to leave it up to the implementor (he usually knows better how to test the code he wrote to get the most bang for the buck)

eth/eventhandler/handlers.go
iurii-ssv: Since there are "valid" reasons to use zero fee in Ethereum - I think we should, maybe even have to, support that behavior (not impose limits), plus things like validating user-inputs should probably be done on the UI side of things

network/topics/controller_test.go
```diff
@@ -406,10 +406,14 @@ func newPeer(ctx context.Context, logger *zap.Logger, t *testing.T, msgValidator
...
 		// TODO: add mock for peers.ScoreIndex
 	}
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
 	db, err := kv.NewInMemory(logger, basedb.Options{})
 	require.NoError(t, err)
 
-	_, validatorStore, err := registrystorage.NewSharesStorage(networkconfig.TestNetwork.Beacon, db, []byte("test"))
+	mockRecipients := mocks.NewMockRecipients(ctrl)
```
iurii-ssv: I'm not sure how `gomock.Controller` works these days, but in my understanding:
- there is no longer a need to finish if with `defer ctrl.Finish()`
- and maybe it's handled differently now but previously by calling `defer ctrl.Finish()` you would render the mocks relying on this `ctrl` not record / verify method calls, I guess ?

Anyway, it doesn't seem right for `defer ctrl.Finish()` to be called here since the test is not finished yet by the time we `return` from this func

operator/fee_recipient/controller.go
```diff
@@ -53,112 +56,106 @@ func NewController(logger *zap.Logger, opts *ControllerOptions) *recipientContro
...
+		validatorProvider:  opts.ValidatorProvider,
 		slotTickerProvider: opts.SlotTickerProvider,
 		operatorDataStore:  opts.OperatorDataStore,
 	}
 }
 
 func (rc *recipientController) Start(ctx context.Context) {
-	rc.listenToTicker(ctx)
+	go rc.listenToTicker(ctx)
+}
```
iurii-ssv: Not that it matters much, but perhaps we'd want to also change `go n.feeRecipientCtrl.Start(ctx)` -> `n.feeRecipientCtrl.Start(ctx)` then ?

On that note, people mostly make `Start` a synchronous call I think, maybe we could call it `StartAsync` ... but I don't have a strong opinion on this (the original way it was done seems like the most straightforward to me)

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -259,20 +253,17 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
 	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
-	}
-	if !found {
-		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
+		return nil, fmt.Errorf("could not get fee recipient for validator %x: %w", validatorPubKey, err)
 	}
 
+	pk := phase0.BLSPubKey{}
+	copy(pk[:], validatorPubKey[:])
```
iurii-ssv: I think we can just type-cast (I assume it would copy the data over) like `phase0.BLSPubKey(validatorPubKey)` below instead of doing `copy` explicitly ? I've seen a relevant comment about it, something like:
```go
	// This copy is required to avoid the "cgo argument has Go pointer to Go pointer" panic.
	// TODO: (Alan) should we do this here or at the source? (wherever the bytes are coming from)
	pkCpy := make([]byte, len(b))
	copy(pkCpy, b)
```
but I don't think it's relevant to this code, or most of the code where we do this kind of pattern (it would be better to resolve it else-where anyway)

Just pointing it out to be on the same page, cc @MatusKysel @nkryuchkov @oleg-ssvlabs @julienh-ssv

registry/storage/recipients.go
```diff
@@ -60,43 +62,84 @@ type recipientDataJSON struct {
...
+// It maintains an in-memory cache of all fee recipients for fast access while
+// persisting data to the underlying database.
 type recipientsStorage struct {
 	logger *zap.Logger
 	db     basedb.Database
 	lock   sync.RWMutex
 	prefix []byte
+
+	// In-memory map of all fee recipients for fast access
+	feeRecipients map[common.Address]bellatrix.ExecutionAddress
```
iurii-ssv: While we are here, I would add a clarifying comment for what the purpose of `lock` field is (what kind of things it's meant to synchronize/protect)

registry/storage/recipients.go
```diff
@@ -60,43 +62,84 @@ type recipientDataJSON struct {
...
+	if !found {
+		return bellatrix.ExecutionAddress{}, fmt.Errorf("fee recipient not found for owner %s", owner.Hex())
+	}
+	return recipient, nil
+}
+
+// GetRecipientsPrefix returns DB prefix
+func (s *recipientsStorage) GetRecipientsPrefix() []byte {
+	return recipientsPrefix
 }
```
iurii-ssv: Seems like this method is not used anymore

registry/storage/recipients.go
```diff
@@ -60,43 +62,84 @@ type recipientDataJSON struct {
...
 
 func (s *recipientsStorage) getRecipientData(r basedb.Reader, owner common.Address) (*RecipientData, bool, error) {
```
iurii-ssv, line 142: Shouldn't `getRecipientData` check if it has the data in in-memory map and return that (currently it always returns whatever is in DB) ?

Although I guess `feeRecipients map[common.Address]bellatrix.ExecutionAddress` would need to be `feeRecipients map[common.Address]*RecipientData` instead ... but still wouldn't that be more straightforward ?

I'm also not sure, btw, whether we even need this in-memory representation for "fast access", I would assume the underlying database itself might have some sort of caching layer of its own (although it might be limited in some ways due to the need to be generic and hold arbitrary data sizes in memory) ... that's why I wouldn't roll out our own caching implementation unless I'm sure that it really improves some performance (since it's more code that we need to make sure works, and maintain).

registry/storage/shares.go
```diff
@@ -187,8 +200,16 @@ func (s *sharesStorage) loadFromDB() error {
 		}
 
 		s.shares[hex.EncodeToString(val.ValidatorPubKey[:])] = share
+		// Owner index removed - not needed anymore
```
iurii-ssv: We probably can delete this comment (and another copy of it as well) ? Since it refers to something that doesn't exist - later on it might be confusing the figure out what it's about

registry/storage/shares.go
```diff
@@ -187,8 +200,16 @@ func (s *sharesStorage) loadFromDB() error {
...
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+
+	// Fee recipients are now loaded by ValidatorStore during handleSharesAdded
```
iurii-ssv: Is this comment correct ? I see that fee recipients are now loaded in `NewRecipientsStorage` ... and seems like it's better to place this comment on top of the following code line (or remove it entirely and let the reader follow the code):
```
...
return recipientReader.GetFeeRecipient(owner)
```

registry/storage/shares.go
```diff
@@ -308,6 +332,8 @@ func (s *sharesStorage) GetValidatorIndicesByPubkeys(vkeys []spectypes.Validator
 }
 
 func (s *sharesStorage) saveToDB(rw basedb.ReadWriter, shares ...*types.SSVShare) error {
+	// Fee recipients are no longer stored in shares - ValidatorStore manages them
+
```
iurii-ssv, line 325: Probably can just omit this comment (we have git history for that in case we need to figure out when/why/how the behavior changed), I'd rather comments describe current behavior that what we did in the past (assuming it bears no relevance for the current code)

registry/storage/shares.go
```diff
@@ -344,18 +370,17 @@ func FromSSVShare(share *types.SSVShare) *Share {
...
+		SharePubKey:     share.SharePubKey,
+		Committee:       committee,
+		DomainType:      share.DomainType,
+		Graffiti:        share.Graffiti,
+		OwnerAddress:    share.OwnerAddress,
+		Liquidated:      share.Liquidated,
+		Status:          uint64(share.Status), //nolint: gosec
+		ActivationEpoch: uint64(share.ActivationEpoch),
+		ExitEpoch:       uint64(share.ExitEpoch),
 	}
```
iurii-ssv, line 373: We can probably also delete `Share.FeeRecipientAddress` field now in this file since it's no longer used. But we'll probably need to re-generate SSZ encodings as well ... and then we won't be able to read the existing data from DB without a dedicated migration (decoding will fail since SSZ won't ignore absent fields but rather error, I think) ... which is probably not worth doing at this point - I would wait for at least 1 more change to justify such a migration.

Would be nice to post that ^ as a comment on top of `FeeRecipientAddress` field for future reference (so that we don't forget and accidentally remove it as unused).

registry/storage/shares.go
```diff
@@ -413,6 +437,8 @@ func (s *sharesStorage) Delete(rw basedb.ReadWriter, pubKey []byte) error {
 		// Remove the share from local storage map
 		delete(s.shares, hex.EncodeToString(pubKey))
 
+		// Owner index removed - not needed anymore
+
```
iurii-ssv: ditto (about comments describing past)

migrations/migration_6_share_exit_epoch.go
```diff
@@ -100,13 +100,12 @@ func mapShare(share *migration_6_OldStorageShare) *storage.Share {
...
-			FeeRecipientAddress: share.FeeRecipientAddress,
-			Graffiti:            share.Graffiti,
-			ValidatorIndex:      phase0.ValidatorIndex(share.ValidatorIndex),
+			ValidatorPubKey: validatorPubKey,
+			SharePubKey:     share.SharePubKey,
+			Committee:       committee,
+			DomainType:      share.DomainType,
+			Graffiti:        share.Graffiti,
+			ValidatorIndex:  phase0.ValidatorIndex(share.ValidatorIndex),
 		},
```
iurii-ssv: Maybe we should leave `migration_6` as it previously was (+ corresponding test as well) ? Otherwise this looks like a breaking change (for those who haven't applied it yet)  ?

registry/storage/shares.go
```diff
@@ -65,10 +73,11 @@ type Shares interface {
...
 type sharesStorage struct {
-	db             basedb.Database
-	storagePrefix  []byte
-	shares         map[string]*types.SSVShare
-	validatorStore *validatorStore
+	db              basedb.Database
+	storagePrefix   []byte
+	shares          map[string]*types.SSVShare
+	validatorStore  *validatorStore
+	recipientReader FeeRecipientReader
```
iurii-ssv: Looks like `recipientReader` is never used. `FeeRecipientReader` could probably be package-private as well ?

iurii-ssv (commented):
Awesome to finally see we got this fee-recipient stuff under control :) Just left some minor suggestions

operator/fee_recipient/controller.go
olegshmuelov: Good catch! I've updated it to maintain consistency with the codebase pattern - keeping go at the call site and removing the internal goroutine from `Start()`. This aligns with how other background services like `HandleMetadataUpdates`, `ReportValidatorStatuses`, and `DoppelgangerHandler` are launched in node.go. The pattern keeps all goroutine management visible at the orchestration level rather than hidden inside methods.

51008bea062adec7e847e43b48c4962027434fac

registry/storage/shares.go
olegshmuelov: https://github.com/ssvlabs/ssv/pull/2457/commits/1b745a79f3bdca160d0b0f89d40458cff28d3f63

migrations/migration_6_share_exit_epoch.go
olegshmuelov: https://github.com/ssvlabs/ssv/pull/2457/commits/1b745a79f3bdca160d0b0f89d40458cff28d3f63

registry/storage/shares.go
olegshmuelov: Good catch! You're right that recipientReader wasn't being used. This observation led to a nice improvement across the codebase:

  1. Removed the unused field from the `Recipients` struct
  2. Refactored from interface to function type: Changed from `feeRecipientReader` interface to `GetFeeRecipientFunc` function type, which is more idiomatic Go for single-method interfaces
  3. Eliminated unnecessary mocks: Removed registry/storage/mocks/recipients.go entirely since we now just pass simple test functions
  4. Simplified test code: Replaced mock implementations with lightweight dummy functions in tests

929d96614562747e117b15a71a01054d0f37c867

migrations/migration_6_share_exit_epoch.go
olegshmuelov: I've carefully reviewed this and we actually should remove the `FeeRecipientAddress` mapping from migration_6. Here's why:

  1. The test was already failing with the `FeeRecipientAddress` mapping in place because `FromSSVShare()` doesn't map this field anyway. The test at commit 1b745a79f3bdca160d0b0f89d40458cff28d3f63 shows this failure.
  2. It's not a breaking change for operators who haven't applied the migration yet:
    - The field is completely unused - operators get fee recipients from `Recipients` storage via `GetFeeRecipient()`, not from `Share` structs
    - Whether the field is empty or has a value doesn't matter since it's never read
    - The migration's purpose is to add `ExitEpoch` field, not manage fee recipients
  3. Why the test fails with the mapping:
  ```go
  // In migration_6, we set:
  FeeRecipientAddress: share.FeeRecipientAddress

  // But FromSSVShare() doesn't map it, so it becomes empty
  // The test then compares old.FeeRecipientAddress with new.FeeRecipientAddress
  // and fails because they don't match (value vs empty)
```
  4. The fix:
    - Remove the `FeeRecipientAddress` mapping from migration (it's unused anyway)
    - Update test to not compare this field
    - Add clear comments explaining this is safe

  This aligns with the broader refactoring where fee recipients are the responsibility of `Recipients` storage, not `Share` structs. The `FeeRecipientAddress` field only exists in `Share` for SSZ backward
  compatibility, but its value is meaningless.
  
  eb7f4e03ca94c2e6f335b71588e0de091eb58235

Comment (iurii-ssv) [reply migrations/migration_6_share_exit_epoch.go]: I see, makes sense 👍

File: operator/fee_recipient/controller.go (hunk 23 above)
Comment (iurii-ssv) [reply operator/fee_recipient/controller.go]: > The pattern keeps all goroutine management visible at the orchestration level rather than hidden inside methods.

Yeah, that's what originally seemed off about it ... good to have it done the same way across the codebase

File: operator/fee_recipient/controller_test.go (hunk 9 above)
Comment (oleg-ssvlabs) [reply operator/fee_recipient/controller_test.go]: I have a slightly different opinion on this — to me, it's in the category of "if you can do it, it doesn’t mean you should."  
If someone feels the need to test a private method, it’s usually a good sign that the code needs to be refactored, often so that the functionality becomes part of another object's public method. 

Although, I believe there is no right and wrong. If tests cover all important scenarios, simple and easy to understand - then it is probably less important how it was achieved.

Comment (iurii-ssv) [reply operator/fee_recipient/controller_test.go]: > to me, it's in the category of "if you can do it, it doesn’t mean you should."

I agree with that statement ^ but I think it's often important to take into considerations practical implications (how much work it is to do it this way vs that, and whether it's better to spend this time on improving something else)

Disclaimer: I haven't looked deep into what's going on with these particular unit-tests in this PR (so I'm speaking in general here)

File: migrations/migration_6_share_exit_epoch.go [hunk 35]
```diff
@@ -100,13 +100,16 @@ func mapShare(share *migration_6_OldStorageShare) *storage.Share {
...
-			ValidatorIndex:      phase0.ValidatorIndex(share.ValidatorIndex),
+			ValidatorPubKey: validatorPubKey,
+			SharePubKey:     share.SharePubKey,
+			Committee:       committee,
+			DomainType:      share.DomainType,
+			// FeeRecipientAddress is not mapped - it's now managed in Recipients storage.
+			// Removing this line doesn't affect the migration since the field is unused.
+			// Operators who haven't run this migration yet will not be affected - the field
+			// will simply remain empty in new shares, which is fine since it's not used.
+			Graffiti:       share.Graffiti,
```
oleg-ssvlabs, line 111: is there a chance we can remove `FeeRecipientAddress` field from `Share` object in `github.com/ssvlabs/ssv-spec` library? It would be good if it part of this PR (package update change)

registry/storage/recipients.go
olegshmuelov: The hybrid caching is intentional based on access patterns:

Why selective caching:
  - Fee recipients: Accessed hundreds of times per epoch (validator registrations + proposal preparations) -> cached to avoid repeated DB reads + GOB decoding
  - Nonces: Only accessed during `ValidatorAdded` events (rare) -> no performance benefit from caching

The "inconsistency" is actually intentional differentiation - fee recipients need performance optimization, nonces don't. This follows the principle of optimizing based on measured usage rather than theoretical consistency.

  Caching cold data (nonces) adds no value but increases complexity. The real win is avoiding repeated GOB decoding for hot-path operations while keeping the cold path simple.

migrations/migration_6_share_exit_epoch.go
olegshmuelov: The ssv-spec library is designed for the SSV specification/protocol definitions. The spec library is meant to define the protocol, not provide production-ready implementations. While we do use these types in production (by embedding them in our own types like SSVShare), we should be careful about which fields we expose and use.

As a suggestion, we could make the embedded `spectypes.Share` private and expose only the fields we actually use:
```go
type SSVShare struct {
    share spectypes.Share  // private

    // Status is validator (this share belongs to) state.
    Status eth2apiv1.ValidatorState
    // ...other fields...
}

// Getters for fields we actually use
func (s *SSVShare) GetValidatorIndex() phase0.ValidatorIndex { return s.share.ValidatorIndex }
func (s *SSVShare) GetValidatorPubKey() spectypes.ValidatorPK { return s.share.ValidatorPubKey }
// ...etc (excluding FeeRecipientAddress)

```
  This would:
  - Prevent any accidental usage of the deprecated `FeeRecipientAddress` field
  - Make it explicit which `spectypes.Share` fields our codebase actually uses
  - Better encapsulation

However, this would require updating all references throughout the codebase.

iurii-ssv: Yeah it's a bit unfortunate that implementations diverge over time, while we still are "forced to conform to certain structures defined in ssv-spec (such as `SSVShare` or `spec-tests`)".

I wonder if instead of getters + private `spectypes.Share` we could just fully re-define the `spectypes.Share` struct in the ssv repo (removing the `FeeRecipientAddress` field, maybe also removing some others):
- this way most of the calls to access fields on `SSVShare struct` will look the same (no need to adjust the code)
- we'll still need to replace the `spectypes.Share` usages in our code (mostly in Runner(s)), but it also doesn't seem too large of a task
- on getters, I feel like getter-pattern is a thing of the past (or at least not as common in Golang land) given how much of an additional work it requires - vs the benefits it brings; Getters that compute something (deriving stuff from struct-fields are still useful ofc to keep the code in one place)

I've created an issue for this here - https://github.com/ssvlabs/ssv/issues/2482 - maybe will give it a try later (definitely out of scope of this PR)

iurii-ssv (approved):
LGTM, just need to decide on this https://github.com/ssvlabs/ssv/pull/2457/files#r2306624623 I guess

oleg-ssvlabs (approved):

scripts/spec-alignment/differ.config.yaml
```diff
@@ -1,4 +1,4 @@
...
...3beaf077","2630b7d3f7463316","33dd6bfa496bc889","a9685eff5edb9118","2fd0165ccc59ddf","de0733313f796fc9","a5458fa616bc6d6d","73244bf5a8b49e23","42052e9213a5780d","26672cc9a6ae0af0","b7ca13196a824fc5","801f7d8a8151c73e","20b4765e446c4803","25c4b276cfa24926","cebe87e9824d213e","be65f009ecd0d4b8","4d1a0519b660e5f","ad8332e881c48041","fbcd6b75d8baa5c1","787162ea04b0c4ed","c97a6f1c933535b","b5b9b0a45f6df79e","e9569ea2dc0f1406","5a3c840d9fc7bec9","7cf54a264ab481f3","cbbb16ca9342dc4c","46636a126e9d32ad","c90ffdb18233f0ea","e5e2698cc827129b","b31079cc52d70820","99ebbf611546fd06","a6c82644fe7b2bf3","c110b52f3fc04964","b2ba608989ac677","373a9525ae8811a8","44a2015641f07e3","d6aafafb59a8dd6e","b7f2e4dec9742e39","bb626b923fe3850","b2315a5d4749f93","f1b684011e369981","fb2604dba9677890","b6908fa26415a154","70247f4b2e200f5b","d3ac29b25686ae94","636c42279c683d53","523ff663791a047a","cee19c2663445d45","819aa7ef21eb1f94","6e7d2035244a01b4","5a73c0f03029acba","96abe9e4636a4d4a","16f41dbe0a2ff32a","e16b89eb783f91bd","c0c8738f2d82c5bd","96c004530466d7ed","ecda760b0ddbdb11","db9d242dc4f80e0a","37176fea3b21eb63","7031f7740a5c0152","8cfc723c9abfd67e","9404f35d52cd90bc","f2eaa816c7adbbd1","db35cde6aead1669"]
```
olegshmuelov, line 1:
```diff
diff --git a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner b/ssv/validator_registration.go@NewValidatorRegistrationRunner
index 96c004530466d7ed..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@NewValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@NewValidatorRegistrationRunner
@@ -1,29 +1,25 @@
 func NewValidatorRegistrationRunner(
-	networkConfig *networkconfig.Network,
+	beaconNetwork BeaconNetwork,
 	share map[phase0.ValidatorIndex]*Share,
-	beacon beacon.BeaconNode,
+	beacon BeaconNode,
 	network Network,
-	signer ekm.BeaconSigner,
-	operatorSigner ssvtypes.OperatorSigner,
-	validatorRegistrationSubmitter ValidatorRegistrationSubmitter,
-	feeRecipientProvider feeRecipientProvider,
+	signer BeaconSigner,
+	operatorSigner *OperatorSigner,
 	gasLimit uint64,
 ) (Runner, error) {
 	if len(share) != 1 {
-		return nil, fmt.Errorf("must have one share")
+		return nil, errors.New("must have one share")
 	}
 	return &ValidatorRegistrationRunner{
 		BaseRunner: &BaseRunner{
 			RunnerRoleType:	RoleValidatorRegistration,
-			NetworkConfig:	networkConfig,
+			BeaconNetwork:	beaconNetwork,
 			Share:		share,
 		},
-		beacon:				beacon,
-		network:			network,
-		signer:				signer,
-		operatorSigner:			operatorSigner,
-		validatorRegistrationSubmitter:	validatorRegistrationSubmitter,
-		feeRecipientProvider:		feeRecipientProvider,
+		beacon:		beacon,
+		network:	network,
+		signer:		signer,
+		operatorSigner:	operatorSigner,
 		gasLimit:	gasLimit,
 	}, nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner b/ssv/validator_registration.go@ValidatorRegistrationRunner
index ecda760b0ddbdb11..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner
@@ -1,11 +1,9 @@
 ValidatorRegistrationRunner struct {
 	BaseRunner	*BaseRunner
-	beacon				beacon.BeaconNode
-	network				Network
-	signer				ekm.BeaconSigner
-	operatorSigner			ssvtypes.OperatorSigner
-	valCheck			ProposedValueCheckF
-	validatorRegistrationSubmitter	ValidatorRegistrationSubmitter
-	feeRecipientProvider		feeRecipientProvider
+	beacon		BeaconNode
+	network		Network
+	signer		BeaconSigner
+	operatorSigner	*OperatorSigner
+	valCheck	ProposedValueCheckF
 	gasLimit	uint64
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.GetRoot b/ssv/json_testutils.go@ValidatorRegistrationRunner.GetRoot
index db9d242dc4f80e0a..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.GetRoot
+++ b/ssv/json_testutils.go@ValidatorRegistrationRunner.GetRoot
@@ -1,8 +1,8 @@
 func (r *ValidatorRegistrationRunner) GetRoot() ([32]byte, error) {
 	marshaledRoot, err := r.Encode()
 	if err != nil {
-		return [32]byte{}, fmt.Errorf("could not encode ValidatorRegistrationRunner: %w", err)
+		return [32]byte{}, errors.Wrap(err, "could not encode ValidatorRegistrationRunner")
 	}
 	ret := sha256.Sum256(marshaledRoot)
 	return ret, nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessConsensus b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessConsensus
index 37176fea3b21eb63..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessConsensus
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessConsensus
@@ -1,3 +1,3 @@
 func (r *ValidatorRegistrationRunner) ProcessConsensus(signedMsg *SignedSSVMessage) error {
-	return fmt.Errorf("no consensus phase for validator registration")
+	return errors.New("no consensus phase for validator registration")
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessPostConsensus b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessPostConsensus
index 7031f7740a5c0152..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessPostConsensus
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessPostConsensus
@@ -1,3 +1,3 @@
 func (r *ValidatorRegistrationRunner) ProcessPostConsensus(signedMsg *PartialSignatureMessages) error {
-	return fmt.Errorf("no post consensus phase for validator registration")
+	return errors.New("no post consensus phase for validator registration")
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessPreConsensus b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessPreConsensus
index 8cfc723c9abfd67e..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.ProcessPreConsensus
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.ProcessPreConsensus
@@ -1,48 +1,34 @@
 func (r *ValidatorRegistrationRunner) ProcessPreConsensus(signedMsg *PartialSignatureMessages) error {
-	ctx, span := tracer.Start(
-		observability.InstrumentName(observabilityNamespace, "runner.process_pre_consensus"),
-		trace.WithAttributes(
-			observability.BeaconSlotAttribute(signedMsg.Slot),
-			observability.ValidatorPartialSigMsgTypeAttribute(signedMsg.Type),
-			observability.ValidatorPublicKeyAttribute(phase0.BLSPubKey(r.GetShare().ValidatorPubKey)),
-		))
-	defer span.End()
-	hasQuorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
+	quorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(r, signedMsg)
 	if err != nil {
-		return traces.Errorf(span, "failed processing validator registration message: %w", err)
+		return errors.Wrap(err, "failed processing validator registration message")
 	}
-	if !hasQuorum {
-		span.AddEvent("no quorum")
-		span.SetStatus(codes.Ok, "")
+	if !quorum {
 		return nil
 	}
 	root := roots[0]
-	span.AddEvent("reconstructing beacon signature", trace.WithAttributes(observability.BeaconBlockRootAttribute(root)))
 	fullSig, err := r.GetState().ReconstructBeaconSig(r.GetState().PreConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
 	if err != nil {
-		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
-		return traces.Errorf(span, "got pre-consensus quorum but it has invalid signatures: %w", err)
+		r.BaseRunner.FallBackAndVerifyEachSignature(r.GetState().PreConsensusContainer, root, r.GetShare().Committee,
+			r.GetShare().ValidatorIndex)
+		return errors.Wrap(err, "got pre-consensus quorum but it has invalid signatures")
 	}
 	specSig := phase0.BLSSignature{}
 	copy(specSig[:], fullSig)
-	registration, err := r.buildValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
+	registration, err := r.calculateValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
 	if err != nil {
-		return traces.Errorf(span, "could not calculate validator registration: %w", err)
+		return errors.Wrap(err, "could not calculate validator registration")
 	}
-	signedRegistration := &api.VersionedSignedValidatorRegistration{
+	signed := &api.VersionedSignedValidatorRegistration{
 		Version:	spec.BuilderVersionV1,
 		V1: &v1.SignedValidatorRegistration{
 			Message:	registration,
 			Signature:	specSig,
 		},
 	}
-	span.AddEvent("submitting validator registration")
-	if err := r.validatorRegistrationSubmitter.Enqueue(signedRegistration); err != nil {
-		return traces.Errorf(span, "could not submit validator registration: %w", err)
+	if err := r.beacon.SubmitValidatorRegistration(signed); err != nil {
+		return errors.Wrap(err, "could not submit validator registration")
 	}
-	const eventMsg = "validator registration submitted successfully"
-	span.AddEvent(eventMsg)
 	r.GetState().Finished = true
-	span.SetStatus(codes.Ok, "")
 	return nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.executeDuty b/ssv/validator_registration.go@ValidatorRegistrationRunner.executeDuty
index 9404f35d52cd90bc..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.executeDuty
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.executeDuty
@@ -1,53 +1,39 @@
 func (r *ValidatorRegistrationRunner) executeDuty(duty Duty) error {
-	_, span := tracer.Start(
-		observability.InstrumentName(observabilityNamespace, "runner.execute_duty"),
-		trace.WithAttributes(
-			observability.RunnerRoleAttribute(duty.RunnerRole()),
-			observability.BeaconSlotAttribute(duty.DutySlot())))
-	defer span.End()
-	vr, err := r.buildValidatorRegistration(duty.DutySlot())
+	vr, err := r.calculateValidatorRegistration(duty.DutySlot())
 	if err != nil {
-		return traces.Errorf(span, "could not calculate validator registration: %w", err)
-	}
-	span.AddEvent("signing beacon object")
-	msg, err := r.BaseRunner.signBeaconObject(
-		r,
-		duty.(*ValidatorDuty),
-		vr,
-		duty.DutySlot(),
-		DomainApplicationBuilder,
-	)
+		return errors.Wrap(err, "could not calculate validator registration")
+	}
+	msg, err := r.BaseRunner.signBeaconObject(r, duty.(*ValidatorDuty), vr, duty.DutySlot(),
+		DomainApplicationBuilder)
 	if err != nil {
-		return traces.Errorf(span, "could not sign validator registration: %w", err)
+		return errors.Wrap(err, "could not sign validator registration")
 	}
 	msgs := &PartialSignatureMessages{
 		Type:		ValidatorRegistrationPartialSig,
 		Slot:		duty.DutySlot(),
 		Messages:	[]*PartialSignatureMessage{msg},
 	}
-	msgID := NewMsgID(r.BaseRunner.NetworkConfig.DomainType, r.GetShare().ValidatorPubKey[:], r.BaseRunner.RunnerRoleType)
+	msgID := NewMsgID(r.GetShare().DomainType, r.GetShare().ValidatorPubKey[:], RunnerRole(r.BaseRunner.RunnerRoleType))
 	encodedMsg, err := msgs.Encode()
 	if err != nil {
-		return traces.Errorf(span, "could not encode validator registration partial sig message: %w", err)
+		return err
 	}
 	ssvMsg := &SSVMessage{
 		MsgType:	SSVPartialSignatureMsgType,
 		MsgID:		msgID,
 		Data:		encodedMsg,
 	}
-	span.AddEvent("signing SSV message")
 	sig, err := r.operatorSigner.SignSSVMessage(ssvMsg)
 	if err != nil {
-		return traces.Errorf(span, "could not sign SSVMessage: %w", err)
+		return errors.Wrap(err, "could not sign SSVMessage")
 	}
 	msgToBroadcast := &SignedSSVMessage{
 		Signatures:	[][]byte{sig},
 		OperatorIDs:	[]OperatorID{r.operatorSigner.GetOperatorID()},
 		SSVMessage:	ssvMsg,
 	}
-	if err := r.GetNetwork().Broadcast(msgID, msgToBroadcast); err != nil {
-		return traces.Errorf(span, "can't broadcast partial randao sig: %w", err)
+	if err := r.GetNetwork().Broadcast(msgToBroadcast.SSVMessage.GetID(), msgToBroadcast); err != nil {
+		return errors.Wrap(err, "can't broadcast partial randao sig")
 	}
-	span.SetStatus(codes.Ok, "")
 	return nil
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.expectedPostConsensusRootsAndDomain b/ssv/validator_registration.go@ValidatorRegistrationRunner.expectedPostConsensusRootsAndDomain
index f2eaa816c7adbbd1..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.expectedPostConsensusRootsAndDomain
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.expectedPostConsensusRootsAndDomain
@@ -1,3 +1,3 @@
 func (r *ValidatorRegistrationRunner) expectedPostConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error) {
-	return nil, [4]byte{}, fmt.Errorf("no post consensus roots for validator registration")
+	return nil, [4]byte{}, errors.New("no post consensus roots for validator registration")
 }
\ No newline at end of file
diff --git a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.expectedPreConsensusRootsAndDomain b/ssv/validator_registration.go@ValidatorRegistrationRunner.expectedPreConsensusRootsAndDomain
index db35cde6aead1669..2222222
--- a/protocol/v2/ssv/runner/validator_registration.go@ValidatorRegistrationRunner.expectedPreConsensusRootsAndDomain
+++ b/ssv/validator_registration.go@ValidatorRegistrationRunner.expectedPreConsensusRootsAndDomain
@@ -1,10 +1,10 @@
 func (r *ValidatorRegistrationRunner) expectedPreConsensusRootsAndDomain() ([]ssz.HashRoot, phase0.DomainType, error) {
 	if r.BaseRunner.State == nil || r.BaseRunner.State.StartingDuty == nil {
-		return nil, DomainError, fmt.Errorf("no running duty to compute preconsensus roots and domain")
+		return nil, DomainError, errors.New("no running duty to compute preconsensus roots and domain")
 	}
-	vr, err := r.buildValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
+	vr, err := r.calculateValidatorRegistration(r.BaseRunner.State.StartingDuty.DutySlot())
 	if err != nil {
-		return nil, DomainError, fmt.Errorf("could not calculate validator registration: %w", err)
+		return nil, DomainError, errors.Wrap(err, "could not calculate validator registration")
 	}
 	return []ssz.HashRoot{vr}, DomainApplicationBuilder, nil
 }
\ No newline at end of file
```

protocol/v2/types/ssvshare.go
```diff
@@ -25,6 +25,13 @@ const (
 )
 
 // SSVShare is a spectypes.Share with extra data that fully describes SSV validator share.
+//
+// WARNING: Do NOT use FeeRecipientAddress from spectypes.Share - it's deprecated.
```
nkryuchkov, line 29: It's still used in `migration_5_gob.go`, do we need to remove it from there?

migrations/migration_6_share_exit_epoch.go
nkryuchkov: Probably it would be easier to get rid of `spectypes.Share` in the implementation — anyway, we already extended it to `types.SSVShare`, which means it doesn't suit our needs. We could just have everything we need in `types.SSVShare`

operator/dutytracer/eviction_test.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 func TestEviction(t *testing.T) {
+
```
nkryuchkov: IIRC, it will trigger the linter after we merge it

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -259,20 +253,16 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
+	feeRecipient, err := r.feeRecipientProvider.GetFeeRecipient(validatorPubKey)
 	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
-	}
-	if !found {
-		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
+		return nil, fmt.Errorf("could not get fee recipient: %w", err)
 	}
 
+	pk := phase0.BLSPubKey(validatorPubKey)
```
nkryuchkov: Can we just inline it?

registry/storage/shares.go
```diff
@@ -89,11 +94,14 @@ const addressLength = 20
...
-	DomainType          [4]byte            `ssz-size:"4"`
+	ValidatorIndex  uint64
+	ValidatorPubKey []byte             `ssz-size:"48"`
+	SharePubKey     []byte             `ssz-max:"48"` // empty for not own shares
+	Committee       []*storageOperator `ssz-max:"13"`
+	DomainType      [4]byte            `ssz-size:"4"`
+	// FeeRecipientAddress is unused - fee recipients are stored in Recipients storage.
+	// DO NOT REMOVE: Removing this field would break SSZ decoding of existing DB data.
+	// Wait for a major migration to batch this removal with other breaking changes.
 	FeeRecipientAddress [addressLength]byte
```
nkryuchkov: can we add this to let Goland know it's deprecated? 
```suggestion
	FeeRecipientAddress [addressLength]byte // DEPRECATED
```

registry/storage/shares.go
```diff
@@ -187,8 +196,14 @@ func (s *sharesStorage) loadFromDB() error {
...
 
 		s.shares[hex.EncodeToString(val.ValidatorPubKey[:])] = share
+
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
nkryuchkov: can we get rid of this?

```go
func (s *sharesStorage) loadFromDB() error {
	return s.db.GetAll(SharesDBPrefix(s.storagePrefix), func(i int, obj basedb.Obj) error {
		val := &Share{}
		if err := val.Decode(obj.Value); err != nil {
			return fmt.Errorf("failed to deserialize share: %w", err)
		}
		share, err := ToSSVShare(val)
		if err != nil {
			return fmt.Errorf("failed to convert storage share to domain share: %w", err)
		}

		s.shares[hex.EncodeToString(val.ValidatorPubKey[:])] = share

		return nil
	})
}
```

registry/storage/validatorstore_test.go
```diff
@@ -1109,7 +1102,326 @@ func createValidatorStore(shares map[spectypes.ValidatorPK]*ssvtypes.SSVShare) *
 			}
 			return share, true
 		},
+		func(owner common.Address) (bellatrix.ExecutionAddress, error) {
+			// For tests, just return owner address as fee recipient
+			var feeRecipient bellatrix.ExecutionAddress
+			copy(feeRecipient[:], owner.Bytes())
+			return feeRecipient, nil
```
nkryuchkov:
```suggestion
			return bellatrix.ExecutionAddress(owner), nil
```

nkryuchkov (approved):

migrations/migration_6_share_exit_epoch.go
iurii-ssv: > Probably it would be easier to get rid of `spectypes.Share` in the implementation — anyway, we already extended it to `types.SSVShare`, which means it doesn't suit our needs. We could just have everything we need in `types.SSVShare`

Just to confirm, that matches what I suggested above (stop using `spectypes.Share` anywhere and just move the fields we use over onto `SSVShare` struct) ?

nkryuchkov: > > Probably it would be easier to get rid of `spectypes.Share` in the implementation — anyway, we already extended it to `types.SSVShare`, which means it doesn't suit our needs. We could just have everything we need in `types.SSVShare`
> 
> Just to confirm, that matches what I suggested above (stop using `spectypes.Share` anywhere and just move the fields we use over onto `SSVShare` struct) ?

yes

protocol/v2/types/ssvshare.go
olegshmuelov: You're right that it's still in `migration_5_gob.go`. However, we must keep it there because:

  1. GOB decoding requires exact struct match - removing the field would break migration for users who haven't migrated yet
  2. Migration 6 handles it differently - we keep the field in the struct (for SSZ compatibility) but don't copy the value, so it becomes empty/zero (more [context](https://github.com/ssvlabs/ssv/pull/2457#discussion_r2308492672))
  3. This is intentional - Migration 5 must read the old data (including the unused `FeeRecipientAddress` field), while Migration 6 onwards leaves it empty since fee recipients have always been stored in Recipients storage (not in `Share`)

The difference is intentional: GOB requires the field for decoding, while in migration 6 we just don't populate this already-unused field.

protocol/v2/ssv/runner/validator_registration.go
olegshmuelov: yeah, see [here](https://github.com/ssvlabs/ssv/pull/2457#discussion_r2307680747)

protocol/v2/types/ssvshare.go
nkryuchkov: @olegshmuelov thanks for the explanation

registry/storage/shares.go
nkryuchkov: This is what it would look like 

image

nkryuchkov (approved):

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -259,18 +253,12 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
...
-	epoch := r.BaseRunner.NetworkConfig.EstimatedEpochAtSlot(slot)
-
-	rData, found, err := r.recipientsStorage.GetRecipientData(nil, r.validatorOwnerAddress)
+	feeRecipient, err := r.feeRecipientProvider.GetFeeRecipient(validatorPubKey)
 	if err != nil {
-		return nil, fmt.Errorf("get recipient data from storage: %w", err)
-	}
-	if !found {
-		return nil, fmt.Errorf("recipient data not found for owner %s", r.validatorOwnerAddress.Hex())
+		return nil, fmt.Errorf("could not get fee recipient: %w", err)
```
iurii-ssv: The error message here could maybe use a `validatorPubKey` in it ?

iurii-ssv (dismissed):

oleg-ssvlabs (dismissed):

oleg-ssvlabs (dismissed):

iurii-ssv (dismissed):

nkryuchkov (approved):

PR #2460, opened 2025-08-25
feat(exporter): enable partial results

julienh-ssv:
Gracefully degrade list endpoints: aggregate entry errors and respond with partial data rather than failing the entire request.

### DIFF
--- a/api/handlers/exporter.go
+++ b/api/handlers/exporter.go
@@ -212,11 +212,7 @@ func (e *Exporter) TraceDecideds(w http.ResponseWriter, r *http.Request) error {
 
 	// if we don't have a single valid participant, return an error
 	if len(participants) == 0 && errs.ErrorOrNil() != nil {
-		// if we only have one error, unwrap it
-		if len(errs.Errors) == 1 {
-			return api.Error(errs.Errors[0])
-		}
-		return api.Error(errs)
+		return toApiError(errs)
 	}
 
 	// otherwise return a partial response with valid participants
@@ -280,10 +276,10 @@ func (e *Exporter) getCommitteeDutiesForSlot(slot phase0.Slot, committeeIDs []sp
 	for _, cmtID := range committeeIDs {
 		duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID)
 		if err != nil {
-			e.logger.Debug("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
 			// if error is not found, nothing to report as we might not have a duty for this role
 			// otherwise report it:
 			if !isNotFoundError(err) {
+				e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
 				errs = multierror.Append(errs, err)
 			}
 			continue
@@ -391,8 +387,7 @@ func (e *Exporter) extractPubKeys(request *validatorRequest) ([]spectypes.Valida
 
 func (e *Exporter) getValidatorDutiesForRoleAndSlot(role spectypes.BeaconRole, slot phase0.Slot, validatorPks []spectypes.ValidatorPK) ([]*dutytracer.ValidatorDutyTrace, error) {
 	if len(validatorPks) == 0 {
-		duties, err := e.traceStore.GetValidatorDuties(role, slot)
-		return duties, err
+		return e.traceStore.GetValidatorDuties(role, slot)
 	}
 
 	duties := make([]*dutytracer.ValidatorDutyTrace, 0, len(validatorPks))
@@ -401,10 +396,10 @@ func (e *Exporter) getValidatorDutiesForRoleAndSlot(role spectypes.BeaconRole, s
 	for _, pubkey := range validatorPks {
 		duty, err := e.traceStore.GetValidatorDuty(role, slot, pubkey)
 		if err != nil {
-			e.logger.Debug("error getting validator duty", zap.Error(err), fields.Slot(slot), fields.Validator(pubkey[:]))
 			// if error is not found, nothing to report as we might not have a duty for this role
 			// otherwise report it:
 			if !isNotFoundError(err) {
+				e.logger.Error("error getting validator duty", zap.Error(err), fields.Slot(slot), fields.Validator(pubkey[:]))
 				errs = multierror.Append(errs, err)
 			}
 			continue
@@ -428,10 +423,10 @@ func (e *Exporter) getValidatorCommitteeDutiesForRoleAndSlot(role spectypes.Beac
 
 		duty, err := e.traceStore.GetCommitteeDuty(slot, committeeID, role)
 		if err != nil {
-			e.logger.Debug("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.BeaconRole(role), fields.PubKey(pubkey[:]))
 			// if error is not found, nothing to report as we might not have a duty for this role
 			// otherwise report it:
 			if !isNotFoundError(err) {
+				e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.BeaconRole(role), fields.PubKey(pubkey[:]))
 				errs = multierror.Append(errs, err)
 			}
 			continue
@@ -487,10 +482,7 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
 	}
 
 	if len(results) == 0 && errs.ErrorOrNil() != nil {
-		if len(errs.Errors) == 1 {
-			return api.Error(errs.Errors[0])
-		}
-		return api.Error(errs)
+		return toApiError(errs)
 	}
 	return api.Render(w, r, toValidatorTraceResponse(results, errs))
 }
@@ -511,4 +503,11 @@ func toValidatorTraceResponse(duties []*dutytracer.ValidatorDutyTrace, errs *mul
 		r.Errors = toStrings(errs.Errors)
 	}
 	return r
+}
+
+func toApiError(errs *multierror.Error) *api.ErrorResponse {
+	if len(errs.Errors) == 1 {
+		return api.Error(errs.Errors[0])
+	}
+	return api.Error(errs)
 }


exporter/store/store.go
```diff
@@ -67,56 +70,70 @@ func (s *DutyTraceStore) SaveValidatorDuties(duties []*model.ValidatorDutyTrace)
...
+	ctx := fmt.Sprintf("role=%s slot=%d index=%d", role, slot, index)
+
 	obj, found, err := s.db.Get(prefix, nil)
 	if err != nil {
-		return nil, fmt.Errorf("get validator duty: %w", err)
+		return nil, fmt.Errorf("get validator duty (%s): %w", ctx, err)
 	}
 	if !found {
-		return nil, ErrNotFound
+		return nil, fmt.Errorf("get validator duty (%s): %w", ctx, ErrNotFound)
```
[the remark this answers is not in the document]
julienharbulot: False positive.

Using fmt.Errorf("...: %w", ErrNotFound) wraps ErrNotFound and preserves the error chain

api/handlers/exporter.go
```diff
@@ -51,28 +52,10 @@ type dutyTraceStore interface {
...
-	var request struct {
-		From    uint64        `json:"from"`
-		To      uint64        `json:"to"`
-		Roles   api.RoleSlice `json:"roles"`
-		PubKeys api.HexSlice  `json:"pubkeys"`
-	}
-	var response struct {
-		Data []*ParticipantResponse `json:"data"`
-	}
+// === Decideds ======================================================================================
```
oleg-ssvlabs, line 55: Instead of separating the file into sections like this, I would rather consider one of the following approaches:

- Restructure files so that each file contains everything related to a specific endpoint request (primarily handler + model).
- Create new packages (per endpoint handler) under `api/handlers`.

Option 1:
```
handlers/
  decideds.go
  committee_traces.go
  validator_traces.go
```
  
Option 2:
```
handlers/
  decideds/
    handler.go
    validation.go
    model.go
  committee_traces/
   ...

File: api/handlers/exporter.go [hunk 3]
```diff
@@ -51,28 +52,10 @@ type dutyTraceStore interface {
...
 
-	if err := api.Bind(r, &request); err != nil {
+func parseAndValidateDecidedRequest(r *http.Request, request *decidedRequest) error {
```
oleg-ssvlabs: I would rather not put the responsibility on this method to return API-related errors.  
It is better if it simply returns errors, and then the caller (handler methods like `Decideds` and `TraceDecideds`) decides what to do with them (e.g. wrap into API `400` errors).

oleg-ssvlabs: Also, since it "parses and validates decided request" it probably would make more sense if it wasn't taking `request *decidedRequest` as parameter, but rather return it. So the function signature could look something like this:

```Go
parseAndValidateDecidedRequest(r *http.Request) (decidedRequest, error)
```
Wdyt?

Another idea is that we could potentially just duplicate the parsing (it’s called from two methods) - it’s only two lines and make this function responsible only for validation:  

`func validateDecidedRequest(request decidedRequest) error`

api/handlers/exporter.go
```diff
@@ -84,19 +67,30 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
-	from := phase0.Slot(request.From)
-	to := phase0.Slot(request.To)
-
-	pubkeys := make([]spectypes.ValidatorPK, 0, len(request.PubKeys))
-	for i, req := range request.PubKeys {
-		var pubkey spectypes.ValidatorPK
-		if len(req) != len(pubkey) {
-			return api.BadRequestError(fmt.Errorf("invalid pubkey length at index %d", i))
+	requiredLength := len(spectypes.ValidatorPK{})
+	for _, req := range request.PubKeys {
```
oleg-ssvlabs, line 67: It seems [previously](https://github.com/ssvlabs/ssv/blob/stage/api/handlers/exporter.go#L92) we performed both validation and pubkey parsing within a single iteration, but now it iterates twice. Do you know the size of the `PubKeys` collection? If it contains many items, this could potentially slow down request processing.

api/handlers/exporter.go
```diff
@@ -122,138 +116,116 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
-	var request struct {
-		From    uint64        `json:"from"`
-		To      uint64        `json:"to"`
-		Roles   api.RoleSlice `json:"roles"`
-		PubKeys api.HexSlice  `json:"pubkeys"`
-	}
-	var response struct {
-		Data []*ParticipantResponse `json:"data"`
-	}
+func noSignersError(pubkey spectypes.ValidatorPK, slot phase0.Slot, role spectypes.BeaconRole) error {
```
oleg-ssvlabs: from what I understand - this is the new error that API can return. I think there is not test coverage for it?

api/handlers/model.go
```diff
@@ -9,11 +9,45 @@ import (
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
+	"github.com/ssvlabs/ssv/api"
 	model "github.com/ssvlabs/ssv/exporter"
 )
 
+// === Decideds ======================================================================
+type ParticipantResponse struct {
```
oleg-ssvlabs: I know this model was exported before, but it seems like it shouldn’t be, since it is only used within this package.

api/handlers/model.go
```diff
@@ -9,11 +9,45 @@ import (
...
+		copy(pubKeys[i][:], pk)
+	}
+	return pubKeys
+}
+
+// === ValidatorTrace ======================================================================
 type validatorTraceResponse struct {
-	Data []validatorTrace `json:"data"`
+	Data   []validatorTrace `json:"data"`
+	Errors []string         `json:"errors,omitempty"`
```
oleg-ssvlabs, line 59: Not in the scope of this PR, but I think it would be good to have an OpenAPI spec for all SSV endpoints. I don’t think we have one today. @MatusKysel @iurii-ssv @julienh-ssv wdyt?

oleg-ssvlabs (commented):
really good changes.
I left few comments

api/handlers/exporter.go
iurii-ssv: Typically separating by file is common for endpoints, I think. But whatever we end up doing is better than no separation at all.

api/handlers/exporter.go
iurii-ssv: Upvote these suggestions, maybe it's better to split `parseAndValidateDecidedRequest` into 2 methods

api/handlers/model.go
iurii-ssv: I think it depends on "if there is a target audience to justify the effort", I don't exactly know if our users would like this feature

IF the API is large enough - perhaps it makes sense to use OpenAPI spec from the very start to generate most of the handler-code (so we can only fill in the implementation details) ... doing this retrospectively though is likely too much effort for no real benefit

api/handlers/exporter.go
julienharbulot: Not really, because we’re not duplicating work: each pass does a different task. It’s still n + n rather than 2n of the same work. The cost of the for _, req := range ... loop itself is negligible in Go.

And even if we disagreed on that, with ~10k pubkeys the overhead of an extra iteration is negligible compared to the network I/O for fetching validators.

I think the benefit of clearly separating validation from processing outweighs any performance concern here — the extra pass is negligible compared to the network I/O, and the separation keeps the code easier to read and maintain.

api/handlers/exporter.go
```diff
@@ -365,97 +362,145 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
...
 	})
 	pubkeys = slices.Compact(pubkeys)
 
+	return pubkeys, errs.ErrorOrNil()
+}
+
+func (e *Exporter) getValidatorDutiesForRoleAndSlot(role spectypes.BeaconRole, slot phase0.Slot, validatorPks []spectypes.ValidatorPK) ([]*dutytracer.ValidatorDutyTrace, error) {
+	if len(validatorPks) == 0 {
+		duties, err := e.traceStore.GetValidatorDuties(role, slot)
+		return duties, err
```
iurii-ssv: Could just `return e.traceStore.GetValidatorDuties(role, slot)` here

api/handlers/exporter.go
```diff
@@ -365,97 +362,145 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
...
+	for _, pubkey := range validatorPks {
+		duty, err := e.traceStore.GetValidatorDuty(role, slot, pubkey)
+		if err != nil {
+			e.logger.Debug("error getting validator duty", zap.Error(err), fields.Slot(slot), fields.Validator(pubkey[:]))
+			// if error is not found, nothing to report as we might not have a duty for this role
+			// otherwise report it:
+			if !isNotFoundError(err) {
+				errs = multierror.Append(errs, err)
+			}
+			continue
```
iurii-ssv: Probably want to log the error if it's `!isNotFoundError(err)`, and also maybe it's better to log it under `Error` (not `Debug`) ?

I guess applies to other places in this PR

api/handlers/exporter.go
julienharbulot: Thanks for the suggestion — I agree that grouping by endpoint/package can make sense once things get larger. At the moment though, each section is only about 5/6 functions, so splitting into multiple files feels like premature structure and would add overhead without much benefit.
For now, I’d prefer to keep everything in one file with simple section comments for readability. If the code grows beyond that, I’m fully on board with reorganizing into separate files or packages.

api/handlers/exporter.go
julienharbulot: Actually I didn't need to create a function for that, there's nothing special about this error compared to any other. I removed it, thanks for pointing it out.

api/handlers/model.go
julienharbulot: good catch

api/handlers/exporter.go
julienharbulot: Good idea I've split the parsing out and changed the error types

api/handlers/exporter.go
```diff
@@ -365,97 +362,145 @@ func (e *Exporter) ValidatorTraces(w http.ResponseWriter, r *http.Request) error
...
 		}
 	}
 
-	return api.Render(w, r, toValidatorTraceResponse(results))
+	if len(results) == 0 && errs.ErrorOrNil() != nil {
+		if len(errs.Errors) == 1 {
+			return api.Error(errs.Errors[0])
+		}
+		return api.Error(errs)
+	}
```
iurii-ssv: This error-handling is kinda ugly, whatever, but since it's duplicated elsewhere perhaps we should have a helper **func** for that code (maybe something like `api.ErrorFromErrors`) ?

iurii-ssv (commented):
Just minor suggestions

api/handlers/exporter.go
julienharbulot: ok

api/handlers/exporter.go
julienharbulot: good idea

iurii-ssv (approved):
LGTM

api/handlers/exporter.go
oleg-ssvlabs: @julienh-ssv what I meant to say is that previously the condition, which now creates an error, was processed [differently](https://github.com/ssvlabs/ssv/blob/stage/api/handlers/exporter.go#L245).  
So it would be good to add a unit test for it. I did not see a test that asserts this new error message.

julienharbulot: @oleg-ssvlabs according to the coverage report this part is already exercised (as shown by the green gutter in the screenshot).

I don’t think we need a dedicated unit test just to check that an error is added to the multi-error. It’s only used for reporting, and the logic doesn’t change any functional behavior. Compared to before, we just surface more details. Testing it explicitly would be similar to writing unit tests for every log line—which we typically don’t do.

Screenshot 2025-08-26 at 4 58 58 pm

iurii-ssv: Just my 2 cents on the general approach (not sure how exactly it applies here), 
- I often find writing unit-tests isn't worth the time (that could be spend on doing reviews/self-reviews or fixing bugs we have elsewhere)
- this is especially true for the code that changes a lot (and you have to rewrite unit-tests over and over as well)
- coverage often doesn't tell us much (I could write a test that has 100% coverage but doesn't assert on anything, and that would be meaningless)

with that in mind, I would say it's worth writing unit-tests:
- for the functionality that we don't expect to change much (so we have some safety-net that would catch accidental modifications)
- for the functionality that has some non-trivial algo involved in it (like computing something), or if there are doubts about how it works (eg. using 3rd party code that you think works certain way but would rather verify with a test)

oleg-ssvlabs (approved):

api/handlers/exporter.go
oleg-ssvlabs: I do not have a very strong opinion here, hence I resolved the conversation.  
However, this particular error will end up as part of the API response.  
While some conditions may not necessarily require testing, it is good to test the outputs, and in this case, the output is the API response.

PR #2463, opened 2025-08-26
pipelines: run CI on merge into stage/main branches

iurii-ssv:
This is a follow-up to https://github.com/ssvlabs/ssv/pull/2362 that should enable pipeline-runs triggered by PR merges (or regular pushes) into **stage** / **main** branches, note:
- after https://github.com/ssvlabs/ssv/pull/2362 has been merged into **stage** we stopped running all CI pipelines there - as can be seen [here](https://github.com/ssvlabs/ssv/commits/stage/)
- https://github.com/ssvlabs/ssv/pull/2362 hasn't made it into **main** branch yet, hence we still [do run pipelines on merges](https://github.com/ssvlabs/ssv/commits/main/) in there
- removing `integration-test` pipeline since we are not using it anymore (see also https://github.com/ssvlabs/ssv/pull/2464)
- updating `ssvsigner-e2e-test.yml` workflow to run pipelines in the same way we run them for **ssv** repo
- as far as I can tell, this PR shouldn't result in duplicate pipelines running anywhere (meaning it shouldn't introduce the issue https://github.com/ssvlabs/ssv/pull/2362 was targeting), but even if it does on PR-merges that's still worth it to be able to spot compilation issues (or other incompatibilities) ASAP


vaclav-ssvlabs (approved):

oleg-ssvlabs (approved):
Good stuff.  

What we could potentially also add is cancellation of workflows that are already running when new triggers occur:  

Something like this:  

```yaml
concurrency:
  group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

PR #2464, opened 2025-08-26
cleanup: remove integration tests

iurii-ssv:
It looks like we are no longer using these integration tests - so I think we can do this clean-up now.


oleg-ssvlabs (approved):
lgtm

PR #2466, opened 2025-08-27
fix(threshold): Add validation to threshold signature split

AgeManning:
We are running a differential fuzzer and in the process fixing issues as they come. 

The go code crashes with an index out of bounds if the threshold is set at 0. I realize this is an unrealistic parameter, but figure it's worth handling the edge case.

### DIFF
--- a/utils/threshold/threshold.go
+++ b/utils/threshold/threshold.go
@@ -9,6 +9,17 @@ import (
 // Create receives a bls.SecretKey hex and count.
 // Will split the secret key into count shares
 func Create(skBytes []byte, threshold uint64, count uint64) (map[uint64]*bls.SecretKey, error) {
+
+	// Validate threshold parameter - must be at least 2 for meaningful threshold schemes
+	if threshold <= 1 {
+		return nil, fmt.Errorf("invalid threshold: threshold must be greater than 1, got %d", threshold)
+	}
+
+	// Validate that we have enough shares for the threshold
+	if count < threshold {
+		return nil, fmt.Errorf("insufficient count: need at least %d shares for threshold %d, got %d", threshold, threshold, count)
+	}
+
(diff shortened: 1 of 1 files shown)


nkryuchkov (approved):

utils/threshold/threshold.go
```diff
@@ -9,6 +9,17 @@ import (
...
 func Create(skBytes []byte, threshold uint64, count uint64) (map[uint64]*bls.SecretKey, error) {
+
+	// Validate threshold parameter - must be at least 2 for meaningful threshold schemes
+	if threshold <= 1 {
+		return nil, fmt.Errorf("invalid threshold: threshold must be greater than 1, got %d", threshold)
+	}
+
+	// Validate that we have enough shares for the threshold
+	if count < threshold {
+		return nil, fmt.Errorf("insufficient shares: need at least %d shares for threshold %d, got %d", threshold, threshold, count)
```
iurii-ssv: Nit: `fmt.Errorf("insufficient shares: need at least %d shares for threshold %d, got %d", threshold, threshold, count)` -> `fmt.Errorf("insufficient shares count: need at least %d shares for threshold of %d, got %d shares", threshold, threshold, count)`

iurii-ssv (approved):
LGTM, but the linter needs to be fixed (no need for additional newline)

utils/threshold/threshold.go
iurii-ssv: @AgeManning any chance you can fix that ?

AgeManning: Yep, sorry for the delay, have broken a finger since writing this.

iurii-ssv (approved):
Thank you!

y0sher (approved):
thanks for this @AgeManning !

PR #2468, opened 2025-08-28
execution-client: add timeouts

iurii-ssv:
This PR sets universal default timeout for every call we make with `ExecutionClient`. This is needed because certain calls can hang for minutes without explicitly enforced timeout (defaulting to TCP-enforced timeout, I believe), and (as recently observed) this further prevents `ExecutionClient` from failing over to another potentially functioning EL (in case multiple ELs are configured for SSV node).

Additionally, while going over this code I've:
- improved logging to make sure EL `address` is added to every erroring client-call we might encounter
- plus removed redundant logging for every erroring client-call (previously we would both "log an error" and "return it to the caller", resulting in 2 log lines for every error we get that are also differing in the actual data they log)


eth/executionclient/defaults.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	DefaultConnectionTimeout           = 10 * time.Second
+	DefaultReqTimeout                  = 15 * time.Second
```
iurii-ssv, line 8: Bumping this value `10s`->`15s` just to be on the safe side, since now it applies to every request `ExecutionClient` makes,

@nkryuchkov @kchojn I think you guys would know better if universal default value of `15s` here makes sense ... or if maybe some specific calls require more/less time ?

eth/executionclient/eth_client.go
```diff
@@ -0,0 +1,81 @@
...
+	// reqTimeout specifies the default timeout to be used with every ethclient.Client call.
+	reqTimeout time.Duration
+}
+
+func newEthClient(client *ethclient.Client, reqTimeout time.Duration) (*ethClient, error) {
+	return &ethClient{
+		client:     client,
+		reqTimeout: reqTimeout,
+	}, nil
+}
```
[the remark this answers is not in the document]
iurii-ssv: Fixed

eth/executionclient/options.go
```diff
@@ -44,17 +44,21 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
...
+		if timeout > 0 {
+			s.reqTimeout = timeout
+		}
 	}
 }
 
-// WithConnectionTimeoutMulti sets timeout for network connection to eth1 node.
-func WithConnectionTimeoutMulti(timeout time.Duration) OptionMulti {
+// WithReqTimeoutMulti sets timeout for network connection to eth1 node.
+func WithReqTimeoutMulti(timeout time.Duration) OptionMulti {
```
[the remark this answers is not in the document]
iurii-ssv: Applied

eth/executionclient/options.go
```diff
@@ -44,17 +44,21 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
...
 
-// WithConnectionTimeout sets timeout for network connection to eth1 node.
-func WithConnectionTimeout(timeout time.Duration) Option {
+// WithReqTimeout sets timeout for RPC requests to eth1 node.
+func WithReqTimeout(timeout time.Duration) Option {
 	return func(s *ExecutionClient) {
-		s.connectionTimeout = timeout
+		if timeout > 0 {
+			s.reqTimeout = timeout
+		}
```
iurii-ssv, line 52: This is an additional safety mechanism because currently we rely on `env-default:"10s"` for timeouts to not be set to 0 by default, which is super unreliable given how far it's configured from the place of actual use (we can easily remove it in the future without noticing)

eth/executionclient/options.go
```diff
@@ -44,17 +44,21 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
 	}
 }
 
-// WithConnectionTimeout sets timeout for network connection to eth1 node.
-func WithConnectionTimeout(timeout time.Duration) Option {
+// WithReqTimeout sets timeout for RPC requests to eth1 node.
+func WithReqTimeout(timeout time.Duration) Option {
```
iurii-ssv, line 48: These funcs are renamed to better reflect that timeout applies to every request `ExecutionClient` makes, rather than to just connect-request.

eth/executionclient/execution_client.go
```diff
@@ -361,17 +351,11 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
...
 	start := time.Now()
 	sp, err := ec.SyncProgress(ctx)
 	if err != nil {
 		recordExecutionClientStatus(ctx, statusFailure, ec.nodeAddr)
-		ec.logger.Error(elResponseErrMsg,
-			zap.String("method", "eth_syncing"),
-			zap.Error(err))
-		return err
+		return ec.errWithDetails(fmt.Errorf("get sync progress: %w", err), "eth_syncing")
 	}
```
iurii-ssv, line 359: Just as an example: `ec.SyncProgress` (the underlying `ethClient.SyncProgress` call) now will take care of the timeout, hence we don't need to `ctx, cancel := context.WithTimeout(ctx, ec.connectionTimeout)` here

eth/executionclient/eth_client.go
```diff
@@ -0,0 +1,81 @@
...
+
+// ethClient is a wrapper around ethclient.Client to add a default timeout for every call we
+// make with ethclient.Client. This allows us to manage timeouts for ethclient.Client in one
+// place such that we don't accidentally forget to set a timeout for some ethclient.Client call.
+type ethClient struct {
+	client *ethclient.Client
+
+	// reqTimeout specifies the default timeout to be used with every ethclient.Client call.
+	reqTimeout time.Duration
+}
```
iurii-ssv, line 21: Note, this allows us to set custom timeout depending on which method we are calling,  if we only need a default timeout that is the same for every method-call we could do [this instead](https://github.com/ssvlabs/ssv/commit/4c4a5ccb7983147c4b4ae6951aa1faee46eda218) ... but either approach is fine IMO (I wouldn't bother switching unless there are other benefits)

beacon/goclient/goclient.go
```diff
@@ -171,8 +171,15 @@ func New(
 	logger *zap.Logger,
 	opt Options,
 ) (*GoClient, error) {
-	logger.Info("consensus client: connecting",
-		fields.Address(opt.BeaconNodeAddr),
+	if opt.BeaconNodeAddr == "" {
+		return nil, fmt.Errorf("no beacon node address provided")
+	}
+
+	// TODO: Decide what symbol to use as a separator. Bootnodes are currently separated by ";". Deployment bot currently uses ",".
```
MatusKysel, line 178: what about this ?

MatusKysel: also fix the comment it's beacon nodes not boot

iurii-ssv: I've completely removed this `TODO` comment now since I don't see it being actionable ... @nkryuchkov let me know if you think we should do something else about it (or if you want this behavior documented somehow)

iurii-ssv:
Closing in favor of https://github.com/ssvlabs/ssv/pull/2469

beacon/goclient/goclient.go
nkryuchkov: @MatusKysel I meant bootnodes use ";" too, so it's consistent with each other

@iurii-ssv Well, I think ";" is the best choice we can make now, and it's too late to change something, so we don't need this TODO anymore. The code looks quite self-explanatory, so IMO it's fine

eth/executionclient/defaults.go
nkryuchkov: We need to check the request duration metrics to decide on this, but even 10s sounds worth switching to another client. Although some requests might be heavy, we need to verify this

PR #2469, opened 2025-08-28
execution-client: add timeouts for every request

iurii-ssv:
This PR is the exact copy of https://github.com/ssvlabs/ssv/pull/2468 but rebased onto **version/optimization** so we can release it under **v2.3.5**

### DIFF
--- a/eth/executionclient/errors.go
+++ b/eth/executionclient/errors.go
@@ -15,9 +15,9 @@ var (
 	ErrSyncing       = fmt.Errorf("syncing")
 )
 
-// errWithDetails wraps provided error adding more details to it.
-func (ec *ExecutionClient) errWithDetails(err error, method string) error {
-	return fmt.Errorf("%s -> %s: %w", ec.nodeAddr, method, err)
+// errSingleClient wraps provided error adding more details to it, useful for single-client errors.
+func (ec *ExecutionClient) errSingleClient(err error, routeName string) error {
+	return fmt.Errorf("single-client request %s -> %s: %w", ec.nodeAddr, routeName, err)
 }
 
 const (
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -100,7 +100,7 @@ func New(ctx context.Context, nodeAddr string, contractAddr ethcommon.Address, o
 	ec.syncProgressFn = func(ctx context.Context) (*ethereum.SyncProgress, error) {
 		sp, err := ec.client.SyncProgress(ctx)
 		if err != nil {
-			return nil, ec.errWithDetails(fmt.Errorf("fetch sync progress: %w", err), "eth_syncing")
+			return nil, ec.errSingleClient(fmt.Errorf("fetch sync progress: %w", err), "eth_syncing")
 		}
 		return sp, nil
 	}
@@ -123,7 +123,7 @@ func (ec *ExecutionClient) Close() error {
 func (ec *ExecutionClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64) (logs <-chan BlockLogs, errors <-chan error, err error) {
 	currentBlock, err := ec.client.BlockNumber(ctx)
 	if err != nil {
-		return nil, nil, ec.errWithDetails(fmt.Errorf("get current block: %w", err), "eth_blockNumber")
+		return nil, nil, ec.errSingleClient(fmt.Errorf("get current block: %w", err), "eth_blockNumber")
 	}
 	if currentBlock < ec.followDistance {
 		return nil, nil, ErrNothingToSync
@@ -237,7 +237,7 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
 	if err == nil {
 		return logs, nil
 	}
-	err = ec.errWithDetails(fmt.Errorf("get filtered logs: %w", err), "eth_getLogs")
+	err = ec.errSingleClient(fmt.Errorf("get filtered logs: %w", err), "eth_getLogs")
 
 	if isRPCQueryLimitError(err) {
 		if q.FromBlock == nil || q.ToBlock == nil {
@@ -360,7 +360,7 @@ func (ec *ExecutionClient) healthy(ctx context.Context) error {
 	sp, err := ec.SyncProgress(ctx)
 	if err != nil {
 		recordExecutionClientStatus(ctx, statusFailure, ec.nodeAddr)
-		return ec.errWithDetails(fmt.Errorf("get sync progress: %w", err), "eth_syncing")
+		return ec.errSingleClient(fmt.Errorf("get sync progress: %w", err), "eth_syncing")
 	}
 	recordRequestDuration(ctx, ec.nodeAddr, time.Since(start))
 
@@ -386,7 +386,7 @@ func (ec *ExecutionClient) healthy(ctx context.Context) error {
 func (ec *ExecutionClient) HeaderByNumber(ctx context.Context, blockNumber *big.Int) (*ethtypes.Header, error) {
 	h, err := ec.client.HeaderByNumber(ctx, blockNumber)
 	if err != nil {
-		return nil, ec.errWithDetails(fmt.Errorf("get header by block number %s: %w", blockNumber, err), "eth_getBlockByNumber")
+		return nil, ec.errSingleClient(fmt.Errorf("get header by block number %s: %w", blockNumber, err), "eth_getBlockByNumber")
 	}
 	return h, nil
 }
@@ -394,7 +394,7 @@ func (ec *ExecutionClient) HeaderByNumber(ctx context.Context, blockNumber *big.
 func (ec *ExecutionClient) SubscribeFilterLogs(ctx context.Context, q ethereum.FilterQuery, ch chan<- ethtypes.Log) (ethereum.Subscription, error) {
 	logs, err := ec.client.SubscribeFilterLogs(ctx, q, ch)
 	if err != nil {
-		return nil, ec.errWithDetails(fmt.Errorf("subscribe to filtered logs (query=%s): %w", q, err), "logs")
+		return nil, ec.errSingleClient(fmt.Errorf("subscribe to filtered logs (query=%s): %w", q, err), "logs")
 	}
 	return logs, nil
 }
@@ -402,7 +402,7 @@ func (ec *ExecutionClient) SubscribeFilterLogs(ctx context.Context, q ethereum.F
 func (ec *ExecutionClient) FilterLogs(ctx context.Context, q ethereum.FilterQuery) ([]ethtypes.Log, error) {
 	logs, err := ec.client.FilterLogs(ctx, q)
 	if err != nil {
-		return nil, ec.errWithDetails(fmt.Errorf("get filtered logs (query=%s): %w", q, err), "eth_getLogs")
+		return nil, ec.errSingleClient(fmt.Errorf("get filtered logs (query=%s): %w", q, err), "eth_getLogs")
 	}
 	return logs, nil
 }
@@ -439,7 +439,7 @@ func (ec *ExecutionClient) streamLogsToChan(ctx context.Context, logCh chan<- Bl
 	// So we can restart from this block once everything is good.
 	sub, err := ec.client.SubscribeNewHead(ctx, heads)
 	if err != nil {
-		return fromBlock, ec.errWithDetails(fmt.Errorf("subscribe new head: %w", err), "newHeads")
+		return fromBlock, ec.errSingleClient(fmt.Errorf("subscribe new head: %w", err), "newHeads")
 	}
 	defer sub.Unsubscribe()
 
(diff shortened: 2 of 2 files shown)


beacon/goclient/goclient.go
```diff
@@ -171,8 +171,15 @@ func New(
...
-	logger.Info("consensus client: connecting",
-		fields.Address(opt.BeaconNodeAddr),
+	if opt.BeaconNodeAddr == "" {
+		return nil, fmt.Errorf("no beacon node address provided")
+	}
+
+	// TODO: Decide what symbol to use as a separator. Bootnodes are currently separated by ";". Deployment bot currently uses ",".
+	beaconAddrList := strings.Split(opt.BeaconNodeAddr, ";")
+
+	logger.Info("consensus client: connecting (multi client)",
```
nkryuchkov, line 180: I'm not sure we need to add `(multi client)` because we have only one implementation. It might be very confusing for someone using only one beacon node, which is the most popular scenario

beacon/goclient/goclient.go
```diff
@@ -171,8 +171,15 @@ func New(
 	logger *zap.Logger,
 	opt Options,
 ) (*GoClient, error) {
-	logger.Info("consensus client: connecting",
-		fields.Address(opt.BeaconNodeAddr),
+	if opt.BeaconNodeAddr == "" {
+		return nil, fmt.Errorf("no beacon node address provided")
+	}
+
+	// TODO: Decide what symbol to use as a separator. Bootnodes are currently separated by ";". Deployment bot currently uses ",".
```
nkryuchkov: Let's remove it as discussed in https://github.com/ssvlabs/ssv/pull/2468/files

eth/executionclient/defaults.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	DefaultConnectionTimeout           = 10 * time.Second
+	DefaultReqTimeout                  = 15 * time.Second
```
nkryuchkov: Moving the conversation from https://github.com/ssvlabs/ssv/pull/2468/files#r2306965090

eth/executionclient/errors.go
```diff
@@ -12,14 +12,13 @@ var (
...
 
-// Domain-specific errors.
-var (
-	errSyncing = fmt.Errorf("syncing")
-)
-
-const elResponseErrMsg = "Execution client returned an error"
+// errWithDetails wraps provided error adding more details to it.
+func (ec *ExecutionClient) errWithDetails(err error, method string) error {
+	return fmt.Errorf("%w (address=%s method=%s)", err, ec.nodeAddr, method)
```
nkryuchkov: nit: this looks a bit inconsistent with how errors are usually rendered, maybe something like this?

If you disagree, please close this

```suggestion
	return fmt.Errorf("%s to %s: %w", method, ec.nodeAddr, err)
```

nkryuchkov (approved):
good job

eth/executionclient/errors.go
iurii-ssv: Yeah I was hesitant to stick the URL in the middle to "not break up the error text" ... but on a second thought it's looking rather nice actually,

updated to `return fmt.Errorf("%s + %s: %w", method, ec.nodeAddr, err)`

beacon/goclient/goclient.go
iurii-ssv: Done (thought I already pushed a commit to remove, but turns out I didn't)

beacon/goclient/goclient.go
iurii-ssv: Oh right, thanks for spotting this (I thought here we are always working with multi-client case, so I wanted the log-message to match what we output in ExecutionClient ... but that's not quite the case),

removed that `(multi client)` part now

eth/executionclient/defaults.go
iurii-ssv: I think we'll need to add more metrics to `ExecutionClient` to get better estimates, currently it seems to only measure `Health` endpoint (which takes just under ~2s on **stage**) ...

So I think 15s is a reasonable safe default we can use for now, we wouldn't want to block the merge of this PR on that - but I'll sketch a separate PR with the added metrics and later on we can re-visit this value (having gathered at least couple of days of data from running it on **stage**)

beacon/goclient/goclient.go
oleg-ssvlabs: it seems the `(multi client)` it still part of the log message, so it should be removed

cli/operator/node.go
```diff
@@ -330,7 +332,7 @@ var StartNodeCmd = &cobra.Command{
 		validatorProvider := nodeStorage.ValidatorStore().WithOperatorID(operatorDataStore.GetOperatorID)
 		validatorRegistrationSubmitter := runner.NewVRSubmitter(cmd.Context(), logger, networkConfig.Beacon, consensusClient, validatorProvider)
 
-		executionAddrList := strings.Split(cfg.ExecutionClient.Addr, ";") // TODO: Decide what symbol to use as a separator. Bootnodes are currently separated by ";". Deployment bot currently uses ",".
```
oleg-ssvlabs, line 333: I assume this comment is valid any longer? Otherwise `,` separator would not work?

eth/executionclient/defaults.go
oleg-ssvlabs: we should probably align this value with default request duration from the [config](https://github.com/ssvlabs/ssv/pull/2469/files#diff-3bb757c4694e898fdb8f40951242228db018bdfaa92da79d9334a01d2f546e98R12) which is `10s`

oleg-ssvlabs: not in scope of this PR, but I've noticed that [these](https://github.com/ssvlabs/ssv/pull/2469/files#diff-e8d4fb87f6b7393d4b6dbe2fbe51c1739da2f99713ab1b6147eabc5a1b673de9R16-R18) constants are unused, maybe worth removing while here

eth/executionclient/eth_client.go
oleg-ssvlabs, line 1: I’m not entirely sure about the idea of this wrapper.  
We already have a wrapper in `execution_client.go`, and we could just add timeouts there instead.  

To avoid accidentally forgetting to set timeouts, we could implement a unit test with right assertions

eth/executionclient/errors.go
```diff
@@ -12,14 +12,13 @@ var (
 	ErrClosed        = fmt.Errorf("closed")
 	ErrBadInput      = fmt.Errorf("bad input")
 	ErrNothingToSync = errors.New("nothing to sync")
```
oleg-ssvlabs: while here, could align error instantiation

oleg-ssvlabs (commented):
good job.

Left few comments

eth/executionclient/defaults.go
iurii-ssv: > not in scope of this PR, but I've noticed that [these](https://github.com/ssvlabs/ssv/pull/2469/files#diff-e8d4fb87f6b7393d4b6dbe2fbe51c1739da2f99713ab1b6147eabc5a1b673de9R16-R18) constants are unused, maybe worth removing while here

I've done that in another PR, so just keeping those around in this one to not create a merge-conflict

> we should probably align this value with default request duration from the [config](https://github.com/ssvlabs/ssv/pull/2469/files#diff-3bb757c4694e898fdb8f40951242228db018bdfaa92da79d9334a01d2f546e98R12) which is `10s`

The thing is - this `10s` value didn't apply to all requests ... so I'm feeling a bit more confident giving it more room for now (`15s`), but I think we can update default config value to match - https://github.com/ssvlabs/ssv/pull/2469/commits/97d0ceec27397c0a7f1b7b3e765ca2fb4e6d8d5d

cli/operator/node.go
iurii-ssv: I'm not 100% sure what the comment means, but my read is "Discord bot uses `,` to split different env-parameters. If we were to use `,` here as well Discord bot would run into trouble parsing our deploy-request"

eth/executionclient/eth_client.go
iurii-ssv: > To avoid accidentally forgetting to set timeouts, we could implement a unit test with right assertions

I'm not sure that's doable with a unit test, could you expand on it ? Lets say we need to start using new method on `ethclient` we didn't previously use - how can we make sure we did set a timeout for it (I guess the only way to do it would be via `ctx` ... [there is another option](https://github.com/ssvlabs/ssv/pull/2468#discussion_r2307313583) but like I mentioned I like this one better since it's more flexible & the cost of adding an extra wrapping-func is negligible)

eth/executionclient/errors.go
iurii-ssv: https://github.com/ssvlabs/ssv/pull/2469/commits/9ce1da193fd4ce6fb837cf852fbfc00984847d3b

eth/executionclient/defaults.go
iurii-ssv: ~Based on the values https://github.com/ssvlabs/ssv/pull/2470 gets us (running for `~3` hours), so far it looks like `~3s` is enough (although I wouldn't set the limit that low), we could probably lower it to `10s`~ this was incorrect, see the update below:

image

I don't quite understand however why `BlockNumber` and `SubscribeNewHead` (and `ChainID` that's used in multi-client only setup, I guess) are shown there as `0s` ... the only thing I can think of it looks like we are doing only 1 of those requests (at SSV node start) and somehow that 1 request is smoothed out to 0 ? @oleg-ssvlabs maybe you have seen something like that with metrics in the past ?

**Update:** actually the chart from above is not correct, it shows "requests per second" rather than "request latency" ... here is the updated chart showing both, and we can see that `~1s` is enough actually for 99th percentile of EL requests:

image

oleg-ssvlabs: @iurii-ssv I am not 100% confident, but I think this is what you are describing - the request was sent only once, and because you are using the `sum` and `rate` functions, it gets flattened to 0

eth/executionclient/eth_client.go
oleg-ssvlabs: I am not familiar with the entire test setup. It seems fairly complex at first glance.  
What I meant was:  
→ Mock the EL response time  
→ Provide a context with a lower timeout  
→ Expect a `DeadlineExceeded` or `Canceled` error type that should arrive no later than context deadline

Can't say how hard if would be to setup this behavior

eth/executionclient/defaults.go
iurii-ssv: It seems if that's the case indeed - the flattening happens in the observability code itself (not in Grafana), it shows 3 points with `0` values there:

image

We could extend `recordRequestDuration` func to accept `logger` for example and also log every request (mainly what we are interested in is "the time it took") ... but if we go for it I think we should do it for both CL and EL (`ExecutionClient` and `GoClient`) - https://github.com/ssvlabs/ssv/issues/2471 (and that's better to do out of scope of this PR)

eth/executionclient/eth_client.go
iurii-ssv: That would only work IF you are writing a unit-test of that form for newly added method-call ... which is rarely the case - nothing prevents the implementor to:
- either forget/bypass writing unit-tests for his newly added code
- or if he writes some he must remember to write that specific kind of test you described

So it's not really reliable :) With the current solution compiler won't let you to call `ethclient.Client` directly - you'd need to implement a wrapper-method for it (similar to other wrapper-methods) ... and the `ethClient struct` is intentionally taking care of timeouts only (we could have moved metric-recording in there, for example, but I think it's better to create another wrapper dedicated to metrics if we want the compiler to help us to not forget to add those as well ... but, unlike with timeouts, forgetting to record a metric isn't a critical bug)

eth/executionclient/defaults.go
iurii-ssv: > We could extend recordRequestDuration func to accept logger for example and also log every request (mainly what we are interested in is "the time it took") ... but if we go for it I think we should do it for both CL and EL (ExecutionClient and GoClient) - https://github.com/ssvlabs/ssv/issues/2471 (and that's better to do out of scope of this PR)

Implemented in https://github.com/ssvlabs/ssv/pull/2472 now,

as for `DefaultReqTimeout` (and similar config default) - [I've set it to `10s`](https://github.com/ssvlabs/ssv/pull/2469/commits/ceaec4d49570e673d1050e47ecc629d2ed8ebbfa) since the trend displayed above has persisted for ~24h now (we could go even lower to 5s for example, but I don't think there is any real benefit in doing that ... and we'd want the request to survive random hitches as well, if possible)

iurii-ssv: > @iurii-ssv I am not 100% confident, but I think this is what you are describing - the request was sent only once, and because you are using the `sum` and `rate` functions, it gets flattened to 0

@oleg-ssvlabs jfyi, now that I've added the logging as well I can see that those are indeed requests we issue only/mostly on startup, and 1-2 at most. So I guess it is a metrics-visibility issue after all.

iurii-ssv: > @oleg-ssvlabs jfyi, now that I've added the logging as well I can see that those are indeed requests we issue only/mostly on startup, and 1-2 at most. So I guess it is a metrics-visibility issue after all.

Super weird stuff btw, here is another relevant example for `SyncCommitteeDuties
` request, in logs I see 3 requests in total (over past 24h):

image

While on the metrics-chart I only see 1 of those 3 requests register (this is using dot-representation, but that doesn't matter) which is **also not even the longest request of those 3**:

image

**Another idea** would be - maybe these metrics are missing because upon cluster re-deploys `prometheus-exporter` is also restarted and is just not ready to accept the request that pushes those metrics to it yet ... this seems to align with the `starting SSV` node data exactly (the request that showed up on metrics chart is not due to SSV node restart, while 2 other do happen due to SSV node restart):

image

I guess it's not something we need (or maybe even can) to address ... just glad to have figured this out.

oleg-ssvlabs (approved):

PR #2470, opened 2025-08-29
execution-client: record request duration for every call

iurii-ssv:
This is a follow-up to https://github.com/ssvlabs/ssv/pull/2469#discussion_r2308447773 - as it turned out, we need more comprehensive metrics for `ExecutionClient` which this PR implements.

The general approach is copied over from CL code (`GoClient`).


eth/executionclient/execution_client.go
```diff
@@ -235,7 +238,7 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
 	default:
 	}
 
-	logs, err := ec.client.FilterLogs(ctx, q)
+	logs, err := ec.FilterLogs(ctx, q)
 	if err == nil {
 		return logs, nil
 	}
```
iurii-ssv, line 238: This makes it so that we re-use `ec.FilterLogs` method that's already instrumented with `recordRequestDuration` call (instead of having to additionally instrument it here).

eth/executionclient/execution_client.go
```diff
@@ -419,7 +426,9 @@ func (ec *ExecutionClient) SubscribeFilterLogs(ctx context.Context, q ethereum.F
 }
 
 func (ec *ExecutionClient) FilterLogs(ctx context.Context, q ethereum.FilterQuery) ([]ethtypes.Log, error) {
+	start := time.Now()
 	logs, err := ec.client.FilterLogs(ctx, q)
+	recordRequestDuration(ctx, "FilterLogs", ec.nodeAddr, http.MethodGet, time.Since(start), err)
```
oleg-ssvlabs: I know it's not very intuitive, but this method is actually a POST method, not GET.  

I'm not sure if writing HTTP verbs into metrics for RPC calls really makes sense, since most (if not all) of them are POST requests, even when the request is just fetching data.

eth/executionclient/execution_client.go
```diff
@@ -560,5 +571,8 @@ func (ec *ExecutionClient) Filterer() (*contract.ContractFilterer, error) {
 }
 
 func (ec *ExecutionClient) ChainID(ctx context.Context) (*big.Int, error) {
-	return ec.client.ChainID(ctx)
+	start := time.Now()
+	chainID, err := ec.client.ChainID(ctx)
+	recordRequestDuration(ctx, "ChainID", ec.nodeAddr, http.MethodGet, time.Since(start), err)
```
oleg-ssvlabs: this is also POST

eth/executionclient/observability.go
```diff
@@ -95,11 +97,26 @@ var (
 			metric.WithDescription("number of times a client was initialized")))
 )
 
-func recordRequestDuration(ctx context.Context, serverAddr string, duration time.Duration) {
+func recordRequestDuration(ctx context.Context, routeName, serverAddr, requestMethod string, duration time.Duration, err error) {
```
oleg-ssvlabs: I know you probably took the implementation from [here](https://github.com/ssvlabs/ssv/blob/stage/beacon/goclient/observability.go#L59) and frankly it was my code which I think could be slightly refactored :)

what about this version:

```Go
attr := []attribute.KeyValue{
		semconv.ServerAddress(serverAddr),
		semconv.HTTPRequestMethodKey.String(requestMethod),
		attribute.String("http.route_name", routeName),
	}

	var rpcErr rpc.Error
	if errors.As(err, &rpcErr) {
		attr = append(attr, attribute.Int("http.response.error_status_code", rpcErr.ErrorCode()))
	}

	requestDurationHistogram.Record(
		ctx,
		duration.Seconds(),
		metric.WithAttributes(attr...))
```

oleg-ssvlabs (commented):
left few comments

eth/executionclient/execution_client.go
iurii-ssv: Yeah I kind of left it to match how we do it in CL (`GoClient`), I guess it's better to remove HTTP verbs then - https://github.com/ssvlabs/ssv/pull/2470/commits/c773a30d3bd504c33fe8d46e40f235c57527c34b

eth/executionclient/execution_client.go
iurii-ssv: Addressed similar to https://github.com/ssvlabs/ssv/pull/2470#discussion_r2309997418

eth/executionclient/observability.go
iurii-ssv: Yup, simplified in https://github.com/ssvlabs/ssv/pull/2470/commits/c773a30d3bd504c33fe8d46e40f235c57527c34b

oleg-ssvlabs (dismissed):

eth/executionclient/execution_client.go
```diff
@@ -518,7 +529,8 @@ func (ec *ExecutionClient) connect(ctx context.Context) error {
 	}
 	ec.client = client
 
-	logger.Info("connected to execution client", zap.Duration("took", time.Since(start)))
+	logger.Info("connected to execution client")
```
nkryuchkov, line 532: Why do we need to remove how long it took?

iurii-ssv: Because now we'll be logging that info (along with some other stats) in `recordRequestDuration` funcs universally for all various client-calls we make - so if you really need this info you still have it there,

we could also log it here as well ... I just don't think it's necessary,

overall I'm trying to clean up errors/logging handling in CL/EL clients so we have better debugging capabilities, https://github.com/ssvlabs/ssv/pull/2472 contains the bulk of the changes (along with some various adjustments to some streaming logic and whatnot) I hope we can merge ~next week

so I'm essentially clearing the ground for the upcoming changes (sometimes it means removing maybe useful things ... but it's nothing major - we can always add back these later)

oleg-ssvlabs (approved):

Issue #2471, opened 2025-08-29
CL/EL: log request duration (for every call)


iurii-ssv:
We have `recordRequestDuration` func that's meant to measure how much time a particular call took (and we have it implemented in 2 places, for `ExecutionClient` and for `GoClient`) - it records those timings as a metric we can view in Prometheus.

It would be nice if we could also log that same data with a `logger` (`recordRequestDuration` func should probably accept `logger` as a param to preserve the relevant context) as well:
- this might simplify the debugging in some cases
- it seems, [sometimes for very rare calls metric doesn't show the actual value recorded (how much time the call took) - but rather just flattens it to 0](https://github.com/ssvlabs/ssv/pull/2469#discussion_r2309982400) ... at least that's what we think is happening, it might as well be some sort of obscure bug we couldn't figure out ... having an explicit log message will help us to investigate that (we'd want to check `BlockNumber` and `SubscribeNewHead` and `ChainID` calls)

PR #2472, opened 2025-08-30
Cl/EL: standardize error handling and logging

iurii-ssv:
This PR resolves https://github.com/ssvlabs/ssv/issues/2471 (see its description) and standardizes the way we log/error for CL requests (to simplify it, de-duplicate logged messages, log missing details such as client-address in some places, and match the approach we use in EL code as was [implemented in this PR](https://github.com/ssvlabs/ssv/pull/2469)).

Also this PR cleans up some code in the area in order to clarify/simplify it, and make single/multi EL client behavior less divergent:
- e.g. single-client case reties event-logs-streaming 3 times I believe, and then stops, while multi-client event-logs-streaming retries perpetually
- multi-client treats single-client case as if it cannot be handled by the generic multi-client logic ... we discussed it in the past and decided to keep it like that "to be on the safe side" back then, I think, but now, since I'm already doing modifications to this code - might as well clean it up (we'll do need to test for regressions ofc, I'll keep an eye out)
- previously, the code would terminate abruptly with `Fatal` (mostly around event-logs-streaming) in order to resolve a situation when EL is stuck and no further progress is possible ... while this is a working solution it's leads to some  unnecessary complexity of handling those scenarios separately from other error-scenarios + makes the code hard to modify; Now I'm relaying errors via `errorsCh` to let the caller decide what to do with them (so that `Fatal` calls are only limited to the outermost caller possible - the rest of unnecessary `Fatal` calls are removed, including those mocks we had in tests)
- I've also [extended a Grafana chart that shows CL / EL stats](https://github.com/ssvlabs/ssv/pull/2472#issuecomment-3240858149) (it's called "Consensus client -> Consensus client stats") - so now in conjunction with the updates from this PR it provides a nice overview for CL / EL requests

### DIFF
--- a/eth/executionclient/execution_client.go
+++ b/eth/executionclient/execution_client.go
@@ -67,7 +67,7 @@ type ExecutionClient struct {
 
 	// healthInvalidationInterval ensures we don't spam EL with health-check type requests too much.
 	healthInvalidationInterval time.Duration
-	lastHeathyTime             atomic.Int64
+	lastHealthyTime            atomic.Int64
 }
 
 // New creates a new instance of ExecutionClient.
@@ -359,7 +359,7 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 		return ErrClosed
 	}
 
-	lastHealthyTime := time.Unix(ec.lastHeathyTime.Load(), 0)
+	lastHealthyTime := time.Unix(ec.lastHealthyTime.Load(), 0)
 	if ec.healthInvalidationInterval != 0 && time.Since(lastHealthyTime) <= ec.healthInvalidationInterval {
 		// Synced recently, reuse the result (only if ec.healthInvalidationInterval is set).
 		return nil
@@ -388,7 +388,7 @@ func (ec *ExecutionClient) Healthy(ctx context.Context) error {
 
 	recordExecutionClientStatus(ctx, statusReady, ec.nodeAddr)
 
-	ec.lastHeathyTime.Store(time.Now().Unix())
+	ec.lastHealthyTime.Store(time.Now().Unix())
 
 	return nil
 }


beacon/goclient/spec.go
```diff
@@ -179,20 +181,16 @@ func (gc *GoClient) fetchBeaconConfig(ctx context.Context, client *eth2clienthtt
...
 
 	return beaconConfig, nil
 }
 
 func (gc *GoClient) getForkData(specResponse map[string]any) (map[spec.DataVersion]phase0.Fork, error) {
-	if specResponse == nil {
-		return nil, fmt.Errorf("spec response is nil")
-	}
-
 	getForkEpoch := func(key string, required bool) (phase0.Epoch, error) {
```
iurii-ssv, line 182: This can't be nil here since we check for that prior, removing to simplify.

beacon/goclient/goclient.go
```diff
@@ -315,32 +311,18 @@ func (gc *GoClient) addSingleClient(ctx context.Context, addr string) error {
...
 			beaconConfig, err := gc.fetchBeaconConfig(ctx, s)
 			if err != nil {
-				logger.Error(clResponseErrMsg,
-					zap.String("api", "fetchBeaconConfig"),
-					zap.Error(err),
-				)
-				return
+				logger.Fatal("couldn't fetch beacon config", zap.Error(err))
+				return // tests may override Fatal's behavior, hence gotta return in addition to the Fatal call
 			}
```
iurii-ssv, line 327: @nkryuchkov would you know if the `Fatal` call I've added here is needed (but was previously missing) or there was some reason to "just return" here (we'd need to document it if that's the case) ? Same applies to `s.NodeVersion` from above (now handled by helper-func `fetchNodeVersion` so we can record the request).

**Update:** it seems that maybe the intent was to "wait for another `OnActive` hook-call so we can retry (instead of crashing as in the case of "unexpected beacon config") ... looking deeper in the `go-eth2-client` code it does seem like connection can become active/inactive multiple times - so doing `Fatal` in case of temporary failure here would be bad, reverted in favor of preserving the old behavior (+ added comments)

beacon/goclient/spec.go
```diff
@@ -27,47 +27,51 @@ const (
...
+	nodeVersionResp, err := client.NodeVersion(ctx, &api.NodeVersionOpts{})
+	recordSingleClientRequest(ctx, gc.log, "NodeVersion", client.Address(), http.MethodGet, time.Since(start), err)
+	if err != nil {
+		return "", errSingleClient(fmt.Errorf("fetch node version response: %w", err), client.Address(), "NodeVersion")
+	}
+	if nodeVersionResp == nil {
+		return "", errSingleClient(fmt.Errorf("node version response is nil"), client.Address(), "NodeVersion")
+	}
+	if nodeVersionResp.Data == "" {
+		return "", errSingleClient(fmt.Errorf("node version response data is nil"), client.Address(), "NodeVersion")
```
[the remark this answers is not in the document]
iurii-ssv: The check itself is correct, adjusted the error message to reflect that it can only be nil (not empty)

beacon/goclient/aggregator.go
```diff
@@ -79,9 +68,6 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
 	switch aggDataResp.Data.Version {
 	case spec.DataVersionElectra:
 		if aggDataResp.Data.Electra == nil {
-			gc.log.Error(clNilResponseForkDataErrMsg,
-				zap.String("api", "AggregateAttestation"),
-			)
 			return nil, DataVersionNil, fmt.Errorf("aggregate attestation electra data is nil")
```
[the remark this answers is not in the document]
iurii-ssv: Thanks for spotting this, fixed.

eth/eventhandler/event_handler.go
```diff
@@ -134,7 +134,6 @@ func (eh *EventHandler) HandleBlockEventsStream(ctx context.Context, logs <-chan
 			logger = logger.With(fields.Type(task))
 			logger.Debug("executing task")
 			if err := task.Execute(); err != nil {
-				// TODO: We log failed task until we discuss how we want to handle this case. We likely need to crash the node in this case.
 				logger.Error("failed to execute task", zap.Error(err))
 			} else {
 				logger.Debug("executed task")
```
iurii-ssv, line 145: From what I understand, every `task` is a best-effort kind of thing, yes we might not be able to process events for some validators but that's what cluster-super-majority is for, 

crashing the node would make it unavailable to serve duties for other validators that are unaffected by the missed event(task) ... so, like we currently do in other places (with validator-startup for example), SSV node should keep working,

hence everything is already working as intended

eth/eventsyncer/event_syncer.go
```diff
@@ -121,14 +126,10 @@ func (es *EventSyncer) SyncHistory(ctx context.Context, fromBlock uint64) (lastP
...
-		lastProcessedBlock, err = es.eventHandler.HandleBlockEventsStream(ctx, fetchLogs, false)
+		lastProcessedBlock, err = es.eventHandler.HandleBlockEventsStream(ctx, fetchLogsCh, fetchErrorCh, false)
 		if err != nil {
 			return 0, fmt.Errorf("handle historical block events: %w", err)
 		}
-		// TODO: (Alan) should it really be here?
-		if err := <-fetchError; err != nil {
-			return 0, fmt.Errorf("error occurred while fetching historical logs: %w", err)
-		}
 		if lastProcessedBlock == 0 {
```
iurii-ssv: On this TODO - `HandleBlockEventsStream` handles it now

eth/executionclient/errors.go
```diff
@@ -39,5 +39,5 @@ func isRPCQueryLimitError(err error) bool {
 
 // isInterruptedError checks if the provided error represents some sort of interruption.
 func isInterruptedError(err error) bool {
-	return errors.Is(err, ErrClosed) || errors.Is(err, rpc.ErrClientQuit) || errors.Is(err, context.Canceled)
+	return errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled)
 }
```
iurii-ssv: I think we need to keep failing over to other clients in case of `rpc.ErrClientQuit` error - I'm not 100% sure, but it might not necessarily mean SSV node process needs to be terminating as well (once we receive this error **from just 1 single EL client**):
- based on what I see in the code, it likely means "this EL has terminated", and seemingly no reconnection can be expected ... but I'm not sure - hence would rather not make that assumption
- and since we don't control how the underlying library decides to implement `rpc.ErrClientQuit` they might change the behavior in the future (without us knowing)

So, for multi-client case, there isn't really much of a downside for treating this as a temporary error + failing over (and it feels more correct semantically). Thoughts ?

eth/executionclient/multi_client.go
```diff
@@ -220,62 +215,75 @@ func (mc *MultiClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64
...
-// StreamLogs subscribes to events emitted by the contract.
-// NOTE: StreamLogs spawns a goroutine which calls os.Exit(1) if no client is available.
-func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) <-chan BlockLogs {
-	logs := make(chan BlockLogs)
+// StreamLogs subscribes to events emitted by the Ethereum SSV contract(s) starting at fromBlock.
+// It spawns a go-routine that spins in a perpetual retry loop, terminating only on unrecoverable
+// interruptions (such as context cancels, client closure, etc.) as defined by isInterruptedError func.
+// Any errors encountered during log-streaming are relayed on errorsCh channel. Both logsCh and errorsCh
+// are closed once the streaming go-routine terminates.
+func (mc *MultiClient) StreamLogs(ctx context.Context, fromBlock uint64) (logsCh chan BlockLogs, errorsCh chan error) {
```
iurii-ssv: Essentially, I'm updating both `MultiClient.StreamLogs` and `ExecutionClient.StreamLogs` to:
- keep the implementations as similar as possible, they didn't diverge much ... but it was weird for example how `MultiClient` implemented perpetual retries while `ExecutionClient` was retrying ~3 times and was giving up after that
- remove the need to call `os.Exit(1)` (I think that comment here actually refers to calling `log.Fatal` - I don't remember the actual `os.Exit(1)` calls anywhere in this code)

eth/executionclient/multi_client.go
```diff
@@ -230,65 +215,79 @@ func (mc *MultiClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64
...
-					mc.logger.Fatal("no healthy clients", zap.Error(err))
-				}
-
-				// fromBlock's value in the outer scope is updated here, so this function needs to be a closure
+				// fromBlock is defined in the outer scope (from f's func perspective), but we have to modify it
+				// inside the f func because in case of failover to another client - we want to keep progressing
+				// from the latest block (not re-read the whole feed from the initial value of fromBlock). Note,
+				// f is called sequentially by a single go-routine, hence there is no need to synchronize concurrent
+				// access to fromBlock variable.
 				f := func(client SingleClientProvider) (any, error) {
```
iurii-ssv: Note, that health-check (`mc.Healthy` call) is done inside of `mc.call` below already (no need to duplicate it here)

eth/executionclient/errors.go
iurii-ssv: So, running into some issues with unit-tests I've settled on defining 2 separate error-categories (to make sure we still keep failing over in case of `rpc.ErrClientQuit` to another EL while terminating multi-client on unrecoverable interruptions such as `context.Canceled` or `ErrClosed`):
```
// isSingleClientInterruptedError checks if the provided error represents some sort of interruption
// an ExecutionClient experienced.
func isSingleClientInterruptedError(err error) bool {
	return errors.Is(err, ErrClosed) || errors.Is(err, rpc.ErrClientQuit) || errors.Is(err, context.Canceled)
}

// isMultiClientInterruptedError checks if the provided error represents some sort of interruption
// a MultiClient experienced.
func isMultiClientInterruptedError(err error) bool {
	// Note, if multi-client encountered ErrClosed (it can only come from ExecutionClient) it's safe to
	// assume we are in some sort of shutdown process when thre is no need to use multi-client failover
	// to try and recover from it.
	return errors.Is(err, ErrClosed) || errors.Is(err, context.Canceled)
}
```

eth/executionclient/execution_client.go
```diff
@@ -293,47 +294,57 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
...
+					// an error by itself.
+					err = fmt.Errorf("streamLogsToChan halted without an error")
 				}
 
 				tries++
-				if tries > 2 {
-					ec.logger.Fatal("failed to stream registry events", zap.Error(err))
+				if tries > maxTries {
+					errCh <- fmt.Errorf("failed to stream registry events even after %d retries: %w", tries, err)
+					return
```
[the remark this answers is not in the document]
iurii-ssv: This is "by design", we could simplify this code a bit to better reflect that - I'm just focusing on other stuff in this PR

eth/executionclient/multi_client.go
```diff
@@ -224,65 +209,81 @@ func (mc *MultiClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64
...
+				_, err := mc.call(contextWithMethod(ctx, "StreamLogs"), f)
+				if isMultiClientInterruptedError(err) {
+					// We are done with log streaming.
 					return
 				}
 
-				if isInterruptedError(err) {
-					mc.logger.Debug("stream logs stopped", zap.Error(err))
+				tries++
+				if tries > maxTries {
```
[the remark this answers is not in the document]
iurii-ssv: See https://github.com/ssvlabs/ssv/pull/2472#discussion_r2343516453

eth/executionclient/multi_client_test.go
```diff
@@ -319,8 +318,8 @@ func TestMultiClient_StreamLogs(t *testing.T) {
 	mockClient2.
 		EXPECT().
 		streamLogsToChan(gomock.Any(), gomock.Any(), uint64(202)). // fromBlock=202
-		Return(uint64(202), nil).                                  // Should not be called
-		Times(0)
+		Return(uint64(202), true, nil).                            // Should not be called
+		Times(1)
```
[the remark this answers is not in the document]
iurii-ssv: Fixed

beacon/goclient/observability.go
```diff
@@ -56,18 +58,85 @@ var (
...
+	err error,
+) {
+	// Log the request, but only if it has errored or if it took long enough that we want to pay attention
+	// to it (there are too many requests being made to log them every time, hence this limiting).
+	if err != nil || duration > 1*time.Millisecond {
+		logger.Debug("CL single-client request done",
+			zap.String("client_addr", clientAddr),
+			zap.String("route_name", routeName),
+			zap.String("http_method", httpMethod),
+			zap.Bool("success", err == nil),
```
nkryuchkov, line 76: Do we need it if we always log `zap.Error(err)`?

beacon/goclient/observability.go
```diff
@@ -56,18 +58,85 @@ var (
...
+func recordSingleClientRequest(
+	ctx context.Context,
+	logger *zap.Logger,
+	routeName, clientAddr, httpMethod string,
+	duration time.Duration,
+	err error,
+) {
+	// Log the request, but only if it has errored or if it took long enough that we want to pay attention
+	// to it (there are too many requests being made to log them every time, hence this limiting).
+	if err != nil || duration > 1*time.Millisecond {
```
nkryuchkov, line 71: Just curious about the limiting: do we have a lot of requests taking <1ms?

beacon/goclient/proposer.go
```diff
@@ -465,25 +457,10 @@ func (gc *GoClient) submitProposalPreparationBatches(
...
 	case submitted > 0:
-		gc.log.Error("partially submitted proposal preparations",
-			zap.Int("submitted", submitted),
-			zap.Int("failed", len(preparations)-submitted),
-			zap.Int("total", len(preparations)),
-			zap.Error(lastErr),
-		)
-		return fmt.Errorf("partially submitted preparations: %d/%d, last error: %w", submitted, len(preparations), lastErr)
-
+		return fmt.Errorf("partially submitted proposal preparations: %d/%d, last error: %w", submitted, len(preparations), jointErr)
```
nkryuchkov: We log `last error` but use `jointErr`. Is it correct? It looks like we log all errors, not just the last one

cli/operator/node.go
```diff
@@ -1167,14 +1179,16 @@ func syncContractEvents(
...
+		// the SSV node cannot work without being up to date with Ethereum events.
 		go func() {
-			err = eventSyncer.SyncOngoing(ctx, fromBlock.Uint64())
-
-			// Crash if ongoing sync has stopped, regardless of the reason.
-			logger.Fatal("failed syncing ongoing registry events",
-				zap.Uint64("last_processed_block", lastProcessedBlock),
-				zap.Error(err))
+			err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64())
+			if !errors.Is(err, context.Canceled) {
```
nkryuchkov: nit: not a part of your changes, but while we are touching this, maybe leaving a comment on why the logic crashes on `nil` error as well could make the code a bit clearer. Or probably we shouldn't crash on `nil` because it's unexpected (like `context.Canceled`)?

eth/executionclient/execution_client.go
```diff
@@ -293,47 +294,57 @@ func (ec *ExecutionClient) subdivideLogFetch(ctx context.Context, q ethereum.Fil
...
+	// handle the error before we can continue further).
+	errCh := make(chan error, 1)
 
 	go func() {
-		defer close(logs)
-		tries := 0
+		defer close(logsCh)
+		defer close(errCh)
+
+		const maxTries = math.MaxUint64 // infinitely many
```
nkryuchkov: Can you please explain the logic with `maxTries`? I'm not sure I understand it

eth/executionclient/execution_client_test.go
```diff
@@ -583,8 +600,10 @@ func TestFetchLogsInBatches(t *testing.T) {
 	t.Run("startBlock is greater than endBlock", func(t *testing.T) {
 		logChan, errChan := env.client.fetchLogsInBatches(env.ctx, 10, 5)
 		select {
-		case <-logChan:
-			require.Fail(t, "Should not receive log when startBlock > endBlock")
+		case log, ok := <-logChan:
+			if ok {
+				require.Fail(t, "Should not receive log when startBlock > endBlock, received log with block number: %d", log.BlockNumber)
+			}
```
nkryuchkov:
```suggestion
			require.Falsef(t, ok, "Should not receive log when startBlock > endBlock, received log with block number: %d", log.BlockNumber)
```

eth/executionclient/multi_client.go
```diff
@@ -224,65 +207,81 @@ func (mc *MultiClient) FetchHistoricalLogs(ctx context.Context, fromBlock uint64
...
+	// All errors are buffered, so we don't block the execution of this func (waiting on the caller to
+	// handle the error before we can continue further).
+	errCh = make(chan error, 1)
 
 	go func() {
-		defer close(logs)
+		defer close(logsCh)
+		defer close(errCh)
+
+		const maxTries = math.MaxUint64 // infinitely many
```
nkryuchkov: ditto

eth/executionclient/observability.go
```diff
@@ -97,17 +99,41 @@ var (
...
+func recordSingleClientRequest(
+	ctx context.Context,
+	logger *zap.Logger,
+	routeName, clientAddr string,
+	duration time.Duration,
+	err error,
+) {
+	// Log the request, but only if it has errored or if it took long enough that we want to pay attention
+	// to it (there are too many requests being made to log them every time, hence this limiting).
+	if err != nil || duration > 1*time.Millisecond {
```
nkryuchkov, line 112: I see this code is somewhat duplicated quite a few times. Do you think we could find a good abstraction and extract all cases, simplifying this code? I'm not sure it would look better, though

eth/executionclient/options.go
```diff
@@ -45,13 +45,15 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
 }
 
 // WithConnectionTimeout sets timeout for network connection to eth1 node.
+// The timeout must be positive, otherwise the default value will be used.
```
nkryuchkov, line 40: I know I asked you to add this, thanks, but it seems this comment is valid only on the `main` branch, as we don't have this logic on `stage`. So perhaps this PR is not the best place to add this comment.

eth/executionclient/options.go
```diff
@@ -45,13 +45,15 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
...
 func WithConnectionTimeout(timeout time.Duration) Option {
 	return func(s *ExecutionClient) {
 		s.connectionTimeout = timeout
 	}
 }
 
 // WithConnectionTimeoutMulti sets timeout for network connection to eth1 node.
+// The timeout must be positive, otherwise the default value will be used.
```
nkryuchkov, line 50: ditto

eth/executionclient/execution_client.go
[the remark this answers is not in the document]
nkryuchkov: @iurii-ssv could you please expand your comment on "by design"? I'm unresolving it temporarily until we clarify this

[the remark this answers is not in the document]
iurii-ssv: @nkryuchkov as per [my comment above](https://github.com/ssvlabs/ssv/pull/2472#discussion_r2312086252) 

> it was weird for example how MultiClient implemented perpetual retries while ExecutionClient was retrying ~3 times and was giving up after that

So, at the very least we should make both implementations (`MultiClient` and `ExecutionClient`) to work the same way,

and then, I think, it makes sense for streaming functions to "implement infinite retries" in general (eg. `Events` function implemented by `go-eth2-client` perpetually resubscribes under the hood in case of any errors are encountered),

- do you agree we want to retry here perpetually ? If the streaming is "stuck for some reason" `eventSyncer` should notice and crash SSV node in this case ... or if we don't want to rely on `eventSyncer` just send every error encountered on `errsCh` (the `StreamLogs` func returns) for the caller to handle as he thinks best (eg. terminate the process after 3 consecutive errors are encountered, without any new block sends) - we'll have to adjust the way the caller is handling those channels returned from `StreamLogs` func accordingly though
- thinking more about it - if we are gonna retry perpetually in `StreamLogs` func we better simplify this code (I've added a commit https://github.com/ssvlabs/ssv/pull/2472/commits/fb58fd5137a1f9a644e4d9eca5d964c4d89dad7f to remind me clean things up once we are on the same page)
- if you think we don't want infinite retries here for some reason, could you explain why ?

beacon/goclient/observability.go
iurii-ssv: It's nice to be able to filter logs by `success=true` / `success=false` (we can probably filter these logs by "error presence" instead ... but it's probably not worth the hassle)

nkryuchkov: @iurii-ssv good point

eth/executionclient/execution_client.go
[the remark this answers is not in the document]
nkryuchkov: @iurii-ssv I don't mind having infinite retries. I just don't understand the logic of going from 0 to `math.MaxUint64` and the `tries > maxTries` comparison because no number exceeds `math.MaxUint64`. Why can't it just be a forever loop?

beacon/goclient/observability.go
iurii-ssv: Yes we do for `Domain`, see [WIP Grafana charts](https://grafana.ops.ssvlabsinternal.com/d/debxu7243gef4f/consensus-client-stats?orgId=1&from=now-6h&to=now&timezone=browser&var-pod=ssv-node-70-0&var-cluster=stage&var-network=hoodi) for details (and later on btw we should probably [look into optimizing some of CL/EL requests](https://github.com/ssvlabs/ssv/issues/2473#issuecomment-3240169367) as well):

image

beacon/goclient/proposer.go
iurii-ssv: Yeah, nice catch, `last error:` mention is not correct here - changed to `encountered errors:` in https://github.com/ssvlabs/ssv/pull/2472/commits/b79bc5bfe257073cc495302730ea4942deeae432

eth/executionclient/execution_client.go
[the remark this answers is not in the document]
iurii-ssv: yeah, it can be a forever loop, and I'm gonna apply that update now - I just wanted to keep the code structure as close to the original one as possible (so the change is better visible in the diff vs 3 retries we were doing previously)

eth/executionclient/execution_client.go
iurii-ssv: See the full explanation in https://github.com/ssvlabs/ssv/pull/2472#discussion_r2343582221

eth/executionclient/multi_client.go
iurii-ssv: See the full explanation in https://github.com/ssvlabs/ssv/pull/2472#discussion_r2343582221

beacon/goclient/observability.go
nkryuchkov: @iurii-ssv Thanks. Can you please leave a short comment about this? I think it might be confusing why we have this condition, so this would make the code much easier to understand

eth/executionclient/observability.go
iurii-ssv: TBH I think it's fine for this code to be replicated multiple times because it contains implementation-specific parts that are hard to abstract away, so I can't really see how we could benefit from it

nkryuchkov: @iurii-ssv Alright, just wanted to know your opinion on this, as you're more familiar with this code. Thanks

eth/executionclient/options.go
iurii-ssv: Resolving to keep the relevant discussion in one place - https://github.com/ssvlabs/ssv/pull/2472#discussion_r2346767899

cli/operator/node.go
iurii-ssv: Yeah we actually don't want to log an error here in case `err == nil` because that scenario (of streaming ending without an error) is/should be handled deeper inside `SyncOngoing` func itself,

fixed that now

eth/executionclient/options.go
iurii-ssv: Right, I was planning to rebase this PR onto the latest **stage** before merging it (which now contains those **main** branch changes) - done, hope it makes sense now (let me know if not)

beacon/goclient/observability.go
iurii-ssv: I'm not sure we want to get more specific than what I already posted in a comment in there:

> // Log the request, but only if it has errored or if it took long enough that we want to pay attention
> // to it (there are too many requests being made to log them every time, hence this limiting).

this comment ^ is generic **intentionally**, we could mention that `Domain` call in there to "provide an example" - but that would make the comment CL-specific (since there is no `Domain` call to be made for EL client) while at the same time we have a very similar block of code with that same comment in EL package as well

so IMO it's a good balance between "the code/comment being generic to be re-used" & "contains enough info to make sense of it" - but if you have something else in mind (on how to rephrase it) let me know

nkryuchkov: @iurii-ssv well, I forgot about the `Domain` call being a no-call, so I was a bit confused why some requests would be <1ms. Can we mention somehow that we are ignoring those "no-calls" with this check? Maybe something like this?

> // Log the request, but only if it has errored or if it has sent a request to CL
> // (some calls don't make a request, and there are too many requests being
> // made to log them every time, hence this limiting).

iurii-ssv: Just to clarify, the "`Domain` call being a no-call" is not entirely true, is it ? It might or might not send a CL request depending on whether there is a cached result available,

but besides that, your suggestion ins't 100% precise - `if it has sent a request to CL` - because some actual networking-calls can take less that 1ms (unlikely, but not impossible),

I think we can rephrase/clarify it as:
```
// Log the request, but only if it has errored or if it took long enough that we want to pay attention
// to it (there are too many requests being made to log them every time, some requests don't even result 
// into a network-based call due to caching implemented for some routes).
```

wdyt ?

nkryuchkov: @iurii-ssv yeah, I like this one

iurii-ssv: Applied in https://github.com/ssvlabs/ssv/pull/2472/commits/9bbfed34c9b308bda512209b8b01318bd3e1a9cc

eth/executionclient/execution_client.go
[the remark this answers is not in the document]
iurii-ssv: Simplified away the retries now (in favor of the perpetual loop, terminating on interrupt-kind of errors such as `context.Cancel` only) - https://github.com/ssvlabs/ssv/pull/2472/commits/dfe0573a350011bfc6a471d1c647d71144778190

nkryuchkov (dismissed):
Great work, let's make sure this gets tested well enough

eth/executionclient/multi_client_test.go
```diff
@@ -18,12 +18,71 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 
 	"github.com/ssvlabs/ssv/observability/log"
```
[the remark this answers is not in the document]
iurii-ssv: We are migrating to `github.com/ssvlabs/ssv/observability/log` gradually, that's fine

PR #2476, opened 2025-08-31
duties: clarify & fix typos

iurii-ssv:
While digging the code in the duty-handlers area I've spotted a couple of typos & rough spots I would like to address in this PR.


operator/duties/committee.go
```diff
@@ -63,7 +64,13 @@ func (h *CommitteeHandler) HandleDuties(ctx context.Context) {
...
 
 			h.logger.Debug("🛠 ticker event", zap.String("period_epoch_slot_pos", buildStr))
-			h.processExecution(ctx, period, epoch, slot)
+
+			func() {
+				tickCtx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(slot+1).Add(100*time.Millisecond))
+				defer cancel()
+
+				h.processExecution(tickCtx, period, epoch, slot)
+			}()
```
iurii-ssv, line 74: I think we need to set a deadline here for committee-duty, like we do for other duties

operator/duties/attester.go
```diff
@@ -84,8 +84,13 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
 
-			h.processExecution(ctx, currentEpoch, slot)
-			h.processFetching(ctx, currentEpoch, slot)
+			func() {
+				tickCtx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(slot+1).Add(100*time.Millisecond))
+				defer cancel()
+
+				h.executeAggregatorDuties(tickCtx, currentEpoch, slot)
+				h.processFetching(tickCtx, currentEpoch, slot)
+			}()
```
iurii-ssv, line 98: For `AttesterHandler` (it handled aggregator duty execution only, I believe) this deadline was previously set in `processFetching`  func ... but I don't see why we can't set it here instead (so we have it done the same way for all duties).

operator/duties/scheduler.go
```diff
@@ -402,29 +402,40 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, duties []*spectypes.Valid
 	defer span.End()
 
 	for _, duty := range duties {
-		duty := duty
 		logger := s.loggerWithDutyContext(duty)
 
 		slotDelay := time.Since(s.beaconConfig.SlotStartTime(duty.Slot))
```
iurii-ssv, line 430: We no longer need this work-around (since go1.22 I believe).

operator/duties/attester.go
```diff
@@ -286,26 +292,31 @@ func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
...
+		// parent-context, and as a result once it decides it's done with the parent-context it will
+		// cancel it (also canceling our operations here). Thus, we create our own context instance.
+		subscriptionCtx, cancel := context.WithDeadline(context.Background(), parentDeadline)
 		defer cancel()
 
 		if err := h.beaconNode.SubmitBeaconCommitteeSubscriptions(subscriptionCtx, subscriptions); err != nil {
 			h.logger.Warn("failed to submit beacon committee subscription", zap.Error(err))
 		}
-	}(h, subscriptions)
+	}()
```
iurii-ssv, line 334: Similar to `duty := duty` construct, we probably don't need to worry about enclosing these values and defining/passing them func-parameters (since go1.22 I believe) ?

operator/duties/proposer.go
```diff
@@ -131,9 +134,6 @@ func (h *ProposerHandler) processFetching(ctx context.Context, epoch phase0.Epoc
 		))
 	defer span.End()
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	span.AddEvent("fetching duties")
 	if err := h.fetchAndProcessDuties(ctx, epoch); err != nil {
 		// Set empty duties to inform DutyStore that fetch for this epoch is done.
```
iurii-ssv, line 141: Removing since this does nothing really.

operator/duties/sync_committee.go
```diff
@@ -148,17 +151,14 @@ func (h *SyncCommitteeHandler) HandleInitialDuties(ctx context.Context) {
...
 			observability.BeaconPeriodAttribute(period),
-			observability.BeaconRoleAttribute(spectypes.BNRoleSyncCommittee),
+			observability.BeaconRoleAttribute(spectypes.BNRoleSyncCommitteeContribution),
 		))
 	defer span.End()
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	if h.fetchCurrentPeriod {
```
iurii-ssv, line 163: Removing since this does nothing really.

operator/duties/proposer.go
julienharbulot: Actually even stronger than "does nothing", I think this immediate cancel was a source of the bug. I've tracked down the runtime downstream to see what this `defer cancel()` actually cancels and this is the goroutine that gets immediately canceled: 

```
go func(ctx, h, subs) {
      subscriptionCtx, cancel := context.WithDeadline(ctx, deadline)
      defer cancel()
      h.beaconNode.SubmitSyncCommitteeSubscriptions(subscriptionCtx, subscriptions)
    }(ctx, h, subscriptions)
```

julienharbulot: Very nice catch @iurii-ssv :)

iurii-ssv: Oh wow! You are right indeed!

operator/duties/attester.go
```diff
@@ -286,26 +289,31 @@ func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
...
+	go func() {
+		// We want to inherit parent-context deadline, but we cannot use the parent-context itself
+		// here because we are now running asynchronously with the go-routine that's managing
+		// parent-context, and as a result once it decides it's done with the parent-context it will
+		// cancel it (also canceling our operations here). Thus, we create our own context instance.
+		subscriptionCtx, cancel := context.WithDeadline(context.Background(), parentDeadline)
 		defer cancel()
 
 		if err := h.beaconNode.SubmitBeaconCommitteeSubscriptions(subscriptionCtx, subscriptions); err != nil {
 			h.logger.Warn("failed to submit beacon committee subscription", zap.Error(err))
```
oleg-ssvlabs: Maybe worth changing the log level from Warning to Error?

operator/duties/attester.go
oleg-ssvlabs: `processFetching` is called in multiple places, so it probably makes sense to have the deadline set within the `processFetching` method.

oleg-ssvlabs: Also, this is a behavior change - previously the execution of aggregator duties had no deadline. However, since this logic runs within the slot ticker, I assume it is actually better to have a deadline?

operator/duties/attester.go
```diff
@@ -159,9 +164,6 @@ func (h *AttesterHandler) processFetching(ctx context.Context, epoch phase0.Epoc
 		))
 	defer span.End()
 
-	ctx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(slot+1).Add(100*time.Millisecond))
```
oleg-ssvlabs, line 164: Because this deadline was removed you won't be able to fetch the deadline [here](https://github.com/ssvlabs/ssv/blob/duties-clarify-and-fix-typos/operator/duties/attester.go#L296)). As I mentioned [here](https://github.com/ssvlabs/ssv/pull/2476/files#r2313375837) - the deadline now is not set for some of the scenarios.

operator/duties/voluntary_exit.go
```diff
@@ -64,7 +64,13 @@ func (h *VoluntaryExitHandler) HandleDuties(ctx context.Context) {
 			next = h.ticker.Next()
 
 			h.logger.Debug("🛠 ticker event", fields.Slot(currentSlot))
-			h.processExecution(ctx, currentSlot)
+
+			func() {
+				tickCtx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(currentSlot+1).Add(100*time.Millisecond))
```
oleg-ssvlabs: since we repeat this deadline calculation in multiple places now - it would be good if we could move it under `baseHandler`. 

maybe something like this:

```Go
func (baseHandler) ExecutionDeadline(ctx context.Context, currentSlot phase0.Slot) (context.Context, context.CancelFunc) {
	slotDeadlineCtx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(currentSlot+1).Add(100*time.Millisecond))

	return slotDeadlineCtx, cancel
}
```

operator/duties/validator_registration.go
```diff
@@ -134,6 +113,36 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
...
+			// Only attesting validators are eligible for registration duties.
+			continue
+		}
+		if uint64(share.ValidatorIndex)%registrationSlots != uint64(slot)%registrationSlots {
+			continue
+		}
+
+		pk := phase0.BLSPubKey{}
+		copy(pk[:], share.ValidatorPubKey[:])
+		h.dutiesExecutor.ExecuteDuties(ctx, []*spectypes.ValidatorDuty{{
```
oleg-ssvlabs: Side question - isn’t it a bit weird doing this one by one? I know it’s async under the hood (running in a separate goroutine in the Scheduler), but still... Wdyt @iurii-ssv @moshe-blox? 

Looking into the implementation - I think it makes more sense to collect all duties and rather execute the entire collection? If there is a specific reason doing it in this way - we need to document it

operator/duties/sync_committee.go
```diff
@@ -221,11 +221,11 @@ func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint
 func (h *SyncCommitteeHandler) fetchAndProcessDuties(ctx context.Context, epoch phase0.Epoch, period uint64, waitForInitial bool) error {
```
oleg-ssvlabs, line 226: This method fetches and stores Sync Committee duties (not Sync Committee Contribution ones) which eventually are fetched from the store and executed by Committee Handler, so you need to revert some changes here.

operator/duties/sync_committee.go
```diff
@@ -184,7 +184,7 @@ func (h *SyncCommitteeHandler) processFetching(ctx context.Context, epoch phase0
 
 func (h *SyncCommitteeHandler) processExecution(ctx context.Context, period uint64, slot phase0.Slot) {
 	ctx, span := tracer.Start(ctx,
-		observability.InstrumentName(observabilityNamespace, "sync_committee.execute"),
+		observability.InstrumentName(observabilityNamespace, "sync_committee_contribution.execute"),
```
oleg-ssvlabs, line 192: this is good change - it should be `sync_committee_contribution`

operator/duties/scheduler.go
```diff
@@ -402,29 +402,40 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, duties []*spectypes.Valid
...
+			span.AddEvent(eventMsg, trace.WithAttributes(
+				attribute.Int64("ssv.beacon.slot_delay_ms", slotDelay.Milliseconds()),
+				observability.BeaconRoleAttribute(duty.Type),
+				observability.RunnerRoleAttribute(duty.RunnerRole())))
 		}
 
 		slotDelayHistogram.Record(ctx, slotDelay.Seconds())
 
-		go func(ctx context.Context) {
-			if duty.Type == spectypes.BNRoleAttester || duty.Type == spectypes.BNRoleSyncCommittee {
```
oleg-ssvlabs, line 442: Seems to be really good catch. These duties will run in `ExecuteCommitteeDuties` method, do I understand it right?

operator/duties/scheduler.go
```diff
@@ -402,29 +402,40 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, duties []*spectypes.Valid
...
-				s.waitOneThirdOrValidBlock(duty.Slot)
-			}
-			recordDutyExecuted(ctx, duty.RunnerRole())
-			s.dutyExecutor.ExecuteDuty(ctx, duty)
-		}(ctx)
+		parentDeadline, ok := ctx.Deadline()
```
oleg-ssvlabs: I am unsure about this.  

This enforces the caller to set the deadline, and if the caller does not set it, duty execution will be completely skipped. I wouldn’t skip duty execution if the deadline isn’t set.  
Maybe it would be better to either skip setting the deadline or set it to some value (perhaps using the same slot-based logic).

operator/duties/scheduler.go
```diff
@@ -450,19 +461,35 @@ func (s *Scheduler) ExecuteCommitteeDuties(ctx context.Context, duties committee
...
+			logger.Warn(eventMsg, zap.Duration("slot_delay", slotDelay))
 			span.AddEvent(eventMsg, trace.WithAttributes(
 				observability.CommitteeIDAttribute(committee.id),
 				attribute.Int64("ssv.beacon.slot_delay_ms", slotDelay.Milliseconds())))
 		}
 
 		slotDelayHistogram.Record(ctx, slotDelay.Seconds())
 
-		go func(ctx context.Context) {
+		parentDeadline, ok := ctx.Deadline()
```
oleg-ssvlabs: [same](https://github.com/ssvlabs/ssv/pull/2476/files#r2313626345)

operator/duties/scheduler.go
oleg-ssvlabs: Optional: maybe we could check if the deadline has already passed?

operator/duties/attester.go
iurii-ssv: > `processFetching` is called in multiple places, so it probably makes sense to have the deadline set within the `processFetching` method.

yeah, I've noticed, but the only other place is `HandleInitialDuties` where it sets its own deadline

> Also, this is a behavior change - previously the execution of aggregator duties had no deadline. However, since this logic runs within the slot ticker, I assume it is actually better to have a deadline?

Yes, I'm changing this behavior because:
- it's better if it's done in the same manner across all the duties (so it would be hard to miss if something is not right with it), and in a place where it's easy to see (rather than buried under layers of helper-func calls)
- regarding the deadline for execution, we are not really setting the `deadline` for the "entire duty execution flow" here but rather we just "start the execution flow here by queueing it up" (this is what `ExecuteDuties` and `ExecuteCommitteeDuties` methods do), and there is no reason not to set a reasonable deadline for these calls now that they accept `ctx` parameter

**Update:** upon further check I found that we also should set the deadline on `case reorgEvent := <-h.reorg:` - fixed now (which was previously set by `processFetching` func) - and [you've spotted it here](https://github.com/ssvlabs/ssv/pull/2476#discussion_r2313498119) as well 👍

operator/duties/attester.go
iurii-ssv: Yeah that would make more sense, done

operator/duties/attester.go
iurii-ssv: Nice catch, fixed as mentioned in https://github.com/ssvlabs/ssv/pull/2476#discussion_r2313672063

operator/duties/sync_committee.go
iurii-ssv: Oh yeah, I got this mixed up a bit, should be fixed now

iurii-ssv: Reverted

operator/duties/validator_registration.go
iurii-ssv: Yes I've also noticed this "being unconventional/suboptimal" for no reason - and I've updated it like you described now so it doesn't pose these questions in the future

operator/duties/voluntary_exit.go
iurii-ssv: Added this to de-dup:
```go
// ctxWithDeadlineOnNextSlot returns the derived context with deadline set to next slot (+ some safety margin
// to account for clock skews).
func (h *baseHandler) ctxWithDeadlineOnNextSlot(ctx context.Context, slot phase0.Slot) (context.Context, context.CancelFunc) {
	return context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(slot+1).Add(100*time.Millisecond))
}
```

operator/duties/scheduler.go
iurii-ssv: Yes, there is no need to have this `if`-statement here because it should never execute after Alan fork has happened (it's dead code)

operator/duties/scheduler.go
iurii-ssv: You are right, this isn't ideal, I've updated it now for "`deadline` to be optional",

I don't really like the idea of setting our own default value here - that could mask the issue.

oleg-ssvlabs (dismissed):
lgtm

operator/duties/attester.go
```diff
@@ -84,8 +84,16 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
+			func() {
+				// Aggregates submissions are rewarded as long as they are finished within 2 slots after the target slot
+				// (the target slot itself, plus the next slot after that), hence we are setting the deadline here to
+				// target slot + 2.
+				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot)
+				defer cancel()
+
+				h.executeAggregatorDuties(tickCtx, currentEpoch, slot)
+				h.processFetching(tickCtx, currentEpoch, slot)
+			}()
```
iurii-ssv, line 98: As per discussion with ChatJPT, I've applied the `target slot + 2` deadline to all 3 duty types:

image

operator/duties/committee.go
```diff
@@ -61,9 +61,17 @@ func (h *CommitteeHandler) HandleDuties(ctx context.Context) {
...
+
+			func() {
+				// Attestations and sync-committee submissions are rewarded as long as they are finished within
+				// 2 slots after the target slot (the target slot itself, plus the next slot after that), hence
+				// we are setting the deadline here to target slot + 2.
+				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot+1)
+				defer cancel()
+
+				h.processExecution(tickCtx, period, epoch, slot)
+			}()
```
iurii-ssv, line 74: See https://github.com/ssvlabs/ssv/pull/2476/files#r2315436567

oleg-ssvlabs (dismissed):

operator/duties/attester.go
iurii-ssv: But also, in this PR I'm currently setting the same ~24s deadline to do both `h.execute...` and `h.process...` - while it probably mostly applies to `h.process...` since fetching can happen whole epoch (or something) in advance ... it's fine I guess

operator/duties/util.go
```diff
@@ -0,0 +1,19 @@
...
+
+// ctxWithDeadlineOnNextSlot returns newly created context setting its deadline to parentCtx
+// deadline if it has one set. This is useful when we want to inherit parentCtx deadline,
+// but not get canceled when parentCtx is canceled (e.g., we want to keep working in the
+// background until the deadline expires).
+func ctxWithParentDeadline(parentCtx context.Context) (ctx context.Context, cancel context.CancelFunc, withDeadline bool) {
+	ctx, cancel = context.WithoutCancel(parentCtx), func() {}
+	parentDeadline, ok := parentCtx.Deadline()
+	if ok {
+		ctx, cancel = context.WithDeadline(context.WithoutCancel(parentCtx), parentDeadline)
```
oleg-ssvlabs, line 15: isn't this technically the same as we would pass `context.Background()` instead of `context.WithoutCancel(parentCtx)`? 

Essentially what we are doing is - please do not inherit anything related to cancellations from `parentCtx`.

I assume this is slightly nicer, because if `parentCtx` carries some data - it will propagate it to the new context. Do I understand it right?

oleg-ssvlabs (dismissed):

operator/duties/util.go
iurii-ssv: Yes, sounds correct, this is per [Julian's suggestion from here](https://github.com/ssvlabs/ssv/pull/2484#issuecomment-3244652441)

operator/duties/util.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+)
+
+// ctxWithDeadlineOnNextSlot returns newly created context setting its deadline to parentCtx
+// deadline if it has one set. This is useful when we want to inherit parentCtx deadline,
+// but not get canceled when parentCtx is canceled (e.g., we want to keep working in the
+// background until the deadline expires).
+func ctxWithParentDeadline(parentCtx context.Context) (ctx context.Context, cancel context.CancelFunc, withDeadline bool) {
```
olegshmuelov: nit: The comment references the wrong function name - it says `ctxWithDeadlineOnNextSlot` but the function is actually `ctxWithParentDeadline`.

olegshmuelov (dismissed):
LGTM

nkryuchkov (commented):
LGTM

PR #2484, opened 2025-09-02
backport: duties: clarify & fix typos & fix sync-committee subscription

iurii-ssv:
This is a backport of https://github.com/ssvlabs/ssv/pull/2476 - but instead of being its full blown carbon-copy it's a minimal working solution to fix the issue for **version/optimization** branch (so it can go into the upcoming v2.3.5 release ASAP without much of extra testing required). 

These changes are "minimal" to keep the future resolving of merge conflicts as simple as possible (once we eventually merge branches like **main**->**stage**).


oleg-ssvlabs (approved):

iurii-ssv:
> Approving, but I think it would be better to use `context.WithoutCancel` instead of `context.Background`?

@julienh-ssv yeah I guess that could be useful for tracing to inhering the values it might store in the `ctx`, but since we don't have tracing deployed on prod (or at least not actively using it there from what I know) it's better to address in https://github.com/ssvlabs/ssv/pull/2476 (I've added a commit in there now to do just that)

PR #2485, opened 2025-09-02
metrics: (duties) add missing details

iurii-ssv:
Just minor clarifications for duty-related metrics we write.

### DIFF
--- a/observability/attributes.go
+++ b/observability/attributes.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ssvlabs/ssv-spec/qbft"
 	"github.com/ssvlabs/ssv-spec/types"
 
+	"github.com/ssvlabs/ssv/observability/utils"
 	"github.com/ssvlabs/ssv/protocol/v2/message"
 	protocolTypes "github.com/ssvlabs/ssv/protocol/v2/types"
 )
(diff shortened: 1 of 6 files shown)


observability/attributes.go
```diff
@@ -38,7 +39,7 @@ func BeaconVersionAttribute(version spec.DataVersion) attribute.KeyValue {
 }
 
 func RunnerRoleAttribute(role types.RunnerRole) attribute.KeyValue {
-	return attribute.String(RunnerRoleAttrKey, role.String())
+	return attribute.String(RunnerRoleAttrKey, utils.FormatRunnerRole(role))
```
oleg-ssvlabs, line 42: I don't think this should really affect the current dashboards, but I would double-check with @zktaiga.

@zktaiga the value of the `ssv.runner.role` attribute will no longer include the `_RUNNER` suffix.  
For example: `PROPOSER_RUNNER` → `PROPOSER`.

oleg-ssvlabs (approved):
lgtm. Left one comment

observability/attributes.go
iurii-ssv: Yeah that's a quality-of-life kind of thing, compare attrs:

image

image

PR #2486, opened 2025-09-02
fix(networkconfig): remove INTERVALS_PER_SLOT, change log level on missing parameter

nkryuchkov:
- remove `INTERVALS_PER_SLOT` as it is a constant in [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/fork-choice.md#constant), so it should not be a variable, especially given some nodes don't expose this parameter
- change the log level to debug when a beacon node doesn't expose some parameter, as we used to have it only once, and now we have it on every reconnection, and some parameters might not be exposed


iurii-ssv (dismissed):
LGTM, however I'm not sure about the change of `Warn`->`Debug` log-level ... @nkryuchkov could you explain "what it means for us to have to use the default value" (if that happens, how bad is it) ?

oleg-ssvlabs (approved):

nkryuchkov:
> > Well, those parameters are the same on all Ethereum networks, and we could just use them as constants in code (and we used to do it like that). The purpose of pulling them from the beacon node is to be able to change anything in a local testnet, and to have some kind of assertion that it's impossible to use the wrong parameters. Also, some parameters will probably be changed in the future forks (e.g., https://eips.ethereum.org/EIPS/eip-7782), so it's safer to pull them from the nodes and assert that nodes have the same values.
> > So nothing will happen if we use the default value, we'll just not be able to assert that the beacon node is configured correctly, but I doubt it could ever happen, I think all nodes expose parameters that might be changed, and if not, we can't do anything with that
> 
> That sounds like `Warn` is more appropriate level to log it at (as it's something that should never happen) ? And it will be simpler to see what's wrong with local **testnet** if we forget to set those params in there. I guess it's a nit either way (can address in another PR if needed).

It's unlikely to happen, but it's possible, and such a case is not an error. I didn't see any requirement regarding certain fields in the spec. Moreover, we return an error if any required fields are missing. If ones we are talking about are missing, it's not an issue, we'll just use the default value. I think WARN could be semantically incorrect here because it's not what we need to pay attention to



> **Update:** I guess we decided that this needs to go into **stage** branch instead

Rebased

iurii-ssv (approved):
Appreciate the detailed explanation on `Warn` vs `Debug` 👍

oleg-ssvlabs (approved):

PR #2488, opened 2025-09-03
node-probe: stabilize

iurii-ssv:
We can make our health-check `Prober` more robust by adding retries to it.

As per the comment in this PR:
```
	// Retry health-check multiple times to make sure we do not classify an occasional glitch (or a network blip)
	// as node being unhealthy. Failing on the very 1st failed request would be too drastic a measure given it
	// may result into SSV node restart.
```

And I think we have a `5s` timeout for CL-health check - which isn't much ... if SSV node has 1 CL configured we might hit it and crash. Note, it seems we cache the health-check results for EL (where request timeout is higher as well - at `10s`) but not CL. Still that just **reduces** the likelihood of SSV node crash due to health-check failure a bit, it doesn't solve for occasional network blips or CL/EL glitches.

Restarting SSV node while CL or EL are "having bad time" probably additionally makes it worse for them (since we send quite a bit of CL/EL requests right after SSV node start, I think).

In this PR I've started with simplifying `Prober` first, the implementation was way way harder than it needs to be to get the same functionality - so I didn't want to throw retries on top of what we had. Plus the logging is somewhat improved now (eg. `context.Canceled` shouldn't be logged as `error`).

### DIFF
--- a/api/handlers/node_test.go
+++ b/api/handlers/node_test.go
@@ -43,11 +43,8 @@ func CreateTestNode(t *testing.T, n int, ctx context.Context) *Node {
 
 	nodeMock := &NodeMock{}
 	nodeMock.HealthyMock.Store(nil)
-	nodeProber := nodeprobe.NewProber(zap.L(), nil, map[string]nodeprobe.Node{
-		"consensus client": nodeMock,
-		"execution client": nodeMock,
-		"event syncer":     nodeMock,
-	})
+	nodeProber := nodeprobe.NewProber(zap.L(), nodeMock, nodeMock)
+	nodeProber.AddEventSyncer(nodeMock)
 
 	return &Node{
 		ListenAddresses: []string{
--- a/api/handlers/node_test.go
+++ b/api/handlers/node_test.go
@@ -43,7 +43,7 @@ func CreateTestNode(t *testing.T, n int, ctx context.Context) *Node {
 
 	nodeMock := &NodeMock{}
 	nodeMock.HealthyMock.Store(nil)
-	nodeProber := nodeprobe.NewProber(zap.L())
+	nodeProber := nodeprobe.New(zap.L())
 	const node1 = "node_1"
 	const node2 = "node_2"
 	const node3 = "node_3"
--- a/api/handlers/node_test.go
+++ b/api/handlers/node_test.go
@@ -43,19 +43,27 @@ func CreateTestNode(t *testing.T, n int, ctx context.Context) *Node {
 
 	nodeMock := &NodeMock{}
 	nodeMock.HealthyMock.Store(nil)
-	nodeProber := nodeprobe.NewProber(zap.L(), nodeMock, nodeMock)
-	nodeProber.AddEventSyncer(nodeMock)
+	nodeProber := nodeprobe.NewProber(zap.L())
+	const node1 = "node_1"
+	const node2 = "node_2"
+	const node3 = "node_3"
+	nodeProber.AddNode(node1, nodeMock, 10*time.Second, 5)
+	nodeProber.AddNode(node2, nodeMock, 10*time.Second, 5)
+	nodeProber.AddNode(node3, nodeMock, 10*time.Second, 5)
 
-	return &Node{
-		ListenAddresses: []string{
+	return NewNode(
+		[]string{
 			fmt.Sprintf("tcp://%s:%d", "localhost", 3030),
 			fmt.Sprintf("udp://%s:%d", "localhost", 3030),
 		},
-		PeersIndex: ln.Nodes[0].(p2pv1.PeersIndexProvider).PeersIndex(),
-		Network:    ln.Nodes[0].(p2pv1.HostProvider).Host().Network(),
-		TopicIndex: ln.Nodes[0].(TopicIndex),
-		NodeProber: nodeProber,
-	}
+		ln.Nodes[0].(p2pv1.PeersIndexProvider).PeersIndex(),
+		ln.Nodes[0].(p2pv1.HostProvider).Host().Network(),
+		ln.Nodes[0].(TopicIndex),
+		nodeProber,
+		node1,
+		node2,
+		node3,
+	)
 }
 
 // createNetworkAndSubscribe creates a local network and subscribes each node to validator topics.
--- a/nodeprobe/mock_node.go
+++ b/nodeprobe/mock_node.go
@@ -1,0 +1,40 @@
+package nodeprobe
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+)
+
+type nodeMock struct {
+	healthy atomic.Pointer[error]
+}
+
+func (m *nodeMock) Healthy(context.Context) error {
+	err := m.healthy.Load()
+	if err != nil {
+		return *err
+	}
+	return nil
+}
+
+type stuckNodeMock struct {
+	healthy atomic.Pointer[error]
+}
+
+func (m *stuckNodeMock) Healthy(ctx context.Context) error {
+	<-ctx.Done() // stuck until the call is canceled
+	return ctx.Err()
+}
+
+type glitchyNodeMock struct {
+	calledCnt int
+}
+
+func (m *glitchyNodeMock) Healthy(ctx context.Context) error {
+	if m.calledCnt > 0 {
+		return nil
+	}
+	m.calledCnt++
+	return fmt.Errorf("got a glitch")
+}
--- a/nodeprobe/mock_node.go
+++ b/nodeprobe/mock_node.go
@@ -28,13 +28,13 @@ func (m *stuckNodeMock) Healthy(ctx context.Context) error {
 }
 
 type glitchyNodeMock struct {
-	calledCnt int
+	calledCnt atomic.Uint64
 }
 
 func (m *glitchyNodeMock) Healthy(ctx context.Context) error {
-	if m.calledCnt > 0 {
+	if m.calledCnt.Load() >= 2 {
 		return nil
 	}
-	m.calledCnt++
+	m.calledCnt.Add(1)
 	return fmt.Errorf("got a glitch")
 }
--- a/nodeprobe/nodeprobe.go
+++ b/nodeprobe/nodeprobe.go
@@ -20,7 +20,9 @@ type node interface {
 // pNode allows configuring the max number of retries intended for the node as well as the healthcheckTimeout
 // to use when checking for health.
 type pNode struct {
-	n                  node
+	name string
+	n    node
+
 	healthcheckTimeout time.Duration
 	retriesMax         int
 }
@@ -93,7 +95,7 @@ func (p *Prober) Probe(ctx context.Context, nodeName string) error {
 	return nil
 }
 
-func (p *Prober) probeNode(ctx context.Context, node pNode) (err error) {
+func (p *Prober) probeNode(ctx context.Context, n pNode) (err error) {
 	defer func() {
 		// Catch panics to present these (however unlikely they are) as a readable error-message.
 		if e := recover(); e != nil {
@@ -104,23 +106,29 @@ func (p *Prober) probeNode(ctx context.Context, node pNode) (err error) {
 	// Retry health-check multiple times to make sure we do not classify an occasional glitch (or a network blip)
 	// as node being unhealthy. Failing on the very 1st failed request would be too drastic a measure given it
 	// may result into SSV node restart.
-	for attempt := 0; attempt <= node.retriesMax; attempt++ {
+	attemptsMax := 1 + n.retriesMax
+	for attempt := 1; attempt <= attemptsMax; attempt++ {
 		err = func() error {
 			healthCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 			defer cancel()
 
-			return node.n.Healthy(healthCtx)
+			return n.n.Healthy(healthCtx)
 		}()
+		if errors.Is(err, context.Canceled) {
+			return nil // probing was canceled (it's not an error then)
+		}
 		if err == nil {
 			return nil // success
 		}
-	}
+		if attempt == attemptsMax {
+			break // all retries failed
+		}
 
-	// All retries failed.
-
-	if errors.Is(err, context.Canceled) {
-		// The caller canceled probing, it's not an error then.
-		return nil
+		p.logger.Debug("health-check failed, gonna retry",
+			zap.String("node", n.name),
+			zap.Int("attempt", attempt),
+			zap.Error(err),
+		)
 	}
 
 	return fmt.Errorf("node is unhealthy: %w", err)
--- a/nodeprobe/nodeprobe.go
+++ b/nodeprobe/nodeprobe.go
@@ -36,7 +36,7 @@ type Prober struct {
 	logger *zap.Logger
 
 	// nodesMu protects access to nodes, needed to handle the case when "event syncer" is added dynamically later on
-	// when the Proper is already running.
+	// when the Prober is already running.
 	nodesMu sync.Mutex
 	nodes   map[nodeName]Node
 }
@@ -85,10 +85,6 @@ func (p *Prober) Probe(ctx context.Context) error {
 
 	var wg sync.WaitGroup
 	errsCh := make(chan error)
-	go func() {
-		wg.Wait()
-		close(errsCh)
-	}()
 
 	p.nodesMu.Lock()
 	nodes := p.nodes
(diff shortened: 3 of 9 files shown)


nodeprobe/mock_node.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (m *stuckNodeMock) Healthy(ctx context.Context) error {
+	<-ctx.Done() // stuck until the call is canceled
+	return ctx.Err()
+}
+
+type glitchyNodeMock struct {
+	calledCnt int
+}
```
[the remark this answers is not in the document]
iurii-ssv: These tests aren't supposed to call this method concurrently, but since it's a simple adjustment I've fixed that

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
-	healthy          atomic.Bool
-	cond             *sync.Cond
-	unhealthyHandler func()
+	logger *zap.Logger
+
+	// nodesMu protects access to nodes, needed to handle the case when "event syncer" is added dynamically later on
+	// when the Prober is already running.
+	nodesMu sync.Mutex
+	nodes   map[nodeName]Node
 }
```
iurii-ssv: Removing a bunch of these fields here - this looks like a relic of over-engineering :)

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
-// Add a method to clients that returns their name or solve this in another way
 func (p *Prober) CheckBeaconNodeHealth(ctx context.Context) error {
 	p.nodesMu.Lock()
 	defer p.nodesMu.Unlock()
-	ctx, cancel := context.WithTimeout(ctx, p.interval)
-	defer cancel()
-	return p.nodes["consensus client"].Healthy(ctx)
+
+	return p.nodes[cl].Healthy(ctx)
 }
```
iurii-ssv: There is no need to set timeout here (and in similar methods) like that because we are making 1 CL call that has its own timeout on it, hence removing

api/handlers/node_test.go
```diff
@@ -43,11 +43,8 @@ func CreateTestNode(t *testing.T, n int, ctx context.Context) *Node {
 
 	nodeMock := &NodeMock{}
 	nodeMock.HealthyMock.Store(nil)
-	nodeProber := nodeprobe.NewProber(zap.L(), nil, map[string]nodeprobe.Node{
```
oleg-ssvlabs, line 46: I do think the previous implementation was slightly overcomplicated. Although, I believe it was designed this way to make it easier to extend dependencies/components that need to be health-checked continuously during runtime.

I think that was somewhat important because Prober does not only probe external dependencies (EL/CL) - which are rather static (it’s unlikely the Node would suddenly gain another external dependency it cannot function without). So this level of extensibility, at the cost of added complexity, might not really be needed. However, Prober also performs health checks for internal components, such as the `event syncer`. 

The semantics are slightly misleading, we call it `nodeprobe` (package name), which indicates it probes nodes (EL/CL), but in reality it also continuously evaluates internal components. Because of that, the extensible design does make sense to some degree, there’s a high chance that in the future we may need to health-check other internal component.

Wdyt @iurii-ssv? 

To me, this refactor is a good one - it made the code more readable, so I don’t have anything against it. Just outlining my thoughts here. Maybe in the future, it would make sense to separate a "node prober" from a "component prober" or so

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+	if errors.Is(err, context.Canceled) {
+		// The caller canceled probing, it's not an error then.
+		return nil
 	}
+
+	return fmt.Errorf("%s is unhealthy: %w", name, err)
 }
 
-func (p *Prober) AddNode(name string, node Node) {
+func (p *Prober) AddEventSyncer(node Node) {
```
oleg-ssvlabs: I somehow preferred the previous implementation more (I mean `AddNode(name string, node Node)`).  
The struct field `nodes map[nodeName]Node` is very agnostic, but the new method isn’t.  

If we were to add another component that needs to be probed, we would have to add a new method.  
With the previous implementation, the `Prober` object wouldn’t need to be modified at all — `AddNode` would just be called one more time.

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+	logger *zap.Logger
+
+	// nodesMu protects access to nodes, needed to handle the case when "event syncer" is added dynamically later on
+	// when the Prober is already running.
+	nodesMu sync.Mutex
+	nodes   map[nodeName]Node
 }
 
-func NewProber(logger *zap.Logger, unhealthyHandler func(), nodes map[string]Node) *Prober {
+func NewProber(logger *zap.Logger, consensusClient, executionClient Node) *Prober {
```
oleg-ssvlabs: this is subjective, but I would not take any nodes in the ctor. I would rather do this:

```Go
prober := NewProber(logger)
prober.Add(consensusClient)
prober.Add(executionClient)

prober.Probe()
...
prober.Add(eventSyncer)

prober.Start()
```

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
 	p.nodesMu.Lock()
 	defer p.nodesMu.Unlock()
 
-	p.nodes[name] = node
+	p.nodes[eventSyncer] = node
 }
 
-// TODO: string constants are error-prone.
-// Add a method to clients that returns their name or solve this in another way
 func (p *Prober) CheckBeaconNodeHealth(ctx context.Context) error {
```
oleg-ssvlabs: I would consider merging `CheckBeaconNodeHealth`, `CheckExecutionNodeHealth` and `CheckEventSyncerHealth` into a single method:

```error
func CheckNodeHealth(node nodeName) error { }
```

nodeprobe/nodeprobe_test.go
oleg-ssvlabs, line 1: If it is easy to assert on `log.Fatal()` (`os.Exit(1)`), it would be good to also have tests for the `Start()` method.

api/handlers/node_test.go
iurii-ssv: Yeah I know that the idea was to make the `Prober` generic and now (with this PR) it "implicitly encodes the components it's expected to work with, pretending as if it's generic" ... but I think now it's at least easy to see what it actually does :)

I will get back to it later to see whether I can address it (either make it generic, or make it more concrete instead)

nodeprobe/nodeprobe.go
iurii-ssv: To [my comment from above](https://github.com/ssvlabs/ssv/pull/2488#discussion_r2318856452) - these methods essentially break the `Prober` attempt to be a well abstracted/generic thing, but yeah looks like it's easy to fix ... on it

nodeprobe/nodeprobe_test.go
iurii-ssv: To me it seems like it isn't worth the effort given how simple the `Start()` method is, we'll probably need to additionally mock-out the `ctx`-timeout value (currently it's a simple globally defined constant) - so it's lots of minor hoops to jump over just to test a `for` loop with a `select` in it

if we add anything in there in the future (that makes the logic meaningfully grow in size/complexity), or use a bit more involved concurrency-primitives ... then sure we'd need to ensure it works as expected (because it wouldn't be easy to tell), but for now I wouldn't really want to do it just to get a 100% coverage

nodeprobe/nodeprobe.go
nkryuchkov: It's not a part of your changes, but it might be worth changing it now
```suggestion
func New(logger *zap.Logger, consensusClient, executionClient Node) *Prober {
```

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
-}
-
-func (p *Prober) Run(ctx context.Context) error {
-	ticker := time.NewTicker(p.interval)
+	ticker := time.NewTicker(probeFrequency)
 	defer ticker.Stop()
 
 	for {
-		p.probe(ctx)
+		func() {
```
nkryuchkov: Why do we need to wrap it with `func() {} ()`?

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+// Node represents a node being probed.
 type Node interface {
 	Healthy(ctx context.Context) error
 }
 
+const (
+	probeFrequency = 60 * time.Second
+)
+
+// Prober probes(monitors) the nodes it is configured with. It can do it periodically in the background via
```
nkryuchkov:
```suggestion
// Prober probes (monitors) the nodes it is configured with. It can do it periodically in the background via
```

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+// Start func (which can crash the process via logger.Fatal call if it finds out that one of the nodes being
+// probed is not healthy), or it can probe on demand via Probe func.
+// It is used to make sure the Ethereum nodes (CL, EL) are up and running, and they are healthy enough to for
```
nkryuchkov: Since it may also probe the event syncer, we need to mention it along with the Ethereum nodes

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
 
-	var healthy atomic.Bool
-	healthy.Store(true)
 	var wg sync.WaitGroup
+	errsCh := make(chan error)
+
 	p.nodesMu.Lock()
-	for name, node := range p.nodes {
+	nodes := p.nodes
+	p.nodesMu.Unlock()
```
nkryuchkov: The mutex guards only the map pointer, which always remains the same. It doesn't guard the map iteration

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+	// may result into SSV node restart.
+	const healthTimeout = 10 * time.Second
+	const maxAttempt = int(probeFrequency / healthTimeout)
+	for attempt := 1; attempt <= maxAttempt; attempt++ {
+		err = func() error {
+			healthCtx, cancel := context.WithTimeout(ctx, healthTimeout)
+			defer cancel()
+
+			return node.Healthy(healthCtx)
+		}()
```
nkryuchkov: Why do we need to wrap it into `func() error { ... } ()`?

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+		if e := recover(); e != nil {
+			err = fmt.Errorf("panic: %v", e)
+		}
+	}()
+
+	// Retry health-check multiple times to make sure we do not classify an occasional glitch (or a network blip)
+	// as node being unhealthy. Failing on the very 1st failed request would be too drastic a measure given it
+	// may result into SSV node restart.
+	const healthTimeout = 10 * time.Second
+	const maxAttempt = int(probeFrequency / healthTimeout)
```
olegshmuelov: Minor suggestion: Consider using ceiling division for future-proofing in case these constants are changed:
```go
const maxAttempt = int((probeFrequency + healthTimeout - 1) / healthTimeout)
```
The current formula works perfectly with 60s/10s = 6, but if someone changes these to values that don't divide evenly (e.g., 65s/10s), we'd get 6 attempts instead of 7.

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,186 @@ package nodeprobe
...
+	for attempt := 1; attempt <= maxAttempt; attempt++ {
+		err = func() error {
+			healthCtx, cancel := context.WithTimeout(ctx, healthTimeout)
+			defer cancel()
+
+			return node.Healthy(healthCtx)
+		}()
+		if err == nil {
+			return nil // success
+		}
```
olegshmuelov, line 122: Consider adding debug logging when retry attempts fail. This would help diagnose transient network issues:

```go
if err != nil && attempt < maxAttempt {
    p.logger.Debug("health check retry",
        zap.String("node", string(name)),
        zap.Int("attempt", attempt),
        zap.Error(err))
}
```
Currently, if a node experiences temporary issues but recovers after retries, there's no visibility. This logging would help with:
- Understanding network stability
- Diagnosing intermittent issues
- Monitoring retry frequency

This is optional but would improve observability significantly.

olegshmuelov (commented):
Great refactoring!

Left 2 minor suggestions (future-proofing integer division & debug logging for retries) but neither are blocking.

nodeprobe/nodeprobe.go
iurii-ssv: Yup, reworked this as you suggested

nodeprobe/nodeprobe_test.go
iurii-ssv: note btw, now I've completely moved the `Start()` out of the `Prober` now since it contains the `Fatal` call that
- prevents us from re-using this `nodeprobe` component elsewhere
- plus it mentions "Ethereum" in the error message (and we can't remove it because we need a clear error message for Operators to make sense of)

Imagine if I had unit-tests to worry about :)

nodeprobe/nodeprobe.go
iurii-ssv: Reworked this as suggested

api/handlers/node_test.go
iurii-ssv: Reworked this as suggested

nodeprobe/nodeprobe.go
iurii-ssv: To be able to call `defer cancel()` just once and never worry about it, I mean ... here the difference isn't that significant - but in general I think it's a useful pattern we can use to keep the code clean 

@oleg-ssvlabs @olegshmuelov any opinions on this ?

nodeprobe/nodeprobe.go
iurii-ssv: https://github.com/ssvlabs/ssv/pull/2488#discussion_r2322684380

nodeprobe/nodeprobe.go
nkryuchkov: Anyway, we call `cancel` when it's never used anymore with this wrapping. Indeed, it would be canceled earier, but I doubt it would change the behavior

nkryuchkov: I looked at the other case while writing the above. Actually, that's not true because of the for loop. Alright, I'm resolving this

nodeprobe/nodeprobe.go
iurii-ssv: Nice catch! Fixed with `hashmap` now

nodeprobe/nodeprobe.go
iurii-ssv: I was actually aiming for the "floor" rounding, but it doesn't matter for this particular case really (given it's 6 vs 7 retries),

now it's no longer relevant since I've reworked it as per [Oleg's suggestion to make `Prober` generic(abstract)](https://github.com/ssvlabs/ssv/pull/2488#discussion_r2318717550) - and the amount of attempts (and the timeout of an attempt) is something `Prober` shouldn't make assumptions about (now it accepts those parameters on per-added-node basis)

nodeprobe/nodeprobe.go
iurii-ssv: I've moved the Ethereum-specfic functionality out of `Prober` now, as per [Oleg's suggestion to make it generic(abstract)](https://github.com/ssvlabs/ssv/pull/2488#discussion_r2318717550)

nodeprobe/nodeprobe.go
iurii-ssv: Added now in https://github.com/ssvlabs/ssv/pull/2488/commits/e344b4c5a18b42980eda1e27664bf0b7351e0a2f

nodeprobe/nodeprobe.go
iurii-ssv: @nkryuchkov also note, now that I've added logging the `func {} ()` is helping with that as well:
```go
		func() {
			logger.Debug("node-prober tick: probing all nodes")
			defer logger.Debug("node-prober tick: probing all nodes done")

			probeCtx, cancel := context.WithTimeout(ctx, probeFrequency)
			defer cancel()

			if err := p.ProbeAll(probeCtx); err != nil {
				logger.Fatal("Ethereum node(s) are either out of sync or down. Ensure the nodes are healthy to resume.", zap.Error(err))
			}
		}()
```

we could potentially move this out into a dedicated `func` defined somewhere below - but since all this code fits in a ~1-2 screens inlining like that reads better I think (no need to jump around between various funcs)

iurii-ssv:
Closing in favor of https://github.com/ssvlabs/ssv/pull/2492 (as per https://github.com/ssvlabs/ssv/pull/2488#issuecomment-3253616975)

PR #2491, opened 2025-09-03
qbft: fix flaky TestTimeoutForRound

iurii-ssv:
This PR should fix the `TestTimeoutForRound` test which failed for me locally as follows:
```
--- FAIL: TestTimeoutForRound (3.42s)
    --- FAIL: TestTimeoutForRound/TimeoutForRound_-_SYNC_COMMITTEE_CONTRIBUTION_RUNNER:_<=_quickTimeoutThreshold (0.19s)
        timer_test.go:89:
            	Error Trace:	/go/src/github.com/ssvlabs/ssv/protocol/v2/qbft/roundtimer/timer_test.go:89
            	            				/go/src/github.com/ssvlabs/ssv/protocol/v2/qbft/roundtimer/timer_test.go:28
            	Error:      	Not equal:
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestTimeoutForRound/TimeoutForRound_-_SYNC_COMMITTEE_CONTRIBUTION_RUNNER:_<=_quickTimeoutThreshold
FAIL
coverage: 72.1% of statements
FAIL	github.com/ssvlabs/ssv/protocol/v2/qbft/roundtimer	3.433s
``` 

I haven't looked into it super deep, but seems to me 10ms is just too optimistic (making it 100ms)

### DIFF
--- a/protocol/v2/qbft/roundtimer/timer_test.go
+++ b/protocol/v2/qbft/roundtimer/timer_test.go
@@ -21,7 +21,9 @@ const (
 	quickTimeout = 300 * time.Millisecond
 	slowTimeout  = 600 * time.Millisecond
 
-	goSchedulerDelayMax = 100 * time.Millisecond
+	// safeTestDelay is a safety buffer (a guesstimate) to use in tests that set expectations making certain
+	// assumptions regarding the go-routine scheduling.
+	safeTestDelay = 100 * time.Millisecond
 )
 
 func TestTimeoutForRound(t *testing.T) {
@@ -95,7 +97,7 @@ func testTimeoutForRound(t *testing.T, role spectypes.RunnerRole, threshold spec
 	require.Equal(t, int32(0), atomic.LoadInt32(&count))
 
 	timer.TimeoutForRound(specqbft.FirstHeight, threshold)
-	<-time.After(timer.RoundTimeout(specqbft.FirstHeight, threshold) + goSchedulerDelayMax)
+	<-time.After(timer.RoundTimeout(specqbft.FirstHeight, threshold) + safeTestDelay)
 	require.Equal(t, int32(1), atomic.LoadInt32(&count))
 }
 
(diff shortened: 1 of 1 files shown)


nkryuchkov (approved):
Should we extract the timeout into a constant to avoid forgetting to change something if we decide to change it again?

protocol/v2/qbft/roundtimer/timer_test.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/ssvlabs/ssv/networkconfig"
 )
 
+const (
+	goSchedulerDelayMax = 100 * time.Millisecond
```
nkryuchkov: Can you please add a comment on how you came up with this value? It looks like a magic number now, so IMO it's worth commenting

nkryuchkov: Well, there's some reason behind other durations in this file too, but I don't think they require commenting. I think only `goSchedulerDelayMax` might be confusing because it's not clear what it refers to: Is it some internal scheduler duration in Go? Or is it some value we came up with as the best one to make tests pass?

nkryuchkov: Or maybe we could just rename it to something like `safeFrameworkDelay` to indicate it's some safe delay to wait for the test framework to finish

iurii-ssv: Well, that's a purely observational estimate - renamed to `safeTestDelay` and added a comment to clarify that it's not a hard-defined constant enforced by Golang (or anything like that):
```
	// safeTestDelay is a safety buffer (a guesstimate) to use in tests that set expectations making certain
	// assumptions regarding the go-routine scheduling.
	safeTestDelay = 100 * time.Millisecond
```

technically go-scheduler is not the only contributing component, sometimes maybe even not the main one, eg. if the go process itself doesn't get scheduled for execution by the OS for long enough - the time is ticking but has nothing to do with the Go Scheduler ... so `goSchedulerDelayMax` isn't really precise indeed.

Based on my past experience of dealing with flaky-tests I feel like `100 * time.Millisecond` is safe enough - I wouldn't want to go any lower than that, and setting it to a higher value increases tests-runtime (so we could bump it to higher value for tests that actually need it, if we observe `100 * time.Millisecond` not being enough)

nkryuchkov (approved):

oleg-ssvlabs (approved):
lgtm

PR #2492, opened 2025-09-05
node-probe: stabilize & cleanup

iurii-ssv:
This is a carbon-copy of https://github.com/ssvlabs/ssv/pull/2488 rebased onto **stage** branch


cli/operator/prober.go
```diff
@@ -0,0 +1,44 @@
...
+	for {
+		func() {
+			logger.Debug("node-prober tick: probing all nodes")
+			defer logger.Debug("node-prober tick: probing all nodes done")
+
+			probeCtx, cancel := context.WithTimeout(ctx, probeFrequency)
+			defer cancel()
+
+			if err := p.ProbeAll(probeCtx); err != nil {
+				logger.Fatal("Ethereum node(s) are either out of sync or down. Ensure the nodes are healthy to resume.", zap.Error(err))
```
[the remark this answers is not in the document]
iurii-ssv: This is working as intended in this case

api/handlers/node_test.go
```diff
@@ -43,22 +43,27 @@ func CreateTestNode(t *testing.T, n int, ctx context.Context) *Node {
...
-
-	return &Node{
-		ListenAddresses: []string{
+	nodeProber := nodeprobe.New(zap.L())
+	const node1 = "node_1"
+	const node2 = "node_2"
+	const node3 = "node_3"
+	nodeProber.AddNode(node1, nodeMock, 10*time.Second, 5)
+	nodeProber.AddNode(node2, nodeMock, 10*time.Second, 5)
+	nodeProber.AddNode(node3, nodeMock, 10*time.Second, 5)
```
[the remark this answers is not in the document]
iurii-ssv: This is just to mock it out for tests

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,142 @@ package nodeprobe
...
-		unhealthyHandler: unhealthyHandler,
-		interval:         probeInterval,
-		nodes:            nodes,
-		cond:             sync.NewCond(&sync.Mutex{}),
-	}
-}
+// pNode allows configuring the max number of retries intended for the node as well as the healthcheckTimeout
+// to use when checking for health.
+type pNode struct {
+	name string
```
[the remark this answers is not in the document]
iurii-ssv: Good find! Tests don't covert this because the `name` field is only used for logging

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,143 @@ package nodeprobe
...
 	}
+
+	return nil
 }
 
-func (p *Prober) AddNode(name string, node Node) {
-	p.nodesMu.Lock()
-	defer p.nodesMu.Unlock()
+func (p *Prober) probeNode(ctx context.Context, n pNode) (err error) {
+	defer func() {
```
nkryuchkov: Why do we need to catch panics here?

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,143 @@ package nodeprobe
...
+			err = fmt.Errorf("panic: %v", e)
+		}
+	}()
 
-	p.nodes[name] = node
-}
+	// Retry health-check multiple times to make sure we do not classify an occasional glitch (or a network blip)
+	// as node being unhealthy. Failing on the very 1st failed request would be too drastic a measure given it
+	// may result into SSV node restart.
+	attemptsMax := 1 + n.retriesMax
```
nkryuchkov: Can you add a comment on why it's 1 more than `n.retriesMax`?

oleg-ssvlabs (dismissed):
I personally really like the new implementation of the prober. Generic and simple at the same time 👍

nkryuchkov (dismissed):

nodeprobe/nodeprobe.go
iurii-ssv: Added a clarifying comment, it all comes down to naming/language we use:
- if you call it retry (retries) it technically means "it is repeating something tried previously" (so this initial request is where 1 comes from)
- we could alternatively `retriesMax` -> `callsMax` ... but it's all not of much importance anyway really

nodeprobe/nodeprobe.go
iurii-ssv: We probably don't - removed now, having considered the implications (I've copied it over as is, to deal with later).

I think it was previously trying to be defensive about this (which I don't really agree with - I'd rather we fail fast and fix the root cause - since in this case it will not be super hard to encounter):
```go
			map[string]nodeprobe.Node{
				"execution client": executionClient,

				// Underlying options.Beacon's value implements nodeprobe.StatusChecker.
				// However, as it uses spec's specssv.BeaconNode interface, avoiding type assertion requires modifications in spec.
				// If options.Beacon doesn't implement nodeprobe.StatusChecker due to a mistake, this would panic early.
				"consensus client": consensusClient,
			},
```

nkryuchkov (dismissed):

nodeprobe/nodeprobe.go
```diff
@@ -2,164 +2,136 @@ package nodeprobe
...
-	return p.nodes["consensus client"].Healthy(ctx)
-}
+func (p *Prober) probeNode(ctx context.Context, n pNode) (err error) {
+	// Retry health-check multiple times to make sure we do not classify an occasional glitch (or a network blip)
+	// as node being unhealthy. Failing on the very 1st failed request would be too drastic a measure given it
+	// may result into SSV node restart.
+	attemptsMax := 1 + n.retriesMax // the initial attempt + retries specified
+	for attempt := 1; attempt <= attemptsMax; attempt++ {
+		err = func() error {
+			healthCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
olegshmuelov: The `healthcheckTimeout` field is stored in the `pNode` struct but never used. The timeout is hardcoded to `10*time.Second` instead.

```suggestion
			healthCtx, cancel := context.WithTimeout(ctx, n.healthcheckTimeout)
```

olegshmuelov (approved):
LGTM!
Just fix the unused `healthcheckTimeout` field issue

nodeprobe/nodeprobe.go
iurii-ssv: Well spotted! That was indeed the intent, applied in https://github.com/ssvlabs/ssv/pull/2492/commits/032af6938d6cf1457aa8f3ecf177ef305511808c (+ added a test to cover this)

oleg-ssvlabs (approved):

PR #2493, opened 2025-09-05
fix(multi client submit) - return wrapped error to the caller

oleg-ssvlabs:
## Background

`multiClientSubmit` is used for submitting attestations and sync committee messages **when** parallel submissions are enabled in the config, and is **always** used for submitting proposals and subscriptions (is that by design? see the [description](https://github.com/ssvlabs/ssv/blob/stage/beacon/goclient/options.go#L15) of the config).  

Previously, `multiClientSubmit` did not return errors (usually originating from `go-eth2-client` calls). Instead, it simply logged them. In many cases, the caller had valuable context for the call (such as the Duty ID), so errors logged only in `multiClientSubmit` were disconnected from the overall flow while looking into the logs.

## Changes
- Return wrapped errors from `multiClientSubmit`  
- Move the `multiClientSubmit` method from `attest.go` to `goclient.go` (a more appropriate location)

### DIFF
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"sync/atomic"
 	"time"
 
 	"github.com/attestantio/go-eth2-client/api"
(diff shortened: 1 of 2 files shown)


oleg-ssvlabs:
Closing in favor of #2472

PR #2494, opened 2025-09-06
CL: correct multi-client usage (Events)

iurii-ssv:
CL `Events`-handling needs to be reworked a bit to address the following issue:

- (the SSV node startup requirements) if Operator has multiple CLs configured with `WithWeightedAttestationData=true` ALL of them must respond to `Events` request on SSV node startup

Instead, just 1 successful CL response is sufficient for SSV node startup to continue (the rest will auto-connect eventually due to retries implemented in `go-eth2-client`). It is just a sanity-check really, we could even continue with SSV node startup **without waiting for any successful response(s)** and let the `Prober` to take care of health-checking ... but I'm not changing that behavior in this PR since it might some additional adjustments somewhere + more testing.


beacon/goclient/events.go
```diff
@@ -84,95 +92,138 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
-	if e.Data == nil {
-		logger.Warn("event data is nil")
-		return
-	}
+// eventHandler creates a handler to process CL events.
+// IMPORTANT: this func is called concurrently by different CL client implementations, hence the handling it
+// performs must be thread-safe and idempotent (it can receive the same event multiple times).
+func (gc *GoClient) newEventHandler() func(*apiv1.Event) {
+	var lastProcessedEventSlotLock sync.Mutex
+	var lastProcessedEventSlot phase0.Slot
```
iurii-ssv: There is no reason for these fields to stay on `GoClient` struct, so I'm moving these here to limit the scope for these (they are only really used to ensure idempotent & thread-safe handling of 1 event-type below)

beacon/goclient/events.go
```diff
@@ -84,95 +92,138 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
-		}
-	} else {
-		if err := gc.multiClient.Events(ctx, opts); err != nil {
+	// By default, we subscribe to CL events via multiClient that only propagates events from a single client
+	// (the 1st active client). If we have weighted attestations enabled, we need to process events from all
+	// clients GoClient is configured with to get the most up-to-date data to score attestations correctly
+	// (and if some clients are unavailable at the moment - it's not a big deal, `go-eth2-client` automatically
+	// periodically resubscribes to Events).
+	// There is no need to issue these calls asynchronously since they spawn their own go-routines are return fast.
+	subscribeToEvents := func() error {
```
nkryuchkov, line 107: Why do we need to define `subscribeToEvents` and then overwrite it conditionally? Can we just run the code inside the if-else?

beacon/goclient/events.go
```diff
@@ -84,95 +92,138 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
+		if err != nil {
 			logger.Error(clResponseErrMsg, zap.String("api", "Events"), zap.Error(err))
 			return err
 		}
+		return nil
 	}
+	if gc.withWeightedAttestationData {
+		subscribeToEvents = func() error {
+			var errs error
+			success := false
```
nkryuchkov, line 118: Do we need `success`? If I understand correctly, `errs` is `nil` on success

iurii-ssv: > If I understand correctly, `errs` is `nil` on success

Nope, when we are performing the `if !success` we might have anywhere from 0 to N errors accumulated in `errs` ( `errs` is `nil` would mean **total** success for every call - but that's too high of a requirement to have)

We could potentially do something like `if (len(errs.Unwrap()) == len(gc.clients)` instead of defining `success` ... but that's ugly and not super self-explanatory

beacon/goclient/events.go
iurii-ssv: True, we could use `if-else`, however I like the func-definition pattern more because:
- it assigns clear name to the block of code we are executing (vs when you read through `else` clause - it might contain whatever)
- here it's not super relevant, but if we were to handle errors with `errors.Is`, for example, we'll be able to dedup this automatically

it mostly comes down to a personal preference (I wouldn't try to use it to replace every `else` clause ofc, but here it seems like a natural candidate for that)

beacon/goclient/events.go
```diff
@@ -84,95 +92,138 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
+}
 
-	if e.Data == nil {
-		logger.Warn("event data is nil")
-		return
-	}
+// eventHandler creates a handler to process CL events.
+// IMPORTANT: this func is called concurrently by different CL client implementations, hence the handling it
+// performs must be thread-safe and idempotent (it can receive the same event multiple times).
+func (gc *GoClient) newEventHandler() func(*apiv1.Event) {
```
julienharbulot, line 150: Doc comment is out of sync with the new function.

beacon/goclient/events.go
nkryuchkov: Ah, I see, indeed

nkryuchkov (approved):

beacon/goclient/events.go
nkryuchkov: Ah, I see, indeed

PR #2496, opened 2025-09-07
metrics: (qbft-instance) simplify and standardize

iurii-ssv:


protocol/v2/qbft/instance/metrics.go
```diff
@@ -42,7 +45,8 @@ func (m *metricsRecorder) RecordRoundChange(ctx context.Context, round qbft.Roun
 		ctx,
 		1,
 		metric.WithAttributes(
-			roleAttribute(m.role),
+			observability.RunnerRoleAttribute(m.runnerRole),
 			observability.DutyRoundAttribute(round),
-			reasonAttribute(reason)))
+			reasonAttribute(reason)),
+	)
```
iurii-ssv, line 52: Getting rid of `roleAttribute` func so we can:
- re-use existing func
- record the metric with trimmed `_RUNNER` suffix (so it looks the same as on other charts)

protocol/v2/qbft/instance/metrics.go
```diff
@@ -5,23 +5,26 @@ import (
...
+func newMetrics(runnerRole spectypes.RunnerRole) *metricsRecorder {
 	return &metricsRecorder{
-		role: role,
+		runnerRole: runnerRole,
 	}
 }
 
-func (m *metricsRecorder) StartStage() {
+// Start records the start of stage(phase) 1 for QBFT instance (it's either a proposal or prepare).
+func (m *metricsRecorder) Start() {
```
julienharbulot, line 27: Quick naming question: what’s the rationale for StartStage → Start while stageStart and EndStage still include Stage? The earlier API was symmetric; this change introduces inconsistency. If the receiver/type already implies “Stage,” I’m +1 on Start/End, but then we should update the field and EndStage accordingly.

iurii-ssv: The thing is `StartStage` doesn't really start a stage (of which we have 3 - proposal, prepare, commit), like I explain in the `// Start` comment - it really only start the 1st stage - so the `StartStage` name is misleading (since it would imply it can start a `commit` stage, for example, but it can't)

the **new** stage is started with the `EndStage` method (the stage 1 is an exception and it is started with `Start` method) ... this is a bit unintuitive/overly-complicated maybe but I'm not planning on changing that behavior in this PR (since it's not trivial to change)

so this should actually resolve inconsistency/misleading-nature of `StartStage` method, not introduce it, does it make sense ?

julienharbulot: This makes sense I hadn't seen that EndStage starts a new stage too

PR #2498, opened 2025-09-08
refactor(exporter): remove pubkeys from collector (low risk PR)

julienh-ssv:
## Summary

Refactors the exporter to prepare for an upcoming fix around the validator store and pubkeys. The main goal is to stop passing full pubkeys through the pipeline and rely on indices instead, reducing memory use and simplifying code paths.

## Review guidelines

Please **focus on regressions and correctness**. This PR is intentionally scoped to mechanical refactors and a small bugfix; broader refactors or nice-to-haves are out of scope. If you spot improvements, please open a follow-up issue/PR so we can keep this change set contained.

## What changed

1 -- Replace validator pubkeys in the collector with indices (leaner memory footprint, simpler logic).
- Push index⇄pubkey conversion to the edges of the system (pubkeys are on track for deprecation).

2 -- Rename package model → exporter for clearer imports and intent.

3 -- `/exporter/decideds` endpoint:
- Accepts additional indices (aligns behavior with other endpoints).
- Fixes a bug where committeeID wasn’t populated in responses.

## Impact

Behavior: No breaking changes expected; the endpoint input is more permissive and the response is now correct for committeeID.

Performance: Lower memory usage by avoiding full pubkey objects in hot paths.

Code health: Clearer boundaries; conversions live at the edges.

## Migration notes

Update imports from …/model to …/exporter.

If any internal callers depended on committeeID being empty (unlikely), they’ll now receive the correct value.

## Next steps

Follow-up PR will address the validator store/pubkey bug fully and complete the pubkey deprecation path.

### DIFF
--- a/api/handlers/exporter.go
+++ b/api/handlers/exporter.go
@@ -449,7 +449,7 @@ func (e *Exporter) getValidatorDutiesForRoleAndSlot(role spectypes.BeaconRole, s
 		if err == nil {
 			result.CommitteeID = &committeeID
 		} else if !isNotFoundError(err) {
-			// if error is not found, nothing to report to prevent unnecessary noise
+			// if error is not found, nothing to report to prevent unnecessary noise, otherwise log the error:
 			e.logger.Debug("error getting committee ID", zap.Error(err), fields.Slot(slot), fields.ValidatorIndex(idx))
 			errs = multierror.Append(errs, err)
 		}
--- a/exporter/model.go
+++ b/exporter/model.go
@@ -119,7 +119,7 @@ func (trace *ProposalTrace) DeepCopy() *ProposalTrace {
 }
 
 type QBFTTrace struct {
-	Round        uint64      // same for
+	Round        uint64
 	BeaconRoot   phase0.Root `ssz-size:"32"`
 	Signer       spectypes.OperatorID
 	ReceivedTime uint64
@@ -165,10 +165,10 @@ func (trace *CommitteeDutyTrace) DeepCopy() *CommitteeDutyTrace {
 	if trace == nil {
 		return nil
 	}
-	// copy operator IDs
+
 	opIDs := make([]spectypes.OperatorID, len(trace.OperatorIDs))
 	copy(opIDs, trace.OperatorIDs)
-	// copy proposal data
+
 	pd := make([]byte, len(trace.ProposalData))
 	copy(pd, trace.ProposalData)
 
@@ -188,7 +188,7 @@ func (trace *CommitteeDutyTrace) DeepCopy() *CommitteeDutyTrace {
 
 type SignerData struct {
 	Signer       spectypes.OperatorID
-	ValidatorIdx []phase0.ValidatorIndex `ssz-max:"1000"`
+	ValidatorIdx []phase0.ValidatorIndex `ssz-max:"3000"`
 	ReceivedTime uint64
 }
 
@@ -197,10 +197,13 @@ func (data *SignerData) DeepCopy() *SignerData {
 	if data == nil {
 		return nil
 	}
-	// Note: ValidatorIdx is intentionally shallow-copied to match existing behavior.
+
+	validatorIdx := make([]phase0.ValidatorIndex, len(data.ValidatorIdx))
+	copy(validatorIdx, data.ValidatorIdx)
+
 	return &SignerData{
 		Signer:       data.Signer,
-		ValidatorIdx: data.ValidatorIdx,
+		ValidatorIdx: validatorIdx,
 		ReceivedTime: data.ReceivedTime,
 	}
 }
@@ -224,7 +227,7 @@ func (trace *ValidatorDutyTrace) DeepCopy() *ValidatorDutyTrace {
 	if trace == nil {
 		return nil
 	}
-	// ProposalData copy
+
 	pd := make([]byte, len(trace.ProposalData))
 	copy(pd, trace.ProposalData)
 
--- a/operator/dutytracer/collector.go
+++ b/operator/dutytracer/collector.go
@@ -593,7 +593,7 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 			index, found := c.validators.ValidatorIndex(validatorPK)
 			if !found {
 				c.logger.Error("validator not found by pubkey", fields.Validator(validatorPK[:]))
-				return nil
+				return fmt.Errorf("validator not found by pubkey: %x", validatorPK[:])
 			}
 
 			trace, late, err := c.getOrCreateValidatorTrace(slot, bnRole, index)
--- a/exporter/model_test.go
+++ b/exporter/model_test.go
@@ -1,0 +1,47 @@
+package exporter
+
+import (
+	"math/rand"
+	"reflect"
+	"testing"
+
+	fuzz "github.com/google/gofuzz"
+)
+
+var deepCopyTypes = []reflect.Type{
+	reflect.TypeOf(DecidedTrace{}),
+	reflect.TypeOf(RoundTrace{}),
+	reflect.TypeOf(RoundChangeTrace{}),
+	reflect.TypeOf(ProposalTrace{}),
+	reflect.TypeOf(QBFTTrace{}),
+	reflect.TypeOf(CommitteeDutyTrace{}),
+	reflect.TypeOf(SignerData{}),
+	reflect.TypeOf(ValidatorDutyTrace{}),
+	reflect.TypeOf(PartialSigTrace{}),
+}
+
+func TestDeepCopy(t *testing.T) {
+	f := fuzz.New().NilChance(0).NumElements(0, 5).RandSource(rand.NewSource(0))
+	for _, typ := range deepCopyTypes {
+		t.Run(typ.Name(), func(t *testing.T) {
+			orig := reflect.New(typ)
+			f.Fuzz(orig.Interface())
+
+			copyVal := orig.MethodByName("DeepCopy").Call(nil)[0]
+
+			if orig.Pointer() == copyVal.Pointer() {
+				t.Fatalf("%s DeepCopy returned same pointer", typ.Name())
+			}
+
+			if !reflect.DeepEqual(orig.Interface(), copyVal.Interface()) {
+				t.Fatalf("%s DeepCopy mismatch", typ.Name())
+			}
+
+			nilVal := reflect.Zero(reflect.PointerTo(typ))
+			nilCopy := nilVal.MethodByName("DeepCopy").Call(nil)[0]
+			if !nilCopy.IsNil() {
+				t.Fatalf("%s DeepCopy nil receiver expected nil", typ.Name())
+			}
+		})
+	}
+}
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/go-chi/render v1.0.2
 	github.com/golang/gddo v0.0.0-20200528160355-8d077c1d8f4c
 	github.com/google/go-cmp v0.7.0
+	github.com/google/gofuzz v1.2.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.3
 	github.com/hashicorp/golang-lru/v2 v2.0.7


exporter/model.go
```diff
@@ -33,6 +33,21 @@ type DecidedTrace struct {
...
+	signers := make([]spectypes.OperatorID, len(trace.Signers))
+	copy(signers, trace.Signers)
+	return &DecidedTrace{
+		Round:        trace.Round,
+		BeaconRoot:   trace.BeaconRoot,
+		Signers:      signers,
+		ReceivedTime: trace.ReceivedTime,
+	}
+}
+
```
julienharbulot, line 50: DeepCopy functions were removed from the `dutytracer/store.go` file to be closer to the corresponding type definitions. Further changes would be out of scope as the goal of this PR is not to improve on this in any way.

exporter/api/decided/stream.go
```diff
@@ -36,22 +37,29 @@ func NewStreamPublisher(logger *zap.Logger, domainType spectypes.DomainType, ws
...
 		participation := qbftstorage.Participation{
 			ParticipantsRangeEntry: qbftstorage.ParticipantsRangeEntry{
-				PubKey:  msg.PubKey,
+				PubKey:  share.ValidatorPubKey,
 				Slot:    msg.Slot,
 				Signers: msg.Signers,
 			},
 			Role:   msg.Role,
-			PubKey: msg.PubKey,
+			PubKey: share.ValidatorPubKey,
```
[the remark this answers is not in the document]
julienharbulot: That's just the way the types were designed. Not in scope for this PR.

api/handlers/exporter.go
```diff
@@ -180,31 +174,50 @@ func (e *Exporter) TraceDecideds(w http.ResponseWriter, r *http.Request) error {
...
+	for _, idx := range request.Indices {
+		indices = append(indices, phase0.ValidatorIndex(idx))
+	}
+	extra, merr := e.pubkeysToIndices(pubkeys)
+	errs = multierror.Append(errs, merr)
+	indices = append(indices, extra...)
+	slices.Sort(indices)
+	indices = slices.Compact(indices)
+
 	for _, r := range request.Roles {
```
iurii-ssv, line 188: Seems like we are dropping `errs` if we get any in case there is at least 1 participant we can return,

maybe it's fine but maybe we want to log `errs` - just calling it out in case it was missed (vs intentionally done this way to keep logs free from these errors)

api/handlers/exporter.go
```diff
@@ -358,65 +385,88 @@ func validateValidatorRequest(request *validatorRequest) error {
...
+		// best effort attempt to fill the CommitteeID field, non blocking if it fails
+		// as the duty itself is still valid without it
+		committeeID, err := e.traceStore.GetCommitteeID(slot, idx)
+		if err == nil {
+			result.CommitteeID = &committeeID
+		} else if !isNotFoundError(err) {
+			// if error is not found, nothing to report to prevent unnecessary noise
+			e.logger.Debug("error getting committee ID", zap.Error(err), fields.Slot(slot), fields.ValidatorIndex(idx))
+			errs = multierror.Append(errs, err)
+		}
```
iurii-ssv: perhaps `nothing to report to prevent unnecessary noise` -> `logging under Debug to prevent unnecessary noise` ?

api/handlers/model.go
```diff
@@ -10,7 +10,7 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
 	"github.com/ssvlabs/ssv/api"
-	model "github.com/ssvlabs/ssv/exporter"
+	"github.com/ssvlabs/ssv/exporter"
```
iurii-ssv, line 13: If `api/handlers/model.go` file contains exporter-relevant stuff only we could also rename `api/handlers/model.go` -> `api/handlers/exporter_model.go` for example

exporter/model.go
```diff
@@ -33,6 +33,21 @@ type DecidedTrace struct {
...
+		return nil
+	}
+	signers := make([]spectypes.OperatorID, len(trace.Signers))
+	copy(signers, trace.Signers)
+	return &DecidedTrace{
+		Round:        trace.Round,
+		BeaconRoot:   trace.BeaconRoot,
+		Signers:      signers,
+		ReceivedTime: trace.ReceivedTime,
+	}
```
iurii-ssv, line 48: Can maybe re-use `deepCopySlice` for `trace.Signers` here ?

exporter/model.go
```diff
@@ -84,12 +160,51 @@ type CommitteeDutyTrace struct {
...
+	if data == nil {
+		return nil
+	}
+	// Note: ValidatorIdx is intentionally shallow-copied to match existing behavior.
+	return &SignerData{
+		Signer:       data.Signer,
+		ValidatorIdx: data.ValidatorIdx,
+		ReceivedTime: data.ReceivedTime,
+	}
+}
```
iurii-ssv: I think we maybe want to change this shallow-copy behavior to avoid shooting ourselves in the foot later on ?
- this `DeepCopy` doesn't seem to be used anywhere for `SignerData`, maybe we could remove the `DeepCopy` method entirely ... it's not like "we must have it"
- it's simpler to debug high memory usage or slow-down somewhere than corrupted data that could result from it

**Update 1:** ah never mind, I see how it's used via `deepCopySlice` now ... so I would just maybe rephrase this `ValidatorIdx is intentionally shallow-copied to match existing behavior.` comment to something like `ValidatorIdx is intentionally shallow-copied to accommodate for better performance, this is OK because slice elements are never modified "in-place"`

**Update 2:** I've also noticed that we have `ssz-max:"1000"` there - I think we should probably bump it up to 3000 while we are here given we are trying to up the amount of supported validators (1000 -> 3000)

operator/dutytracer/collector.go
```diff
@@ -591,7 +589,14 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 				return err
 			}
 
-			trace, late, err := c.getOrCreateValidatorTrace(slot, bnRole, validatorPK)
+			// map pubkey to validator index for internal storage
+			index, found := c.validators.ValidatorIndex(validatorPK)
+			if !found {
+				c.logger.Error("validator not found by pubkey", fields.Validator(validatorPK[:]))
+				return nil
+			}
```
iurii-ssv: Why log error & `return nil` ? Shouldn't we just `return err` (wrapping it with `"validator not found by pubkey: %s: %w") ?

api/handlers/exporter.go
julienharbulot: Actually, the test says `if !isNotFoundError(err)` (note the `!`) and the comment explains why we filter out this case. I'll add "otherwise log the error" to the comment to be more explicit

api/handlers/model.go
julienharbulot: Yes but we've already planned to refactor the api/handlers to create one package per handler so this will change very soon anyways

exporter/model.go
julienharbulot: Not possible because OperatorID is an alias to a native type and does not implement hasDeepCopy

exporter/model.go
julienharbulot: Good suggestions, I made the changes but honestly I had explicitly said that improving these methods was not in scope for this PR ^^

api/handlers/exporter.go
julienharbulot: Not sure I understand correctly what you mean: `merr` is not dropped it's accumulated in `errs` like all the other errors of this function. `errs` is then merge in the response under best effort batch processing with partial results.

operator/dutytracer/collector.go
julienharbulot: nice catch thanks

exporter/model.go
iurii-ssv: Ah, okay, 

for completeness kinda feels like we need a twin-func to complement `deepCopySlice` (call it `copySlice` maybe) IF we can make it generic - we are using `copy` multiple times

exporter/model.go
```diff
@@ -84,12 +160,54 @@ type CommitteeDutyTrace struct {
...
 }
 
+// DeepCopy returns a deep copy of the SignerData.
+func (data *SignerData) DeepCopy() *SignerData {
+	if data == nil {
+		return nil
+	}
+
+	validatorIdx := make([]phase0.ValidatorIndex, len(data.ValidatorIdx))
+	copy(validatorIdx, data.ValidatorIdx)
```
iurii-ssv, line 202: Do you think we have to copy it though ? From perf perspective this might be quite a bit of overhead, but it seems we are never actually modifying any of the slice objects "in-place"

operator/dutytracer/collector.go
```diff
@@ -591,7 +589,14 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 				return err
 			}
 
-			trace, late, err := c.getOrCreateValidatorTrace(slot, bnRole, validatorPK)
+			// map pubkey to validator index for internal storage
+			index, found := c.validators.ValidatorIndex(validatorPK)
+			if !found {
+				c.logger.Error("validator not found by pubkey", fields.Validator(validatorPK[:]))
+				return fmt.Errorf("validator not found by pubkey: %x", validatorPK[:])
+			}
```
iurii-ssv, line 597: probably no need to `logger.Error` if we are returning it ?

api/handlers/exporter.go
```diff
@@ -358,65 +385,88 @@ func validateValidatorRequest(request *validatorRequest) error {
...
+
+		// best effort attempt to fill the CommitteeID field, non blocking if it fails
+		// as the duty itself is still valid without it
+		committeeID, err := e.traceStore.GetCommitteeID(slot, idx)
+		if err == nil {
+			result.CommitteeID = &committeeID
+		} else if !isNotFoundError(err) {
+			// if error is not found, nothing to report to prevent unnecessary noise, otherwise log the error:
+			e.logger.Debug("error getting committee ID", zap.Error(err), fields.Slot(slot), fields.ValidatorIndex(idx))
+			errs = multierror.Append(errs, err)
```
iurii-ssv, line 454: I don't really understand why we need to log error here (also why as `logger.Debug` and not `logger.Error`) if we are gonna return it in this func below as `errs.ErrorOrNil()` for the caller to deal with (to log, etc.)

This is probably how it was originally implemented, but we are also logging the error & returning it at the same time in multiple various handlers ... this duplication can be confusing to debug - I would personally prefer to only return the error (wrapping it with the necessary details) for the caller to deal with at the outer-most layer (typically log it, or type-check and maybe log it)

wdyt @julienh-ssv ?

exporter/model.go
julienharbulot: Method is called DeepCopy, not copying was a bug

operator/dutytracer/collector.go
julienharbulot: I'll change it in the next PR when I change the validators store

api/handlers/exporter.go
julienharbulot: Logging it at Debug because it's a non-blocking error that people don't need to worry about (so not error) but that I still want to be able to monitor when I look at the logs

exporter/model.go
iurii-ssv: A bug or an optimization : )

api/handlers/exporter.go
iurii-ssv: Regarding Debug - makes sense,

regarding logging it here - like I mentioned, seems better to log at the outer-most layer (at the caller) once because otherwise we are potentially logging it 2 times ... but that's out of scope of this PR, just a convention I think we should follow

iurii-ssv (approved):
LGTM, just keep an eye on potentially increased memory usage [due to this bug-fix/optimization (whatever we call it, lol)](https://github.com/ssvlabs/ssv/pull/2498#discussion_r2332150864)

api/handlers/exporter.go
```diff
@@ -123,46 +122,41 @@ func (e *Exporter) Decideds(w http.ResponseWriter, r *http.Request) error {
...
-	participants, err := e.traceStore.GetValidatorDecideds(role, slot, pubkeys)
+	entries, err := e.traceStore.GetValidatorDecideds(role, slot, indices)
 	if err != nil {
 		errs = multierror.Append(errs, err)
 	}
-	return participants, errs
+	return entries, errs
 }
 
 func (e *Exporter) TraceDecideds(w http.ResponseWriter, r *http.Request) error {
```
oleg-ssvlabs, line 162: Question:
Have you considered switching from `pubkeys` to `indices` for the `/v1/exporter/decideds` endpoint handler (`Decideds` method in this file)? I’m wondering why you made the change here but not there?

api/handlers/exporter.go
oleg-ssvlabs: I agree with @iurii-ssv regarding these two issues: logging with `Error` level (or at least `Warning`), and logging by the caller unless the “inner” layer has some additional context it doesn’t need or want to share with the caller.  

`Debug` level logs won’t be noticeable, while `Warning` or `Error` will be. If `CommitteeID` cannot continuously be fetched, I assume we want to notice that and treat it as an indicator of a potential issue? 

not sure guys what you agreed upon. @julienh-ssv maybe you are planning to look into it in next PRs? I see the log was already there before your changes

api/handlers/model.go
```diff
@@ -30,10 +30,11 @@ type decidedResponse struct {
...
 type decidedRequest struct {
-	From    uint64        `json:"from"`
-	To      uint64        `json:"to"`
-	Roles   api.RoleSlice `json:"roles"`
-	PubKeys api.HexSlice  `json:"pubkeys"`
+	From    uint64          `json:"from"`
+	To      uint64          `json:"to"`
+	Roles   api.RoleSlice   `json:"roles"`
+	PubKeys api.HexSlice    `json:"pubkeys"`
+	Indices api.Uint64Slice `json:"indices"`
```
oleg-ssvlabs, line 37: It would be good if you could mention this in the PR description - that the API endpoint request body was changed and a new optional field was added. I think this is a fairly important change

api/handlers/exporter.go
iurii-ssv: Regarding log-level I'm fine with either (depending on the urgency of it - if we only ever want to see this error when debugging **when specifically looking for it**, as opposed to being able to just notice it, then `Debug` is fine I think)

As for double-logging of the error, this can be done in a follow-up PRs for sure, I just want us to be on the same page about it (this is really similar in nature to "pushing pubkey/index conversion to API boundaries", but for logging)

julienharbulot: Let's leave this for another PR focused on logs and error management. 

I agree that double logging is not necessary. Currently the "outer layer" does not log these errors (the error list is sent back in the API response for humans to understand what's up but we don't store the response in our logs at the moment) but I had a plan add it in the future when I focus on API traces so we can re-discuss when we get to it.

Regarding logging at DEBUG or ERROR, both make total sense depending on the goal. Think about errors developers are interested in and errors that clients are interested in. clients running an operator at the usual INFO level will not see DEBUG logs but will see (and freak out) at the ERROR level. Not all errors are equal.

api/handlers/model.go
julienharbulot: It's already mentioned in "what's changed" section:

> 3 -- /exporter/decideds endpoint:
>
> Accepts additional indices (aligns behavior with other endpoints).
> Fixes a bug where committeeID wasn’t populated in responses.

api/handlers/exporter.go
julienharbulot: So the endpoint you mention `/v1/exporter/decideds` is actually mapped to this method `TraceDediceds` in our current exporter v2 (fullExporter) as show in the code excerpt below:

```
router.Get("/v1/exporter/decideds", api.Handler(s.exporter.TraceDecideds))
router.Post("/v1/exporter/decideds", api.Handler(s.exporter.TraceDecideds))
```

The older implementations (such as `s.exporter.Decideds`) are end of life and will be removed. We don't update them anymore and they are only reachable when running in the old "exporter v1" mode.

oleg-ssvlabs (approved):

PR #2499, opened 2025-09-08
backport: CL: correct multi-client usage (Events)

iurii-ssv:
This is a backport of https://github.com/ssvlabs/ssv/pull/2494 onto **version/optimization** branch (so we can release it in **v2.3.5**),

it is a "minimum viable fix" to minimize future merge conflicts

### DIFF
--- a/beacon/goclient/events.go
+++ b/beacon/goclient/events.go
@@ -2,6 +2,7 @@ package goclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


beacon/goclient/events.go
```diff
@@ -90,11 +91,18 @@ func (gc *GoClient) startEventListener(ctx context.Context) error {
...
+		var errs error
+		success := false
 		for _, client := range gc.clients {
 			if err := client.Events(ctx, opts); err != nil {
 				logger.Error(clResponseErrMsg, zap.String("api", "Events"), zap.Error(err))
-				return err
+				errs = errors.Join(errs, err)
+				continue
 			}
+			success = true
```
oleg-ssvlabs, line 102: Does this mean the errors can be ignored as long as at least one client successfully subscribes to the events?

I think it would be worth writing a warning log if errs is not nil and there is at least one success.

iurii-ssv: We are:
- logging every error with `logger.Error(clResponseErrMsg, zap.String("api", "Events"), zap.Error(err))`
- also returning `errs` if all the calls errored

also note, that it means in case all of the calls errored - we'll essentially log the errors twice - I'm hoping to address duplicate error-logging in https://github.com/ssvlabs/ssv/pull/2472 a bit later (what I'm doing in this PR is 100% consistent with the current approach)

oleg-ssvlabs: ah I see, I missed the line where the error is actually logged.

PR #2501, opened 2025-09-08
storage: ListOperatorsAll method

iurii-ssv:
This PR is addressing a `TODO` comment I've spotted while doing something else.

BTW, I'm not 100% sure we need to print all operators (their pubkeys) on SSV node startup ... maybe it's useful but it's ~1k extra log-lines

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -1142,7 +1142,7 @@ func syncContractEvents(
 
 		// Print registry stats.
 		shares := nodeStorage.Shares().List(nil)
-		operators, err := nodeStorage.ListOperators(nil, 0, 0)
+		operators, err := nodeStorage.ListOperatorsAll(nil)
 		if err != nil {
 			logger.Error("failed to get operators", zap.Error(err))
 		}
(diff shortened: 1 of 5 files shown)


cli/operator/node.go
```diff
@@ -1142,7 +1142,7 @@ func syncContractEvents(
 
 		// Print registry stats.
 		shares := nodeStorage.Shares().List(nil)
-		operators, err := nodeStorage.ListOperators(nil, 0, 0)
+		operators, err := nodeStorage.ListOperatorsAll(nil)
```
nkryuchkov, line 1145: `ListOperatorsAll` looks a bit weird IMO. What about `ListAllOperators`?

iurii-ssv: I think `ListOperatorsAll` name is a bit more consistent with `ListOperators` method (which essentially can be called `ListOperatorsFromTo` if we were to expand its name to include the actual filtering params it expects)

TBH this seems like an opinionated thing that's hard to agree on (so I wouldn't try to "enforce" any conventions around it)

PR #2515, opened 2025-09-10
git: testing how git is treating squashed merge-commits (target branch main)

iurii-ssv:


iurii-ssv:
So, looks like just "having a merge-commit as your last commit in the PR" is not OK for git to figure out the conflicts on its own

PR #2521, opened 2025-09-11
fix(exporter): fix retro-compatibility issue in api

julienh-ssv:
Main quest:

This PR implements an API fix for the exporter to fix retro-compatibility issue with existing clients:

- The main retro-compat change is that `NotFoundErr` should never trigger an error status code in the API response on its own. Instead, we should return a response with empty data:

```go
errs = filterOutDutyNotFoundErrors(errs)

// if we don't have a single valid result and we have at least one meaningful error, return an error
if len(all) == 0 && errs.ErrorOrNil() != nil {
  e.logger.Error("error serving SSV API request", zap.Any("request", request), zap.Error(errs))
  return toApiError(errs)
}
```

- Another important fix is that we need to make sure that we don't misinterpret a filter request for a get-all request. This is implemented as follows:

```go
// if the request was for a specific set of participants and we couldn't resolve any, we're done
if request.hasFilters() && len(indices) == 0 { return... }
```

- I've also updated the tests to add regression tests and make sure we don't loose this semantic in the future.

Side quest:
- I've also removed a bit of duplication when parsing pubkeys and indices in the `/decided` and `/validator` endpoints

### DIFF
--- a/api/handlers/exporter/exporter.go
+++ b/api/handlers/exporter/exporter.go
@@ -56,7 +56,7 @@ func toApiError(errs *multierror.Error) *api.ErrorResponse {
 }
 
 func toStrings(err *multierror.Error) []string {
-	if err == nil || err.ErrorOrNil() == nil {
+	if err.ErrorOrNil() == nil {
 		return nil
 	}
 	errs := err.Errors


api/handlers/exporter/committee.go
```diff
@@ -29,12 +29,23 @@ func (e *Exporter) CommitteeTraces(w http.ResponseWriter, r *http.Request) error
...
 	}
+
+	// by design, not found duties are expected and not considered as API errors
+	errs = filterOutDutyNotFoundErrors(errs)
+
+	// if we don't have a single valid result and we have at least one meaningful error, return an error
+	if len(all) == 0 && errs.ErrorOrNil() != nil {
+		e.logger.Error("error serving SSV API request", zap.Any("request", request), zap.Error(errs))
+		return toApiError(errs)
+	}
```
julienharbulot, line 47: This is the main change to preserve previous API behavior: NotFoundErrors are not considered "errors" at all from the API point of view.

This makes sense because a lot of slot don't have duties and the store responds with "NotFoundError". But it does not mean that "something went wrong".

api/handlers/exporter/committee.go
```diff
@@ -64,12 +75,8 @@ func (e *Exporter) getCommitteeDutiesForSlot(slot phase0.Slot, committeeIDs []sp
...
 		duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID)
 		if err != nil {
-			// if error is not found, nothing to report as we might not have a duty for this role
-			// otherwise report it:
-			if !isNotFoundError(err) {
-				e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
-				errs = multierror.Append(errs, err)
-			}
+			e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
+			errs = multierror.Append(errs, err)
```
julienharbulot, line 79: Removing this special check for NotFound because it's not needed anymore (we handle it more uniformly by filtering NotFound in the API handler instead).

api/handlers/exporter/common_model.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/ssvlabs/ssv/exporter"
 )
 
+type filterRequest interface {
+	pubKeys() []spectypes.ValidatorPK
+	indices() []uint64
+	hasFilters() bool
+}
```
julienharbulot, line 18: Little abstraction to reduce duplication and allow us to process decided and validator requests in the same way

api/handlers/exporter/decided.go
```diff
@@ -27,20 +27,16 @@ func (e *Exporter) TraceDecideds(w http.ResponseWriter, r *http.Request) error {
...
-	indices := make([]phase0.ValidatorIndex, 0, len(request.Indices)+len(pubkeys))
-	for _, idx := range request.Indices {
-		indices = append(indices, phase0.ValidatorIndex(idx))
+	indices, err := e.extractIndices(&request)
+	errs = multierror.Append(errs, err)
+
+	// if the request was for a specific set of participants and we couldn't resolve any, we're done
+	if request.hasFilters() && len(indices) == 0 {
+		return toApiError(errs)
 	}
```
julienharbulot, line 39: Fail early if we can't resolve the requested filters. This also prevents mis-interpreting a filter request for a get all request later.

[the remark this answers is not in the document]
julienharbulot: We haven't looked for duties yet so there won't be duties not found errors

api/handlers/exporter/exporter.go
```diff
@@ -53,7 +55,11 @@ func toApiError(errs *multierror.Error) *api.ErrorResponse {
 	return api.Error(errs)
 }
 
-func toStrings(errs []error) []string {
+func toStrings(err *multierror.Error) []string {
+	if err == nil || err.ErrorOrNil() == nil {
```
iurii-ssv: This `if err == nil || err.ErrorOrNil() == nil` check could be simplified down to `if err.ErrorOrNil() == nil`,

pointing this out because I've noticed in `extractIndices` we do rely on `ErrorOrNil` checking for `nil` internally - so it would be nice to rely on it consistently everywhere (or not rely everywhere, if you prefer that)

api/handlers/exporter/exporter.go
```diff
@@ -66,3 +72,51 @@ func toStrings(errs []error) []string {
...
+		return nil
+	}
+	var filteredErrs *multierror.Error
+	for _, err := range e.Errors {
+		if !isNotFoundError(err) {
+			filteredErrs = multierror.Append(filteredErrs, err)
+		}
+	}
+	return filteredErrs
+}
```
iurii-ssv, line 87: Regarding `filterOutDutyNotFoundErrors` func, you've probably considered it - but wouldn't it be "more natural" to not include those not-found errors in places where we receive them ? This would also make it transparent which code-path return which not-found errors.

Anyway, that's just a minor suggestion - feel free to skip (or implement later on)

iurii-ssv (approved):
Awesome, just couple suggestions

api/handlers/exporter/exporter.go
julienharbulot: Actually this PR just to opposite way of what you suggest by removing the "in-place" special cases and filtering once at the end. It leads to simpler more consistent code. 

Also not sure but I think this function was removed in the subsequent PR for pubkeys fix.

api/handlers/exporter/exporter.go
julienharbulot: ok

y0sher (approved):

PR #2522, opened 2025-09-11
stage <- main(v2.3.5)

iurii-ssv:
This PR brings some of the changes we have in **main** branch into **stage** branch (some of these were only merged into **main** branch without any counterpart in **stage** branch, while others were merged into both - such that **main** branch would typically only receive a minimal hot-fix while **stage** branch would get a proper/complete/well-documented solution).

This PR is a prerequisite for https://github.com/ssvlabs/ssv/pull/2520 (to resolve merge conflicts https://github.com/ssvlabs/ssv/pull/2520 currently has).


operator/validator/controller_test.go
```diff
@@ -798,6 +917,17 @@ func buildOperatorData(id uint64, ownerAddress string) *registrystorage.Operator
...
+func buildFeeRecipient(owner string, feeRecipient string) *registrystorage.RecipientData {
+	feeRecipientSlice := []byte(feeRecipient) // Assuming FeeRecipient is a string or similar
+	var executionAddress bellatrix.ExecutionAddress
+	copy(executionAddress[:], feeRecipientSlice)
+	return &registrystorage.RecipientData{
+		Owner:        common.BytesToAddress([]byte(owner)),
+		FeeRecipient: executionAddress,
+		Nonce:        nil,
+	}
+}
```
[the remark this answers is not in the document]
iurii-ssv: That's actually dead code now that made it here from **main** branch as the result of `git merge main`, removed it now.

ssvsigner/keystore/file.go
```diff
@@ -53,7 +55,11 @@ func EncryptKeystore(privkey []byte, pubKeyBase64, password string) ([]byte, err
 		return nil, fmt.Errorf("encrypt private key: %w", err)
 	}
 
-	encryptedKeystoreJSON["pubkey"] = pubKeyBase64
+	if legacyPubkey {
```
oleg-ssvlabs, line 58: I am not sure I understand this if else, it seems we assign the same value to the same variable

iurii-ssv: The key is different `pubKey` vs `pubkey` - see `K` and `k`, this is intended change, I just hope I copied it over correctly (without forgetting anything) - cc @nkryuchkov

oleg-ssvlabs: I see now! Thank you, sorry it was hard to notice

beacon/goclient/goclient.go
```diff
@@ -171,8 +171,14 @@ func New(
...
 ) (*GoClient, error) {
-	logger.Info("consensus client: connecting",
-		fields.Address(opt.BeaconNodeAddr),
+	if opt.BeaconNodeAddr == "" {
+		return nil, fmt.Errorf("no beacon node address provided")
+	}
+
+	beaconAddrList := strings.Split(opt.BeaconNodeAddr, ";")
+
+	logger.Info("consensus client: connecting (multi client)",
```
nkryuchkov, line 180: Continuing the discussion from https://github.com/ssvlabs/ssv/pull/2469#discussion_r2308435517, we were going to remove the `(multi client)` part, but it seems to have been merged into `main`

eth/executionclient/options.go
```diff
@@ -44,17 +44,21 @@ func WithFollowDistanceMulti(offset uint64) OptionMulti {
 	}
 }
 
-// WithConnectionTimeout sets timeout for network connection to eth1 node.
-func WithConnectionTimeout(timeout time.Duration) Option {
+// WithReqTimeout sets timeout for RPC requests to eth1 node.
```
nkryuchkov, line 47: It's out of the scope of the PR, but since we made the behavior ignore zero timeout, it might be worth mentioning in the comment (same for the function below). Probably, we should do it in another PR, though

iurii-ssv: I'm not sure though how exactly we'd want to document such a behavior (or whether we'd even want to) though:
- it seems to me this is just a basic sanity-check that requires no clarifications because we really do not want a timeout of 0
- if 0 meant "never timeout" (in this case it doesn't) then I would for sure want to call out that "we do not allow infinite timeout"

could you expand on what kind of comment you have in mind ?

beacon/goclient/goclient.go
iurii-ssv: Hmm, good you spotted that, not sure how I missed the original discussion (even though I commented on it, lol),

TBH I think we want to put more effort into our logging in general (and I have couple PRs that should improve that a bit - almost ready for review), for example we probably want to differentiate between:
- differentiate between logs intended for the end user (INFO) - these should use CL / EL terminology (not `multi client` which is a vague implementation detail) **vs** logs we write for debugging (DEBUG) - these can/should use all the data the  developers can leverage
- log more things/details in general
- de-duplicate the stuff we log as much as possible (single source of truth) - log things at the edges (at the outer-most caller possible) so it's done **once** in one place ... that means prefer returning `error` over logging it pretty much

as for this `(multi client)` thing - I've pushed [a commit into another PR](https://github.com/ssvlabs/ssv/pull/2472/commits/aee4e2f328329d631bda8cd246ef5c11c0b744df) to resolve it to keep this one free of changes that update the behavior (this PR just needs to reconcile **stage** vs **main** branches)

cc @oleg-ssvlabs @y0sher ^

eth/executionclient/options.go
nkryuchkov: I'd just add something like `The timeout must be positive. Zero and negative timeouts are ignored.` to make sure no one tries to disable the timeout with `WithReqTimeout(0)`. But it's kind of a nit, maybe it could make the `WithReqTimeout` documentation a bit clearer, although it's not very important. Resolve this if you disagree

iurii-ssv: Added a commit to clarify this in https://github.com/ssvlabs/ssv/pull/2472

beacon/goclient/goclient.go
y0sher: ACK, I agree we should do some logs clean up. Though it's a long term effort we should probably pair it with some clear guidelines that should be enforced in PRs or even a lint checker.

y0sher (approved):

y0sher:
Good job on this!

PR #2525, opened 2025-09-12
2480: iurii-pass2 feature/p2p/pinned-peers

iurii-ssv:
Simplifying a bit.

### DIFF
--- a/api/handlers/p2p/model.go
+++ b/api/handlers/p2p/model.go
@@ -1,4 +1,4 @@
-package pinned_peers
+package p2p
 
 import (
 	"net/http"
--- a/api/handlers/p2p/pinned_peers.go
+++ b/api/handlers/p2p/pinned_peers.go
@@ -1,10 +1,9 @@
-package pinned_peers
+package p2p
 
 import (
 	"fmt"
 	"net/http"
 
-	p2pnetwork "github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 
 	"github.com/ssvlabs/ssv/api"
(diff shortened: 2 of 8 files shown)


api/handlers/p2p/model.go
```diff
@@ -1,4 +1,4 @@
-package pinned_peers
+package p2p
```
iurii-ssv, line 1: I feel like we can just use `p2p` as a package name

api/handlers/p2p/pinned_peers.go
```diff
@@ -1,10 +1,9 @@
-package pinned_peers
+package p2p
 
 import (
 	"fmt"
 	"net/http"
 
-	p2pnetwork "github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
```
iurii-ssv, line 7: Ideally **handlers** shouldn't know about implementation details such as `libp2p`

api/handlers/p2p/pinned_peers.go
```diff
@@ -96,22 +98,8 @@ func (h *Handler) Remove(w http.ResponseWriter, r *http.Request) error {
...
-// NewLibp2pConnChecker wraps a libp2p network with a connection checker.
-func NewLibp2pConnChecker(net p2pnetwork.Network) *libp2pConnChecker {
-	return &libp2pConnChecker{net: net}
-}
-
-func (c *libp2pConnChecker) IsConnected(id peer.ID) bool {
-	return c.net.Connectedness(id) == p2pnetwork.Connected
-}
+// peerConnectionCheckFn provides a minimal, testable interface for connection checks.
+type peerConnectionCheckFn func(peer.ID) bool
```
iurii-ssv, line 102: No need for an interface where `func` will do, perhaps this is a bit opinionated but it trims the code quite a bit

network/p2p/p2p.go
```diff
@@ -154,32 +155,31 @@ func New(
 		),
 	}
 
-	if err := n.parsePinnedPeers(); err != nil {
+	if err := n.initPinnedPeers(); err != nil {
 		return nil, err
 	}
-	if err := n.parseTrustedPeers(); err != nil {
+	if err := n.initTrustedPeers(); err != nil {
```
iurii-ssv, line 162: Lets make the side-effects of these funcs explicit

PR #2528, opened 2025-09-16
Fulu support

olegshmuelov:
Add Fulu fork support including aligning to https://github.com/ssvlabs/ssv-spec/compare/v1.1.3...v1.1.4


liorrutenberg (approved):

beacon/goclient/proposer.go
```diff
@@ -82,7 +83,7 @@ func (gc *GoClient) GetBeaconBlock(
 	slot phase0.Slot,
 	graffitiBytes []byte,
 	randao []byte,
-) (ssz.Marshaler, spec.DataVersion, error) {
+) (*api.VersionedProposal, ssz.Marshaler, error) {
```
oleg-ssvlabs, line 83: nit: considering that `ssz.Marshaler` is actually a versioned block representation, I think it would be worth either using named return values or including this information in the comment above the method

beacon/goclient/proposer.go
```diff
@@ -96,109 +97,52 @@ func (gc *GoClient) GetBeaconBlock(
...
+			return beaconBlock, beaconBlock.FuluBlinded, nil
 		}
-		if beaconBlock.Electra.Block.Body.ExecutionPayload == nil {
-			return nil, DataVersionNil, fmt.Errorf("electra block execution payload is nil")
-		}
-		return beaconBlock.Electra, beaconBlock.Version, nil
+		return beaconBlock, beaconBlock.Fulu, nil
 	default:
-		return nil, DataVersionNil, fmt.Errorf("beacon block version %s not supported", beaconBlock.Version)
+		return nil, nil, fmt.Errorf("beacon blinded block version %s not supported", beaconBlock.Version)
```
oleg-ssvlabs: nit: should `blinded` be in this error message? I mean the block can be both blinded and non-blinded

beacon/goclient/proposer.go
```diff
@@ -262,126 +206,178 @@ func (gc *GoClient) getProposalParallel(
...
+			SignedBlock: &electra.SignedBeaconBlock{
+				Message:   block.Fulu.Block,
+				Signature: sig,
+			},
+			KZGProofs: block.Fulu.KZGProofs,
+			Blobs:     block.Fulu.Blobs,
+		}
 	default:
-		return fmt.Errorf("unknown block version")
+		return fmt.Errorf("unknown block version %d", version)
```
oleg-ssvlabs, line 376: nit: it would be more readable to write version as a string:

```Go
fmt.Errorf("unknown block version %s", version.String())
```

protocol/v2/qbft/controller/controller.go
```diff
@@ -230,6 +230,18 @@ func (c *Controller) Decode(data []byte) error {
 	return nil
 }
 
+// UnmarshalJSON is a custom JSON unmarshaller for Controller
+func (c *Controller) UnmarshalJSON(data []byte) error {
```
oleg-ssvlabs: may I ask why this was needed? It does look like no-op unmarshaller (I mean it does not change the default behavior)

beacon/goclient/proposer_test.go
```diff
@@ -183,16 +181,11 @@ func TestGetBeaconBlock_FeeRecipientValidation(t *testing.T) {
 		randao := getTestRANDAO()
 
 		// Should get block successfully but fee recipient should be zero
-		block, version, err := client.GetBeaconBlock(t.Context(), slot, graffiti, randao)
+		versionedProposal, _, err := client.GetBeaconBlock(t.Context(), slot, graffiti, randao)
```
oleg-ssvlabs: maybe worth not ignoring second returned value (`ssz.Marshaler`). We could add `require.NotNil(t, obj)` or so. I don't think it is expected to be `nil`. Same in other modified tests

protocol/v2/ssv/runner/proposer.go
```diff
@@ -251,18 +258,15 @@ func (r *ProposerRunner) ProcessConsensus(ctx context.Context, logger *zap.Logge
 		observability.ValidatorPublicKeyAttribute(cd.Duty.PubKey),
 	)
 
-	if r.decidedBlindedBlock() {
+	versionedBlock, blkToSign, err := cd.GetBlockData()
+	if err != nil {
+		return errors.Wrap(err, "could not get block data")
```
oleg-ssvlabs: it is better to use `traces.Errorf()` here

protocol/v2/ssv/runner/proposer.go
```diff
@@ -372,70 +376,54 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
...
-		}
 
-		bSummary, summarizeErr = summarizeBlock(vBlindedBlk)
-		logger = logger.With(
-			fields.BlockHash(bSummary.Hash),
-			zap.NamedError("summarize_err", summarizeErr),
-		)
+	vBlk, _, err := validatorConsensusData.GetBlockData()
+	if err != nil {
+		return errors.Wrap(err, "could not get block")
```
oleg-ssvlabs: same here, better to use `traces.Errorf`

oleg-ssvlabs (commented):
good job. Left few comments/questions

beacon/goclient/aggregator.go
```diff
@@ -35,8 +35,14 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
...
+	// Pre-Electra: Index = committeeIndex | Electra+ (EIP-7549): Index = 0
 	dataVersion, _ := gc.beaconConfig.ForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(attData.Slot))
-	if dataVersion < spec.DataVersionElectra {
+	if dataVersion >= spec.DataVersionElectra {
+		// EIP-7549: For Electra and later, index must always be 0
+		attData.Index = 0
+	} else {
+		// Pre-Electra uses committee index
 		attData.Index = committeeIndex
 	}
```
iurii-ssv, line 43: Nit (coding convention): what do you think of implementing `if-else` as just `if` instead:
```go
// EIP-7549: For Electra and later, index must always be 0, pre-Electra uses committee index.
attData.Index = 0
if dataVersion < spec.DataVersionElectra {
    attData.Index = committeeIndex
}
```
I think it sort-of aligns well with Golang's “happy path” idiom (sometimes it's "computationally expensive" to do it like that ... but 95% of cases still can be written like that) cc @oleg-ssvlabs @y0sher @nkryuchkov @julienh-ssv just to get on the same page

beacon/goclient/aggregator.go
```diff
@@ -76,79 +82,80 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
...
-	switch aggDataResp.Data.Version {
-	case spec.DataVersionElectra:
-		if aggDataResp.Data.Electra == nil {
-			gc.log.Error(clNilResponseForkDataErrMsg,
-				zap.String("api", "AggregateAttestation"),
-			)
-			return nil, DataVersionNil, fmt.Errorf("aggregate attestation electra data is nil")
+	version := aggDataResp.Data.Version
+	switch version {
+	case spec.DataVersionPhase0:
```
iurii-ssv: Nit: perhaps it would let us "shave off more bulk" in the code-block below if we define `r := aggDataResp.Data` instead of just `version := aggDataResp.Data.Version` ?
- `version` becomes `r.Version`
- `aggDataResp.Data.Altair` -> `r.Altair`
- etc.

beacon/goclient/proposer.go
```diff
@@ -262,126 +206,178 @@ func (gc *GoClient) getProposalParallel(
...
+		}
+		// Fulu reuses Electra's block types as per consensus spec
+		signedBlock.Fulu = &apiv1fulu.SignedBlockContents{
+			SignedBlock: &electra.SignedBeaconBlock{
+				Message:   block.Fulu.Block,
+				Signature: sig,
+			},
+			KZGProofs: block.Fulu.KZGProofs,
+			Blobs:     block.Fulu.Blobs,
+		}
```
iurii-ssv: I'm not sure what this `Fulu reuses Electra's block types as per consensus spec` comment is trying to describe, maybe expand it ?

beacon/goclient/spec.go
```diff
@@ -244,17 +244,30 @@ func (gc *GoClient) getForkData(specResponse map[string]any) (map[spec.DataVersi
...
 	if err != nil {
 		return nil, err
 	}
+	// After Fulu fork happens on all networks,
+	// - Fulu check should become required
+	// - Gloas check might be added as non-required
+	fuluEpoch, err := getForkEpoch("FULU_FORK_EPOCH", false)
+	if err != nil {
+		return nil, err
+	}
```
iurii-ssv, line 259: Just to clarify I understand it correctly, we are erroring here if CL is not aware of `FULU_FORK_EPOCH`, right ? I assume most CLs should already return `FULU_FORK_EPOCH` in their responses in the latest versions, 

but if we are to ship this PR well before the Fulu fork happens Operators might update to the latest SSV node version **while still running with older** CLs ... perhaps we shouldn't rely on that and treat error here the same way we treat `fuluEpoch  == FarFutureEpoch` case ?

If we need to do that ^ we'd also need to adjust the error check here as well, I guess:
```go
		fuluForkVersion, err = getForkVersion("FULU_FORK_VERSION")
		if err != nil {
			return nil, err
		}
```

go.mod
```diff
@@ -285,4 +285,6 @@ replace github.com/google/flatbuffers => github.com/google/flatbuffers v1.11.0
 
 replace github.com/dgraph-io/ristretto => github.com/dgraph-io/ristretto v0.1.1-0.20211108053508-297c39e6640f
 
-replace github.com/attestantio/go-eth2-client => github.com/ssvlabs/go-eth2-client v0.6.31-0.20250807154556-0c7614aa26d4
+// TODO(fulu): Once Fulu mainnet epoch is scheduled, merge upstream official release into SSV fork
+// and update to new version tag. Current fork includes both Fulu support and SSV-specific changes.
+replace github.com/attestantio/go-eth2-client => github.com/ssvlabs/go-eth2-client v0.6.31-0.20250916050033-a3418b3bf251
```
iurii-ssv: Nit: too many different forks :) `Current fork` -> `Our current go-eth2-client fork`

beacon/goclient/proposer.go
olegshmuelov: Keeping `%d` is better than `version.String()` because:

When an unknown version comes through, `%d` shows the actual number (e.g., "unknown block version 7"), while `version.String()` would output "unknown block version unknown" which is redundant and less informative.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -372,70 +376,54 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
...
 		fields.QBFTHeight(r.BaseRunner.QBFTController.Height),
 		fields.QBFTRound(r.GetState().RunningInstance.State.Round),
-		zap.String("block_hash", bSummary.Hash.String()),
-		zap.Bool("blinded", bSummary.Blinded),
 		fields.Took(time.Since(start)),
-		zap.NamedError("summarize_err", summarizeErr),
 		fields.TotalConsensusTime(r.measurements.TotalConsensusTime()),
 		fields.TotalDutyTime(r.measurements.TotalDutyTime()),
-	)
+	}
```
iurii-ssv: Seems like we can just "inline" these fields with the `✅ successfully submitted block proposal` message (and remove the `// Build success log fields conditionally` comment) ? Unlike for `loggerFields` I don't see why these need to be "conditional"

protocol/v2/ssv/spectest/ssv_mapping_test.go
```diff
@@ -239,11 +239,14 @@ func newRunnerDutySpecTestFromMap(t *testing.T, m map[string]interface{}) *Start
...
 	outputMsgs := make([]*spectypes.PartialSignatureMessages, 0)
-	for _, msg := range m["OutputMessages"].([]interface{}) {
-		byts, _ := json.Marshal(msg)
-		typedMsg := &spectypes.PartialSignatureMessages{}
-		require.NoError(t, json.Unmarshal(byts, typedMsg))
-		outputMsgs = append(outputMsgs, typedMsg)
+	// Handle null/empty OutputMessages from spec (empty arrays are now null in JSON)
+	if m["OutputMessages"] != nil {
+		for _, msg := range m["OutputMessages"].([]interface{}) {
+			byts, _ := json.Marshal(msg)
```
iurii-ssv: Nit: could check for error here (and couple other places in this file) with `require.NoError(t, ...`

ssvsigner/go.mod
```diff
@@ -157,3 +157,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.3.0 // indirect
 )
+
+replace github.com/ssvlabs/ssv => ../
```
iurii-ssv, line 165: This could probably use a comment

iurii-ssv (commented):
Love the explanation-comments and ssvsigner-checklist 👍 Left couple questions/suggestions

beacon/goclient/proposer.go
iurii-ssv: Perhaps we could even return `[]bytes` instead of `ssz.Marshaler` ? I don't see why we can't just do that

beacon/goclient/proposer.go
iurii-ssv: Yeah it looks like a shortcoming of `version.String()` (swallowing these details)

protocol/v2/qbft/controller/controller.go
olegshmuelov: spec alignment no-ops. deleted

beacon/goclient/spec.go
olegshmuelov: The current implementation uses `getForkEpoch("FULU_FORK_EPOCH", false)` intentionally - this follows the pattern we had for Electra before it went live (as you can see in the git history, Electra also used false initially with a comment about changing it to true after the fork happens).

  The pattern is:
  - Active/deployed forks (Altair through Electra): Use `required: true`
  - Future forks (currently Fulu): Use `required: false` for backward compatibility

  This allows operators to upgrade SSV nodes before their CL clients support Fulu. When the CL doesn't know about `FULU_FORK_EPOCH`, it won't cause an error - effectively treating it as not scheduled yet.

  I've added this to the FULU_MAINNET_CHECKLIST.md as a TODO item to change it to true one step before the mainnet fork when Fulu support is widespread in CL clients.

beacon/goclient/proposer.go
olegshmuelov: We're aligned with the SSV spec interfaces - the spec's `ProposerCalls.GetBeaconBlock` interface defines the return type as `ssz.Marshaler`, which we're following exactly. While we could change it to return `[]byte` directly, that would mean diverging from the spec interface definition and wouldn't provide any real benefit. The current approach maintains spec compliance.

oleg-ssvlabs (dismissed):

beacon/goclient/spec.go
iurii-ssv: > When the CL doesn't know about FULU_FORK_EPOCH, it won't cause an error - effectively treating it as not scheduled yet.

Yeah I just wanted to clarify that's accounted for

iurii-ssv (dismissed):

nkryuchkov (dismissed):

oleg-ssvlabs (dismissed):

PR #2534, opened 2025-09-17
node-probe: introduce delay between retries

iurii-ssv:
This is a follow-up to https://github.com/ssvlabs/ssv/pull/2492 to introduce some delay between the actual retry attempts, 

this is pretty much necessary to ensure we don't exhaust all retry attempts too rapidly (eg. if the Node replies immediately - perhaps due to caching its health-value), which is what we observed happen during recent testing:

image

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -569,8 +569,8 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		nodeProber := nodeprobe.New(logger)
-		nodeProber.AddNode(clNodeName, consensusClient, 10*time.Second, 5)
-		nodeProber.AddNode(elNodeName, executionClient, 10*time.Second, 5)
+		nodeProber.AddNode(clNodeName, consensusClient, 10*time.Second, 5, 10*time.Second)
+		nodeProber.AddNode(elNodeName, executionClient, 10*time.Second, 5, 10*time.Second)
 
 		probeCtx, cancel := context.WithTimeout(cmd.Context(), 30*time.Second)
 		defer cancel()
@@ -593,7 +593,7 @@ var StartNodeCmd = &cobra.Command{
 			doppelgangerHandler,
 		)
 		if len(cfg.LocalEventsPath) == 0 {
-			nodeProber.AddNode(eventSyncerNodeName, eventSyncer, 10*time.Second, 5)
+			nodeProber.AddNode(eventSyncerNodeName, eventSyncer, 10*time.Second, 5, 10*time.Second)
 		}
 		go startNodeProber(cmd.Context(), logger, nodeProber)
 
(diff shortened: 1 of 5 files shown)


nodeprobe/nodeprobe.go
```diff
@@ -122,16 +123,19 @@ func (p *Prober) probeNode(ctx context.Context, n pNode) (err error) {
 			zap.Int("attempt", attempt),
 			zap.Error(err),
 		)
+
+		time.Sleep(n.retryDelay) // wait before the next retry attempt
```
nkryuchkov: should we cancel it on context cancelation?

iurii-ssv: Yeah that would be nice actually, applied in https://github.com/ssvlabs/ssv/pull/2534/commits/8ef8b013e47f7a16276cc728583486487a6f1fdb

nkryuchkov (approved):

cli/operator/node.go
```diff
@@ -569,8 +569,8 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		nodeProber := nodeprobe.New(logger)
-		nodeProber.AddNode(clNodeName, consensusClient, 10*time.Second, 5)
-		nodeProber.AddNode(elNodeName, executionClient, 10*time.Second, 5)
+		nodeProber.AddNode(clNodeName, consensusClient, 10*time.Second, 5, 10*time.Second)
+		nodeProber.AddNode(elNodeName, executionClient, 10*time.Second, 5, 10*time.Second)
```
y0sher: I think it's worth to set these values in constants just for readability and because they are repeated. so all 3 timeout, retry num and retry delay.

iurii-ssv: Yeah, would be nice to make sure "these values aren't modified by accident without anybody noticing",

generally speaking these are independent though (they just happen to match because we are not super picky about each of those values),

applied in https://github.com/ssvlabs/ssv/pull/2534/commits/1ef3494ed529e84a56123fd7cbf9f47fffd41e6d

oleg-ssvlabs (approved):
lgtm

nodeprobe/nodeprobe.go
```diff
@@ -122,16 +123,24 @@ func (p *Prober) probeNode(ctx context.Context, n pNode) (err error) {
 			zap.Int("attempt", attempt),
 			zap.Error(err),
 		)
+
+		// Wait before the next attempt.
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(n.retryDelay):
```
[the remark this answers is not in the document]
iurii-ssv: Right, not super important for this functionality, but I'll adjust it in some later PR (to not block the merging of this one)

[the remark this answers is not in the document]
iurii-ssv: Actually we are already checking for this with:
```go
		if attempt == attemptsMax {
			break // all retries failed
		}
```

PR #2535, opened 2025-09-17
feat: Add OpenApi generator

julienh-ssv:
No more "someone created it but we lost it". It's now a feature of the codebase :)

Changes:
- Added open api generator and linter (makefile)
- Added open api info to exporter API funcs and types

Notes:
- Model types must be exported for OpenApi generator to use them.

### DIFF
--- a/protocol/v2/ssv/runner/measurements.go
+++ b/protocol/v2/ssv/runner/measurements.go
@@ -85,7 +85,7 @@ func (cm *measurementsStore) EndPostConsensus() {
 	}
 }
 
-// StartDutyFullFlow stores duty full flow start time.
+// StartDutyFlow stores duty full flow start time.
 func (cm *measurementsStore) StartDutyFlow() {
 	if cm != nil {
 		cm.dutyStart = time.Now()
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -119,7 +119,7 @@ func (b *BaseRunner) SetHighestDecidedSlot(slot phase0.Slot) {
 	b.highestDecidedSlot = slot
 }
 
-// setupForNewDuty is sets the runner for a new duty
+// baseSetupForNewDuty is sets the runner for a new duty
 func (b *BaseRunner) baseSetupForNewDuty(duty spectypes.Duty, quorum uint64) {
 	// start new state
 	// start new state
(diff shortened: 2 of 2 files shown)


api/server/openapi_info.go
```diff
@@ -0,0 +1,23 @@
+package server
+
+// General OpenAPI/Swagger metadata for the SSV HTTP API.
+//
+// @title           SSV Node API
+// @version         0.1.0
+// @description     HTTP endpoints exposed by the SSV node, including exporter traces and validator metadata.
+// @contact.name    SSV Labs
+// @contact.url     https://ssv.network
+// @contact.email   support@ssv.network
```
y0sher: ssv labs is at ssvlabs.io , ssv.network is the address of ssv.network DAO or proejct. 

ssv node is developed for ssv.network by ssvlabs.

short: change to ssvlabs.

julienharbulot: ok, I checked on the ssvlabs.io website and for support they redirect to `info at ssv.network` so I also changed the email

y0sher (dismissed):
lgtm

api/handlers/exporter/validator_model.go
```diff
@@ -9,48 +9,70 @@ import (
...
-	Rounds      []round               `json:"consensus"`
-	Decideds    []decided             `json:"decideds"`
-	Pre         []message             `json:"pre"`
-	Post        []message             `json:"post"`
-	Proposal    string                `json:"proposalData,omitempty"`
+// ValidatorTrace captures the consensus trace information for a single
+// validator duty.
+type ValidatorTrace struct {
+	// Slot is the duty slot for this validator trace.
+	Slot phase0.Slot `json:"slot" swaggertype:"integer" format:"int64" example:"123456"`
```
oleg-ssvlabs: I think this is actually a `string`, not an `integer` (marshaller is overridden in the lib)

api/handlers/exporter/validator_model.go
```diff
@@ -9,48 +9,70 @@ import (
...
+	// Role is the beacon role for this duty (e.g., ATTESTER).
+	Role string `json:"role" example:"ATTESTER"`
+	// Validator is the validator index for the duty.
+	Validator phase0.ValidatorIndex `json:"validator" swaggertype:"integer" format:"int64" example:"123"`
```
oleg-ssvlabs: this is also a `string` type.

I see in the example you provide it as a string, so maybe you meant to say - it's a string type, but it should be convertible to `int64`?

oleg-ssvlabs (commented):
Good job. Left two questions regarding open API types

api/handlers/exporter/validator_model.go
julienharbulot: Good catch thank you!

api/handlers/exporter/validator_model.go
julienharbulot: Same as the other one, fixed. Thank you

PR #2540, opened 2025-09-19
node: should start up with 0 active validators

iurii-ssv:
This PR is addressing a bug (edge-case) introduced in v2.3.5,

if a Share is registered with an Operator that doesn't have any other active validators besides that one ... and then Operator restarts before this newly added Share is activated ... Operator wouldn't be able to start (it will keep crashing)

### DIFF
--- a/operator/node.go
+++ b/operator/node.go
@@ -137,20 +137,42 @@ func (n *Node) Start(ctx context.Context) error {
 
 	n.validatorsCtrl.StartNetworkHandlers()
 
+	// IMPORTANT: We must initialize validators regardless of whether we are running exporter or
+	// a regular SSV node.
+	validatorsInitialized, err := n.validatorsCtrl.InitValidators()
+	if err != nil {
+		return fmt.Errorf("init validators: %w", err)
+	}
+
 	if n.exporterOptions.Enabled {
-		// Subscribe to all subnets.
+		// For exporter, we connect to all subnets.
+
 		err := n.net.SubscribeAll()
 		if err != nil {
 			n.logger.Error("failed to subscribe to all subnets", zap.Error(err))
 		}
+	} else {
+		// For regular SSV node, we either connect to the subnets validators we manage need or any random
+		// subnet just to participate in the network.
+
+		if len(validatorsInitialized) == 0 {
+			// If no validators were initialized - we're not subscribed to any subnets, we
+			// have to subscribe to at least 1 random subnet to participate in the network.
+			if err := n.net.SubscribeRandoms(1); err != nil {
+				return fmt.Errorf("subscribe to 1 random subnet: %w", err)
+			}
+			n.logger.Info("no validators to start, successfully subscribed to random subnet")
+			return nil
+		}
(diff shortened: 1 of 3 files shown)


operator/validator/controller.go
```diff
@@ -492,54 +492,53 @@ func (c *controller) StartValidators(ctx context.Context) error {
...
 	}
 	return nil
 }
 
-// setupValidators initializes validators for the provided shares.
+// initValidators initializes validators for the provided shares.
 // Share w/o validator's metadata won't start, but the metadata will be fetched and the validator will start afterward.
-func (c *controller) setupValidators(shares []*ssvtypes.SSVShare) ([]*validator.Validator, []*validator.Committee) {
-	c.logger.Info("initializing validators ...", zap.Int("shares count", len(shares)))
+func (c *controller) initValidators(shares []*ssvtypes.SSVShare) ([]*validator.Validator, error) {
```
iurii-ssv: Removing `[]*validator.Committee` from the return-values here since it's some relic of the past that no longer makes sense, it is only used to print this log line and the value printed is incorrect (`"started":5740`):
```
{"level":"\u001b[34mINFO\u001b[0m","time":"2025-09-18T11:24:54.034619Z","name":"Controller","msg":"start validators done","map size":2870,"failures":0,"shares":2870,"started":5740}
```

nkryuchkov (approved):

kchojn (dismissed):

operator/node.go
```diff
@@ -140,21 +137,43 @@ func (n *Node) Start() error {
...
+		}
 	}
+
 	go n.net.UpdateSubnets()
 	go n.net.UpdateScoreParams()
 
-	err := n.validatorsCtrl.StartValidators(ctx)
-	if err != nil {
-		return fmt.Errorf("start validators: %w", err)
-	}
```
iurii-ssv, line 156: I think it's better to start these go-routines **after** we've initialized/started validators (since those calculate subnets we are interested in based on which validators are active/running) - otherwise we are very likely to skip the updates on 1st tick

hence I've moved validator init/startup a bit higher here ... although since it ticks every 1s it's probably not a big deal

operator/validator/controller.go
```diff
@@ -492,54 +492,52 @@ func (c *controller) StartValidators(ctx context.Context) error {
...
+		zap.Int("initialized", len(validatorsInitialized)),
 		zap.Int("missing_metadata", len(fetchMetadata)),
-		zap.Int("shares", len(shares)),
-		zap.Int("initialized", len(validators)),
+		zap.Int("failures", len(errs)),
+		zap.Int("committees", c.validatorsMap.SizeCommittees()),
 	)
-	return validators, committees
+
+	if len(errs) == len(shares) {
```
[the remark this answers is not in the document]
iurii-ssv: Good catch, resolved now in https://github.com/ssvlabs/ssv/pull/2540/commits/c7564a68c7f4ec024caa7ebd4192ac877d05ae2b

PR #2550, opened 2025-09-29
CL/EL: extend & simplify logs

iurii-ssv:
This PR:
- simplifies the way we keep track of CL/EL requests in our logs (instead of `CL single-client` and `CL multi-client` we now have only `CL client` to watch out for)
- logs the actual CL being used for requests (previously known as `CL multi-client` requests) that are potentially performing a fallback

it also makes the CL and EL request logging more similar to each other


beacon/goclient/aggregator.go
```diff
@@ -54,7 +54,7 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
 		AttestationDataRoot: root,
 		CommitteeIndex:      committeeIndex,
 	})
-	recordMultiClientRequest(ctx, gc.log, "AggregateAttestation", http.MethodGet, time.Since(aggDataReqStart), err)
+	recordRequest(ctx, gc.log, "AggregateAttestation", http.MethodGet, gc.multiClient.Name(), true, time.Since(aggDataReqStart), err)
```
nkryuchkov: The order seems to be wrong. Also, should we use `gc.multiClient.Address()` instead?

```go
func recordRequest(
	ctx context.Context,
	logger *zap.Logger,
	routeName, clientAddr, httpMethod string,
	maybeFallback bool,
	duration time.Duration,
	err error,
)
```

```suggestion
	recordRequest(ctx, gc.log, "AggregateAttestation", gc.multiClient.Address(), http.MethodGet, true, time.Since(aggDataReqStart), err)
```

beacon/goclient/observability.go
```diff
@@ -58,73 +58,32 @@ var (
 			metric.WithDescription("beacon client selections for attestation data")))
 )
 
-func recordSingleClientRequest(
+func recordRequest(
 	ctx context.Context,
 	logger *zap.Logger,
 	routeName, clientAddr, httpMethod string,
+	maybeFallback bool,
```
nkryuchkov: Can you please add a comment on what this means?

nkryuchkov (changes requested):

beacon/goclient/aggregator.go
nkryuchkov: Nit: Perhaps we could also match the order of input parameters and logs while we are touching this to reduce the probability of a mistake

eth/executionclient/observability.go
```diff
@@ -99,7 +99,7 @@ var (
 			metric.WithDescription("number of times a client was initialized")))
 )
 
-func recordSingleClientRequest(
+func recordRequest(
```
nkryuchkov, line 102: Can we convert it to a method? It receives `ec.logger` and `ec.nodeAddr` for each call. Then we'll be able to simplify all calls.

Before:
```go
recordRequest(ctx, ec.logger, "BlockNumber", ec.nodeAddr, time.Since(start), err)
```

After:
```go
ec.recordRequest(ctx, "BlockNumber", time.Since(start), err)
```

beacon/goclient/aggregator.go
iurii-ssv: > Also, should we use gc.multiClient.Address() instead?

Yeah we should, nice catch

> Nit: Perhaps we could also match the order of input parameters and logs while we are touching this to reduce the probability of a mistake

yes, good idea, applied

beacon/goclient/observability.go
iurii-ssv: Yeah it's a bit unfortunate we don't have more granular control over fallback/failover mechanism for CL multi-client implementation, added a comment like this - does it make sense ?
```go
// whether this request might have undergone a fallback (true means maybe, not a 100% yes)
```

beacon/goclient/aggregator.go
nkryuchkov: Nit: Maybe we could also change the address type from `string` to `interface { Address() }`, just because several strings are error-prone

eth/executionclient/observability.go
iurii-ssv: I think we probably want to leave the `logger` here as a param because:
- similar `recordRequest` func in `GoClient` sometimes passes a different logger
- I'm planning to update the way CL/EL clients work with their `logger`(s) to make it so that we can use caller's logger (eg. a `logger` with duty-related data on it, such as `duty_id`) when logging a request ... this will likely mean `recordRequest` will need to accept `logger` as a parameter anyway

beacon/goclient/aggregator.go
iurii-ssv: > Nit: Maybe we could also change the address type from string to interface { Address() }, just because several strings are error-prone

Yeah that should prevent a bunch of typos, applied in https://github.com/ssvlabs/ssv/pull/2550/commits/a62b3d2f47c781cb7cbf7755ee7c84209f58aa85

y0sher (approved):

PR #2551, opened 2025-09-30
2358: iurii refactor

iurii-ssv:
This PR implements [this suggestion](https://github.com/ssvlabs/ssv/pull/2358#discussion_r2379043751) + some minor adjustments borrowed from the draft of a diff Nikita shared with me in Discord

we could polish this further, but it could also be done later on (in separate PRs) since the general code structure should stay roughly the same

### DIFF
--- a/protocol/v2/qbft/spectest/controller_type.go
+++ b/protocol/v2/qbft/spectest/controller_type.go
@@ -166,10 +166,6 @@ func runInstanceWithData(
 
 	testBroadcastedDecided(t, contr.GetConfig().(*qbft.Config), contr.Identifier, runData, contr.CommitteeMember.Committee)
 
-	for _, inst := range contr.StoredInstances {
-		inst.ValueChecker = nil
-	}
-
 	// test root
 	r, err := contr.GetRoot()
 	require.NoError(t, err)
--- a/protocol/v2/ssv/spectest/committee_msg_processing_type.go
+++ b/protocol/v2/ssv/spectest/committee_msg_processing_type.go
@@ -71,12 +71,6 @@ func (test *CommitteeSpecTest) RunAsPartOfMultiTest(t *testing.T) {
 	// test beacon broadcasted msgs
 	spectestingutils.CompareBroadcastedBeaconMsgs(t, test.BeaconBroadcastedRoots, broadcastedRoots)
 
-	for _, r := range test.Committee.Runners {
-		if r.BaseRunner.State.RunningInstance != nil {
-			r.BaseRunner.State.RunningInstance.ValueChecker = nil
-		}
-	}
-
 	// post root
 	postRoot, err := test.Committee.GetRoot()
 	require.NoError(t, err)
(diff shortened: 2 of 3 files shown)


operator/validator/controller.go
```diff
@@ -1055,18 +1054,15 @@ func SetupCommitteeRunners(
...
+			attestingValidators,
+			buildController(spectypes.RoleCommittee),
 			options.Beacon,
 			options.Network,
 			options.Signer,
 			options.OperatorSigner,
-			valCheck,
 			dutyGuard,
 			options.DoppelgangerHandler,
 		)
```
iurii-ssv, line 1068: We can initialize `CommitteeRunner` with the `attestingValidators` we got when setting up the duty, or we could define a separate func (`attestingValidators` and calculate this dynamically in `executeDuty`),

the end result is the same, but the 2nd option is slightly less performant (since it's basically gonna iterate over `duty.ValidatorDuties` 1 more time)

protocol/v2/qbft/config.go
```diff
@@ -21,8 +18,6 @@ type signing interface {
 
 type IConfig interface {
 	signing
-	// GetValueChecker returns value checker
-	GetValueChecker() ssv.ValueChecker
```
iurii-ssv, line 25: We don't need `GetValueChecker()` here

protocol/v2/ssv/runner/runner.go
```diff
@@ -35,7 +35,6 @@ type Getters interface {
 	GetLastRound() specqbft.Round
 	GetStateRoot() ([32]byte, error)
 	GetBeaconNode() beacon.BeaconNode
-	GetValChecker() ssv.ValueChecker
 	GetSigner() ekm.BeaconSigner
 	GetOperatorSigner() ssvtypes.OperatorSigner
 	GetNetwork() specqbft.Network
```
iurii-ssv, line 40: We don't need `GetValChecker()` here

operator/validator/controller.go
```diff
@@ -1123,15 +1118,15 @@ func SetupRunners(
 		switch role {
 		case spectypes.RoleProposer:
 			proposedValueCheck := ssv.NewProposerChecker(options.Signer, options.NetworkConfig.Beacon, share.ValidatorPubKey, share.ValidatorIndex, phase0.BLSPubKey(share.SharePubKey))
-			qbftCtrl := buildController(spectypes.RoleProposer, proposedValueCheck)
+			qbftCtrl := buildController(spectypes.RoleProposer)
 			runners[role], err = runner.NewProposerRunner(logger, options.NetworkConfig, shareMap, qbftCtrl, options.Beacon, options.Network, options.Signer, options.OperatorSigner, options.DoppelgangerHandler, proposedValueCheck, 0, options.Graffiti, options.ProposerDelay)
```
iurii-ssv, line 1122: Note, for runners other than `CommitteeRunner` we can initialize the `ValueChecker` once per runner (like we are doing it here already) - there is no need to instantiate it inside of `executeDuty` dynamically ... but we could for the sake of keeping the code across all the runners similar

PR #2552, opened 2025-10-01
feat(exporter): add schedule to store

julienh-ssv:
This PR:

- Fetches the list of scheduled duties from the beacon node, using an optimized beacon adapter that is only used for the exporters (safe switch since exporter does not execute any real duties) --> optimised adapter allows to query duties for all validators in a single call .
- Stores these duties efficiently on disk using roaring bitmap
- Exposes these duties in the API responses under `schedule` key.

### DIFF
--- a/api/handlers/exporter/exporter.go
+++ b/api/handlers/exporter/exporter.go
@@ -47,8 +47,6 @@ type dutyTraceStore interface {
 	GetAllValidatorDecideds(role spectypes.BeaconRole, slot phase0.Slot) ([]dutytracer.ParticipantsRangeIndexEntry, error)
 	GetCommitteeDecideds(slot phase0.Slot, index phase0.ValidatorIndex, roles ...spectypes.BeaconRole) ([]dutytracer.ParticipantsRangeIndexEntry, error)
 	GetAllCommitteeDecideds(slot phase0.Slot, roles ...spectypes.BeaconRole) ([]dutytracer.ParticipantsRangeIndexEntry, error)
-
-	// Scheduled duties (read-only for exporter)
 	GetScheduled(slot phase0.Slot) (map[phase0.ValidatorIndex]rolemask.Mask, error)
 }
 
--- a/exporter/store/store.go
+++ b/exporter/store/store.go
@@ -266,8 +266,6 @@ func (s *DutyTraceStore) GetCommitteeDuty(slot phase0.Slot, committeeID spectype
 	return
 }
 
-// Scheduled duties: per (slot, role) roaring64 bitmap of validator indices
-
 // SetScheduledRole overwrites scheduled indices for a (slot, role).
 func (s *DutyTraceStore) SetScheduledRole(slot phase0.Slot, role spectypes.BeaconRole, indices []phase0.ValidatorIndex) error {
 	prefix := s.makeScheduledRolePrefix(slot, role)
--- a/observability/log/fields/fields.go
+++ b/observability/log/fields/fields.go
@@ -296,14 +296,16 @@ func FeeRecipient(pubKey []byte) zap.Field {
 	return zap.Stringer(FieldFeeRecipient, stringer.HexStringer{Val: pubKey})
 }
 
-func Duties(epoch phase0.Epoch, duties []*spectypes.ValidatorDuty) zap.Field {
+// Duties formats a list of duties as a single log field, truncating the output if needed.
+// Use truncateAfter<=0 to disable truncation.
+func Duties(epoch phase0.Epoch, duties []*spectypes.ValidatorDuty, truncateAfter int) zap.Field {
 	var b strings.Builder
 	for i, duty := range duties {
 		if i > 0 {
 			b.WriteString(", ")
 		}
 		b.WriteString(BuildDutyID(epoch, duty.Slot, duty.RunnerRole(), duty.ValidatorIndex))
-		if i >= 10 {
+		if truncateAfter > 0 && i+1 >= truncateAfter {
 			b.WriteString(", (truncated) ...")
 			break
 		}
--- a/operator/duties/beacon_adapter.go
+++ b/operator/duties/beacon_adapter.go
@@ -17,14 +17,13 @@ import (
 	beaconprotocol "github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 )
 
-// prefetchingBeacon adapts the scheduler BeaconNode interface, prefetching
-// epoch/period data and serving handler calls from cache. It instruments
-// network request duration and processing time.
-
 type validatorPubkeyProvider interface {
 	ValidatorPubkey(index phase0.ValidatorIndex) (spectypes.ValidatorPK, bool)
 }
 
+// prefetchingBeacon adapts the scheduler BeaconNode interface, prefetching
+// epoch/period data and serving handler calls from cache. It instruments
+// network request duration and processing time.
 type prefetchingBeacon struct {
 	log    *zap.Logger
 	inner  beaconprotocol.BeaconNode
--- a/operator/dutytracer/collector.go
+++ b/operator/dutytracer/collector.go
@@ -1032,8 +1032,6 @@ func (c *Collector) computeAndPersistScheduleForSlot(slot phase0.Slot) error {
 	return c.store.SaveScheduled(slot, schedule)
 }
 
-// (backfill helpers removed; schedule is enqueued only for current and previous slot via worker)
-
 // runScheduleWorker performs schedule computations and DB writes off the hot path.
 func (c *Collector) runScheduleWorker(ctx context.Context) {
 	for {
--- a/operator/duties/attester.go
+++ b/operator/duties/attester.go
@@ -277,10 +277,14 @@ func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 	span.AddEvent("storing duties", trace.WithAttributes(observability.DutyCountAttribute(len(storeDuties))))
 	h.duties.Set(epoch, storeDuties)
 
+	truncate := -1
+	if h.exporterMode {
+		truncate = 10
+	}
 	h.logger.Debug("🗂 got duties",
 		fields.Count(len(duties)),
 		fields.Epoch(epoch),
-		fields.Duties(epoch, specDuties),
+		fields.Duties(epoch, specDuties, truncate),
 		fields.Duration(start))
 
 	// Further processing is not needed in exporter mode, terminate early
--- a/operator/duties/proposer.go
+++ b/operator/duties/proposer.go
@@ -235,10 +235,14 @@ func (h *ProposerHandler) fetchAndProcessDuties(ctx context.Context, epoch phase
 	span.AddEvent("storing duties", trace.WithAttributes(observability.DutyCountAttribute(len(storeDuties))))
 	h.duties.Set(epoch, storeDuties)
 
+	truncate := -1
+	if h.exporterMode {
+		truncate = 10
+	}
 	h.logger.Debug("📚 got duties",
 		fields.Count(len(duties)),
 		fields.Epoch(epoch),
-		fields.Duties(epoch, specDuties),
+		fields.Duties(epoch, specDuties, truncate),
 		fields.Duration(start))
 
 	span.SetStatus(codes.Ok, "")
--- a/operator/duties/scheduler.go
+++ b/operator/duties/scheduler.go
@@ -467,7 +467,7 @@ func (s *Scheduler) ExecuteCommitteeDuties(ctx context.Context, duties committee
 
 		const eventMsg = "🔧 executing committee duty"
 		dutyEpoch := s.beaconConfig.EstimatedEpochAtSlot(duty.Slot)
-		logger.Debug(eventMsg, fields.Duties(dutyEpoch, duty.ValidatorDuties))
+		logger.Debug(eventMsg, fields.Duties(dutyEpoch, duty.ValidatorDuties, -1))
 		span.AddEvent(eventMsg, trace.WithAttributes(
 			observability.CommitteeIDAttribute(committee.id),
 			observability.DutyCountAttribute(len(duty.ValidatorDuties)),


observability/log/fields/fields.go
```diff
@@ -303,6 +303,10 @@ func Duties(epoch phase0.Epoch, duties []*spectypes.ValidatorDuty) zap.Field {
 			b.WriteString(", ")
 		}
 		b.WriteString(BuildDutyID(epoch, duty.Slot, duty.RunnerRole(), duty.ValidatorIndex))
+		if i > 10 {
```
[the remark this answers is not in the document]
julienharbulot: sure ^^

operator/dutytracer/store.go
```diff
@@ -38,6 +38,10 @@ type DutyTraceStore interface {
 	GetCommitteeDutyLinks(slot phase0.Slot) ([]*exporter.CommitteeDutyLink, error)
 	GetValidatorDuty(slot phase0.Slot, role spectypes.BeaconRole, index phase0.ValidatorIndex) (*exporter.ValidatorDutyTrace, error)
 	GetValidatorDuties(role spectypes.BeaconRole, slot phase0.Slot) ([]*exporter.ValidatorDutyTrace, error)
+
+	// Compact scheduled duties I/O
+	SaveScheduled(slot phase0.Slot, schedule map[phase0.ValidatorIndex]uint8) error
+	GetScheduled(slot phase0.Slot) (map[phase0.ValidatorIndex]uint8, error)
```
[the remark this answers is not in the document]
julienharbulot: False positive...

exporter/rolemask/rolemask.go
```diff
@@ -0,0 +1,52 @@
...
+
+// All returns canonical list of roles represented in the mask.
+func All() []spectypes.BeaconRole {
+	return []spectypes.BeaconRole{
+		spectypes.BNRoleAttester,
+		spectypes.BNRoleAggregator,
+		spectypes.BNRoleProposer,
+		spectypes.BNRoleSyncCommittee,
+		spectypes.BNRoleSyncCommitteeContribution,
+	}
```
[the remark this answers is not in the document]
julienharbulot: 1. I don't think it worth the possible added risk since the map has dynamic keys updates

operator/duties/beacon_adapter.go
```diff
@@ -0,0 +1,272 @@
...
+		start := time.Now()
+		comm, err := gc2.CommitteesForEpoch(ctx, epoch)
+		p.log.Debug("fetch CommitteesForEpoch(GoClient)", zap.Uint64("epoch", uint64(epoch)), zap.Duration("took", time.Since(start)), zap.Error(err))
+		if err != nil {
+			return nil, err
+		}
+		_ = p.committees.Set(epoch, comm, ttlcache.DefaultTTL)
+		return comm, nil
+	}
+	return nil, context.DeadlineExceeded
```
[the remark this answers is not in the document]
julienharbulot: ok

exporter/store/store.go
```diff
@@ -260,6 +264,162 @@ func (s *DutyTraceStore) GetCommitteeDuty(slot phase0.Slot, committeeID spectype
...
+		}
+		if err := s.db.Set(prefix, nil, buf.Bytes()); err != nil {
+			return fmt.Errorf("save scheduled (slot=%d role=%d): %w", slot, role, err)
+		}
+	}
+	return nil
+}
+
+// GetScheduled returns compact schedule map for a slot.
+func (s *DutyTraceStore) GetScheduled(slot phase0.Slot) (map[phase0.ValidatorIndex]uint8, error) {
```
oleg-ssvlabs: I think it would be better if this method returned `map[phase0.ValidatorIndex]rolemask.Mask` instead of `map[phase0.ValidatorIndex]uint8`.

exporter/rolemask/rolemask.go
```diff
@@ -0,0 +1,52 @@
+package rolemask
+
+import (
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+)
+
+// Mask is a compact bitfield of scheduled BN roles.
```
oleg-ssvlabs: It would be good to add description explaining what is the purpose of this type/alias (I mean `Mask`)

exporter/store/store.go
```diff
@@ -260,6 +264,162 @@ func (s *DutyTraceStore) GetCommitteeDuty(slot phase0.Slot, committeeID spectype
...
+	out := make(map[phase0.ValidatorIndex]uint8)
+	var errs *multierror.Error
+	for _, role := range rolemask.All() {
+		bit, _ := rolemask.BitOf(role)
```
oleg-ssvlabs: **Nit:** We iterate over `rolemask.All()` and then retrieve the bit for each role, skipping the boolean return value from `rolemask.BitOf`. This makes sense, as it's practically impossible for a `Mask` to not exist for a given role.

Maybe it is worth considering `rolemask.All()` to return a `map[spectypes.BeaconRole]Mask` instead, or alternatively, adding another exported function that returns such a type.  
That way, you could do something like this, which is useful since you often need both the `role` and its corresponding `mask`:

```Go
for role, mask := range rolemask.All(){
}
```

exporter/store/store.go
```diff
@@ -292,6 +452,14 @@ func (s *DutyTraceStore) makeValidatorCommitteePrefix(slot phase0.Slot) []byte {
 	return append(prefix, slotToByteSlice(slot)...)
 }
 
+func (s *DutyTraceStore) makeScheduledRolePrefix(slot phase0.Slot, role spectypes.BeaconRole) []byte {
+	prefix := make([]byte, 0, len(scheduledDutyKey)+4+1)
+	prefix = append(prefix, []byte(scheduledDutyKey)...)
+	prefix = append(prefix, slotToByteSlice(slot)...)
+	prefix = append(prefix, byte(role&0xff))
```
oleg-ssvlabs: I believe this line works as long as the role value does not exceed `255`.  
However, we already have the `BeaconRole.Unknown`, which is set to `math.MaxUint64`.  
While this isn’t currently an issue, it could become one if we later decide to extend the list of supported roles in [`rolemask.go`](https://github.com/ssvlabs/ssv/blob/feat/exporter/schedule-v2/exporter/rolemask/rolemask.go#L30).

Let me know if I’m misunderstanding something here

exporter/store/store.go
```diff
@@ -292,6 +452,14 @@ func (s *DutyTraceStore) makeValidatorCommitteePrefix(slot phase0.Slot) []byte {
 	return append(prefix, slotToByteSlice(slot)...)
 }
 
+func (s *DutyTraceStore) makeScheduledRolePrefix(slot phase0.Slot, role spectypes.BeaconRole) []byte {
+	prefix := make([]byte, 0, len(scheduledDutyKey)+4+1)
```
oleg-ssvlabs: **Optional:**  
This is unrelated to the PR, and it seems you were just following the existing pattern with `len(scheduledDutyKey) + 4 + 1`.  
However, could we consider moving the slot size (4 bytes) into a named constant or returning it from a helper function?  
Right now, it's a bit too implicit and might be unclear to others reading the code.

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,62 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+		filter = make(map[string]struct{}, len(req.CommitteeIDs))
+		for _, c := range req.CommitteeIDs {
+			filter[hex.EncodeToString(c)] = struct{}{}
+		}
+	}
+
+	for s := req.From; s <= req.To; s++ {
+		slot := phase0.Slot(s)
+		sched, err := e.traceStore.GetScheduled(slot)
+		if err != nil || len(sched) == 0 {
```
oleg-ssvlabs: If the error is not `nil`, I would log it.  
Further down the call chain, errors also don’t appear to be logged, so if we don’t log them at least at the top level caller, they’ll be completely invisible.

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,62 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+			continue
+		}
+		links, err := e.traceStore.GetCommitteeDutyLinks(slot)
+		if err != nil || len(links) == 0 {
```
oleg-ssvlabs: Same here

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,62 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+			continue
+		}
+		// committeeID(hex) -> role -> indices
+		grouped := make(map[string]map[string][]uint64)
```
oleg-ssvlabs: I think it would be easier to read if specific types were used instead, like: `spectypes.CommitteeID`, `spectypes.BeaconRole`, `phase0.ValidatorIndex`

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,62 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+			if filter != nil {
+				if _, ok := filter[cid]; !ok {
+					continue
+				}
+			}
+			if grouped[cid] == nil {
+				grouped[cid] = make(map[string][]uint64)
+			}
+			// Populate roles for bits present
+			for _, role := range []spectypes.BeaconRole{
```
oleg-ssvlabs: here you can instead iterate over `rolemask.All()` function

exporter/rolemask/rolemask.go
```diff
@@ -0,0 +1,52 @@
...
+var roleToBit = map[spectypes.BeaconRole]Mask{
+	spectypes.BNRoleAttester:                  BitAttester,
+	spectypes.BNRoleAggregator:                BitAggregator,
+	spectypes.BNRoleProposer:                  BitProposer,
+	spectypes.BNRoleSyncCommittee:             BitSyncCommittee,
+	spectypes.BNRoleSyncCommitteeContribution: BitSyncContribution,
+}
+
+// All returns canonical list of roles represented in the mask.
+func All() []spectypes.BeaconRole {
```
oleg-ssvlabs: small optimization suggestion:  the beacon role collection could be defined as a static package-level variable, and `All()` could return it.  
This would avoid unnecessary allocations on each call.  

The downside is that if any caller mutates the returned collection, it could lead to unintended side effects.

So up for you to decide if you want to apply this change

api/handlers/exporter/validator.go
```diff
@@ -198,6 +204,61 @@ func toValidatorTraceResponse(duties []validatorDutyTraceWithCommitteeID, errs *
...
+		roleWanted[spectypes.BeaconRole(r)] = struct{}{}
+	}
+
+	// If no filters provided, we’ll include all indices present in the schedule per slot.
+	filter := req.hasFilters()
+
+	for s := req.From; s <= req.To; s++ {
+		slot := phase0.Slot(s)
+		sched, err := e.traceStore.GetScheduled(slot)
+		if err != nil {
```
oleg-ssvlabs, line 224: maybe log the error here

api/handlers/exporter/validator.go
```diff
@@ -198,6 +204,61 @@ func toValidatorTraceResponse(duties []validatorDutyTraceWithCommitteeID, errs *
...
+			if !ok {
+				continue
+			}
+			roles := make([]string, 0, len(req.Roles))
+			for role := range roleWanted {
+				if rolemask.Has(mask, role) {
+					roles = append(roles, role.String())
+				}
+			}
+			if len(roles) == 0 {
```
oleg-ssvlabs, line 251: I think it might be worth including the validator in the response, even if no roles are assigned?

```
schedule: {
  slot: 100,
  validator: 101,
  roles: [],
}
```

beacon/goclient/committees.go
```diff
@@ -0,0 +1,60 @@
...
+	cfg := gc.getBeaconConfig()
+	if cfg == nil {
+		return nil, fmt.Errorf("no beacon config set")
+	}
+	epoch := cfg.EstimatedEpochAtSlot(slot)
+	all, err := gc.CommitteesForEpoch(ctx, epoch)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]*eth2apiv1.BeaconCommittee, 0, 8)
```
oleg-ssvlabs: I am really not sure why capacity is set to `8` specifically?

To my knowledge `MAX_COMMITTEES_PER_SLOT` for both testnets and mainnet is equal to `64`. Unfortunately we do not have that value in `beaconConfig`, I assume it was not really needed, otherwise we could use it for setting the right capacity.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/ssvlabs/ssv
 go 1.24.0
 
 require (
+	github.com/RoaringBitmap/roaring v1.9.4
```
oleg-ssvlabs: I think it [isn't the latest version](https://github.com/RoaringBitmap/roaring/releases/tag/v2.10.0). Was there a reason not to use V2?

oleg-ssvlabs (commented):
Good job Julien!

I have not reviewed majority, but not all files. It's a large PR. Submitting first batch of comments and questions. Once they are resolved, I'll continue reviewing

exporter/store/store.go
julienharbulot: ok

exporter/store/store.go
julienharbulot: ok

exporter/store/store.go
julienharbulot: ok

api/handlers/exporter/committee.go
julienharbulot: ok

api/handlers/exporter/committee.go
julienharbulot: ok

api/handlers/exporter/committee.go
julienharbulot: ok

api/handlers/exporter/committee.go
julienharbulot: ok

exporter/rolemask/rolemask.go
julienharbulot: ok

api/handlers/exporter/validator.go
julienharbulot: ok

api/handlers/exporter/validator.go
julienharbulot: ok

beacon/goclient/committees.go
julienharbulot: ok

go.mod
julienharbulot: ok

exporter/rolemask/rolemask.go
julienharbulot: ok

exporter/store/store.go
julienharbulot: Switched scheduled key suffix to the mask bit to decouple from raw role values (avoids >255 concerns).

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,75 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+			copy(id[:], c)
+			filter[id] = struct{}{}
+		}
+	}
+
+	for s := req.From; s <= req.To; s++ {
+		slot := phase0.Slot(s)
+		sched, err := e.traceStore.GetScheduled(slot)
+		if err != nil {
+			e.logger.Debug("get scheduled failed", zap.Error(err), fields.Slot(slot))
```
oleg-ssvlabs: I would probably change the log level to `warning` or `error`

api/handlers/exporter/committee.go
```diff
@@ -107,3 +112,75 @@ func toCommitteeTraceResponse(duties []*exporter.CommitteeDutyTrace, errs *multi
...
+			continue
+		}
+		if len(sched) == 0 {
+			continue
+		}
+		links, err := e.traceStore.GetCommitteeDutyLinks(slot)
+		if err != nil {
+			e.logger.Debug("get committee links failed", zap.Error(err), fields.Slot(slot))
```
oleg-ssvlabs: here log level as well

api/handlers/exporter/validator.go
```diff
@@ -198,6 +204,67 @@ func toValidatorTraceResponse(duties []validatorDutyTraceWithCommitteeID, errs *
 	return r
 }
 
+// buildValidatorSchedule reads the compact on-disk schedule and returns a filtered
+// per-validator schedule for the requested roles and slot range.
+func (e *Exporter) buildValidatorSchedule(req *ValidatorTracesRequest, indices []phase0.ValidatorIndex) []ValidatorSchedule {
+	out := make([]ValidatorSchedule, 0)
+	// Build a quick lookup for requested roles.
+	roleWanted := map[spectypes.BeaconRole]struct{}{}
```
oleg-ssvlabs, line 213: I don’t think this map is necessary.  
You never access it via the key, so you do not really utilize `O(1)` lookup.

[Here](https://github.com/ssvlabs/ssv/blob/feat/exporter/schedule-v2/api/handlers/exporter/validator.go#L245), couldn’t you just iterate over `req.Roles`?

oleg-ssvlabs (commented):
Good job Julien.
Left just few nit comments

api/handlers/exporter/validator.go
julienharbulot: The map is the standard way to dedup. I updated the comment to make it clearer.

oleg-ssvlabs: ah, I see. I think I misunderstood the purpose from the comment, thank you for explaining

oleg-ssvlabs (dismissed):

api/handlers/exporter/exporter.go
```diff
@@ -41,10 +42,14 @@ type dutyTraceStore interface {
 	GetCommitteeDuty(slot phase0.Slot, committeeID spectypes.CommitteeID, role ...spectypes.BeaconRole) (*exporter.CommitteeDutyTrace, error)
 	GetCommitteeDuties(slot phase0.Slot, roles ...spectypes.BeaconRole) ([]*exporter.CommitteeDutyTrace, error)
 	GetCommitteeID(slot phase0.Slot, index phase0.ValidatorIndex) (spectypes.CommitteeID, error)
+	GetCommitteeDutyLinks(slot phase0.Slot) ([]*exporter.CommitteeDutyLink, error)
 	GetValidatorDecideds(role spectypes.BeaconRole, slot phase0.Slot, indices []phase0.ValidatorIndex) ([]dutytracer.ParticipantsRangeIndexEntry, error)
 	GetAllValidatorDecideds(role spectypes.BeaconRole, slot phase0.Slot) ([]dutytracer.ParticipantsRangeIndexEntry, error)
 	GetCommitteeDecideds(slot phase0.Slot, index phase0.ValidatorIndex, roles ...spectypes.BeaconRole) ([]dutytracer.ParticipantsRangeIndexEntry, error)
 	GetAllCommitteeDecideds(slot phase0.Slot, roles ...spectypes.BeaconRole) ([]dutytracer.ParticipantsRangeIndexEntry, error)
+
+	// Scheduled duties (read-only for exporter)
```
kchojn: can we have a proper comment here with  `GetScheduled` prefix ?

operator/dutytracer/collector.go
```diff
@@ -928,3 +944,106 @@ func (c *Collector) signersToKey(signers []spectypes.OperatorID) string {
...
+		}
+	}
+
+	if len(schedule) == 0 {
+		return nil
+	}
+	return c.store.SaveScheduled(slot, schedule)
+}
+
+// (backfill helpers removed; schedule is enqueued only for current and previous slot via worker)
```
kchojn: we can delete comments like this

kchojn (dismissed):

api/handlers/exporter/exporter.go
julienharbulot: Okay will open a new PR for comment fixes

operator/dutytracer/collector.go
julienharbulot: Okay I will open a new PR for comment fixes

api/handlers/exporter/exporter.go
y0sher: why? just fix the comments and we re-approve.

operator/dutytracer/collector.go
y0sher: why? just fix the comments and we re-approve.

observability/log/fields/fields.go
```diff
@@ -303,6 +303,10 @@ func Duties(epoch phase0.Epoch, duties []*spectypes.ValidatorDuty) zap.Field {
 			b.WriteString(", ")
 		}
 		b.WriteString(BuildDutyID(epoch, duty.Slot, duty.RunnerRole(), duty.ValidatorIndex))
+		if i >= 10 {
+			b.WriteString(", (truncated) ...")
+			break
+		}
```
y0sher: this list might be too long when on exporter, but we are currently relying on finding the duty ids in the logs to track back their flow, so its a tricky one, I might truncate this only on exporter or in a much longer list? even in exporter logs this is how we prove exporter had observed this duty.

julienharbulot: what if we add the exporter mode flag to it?

```
func Duties(epoch phase0.Epoch, duties []*spectypes.ValidatorDuty, exporterMode bool) zap.Field {
   if i >= 10 && exporterMode {
       b.WriteString(", (truncated) ...")
}
```

julienharbulot: Initially I didn't truncate it but the logs are unreadable in exporter mode because we have way too many duties to list them all

y0sher: Yeah I guess this works. might be easier to create a specific ExporterDuties function so we don't have to change `Duties` usage all around with "false".

operator/dutytracer/collector.go
julienharbulot: fixed

api/handlers/exporter/exporter.go
julienharbulot: fixed

observability/log/fields/fields.go
julienharbulot: fixed

PR #2553, opened 2025-10-01
operator-id: clarify & simplify

iurii-ssv:
I've noticed we could simplify some things related to **operator-id** (and perhaps fix a bug while at it)

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -793,28 +793,26 @@ func (c *Controller) committeeMemberFromShare(share *ssvtypes.SSVShare) (*specty
 	var activeOperators uint64
 
 	for _, cm := range share.Committee {
-		opdata, found, err := c.operatorsStorage.GetOperatorData(nil, cm.Signer)
+		operator := &spectypes.Operator{
+			OperatorID: cm.Signer,
+		}
+
+		opdata, found, err := c.operatorsStorage.GetOperatorData(nil, operator.OperatorID)
 		if err != nil {
 			return nil, fmt.Errorf("could not get operator data: %w", err)
 		}
-
 		if !found {
 			c.logger.Warn(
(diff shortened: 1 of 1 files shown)


cli/operator/node.go
```diff
@@ -692,8 +691,8 @@ func ensureNoMissingKeys(
 	operatorDataStore operatordatastore.OperatorDataStore,
 	ssvSignerClient *ssvsigner.Client,
 ) {
-	if operatorDataStore.GetOperatorID() == 0 {
-		logger.Fatal("operator ID is not ready")
+	if !operatorDataStore.OperatorIDReady() {
+		return
 	}
```
iurii-ssv, line 697: It seems incorrect to me to terminate with `Fatal` here because "not having an operator ID at this stage" is still possible if an Operator has started a node but hasn't registered with SSV smart-contract yet ?

y0sher (approved):
lgtm, due to different ordering of startup sequence in the past, operator ID would've had to change for some components after they started, this is where we introduced AwaitOperatorID which isn't used anymore after we fixed the ordering to make more sense.

operator/validator/controller.go
```diff
@@ -798,26 +798,28 @@ func (c *Controller) committeeMemberFromShare(share *ssvtypes.SSVShare) (*specty
...
 			c.logger.Warn(
 				"operator data not found, validator will only start if the number of available operators is greater than or equal to the committee quorum",
 				fields.OperatorID(cm.Signer),
 				fields.CommitteeID(share.CommitteeID()),
 			)
-		} else {
-			activeOperators++
+			operators = append(operators, &spectypes.Operator{
+				OperatorID: cm.Signer,
+			})
```
nkryuchkov: If we keep 
```go
		operator := &spectypes.Operator{
			OperatorID: cm.Signer,
		}
```
before this block, these lines could be simplified to `operators = append(operators, operator)`

PR #2559, opened 2025-10-22
Exporter: clear beacon adapter cache periodically

julienh-ssv:
Simple fix to prevent memory leaks: clears the cache periodically. 

**This is currently running on hoodi on both stage & prod.**

### DIFF
--- a/operator/duties/beacon_adapter.go
+++ b/operator/duties/beacon_adapter.go
@@ -105,12 +105,12 @@ func (p *prefetchingBeacon) evictOldAttesterCaches(anchor phase0.Epoch) {
 		keepMin = anchor - 1
 	}
 	p.muAtt.Lock()
+	defer p.muAtt.Unlock()
 	for e := range p.attester {
 		if e < keepMin {
 			delete(p.attester, e)
 		}
 	}
-	p.muAtt.Unlock()
 }
 
 // evictOldProposerCaches evicts proposer epochs older than (anchor-1).
@@ -120,24 +120,24 @@ func (p *prefetchingBeacon) evictOldProposerCaches(anchor phase0.Epoch) {
 		keepMin = anchor - 1
 	}
 	p.muProp.Lock()
+	defer p.muProp.Unlock()
 	for e := range p.proposer {
 		if e < keepMin {
 			delete(p.proposer, e)
 		}
 	}
-	p.muProp.Unlock()
 }
 
 // evictOldSyncCommitteePeriods evicts sync-committee periods older than anchorPeriod.
 func (p *prefetchingBeacon) evictOldSyncCommitteePeriods(anchorPeriod uint64) {
 	keepMin := anchorPeriod
 	p.muSync.Lock()
+	defer p.muSync.Unlock()
 	for period := range p.sync {
 		if period < keepMin {
 			delete(p.sync, period)
 		}
 	}
-	p.muSync.Unlock()
 }
 
 func (p *prefetchingBeacon) ensureAttesterEpoch(ctx context.Context, epoch phase0.Epoch) error {


operator/duties/beacon_adapter.go
```diff
@@ -98,6 +98,50 @@ func (p *prefetchingBeacon) committeesForEpoch(ctx context.Context, epoch phase0
...
+
+	// Proposer
+    p.muProp.Lock()
+    for e := range p.proposer {
+        if e < keepMin || e > keepMax {
+            delete(p.proposer, e)
+        }
+    }
+    p.muProp.Unlock()
+}
```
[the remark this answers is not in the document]
julienharbulot: False positive.
evictOldEpochCaches doesn’t hold both locks at once; it locks muAtt, unlocks, then locks
        muProp. There’s no overlap, so no ordering deadlock is possible. Other code paths don’t acquire
        both locks simultaneously either. We also maintain a consistent conceptual order (attester
        before proposer) if that ever changes.

operator/duties/beacon_adapter.go
```diff
@@ -209,8 +257,10 @@ func (p *prefetchingBeacon) ensureSyncPeriod(ctx context.Context, epoch phase0.E
 	}
 	p.muSync.Lock()
 	p.sync[period] = collected
-	p.muSync.Unlock()
-	return nil
+    p.muSync.Unlock()
+    // Trim caches opportunistically after populating.
+    p.evictOldSyncPeriods(period)
+    return nil
```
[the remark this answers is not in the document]
julienharbulot: False positive.

operator/duties/beacon_adapter.go
```diff
@@ -176,8 +222,10 @@ func (p *prefetchingBeacon) ensureProposerEpoch(ctx context.Context, epoch phase
 	}
 	p.muProp.Lock()
 	p.proposer[epoch] = collected
-	p.muProp.Unlock()
-	return nil
+    p.muProp.Unlock()
+    // Trim caches opportunistically after populating.
+    p.evictOldEpochCaches(epoch)
+    return nil
```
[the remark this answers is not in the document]
julienharbulot: False positive.
Not a correctness issue for our usage. The adapter is a best‑effort cache. We evict based on the epoch we just populated; that eviction never removes the same epoch and keeps the adjacent one. The scheduler only requests current/next epochs, not epochs older than previous. If another goroutine populates an older-than-previous epoch in the small window, eviction may remove it by design; calls asking for such an epoch are out of scope for exporter operation.

operator/duties/beacon_adapter.go
```diff
@@ -143,9 +187,11 @@ func (p *prefetchingBeacon) ensureAttesterEpoch(ctx context.Context, epoch phase
...
 	p.muAtt.Lock()
 	p.attester[epoch] = dutyMap
-	p.muAtt.Unlock()
-	p.log.Debug("build attester cache", zap.Uint64("epoch", uint64(epoch)), zap.Int("indices", len(dutyMap)), zap.Duration("took", time.Since(start)))
-	return nil
+    p.muAtt.Unlock()
+    p.log.Debug("build attester cache", zap.Uint64("epoch", uint64(epoch)), zap.Int("indices", len(dutyMap)), zap.Duration("took", time.Since(start)))
+    // Trim caches opportunistically after populating.
+    p.evictOldEpochCaches(epoch)
+    return nil
```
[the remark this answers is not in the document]
julienharbulot: False positive. Eviction holds its own lock.

operator/duties/beacon_adapter.go
```diff
@@ -98,6 +98,50 @@ func (p *prefetchingBeacon) committeesForEpoch(ctx context.Context, epoch phase0
 	return nil, ErrCommitteesUnsupported
 }
 
+// evictOldEpochCaches trims attester/proposer caches to a small sliding window
+// around the provided anchor epoch (the epoch being populated) to prevent
+// unbounded growth during long runs. Keeps [anchor-1, anchor, anchor+1].
+func (p *prefetchingBeacon) evictOldEpochCaches(anchor phase0.Epoch) {
```
iurii-ssv: I think it would be cleaner to split this func into `evictOldAttesterCaches` and `evictOldProposerCaches` because
- it seems these can be done separately (and it is confusing to clean the "ProposerCaches" inside of `ensureAttesterEpoch` - which I assume has nothing to do with proposer(s))
- it will also match the `evictOldSyncPeriods` which mentions that it is "sync-specific"

operator/duties/beacon_adapter.go
```diff
@@ -98,6 +98,50 @@ func (p *prefetchingBeacon) committeesForEpoch(ctx context.Context, epoch phase0
...
+		if e < keepMin || e > keepMax {
+			delete(p.proposer, e)
+		}
+	}
+	p.muProp.Unlock()
+}
+
+// evictOldSyncPeriods trims sync-committee cache around the provided anchor
+// period. Keeps [anchor, anchor+1].
+func (p *prefetchingBeacon) evictOldSyncPeriods(anchorPeriod uint64) {
```
iurii-ssv: Nit: I would expand the name to be `evictOldSyncCommitteePeriods` because `Sync` can mean a lot of things by itself (there is no need to keep this name short really)

operator/duties/beacon_adapter.go
```diff
@@ -98,6 +98,50 @@ func (p *prefetchingBeacon) committeesForEpoch(ctx context.Context, epoch phase0
...
 
+// evictOldEpochCaches trims attester/proposer caches to a small sliding window
+// around the provided anchor epoch (the epoch being populated) to prevent
+// unbounded growth during long runs. Keeps [anchor-1, anchor, anchor+1].
+func (p *prefetchingBeacon) evictOldEpochCaches(anchor phase0.Epoch) {
+	var keepMin phase0.Epoch
+	if anchor > 0 {
+		keepMin = anchor - 1
+	}
+	keepMax := anchor + 1
```
iurii-ssv: Could you clarify why we need to evict epochs higher than `anchor + 1` (anchor + 2, anchor + 3, etc) ? I don't have lots of context on this code but I think epochs we are working with are always going up. 

This also maybe contradicts func name slightly - `evictOldEpochCaches` - where I would think `Old = anchor - 1, anchor - 2, etc.`

iurii-ssv (commented):
Looks good, just couple questions/suggestions

operator/duties/beacon_adapter.go
julienharbulot: okay I will split it

operator/duties/beacon_adapter.go
julienharbulot: yes we don't really need to evict the future. Will change it.

operator/duties/beacon_adapter.go
julienharbulot: ok

iurii-ssv (dismissed):
👍

y0sher (approved):

PR #2562, opened 2025-10-23
feat: blind all blocks

y0sher:
### Description

Soon in Fulu fork is going to go through and then the BPO forks will increase block sizes, this might make first round of consensus for proposers slower and prone to failure, we decided to blind all blocks at the leader and send only blinded blocks as the signed data.

Blinding is done by copying the required data from the full block and the execution payload into a blinded block structure, replacing txs and withdrawals with hashes.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -17,6 +17,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          lfs: true
 
       - name: Setup make
         run: sudo apt-get update && sudo apt-get install make
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ full-test:
 .PHONY: unit-test
 unit-test:
 	@echo "Running unit tests"
-	@go test -tags blst_enabled -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 `go list ./... | grep -ve "spectest\|ssv/scripts/"`
+	@go test -tags "blst_enabled lfs" -timeout 20m -race -covermode=atomic -coverprofile=coverage.out -p 1 `go list ./... | grep -ve "spectest\|ssv/scripts/"`
 	@$(MAKE) ssvsigner-test
 
 .PHONY: ssvsigner-test
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -1599,6 +1599,52 @@ func TestExporterBuildValidatorSchedule_Filtered(t *testing.T) {
 	assert.ElementsMatch(t, []string{"ATTESTER", "AGGREGATOR"}, entry.Roles)
 }
 
+func TestExporter_GetValidatorCommitteeDutiesForRoleAndSlot_MembershipGating(t *testing.T) {
+	store := newMockTraceStore()
+	exp := &Exporter{traceStore: store, logger: zap.NewNop()}
+
+	slot := phase0.Slot(42)
+	idx := phase0.ValidatorIndex(10)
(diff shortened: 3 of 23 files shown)


protocol/v2/blockchain/beacon/blind/blind.go
```diff
@@ -0,0 +1,285 @@
...
+				BLSToExecutionChanges:  p.Electra.Block.Body.BLSToExecutionChanges,
+				BlobKZGCommitments:     p.Electra.Block.Body.BlobKZGCommitments,
+				ExecutionRequests:      p.Electra.Block.Body.ExecutionRequests,
+			},
+		}
+
+		return &api.VersionedProposal{Version: p.Version, Blinded: true, ElectraBlinded: bb}, bb, nil
+
+	case spec.DataVersionFulu:
+		// Fulu reuses Electra block structures in this codebase.
```
nkryuchkov, line 231: nit: perhaps having this comment above `bb := &apiv1electra.BlindedBeaconBlock{` would be clearer as we don't reference electra on the line below

protocol/v2/ssv/runner/proposer.go
```diff
@@ -389,11 +406,37 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
 		fields.QBFTRound(r.GetState().RunningInstance.State.Round),
 	)
 
+	// Use the locally cached block for submission, but first verify the root
```
nkryuchkov: Where do we verify the root?

protocol/v2/ssv/runner/proposer.go
```diff
@@ -389,11 +406,37 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
...
+	// TODO: should we send the block at all if we're not the leader? It's probably not effective but
+	//		I left it for now to keep backwards compatibility.
+	if r.originalFullProposal != nil {
+		inst := r.GetState().RunningInstance
+		if inst != nil && inst.State != nil {
+			proposerF := inst.GetConfig().GetProposerF()
+			leaderID := proposerF(inst.State, inst.State.Round)
+			if leaderID == r.operatorSigner.GetOperatorID() {
+				logger.Debug("retrieved original full proposal for leader submission")
+				vBlk = r.originalFullProposal
```
nkryuchkov: Do we need to check the root match here, as mentioned in the comment above?

protocol/v2/ssv/runner/proposer.go
```diff
@@ -389,11 +406,37 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
 		fields.QBFTRound(r.GetState().RunningInstance.State.Round),
 	)
 
+	// Use the locally cached block for submission, but first verify the root
+	// matches the decided one.
+	validatorConsensusData := &spectypes.ValidatorConsensusData{}
+	err = validatorConsensusData.Decode(r.GetState().DecidedValue)
+	if err != nil {
+		return traces.Errorf(span, "could not decode consensus data: %w", err)
+	}
```
iurii-ssv: > but first verify the root matches the decided one.

I don't see this step, and we actually probably need it to ensure "we didn't change the round (changing our view on who the Leader is) right before the decided message arrived" (if that scenario is even possible)

I've added this check in https://github.com/ssvlabs/ssv/pull/2567 (see `cachedBlindedSSZ`, maybe we can optimize it further by using a **hash** of `cachedBlindedSSZ` value ... but that's probably a marginal improvement in terms of performance)

iurii-ssv (approved):

y0sher:
merged on different PR

PR #2565, opened 2025-10-24
refactor: replace Split in loops with more efficient SplitSeq

reddaisyy:
strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

More info: https://github.com/golang/go/issues/61901


iurii-ssv (approved):
Interesting!

y0sher (approved):
Approving even though the replaced slices are pretty slow hence there's no too much concern about resources when handling them.

PR #2567, opened 2025-10-25
2562: iurii suggestions

iurii-ssv:
Suggested adjustments for https://github.com/ssvlabs/ssv/pull/2562

### DIFF
--- a/protocol/v2/ssv/spectest/msg_processing_type.go
+++ b/protocol/v2/ssv/spectest/msg_processing_type.go
@@ -180,7 +180,7 @@ func (test *MsgProcessingSpecTest) RunAsPartOfMultiTest(t *testing.T, logger *za
 	spectestingutils.ComparePartialSignatureOutputMessages(t, test.OutputMessages, network.BroadcastedMsgs, committee)
 
 	// test beacon broadcasted msgs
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/runner/proposer.go
```diff
@@ -406,34 +411,33 @@ func (r *ProposerRunner) ProcessPostConsensus(ctx context.Context, logger *zap.L
...
-		inst := r.GetState().RunningInstance
-		if inst != nil && inst.State != nil {
-			proposerF := inst.GetConfig().GetProposerF()
-			leaderID := proposerF(inst.State, inst.State.Round)
-			if leaderID == r.operatorSigner.GetOperatorID() {
-				logger.Debug("retrieved original full proposal for leader submission")
-				vBlk = r.originalFullProposal
-			}
+	leaderID := r.GetState().RunningInstance.Proposer()
+	if r.cachedFullBlock != nil && leaderID == r.operatorSigner.GetOperatorID() {
```
iurii-ssv, line 431: Regarding `if inst != nil && inst.State != nil` checks,

`ValidatePostConsensusMsg` already validates this (and it makes more sense to validate it early, rather than here - since we wouldn't want to validate it every time we need to use the `Instance`)

PR #2571, opened 2025-10-27
update spec to latest version and align

iurii-ssv:


nkryuchkov (approved):

protocol/v2/ssv/value_check.go
```diff
@@ -47,10 +47,6 @@ func (v *voteChecker) CheckValue(value []byte) error {
 		return spectypes.WrapError(spectypes.DecodeBeaconVoteErrorCode, fmt.Errorf("failed decoding beacon vote: %w", err))
 	}
 
-	if bv.Target.Epoch > v.estimatedCurrentEpoch+1 {
```
y0sher, line 50: now we enforce that target is exactly equal to what our beacon node sees.
TODO: consider failing earlier like here. but removed now to satisfy spec tests and deps.

PR #2573, opened 2025-11-05
chore: fix mismatched test function comment in TestWeb3SignerRestart

hustrust:
Corrected the comment for TestWeb3SignerRestart function in proposer_test.go to match the actual function name. The comment previously referred to the function as "TestBlockRestart" which was inconsistent with the real function name "TestWeb3SignerRestart". 

This change ensures the code documentation accurately reflects the implementation.


nkryuchkov (approved):
Thank you for the contribution

PR #2574, opened 2025-11-05
fix(exporter): correct quorum detection for committee duties

olegshmuelov:
Fixes quorum detection issues introduced in #2558.

  **Problems Fixed:**

  1. **False positives:** Signatures were checked against both roles, causing attestation signatures to incorrectly trigger sync committee quorum
  2. **Missing detection:** When signatures arrived before proposal, quorum was never checked after flushing pending signatures

  **Solution:**

  - Use role-specific root matching in `checkAndPublishQuorum()` to prevent cross-role false positives
  - Add `checkQuorumAfterFlush()` to detect quorum immediately after `flushPending()` reclassifies buffered signatures
  - Refactor for better code quality and debuggability
  - Comprehensive test coverage for all edge cases

### DIFF
--- a/protocol/v2/qbft/controller/error.go
+++ b/protocol/v2/qbft/controller/error.go
@@ -7,10 +7,10 @@ import (
 
 var (
 	// ErrFutureConsensusMsg means the message we've got is "from the future"; it can happen if we haven't started
-	// QBFT instance for the slot the message is targeting yet, while another operator already did + sent this
+	// QBFT instance for the slot the message is targeting yet, while another operator already has + sent this
 	// message to us.
 	ErrFutureConsensusMsg = fmt.Errorf("future consensus msg")
-	// ErrInstanceNotFound means we might not have started the QBFT instance yet, while another operator already did
+	// ErrInstanceNotFound means we might not have started the QBFT instance yet, while another operator already has
 	// + sent this message to us.
 	ErrInstanceNotFound = fmt.Errorf("instance not found")
 )
--- a/protocol/v2/qbft/instance/error.go
+++ b/protocol/v2/qbft/instance/error.go
@@ -7,7 +7,7 @@ import (
 
 var (
 	// ErrWrongMsgRound means we might not have changed the round to the next one yet, while another operator
-	// already did + sent this message to us.
+	// already has + sent this message to us.
 	ErrWrongMsgRound = fmt.Errorf("wrong msg round")
 	// ErrNoProposalForCurrentRound means we might not have received a proposal-message yet, while another operator already
 	// did and started preparing it.
--- a/protocol/v2/ssv/runner/aggregator.go
+++ b/protocol/v2/ssv/runner/aggregator.go
@@ -114,6 +114,11 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 	defer span.End()
 
 	hasQuorum, roots, err := r.BaseRunner.basePreConsensusMsgProcessing(ctx, logger, r, signedMsg)
+	if errors.Is(err, ErrNoDutyAssigned) || errors.Is(err, ErrRunningDutyFinished) {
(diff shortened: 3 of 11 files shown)


operator/dutytracer/collector.go
```diff
@@ -616,8 +618,27 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
+								pendingCount += len(idxs)
+							}
+						}
+					}
+					c.logger.Error("CRITICAL: failed to compute role roots from proposal - pending signatures will be dropped",
+						zap.Error(err),
+						fields.Slot(slot),
+						fields.CommitteeID(committeeID),
+						zap.Int("pending_signature_count", pendingCount),
+						zap.Int("pending_roots_count", pendingRoots))
```
iurii-ssv: If this is indeed critical, maybe we should log the roots and signatures (and whatever other data we might need) for further debugging ?

We also have the `Warn loudly if we are about to drop pending buffered entries ...` below that seems like it needs a "similar CRITICAL treatment" ?

operator/dutytracer/eviction.go
```diff
@@ -68,9 +78,11 @@ func (c *Collector) dumpCommitteeToDBPeriodically(slot phase0.Slot) (totalSaved
 			// We intentionally drop pending entries here; they are not persisted.
 			trace.pendingByRoot = nil
 		}
+
+		// Deep copy while still holding the lock
+		data := trace.DeepCopy()
 		trace.Unlock()
```
iurii-ssv, line 80: The `DeepCopy` method could use a comment about the "need to hold the lock",

and there is also a `trace()` method that maybe could use a re-name into `DeepCopyLocked` or something

operator/dutytracer/eviction.go
```diff
@@ -80,23 +92,25 @@ func (c *Collector) dumpCommitteeToDBPeriodically(slot phase0.Slot) (totalSaved
...
+	// Delete from map FIRST while collecting data to prevent race condition
+	// where new messages arrive between reading and deletion
 	c.validatorTraces.Range(func(idx phase0.ValidatorIndex, slotToTraceMap *hashmap.Map[phase0.Slot, *validatorDutyTrace]) bool {
 		trace, found := slotToTraceMap.Get(slot)
 		if !found {
 			return true
 		}
 
+		// Remove from map immediately to prevent new messages from modifying this trace
+		slotToTraceMap.Delete(slot)
```
iurii-ssv: Nit: I'm not sure if this specific comment (and similar ones) is entirely accurate:
- the `trace` object has a dedicated lock that prevents all concurrent modifications ... so it alone should be able to take care of that, no ?
- the `slotToTraceMap.Delete` here does reduce the likelihood of concurrent actor "acquiring the `slotToTraceMap` object" but doesn't solve it entirely (I think only mutex can solve it entirely)

I do like the additional comments on this PR, just want to make sure they don't "create a false sense of security" about this stuff being atomic (it seems to be "best effort kind of thingy")

operator/dutytracer/collector.go
olegshmuelov: Added root logging: Implemented using the existing `pendingDetails()` helper.
Kept eviction without "CRITICAL:" prefix since it's cleanup, not an active failure, but both provide the same detailed debugging data.

operator/dutytracer/eviction.go
olegshmuelov: Exactly right on both points!

  1. Yes, the trace's lock is what provides concurrency safety - not the map deletion. The map deletion is purely about lifecycle management (preventing new references).

  2. Correct - deletion doesn't guarantee exclusivity. A goroutine could have already acquired a reference via `Get(slot)` before we delete.
  Updated comments in all 3 dump functions to reflect this
  
8b6781d2396b0d32e2d7bc0431dba5464c0b71b0

PR #2576, opened 2025-11-05
traces: droppable spans

iurii-ssv:
This PR adds a tracing middleware that allows us to filter out certain spans (marked with `DroppableSpan` attribute) before sending the data to the server.

This is needed because tracing is typically done for the entire func (meaning for **every** func-call a span will be generated, recorded & sent to the server), while for some funcs we want to record & send only **certain** func-call outcomes (not each one of them); For example, we don't want to record & send every retry-call because this will generate a lot of useless data.

Implementation-wise, seems like there is no `span.Discard()` (or similar) we could use on a `span` object directly ... so we have to use an attribute to "mark a particular span".

### DIFF
--- a/observability/traces/provider.go
+++ b/observability/traces/provider.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/contrib/exporters/autoexport"
+	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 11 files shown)


oleg-ssvlabs (approved):
I think it’s an interesting implementation and that it will scale well. Good job!

oleg-ssvlabs (approved):

PR #2577, opened 2025-11-05
p2p: calculate best-peer-subnets correctly

iurii-ssv:
There was a typo in the following discovery-related piece of code:
```
subnets, _ := n.PeersIndex().GetPeerSubnets(bestPeer.ID)
for subnet, v := range subnets.SubnetList() {
    bestPeerSubnets[subnet] = uint16(v) // #nosec G115 -- subnets has a constant max len of 128
}
```
the `subnets.SubnetList` returns a list of subnet-numbers in 0-127 range (eg. 32, 55, 99) while the code here ^ seemingly assumes it returns "a slice of up to 128 elements with each element value being = how many peers a subnet has".

While browsing this code I've also cleaned-up/clarified some things (hence this PR isn't a 1-liner fix).


network/commons/subnets.go
```diff
@@ -137,26 +137,36 @@ func (s *Subnets) Clear(i uint64) {
 	s.v[byteIndex] &^= 1 << bitIndex
 }
 
+// String returns human-readable subnets representation.
 func (s *Subnets) String() string {
+	var result strings.Builder
```
nkryuchkov: Can we use something like `strings.Join` for simplicity?

nkryuchkov (approved):

network/commons/subnets.go
iurii-ssv: `strings.Join` expects a slice of strings while `s.SubnetList()` returns a `[]uint64` (so we would still need to iterate over it and covert each element),

but we can actually use `fmt.Sprint(s.SubnetList())` to simplify this code (applied)

PR #2603, opened 2025-11-24
majority fork protection: remove target root check

nkryuchkov:
While checking the target root is correct, it seems to cause some performance drops on reorgs.

This PR resolves the performance drops by removing the target root check until we come up with a better solution.

### DIFF
--- a/protocol/v2/ssv/value_check.go
+++ b/protocol/v2/ssv/value_check.go
@@ -69,6 +69,7 @@ func (v *voteChecker) CheckValue(value []byte) error {
 	}
 
 	// Implemented according to https://github.com/ssvlabs/SIPs/pull/69
+	// Not checking target root due to possible orphaned blocks.
 	if bv.Source.Epoch != v.expectedVote.Source.Epoch {
 		return fmt.Errorf("unexpected source epoch %v, expected %v", bv.Source.Epoch, v.expectedVote.Source.Epoch)
 	}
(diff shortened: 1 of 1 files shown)


iurii-ssv (dismissed):
As an alternative to turning this check off entirely we maybe could just relax the check to allow any of:
- bv.Target.Root == v.expectedVote.Target.Root
- bv.Target.Root == PREVIOUSLY_CACHED_BLOCK_ROOT (if it is known)
- bv.Target.Root == ANY (if PREVIOUSLY_CACHED_BLOCK_ROOT is not known)

that should prevent forks but still allow for delays in block propagation ? This would need some testing though (so if we want to go this route it's probably better to add it later on - on top of this PR).

protocol/v2/ssv/value_check.go
```diff
@@ -69,6 +69,7 @@ func (v *voteChecker) CheckValue(value []byte) error {
 	}
 
 	// Implemented according to https://github.com/ssvlabs/SIPs/pull/69
+	// Not checking target root due to possible orphaned blocks.
```
moshe-blox: misplaced comment?

moshe-blox (dismissed):

protocol/v2/ssv/value_check.go
nkryuchkov: We don't have that check anymore, so commenting where I deleted it wouldn't make sense. Also, https://github.com/ssvlabs/SIPs/pull/69 checks the whole checkpoint, so the comment adds a note that we removed this check. Perhaps I need to rewrite the comment a bit though

moshe-blox (dismissed):

moshe-blox (approved):

iurii-ssv (approved):

PR #2605, opened 2025-11-25
majority fork protection: add the strict mode with a flag

nkryuchkov:
Add a flag for the target root check.

**How to Enable**

  ```yaml
  ssv:
    ValidatorOptions:
      MajorityForkProtectionStrict: true
  ```

  *Environment variable:*

  ```bash
  MAJORITY_FORK_PROTECTION_STRICT=true
  ```

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -171,7 +171,8 @@ type Controller struct {
 
 // NewController creates a new validator controller instance.
 func NewController(logger *zap.Logger, options ControllerOptions, exporterOptions exporter.Options) *Controller {
-	logger.Debug("setting up validator controller")
+	logger.Debug("setting up validator controller",
+		zap.Bool("mfp_strict", options.MajorityForkProtectionStrict))
 
 	// lookup in a map that holds all relevant operators
 	operatorsIDs := &sync.Map{}
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -89,9 +89,10 @@ type ControllerOptions struct {
 	ProposerDelay                  time.Duration
 
 	// worker flags
-	WorkersCount    int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
-	QueueBufferSize int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
-	GasLimit        uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	WorkersCount                 int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
+	QueueBufferSize              int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
+	GasLimit                     uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	MajorityForkProtectionStrict bool   `yaml:"MajorityForkProtectionStrict" env:"MAJORITY_FORK_PROTECTION_STRICT" env-description:"Check if the attestation target root matches the proposed one. It should improve Ethereum network split detection in case of global network incidents. However, it can worsen the performance on reorgs."`
 }
 
 type Nonce uint16
@@ -197,6 +198,7 @@ func NewController(logger *zap.Logger, options ControllerOptions, exporterOption
 		options.MessageValidator,
 		options.Graffiti,
 		options.ProposerDelay,
+		options.MajorityForkProtectionStrict,
 	)
 
 	cacheTTL := 2 * options.NetworkConfig.EpochDuration() // #nosec G115
@@ -1050,6 +1052,7 @@ func SetupCommitteeRunners(
 			options.OperatorSigner,
 			dutyGuard,
 			options.DoppelgangerHandler,
+			options.MajorityForkProtectionStrict,
 		)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


operator/validator/controller.go
```diff
@@ -170,7 +171,8 @@ type Controller struct {
 
 // NewController creates a new validator controller instance.
 func NewController(logger *zap.Logger, options ControllerOptions, exporterOptions exporter.Options) *Controller {
-	logger.Debug("setting up validator controller")
+	logger.Debug("setting up validator controller",
+		zap.Bool("mfp_strict", options.MajorityForkProtectionStrict))
```
iurii-ssv, line 175: Maybe we could log this `mfp_strict` value along with the rest of `options` (eg. `options.ProposerDelay`) somewhere else ? Seems kind of "out of place" to log it here.

protocol/v2/ssv/runner/committee.go
```diff
@@ -54,6 +54,7 @@ type CommitteeRunner struct {
 	DutyGuard           CommitteeDutyGuard
 	doppelgangerHandler DoppelgangerProvider
 	measurements        *dutyMeasurements
+	mfpStrict           bool
```
iurii-ssv, line 57: Could use a 1-liner comment for what it is

protocol/v2/ssv/value_check.go
```diff
@@ -23,6 +23,7 @@ type voteChecker struct {
 	sharePublicKeys       []phase0.BLSPubKey
 	estimatedCurrentEpoch phase0.Epoch
 	expectedVote          *spectypes.BeaconVote
+	mfpStrict             bool
```
iurii-ssv, line 26: Could use a 1-liner comment for what it is

protocol/v2/ssv/value_check.go
```diff
@@ -68,8 +71,8 @@ func (v *voteChecker) CheckValue(value []byte) error {
 		}
 	}
 
-	// Implemented according to https://github.com/ssvlabs/SIPs/pull/69, except the target root check is deleted
-	// because reorgs/orphaned blocks will trigger the protection and cause performance drops.
+	// Implemented according to https://github.com/ssvlabs/SIPs/pull/69, except the target root check is optional,
+	// because reorgs/orphaned blocks will trigger the target root mismatch and cause performance drops.
```
iurii-ssv: `performance drops` -> `attestation miss(es)` to be precise ?

operator/validator/controller.go
```diff
@@ -89,9 +89,10 @@ type ControllerOptions struct {
...
-	GasLimit        uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	WorkersCount                 int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
+	QueueBufferSize              int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
+	GasLimit                     uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	MajorityForkProtectionStrict bool   `yaml:"MajorityForkProtectionStrict" env:"MAJORITY_FORK_PROTECTION_STRICT" env-description:"Check if the attestation target root matches the proposed one. It should improve Ethereum network split detection in case of global network incidents. However, it can worsen the performance on reorgs."`
```
iurii-ssv: Maybe `It should improve Ethereum network split detection in case of global network incidents. However, it can worsen the performance on reorgs.` -> `It should prevent Ethereum network splits in case of global network incidents. However, it can result in attestation misses on reorgs (and when block-propagation has been delayed).`

iurii-ssv (dismissed):
Minor suggestions

protocol/v2/ssv/runner/committee.go
nkryuchkov: I didn't like using `majorityForkProtectionStrict`. I thought it was too long, given that passing it is easy to follow. I'd rather rename it to `majorityForkProtectionStrict` than comment because I don't like copying the explanation from `operator/validator/controller.go`. Do you want me to rename it?

operator/validator/controller.go
nkryuchkov: I kind of agree that there might be a better place, but honestly, I didn't find it. Perhaps `CommitteeRunner` would be the best, but I don't like adding a new log just to log this variable

protocol/v2/ssv/value_check.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/2605#discussion_r2564143201

operator/validator/controller.go
nkryuchkov: @iurii-ssv We cannot prevent a network split. We can just stop attesting in this case

iurii-ssv: Yeah, right, that's what I meant :)

It's just that "performance" typically refers to attestation Correctness/Effectiveness I think, not Misses.

protocol/v2/ssv/runner/committee.go
iurii-ssv: The main reason I'd like it to be clarified (long name is also fine to me, but I think it's fine to copy-paste a comment as well in these cases) is because this variable is NOT limited to some small scope.

operator/validator/controller.go
iurii-ssv: Yeah it doesn't seem like we log the entire config anywhere (eg. `ProposerDelay` is only logged selectively in certain places) ... so I guess we can add it later and move it there

iurii-ssv (dismissed):

operator/validator/controller.go
moshe-blox: > Refuse to sign attestations with different target roots than your own. This does not increase validator safety. Instead, this is potentially more beneficial for Ethereum (area of active research), at the cost of potentially decrease validator performance during transitory disagreements between operators.

this might be more understandable for operators who read this (it says what it actually means for them)

nkryuchkov: @moshe-blox LGTM, except I'd mention that performance may be decreased on orphaned blocks/reorgs, otherwise it's not clear where it comes from

iurii-ssv (approved):

moshe-blox (approved):

Issue #2526, opened 2025-09-15
validator-registrations: ensure safe duty scheduling/execution


iurii-ssv:
Now that https://github.com/ssvlabs/ssv/pull/2130 is merged we are scheduling validator-registration duties for execution every time we get a "fee-recipient updated" event (an EL event),

this result in Validator fee recipient changing much faster (in the next 2 epochs at most, vs the next 10 epochs as it was previously), but also introduces a potential issue of overwhelming SSV node with the sheer amount of validator-registration duties scheduled for execution - this is because:
- 1 validator registration duty is scheduled per 1 fee-recipient change
- additionally/independently, periodically we send registrations for every Validator SSV Operator manages (say 3000 registrations spread out throughout 10 epochs = 300 registrations per epoch = 10 registrations per slot)

if all 3000 validators change fee recipient at the same time - this might cause SSV node to overload temporarily until all the corresponding validator-registration duties are processed/dropped (we have some rate limits in place),

during QA testing of https://github.com/ssvlabs/ssv/pull/2130:
- we found no critical issues when updating ~300 fee-recipients at a time (this is the largest number we could test this case with), all other duties were executing successfully, there was no significant increase in CPU/Memory usage - the main resource-usage spike was observed in the bandwidth department (to handle all the p2p messages sent around)
- we did notice some increase in latency for all duty executions though (attestations, aggregations, but other duty types are also probably affected to some extent as well) which might have been more of an issue with **stage** env than anything else ... but allowing all 3000 validators to change fee-recipients that will trigger 3000 validator-registrations duties might have more dramatic effect

the simplest way to solve this would be to just limit the amount of validator-registrations we are willing to process "in real time" (those resulting from fee-recipient change events) - 300 value we've tested with seems like a reasonable limit

iurii-ssv:
Some useful notes (on how validator-registrations are handled):
- Ethereum events comes in with an `owner` on it
- we iterate over all our validators to find those shares that have that `owner`, lets say we found 100 of them
- we generate 100 validator-registration duties (they will execute in 4 slots from the current one to make sure every node in SSV cluster has had enough time to receive the same event)
- each duty executes, maybe some fail/time-out if there are too many - this is a best effort kind of thing (we won't be able to process 1000+ duties but we will be able to process ~10-100 I think)
[...]

PR #2608, opened 2025-11-26
fee-recipient-update: self-throttle on-demand validator-registrations

iurii-ssv:
This PR implements the self-throttle suggestion from https://github.com/ssvlabs/ssv/issues/2526

while we didn't discover any prominent issues during our recent testing, I still think it's good to have this implemented:
- we do not want to flood the corresponding p2p topic with "low-priority" validator-registration messages (and it also affects other clusters that are using the same p2p topic) - while COMMITTEE/AGGREGATOR duties are still executing successfully they noticeably drop in efficiency and PROPOSALs (if any happen during the affected slot(s)) will likely fail due to the added p2p delays
- for mainnet, the effects are hard to predict (perhaps for already congested topics the added delays will become "too much to handle")

Closes ssvlabs/ssv-node-board#1048


operator/validator/task_executor.go
```diff
@@ -81,47 +82,65 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+			select {
+			case <-c.ctx.Done():
+				vLogger.Debug("context is done - not gonna schedule validator registration")
+			case c.validatorRegistrationCh <- regDesc:
+				vLogger.Debug("added validator registration task to pipeline")
+			case <-time.After(2 * c.networkConfig.SlotDuration):
+				vLogger.Error("failed to schedule validator registration duty!")
 			}
 		}()
 	}
```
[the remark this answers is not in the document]
iurii-ssv: Should be fixed via https://github.com/ssvlabs/ssv/pull/2608/commits/4603aa33f792551827fa1e7fd322ab2b73b4d8f0

operator/validator/task_executor.go
```diff
@@ -81,47 +82,65 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+	)
+
+	// scheduledValidatorRegsLimit limits how many validator-registration duties (per owner) we are allowed to
+	// schedule. Ideally we'd want to schedule them all ... but this self-throttling ensures we do not overwhelm
+	// ourselves and the p2p network with lots of duties/messages (updating fee-recipient via validator-registrations
+	// is a low priority thing compared to other duties validator performs, so we are gonna do it on a best-effort
+	// basis).
+	// Note, we MUST sort shares for the target owner here somehow so that all SSV nodes in the cluster schedule
+	// the "same duty" (otherwise we won't have a majority of the cluster for some duties to work with).
+	const scheduledValidatorRegsLimit = 100
```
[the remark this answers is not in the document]
iurii-ssv: I set it to 100 to keep it "conservative", keep in mind that a single cluster might need to process fee-recipient updates from **different** owners = that means this 100 will be multiplied by however many owners are updating fee-recipient concurrently (that's why it is "conservative").

operator/validator/task_executor.go
```diff
@@ -81,47 +82,66 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+		go scheduledValidatorReg(vLogger, r)
+	}
+
+	// Notify (without blocking) the fee recipient controller about the fee recipient address change
+	// so it can submit updated proposal preparations with the new fee recipient.
+	go func() {
+		if !c.reportFeeRecipientChange(c.ctx) {
+			logger.Error("failed to notify fee recipient change")
+		}
+	}()
```
iurii-ssv: Note, if we are executing this task here - it already means some fee-recipient has been updated (because [we only spawn this task when this is the case](https://github.com/ssvlabs/ssv/blob/998d9e2cf699f0e85dfabbc271a64ce9aef65cb8/eth/eventhandler/event_handler.go#L379-L405)) ... so this additional `if updated` check we had here previuosly isn't needed

iurii-ssv:
@greptile could you review again ?

operator/validator/task_executor.go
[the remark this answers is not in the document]
iurii-ssv: I've explained why it's okay to do it like that (why we don't need this additional check) in a comment above - https://github.com/ssvlabs/ssv/pull/2608#discussion_r2564682980

> Note, if we are executing this task here - it already means some fee-recipient has been updated (because [we only spawn this task when this is the case](https://github.com/ssvlabs/ssv/blob/998d9e2cf699f0e85dfabbc271a64ce9aef65cb8/eth/eventhandler/event_handler.go#L379-L405)) ... so this additional `if updated` check we had here previuosly isn't needed

does it make sense ?

operator/validator/task_executor.go
```diff
@@ -81,47 +82,66 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+		case <-c.ctx.Done():
+			logger.Debug("context is done - not gonna schedule validator registration")
+		case c.validatorRegistrationCh <- r:
+			logger.Debug("added validator registration task to pipeline")
+		case <-time.After(2 * c.networkConfig.SlotDuration):
+			logger.Error("failed to schedule validator registration duty!")
+		}
 	}
 
+	for _, s := range ownerShares[:min(scheduledValidatorRegsLimit, len(ownerShares))] {
```
nkryuchkov, line 129: Where do we process the rest if `len(ownerShares) > scheduledValidatorRegsLimit`?

iurii-ssv: We don't. Basically as per what the comment on `scheduledValidatorRegsLimit` explains:
```go
	// scheduledValidatorRegsLimit limits how many validator-registration duties (per owner) we are allowed to
	// schedule. Ideally we'd want to schedule them all ... but this self-throttling ensures we do not overwhelm
	// ourselves and the p2p network with lots of duties/messages (updating fee-recipient via validator-registrations
	// is a low priority thing compared to other duties validator performs, so we are gonna do it on a best-effort
	// basis).
	// Note, we MUST sort shares for the target owner here somehow so that all SSV nodes in the cluster schedule
	// the "same duty" (otherwise we won't have a majority of the cluster for some duties to work with).
	const scheduledValidatorRegsLimit = 100
```

I've added a newline in between this comment and `scheduledValidatorRegsLimit` now to make it clear that the comment is about this whole block of code below it (rather than specific to `scheduledValidatorRegsLimit` itself)

operator/validator/task_executor.go
```diff
@@ -81,47 +82,67 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+		case <-c.ctx.Done():
+			logger.Debug("context is done - not gonna schedule validator registration")
+		case c.validatorRegistrationCh <- r:
+			logger.Debug("added validator registration task to pipeline")
+		case <-time.After(2 * c.networkConfig.SlotDuration):
+			logger.Error("failed to schedule validator registration duty!")
+		}
+	}
+
+	for _, s := range ownerShares[:min(scheduledValidatorRegsLimit, len(ownerShares))] {
```
olegshmuelov, line 129: When an owner has more than 100 validators, the excess are silently deferred to the periodic handler (~64 min).

Consider logging when throttling kicks in so operators are aware:

```go
if len(ownerShares) > scheduledValidatorRegsLimit {
    logger.Info("throttling validator registrations",
        zap.Int("total", len(ownerShares)),
        zap.Int("scheduled", scheduledValidatorRegsLimit))
}
```

operator/validator/task_executor.go
```diff
@@ -81,47 +82,67 @@ func (c *Controller) ReactivateCluster(owner common.Address, operatorIDs []spect
...
+		go scheduledValidatorReg(vLogger, r)
 	}
 
+	// Notify (without blocking) the fee recipient controller about the fee recipient address change
+	// so it can submit updated proposal preparations with the new fee recipient.
+	go func() {
+		if !c.reportFeeRecipientChange(c.ctx) {
+			logger.Error("failed to notify fee recipient change")
+		}
+	}()
```
olegshmuelov: The event handler only creates this task when a real on-chain update happened, but `updated == true` means "the fee recipient changed for this owner," not "this node has validators for this owner."

When the event is for an owner with no local validators, `ownerShares` is empty, no registrations are scheduled, yet `reportFeeRecipientChange` still fires, triggering a full `SubmitProposalPreparations` cycle (fetches all active shares, builds preparations, beacon RPC) for unrelated validators.

Consider restoring the guard:

```go
if len(ownerShares) > 0 {
    go func() {
        if !c.reportFeeRecipientChange(c.ctx) {
            logger.Error("failed to notify fee recipient change")
        }
    }()
}
```

operator/validator/task_executor.go
iurii-ssv: This actually isn't entirely correct, see https://github.com/ssvlabs/ssv/pull/2608#discussion_r2837773655 as a continuation of this topic

operator/validator/task_executor.go
iurii-ssv: Hmm, you are right, 

for some reason I thought the event-handler takes care of events that are not relevant for our Operator (skipping them without creating a `Task`) ... but looks like we "always" create a `Task` and "expect" to handle the skipping inside of it,

added the `if len(ownerShares) > 0` check you've suggested in https://github.com/ssvlabs/ssv/pull/2608/changes/804cf15d4630fd2213baf01471af02e5efe080b4

olegshmuelov (approved):
LGTM.

y0sher (approved):

y0sher (approved):

PR #2609, opened 2025-11-26
qbft-instance: log stage durations

iurii-ssv:
This PR adds the missing logging for every **stage** of QBFT instance to complement the metrics we record (while metrics show a general picture of how things look, they don't help when debugging a specific QBFT instance misbehavior).

Also, `round_change` stage was added so we can track it as well (instead of "bundling" it with the rest of the stages).


protocol/v2/qbft/controller/decided.go
```diff
@@ -23,7 +24,7 @@ func (c *Controller) UponDecided(msg *specqbft.ProcessingMessage) (*spectypes.Si
 	isFutureDecided := msg.QBFTMessage.Height > c.Height
 
 	if inst == nil {
-		i := instance.NewInstance(c.GetConfig(), c.CommitteeMember, c.Identifier, msg.QBFTMessage.Height, c.OperatorSigner)
+		i := instance.NewInstance(zap.NewNop(), c.GetConfig(), c.CommitteeMember, c.Identifier, msg.QBFTMessage.Height, c.OperatorSigner)
```
iurii-ssv, line 28: Since we are going to get rid of "decided" messages anyway I'm just passing the `zap.NewNop()` here for simplicity.

protocol/v2/qbft/instance/metrics.go
```diff
@@ -7,37 +7,57 @@ import (
...
 
-func (m *metricsRecorder) EndStage(ctx context.Context, round specqbft.Round, s stage) {
+func (m *metricsRecorder) EndStage(ctx context.Context, round specqbft.Round) {
+	took := time.Since(m.stageStart)
+
+	m.logger.Debug("stage finished",
+		fields.QBFTRound(round),
+		zap.String("stage", string(m.stage)),
+		fields.Took(took),
+	)
```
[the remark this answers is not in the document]
iurii-ssv: That's kind of intentional, in a sense that it's hard to "cover all the cases", we just need to monitor we never hit the `undefinedStage` case in our logs/metrics

y0sher (dismissed):

protocol/v2/qbft/controller/decided.go
nkryuchkov: @iurii-ssv I think it needs to be commented, because after merging it won't be clear why we decided to do this

nkryuchkov (dismissed):

protocol/v2/qbft/controller/decided.go
iurii-ssv: Added the comment 👍

nkryuchkov (approved):

protocol/v2/qbft/instance/metrics.go
```diff
@@ -7,37 +7,57 @@ import (
...
 		ctx,
-		time.Since(m.stageStart).Seconds(),
+		took.Seconds(),
 		metric.WithAttributes(
-			stageAttribute(s),
+			stageAttribute(m.stage),
 			observability.RunnerRoleAttribute(m.runnerRole),
 			observability.DutyRoundAttribute(round)),
 	)
-	m.stageStart = time.Now()
```
vyzo, line 40: so where is this initialized now?

vyzo (commented):
all this looks reasonable to me, although it is hard to say if it misses something that should be logged.

left a comment.

protocol/v2/qbft/instance/metrics.go
iurii-ssv: Now we call `StartStage` explicitly every time a new stage starts (which does the `m.stageStart = time.Now()` as well), while previously we'd do it here implicitly in the `EndStage` func. 

So the behavior stays the same in that respect, it just is a bit easier to follow I think.

vyzo: this is much better, indeed.

vyzo (approved):

y0sher (approved):
lgtm. the only issue this might add more load on our logging infra ( cc: @zktaiga )

iurii-ssv:
> lgtm. the only issue this might add more load on our logging infra ( cc: @zktaiga )

Shouldn't be too bad since it only logs more stuff for QBFT-related messages (and not pre/post consensus messages, for example)

PR #2613, opened 2025-11-28
operator/duties: fix missing duty fetch on an epoch boundary

nkryuchkov:
Eliminates missed attestations/sync messages after restarting near a boundary.

Changes:

  - Add startup prefetch when starting in the last 2 slots of an epoch/period.
  - Add tick‑time “fetch‑on‑miss” to immediately fetch duties if the new epoch/period store is empty.
  - Applies to attester and sync‑committee handlers.

### DIFF
--- a/operator/duties/attester.go
+++ b/operator/duties/attester.go
@@ -84,6 +84,11 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
 			slot := h.ticker.Slot()
 			next = h.ticker.Next()
 			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(slot)
+			// If we don't yet have duties recorded for the current epoch (e.g., restart right
+			// before/after an epoch boundary), force a fetch of current-epoch duties.
+			if !h.duties.IsEpochSet(currentEpoch) {
+				h.fetchCurrentEpoch = true
+			}
 			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, slot, slot%32+1)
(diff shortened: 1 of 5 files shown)


operator/duties/attester.go
```diff
@@ -84,6 +84,11 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
 			slot := h.ticker.Slot()
 			next = h.ticker.Next()
 			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(slot)
+			// If we don't yet have duties recorded for the current epoch (e.g., restart right
+			// before/after an epoch boundary), force a fetch of current-epoch duties.
+			if !h.duties.IsEpochSet(currentEpoch) {
+				h.fetchCurrentEpoch = true
+			}
```
iurii-ssv, line 91: Can we move this right below the `🛠 ticker event` line ? I think we should log the `🛠 ticker event` first, and then do all the logic related to this tick

**But actually** I think this code belongs into `HandleInitialDuties` func since it really only is needed upon SSV-node start, right ? I guess another edge-case is related to reorgs, but those seem to be addressed under `case reorgEvent` separately/explicitly

iurii-ssv (commented):
Just 1 suggestion,

Also, I see the change for **attester** and **sync_committee** duties ... but don't we also need this for other duty-types (**proposer**, not sure about **aggregator** - I think it is covered by **attester** duty-fetcher/handler already?)

nkryuchkov:
superseded by #2614

PR #2614, opened 2025-12-01
duties: fix fetch upon node start (at epoch boundary)

iurii-ssv:
Similar (replaces) to https://github.com/ssvlabs/ssv/pull/2613 but 
- better fits with the current implementation, I think
- and makes some adjustments to request timeouts(deadline)

To summarize the issue this PR solves:
- the `slotTicker` starts ticking on the slot that's coming **right after** the slot it has been initialized on
- in **stage** branch `HandleInitialDuties` only fetches duties for the current epoch (but NOT also for the next epoch)
- the `slotTicker` attempts to fetch duties for the next epoch on **every** tick (it also serves as a sort-of retry mechanism)
- hence, if SSV node starts at the last slot of the epoch, we will NOT fetch duties for the next epoch in `HandleInitialDuties` and since the `slotTicker` will happen only on the 1st slot of the next epoch it will also NOT fetch duties for that (now current) epoch because of how the code is written (how we update `h.fetchCurrentEpoch` value)

Solution:
- `HandleInitialDuties` should attempt (express the intent) to fetch duties for both current and next epochs

For testing, see https://github.com/ssvlabs/ssv/compare/stage...duties-fix-fetch-upon-node-start-testing (it's rebased onto this PR, with all the testing-utilities added in the last commit)

TODOs:
- [x] for **proposer** duty, we need add similar kind of improvement ... but it seems its code is a bit different - need to figure this out
- [x] for **sync_comittee** duty, we could apply similar optimization that allows for duty-fetching for the next period only when we are "close" to it (has passed beyound 1/2 of the current period mark)
- [x] see if we can copy the tests added in https://github.com/ssvlabs/ssv/pull/2613 over here


operator/duties/attester.go
```diff
@@ -91,7 +90,7 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
 				// Aggregates submissions are rewarded as long as they are finished within 2 slots after the target slot
 				// (the target slot itself, plus the next slot after that), hence we are setting the deadline here to
 				// target slot + 2.
-				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot)
+				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot+1)
```
iurii-ssv, line 97: This seems like a typo, the comment says that we need to allow for a 2-slot deadline (I think the comment is correct here), so bumping it up.

operator/duties/attester.go
```diff
@@ -156,10 +155,16 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
+// HandleInitialDuties fetches duties for the current and next epochs.
+// Fetching duties for the next epoch is necessary if we are starting close to epoch-boundary because
+// our ticker might "miss" that rollover otherwise.
 func (h *AttesterHandler) HandleInitialDuties(ctx context.Context) {
-	ctx, cancel := context.WithTimeout(ctx, h.beaconConfig.SlotDuration/2)
+	ctx, cancel := context.WithTimeout(ctx, h.beaconConfig.SlotDuration)
 	defer cancel()
 
+	h.fetchCurrentEpoch = true
+	h.fetchNextEpoch = true
```
iurii-ssv, line 170: I think we need to bump this timeout(deadline) here a bit just in case (also because it now potentially covers 2 fetch-calls instead of just 1). Applied to all duty-types.

operator/duties/scheduler.go
```diff
@@ -213,7 +213,11 @@ func (s *Scheduler) Start(ctx context.Context) error {
 		)
 
 		// This call is blocking.
-		handler.HandleInitialDuties(ctx)
+		err := handler.HandleInitialDuties(ctx)
+		if err != nil {
+			return fmt.Errorf("handle initial duties: %w", err)
+		}
```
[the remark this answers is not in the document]
iurii-ssv: This was a typo that is fixed now (force-pushed)

iurii-ssv:
> This PR cannot be merged as-is due to a compilation error

This was a typo, should be fixed now.

nkryuchkov:

iurii-ssv:
> for **proposer** duty, we need add similar kind of improvement ... but it seems its code is a bit different - need to figure this out

Because Proposer-duty is implemented slightly differently, it seems the issue this PR is aiming to solve doesn't apply there. Although Proposer-duty could use some other improvements - https://github.com/ssvlabs/ssv/issues/2619

> see if we can copy the tests added in https://github.com/ssvlabs/ssv/pull/2613 over here

It's not super easy to adjust those test for this PR, so I don't think it's worth the time (also because this code is likely to change - we'd need to re-write unit-tests anyway).

nkryuchkov (dismissed):

operator/duties/sync_committee.go
```diff
@@ -25,9 +25,14 @@ import (
...
 
-	duties             *dutystore.SyncCommitteeDuties
+	duties *dutystore.SyncCommitteeDuties
+
+	// fetchCurrentPeriod stores the intent to fetch duties for the current period, while
+	// processFetching func uses this value to decide on whether the fetch is needed.
 	fetchCurrentPeriod bool
-	fetchNextPeriod    bool
+	// fetchCurrentPeriod stores the intent to fetch duties for the next period, while
+	// processFetching func uses this value to decide on whether the fetch is needed.
```
olegshmuelov: The comment for `fetchNextPeriod` incorrectly says "fetchCurrentPeriod" instead of "fetchNextPeriod". This appears to be a copy-paste error from the line above.

  ```go
  // fetchNextPeriod stores the intent to fetch duties for the next period, while
  // processFetching func uses this value to decide on whether the fetch is needed.
  fetchNextPeriod bool
  ```

operator/duties/sync_committee.go
```diff
@@ -142,10 +138,23 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
 	}
 }
 
+// HandleInitialDuties fetches duties for the current and next epochs.
+// Fetching duties for the next epoch is necessary if we are starting close to epoch-boundary because
+// our ticker might "miss" that rollover otherwise.
```
olegshmuelov: The function comment says "fetches duties for the current and next epochs" but sync committee duties are period-based, not epoch-based. The comment should say "current and next periods" for consistency with the actual implementation.

  ```go
  // HandleInitialDuties fetches duties for the current and (if appropriate) next periods.
  // Fetching duties for the next period is necessary if we are starting close to period-boundary because
  // our ticker might "miss" that rollover otherwise.
  ```

operator/duties/proposer.go
```diff
@@ -119,8 +119,11 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
 	}
 }
 
+// HandleInitialDuties fetches duties for the current and next epochs.
+// Fetching duties for the next epoch is necessary if we are starting close to epoch-boundary because
+// our ticker might "miss" that rollover otherwise.
```
olegshmuelov: The comment says "fetches duties for the current and next epochs" but the implementation only fetches duties for the current epoch. The comment appears to be copy-pasted from the attester handler. 

  ```go
  // HandleInitialDuties fetches duties for the current epoch with an extended timeout.
  ```

olegshmuelov (dismissed):

operator/duties/proposer.go
iurii-ssv: Correct 👍 - I copy-pasted and forgot to update :) Fixed in https://github.com/ssvlabs/ssv/pull/2614/commits/9d3d0ce04574378fac5b718197e176cb545c3adb

operator/duties/sync_committee.go
iurii-ssv: Fixed in https://github.com/ssvlabs/ssv/pull/2614/commits/9d3d0ce04574378fac5b718197e176cb545c3adb

operator/duties/sync_committee.go
iurii-ssv: Yep, fixed in https://github.com/ssvlabs/ssv/pull/2614/commits/9d3d0ce04574378fac5b718197e176cb545c3adb

iurii-ssv:
Bumped the `baseDuration` (from 100ms to 200ms) since it's likely this failure https://github.com/ssvlabs/ssv/actions/runs/20102727352/job/57677827914?pr=2614 is cased by it (and it was [reported previously as case 2 here](https://github.com/ssvlabs/ssv/issues/1577) too)

olegshmuelov (approved):

nkryuchkov (approved):

y0sher (approved):

PR #2622, opened 2025-12-08
CL/EL: clean handling for context cancel

iurii-ssv:
The `context canceled` error means we've canceled request on our own, so we shouldn't be treating/recording it as an error (to keep error-list clean and legible).


y0sher (approved):
lgtm

nkryuchkov (approved):

Issue #2516, opened 2025-09-10
Build process sets incorrect version for branches not descended from latest tag


olegshmuelov:
**Description:**
The current build process in [Makefile:119](https://github.com/ssvlabs/ssv/blob/8fc1632c2d5278042ede70e1b8d46d52375bfa1a/Makefile#L119) uses:
```bash
git describe --tags $(git rev-list --tags --max-count=1)
```

  This always returns the latest tag in the repository, regardless of the current branch. This causes builds from unrelated branches to show incorrect version numbers.

  **Example:**
  - Latest tag in repo: v2.3.3
  - Building from feature/xyz branch (based on v2.2.0)
  - Result: binary shows v2.3.3-<commit> instead of correct version

  **Impact:**
  - Multiple nodes can show same version tag while running different code
  - Complicates debugging and version tracking
  - Potential consensus issues if operators think they're running same version

  **Suggested fix:**
  Replace with:
```bash
  git describe --tags --always
```

  This correctly shows the version relative to the current branch's history.

  Affected files:
  - [Makefile:119](https://github.com/ssvlabs/ssv/blob/8fc1632c2d5278042ede70e1b8d46d52375bfa1a/Makefile#L119)
  - [Dockerfile:40](https://github.com/ssvlabs/ssv/blob/8fc1632c2d5278042ede70e1b8d46d52375bfa1a/Dockerfile#L44)
  - [ssvsigner/Dockerfile:40](https://github.com/ssvlabs/ssv/blob/8fc1632c2d5278042ede70e1b8d46d52375bfa1a/Dockerfile#L40)

PR #2626, opened 2025-12-10
git: stabilize tag versioning

iurii-ssv:
This PR corrects `VERSION` attribute initialization such that it will always be:
- either initialized with the corresponding tag via `git describe --tags` lookup (if such tag exists)
- or explicitly mention that there is no tag associated with that commit (**untagged**), while previously it would assign the latest tag available (which is confusing in some situations since that tag would correspond to a different commit, not the commit we are actually running ... which is what `COMMIT` attribute is assigned to)

Example output (for untagged commit):
- before
```
{"level":"\u001b[34mINFO\u001b[0m","time":"2025-12-10T09:35:12.507697Z","msg":"starting SSV-Node:v2.3.9-unstable.2-4be3b68507921713e0fcc3c6de7e84864ec60076"}
```
- after
```
{"level":"\u001b[34mINFO\u001b[0m","time":"2025-12-10T12:05:13.977855Z","msg":"starting SSV-Node:untagged-5b117a30027d1b14388c1d10bab903bbf87edf66"}
```

Note, for `git describe --tags` to work properly we need to create "annotated" tags (otherwise the tag list corresponding to the same commit cannot be sorted correctly by git, potentially resulting in `VERSION` set to something like `v2.3.9-unstable.2` instead of `v2.3.9` ... which is technically correct, but a bit confusing still). I don't know of a simple way to improve that part.

Resolves https://github.com/ssvlabs/ssv/issues/2516

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -6,17 +6,17 @@ set -eo pipefail
 export DOCKER_BUILDKIT=1
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
+# Use HEAD commit if none was specified (eg. when running locally)
+COMMIT="HEAD"
 if [ -n "${SOURCE_COMMIT}" ]; then
-  APP_VERSION=$(git describe --tags --exact-match "${SOURCE_COMMIT}" 2>/dev/null || echo "untagged")
+  COMMIT="${SOURCE_COMMIT}"
   echo "Using SOURCE_COMMIT=${SOURCE_COMMIT}"
-else
-  # We are running locally, use HEAD commit
-  APP_VERSION=$(git describe --tags --exact-match HEAD 2>/dev/null || echo "untagged")
 fi
+APP_VERSION=$(git describe --tags --exact-match "${COMMIT}" 2>/dev/null || echo "untagged")
 
 echo "Build Configuration:"
 echo "  Image: ${IMAGE_NAME}"
-echo "  Commit: ${SOURCE_COMMIT}"
+echo "  Commit: ${COMMIT}"
 echo "  Version: ${APP_VERSION}"
 
 docker buildx rm ssv-multiplatform-builder 2>/dev/null || true
--- a/hooks/build
+++ b/hooks/build
@@ -7,11 +7,11 @@ export DOCKER_BUILDKIT=1
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
 if [ -n "${SOURCE_COMMIT}" ]; then
-  APP_VERSION=$(git describe --tags "${SOURCE_COMMIT}")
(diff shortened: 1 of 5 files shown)


hooks/build
```diff
@@ -7,11 +7,11 @@ export DOCKER_BUILDKIT=1
...
 
 if [ -n "${SOURCE_COMMIT}" ]; then
-  APP_VERSION=$(git describe --tags "${SOURCE_COMMIT}")
+  APP_VERSION=$(git describe --tags --exact-match "${SOURCE_COMMIT}" 2>/dev/null || echo "untagged")
   echo "Using SOURCE_COMMIT=${SOURCE_COMMIT}"
 else
-  # Extract Git tag when running locally
-  APP_VERSION=$(git describe --tags 2>/dev/null || echo "latest")
+  # We are running locally, use HEAD commit
+  APP_VERSION=$(git describe --tags --exact-match HEAD 2>/dev/null || echo "untagged")
```
olegshmuelov: nit: before this PR, the two branches had different logic (different flags, different fallbacks). Now they're nearly identical—the only difference is `"${SOURCE_COMMIT}"` vs `HEAD`. You could consolidate them:

```bash
# Use SOURCE_COMMIT if provided, otherwise default to HEAD
COMMIT_REF="${SOURCE_COMMIT:-HEAD}"
APP_VERSION=$(git describe --tags --exact-match "${COMMIT_REF}" 2>/dev/null || echo "untagged")

if [ -n "${SOURCE_COMMIT}" ]; then
  echo "Using SOURCE_COMMIT=${SOURCE_COMMIT}"
fi
```

olegshmuelov (dismissed):
LGTM!
minor suggestion

hooks/build
iurii-ssv: Yep, should be resolved in ~~https://github.com/ssvlabs/ssv/pull/2626/commits/b02506147a30d4201a6cabf03633d8fb33507078~~ https://github.com/ssvlabs/ssv/pull/2626/commits/58beeb871966afdd083f2d9d019100de80645752

iurii-ssv:
~~Bumped the `baseDuration` (from 100ms to 200ms) since it's likely this failure https://github.com/ssvlabs/ssv/actions/runs/20102727352/job/57677827914?pr=2614 is cased by it (and it was [reported previously as case 2 here](https://github.com/ssvlabs/ssv/issues/1577) too)~~ wrong PR, moved the commit over where it was meant to go - https://github.com/ssvlabs/ssv/pull/2614

olegshmuelov (approved):

julienh-ssv (approved):

nkryuchkov (approved):

y0sher (approved):

PR #2629, opened 2025-12-18
logs: trim unnecessary log-lines

iurii-ssv:
In this PR I went over the most frequently occurring log-lines to see what we can do away with. 

While at it I also noticed (and as Yosher suggested below) we could:
- add certain details (such as roots/validator-indices) to certain log-lines where it makes sense (eg. `got pre-consensus message` and `got pre-consensus quorum`)
- leverage/prefer log-lines that are universal across all duties when possible (vs every duty logging its own version of "msg received" kind of log-lines)


zktaiga (dismissed):

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -137,27 +137,19 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
...
 
-	signer := ssvtypes.PartialSigMsgSigner(signedMsg)
 	duty := r.state().CurrentDuty.(*spectypes.ValidatorDuty)
 	span.SetAttributes(
 		observability.CommitteeIndexAttribute(duty.CommitteeIndex),
 		observability.ValidatorIndexAttribute(duty.ValidatorIndex),
 	)
 
-	const gotPartialSignaturesEvent = "🧩 got partial aggregator selection proof signatures"
-	span.AddEvent(gotPartialSignaturesEvent, trace.WithAttributes(observability.ValidatorSignerAttribute(signer)))
```
y0sher, line 148: I don't think the point was to delete traces as well.

protocol/v2/ssv/runner/committee.go
```diff
@@ -508,22 +508,6 @@ func (r *CommitteeRunner) signAttesterDuty(
 		return false, partialMsg, fmt.Errorf("failed signing attestation data: %w", err)
 	}
 
-	attDataRoot, err := attestationData.HashTreeRoot()
-	if err != nil {
-		return false, partialMsg, fmt.Errorf("failed to hash attestation data: %w", err)
-	}
-
-	const eventMsg = "signed attestation data"
```
y0sher, line 516: IMO this data is useful, we should at least have it on some other logs, specifically signing root and attestation data.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -137,27 +137,19 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
...
 	}
 
-	signer := ssvtypes.PartialSigMsgSigner(signedMsg)
 	duty := r.state().CurrentDuty.(*spectypes.ValidatorDuty)
 	span.SetAttributes(
 		observability.CommitteeIndexAttribute(duty.CommitteeIndex),
 		observability.ValidatorIndexAttribute(duty.ValidatorIndex),
 	)
 
-	const gotPartialSignaturesEvent = "🧩 got partial aggregator selection proof signatures"
```
y0sher, line 147: Maybe it'll be smart to log who participated in the pre consensus quorum of we eventually happen to be aggregators

protocol/v2/ssv/runner/aggregator.go
iurii-ssv: I mean, these logs are not really needed - that's why I'm deleting these,

and that applies to traces/spans too - we should either both log & trace it or neither log and trace it (doesn't make sense to trace it but not log it),

In https://github.com/ssvlabs/ssv/pull/2629/changes/91b8f649ecf43b97168e6394c53f7f85d0091696 I've done some cleanup/improvements for the way we log roots/signers so it's comprehensive and consistent across all duty-types ... and as the result also removed some other log-lines that pretty much duplicate log-lines we already have.

protocol/v2/ssv/runner/aggregator.go
iurii-ssv: Yeah, it's a good idea, and we can actually do it for `got pre-consensus quorum` log-line (and all other "got ... quorum" lines we log too), that will cover all different duty-types, 

applied and cleaned up some other relevant logs.

protocol/v2/ssv/runner/committee.go
iurii-ssv: Somehow I thought we log it below as a summary (so wouldn't be necessary to log per every validator), but upon further checking ... it's different for every validator - so we indeed can't/don't log the summary, reverted.

protocol/v2/ssv/runner/committee.go
```diff
@@ -527,34 +527,19 @@ func (r *CommitteeRunner) signAttesterDuty(
 	return false, partialMsg, nil
 }
 
-// TODO finish edge case where some roots may be missing
```
vyzo, line 530: has this actually been adressed?

protocol/v2/ssv/runner/committee.go
```diff
@@ -574,17 +559,11 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
-		deduplicatedRoots[root] = struct{}{}
-	}
-
 	var executionErr error
 
-	span.SetAttributes(observability.BeaconBlockRootCountAttribute(len(deduplicatedRoots)))
+	span.SetAttributes(observability.BeaconBlockRootCountAttribute(len(roots)))
 	// For each root that got at least one quorum, find the duties associated to it and try to submit
-	for root := range deduplicatedRoots {
+	for _, root := range roots {
```
vyzo, line 566: is it possible to actually have duplicates here?

the pr does remove deduplication code, was that extraneous?

protocol/v2/ssv/runner/runner.go
```diff
@@ -309,39 +320,51 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
...
 	container *ssv.PartialSigContainer,
-) (bool, [][32]byte) {
-	roots := make([][32]byte, 0)
-	quorumReached := false
+) (gotAnyQuorum bool, roots map[[32]byte]map[phase0.ValidatorIndex][]spectypes.OperatorID) {
+	roots = make(map[[32]byte]map[phase0.ValidatorIndex][]spectypes.OperatorID)
 
 	for _, msg := range signedMsg.Messages {
-		quorumReachedPreviously := container.HasQuorum(msg.ValidatorIndex, msg.SigningRoot)
+		quorumReachedPreviously, _ := container.HasQuorum(msg.ValidatorIndex, msg.SigningRoot)
```
vyzo, line 367: is that an ignored error in the `_`?

observability/log/fields/fields.go
```diff
@@ -305,6 +306,18 @@ func BlockRoot(r [32]byte) zap.Field {
...
 
+// QuorumRoots turns roots map into zap.Field, we can't just use zap.Any instead of this method
+// because zap will use the json-encoder that can't digest [32]byte (so we convert it to string
+// here explicitly before json-encoding the roots).
+func QuorumRoots(roots map[[32]byte]map[phase0.ValidatorIndex][]spectypes.OperatorID) zap.Field {
+	tmp := make(map[string]map[phase0.ValidatorIndex][]spectypes.OperatorID, len(roots))
+	for k, v := range roots {
+		tmp["0x"+hex.EncodeToString(k[:])] = v
+	}
+	result, _ := json.Marshal(tmp)
```
vyzo: error is ignored.

exporter/api/conn.go
```diff
@@ -224,24 +224,24 @@ func (c *conn) sendMsg(msg []byte) (int, error) {
...
-	if !ok {
-		return
-	}
-	filter, err := fraw.MarshalJSON()
-	if err != nil {
-		logger.Error("could not parse filter", zap.Error(err))
-	}
-	logger.Debug("ws msg was sent", zap.Int("bytes", byteWritten), zap.ByteString("filter", filter))
-}
+//func (c *conn) logMsg(logger *zap.Logger, message []byte, byteWritten int) {
```
vyzo: add a comment line before this please, explaining why we are keeping this around.

protocol/v2/ssv/runner/runner.go
vyzo: i think not, but just checking... ignored last values make me uneasy.

protocol/v2/ssv/runner/runner.go
```diff
@@ -350,16 +373,18 @@ func (b *BaseRunner) basePartialSigMsgProcessing(
...
-			// Notify about first quorum only
-			roots = append(roots, msg.SigningRoot)
-			quorumReached = true
+		// We are interested in any quorum that occurs for the very first time (for this root).
+		hasQuorum, quorumSigners := container.HasQuorum(msg.ValidatorIndex, msg.SigningRoot)
+		if !quorumReachedPreviously && hasQuorum {
+			if roots[msg.SigningRoot] == nil {
+				roots[msg.SigningRoot] = make(map[phase0.ValidatorIndex][]spectypes.OperatorID)
+			}
+			roots[msg.SigningRoot][msg.ValidatorIndex] = quorumSigners
```
vyzo, line 382: i need to ask here: this slice came from a container,  is it fresh and owned by the callee or some internal subslice?

y0sher (dismissed):

vyzo (commented):
Left some comments.

Hopefully no log monitor will break from this, but hard to tell.

protocol/v2/ssv/runner/runner.go
iurii-ssv: It's not an error that `container.HasQuorum` returns but a "list of signers/operators" ... we just don't need these signers from this specific call here (in other places these signers/operators are used to print detailed debug log-lines)

protocol/v2/ssv/runner/committee.go
iurii-ssv: I believe it has been (not in this PR, but elsewhere) ... otherwise we'd know about it resulting in some issues - it's been more than 1y now,

TBH I'm not entirely sure, @y0sher do you remember anything about it ? From what I know if some roots are missing (from p2p message?) we'll just not be able to get a quorum for those roots, and that's an expected behavior (if Operators don't agree on a certain root, there won't be a quorum for that root for any validator)

I removed the comment because it seems to be stale by now, and just draws unnecessary attention.

observability/log/fields/fields.go
iurii-ssv: That's sort of by design, there isn't an easy way to log that error in a meaningful way here (we don't use global loggers) ... and we wouldn't want to return an `error` from this func either (as it wouldn't be very usable),

we could `panic` alternatively ... but since this is nothing mission-critical I think just dropping the error and dealing with consequences is the best we can do, any other suggestions ?

protocol/v2/ssv/runner/runner.go
iurii-ssv: This slice has been [created via `slices.Collect(maps.Keys(rootSigs))` here](https://github.com/ssvlabs/ssv/blob/a0c2f1c0610e0810c1c6aa6e6ef471e81f7ad36d/protocol/v2/ssv/partial_sig_container.go#L152-L152) - so `HasQuorum` returns(copying over the data `container` owns) that fresh copy to the caller for the caller to decide what to do with it

protocol/v2/ssv/runner/committee.go
iurii-ssv: Duplicates are possible,

previously:
- we'd do deduplication explicitly here (as `deduplicatedRoots` signifies)
- for some reason the committee-runner was the only duty doing that deduplication (I guess because other runners only work with 1 root ... but that's an assumption I would rather not make)

currently (this PR):
- deduplication happens in `basePartialSigMsgProcessing` (it returns `roots map[[32]byte]map[phase0.ValidatorIndex][]spectypes.OperatorID` where root is a **key** in that map)
- so every duty runner gets deduplicated roots already

exporter/api/conn.go
iurii-ssv: I've left a comment about it at the caller-site, but I guess can copy it over here too, done.

exporter/api/conn.go
```diff
@@ -137,13 +136,14 @@ func (c *conn) WriteLoop(logger *zap.Logger) {
...
-			n, err := c.sendMsg(message)
+			_, err := c.sendMsg(message)
 			c.writeLock.Unlock()
 			if err != nil {
 				logger.Warn("failed to send message", zap.Error(err))
 				return
 			}
-			c.logMsg(logger, message, n)
+			// Use this logging here for debugging only (since it produces lots of log-lines).
+			//c.logMsg(logger, message, bytesWritten)
```
nkryuchkov: Can we have a flag to enable this instead of commenting out? Or can we remove `c.logMsg` at all? It doesn't look complex, so it will be quick to implement if it's needed for debugging

observability/log/fields/fields.go
vyzo: oh no, dont panic for this.  maybe we could use something in the value to denote that there was an error instead of just empty?

not a big deal though.

protocol/v2/ssv/runner/committee.go
vyzo: ok, thats good.

vyzo (dismissed):
looks fine, lets just find out about that comment if we can.

observability/log/fields/fields.go
iurii-ssv: Yeah we can do something like https://github.com/ssvlabs/ssv/pull/2629/changes/8e362c7c9acaeda598178e57492d9e2d52b5b7bb

vyzo: yeah, thats good.

protocol/v2/ssv/runner/committee.go
y0sher: Yeah seems to be an old comment I added when we first implemented alan and committee runner rapidly. I can't recall any case of this recently so I don't think it's still valid.

y0sher (dismissed):

protocol/v2/ssv/runner/committee.go
```diff
@@ -574,17 +559,11 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 	attestationsToSubmit := make(map[phase0.ValidatorIndex]*spec.VersionedAttestation)
 	syncCommitteeMessagesToSubmit := make(map[phase0.ValidatorIndex]*altair.SyncCommitteeMessage)
 
-	// Get unique roots to avoid repetition
-	deduplicatedRoots := make(map[[32]byte]struct{})
```
nkryuchkov, line 578: Are you sure we can remove this? IIRC, we have root deduplication [in spec](https://github.com/ssvlabs/ssv-spec/blob/8f5ddfc8b9a7cf82d03c978bd2f9b46a1f4a8fc5/ssv/committee_runner.go#L179) too

protocol/v2/ssv/runner/runner.go
```diff
@@ -309,39 +320,51 @@ func (b *BaseRunner) baseConsensusMsgProcessing(ctx context.Context, logger *zap
 }
 
 // basePostConsensusMsgProcessing is a base func that all runner implementation can call for processing a post-consensus msg
-func (b *BaseRunner) basePostConsensusMsgProcessing(ctx context.Context, logger *zap.Logger, runner Runner, signedMsg *spectypes.PartialSignatureMessages) (bool, [][32]byte, error) {
+func (b *BaseRunner) basePostConsensusMsgProcessing(
+	ctx context.Context,
+	logger *zap.Logger,
+	runner Runner,
+	signedMsg *spectypes.PartialSignatureMessages,
+) (ok bool, roots [][32]byte, err error) {
```
nkryuchkov, line 328: Should we swap `ok` with `roots`? `ok` usually goes after the result

y0sher: @nkryuchkov is right but I rather refrain from a function signature change in a logs PR. also this is a change for spec I think as well, we can create a task for it.

exporter/api/conn.go
y0sher: I agree but can be done in another PR, also this log is only for exporter, is it still producing so much logs?

protocol/v2/ssv/runner/committee.go
y0sher: basePartialSigMsgProcessing seems to do the deduplication with the map as @iurii-ssv mentioned.

iurii-ssv: We are doing it now in a more universal manner -  https://github.com/ssvlabs/ssv/pull/2629#discussion_r2642688500

nkryuchkov: @y0sher @iurii-ssv great, thanks

exporter/api/conn.go
nkryuchkov: @y0sher if we can comment it out in this PR, why can't we delete it here?

vyzo (dismissed):

exporter/api/conn.go
nkryuchkov: We used to have some commented-out code in the past, and it became outdated very quickly, so we had to delete it eventually. So I think we shouldn't keep doing that

iurii-ssv: Yeah I guess commenting out is not ideal, removed.

protocol/v2/ssv/runner/runner.go
iurii-ssv: I'll have a follow-up PR with some further cleanup I've notice we can use, I'll apply that suggestion in there.

nkryuchkov (approved):

y0sher (approved):

PR #2630, opened 2025-12-18
logs: standardize default log level

iurii-ssv:
We are using `DEBUG` as default log level almost everywhere (for our own deployments, as well as 3rd party Operators are also running with that level), and `INFO` produces way too little information for troubleshooting anyway,

so it's better to be explicit and consistent about it - this PR ensures we always default to `DEBUG` log level every time (unless it's explicitly overridden with a corresponding configuration setting).


nkryuchkov:
I think before we consider doing this, we should try to review the logs, think about them, and raise the level of important logs to make the node somewhat debuggable with the INFO level.

iurii-ssv:
The original intent of this PR was to standardize logging.

Upon further digging, it looks like we do have 2 different ways to emit logs:
```
stdout - INFO is the default
rotated log files (up to 3 x 500mb files) - DEBUG is the default (and only option)
```
so it works a bit differently than what this PR assumes, hence I'll close it for now since it would indeed change that default behavior for `stdout` logs.

PR #2628, opened 2025-12-17
CL: misc WAD-related improvements

iurii-ssv:
Depends on https://github.com/ssvlabs/ssv/issues/1825 in a sense that until QBFT instance can work with "absent" value(s) there is no point in having a hard timeout/deadline for a CL request ... hence marking this PR as draft for now.

This PR improves couple of WAD-related things, mostly:
- improves logs & adds/clarifies some comments
- `record` the `BeaconBlockHeader` request so it's visible on Grafana charts that track CL-related metrics
- wraps the errors `BeaconBlockHeader` returns such that we log/track these correctly
- and most importantly, replaces the previous "soft-hard timeout" approach with a more fitting/precise deadline-based one

To expand on that last point, the issue(s) we want to solve:
- we want to make sure we are done with attestation-fetch request well before `slot_start_time + 6s` mark, so that attestation-duty can participate in and complete its QBFT in round 1 which times out at exactly 6s since slot start ... while with the current implementation we are setting **the same request timeout** (currently we use soft-timeout = 2s, hard-timeout = 4s) **regardless of when** the attestation-fetch request has started, so we are either aborting requests sooner than necessary or skipping round 1 entirely (which isn't great because of https://github.com/ssvlabs/ssv/issues/1825)
- we want to let the caller of `GoClient.GetAttestationData` to specify the deadline (think of it as a hard-deadline) because we call this method not only for attestation-duty but also for aggregator-duty (and we wouldn't want to make any unnecessary assumptions about the callers anyway)
- see also [a concrete example](https://github.com/ssvlabs/ssv/pull/2628#issuecomment-3689244396) presented below

the solution:
- the caller configures the hard-deadline on the `ctx` passed to `GetAttestationData`
- for attestation-duty that hard-deadline is `slot_start_time + 6s - pre_consensus_duration - qbft_consensus_duration - misc_time` (where `pre_consensus_duration=0ms` since it's absent for this duty-type, `qbft_consensus_duration=350ms` as [per our previous estimates here](https://github.com/ssvlabs/ssv/blob/af176b531548c20e444495423ce3e8a4a7810f84/docs/MEV_CONSIDERATIONS.md#L94-L94), `misc_time=150ms` to account for miscellaneous delays); The deadline effectively comes out to be = `slot_start_time + 5.5s` which gives us `~1.5s` to fetch attestation data in the worst-case scenario (when we started the duty `4s` since the slot start)
[...]

PR #2631, opened 2025-12-18
implement proposal scoring and best selection

vyzo:
This implements proposal selection from multiple beacons.

The behavior changes from selecting the first received one, as it has been observed in practice that it is not always the most profitable. A small time is allotted for selection, during which proposals are collected to select the one with the best score. If no valid proposal is collected at the allotted time, it falls back to the first received to avoid increasing latency too much.

The proposal score is the one used in vouch -- see https://github.com/attestantio/vouch/blob/9e03b5be9ce3c4553244b628f3d662561c1d20bc/strategies/beaconblockproposal/best/score.go#L25

The timeout strategy is also similar to vouch.

## Remaining Work

- [ ] unit test; an LLM can be coaxed to do this.
- [x] test in our staging environment


beacon/goclient/proposer.go
```diff
@@ -161,6 +169,12 @@ func (gc *GoClient) getProposalParallel(
 	parallelCtx, cancelParallel := context.WithCancel(ctx)
 	defer cancelParallel()
 
+	// Create a contet that we 'll use to collect and evaluate proposals for a short time
```
iurii-ssv: `contet that we 'll` -> `context that we'll`

beacon/goclient/proposer.go
```diff
@@ -181,28 +199,96 @@ func (gc *GoClient) getProposalParallel(
...
+		case res := <-resultCh:
+			pendingClients--
+
+			if res.err != nil {
+				errs = errors.Join(errs, res.err)
+				continue
+			}
+
 			// Got a successful response, cancel other requests and return.
 			gc.log.Debug("received proposal, canceling other requests",
```
oleg-ssvlabs: It would be good to have a log similar to [this](https://github.com/ssvlabs/ssv/pull/2631/commits/55688adf27b7801e2df6403c27608832b00a3446#diff-b36ec26955c8bcb78a560dba7283cdf962845b93795219ee0b6665da59348757R247) here, so we can easily search the logs to see which proposal was selected, regardless of whether it was chosen on a soft or hard timeout.

beacon/goclient/proposer.go
```diff
@@ -181,28 +199,96 @@ func (gc *GoClient) getProposalParallel(
...
+			return nil, hardTimeoutCtx.Err()
 		}
 	}
 
 	return nil, fmt.Errorf("all %d clients failed to get proposal for slot %d, encountered errors: %w", len(gc.clients), slot, errs)
 }
 
+// scoreProposal computes a score for a beacon proposal.
+// see https://github.com/attestantio/vouch/blob/master/strategies/beaconblockproposal/best/score.go as well
+func (gc *GoClient) scoreProposal(
```
oleg-ssvlabs, line 317: Is there a way to know whether a score is the **maximum possible score**?
If so, we could cancel the remaining outgoing HTTP requests, which would reduce the overall execution time for proposal fetching.

With the current implementation, fetching a proposal takes **at least the soft timeout**.
If we could detect that the score is already at its maximum, we could reduce the minimum execution time to the **fastest CL response time**, which could be significantly faster.

beacon/goclient/proposer.go
```diff
@@ -161,6 +169,16 @@ func (gc *GoClient) getProposalParallel(
 	parallelCtx, cancelParallel := context.WithCancel(ctx)
```
oleg-ssvlabs: Do we really still need `parallelCtx`?

The context deadline is `min(parent, child)`, so if the caller cancels the parent context earlier than the soft/hard timeout, it would also cancel both `collectCtx` and `hardTimeoutCtx`, right?

Could we just pass `ctx` [here](https://github.com/ssvlabs/ssv/blob/b5004cf5af194854a991debb8ff1f2e7a246f844/beacon/goclient/proposer.go#L192) and avoid listening to `parallelCtx` or `ctx` cancellations separately?

oleg-ssvlabs (commented):
good job, looks clean!
Left few comments

beacon/goclient/proposer.go
vyzo: no, there is no way to know in advance without asking the beacons.

beacon/goclient/proposer.go
vyzo: i think we still need it, we have a way to cancel background calls in a timely manner if we return early.

The request context is external, no way to know when it will be done.

beacon/goclient/proposer.go
vyzo: done.

vyzo: https://github.com/ssvlabs/ssv/pull/2631/commits/d057aff7df40b4174a40c0d1be86efec549e6ec6

vyzo: and adjusted in https://github.com/ssvlabs/ssv/pull/2631/commits/624d1bb6204ca88d7e3834f1870a9c4a5ced3f6f so that we have a way to discriminate whether the proposal was in the fallback path, after the soft timeout. That would be a useful metric to tells us we need to adjust the soft timeout if need be.

beacon/goclient/proposer.go
vyzo: i think we should adjust the comment though, it is misleading now.

vyzo: adjusted comment; https://github.com/ssvlabs/ssv/pull/2631/commits/de6435c30d6b748f51829e496048048233f8435e

iurii-ssv (commented):
The approach looks sound to me, we just need to pick a good `proposalCollectTimeout` value to use for now (while we are using timeouts)

and eventually I think we'd want to tighten this further by using precise deadlines (similar to how https://github.com/ssvlabs/ssv/pull/2628 is doing it for WAD)

Attaching some estimates for choosing timeout/deadline (applies to soft timeout/deadline if we are not gonna set a hard one, I guess):
```
# constraints to succeed in time for round 1
RANDAOTime + ProposerDelay + BlockFetchTime + QBFTTime + MiscellaneousTime < 2s (round 1 timeout)

# constraints to succeed in time for round 2
RANDAOTime + ProposerDelay + BlockFetchTime + QBFTTime + MiscellaneousTime + BlockSubmissionTime < 4s (duty becomes irrelevant)

# some reasonable estimates for those values:
const RANDAOTime = 100 * time.Millisecond
const QBFTTime = 350 * time.Millisecond
const MiscellaneousTime = 150 * time.Millisecond
const BlockSubmissionTime = 1000 * time.Millisecond
```
those mean that if we want our soft timeout/deadline to allow us to be in time for participation in:
- round 1, we need to define it roughly as: ~`1.4s - ProposerDelay`
- round 2, we need to define it roughly as: ~`2.4s - ProposerDelay`

beacon/goclient/proposer.go
vyzo: oops, fixed.

vyzo:
I added it to the options in https://github.com/ssvlabs/ssv/pull/2631/commits/ae96a066715817f17e2201798477cdcb7dbf2ddb and set defaults for 1s and 2s respectively.

beacon/goclient/proposer.go
iurii-ssv: Now that we have added the `hardTimeoutCtx` I think it can completely replace the `parallelCtx` ?

We could probably also use the `softCtx` and `hardCtx` names for better cohesion (and have a consistent naming vs the one we use in WAD code) ... but we'd also need to adjust the comments that mention "collection" I guess

vyzo: actually you are right, hard can replace the parallel.

beacon/goclient/proposer.go
```diff
@@ -181,28 +200,101 @@ func (gc *GoClient) getProposalParallel(
...
 				fields.Slot(slot),
 			)
-			cancelParallel()
 			return res.proposal, nil
-		case <-ctx.Done():
-			return nil, ctx.Err()
+
+		case <-hardTimeoutCtx.Done():
+			return nil, hardTimeoutCtx.Err()
 		}
```
iurii-ssv: Do you think we maybe can just omit this `case <-hardTimeoutCtx.Done()` and let all the requests run to completion ? If we are passing the `hardTimeoutCtx` into `gc.fetchProposal` it should return pretty much immediately once we hit that hard deadline

vyzo: i think it is safer to just leave it there, less cognitive load when we read it 6 months from now.

beacon/goclient/proposer.go
vyzo: done. Removed the now unnecessary parallel context, and renamed the others to soft and hard. I don't think it causes confusion (we have good comments) and it is consistent.

https://github.com/ssvlabs/ssv/pull/2631/commits/f67c9abb52a61958f8f8873b5948519a6a5b6a48

beacon/goclient/proposer.go
vyzo: I mean, pretty much all (good) go code out there does it; if we don't, even if it is correct, it will cause the next reader to scratch their head.

beacon/goclient/goclient.go
```diff
@@ -134,6 +138,15 @@ type GoClient struct {
...
 	weightedAttestationDataHardTimeout time.Duration
 
+	// proposalCollectTimeout is the maximum time to wait for multiple block proposals in
+	// order to select the best one from multiple clients.
+	// If no proposal has been received after this time elapses, the first valid proposal
+	// seen is returned
+	// proposalHardTimeout is the hard timeout for retrieving any proposals.
+	// TODO these should be configurable by the operator
+	proposalCollectTimeout time.Duration
+	proposalHardTimeout    time.Duration
```
iurii-ssv: I think a better place to move this TODO note to be in the `Options struct` where the respective parameters were added (so we can later on turn them into config-based ones if we need/want)

iurii-ssv (commented):
Besides minor nits from above, the only thing we need to take care of is to account for the `ProposerDelay` value (as per those [estimates I've added](https://github.com/ssvlabs/ssv/pull/2631#pullrequestreview-3597768085)).

The `GoClient` doesn't know anything about `ProposerDelay` though, so I guess we need to calculate the resulting `proposalSoftTimeout` and `proposalHardTimeout` values and pass those to `GoClient` via `Options struct`. 

That also makes me think the default `DefaultProposalSoftTimeout` and `DefaultProposalHardTimeout` values should be defined outside of `package goclient` (while the `GoClient` should validate the `proposalSoftTimeout` and `proposalHardTimeout` values are not set to 0, making these non-optional).

Also note, we probably want to ensure/validate `ProposerDelay` is not set to be too high that it makes `proposalSoftTimeout` and `proposalHardTimeout` values are too small (even though we already do some validation on what `ProposerDelay` is allowed to be ... that's probably not enough).

beacon/goclient/proposer.go
iurii-ssv: I don't have a strong opinion on this, guess it does make it a bit more explicit :)

beacon/goclient/goclient.go
vyzo: agreed; done.

vyzo:
I agree, but i am not sure we need to solve it in this pr; we should probably make a follow up issue to get this right.

iurii-ssv:
> I agree, but i am not sure we need to solve it in this pr; we should probably make a follow up issue to get this right.

Yup, we can do a follow-up PR for that, just seems kinda logical to me to group these "features" together (even though nothing really makes it "necessary" to do so).

vyzo:
lets get some logs first and then we can decide.

beacon/goclient/options.go
```diff
@@ -16,4 +16,9 @@ type Options struct {
 
 	CommonTimeout time.Duration // Optional.
 	LongTimeout   time.Duration // Optional.
+
+	// TODO these should be configurable by the operator
+	ProposalSoftTimeout time.Duration // Optional
+	ProposalHardTimeout time.Duration // Optional
+	ProposerDelay       time.Duration // Optional
```
y0sher: I'm pretty sure proposerDelay is already configurable just not passed into goclient

liorrutenberg:
Hey, nice work on this! I've been thinking about how this interacts with proposer timing games.

  Question about the tradeoff:

  I'm trying to understand the expected gain from the collection window vs. the cost of reduced builder time. With the ~1.6s soft timeout, we're effectively reducing the proposerDelay budget available for block building.

  For typical setups (shared MEV-boost or same relays across MEV-boosts), I'd expect all beacon clients to return essentially the same block from the same winning relay bid. In that case, time seems like the primary value driver rather than selection between near-identical proposals.

  An alternative approach:

  What about differentiating based on what comes back first?
  - First response is MEV (blinded): Take it immediately, maximize builder time. The proposals from other clients will likely be the same anyway.
  - First response is vanilla: Wait briefly (soft timeout) in case another client returns an MEV block - vanilla first might indicate one client's MEV-boost had issues.

  This way, we get the best of both worlds - maximum builder time for MEV while still having a safety net to catch MEV blocks when the first response happens to be vanilla.

liorrutenberg:
@andrew-blox also suggested checking another optimization: request vanilla blocks in parallel while waiting for MEV. That way if MEV fails or times out, we already have vanilla ready without the extra round-trip latency.

vyzo:
I like the idea of taking MEV block immediately if it arrives within the soft timeout.  I will add it tomorrow.

The requests are all parallel, so we dont have to do anything more.

Roy-blox:
@vyzo I looked through the changes on the stage environment and it seems we manage to propose and submit most proposals with MEV but we seem to fail quite a high percentage also, I see the proposals suffer some sort of 3 second "hang" on submittions, can't put my finger on what exactly the reason for this.
I also looked at the `waited out proposer delay` and saw delay was 0ms on every proposal, not sure if it's an issue but it seems we might be missing something

vyzo:
ok, these are findings we need to address.

beacon/goclient/options.go
iurii-ssv: Yup, we should pass `ProposerDelay` here, we should pass the final `ProposalSoftTimeout` and `ProposalHardTimeout` values that **already** include it (account for it)

vyzo: ok, removing this ProposerDelay.

vyzo: done

beacon/goclient/proposer.go
```diff
@@ -171,38 +185,127 @@ func (gc *GoClient) getProposalParallel(
...
+				bestScore = proposalScore
+				bestClient = res.client
+			}
+
+			if res.proposal.Blinded {
+				// We immediately return as an optimization, under the assumption
+				// that this is a MEV block; it is a reasonable assumption to make in
+				// the usual operating environment.
+				// Note: We may want to add an operator option to disable this behavior
+				// in the future.
```
iurii-ssv, line 254: Nit: I would maybe describe this as something like:
```
Returning as soon as we fetch at least 1 MEV block is good enough, see https://github.com/ssvlabs/ssv/pull/2631#issuecomment-3678879204 for details.
```

vyzo: ok, lets keep pointers to the reasoning

vyzo: done, https://github.com/ssvlabs/ssv/pull/2631/commits/c41a108bd24987598254d5bfd5440abf2ef98e95

beacon/goclient/goclient.go
```diff
@@ -180,6 +192,14 @@ func New(ctx context.Context, logger *zap.Logger, opt Options) (*GoClient, error
...
 		longTimeout = DefaultLongTimeout
 	}
+	proposalSoftTimeout := opt.ProposalSoftTimeout
+	if proposalSoftTimeout == 0 {
+		proposalSoftTimeout = DefaultProposalSoftTimeout
+	}
+	proposalHardTimeout := opt.ProposalHardTimeout
+	if proposalHardTimeout == 0 {
+		proposalHardTimeout = DefaultProposalHardTimeout
+	}
```
iurii-ssv: I think setting defaults here is no longer needed if we are doing it also in `NewOptions`, having a separate place where defaults apply also adds unnecessary complexity (for somebody trying to figure out what the end value is gonna be)

beacon/goclient/goclient.go
```diff
@@ -38,6 +38,10 @@ const (
 	DefaultCommonTimeout = time.Second * 5  // For dialing and most requests.
 	DefaultLongTimeout   = time.Second * 60 // For long requests.
 
+	// Proposal timeouts
+	DefaultProposalSoftTimeout = time.Millisecond * 1600
+	DefaultProposalHardTimeout = time.Millisecond * 2600
```
iurii-ssv: What do you think of making these consts local to the place where these are used ? I don't see much value in putting these here because even if these are used they are also adjusted/offset by ProposerDelay value (which means just looking here creates a wrong impressing that these might be the final values)

beacon/goclient/options.go
```diff
@@ -16,4 +17,34 @@ type Options struct {
 
 	CommonTimeout time.Duration // Optional.
 	LongTimeout   time.Duration // Optional.
+
+	ProposalSoftTimeout time.Duration `yaml:"ProposalSoftTimeout"`
+	ProposalHardTimeout time.Duration `yaml:"ProposalHardTimeout"`
```
iurii-ssv: To match other options, we probably should:
- provide env-variables for these
- add basic description
- default value I wouldn't add since we already have those in the code

beacon/goclient/options.go
```diff
@@ -16,4 +17,34 @@ type Options struct {
...
+}
+
+func NewOptions(base Options, proposerDelay time.Duration) (Options, error) {
+	options := base
+	if options.ProposalSoftTimeout == 0 {
+		options.ProposalSoftTimeout = DefaultProposalSoftTimeout
+	}
```
iurii-ssv: Nit: would be cleaner to have a newline after the `options := base` statement

beacon/goclient/goclient.go
vyzo: definitely, we should remove. I left it for now to avoid breaking tests.

vyzo: definitely, we should remove. I left it for now to avoid breaking tests.

vyzo: i'll add a comment and remove once we are confident nothing breaks.

beacon/goclient/goclient.go
vyzo: i think it is better to have them closer to the source, but no particulalry strong opinions.

beacon/goclient/options.go
vyzo: indeed, i will add.

beacon/goclient/goclient.go
vyzo: added comment. I don't think a human has to do anything about this, we can get the LLM to do it.

beacon/goclient/options.go
vyzo: done.l

beacon/goclient/options.go
vyzo: done.

vyzo:
I have marked the pr as ready for review.

There are two pieces of work that remain to be done, but these are best done by LLMs and i think they should be in a separate part, as i don't like to mix human written code with LLM slop.

beacon/goclient/goclient.go
iurii-ssv: I don't really see a reason for leaving these in, we still kinda need to do deploy & do the final testing of the PR we end up with (to ensure that nothing broke)

vyzo: alright, let's remove.

beacon/goclient/goclient.go
iurii-ssv: Lets move these also then ?

beacon/goclient/goclient.go
vyzo: done.

beacon/goclient/goclient.go
vyzo: ok, where do you want to put them?

iurii-ssv: The only place where these are used now - `func NewOptions`

iurii-ssv (dismissed):
The functionality looks good to me, there are just a couple of minor discussions left to address

beacon/goclient/goclient.go
vyzo: sounds good.

vyzo: sounds good.

vyzo: moved the other two default timeout constants as well, for consistency. I also fixed the test.

iurii-ssv (dismissed):

beacon/goclient/options.go
```diff
@@ -14,6 +25,45 @@ type Options struct {
...
+		options.ProposalHardTimeout = DefaultProposalHardTimeout
+	}
+
+	if proposerDelay > 0 {
+		options.ProposalSoftTimeout -= proposerDelay
+		options.ProposalHardTimeout -= proposerDelay
+	}
+
+	if options.ProposalSoftTimeout < 0 {
+		return Options{}, fmt.Errorf("invalid proposal soft timeout: %s", options.ProposalSoftTimeout)
```
liorrutenberg: ⚠️ Validation catches negative timeouts but does not warn when timing budget is tight. If `proposerDelay + HardTimeout > ~1750ms`, round 1 is at risk.

Consider adding a log warning when the effective hard end exceeds the round 1 deadline (~1750ms from slot start).

Also, the error message showing negative duration is confusing. Suggest:
```go
fmt.Errorf("proposerDelay (%s) exceeds soft timeout budget (%s)", proposerDelay, DefaultProposalSoftTimeout)
```

beacon/goclient/options.go
```diff
@@ -1,11 +1,22 @@
...
 )
 
+const (
+	// Client timeouts.
+	DefaultCommonTimeout = time.Second * 5  // For dialing and most requests.
+	DefaultLongTimeout   = time.Second * 60 // For long requests.
+
+	// Proposal timeouts
+	DefaultProposalSoftTimeout = time.Millisecond * 1600
+	DefaultProposalHardTimeout = time.Millisecond * 2600
```
liorrutenberg: ❓ Hard timeout of 2600ms seems generous. If we need it, we have already exceeded round 1 budget (~2s) and a round change is likely. Is this intentional as a "last chance" fallback, or should it be tighter (~2000ms) to fail faster?

liorrutenberg:
unit tests are marked as remaining work in the PR description. For timing-critical proposal code, I'd recommend covering these cases before merge:

  1. **Scoring selection** - Multiple proposals received → highest score selected
  2. **Early MEV exit** - First blinded proposal → immediate return triggered
  3. **Soft timeout fallback** - No proposals during soft timeout → waits for first valid
  4. **Hard timeout error** - All clients fail/timeout → returns error with joined errors
  5. **ProposerDelay adjustment** - Timeouts correctly reduced by ProposerDelay value
  6. **Validation edge cases** - ProposerDelay too large → returns descriptive error

andrew-blox:
Just looking at this diagram, small suggestion, basically everything runs then we check to see if fee recipient is good, why not just add fee recipient validation to the score, cause like lets say we get 2 MEV both better than vanila, then the best one has a better score but a bad fee recipient, shame we didnt go with the MEV that was more profitable than vanila and with a correct fee recipient @vyzo

vyzo:
i am not convinced that complicating the score logic is a good idea. Is there actual data to support this or just a guess?

if the latter, i say no, lets do what we know generally works.

beacon/goclient/options.go
vyzo: ok, reasonable.  we can reduce to 2s.

vyzo:
that last comment in response to @andrew-blox comment about recipient fee validation.

beacon/goclient/options.go
MatheusFranco99: I agree the hard shouldn't go above the 2s. Even because if the operator is the leader now (1st round) and doesn't send anything on it, its block (only obtained after the 2s) won't have any importance for the next rounds.

Btw, this is only done for the 1st round leader, right? If for others, we should guarantee that this is at least parallel to the QBFT thread. Otherwise, a non-leader operator would be delaying the protocol execution for no reason.

MatheusFranco99: Also, in case this is a parallel thread to the QBFT one, I'm wondering if we even need the hard timeout. I mean, could we just wait indefinitely until the duty finishes?

MatheusFranco99: Another point of view is that, for the operators that are not the 1st round leader, you can actually wait quite a decent amount of time to select the best block, delivering it only once the QBFT thread asks for it. For example, the 2nd round starts and you're the leader. What do you think?

beacon/goclient/options.go
```diff
@@ -1,11 +1,22 @@
...
 	"github.com/ssvlabs/ssv/networkconfig"
 )
 
+const (
+	// Client timeouts.
+	DefaultCommonTimeout = time.Second * 5  // For dialing and most requests.
+	DefaultLongTimeout   = time.Second * 60 // For long requests.
+
+	// Proposal timeouts
+	DefaultProposalSoftTimeout = time.Millisecond * 1600
```
MatheusFranco99: I wonder if this isn't too much. Maybe I just don't know enough about the current timing games that happen in the proposer duty.

But, for example, if the current response time is pretty much a network delay < 200ms, this may be too long and may impact the duty's success.

It would be good to have some real data about how long we usually wait for, how long it takes for other proposals to arrive, and until how long we can safely wait to start the protocol.

beacon/goclient/proposer.go
```diff
@@ -171,38 +193,129 @@ func (gc *GoClient) getProposalParallel(
...
 	return nil, fmt.Errorf("all %d clients failed to get proposal for slot %d, encountered errors: %w", len(gc.clients), slot, errs)
 }
 
+// scoreProposal computes a score for a beacon proposal.
+// see https://github.com/attestantio/vouch/blob/master/strategies/beaconblockproposal/best/score.go as well
+func (gc *GoClient) scoreProposal(
+	proposal *api.VersionedProposal,
+) float64 {
+	score, _ := new(big.Int).Add(proposal.ConsensusValue, proposal.ExecutionValue).Float64()
+	return score
```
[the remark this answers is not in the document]
iurii-ssv: I believe this precision loss here would be negligible for our purposes, so there is no need to worry about it

beacon/goclient/options.go
y0sher: 600ms-1.3s is very typical.

y0sher: so we gave some buffer

y0sher:
another PR with the same branch was merged 🙏

PR #2634, opened 2025-12-24
Enrich proposer logs with duty ID

y0sher:
Retrieve duty ID from trace context and attach it to the proposer logger, passing the enriched logger into getProposalParallel. Add DutyIDFromContext and store the duty ID in traces.Context for later retrieval.


vyzo (approved):
looks good, and a good idea.

The only question is whether the duty id is actually in the context, we'll find out soon.

Issue #2632, opened 2025-12-22
p2p: enhanced message-validation for duplicate nodes


iurii-ssv:
Supersedes https://github.com/ssvlabs/ssv/issues/2616, upon further investigation, the issue we observe probably happens something like this:

- say we have node1, node2, node3, node4
- and then we duplicate node1 (node1A, node1B - each of these will reveal itself under different/unique peerID)
- node2 receives & accepts a message with the same msgID in 2 different forms - from node1A and from node1B (because these have different peerIDs), and it broadcasts it to node3 twice
- node3 sees 2 messages coming from the same peerID(node2) having the same msgID and it rejects the 2nd message penalizing the innocent node2

and if we can't resolve it via peerID-authentication - then I think the best we can do (and it will also work correctly in case of duplicate nodes) in terms of rules for p2p messages is something like this:
1) if we receive a message with new ID we **accept** it and broadcast it once (+ record the peer we received it from)
2) if we receive a message with the same ID from the same peer - **reject** it
3) if we receive a message with the same ID from different peer - **ignore** it (this is where current rule set differs in that it **accepts** & propagates the message)

For the above rules to work, message ID needs to be unique (seems like the msgID we currently use is not unique on its own ... it's = Operator+DutyType), so we need to rearrange our code to calculate the "full message ID" as:
- (for consensus message) messageID=msgID+slot+round
- (for partial-sig message) messageID=msgID+slot

Note, malicious actor would still be able to spam messages using new peerIDs if he can generate/connect those fast enough (just as he can today) ... but that's probably fine.

iurii-ssv:
Resolved in https://github.com/ssvlabs/ssv/pull/2677

PR #2635, opened 2025-12-24
p2p: rework duplicate message rejections

iurii-ssv:
Resolves https://github.com/ssvlabs/ssv/issues/2632

This PR changes the validation-rules treatment of duplicate messages. A duplicate message is a message that should not be applied more than 1 time from QBFT-logic perspective (and it is "additionally" enforced via message-validation rules). SSV node wants to detect duplicate messages in order to:
- reduce the amount of unnecessary traffic in p2p network as much as possible
- actively punish peers (through libp2p reputation system) who don't follow protocol rules

The behavior this PR is aiming to change:
- previously, a duplicate message will not be detected correctly if it comes from a different peer (a peer with different peerID) ... and so [SSV node will accept such message and broadcast it to other peers getting punished for it](https://github.com/ssvlabs/ssv/issues/2632)
- with this PR, a duplicate message is detected always, and then depending on whether it comes from the peer who already sent that/similar message to us, or from another/different peer - it is rejected/ignored respectively (punishing only those peers who "knowingly" broadcast duplicate messages, the "knowingly" part comes from the libp2p property of every node in p2p pubsub being fully responsible for validating messages received from other peers it chooses to re-broadcasts ... so it must validate the message isn't a duplicate before deciding to re-broadcast it)

In order to be able to tell whether a duplicate p2p message is coming from new/different peer (or from the same peer we've received that/similar message in the past) we need to store & update the `message->peer` mapping ... that is however somewhat expensive (and doesn't fit well with the current minimalistic/optimized implementation we have for message-validation). Instead, in this PR, we keep track of "seen validation-rule violations":
- 1 violation is always allowed (if some peer has already sent some message, and then he sends the same/similar message 1 more time - he will not be punished by message rejection that 1 time, the receiving SSV node will simply ignore such a message for the 1st time)
- 2nd+ violations are spotted via a check against "seen validation-rule violations" (the new structure we now track per peer, on a need-to basis), and if the violating message is coming from the peer who already had been observed to have this/similar kind of violation in the past (for this Operator+DutyType+slot) we reject the message (otherwise we ignore that violating message since it would be the 1st violation of this type for this peer)

Additional considerations:
- previously `messageValidator.state` and `messageValidator.validationLockCache` maps were keyed by `spectypes.MessageID+peerID` ... that doesn't look correct to me because the "state" doesn't/shouldn't know/depend on which peer a message is received from
- in this PR `messageValidator.state` and `messageValidator.validationLockCache` maps are keyed by `spectypes.MessageID` (which is basically = Operator+DutyType) only, meaning messages with the same `spectypes.MessageID` targeting **different** slots will acquire the same validation-lock and will be validated sequentially ... I think we could potentially optimize that further making validation for messages targeting different slots run concurrently, but it would be hard to implement this correctly on top of existing validation code (and this is out of scope of this PR anyway)
- `SignerState.SeenMsgTypes` keeps track of both: messages related to QBFT-instance as well as pre-consensus phase messages, and upon round-change `SignerState.Reset` call fully resets not only QBFT-instance-related stuff but also pre-consensus phase messages as well. This means we are "allowing the receival of pre-consensus phase messages again once round-change happened" which doesn't make much sense. But it's not really abusable either, so there is no need to address that - just wanted to document that behavior
- we also have a bunch of errors this PR doesn't touch that we classify as ignore (and never as reject), we might want to penalize peers sending lots of duplicate messages that end up as those errors as well as a potential future improvement


message/validation/signer_state.go
```diff
@@ -37,5 +49,25 @@ func (s *SignerState) Reset(slot phase0.Slot, round specqbft.Round) {
...
+		s.SeenViolations = make(map[peer.ID]map[Error]struct{})
+	}
+	if s.SeenViolations[receivedFrom] == nil {
+		s.SeenViolations[receivedFrom] = make(map[Error]struct{})
+	}
+	_, seenViolation := s.SeenViolations[receivedFrom][ignoreErr]
+	if seenViolation {
+		return rejectErr
+	}
+	s.SeenViolations[receivedFrom][ErrTooManyPartialSigMessage] = struct{}{}
```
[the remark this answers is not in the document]
iurii-ssv: Good catch, fixed in https://github.com/ssvlabs/ssv/pull/2635/changes/65bea82dc61ddc958a2b476a6d4df9ffb96ee985

message/validation/consensus_validation.go
```diff
@@ -11,7 +11,6 @@ import (
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/libp2p/go-libp2p/core/peer"
-
```
vyzo, line 14: keep this line please, it is good practice to separate our own libraries from external deps.  makes audit a tad easier.

message/validation/consensus_validation.go
```diff
@@ -342,14 +343,14 @@ func (mv *messageValidator) processSignerState(
 
 	signerCount := len(signedSSVMessage.OperatorIDs)
 	if signerCount > 1 {
+		if signerState.SeenSigners == nil {
+			signerState.SeenSigners = make(map[SignersBitMask]struct{}) // lazy init on demand to reduce mem consumption
```
vyzo, line 348: do we need to garbage collect cold entries? Maybe also limit size?

I think it might be a good idea, i am concerned it might open a dos attack vector by making us use potentially unlimited memory in this map.

message/validation/partial_validation.go
```diff
@@ -9,7 +9,6 @@ import (
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/libp2p/go-libp2p/core/peer"
-
```
vyzo, line 12: keep line

vyzo (commented):
first pass, approach is pretty reasonable, i will do another pass.

Left a comment.

message/validation/consensus_validation.go
iurii-ssv: For `SeenSigners` and `SeenViolations` we basically drop the reference(s) to those in `SignerState.Reset` letting the Golang GC to collect those maps we used in the past.

So it is something to consider, but from what I see in the code it looks like it's working correctly.

vyzo: ok, this limits the attack vector considerably.  I think it should be ok too, i just get uneasy when i see maps reachable from the network.

message/validation/consensus_validation.go
iurii-ssv: This is actually still consistent with the current version of our `make format` command:
- if you run it after removing all the newlines,
- but once those newlines are already there it **doesn't** remove those

so we can end up with 2 valid but different-looking imports sections (both of which are valid from `make format` & `make lint` perspective)

Note also that currently we classify `ssv-spec` as any other 3rd-party repo (we don't bundle `ssv-spec` and `ssv` packages together in under imports section).

I've pushed a commit to revert these imports-affecting changes for now, but in general I think we shouldn't worry much about it (or maybe use something like `gci` to enforce certain rules) - https://github.com/ssvlabs/ssv/pull/2635/changes/e3f2cfd3e45961d03cb03ebcafacf7940ed7d915

MatheusFranco99:
> 1 violation is always allowed (if some peer has already sent some message, and then he sends the same/similar message 1 more time - he will not be punished by message rejection that 1 time, the receiving SSV node will simply ignore such a message for the 1st time)

Hmm, I don't see a clear reason for allowing this. I think that's already what we want to avoid: the same peer sending the same logical message twice ("duplicated"). I think it could be rejected already on its first duplicated.

---

> - previously messageValidator.state and messageValidator.validationLockCache maps were keyed by spectypes.MessageID+peerID ... that doesn't look correct to me because the "state" doesn't/shouldn't know/depend on which peer a message is received from
> - in this PR messageValidator.state and messageValidator.validationLockCache maps are keyed by spectypes.MessageID (which is basically = Operator+DutyType) only, meaning messages with the same spectypes.MessageID targeting different slots will acquire the same validation-lock and will be validated sequentially ... I think we could potentially optimize that further making validation for messages targeting different slots run concurrently, but it would be hard to implement this correctly on top of existing validation code (and this is out of scope of this PR anyway)

Btw, this is a critical change. Per-peer state (btw maybe it's easy to reason about it as "network view") was introduced so that we could penalize peers more accurately, as it's only responsible by its own state.
Still, at least, I think having a common state could only hurt the logical rules (as semantic rules don't matter here). For example, these come to mind:

*(from knowledge-base)*

**QBFT Logic**
| Verification | Error | Classification | Description |
| ------------|------|--------------|-------------|
| Double Proposal                         | ErrDuplicatedMessage                   | Reject         | Signer already sent a proposal for round.                              |
| = for Prepare, Commit, and RC                          |     -               | -  | -  |
| Already advanced round                  | ErrRoundAlreadyAdvanced                | Ignore         | Signer is already in a future round.                                   |

**PSig Logic**
| Invalid signature type count | ErrInvalidPartialSignatureTypeCount                | Reject         | It's allow only:<br> 1 PostConsensusPartialSig, for Committee duty,<br> 1 RandaoPartialSig and 1 PostConsensusPartialSig for Proposer,<br> 1 SelectionProofPartialSig and 1 PostConsensusPartialSig for Aggregator,<br> 1 SelectionProofPartialSig and 1 PostConsensusPartialSig for Sync committee contribution,<br> 1 ValidatorRegistrationPartialSig for Validator Registration,<br> 1 VoluntaryExitPartialSig for Voluntary Exit. |
| ------------|------|--------------|-------------|


**Duty Logic**

| Already advanced slot     | ErrSlotAlreadyAdvanced        | Ignore         | (Non-committee roles) Signer already advanced to later slot.                                                                                                                                         |
| ------------|------|--------------|-------------|
| Too many duties per epoch | ErrTooManyDutiesPerEpoch      | Ignore         | If role is either aggregator, voluntary exit and validator registration,<br> it's allowed 2 duties per epoch. Else if committee,<br> 2*V (if no validator is doing sync committee).<br> Else accept. |


For the duplicated cases, we already already solving it with this change. For the `*AlreadyAdvanced` and `ErrTooManyDutiesPerEpoch`, I don't think an attacker can do much (e.g. trying to populate slots for a small committee so that it wouldn't be able to send more msgs)  as the msgs also need to be correctly signed, so the attacker would need to be in the committee itself.

I think it may be good for @GalRogozinski to take a look here as well

iurii-ssv:
@MatheusFranco99 thanks for the review,

> Hmm, I don't see a clear reason for allowing this. I think that's already what we want to avoid: the same peer sending the same logical message twice ("duplicated"). I think it could be rejected already on its first duplicated.

As I mentioned in PR description, it would be hard/expensive to implement the necessary logic to track & correctly punish the very 1st violation. So instead, we just "record" the violation 1st time it happens and only "punish" (by message reject) if it happens again.

Does it make sense ? I agree that it would be better to not go for this "shortcut", but it's a good practical trade-off to take, I think (it's not like it can be abused by the attacker).

> Still, at least, I think having a common state could only hurt the logical rules (as semantic rules don't matter here)

I'm not sure I understand what this ^ part refers to, could you expand ? Are the changes in this PR fine to do or not ?

MatheusFranco99:
> it would be hard/expensive to implement the necessary logic to track & correctly punish the very 1st violation. So instead, we just "record" the violation 1st time it happens and only "punish" (by message reject) if it happens again.

Hmm, sorry I'm a bit lost here. When you say `we just "record" the violation 1st time it happens`, isn't it already `the necessary logic to track` it being used? If not, how do you detect it? Or is this detection by some other "approximation" mechanism?

I think I didn't completely understood it by looking at the code, but is the this duplication counter (for a certain peer) per logical message step (like `committee:X, duty type: D, slot: S, round: R, prepare`), or is it a counter for all messages?

> I'm not sure I understand what this ^ part refers to, could you expand ? Are the changes in this PR fine to do or not ?

This is just a side-thought/concern, just adding it here for others to also think about it and confirm I didn't miss something

iurii-ssv:
> Hmm, sorry I'm a bit lost here. When you say `we just "record" the violation 1st time it happens`, isn't it already `the necessary logic to track` it being used? If not, how do you detect it? Or is this detection by some other "approximation" mechanism?

Yeah, sorry, it's a bit confusing, basically
- in my terminology, violation is a validation-logic error **before** we classify that error as `ignore` or `reject`
- so we need to figure out whether the violation is intentional (same peer knowingly sending us the bad message 2+ times, in which case we want to `reject` the 2nd, 3rd, etc. messages) or unintentional (due to the way libp2p works, any peer can unknowingly send you a violating message 1 time, so we need to `ignore` it as long as it is their 1st time)
- prior violations are tracked via `SeenViolations` structure in this PR
- also note, this PR targets (properly classifies between `ignore` and `reject`, instead of just always doing `ignore`) only subset of possible duplicate messages SSV node might receive (specifically those observed to be sent when running a duplicate SSV node during my testing) ... ideally we'd want to treat all errors with the same approach, but maybe it's just not necessary in practice

GalRogozinski (changes requested):
I am rejecting not because the change isn't good, I actually don't know if it is good.
It is simply something the spec team need to schedule time for correct analysis.

The rejection is just to hold off the merge for now

cc @Tom-ssvlabs

message/validation/consensus_validation.go
```diff
@@ -179,24 +176,24 @@ func (mv *messageValidator) validateQBFTLogic(
...
+			return e
 		}
 	}
 
 	msgSlot := phase0.Slot(consensusMessage.Height)
 	for _, signer := range signedSSVMessage.OperatorIDs {
-		signerStateBySlot := state.Signer(committeeInfo.signerIndex(signer))
-		signerState := signerStateBySlot.GetSignerState(msgSlot)
+		signerState := state.OperatorState(committeeInfo.signerIndex(signer)).GetSignerStateForSlot(msgSlot)
 		if signerState == nil {
```
oleg-ssvlabs, line 197: I think it would worth adding a comment explaining what `signerState` equals to `nil` mean. I assume it happens when it validates the first message from this particular _operator_ for this particular _slot_ ?

message/validation/consensus_validation.go
```diff
@@ -207,22 +204,26 @@ func (mv *messageValidator) validateQBFTLogic(
...
-				return err
+				e := ErrRoundAlreadyAdvanced
+				e.want = signerState.Round
+				e.got = consensusMessage.Round
+				return e
 			}
 
 			if consensusMessage.Round == signerState.Round {
 				// Rule: Peer must not send two proposals with different data
 				if len(signedSSVMessage.FullData) != 0 && signerState.HashedProposalData != nil {
```
oleg-ssvlabs, line 215: side question:
I don't think there is validation somewhere for the length of the `FullData` field? Shouldn't it always be non empty? And if it is empty - validation should return an error?

message/validation/consensus_validation.go
```diff
@@ -207,22 +204,26 @@ func (mv *messageValidator) validateQBFTLogic(
...
-					if *signerState.HashedProposalData != sha256.Sum256(signedSSVMessage.FullData) {
-						return ErrDifferentProposalData
+					msgHashedProposalData := sha256.Sum256(signedSSVMessage.FullData)
```
oleg-ssvlabs, line 216: [here](https://github.com/ssvlabs/ssv/blob/p2p-rework-duplicate-message-rejections/message/validation/consensus_validation.go#L119) there is a validation which ensures `consensusMessage.Root` equals to hash FullData (same SHA256 hashing under the `specabft.HashDataRoot()` method). This validation runs before this code.

This basically means you don't need to hash `FullData` here for comparison, you can do this instead:

```Go
*signerState.HashedProposalData != consensusMessage.Root
```

Hashing is not very expensive, but `FullData` max size is 8Mb (considering `ssz-max` tag), so this minor optimization might introduce some performance improvements.

oleg-ssvlabs: You can also change [here](https://github.com/ssvlabs/ssv/blob/p2p-rework-duplicate-message-rejections/message/validation/consensus_validation.go#L341) from:
```Go
fullDataHash := sha256.Sum256(signedSSVMessage.FullData)
signerState.HashedProposalData = &fullDataHash
```
to:
```Go
 signerState.HashedProposalData = &consensusMessage.Root 
```

message/validation/signer_state.go
```diff
@@ -3,25 +3,40 @@ package validation
...
 	// Max possible map size for committee sizes:
 	//  4 (f=1): C(4,3)+C(4,4)=5
 	//  7 (f=2): C(7,5)+C(7,6)+C(7,7)=29
 	// 10 (f=3): C(10,7)+C(10,8)+C(10,9)+C(10,10)=176
 	// 13 (f=4): C(13,9)+C(13,10)+C(13,11)+C(13,12)+C(13,13)=1093
 	SeenSigners map[SignersBitMask]struct{}
+
+	// SeenViolations keeps track of validation violations by peers we get messages for this signer from detected
+	// so far (mapping peer-id -> error-text).
+	SeenViolations map[peer.ID]map[string]struct{}
```
oleg-ssvlabs, line 39: I don’t think there’s a need to export this field, since it’s only used within the `SignerState` object. 
Actually, the entire `SignerState` could be private as well with all the fields?

oleg-ssvlabs (commented):
gj!
Left couple of comments

message/validation/consensus_validation.go
iurii-ssv: TBH I didn't look into this cause this behavior isn't changing in this PR ...

message/validation/consensus_validation.go
iurii-ssv: I'm pretty sure it can be empty for some message types, but I haven't looked into it deeper.

message/validation/signer_state.go
iurii-ssv: Probably, I just went for the "minimal viable change", I think if we want to refactor it it's better to do in a separate PR on top of this one later on (to keep the changes in this one limited in size and easy to review ... I kinda already had to do some renaming I wasn't planning for in this PR that touches lots of files - to figure out how it works)

oleg-ssvlabs: sure, makes sense @iurii-ssv

GalRogozinski:
I am glad that we are removing `peerID`

Regarding letting the first violation through.
@MatheusFranco99 is right, from a protocol perspective it is not correct (or not needed)

Running two nodes is byzantine behavior.
If one operator become byzantine and we ignore it, and then a little later another operator becomes byzantine (even by mistake) then our guarantees are REKT.

iurii-ssv:
> Regarding letting the first violation through. @MatheusFranco99 is right, from a protocol perspective it is not correct (or not needed)
> 
> Running two nodes is byzantine behavior. If one operator become byzantine and we ignore it, and then a little later another operator becomes byzantine (even by mistake) then our guarantees are REKT.

@GalRogozinski while I agree that it would be nice to not leave even 1 violation unpunished ... I don't see why it would matter in practice, in the worst-case scenario we'll get twice as many the number of messages from malicious peer (and end up ignoring half of them). But SSV node should be able to easily handle that (maybe if it was 10x we should be worried, 2x doesn't seem like a big deal). You could think of it as "optimistic" solution.

We can implement it the way you are asking, it just seems like an unnecessary effort and performance hit.

GalRogozinski:
@iurii-ssv 
I wasn't talking about DoS in the last comment.
I was talking about one node becomes byzantine (runs two instances) undetected and then later another node doing this.
If we have two nodes that do this in a committee of 4 then we are in trouble... I think this can cause slashing.

message/validation/consensus_validation.go
iurii-ssv: Will be addressed in https://github.com/ssvlabs/ssv/pull/2673 - seems like a good idea to apply this optimization

iurii-ssv:
Closing in favor of https://github.com/ssvlabs/ssv/pull/2677

PR #2637, opened 2026-01-04
Feat/spec proposer round robin for Boole fork

julienh-ssv:
This PR updates the proposer selection function to match the latest spec for Boole Fork. Since the function was duplicated in our codebase, I also refactored the code to use a single source of truth implementation.

1. [Spec PR](https://github.com/ssvlabs/ssv-spec/pull/591)
2. [SIP](https://github.com/ssvlabs/SIPs/pull/74)


protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +25,20 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
 	index := (firstRoundIndex + uint64(round) - uint64(specqbft.FirstRound)) % uint64(len(state.CommitteeMember.Committee))
 	return state.CommitteeMember.Committee[index].OperatorID
 }
+
+// RoundRobinProposer_Boole returns the proposer for the round.
+// Each new height starts with the first proposer and increments by 1 with each following round.
+// Each new height has a different first round proposer which is +1 from the previous height.
+// Also, the current Ethereum epoch is taken into account to introduce variability through epochs
+// (mostly for committees with 4 operators, as 32%4 = 0 as the epochs would "repeat" otherwise).
```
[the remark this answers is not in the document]
julienh-ssv: Implementation taken from spec.

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +25,20 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
+// First height starts with index 0
+func RoundRobinProposer_Boole(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
+	firstRoundIndex := 0
+	if state.Height != specqbft.FirstHeight {
+		firstRoundIndex += int(state.Height) % len(state.CommitteeMember.Committee)
+	}
+	ethEpoch := int(state.Height) / 32
+
+	index := (firstRoundIndex + int(round) - int(specqbft.FirstRound) + ethEpoch) % len(state.CommitteeMember.Committee)
+	return state.CommitteeMember.Committee[index].OperatorID
```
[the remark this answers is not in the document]
julienh-ssv: Yes can fix

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +25,20 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
+// Each new height has a different first round proposer which is +1 from the previous height.
+// Also, the current Ethereum epoch is taken into account to introduce variability through epochs
+// (mostly for committees with 4 operators, as 32%4 = 0 as the epochs would "repeat" otherwise).
+// First height starts with index 0
+func RoundRobinProposer_Boole(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
+	firstRoundIndex := uint64(0)
+	if state.Height != specqbft.FirstHeight {
+		firstRoundIndex += uint64(state.Height) % uint64(len(state.CommitteeMember.Committee))
+	}
+	ethEpoch := uint64(state.Height) / 32
```
y0sher: 32 should be derived from cfg (slots per epoch) to support devnets and special networks.

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +25,20 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
 	return state.CommitteeMember.Committee[index].OperatorID
 }
+
+// RoundRobinProposer_Boole returns the proposer for the round.
+// Each new height starts with the first proposer and increments by 1 with each following round.
+// Each new height has a different first round proposer which is +1 from the previous height.
+// Also, the current Ethereum epoch is taken into account to introduce variability through epochs
+// (mostly for committees with 4 operators, as 32%4 = 0 as the epochs would "repeat" otherwise).
+// First height starts with index 0
+func RoundRobinProposer_Boole(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
```
y0sher:
```suggestion
func RoundRobinProposerBoole(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
```

networkconfig/network.go
```diff
@@ -30,3 +30,7 @@ func (n Network) StorageName() string {
 func (n Network) GasLimit36Fork() bool {
 	return n.EstimatedCurrentEpoch() >= n.SSV.Forks.GasLimit36
 }
+
+func (n Network) BooleFork() bool {
```
y0sher: it's better to pass in the slot/epoch instead of using current, for example some messages in pre-fork could arrive into fork epoch but should still be treated as prefork as they belong to the slot before.

In some places in our code this won't be relevant, but for this example when calculating the leader I think my point is valid.

y0sher: So we have height when determining leader and it's better to use it than base on time.Now (EstimatedCurrentEpoch)

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -5,6 +5,13 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
+func ForkAwareRoundRobinProposer(state *specqbft.State, round specqbft.Round, booleFork bool) spectypes.OperatorID {
```
y0sher: imo this code should live in controller because it's not qbft related but configuration and network state related.

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: done

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: ok

y0sher:
@julienh-ssv please add description and also link to relevant Spec and SIP PRs.

y0sher (commented):
To be honest, I think the scope of this change should be as limited as we can to the required change. So even though I think the refactor is correct and better looking, I wouldn't want to tie it with the scope of this change, So I'm willing to live with the duplication of the forking logic:
```go
if netCfg.BooleForkAtSlot { 
   // with epoch leader 
} else {
  // no epoch leader 
}
```
It's going to be duplicated similarly around the code for many use cases anyway. but if possible to reduce the duplication in `consensus_validation` by importing the logic from the `qbft` package (and wrapping with the ifs like I mentioned before).

Just to clarify my main goal is to keep the change as minimal as possible while achieving the expected result. 

I'd be up to do a refactor in how we do leaders in a future PR with a clear scope for that specifically.

y0sher (dismissed):

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -1,20 +1,35 @@
...
+}
+
 // RoundRobinProposer returns the proposer for the round.
 // Each new height starts with the first proposer and increments by 1 with each following round.
 // Each new height has a different first round proposer which is +1 from the previous height.
-// First height starts with index 0
-func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
+// First height starts with index 0.
+// Boole fork adds an epoch-derived offset (from network config) to introduce additional variability.
+func RoundRobinProposer(height specqbft.Height, round specqbft.Round, committee []spectypes.OperatorID, netCfg networkConfig) spectypes.OperatorID {
```
vyzo: vary minor nit, but this argument line is an eyesore, you could break it to an argument per line.

vyzo (dismissed):
just a minor formatting nit, looks good.

julienh-ssv:
@vyzo @y0sher fixed the formatting and rebased on top of stage to get the linter fixes

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: I think we usually use very long lines in this project. But sure I can change it

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -19,7 +19,11 @@ type networkConfig interface {
 // Each new height has a different first round proposer which is +1 from the previous height.
 // First height starts with index 0.
 // Boole fork adds an epoch-derived offset (from network config) to introduce additional variability.
-func RoundRobinProposer(height specqbft.Height, round specqbft.Round, committee []spectypes.OperatorID, netCfg networkConfig) spectypes.OperatorID {
+func RoundRobinProposer(
+	height specqbft.Height,
+	round specqbft.Round,
+	committee []spectypes.OperatorID,
+	netCfg networkConfig) spectypes.OperatorID {
```
iurii-ssv: Nit: I think it's a bit easier to read when return-values are on a separate line like this
```go
func RoundRobinProposer(
	height specqbft.Height,
	round specqbft.Round,
	committee []spectypes.OperatorID,
	netCfg networkConfig,
) spectypes.OperatorID {
```

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -1,20 +1,48 @@
...
+}
+
 // RoundRobinProposer returns the proposer for the round.
 // Each new height starts with the first proposer and increments by 1 with each following round.
 // Each new height has a different first round proposer which is +1 from the previous height.
-// First height starts with index 0
-func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
+// First height starts with index 0.
+// Boole fork adds an epoch-derived offset (from network config) to introduce additional variability.
+func RoundRobinProposer(
```
iurii-ssv, line 32: Nit: would be nice to copy over slightly more detailed comment from ssv-spec repo ?
```go
// RoundRobinProposer returns the proposer for the round.
// Each new height starts with the first proposer and increments by 1 with each following round.
// Each new height has a different first round proposer which is +1 from the previous height.
// Also, the current Ethereum epoch is taken into account to introduce variability through epochs
// (mostly for committees with 4 operators, as 32%4 = 0 as the epochs would "repeat" otherwise).
```

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -1,20 +1,48 @@
...
+	height specqbft.Height,
+	round specqbft.Round,
+	committee []spectypes.OperatorID,
+	netCfg networkConfig) spectypes.OperatorID {
+	if !sort.SliceIsSorted(committee, func(i, j int) bool { return committee[i] < committee[j] }) {
+		sorted := make([]spectypes.OperatorID, len(committee))
+		copy(sorted, committee)
+		sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
+		committee = sorted
+	}
```
iurii-ssv, line 44: I'm not sure I understand all the implications of this sorting here, could you explain for me why
- in [spec repo they don't sort](https://github.com/ssvlabs/ssv-spec/pull/591/changes)
- the sorting is happening regardless of `netCfg.BooleForkAtEpoch(epoch)` (assuming the sorting is needed) ... so once some operators update it would technically "fork them immediately", I don't think this is intended ?

iurii-ssv (commented):
Nice changes, just [some questions](https://github.com/ssvlabs/ssv/pull/2637/changes#r2681534105) & minor suggestions from me.

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: Sorting is needed because we removed duplication between the two packages, not because we're forking. So even the old code needs it.

Since the selection algorithm outputs a slice index, (e.g. `1`), when I call the function with the committee `a = [4,5,6,7]` and the committee `b = [5,4,6,7]`, even though both `a` and `b` represent the exact same committee, the leader selection will be different since the slices are in different order: `a[1] != b[1]`.

The spec repo implicitly assumes that the committee is in a deterministic order that does not change between calls (e.g. same `state` -> same order). With duplication removal we actually use the same committee but not the same slice so we need a way to choose a canonical deterministic order.

So now that we use this function in different modules that build the committee slice in any order, we ensure robustness by sorting our own local copy (and the computation cost of sorting is nothing here).

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: Honestly in this case, I really don't see how this comment helps since it's just repeating what the code already says.

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: Sure, I can change it. – I wish we had a better formatter to take care of these issues.

protocol/v2/qbft/round_robin_proposer.go
julienh-ssv: Also it's not just a theoretical guardrail: I had initially deployed without sorting and messages were all rejected due to wrong signer/leader.

julienh-ssv: Although now that we’re discussing it, I’m thinking that if nodes previously relied on a deterministic but unsorted slice order, sorting could affect retro-compatibility. 🤔 To confirm pre-fork behavior remains compatible, I’ll deploy a cluster of old nodes and do some testing.

iurii-ssv: > Although now that we’re discussing it, I’m thinking that if nodes previously relied on a deterministic but unsorted slice order, sorting could affect retro-compatibility.

Yeah, that's what I'm basically calling out here :) I understand that we need a deterministic order between the calls, but we ALSO need this determinism preserved between old/new versions (to not fork by accident).

We probably want to think this part through because just testing it might "work fine by accident/luck" (depending on how it's currently implemented).

julienh-ssv: Today's tests confirm that it's working on two different clusters, each running a mix of stage (old code) and this branch (new code, pre-Boole). 

But I agree that it could just be working by luck. I'll push some changes to restore the previous code in pre-fork mode. This should remove all uncertainty.

iurii-ssv: Yeah, briefly looking at the code involved in setting up that shares-slice it doesn't look like we sort it, so I wouldn't rely on it arriving sorted from the corresponding smart-contract event (seems is where it is coming from).

iurii-ssv (approved):
LGTM!

y0sher (approved):

PR #2638, opened 2026-01-05
Rework Scored Proposal logic

vyzo:
Given our observations from QA testing, it seems our original logic is not robust. This reworks the logic so to properly account for the proposal delay.

The new logic is as follows:
1. If there is a proposer delay, prefetch the best block before waiting.
2. Wait for the proposer delay
3. Fetch the best block.

The soft/hard timeout distinction is also gone, no longer sensible.

### DIFF
--- a/beacon/goclient/proposer.go
+++ b/beacon/goclient/proposer.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/ssvlabs/ssv/observability/log/fields"
 	"github.com/ssvlabs/ssv/observability/traces"
+	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
 )
 
 // ProposerDuties returns proposer duties for the given epoch.
@@ -68,9 +69,9 @@ func (gc *GoClient) GetBeaconBlock(
 	graffiti := [32]byte{}
 	copy(graffiti[:], graffitiBytes[:])
 
-	proposeTime, ok := ctx.Value("ProposeTime").(time.Time)
+	proposeTime, ok := ctx.Value(beacon.ProposeTimeContextKey{}).(time.Time)
 	if !ok {
-		// short circut the prefetch
+		// short circuit the prefetch
 		proposeTime = time.Now()
 	}
 
--- a/beacon/goclient/proposer.go
+++ b/beacon/goclient/proposer.go
@@ -49,34 +49,6 @@ func (gc *GoClient) ProposerDuties(ctx context.Context, epoch phase0.Epoch, vali
 	return resp.Data, nil
 }
 
-// fetchProposal fetches a proposal from a single client and records metrics
-func (gc *GoClient) fetchProposal(
-	ctx context.Context,
-	client Client,
-	slot phase0.Slot,
-	sig phase0.BLSSignature,
-	graffiti [32]byte,
-) (*api.VersionedProposal, error) {
-	reqStart := time.Now()
-	resp, err := client.Proposal(ctx, &api.ProposalOpts{
-		Slot:         slot,
-		RandaoReveal: sig,
-		Graffiti:     graffiti,
-	})
-	recordRequest(ctx, gc.log, "Proposal", client, http.MethodGet, false, time.Since(reqStart), err)
-	if err != nil {
-		return nil, errSingleClient(fmt.Errorf("fetch proposal: %w", err), client.Address(), "Proposal")
-	}
-	if resp == nil {
-		return nil, errSingleClient(fmt.Errorf("proposal response is nil"), client.Address(), "Proposal")
-	}
-	if resp.Data == nil {
-		return nil, errSingleClient(fmt.Errorf("proposal response data is nil"), client.Address(), "Proposal")
-	}
-
-	return resp.Data, nil
-}
-
 // GetBeaconBlock implements ProposerCalls.GetBeaconBlock
 func (gc *GoClient) GetBeaconBlock(
 	ctx context.Context,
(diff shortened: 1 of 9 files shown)


beacon/goclient/proposer.go
```diff
@@ -49,34 +49,6 @@ func (gc *GoClient) ProposerDuties(ctx context.Context, epoch phase0.Epoch, vali
 	return resp.Data, nil
 }
 
-// fetchProposal fetches a proposal from a single client and records metrics
-func (gc *GoClient) fetchProposal(
```
vyzo, line 53: this was moved below for better related function grouping.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -153,31 +153,20 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
-		select {
-		case <-time.After(timeLeft):
-		case <-ctx.Done():
-			return ctx.Err()
-		}
-	}
-
-	waitedOutProposerDelayEvent := fmt.Sprintf("waited out proposer delay of %dms", r.proposerDelay.Milliseconds())
-	logger.Debug(waitedOutProposerDelayEvent)
-	span.AddEvent(waitedOutProposerDelayEvent)
```
vyzo, line 169: if we want this span, it will have to move inside the beacon client.

protocol/v2/ssv/runner/proposer.go
```diff
@@ -153,31 +153,20 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
...
-
-	duty = r.state().CurrentDuty.(*spectypes.ValidatorDuty)
```
vyzo, line 171: some concern about this one, although it think it should be ok.

protocol/v2/blockchain/beacon/context.go
```diff
@@ -0,0 +1,3 @@
+package beacon
+
+type ProposeTimeContextKey struct{}
```
vyzo, line 3: maybe not the best place to put this, feel free to suggest a better package.

beacon/goclient/proposer.go
```diff
@@ -96,23 +69,78 @@ func (gc *GoClient) GetBeaconBlock(
 	graffiti := [32]byte{}
 	copy(graffiti[:], graffitiBytes[:])
 
+	proposeTime, ok := ctx.Value(beacon.ProposeTimeContextKey{}).(time.Time)
```
oleg-ssvlabs, line 72: it's interesting approach.
Have you considered passing `proposeTime` as `GetBeaconBlock` method argument instead? 
Could be something like(to make it optional):

```Go
 GetBeaconBlock(ctx context.Context, slot phase0.Slot, graffiti, randao []byte, opts *ProposalOptions)
```

was it important not to change an interface, hence you decided to pass it via context?

beacon/goclient/proposer.go
```diff
@@ -96,23 +69,78 @@ func (gc *GoClient) GetBeaconBlock(
...
+				zap.Bool("blinded", beaconBlock.Blinded),
+				fields.Slot(slot),
+			)
 		}
-	} else {
-		// For multiple clients, race them in parallel for the fastest response
-		beaconBlock, err = gc.getProposalParallel(ctx, logger, slot, sig, graffiti)
-		if err != nil {
+
+		// wait out the prposer delay
```
oleg-ssvlabs, line 107: typo `prposer` -> `proposer`

beacon/goclient/proposer.go
```diff
@@ -96,23 +69,78 @@ func (gc *GoClient) GetBeaconBlock(
...
+	if err != nil {
+		logger.Warn("failed to fetch final block proposal", zap.Error(err))
+
+		if beaconBlock == nil {
 			return nil, nil, err
 		}
 	}
 
+	if latestBeaconBlock != nil {
+		beaconBlock = latestBeaconBlock
```
oleg-ssvlabs, line 131: Does this mean the previously fetched block is always discarded without comparing the scores?
If so, does that imply that the later a block is fetched, the higher its score would be (or at least that it cannot be lower)?

beacon/goclient/proposer.go
```diff
@@ -152,37 +180,68 @@ func (gc *GoClient) GetBeaconBlock(
...
+		return block, client.Address(), err
+	}
+
+	return gc.fetchProposalParallel(ctx, logger, slot, sig, graffiti)
+}
+
+// fetchProposal fetches a proposal from a single client and records metrics
+func (gc *GoClient) fetchProposal(
+	ctx context.Context,
+	logger *zap.Logger,
```
oleg-ssvlabs, line 202: nit: `logger` param is unused here

beacon/goclient/proposer.go
```diff
@@ -193,23 +252,22 @@ func (gc *GoClient) getProposalParallel(
 
 	for _, client := range gc.clients {
 		go func(c Client) {
```
oleg-ssvlabs, line 254: nit:
from what I remember, since `1.22`, the loop variables are per-iteration by default, so there is no need passing them as function parameter.  You decide if you want to change it

beacon/goclient/proposer.go
```diff
@@ -193,23 +252,22 @@ func (gc *GoClient) getProposalParallel(
...
-			proposal, err := gc.fetchProposal(hardCtx, c, slot, sig, graffiti)
+			proposal, err := gc.fetchProposal(ctx, logger, c, slot, sig, graffiti)
 			select {
 			case resultCh <- result{proposal: proposal, err: err, client: c.Address()}:
```
oleg-ssvlabs, line 257: Is there a need for this select statement?

It seems its purpose is to guard against a blocked `resultCh`, allowing the function to exit when the context is cancelled first. However, I don’t think `resultCh` can actually become blocked, since it’s a buffered channel with a buffer size equal to the number of clients. Wdyt?

beacon/goclient/proposer.go
```diff
@@ -230,14 +288,11 @@ collect:
...
 			)
 
-			if bestProposal == nil ||
-				proposalScore > bestScore ||
-				// this condition prefers the blinded proposal even if same score
-				// as the best we have observed so far
-				(res.proposal.Blinded && proposalScore == bestScore) {
-				bestProposal = res.proposal
-				bestScore = proposalScore
+			bestProposal = gc.selectProposal(bestProposal, res.proposal)
```
oleg-ssvlabs, line 291: I think we are scoring the proposals way too many times. It is cheap operation, but still.  First time [here](https://github.com/ssvlabs/ssv/blob/7b782b39c6425e0c8035e3f2d493683013107516/beacon/goclient/proposer.go#L281), then [here](https://github.com/ssvlabs/ssv/blob/7b782b39c6425e0c8035e3f2d493683013107516/beacon/goclient/proposer.go#L354) and then also [here](https://github.com/ssvlabs/ssv/blob/7b782b39c6425e0c8035e3f2d493683013107516/beacon/goclient/proposer.go#L135) (not all responses, only the one selected).

Could we maybe refactor it? I think we could simplify the `selectProposal` method or do something like this(and remove `selectProposal`):

```Go
if bestProposal == nil ||                                                                                                                                                                                                                       
     proposalScore > bestScore ||                                                                                                                                                                                                                 
     (proposalScore == bestScore && res.proposal.Blinded && !bestProposal.Blinded) {                                                                                                                                                              
      bestProposal = res.proposal                                                                                                                                                                                                                 
      bestClient = res.client                                                                                                                                                                                                                     
      bestScore = proposalScore                                                                                                                                                                                                                   
  }
```

oleg-ssvlabs (commented):
gj!
Left couple of comments on the PR

beacon/goclient/proposer.go
vyzo: I did, but this is a more invasive change i didnt want to make.

I don't like changing interfaces willy nilly.

vyzo: I did, but this is a more invasive change i didnt want to make.

I don't lile changing interfaces willy nilly.

beacon/goclient/proposer.go
vyzo: oops, will fix.

beacon/goclient/proposer.go
vyzo: yes, although i did initially compare the scores. 

Maybe i should restore that logic.

beacon/goclient/proposer.go
vyzo: yeah, but i wanted interface consistency.

beacon/goclient/proposer.go
vyzo: i know, but old habbits die hard, and there are a lot of people that might scratch their heads.

no strong opinions.

beacon/goclient/proposer.go
vyzo: looks like it is not really needed, we can just put to the channel directly.

beacon/goclient/proposer.go
vyzo: the compiler will probably inline that, but we could.  Just easier to read this way.

vyzo: no strong opinions here either, thats a change we can make.

beacon/goclient/proposer.go
oleg-ssvlabs: I’m still not fully convinced about this idea. It feels like an optional parameter, and I think optional parameters should be more explicit and visible.

I usually associate context.Context with passing values of a different purpose - such as trace IDs, authentication tokens, or request-scoped identifiers (something that isn't business logic related).

I don't really have strong opinion here, would be good to hear some other opinions. @y0sher @iurii-ssv

beacon/goclient/proposer.go
oleg-ssvlabs: I mean, it's non exported method, I don't think interface consistency is really important here. Wdyt?

beacon/goclient/proposer.go
iurii-ssv: Generally I would agree with Oleg on this, unless there is a very very strong reason to not change that interface.

y0sher:
So we ended up taking a different approach 🙏

PR #2639, opened 2026-01-07
Proposal: only soft timeout and wait. use 1.8s default and allow flag to change.

y0sher:
---
 ### Summary

  - Remove hard timeout for block proposals - parent context (duty deadline) now serves as the ultimate deadline
  - Reduce default proposal soft timeout from 1.6s to 1.8s
  - Adjust soft timeout dynamically based on proposer delay to maintain consistent timing
  - Add power user mode for explicit timeout override via environment variable

 ### Problem

  The previous timeout calculation could result in insufficient time for block fetching when combined with proposer delay. For example, with a 1.5s proposer delay and 1.6s soft timeout, operators had only 100ms remaining before hitting the hard timeout, causing missed proposals.

### Solution

  1. No hard timeout: The parent context from the duty runner (bounded by slot timing) serves as the ultimate deadline. We never give up early on getting a block.
  2. Dynamic soft timeout: Default 1800ms, reduced by proposer delay with 500ms minimum:
    - 0ms delay → 1800ms collection
    - 500ms delay → 1300ms collection
    - 1000ms delay → 800ms collection
    - 1300ms+ delay → 500ms collection (minimum)
  3. Power user mode: When WITH_PROPOSAL_SOFT_TIMEOUT is explicitly set, the value is used as-is without proposer delay reduction.

 ### Test plan

  - Unit tests pass
  - Deploy to stage with proposer delay and verify proposals complete successfully
  - Monitor proposal timing and success rates

### DIFF
--- a/beacon/goclient/options.go
+++ b/beacon/goclient/options.go
@@ -1,7 +1,6 @@
 package goclient
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/ssvlabs/ssv/networkconfig"
@@ -12,9 +11,23 @@ const (
 	DefaultCommonTimeout = time.Second * 5  // For dialing and most requests.
 	DefaultLongTimeout   = time.Second * 60 // For long requests.
 
-	// Proposal timeouts
-	DefaultProposalSoftTimeout = time.Millisecond * 1600
-	DefaultProposalHardTimeout = time.Millisecond * 2600
+	// DefaultProposalSoftTimeout is the base collection period during which we
+	// gather proposals from multiple beacon nodes to select the best one.
+	// This value is reduced by the proposer delay to maintain consistent timing.
+	// After the soft timeout, we return the best proposal seen so far, or wait
+	// for the first valid proposal if none received yet.
+	// The parent context (duty deadline) serves as the hard timeout.
+	//
+	// Note: MEV (blinded) blocks return immediately, so this timeout mainly
+	// affects how long we wait for MEV when we first receive a vanilla block.
+	//
+	// Can be overridden via WITH_PROPOSAL_SOFT_TIMEOUT env var. When explicitly
+	// set, the value is used as-is without proposer delay reduction (power user mode).
+	DefaultProposalSoftTimeout = time.Millisecond * 3000
+
+	// MinProposalSoftTimeout is the minimum collection period, even with maximum
+	// proposer delay. This ensures we always have some time to compare proposals.
+	MinProposalSoftTimeout = time.Millisecond * 500
 )
 
 // Options defines beacon client options
@@ -28,8 +41,7 @@ type Options struct {
 	CommonTimeout time.Duration `yaml:"CommonTimeout" env:"WITH_COMMON_TIMEOUT" env-description:"Specifies the common timeout for network operations"`
 	LongTimeout   time.Duration `yaml:"LongTimeout" env:"WITH_LONG_TIMEOUT" env-description:"Specifies the long timeout for network operations"`
 
-	ProposalSoftTimeout time.Duration `yaml:"ProposalSoftTimeout" env:"WITH_PROPOSAL_SOFT_TIMEOUT" env-description:"Specifies the beacon proposal collection soft timeout; it will be adjusted for the proposer delay"`
-	ProposalHardTimeout time.Duration `yaml:"ProposalHardTimeout" env:"WITH_PROPOSAL_HARD_TIMEOUT" env-description:"Specifies the beacon proposal collection hard timeout; it will be adjusted for the proposer delay"`
+	ProposalSoftTimeout time.Duration `yaml:"ProposalSoftTimeout" env:"WITH_PROPOSAL_SOFT_TIMEOUT" env-description:"Specifies the beacon proposal collection soft timeout (collection period for comparing proposals from multiple beacon nodes)"`
 }
 
 func NewOptions(base Options, proposerDelay time.Duration) (Options, error) {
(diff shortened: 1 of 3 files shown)


vyzo (commented):
pretty reasonable if it works reliably.

What abour the alternate approach? it is similar to this, but pltentially more reliable at the expense of calling the beacons twice.

beacon/goclient/options.go
```diff
@@ -12,9 +11,23 @@ const (
...
+	// After the soft timeout, we return the best proposal seen so far, or wait
+	// for the first valid proposal if none received yet.
+	// The parent context (duty deadline) serves as the hard timeout.
+	//
+	// Note: MEV (blinded) blocks return immediately, so this timeout mainly
+	// affects how long we wait for MEV when we first receive a vanilla block.
+	//
+	// Can be overridden via WITH_PROPOSAL_SOFT_TIMEOUT env var. When explicitly
+	// set, the value is used as-is without proposer delay reduction (power user mode).
+	DefaultProposalSoftTimeout = time.Millisecond * 3000
```
iurii-ssv: This whole comment would probably make more sense if put on the `Options.ProposalSoftTimeout` instead, since this is just a default value we define here.

Also, do I understand it right that when we have ProposerDelay = 1.5s and WITH_PROPOSAL_SOFT_TIMEOUT=2s set it means: we'll wait for 1.5s+2s before we can start QBFT **when we got only vanilla blocks** during all that 2s waiting ?

beacon/goclient/options.go
```diff
@@ -12,9 +11,23 @@ const (
...
+
+	// MinProposalSoftTimeout is the minimum collection period, even with maximum
+	// proposer delay. This ensures we always have some time to compare proposals.
+	MinProposalSoftTimeout = time.Millisecond * 500
```
iurii-ssv: Maybe to clarify: `even with maximum proposer delay` -> `regardless of ProposerDelay or WithProposalSoftTimeout values`

beacon/goclient/options.go
```diff
@@ -12,9 +11,23 @@ const (
 	DefaultCommonTimeout = time.Second * 5  // For dialing and most requests.
 	DefaultLongTimeout   = time.Second * 60 // For long requests.
 
-	// Proposal timeouts
-	DefaultProposalSoftTimeout = time.Millisecond * 1600
-	DefaultProposalHardTimeout = time.Millisecond * 2600
+	// DefaultProposalSoftTimeout is the base collection period during which we
+	// gather proposals from multiple beacon nodes to select the best one.
+	// This value is reduced by the proposer delay to maintain consistent timing.
```
iurii-ssv: I would use the same naming for "proposer delay" and "WITH_PROPOSAL_SOFT_TIMEOUT" in these comments (here and below) to make it clear that both refer to specific configuration setting,

eg. "PROPOSER_DELAY" and "WITH_PROPOSAL_SOFT_TIMEOUT" (or ProposerDelay and WithProposalSoftTimeout)

iurii-ssv (commented):
Just to make sure I understand, the soft timeout = 3s when waited out fully pretty much means we won't have enough time to do a round 2, that's by design ?

beacon/goclient/options.go
y0sher: Basically yes. setting the soft timeout value manually cancels it effect on the proposer delay.
Operators that setup proposer delay optimize for mev blocks

beacon/goclient/options.go
y0sher: not really, WithProposalSoftTimeout will cancel the min timeout and we'll use the passed value.

beacon/goclient/options.go
```diff
@@ -23,7 +23,7 @@ const (
 	//
 	// Can be overridden via WITH_PROPOSAL_SOFT_TIMEOUT env var. When explicitly
 	// set, the value is used as-is without proposer delay reduction (power user mode).
-	DefaultProposalSoftTimeout = time.Millisecond * 3000
+	DefaultProposalSoftTimeout = time.Millisecond * 1800
```
iurii-ssv: Probably worth adding a note about why this value is 1800 and not 3000, eg. `This value shouldn't be too high because Operator might not be able to participate in QBFT round 2 (or finish it in time) if it is roughly > 2000ms`.

beacon/goclient/options.go
iurii-ssv: Ah, I see, the `WithProposalSoftTimeout` overrides that `MinProposalSoftTimeout` value ... 

I'm not sure though why have that discrepancy:
- we enforce that `MinProposalSoftTimeout` when `ProposerDelay` is set
- we don't enforce `MinProposalSoftTimeout` when `WithProposalSoftTimeout` is set

to me it seems we should enforce for both cases some reasonably minimal value the same way (or maybe not have that min value at all) ?

**Update:** ah, never mind, I see where the discrepancy comes from, it's basically because `ProposalSoftTimeout` and `WithProposalSoftTimeout` have 2 different meanings. 

I wish we could make it a bit more explicit in the code though - `Options.ProposalSoftTimeout` should actually be called `Options.WithProposalSoftTimeout` if it is a user-configured setting, wdyt ?

beacon/goclient/options.go
```diff
@@ -28,8 +41,7 @@ type Options struct {
 	CommonTimeout time.Duration `yaml:"CommonTimeout" env:"WITH_COMMON_TIMEOUT" env-description:"Specifies the common timeout for network operations"`
 	LongTimeout   time.Duration `yaml:"LongTimeout" env:"WITH_LONG_TIMEOUT" env-description:"Specifies the long timeout for network operations"`
 
-	ProposalSoftTimeout time.Duration `yaml:"ProposalSoftTimeout" env:"WITH_PROPOSAL_SOFT_TIMEOUT" env-description:"Specifies the beacon proposal collection soft timeout; it will be adjusted for the proposer delay"`
-	ProposalHardTimeout time.Duration `yaml:"ProposalHardTimeout" env:"WITH_PROPOSAL_HARD_TIMEOUT" env-description:"Specifies the beacon proposal collection hard timeout; it will be adjusted for the proposer delay"`
+	ProposalSoftTimeout time.Duration `yaml:"ProposalSoftTimeout" env:"WITH_PROPOSAL_SOFT_TIMEOUT" env-description:"Specifies the beacon proposal collection soft timeout (collection period for comparing proposals from multiple beacon nodes)"`
```
iurii-ssv: Maybe we should fix the discrepancy between `ProposalSoftTimeout` and `WITH_PROPOSAL_SOFT_TIMEOUT` ? See more on that in https://github.com/ssvlabs/ssv/pull/2639#discussion_r2681153634

beacon/goclient/options.go
iurii-ssv: Ah, we also have the naming discrepancy for `CommonTimeout` and `LongTimeout` ... so I guess we should leave the names as they currently are (`ProposalSoftTimeout` & `WITH_PROPOSAL_SOFT_TIMEOUT`) too.

But I think the min value should still either apply to **both** scenarios (which is what I've done in https://github.com/ssvlabs/ssv/pull/2647) or to **neither**.

beacon/goclient/options.go
iurii-ssv: Resolved in https://github.com/ssvlabs/ssv/pull/2647 by simplifying this comment

beacon/goclient/options.go
iurii-ssv: Addressed in https://github.com/ssvlabs/ssv/pull/2647

PR #2645, opened 2026-01-12
Exporter: Refactor WS endpoint to use the new exporter2 package (remove duplication)

julienh-ssv:
This is part 1 of the WS endpoint refactoring. 

This PR aims to 
1. keep diff minimal
2. while using most of the processing logic from exporter2 package. 
 
Follow up PR will address finer refactoring of the leftover processing logic that will require a larger diff with more fine-grained behaviour tweaks. In particular, I have left some glue code `operator/node.go` that we can refactor in the follow up PR.

Known change in functional behaviour that we consider acceptable:

- WS `/query` now omits duties with empty signers.
- ~WS `/query` error handling internal errors now surfaces an aggregated multi-error rather than last error only~ (edit: old behaviour reestablished in latest commit for retro-compat)

### DIFF
--- a/operator/node.go
+++ b/operator/node.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ssvlabs/ssv/exporter"
 	"github.com/ssvlabs/ssv/exporter/api"
 	exporterstore "github.com/ssvlabs/ssv/exporter/store"
+	"github.com/ssvlabs/ssv/exporter2"
 	qbftstorage "github.com/ssvlabs/ssv/ibft/storage"
 	"github.com/ssvlabs/ssv/network"
 	"github.com/ssvlabs/ssv/networkconfig"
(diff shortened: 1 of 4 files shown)


operator/node.go
```diff
@@ -79,7 +74,7 @@ type Node struct {
 	ws        api.WebSocketServer
 	wsAPIPort int
 
-	traceCollector dutyTraceDecidedsProvider
+	exporterRead *exporter2.Exporter
```
iurii-ssv, line 77: Maybe `exporterRead`->`exporterReadOnly` (or maybe just call it `exporterV2`) ?

iurii-ssv (approved):
LGTM

operator/node.go
julienh-ssv: Yeah actually this will likely change when I complete the refactoring and add the write version too. Although `exporterReadOnly` would be too long for not much added meaning compared to `exporterRead`.

y0sher (approved):
lgtm

PR #2649, opened 2026-01-13
feat(github): increase stale time

zktaiga:


.github/workflows/stale.yml
```diff
@@ -17,8 +17,8 @@ jobs:
     steps:
       - uses: actions/stale@v9.1.0
         with:
-          days-before-stale: 60
-          days-before-close: 30
+          days-before-stale: 90
+          days-before-close: 90
```
iurii-ssv, line 21: I think it's better to have some delay between `days-before-stale` and `days-before-close` for issue author to be able to react and remove the "stale" label (to prevent issue closure),

would suggest these values:
- days-before-stale = 90
- days-before-close = 120

@zktaiga ^

nkryuchkov: @iurii-ssv doesn't it work like 90+90=180? After 90 days, it's stale, then after 90 more days, it's closed

iurii-ssv: Oh, hmm, you are right, but then I think it would be more convenient to do something like this(because there is no need to ping every 3 month if the closure will only happen in 6 month):
- days-before-stale = 180
- days-before-close = 30

nkryuchkov: I agree 180/30 is better. I think 90 is too early to consider issues/PRs stale because many PRs keep being worked on longer than that. Also, we could probably increase it even more for issues, as some of them can easily remain open for more than a year. E.g., long-term protocol changes. I'd suggest 360/30 for issues

iurii-ssv: Yeah, I would also prefer 360/30 even more so.

y0sher (approved):

Issue #2437, opened 2025-08-11
unit-tests: TestRun_ActualExecution is racy


iurii-ssv:
Looks like there is a race in `TestRun_ActualExecution` test(s), running `make docker-unit-test` locally on Mac I got the following:
```
ok  	github.com/ssvlabs/ssv/api/handlers	4.310s	coverage: 96.3% of statements
==================
WARNING: DATA RACE
Read at 0x00c000132fb0 by goroutine 12:
  github.com/ssvlabs/ssv/api/server.TestRun_ActualExecution()
      /go/src/github.com/ssvlabs/ssv/api/server/server_test.go:199 +0x6c8
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1792 +0x180
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1851 +0x40

Previous write at 0x00c000132fb0 by goroutine 14:
  github.com/ssvlabs/ssv/api/server.(*Server).Run()
      /go/src/github.com/ssvlabs/ssv/api/server/server.go:166 +0x17fc
  github.com/ssvlabs/ssv/api/server.TestRun_ActualExecution.func1()
      /go/src/github.com/ssvlabs/ssv/api/server/server_test.go:178 +0x34

Goroutine 12 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1851 +0x684
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:2279 +0x7c
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1792 +0x180
  testing.runTests()
      /usr/local/go/src/testing/testing.go:2277 +0x77c
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:2142 +0xb68
  main.main()
      _testmain.go:65 +0x110

Goroutine 14 (running) created at:
  github.com/ssvlabs/ssv/api/server.TestRun_ActualExecution()
      /go/src/github.com/ssvlabs/ssv/api/server/server_test.go:177 +0x558
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1792 +0x180
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1851 +0x40
==================
==================
WARNING: DATA RACE
Write at 0x00c000122190 by goroutine 12:
  ??()
      -:0 +0x1217bec
  sync/atomic.StoreUint32()
[...]

PR #2650, opened 2026-01-13
exporter: fix racy API tests

iurii-ssv:
Resolves https://github.com/ssvlabs/ssv/issues/2437

The issue stems from the way `Server.Run` method is implemented (besides running the server, it also initializes some stuff on `Server` struct ... which makes it tricky to call `Server.Run` in a separate go-routine).

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -19,14 +19,9 @@ import (
 // Server represents the HTTP API server for SSV.
 type Server struct {
 	logger *zap.Logger
-	addr   string
 
-	node       *nodeHandlers.Node
-	validators *validatorsHandlers.Validators
-	exporter   *exporterHandlers.Exporter
+	addr       string
 	httpServer *http.Server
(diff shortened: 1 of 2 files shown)


api/server/server.go
```diff
@@ -19,14 +19,9 @@ import (
...
-	addr   string
 
-	node       *nodeHandlers.Node
-	validators *validatorsHandlers.Validators
-	exporter   *exporterHandlers.Exporter
+	addr       string
 	httpServer *http.Server
-
-	fullExporter bool
 }
```
iurii-ssv, line 25: Removed some of these fields on this struct because we can just use them "as is" in `New` func (storing these here only complicates this code for no reason)

julienh-ssv (approved):

nkryuchkov (approved):

y0sher (approved):

PR #2652, opened 2026-01-13
tests: run on Boole fork

nkryuchkov:
Currently, the Boole fork epoch on `TestNetwork` is set to `math.MaxUint64`, so the unit tests run on the Alan fork. We need to run them on the Boole fork to make sure everything passes. Also, the Alan logic will be removed soon, so we need to make sure in advance that tests work there.

The PR sets the Boole fork epoch on `TestNetwork` to 0 to run all tests on the Boole fork.

The change broke message validation tests, so the PR has also fixed them.

### DIFF
--- a/message/validation/consensus_validation.go
+++ b/message/validation/consensus_validation.go
@@ -16,7 +16,7 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
 	"github.com/ssvlabs/ssv/protocol/v2/message"
-	qbft "github.com/ssvlabs/ssv/protocol/v2/qbft"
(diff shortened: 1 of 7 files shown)


julienh-ssv (changes requested):
In the other PR for RoundRobinProposer we decided to not change any of the legacy code to prevent introducing any bug in the pre-fork behavior. (So, pre-fork -> no duplication removal, no signature changes, nothing basically).

I'll work on finishing the fix in this PR: https://github.com/ssvlabs/ssv/pull/2654

nkryuchkov:
Superseded with https://github.com/ssvlabs/ssv/pull/2654#pullrequestreview-3659612329

PR #2654, opened 2026-01-14
Boole Fork: enable bool fork for tests CI

julienh-ssv:


nkryuchkov (dismissed):

network/p2p/p2p_test.go
```diff
@@ -202,6 +204,15 @@ func generateCommitteeMsg(ks *spectestingutils.TestKeySet, round specqbft.Round)
...
 	share := spectestingutils.TestingShare(ks, 1)
+	netCfg := networkconfig.TestNetwork
+
+	if netCfg.BooleForkAtSlot(phase0.Slot(height)) {
+		committee := make([]spectypes.OperatorID, 0, len(share.Committee))
+		for _, c := range share.Committee {
+			committee = append(committee, c.Signer)
+		}
+		return qbft.RoundRobinProposer(height, round, committee, netCfg)
+	}
```
iurii-ssv, line 215: Does this `if netCfg.BooleForkAtSlot(phase0.Slot(height))` check do anything ? Now that we've enabled Bool fork (by setting `Boole:      0`) ? Seems like it's always `true`

message/validation/validation_test.go
```diff
@@ -1987,12 +2007,30 @@ func generateSignedMessage(
...
+	signedSSVMessage := spectestingutils.SignQBFTMsg(ks.OperatorKeys[leader], leader, qbftMessage)
 	signedSSVMessage.FullData = fullData
 
 	return signedSSVMessage
 }
 
+var testLeaderNetCfg *networkconfig.Network
+var testLeaderCommittee []spectypes.OperatorID
+
+func leaderForTest(height specqbft.Height, round specqbft.Round) spectypes.OperatorID {
```
iurii-ssv: I'm not sure I understand why these are defined as "globals", can't we just pass these as params to `leaderForTest` func ?

If there is a specific purpose for it - these global definitions could use a comment, I think.

network/p2p/p2p_test.go
nkryuchkov: We can change the epoch to verify quickly that tests pass pre-fork. I think it's nice to have this check

iurii-ssv: Feels like there is no need to merge code like this into stage/main though (to keep things simple) ?

Why not have an open PR (marked as Draft) with that specific change, so we can use that PR to run all the tests against the latest code with disabled Bool fork (just need to rebase such a PR onto the latest stage/main branch version whenever we want to trigger a test-run).

nkryuchkov: IMO, it's fine to have this check both in the code and tests. It keeps things aligned. I think tests should pass with any Boole epoch set, and the epoch should be 0 by default. After the fork, we'll just clean up both code and tests

iurii-ssv: Okay, if you put it like that - makes sense.

message/validation/validation_test.go
julienh-ssv: I pushed a commit to remove the global variables

iurii-ssv (approved):
LGTM

nkryuchkov (approved):

network/p2p/p2p_test.go
y0sher: I still think it's important that we run the tests both with prefork epoch and post fork epoch. triggering manually isn't a very good solution for that.

iurii-ssv: cc @julienh-ssv ^

nkryuchkov: It seems to have already been addressed by @julienh-ssv in https://github.com/ssvlabs/ssv/pull/2655

julienh-ssv: I ran the tests on both versions and there is a separate PR to enable running the CI on both. Once we merge both the PR CI will run both old and new and pass both. @y0sher

julienh-ssv:
Closing this PR. We will merge with https://github.com/ssvlabs/ssv/pull/2665 instead.

PR #2503, opened 2025-09-09
Aggregator, Sync Committee Contribution - Committee Runner

olegshmuelov:
align to https://github.com/ssvlabs/ssv-spec/pull/572
SIP: https://github.com/ssvlabs/SIPs/pull/61

PR #2658, opened 2026-01-15
committee-duty: clarify BLS signature reconstruction (edge-cases)

iurii-ssv:
This PR clarifies some edge-case behavior related to BLS signature reconstruction in committee-duty to document the findings from https://github.com/ssvlabs/ssv/pull/2503#discussion_r2650673131 (thanks to @MatheusFranco99 chiming in),

the only "behavior change" (an optimization) in this PR is explained in this comment:

```
					// 2) we need to verify partial signatures only for the roots we haven't tried reconstructing
					//    signatures for (hence roots[i:])
```

`for _, root := range roots` -> `for _, root := range roots[i:]`

@nkryuchkov once we merge these changes, it would be nice if you could apply the same to ]aggregator-committee runner](https://github.com/ssvlabs/ssv/pull/2503).


protocol/v2/ssv/runner/committee.go
```diff
@@ -638,7 +654,7 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 					validatorIndex: validatorIndex,
 					signature:      (phase0.BLSSignature)(sig),
 				}
-			}(validator, root, roots)
+			}(validator, root)
```
iurii-ssv, line 657: Using `roots` slice in the closure-func "as is" now (instead of passing as func param) since it's the same for every go-routine

oleg-ssvlabs (approved):
lgtm

y0sher (approved):

PR #2663, opened 2026-01-19
Adjustments by iurii to #2659 (part 2)

iurii-ssv:
This should resolve a potential concurrency bug we have in `BaseRunner` code (+ documenting some assumptions that code relies on),

I'm saying "potential" because in practice we seemingly always call at least 1 of `BaseRunner.hasRunningDuty`, `BaseRunner.hasDutyAssigned`, `BaseRunner.hasDutyFinished` prior to reading/writing the `BaseRunner.State`, which means it is safe to access `BaseRunner.State` (read & write to it) as long as it's done sequentially after any of those 3 calls. That's most likely why we don't see any of our tests failing with "race detected". But it's a very implicit and bug-prone way of going about it ... so in this PR I suggest we explicitly use `atomic.Pointer` instead.

This PR is also fixing the https://github.com/ssvlabs/ssv/pull/2659#discussion_r2694792480 in similar manner.

### DIFF
--- a/protocol/v2/ssv/runner/aggregator.go
+++ b/protocol/v2/ssv/runner/aggregator.go
@@ -218,7 +218,7 @@ func (r *AggregatorRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 	msg, err := signBeaconObject(
 		ctx,
 		r,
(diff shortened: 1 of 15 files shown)


iurii-ssv:
Closing in favor of https://github.com/ssvlabs/ssv/pull/2664

PR #2664, opened 2026-01-19
Adjustments by iurii to #2659 (part 3)

iurii-ssv:
As per https://github.com/ssvlabs/ssv/pull/2659#discussion_r2704528406 there is no need for `BaseRunner.mtx` to exist,

this PR:
- removes `BaseRunner.mtx`
- documents the assumptions `BaseRunner`relies on to work correctly in the concurrent setting

This PR also supercedes https://github.com/ssvlabs/ssv/pull/2663 (+ #2663 would be an incomplete solution anyway).

### DIFF
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -81,6 +81,15 @@ type DoppelgangerProvider interface {
 var _ Runner = new(CommitteeRunner)
 
 type BaseRunner struct {
+	// State stores the current runner state, this state corresponds to 1 particular duty the runner is
+	// currently busy with at the moment. The BaseRunner is not responsible for synchronizing any updates
+	// State might need to record - the caller is responsible to ensure the updates/reads (these can happen
+	// whenever runner's method is called to process a p2p message, or an event) are applied sequentially,
+	// plus the caller is also responsible for ensuring there is no race with moving on to the next duty
+	// (the baseSetupForNewDuty call).
+	// Note, the current implementation achieves concurrent safety by making sure every State read/update
+	// is done by the same go-routing, handling all the messages in queue.SSVMessage (p2p messages and events)
+	// sequentially.
 	State *State
 
 	Share          map[phase0.ValidatorIndex]*spectypes.Share
--- a/protocol/v2/ssv/runner/runner.go
+++ b/protocol/v2/ssv/runner/runner.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"maps"
 	"slices"
-	"sync"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	ssz "github.com/ferranbt/fastssz"
@@ -82,8 +81,8 @@ type DoppelgangerProvider interface {
 var _ Runner = new(CommitteeRunner)
 
 type BaseRunner struct {
-	mtx            sync.RWMutex
-	State          *State
+	State *State
+
 	Share          map[phase0.ValidatorIndex]*spectypes.Share
 	QBFTController *controller.Controller
 	NetworkConfig  *networkconfig.Network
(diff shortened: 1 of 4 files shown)


protocol/v2/ssv/runner/runner.go
```diff
@@ -82,8 +81,17 @@ type DoppelgangerProvider interface {
...
-	State          *State
+	// State stores the current runner state, this state corresponds to 1 particular duty the runner is
+	// currently busy with at the moment. The BaseRunner is not responsible for synchronizing any updates
+	// State might need to record - the caller is responsible to ensure the updates/reads (these can happen
+	// whenever runner's method is called to process a p2p message, or an event) are applied sequentially,
+	// plus the caller is also responsible for ensuring there is no race with moving on to the next duty
+	// (the baseSetupForNewDuty call).
+	// Note, the current implementation achieves concurrent safety by making sure every State read/update
+	// is done by the same go-routing, handling all the messages in queue.SSVMessage (p2p messages and events)
+	// sequentially.
```
momosh-ssv, line 92: Should we keep the RWMutex for the State read-path methods? Both `Validator.onTimeout` (timer.go:37) and `Committee.onTimeout` (timer.go:98) call `HasRunningDuty()` from a timer goroutine, which reads `b.State` concurrently with the queue goroutine writing it in `baseSetupForNewDuty` and setting `b.State.Finished`. The removed `b.mtx.RLock` in `hasRunningDuty`/`hasDutyAssigned`/`hasDutyFinished` was protecting exactly this access pattern. Alternatively, the timer could route `HasRunningDuty` through the queue so it never reads State directly.

iurii-ssv: Due to `onTimeout` being processed as an **event** that is handled sequentially with respect to all p2p messages Runner receives, there is no races or memory corruption once the mutex is removed, explained in more detail in https://github.com/ssvlabs/ssv/pull/2659#discussion_r2704528406

Also, I think https://github.com/ssvlabs/ssv/pull/2659 is probably no longer relevant - so I'll swap the branch for this PR to target **stage** instead

momosh-ssv: ... because I was also looking at the `HasRunningDuty()` which is also reading from the timer goroutine, and wanted to know if this was acceptable. 
Seems like it's a read-only guard with benign stale-read semantics, right?

momosh-ssv (approved):

iurii-ssv:
Converted to draft for now, until I can get back to https://github.com/ssvlabs/ssv/pull/2664#discussion_r2983144107

protocol/v2/ssv/runner/runner.go
iurii-ssv: > because I was also looking at the `HasRunningDuty()` which is also reading from the timer goroutine,

That's a good observation, this `b.mtx` actually **does ensure** that we read "the correct" runner state when processing timeout. But since it's the only such place, and the way `b.mtx` works is non-trivial - I still would like to get rid of it ... we just need to move that `HasRunningDuty()` check elsewhere. I've created a PR into **stage** branch https://github.com/ssvlabs/ssv/pull/2771 that does roughly this:
- we do not perform the `HasRunningDuty()` check in the `onTimeout` func(s) themselves, we just issue the timeout event every time (it's pretty rare, so the overhead from the added events is negligible)
- these timeout events are processed within the same for-loop (`StartQueueConsumer` and `ConsumeQueue` funcs) with every other duty-related message/event, they are all processed **sequentially** (due to reading from same thread-safe queue, in the same for-loop)
- the `HasRunningDuty()` check is performed during the actual processing of those timeout events (see `case types.Timeout:`)

iurii-ssv:
This PR got ported onto **stage** branch https://github.com/ssvlabs/ssv/pull/2771 - since the base branch of this PR is no longer relevant.

PR #2673, opened 2026-01-26
message-validation: cleanup & clarify

iurii-ssv:
This PR implements simple refactoring of some message-validation code (pulled out of https://github.com/ssvlabs/ssv/pull/2635) preserving the same behavior, with the goal of making it more consistent, readable and better documented (clearing the ground for the follow-up validation-rules updates).

The only notable change (an optimization) in this PR is that, per Oleg's suggestion, we can [re-use the readily-available `consensusMessage.Root` value instead of re-computing fhe `FullDatasignedSSVMessage.FullData` hash from scratch](https://github.com/ssvlabs/ssv/pull/2635#discussion_r2672156832)

### DIFF
--- a/message/validation/consensus_validation.go
+++ b/message/validation/consensus_validation.go
@@ -4,7 +4,6 @@ package validation
 
 import (
 	"bytes"
-	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"time"
@@ -154,10 +153,10 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
 	}
 
 	if consensusMessage.Round > maxRound {
-		err := ErrRoundTooHigh
-		err.got = fmt.Sprintf("%v (%v role)", consensusMessage.Round, message.RunnerRoleToString(role))
-		err.want = fmt.Sprintf("%v (%v role)", maxRound, message.RunnerRoleToString(role))
-		return err
+		e := ErrRoundTooHigh
+		e.got = fmt.Sprintf("%v (%v role)", consensusMessage.Round, message.RunnerRoleToString(role))
+		e.want = fmt.Sprintf("%v (%v role)", maxRound, message.RunnerRoleToString(role))
+		return e
 	}
 
 	// Rule: consensus message must have the same identifier as the ssv message's identifier
@@ -186,17 +185,16 @@ func (mv *messageValidator) validateQBFTLogic(
 		// Rule: Signer must be the leader
 		leader := mv.roundRobinProposer(consensusMessage.Height, consensusMessage.Round, committeeInfo.committee)
 		if signedSSVMessage.OperatorIDs[0] != leader {
-			err := ErrSignerNotLeader
-			err.got = signedSSVMessage.OperatorIDs[0]
-			err.want = leader
-			return err
+			e := ErrSignerNotLeader
+			e.got = signedSSVMessage.OperatorIDs[0]
+			e.want = leader
+			return e
 		}
 	}
 
 	msgSlot := phase0.Slot(consensusMessage.Height)
 	for _, signer := range signedSSVMessage.OperatorIDs {
-		signerStateBySlot := state.Signer(committeeInfo.signerIndex(signer))
-		signerState := signerStateBySlot.GetSignerState(msgSlot)
+		signerState := state.OperatorState(committeeInfo.signerIndex(signer)).GetSignerStateForSlot(msgSlot)
 		if signerState == nil {
 			continue
 		}
@@ -207,22 +205,25 @@ func (mv *messageValidator) validateQBFTLogic(
 				// Signers aren't allowed to decrease their round.
 				// If they've sent a future message due to clock error,
 				// they'd have to wait for the next slot/round to be accepted.
-				err := ErrRoundAlreadyAdvanced
-				err.want = signerState.Round
-				err.got = consensusMessage.Round
-				return err
+				e := ErrRoundAlreadyAdvanced
+				e.want = signerState.Round
+				e.got = consensusMessage.Round
+				return e
 			}
 
 			if consensusMessage.Round == signerState.Round {
 				// Rule: Peer must not send two proposals with different data
 				if len(signedSSVMessage.FullData) != 0 && signerState.HashedProposalData != nil {
-					if *signerState.HashedProposalData != sha256.Sum256(signedSSVMessage.FullData) {
-						return ErrDifferentProposalData
+					if *signerState.HashedProposalData != consensusMessage.Root {
+						e := ErrDifferentProposalData
+						e.want = "0x" + hex.EncodeToString((*signerState.HashedProposalData)[:])
+						e.got = "0x" + hex.EncodeToString(consensusMessage.Root[:])
+						return e
 					}
 				}
 
 				// Rule: Peer must send only 1 proposal, 1 prepare, 1 commit, and 1 round-change per round
-				if err := signerState.SeenMsgTypes.ValidateConsensusMessage(signedSSVMessage, consensusMessage); err != nil {
+				if err := validateConsensusMessageLimit(signedSSVMessage, consensusMessage, signerState); err != nil {
 					return err
 				}
 			}
@@ -235,7 +236,8 @@ func (mv *messageValidator) validateQBFTLogic(
 			// Rule: Decided msg can't have the same signers as previously sent before for the same duty
 			if signerState.SeenSigners != nil {
 				if _, ok := signerState.SeenSigners[quorum.ToBitMask()]; ok {
-					return ErrDecidedWithSameSigners
+					e := ErrDecidedWithSameSigners
+					return e
 				}
 			}
 		}
@@ -263,8 +265,8 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
 	// Rule: Height must not be "old". I.e., signer must not have already advanced to a later slot.
 	if role != spectypes.RoleCommittee { // Rule only for validator runners
 		for _, signer := range signedSSVMessage.OperatorIDs {
-			signerStateBySlot := state.Signer(committeeInfo.signerIndex(signer))
-			if maxSlot := signerStateBySlot.MaxSlot(); maxSlot > phase0.Slot(consensusMessage.Height) {
+			operatorState := state.OperatorState(committeeInfo.signerIndex(signer))
+			if maxSlot := operatorState.MaxSlot(); maxSlot > phase0.Slot(consensusMessage.Height) {
 				e := ErrSlotAlreadyAdvanced
 				e.got = consensusMessage.Height
 				e.want = maxSlot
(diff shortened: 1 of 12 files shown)


message/validation/consensus_validation.go
```diff
@@ -207,22 +205,25 @@ func (mv *messageValidator) validateQBFTLogic(
...
 			}
 
 			if consensusMessage.Round == signerState.Round {
 				// Rule: Peer must not send two proposals with different data
 				if len(signedSSVMessage.FullData) != 0 && signerState.HashedProposalData != nil {
-					if *signerState.HashedProposalData != sha256.Sum256(signedSSVMessage.FullData) {
-						return ErrDifferentProposalData
+					if *signerState.HashedProposalData != consensusMessage.Root {
+						e := ErrDifferentProposalData
+						e.want = "0x" + hex.EncodeToString((*signerState.HashedProposalData)[:])
```
nkryuchkov:
```suggestion
						e.want = hexutil.Bytes((*signerState.HashedProposalData)[:]).String()
```

message/validation/consensus_validation.go
```diff
@@ -336,20 +338,19 @@ func (mv *messageValidator) processSignerState(
 	signerState *SignerState,
 ) error {
 	if len(signedSSVMessage.FullData) != 0 && consensusMessage.MsgType == specqbft.ProposalMsgType {
-		fullDataHash := sha256.Sum256(signedSSVMessage.FullData)
-		signerState.HashedProposalData = &fullDataHash
+		signerState.HashedProposalData = &consensusMessage.Root
```
nkryuchkov: Let's make sure we don't need to modify the condition in the if-block

nkryuchkov (dismissed):

message/validation/consensus_validation.go
iurii-ssv: As far as I can tell - we don't, we simply optimize away the hash-calculation (re-use readily-available hash value we verified previously, as [Oleg explained here](https://github.com/ssvlabs/ssv/pull/2635#discussion_r2672156832))

message/validation/consensus_validation.go
iurii-ssv: Addressed in https://github.com/ssvlabs/ssv/pull/2673/changes/c9b48e111328d35d9c27e834447e6c83c1c0fc6b

nkryuchkov (dismissed):

message/validation/consensus_validation.go
```diff
@@ -336,20 +339,19 @@ func (mv *messageValidator) processSignerState(
...
```
olegshmuelov: Taking a pointer to `consensusMessage.Root` will retain the entire `consensusMessage` in memory (including justification slices). Consider copying the value:
```go
root := consensusMessage.Root
signerState.HashedProposalData = &root
```

message/validation/errors.go
```diff
@@ -119,23 +118,24 @@ var (
...
 	ErrFullDataNotInConsensusMessage           = Error{text: "full data not in consensus message", reject: true}
 	ErrTripleValidatorIndexInPartialSignatures = Error{text: "triple validator index in partial signatures", reject: true}
 	ErrZeroRound                               = Error{text: "zero round", reject: true}
-	ErrDuplicatedMessage                       = Error{text: "message is duplicated", reject: true}
-	ErrInvalidPartialSignatureTypeCount        = Error{text: "sent more partial signature messages of a certain type than allowed", reject: true}
-	ErrTooManyPartialSignatureMessages         = Error{text: "too many partial signature messages", reject: true}
+	ErrDuplicatedMessage                       = Error{text: "got duplicate message", reject: true}
+	ErrTooManyPartialSigMessage                = Error{text: "got more partial signature messages of a certain type than allowed", reject: true}
+	ErrDifferentProposalData                   = Error{text: "got different proposal data", reject: true}
+	ErrDecidedWithSameSigners                  = Error{text: "decided with same number of signers", reject: true}
```
olegshmuelov: Moving `ErrDecidedWithSameSigners` to rejected changes this from `ValidationIgnore` to `ValidationReject`, penalizing peers for re-gossiping decided messages with the same signers. These duplicates are expected in gossip protocols. If intentional (prep for duplicate operator fix?), consider adding a brief comment explaining the rationale.

message/validation/consensus_validation.go
```diff
@@ -235,7 +237,8 @@ func (mv *messageValidator) validateQBFTLogic(
 			// Rule: Decided msg can't have the same signers as previously sent before for the same duty
 			if signerState.SeenSigners != nil {
 				if _, ok := signerState.SeenSigners[quorum.ToBitMask()]; ok {
-					return ErrDecidedWithSameSigners
+					e := ErrDecidedWithSameSigners
+					return e
```
olegshmuelov: nit: could just be `return ErrDecidedWithSameSigners`

message/validation/consensus_validation.go
iurii-ssv: The `signerState` is short-lived though (I think its lifetime is somewhere around ~1 epoch), so it doesn't seem like it matters in this case then ?

olegshmuelov: Fair enough, the short lifetime mitigates it. Could still add up with many validators, but optional nit.

message/validation/errors.go
iurii-ssv: Good catch! Actually we probably need to consult with @GalRogozinski and @MatheusFranco99 on this,

from what I understand about this `decided with same number of signers` error:
- we are validating against what's known as **peer-state** here (vs the **global-state** that will be added in the future PRs, for now we don't have it)
- so if we got this error that would mean peer is sending us a duplicate message ? That's how I interpret it and why I changed it in my original PR (the PR about "duplicate-node issue") ... here it just got copied-over as is
- handling it as REJECT (as opposed to IGNORE as we currently do in **stage** branch) would match the handling we have for other errors like: `got duplicate message`, `got more partial signature messages of a certain type than allowed`, etc.

message/validation/consensus_validation.go
iurii-ssv: Applied in https://github.com/ssvlabs/ssv/pull/2673/changes/3d0f32567e00125011c7872f0b8bf7f6a9bbaf1c

message/validation/consensus_validation.go
iurii-ssv: Okay, applied in https://github.com/ssvlabs/ssv/pull/2673/changes/3d0f32567e00125011c7872f0b8bf7f6a9bbaf1c just in case.

message/validation/errors.go
MatheusFranco99: If I remember correctly, for an example, this rule will ignore a decided for signers `[1,2,4]` from a peer if it already received from this peer, for this duty, another decided with the same number of signers (e.g. `[1,2,3]`). Correct?

If so, I don't think it's a malicious behaviour (as a duplicated msg) to send these two. E.g. node `A` may receive `decided([1,2,3])` from B and propagate, and may receive `decided([1,2,4])` from `C` and propagate.
So I think it shouldn't be punished (rejected).

Wdyt ?
cc @GalRogozinski

iurii-ssv: @MatheusFranco99 the [rule seemingly states that it is checking for the exact match](https://github.com/ssvlabs/ssv/blob/6c6941a26b86898af97daeaaaa65daaa29ba4c0f/message/validation/consensus_validation.go#L235-L240):
```go
			// Rule: Decided msg can't have the same signers as previously sent before for the same duty
			if signerState.SeenSigners != nil {
				if _, ok := signerState.SeenSigners[quorum.ToBitMask()]; ok {
					return ErrDecidedWithSameSigners
				}
			}
```
and as far as I can tell, the way this code works is:
- suppose we got a message `decided([1,2,3])` already
- if we get `decided([1,2,4])` after that - that's fine, no rule is triggered by that, it's different signers (even though the number of signers is the same as came with the previous message)
- if we get `decided([1,2,3])` after that - this rule will produce `decided with same number of signers` error ... so either the error-message is incorrect (it should mention about exact signers matching, not just the number of them being the same) or the implementing code is incorrect, seems to me

iurii-ssv: Basically there are 2 cases:
- signers amount matches prev message, signers DOES match prev message exactly
- signers amount matches prev message, signers do NOT match prev message exactly

and the `ErrDecidedWithSameSigners           = Error{text: "decided with same number of signers"}` sort of refers to both ... which is confusing,

and I'm not sure what happens in other cases of signers amount does NOT match ... but I assume that's covered elsewhere ?

MatheusFranco99: Hmm, the description and error name are confusing yes. I got a bit confused, I believe.

> what happens in other cases of signers amount does NOT match

This is just fine and can be accepted.

For the cases you listed for diff msgs:
- `signers DOES match`: then this seems odd indeed, and may be a malicious behaviour I believe. Always, for rejection rules, I prefer re-thinking them over and over before making a decision, but a quick answer would be that this is indeed rejectable.
- `signers do NOT match`: this is valid, but can be ignored.

Wdyt @GalRogozinski @iurii-ssv ?

MatheusFranco99: I actually think this name and description confusion preverses a little of history.

As I remember, this ignore rule was originally created only for the case of "same number of signers". It was created to avoid getting **tons** of decided messages. Note that, indeed, many valid combinations are possible (grows combinatorially with the number of operators in the committee) and we wanted to avoid such an attack (that can't be penalized as it only sends valid msgs).

It was fine to "ignore" and drop it since the extra ones were irrelevant for the `decided` msg purpose, while we would allow new msgs with more signers for explorer purposes (operator's score).

iurii-ssv: > For the cases you listed for diff msgs:
> 
> * `signers DOES match`: then this seems odd indeed, and may be a malicious behaviour I believe. Always, for rejection rules, I prefer re-thinking them over and over before making a decision, but a quick answer would be that this is indeed rejectable.
> * `signers do NOT match`: this is valid, but can be ignored.
> 
> Wdyt @GalRogozinski @iurii-ssv ?

Yes that would make sense, so we currently check just for 1 of those cases (and also handle it poorly I believe) ... but we also need to handle the other case.

So what I would propose is this:
- in this PR I'll revert to the previous behavior where we do this: `if signers DOES match -> IGNORE`
- in a follow-up PR (the PR where I'm working on a "duplicate SSV node issue") I will also update the way we handle this `ErrDecidedWithSameSigners` so that it works as expected: 1) if `signers DOES match` then REJECT the message, and 2) if `signers do NOT match` then IGNORE the message

@MatheusFranco99 correct ?

MatheusFranco99: Agree

iurii-ssv: > * in this PR I'll revert to the previous behavior where we do this: `if signers DOES match -> IGNORE`

Okay, https://github.com/ssvlabs/ssv/pull/2673/changes/c44907f400bee4a9594f42c1bf8a64d71367153b does that ^

olegshmuelov (approved):
LGTM.

y0sher (approved):
great job!

PR #2674, opened 2026-01-27
v2.4.0: stage -> main

iurii-ssv:
As per the title. There is no changes in **main** branch that we'd need to apply to **stage** branch first, so this PR should be clean to merge.


protocol/v2/ssv/runner/committee.go
```diff
@@ -628,18 +608,36 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
+					// 1) FallBackAndVerifyEachSignature call may also lead to a certain root+validator pairs
+					//    in PostConsensusContainer not having quorum anymore since it previously was computed
+					//    optimistically.
+					// 2) we need to verify partial signatures only for the roots we haven't tried reconstructing
+					//    signatures for (hence roots[i:])
+					// 3) since this code is running a bunch of concurrent go-routines, we need to be careful to
+					//    not call FallBackAndVerifyEachSignature for the same root+validator pair multiple times -
+					//    this is why we are parallelizing by validators only (and not by root+validator), processing
+					//    each root sequentially
+					for _, root := range roots[i:] {
```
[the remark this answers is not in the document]
iurii-ssv: This shouldn't be an issue with the latest Golang versions, right ?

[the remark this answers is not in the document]
iurii-ssv: Good thing we've upgraded to Go 1.22+ a while ago.

oleg-ssvlabs (dismissed):

protocol/v2/ssv/runner/runner.go
```diff
@@ -234,19 +237,27 @@ func (b *BaseRunner) basePreConsensusMsgProcessing(ctx context.Context, logger *
...
 
 	if hasQuorum {
 		const gotPreConsensusQuorumEvent = "🎯 got pre-consensus quorum"
-		logger.Debug(gotPreConsensusQuorumEvent)
+		logger.Debug(gotPreConsensusQuorumEvent, fields.QuorumRoots(quorumRoots))
 		span.AddEvent(gotPreConsensusQuorumEvent)
 	}
 
-	return hasQuorum, roots, nil
+	return hasQuorum, slices.Collect(maps.Keys(quorumRoots)), nil
```
olegshmuelov, line 260: Potential ordering issue with quorum roots

Using `maps.Keys()` here returns roots in nondeterministic order per Go's map iteration semantics.

However, `sync_committee_contribution.go:136` assumes positional alignment between `roots[i]` and `ValidatorSyncCommitteeIndices[i]` - that relationship only holds if roots are returned in the same order they were originally generated in `expectedPreConsensusRootsAndDomain()`.

When a validator occupies multiple sync committee positions, this ordering mismatch can cause:
1. Selection proofs paired with incorrect subnets
2. Beacon node rejecting the resulting contributions
3. Contributions submitted to wrong subcommittees

**Suggested fix:** Ensure roots are returned in deterministic order matching `ValidatorSyncCommitteeIndices`. This could be achieved by iterating over the expected roots from `expectedPreConsensusRootsAndDomain()` and filtering for those present in the quorum map, rather than relying on `maps.Keys()`.

y0sher: good point, maybe we could just sort ?

olegshmuelov: Sorting would make it deterministic, but the real issue is that `roots[i]` needs to match `ValidatorSyncCommitteeIndices[i]` not just be in some consistent order.

Say we have indices `[100, 200]` producing roots `[rootA, rootB]`. If we sort by hash we might get `[rootB, rootA]`, so `roots[0]` still gets paired with the wrong index.

I think we'd need to iterate over the expected roots from `expectedPreConsensusRootsAndDomain()` and keep only the ones that reached quorum, that preserves the original alignment.

iurii-ssv: > I think we'd need to iterate over the expected roots from expectedPreConsensusRootsAndDomain() and keep only the ones that reached quorum, that preserves the original alignment.

Yeah that sounds like what we need, I'll sketch a PR into **stage** branch with that fix (and then we can rebase this branch on top of **stage** branch once the fix is merged),

Good find! This issue was present even in v2.3.9 and prior versions, right (since the amount of roots returned by `basePreConsensusMsgProcessing` can be anywhere from 0 to N and we are iterating over them like that: `for i, root := range roots` we can't just "blindly" access `ValidatorSyncCommitteeIndices[i]` by `i` in cases when only some roots got the quorum ... even with the matching/correct order) ? Need to make this code a bit more explicit about these assumptions anyway.

iurii-ssv: https://github.com/ssvlabs/ssv/pull/2675

nkryuchkov (dismissed):

y0sher (dismissed):

olegshmuelov (approved):

nkryuchkov (approved):

julienh-ssv (approved):

y0sher (approved):

PR #2677, opened 2026-02-03
message-validation: rework duplicate message rejections

iurii-ssv:
Resolves https://github.com/ssvlabs/ssv/issues/2632, supersedes https://github.com/ssvlabs/ssv/pull/2635

This PR implements p2p message-validation enhancements described in https://github.com/ssvlabs/knowledge-base/pull/3


message/validation/consensus_validation.go
```diff
@@ -213,32 +210,46 @@ func (mv *messageValidator) validateQBFTLogic(
...
+			if len(signedSSVMessage.OperatorIDs) <= signerState.Peer(receivedFrom).SeenDecidedMsgSignersCount {
+				e := ErrDecidedMessageWithUnexpectedNumberOfSigners
+				e.reject = true
+				return e
+			}
+			// Check if a different peer is sending us a "logical duplicate" message, ignore message since this
+			// is expected occasionally.
+			if len(signedSSVMessage.OperatorIDs) <= signerState.World.SeenDecidedMsgSignersCount {
+				e := ErrDecidedMessageWithUnexpectedNumberOfSigners
+				return e
```
iurii-ssv, line 246: This change implements the conclusion we came to in [this discussion](https://github.com/ssvlabs/ssv/pull/2673#discussion_r2741990708), 

previuosly (in **stage** branch) we would:
- ACCEPT decided messages unless we already saw a decided message with exactly the same signers. For example, we saw [1,2,3] and [1,2,4] arrives - we ACCEPT, [1, 2] arrives - we ACCEPT, etc.
- IGNORE otherwise (we never REJECT even if it's the same peer sending us the same message over and over). For example, we saw [1,2,3] and [1,2,3] arrives - we IGNORE

in this PR we would:
- ACCEPT only decided messages with increasing amount of signers (I think Matheus said it's needed for exporter)
- IGNORE/REJECT decided messages with the same amount of singers or less, we REJECT only when it's clear that the same peer is sending us a 2nd, 3rd, etc. message (and IGNORE otherwise)

message/validation/signer_state.go
```diff
@@ -4,46 +4,64 @@ package validation
...
-	//  7 (f=2): C(7,5)+C(7,6)+C(7,7)=29
-	// 10 (f=3): C(10,7)+C(10,8)+C(10,9)+C(10,10)=176
-	// 13 (f=4): C(13,9)+C(13,10)+C(13,11)+C(13,12)+C(13,13)=1093
-	SeenSigners map[SignersBitMask]struct{}
+func (s *SignerStateForSlotRound) Peer(peerID peer.ID) *SignerState {
+	state := s.Peers[peerID]
+	if state == nil {
+		s.Peers[peerID] = &SignerState{}
+	}
+	return s.Peers[peerID]
```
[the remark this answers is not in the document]
iurii-ssv: Not sure I understand your concern/comment, isn't the `return s.Peers[peerID]` I'm doing exactly the same as `return state` in the code you wrote ? Seems like it does the same thing to me.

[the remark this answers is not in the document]
iurii-ssv: Can you re-review the PR then (and maybe update the "Confidence Score" accordingly) ?

iurii-ssv:
@greptile can you review again please ?

message/validation/consensus_validation.go
MatheusFranco99: About the case "decided with fewer signers than before" and "with the same amount but with different peers", I think this isn't rejectable (as it's not a clear malicious behaviour, e.g. msgs can get reordered). It would be rejectable only if you find the same signers but with diff decided data. For the cases I mentioned, I think ignore is better. Wdyt?

iurii-ssv: Hmm, yeah I didn't think about message-reordering ... that could result in those false-positive REJECTs,

it seems we should do something like this:
- always ACCEPT a message with greater amount of signers seen so far
- (for same amount of signers) REJECT only when we've seen a message with different signers from the same peer, IGNORE otherwise
- (for fewer signers) always IGNORE, regardless of what peer the message came from

but it seems like we don't gain much by implementing this ^ precisely (unless we also want to expand the 3rd rule to also REJECT duplicates of messages with fewer signers sent by the same peer ...), so I would suggest we should just always IGNORE instead ... @MatheusFranco99 wdyt ?

MatheusFranco99 (changes requested):
Great job!! Everything looks fine. The only issue is the comment you raised, for which I proposed a change above.

A side comment: the implementation deviates a bit from spec, but I found it a good change. As I remember, spec says something like "check **all** rules against the peer state, **then** check **all** rules against the world state". Instead, I see that the implementation already checks a rule against the world state just after it's checked against the peer state. This is good as: (i) indeed most rules don't need to be checked again (as they don't depend on state); (ii) there seems to be no need to check all peer rules before starting to check the world ones.
For alignment purposes, I think we can add this change to spec, and specify/mark all the rules that need the world check.
cc @GalRogozinski

message/validation/consensus_validation.go
MatheusFranco99: Hmm this seems delicate...

Ignores are fine whenever there's no maliciousness (that we can prove). However, ignoring a msg is less helpful for our system than rejecting it. For example, if a malicious peer is able to send `decided(signers=[1,2,3],data=x)` and keep changing `x = 1, x = 2, ...` and just get ignored, than it's easily able to spam our node. Ofc we have a limiter for amount/rate of ignores per peer, however we don't want to reach this point if we may prevent it, and, even further, the peer can try to avoid the limiter by sending just a bit less per time window then the limit.

Tbh, I think the best is to be super precise here. I would say even to keep the map `map(decided_signers_subset) -> sent or not`, and reject if the same subset is reached.

A further extension is that maybe `decided(signers=[1,2,3],data="A")` and `decided(signers=[1,2,3,4],data="B")` is rejectable, as `(1,2,3)` would need to sign both `"A"` and `"B"`. I think we should add it.

We don't really need to store every subset combination (which is big for `N=13`, and as I did in `map(decided_signers_subset)`). It's fine to have a linear space `map[signer]->data_hash`, and then do
```text
any_new = false
FOR signer in decided_msg.signers DO:
    IF map[signer] is not empty and map[signer] != decided_msg.data_hash THEN
        Reject
   ELSE IF map[signer] is empty THEN
        any_new = true
   ENDIF

IF not any_new THEN
    Reject (? correct here? I need to think more)
ENDIF

IF len(decided_msg.signers) > state.max_decided_size THEN
    Accept
ELSE
    Ignore
ENDIF
```

@GalRogozinski @iurii-ssv  what do you think?

GalRogozinski: I think this should be probably another PR since here we just wanted to solve the double running operator issue.

But to the point:
1. I don't understand the `any_new` reject.
2. The size comparison rule can also be a Reject, no?

Maybe let's concentrate on Friday on this matter @MatheusFranco99 as a separate task?

iurii-ssv: Okay, leaving the behavior for this error unchanged (always IGNORE) - https://github.com/ssvlabs/ssv/pull/2677/changes/2b6bb912224f26e9bd48b4d2f1453e73fbf3b81f

**Edit:** the behavior did actually change,
- previously: we would only IGNORE the exact message match
- in this PR: we'll IGNORE many more combinations = potentially reducing the overall p2p traffic (the more operators we have in the cluster, the more combinations we'll now ignore/drop for those decided messages)

MatheusFranco99 (dismissed):
Good work!!

message/validation/consensus_validation.go
```diff
@@ -213,32 +210,40 @@ func (mv *messageValidator) validateQBFTLogic(
...
-			// Rule: Decided msg can't have the same signers as previously sent before for the same duty
-			if signerState.SeenSigners != nil {
-				if _, ok := signerState.SeenSigners[quorum.ToBitMask()]; ok {
-					return ErrDecidedWithSameSigners
-				}
+			// For simplicity (because we are gonna get rid of decided messages soon anyway), we just check
+			// the world state here to see if we want to accept or ignore this message.
+			if len(signedSSVMessage.OperatorIDs) <= signerState.World.SeenDecidedMsgSignersCount {
+				e := ErrDecidedMessageWithTooFewSigners
+				return e
```
nkryuchkov, line 246:
```suggestion
				return ErrDecidedMessageWithTooFewSigners
```

nkryuchkov (dismissed):

message/validation/consensus_validation.go
iurii-ssv: I left this code like that so its shape is consistent/same/similar to cases when we also need to add the `e.reject = true`

message/validation/validation.go
```diff
@@ -315,16 +306,16 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
 	return newCommitteeInfo(share.CommitteeID(), operators, indices), nil
 }
 
-func (mv *messageValidator) validatorState(key peerIDWithMessageID, committee []spectypes.OperatorID) *ValidatorState {
-	if v := mv.state.Get(key); v != nil {
+func (mv *messageValidator) validatorState(key spectypes.MessageID, committee []spectypes.OperatorID) *ValidatorState {
```
olegshmuelov: For non-committee duties, if a validator is removed and re-added with a different operator set within the state TTL, `validatorState()` returns the stale cached entry (same `MessageID`, same cache key).
Since the cache is now keyed by `messageID` only, all peers share the same stale entry.

Two failure modes:
- **Larger committee** (e.g., 4->7): `signerIndex(newOp)` exceeds `len(cs.operators)` -> index-out-of-bounds panic in `OperatorState()`, no `recover()`.
- **Different operators, same size** (e.g., [1,2,3,4]->[5,6,7,8]): op5 gets index 0 in the new committee, but index 0 still holds op1's old state, so op5's first message is falsely rejected as a duplicate.

**Suggestion:** consider either storing the committee in `ValidatorState` and recreating state on mismatch, or evicting the cache on `ValidatorRemoved` events (currently no eviction happens, the cache relies solely on TTL expiration).

message/validation/validation_test.go
```diff
@@ -1472,7 +1472,7 @@ func Test_ValidateSSVMessage(t *testing.T) {
 		require.NoError(t, err)
 
 		_, err = validator.handleSignedSSVMessage(signedSSVMessage, topicID, peerID, receivedAt)
-		require.ErrorIs(t, err, ErrDecidedWithSameSigners)
+		require.ErrorIs(t, err, ErrDecidedMessageWithTooFewSigners)
```
olegshmuelov, line 1610: All duplicate-detection tests use a single `peerID`. The core new behavior,
different peer sends same message -> IGNORE (not REJECT) has no test coverage.

Consider adding tests that:
1. Send a message from Peer1 (accepted), then the same from Peer2 -> verify IGNORED (not rejected)
2. Send a message from Peer1 twice -> verify REJECTED
3. Verify `reject == true` for same-peer and `reject == false` for different-peer duplicates

message/validation/consensus_validation.go
```diff
@@ -334,28 +340,31 @@ func (mv *messageValidator) updateConsensusState(
...
+		signerState.World.HashedProposalData = &rootCopy
 	}
 
 	signerCount := len(signedSSVMessage.OperatorIDs)
-	if signerCount > 1 {
-		if signerState.SeenSigners == nil {
-			signerState.SeenSigners = make(map[SignersBitMask]struct{}) // lazy init on demand to reduce mem consumption
-		}
+	if signerCount > 1 { // checks if this is a "decided" message
+		signerState.Peer(receivedFrom).SeenDecidedMsgSignersCount = signerCount
```
olegshmuelov, line 354: nit: `Peer(receivedFrom).SeenDecidedMsgSignersCount` is written here but never read

message/validation/signer_state.go
y0sher, line 32: shouldn't we at some cases, not even create state for a peer? otherwise we could be spammed to OOM with new identities? @MatheusFranco99 @iurii-ssv @GalRogozinski ?

message/validation/validation_test.go
y0sher, line 1: shouldn't we add some specific tests for a duplicate message acting different if it same sender or not?

message/validation/signer_state.go
y0sher: maybe we can only check for existence not create new state in some of the cases and that would be enough not be spammed.

MatheusFranco99: I think it's always good if we can do so, but I wonder about these cases in which we don't need to store the state for a peer.

About this memory mgnt, I agree we could do some cleanup even because this state should be very transient. Once the duty is done we can delete it.

But I don't know how to differentiate legitimate peers from bad ones. One mechanism gossipsub has against sybil attacks is the IP limiter, but I agree this is not sound. In any case, sybil attacks are not easily solvable :S

iurii-ssv: This "issue" btw applies to our current implementation (main/stage branch) too,

but I'm not sure it is an "issue" because the `peerID` we use here is = "our immediate neighbor we received the message from":
- there is a limited amount of those immediate neighbors we connect at any given time (up to `MaxPeers` or something like that)
- those swap out slowly over time ... we basically need to connect them(peers) first before we can receive a p2p pubsub message ... so it's rate limited in this way

if the `peerID` here was the original message initiator - then your concern would be something to worry about, @y0sher, I think

Another considerations is - should we clean the map to prevent infinite growth ? And to me it doesn't seems like this is worth the code-complexity given SSV node has on the order of 100 peers at any given time, and those rotate several peers per minute (even if under "attack" for days, the size of the map shouldn't be much larger than ~10000-100000 entries).

message/validation/consensus_validation.go
iurii-ssv: Right, this is because we simplified [this rule we discussed with Matheus here](https://github.com/ssvlabs/ssv/pull/2677#discussion_r2759396557) ... and as the result we do not need to query peer-specific state in the current implementation,

but if we were to "implement it properly" then we'd need to query this data, hence I think recording it regardless is a right/good/correct thing to do,

but it's a good catch for me to double-check anyway 👍

message/validation/validation.go
iurii-ssv: Interesting, this however also seems to apply for our current implementation (stage/main branches) too ?

But you are right, validator always belongs in a specific committee and `states *ttlcache.Cache[spectypes.MessageID, *ValidatorState]` doesn't know anything about that ... I think the cleanest way to solve this would be to modify `states` map **key** to be `spectypes.MessageID+spectypes.CommitteeID` (not just `spectypes.MessageID`), @olegshmuelov wdyt ?

cc @MatheusFranco99 - even though this is probably an implementation detail, we'd need your review/approval on the approach chosen too

olegshmuelov: You're right that this is pre-existing. This PR makes it worse though, previously keyed by
`(peerID, messageID)`, so only the same peer hit the stale entry. Now with `messageID` only,
all peers share it.

A potentially lighter alternative to `MessageID + CommitteeID` is to add a `committee` field to `ValidatorState` and compare on cache hit:

something like
```diff
 func (mv *messageValidator) validatorState(key spectypes.MessageID, committee []spectypes.OperatorID) *ValidatorState {
 	if v := mv.states.Get(key); v != nil {
-		return v.Value()
+		if slices.Equal(v.Value().committee, committee) {
+			return v.Value()
+		}
 	}
```
It avoids changing the key type and the SHA256 per call.
The committee comes from the validator store (on-chain state), so recreation can only be
triggered by legitimate contract operations.

iurii-ssv: > A potentially lighter alternative to `MessageID + CommitteeID` is to add a `committee` field to `ValidatorState` and compare on cache hit

This is indeed more light-weight, but I wonder if we could get in trouble when a delayed message with an old committee on it arrives **after** we've received some messages with new committee on them (so we'll basically delete the state with new committee ... and probably will re-create it again once more messages with new committee on them arrive),

that's why I thought making the key explicitly mention the target committee would be the way to go

since we are storing entries in `states *ttlcache.Cache[spectypes.MessageID, *ValidatorState]` map for ~ 1 epoch the memory usage shouldn't be an issue ... so I'd rather go with a bit more explicit/robust implementation all else being equal ?

> It avoids changing the key type and the SHA256 per call

I think we can re-use the `CommitteeInfo.committeeID` field that's populated already (no need to re-compute SHA256)

olegshmuelov: I think the compare-on-hit approach might actually be cleaner here
it's ~5 lines vs a new key type + signature changes across multiple files. It also replaces the
stale entry in-place rather than leaving a ghost entry until TTL.

The delayed-message concern shouldn't apply, the committee always comes from the
validator store at validation time, not from the message itself. After re-add, all messages get the current committee, and old-committee operators are rejected at `belongsToCommittee` before reaching `validatorState()`.

That said, both work correctly. Happy to go with either if you feel strongly about the key-based approach.

message/validation/validation_test.go
iurii-ssv: For this specific `ErrDecidedMessageWithTooFewSigners` case we actually always IGNORE now because [we are gonna remove "decided messages" soon anyway and it's simpler to always IGNORE](https://github.com/ssvlabs/ssv/pull/2677#discussion_r2759396557)

as for testing methodology overall - you are right, we should cover those added scenarios where we now differentiate between IGNORE and REJECT on a more granular basis - https://github.com/ssvlabs/ssv/pull/2677/changes/c40f2beb459a44ffb9507927415cf2b390260ff9

message/validation/validation.go
iurii-ssv: Ah, true, that sounds good ^

Added this optimistic check you've suggested above like this: https://github.com/ssvlabs/ssv/pull/2677/changes/22dc4e4da4dc9fade4fde21cc3d4d82cbd8290b8

message/validation/validation_test.go
iurii-ssv: Added unit-tests here - https://github.com/ssvlabs/ssv/pull/2677#discussion_r2782180454

olegshmuelov (approved):

message/validation/signer_state.go
y0sher: ok sounds fair. @GalRogozinski we might want to add this test to some attack simulator

y0sher (approved):

PR #2678, opened 2026-02-04
feat(beacon): detect and recover from stale attestation data

olegshmuelov:
## Summary

Adds detection and automatic recovery for stale attestation data returned by beacon nodes. When the beacon node returns attestation data with a head root that doesn't match the latest HeadEvent, the client waits briefly and refetches to get fresh data.

**Key features:**
- **Attestation data verification**: Compares `beacon_block_root` against cached HeadEvent, refetches on mismatch
- **Proposal parent verification**: Observability-only metrics for detecting parent root mismatches (no refetch)
- **Time-budget aware**: Skips retry if insufficient time remains before deadline

## How it works

1. On `GetAttestationData`, fetch from beacon node
2. Compare returned `beacon_block_root` against cached HeadEvent for that slot
3. If mismatch: wait 50ms (let BN cache update), then refetch
4. If still mismatch: return refetched data anyway (possible reorg)
5. Metrics track each outcome for monitoring

## Metrics Added

**Attestation (with recovery):**
| Metric | Description |
|--------|-------------|
| `ssv_cl_attestation_data_head_verify_total` | Verification attempts |
| `ssv_cl_attestation_data_head_match_total` | Head matched cache |
| `ssv_cl_attestation_data_head_mismatch_total` | Stale data detected |
| `ssv_cl_attestation_data_head_cache_miss_total` | No cached head for slot |
| `ssv_cl_attestation_data_refetch_success_total` | Refetch got correct head |
| `ssv_cl_attestation_data_refetch_failed_total` | Refetch error/timeout |
| `ssv_cl_attestation_data_refetch_still_mismatch_total` | Possible reorg |
| `ssv_cl_attestation_data_refetch_skipped_total` | Insufficient time |

**Proposal (observability only):**
| Metric | Description |
|--------|-------------|
| `ssv_cl_proposal_parent_verify_total` | Verification attempts |
| `ssv_cl_proposal_parent_match_total` | Parent matched cache |
| `ssv_cl_proposal_parent_mismatch_total` | Parent mismatch detected |
| `ssv_cl_proposal_parent_cache_miss_total` | No cached head for parent slot |

## Validation

Tested on stage environment with ~9200 validators for 5 days:
- 6 mismatches detected → 6 successful refetches (100% recovery)
- No negative side effects observed

## Test plan
- [x] Unit tests for all verification paths
- [x] Stage environment validation (3 days)
- [ ] Monitor metrics in production for `head_mismatch` and `refetch_*` counters


beacon/goclient/attest.go
```diff
@@ -55,36 +65,26 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
 
 // GetAttestationData returns attestation data for a given slot.
 // Multiple calls for the same slot are joined into a single request, after which
-// the result is cached for a short duration, deep copied and returned
+// the result is cached for a short duration, deep copied and returned.
+// Verifies the returned head against cached HeadEvent root and re-fetches if stale.
```
iurii-ssv: Maybe to expand/clarify this comment a bit: `Verifies the returned ...` -> `It also verifies the returned ...`

beacon/goclient/attest.go
```diff
@@ -93,6 +93,95 @@ func (gc *GoClient) GetAttestationData(ctx context.Context, slot phase0.Slot) (*
...
+// If mismatch detected, waits briefly then re-fetches.
+// Returns (attestationData, shouldCache).
+func (gc *GoClient) verifyAndRefetchIfStale(
+	ctx context.Context,
+	slot phase0.Slot,
+	attData *phase0.AttestationData,
+) (result *phase0.AttestationData, shouldCache bool) {
+	if gc.headCache == nil {
+		return attData, true
+	}
```
iurii-ssv: Seems like this nil-check is redundant (we always want to use this feature) ?

beacon/goclient/attest.go
```diff
@@ -93,6 +93,95 @@ func (gc *GoClient) GetAttestationData(ctx context.Context, slot phase0.Slot) (*
...
+}
+
+// verifyAndRefetchIfStale checks attestation data against cached head root.
+// If mismatch detected, waits briefly then re-fetches.
+// Returns (attestationData, shouldCache).
+func (gc *GoClient) verifyAndRefetchIfStale(
+	ctx context.Context,
+	slot phase0.Slot,
+	attData *phase0.AttestationData,
+) (result *phase0.AttestationData, shouldCache bool) {
```
iurii-ssv: This `verifyAndRefetchIfStale` func doesn't know anything about caching (it's responsible for fetching & checking "stale or not" ... it does consult another cache `headCache`, but that's a different one), 

it would be simpler if it just returned `stale bool` (instead of `shouldCache bool`) - so the caller decides when to cache or not

beacon/goclient/attest.go
```diff
@@ -93,6 +93,95 @@ func (gc *GoClient) GetAttestationData(ctx context.Context, slot phase0.Slot) (*
...
+	case <-time.After(refetchDelay):
+	}
+
+	refetchCtx, cancel := context.WithTimeout(ctx, refetchTimeout)
+	defer cancel()
+
+	fetchFn := gc.fetchAttestationData
+	if gc.fetchAttestationDataFunc != nil {
+		fetchFn = gc.fetchAttestationDataFunc
+	}
```
iurii-ssv: Nit: it's probably cleaner to initialize `fetchAttestationDataFunc` in the constructor (to a default value, if there isn't any override) ... also because if we want to call this func from another place we won't need to duplicate this kind of handling there

beacon/goclient/events.go
```diff
@@ -191,6 +191,11 @@ func (gc *GoClient) newEventHandler() func(*apiv1.Event) {
 			lastProcessedEventSlot = eventData.Slot
 			lastProcessedEventSlotLock.Unlock()
 
+			// Must be after monotonicity gate to ensure we only cache the latest head.
+			if gc.headCache != nil {
+				gc.headCache.Set(eventData.Slot, eventData.Block, ttlcache.DefaultTTL)
+			}
```
iurii-ssv: Also here nil-check seems redundant (it's probably easier to spot that we "forgot to initialize it" with just a `panic`)

beacon/goclient/proposer.go
```diff
@@ -321,6 +324,49 @@ func (gc *GoClient) scoreProposal(
...
+// This is observability only - no re-fetch, just metrics and logging.
+func (gc *GoClient) verifyProposalParent(
+	ctx context.Context,
+	logger *zap.Logger,
+	slot phase0.Slot,
+	proposal *api.VersionedProposal,
+) {
+	if gc.headCache == nil {
+		return
+	}
```
iurii-ssv: ditto

beacon/goclient/attest.go
```diff
@@ -19,6 +19,16 @@ import (
...
 
+const (
+	// refetchDelay gives the BN time to update its cache after HeadEvent.
+	// Value is a heuristic; monitor refetch_success vs refetch_still_mismatch.
+	refetchDelay = 50 * time.Millisecond
+	// refetchTimeout bounds re-fetch to preserve QBFT consensus budget.
+	refetchTimeout = 200 * time.Millisecond
+	// minTimeForRetry is minimum time needed before deadline to attempt retry.
+	minTimeForRetry = refetchDelay + refetchTimeout
+)
```
iurii-ssv, line 30: One thing to consider is: 
- for WAD-enabled setups, we have those soft-timeout / hard-timeout values that are currently set to 2.5s and 5s respectively - we probably could re-distribute this time more evenly between 1st-fest and re-fetch this PR implements ? Maybe not 1:1 but something like 2:1 or 3:1 (not > 10:1)
- for non-WAD case, it seemingly is using `commonTimeout` (which is 5s) as its hard-timeout ... previously that didn't really matter (there was no point in having a hard-timeout since duty won't work without attestation data), but now we are adding the re-fetch - and since we do, similar to the WAD case, it also probably makes more sense to allocate the time more evenly between 1st-fest and re-fetch

iurii-ssv (commented):
LGTM, some minor suggestions from me

beacon/goclient/attest.go
olegshmuelov: Good point. I increased the delay from 50ms to 100ms to give BNs more time to update their cache after processing the new block.
b96203e0e076a4184aae99d31cd36b3f97f0e400

iurii-ssv: Do you think `refetchTimeout` is good at `200ms` ? 

Wouldn't it be better to set it to `600ms` for example while also lowering the other timeouts (these are defined somewhat differently for WAD, and for non-WAD cases) by `400ms` ? That is to change soft-timeout = 2100ms (instead of 2500ms), and hard-timeout = 4600ms (instead of 5000ms). Or something like that.

olegshmuelov: Bumped `refetchTimeout` to 500ms (2.5x more headroom for BN response under load).

I kept first-fetch timeouts unchanged since the refetch runs under the epoch-wide duty context (not the WAD context), so there's no time pressure to redistribute. The mismatch is rare enough (~6 events/5 days) that the extra worst-case ~300ms is negligible. Happy to revisit if monitoring shows otherwise.

iurii-ssv: 500ms is better than 200ms I think :)

iurii-ssv (dismissed):
Nice!

iurii-ssv (approved):

beacon/goclient/proposer.go
```diff
@@ -321,6 +324,45 @@ func (gc *GoClient) scoreProposal(
 	return score
 }
 
+// verifyProposalParent checks the proposal's parent root against cached HeadEvent.
+// This is observability only - no re-fetch, just metrics and logging.
+func (gc *GoClient) verifyProposalParent(
+	ctx context.Context,
```
MatheusFranco99, line 330: Has any occurrence been observed?

beacon/goclient/attest.go
```diff
@@ -93,6 +93,86 @@ func (gc *GoClient) GetAttestationData(ctx context.Context, slot phase0.Slot) (*
...
+	if item == nil {
+		attestationDataHeadCacheMissCounter.Add(ctx, 1)
+		return attData, false
+	}
+	expectedRoot := item.Value()
+
+	if attData.BeaconBlockRoot == expectedRoot {
+		attestationDataHeadMatchCounter.Add(ctx, 1)
+		return attData, false
+	}
```
MatheusFranco99, line 124: It may be worth separately detecting the case in which the data received matched the head of slot `slot-1`, to understand if the BN missed by one single block, or if it's "totally wrong" about the data.
What do you think?

beacon/goclient/attest.go
```diff
@@ -19,6 +19,16 @@ import (
 	"github.com/ssvlabs/ssv/observability/log/fields"
 )
 
+const (
+	// refetchDelay gives the BN time to update its cache after HeadEvent.
+	// Value is a heuristic; monitor refetch_success vs refetch_still_mismatch.
+	refetchDelay = 100 * time.Millisecond
```
MatheusFranco99, line 25: This value was obtained from this monitoring? Or is this monitoring a TODO?

beacon/goclient/proposer.go
olegshmuelov: Not yet, proposal parent verification is observability-only so we're collecting data.
Proposals are rare enough on the current deployment (~9200 validators) that we haven't had meaningful coverage. We'd need to run this across all stage nodes to increase the chance of catching these scenarios.
Once we have enough data we'll decide if proposal recovery is needed.

beacon/goclient/attest.go
olegshmuelov: Good idea. For now the mismatch logs already include both `expected_root` and `got_root`, so we can cross-reference with HeadEvent logs to determine if the BN was one block behind. If the mismatch rate turns out to be higher in production I'll add an explicit `slot-1` comparison as a follow-up.

beacon/goclient/attest.go
olegshmuelov: The 100ms value was chosen to give beacon nodes enough time to update their internal cache after processing a new block. Validated in staging with ~9200 validators - 100% recovery rate on all observed mismatches.

Once merged, all stage environments will run with this fix so we can gather broader metrics data (across different BN setups and load conditions) and tune the timing parameters if needed.

iurii-ssv (dismissed):

MatheusFranco99 (dismissed):

beacon/goclient/attest.go
```diff
@@ -93,6 +93,86 @@ func (gc *GoClient) GetAttestationData(ctx context.Context, slot phase0.Slot) (*
 	return result, spec.DataVersionPhase0, nil
 }
```
momosh-ssv, line 94: What do you think about switching to `time.NewTimer` + `defer timer.Stop()` here? If `ctx.Done()` fires first, the timer from `time.After` leaks until it expires. Should be the idiomatic Go pattern for timers in selects.

beacon/goclient/proposer.go
```diff
@@ -321,6 +324,45 @@ func (gc *GoClient) scoreProposal(
...
+	logger *zap.Logger,
+	slot phase0.Slot,
+	proposal *api.VersionedProposal,
```
momosh-ssv, line 333: Should we guard against `slot == 0` here? `phase0.Slot` is uint64, so `0 - 1` wraps to MaxUint64. It would be a harmless cache miss I think, but an early return would make the intent clearer.

momosh-ssv (commented):
Two minor observations — neither blocking.

1. **`time.After` timer leak** (`attest.go:144-148`) — if `ctx.Done()` wins the select, the timer from `time.After(refetchDelay)` is not stopped and leaks until it fires. Not sure if this is negligible.

2. **`parentSlot := slot - 1` uint64 underflow** (`proposer.go:344`) — `phase0.Slot` is uint64, so slot 0 wraps to MaxUint64. Probably benign, but a guard would be cleaner.

momosh-ssv (dismissed):

beacon/goclient/attest.go
iurii-ssv: > the timer from time.After leaks until it expires

I thought this has been improved in some recent versions of Golang (making the firing not necessary for the timer to get GCed), but either way to me using `time.After` seems like a cleaner way (I would "optimize" that only if it's really needed)

beacon/goclient/proposer.go
olegshmuelov: Slot 0 is the genesis slot, the block is hardcoded by the spec, not proposed by a validator. No beacon node will ever issue a proposer duty for it, so this path is unreachable with slot == 0.

beacon/goclient/attest.go
```diff
@@ -55,36 +65,26 @@ func (gc *GoClient) AttesterDuties(ctx context.Context, epoch phase0.Epoch, vali
...
-			}
+
+		attData, err := gc.fetchAttestationData(ctx, slot)
+		if err != nil {
+			return nil, err
 		}
 
-		// Caching resulting value here (as part of inflight request) guarantees only 1 request
-		// will ever be done for a given slot.
-		gc.attestationDataCache.Set(slot, attestationData, ttlcache.DefaultTTL)
```
jose-ssv-labs, line 85: When `HeadEvent` for slot N hasn't arrived yet, `verifyAndRefetchIfStale` returns `(attData, stale=false)`. Then this code caches the result, right?

Now, what if `GetAttestationData` is called before `HeadEvent` for this slot has arrived (this is possible given that  the 95 percentile of block propagation is around ~1.5-3s from slot start). Isn't possible that the potentially stale result gets cached and all subsequent callers for that slot hit the attestation cache and bypass `verifyAndRefetchIfStale`?

jose-ssv-labs (commented):
In general, this PR looks good to me. After `GetAttestationData` returns data with stale head, it detects the mismatch against the cached HeadEvent and waits 100ms for the BN's internal attestation data cache to refresh before refetching. For what I see, the `refetchDelay=100ms` is appropriate for this specific purpose. It is not waiting for block propagation, just for the BN cache to catch up. 

A broader solution that handles [SIP 83](https://github.com/ssvlabs/SIPs/discussions/83) is still needed, but probably outside this PR.

beacon/goclient/attest.go
olegshmuelov: Good catch. When `HeadEvent` hasn't arrived, the unverified result gets cached and the aggregator at 2/3 slot hits stale data, `AggregateAttestation` finds no matching aggregate since the network's pool is keyed by the correct root.

Fixed: returning `stale=true` on head-cache miss skips caching. Attesters are unaffected (singleflight), aggregator re-fetches at T=8s with verification.

iurii-ssv (approved):

jose-ssv-labs (approved):

ljuba-ssv (approved):
Looks good — solid detection logic, timing heuristics are reasonable, existing review threads cover the important points. Two minor nits:

1. **Missing tests for `verifyProposalParent`** — would be straightforward to add (match, cache miss, mismatch, ParentRoot error paths). Not blocking.

2. **`Warn` → `Info` for proposal parent mismatch** (proposer.go:356) — since this is observability-only with no corrective action, `Warn` may cause alert fatigue. `Info` seems more appropriate here.

GalRogozinski (approved):

Roy-blox:
## QA Verification — PASS ✅

Tested on stage-hoodi (stage branch, operators 1/3/4) via metrics and e2m.

**Key numbers (last 1h rates):**
- Stale detected on ~0.6% of attestation fetches
- 92% of stale cases recovered successfully via refetch
- `refetch_skipped` / `refetch_failed` / `refetch_still_mismatch`: negligible (1 each lifetime)
- 14% `head_cache_miss` — HeadEvent timing race, expected and handled correctly

**Attestation effectiveness:** stable at 97–99% post-merge with no regression attributable to this change.

Tracked on board: ssvlabs/ssv-node-board#1033

PR #2679, opened 2026-02-06
Boole domain

nkryuchkov:
~~The PR must be rebased onto the `boole-fork` branch after both https://github.com/ssvlabs/ssv/pull/2302 and https://github.com/ssvlabs/ssv/pull/2503 are merged~~
~~The PR must be rebased onto the `boole-fork` branch after https://github.com/ssvlabs/ssv/pull/2503 is merged~~


protocol/v2/ssv/runner/aggregator.go
```diff
@@ -234,7 +234,8 @@ func (r *AggregatorRunner) ProcessConsensus(ctx context.Context, logger *zap.Log
 		Messages: []*spectypes.PartialSignatureMessage{msg},
 	}
 
-	msgID := spectypes.NewMsgID(r.BaseRunner.NetworkConfig.DomainType, r.GetShare().ValidatorPubKey[:], r.BaseRunner.RunnerRoleType)
+	domain := r.BaseRunner.NetworkConfig.DomainTypeAtSlot(decidedValue.Duty.Slot)
+	msgID := spectypes.NewMsgID(domain, r.GetShare().ValidatorPubKey[:], r.BaseRunner.RunnerRoleType)
```
y0sher, line 238: should we send old aggregator runner messages using the new domain type after fork? I'm not sure about the change here..

nkryuchkov: It shouldn't be possible because we won't have any aggregator runner duties during Boole epochs. We could have used `r.BaseRunner.NetworkConfig.DomainType` but I like the current code better, because it keeps the runner choice logic abstract and independent of the runner code, otherwise it would look like a hardcoded value

network/p2p/p2p_setup.go
```diff
@@ -218,10 +218,25 @@ func (n *p2pNetwork) setupPeerServices() error {
...
+				allowedDomains = append(allowedDomains, nextDomainString)
+			}
+		}
+
+		var networkFilter connections.HandshakeFilter
+		if len(allowedDomains) == 1 {
+			networkFilter = connections.NetworkIDFilter(allowedDomains[0])
+		} else {
+			networkFilter = connections.NetworkIDFilterAny(allowedDomains...)
+		}
```
iurii-ssv: Nit: I think it would be cleaner to have just 1 `NetworkIDFilterAny` func that accepts `allowedDomains...` and checks for its `len(allowedDomains)` inside of it

networkconfig/network.go
```diff
@@ -32,6 +33,23 @@ func (n Network) StorageName() string {
...
+func (n Network) CurrentDomainType() spectypes.DomainType {
+	return n.DomainTypeAtSlot(n.EstimatedCurrentSlot())
+}
+
+func (n Network) NextDomainTypeAtSlot(slot phase0.Slot) spectypes.DomainType {
+	if n.BooleForkAtSlot(slot) {
+		return n.DomainTypeAtSlot(slot)
+	}
+	return n.NextDomainType
+}
```
iurii-ssv: I think `NextDomainTypeAtSlot` always returns `n.NextDomainType` (seems unit-tests don't fully cover all edge-cases) ... it should probably be something like this instead ?
```go
func (n Network) NextDomainTypeAtSlot(slot phase0.Slot) spectypes.DomainType {
	if n.BooleForkAtSlot(slot) || n.InBooleTransitionWindow(slot) {
		return n.NextDomainType
	}
	return n.DomainType
}
```

I would also add a documenting comment for "what exactly this func is for".

Also, once we are **past the Boole-fork** we'll have both `currentDomain` and `nextDomain` equal (both set to Boole) ... which is probably gonna work fine, but maybe it's better to have `nextDomain` to be explicitly set to `spectypes.DomainType{}` after transition window (and handle it explicitly wherever the handling is needed).

iurii-ssv (commented):
LGTM, with maybe a minor adjustment

networkconfig/network.go
nkryuchkov: The next domain represents the next fork domain, so it's always the same, because we don't have any next scheduled forks after Boole. I think that zeroing it out after the fork is error-prone and doesn't give us any benefits. Current domain equal to next domain after fork looks fine to me, because next domain will become redundant after fork (we already used to have next domain once, and then we removed it).

As for issues that might be caused by time drift, we allow the next domain for `InBooleTransitionWindow` during a handshake: https://github.com/ssvlabs/ssv/blob/boole-domain/network/p2p/p2p_setup.go#L224

So I'd keep it as it is now

iurii-ssv: > The next domain represents the next fork domain ...

Yeah, it's fine as is (I just don't like to rely on "implicit" behavior even if it works well).

But my comment is more about how this function is implemented, currently it is equivalent to just this (which seems strange to me) ?
```go
func (n Network) NextDomainTypeAtSlot(slot phase0.Slot) spectypes.DomainType {
	return n.NextDomainType
}
```

nkryuchkov: @iurii-ssv Yeah, I didn't think it through well enough after refactoring. Perhaps we can delete this method

nkryuchkov: removed

network/discovery/dv5_service.go
```diff
@@ -235,26 +235,20 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 		return false
 	}
 
-	nextDomain := dvs.nextDomainType()
-	return nodeDomainType == nextDomain || nodeNextDomainType == nextDomain
+	// During transition, accept either configured fork domain for compatibility.
+	return nodeDomainType == dvs.netCfg.DomainType ||
+		nodeNextDomainType == dvs.netCfg.DomainType ||
+		nodeDomainType == dvs.netCfg.NextDomainType ||
+		nodeNextDomainType == dvs.netCfg.NextDomainType
```
iurii-ssv: I wonder if we could simplify the **current**/**next** domain handling to something like this ?
- publisher will only publish the **current** domain
- the receiver would check that ^ against its **current**/**next** domain by doing something like:
```go
func (dvs *DiscV5Service) isDomainCompatible(nodeDomainType spectypes.DomainType) bool {
	if nodeDomainType == dvs.netCfg.CurrentDomainType() {
		return true
	}

	if dvs.netCfg.InBooleTransitionWindow(dvs.netCfg.EstimatedCurrentSlot()) {
		return nodeDomainType == dvs.netCfg.DomainType || nodeDomainType == dvs.netCfg.NextDomainType
	}

	return false
}
```

iurii-ssv (approved):
Looks good, but maybe we could [simplify the current/next domain handling a bit](https://github.com/ssvlabs/ssv/pull/2679#discussion_r2822708204)

network/discovery/dv5_service.go
nkryuchkov: @iurii-ssv probably, it might be good to check `nodeNextDomainType == dvs.netCfg.DomainType || nodeDomainType == dvs.netCfg.NextDomainType` too to avoid any clock skew. I think it's still possible to be in the fork and get a handshake from a node that was sent before the fork

iurii-ssv: I'm thinking: the `InBooleTransitionWindow` is designed to take care of the clock-skews,

> I think it's still possible to be in the fork and get a handshake from a node that was sent before the fork

so once our node is past Boole-fork we'll have 1 slot of time accepting peers who broadcast the older domainType ... after that 1 slot passes those peers must switch to broadcasting the nextDomainType. It is possible that we'll ignore some peers during the transition because of p2p delays, but it doesn't seem like it should cause any real issues (while the simpler code is always preferable).

But maybe it's not worth changing it if you already have tested the current solution in this PR to work well.

iurii-ssv: Thinking more about it, maybe `InBooleTransitionWindow` isn't even the "right" window to use here,

what we really need is to wait a sufficient amount of time (however long it takes for all SSV nodes to start publishing/advertising themselves with the `nextDomainType` value) after Boole-fork epoch starts ... I don't think 1 slot is enough

^ we'd need that ^ if we were to simplify as I suggested above, your current solution in this PR probably works already regardless

iurii-ssv (approved):

network/discovery/dv5_service.go
nkryuchkov: Well, actually, we QA'd the version you suggested. I decided to extend it, but thinking more about it, it's likely redundant. I simplified it

operator/validator/controller.go
```diff
@@ -1043,8 +1043,6 @@ func SetupCommitteeRunners(
 ) validator.CommitteeRunnerFunc {
 	buildController := func(role spectypes.RunnerRole) *qbftcontroller.Controller {
 		config := &qbft.Config{
-			BeaconSigner: options.Signer,
-			Domain:       options.NetworkConfig.DomainType,
```
olegshmuelov, line 1047: Now that `BeaconSigner` and `Domain` are no longer set, the fields, their getters (`GetShareSigner`, `GetSignatureDomainType`), and the `signing` interface in `config.go` are dead code

network/discovery/dv5_service.go
```diff
@@ -233,6 +235,18 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 }
 
+func (dvs *DiscV5Service) isDomainCompatible(nodeDomainType spectypes.DomainType, nodeNextDomainType spectypes.DomainType) bool {
+	if nodeDomainType == dvs.netCfg.CurrentDomainType() {
+		return true
+	}
+
+	if !dvs.netCfg.InBooleTransitionWindow(dvs.netCfg.EstimatedCurrentSlot()) {
+		return false
+	}
```
olegshmuelov, line 245: After the 1-slot subsequent window, peers whose cached ENR still shows `DomainType=Alan` are rejected even if `NextDomainType=Boole` signals fork compatibility.

`PublishENR` fires at fork time via subnet changes, but discv5 ENR propagation isn't instant - peers discovered before the fork can have stale cached ENRs beyond the 1-slot window.

Consider also accepting `nodeNextDomainType == CurrentDomainType()` outside the transition window:

```go
if nodeDomainType == dvs.netCfg.CurrentDomainType() || nodeNextDomainType == dvs.netCfg.CurrentDomainType() {
    return true
}
```

message/validation/signed_ssv_message.go
```diff
@@ -121,15 +119,6 @@ func (mv *messageValidator) validateSSVMessage(ssvMessage *spectypes.SSVMessage)
 		return e
 	}
 
-	// Rule: If domain is different then self domain
```
olegshmuelov, line 124: Removing this lets messages with random domains reach the per-`MessageID` validation locks and committee lookups before `validateDomainAtSlot` rejects them - potential amplification vector since each random domain creates a unique lock entry (TTL = 2 epochs).

Consider a cheap slot-free early gate:

```go
domain := ssvMessage.GetID().GetDomain()
if !bytes.Equal(domain, mv.netCfg.DomainType[:]) && !bytes.Equal(domain, mv.netCfg.NextDomainType[:]) {
    return ErrWrongDomain
}
```

olegshmuelov (approved):

nkryuchkov:
The current state of the branch has been passed to the QA team, as this branch is the only thing left to make the `boole-fork` branch work overall.

Any further changes to this PR are unwanted unless they are critical

y0sher (approved):

PR #2680, opened 2026-02-09
boole: use ssv config for topic name

nkryuchkov:
Use SSV config name that considers environment, e.g., `/ssv/hoodi/boole/1` => `/ssv/hoodi-stage/boole/1`


nkryuchkov:
@greptileai review this, considering it won't cause network partitioning, because the base branch isn't in use, other than testing

y0sher (approved):

PR #2682, opened 2026-02-09
aggregator committee: extend value checks

nkryuchkov:
On hold until it's implemented in spec due to broken spec tests

### DIFF
--- a/protocol/v2/ssv/runner/runner_validations.go
+++ b/protocol/v2/ssv/runner/runner_validations.go
@@ -10,6 +10,7 @@ import (
 	ssz "github.com/ferranbt/fastssz"
 	"github.com/pkg/errors"
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
+	"go.uber.org/zap"
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
@@ -18,6 +19,7 @@ import (
 
 func (b *BaseRunner) ValidatePreConsensusMsg(
 	ctx context.Context,
+	logger *zap.Logger,
 	runner Runner,
 	psigMsgs *spectypes.PartialSignatureMessages,
 ) error {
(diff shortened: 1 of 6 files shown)


nkryuchkov:
@olegshmuelov @GalRogozinski @MatheusFranco99 https://github.com/ssvlabs/ssv/pull/2503#discussion_r2783349025

protocol/v2/ssv/value_check.go
```diff
@@ -96,6 +133,46 @@ func (v *aggregatorCommitteeChecker) CheckValue(value []byte) error {
...
+			fmt.Errorf("no aggregators or sync committee contributors in consensus data"),
+		)
+	}
+
+	for _, agg := range cd.Aggregators {
+		allowedByValidator, ok := v.allowedAggregators[agg.ValidatorIndex]
+		if !ok {
+			return spectypes.NewError(
+				spectypes.QBFTValueInvalidErrorCode,
+				fmt.Sprintf("unexpected aggregator validator %d", agg.ValidatorIndex),
```
shane-moore, line 148: hi, bringing back the divergent views concern from the parent PR (#2503, https://github.com/ssvlabs/ssv/pull/2503#discussion_r2768138219 by @MatheusFranco99).                                                                               
                                                                                             
Since operators can have different views of the aggregator lists (i.e. different or failed BN responses, or some operators haven't processed a ValidatorAdded event yet), strict rejection here would stall consensus. The same issue applies to verifyExpectedSigningRoots I think -> an unknown validator's signing root causes the entire partial sig batch to be dropped, including valid sigs for validators the operator does know about.               

 Matheus's suggestion of only rejecting validators that are known to the operator but shouldn't have that duty rather than simply unknown seems like an ok tradeoff

nkryuchkov: @shane-moore Hi, thanks
@MatheusFranco99 Can we implement this in spec? I'll adjust the implementation afterwards

protocol/v2/ssv/runner/runner_validations.go
```diff
@@ -28,27 +30,40 @@ func (b *BaseRunner) ValidatePreConsensusMsg(
...
+		for _, root := range aggregatorMap {
+			expectedRoots[root] = struct{}{}
+		}
+		for _, indexMap := range contributionMap {
+			for _, root := range indexMap {
+				expectedRoots[root] = struct{}{}
+			}
+		}
+
+		return b.verifyExpectedSigningRoots(psigMsgs, expectedRoots)
```
GalRogozinski, line 58: I understand that you are trying to fend off attacks where a single validator can complete quorum on a single root, thus preventing selection proofs on all other roots?

We had a discussion on this with @liorrutenberg and it was decided that we [allow this attack](https://github.com/ssvlabs/SIPs/pull/61/changes/22a4f923ba0875f76e84b5afce6e525e7b2e2224) to allow operators to have divergent views of validators (in case of EL issues).

I just discussed with @MatheusFranco99 and maybe the correct solution is to actually wait on the preconsensus stage if no aggregator passes the selection proof

protocol/v2/ssv/runner/runner_validations.go
```diff
@@ -159,7 +179,31 @@ func (b *BaseRunner) ValidatePostConsensusMsg(ctx context.Context, runner Runner
...
+			for _, root := range aggregatorMap {
+				expectedRoots[root] = struct{}{}
+			}
+			for _, roots := range contributionMap {
+				for _, root := range roots {
+					expectedRoots[root] = struct{}{}
+				}
+			}
+
+			return b.verifyExpectedSigningRoots(psigMsgs, expectedRoots)
```
GalRogozinski, line 206: I think whatever conclusion we have on preconsensus validation will be also valid here

protocol/v2/ssv/value_check.go
```diff
@@ -96,6 +133,46 @@ func (v *aggregatorCommitteeChecker) CheckValue(value []byte) error {
...
+			)
+		}
+		subnetID := contrib.CommitteeIndex
+		if _, ok := allowedByValidator[subnetID]; !ok {
+			return spectypes.NewError(
+				spectypes.QBFTValueInvalidErrorCode,
+				fmt.Sprintf("unexpected contributor subnet %d for validator %d", subnetID, contrib.ValidatorIndex),
+			)
+		}
+	}
```
GalRogozinski, line 174: do you see any good motivation for having this check @MatheusFranco99?

We need more than `F` bad nodes for this to fail I think

GalRogozinski (changes requested):
I don't think the changes are correct so rejecting for now

protocol/v2/ssv/value_check.go
MatheusFranco99: Correct, we can't do validator id checks 👍 
Regarding not having a duty for the validator, I think this is naturally handled by spec code.
Still, I'm adding a PR for making this clear. Will ask reviews soon.

protocol/v2/ssv/runner/runner_validations.go
MatheusFranco99: Yes, I'm adding a PR for it.

protocol/v2/ssv/value_check.go
MatheusFranco99: Agree

PR #2683, opened 2026-02-10
aggregator committee: exporter

nkryuchkov:
Support aggregator committee role in exporter


exporter/store/store.go
```diff
@@ -431,17 +439,20 @@ func (s *DutyTraceStore) makeValidatorPrefix(slot phase0.Slot, role spectypes.Be
...
 	return prefix
 }
 
-func (s *DutyTraceStore) makeCommitteePrefix(slot phase0.Slot, id spectypes.CommitteeID) []byte {
-	prefix := make([]byte, 0, len(committeeDutyTraceKey)+4+32)
+func (s *DutyTraceStore) makeCommitteePrefix(slot phase0.Slot, role spectypes.RunnerRole, id spectypes.CommitteeID) []byte {
+	prefix := make([]byte, 0, len(committeeDutyTraceKey)+4+1+32)
 	prefix = append(prefix, []byte(committeeDutyTraceKey)...)
 	prefix = append(prefix, slotToByteSlice(slot)...)
+	prefix = append(prefix, byte(role&0xff))
```
iurii-ssv: This line could use a similar comment from above: `// Include runner role to avoid collisions between committee and aggregator-committee duties.`

operator/dutytracer/collector.go
```diff
@@ -400,15 +409,36 @@ func (c *Collector) processPartialSigCommittee(receivedAt uint64, msg *spectypes
 	var attIdxs []phase0.ValidatorIndex
 	var scIdxs []phase0.ValidatorIndex
 
+	if msg.Type == spectypes.AggregatorCommitteePartialSig && !trace.aggSelectionRootReady {
+		root, err := c.getAggregatorSelectionRoot(ctx, msg.Slot)
+		if err != nil {
+			c.logger.Debug("failed to compute aggregator selection root",
+				zap.Error(err),
+				fields.Slot(msg.Slot),
+				fields.CommitteeID(committeeID))
```
iurii-ssv: Should this be logged under Warn or Error ?

operator/dutytracer/collector.go
```diff
@@ -591,34 +694,50 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
-					c.logger.Error("CRITICAL: failed to compute role roots from proposal - pending signatures will be dropped",
-						zap.Error(err),
-						fields.Slot(slot),
-						fields.CommitteeID(committeeID),
-						zap.Int("pending_signature_count", pendingCount),
-						pendingDetails(trace.pendingByRoot))
+				case spectypes.RoleAggregatorCommittee:
+					rolesByRoot, err := c.computeAggregatorCommitteePostConsensusRoles(ctx, slot, msg.SignedSSVMessage.FullData)
+					if err != nil {
+						c.logger.Error("failed to compute aggregator committee roots from proposal",
```
iurii-ssv: We probably want to "align" this error message with the one above (it has "CRITICAL" in it while this one doesn't): `CRITICAL: failed to compute role roots from proposal - pending signatures will be dropped`

operator/dutytracer/collector.go
```diff
@@ -591,34 +694,50 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
+							fields.Slot(slot),
+							fields.CommitteeID(committeeID))
+					} else {
+						trace.aggPostConsensusRoles = rolesByRoot
+						trace.flushPending()
+						c.checkQuorumAfterFlush(c.logger, committeeID, slot, trace)
+					}
+				case spectypes.RoleProposer, spectypes.RoleValidatorRegistration, spectypes.RoleVoluntaryExit, spectypes.RoleUnknown:
+					c.logger.Warn("unexpected committee role on committee trace path", zap.Int32("role", int32(role)))
 				}
```
iurii-ssv: Seems like we should use `default` here instead of this `case` that's trying to list every role explicitly (because it would be simpler, and also we might not be validating `role` prior ... so it might have whatever value)

operator/dutytracer/collector.go
```diff
@@ -577,12 +680,12 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 		executorID := msgID.GetDutyExecutorID()
 
 		switch role := msgID.GetRoleType(); role {
-		case spectypes.RoleCommittee:
+		case spectypes.RoleCommittee, spectypes.RoleAggregatorCommittee:
```
iurii-ssv: Seems like we are checking if the `role` is either of `spectypes.RoleCommittee, spectypes.RoleAggregatorCommittee` is a bunch of places in this package, probably want to define a helper method for it (eg. `isCommiteeRole`) ?

operator/dutytracer/collector.go
```diff
@@ -486,6 +516,79 @@ func (c *Collector) getSyncCommitteeRoot(ctx context.Context, slot phase0.Slot,
...
+
+	dAgg, err := c.client.DomainData(ctx, epoch, spectypes.DomainAggregateAndProof)
+	if err != nil {
+		return nil, fmt.Errorf("get aggregate and proof domain data: %w", err)
+	}
+	for _, aggAndProof := range aggregateAndProofs {
+		hashRoot, err := spectypes.GetAggregateAndProofHashRoot(aggAndProof)
+		if err != nil {
+			continue
+		}
```
iurii-ssv, line 582: I'd maybe log these errors here (and similar) just in case because it might be hard to spot/debug the issue otherwise

api/handlers/exporter/common_model.go
```diff
@@ -184,6 +184,16 @@ func toBeaconRoles(rs []api.Role) []spectypes.BeaconRole {
 	return out
 }
 
+// toRunnerRoles converts an API RunnerRoleSlice into a slice of runner roles.
+// It is used by HTTP adapters when constructing core query models.
+func toRunnerRoles(rs []api.RunnerRole) []spectypes.RunnerRole {
```
julienh-ssv: I think the input type should be `api.RunnerRoleSlice` for consistency with `toValidatorIndices`

exporter/store/store.go
```diff
@@ -431,17 +439,20 @@ func (s *DutyTraceStore) makeValidatorPrefix(slot phase0.Slot, role spectypes.Be
...
 }
 
-func (s *DutyTraceStore) makeCommitteeSlotPrefix(slot phase0.Slot) []byte {
-	prefix := make([]byte, 0, len(committeeDutyTraceKey)+4)
+func (s *DutyTraceStore) makeCommitteeSlotRolePrefix(slot phase0.Slot, role spectypes.RunnerRole) []byte {
+	prefix := make([]byte, 0, len(committeeDutyTraceKey)+4+1)
 	prefix = append(prefix, []byte(committeeDutyTraceKey)...)
 	prefix = append(prefix, slotToByteSlice(slot)...)
+	// Include runner role to avoid collisions between committee and aggregator-committee duties.
+	prefix = append(prefix, byte(role&0xff))
```
julienh-ssv: IMHO Here we shouldn't depend on an external value to decide our internal storage representation. If the spec changes the value for some reason, our previously written value could become orphaned or collide with another.

It would be better to have a dedicated function to map roles to explicit, stable prefix bytes in this package e.g. 

```
func runnerRoleToPrefix(role spectypes.RunnerRole) (byte, error) {
	switch role {
	case spectypes.RoleCommittee:
		return 1, nil
	case spectypes.RoleAggregator:
		return 2, nil
	default:
		return 0, fmt.Errorf("unknown runner role: %v", role)
	}
}
```

exporter/store/store.go
julienh-ssv: Same as previous comment about `func runnerRoleToPrefix(role spectypes.RunnerRole) (byte, error)`

exporter/model.go
```diff
@@ -150,6 +150,7 @@ type CommitteeDutyTrace struct {
 	ConsensusTrace
 
 	Slot phase0.Slot
+	Role uint64
```
julienh-ssv: I think it would be better to use the spectype, here. Is there a reason to use a naked uint64? The other members are using spectypes.

iurii-ssv: I would leave a clarifying comment to explain that `CommitteeDutyTrace struct` might "represent" either of: Committee and AggregatorCommitee kinds of traces (specified/defined by `Role`)

I kinda feel like it would be better to define a separate struct for it (imagine if it wasn't that easy to represent both with the same struct) ... but we can do it later if necessary, I guess.

operator/dutytracer/store.go
```diff
@@ -215,25 +188,71 @@ func hasSignersForRoles(duty *exporter.CommitteeDutyTrace, roles ...spectypes.Be
...
+	var wantAggregator bool
+	for _, role := range roles {
+		switch role {
+		case spectypes.BNRoleAttester, spectypes.BNRoleSyncCommittee:
+			wantCommittee = true
+		case spectypes.BNRoleAggregator, spectypes.BNRoleSyncCommitteeContribution:
+			wantAggregator = true
+		case spectypes.BNRoleProposer, spectypes.BNRoleValidatorRegistration, spectypes.BNRoleVoluntaryExit, spectypes.BNRoleUnknown:
+			// Not committee-related beacon roles.
+		}
```
iurii-ssv: Can probably omit this last `case` clause here ?

operator/dutytracer/store.go
```diff
@@ -128,85 +128,58 @@ func (c *Collector) getValidatorDutyFromDiskIndex(role spectypes.BeaconRole, slo
...
-// since we don't store a boolean flag to separate duties by their role in the db
-// we rely on the fact that during collection we separate the signers in their
-// corresponding fields (Attester and SyncCommittee) based on the role
+// hasSignersForRoles checks if the duty has signers for the given beacon role(s).
+// Committee duties are keyed by runner role, but we still use the per-role signer
+// buckets to distinguish between attester vs aggregator and sync committee vs SCC.
 func hasSignersForRoles(duty *exporter.CommitteeDutyTrace, roles ...spectypes.BeaconRole) bool {
 	if len(roles) == 0 {
 		return true
 	}
```
iurii-ssv: Not 100% related to this PR, but I've noticed we `return true` whenever `len(roles) == 0` here ... this seems strange/unintuitive to me - is this the expected behavior ? If yes, can we document why it works like that ?

cc @julienh-ssv to double-check

iurii-ssv (commented):
LGTM if it works, left minor questions/suggestions

operator/dutytracer/store.go
```diff
@@ -170,6 +170,7 @@ func (c *Collector) GetCommitteeDuty(slot phase0.Slot, committeeID spectypes.Com
 // buckets to distinguish between attester vs aggregator and sync committee vs SCC.
 func hasSignersForRoles(duty *exporter.CommitteeDutyTrace, roles ...spectypes.BeaconRole) bool {
 	if len(roles) == 0 {
+		// No role filter means "match all committee duties".
 		return true
```
iurii-ssv: This kinda suggests maybe `hasSignersForRoles` is not the best place to do this `len(roles) == 0` check ?

operator/dutytracer/store.go
```diff
@@ -215,25 +189,71 @@ func hasSignersForRoles(duty *exporter.CommitteeDutyTrace, roles ...spectypes.Be
...
+	ctx := fmt.Sprintf("slot=%d role=%d committeeID=%x", slot, role, committeeID)
+	trace, err := c.store.GetCommitteeDuty(slot, role, committeeID)
 	if err != nil {
 		return nil, fmt.Errorf("get committee duty from disk (%s): %w", ctx, err)
 	}
 
 	return trace, nil
 }
 
+func runnerRolesForBeaconRoles(roles ...spectypes.BeaconRole) []spectypes.RunnerRole {
```
iurii-ssv: maybe rename to make it explicit: `runnerRolesForBeaconRoles` -> `commiteeRunnerRolesForBeaconRoles`

operator/dutytracer/collector.go
```diff
@@ -768,69 +778,67 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
-					var data = new(spectypes.ProposerConsensusData)
-					if err := data.Decode(msg.SignedSSVMessage.FullData); err == nil {
-						func() {
-							trace.Lock()
-							defer trace.Unlock()
+		// map pubkey to validator index for internal storage
+		index, found := c.validators.ValidatorIndex(validatorPK)
+		if !found {
+			c.logger.Error("validator not found by pubkey", fields.Validator(validatorPK[:]))
+			return fmt.Errorf("validator not found by pubkey: %x", validatorPK[:])
```
iurii-ssv: Do we need to log the error if we are also returning it ?

iurii-ssv (approved):
LGTM, with minor nits

iurii-ssv (approved):

iurii-ssv (approved):

nkryuchkov:
~~Blocked by https://github.com/ssvlabs/ssv/pull/2694~~

api/handlers/exporter/committee_model.go
```diff
@@ -108,6 +116,7 @@ func toCommitteeTrace(t *exporter.CommitteeDutyTrace) CommitteeTrace {
 	return CommitteeTrace{
 		// consensus trace
 		Slot:          uint64(t.Slot),
+		Role:          obsutils.FormatRunnerRole(t.Role),
```
y0sher, line 119: isn't this a breaking change? we should do some prep for the explorer team CC: @nir-ssvlabs

nkryuchkov: It's just a field in a JSON output, so it shouldn't break anything

operator/dutytracer/store.go
nkryuchkov: I changed this

operator/dutytracer/store.go
nkryuchkov: fixed

operator/dutytracer/collector.go
nkryuchkov: fixed

operator/dutytracer/collector.go
```diff
@@ -400,15 +409,36 @@ func (c *Collector) processPartialSigCommittee(receivedAt uint64, msg *spectypes
 	var attIdxs []phase0.ValidatorIndex
 	var scIdxs []phase0.ValidatorIndex
 
+	if msg.Type == spectypes.AggregatorCommitteePartialSig && !trace.aggSelectionRootReady {
+		root, err := c.getAggregatorSelectionRoot(ctx, msg.Slot)
```
olegshmuelov: `getAggregatorSelectionRoot` makes a `DomainData` RPC while the trace lock is held (line 882).

On success the result is cached via `aggSelectionRootReady`, but on failure the flag stays `false`, so every subsequent aggregator partial sig for this trace retries the RPC under lock.

Consider caching the root per-slot on the collector (outside the trace lock) or adding a retry backoff to avoid repeated RPC calls while beacon is temporarily down.

operator/dutytracer/eviction.go
```diff
@@ -79,16 +79,21 @@ func (c *Collector) dumpCommitteeToDBPeriodically(slot phase0.Slot) (totalSaved
...
-	if err := c.store.SaveCommitteeDuties(slot, duties); err != nil {
-		c.logger.Error("save committee duties to disk", zap.Error(err))
-		return 0
+	for role, duties := range dutiesByRole {
+		if len(duties) == 0 {
+			continue
+		}
+		if err := c.store.SaveCommitteeDuties(slot, role, duties); err != nil {
+			c.logger.Error("save committee duties to disk", zap.Error(err), fields.RunnerRole(role))
+			return 0
```
olegshmuelov: If saving one role fails after another role's duties were already persisted, this returns 0 instead of the partial count.

The DB data is fine but the metric/log count is wrong. Consider accumulating errors and returning the partial count:

```go
var saveErr error
for role, duties := range dutiesByRole {
    if len(duties) == 0 {
        continue
    }
    if err := c.store.SaveCommitteeDuties(slot, role, duties); err != nil {
        c.logger.Error("save committee duties to disk", zap.Error(err), fields.RunnerRole(role))
        saveErr = err
        continue
    }
    totalSaved += len(duties)
}
```

exporter/store/store_test.go
```diff
@@ -108,13 +108,13 @@ func TestSaveCommitteeDuties(t *testing.T) {
...
 
-	duty, err := store.GetCommitteeDuty(phase0.Slot(1), [32]byte{'a'})
+	duty, err := store.GetCommitteeDuty(phase0.Slot(1), spectypes.RoleCommittee, [32]byte{'a'})
 	require.NoError(t, err)
 	assert.True(t, committeeDutiesAreEqual(traces[0], duty))
 
-	duty, err = store.GetCommitteeDuty(phase0.Slot(1), [32]byte{'b'})
+	duty, err = store.GetCommitteeDuty(phase0.Slot(1), spectypes.RoleCommittee, [32]byte{'b'})
 	require.NoError(t, err)
 	assert.True(t, committeeDutiesAreEqual(traces[1], duty))
```
julienh-ssv, line 154: It would be good to add a new test where we store duties with two different roles, and make sure we don't get collisions.

exporter/model.go
```diff
@@ -145,11 +145,15 @@ type PartialSigTrace struct {
...
 
-// Committee
+// CommitteeDutyTrace represents committee-scoped duty traces for both
+// COMMITTEE and AGGREGATOR_COMMITTEE runner roles.
 type CommitteeDutyTrace struct {
 	ConsensusTrace
 
 	Slot phase0.Slot
+	// Role stores the runner role value for this committee trace
+	// (COMMITTEE or AGGREGATOR_COMMITTEE).
```
julienh-ssv: I think this comment is not necessary because none of the other fields have comments on them.

exporter2/validator.go
```diff
@@ -35,12 +35,7 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
 	for s := request.From; s <= request.To; s++ {
 		slot := phase0.Slot(s)
 		for _, role := range request.Roles {
-			providerFunc := e.getValidatorDutiesForRoleAndSlot
-			if isCommitteeDuty(role) {
-				providerFunc = e.getValidatorCommitteeDutiesForRoleAndSlot
-			}
-
-			duties, err := providerFunc(role, slot, indices)
+			duties, err := e.getValidatorTracesForRoleAndSlot(role, slot, indices)
```
julienh-ssv: I think that it would be better to change the `isCommitteeDuty` function to take the fork into account e.g. `e.isCommitteeDutyAtSlot(role, slot)`

This would allow not changing this code, avoid creating a new function `getValidatorTracesForRoleAndSlot` and avoid changing the code of `validateValidatorRequest`. 

The current implementation duplicates the check to decide if AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are committee duties in two locations: once in `getValidatorTracesForRoleAndSlot` and once in `validateValidatorRequest`.

I think that editing `isCommitteeDuty` instead would prevent this duplication.

exporter2/validator.go
```diff
@@ -55,7 +50,23 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
...
+func (e *Exporter) getValidatorTracesForRoleAndSlot(role spectypes.BeaconRole, slot phase0.Slot, indices []phase0.ValidatorIndex) ([]ValidatorCommitteeTrace, error) {
+	switch role {
+	case spectypes.BNRoleAttester, spectypes.BNRoleSyncCommittee:
+		return e.getValidatorCommitteeDutiesForRoleAndSlot(role, slot, indices)
+	case spectypes.BNRoleAggregator, spectypes.BNRoleSyncCommitteeContribution:
+		// Alan: AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are validator duties.
+		// Boole: AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are committee duties.
+		if e.networkConfig.BooleForkAtSlot(slot) {
+			return e.getValidatorCommitteeDutiesForRoleAndSlot(role, slot, indices)
+		}
```
julienh-ssv: see my previous comment about `isCommitteeDuty` and duplication

exporter2/validator.go
```diff
@@ -68,7 +79,10 @@ func validateValidatorRequest(request *ValidatorTracesQuery) error {
 	if len(request.PubKeys) == 0 && len(request.Indices) == 0 {
 		for _, role := range request.Roles {
 			if isCommitteeDuty(role) {
-				return fmt.Errorf("role %s is a committee duty, please provide either pubkeys or indices to filter the duty for a specific validators subset or use the /committee endpoint to query all the corresponding duties", role.String())
+				// Aggregator/SCC are validator duties pre-fork and committee duties post-fork.
+				if !isAggregatorCommitteeRole(role) || e.networkConfig.BooleForkAtSlot(phase0.Slot(request.To)) {
+					return fmt.Errorf("role %s is a committee duty, please provide either pubkeys or indices to filter the duty for a specific validators subset or use the /committee endpoint to query all the corresponding duties", role.String())
```
julienh-ssv: see my previous comment about `isCommitteeDuty` and duplication

operator/dutytracer/store.go
```diff
@@ -215,25 +186,71 @@ func hasSignersForRoles(duty *exporter.CommitteeDutyTrace, roles ...spectypes.Be
...
+	if wantCommittee {
+		out = append(out, spectypes.RoleCommittee)
+	}
+	if wantAggregator {
+		out = append(out, spectypes.RoleAggregatorCommittee)
+	}
+	return out
+}
+
+func runnerRoleAllowed(role spectypes.RunnerRole, allowed []spectypes.RunnerRole) bool {
```
julienh-ssv: This seems like a duplicate of the standard function `slices.Contains`?

exporter2/validator.go
```diff
@@ -116,9 +130,16 @@ func (e *Exporter) getValidatorDutiesForRoleAndSlot(role spectypes.BeaconRole, s
...
+// getValidatorCommitteeDutiesForRoleAndSlot reads committee-backed duties.
+// Callers should route AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION here only on Boole slots
+// (pre-Boole these are validator duties). ATTESTER/SYNC_COMMITTEE are committee duties on all forks.
 func (e *Exporter) getValidatorCommitteeDutiesForRoleAndSlot(role spectypes.BeaconRole, slot phase0.Slot, indices []phase0.ValidatorIndex) ([]ValidatorCommitteeTrace, error) {
 	results := make([]ValidatorCommitteeTrace, 0, len(indices))
 	var errs *multierror.Error
+	runnerRole := spectypes.RoleCommittee
+	if role == spectypes.BNRoleAggregator || role == spectypes.BNRoleSyncCommitteeContribution {
+		runnerRole = spectypes.RoleAggregatorCommittee
+	}
```
julienh-ssv, line 131: I think it would be better to use a standard mapping function BNRole --> RunnerRole:

```
runnerRole, ok := ssvtypes.CommitteeRunnerRoleForBeaconRole(role)
	if !ok {
		return nil, fmt.Errorf("unexpected committee-backed beacon role: %s", role.String())
	}
```

to prevent duplicating the mapping logic everywhere.

operator/dutytracer/store.go
```diff
@@ -128,85 +128,56 @@ func (c *Collector) getValidatorDutyFromDiskIndex(role spectypes.BeaconRole, slo
...
-		if role == spectypes.BNRoleAttester {
+		if role == spectypes.BNRoleAttester || role == spectypes.BNRoleAggregator {
 			if len(duty.Attester) == 0 {
 				return false
 			}
 		}
-		if role == spectypes.BNRoleSyncCommittee {
+		if role == spectypes.BNRoleSyncCommittee || role == spectypes.BNRoleSyncCommitteeContribution {
 			if len(duty.SyncCommittee) == 0 {
 				return false
```
julienh-ssv, line 187: Here we could encapsulate the mapping of BNRole to (Attester or SyncCommittee) in a helper function that we can also reuse in `getValidatorCommitteeDutiesForRoleAndSlot` e.g.

```
const (
	CommitteeSignerBucketUnknown CommitteeSignerBucket = iota
	CommitteeSignerBucketAttester
	CommitteeSignerBucketSyncCommittee
)

// CommitteeSignerBucketForBeaconRole returns which CommitteeDutyTrace signer bucket should
// contain signers for the given beacon role.
func CommitteeSignerBucketForBeaconRole(role spectypes.BeaconRole) (CommitteeSignerBucket, bool) {
	switch role {
	case spectypes.BNRoleAttester, spectypes.BNRoleAggregator:
		return CommitteeSignerBucketAttester, true
	case spectypes.BNRoleSyncCommittee, spectypes.BNRoleSyncCommitteeContribution:
		return CommitteeSignerBucketSyncCommittee, true
	default:
		return CommitteeSignerBucketUnknown, false
	}
}
```

This would avoid duplicating the switch in several locations

julienh-ssv (changes requested):
Nice work overall — the changes look solid.

I think this PR would be **significantly improved** by **centralizing all "role classification" logic** into a single helper layer (e.g. *beacon role -> committee runner role*, *beacon role -> signer bucket*, *fork-aware committee-backed-ness*). Then call sites can stay simple and consistent, only asking:

- **"Is this committee-backed at slot/range X?"**
- **"Which runner role / signer bucket does this beacon role map to?"**

This should reduce duplicated `switch` logic, make fork-awareness explicit in one place, and prevent the possibility of subtle inconsistencies between Decideds vs Validator query paths.

Here are a few suggestions:

---

## A) Fork-aware "committee duty at slot" helper

Add an `e.isCommitteeDutyAtSlot(role, slot)` helper that both Decideds and Validator queries can use, avoiding duplicated role switching. This should also allow removing `getValidatorTracesForRoleAndSlot`.

```go
func (e *Exporter) isCommitteeDutyAtSlot(role spectypes.BeaconRole, slot phase0.Slot) bool {
	rr, ok := ssvtypes.CommitteeRunnerRoleForBeaconRole(role)
	if !ok {
		return false
	}
	if rr == spectypes.RoleAggregatorCommittee {
		// Alan: AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are validator duties.
		// Boole: AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are committee duties.
		return e.networkConfig.BooleForkAtSlot(slot)
	}
	return true
}
```

This keeps the fork boundary logic in one place and makes downstream code just ask "committee-backed?" rather than re-deriving it.

---

## B) Standard mapper: BN role <-> runner role

Use a single canonical mapper for this conversion everywhere:

```go
ssvtypes.CommitteeRunnerRoleForBeaconRole(role)
```

Having one source of truth here helps prevent drift when roles evolve.

---

## C) Centralize role -> signer bucket mapping

Likewise, the mapping of BN roles to "attester vs sync committee" buckets can be centralized:

```go
// CommitteeSignerBucketForBeaconRole returns which CommitteeDutyTrace signer bucket should
// contain signers for the given beacon role.
func CommitteeSignerBucketForBeaconRole(role spectypes.BeaconRole) (CommitteeSignerBucket, bool) {
	switch role {
	case spectypes.BNRoleAttester, spectypes.BNRoleAggregator:
		return CommitteeSignerBucketAttester, true
	case spectypes.BNRoleSyncCommittee, spectypes.BNRoleSyncCommitteeContribution:
		return CommitteeSignerBucketSyncCommittee, true
	default:
		return CommitteeSignerBucketUnknown, false
	}
}
```

Then call sites can just ask for the bucket, instead of repeating classification logic.

api/handlers/exporter/committee_model.go
julienh-ssv: Yes as a general rule, adding new fields to a json response should not break clients as most json parsers simply ignore it.

exporter/store/store.go
nkryuchkov: Well, we rely a lot on the spec values, and if the spec changes the value, it will break everything throughout the project. I'm going to implement your suggestion, but I'm still not sure we should proceed with it. I think `runnerRoleToPrefix` would be just a workaround, and it won't give us any benefits (it will give benefits only for the `store`, which is not enough). I agree there is some risk the spec changes indices, but the values and their order are part of the spec, and if we want to protect the implementation from such risks, we need a more systematic approach

exporter/model.go
nkryuchkov: Yeah, I initially picked `uint64` for simplicity, but I have already changed this

julienh-ssv (approved):
Nice fixes!

exporter/store/store.go
julienh-ssv: I think the codebase relies on the named variable i.e. `spectypes.RoleCommittee` for checks like `if role == spectypes.RoleCommittee` but not on the actual value of that variable (e.g. `5`). So if the value changes, the code should not break. Testing for value, i.e. `if role == 5` would be a brittle way to use the spec value.

nkryuchkov: We also use the values when encoding messages, e.g., in `NewMsgID`

julienh-ssv: I agree that some serialized things must depend on spec values (the `NewMsgID(..., spectypes.RoleCommittee)` example is a good one) — because that’s literally the wire/protocol surface the spec defines. If the spec ever changes those numeric assignments, that’s a **breaking protocol change** and we’ll have to handle it explicitly (version bump, coordinated upgrade, maybe temporarily accepting both encodings).

But that’s exactly why I still think we shouldn’t reuse those same numeric values as **permanent DB prefix bytes**:

- **Different blast radius**: protocol breakage is "everyone upgrades together"; DB breakage is "your local data becomes unreadable/orphaned/collides silently" even if the semantic role didn’t change.

- **Spec refactor vs schema stability**: upstream can reorder/renumber enums for internal reasons (or language/codegen reasons) while keeping semantics. That shouldn’t force us into a KV schema migration.

- **Prefixes are our schema contract:** DB prefixes are part of our storage layout. They should be explicit, stable, and owned here — same as we wouldn’t base store keys on someone else’s iota without pinning it forever.

So runnerRoleToPrefix isn’t a workaround, it’s a **separation-of-concerns boundary**: we keep using spec values where the spec is authoritative (wire/message encoding), and we keep storage prefixes authoritative in our module (DB schema). Add a small test to guarantee the mapping is one-to-one and never changes accidentally, and we’ve bought ourselves safety with almost zero complexity.

If we think that we don't have a systematic approach and want stronger guarantees, it can be: protocol serialization follows spec, storage serialization follows our schema constants, and any intentional change to either comes with an explicit migration/compat plan.

nkryuchkov: Alright, let's keep it

y0sher (approved):

PR #2685, opened 2026-02-11
fix(ci): remove make tools to prevent uncontrolled tool upgrades

olegshmuelov:
## Summary

- Removed `make tools` target that ran `go get -tool` for every tool on each CI run, upgrading them to latest unconditionally
- Removed `make tools` step from `lint.yml` — unnecessary with Go 1.24+ tool directives
- Added `tool.sum` to CI cache for faster builds

## Why

`golangci-lint` v2.9.0 was released requiring Go >= 1.25.0, which broke CI for all branches because `make tools` auto-upgraded it from the pinned v2.8.0.

With Go 1.24+ tool directives, `go tool -modfile=tool.mod <name>` automatically downloads and builds the tool at the version pinned in `tool.mod`. The `go get` calls were redundant and harmful.

## How tool upgrades work now

**Automated:** Renovate will open PRs when new tool versions are available — reviewed, tested, merged deliberately.

**Manual:**
```bash
go get -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.9.0
# commit tool.mod + tool.sum
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,10 +25,8 @@ jobs:
           cache: true
           cache-dependency-path: |
             go.sum
+            tool.sum
             ssvsigner/go.sum
-
-      - name: Run tools
-        run: make tools
 
       - name: Run make lint
         run: make lint
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ ifeq ($(COVERAGE),true)
 	COV_CMD=-coverpkg=./... -covermode="atomic" -coverprofile="coverage.out"
 endif
 
-GET_TOOL=go get -modfile=tool.mod -tool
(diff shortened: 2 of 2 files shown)


julienh-ssv (commented):
I'm fine with the pipeline change (I had opened a similar PR a few weeks ago with version pinning, see: https://github.com/ssvlabs/ssv/pull/2642)

But, I think it would be useful to keep the makefile target to run locally?

iurii-ssv (approved):
👍

y0sher (approved):

olegshmuelov:
> But, I think it would be useful to keep the makefile target to run locally?

`go tool` already downloads and caches the pinned version on first use. `go mod download` would only pre-cache sources, not compiled binaries, so `go tool` still builds on first run anyway.

PR #2691, opened 2026-02-16
aggregator committee: improve pre-consensus completeness

nkryuchkov:
https://github.com/ssvlabs/ssv-spec/pull/604


protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -350,12 +361,28 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
 	if err != nil {
 		return fmt.Errorf("failed processing selection proof message: %w", err)
 	}
 	// quorum returns true only once (first time quorum achieved)
-	if !hasQuorum {
+	if !hasNewQuorum {
+		// If no new quorum, didn't start QBFT, and received the last message, then terminate.
+		if r.HasSeenAllPreConsensusSigners() && !r.HasStartedConsensus() {
+			r.BaseRunner.State.Finished = true
+		}
```
iurii-ssv: Since we've checked for `r.HasStartedConsensus()` above already there is no need to also check for it here ? Because the `!r.HasStartedConsensus()` is basically always `true` once we got here.

iurii-ssv: Also, to "finish the duty correctly" I think we need to do a bunch of things, and logging the fact that it finished would be nice to (not just set `r.state().Finished = true`):
```go
			r.state().Finished = true
			r.measurements.EndDutyFlow()
			recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
			const dutyFinishedNoMessages = "✔️successfully finished duty processing (got no quorums in pre-consensus)"
			logger.Info(dutyFinishedNoMessages,
				fields.PreConsensusTime(r.measurements.PreConsensusTime()),
				fields.TotalConsensusTime(r.measurements.TotalConsensusTime()),
				fields.TotalDutyTime(r.measurements.TotalDutyTime()),
			)
			span.AddEvent(dutyFinishedNoMessages)
			return nil
```

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -516,13 +545,18 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
+		if r.HaveCheckedAllDutiesForSelection(aggregatorMap, contributionMap) || r.HasSeenAllPreConsensusSigners() {
+			r.state().Finished = true
+			r.measurements.EndDutyFlow()
+			recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+			return nil
+		}
 
-		return nil
+		// If no validator was selected, but there are more possible messages, keep waiting for more messages.
+		return anyErr
```
olegshmuelov: When `anyErr != nil` here (e.g., `GetSyncCommitteeContribution` failed for a selected contributor), the error is silently discarded.
Consider `return anyErr` instead of `return nil` - the duty should still terminate (`Finished = true` ensures that), but propagating the error gives the queue consumer a chance to log it (DEBUG + span), making transient beacon failures debuggable.

Note: "only terminate when anyErr == nil" would be worse - it would cause the duty to hang indefinitely since all operators already sent their messages and no more calls to `ProcessPreConsensus` will come.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -1691,3 +1720,75 @@ func (r *AggregatorCommitteeRunner) GetSigner() ekm.BeaconSigner {
...
+	return r.BaseRunner.QBFTController.StoredInstances.FindInstance(
+		specqbft.Height(r.BaseRunner.State.CurrentDuty.DutySlot()),
+	) != nil
+}
+
+// MarkPreConsensusSignerAsSeen marks the signer of the given pre-consensus message as seen.
+func (r *AggregatorCommitteeRunner) MarkPreConsensusSignerAsSeen(signedMsg *spectypes.PartialSignatureMessages) {
+	for _, msg := range signedMsg.Messages {
+		r.preConsensusSeenSigners[msg.Signer] = struct{}{}
+		break // all message signers are equal, no need to inspect more messages
```
iurii-ssv: `all message signers are equal` -> `all messages have the same signer` ?

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -1691,3 +1720,75 @@ func (r *AggregatorCommitteeRunner) GetSigner() ekm.BeaconSigner {
...
+	if r.BaseRunner.State == nil {
+		return false
+	}
+	if r.BaseRunner.State.CurrentDuty == nil {
+		return false
+	}
+	return r.BaseRunner.QBFTController.StoredInstances.FindInstance(
+		specqbft.Height(r.BaseRunner.State.CurrentDuty.DutySlot()),
+	) != nil
+}
```
iurii-ssv, line 1744: The implementation of this func can probably be just this instead:
```go
// HasStartedConsensus checks if consensus has already started (or maybe even finished) for the current duty.
func (r *AggregatorCommitteeRunner) HasStartedConsensus() bool {
	if r.BaseRunner.State == nil {
		return false
	}
	return r.BaseRunner.State.RunningInstance != nil || r.BaseRunner.State.DecidedValue != nil
}
```

iurii-ssv (commented):
LGTM, just minor suggestions

protocol/v2/ssv/runner/aggregator_committee.go
nkryuchkov: I copied this from https://github.com/ssvlabs/ssv-spec/pull/604. Can you please leave the suggestions from this review there? It would be great to simplify it in the spec to avoid code discrepancies. I'll backport the changes once they are applied

iurii-ssv: https://github.com/ssvlabs/ssv-spec/pull/604/changes#r2833415312

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -516,13 +545,18 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
-		recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+	// No aggregators selected: decide whether to finish or wait for more messages.
+	if !hasAnyAggregatorForNewQuorum {
+		// If all duties have been tested for selection or all messages (from all operators) have been seen, terminate.
+		if r.HaveCheckedAllDutiesForSelection(aggregatorMap, contributionMap) || r.HasSeenAllPreConsensusSigners() {
+			r.state().Finished = true
+			r.measurements.EndDutyFlow()
+			recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+			return nil
+		}
```
olegshmuelov, line 570: nit: The `processSyncCommitteeSelectionProof` error at line 537 is only stored in `anyErr` with no logging.
Unlike BLS reconstruction failures (which are logged at Error level at line 492), a failed `GetSyncCommitteeContribution` would be invisible if the duty terminates via the all-signers-seen path.

Consider adding a `logger.Warn(...)` at the error site for observability.

protocol/v2/ssv/runner/aggregator_committee.go
nkryuchkov: I applied the latest changes from spec. Closing this

protocol/v2/ssv/runner/aggregator_committee.go
nkryuchkov: @iurii-ssv applied, please review

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -501,10 +535,16 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
 						if isAggregator {
-							hasAnyAggregator = true
+							hasAnyAggregatorForNewQuorum = true
 						}
 					} else {
+						logger.Warn("failed to process sync committee selection proof",
+							zap.Error(err),
+							fields.Slot(vDuty.Slot),
+							fields.ValidatorIndex(vDuty.ValidatorIndex),
+							fields.Root(root))
```
iurii-ssv: I believe we don't need to add any of: `fields.Slot(vDuty.Slot)`, `fields.ValidatorIndex(vDuty.ValidatorIndex)`, `fields.Root(root)` here:
- the `fields.Slot(vDuty.Slot)` is already propagated with the `logger` itself (true for every p2p-message-handling func such as `ProcessPreConsensus`, true for every runner)
- the `fields.ValidatorIndex(vDuty.ValidatorIndex)`, `fields.Root(root)` are already added to the logger in the code above

iurii-ssv (commented):
LGTM, with minor comment

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -516,13 +556,18 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
-		r.measurements.EndDutyFlow()
-		recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+	// No aggregators selected: decide whether to finish or wait for more messages.
+	if !hasAnyAggregatorForNewQuorum {
+		// If all duties have been tested for selection or all messages (from all operators) have been seen, terminate.
+		if r.HaveCheckedAllDutiesForSelection(aggregatorMap, contributionMap) || r.HasSeenAllPreConsensusSigners() {
+			r.state().Finished = true
+			r.measurements.EndDutyFlow()
+			recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+			return anyErr
```
iurii-ssv, line 569: We should also log & record all the data here too, similar to how it's done in other places with `.Finished = true`:
```
		r.state().Finished = true
		r.measurements.EndDutyFlow()
		recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
		const dutyFinishedNoAggregators = "✔️successfully finished duty processing (no validator is aggregator or sync committee contributor)"
		logger.Info(dutyFinishedNoAggregators,
			fields.PreConsensusTime(r.measurements.PreConsensusTime()),
			fields.TotalDutyTime(r.measurements.TotalDutyTime()),
		)
		span.AddEvent(dutyFinishedNoAggregators)
		return anyErr
```

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -350,12 +361,34 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
+		if r.HasSeenAllPreConsensusSigners() {
+			r.BaseRunner.State.Finished = true
+			r.measurements.EndDutyFlow()
+			recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregatorCommittee, 0)
+			const dutyFinishedNoMessages = "✔️successfully finished duty processing (got no quorums in pre-consensus)"
+			logger.Info(dutyFinishedNoMessages,
+				fields.PreConsensusTime(r.measurements.PreConsensusTime()),
+				fields.TotalConsensusTime(r.measurements.TotalConsensusTime()),
+				fields.TotalDutyTime(r.measurements.TotalDutyTime()),
+			)
```
iurii-ssv, line 390: Since we couldn't have had the consensus phase (if we ended up here), we shouldn't be logging `fields.TotalConsensusTime(r.measurements.TotalConsensusTime()),` at all (otherwise its value will be `0` which is confusing) ... so I would remove this field from this log-line.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -350,12 +361,34 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
...
 	if err != nil {
 		return fmt.Errorf("failed processing selection proof message: %w", err)
 	}
 	// quorum returns true only once (first time quorum achieved)
-	if !hasQuorum {
+	if !hasNewQuorum {
+		// If didn't get any new quorum, didn't yet start QBFT (checked above), and has received the last message, then terminate.
+		if r.HasSeenAllPreConsensusSigners() {
+			r.BaseRunner.State.Finished = true
+			r.measurements.EndDutyFlow()
```
iurii-ssv, line 384: We should also add these here right before the `r.measurements.EndDutyFlow()` call, to make sure we log the correct value (and actually record the metric) when terminating duty like that:
```go
	r.measurements.EndPreConsensus()
	recordPreConsensusDuration(ctx, r.measurements.PreConsensusTime(), spectypes.RoleAggregatorCommittee)
```

iurii-ssv (approved):

PR #2695, opened 2026-02-19
eth: sync finality

nkryuchkov:
Supersedes https://github.com/ssvlabs/ssv/pull/2184

SIP: https://github.com/ssvlabs/SIPs/pull/67

### DIFF
--- a/cli/GENERATE_CONFIG.md
+++ b/cli/GENERATE_CONFIG.md
@@ -21,23 +21,22 @@ The `generate-config` command allows you to generate a YAML configuration file b
 
 ### Flags
 
-| Flag                                 | Type   | Default                           | Description                                                                |
-|--------------------------------------|--------|-----------------------------------|----------------------------------------------------------------------------|
-| `--output-path`                      | string | `./config/config.local.yaml`      | Output path for the generated configuration file.                          |
-| `--log-level`                        | string | `info`                            | Sets the logging level (e.g., `debug`, `info`, `warn`, `error`).           |
-| `--db-path`                          | string | `./data/db`                       | Path to the database directory.                                            |
-| `--discovery`                        | string | `mdns`                            | Discovery method.                                                          |
-| `--consensus-client`                 | string | _Mandatory_                       | Address of the consensus client (e.g., `http://localhost:9000`).           |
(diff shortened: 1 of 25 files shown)


y0sher:
As mentioned in the [SIP](https://github.com/ssvlabs/SIPs/pull/67) comments, we are still debating whether this feature will need a fork or not. Also, we are discussing adding a "special treatment" for some events like exits/removals to allow a nice UX. So I think this pends some discussion with spec (CC: @GalRogozinski @MatheusFranco99 ).

nkryuchkov:
> As mentioned in the [SIP](https://github.com/ssvlabs/SIPs/pull/67) comments, we are still debating whether this feature will need a fork or not. Also, we are discussing adding a "special treatment" for some events like exits/removals to allow a nice UX. So I think this pends some discussion with spec (CC: @GalRogozinski @MatheusFranco99 ).

Yes, I wanted the first version to be ported from the original PR as it is, just with minor changes. I tested it on stage, and it worked fine. I have also been thinking about the fork, and I don't see the reason why we'd need it. So, I want to remove the fork code, but I decided to keep it until the discussion on it has been finalized.

I'm currently working on the special treatment of `ValidatorExited`, I'll likely push it in the next commit

GalRogozinski (commented):
There could be issues where views differ of registered validators. So some operators may send messages that will be rejected by others. This can cause temporary scoring issues...

I think ValidatorExits may have issues, but it is not so important since this can be activated via EL

MatheusFranco99:
Overall I like moving event sync from 8 EL blocks to CL finality, but we need to be sure that this has no impact on duty loss, or signing anything on behalf of exited validators.

Also, I wanted to clarify something about the fork transition.
Suppose $F$ is the fork epoch.
Let $f.s$ be the slot $s$ (1 to 32) of epoch $f$.
- During slot $F-1.32$, operators will sync with events of block with slot up to $F-1.24$.
- Then, at slot $F.1$ it won't sync anymore with events of slot $F-1.25$ since it's already applying the new rule, and it will only sync with all events of epoch $F-1$ once (at best) it's in $F+1$?

nkryuchkov:
On hold until we decide if it requires a fork

PR #2699, opened 2026-02-23
Pre-fetch proposer duties mid-epoch and Retry on failure

### LINKED ISSUE #2619: proposer-duty: duty-fetch improvements
For Proposer-duty, we fetch duties on the very 1st slot of the current epoch (or on the slot that comes after the slot SSV node started at)

- if we have a Proposer-duty to execute in that same slot of that epoch we'll delay it for longer than necessary waiting for that fetch to happen (for attestations we fetch duties ~half an epoch ahead of time, for example)
- if we are not able to fetch the Proposer-duties from CL in some epoch - there will be no retrying of that fetch for that epoch (we'll just log the `failed to fetch duties for current epoch` error and won't execute any duties in that epoch)

Related https://github.com/ssvlabs/ssv-node-board/issues/676

10234567Z:
## Problem

Two issues with proposer duty fetching:

1. Duties are fetched at the **first slot of each epoch**. If a proposer duty falls on that same slot, execution is delayed waiting for the fetch, unlike attestations which pre-fetch ~half an epoch ahead.

2. If the fetch **fails**, no retry happens for the rest of the epoch (~6 min of missed duties).

## Solution

Align proposer handler with the existing attester pattern:

- **Pre-fetch at mid-epoch** (`slotsPerEpoch/2 - 1`) so duties are ready before the epoch boundary
- **Retry on failure** by keeping the fetch flag set, so the next slot tick (12s) retries automatically
- Also pre-fetch next epoch in `HandleInitialDuties` if node starts past mid-epoch

## Files Changed

- `operator/duties/proposer.go` : core logic + typo fix (`processFetteching`)
- `operator/duties/proposer_test.go` : updated one test to expect the new pre-fetch signal

Closes #2619


operator/duties/proposer.go
```diff
@@ -80,21 +93,49 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+					if h.fetchCurrentEpoch {
+						if h.processFetching(tickCtx, currentEpoch) {
+							h.fetchCurrentEpoch = false
+						}
+					}
+					if h.fetchNextEpoch {
+						if h.processFetching(tickCtx, currentEpoch+1) {
+							h.fetchNextEpoch = false
+						}
 					}
```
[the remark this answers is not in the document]
10234567Z: Fixed it with new commit

10234567Z:
Fixed in the third commit. At the last slot of the epoch, if fetchNextEpoch is still set (all retries failed), it's converted to fetchCurrentEpoch. This way, after the epoch rolls over, the handler fetches the correct epoch instead of skipping it by targeting currentEpoch+1 (which would be N+2 at that point).

operator/duties/proposer.go
```diff
@@ -119,16 +168,26 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-	epoch := h.beaconConfig.EstimatedCurrentEpoch()
+	slot := h.beaconConfig.EstimatedCurrentSlot()
+	epoch := h.beaconConfig.EstimatedEpochAtSlot(slot)
 	h.processFetching(ctx, epoch)
+
+	// If starting past mid-epoch, pre-fetch next epoch duties since
+	// the mid-epoch trigger for this epoch has already passed.
+	if uint64(slot)%h.beaconConfig.SlotsPerEpoch >= h.beaconConfig.SlotsPerEpoch/2 {
+		h.processFetching(ctx, epoch+1)
+	}
```
iurii-ssv: I think it's better to explicitly ignore the returned value:
```
_ = h.processFetching(ctx, epoch)

...

_ = h.processFetching(ctx, epoch+1)
```

operator/duties/proposer.go
```diff
@@ -43,6 +50,7 @@ func (h *ProposerHandler) Name() string {
 // On First Run:
 //  1. Fetch duties for the current epoch.
 //  2. Execute duties.
+//  3. If past mid-epoch, schedule next-epoch pre-fetch.
```
iurii-ssv: I think we are doing `If past mid-epoch, schedule next-epoch pre-fetch.` before `Execute duties.` on first run ? If so we should swap 2 and 3

**Update:** actually, maybe not ... it's a bit hard to tell with the current code given we have various conditions spread out across multiple places ... which is basically why I'd rather use the approach `AttesterHandler` follows (instead of having multiple `if uint64(slot)%slotsPerEpoch == slotsPerEpoch/2-1` checks in various places and an extra `fetchFirst` flag)

iurii-ssv (commented):
Hey @10234567Z, thanks for looking into this,

the diff in this PR looks reasonable, I would however suggest we adopt an approach that closely mirrors that of other duty-handlers (eg. `AttesterHandler`) to keep things uniform/simple,

the only downside to doing that I can think of would be: we'd be fetching duties for the current **and** next epoch on the `HandleInitialDuties` call **before** we proceed to executing them (meaning we'd be delaying the duty execution for the current/upcoming slot for no good reason - with the current implementation, and the implementation from your PR, I think we are fetching the duties for the next epoch only after we execute the duties for the current slot) ... while extra wait-time upon SSV node startup is undesirable - it isn't a big issue because SSV cluster typically has redundant nodes to perform duties even with 1 node being down for some time. I think if we want to address this "extra" downtime caused by fetching duties for the next epoch - we should do it universally for all duty-types in the same manner (in a separate PR later on) ... I've created a separate issue to track it - https://github.com/ssvlabs/ssv/issues/2705

Do you think we could adopt/adapt the implementation we currently have in `AttesterHandler` for `ProposerHandler` @10234567Z ?

Btw, we use **stage** as the base/development branch (create PRs against it), and merge **stage** branch into **main** every so often to make a release. So we should have this PR open against & based on the **stage** branch.

10234567Z:
@iurii-ssv Yeah I agree going the universal way would be a correct choice of things. Would be implementing the same way it is being handled inside `AttesterHandler` . Once this is done, I can even move forward to fixing the universal fetching logic which you mentioned in the new issue.

Will let you know once done, thanks for the quick input!

protocol/v2/ssv/spectest/msg_processing_type.go
```diff
@@ -148,8 +148,13 @@ func (test *MsgProcessingSpecTest) runPreTesting(ctx context.Context, logger *za
 				lastErr = err
 			}
 			if test.DecidedSlashable && IsQBFTProposalMessage(msg) {
+				consensusMsg, err := specqbft.DecodeMessage(msg.SSVMessage.Data)
+				if err != nil {
+					panic(err)
```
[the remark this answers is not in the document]
10234567Z: Obsolete Changes, don't mind these!

10234567Z:
@iurii-ssv Hey! I've done a full rewrite of `ProposerHandler` to mirror the `AttesterHandler` pattern as you suggested. Here's a summary of the changes:

**`operator/duties/proposer.go`** - Full rewrite to align with `AttesterHandler`:
- Replaced `fetchFirst` flag with `fetchCurrentEpoch` / `fetchNextEpoch` flags (same as attester)
- New `processFetching(ctx, epoch, slot)` method handles both current and next epoch fetching internally using the flag guards, no more scattered fetch logic
- Retry on failure: if a fetch fails, the flag stays set (`= false` is only reached on success), so the next slot tick automatically retries
- Mid-epoch pre-fetch: at slot `slotsPerEpoch/2 - 1`, `fetchNextEpoch` is set to `true`, with an additional `shouldFetchNexEpoch(slot)` guard inside `processFetching` to delay the actual fetch until the appropriate slot range
- `HandleInitialDuties` sets both flags and calls `processFetching`, same as attester
- Removed the old `h.duties.Set(epoch, []StoreDuty{})` fallback on error (this was the root cause of no-retry, it signaled "done" to the store even on failure)
- `ResetEpoch(currentEpoch - 1)` GC at last slot of epoch, same as attester
- Execution (`processExecution`) runs before fetching in the ticker, so stored duties execute without delay

**`operator/duties/proposer_test.go`** - All 6 tests rewritten to match the new pattern:
- `setupProposerDutiesMock` now uses `waitForDuties *SafeValue[bool]` gating (same as attester tests)
- Duties are set up before `startScheduler`, and `waitForDuties` controls when fetches are signaled
- Reorg tests account for mid-epoch next-epoch fetch signals
- All 6 tests pass consistently

**`operator/duties/base_handler.go`** - Removed unused `indicesChanged` field (was flagged by linter)

Let me know if this looks good or if there's anything else to adjust!

EDIT: Just want to add, that the main branch of my fork is just stage branch of SSV, so as suggested by you I did the PR against/for stage instead of main now.

operator/duties/proposer.go
```diff
@@ -77,25 +86,23 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-				h.duties.ResetEpoch(currentEpoch - 1)
-				h.fetchFirst = true
-			}
+				h.processFetching(tickCtx, currentEpoch, slot)
+
+				// last slot of epoch
+				if uint64(slot)%slotsPerEpoch == slotsPerEpoch-1 {
+					h.duties.ResetEpoch(currentEpoch - 1)
+				}
+			}()
```
iurii-ssv, line 151: For this block of code, I think it's better to follow the `AttesterHandler` exactly (is there any reasons to have these lines in a slightly different order, or terminate the `func` that wraps the `tickCtx` later than here ?)`:
```
			func() {
				tickCtx, cancel := h.ctxWithDeadlineInOneEpoch(ctx, slot)
				defer cancel()

				h.executeAggregatorDuties(tickCtx, currentEpoch, slot)
				h.processFetching(tickCtx, currentEpoch, slot)
			}()

			slotsPerEpoch := h.beaconConfig.SlotsPerEpoch

			// If we have reached the mid-point of the epoch, fetch the duties for the next epoch in the next slot.
			// This allows us to set them up at a time when the beacon node should be less busy.
			if uint64(slot)%slotsPerEpoch == slotsPerEpoch/2-1 {
				h.fetchNextEpoch = true
			}

			// last slot of epoch
			if uint64(slot)%slotsPerEpoch == slotsPerEpoch-1 {
				h.duties.ResetEpoch(currentEpoch - 1)
			}
```

operator/duties/proposer.go
```diff
@@ -114,36 +125,62 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
 			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, slot, slot%32+1)
 			h.logger.Info("🔁 indices change received", zap.String("epoch_slot_pos", buildStr))
 
-			h.indicesChanged = true
+			h.fetchCurrentEpoch = true
+
+			// reset next epoch duties if in appropriate slot range
+			if h.shouldFetchNexEpoch(slot) {
+				h.fetchNextEpoch = true
+			}
```
iurii-ssv: I think we need to mirror exactly what the `AttesterHandler` does for both: `case reorgEvent := <-h.reorg` and `case <-h.indicesChange`.

**Update:** actually the `case reorgEvent := <-h.reorg` should be slightly simpler since for the proposer duties only the duties for the next epoch can be affected by the reorg.

operator/duties/proposer.go
```diff
@@ -42,20 +48,23 @@ func (h *ProposerHandler) Name() string {
...
 //  1. Execute duties.
 //  2. ResetEpoch duties for the current epoch.
 //  3. Fetch duties for the current epoch.
+//  4. If necessary, fetch duties for the next epoch.
 //
 // On Ticker event:
 //  1. Execute duties.
-//  2. If necessary, fetch duties for the current epoch.
+//  2. If necessary, fetch duties for the next epoch.
 func (h *ProposerHandler) HandleDuties(ctx context.Context) {
```
iurii-ssv, line 66: Once we align the code with `AttesterHandler` we should make this comment here to also match the comment from `AttesterHandler` exactly (unless there are some differences that need to be there for `ProposerHandler` ... but I don't see why there would be anything different about it)

iurii-ssv (commented):
Hey @10234567Z, good job!

I think the PR is pretty close to what we need, but I don't see why we can't make the `ProposerHandler` to look exactly as `AttesterHandler` does (with the main difference being them having their own `fetchAndProcessDuties` funcs doing slightly different things, fetching slightly different data).

Note, I think we should keep the code structures for attester and proposer separate (rather than trying to make a generic struct, or something) ... but I'm thinking that the code in `HandleDuties` func should pretty much look almost the same for both ?

I would imagine we could also copy over the unit-tests we have for `AttesterHandler` and adjust them accordingly (to replace the `ProposerHandler` unit-tests that will probably become stale after that change).

iurii-ssv:
Also, @10234567Z I've been looking at this code and I think we might want to simplify `AttesterHandler` first before proceeding with the change in this PR (since it's gonna be mostly a copy-paste from what `AttesterHandler` code does) ... so please wait a bit with this PR before proceeding further - I'll post an update about it here.

10234567Z:
@iurii-ssv Thanks for the detailed review! You're right, there's no reason for these to differ from AttesterHandler. The differences were unintentional - I'll fix all three:

1. **Ticker block**: Will move mid-epoch trigger and `ResetEpoch` outside the `func()` closure, and use `ctxWithDeadlineInOneEpoch` instead of `ctxWithDeadlineOnNextSlot` - matching attester exactly.

2. **Reorg block**: Will add the `func()` wrapper with `ctxWithDeadlineInOneEpoch`, handle both `reorgEvent.Previous` and `reorgEvent.Current` separately, and call `processFetching` inside the reorg case - matching attester exactly.

3. **Comment block**: Will copy attester's comment and adjust for proposer.

I saw your note about potentially simplifying AttesterHandler first..so yeah I'll hold off on pushing changes until you give the go-ahead. Let me know when I should proceed!

iurii-ssv:
> Also, @10234567Z I've been looking at this code and I think we might want to simplify `AttesterHandler` first before proceeding with the change in this PR (since it's gonna be mostly a copy-paste from what `AttesterHandler` code does) ... so please wait a bit with this PR before proceeding further - I'll post an update about it here.

@10234567Z to follow up on this ^

I've [opened a PR](https://github.com/ssvlabs/ssv/pull/2712) with some improvements & cleanup for `AttesterHandler`, once I finish it (we need to review & test & merge it) - we can get back to this PR.

iurii-ssv:
@greptile could you re-review this PR (now that it has been pretty much reworked from scratch) ?

operator/duties/proposer.go
```diff
@@ -21,16 +21,20 @@ import (
...
 }
 
 func NewProposerHandler(duties *dutystore.Duties[eth2apiv1.ProposerDuty], exporterMode bool) *ProposerHandler {
 	return &ProposerHandler{
-		duties:       duties,
-		fetchFirst:   true,
-		exporterMode: exporterMode,
+		duties:           duties,
+		exporterMode:     exporterMode,
+		dutyFetchIntents: make(map[phase0.Epoch]bool),
```
iurii-ssv, line 36: Nit: lets make sure the initialization order here follows the order in which these fields are defined on the `ProposerHandler struct` (this way it's easier to spot a missing initialization, for example) ... meaning `exporterMode:     exporterMode,` should be the last initialized fields here

operator/duties/proposer.go
```diff
@@ -84,81 +96,178 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-			h.indicesChanged = true
+				// 1) Declare intents.
+				if !reorgEvent.Current {
+					// Reorg on the previous epoch means the duties for the current epoch might have changed, so
+					// we want to re-fetch them.
+					h.dutyFetchIntents[currentEpoch] = false
+				}
+				// Reorg on the previous or current epoch means the duties for the next epoch might have changed, so
+				// we want to re-fetch them.
+				h.dutyFetchIntents[nextEpoch] = false
```
iurii-ssv: I've looked a bit more into how Proposer duties re-fetch should work during reorgs - seems like we need to update this handling to be more like this:
```go
// Reorg with "CurrentDutyDependentRoot changed in the same epoch", or "PreviousDutyDependentRoot changed during the epoch transition" potentially affects duties for the current epoch.
// Any reorg always potentially affects duties for the next epoch.
refetchCurrentEpoch := reorgEvent.CurrentDutyDependentRootChanged ||
    (reorgEvent.PreviousDutyDependentRootChanged && reorgEvent.EpochTransition)
if refetchCurrentEpoch {
    h.dutyFetchIntents[currentEpoch] = false
}
h.dutyFetchIntents[nextEpoch] = false
```
In https://github.com/ssvlabs/ssv/pull/2712 I've simplified the `ReorgEvent struct` to define just `Current`, but seems like we actually need to change to be this instead (and initialize corresponding fields correctly too ofc):
```go
type ReorgEvent struct {
    Slot phase0.Slot

    CurrentDutyDependentRootChanged  bool
    PreviousDutyDependentRootChanged bool
    EpochTransition bool
}
```
And we'll need to do something like this in `AttesterHandler` (to account for this change):
```go
// Reorg with "PreviousDutyDependentRoot changed" potentially affects duties for the current epoch.
// Any reorg always potentially affects duties for the next epoch.
refetchCurrentEpoch := reorgEvent.PreviousDutyDependentRootChanged
if refetchCurrentEpoch {
    h.dutyFetchIntents[currentEpoch] = false
}
h.dutyFetchIntents[nextEpoch] = false
```

iurii-ssv (commented):
Looks great, but we probably need to adjust the reorg behavior a bit (see the corresponding comment on it)

10234567Z:
- Moves the `refetchCurrentEpoch` bool computation outside the function closure (but before the logger) in `attester.go` and `proposer.go`, so it can be reused by both the logger and the intent-setting logic inside the closure.

- Changes the `ReorgEvent` struct in `scheduler.go` to have separate fields (`CurrentDutyDependentRootChanged`, `PreviousDutyDependentRootChanged`, `EpochTransition`) instead of a single `Current` bool, allowing each duty handler to decide independently whether to refetch based on which root changed and whether an epoch transition occurred.

- Updates proposer reorg logic: refetch current epoch on `CurrentDutyDependentRootChanged` or on `PreviousDutyDependentRootChanged && EpochTransition` (attester logic unchanged: refetch on `PreviousDutyDependentRootChanged`).

- Fixes the 4 proposer reorg tests to align with the new logic (`Reorg_Previous` / `Reorg_Previous_Indices_Changed` now expect no refetch; `Reorg_Current` / `Reorg_Current_Indices_Changed` now expect an immediate refetch).

- Aligns `ReorgEvent` struct literal fields in `sync_committee.go`.

operator/duties/proposer.go
```diff
@@ -157,6 +157,8 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
 
 			slotNumber := uint64(currentSlot)%h.beaconConfig.SlotsPerEpoch + 1
 			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, currentSlot, slotNumber)
+			refetchCurrentEpoch := reorgEvent.CurrentDutyDependentRootChanged ||
+				(reorgEvent.PreviousDutyDependentRootChanged && reorgEvent.EpochTransition)
 			logger := h.logger.With(
```
iurii-ssv, line 163: Let's add a blank-line between `refetchCurrentEpoch := ` and `logger := ` to keep this code cleanly-separated (like in `AttesterHandler`)

operator/duties/proposer.go
```diff
@@ -177,14 +179,9 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-				// 1) Declare intents.
-				if !reorgEvent.Current {
-					// Reorg on the previous epoch means the duties for the current epoch might have changed, so
-					// we want to re-fetch them.
+				if refetchCurrentEpoch {
 					h.dutyFetchIntents[currentEpoch] = false
 				}
-				// Reorg on the previous or current epoch means the duties for the next epoch might have changed, so
-				// we want to re-fetch them.
 				h.dutyFetchIntents[nextEpoch] = false
```
iurii-ssv, line 188: Seems like this section is missing a comment (compared to `AttesterHandler`), could add this one from https://github.com/ssvlabs/ssv/pull/2699#discussion_r2976408498:
```go
// Reorg with "CurrentDutyDependentRoot changed in the same epoch", or "PreviousDutyDependentRoot changed during the epoch transition" potentially affects duties for the current epoch.
// Any reorg always potentially affects duties for the next epoch.
```

iurii-ssv (commented):
Almost there, needs minor adjustments (see open discussions below/above).

Would be also nice to add a test to cover `reorgEvent.EpochTransition == true` edge-case for `ProposerHandler` (seems to be missing).

PR #2703, opened 2026-02-26
feat(api): add pagination to validators list

julienharbulot:
Add optional pagination to the validators http api with retro-compatibility

```
curl --request GET \
    --url 'http://ssv-node-exporter.stage.vnet.ops.ssvlabsinternal.com/v1/validators' \
    --header 'Content-Type: application/json' \
    --data '{
      "clusters": [
        [73,74,75,76]
      ],
      "pagination": {
        "page": 1,
        "per_page": 10
      }
    }'
```

Closes ssvlabs/ssv-node-board#1047

### DIFF
--- a/api/handlers/validators/validators.go
+++ b/api/handlers/validators/validators.go
@@ -2,7 +2,9 @@ package validators
 
 import (
 	"bytes"
+	"fmt"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -18,21 +20,37 @@ type Validators struct {
 }
 
 func (h *Validators) List(w http.ResponseWriter, r *http.Request) error {
-	var request struct {
-		Owners      api.HexSlice    `json:"owners" form:"owners"`
-		Operators   api.Uint64Slice `json:"operators" form:"operators"`
-		Clusters    requestClusters `json:"clusters" form:"clusters"`
-		Subclusters requestClusters `json:"subclusters" form:"subclusters"`
-		PubKeys     api.HexSlice    `json:"pubkeys" form:"pubkeys"`
-		Indices     api.Uint64Slice `json:"indices" form:"indices"`
+	const (
+		defaultPerPage = uint64(1000)
+		maxPerPage     = uint64(10000)
+	)
+
+	if r.URL.Query().Has("page") || r.URL.Query().Has("per_page") || r.URL.Query().Has("pagination") {
+		return api.BadRequestError(fmt.Errorf("pagination must be provided in request body"))
 	}
-	var response struct {
-		Data []*validatorJSON `json:"data"`
+
+	var request struct {
+		Owners      api.HexSlice           `json:"owners" form:"owners"`
+		Operators   api.Uint64Slice        `json:"operators" form:"operators"`
+		Clusters    requestClusters        `json:"clusters" form:"clusters"`
+		Subclusters requestClusters        `json:"subclusters" form:"subclusters"`
+		PubKeys     api.HexSlice           `json:"pubkeys" form:"pubkeys"`
+		Indices     api.Uint64Slice        `json:"indices" form:"indices"`
+		Pagination  api.OptionalPagination `json:"pagination" form:"pagination"`
 	}
 
 	if err := api.Bind(r, &request); err != nil {
-		return err
+		return api.BadRequestError(err)
 	}
+
+	paginationRequest, err := request.Pagination.ToRequest(api.PaginationOptions{
+		DefaultPerPage: defaultPerPage,
+		MaxPerPage:     maxPerPage,
+	})
+	if err != nil {
+		return api.BadRequestError(err)
+	}
+	paginationRequested := paginationRequest != nil
 
 	var filters []registrystorage.SharesFilter
 	if len(request.Owners) > 0 {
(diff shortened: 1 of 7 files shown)


api/pagination.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func PaginationFromRequest(p PaginationRequest, total uint64) Pagination {
+	out := Pagination{
+		Page:    p.Page,
+		PerPage: p.PerPage,
+		Total:   total,
+	}
+	if total > 0 && p.PerPage > 0 {
+		out.TotalPages = (total + p.PerPage - 1) / p.PerPage
```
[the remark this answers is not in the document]
julienh-ssv: ok

api/pagination_test.go
```diff
@@ -0,0 +1,61 @@
...
+		require.Error(t, err)
+	})
+
+	t.Run("per_page too large invalid", func(t *testing.T) {
+		t.Parallel()
+
+		_, err := OptionalPaginationFromQuery(url.Values{"per_page": []string{"10001"}}, opts)
+		require.Error(t, err)
+	})
+}
```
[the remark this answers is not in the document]
julienh-ssv: ok

api/handlers/validators/validators.go
```diff
@@ -55,10 +67,39 @@ func (h *Validators) List(w http.ResponseWriter, r *http.Request) error {
...
+	if !paginationRequested {
+		var response struct {
+			Data []*validatorJSON `json:"data"`
+		}
+		response.Data = make([]*validatorJSON, len(shares))
+		for i, share := range shares {
+			response.Data[i] = validatorFromShare(share)
+		}
+		return api.Render(w, r, response)
+	}
```
iurii-ssv, line 89: Maybe could leave a brief comment about this code ensuring backward-compatibility of this API (as the main reason for it to be here)

iurii-ssv (dismissed):
LGTM

api/handlers/validators/validators.go
julienh-ssv: ok

api/handlers/validators/validators.go
```diff
@@ -18,22 +20,38 @@ type Validators struct {
...
-		Clusters    requestClusters `json:"clusters" form:"clusters"`
-		Subclusters requestClusters `json:"subclusters" form:"subclusters"`
-		PubKeys     api.HexSlice    `json:"pubkeys" form:"pubkeys"`
-		Indices     api.Uint64Slice `json:"indices" form:"indices"`
+	const (
+		defaultPerPage = uint64(1000)
+		maxPerPage     = uint64(10000)
+	)
+
+	if r.URL.Query().Has("page") || r.URL.Query().Has("per_page") || r.URL.Query().Has("pagination") {
```
oleg-ssvlabs, line 40: I am not sure I fully understand this validation.

1) Error message mentions request **body**, but we validate URL **query parameters**, not request body.
2) I assume `if r.URL.Query().Has("page") ...`should rather contain logical NOT, like `!r.URL.Query().Has("page")`?
3) I think we agreed that pagination should be optional, so this validation should rather not exist? So when it does - the response is paginated, if it does not - return non paginated response.

Do I understand it right?

oleg-ssvlabs: Ah, I probably understand it now.

So the pagination related values **must** be provided in request body and not query parameters and you validate it

api/pagination.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"math/bits"
+)
+
+type PaginationOptions struct {
+	DefaultPerPage uint64
+	MaxPerPage     uint64
+}
+
+type OptionalPagination struct {
```
oleg-ssvlabs: nit:

I think it would be easier to differentiate which data structures are request-related and which are response-related if we reflect that in their names, for example:

- `OptionalPagination` → `OptionalPaginationRequest`
- `Pagination` → `PaginationResponse`

oleg-ssvlabs (commented):
gj!

I think we also need to update [Open API spec](https://github.com/ssvlabs/ssv/tree/stage/docs/api)

api/pagination.go
julienh-ssv: There is no Open API Spec for that endpoint because it's not part of the exporter api. The generator does not auto-generate outside of the exporter API and manual edits won't pass the linter.

I can rename the fields.

EDIT I'll see what I can do about the open api generation but this is likely for another PR

oleg-ssvlabs (dismissed):
lgtm

api/handlers/validators/validators.go
iurii-ssv: Were you gonna to add the comment ? I think it's still missing.

Not a blocker for merging though :)

y0sher:
@julienh-ssv This is a breaking change in the API, can we still use `v1/validators` without pagination the same way?

api/pagination.go
```diff
@@ -0,0 +1,126 @@
...
+	PerPage    uint64 `json:"per_page"`
+	Total      uint64 `json:"total"`
+	TotalPages uint64 `json:"total_pages"`
+}
+
+// SliceBounds returns safe [start,end) bounds for slicing a collection with the given total length.
+func (p Pagination) SliceBounds(total uint64) (start, end uint64) {
+	if p.Page == 0 || p.PerPage == 0 {
+		return 0, 0
+	}
```
iurii-ssv: Since these values are validated to not be `0` upon construction of `Pagination` object (in `ToPagination`) we could probably just `panic` here to explicitly catch any bugs in our code (instead of returning zeroes which can make the debugging harder)

api/handlers/validators/validators.go
julienh-ssv: Done

julienh-ssv:
> @julienh-ssv This is a breaking change in the API, can we still use `v1/validators` without pagination the same way?

@y0sher yes it's fully retro-compatible

api/handlers/validators/validators.go
```diff
@@ -55,10 +76,34 @@ func (h *Validators) List(w http.ResponseWriter, r *http.Request) error {
...
+		for i, share := range shares {
+			response.Data[i] = validatorFromShare(share)
+		}
+		return api.Render(w, r, response)
+	}
+
+	// Ensure deterministic ordering for pagination.
+	sort.Slice(shares, func(i, j int) bool {
+		return bytes.Compare(shares[i].ValidatorPubKey[:], shares[j].ValidatorPubKey[:]) < 0
+	})
```
iurii-ssv, line 94: There is another edge-case (besides "deterministic ordering") - a share is added/removed to/from the collection while somebody is paginating over it (a share is added in between requests for different pages) ... do we want to address that somehow ?

I'm not sure if a duplicate/skipped share in the response resulting from this is a critical issue or not.

iurii-ssv (commented):
Looks good, got just [1 question about an edge-case](https://github.com/ssvlabs/ssv/pull/2703#discussion_r2877315043)

api/handlers/validators/validators.go
julienh-ssv: No, I thought about it but I don't think it's worth the complexity. The goal of pagination here is to split a slow query into several faster queries but in the 5 second time-window that either would take just assume that you don't get new data and re-query 5 seconds later if you need an update. In practice there is so much lag introduced by the blockchain during validator registration that these 5 seconds don't matter

oleg-ssvlabs (dismissed):

api/pagination.go
julienh-ssv: Yeah but we need a stronger validation at construction to actually support it. I can add it but it's more work than I though was worth it. I pushed a new commit with the changes.

iurii-ssv (approved):

oleg-ssvlabs (approved):

y0sher (approved):

PR #2709, opened 2026-03-02
switch ssv-node storage to Pebble DB

nkryuchkov:
The PR applies Pebble DB usage for both ssv-node and exporter instead of Badger DB for ssv-node and Pebble DB for exporter.


nkryuchkov:
@greptileai please review this PR

nkryuchkov:

storage/migration/db_migration.go
```diff
@@ -0,0 +1,554 @@
...
+
+			copied++
+			pending++
+
+			if pending >= batchSize {
+				if err := flushBatch(); err != nil {
+					return err
+				}
+			}
+		}
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai fixed, please review it again

nkryuchkov:
@greptileai review it again

storage/migration/db_migration.go
```diff
@@ -0,0 +1,574 @@
...
+		return bdb, false, nil
+	}
+	if !isBadgerTruncateRequiredError(err) {
+		return nil, false, err
+	}
+
+	recoverOpt := badgerdb.DefaultOptions(path)
+	recoverOpt.ReadOnly = false
+	recoverOpt.Logger = nil
+
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai

Your suggestion doesn't work

```
storage/migration/db_migration.go:385:13: recoverOpt.Truncate undefined (type badger.Options has no field or method Truncate)
```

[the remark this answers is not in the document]
nkryuchkov: @greptileai `ErrTruncateNeeded` is returned only when `Readonly` is `true`, but we have `recoverOpt.ReadOnly = false`. So why do we need to handle `ErrTruncateNeeded`?

```go
func (mt *memTable) UpdateSkipList() error {
	if mt.wal == nil || mt.sl == nil {
		return nil
	}
	endOff, err := mt.wal.iterate(true, 0, mt.replayFunction(mt.opt))
	if err != nil {
		return y.Wrapf(err, "while iterating wal: %s", mt.wal.Fd.Name())
	}
	if endOff < mt.wal.size.Load() && mt.opt.ReadOnly {
		return y.Wrapf(ErrTruncateNeeded, "end offset: %d < size: %d", endOff, mt.wal.size.Load())
	}
	return mt.wal.Truncate(int64(endOff))
}
```

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR. Don't limit the number of comments to 3 — post all the comments you have

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again

nkryuchkov:
@greptileai review this again

y0sher:
## Staging Test Report (operators 93-96, BadgerDB → PebbleDB migration)

**Branch:** `pebbledb-all`
**Baseline:** operators 29-32 running `stage` (BadgerDB)

### Migration: PASS

All 4 nodes migrated from BadgerDB to PebbleDB on startup:

| Node | Keys Migrated | Status |
|------|--------------|--------|
| ssv-node-93 | 93,618 | Success |
| ssv-node-94 | 93,618 | Success |
| ssv-node-95 | 93,624 | Success |
| ssv-node-96 | 100,359 | Success |

All 9 schema migrations preserved. Logs confirm `"using pebble db"` and `"migrated legacy badger db to pebble"`.

### Memory: 55% REDUCTION

| Node Type | Memory (RSS) |
|-----------|-------------|
| **PebbleDB avg** | **~550 MB** |
| **BadgerDB avg** | **~1,280 MB** |

| Node | Memory |
|------|--------|
| PebbleDB node-93 | 530 MB |
| PebbleDB node-94 | 537 MB |
| PebbleDB node-95 | 562 MB |
| PebbleDB node-96 | 570 MB |
| BadgerDB node-29 | 1,186 MB |
| BadgerDB node-30 | 1,360 MB |
| BadgerDB node-31 | 1,269 MB |
| BadgerDB node-32 | 1,299 MB |

⚠️ PebbleDB nodes were freshly restarted (~20 min uptime). **Will re-check after 1h+ to confirm memory holds.**

### Goroutines: ~100 fewer per node

| PebbleDB (93-96) | BadgerDB (29-32) |
|---|---|
| 18,530 – 18,581 | 18,654 – 18,657 |

Likely from eliminated BadgerDB background GC goroutines.

### Functional Correctness: PASS

- Duties executing: AGGREGATOR, SYNC_COMMITTEE_CONTRIBUTION, VALIDATOR_REGISTRATION
- Consensus completing in round 1-2
- Event syncer tracking blocks at 100% progress
- Zero storage errors
- E2M epoch 84379 effectiveness: 99.20% (network-wide, no regression)

### Errors: PASS (all pre-existing)

- `"could not publish p2p message: validation ignored"` — P2P, not storage
- `"failed to find event by ID"` — appeared on both PebbleDB AND BadgerDB nodes at same time (network-wide event)

### P2P Peers: Converging

PebbleDB nodes at 3-9 peers (just restarted), baseline at 10-11. Expected to converge.

### Note on path

Nodes log a WARN: `"using legacy pebble database path"` — configured path is `db-holesky-stage-new` but actual pebble path is `db-holesky-stage-new-pebble`. Not an error, just informational.

### Next Steps

- [ ] Re-check memory after 1h+ uptime to confirm 55% reduction holds
- [ ] Re-check P2P peers have converged to 10+
- [ ] Consider longer soak (24h+) before mainnet

---

olegshmuelov (dismissed):
LGTM!

One thing for release notes @y0sher: the migration requires roughly 2x the current database size in free disk space (both Badger and Pebble coexist during and after migration). Operators with large DBs should be warned before upgrading. The code handles ENOSPC gracefully (clear error, resumable), but better to warn proactively than have operators discover it at runtime.

cli/operator/node.go
```diff
@@ -830,37 +824,64 @@ func init() {
...
+			zap.Error(closeErr),
+		)
+		return errors.Join(setupErr, fmt.Errorf("close db after setup error: %w", closeErr))
+	}
+	if dbPath != cfg.DBOptions.Path {
+		logger.Warn("using legacy pebble database path",
+			zap.String("configured_path", cfg.DBOptions.Path),
+			zap.String("legacy_path", dbPath),
+		)
+	}
```
iurii-ssv: Nit: I would move this check above, to go right after the `dbPath := plan.PebblePath` line, it feels out of place here

Also, would be nice to either expand the `using legacy pebble database path` log-line or add a clarifying comment here for "what is our intent here with respect to `dbPath`" (basically it's not entirely clear what "legacy" means for when `cfg.DBOptions.Path` is also configured)

network/p2p/testutils.go
```diff
@@ -133,34 +136,78 @@ func (mockSignatureVerifier) VerifySignature(operatorID spectypes.OperatorID, me
...
 		return nil, err
 	}
+	dbCloser := newOnceCloser(db.Close)
+	t.Cleanup(func() {
+		_ = dbCloser.Close()
+	})
+	closeOnErr := func(err error) (network.P2PNetwork, error) {
+		_ = dbCloser.Close()
+		return nil, err
+	}
```
iurii-ssv: Could you clarify why we need to call `dbCloser.Close()` both: from `closeOnErr` as well as upon `t.Cleanup` ? Seems like that would result in a duplicate call.

Additionally, we also set the `p2pNetworkWithDB.dbCloser` that seemingly will result into 3rd `.Close()` call ?

ssvsigner/e2e/testenv/key_managers.go
```diff
@@ -45,37 +43,54 @@ func (env *TestEnvironment) createLocalKeyManager(logger *zap.Logger) error {
 		env.operatorKey,
 	)
 	if err != nil {
+		if closeErr := localDB.Close(); closeErr != nil {
+			return errors.Join(
+				fmt.Errorf("failed to create local key manager: %w", err),
+				fmt.Errorf("close local database after key manager setup failure: %w", closeErr),
+			)
+		}
 		return fmt.Errorf("failed to create local key manager: %w", err)
```
iurii-ssv: Nit: would be nice to de-dup `failed to create local key manager` error

ssvsigner/e2e/testenv/key_managers.go
```diff
@@ -45,37 +43,54 @@ func (env *TestEnvironment) createLocalKeyManager(logger *zap.Logger) error {
...
 	if err != nil {
+		if createdRemoteDB {
+			if closeErr := remoteDB.Close(); closeErr != nil {
+				return errors.Join(
+					fmt.Errorf("failed to create remote key manager: %w", err),
+					fmt.Errorf("close remote database after key manager setup failure: %w", closeErr),
+				)
+			}
+		}
 		return fmt.Errorf("failed to create remote key manager: %w", err)
```
iurii-ssv: Nit: would be nice to de-dup `failed to create remote key manager` error

ssvsigner/ekm/signer_storage_test.go
```diff
@@ -29,21 +29,26 @@ func _byteArray(input string) []byte {
...
 	if err != nil {
 		return nil, func() {}
 	}
 
 	s := NewSignerStorage(db, networkconfig.TestNetwork.Beacon, logger)
-	return s, func() {
-		db.Close()
-	}
+	return s, func() {}
 }
```
iurii-ssv: There is no longer a need to return 2nd value from this func - since it's always a no-op now

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	badgerImportDoneMarkerFileName       = ".ssv-badger-import.done.json"
+	badgerImportInProgressMarkerFileName = ".ssv-badger-import.inprogress.json"
+	defaultImportBatchSize               = 10_000
+	importProgressLogEveryBatches        = 10
+)
+
+type importBatchCommitHook func(committedBatches int, copiedKeys int) error
+
+type badgerImportMarker struct {
+	SourcePath  string `json:"source_path"`
```
iurii-ssv: Not sure what "Source" refers to, I would maybe `SourcePath`->`BadgerDirPath` to clarify

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+const (
+	badgerImportDoneMarkerFileName       = ".ssv-badger-import.done.json"
+	badgerImportInProgressMarkerFileName = ".ssv-badger-import.inprogress.json"
+	defaultImportBatchSize               = 10_000
+	importProgressLogEveryBatches        = 10
+)
+
+type importBatchCommitHook func(committedBatches int, copiedKeys int) error
+
+type badgerImportMarker struct {
```
iurii-ssv, line 46: Would be nice to leave brief comments for some fields on `badgerImportMarker`, specifically I'm wondering what `Keys` is for

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+
+func markerExistsForSource(path string, badgerPath string) (bool, error) {
+	marker, exists, err := readMarker(path)
+	if err != nil {
+		return false, err
+	}
+	if !exists {
+		return false, nil
+	}
+	return marker.SourcePath == "" || samePath(marker.SourcePath, badgerPath), nil
```
iurii-ssv: This `marker.SourcePath == ""` doesn't make sense to me, could you explain (and leave a clarifying comment in the code) ?

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	badgerIsNonEmpty := false
+	if badgerStateKnown {
+		hasBadgerData = knownBadgerExists
+		badgerIsNonEmpty = knownBadgerNonEmpty
+	} else {
+		hasBadgerData, badgerIsNonEmpty, err = badgerDirState(logger, badgerPath)
+		if err != nil {
+			return false, 0, err
+		}
+	}
```
iurii-ssv: I'm not sure why we don't consolidate the logic of figuring out those `hasBadgerData` and `badgerIsNonEmpty` values between "plan-parsing" and "here" ... would be more straightforward to have it parsed in one place

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("remove in-progress import marker: %w", err)
+	}
+	if err == nil {
+		if err := syncDirectory(filepath.Dir(inProgressMarkerPath(pebblePath))); err != nil {
+			return fmt.Errorf("fsync marker directory after marker removal: %w", err)
+		}
+	}
+	return nil
+}
```
iurii-ssv, line 677: The `os.Remove` error is seemingly swallowed here, while `syncDirectory` is not ... we should probably be consistent about it and return error for every outcome (or remove `error` from `removeBadgerImportInProgressMarker` func declaration entirely)

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+		return false, 0, nil
+	}
+	if !badgerIsNonEmpty {
+		logger.Info("legacy badger database is empty, skipping import", zap.String("path", badgerPath))
+		return false, 0, nil
+	}
+
+	if err := ctx.Err(); err != nil {
+		return false, 0, err
+	}
```
iurii-ssv: This `ctx.Err()` check here, and a similar one below doesn't seem to do much (just complicate the code) ... it would maybe make sense if it was done in-between 2 operations that BOTH take long time - but it seems `copyBadgerToPebble` is the only operation that might take long here

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+		return false, 0, wrapNoSpaceImportError(err, badgerPath, pebblePath)
+	}
+	if err := ctx.Err(); err != nil {
+		return false, 0, err
+	}
+	if err := writeBadgerImportDoneMarker(pebblePath, badgerPath, copied); err != nil {
+		return false, 0, wrapNoSpaceImportError(err, badgerPath, pebblePath)
+	}
+	if err := removeBadgerImportInProgressMarker(pebblePath); err != nil {
+		logger.Warn("failed to remove in-progress badger import marker after successful migration; marker will be retried on next startup",
```
iurii-ssv: `marker will be retried` -> `marker will be removed`

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	pending := 0
+	committedBatches := 0
+	batch := db.NewBatch()
+	defer func() {
+		if batch != nil {
+			_ = batch.Close()
+		}
+	}()
+
+	flushBatch := func() error {
```
iurii-ssv: I feel like `batch` lifecycle is smeared/mixed between this code and the code inside of `flushBatch`, would be nice if new/close happened in one place instead

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	if recovered {
+		logger.Warn(
+			"badger value log required truncation during pebble import; proceeding with recovered badger state",
+			zap.String("badger_path", badgerPath),
+		)
+	}
+
+	batchSize := defaultImportBatchSize
+	copied := 0
+	pending := 0
```
iurii-ssv: Could probably just use `batch.Count()` instead of `pending` for clarity

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+		}
+		if hasKeyRegistry && hasValueLog {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func copyBadgerToPebble(ctx context.Context, logger *zap.Logger, badgerPath string, db *pebble.DB, hook importBatchCommitHook) (int, error) {
```
iurii-ssv: `hook`->`onBatchCommitHook` would make the code a bit more informative

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+	defer func() { _ = bdb.Close() }()
+	if recovered {
+		logger.Warn(
+			"badger value log required truncation during pebble import; proceeding with recovered badger state",
+			zap.String("badger_path", badgerPath),
+		)
+	}
+
+	batchSize := defaultImportBatchSize
+	copied := 0
```
iurii-ssv: `copied`->`keysCopiedTotal`

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+		return 0, err
+	}
+	if err := flushBatch(); err != nil {
+		return 0, err
+	}
+
+	return copied, nil
+}
+
+func openBadgerForImport(path string) (*badgerdb.DB, bool, error) {
```
iurii-ssv: Perhaps `openBadgerForImport`->`openBadgerReadOnly` since "ForImport" makes me think we are importing data INTO Badger

iurii-ssv (commented):
Looks quite comprehensive, left some minor suggestions.

Although I'm struggling a bit to get the full picture, it maybe feels like the intermediary `PebbleDBPlan` concept complicates things more than helps (or maybe it just could use some restructuring) ... I get the idea to separate the "parsing of the current state/plan" (represented by `PebbleDBPlan`) from the "data-migration execution" ... it's just a bit hard to map it onto a simple list of scenarios I would expect us to have to implement such a migration, which would be something like:
- pebble exists, badger does not -> just use pebble
- pebble exists, badger exists -> resolve which one is the source of truth and finish the migration if needed
- pebble doesn't exist, badger exists -> run the migration
- ...

maybe the code in `migrateBadgerToPebbleIfNeeded` func could be restructured a bit & supplied with step-by-step comments to clarify where/when each of those cases (from my list above) takes place - currently it kinda mixed a lot of things together making it hard to see what's going on

storage/migration/db_migration.go
```diff
@@ -0,0 +1,812 @@
...
+
+// PebbleDBPlan describes which Pebble path to open and whether legacy Badger
+// data should be imported into it on startup.
+type PebbleDBPlan struct {
+	PebblePath       string
+	BadgerImportPath string
+	BadgerStateKnown bool
+	BadgerExists     bool
+	BadgerNonEmpty   bool
+}
```
iurii-ssv: To complement my other comments regarding `PebbleDBPlan` struct,

I wonder if we could make it a bit more straightforward with respect to actually "define a plan that needs to be done", in its current form it doesn't look like a "Plan" to me, namely:
- `PebblePath` and `BadgerImportPath` are just data, they don't "tell what to do (like a plan would", so maybe these should be converted to `bools` (that tell "what to do") while the paths themselves could be passed around as func-params (like they already are in a bunch of places)
- `BadgerStateKnown` probably needs a comment at least, because I am not sure what "unknown badger state" would mean (or how it would affect our execution "Plan")
- do we need both of `BadgerExists` and `BadgerNonEmpty` ? Shouldn't we just care about a single concept such as `BadgerDataExists` ? There probably is a reason to have 2 separate flags, it's just hard to figure out just by looking at the code (so I'm thinking maybe there is a way to simplify it down to just 1 flag)

also, I'm not sure why it is `PebbleDBPlan` and not `MigrationPlan` ... seems like we are mixing/merging 2 separate things together here (the migration, and the pebble-db instance to be used after migration has completed) - I'd rather: 1) plan migration -> 2) prepare db-instance(s) involved -> 3) do migration -> 4) choose pebble-db instance to use for SSV node startup/operations

operator/fee_recipient/controller_test.go
```diff
@@ -313,8 +311,9 @@ func TestSubmitProposal(t *testing.T) {
 
 func createStorage(t *testing.T) (basedb.Database, registrystorage.Shares) {
 	logger := log.TestLogger(t)
-	db, err := kv.NewInMemory(logger, basedb.Options{})
+	db, err := pebble.NewTemporary(logger, basedb.Options{})
 	require.NoError(t, err)
+	t.Cleanup(func() { _ = db.Close() })
```
iurii-ssv: This func doesn't need to return it's first result-value since it's never used anymore

storage/pebble/pebble.go
```diff
@@ -14,7 +16,8 @@ var _ basedb.Database = &DB{}
 
 type DB struct {
 	*pebble.DB
-	logger *zap.Logger
+	logger      *zap.Logger
+	cleanupPath string
 }
```
iurii-ssv, line 18: It's a bit hacky to define `cleanupPath` here since it's only used for when the DB is created "as temporary" ... I would define a separate wrapper-struct for it to keep it clean & straightforward

https://github.com/ssvlabs/ssv/pull/2770 could do it

nkryuchkov:
@greptileai please review the latest changes

network/p2p/testutils.go
nkryuchkov: We use `sync.Once`, so the 2nd and 3rd operations will be no-op. I wouldn't use it for non-test code, but it looks convenient for me for tests. `t.Cleanup` indeed looks redundant here, I removed it

storage/migration/db_migration.go
nkryuchkov: We ignore only `os.ErrNotExist`, on other errors we return `fmt.Errorf("remove in-progress import marker: %w", err)`

storage/migration/db_migration.go
nkryuchkov: It's become a method, but I think `Pending()` fits better

storage/migration/db_migration.go
nkryuchkov: renamed

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+		return resolveSingleNonEmptyPebblePathLayout(layout.canonical.path, layout.canonical.path, badgerFilesPresent)
+	case layout.legacy.nonEmpty && !layout.canonical.nonEmpty:
+		return resolveSingleNonEmptyPebblePathLayout(layout.legacy.path, layout.canonical.path, badgerFilesPresent)
+	default:
+		return DBLayout{}, false, nil
+	}
+}
+
+func resolveSingleNonEmptyPebblePathLayout(pebblePath string, badgerPath string, badgerFilesPresent bool) (DBLayout, bool, error) {
+	done, inProgress, err := badgerImportMarkerState(pebblePath, badgerPath)
```
iurii-ssv: This func always returns `true` for its second return-value, can remove it ?

cli/operator/node.go
```diff
@@ -889,37 +883,63 @@ func init() {
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to open db: %w", err)
+		return nil, fmt.Errorf("resolve database layout: %w", err)
 	}
-
-	if err := applyMigrations(logger, beaconConfig, operatorPrivKey, db, cfg.DBOptions.Path); err != nil {
-		return nil, fmt.Errorf("apply migrations: %w", err)
+	// dbPath may differ from the configured base path when startup resolves a
+	// legacy Badger/Pebble on-disk layout to the correct Pebble directory.
+	dbPath := layout.PebblePath
```
iurii-ssv: I would skip defining the `dbPath` here and use/inline `layout.PebblePath` instead - it would make for a clearer code down below,

eg. here you would immediately see that `dbPath` is actually always a "pebble path: `layout.PebblePath`" (not "badger path"):
```go
			logger.Info("migrated legacy badger db to pebble",
				zap.String("from_path", layout.BadgerImportPath),
				zap.String("to_path", dbPath),
				zap.Int("keys", migratedKeys),
			)
```

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+	pebbleEmpty, err := isPebbleEmpty(db)
+	if err != nil {
+		return false, 0, err
+	}
+
+	badgerExists, badgerNonEmpty, err := badgerDirState(logger, badgerPath)
+	if err != nil {
+		return false, 0, err
+	}
+	badger := badgerState{exists: badgerExists, nonEmpty: badgerNonEmpty}
```
iurii-ssv: Nit: we should use consistent form for both (not mix these):
- either `pebbleEmpty` + `badgerEmpty` (I would choose this one)
- or `pebbleNonEmpty` + `badgerNonEmpty`

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+// 4. Pebble is empty and Badger has data -> run the import.
+// 5. Pebble is empty and Badger is gone/empty -> nothing to migrate.
+func MigrateBadgerToPebbleIfNeeded(
+	ctx context.Context,
+	logger *zap.Logger,
+	badgerPath string,
+	pebblePath string,
+	db *pebble.DB,
+) (bool, int, error) {
+	return migrateBadgerToPebbleIfNeeded(ctx, logger, badgerPath, pebblePath, db, nil)
```
iurii-ssv, line 222: Both `MigrateBadgerToPebbleIfNeeded` and `migrateBadgerToPebbleIfNeeded` could use named variables

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+				pebblePath,
+				badgerPath,
+			)
+		default:
+			logger.Warn("resuming interrupted badger-to-pebble import",
+				zap.String("badger_path", badgerPath),
+				zap.String("pebble_path", pebblePath),
+			)
+		}
+	}
```
iurii-ssv: I think this would read better as a bunch of `if ...`s instead of that large `switch` ... `switch` is best when it covers ALL the options "at once", but here we are just proceeding to another `switch` after this one (which also btw could be represented as a couple of if-statements instead)

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+
+type badgerState struct {
+	exists   bool
+	nonEmpty bool
+}
+
+type pebbleDirState struct {
+	path     string
+	exists   bool
+	nonEmpty bool
```
iurii-ssv: maybe it's me, but I think it's more common to use the `empty` for things like this because the double-negation `!nonEmpty` is hard to read

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+	if err != nil {
+		return DBLayout{}, err
+	}
+	if err := validateEmptyPebbleDoneMarkers(basePath, layout); err != nil {
+		return DBLayout{}, err
+	}
+	badgerFilesPresent, err := hasBadgerFiles(basePath)
+	if err != nil {
+		return DBLayout{}, fmt.Errorf("check badger path %q: %w", basePath, err)
+	}
```
iurii-ssv, line 76: Badger opened on every post-migration startup. Even after the done marker is written, [badgerDirState](https://github.com/ssvlabs/ssv/blob/1452a990be4501e32f952d76dd2c994a97f454a6/storage/migration/db_migration.go#L380) still stats the legacy directory and (depending on branch) opens it read-only. For already-migrated nodes this is pure IO waste — the done marker alone should short-circuit the layout resolution. Consider returning early in ResolveDBLayout the moment the done marker is observed.

cli/operator/node.go
```diff
@@ -889,37 +883,63 @@ func init() {
...
+		}
+		logger.Warn("failed to close db after setup error",
+			zap.String("path", dbPath),
+			zap.Error(closeErr),
+		)
+		return errors.Join(setupErr, fmt.Errorf("close db after setup error: %w", closeErr))
+	}
+
+	if layout.BadgerImportPath != "" {
+		migrated, migratedKeys, err := storagemigration.MigrateBadgerToPebbleIfNeeded(
```
iurii-ssv, line 921: During import, both Badger and Pebble directories coexist. This is mentioned in the PR description but not in user-facing startup logs; operators on tight disks will discover this at the worst moment. Suggest a preflight statfs check at [copyBadgerToPebble](https://github.com/ssvlabs/ssv/blob/1452a990be4501e32f952d76dd2c994a97f454a6/storage/migration/db_migration.go#L497) entry with a clear error if free space < badger size × 1.2.

Probably a good idea to add a check like this.

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+	if err != nil {
+		return fmt.Errorf("marshal marker: %w", err)
+	}
+
+	tmpPath := path + ".tmp"
+	// #nosec G304 -- marker path is derived from configured DB path plus fixed marker filenames.
+	tempFile, err := os.OpenFile(tmpPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0o600)
+	if err != nil {
+		return fmt.Errorf("open marker temp file: %w", err)
+	}
```
iurii-ssv, line 731: Nit: [writeMarker](https://github.com/ssvlabs/ssv/blob/1452a990be4501e32f952d76dd2c994a97f454a6/storage/migration/db_migration.go#L710) does not os.Remove the temp file if fsync or rename fails partway, leaving a stray .tmp next to the marker. defer os.Remove(tmp) after the initial create, cleared on success, would be robust.

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+func (b *pebbleImportBatch) Close() error {
+	if b.batch == nil {
+		return nil
+	}
+	err := b.batch.Close()
+	b.batch = nil
+	return err
+}
+
+func copyBadgerToPebble(ctx context.Context, logger *zap.Logger, badgerPath string, db *pebble.DB, onBatchCommitHook importBatchCommitHook) (int, error) {
```
iurii-ssv, line 503: This func could use some error-wrapping

iurii-ssv (commented):
Looks much cleaner! Left some minor suggestions. 

Also, we should probably create a task to "remove the Badger directory once the done marker has been present for, say, 24 hours" ?

cli/operator/node.go
iurii-ssv, line 981: Pre-existing:

`FullGC` passes the first and last keys directly to `DB.Compact`, but Pebble requires `start < end`. An empty DB calls `Compact(nil, nil, ...)`, and a single-key DB calls `Compact(k, k, ...)`, both of which return an error. Because [cli/operator/node.go#L970-L982](https://github.com/ssvlabs/ssv/blob/pebbledb-all/cli/operator/node.go#L970-L982) invokes `db.FullGC()` after migrations, a migration on a very small DB can fail startup even though the migration itself succeeded.

We should probably add a safe-guard inside of `FullGC` itself to not call `Compact` when `start < end`.

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+	// no usable Badger data.
+	switch {
+	case !badger.exists:
+		return false, 0, nil
+	case !badger.nonEmpty:
+		logger.Info("legacy badger database is empty, skipping import", zap.String("path", badgerPath))
+		return false, 0, nil
+	default:
+		return runBadgerImport(ctx, logger, badgerPath, pebblePath, db, inProgressMarkerExists, onBatchCommitHook)
+	}
```
iurii-ssv: The `Pebble is empty` part of this comment seems incorrect ? We can end up here, having logged the `resuming interrupted badger-to-pebble import` above (which basically contradicts this comment)

storage/migration/db_migration.go
```diff
@@ -0,0 +1,815 @@
...
+	pebblePath string,
+	db *pebble.DB,
+	inProgressMarkerExists bool,
+	onBatchCommitHook importBatchCommitHook,
+) (bool, int, error) {
+	if !inProgressMarkerExists {
+		if err := writeBadgerImportInProgressMarker(pebblePath, badgerPath); err != nil {
+			return false, 0, wrapNoSpaceImportError(err, badgerPath, pebblePath)
+		}
+	}
```
iurii-ssv, line 349: I'm not sure I understand what the purpose of `InProgressMarker` is:
- if it is it to prevent concurrent migrations from running (multiple SSV nodes accessing the same DB) ... it doesn't seem like it works for that
- if it is to "continue from where we left off", it doesn't track where we stopped (and it's probably an overkill for this anyway)

so it kinda seems like we could just use the `DoneMarker` alone and get the same results ?

storage/migration/db_migration.go
nkryuchkov: @iurii-ssv I'm not sure I understood the part about `badgerDirState`, can you please clarify it?

cli/operator/node.go
nkryuchkov: @iurii-ssv I had this check, but I decided to remove it because it's not reliable and the code is already complicated. Anyway, if migration fails due to a lack of space, it should interrupt the migration

storage/migration/db_migration.go
nkryuchkov: What paths do you mean where it's not removed?

storage/migration/db_migration.go
nkryuchkov: I'm not sure `DoneMarker` alone would be enough here. Historically we could already have both DBs populated without any migration marker at all, because Pebble was used under `cfg.DBOptions.Path + "-pebble"` while Badger stayed at `cfg.DBOptions.Path`. `InProgressMarker` is what lets us distinguish that kind of overlap from an actual interrupted migration that is safe to resume.

cli/operator/node.go
iurii-ssv: Why do you think it's not reliable ?

I think it makes sense to check the available space not only to make sure the migration succeeds, but also to ensure the SSV node is able to keep functioning after the migration too (e.g if we have 20 GB of space on disk with Badger occupying 10GB, and newly created Pebble occupying 9.5GB after migration is done ... we have 0.5GB left to work with - which might result in a failure shortly after, even before the Operator can spot & resolve this manually), right ?

I don't think we need to do the "space check" on every migration as a habit (although we maybe could add this as a generic sanity-check on startup? So we fail to start the node cleanly, instead of letting it run out of disk-space), but for this one specifically seems like it might prevent some pain for Operators.

nkryuchkov: I don't think it's reliable because we don't know how much space the Pebble DB will take. If we check for 1.2x of Badger, but Pebble uses 1.5x, then the check will be useless. If we check for 1.2x but Pebble uses 0.5x, the check could prevent migration even with enough space. Even if the size is roughly the same, but a disk has 15% extra space, not 20%, it might be fine, because we don't need Pebble DB after successful migration.

iurii-ssv: We could probably estimate the relative size (Badger vs Pebble) by running this migration once on some operator from our stage env (so we'll know what *.*x value is reasonable to use).

I mean, I agree that it's a heuristic/judgement that might result in false-positives ... I just think it's worthwhile even so.

cli/operator/node.go
```diff
@@ -889,37 +883,62 @@ func init() {
...
+				zap.String("from_path", layout.BadgerImportPath),
+				zap.String("to_path", layout.PebblePath),
+				zap.Int("keys", migratedKeys),
+			)
+		}
+	}
 
-	if err := applyMigrations(logger, beaconConfig, operatorPrivKey, db, dbPath); err != nil {
-		return nil, fmt.Errorf("apply migrations: %w", err)
+	if err := applyMigrations(logger, beaconConfig, operatorPrivKey, db, layout.PebblePath); err != nil {
```
momosh-ssv, line 940: What do you think about deferring the done-marker write until after `applyMigrations` succeeds? I fear that if `applyMigrations` crashes mid-run on a fresh import, the next start sees the done marker present plus a half-transformed Pebble. The resolver then trusts Pebble as the source of truth, and recovery depends on each schema migration being individually idempotent — which is a per-author guarantee rather than a structural one. Either move the done-marker write after `applyMigrations`, or add a separate `schema-applied` sentinel and refuse to start when the done marker exists but the schema sentinel doesn't.

storage/migration/db_migration.go
```diff
@@ -0,0 +1,821 @@
...
+	if !exists {
+		return false, nil
+	}
+	// Empty BadgerDirPath is accepted for backward compatibility with early marker
+	// files that were written before the source path was persisted.
+	return marker.BadgerDirPath == "" || samePath(marker.BadgerDirPath, badgerPath), nil
+}
+
+func samePath(left string, right string) bool {
```
momosh-ssv, line 698: Could this lead to spurious re-imports on operator setups that bind-mount or symlink the data dir? Seems that `samePath` only does `Clean`+`Abs` but not symlink resolution, so e.g. `/var/lib/ssv -> /mnt/data/ssv` will produce a different absolute path on subsequent runs, `markerExistsForSource` returns false, and the migration tries to re-import on top of an already-imported Pebble. Maybe worth using `filepath.EvalSymlinks` with a fallback to `Clean`+`Abs` when EvalSymlinks fails (e.g. path doesn't exist yet).

cli/operator/node.go
```diff
@@ -889,37 +883,62 @@ func init() {
...
-
-func setupPebbleDB(
-	logger *zap.Logger,
-	beaconConfig *networkconfig.Beacon,
-	operatorPrivKey keys.OperatorPrivateKey,
-) (*pebble.DB, error) {
-	dbPath := cfg.DBOptions.Path + "-pebble" // opinionated approach to avoid corrupting old db location
-
-	db, err := pebble.New(logger, dbPath, &cockroachdb.Options{})
+	db, err := pebble.New(logger, layout.PebblePath, &cockroachdb.Options{})
```
momosh-ssv, line 904: Any reason we're not wiring zap into Pebble's Options here? I fear this is an observability regression vs Badger — Pebble's `WriteStallBegin/End`, `DiskSlow`, compaction stalls and manifest recovery events would otherwise go nowhere, and migrating the operator-facing data layer makes those signals more important, not less. Could we add a `pebble.Logger` adapter over zap on `Options.Logger`, an `EventListener` for stall/slow events, and expose `db.Metrics()` via the existing metrics handler?

momosh-ssv (commented):
One more that didn't fit inline (DropPrefix sits outside the diff hunks):

**`storage/pebble/pebble.go` `DropPrefix`** — could this lead to keys persisting after the call returns? Seems that the Badger version ran inside a transaction, while the Pebble version here uses a non-indexed batch plus a live-DB iterator, so writes that race with the loop can persist after `DropPrefix` returns. Worth considering `db.DeleteRange(prefix, upperBound, pebble.Sync)` — single atomic operation, and faster too.

PR #2710, opened 2026-03-03
ssvsigner: break circular dependency with ssv-node

nkryuchkov:
as per title

Closes ssvlabs/ssv-node-board#1057


nkryuchkov:

nkryuchkov:
@greptileai review this again, considering all comments on this PR

nkryuchkov:
@greptileai review this again, considering all comments on this PR

ekmadapter/database.go
```diff
@@ -0,0 +1,90 @@
...
+func (a *databaseAdapter) GetAll(txn ekm.ReadTxn, prefix []byte, handler func(int, ekm.Obj) error) error {
+	reader, err := asReader(txn)
+	if err != nil {
+		return err
+	}
+
+	return a.db.UsingReader(reader).GetAll(prefix, func(i int, obj basedb.Obj) error {
+		return handler(i, ekm.Obj{Key: obj.Key, Value: obj.Value})
+	})
+}
```
iurii-ssv, line 44: Probably better to keep these method-definition next to the `databaseAdapter` definition (so it's easy to find all struct-related methods)

ekmadapter/database.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+
+	"github.com/ssvlabs/ssv/ssvsigner/ekm"
+	"github.com/ssvlabs/ssv/storage/basedb"
+)
+
+// NewDatabaseAdapter bridges node's basedb.Database to ekm.Database.
+func NewDatabaseAdapter(db basedb.Database) ekm.Database {
+	return &databaseAdapter{db: db}
+}
```
iurii-ssv, line 13: I'm not sure I understand why this package is:
- not residing under `ssvsigner` directory (it does reference things like `ekm.Database` that are under `ssvsigner` ... so might as well move it there too)
- has "adapter" in the package/primitives names ... I see that it basically "converts one interface to another" but seems like we could just call it/use it as `ekm.NewDatabase(...` passing in the underlying storage too (the "adapter" naming adds unnecessary complexity IMO)

iurii-ssv (dismissed):
Good stuff!

ekmadapter/database.go
nkryuchkov: This package stays in `ssv` intentionally because it depends on node-only `storage/basedb`, while `ssvsigner` must remain importable without any dependency back on the root module.

The adapter naming is also intentional: its job is specifically to translate the node's `basedb.Database` into `ekm.Database` without teaching `ekm` about root-module storage types. Moving this into `ssvsigner` or folding it into `ekm` would reintroduce the boundary leak we split out in this PR.

nkryuchkov: Added a `doc.go` with an explanation

iurii-ssv: Got it.

Makefile
```diff
@@ -53,11 +53,13 @@ full-test:
 	@echo "Running all tests"
 	@go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 ./...
 	@cd ssvsigner && go test -tags blst_enabled -timeout 20m ${COV_CMD} -p 1 ./...
-
 .PHONY: unit-test
```
iurii-ssv: Seems like we should keep this newline (typo, I assume) ?

iurii-ssv (dismissed):
Just 1 nit

olegshmuelov (dismissed):
LGTM.

One nit: the new `full-test` target doesn't include `-race`, `unit-test` does. Worth adding for consistency.

nkryuchkov:
> One nit: the new `full-test` target doesn't include `-race`, `unit-test` does. Worth adding for consistency.

I accidentally added `full-test` while merging, although we don't need it. Deleted

Makefile
nkryuchkov: We need to delete `full-test`, I fixed a merge conflict here incorrectly

iurii-ssv (dismissed):

olegshmuelov (dismissed):

iurii-ssv (dismissed):

scripts/ssvsigner_boundary.sh
```diff
@@ -0,0 +1,20 @@
...
+set -eo pipefail
+
+echo -e "\033[36mINFO\033[0m [ssvsigner-boundary] Verifying ssvsigner does not import root ssv packages"
+
+OUT="$(
+  cd ssvsigner
+  go list -f '{{range .Imports}}{{println .}}{{end}}{{range .TestImports}}{{println .}}{{end}}{{range .XTestImports}}{{println .}}{{end}}' ./... \
+    | sort -u \
+    | grep '^github.com/ssvlabs/ssv/' \
+    | grep -Ev '^github.com/ssvlabs/ssv/ssvsigner($|/)' || true
```
olegshmuelov: The `|| true` at the end of the pipeline prevents failure when `grep` finds no forbidden imports (the success case), but it also swallows `go list` failures. If `go list` errors out (syntax error, unresolved dependency), `OUT` is empty and the script reports "Import boundary check passed" — making the guard fail-open.

Low severity because CI also runs build and tests which would catch broken packages, but the boundary check itself is unreliable in that scenario.

Suggestion:
```bash
cd ssvsigner

# Fail fast if go list itself errors.
ALL_IMPORTS="$(go list -f '{{range .Imports}}{{println .}}{{end}}{{range .TestImports}}{{println .}}{{end}}{{range .XTestImports}}{{println .}}{{end}}' ./...)"

OUT="$(echo "$ALL_IMPORTS" \
  | sort -u \
  | grep '^github.com/ssvlabs/ssv/' \
  | grep -Ev '^github.com/ssvlabs/ssv/ssvsigner($|/)' || true
)"
```

olegshmuelov (dismissed):
One minor comment

ssvsigner/ekm/contracts.go
```diff
@@ -0,0 +1,59 @@
...
+
+// NetworkContext groups network parameters used by key managers.
+// Name must match the network identifier represented by Beacon (for example
+// networkconfig.Beacon.Name), because Name is used as the storage key prefix
+// for slashing-protection data.
+type NetworkContext struct {
+	Name                  string
+	Beacon                BeaconNetwork
+	GenesisValidatorsRoot phase0.Root
+}
```
ljuba-ssv: Nit: before this PR, a Name/Beacon mismatch was structurally impossible (single `*networkconfig.Beacon` carried both). Now they are separate fields set by hand at ~8 call sites. A mismatch would silently write slashing-protection data under the wrong storage prefix.

All current call sites are correct (`Name` always derived from the same config as `Beacon`), so no active bug. But since the consequence is severe (bypassed slashing protection), consider a constructor that validates the coupling, or at minimum, make `Name` a computed property if `BeaconNetwork` can expose it.

ljuba-ssv (dismissed):

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2711, opened 2026-03-04
exporter: disable non-essential services

nkryuchkov:
## Summary

Clean up exporter mode by disabling services and flows that are only needed for operator nodes.

## Changes

- stop requiring `OPERATOR_KEY` in exporter mode
- skip signer/key manager setup for exporter
- skip fee recipient, doppelganger, validator status reporting, and validator registration paths
- skip the duty scheduler in exporter `standard` mode
- keep the required scheduler/network paths in exporter `archive` mode
- relax config lock compatibility for exporter so existing DBs still work
- remove exporter-related operator identity plumbing that is no longer needed

Closes ssvlabs/ssv-node-board#1050


nkryuchkov:

cli/operator/node.go
```diff
@@ -727,6 +727,27 @@ func ensureNoMissingKeys(
...
+) {
+	if !cfg.ExporterOptions.Enabled || !operatorDataStore.OperatorIDReady() {
+		return
+	}
+
+	operatorID := operatorDataStore.GetOperatorID()
+	shares := nodeStorage.Shares().List(nil, registrystorage.ByOperatorID(operatorID))
+	if len(shares) == 0 {
+		return
+	}
```
y0sher: operator doesn't really have operator ID so I don't really understand this code. Unless `operatorDataStore.OperatorIDReady` returns some data for exporter which is incorrect.

nkryuchkov: It's still a draft

operator/node.go
```diff
@@ -214,16 +214,19 @@ func (n *Node) Start(ctx context.Context) error {
...
-
-	go func() {
-		if err := n.validatorOptions.DoppelgangerHandler.Start(ctx); err != nil {
-			n.logger.Error("Doppelganger monitoring exited with error", zap.Error(err))
-		}
-	}()
+	if n.exporterOptions.Enabled {
+		n.logger.Info("exporter mode: skipping fee recipient, validator status reporting and doppelganger services")
+	} else {
+		go n.feeRecipientCtrl.Start(ctx)
```
y0sher, line 244: we might want to have correct fee recipient for exporter as well just so we have complete and correct data on /validators endpoint

nkryuchkov:

cli/operator/node.go
nkryuchkov: removed, resolving

operator/node.go
nkryuchkov: We don't return fee recipient on `/validators`:

```go
type validatorJSON struct {
	PubKey          api.Hex                `json:"public_key"`
	Index           phase0.ValidatorIndex  `json:"index"`
	Status          string                 `json:"status"`
	ActivationEpoch phase0.Epoch           `json:"activation_epoch"`
	ExitEpoch       phase0.Epoch           `json:"exit_epoch"`
	Owner           api.Hex                `json:"owner"`
	Committee       []spectypes.OperatorID `json:"committee"`
	Quorum          uint64                 `json:"quorum"`
	PartialQuorum   uint64                 `json:"partial_quorum"`
	Graffiti        string                 `json:"graffiti"`
	Liquidated      bool                   `json:"liquidated"`
}
```

nkryuchkov:

nkryuchkov:

nkryuchkov:

cli/operator/node.go
```diff
@@ -322,29 +328,34 @@ var StartNodeCmd = &cobra.Command{
...
+		} else if !cfg.ExporterOptions.Enabled {
 			if err := ensureOperatorPrivateKey(nodeStorage, operatorPrivKey, operatorPrivKeyPEM); err != nil {
 				logger.Fatal("could not save operator private key", zap.Error(err))
 			}
 		}
 
-		logger.Info("successfully loaded operator keys", zap.String(fields.FieldPubKey, operatorPubKeyBase64))
+		if !cfg.ExporterOptions.Enabled {
+			logger.Info("successfully loaded operator keys", zap.String(fields.FieldPubKey, operatorPubKeyBase64))
+		}
```
iurii-ssv: Perhaps we could have just 1 `if !cfg.ExporterOptions.Enabled` block with all the relevant code in it (instead of multiple separate checks spread all over the code) ?

Maybe we'll still have multiple of these `if !cfg.ExporterOptions.Enabled` check, but I think we could "merge" some

operator/node.go
```diff
@@ -157,14 +177,19 @@ func (n *Node) Start(ctx context.Context) error {
...
-	if err := n.dutyScheduler.Start(ctx); err != nil {
-		return fmt.Errorf("failed to run duty scheduler: %w", err)
+	// Start the duty scheduler in modes that use it.
+	if n.dutyScheduler != nil {
+		if err := n.dutyScheduler.Start(ctx); err != nil {
+			return fmt.Errorf("failed to run duty scheduler: %w", err)
+		}
+	} else if n.exporterOptions.Enabled && n.exporterOptions.Mode == exporter.ModeStandard {
+		n.logger.Info("exporter standard mode: skipping duty scheduler")
 	}
```
iurii-ssv: We can probably omit these additional conditions (`n.exporterOptions.Enabled && n.exporterOptions.Mode == exporter.ModeStandard`) in this `else` because just from `n.dutyScheduler != nil` already "contains this info"

iurii-ssv (dismissed):
LGTM, just minor suggestions

cli/operator/node.go
nkryuchkov: Initially, I wanted to put everything in one branch, but I didn't like what it looked like because the setup code became duplicated

iurii-ssv (dismissed):

operator/node.go
```diff
@@ -212,21 +237,28 @@ func (n *Node) Start(ctx context.Context) error {
...
 
-	if err := n.dutyScheduler.Wait(); err != nil {
-		n.logger.Fatal("duty scheduler exited with error", zap.Error(err))
+	if n.dutyScheduler != nil {
+		if err := n.dutyScheduler.Wait(); err != nil {
+			n.logger.Fatal("duty scheduler exited with error", zap.Error(err))
+		}
+	} else {
+		<-ctx.Done()
 	}
```
olegshmuelov, line 265: The `else` branch is intentionally reached in exporter standard mode. But if `shouldRunDutyScheduler` is the only thing preventing a regular operator node from having a nil `dutyScheduler`, any future refactoring that breaks that invariant would cause a regular node to silently skip duty scheduling with no log or error — blocking on `ctx.Done()` instead of running duties.

A small defensive check makes the intent explicit:

```go
if n.dutyScheduler != nil {
    if err := n.dutyScheduler.Wait(); err != nil {
        n.logger.Fatal("duty scheduler exited with error", zap.Error(err))
    }
} else if !n.exporterOptions.Enabled || n.exporterOptions.Mode != exporter.ModeStandard {
    n.logger.Fatal("duty scheduler is nil for non-exporter-standard node — this should never happen")
} else {
    <-ctx.Done()
}
```

operator/validator/controller.go
olegshmuelov, line 1014: `SetupRunners` (line 1071) has an explicit exporter guard. `SetupCommitteeRunners` does not, yet it uses `options.OperatorSigner` which is nil in exporter mode. Currently safe because `InitValidators` returns early and `ByOperatorID(0)` filters to an empty set. But the asymmetry is a maintenance risk — a future caller would hit a nil-pointer panic with no clear error.

Consider a matching guard for symmetry:

```go
func SetupCommitteeRunners(ctx context.Context, options *validator.Options) validator.CommitteeRunnerFunc {
    if options.ExporterOptions.Enabled {
        return func(...) (*runner.CommitteeRunner, error) {
            return nil, fmt.Errorf("cannot set up committee runners in exporter mode")
        }
    }
    // ...existing code...
}
```

olegshmuelov (approved):

iurii-ssv (approved):

ljuba-ssv (approved):

Issue #2705, opened 2026-02-26
duty-fetching could be more efficient during SSV node startup


iurii-ssv:
Desirable to implement after https://github.com/ssvlabs/ssv/issues/2619 has been addressed (so it's easy to apply/replicate across all duty-types).

From https://github.com/ssvlabs/ssv/pull/2699#pullrequestreview-3861766927, the current implementation of duty-fetching for most duty-types (all except for `ProposerHandler `, until https://github.com/ssvlabs/ssv/issues/2619 is resolved):

> the only downside to doing that I can think of would be: we'd be fetching duties for the current **and** next epoch on the `HandleInitialDuties` call **before** we proceed to executing them (meaning we'd be delaying the duty execution for the current/upcoming slot for no good reason - with the current implementation, and the implementation from your PR, I think we are fetching the duties for the next epoch only after we execute the duties for the current slot) ... while extra wait-time upon SSV node startup is undesirable - it isn't a big issue because SSV cluster typically has redundant nodes to perform duties even with 1 node being down for some time.

iurii-ssv:
Resolved already by those PRs linked above.

PR #2712, opened 2026-03-04
duty-fetching: simplify AttesterHandler

iurii-ssv:
This PR aims to clarify/document/simplify the behavior of `AttesterHandler` (also making the handling of reorgs and indices-changes more precise, especially at epoch-boundaries).

Notable changes:
- instead of having `fetchCurrentEpoch/fetchNextEpoch` dictating when to fetch duties we now have the exact epoch->intent mapping (`dutyFetchIntents`) that works 100% well when crossing epoch-boundary (while the meaning of `fetchCurrentEpoch/fetchNextEpoch` depends on what the current epoch is - hence sometimes it might lead to intents being interpreted incorrectly at epoch-boundaries) ... this is especially critical to our fetch-retries issuing retry-requests on the next slot (and, if necessary, the next slot after that, etc.) potentially crossing that epoch-boundary with outdated/incorrect fetch-intents
- on reorgs and indices-change events, we now check whether we are at the last slot of the current epoch - if we are, we don't re-fetch the duties for the current epoch anymore since the next slot-ticker tick will be on the very 1st slot of the next epoch (so, those duties will become irrelevant anyway)
- on indices-change events, Previously: we will not immediately re-fetch the duties (for the current/next epochs) so we can execute the correct(updated) duties on the very next slot; Now: we'll re-fetch the relevant duties immediately to have them ready for execution on the very next slot (the next slot-ticker tick)
- on indices-change events, these events are timed to be processed early into the slot (or wait until the next slot, and be processed early in the next slot) such that the duty re-fetching/processing they might result into does not "delay" the next tick of the slot-ticker; This is mostly relevant to accommodate the previous step described above, but also to get rid of the assumption that "the processing of indices-change events is very fast (so it doesn't interfere with the slot-ticker ticks just because it happens to be very fast)"
- additionally, to keep things clear and extendable in the future, this PR defines 2 separate deadlines (for slot-ticker ticks, and for duty executions) instead of using a single deadline for both - which is what we had previously (and that basically meant we had to use the `max` of these 2 deadlines ... which is suboptimal and confusing)

If this approach is looking good, we should apply it also to `SyncCommitteeHandler` (and to `ProposerHandler` - see also a relevant PR with some prior work towards that: https://github.com/ssvlabs/ssv/pull/2699)

Resolves https://github.com/ssvlabs/ssv/issues/2705 for `AttesterHandler`.


operator/duties/attester.go
```diff
@@ -84,127 +84,111 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
+			// 1. Process the tick.
+
+			currentSlot := h.ticker.Slot()
+			next = h.ticker.Next() // advances h.ticker
+			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(currentSlot)
+
+			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, currentSlot, currentSlot%32+1)
+			logger := h.logger.With(zap.String("epoch_slot_pos", buildStr))
+
+			logger.Debug("🛠 ticker event", zap.String("epoch_slot_pos", buildStr))
```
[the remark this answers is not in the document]
iurii-ssv: fixed now

operator/duties/attester.go
```diff
@@ -84,127 +84,111 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
-				tickCtx, cancel := h.ctxWithDeadlineInOneEpoch(ctx, slot)
+				tickCtx, cancel := h.ctxWithDeadlineInOneEpoch(ctx, currentSlot)
 				defer cancel()
 
-				h.executeAggregatorDuties(tickCtx, currentEpoch, slot)
-				h.processFetching(tickCtx, currentEpoch, slot)
+				h.executeAggregatorDuties(tickCtx, currentEpoch, currentSlot)
+
+				h.prepareNextEpoch(ctx, currentEpoch, currentSlot)
 			}()
```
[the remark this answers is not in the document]
iurii-ssv: This has been addressed now.

operator/duties/attester.go
```diff
@@ -84,127 +84,111 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
-		}
-		h.fetchNextEpoch = false
-	}
+	h.dutyFetchIntents[currentEpoch] = struct{}{}
+	h.dutyFetchIntents[currentEpoch+1] = struct{}{}
 
-	span.SetStatus(codes.Ok, "")
+	h.prepareCurrentEpoch(ctx, currentEpoch, currentSlot)
+	h.prepareNextEpoch(ctx, currentEpoch, currentSlot)
 }
```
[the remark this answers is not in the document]
iurii-ssv: Restored now

operator/duties/attester.go
```diff
@@ -84,127 +84,111 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
+				// (the current epoch will have been passed upon the next slot-tick). Otherwise, pre-fetch & prepare
+				// the duties for the current epoch.
+				if h.lastTickedSlotAtEpochBoundary() {
+					h.prepareNextEpoch(ctx, currentEpoch, currentSlot)
+				} else {
+					h.prepareCurrentEpoch(ctx, currentEpoch, currentSlot)
+				}
+			case <-time.After(time.Until(indicesChangeDeadline)):
+				// It's too late(risky) to handle indices change on the current slot, we'll do it on the next slot.
+			}
```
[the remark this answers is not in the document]
iurii-ssv: Makes sense, added the `case <-ctx.Done()` clause now.

iurii-ssv:
@greptile could you re-review this PR ?

operator/duties/attester.go
```diff
@@ -417,7 +494,13 @@ func toBeaconCommitteeSubscription(duty *eth2apiv1.AttesterDuty, role spectypes.
...
+	slotsPerEpoch := h.beaconConfig.SlotsPerEpoch
+	return uint64(currentSlot)%slotsPerEpoch > slotsPerEpoch/2-2
+}
+
+func (h *AttesterHandler) atLastSlotOfCurrentEpoch() bool {
+	currentSlot := h.beaconConfig.EstimatedCurrentSlot()
 	slotsPerEpoch := h.beaconConfig.SlotsPerEpoch
-	return uint64(slot)%slotsPerEpoch > slotsPerEpoch/2-2
+	return uint64(currentSlot+1)%slotsPerEpoch == 0
 }
```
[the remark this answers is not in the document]
iurii-ssv: True, but I don't see this as an issue because the code that relies on `atLastSlotOfCurrentEpoch` does handle this "inconsistency" fine regardless.

operator/duties/attester.go
```diff
@@ -243,41 +281,81 @@ func (h *AttesterHandler) executeAggregatorDuties(ctx context.Context, epoch pha
...
+
+		err := h.fetchAndProcessDuties(ctx, logger, currentEpoch, currentSlot)
+		if err != nil {
+			h.logger.Error("failed to prepare duties for current epoch", zap.Error(err))
+			return
+		}
+		h.dutyFetchIntents[currentEpoch] = true // the intent has been fulfilled
+
+		logger.Info("fetching duties for the current epoch succeeded")
+	}
```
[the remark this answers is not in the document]
iurii-ssv: good find, fixed in https://github.com/ssvlabs/ssv/pull/2712/changes/95bd1fc57ccfadcc5f33de913a000197124f5be5

operator/duties/attester.go
```diff
@@ -84,127 +82,167 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
-		case reorgEvent := <-h.reorg:
-			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(reorgEvent.Slot)
-			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, reorgEvent.Slot, reorgEvent.Slot%32+1)
-			h.logger.Info("🔀 reorg event received", zap.String("epoch_slot_pos", buildStr), zap.Any("event", reorgEvent))
+				slotsPerEpoch := h.beaconConfig.SlotsPerEpoch
 
-			func() {
-				tickCtx, cancel := h.ctxWithDeadlineInOneEpoch(ctx, reorgEvent.Slot)
-				defer cancel()
+				// Clean up the irrelevant data to prevent infinite memory growth at the 1 slot of the epoch.
```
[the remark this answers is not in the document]
iurii-ssv: it should actually be the very 1st slot of the epoch, fixed in https://github.com/ssvlabs/ssv/pull/2712/changes/95bd1fc57ccfadcc5f33de913a000197124f5be5

operator/duties/proposer.go
```diff
@@ -75,27 +75,22 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-			func() {
-				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot)
-				defer cancel()
-
-				if h.fetchFirst {
-					h.fetchFirst = false
+			if h.fetchFirst {
+				h.fetchFirst = false
+				h.indicesChanged = false
+				h.processFetching(ctx, currentEpoch)
```
olegshmuelov: This `ctx` is the long-lived scheduler context (no deadline). Previously this call was inside `ctxWithDeadlineOnNextSlot` which bounded it to ~12s.
Now if `beaconNode.ProposerDuties()` stalls, this handler's event loop blocks indefinitely - no future ticks, reorgs, or indices changes get processed.
The attester handler correctly creates a per-tick `tickCtx` for this. Consider wrapping in a similar deadline here.

operator/duties/sync_committee.go
```diff
@@ -99,13 +99,8 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
...
 
-			func() {
-				tickCtx, cancel := h.ctxWithDeadlineOnNextSlot(ctx, slot)
-				defer cancel()
-
-				h.processExecution(tickCtx, period, slot)
-				h.processFetching(tickCtx, epoch, period, true)
-			}()
+			h.processExecution(ctx, period, slot)
+			h.processFetching(ctx, epoch, period, true)
```
olegshmuelov: `processFetching` now runs with the long-lived scheduler context (no deadline). Previously this was inside `ctxWithDeadlineOnNextSlot`.
If `beaconNode.SyncCommitteeDuties()` stalls, this handler's event loop blocks indefinitely.
Additionally, `fetchAndProcessDuties` spawns a `SubmitSyncCommitteeSubscriptions` goroutine via `CtxWithParentDeadline(ctx)` - since `ctx` has no deadline, that goroutine inherits none, which can block `WaitShutdown()`.

operator/duties/scheduler.go
```diff
@@ -473,13 +474,8 @@ func (s *Scheduler) ExecuteDuties(ctx context.Context, duties []*spectypes.Valid
 		go func() {
 			defer s.backgroundTasks.Done()
 
-			// Cannot use parent-context itself here, have to create independent instance
-			// to be able to continue working in background.
-			dutyCtx, cancel, withDeadline := utils.CtxWithParentDeadline(ctx)
+			dutyCtx, cancel := context.WithDeadline(ctx, dutyDeadline)
```
olegshmuelov: The attester passes `tickCtx` (deadline ~12s) to `ExecuteDuties`, but `dutyDeadline` is ~6.4 min. Since `context.WithDeadline(tickCtx, dutyDeadline)` takes the minimum, the effective deadline is 12s, making the `dutyDeadline` parameter dead code for the attester. Also, `defer cancel()` on the tick closure cascades to all spawned duty contexts.

The old code used `CtxWithParentDeadline` -> `context.WithoutCancel` to detach duties from parent cancellation. Only the attester is affected (other handlers pass the scheduler context which has no deadline).

Is it intentional that aggregation duties are bounded to ~12s? If so, the 6.4 min deadline and its spec comment are misleading. If not:
```go
dutyCtx, cancel := context.WithDeadline(context.WithoutCancel(ctx), dutyDeadline)
```

Same pattern in `ExecuteCommitteeDuties` at line 527 (not currently affected).

operator/duties/attester.go
```diff
@@ -251,41 +284,81 @@ func (h *AttesterHandler) executeAggregatorDuties(ctx context.Context, epoch pha
...
+	// We set the deadline to target slot + SLOTS_PER_EPOCH + 1 (since the deadline slot itself is excluded).
+	slotsPerEpoch := phase0.Slot(h.beaconConfig.SlotsPerEpoch)
+	dutyDeadline := h.beaconConfig.SlotStartTime(slot + slotsPerEpoch + 1)
+	h.dutiesExecutor.ExecuteDuties(ctx, toExecute, dutyDeadline)
 
 	span.SetStatus(codes.Ok, "")
 }
 
-func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase0.Epoch, slot phase0.Slot) error {
+func (h *AttesterHandler) prepareCurrentEpoch(ctx context.Context, logger *zap.Logger, currentEpoch phase0.Epoch, currentSlot phase0.Slot) {
```
olegshmuelov, line 299: Nit: `fetchCurrentEpochDuties` would be more descriptive - "prepare" doesn't convey what the function does (fetch + store duties from the beacon node). Same for `prepareNextEpoch` below.

operator/duties/attester.go
```diff
@@ -428,7 +500,13 @@ func toBeaconCommitteeSubscription(duty *eth2apiv1.AttesterDuty, role spectypes.
 	}
 }
 
-func (h *AttesterHandler) shouldFetchNexEpoch(slot phase0.Slot) bool {
+func (h *AttesterHandler) goodTimeToFetchDutiesForNextEpoch(currentSlot phase0.Slot) bool {
```
olegshmuelov: Nit: `shouldFetchNextEpoch` - 3 words instead of 7, and the `should` prefix is idiomatic Go for timing predicates.

operator/duties/attester.go
```diff
@@ -428,7 +500,13 @@ func toBeaconCommitteeSubscription(duty *eth2apiv1.AttesterDuty, role spectypes.
...
+	slotsPerEpoch := h.beaconConfig.SlotsPerEpoch
+	return uint64(currentSlot)%slotsPerEpoch > slotsPerEpoch/2-2
+}
+
+func (h *AttesterHandler) atLastSlotOfCurrentEpoch() bool {
```
olegshmuelov: This re-derives the current slot from the wall clock, but every caller already has `currentSlot`. Consider `lastSlotOfEpoch(slot phase0.Slot)` - it becomes a pure function, trivially testable, deterministic, and consistent with `goodTimeToFetchDutiesForNextEpoch` which already accepts `currentSlot`.

operator/duties/attester.go
```diff
@@ -92,127 +84,168 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
...
-			next = h.ticker.Next()
-			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(slot)
-			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, slot, slot%32+1)
-			h.logger.Debug("🛠 ticker event", zap.String("epoch_slot_pos", buildStr))
+			currentSlot := h.ticker.Slot()
+			next = h.ticker.Next() // advances h.ticker
+			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(currentSlot)
+			nextEpoch := currentEpoch + 1
+
+			buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, currentSlot, currentSlot%32+1)
```
olegshmuelov: pre-existing:
`%32` hardcodes mainnet's `SlotsPerEpoch`. Should be `currentSlot%phase0.Slot(h.beaconConfig.SlotsPerEpoch)+1`.

operator/duties/attester.go
```diff
@@ -251,41 +284,81 @@ func (h *AttesterHandler) executeAggregatorDuties(ctx context.Context, epoch pha
...
+	dutyDeadline := h.beaconConfig.SlotStartTime(slot + slotsPerEpoch + 1)
+	h.dutiesExecutor.ExecuteDuties(ctx, toExecute, dutyDeadline)
 
 	span.SetStatus(codes.Ok, "")
 }
 
-func (h *AttesterHandler) fetchAndProcessDuties(ctx context.Context, epoch phase0.Epoch, slot phase0.Slot) error {
+func (h *AttesterHandler) prepareCurrentEpoch(ctx context.Context, logger *zap.Logger, currentEpoch phase0.Epoch, currentSlot phase0.Slot) {
+	if fulfilled, ok := h.dutyFetchIntents[currentEpoch]; ok && !fulfilled {
+		logger.Info("fetching duties for the current epoch")
```
olegshmuelov: Nit: Both "fetching duties..." (here) and "...succeeded" (line 310) at Info level is verbose for something that runs every few slots. Consider keeping only the success log or demoting one to Debug.

operator/duties/proposer.go
iurii-ssv: Oh yeah, I meant to "rework" `ProposerHandler` in a similar manner in a follow-up PR, but you are right - it's better to keep using `tickCtx` here for now to not have any regressions ... applied in https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

operator/duties/sync_committee.go
iurii-ssv: > `processFetching` now runs with the long-lived scheduler context (no deadline). Previously this was inside `ctxWithDeadlineOnNextSlot`.
> If `beaconNode.SyncCommitteeDuties()` stalls, this handler's event loop blocks indefinitely.

similar to the comment from above, using `tickCtx` again now https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

> Additionally, `fetchAndProcessDuties` spawns a `SubmitSyncCommitteeSubscriptions` goroutine via `CtxWithParentDeadline(ctx)` - since `ctx` has no deadline, that goroutine inherits none, which can block `WaitShutdown()`.

so now we'll be basically using the deadline on `tickCtx` (like we were previously), I think it should be sufficient in practice

operator/duties/scheduler.go
iurii-ssv: Oh yeah, good catch, I meant to create a brand-new context here like: `dutyCtx, cancel := context.WithDeadline(context.Background(), dutyDeadline)`,

but since we also want to be able to terminate those spawned go-routines upon SSV node shutdown - I'm storing the context `Scheduler` was started with in `s.ctx` and use that now https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

operator/duties/attester.go
iurii-ssv: The thing is: `prepareCurrentEpoch` does both "fetch" and "process" (it calls the func called `fetchAndProcessDuties` which does other things besides the fetching) - so in that sense I would rather leave a more "vague" name (such as `prepareCurrentEpoch`) that kinda "forces you to go look at what's really going on" rather than mislead the reader into thinking that "fetch is all that happens there"

operator/duties/attester.go
iurii-ssv: I wanted to emphasize that this func specifically looks at "whether it's a good time" (rather than checking for something else) ... but I guess we can add a describing comment for that instead, applied in https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

operator/duties/attester.go
iurii-ssv: Right, I kinda left it like that to "emphasize" that wall-clock value is all we really need here (no need to pass the exact slot), but I guess the consistency with `goodTimeToFetchDutiesForNextEpoch` func is indeed nice to have, applied in https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61 (also moved both of these funcs to `baseHandler` as different handler-implementations will want to re-use them)

operator/duties/attester.go
iurii-ssv: Addressed in https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

operator/duties/attester.go
iurii-ssv: Demoted to `Debug` in https://github.com/ssvlabs/ssv/pull/2712/changes/d599937523fb4ab63c3b3d87445c8c4df3a8aa61

although it is a bit redundant, I think it's useful to preserve both lines just because this would allow to grep by `fetching duties for` and see all the outcomes (+ allows us to quickly compare the timestamps of those log lines to get an idea of how long it took),

overall the total amount of these lines written is quite moderate (it's ~ several per epoch):

image

operator/duties/sync_committee.go
iurii-ssv: > > Additionally, `fetchAndProcessDuties` spawns a `SubmitSyncCommitteeSubscriptions` goroutine via `CtxWithParentDeadline(ctx)` - since `ctx` has no deadline, that goroutine inherits none, which can block `WaitShutdown()`.
> 
> so now we'll be basically using the deadline on `tickCtx` (like we were previously), I think it should be sufficient in practice

Come to think of it, using the deadline of `tickCtx` isn't really appropriate here (the fact that we spin off a separate go-routine kinda hints at that too) ... we should really create a dedicated context for that background task:
- it will inherit from `Handler.ctx` to terminate upon SSV node shutdown
- we could define a dedicated deadline for it, but I don't really see a reason to do it given that beacon-client force-sets request timeouts anyway

applied these ^ in https://github.com/ssvlabs/ssv/pull/2712/changes/f37813a3b475e66c3442f4193a1355b8dc82e257

Also, we could implement retries for these background procedures (in case Beacon requests fail), do you think we should @olegshmuelov ? Unlike for duty-fetches that are retried on subsequent slot-ticks, these background tasks (`SubmitBeaconCommitteeSubscriptions`, `SubmitSyncCommitteeSubscriptions` calls) aren't ever repeated, seems like

olegshmuelov: Good call on `h.ctx` for the background tasks.

Re retries: subscriptions are re-submitted on each duty-fetch cycle (indices change, reorg, next epoch), so transient failures get implicitly retried. For persistent failures (beacon node down), explicit retries wouldn't help either. I'd skip retry logic here unless production data shows subscription failures correlating with missed aggregations.

Minor: `shouldFetchNextEpoch` and `atLastSlotOfCurrentEpoch` were moved to `base_handler.go` but still have `*AttesterHandler` receiver - other handlers can't reuse them. If reuse is the intent, receiver should be `*baseHandler`.

iurii-ssv: > Re retries: subscriptions are re-submitted on each duty-fetch cycle (indices change, reorg, next epoch), so transient failures get implicitly retried. For persistent failures (beacon node down), explicit retries wouldn't help either. I'd skip retry logic here unless production data shows subscription failures correlating with missed aggregations.

Sounds good.

> Minor: `shouldFetchNextEpoch` and `atLastSlotOfCurrentEpoch` were moved to `base_handler.go` but still have `*AttesterHandler` receiver - other handlers can't reuse them. If reuse is the intent, receiver should be `*baseHandler`.

Oh, kinda thought it's on the compiler to catch that :) Fixed now in https://github.com/ssvlabs/ssv/pull/2712/changes/65526503f88446fe012cc42fde8ae8ca5f7dc293

olegshmuelov (approved):
LGTM!

ljuba-ssv (approved):

PR #2714, opened 2026-03-10
duties: temporary work-around for pool-related panic

iurii-ssv:
The `pool.Pool` panics when tasks are being sent to it while/after it is in the shut-down process, so we need to be a bit more careful with the way we are using it (the reason to use it is to have a graceful shutdown).


iurii-ssv:
Closing since we decided to revert the offending PR for now instead - https://github.com/ssvlabs/ssv/pull/2715

PR #2719, opened 2026-03-12
eth/contract: update contract ABI to v2.0.0

nkryuchkov:
Sometimes we have `failed to find event by ID` logs due to the old ABI not recognizing new events.

~~As of now, the latest available contract version is [v1.2.0 for mainnet](https://github.com/ssvlabs/ssv-network/tree/contract-abi/docs/mainnet) and [v2.0.0 for testnet](https://github.com/ssvlabs/ssv-network/tree/contract-abi/docs/testnet), so perhaps we need to put this PR on hold until v2.0.0 is released for mainnet~~

The PR updates the contract version to the latest one, v2.0.0: https://github.com/ssvlabs/ssv-network/blob/contract-abi/docs/mainnet/v2.0.0/abi/SSVNetwork.json

Closes ssvlabs/ssv-node-board#1054


olegshmuelov (changes requested):
v2.0.0 ABI is already out, let's retarget this PR to v2.0.0 directly.

iurii-ssv (dismissed):

olegshmuelov (dismissed):
LGTM.

One follow-up worth tracking: the 26+ new events recognized by v2.0.0 will each produce a `Warn("unknown event name")` log line. Some of these (e.g. `OperatorMultipleWhitelistUpdated`, `ModuleUpgraded`) fire frequently on-chain and could get noisy. Worth a follow-up to add explicit no-op cases for the common ones to reduce log clutter without adding behavior.

ljuba-ssv (dismissed):
LGTM. One minor nit:

`eth/contract/generate.go:1-4` — the package comment still says "v1 contract" and the source URL still points to `v1.0.2/abi/SSVNetwork.json`. These should be updated to reflect the v2.0.0 ABI.

mtabasco (approved):

nkryuchkov:
> LGTM. One minor nit:
> 
> `eth/contract/generate.go:1-4` — the package comment still says "v1 contract" and the source URL still points to `v1.0.2/abi/SSVNetwork.json`. These should be updated to reflect the v2.0.0 ABI.

The source URL uses v2 but the package comment is indeed confusing. Fixed

ljuba-ssv (approved):

iurii-ssv (approved):

PR #2722, opened 2026-03-16
feat(el): bloom filter cross-check for log integrity

y0sher:
## Summary
- Adds bloom filter cross-check in `fetchLogsInBatches` to detect and recover events silently dropped by execution clients (known Geth v1.17.1 bug)
- For each block with no logs, fetches the block header and checks if the bloom filter matches the SSV contract address — if it does, retries `FilterLogs` for that single block (up to 3x, 2s delay)
- All blocks in a chunk must pass verification before any `BlockLogs` are emitted downstream, preserving atomicity
- **Scoped to streaming path only** — bloom verification runs near chain tip where the bug is most impactful; skipped during historical sync to avoid ~200 extra header RPCs per batch
- **Single consolidated metric** `ssv.el.bloom.checks` with outcome attribute (`recovery` / `false_positive`)

## Context
Geth intermittently returns `[]` from `eth_getLogs` for blocks that contain SSV contract events (confirmed: block 24625387 had 80 events via Nethermind but 0 via Geth). The SSV node silently accepts this as "no events" and permanently skips the block's data. Bloom filters in block headers are consensus-validated with **no false negatives**, making them a reliable cross-check.

## Changes
- **`bloom.go`** — `verifyLogsWithBloom()` and `retryBlockLogs()` methods on `ExecutionClient`
- **`execution_client.go`** — `fetchLogsInBatches` takes a `verifyBloom bool` parameter; `FetchHistoricalLogs` passes `false`, `streamLogsToChan` passes `true`
- **`defaults.go`** — `DefaultBloomRetryAttempts = 3`, `DefaultBloomRetryDelay = 2s`
- **`observability.go`** — `ssv.el.bloom.checks` counter with `ssv.el.bloom.outcome` attribute
- **`execution_client_test.go`** — updated `fetchLogsInBatches` calls with new parameter

## Test plan
- [x] `go build ./eth/executionclient/...` compiles
- [x] `go test ./eth/executionclient/...` passes (all existing + 5 new tests)
- [x] `make lint` passes
- [x] Verify on stage with simulated 0-log response that bloom check triggers recovery
- [x] Monitor `ssv.el.bloom.checks` counter in Grafana after deploy

Closes ssvlabs/ssv-node-board#1046

### DIFF
--- a/eth/executionclient/bloom.go
+++ b/eth/executionclient/bloom.go
@@ -18,6 +18,10 @@ import (
 // verifyLogsWithBloom cross-checks the logs returned by FilterLogs against each block's
 // bloom filter. Blocks that returned no logs but whose bloom filter matches the contract
 // address are retried individually to recover potentially dropped events (a known Geth bug).
+//
+// This is a block-level check only — it assumes all-or-nothing drops (an entire block's
+// logs are missing). Partial drops (some logs present, some missing within the same block)
+// have not been observed and are not handled here.
 func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethtypes.Log, fromBlock, toBlock uint64) ([]ethtypes.Log, error) {
 	// Build a set of blocks that already have logs in the result.
 	blocksWithLogs := make(map[uint64]bool)
--- a/eth/executionclient/bloom.go
+++ b/eth/executionclient/bloom.go
@@ -2,7 +2,9 @@ package executionclient
 
 import (
 	"context"
+	"fmt"
 	"math/big"
+	"sort"
 	"time"
 
 	"github.com/ethereum/go-ethereum"
@@ -23,6 +25,7 @@ func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethty
 		blocksWithLogs[logs[i].BlockNumber] = true
 	}
 
+	recovered := false
 	for blockNum := fromBlock; blockNum <= toBlock; blockNum++ {
 		if blocksWithLogs[blockNum] {
 			continue // block already has logs, nothing to verify
@@ -30,7 +33,7 @@ func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethty
 
 		header, err := ec.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNum))
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("bloom check: fetch header for block %d: %w", blockNum, err)
 		}
 
 		if !header.Bloom.Test(ec.contractAddress.Bytes()) {
@@ -38,27 +41,38 @@ func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethty
 		}
 
 		// Bloom matched but FilterLogs returned nothing — suspected EL bug.
-		ec.logger.Warn("bloom filter mismatch: block bloom matches contract but no logs returned, retrying",
+		ec.logger.Warn("bloom filter mismatch during block-logs fetching: block bloom matches contract but EL returned no logs for this block, gonna retry to fetch block-logs for this specific block",
 			fields.BlockNumber(blockNum),
 			zap.String("contract", ec.contractAddress.Hex()),
 		)
 
-		recovered, err := ec.retryBlockLogs(ctx, blockNum)
+		blockLogs, err := ec.retryBlockLogs(ctx, blockNum)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("bloom check: retry block %d logs: %w", blockNum, err)
 		}
 
-		if len(recovered) > 0 {
+		if len(blockLogs) > 0 {
 			ec.logger.Warn("bloom cross-check recovered missing events",
 				fields.BlockNumber(blockNum),
-				zap.Int("recovered_events", len(recovered)),
+				zap.Int("recovered_events", len(blockLogs)),
 			)
 			recordBloomRecovery(ctx)
-			logs = append(logs, recovered...)
+			logs = append(logs, blockLogs...)
+			recovered = true
 		} else {
 			// Bloom false positive — address was in bloom but no actual logs for our contract.
 			recordBloomFalsePositive(ctx)
 		}
+	}
+
+	// Re-sort if we appended recovered logs so downstream receives them in block/tx order.
+	if recovered {
+		sort.Slice(logs, func(i, j int) bool {
+			if logs[i].BlockNumber != logs[j].BlockNumber {
+				return logs[i].BlockNumber < logs[j].BlockNumber
+			}
+			return logs[i].TxIndex < logs[j].TxIndex
+		})
 	}
 
 	return logs, nil
@@ -93,7 +107,7 @@ func (ec *ExecutionClient) retryBlockLogs(ctx context.Context, blockNumber uint6
 				zap.Error(err),
 			)
 			if attempt == DefaultBloomRetryAttempts {
-				return nil, err
+				return nil, fmt.Errorf("bloom retry: filter logs for block %d: %w", blockNumber, err)
 			}
 			continue
 		}
--- a/eth/executionclient/bloom.go
+++ b/eth/executionclient/bloom.go
@@ -1,0 +1,108 @@
+package executionclient
+
+import (
+	"context"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/observability/log/fields"
+)
+
+// verifyLogsWithBloom cross-checks the logs returned by FilterLogs against each block's
+// bloom filter. Blocks that returned no logs but whose bloom filter matches the contract
+// address are retried individually to recover potentially dropped events (a known Geth bug).
+func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethtypes.Log, fromBlock, toBlock uint64) ([]ethtypes.Log, error) {
+	// Build a set of blocks that already have logs in the result.
+	blocksWithLogs := make(map[uint64]bool)
+	for i := range logs {
+		blocksWithLogs[logs[i].BlockNumber] = true
+	}
+
+	for blockNum := fromBlock; blockNum <= toBlock; blockNum++ {
+		if blocksWithLogs[blockNum] {
+			continue // block already has logs, nothing to verify
+		}
+
+		header, err := ec.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNum))
+		if err != nil {
+			return nil, err
+		}
+
+		if !header.Bloom.Test(ec.contractAddress.Bytes()) {
+			continue // bloom says no SSV events in this block — genuine empty
+		}
+
+		// Bloom matched but FilterLogs returned nothing — suspected EL bug.
+		ec.logger.Warn("bloom filter mismatch: block bloom matches contract but no logs returned, retrying",
+			fields.BlockNumber(blockNum),
+			zap.String("contract", ec.contractAddress.Hex()),
+		)
+
+		recovered, err := ec.retryBlockLogs(ctx, blockNum)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(recovered) > 0 {
+			ec.logger.Warn("bloom cross-check recovered missing events",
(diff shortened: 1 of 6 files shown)


eth/executionclient/bloom.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		header, err := ec.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNum))
+		if err != nil {
+			return nil, err
+		}
+
+		if !header.Bloom.Test(ec.contractAddress.Bytes()) {
+			continue // bloom says no SSV events in this block — genuine empty
+		}
```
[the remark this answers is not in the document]
y0sher: Addressed — bloom verification is now scoped to the streaming path only (`verifyBloom` flag on `fetchLogsInBatches`). Historical sync passes `false`, streaming passes `true`. Zero extra RPCs during initial sync.

eth/executionclient/observability.go
```diff
@@ -97,6 +97,16 @@ var (
...
+
+	bloomRecoveryCounter = metrics.New(
+		meter.Int64Counter(
+			observability.InstrumentName(observabilityNamespace, "bloom.recoveries"),
+			metric.WithDescription("number of times bloom cross-check recovered missing events from the execution client")))
+
+	bloomFalsePositiveCounter = metrics.New(
+		meter.Int64Counter(
+			observability.InstrumentName(observabilityNamespace, "bloom.false_positives"),
+			metric.WithDescription("number of bloom filter false positives where bloom matched but no contract events existed")))
```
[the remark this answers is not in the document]
y0sher: Addressed — consolidated into a single `ssv.el.bloom.checks` counter with an `ssv.el.bloom.outcome` attribute (`"recovery"` / `"false_positive"`).

eth/executionclient/bloom.go
```diff
@@ -0,0 +1,108 @@
...
+
+	for blockNum := fromBlock; blockNum <= toBlock; blockNum++ {
+		if blocksWithLogs[blockNum] {
+			continue // block already has logs, nothing to verify
+		}
+
+		header, err := ec.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNum))
+		if err != nil {
+			return nil, err
+		}
```
iurii-ssv, line 41: Probably want to wrap this error (and similar below) with the additional context, something like: `return nil, fmt.Erorrf("fetch header by block number: %w", err)`

y0sher:
> ### Greptile Summary
> 
> Adds a bloom filter cross-check in `fetchLogsInBatches` to detect and recover events silently dropped by execution clients (targeting a known Geth v1.17.1 bug). For each block with no returned logs, the block header's bloom filter is checked against the SSV contract address — if it matches, `FilterLogs` is retried for that single block up to 3 times. This is a sound defensive measure since Ethereum bloom filters are consensus-validated with no false negatives.
> 
>     * The integration point in `fetchLogsInBatches` is clean — verification happens after `subdivideLogFetch` but before any emission, preserving batch atomicity
> 
>     * Retry logic in `retryBlockLogs` properly respects context cancellation and client closure via `select`
> 
>     * The `Bloom.Test(address.Bytes())` call is correct — go-ethereum's bloom API hashes the input internally using Keccak-256, consistent with how addresses are added to the bloom during block creation
> 
>     * Both historical sync (`FetchHistoricalLogs`) and live streaming (`StreamLogs` → `streamLogsToChan`) paths are covered since both use `fetchLogsInBatches`
> 
>     * `PackLogs` downstream sorts by block number, so recovered logs appended out-of-order will be correctly reordered
> 
>     * **Performance note**: fetching a block header for every empty block in each 200-block batch may add significant RPC overhead during long historical syncs; consider gating this to near-tip blocks only
> 
>     * Test coverage is thorough with 5 new tests simulating the Geth bug via HTTP proxy interception
> 
> 
> ### Confidence Score: 4/5
> 
>     * This PR is safe to merge — the bloom cross-check is a defensive addition with correct logic and good test coverage, with only a performance concern during historical sync.
> 
>     * The implementation is logically sound: Bloom.Test() usage is correct, retry logic handles cancellation properly, error propagation preserves atomicity, and the feature is well-tested. The only concern is the potential performance overhead of per-block header fetches during long historical syncs, which is a non-blocking style issue.
> 
>     * Pay attention to `eth/executionclient/bloom.go` for potential sync performance impact from per-block header fetches.
> 
> 
> ### Important Files Changed
> Filename 	Overview
> eth/executionclient/bloom.go 	New bloom filter cross-check logic: verifyLogsWithBloom iterates empty blocks, fetches headers, and retries FilterLogs on bloom mismatch. Correct bloom API usage, proper context/cancellation handling, but header-per-block fetching may be costly during historical sync.
> eth/executionclient/bloom_test.go 	Comprehensive test suite with 5 tests covering: recovery of dropped logs, no false recovery, empty ranges, exhausted retries, and header fetch optimization for blocks with existing logs. Tests use HTTP proxy interception to simulate the Geth bug.
> eth/executionclient/defaults.go 	Adds DefaultBloomRetryAttempts (3) and DefaultBloomRetryDelay (2s) constants, consistent with existing pattern of extracting config values into named constants.
> eth/executionclient/execution_client.go 	7-line insertion in fetchLogsInBatches to call verifyLogsWithBloom after subdivideLogFetch and before emission. Clean integration that preserves atomicity — errors abort the batch.
> eth/executionclient/observability.go 	Adds two new counters (bloom.recoveries and bloom.false_positives) with proper namespacing. Could be consolidated into a single metric with an outcome attribute per project conventions.
> ### Flowchart
> 
> ```mermaid
> %%{init: {'theme': 'neutral'}}%%
> flowchart TD
>     A[fetchLogsInBatches: batch of 200 blocks] --> B[subdivideLogFetch: get logs from EL]
>     B --> C[verifyLogsWithBloom]
>     C --> D{For each block in range}
>     D -->|Block has logs| E[Skip - already verified]
>     D -->|Block has no logs| F[Fetch block header via HeaderByNumber]
>     F --> G{Bloom.Test contract address}
>     G -->|No match| H[Genuine empty - skip]
>     G -->|Match!| I[Suspected EL bug - retry]
>     I --> J[retryBlockLogs: up to 3 attempts, 2s delay]
>     J -->|Logs recovered| K[Append to results + record recovery metric]
>     J -->|No logs after retries| L[Bloom false positive - record metric]
>     J -->|RPC error after retries| M[Return error - abort batch]
>     E --> D
>     H --> D
>     K --> D
>     L --> D
>     D -->|All blocks checked| N[Return verified logs]
>     N --> O[PackLogs + emit to logCh]
> ```
> 
> Last reviewed commit: [3072ec4](https://github.com/ssvlabs/ssv/commit/3072ec4bea99009f46a7bf8b98b3cb7eea3fe54b)

can you please review again

eth/executionclient/bloom.go
```diff
@@ -0,0 +1,108 @@
...
+
+		if len(recovered) > 0 {
+			ec.logger.Warn("bloom cross-check recovered missing events",
+				fields.BlockNumber(blockNum),
+				zap.Int("recovered_events", len(recovered)),
+			)
+			recordBloomRecovery(ctx)
+			logs = append(logs, recovered...)
```
iurii-ssv: Shouldn't we reconstruct the resulting slice `verifyLogsWithBloom` will return from scratch - to ensure the sorting order is preserved ? Seems like "by just appending to the end of `logs` slice we won't get the fully sorted outcome" (and seems like we must process events strictly sequentially, in the order they happen from block to block)

eth/executionclient/bloom.go
```diff
@@ -0,0 +1,108 @@
...
+
+		if !header.Bloom.Test(ec.contractAddress.Bytes()) {
+			continue // bloom says no SSV events in this block — genuine empty
+		}
+
+		// Bloom matched but FilterLogs returned nothing — suspected EL bug.
+		ec.logger.Warn("bloom filter mismatch: block bloom matches contract but no logs returned, retrying",
+			fields.BlockNumber(blockNum),
+			zap.String("contract", ec.contractAddress.Hex()),
+		)
```
iurii-ssv, line 51: Perhaps to make this log-line a bit more detailed:
```
ec.logger.Warn("bloom filter mismatch during block-logs fetching: block bloom matches contract but EL returned no logs for this block, gonna retry to fetch block-logs for this specific block",
	fields.BlockNumber(blockNum),
	zap.String("contract", ec.contractAddress.Hex()),
)
```

iurii-ssv (commented):
Looks good, check out [the question/comment about sorting](https://github.com/ssvlabs/ssv/pull/2722#discussion_r2940228912)

eth/executionclient/bloom.go
y0sher: Done — both error returns now wrapped with context (`bloom check: fetch header for block %d` and `bloom check: retry block %d logs`). Also wrapped the final retry error in `retryBlockLogs`.

eth/executionclient/bloom.go
y0sher: Good catch — fixed. The slice is now re-sorted by block number + tx index after appending recovered logs, so the output is always in correct order regardless of where the recovery happened. (Note: `PackLogs` downstream also sorts, but better to return a correct result from this function too.)

eth/executionclient/bloom.go
y0sher: Updated to your suggested wording.

iurii-ssv (dismissed):
Nice!

eth/executionclient/bloom_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+	// The proxy should have been called more than once for eth_getLogs:
+	// 1 initial wide query + at least 1 single-block retry.
+	require.Greater(t, filterCallCount.Load(), int32(1), "expected at least one retry for bloom mismatch")
+}
+
+// TestBloomCrossCheck_NoFalseRecovery verifies that when no logs are dropped
+// (all blocks genuinely have no events or already have logs), the bloom check
+// does not inject spurious events.
+func TestBloomCrossCheck_NoFalseRecovery(t *testing.T) {
```
olegshmuelov, line 128: `TestBloomCrossCheck_NoFalseRecovery` calls `FetchHistoricalLogs` (L146) which hardcodes `verifyBloom=false`. The test doesn't exercise bloom verification - it's testing the historical (non-bloom) path. Consider using `fetchLogsInBatches` directly with `verifyBloom=true` (like the other bloom tests do) to actually verify that bloom checks don't inject false recoveries.

eth/executionclient/bloom.go
```diff
@@ -0,0 +1,122 @@
...
+func (ec *ExecutionClient) verifyLogsWithBloom(ctx context.Context, logs []ethtypes.Log, fromBlock, toBlock uint64) ([]ethtypes.Log, error) {
+	// Build a set of blocks that already have logs in the result.
+	blocksWithLogs := make(map[uint64]bool)
+	for i := range logs {
+		blocksWithLogs[logs[i].BlockNumber] = true
+	}
+
+	recovered := false
+	for blockNum := fromBlock; blockNum <= toBlock; blockNum++ {
+		if blocksWithLogs[blockNum] {
```
olegshmuelov, line 34: The bloom check catches blocks with zero logs - is there any evidence the Geth bug can do partial drops (some logs returned, some missing for the same block)? If not, might be worth adding a brief comment here noting the scope: "block-level check only, assumes all-or-nothing drops."

olegshmuelov (dismissed):
LGTM, few minor comments.

eth/executionclient/bloom.go
y0sher: Thanks, I haven't observed this issue ever with partial drops, so trying to work on it form this perspective might require a whole new approach. We'll do it if we observe the issue happens this way.
will add a comment

eth/executionclient/bloom_test.go
y0sher: Fixed in c882ec8 — test now uses fetchLogsInBatches with verifyBloom=true so it actually exercises the bloom path.

eth/executionclient/bloom.go
y0sher: Added scope comment in c882ec8 — block-level check only, assumes all-or-nothing drops. Will revisit if partial drops are observed.

y0sher:
Addressed Oleg's review comments. Two things still blocking this from my end:

1. **Testing with a faulty Geth client** — we should verify this actually recovers dropped logs in a realistic scenario, not just unit tests with simulated backends. Need to reproduce the Geth bug with a real (or patched) node.
2. **Sync time impact** — bloom verification adds per-block header fetches. We need to measure the overhead on a full historical sync to make sure it's acceptable.

Once we have data on both, this is good to go.

iurii-ssv (approved):

olegshmuelov (approved):

momosh-ssv (approved):

PR #2723, opened 2026-03-17
runners: handle unexpected messages gracefully

iurii-ssv:
There are several places in the code where we "expect" a p2p message to be of a certain type (or to be non-`nil`), while this should be checked for in the validation-layer that comes prior ... we still should be defensive in with our type-casts in case there is a missing validation-check because, even though it is a bug in our code that we want to see and fix ASAP, panic-ing in these cases might lead to never-ending crash-loops or potentially open DOS attack vectors.

Additionally(prompted by further investigation of those Greptile review suggestions), `mKey` func was updated to accommodate the "events" which were previously missing from there resulting in different events potentially sharing the same retry-context (which doesn't seem critical, but we'd want to fix it either way just in case).

Resolves `F-ssv-046` (raised by Claude).


iurii-ssv:
> * `validator.go` `ProcessMessage`: the `SSVPartialSignatureMsgType` branch still uses `if !ok` without the `|| signedMsg == nil` guard that every other branch in this PR received, leaving a potential nil-pointer panic path.

Well spotted, fixed in https://github.com/ssvlabs/ssv/pull/2723/changes/2314cee0aa62fcb3430466e029966780224b69f4

> * `mKey` in `msg_queue.go`: all invalid messages of both types fall back to the same constant `"undefined"`, so distinct malformed messages share a key in `msgStates`, potentially mixing retry counters and tracing spans.

Right, it's better to update `mKey` to return an `error` explicitly for the caller to handle it, implemented in https://github.com/ssvlabs/ssv/pull/2723/changes/2314cee0aa62fcb3430466e029966780224b69f4

@greptile could you re-review ?

iurii-ssv:
> * The PR correctly addresses its stated goal across 8 of 9 changed files. The one omission — scoreMessageType — has the same nil-dereference potential (typed nil EventMsg body in a type-switch) that was fixed in all neighbouring functions in the same file. In practice this would only be triggered if Body is somehow set to a typed nil, which current decode paths do not produce, so the risk is low but real enough to warrant a follow-up before closing the issue.
> * protocol/v2/ssv/queue/messages.go — scoreMessageType was not updated with a nil guard, unlike every other scoring helper in the file.

This should be fixed now with https://github.com/ssvlabs/ssv/pull/2723/changes/5a371b99b21bbd8d01feb970e0ca17a2adf90e7d

@greptile pls re-review

y0sher:
Changes look solid. I would still want to run this for a while in stage to see message flow is not hurt since this is a critical area. please report back test results.

alok-ssv (dismissed):

nkryuchkov (dismissed):

ljuba-ssv (dismissed):
Root cause fix is correct — Fatal replaced with error return. The broader hardening of type assertions across the message pipeline is thorough. mKey refactoring to return errors and handle events properly is a genuine improvement.

Minor suggestions for follow-up:
- Add `|| eventMsg == nil` to committee.go:314 for consistency
- Add `|| e == nil` to queue_validator.go:130 filter

protocol/v2/ssv/validator/msg_queue.go
```diff
@@ -34,43 +35,77 @@ type messageKey string
 
 // mKey returns an ID that represents a potentially retryable message (msg.ID is the same for messages
 // with different signers, slots, types, rounds, etc. - so we can't use just msg.ID as a unique identifier)
-func mKey(msg *queue.SSVMessage, logger *zap.Logger) messageKey {
-	const idUndefined = "undefined"
+func mKey(msg *queue.SSVMessage) (messageKey, error) {
+	const mKeyUndefined = "undefined"
+
 	msgSlot, err := msg.Slot()
```
momosh-ssv, line 39: If `Body` is a nil typed pointer (e.g. `(*specqbft.Message)(nil)`), `Slot()` will match the type-switch case and dereference nil before this error check runs. The nil guards added throughout this PR are bypassed because `Slot()` panics first. Each case arm in `Slot()` needs `if m == nil { return 0, ErrUnknownMessageType }`.

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -137,7 +137,7 @@ func (v *Validator) StartQueueConsumer(
 				// for the current height and round.
 				filter = func(m *queue.SSVMessage) bool {
 					qbftMsg, ok := m.Body.(*specqbft.Message)
-					if !ok {
+					if !ok || qbftMsg == nil {
```
momosh-ssv, line 140: This filter was fixed, but the adjacent filter at line 129 (`e, ok := m.Body.(*types.EventMsg)`) was not. If `e` is a nil typed pointer, `e.Type` at line 133 will panic. Add `|| e == nil` to the `!ok` check at line 130 for consistency.

momosh-ssv (commented):
Good direction — systematically eliminates panic vectors from unsafe type assertions across the message pipeline. 
Two items that should be fixed, and two to consider:

**Should fix:**

1. **`Slot()` nil-deref defeats the PR's purpose.** `Slot()` in `queue/messages.go:30-56` uses a type switch that matches nil typed pointers (e.g. `(*specqbft.Message)(nil)`) and immediately dereferences them (`m.Height`, `m.Slot`, `m.GetTimeoutData()`). Since `mKey()`, `EnqueueMessage()`, and other callers invoke `Slot()` before reaching the new nil checks, the panic still happens — just one call frame earlier. Each case arm in `Slot()` needs a nil guard.

2. **Missed filter lambda.** See inline comment on `queue_validator.go`.

**Consider:**

3. `committee.go:313-314` — `eventMsg, ok := msg.Body.(*types.EventMsg)` is missing `|| eventMsg == nil`, unlike every other similar pattern in this PR.

4. `queue/messages.go` `scoreMessageSubtype` / `scoreCommitteeMessageSubtype` — `isConsensusMessage` is set to `true` for nil typed pointers since only the bool is checked, not the value. Low risk but inconsistent with the defensive pattern.

protocol/v2/ssv/validator/msg_queue.go
```diff
@@ -34,43 +35,77 @@ type messageKey string
 
 // mKey returns an ID that represents a potentially retryable message (msg.ID is the same for messages
 // with different signers, slots, types, rounds, etc. - so we can't use just msg.ID as a unique identifier)
-func mKey(msg *queue.SSVMessage, logger *zap.Logger) messageKey {
-	const idUndefined = "undefined"
+func mKey(msg *queue.SSVMessage) (messageKey, error) {
+	const mKeyUndefined = "undefined"
```
oleg-ssvlabs: since `error` as return value was added - I don't think `mKeyUndefined` variable serves any purpose? All the callers do error check and if it is not equal to nil - they do not read message key. Also, `mKeyUndefined` is only returned with `error` (error is not `nil`). 
So, I believe, this constant could be just removed

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -277,28 +281,31 @@ func (c *Committee) logWithMessageFields(logger *zap.Logger, msg *queue.SSVMessa
...
+		logger = logger.With(fields.QBFTRound(qbftMsg.Round), fields.QBFTHeight(qbftMsg.Height))
+	}
 
 	return logger, nil
 }
 
 // mKey is a wrapper that provides a logger to report errors (if any).
-func (c *Committee) mKey(msg *queue.SSVMessage) messageKey {
-	return mKey(msg, c.logger)
+func (c *Committee) mKey(msg *queue.SSVMessage) (messageKey, error) {
```
oleg-ssvlabs: maybe this method could be entirely removed? Also, on Validator object (`func (v *Validator) mKey(...)`) in `queue_validator.go`

oleg-ssvlabs (commented):
gj!
Left few comments

protocol/v2/ssv/validator/msg_queue.go
iurii-ssv: Good find, added missing nil-checks in https://github.com/ssvlabs/ssv/pull/2723/changes/2aaf47b0566a377c2632ea62f7bcec4407430b0c

protocol/v2/ssv/validator/queue_validator.go
iurii-ssv: resolved, along with other comments from above in https://github.com/ssvlabs/ssv/pull/2723/changes/2aaf47b0566a377c2632ea62f7bcec4407430b0c

protocol/v2/ssv/validator/msg_queue.go
iurii-ssv: True, applied in https://github.com/ssvlabs/ssv/pull/2723/changes/2aaf47b0566a377c2632ea62f7bcec4407430b0c

protocol/v2/ssv/validator/committee_queue.go
iurii-ssv: Right, now we no longer need to pass in the `logger` ... so it's no longer needed, removed in https://github.com/ssvlabs/ssv/pull/2723/changes/2aaf47b0566a377c2632ea62f7bcec4407430b0c

olegshmuelov (approved):

oleg-ssvlabs (approved):

julienh-ssv (approved):

ljuba-ssv (approved):

y0sher:
## Staging Test Report (operators 89-92, ~20 min soak)

**Branch:** `runners-handle-unexpected-messages-gracefully`
**Baseline:** operators 29-32 running `stage`

### Results: PASS

| Check | Result |
|-------|--------|
| Panics | **Zero** across all 4 nodes |
| Rejected messages | **Zero** — nil checks not rejecting valid messages |
| Duties completing | All roles: attestation, aggregation, sync committee, validator registration |
| Consensus rounds | Round 1 consistently — no performance regression |
| P2P peers | 5-6 per node (healthy) |
| Dropped messages | DEBUG level only, expected reasons (late consensus, stale partial sigs) |

### Message Validation Throughput (accepted, since restart)

| Node | AGGREGATOR | COMMITTEE | SYNC_COMMITTEE | VALIDATOR |
|------|-----------|-----------|----------------|-----------|
| 89 | 2,842 | 208 | 157 | 320 |
| 90 | 2,842 | 208 | 157 | 320 |
| 91 | 2,264 | 160 | 137 | 256 |
| 92 | 3,582 | 260 | 199 | 396 |

### Errors Observed (unrelated to PR)

- `"could not publish p2p message — validation ignored"` — sporadic P2P publishing issue, pre-existing
- `"failed to find event by ID"` — on-chain event sync issue, burst at 12:31:24Z, pre-existing

### Notes

Ignored messages are minimal (2-10 per node), all with reason `"got decided message with too few signers"` — normal QBFT behavior. Zero nil-pointer related issues.

Recommend extending soak to 24h+ before merging to main, but initial results are clean.

---

PR #2726, opened 2026-03-18
fix(leaks) - interval timer leak and p2p context cancellations

oleg-ssvlabs:
# Summary

This PR fixes goroutine and timer leak paths in the network layer by making periodic background loops respect shutdown and by stopping the interval ticker when the loop exits.

Specifically:

- `utils/async/interval.go`: stop the internal `time.Ticker` on exit to avoid leaking timer resources
- `network/p2p/p2p.go`: make `UpdateSubnets` and `UpdateScoreParams` exit when the P2P network context is cancelled

This keeps shutdown behavior aligned with `p2pNetwork.Close()`, which already cancels `n.ctx`.

# Problem

Two periodic loop patterns were not fully tied to shutdown:

- `Interval(...)` created a `time.Ticker` but never called `ticker.Stop()`
- `UpdateSubnets()` and `UpdateScoreParams()` were launched as background goroutines, but their loop wait paths did not consistently observe `n.ctx.Done()`

### DIFF
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -472,10 +472,8 @@ func (n *p2pNetwork) UpdateSubnets() {
 
 	// Run immediately and then every second.
 	for {
-		select {
-		case <-n.ctx.Done():
+		if n.ctx.Err() != nil {
 			return
-		default:
 		}
 
 		start := time.Now()
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -471,7 +471,13 @@ func (n *p2pNetwork) UpdateSubnets() {
 	defer ticker.Stop()
 
 	// Run immediately and then every second.
-	for ; true; <-ticker.C {
+	for {
+		select {
+		case <-n.ctx.Done():
+			return
+		default:
+		}
+
 		start := time.Now()
 
 		updatedSubnets := n.SubscribedSubnets()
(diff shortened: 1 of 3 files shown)


network/p2p/p2p.go
```diff
@@ -471,7 +471,13 @@ func (n *p2pNetwork) UpdateSubnets() {
 	defer ticker.Stop()
 
 	// Run immediately and then every second.
-	for ; true; <-ticker.C {
+	for {
+		select {
+		case <-n.ctx.Done():
+			return
+		default:
+		}
```
iurii-ssv: Nit: something like this reads a bit better to me (cause shorter):
```go
if n.ctx.Err() != nil {
	return
}
```

network/p2p/p2p_shutdown_test.go
```diff
@@ -0,0 +1,105 @@
...
+	go func() {
+		defer close(done)
+		n.UpdateSubnets()
+	}()
+
+	cancel()
+
+	select {
+	case <-done:
+	case <-time.After(200 * time.Millisecond):
```
iurii-ssv: I would bump this "timeout" to at least `400 * time.Millisecond` - just an empirical value I found out preventing the tests from being flaky when running concurrently (since https://github.com/ssvlabs/ssv/pull/2716),

I guess applies to other similar test-scenarios below.

utils/async/interval.go
```diff
@@ -6,11 +6,11 @@ import (
 )
 
 // Interval runs the provided function periodically every period
-func Interval(pctx context.Context, interval time.Duration, f func()) {
+func Interval(ctx context.Context, interval time.Duration, f func()) {
 	ticker := time.NewTicker(interval)
 	go func() {
-		ctx, cancel := context.WithCancel(pctx)
-		defer cancel()
+		defer ticker.Stop()
```
iurii-ssv: Just for simplicity, I would move the `ticker := time.NewTicker(interval)` inside of `go func()...`

iurii-ssv (dismissed):
Looks good, just minor suggestions

network/p2p/p2p.go
oleg-ssvlabs: agree, looks simpler. Fixed b8c683faea1af9fcb9bfb73335c2b525e5f134aa

network/p2p/p2p_shutdown_test.go
oleg-ssvlabs: fixed b8c683faea1af9fcb9bfb73335c2b525e5f134aa

utils/async/interval.go
oleg-ssvlabs: fixed b8c683faea1af9fcb9bfb73335c2b525e5f134aa

iurii-ssv (approved):
Nice!

y0sher (approved):

y0sher (approved):

PR #2727, opened 2026-03-18
Peer scoring algorithm

julienharbulot:
O(n^2) Peer Scoring During Trimming

Improve peer scoring performance from O(n^2) to O(n log(n)) by precomputing peer scores before sorting.

### DIFF
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -595,8 +595,9 @@ func (n *p2pNetwork) buildPeerTrimScores(peerIDs []peer.ID) map[peer.ID]float64 
 	trimSnapshot := n.newPeerTrimSnapshot()
 	scores := make(map[peer.ID]float64, len(peerIDs))
 	for _, peerID := range peerIDs {
+		peerContribution := trimSnapshot.peerSubnetPeers[peerID]
 		peerSubnets, _ := n.PeersIndex().GetPeerSubnets(peerID)
-		scores[peerID] = trimSnapshot.score(peerID, peerSubnets)
+		scores[peerID] = trimSnapshot.score(peerContribution, peerSubnets)
 	}
 	return scores
 }
@@ -628,6 +629,9 @@ func (n *p2pNetwork) newPeerTrimSnapshot() peerTrimSnapshot {
 		for _, peerID := range peers {
 			snapshot.totalSubnetPeers[subnet]++
 
+			// peerSubnetPeers is populated from the same topic membership snapshot as
+			// totalSubnetPeers, so each peer contribution count is a subset of the
+			// corresponding total and can be subtracted safely when scoring.
 			peerSubnetPeers := snapshot.peerSubnetPeers[peerID]
 			peerSubnetPeers[subnet]++
 			snapshot.peerSubnetPeers[peerID] = peerSubnetPeers
@@ -637,8 +641,8 @@ func (n *p2pNetwork) newPeerTrimSnapshot() peerTrimSnapshot {
 	return snapshot
 }
 
-func (s peerTrimSnapshot) score(peerID peer.ID, peerSubnets commons.Subnets) float64 {
-	subnetPeersExcluding := s.totalSubnetPeers.Subtract(s.peerSubnetPeers[peerID])
+func (s peerTrimSnapshot) score(peerContribution SubnetPeers, peerSubnets commons.Subnets) float64 {
+	subnetPeersExcluding := s.totalSubnetPeers.Subtract(peerContribution)
 	return subnetPeersExcluding.Score(s.ownSubnets, peerSubnets)
 }
 
@@ -665,9 +669,17 @@ func (a SubnetPeers) Add(b SubnetPeers) SubnetPeers {
 	return sum
 }
 
+// Subtract returns the element-wise difference between two subnet peer snapshots.
+// In trim scoring, b is expected to be a per-peer contribution derived from the
+// same topic membership snapshot as a, so each b[i] should be less than or equal
+// to a[i]. If a future caller violates that relationship, the result is clamped
+// to 0 for that subnet instead of allowing uint16 underflow.
 func (a SubnetPeers) Subtract(b SubnetPeers) SubnetPeers {
 	var diff SubnetPeers
 	for i := range a {
+		if b[i] > a[i] {
+			continue
+		}
 		diff[i] = a[i] - b[i]
 	}
 	return diff
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -381,7 +381,7 @@ func (n *p2pNetwork) peersTrimming() func() {
 }
 
 // choosePeersToTrim returns a map of peers that are least-valuable to us based on how much
-// (dead/solo/duo) they contribute to us (as defined by peerScore func).
+// (dead/solo/duo) they contribute to us.
 func (n *p2pNetwork) choosePeersToTrim(trimCnt int, trimInboundOnly bool) map[peer.ID]struct{} {
 	myPeers, err := n.topicsCtrl.Peers("")
 	if err != nil {
@@ -594,16 +594,22 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
 // buildPeerTrimScores snapshots topic membership once and computes trim scores
 // for the given peers.
 //
-// ownSubnets is our local subscription bitset: which subnets this node cares
-// about when evaluating a peer's usefulness.
+// The peer-scores are calculated based on:
+//   - ownSubnets is the desired set of subnets we want to be connected to
+//   - ownSubnetPeers is our own currently connected peer count per subnet across all
+//     peers in our topic mesh
+//   - peerSubnets tracks all the subnets each of our peers is connected to
 //
-// totalSubnetPeers is the current connected peer count per subnet across all
-// peers in our topic mesh. For each candidate peer we subtract only that peer's
-// own contribution from this total before scoring it.
+// Algo:
+//   - calculate (take snapshot of) ownSubnets, ownSubnetPeers, peerSubnets
+//   - for each candidate peer we need to score:
+//   - calculate subnetPeersExcluding (currently connected subnets IF that candidate peer is excluded/disconnected)
+//   - use SubnetPeers.Score to calculate the final peer-score for each peer (based on: subnetPeersExcluding, desired
+//     set of subnets, subnets this peer is connected to)
 func (n *p2pNetwork) buildPeerTrimScores(peerIDs []peer.ID) map[peer.ID]float64 {
 	ownSubnets := n.SubscribedSubnets()
-	totalSubnetPeers := newSubnetPeers()
-	peerSubnetPeers := make(map[peer.ID]SubnetPeers)
+	ownSubnetPeers := newSubnetPeers()
+	peerSubnets := make(map[peer.ID]commons.Subnets)
 
 	for topic, peers := range n.PeersByTopic() {
 		subnet, ok := n.topicSubnet(topic)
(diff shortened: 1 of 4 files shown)


network/p2p/p2p.go
```diff
@@ -590,34 +591,59 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
-				zap.String("topic", topic), zap.Error(err))
+			n.logger.Error("failed to parse topic", zap.String("topic", topic), zap.Error(err))
 			continue
 		}
 		if subnet < 0 || subnet >= commons.SubnetsCount {
-			n.logger.Error("invalid topic",
-				zap.String("topic", topic), zap.Int("subnet", int(subnet)))
+			n.logger.Error("invalid topic", zap.String("topic", topic), zap.Int("subnet", int(subnet)))
 			continue
 		}
```
iurii-ssv, line 618: seems like this block of code is duplicated here and in [p2p_discovery.go](https://github.com/ssvlabs/ssv/blob/e4623ec2a1a4dbcf84e1000e2b55e3b7a3b0c15d/network/p2p/p2p_discovery.go#L71-L83):
```go
		subnet, err := strconv.ParseInt(commons.GetTopicBaseName(topic), 10, 64)
		if err != nil {
			n.logger.Error("failed to parse topic",
				zap.String("topic", topic), zap.Error(err))
			n.logger.Error("failed to parse topic", zap.String("topic", topic), zap.Error(err))
			continue
		}
		if subnet < 0 || subnet >= commons.SubnetsCount {
			n.logger.Error("invalid topic",
				zap.String("topic", topic), zap.Int("subnet", int(subnet)))
			n.logger.Error("invalid topic", zap.String("topic", topic), zap.Int("subnet", int(subnet)))
			continue
		}
```
we should probably create a helper-func for it

network/p2p/p2p.go
```diff
@@ -590,34 +591,59 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
+		peerSubnets, _ := n.PeersIndex().GetPeerSubnets(peerID)
+		scores[peerID] = trimSnapshot.score(peerContribution, peerSubnets)
+	}
+	return scores
+}
+
+type peerTrimSnapshot struct {
+	ownSubnets       commons.Subnets
+	totalSubnetPeers SubnetPeers
+	peerSubnetPeers  map[peer.ID]SubnetPeers
```
iurii-ssv: `totalSubnetPeers`->`ownSubnetPeers`, to keep it consistent with other names here ?

But also, I think there is no need to wrap all these fields into a `peerTrimSnapshot struct` since it's doesn't add much of anything, but creates additional syntax clutter to worry about ... I'd rather inline all this snapshot-related stuff directly into `buildPeerTrimScores` func ... I understand the intent to make `buildPeerTrimScores` "high-level and simple" - but I wouldn't say it is (you still need to figure out how "peerTrimSnapshot" is constructed to make sense of it)

iurii-ssv (commented):
Good to have! Would like to see [a bit of code consolidation](https://github.com/ssvlabs/ssv/pull/2727#discussion_r2960628618) though (kinda hard to wrap my head around the change in its current form)

network/p2p/p2p.go
iurii-ssv: I'm also not entirely sure why we need even need to define something like  `SubnetPeers.Subtract`, 

from what I understand, in this PR, we basically just need to move `n.peerScore` computations out of the `slices.SortFunc` call to create a mapping we can reference (and use that) ... kinda worried we might change the logic here if we are not careful with how we move this code around (I'm not against refactoring/clarifying it, just want to make sure we do it a tiny-step at a time)

network/p2p/p2p.go
julienh-ssv: Agreed. I extracted a shared topicSubnet helper and switched both trim scoring and discovery to use it.

network/p2p/p2p.go
julienh-ssv: On totalSubnetPeers rename let's keep totalSubnetPeers because it counts all connected peers on each subnet; ownSubnets already means our local subscription set, so ownSubnetPeers would be ambiguous.
On peerTrimSnapshot / Subtract: Agreed. I inlined the snapshot bookkeeping into buildPeerTrimScores and removed the extra struct/helper so the change stays closer to the original scoring flow.

julienh-ssv:
```
    // ownSubnets is our local subscription bitset: which subnets this node cares
    // about when evaluating a peer's usefulness.
    //
    // totalSubnetPeers is the current connected peer count per subnet across all
    // peers in our topic mesh. For each candidate peer we subtract only that peer's
    // own contribution from this total before scoring it.
```

network/p2p/p2p.go
```diff
@@ -590,34 +591,67 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
+		subnet, ok := n.topicSubnet(topic)
+		if !ok {
 			continue
 		}
-		if subnet < 0 || subnet >= commons.SubnetsCount {
-			n.logger.Error("invalid topic",
-				zap.String("topic", topic), zap.Int("subnet", int(subnet)))
-			continue
+
+		totalSubnetPeers[subnet] = uint16(len(peers)) //nolint: gosec
```
momosh-ssv: Should we double-check whether `PeersByTopic()` can ever return two distinct topic strings that both parse to the same subnet index? This line overwrites `totalSubnetPeers[subnet]` rather than accumulating, so if there's a collision the second topic silently replaces the first. Meanwhile `peerSubnetPeers` increments per-appearance across all topics — so the subtraction on line 635 could over-subtract (hitting the clamp-to-zero path) and inflate scores.

The old code accumulated with `subnetPeersExcluding[subnet]++` which was additive by design. This semantic difference matters if topic collisions are possible.

network/p2p/p2p.go
```diff
@@ -590,34 +591,67 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
+}
 
-	ownSubnets := n.SubscribedSubnets()
-	peerSubnets, _ := n.PeersIndex().GetPeerSubnets(peerID)
-	return subnetPeersExcluding.Score(ownSubnets, peerSubnets)
+// topicSubnet parses a topic name into a subnet index and logs malformed topics.
+func (n *p2pNetwork) topicSubnet(topic string) (int, bool) {
+	subnet, err := strconv.ParseInt(commons.GetTopicBaseName(topic), 10, 64)
+	if err != nil {
+		n.logger.Error("failed to parse topic", zap.String("topic", topic), zap.Error(err))
```
momosh-ssv, line 652: What do you think about downgrading these to `Debug`? Non-subnet topics (committee/validator pubsub topics) will hit this path every trim cycle, producing Error-level log noise in production. Since `topicSubnet` is now a shared helper called from both `buildPeerTrimScores` and `startDiscovery`, fixing once here cleans it up everywhere.

network/p2p/p2p_trimming_test.go
```diff
@@ -0,0 +1,449 @@
...
+				"1": {candidate, other1},
+				"2": {candidate, other1, other2},
+				"3": {candidate, other1, other2, other3},
+			},
+		},
+		idx,
+		subnetsFor(0, 1, 2, 3),
+	)
+
+	assert.Equal(t, 21.0, singleTrimScore(network, candidate))
```
momosh-ssv: What do you think about adding a short comment showing the derivation? Something like `// 16 (dead, subnet 0) + 4 (solo, subnet 1) + 1 (duo, subnet 2) + 0 (trio, subnet 3) = 21`. The test name says "characterizes" which is great as a safety net, but the magic number will be hard to maintain if scoring weights ever change.

momosh-ssv (commented):
This touches the peer trimming hot path which directly affects network health and validator performance. I'd suggest we take the time to be thorough here rather than rush it through — a subtle scoring bug could silently degrade peer selection across the fleet.

The O(n²) → O(n) precomputation approach is sound. A few things to verify and clean up below.

iurii-ssv (changes requested):
Looks much cleaner now, but we can simplify & improve it further with https://github.com/ssvlabs/ssv/pull/2737 (lets try to merge it into this PR ?)

network/p2p/p2p_trimming_test.go
iurii-ssv: Already done in https://github.com/ssvlabs/ssv/pull/2737

network/p2p/p2p.go
iurii-ssv: > Non-subnet topics (committee/validator pubsub topics) will hit this path every trim cycle

Not sure what you mean there, every topic this func will handle is supposed to map onto 1 (out of 128 total) subnet,

the validation in `topicSubnet` that logs errors seems very generic to me (it doesn't seem like our code CAN pass in an invalid topic name here - and if it does, there is a bug somewhere in the caller we should know about ASAP ... so logging as Error is very appropriate due to that)

network/p2p/p2p.go
iurii-ssv: > Should we double-check whether PeersByTopic() can ever return two distinct topic strings that both parse to the same subnet index?

As far as I know, distinct topic-names should map onto different subnets (a subnet is just a number in 0...127 range, and a topic-name is a string-representation for a subnet ... we use them interchangeably), 

so, there should not be any issues like this concern describes (it is a valid concern, just doesn't really apply to our code) ... unless we are undergoing a protocol fork that renames topics producing 2 different topic-names per subnet - which we are planning to do soon ... let me check if that's handled somehow there

iurii-ssv: > unless we are undergoing a protocol fork that renames topics producing 2 different topic-names per subnet - which we are planning to do soon ... let me check if that's handled somehow there

for the upcoming Boole-fork (the `boole-fork` branch) it seems to be an real issue ... this place Momo spotted + another one denoted by `// Compute the number of peers we're connected to for each subnet.` comment are the only parts affected as far as I can tell

I suggest we:
- merge this PR as is since it doesn't know/care about Boole fork
- pull this code into `boole-fork` branch
- create a PR to fix the issue for both places we spotted (targeting `boole-fork` branch) - I've created an issue for this https://github.com/ssvlabs/ssv/issues/2740

cc @julienh-ssv @nkryuchkov fyi

julienh-ssv: @momosh-ssv thanks for pointing it out. I reestablished the old behavior in my latest commit. Seems safer since changing the scoring algorithm is not in scope for this ticket -- let's just focus on the caching optimization.

network/p2p/p2p.go
julienh-ssv: I suggest to keep Error as hitting an invalid topic here would indicate a bug, unless I'm mistaken.

network/p2p/p2p.go
```diff
@@ -639,7 +639,8 @@ func (n *p2pNetwork) buildPeerTrimScores(peerIDs []peer.ID) map[peer.ID]float64
 			}
 		}
 
-		scores[peerID] = subnetPeersExcluding.Score(ownSubnets, pSubnets)
+		advertisedSubnets, _ := n.PeersIndex().GetPeerSubnets(peerID)
+		scores[peerID] = subnetPeersExcluding.Score(ownSubnets, advertisedSubnets)
```
iurii-ssv, line 643: As I explained in https://github.com/ssvlabs/ssv/pull/2737 I think it's better to use `pSubnets` here directly ... but I can sketch a separate PR for this specific modification if you want to keep this refactoring free from that change. Just pointing it out for historical purposes.

iurii-ssv (approved):
LGTM

network/p2p/p2p.go
julienh-ssv: Sounds good, let's stash that functional change under a different ticket.

network/p2p/p2p.go
momosh-ssv: > Not sure what you mean there, every topic this func will handle is supposed to map onto 1 (out of 128 total) subnet...

ahh my bad, I traced the subscribe paths and `CommitteeTopicID` resolves to `fmt.Sprintf("%d", CommitteeSubnet(cid))`...
So committee topics are just numeric subnet IDs. There are no non-numeric topics hitting `topicSubnet`. My mistake.

momosh-ssv (approved):
really gj here!

ljuba-ssv (approved):

y0sher:
## Staging P2P Health Test — `stage` branch (includes PR #2727)

Deployed to nodes **89-92** on `stage-hoodi` and compared against baseline nodes.

### Peers Connected
| Nodes | Peers |
|-------|-------|
| **89-92 (stage)** | 4-6 (ramping, ~5 min uptime) |
| **1,3,4 (baseline)** | 10 (long-running) |
| **33,34 (baseline)** | 3 |

Peer count normal for freshly started nodes.

### Message Validation Duration (avg, seconds)
| Nodes | AGGREGATOR | COMMITTEE | SYNC_COMMITTEE | VALIDATOR_REG |
|-------|-----------|-----------|----------------|---------------|
| **89-92 (stage)** | 0.0008–0.0022 | 0.0004–0.0005 | 0.0004–0.0026 | 0.0011–0.0024 |
| **Baseline** | 0.0004–0.0005 | 0.0004–0.0006 | N/A | 0.0007–0.0009 |

All sub-millisecond — no regression.

### Rejections & Ignores
- **No rejected messages** on either deployed or baseline (metric doesn't exist on either)
- **Ignore rates** comparable: deployed ~0.02–0.03/s vs baseline ~0.02–0.08/s (same "too few signers" pattern, normal)
- **Filtered connections**: near-zero on both

### Consensus Duration (seconds)
| Nodes | AGGREGATOR | COMMITTEE |
|-------|-----------|-----------|
| **89-92 (stage)** | 0.01–0.024 | 0.018–0.020 |
| **1,3,4 (baseline)** | 0.35–0.57 | 1.95–2.25 |
| **33,34 (baseline)** | 0.03 | 0.02 |

### Errors
Only "validation ignored" on topic 103 — common for freshly started nodes, not a concern.

### Verdict
**P2P is healthy.** No new rejections or ignores introduced. Scoring works correctly with sub-millisecond validation durations. The O(n²) → O(n log n) peer scoring change shows no regressions.

PR #2728, opened 2026-03-19
message/validation: move signature verification outside validation lock

nkryuchkov:
**Summary**
Move message signature verification out of the per-`MessageID` validation critical section.

Previously, incoming consensus and partial-signature messages were serialized under the same per-`MessageID` mutex for the full validation flow, including RSA signature verification. This meant messages for the same validator/role could queue behind each other while spending time in crypto verification.

This change keeps the existing stateful validation and state update serialized, but performs signature verification before acquiring the validation lock. That reduces lock hold time and allows signature verification to proceed in parallel for messages sharing the same `MessageID`.

**What Changed**
- Consensus validation now performs:
  1. decode + semantic validation
  2. signature verification
  3. locked stateful validation + state update
- Partial-signature validation now follows the same pattern
- The outer dispatcher no longer holds the validation lock across the whole message flow
- Added regression tests to verify signature verification can proceed while the per-`MessageID` validation lock is held

**Why**
This reduces contention in the message validation pipeline during active consensus, where multiple operators send messages nearly simultaneously for the same committee/role. Expensive signature verification no longer blocks other messages from progressing to their own signature checks.

**Behavioral Notes**
- Stateful validation order remains unchanged within the locked phase
- We intentionally keep all mutable-state checks and updates under a single lock acquisition
- Cheap stateful rejections may still happen after signature verification; this preserves simpler semantics while still shrinking the critical section

### DIFF
--- a/message/validation/consensus_validation.go
+++ b/message/validation/consensus_validation.go
@@ -47,28 +47,23 @@ func (mv *messageValidator) validateConsensusMessage(
 		return consensusMessage, err
 	}
 
-	state := mv.validatorState(ssvMessage.GetID(), committeeInfo)
-
-	if err := mv.validateQBFTLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, receivedAt, state); err != nil {
+	if err := mv.verifyConsensusMessageSignatures(signedSSVMessage); err != nil {
 		return consensusMessage, err
 	}
 
-	if err := mv.validateQBFTMessageByDutyLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedAt, state); err != nil {
-		return consensusMessage, err
-	}
+	if err := mv.withValidationLock(ssvMessage.GetID(), func() error {
+		state := mv.validatorState(ssvMessage.GetID(), committeeInfo)
 
-	for i := range signedSSVMessage.Signatures {
-		operatorID := signedSSVMessage.OperatorIDs[i]
-		signature := signedSSVMessage.Signatures[i]
-
-		if err := mv.signatureVerifier.VerifySignature(operatorID, ssvMessage, signature); err != nil {
-			e := ErrSignatureVerification
-			e.innerErr = fmt.Errorf("verify opid: %v signature: %w", operatorID, err)
-			return consensusMessage, e
+		if err := mv.validateQBFTLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, receivedAt, state); err != nil {
+			return err
 		}
-	}
 
-	if err := mv.updateConsensusState(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, state); err != nil {
+		if err := mv.validateQBFTMessageByDutyLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedAt, state); err != nil {
+			return err
+		}
+
+		return mv.updateConsensusState(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, state)
+	}); err != nil {
 		return consensusMessage, err
 	}
 
@@ -166,6 +161,22 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
 
 	if err := mv.validateJustifications(consensusMessage); err != nil {
 		return err
+	}
+
+	return nil
+}
+
+func (mv *messageValidator) verifyConsensusMessageSignatures(signedSSVMessage *spectypes.SignedSSVMessage) error {
+	ssvMessage := signedSSVMessage.SSVMessage
+	for i := range signedSSVMessage.Signatures {
+		operatorID := signedSSVMessage.OperatorIDs[i]
+		signature := signedSSVMessage.Signatures[i]
+
+		if err := mv.signatureVerifier.VerifySignature(operatorID, ssvMessage, signature); err != nil {
+			e := ErrSignatureVerification
+			e.innerErr = fmt.Errorf("verify opid: %v signature: %w", operatorID, err)
+			return e
+		}
 	}
 
 	return nil
@@ -281,7 +292,9 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
 	}
 
 	msgSlot := phase0.Slot(consensusMessage.Height)
+
 	randaoMsg := false
+	// Rule: Message must correspond to a known beacon duty when the role requires it.
 	if err := mv.validateBeaconDuty(signedSSVMessage.SSVMessage.GetID().GetRoleType(), msgSlot, committeeInfo.validatorIndices, randaoMsg); err != nil {
 		return err
 	}
--- a/message/validation/consensus_validation.go
+++ b/message/validation/consensus_validation.go
@@ -294,7 +294,6 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
 	msgSlot := phase0.Slot(consensusMessage.Height)
 
 	randaoMsg := false
-	// Rule: Message must correspond to a known beacon duty when the role requires it.
 	if err := mv.validateBeaconDuty(signedSSVMessage.SSVMessage.GetID().GetRoleType(), msgSlot, committeeInfo.validatorIndices, randaoMsg); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


nkryuchkov:

message/validation/consensus_validation.go
```diff
@@ -47,28 +47,23 @@ func (mv *messageValidator) validateConsensusMessage(
...
-			e.innerErr = fmt.Errorf("verify opid: %v signature: %w", operatorID, err)
-			return consensusMessage, e
+		if err := mv.validateQBFTMessageByDutyLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedAt, state); err != nil {
+			return err
 		}
-	}
 
-	if err := mv.updateConsensusState(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, state); err != nil {
+		return mv.updateConsensusState(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, state)
+	}); err != nil {
```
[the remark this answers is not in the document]
iurii-ssv: > but it may be worth considering whether a subset of the cheapest stateful checks (e.g. `ErrRoundAlreadyAdvanced`) could be moved back before signature verification to provide an inexpensive early-exit for replay traffic, without reintroducing lock contention for the common case.

TBH I don't think this is worth the added code complexity.

iurii-ssv (approved):
LGTM, I would also ask @MatheusFranco99 to take a look (specifically at [this concern AI bot raised](https://github.com/ssvlabs/ssv/pull/2728#discussion_r2960167297)) ... IMO this PR doesn't pose much of an attack-danger because we REJECT (not INGORE) those invalid-signature messages and duplicate-peer messages - so an attacker won't be able to "sustain" his attack for a meaningfully long period of time that would result in issues for SSV node in practice.

y0sher:
## Staging Test Report

| | |
|---|---|
| **Branch** | `msgval-sig-lock-split` |
| **Nodes** | `ssv-node-71`, `ssv-node-72` (stage-hoodi) |
| **Status** | ✅ **PASS** |

### Results

- ✅ **Startup**: Node started successfully
- ✅ **Duties**: All duties completing successfully with `consensus_rounds=1`
- ✅ **No panics/crashes**: Only benign `validation ignored` p2p messages
- ✅ **No race conditions**: Zero deadlock, race, or concurrent access issues observed
- ✅ **Consensus timing**: Healthy ~0.04-0.18s total consensus time

### Fix Verification

The PR splits the validation flow: signature verification (CPU-intensive RSA ops) now runs **before** acquiring the per-MessageID validation lock, while stateful checks (QBFT logic, duty logic, validator state) remain under the lock via the new `withValidationLock()` helper. Both consensus and partial signature paths were restructured with dedicated `verifyConsensusMessageSignatures()` and `verifyPartialSignatureMessageSignature()` functions that execute lock-free. A new `validation_lock_test.go` (224 lines) verifies the lock behavior. Round changes seen only on stale aggregator duties (expected). No lock contention issues observed.

---

nkryuchkov:
> LGTM, I would also ask @MatheusFranco99 to take a look (specifically at [this concern AI bot raised](https://github.com/ssvlabs/ssv/pull/2728#discussion_r2960167297)) ... IMO this PR doesn't pose much of an attack-danger because we REJECT (not INGORE) those invalid-signature messages and duplicate-peer messages - so an attacker won't be able to "sustain" his attack for a meaningfully long period of time that would result in issues for SSV node in practice.

~~I opened [a draft PR with an actor-based approach](https://github.com/ssvlabs/ssv/pull/2736) as an alternative solution. We can discuss it there, and if we agree that actors are better in this case, I'll finish that PR and merge it into this one.~~

UPD: I closed https://github.com/ssvlabs/ssv/pull/2736 because I don't like that solution

nkryuchkov:
Converting to draft until we make sure we need it

message/validation/consensus_validation.go
```diff
@@ -171,6 +166,22 @@ func (mv *messageValidator) validateConsensusMessageSemantics(
 	return nil
 }
 
+func (mv *messageValidator) verifyConsensusMessageSignatures(signedSSVMessage *spectypes.SignedSSVMessage) error {
```
olegshmuelov, line 169: The board finding (F-ssv-173) recommended three optimizations:

1. Move sig verification outside the lock — **this PR**
2. Verify multiple signatures in parallel for decided messages (up to 13 sigs × ~1.5ms = ~20ms sequential) — **not addressed**
3. Cache recently verified (operatorID, messageHash, signature) tuples to skip re-verification on retransmitted messages — **not addressed**

This PR reduces lock hold time but doesn't reduce the total CPU cost of verification itself. A decided message with 4 signatures still takes ~6ms of sequential RSA verification before acquiring the lock. Consider whether 2 (parallel verification in `verifyConsensusMessageSignatures` via a goroutine pool) should be part of this PR or tracked as a follow-up.

nkryuchkov: (2) and (3) depend on (1), so I want to decide on (1) first before continuing. It moves the expensive signature verification before the light checks, which might be worse than keeping the lock

Once we decide on it, (2) and (3) will be ready for implementation

iurii-ssv: > 3\. Cache recently verified (operatorID, messageHash, signature) tuples to skip re-verification on retransmitted messages — **not addressed**

**libp2p** probably does some basic de-duplication for us already (not sure if by messageHash or somehow else) ? Probably want to check if our own caching will even catch any such duplicates before proceeding to it

message/validation/consensus_validation.go
```diff
@@ -47,28 +47,23 @@ func (mv *messageValidator) validateConsensusMessage(
 		return consensusMessage, err
 	}
 
-	state := mv.validatorState(ssvMessage.GetID(), committeeInfo)
-
-	if err := mv.validateQBFTLogic(signedSSVMessage, consensusMessage, committeeInfo, receivedFrom, receivedAt, state); err != nil {
+	if err := mv.verifyConsensusMessageSignatures(signedSSVMessage); err != nil {
```
momosh-ssv, line 50: What do you think about per-peer CPU amplification here? I fear a single peer could send multiple structurally-different messages with the same MsgID (varying Round, FullData, etc.) that all pass semantic validation and enter `verifyConsensusMessageSignatures` in parallel before the per-peer `SeenMsgTypes` limits kick in inside the lock. Pre-PR the outer lock capped simultaneous RSA work per MsgID at 1; now it's bounded only by libp2p's validation-worker pool. Maybe worth a follow-up: a cheap pre-verify per-peer counter keyed on `(MsgID, peerID)`, or rely on topic scoring — could be bundled with the F-ssv-173 (2)/(3) follow-ups. Same applies to `partial_validation.go` verify path.

message/validation/validation_lock_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}()
+
+	done := make(chan error, 1)
+	go func() {
+		_, err := env.validator.handleSignedSSVMessage(signedSSVMessage, topicID, peerID, env.netCfg.SlotStartTime(slot))
+		done <- err
+	}()
+
+	select {
+	case <-env.validator.signatureVerifier.(*observingSignatureVerifier).called:
```
momosh-ssv, line 155: Worth considering whether this test actually proves the parallel-verify claim? Seems that it asserts verify runs while the test goroutine externally holds the lock, but the operational value of the PR is that two handler goroutines for the same MsgID can both be in verify simultaneously. A test that spawns two `handleSignedSSVMessage` goroutines (no external lock) and asserts both reach `VerifySignature` before either records state would be a stronger regression guard. Could be a follow-up alongside the F-ssv-173 items.

message/validation/partial_validation.go
```diff
@@ -43,26 +43,37 @@ func (mv *messageValidator) validatePartialSignatureMessage(
...
+		return mv.updatePartialSignatureState(partialSignatureMessages, receivedFrom, state, signer, committeeInfo)
+	}); err != nil {
 		return partialSignatureMessages, err
 	}
 
 	return partialSignatureMessages, nil
 }
 
+func (mv *messageValidator) verifyPartialSignatureMessageSignature(signedSSVMessage *spectypes.SignedSSVMessage) error {
+	signature := signedSSVMessage.Signatures[0]
```
momosh-ssv, line 66: Might be worth a defensive `len` guard or a short comment here? The `[0]` indexing is safe today because `validateSignedSSVMessage` (len≥1, equal arities) and `validatePartialSignatureMessageSemantics` (len==1) run before verify. A reorder in a future refactor would silently break it — a one-line `// semantics above guarantees len==1` makes the invariant explicit.

PR #2730, opened 2026-03-22
fix(p2p): replace Fatal calls with graceful error handling

olegshmuelov:
## Summary
- Replace `logger.Fatal` in `Start()` with error return — caller decides how to handle
- Replace `logger.Fatal` in `bootstrapDiscovery()` with `logger.Error` — goroutine logs instead of crashing
- Add `defer` state rollback in `Start()` — resets `stateReady` to `stateClosed` on failure so retries aren't masked
- Add `defer close(connector)` to `bootstrapDiscovery` — prevents consumer goroutine leak on both error and normal shutdown

## Finding
F-ssv-049 — `Fatal` calls in p2p network startup

## Test
No logic change. Error paths remain the same, just handled gracefully instead of crashing.

### DIFF
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -257,8 +257,7 @@ func (n *p2pNetwork) getConnector() (chan peer.AddrInfo, error) {
 // Start starts the discovery service, garbage collector (peer index), and reporting.
 func (n *p2pNetwork) Start() (err error) {
 	if atomic.SwapInt32(&n.state, stateReady) == stateReady {
-		// return errors.New("could not setup network: in ready state")
-		return nil
+		return fmt.Errorf("network already started")
 	}
 	defer func() {
 		if err != nil {
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -255,11 +255,16 @@ func (n *p2pNetwork) getConnector() (chan peer.AddrInfo, error) {
 }
 
 // Start starts the discovery service, garbage collector (peer index), and reporting.
-func (n *p2pNetwork) Start() error {
+func (n *p2pNetwork) Start() (err error) {
 	if atomic.SwapInt32(&n.state, stateReady) == stateReady {
 		// return errors.New("could not setup network: in ready state")
 		return nil
 	}
+	defer func() {
+		if err != nil {
+			atomic.StoreInt32(&n.state, stateClosed)
+		}
+	}()
 
 	pAddrs, err := peer.AddrInfoToP2pAddrs(&peer.AddrInfo{
 		ID:    n.host.ID(),
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -266,7 +266,7 @@ func (n *p2pNetwork) Start() error {
 		Addrs: n.host.Addrs(),
 	})
 	if err != nil {
-		n.logger.Fatal("could not get my address", zap.Error(err))
+		return fmt.Errorf("resolve p2p address: %w", err)
 	}
 	maStrs := make([]string, len(pAddrs))
 	for i, ima := range pAddrs {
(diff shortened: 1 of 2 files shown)


network/p2p/p2p.go
```diff
@@ -453,7 +459,8 @@ func (n *p2pNetwork) bootstrapDiscovery(connector chan peer.AddrInfo) {
 		})
 	}, 3)
 	if err != nil {
-		n.logger.Fatal("could not setup discovery", zap.Error(err))
+		n.logger.Error("could not setup discovery", zap.Error(err))
```
y0sher, line 461: this silent return might be an issue that makes us unaware that discovery isn't functioning well

y0sher (changes requested):
one comment

y0sher:
## Staging Test Report

| | |
|---|---|
| **Branch** | `fix/replace-fatal-p2p-startup` |
| **Nodes** | `ssv-node-89`, `ssv-node-90` (stage-hoodi) |
| **Status** | ✅ **PASS** |

### Results

- ✅ **Startup**: Node started successfully — critical since this PR changes the startup code path
- ✅ **Duties**: Attestations, aggregations, and validator registrations all completing successfully
- ✅ **Zero Fatal log entries**: Confirmed `Fatal` calls have been replaced with graceful error handling
- ✅ **P2P network operational**: Peer discovery working (4 peers connected), subnet scoring active

### Fix Verification

The PR changes `Start()` to return named error with deferred state rollback, replaces `logger.Fatal("could not setup discovery")` with `logger.Error` + return, and adds `defer close(connector)` to prevent goroutine leaks. All three changes function correctly — node starts, discovers peers, and processes duties without any Fatal-level log output.

---

network/p2p/p2p.go
olegshmuelov: Valid concern. `bootstrapDiscovery` runs as a goroutine (`go n.bootstrapDiscovery(...)` in `p2p_discovery.go:29`) so we can't propagate the error to the caller. `Fatal` kills the process without defers or cleanup - not great either.

`logger.Error` matches the existing pattern for goroutine errors in this package (see lines 189, 202, 388, 414, 423 - all use `logger.Error`). In production this path is effectively unreachable - `DiscV5Service.Bootstrap` blocks inside `discover()` and never returns an error.

That said, the broader issue of detecting degraded goroutines (discovery, peer trimming, etc.) through health checks could be valuable as a follow-up - but that's a larger effort beyond this finding.

nkryuchkov: It might be useful to propagate errors like this to the node health checker by passing some kind of health check reporter to the components that may report failures. That would be another major task, though

nkryuchkov (dismissed):
LGTM, but I'd create an issue for https://github.com/ssvlabs/ssv/pull/2730/changes#r2973870478 before merging this, unless there are any objections

network/p2p/p2p.go
olegshmuelov: Agreed, a health check reporter for goroutine failures would be the right long-term approach.
For this PR, `logger.Error` keeps it consistent with the rest of the p2p package.

network/p2p/p2p.go
```diff
@@ -255,18 +255,23 @@ func (n *p2pNetwork) getConnector() (chan peer.AddrInfo, error) {
 }
 
 // Start starts the discovery service, garbage collector (peer index), and reporting.
-func (n *p2pNetwork) Start() error {
+func (n *p2pNetwork) Start() (err error) {
 	if atomic.SwapInt32(&n.state, stateReady) == stateReady {
 		// return errors.New("could not setup network: in ready state")
 		return nil
```
iurii-ssv: What's this commented-out line ? Seems to me we should uncomment & use it (instead of just `return nil`) ?  We do something like that in `p2pNetwork.Setup` already btw

iurii-ssv (dismissed):
Left minor comment + we really should create a follow-up issue (as Nikita mentioned above)

iurii-ssv (approved):

network/p2p/p2p.go
y0sher: Ok I'm resolving it and we'll create another task to report goroutine health status. and maybe even have a restart mechanism?

y0sher (approved):

network/p2p/p2p.go
olegshmuelov: Created: #2738

Phase 1: add `Healthy()` to p2p and register with the existing `nodeprobe.Prober` — same pattern as CL/EL/EventSyncer. Track `discoveryFailed` state so the prober catches it on the next 60s cycle.

Phase 2: restart mechanism with backoff for transient failures.

PR #2731, opened 2026-03-22
fix(ttl): add context support to prevent goroutine leak

olegshmuelov:
## Summary
- Add `context.Context` to `ttl.New()` — cleanup goroutine exits when context is canceled
- Override `Delete` and `GetAndDelete` to also clean timestamp index — fixes pre-existing metadata leak
- Resolve both TODOs in the code (`ctx.Done()` support and Go version comment)
- Update all callers (2 production, 4 test) to pass context
- Add `TestCleanupStopsOnCancel` and `TestDeleteCleansTimestampIndex`

## Finding
F-ssv-036 — TTL Map goroutine leak, no cancellation mechanism

## Test
All `utils/ttl` tests pass. Two new tests: cancellation behavior and timestamp index cleanup.
Lint clean (`golangci-lint`).

### DIFF
--- a/utils/ttl/map.go
+++ b/utils/ttl/map.go
@@ -33,7 +33,6 @@ func New[Key comparable, Value any](ctx context.Context, lifespan, cleanupInterv
 			case <-ticker.C:
 				m.idxLastUpdatedAt.Range(func(key Key, t time.Time) bool {
 					if time.Since(t) > lifespan {
-						m.idxLastUpdatedAt.Delete(key)
 						m.Delete(key)
 					}
 					return true
--- a/utils/ttl/map.go
+++ b/utils/ttl/map.go
@@ -23,14 +23,11 @@ func New[Key comparable, Value any](ctx context.Context, lifespan, cleanupInterv
 		idxLastUpdatedAt: hashmap.Map[Key, time.Time]{},
 	}
 	go func() {
-		ticker := time.NewTicker(cleanupInterval)
-		defer ticker.Stop()
-
 		for {
 			select {
 			case <-ctx.Done():
 				return
-			case <-ticker.C:
+			case <-time.After(cleanupInterval):
 				m.idxLastUpdatedAt.Range(func(key Key, t time.Time) bool {
 					if time.Since(t) > lifespan {
 						m.Delete(key)
--- a/utils/ttl/map.go
+++ b/utils/ttl/map.go
@@ -1,36 +1,44 @@
 package ttl
 
 import (
+	"context"
 	"time"
 
 	"github.com/ssvlabs/ssv/utils/hashmap"
 )
 
-// Map implements a thread-safe map with a sync.Map under the hood.
+// Map implements a thread-safe map with automatic TTL-based expiry.
 type Map[Key comparable, Value any] struct {
 	*hashmap.Map[Key, Value]
 	idxLastUpdatedAt hashmap.Map[Key, time.Time]
 }
 
-func New[Key comparable, Value any](lifespan, cleanupInterval time.Duration) *Map[Key, Value] {
+// New creates a TTL map that automatically removes entries older than lifespan.
+// The cleanup goroutine runs every cleanupInterval and exits when ctx is canceled.
+// After ctx is canceled, the map remains readable but should not be used for new writes —
+// cleanup is permanently stopped and timestamp metadata from new writes will not be reclaimed.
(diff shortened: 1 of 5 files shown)


y0sher (dismissed):

y0sher:
## Staging Test Report

| | |
|---|---|
| **Branch** | `fix/ttl-map-goroutine-leak` |
| **Nodes** | `ssv-node-69`, `ssv-node-70` (stage-hoodi) |
| **Status** | ✅ **PASS** |

### Results

- ✅ **Startup**: Node started successfully (`SSV-Node:untagged-85f08cc...`)
- ✅ **Duties**: Attestations and sync committee processing normally
- ✅ **No goroutine errors**: Zero goroutine-related log entries found
- ✅ **No panics/crashes**: Only benign `validation ignored` p2p messages (common in staging)

### Fix Verification

The PR adds `context.Context` to `ttl.New()`, replacing the old design where cleanup goroutines ran indefinitely. The cleanup loop now selects on `ctx.Done()` to exit when the context is cancelled, preventing goroutine leaks. Also adds a proper `Delete` method that cleans up the timestamp index. Node ran stable with all duties processing normally.

---

utils/ttl/map.go
```diff
@@ -1,36 +1,43 @@
...
+// The cleanup goroutine runs every cleanupInterval and exits when ctx is canceled.
+// After ctx is canceled, the map remains readable but should not be used for new writes —
+// cleanup is permanently stopped and timestamp metadata from new writes will not be reclaimed.
+func New[Key comparable, Value any](ctx context.Context, lifespan, cleanupInterval time.Duration) *Map[Key, Value] {
 	m := &Map[Key, Value]{
 		Map:              hashmap.New[Key, Value](),
 		idxLastUpdatedAt: hashmap.Map[Key, time.Time]{},
 	}
 	go func() {
-		// TODO: use time.After when Go is updated to 1.23
```
nkryuchkov, line 21: Can we use `time.After` as mentioned in the comment?

nkryuchkov (dismissed):

utils/ttl/map.go
olegshmuelov: applied

utils/ttl/map.go
```diff
@@ -1,36 +1,40 @@
...
+				return
+			case <-time.After(cleanupInterval):
+				m.idxLastUpdatedAt.Range(func(key Key, t time.Time) bool {
+					if time.Since(t) > lifespan {
+						m.Delete(key)
+					}
+					return true
+				})
+			}
 		}
```
iurii-ssv, line 38: Previously this would also clean up the residual last-updated entry with `m.idxLastUpdatedAt.Delete(key)` ... not sure if it's "omitted as an optimization" - it doesn't seem too important anyway, but I would clean it up (for consistency)

utils/ttl/map_test.go
```diff
@@ -550,6 +551,63 @@ func TestIssue1682(t *testing.T) {
...
+	// Add new entries to verify cleanup still works.
+	m.Set(3, "c")
+	require.Eventually(t, func() bool {
+		return m.SlowLen() == 0
+	}, time.Second, 10*time.Millisecond, "new entries should also be cleaned up")
+
+	// Cancel and verify no panic — cleanup goroutine should exit cleanly.
+	cancel()
+	select {
+	case <-time.After(50 * time.Millisecond):
```
iurii-ssv: I would use `400 * time.Millisecond` (or higher) for sleep-type of waits in tests, I've encountered lots of flakiness due to shorter sleep-times we had in many unit-tests after we started to run them concurrently:
- since it can spawn tens (or even hundreds of tests at a time), some might be delayed by a lot longer than just 50ms
- since many tests are executing concurrently, sleeping for longer doesn't "cost us anything" in many cases (so, sleeping for many seconds likely won't affect the total execution time of the whole test-suite much)

iurii-ssv (dismissed):
LGTM, just couple suggestions

utils/ttl/map.go
olegshmuelov: Intentional - the `Delete` override (line 62) handles both maps:
```go
func (m *Map[Key, Value]) Delete(key Key) bool {
    m.idxLastUpdatedAt.Delete(key)
    return m.Map.Delete(key)
}
```
So `m.Delete(key)` in the cleanup loop already cleans `idxLastUpdatedAt`. Adding it back explicitly would be redundant.

nkryuchkov (approved):

iurii-ssv (approved):

y0sher (approved):

PR #2732, opened 2026-03-22
fix(topics): replace error-swallowing Get with read-only Lookup

olegshmuelov:
## Summary
- Remove `Get()` which silently discarded `Join()` errors (`topic, _ := tc.Join(name)`)
- Replace with read-only `Lookup()` that reads from the topic cache without calling `Join`
- `Get` called `Join` internally, which has a create side effect — inappropriate for read-only callers like `Peers` and `UpdateScoreParams`. During topic leave/rejoin, `Join` can race with libp2p's async close and fail with "topic already exists"
- `Lookup` avoids this entirely: read-only cache access with `RLock`, no side effects

## Finding
F-ssv-051 — Swallowed error in topic container `Get`

## Test
Build clean. Lint clean. `Peers` returns `nil, nil` on cache miss — same behavior as before, no caller changes needed.

### DIFF
--- a/network/topics/container.go
+++ b/network/topics/container.go
@@ -33,9 +33,11 @@ func newTopicsContainer(ps *pubsub.PubSub, onTopicJoined onTopicJoined) *topicsC
 	}
 }
 
-func (tc *topicsContainer) Get(name string) *pubsub.Topic {
-	topic, _ := tc.Join(name)
-	return topic
+// Lookup returns a cached topic by name, or nil if not found.
+func (tc *topicsContainer) Lookup(name string) *pubsub.Topic {
+	tc.lock.RLock()
+	defer tc.lock.RUnlock()
+	return tc.topics[name]
 }
 
 func (tc *topicsContainer) Leave(name string) error {
--- a/network/topics/controller.go
+++ b/network/topics/controller.go
@@ -115,9 +115,9 @@ func (ctrl *topicsCtrl) UpdateScoreParams() error {
 	var errs error
 	topics := ctrl.ps.GetTopics()
 	for _, topicName := range topics {
-		topic := ctrl.container.Get(topicName)
+		topic := ctrl.container.Lookup(topicName)
 		if topic == nil {
-			errs = errors.Join(errs, fmt.Errorf("topic %s is not ready; ", topicName))
+			errs = errors.Join(errs, fmt.Errorf("topic %s not found in cache", topicName))
 			continue
 		}
 		p := ctrl.scoreParamsFactory(topicName)
@@ -149,7 +149,7 @@ func (ctrl *topicsCtrl) Peers(name string) ([]peer.ID, error) {
 		return ctrl.ps.ListPeers(""), nil
 	}
 	name = commons.GetTopicFullName(name)
-	topic := ctrl.container.Get(name)
+	topic := ctrl.container.Lookup(name)
 	if topic == nil {
 		return nil, nil
 	}
--- a/network/topics/controller.go
+++ b/network/topics/controller.go
@@ -16,11 +16,6 @@ import (
 	"github.com/ssvlabs/ssv/protocol/v2/ssv/queue"
 )
 
-var (
-	// ErrTopicNotReady happens when trying to access a topic which is not ready yet
-	ErrTopicNotReady = errors.New("topic is not ready")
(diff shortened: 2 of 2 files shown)


y0sher:
## Staging Test Report

| | |
|---|---|
| **Branch** | `fix/topic-container-swallowed-error` |
| **Nodes** | `ssv-node-93`, `ssv-node-94` (stage-hoodi) |
| **Status** | ✅ **PASS** |

### Results

- ✅ **Startup**: Node started successfully (`SSV-Node:untagged-5d986be...`)
- ✅ **Duties**: Both operators successfully submitting attestations, sync committee, and validator registrations every slot
- ✅ **No panics/crashes**: Only benign `validation ignored` p2p messages
- ✅ **No swallowed errors**: Lookup() does not trigger Join() side effects

### Fix Verification

The old `Get()` method (which called `Join()` internally and discarded its error via `topic, _ := tc.Join(name)`) has been replaced with `Lookup()`, a read-only cache access using `RLock`. The callers (`Peers`, `UpdateScoreParams`) now get nil on cache miss without any hidden `Join()` calls that could race with libp2p's async close.

---

nkryuchkov (approved):

network/topics/controller.go
```diff
@@ -115,18 +110,18 @@ func (ctrl *topicsCtrl) UpdateScoreParams() error {
 	var errs error
 	topics := ctrl.ps.GetTopics()
 	for _, topicName := range topics {
-		topic := ctrl.container.Get(topicName)
+		topic := ctrl.container.Lookup(topicName)
```
momosh-ssv, line 113: What do you think about adding a unit test for this cache-miss path? `ps.GetTopics()` and `tc.topics` can momentarily diverge during leave/rejoin, so this branch is reachable in practice — a test would document that it's expected, not a bug.

network/topics/controller.go
```diff
@@ -149,7 +144,7 @@ func (ctrl *topicsCtrl) Peers(name string) ([]peer.ID, error) {
 		return ctrl.ps.ListPeers(""), nil
 	}
 	name = commons.GetTopicFullName(name)
-	topic := ctrl.container.Get(name)
+	topic := ctrl.container.Lookup(name)
 	if topic == nil {
 		return nil, nil
```
momosh-ssv, line 149: Should we return a sentinel error here instead of `nil, nil`? Right now callers can't distinguish "topic not joined" from "joined but zero peers." Not a regression (old `Get` had the same behavior), but with `Lookup` being explicitly read-only this seems like a good moment to clarify the contract.

network/topics/container.go
```diff
@@ -33,9 +33,11 @@ func newTopicsContainer(ps *pubsub.PubSub, onTopicJoined onTopicJoined) *topicsC
...
 }
 
-func (tc *topicsContainer) Get(name string) *pubsub.Topic {
-	topic, _ := tc.Join(name)
-	return topic
+// Lookup returns a cached topic by name, or nil if not found.
+func (tc *topicsContainer) Lookup(name string) *pubsub.Topic {
+	tc.lock.RLock()
+	defer tc.lock.RUnlock()
+	return tc.topics[name]
```
iurii-ssv, line 40: Nit: while at it, we could rename `lock` -> `topicsLock` (or similar) to clarify what this lock is responsible for (that would match `subLock` naming too)

network/topics/controller.go
iurii-ssv: Good point, however from what I see in the code - "there is no need to differentiate" for any of the callers, I would keep it as is until we get a use-case that needs/relies on such a sentinel error

iurii-ssv (approved):

network/topics/controller.go
olegshmuelov: The nil-on-miss is intentional and unchanged from old `Get`. Testing this path meaningfully requires mocking `ps.GetTopics()` to diverge from the container, better suited for an integration test than a unit test.

network/topics/container.go
olegshmuelov: `tc` is `topicsContainer` - `tc.lock` already implies it protects topics. Renaming to `tc.topicsLock` would be redundant.

momosh-ssv (approved):

PR #2734, opened 2026-03-23
fix(streams): preserve write error and handle short writes in WriteWithTimeout

olegshmuelov:
## Summary
- Preserve the original `Write` error on failure instead of discarding it with a generic "bytes don't match" message
- Add defensive `io.ErrShortWrite` fallback for contract-violating short writes with nil error
- Migrate `pkg/errors` to `fmt.Errorf` with `%w`
- Fix doc comment ("reads" → "writes")
- Add regression tests for all three error paths

## Finding
F-ssv-052 — `WriteWithTimeout` discards original write error on short write

## Test
New `TestWriteWithTimeout_ErrorPaths` covers: write error preserves root cause, short write returns `io.ErrShortWrite`, deadline failure is wrapped. All stream tests pass. Lint clean.

### DIFF
--- a/network/streams/stream_test.go
+++ b/network/streams/stream_test.go
@@ -2,6 +2,8 @@ package streams
 
 import (
 	"context"
+	"errors"
+	"io"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


nkryuchkov (approved):

network/streams/stream_test.go
```diff
@@ -14,6 +16,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type stubStream struct {
+	core.Stream
```
momosh-ssv, line 20: Non-blocking: `stubStream` embeds `core.Stream` as a nil interface, so any method besides `Write`/`SetWriteDeadline` will panic with a nil dereference instead of a clear test failure. Not a problem today since `WriteWithTimeout` only calls those two, but it makes the stub fragile against future changes to the function under test.

network/streams/stream_test.go
```diff
@@ -14,6 +16,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type stubStream struct {
```
momosh-ssv, line 19: Non-blocking: `controller.go` in this same package still uses `pkg/errors`. Might be worth a follow-up to migrate the rest of the package for consistency, since `stream.go` now uses `fmt.Errorf` exclusively.

momosh-ssv (commented):
Looks good overall. Two non-blocking observations below.

momosh-ssv (approved):

y0sher (approved):

PR #2735, opened 2026-03-23
network/streams: cap stream reads to prevent OOM

nkryuchkov:
## Summary

Addresses `F-ssv-122`.

`ReadWithTimeout` used `io.ReadAll` directly on libp2p streams, which allowed a remote peer to send an arbitrarily large payload and force unbounded memory allocation. This affected stream-based protocols such as the node-info handshake.

This change adds a hard `1 MiB` limit to stream reads in `network/streams` and returns an explicit error when the payload exceeds that limit.

## Changes

- add `maxStreamMessageSize` in `network/streams/stream.go`
- read with `io.LimitReader(..., maxStreamMessageSize+1)` to detect oversize payloads
- reject oversized stream messages with `errStreamMessageTooLarge`
- add a regression test covering oversized payload rejection

## Why `+1`

Using `io.LimitReader(..., maxStreamMessageSize)` alone could silently truncate oversized payloads. Reading one byte past the limit lets us detect overflow and reject the message instead of processing partial input.


nkryuchkov:

network/streams/stream.go
```diff
@@ -38,7 +42,19 @@ func (ts *streamWrapper) ReadWithTimeout(timeout time.Duration) ([]byte, error)
...
+
+	// Read one byte past the limit so oversized payloads are rejected instead of silently truncated.
+	data, err := io.ReadAll(io.LimitReader(ts.Stream, maxStreamMessageSize+1))
+	if err != nil {
+		return nil, err
+	}
+	if len(data) > maxStreamMessageSize {
+		// Best-effort abort: the payload is invalid, and the oversize error below is the one callers should handle.
+		_ = ts.Reset()
+		return nil, ErrStreamMessageTooLarge
```
momosh-ssv, line 58: Should we add some observability here? When a peer sends an oversized payload there's no metric or structured log to surface it. Since `ReadWithTimeout` doesn't have logger/peer context, the callers in `controller.go` could check `errors.Is(err, ErrStreamMessageTooLarge)` and emit a warn-level log with the peer ID and a counter metric. Without that, you won't know when this defense triggers in production.

network/streams/stream.go
```diff
@@ -38,7 +42,19 @@ func (ts *streamWrapper) ReadWithTimeout(timeout time.Duration) ([]byte, error)
...
-	return io.ReadAll(ts.Stream)
+
+	// Read one byte past the limit so oversized payloads are rejected instead of silently truncated.
+	data, err := io.ReadAll(io.LimitReader(ts.Stream, maxStreamMessageSize+1))
+	if err != nil {
+		return nil, err
+	}
+	if len(data) > maxStreamMessageSize {
+		// Best-effort abort: the payload is invalid, and the oversize error below is the one callers should handle.
+		_ = ts.Reset()
```
momosh-ssv, line 57: What do you think about adding a short comment here noting that callers are expected to suppress `ErrReset` on their deferred `Close()`? The contract works correctly today (the `!errors.Is(err, libp2pnetwork.ErrReset)` guards in controller.go handle it), but the coupling is implicit and a future reader might remove those guards without realizing `Reset()` is called here.

momosh-ssv (commented):
Solid fix. Two suggestions below, neither is a blocker.

iurii-ssv (dismissed):
Looks good, I'd throw in [this minor refactoring](https://github.com/ssvlabs/ssv/pull/2751) to standardize `Request` and `HandleStream` funcs a bit

network/streams/stream.go
```diff
@@ -1,13 +1,18 @@
...
 import (
+	"errors"
 	"fmt"
 	"io"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core"
 )
 
+const maxStreamMessageSize = 1 << 20 // 1 MiB
```
iurii-ssv, line 14: I don't know if 1MiB is large enough to not fail any legitimate reads ... perhaps using something like 16(or 64) MiB would still achieve the goal while provide us a higher error-margin. 

If you did some research into "what's the reasonable limit" we could leave a rationale-describing comment here.

network/streams/stream_test.go
```diff
@@ -89,6 +90,33 @@ func TestStream(t *testing.T) {
 	wg.Wait()
 }
 
+func TestStreamRejectsOversizedPayload(t *testing.T) {
```
iurii-ssv, line 93: Could maybe add a test to cover the boundary `case len(payload) == maxStreamMessageSize` is still accepted.

momosh-ssv (dismissed):
seems good now, gj!

network/streams/stream.go
```diff
@@ -1,13 +1,18 @@
...
+
+var ErrStreamMessageTooLarge = errors.New("stream message exceeds max size")
```
oleg-ssvlabs, line 16: there is no need to export this error since it's used only within this pacakge

oleg-ssvlabs (commented):
gj. Left 1 minor comment

network/streams/stream.go
nkryuchkov: @greptileai [asked to export it](https://github.com/ssvlabs/ssv/pull/2735#discussion_r2974637387)

oleg-ssvlabs: I see. I don't have strong opinion here, having it exported does not harm, even though does not provide any value today either (outside packages do not check the error type via `errors.Is`). Resolving

oleg-ssvlabs (dismissed):

network/streams/stream.go
iurii-ssv: I think both `ErrStreamMessageTooLarge` and `Stream interface` need to be of the same visibility-scope (both public, or both private)

iurii-ssv (approved):

momosh-ssv (approved):

PR #2739, opened 2026-03-24
fix: replace panic with error

alok-ssv:
Addresses `F-ssv-045`

### DIFF
--- a/protocol/v2/ssv/runner/runner_state.go
+++ b/protocol/v2/ssv/runner/runner_state.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"crypto/sha256"
 	"encoding/json"
+	"fmt"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/pkg/errors"
@@ -128,7 +129,7 @@ func (pcs *State) UnmarshalJSON(data []byte) error {
 	} else if aux.CommitteeDuty != nil {
 		pcs.CurrentDuty = aux.CommitteeDuty
 	} else {
-		panic("no starting duty")
+		return fmt.Errorf("no starting duty in state JSON")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


protocol/v2/ssv/runner/runner_state.go
```diff
@@ -128,7 +129,7 @@ func (pcs *State) UnmarshalJSON(data []byte) error {
 	} else if aux.CommitteeDuty != nil {
 		pcs.CurrentDuty = aux.CommitteeDuty
 	} else {
-		panic("no starting duty")
+		return fmt.Errorf("no starting duty in state JSON")
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai `github.com/pkg/errors` is discontinued, and we should not use it for any new code. So let's not do this

nkryuchkov, line 132: We can probably get rid of the `else` clause:

```go
	if aux.ValidatorDuty != nil {
		pcs.CurrentDuty = aux.ValidatorDuty
	} else if aux.CommitteeDuty != nil {
		pcs.CurrentDuty = aux.CommitteeDuty
	}

	return fmt.Errorf("no starting duty in state JSON")
```

But I don't mind keeping it as it might be a bit out of the scope of the PR

nkryuchkov (dismissed):

protocol/v2/ssv/runner/runner_state.go
alok-ssv: I dont think we can. We will need to return in all the above if cases.

nkryuchkov: Yeah, sorry, I missed moving `return nil` into both cases while I was writing that. Never mind, let's keep it as-is

iurii-ssv (dismissed):

nkryuchkov (approved):

iurii-ssv (approved):

oleg-ssvlabs (approved):
lgtm

ljuba-ssv (approved):

PR #2741, opened 2026-03-24
protocol/v2/qbft/instance: add missing unit tests

nkryuchkov:
## Summary

Adds missing unit test coverage for QBFT instance logic in `protocol/v2/qbft/instance`.

Covers:
- `T-ssv-001` Proposal processing
- `T-ssv-002` Prepare processing
- `T-ssv-003` Commit processing
- `T-ssv-004` Round change processing
- `T-ssv-005` Timeout handling
- `T-ssv-006` `ProcessMsg` dispatch and `BaseMsgValidation`

## What’s Included

Added focused test files for:
- proposal handling and justification validation
- prepare quorum/validation paths
- commit quorum, aggregation, and validation paths
- round change orchestration and validation paths
- timeout handling, including deferred round bump behavior
- `ProcessMsg` dispatch, past-round handling, and concurrent access behavior

Also added shared test helpers for building signed QBFT messages and asserting timer/network behavior.

## Key Behaviors Covered

- future-round proposal bump and prepare broadcast
- wrong leader, wrong height, signer, root, and justification rejection
- prepare quorum transition, early return when quorum already existed, and round-change justification extraction
- commit quorum detection and `aggregateCommitMsgs` operator/signature ordering
- partial quorum round change flow, justified leader proposal flow, `highestPrepared`, `CreateRoundChange`, and round-change validation branches
- timeout behavior where `CreateRoundChange` runs against old round state and the round bump happens after broadcast
- `BaseMsgValidation` past-round behavior and `ProcessMsg` state update/dispatch behavior

## Notes

- The exact `newRound <= prevRound` guard in `round_change.go` appears structurally unreachable through normal inputs because `hasReceivedPartialQuorum()` only returns rounds strictly greater than the current instance round. The surrounding partial-quorum behavior is covered.
- For `T-ssv-006`, the code currently bypasses the generic past-round rejection for quorum commit messages, but still applies commit-specific validation afterward. The tests reflect the implemented behavior.

### DIFF
--- a/protocol/v2/qbft/instance/process_msg_test.go
+++ b/protocol/v2/qbft/instance/process_msg_test.go
@@ -26,6 +26,7 @@ func TestBaseMsgValidationPastRoundDecidedCommitBypassesPastRoundGuard(t *testin
 	err := env.inst.BaseMsgValidation(msg)
 	require.Error(t, err)
 	require.ErrorContains(t, err, "msg allows 1 signer")
+	require.NotContains(t, err.Error(), "past round")
 }
 
 func TestBaseMsgValidationPastRoundNonDecidedRejected(t *testing.T) {
--- a/protocol/v2/qbft/instance/process_msg_test.go
+++ b/protocol/v2/qbft/instance/process_msg_test.go
@@ -88,6 +88,8 @@ func TestProcessMsgConcurrentAccess(t *testing.T) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+			// Intentionally reuse the same message pointer to exercise dedup/idempotent
+			// processing while ProcessMsg serializes handler execution through processMsgF.
 			_, _, _, err := env.inst.ProcessMsg(context.Background(), zap.NewNop(), msg)
 			errs <- err
 		}()
--- a/protocol/v2/qbft/instance/process_msg_test.go
+++ b/protocol/v2/qbft/instance/process_msg_test.go
@@ -1,0 +1,110 @@
+package instance
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	specqbft "github.com/ssvlabs/ssv-spec/qbft"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestBaseMsgValidationPastRoundDecidedCommitBypassesPastRoundGuard(t *testing.T) {
+	env := newInstanceTestEnv(t, 4)
+	fullData := []byte("commit-value")
+	root := env.hash(fullData)
+	env.inst.State.Round = 2
+	env.inst.State.ProposalAcceptedForCurrentRound = env.proposal(2, 1, fullData, root, nil, nil)
+
+	msg := env.aggregateMessages(
+		env.commit(1, 1, root),
+		env.commit(1, 2, root),
+		env.commit(1, 3, root),
+	)
+
+	err := env.inst.BaseMsgValidation(msg)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "msg allows 1 signer")
+}
+
+func TestBaseMsgValidationPastRoundNonDecidedRejected(t *testing.T) {
+	env := newInstanceTestEnv(t, 2)
+	root := env.hash([]byte("proposal-value"))
+	env.inst.State.Round = 2
+	env.inst.State.ProposalAcceptedForCurrentRound = env.proposal(2, 1, []byte("proposal-value"), root, nil, nil)
+
+	err := env.inst.BaseMsgValidation(env.prepare(1, 1, root))
+	require.ErrorContains(t, err, "past round")
+
+	err = env.inst.BaseMsgValidation(env.commit(1, 1, root))
+	require.ErrorContains(t, err, "past round")
+}
+
+func TestProcessMsgDispatchCommitUpdatesDecision(t *testing.T) {
+	env := newInstanceTestEnv(t, 4)
+	env.setLeader(1)
+
+	fullData := []byte("commit-value")
+	root := env.hash(fullData)
+	env.inst.State.ProposalAcceptedForCurrentRound = env.proposal(1, 1, fullData, root, nil, nil)
+	env.addMessages(
+		env.inst.State.CommitContainer,
+		env.commit(1, 1, root),
+		env.commit(1, 2, root),
+	)
+
+	decided, decidedValue, aggregated, err := env.inst.ProcessMsg(
+		context.Background(),
+		zap.NewNop(),
+		env.commit(1, 3, root),
+	)
+	require.NoError(t, err)
+	require.True(t, decided)
+	require.Equal(t, fullData, decidedValue)
+	require.NotNil(t, aggregated)
+	require.True(t, env.inst.State.Decided)
+	require.Equal(t, fullData, env.inst.State.DecidedValue)
+}
+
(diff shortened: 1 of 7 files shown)


nkryuchkov:

protocol/v2/qbft/instance/process_msg_test.go
```diff
@@ -0,0 +1,110 @@
...
+	errs := make(chan error, workers)
+	var wg sync.WaitGroup
+	for i := 0; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_, _, _, err := env.inst.ProcessMsg(context.Background(), zap.NewNop(), msg)
+			errs <- err
+		}()
+	}
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai Kept this as-is intentionally. The goal of this test is to exercise concurrent calls into `ProcessMsg` with the same `*ProcessingMessage` instance and verify that `processMsgF` serializes handling and preserves idempotent/deduplicated behavior. Giving each goroutine its own copy would be a slightly different test. I added a comment in the test to make that intent explicit.

nkryuchkov:

protocol/v2/qbft/instance/proposal_test.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+func TestUponProposalFutureRoundBumpsAndBroadcastsPrepare(t *testing.T) {
+	env := newInstanceTestEnv(t, 2)
+	env.setLeader(1)
+
+	fullData := []byte("proposal-value")
+	proposal := env.proposal(2, 1, fullData, env.hash(fullData), nil, nil)
+
+	err := env.inst.uponProposal(context.Background(), zap.NewNop(), proposal)
```
iurii-ssv: Nit: probably can use `t.Context()` in these tests (unless such test spawns a go-routine - we can't use `t` from inside of those)

iurii-ssv (dismissed):
Looks good!

protocol/v2/qbft/instance/timeout_test.go
```diff
@@ -0,0 +1,69 @@
...
+	env := newInstanceTestEnv(t, 2)
+	env.inst.ForceStop()
+
+	err := env.inst.UponRoundTimeout(context.Background(), zap.NewNop())
+	require.ErrorContains(t, err, "instance stopped processing timeouts")
+}
+
+func TestUponRoundTimeoutAtCutOffRound(t *testing.T) {
+	env := newInstanceTestEnv(t, 2)
+	env.config.CutOffRound = env.inst.State.Round
```
momosh-ssv: Should we reconsider what this test actually covers? Setting `CutOffRound = Round` means `Round < Round` is already false, so `CanProcessMessages()` returns false before `UponRoundTimeout` runs any timeout logic. This makes it functionally identical to `TestUponRoundTimeoutForceStoppedInstance` — both hit the same `!CanProcessMessages()` early return.

The actual cut-off boundary (where a timeout increments the round past the cut-off) is never exercised. A test with `CutOffRound = Round + 1` that lets the timeout fire and then asserts the next call is rejected would cover that edge.

protocol/v2/qbft/instance/process_msg_test.go
```diff
@@ -0,0 +1,112 @@
...
+	env.inst.State.ProposalAcceptedForCurrentRound = env.proposal(2, 1, fullData, root, nil, nil)
+
+	msg := env.aggregateMessages(
+		env.commit(1, 1, root),
+		env.commit(1, 2, root),
+		env.commit(1, 3, root),
+	)
+
+	err := env.inst.BaseMsgValidation(msg)
+	require.Error(t, err)
```
momosh-ssv, line 27: What do you think about adding `require.NotContains(t, err.Error(), "past round")` here? The test proves the bypass exists (the error is `"msg allows 1 signer"` from `validateCommit`, not `"past round"`), but that intent is implicit. Without the negative assertion, this reads like a plain `validateCommit` multi-signer rejection test.

protocol/v2/qbft/instance/round_change_test.go
```diff
@@ -0,0 +1,380 @@
...
+func TestUponRoundChangeReturnsEarlyWhenQuorumAlreadyExists(t *testing.T) {
+	env := newInstanceTestEnv(t, 2)
+	env.addMessages(
+		env.inst.State.RoundChangeContainer,
+		env.roundChange(2, 1, specqbft.NoRound, [32]byte{}, nil, nil),
+		env.roundChange(2, 3, specqbft.NoRound, [32]byte{}, nil, nil),
+		env.roundChange(2, 4, specqbft.NoRound, [32]byte{}, nil, nil),
+	)
+	err := env.inst.uponRoundChange(context.Background(), zap.NewNop(), env.roundChange(2, 2, specqbft.NoRound, [32]byte{}, nil, nil))
+	require.NoError(t, err)
```
momosh-ssv, line 200: Should we also assert `env.inst.State.Round` is unchanged here? The early-return claim is currently proven only by the absence of broadcast. Adding a round check would make the test resilient to future refactors that might broadcast on a different path.

protocol/v2/qbft/instance/round_change_test.go
```diff
@@ -0,0 +1,380 @@
...
+		env.roundChange(2, 3, specqbft.NoRound, [32]byte{}, nil, nil),
+	)
+	require.NoError(t, err)
+
+	require.Equal(t, specqbft.Round(2), env.inst.State.Round)
+	require.Nil(t, env.inst.State.ProposalAcceptedForCurrentRound)
+	require.Equal(t, 1, env.timer.State.Timeouts)
+	require.Equal(t, specqbft.Round(2), env.timer.State.Round)
+
+	msg := env.broadcastedProcessingMessage(0)
```
momosh-ssv, line 172: What do you think about asserting `msg.QBFTMessage.DataRound` on the broadcast message? Since `LastPreparedRound`/`LastPreparedValue` are not set, this is the unprepared path — a `require.Equal(t, specqbft.NoRound, msg.QBFTMessage.DataRound)` would make that intent explicit and catch regressions if the default changes.

protocol/v2/qbft/instance/round_change_test.go
```diff
@@ -0,0 +1,380 @@
...
+	env := newInstanceTestEnv(t, 1)
+	env.setLeader(1)
+	env.inst.State.Round = 2
+	value := []byte("proposal-value")
+	roundChanges := []*specqbft.ProcessingMessage{
+		env.roundChange(2, 2, specqbft.NoRound, [32]byte{}, nil, nil),
+		env.roundChange(2, 3, specqbft.NoRound, [32]byte{}, nil, nil),
+		env.roundChange(2, 4, specqbft.NoRound, [32]byte{}, nil, nil),
+	}
+	roundChangeMsg := roundChanges[0]
```
momosh-ssv: Should we create a fresh `env` per subtest here? The subtests share the outer `env` and mutate `ProposalAcceptedForCurrentRound` sequentially — the second subtest resets it to `nil`. This works because subtests run in order, but if they are ever parallelized with `t.Parallel()` it would break silently.

momosh-ssv (commented):
A few suggestions to possibly tighten assertions and close a coverage gap, maybe?

y0sher (dismissed):

ljuba-ssv (approved):

iurii-ssv (approved):

PR #2742, opened 2026-03-24
event_parser: fix misleading error message

julienh-ssv:
F-ssv-047


momosh-ssv (approved):
LGTM — the old error wrapped a guaranteed-nil `err` from the earlier `Unpack` call, producing a misleading message. The new format string with `%T` correctly surfaces the actual unexpected type.

nkryuchkov (approved):

PR #2743, opened 2026-03-24
cli/operator: add configurable SSV API address warning

nkryuchkov:
## Summary

Adds a configurable listen address for the SSV HTTP API.

Until now, enabling `SSVAPIPort` always started the API on `:<port>`, which listens on all interfaces. This change adds `SSVAPIAddress` so operators can explicitly bind the API to `127.0.0.1` or another address without changing the port behavior.

To avoid breaking existing deployments, the default remains unchanged:
- if `SSVAPIAddress` is unset, the API still listens on all interfaces
- when that happens, the node now logs a warning recommending `127.0.0.1`

## Changes

- Added `SSVAPIAddress` config/env support in `cli/operator`
- Wired API server startup to use `SSVAPIAddress + SSVAPIPort`
- Added a warning when the API is enabled and `SSVAPIAddress` is not set
- Updated example configs and exporter release notes to recommend `SSVAPIAddress: 127.0.0.1`

## Compatibility

This is intentionally non-breaking.
Existing setups that only configure `SSVAPIPort` continue to behave the same way.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -9,9 +9,11 @@ import (
 	"fmt"
 	"log"
 	"math/big"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -115,6 +117,7 @@ type config struct {
 	NetworkPrivateKey            string                  `yaml:"NetworkPrivateKey" env:"NETWORK_PRIVATE_KEY" env-description:"Private key for P2P network identity"`
 	WsAPIPort                    int                     `yaml:"WebSocketAPIPort" env:"WS_API_PORT" env-description:"Port for WebSocket API server"`
 	WithPing                     bool                    `yaml:"WithPing" env:"WITH_PING" env-description:"Enable WebSocket ping messages"`
+	SSVAPIAddress                string                  `yaml:"SSVAPIAddress" env:"SSV_API_ADDRESS" env-description:"Listen address for SSV API server. Leave empty to listen on all interfaces; use 127.0.0.1 to keep it local-only"`
 	SSVAPIPort                   int                     `yaml:"SSVAPIPort" env:"SSV_API_PORT" env-description:"Port for SSV API server"`
 	LocalEventsPath              string                  `yaml:"LocalEventsPath" env:"EVENTS_PATH" env-description:"Path to local events file"`
 	EnableDoppelgangerProtection bool                    `yaml:"EnableDoppelgangerProtection" env:"ENABLE_DOPPELGANGER_PROTECTION" env-description:"Enable doppelganger protection for validators"`
@@ -655,9 +658,16 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		if cfg.SSVAPIPort > 0 {
+			if cfg.SSVAPIAddress == "" {
+				logger.Warn("SSV API address not configured; listening on all interfaces",
+					zap.Int("port", cfg.SSVAPIPort),
+					zap.String("config_key", "SSVAPIAddress"),
+					zap.String("recommended_address", "127.0.0.1"),
+				)
+			}
 			apiServer := apiserver.New(
 				logger,
-				fmt.Sprintf(":%d", cfg.SSVAPIPort),
+				net.JoinHostPort(cfg.SSVAPIAddress, strconv.Itoa(cfg.SSVAPIPort)),
 				hnode.NewNode(
 					// TODO: replace with narrower interface! (instead of accessing the entire PeersIndex)
 					[]string{fmt.Sprintf("tcp://%s:%d", cfg.P2pNetworkConfig.HostAddress, cfg.P2pNetworkConfig.TCPPort), fmt.Sprintf("udp://%s:%d", cfg.P2pNetworkConfig.HostAddress, cfg.P2pNetworkConfig.UDPPort)},
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -658,13 +658,7 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		if cfg.SSVAPIPort > 0 {
-			if cfg.SSVAPIAddress == "" {
-				logger.Warn("SSV API address not configured; listening on all interfaces",
-					zap.Int("port", cfg.SSVAPIPort),
-					zap.String("config_key", "SSVAPIAddress"),
-					zap.String("recommended_address", "127.0.0.1"),
-				)
-			}
+			warnIfSSVAPIAddressUnset(logger, cfg.SSVAPIAddress, cfg.SSVAPIPort)
 			apiServer := apiserver.New(
 				logger,
 				net.JoinHostPort(cfg.SSVAPIAddress, strconv.Itoa(cfg.SSVAPIPort)),
@@ -696,6 +690,18 @@ var StartNodeCmd = &cobra.Command{
 			logger.Fatal("failed to start SSV node", zap.Error(err))
 		}
 	},
+}
+
+func warnIfSSVAPIAddressUnset(logger *zap.Logger, address string, port int) {
+	if address != "" {
+		return
+	}
+
+	logger.Warn("SSV API address not configured; listening on all interfaces",
+		zap.Int("port", port),
+		zap.String("config_key", "SSVAPIAddress"),
+		zap.String("recommended_address", "127.0.0.1"),
+	)
 }
 
 func ensureNoMissingKeys(
--- a/config/config.example.yaml
+++ b/config/config.example.yaml
@@ -54,7 +54,7 @@ MetricsAPIPort: 15000
 EnableTraces: false
 
 # These settings enable the SSV API at the specified address+port. Refer to the documentation at https://bloxapp.github.io/ssv/
-# Prefer 127.0.0.1 address and use 0.0.0.0 only when remote access is intentional and protected by network controls (to prevent potential resource-intensive attacks).
-# SSVAPIAddress defaults to 0.0.0.0 if not specified (for backward compatibility).
+# Prefer 127.0.0.1 and listen on all interfaces only when remote access is intentional and protected by network controls (to prevent potential resource-intensive attacks).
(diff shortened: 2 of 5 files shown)


docs/release-notes_exporter_v2.0.0-rc.1.md
```diff
@@ -114,6 +116,8 @@ WithPing: true
 
 Please refer to the new [OpenApi spec](https://github.com/ssvlabs/ssv/tree/stage/docs/api) (`docs/api/ssvnode.openapi.yaml|json`) for details. Below is a high-level overview of the new HTTP API endpoints.
 
+For operator deployments, prefer binding the HTTP API to `127.0.0.1` and only expose `0.0.0.0` when remote access is intentional and protected by network controls such as a private network, firewall, or reverse proxy.
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai The config key is shown in the configuration examples, while the added paragraph is API exposure guidance, so the HTTP API section is an appropriate place for it

config/config.example.yaml
```diff
@@ -54,5 +54,7 @@ MetricsAPIPort: 15000
 EnableTraces: false
 
 # This enables the SSV API at the specified port. Refer to the documentation at https://bloxapp.github.io/ssv/
+# Bind it to 127.0.0.1 by default and expose 0.0.0.0 only when remote access is intentional and protected by network controls.
```
iurii-ssv: Nit: could adjust this phrasing to something like:
```
# These settings enable the SSV API at the specified address+port. Refer to the documentation at https://bloxapp.github.io/ssv/
# Prefer 127.0.0.1 address and use 0.0.0.0 only when remote access is intentional and protected by network controls (to prevent potential resource-intensive attacks). 
# SSVAPIAddress defaults to 0.0.0.0 if not specified (for backward compatibility).
```

Similar applies to the other config file(s).

iurii-ssv (dismissed):
Cool!

y0sher (dismissed):
we need to make sure to mention this API and env change in the release.

config/config.example.yaml
y0sher: Adopted your suggestion in ffebd803d — updated both config files.

iurii-ssv (dismissed):

cli/operator/node.go
```diff
@@ -655,9 +658,16 @@ var StartNodeCmd = &cobra.Command{
...
+				logger.Warn("SSV API address not configured; listening on all interfaces",
+					zap.Int("port", cfg.SSVAPIPort),
+					zap.String("config_key", "SSVAPIAddress"),
+					zap.String("recommended_address", "127.0.0.1"),
+				)
+			}
 			apiServer := apiserver.New(
 				logger,
-				fmt.Sprintf(":%d", cfg.SSVAPIPort),
+				net.JoinHostPort(cfg.SSVAPIAddress, strconv.Itoa(cfg.SSVAPIPort)),
```
momosh-ssv, line 664: What do you think about adding a `net.ParseIP` check on `cfg.SSVAPIAddress` before this line? Right now a typo like `"not-an-ip"` or an accidental `"127.0.0.1:8080"` passes through `JoinHostPort` silently — the error only surfaces inside the goroutine at `ListenAndServe` time (and the warning is suppressed since the string is non-empty). An early validation with a clear fatal message would make misconfiguration much easier to diagnose.

cli/operator/node.go
```diff
@@ -655,9 +658,16 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		if cfg.SSVAPIPort > 0 {
+			if cfg.SSVAPIAddress == "" {
+				logger.Warn("SSV API address not configured; listening on all interfaces",
+					zap.Int("port", cfg.SSVAPIPort),
```
momosh-ssv: Worth considering adding a few tests for this code path — e.g. warning fires when `SSVAPIAddress` is empty, warning is suppressed when set, and `JoinHostPort` produces the expected listen string. `zaptest/observer` is already available in the test file, so it should be straightforward.

config/config.example.yaml
```diff
@@ -53,6 +53,8 @@ MetricsAPIPort: 15000
 # This enables Open Telemetry traces
 EnableTraces: false
 
-# This enables the SSV API at the specified port. Refer to the documentation at https://bloxapp.github.io/ssv/
-# It's recommended to keep this port private to prevent potential resource-intensive attacks.
-# SSVAPIPort: 16000
\ No newline at end of file
+# These settings enable the SSV API at the specified address+port. Refer to the documentation at https://bloxapp.github.io/ssv/
+# Prefer 127.0.0.1 address and use 0.0.0.0 only when remote access is intentional and protected by network controls (to prevent potential resource-intensive attacks).
+# SSVAPIAddress defaults to 0.0.0.0 if not specified (for backward compatibility).
```
momosh-ssv: Might be worth adjusting this to something like "defaults to listening on all interfaces" rather than "defaults to 0.0.0.0". The actual default is an empty string, and `net.JoinHostPort("", port)` produces `:port` — which Go interprets as all interfaces. They behave the same on most systems, but on dual-stack hosts the distinction can matter, and the current wording implies the field is literally set to `0.0.0.0`.

docs/release-notes_exporter_v2.0.0-rc.1.md
```diff
@@ -71,6 +71,7 @@ exporter:
   Enabled: true
   Mode: archive
 
+SSVAPIAddress: 127.0.0.1
```
momosh-ssv, line 74: Should we consider adding an inline comment like `# optional, omit to bind all interfaces` here? The snippet reads as a drop-in config block, and operators copying it verbatim will get `127.0.0.1` binding — which is fine — but those upgrading existing configs who skip this line will get the new warning. A small hint would reduce surprise.

cli/operator/node.go
nkryuchkov: `net.ParseIP` would forbid network names like `localhost`

docs/release-notes_exporter_v2.0.0-rc.1.md
nkryuchkov: We want to make `127.0.0.1` the new default value without changing the default value to avoid breaking changes. So the ones who create a new config will use `127.0.0.1`, but the ones with already existing config will continue using `0.0.0.0` without any breaking changes, and they will get a warning that encourages them to use `127.0.0.1` instead.

iurii-ssv (approved):

olegshmuelov (approved):

ljuba-ssv (approved):

momosh-ssv (approved):

Roy-blox:
## QA — verified on stage-hoodi

Tested the `SSVAPIAddress` config end-to-end on v2.4.3 stage with op 76 as the sacrificial sensor. Three-state roundtrip: default → `127.0.0.1` → default.

### Test setup

- **Environment:** stage-hoodi
- **Sacrificial op:** 76 (port 16000)
- **Phases:**
  - A: log scan across 12 default-config v2.4.3 ops for the boot-time warning
  - B: redeploy op 76 with `SSV_API_ADDRESS=127.0.0.1`, verify warning absent + bind changed + node still healthy
  - C: restore default config, confirm warning returns

### Results

| Phase | Check | Expected | Observed | Verdict |
|---|---|---|---|---|
| A | Warning fires on default-config ops (12 sampled) | ≥1 hit per op | 6/12 hit; 6/12 absent but explained (4 stable-pod, 2 on pre-#2743 branch) | **PASS** |
| B.1 | Warning absent when `SSV_API_ADDRESS=127.0.0.1` | 0 hits | 0 hits in 10m window post-restart | **PASS** |
| B.2 | External connect to op 76 API port refused | refused / timeout | **INCONCLUSIVE-harness** — vnet has no path to op 76:16000 (baseline already unreachable). Compensated by positive bind-log evidence below. | **PASS** (alternate evidence) |
| B.3 | Op 76 still healthy after redeploy | 0 panics, ≥1 attestation | 0 panics, 363 attestations in 10m | **PASS** |
| C | Default config restored, warning returns | ≥1 hit | warning fired at restart; `addr=:16000` bind restored | **PASS** |

### Direct evidence chain

```
# Default config (no env-vars)
WARN  SSV API address not configured; listening on all interfaces
      port=16000 config_key=SSVAPIAddress recommended_address=127.0.0.1
INFO  Serving SSV API addr=:16000          ← all interfaces

# SSV_API_ADDRESS=127.0.0.1 (env-var override)
INFO  Serving SSV API addr=127.0.0.1:16000  ← localhost only
(no warning emitted)

# Default config restored
WARN  SSV API address not configured; listening on all interfaces ...
INFO  Serving SSV API addr=:16000          ← back to all interfaces
```

The three-state roundtrip cleanly demonstrates the config path is end-to-end functional: warning fires only when address is unset, bind-address log changes accordingly, node restarts pick up the new config without issue.

### Caveats

1. **B.2 external-reachability probe was harness-limited.** vnet doesn't expose per-SSV-node API ports (only the exporter at port 31601 has a vnet-routable hostname). Both the public IP and the pod IP for op 76:16000 were unreachable from the test host even before the config change. Go's stdlib socket bind semantics are OS-enforced, so the `addr=127.0.0.1:16000` log is reliable circumstantial evidence — but a true end-to-end probe (e.g., `kubectl exec` from another op's pod to op 76's pod IP) would close the loop definitively. Worth doing if there's pod-shell access.
2. **Phase A oracle was overstrict.** "≥1 hit per op in the last 24h" misses long-running stable pods. Of the 6 zero-hit ops, 4 hadn't restarted in 14+ days and 2 were on a pre-#2743 branch — neither is a logic break. Relaxed oracle: "≥1 hit per op in its most-recent restart, or pod uptime > Loki retention."

### Verdict

**PR #2743 is `PASS-hardening` on all available evidence.** The `SSVAPIAddress` config is correctly read, applied to the listener bind, and the warning correctly toggles based on whether the address is set. No SUT regression — security-relevant binding behaves as designed.

**Related board finding:** [#968 SYS-3 Unauthenticated API Surface](https://github.com/ssvlabs/ssv-node-board/issues/968). This PR is a partial mitigation (operators can now restrict bind address), not the full fix (which would require authentication).

PR #2746, opened 2026-03-25
network/peers: add connection management tests

nkryuchkov:
## Summary
Add test coverage for peer connection management across the networking stack.

Addresses:

`T-ssv-022` Zero Tests for ConnectionGater
`T-ssv-023` Zero Tests for ConnectionHandler
`T-ssv-024` No Tests for ConnManager

This PR adds coverage for:
- `connGater` address dial filtering, inbound/max-peer limits, IP rate limiting, and recently-trimmed/bad-peer rejection
- `connHandler` outbound handshake flow, concurrent handshake deduplication, inbound handshake waiting, subnet checks, and disconnect handling
- `ConnManager` peer trimming, bad-peer disconnection, and irrelevant-peer disconnection with quota handling

### DIFF
--- a/network/peers/connections/conn_gater_test.go
+++ b/network/peers/connections/conn_gater_test.go
@@ -1,0 +1,98 @@
+package connections
+
+import (
+	"testing"
+	"time"
+
+	libp2pnetwork "github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	leakybucket "github.com/prysmaticlabs/prysm/v4/container/leaky-bucket"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/utils/ttl"
+)
+
+func TestConnGaterInterceptAddrDial(t *testing.T) {
+	gater := &connGater{
+		logger:          zap.NewNop(),
+		isBadPeer:       func(id peer.ID) bool { return id == "bad-peer" },
+		trimmedRecently: ttl.New[peer.ID, struct{}](t.Context(), time.Minute, time.Minute),
+	}
+
+	require.False(t, gater.InterceptAddrDial("bad-peer", mustMultiaddr("/ip4/127.0.0.1/tcp/13000")))
+	require.True(t, gater.InterceptAddrDial("good-peer", mustMultiaddr("/ip4/127.0.0.1/tcp/13000")))
+}
+
+func TestConnGaterInterceptAcceptHonorsLimits(t *testing.T) {
+	t.Run("disabled bypasses limits", func(t *testing.T) {
+		gater := &connGater{
+			logger:          zap.NewNop(),
+			disable:         true,
+			atMaxPeersLimit: func() bool { return true },
+			atInboundLimit:  func() bool { return true },
+			trimmedRecently: ttl.New[peer.ID, struct{}](t.Context(), time.Minute, time.Minute),
+		}
+
+		require.True(t, gater.InterceptAccept(&testConn{remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13000")}))
+	})
+
+	t.Run("rejects at inbound limit", func(t *testing.T) {
+		gater := &connGater{
+			logger:          zap.NewNop(),
+			atMaxPeersLimit: func() bool { return false },
+			atInboundLimit:  func() bool { return true },
+			trimmedRecently: ttl.New[peer.ID, struct{}](t.Context(), time.Minute, time.Minute),
+		}
+
+		require.False(t, gater.InterceptAccept(&testConn{remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13000")}))
+	})
+
+	t.Run("rejects at max peers limit after a valid dial", func(t *testing.T) {
+		gater := &connGater{
+			logger:          zap.NewNop(),
+			atMaxPeersLimit: func() bool { return true },
+			atInboundLimit:  func() bool { return false },
+			ipLimiter:       leakybucket.NewCollector(ipLimitRate, ipLimitBurst, ipLimitPeriod, true),
+			trimmedRecently: ttl.New[peer.ID, struct{}](t.Context(), time.Minute, time.Minute),
+		}
+
+		require.False(t, gater.InterceptAccept(&testConn{remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13000")}))
+	})
+}
+
+func TestConnGaterInterceptAcceptRateLimitsByIP(t *testing.T) {
+	gater := &connGater{
+		logger:          zap.NewNop(),
+		atMaxPeersLimit: func() bool { return false },
+		atInboundLimit:  func() bool { return false },
+		ipLimiter:       leakybucket.NewCollector(ipLimitRate, ipLimitBurst, ipLimitPeriod, true),
+		trimmedRecently: ttl.New[peer.ID, struct{}](t.Context(), time.Minute, time.Minute),
+	}
+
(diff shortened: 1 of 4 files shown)


nkryuchkov:
@greptileai please review this again

y0sher (dismissed):

network/peers/connections/conn_handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}, 250*time.Millisecond, 10*time.Millisecond)
+
+	close(handshaker.release)
+
+	require.Eventually(t, func() bool {
+		return peerInfos.State(pid) == peers.StateConnected
+	}, 3*time.Second, 10*time.Millisecond)
+}
+
+func TestConnHandlerHandleInboundConnectionWaitsForHandshake(t *testing.T) {
```
olegshmuelov, line 166: The test name claims to verify that the inbound path "waits for handshake," but it doesn't actually prove the waiting behavior. The `testHandshaker` is passed but the test never asserts `handshaker.CallCount() == 0`. If someone regressed the inbound branch to call `Handshake()` directly (like the outbound path does), `testHandshaker.Handshake()` returns nil immediately -> `StateConnected` is reached -> test still passes.

To actually prove the inbound path waits rather than initiating, the test needs:
1. Assert `handshaker.CallCount() == 0` - proves the handler didn't initiate the handshake
2. Assert the peer is NOT `StateConnected` before the synthetic handshake injection - proves it actually waited

Suggestion:
```go
// After ConnectedF, verify we're NOT connected yet (still waiting).
time.Sleep(50 * time.Millisecond) // let goroutine enter Wait loop
require.NotEqual(t, peers.StateConnected, peerInfos.State(pid))

// Inject synthetic handshake...
// (existing time.AfterFunc code)

// After Eventually(StateConnected):
require.Equal(t, 0, handshaker.CallCount(), "inbound path must not initiate handshake")
```

network/peers/connections/conn_handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+	net.connectedness[pid] = libp2pnetwork.Connected
+	conn := &testConn{
+		remotePeer:      pid,
+		remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13001"),
+		stats:           libp2pnetwork.ConnStats{Stats: libp2pnetwork.Stats{Direction: libp2pnetwork.DirInbound}},
+	}
+
+	handler.Handle().ConnectedF(net, conn)
+
+	time.AfterFunc(100*time.Millisecond, func() {
```
olegshmuelov: The `time.AfterFunc(100ms)` fires relative to when the *test* runs, not when the *goroutine* sets `start := time.Now()` (conn_handler.go:120). The code only accepts `LastHandshake.After(start)`. If the background goroutine is scheduled late (loaded CI), the timer fires first, sets `LastHandshake = T+100ms`, then the goroutine starts at `T+150ms` with `start = T+150ms`, and `LastHandshake.After(start)` is false. The handshake is never recognized, the 20s deadline fires, and `require.Eventually(3s)` fails.

This affects both `TestConnHandlerHandleInboundConnectionWaitsForHandshake` and `TestConnHandlerHandleInboundConnectionRejectsPeerWithoutSharedSubnets`.

Fix: inject the handshake only after confirming the goroutine has entered the Wait loop (e.g., via a signal channel or by polling for `StateDisconnected` which `AddPeerInfo` sets). Alternatively, set `LastHandshake` to a far-future time like `time.Now().Add(time.Hour)`.

network/peers/connections/conn_handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+	net.connectedness[pid] = libp2pnetwork.Connected
+	conn := &testConn{
+		remotePeer:      pid,
+		remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13002"),
+		stats:           libp2pnetwork.ConnStats{Stats: libp2pnetwork.Stats{Direction: libp2pnetwork.DirInbound}},
+	}
+
+	handler.Handle().ConnectedF(net, conn)
+
+	time.AfterFunc(100*time.Millisecond, func() {
```
olegshmuelov: ditto

network/peers/connections/conn_handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+		stats:           libp2pnetwork.ConnStats{Stats: libp2pnetwork.Stats{Direction: libp2pnetwork.DirOutbound}},
+	}
+
+	handler.Handle().ConnectedF(net, conn)
+
+	require.Eventually(t, func() bool {
+		return peerInfos.State(pid) == peers.StateConnected
+	}, 3*time.Second, 10*time.Millisecond)
+	require.Equal(t, 1, handshaker.CallCount())
+	require.False(t, discoveredPeers.Has(pid))
```
olegshmuelov: In the goroutine, `SetState(StateConnected)` fires before `discoveredPeersPool.Delete()` (with a `logger.Debug` between them). The test observes `StateConnected` via `require.Eventually`, then immediately asserts `discoveredPeers.Has(pid) == false`. There's no formal synchronization guaranteeing `Delete` has executed by then.

Practically safe with `zap.NewNop()` - the goroutine is already past `SetState` and `Delete` follows with no blocking. But wrapping it in `require.Eventually` would make it unconditionally robust.

Suggestion:
```go
require.Eventually(t, func() bool {
    return !discoveredPeers.Has(pid)
}, 3*time.Second, 10*time.Millisecond)
```

network/peers/connections/conn_handler_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	net := newTestNetwork()
+	conn := &testConn{
+		remotePeer:      pid,
+		remoteMultiaddr: mustMultiaddr("/ip4/127.0.0.1/tcp/13003"),
+		stats:           libp2pnetwork.ConnStats{Stats: libp2pnetwork.Stats{Direction: libp2pnetwork.DirInbound}},
+	}
+
+	net.connectedness[pid] = libp2pnetwork.Connected
+	handler.Handle().DisconnectedF(net, conn)
```
olegshmuelov: nit: `TestConnHandlerDisconnectedF` calls `handler.Handle()` twice, creating two separate `NotifyBundle` instances. `DisconnectedF` doesn't depend on any per-bundle state, so it works, but reusing a single bundle would be slightly cleaner:

```go
bundle := handler.Handle()

net.connectedness[pid] = libp2pnetwork.Connected
bundle.DisconnectedF(net, conn)
// ...

net.connectedness[pid] = libp2pnetwork.NotConnected
bundle.DisconnectedF(net, conn)
```

olegshmuelov (commented):
**Suggested additional coverage**

A few paths in `connHandler` and `connManager` that would be good to cover while the test infrastructure is fresh:

- **Outbound handshake failure** — `testHandshaker.err` already supports this. Verify the peer gets disconnected and never reaches `StateConnected`.
- **`AtLimit` rejection** — outbound connection that passes handshake but `connIdx.AtLimit()` returns true. Verify disconnect.
- **State dedup in `acceptConnection`** — connect a peer that's already `StateConnected` or `StateConnecting`. Verify the connection is ignored (no handshake attempt).
- **`DisconnectFromIrrelevantPeers` error path** — when `ClosePeer` returns an error, the peer should not count against `disconnectQuota`. Curren code increments the counter only on success, but there's no test enforcing that.

network/peers/connections/conn_gater_test.go
oleg-ssvlabs, line 1: Maybe worth adding tests for `InterceptUpgraded` and `InterceptPeerDial` methods? They do not really do anything and not being called from what I can see, but still they exist and they are public.

oleg-ssvlabs (commented):
gj. Left one minor comment

olegshmuelov (approved):
LGTM!

ljuba-ssv (approved):

oleg-ssvlabs (approved):

y0sher (approved):
Gj, you can squash merge

PR #2747, opened 2026-03-25
eth/{eventhandler,eventsyncer}: add tests for inferior block

nkryuchkov:
## Summary

Addresses `T-ssv-034` by adding regression coverage for the inferior-block path in the Ethereum event pipeline.

## Changes

- add an `eventhandler` test that verifies an inferior block:
  - returns `ErrInferiorBlock`
  - reports no progress
  - leaves `lastProcessedBlock` unchanged
- add an `eventsyncer` test that verifies `SyncOngoing`:
  - does not recover internally from `ErrInferiorBlock`
  - propagates the wrapped error to the caller
- clarify the `node.go` comment around ongoing sync shutdown on block-order violations

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -1190,6 +1190,9 @@ func syncContractEvents(
 
 		// Sync ongoing registry events in the background, crash if ongoing sync has stopped because
 		// the SSV node cannot work without being up to date with Ethereum events.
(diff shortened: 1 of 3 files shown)


y0sher (approved):

iurii-ssv (approved):
👍

PR #2748, opened 2026-03-25
feat(p2p): register discovery health with node prober

### LINKED ISSUE #2738: feat(p2p): register discovery health with node prober
## Context

PR #2730 replaced `logger.Fatal` in `bootstrapDiscovery` with `logger.Error` to prevent ungraceful crashes. This is correct — Fatal kills without running defers — but discovery now fails silently. If bootstrap exhausts its 3 retries, the goroutine exits and nothing notices.

The node already has `nodeprobe.Prober` that monitors CL, EL, and EventSyncer via the `Healthy(ctx) error` interface. P2P is the only core subsystem not registered with it. The `/v1/node/health` endpoint checks peer count, but that's a symptom check — it won't catch "discovery died 10 minutes ago and peer count is slowly decaying."

## Problem

`bootstrapDiscovery` is fire-and-forget (`go n.bootstrapDiscovery(...)` in `p2p_discovery.go`). On failure after 3 retries, it logs and returns. No health is reported. The prober doesn't know. The health endpoint doesn't know. The node runs with degraded networking until peers naturally churn away.

Same pattern exists for other p2p goroutines (backoffConnector, peer trimming, peer selection) — all fire-and-forget via `async.Interval` with no error surfacing.

### Phase 1: Plug into existing prober (minimal, immediate)

Add `Healthy()` to `p2pNetwork` and register with the prober:

```go
func (n *p2pNetwork) Healthy(ctx context.Context) error {
    if !n.isReady() {
        return fmt.Errorf("p2p network not ready")
    }
    if n.discoveryFailed.Load() {
        return fmt.Errorf("discovery bootstrap failed")
    }
    // Could also check: time since last peer discovered,
    // minimum peer count, discovery listener alive
    return nil
}
```

Track discovery state via an `atomic.Bool` set in `bootstrapDiscovery` on terminal failure. Register in `startNodeProber()` alongside CL/EL/EventSyncer.

### Phase 2: Restart mechanism (follow-up)

Wrap `bootstrapDiscovery` with retry-on-failure with exponential backoff, rather than the current "3 retries then die" approach. Options range from simple (relaunch goroutine with backoff in a wrapper) to structured (suture-style supervisor for p2p goroutines).

The current node pattern is "crash and let k8s restart" — phase 1 preserves this by making the prober detect the failure and Fatal. Phase 2 would add in-process recovery for transient failures.

## Scope

- **Phase 1** — p2p discovery only. Minimal: add `Healthy()`, track state, register with prober.
- **Phase 2** — restart mechanism + potentially extend to other fire-and-forget goroutines (backoffConnector, peer trimming, etc.).

## References

- #2730 — fix(p2p): replace Fatal calls with graceful error handling
- Finding: F-ssv-049
- `nodeprobe/nodeprobe.go` — existing Prober infrastructure
- `network/p2p/p2p_discovery.go` — `bootstrapDiscovery` (fire-and-forget)
- `api/handlers/node/node.go` — `Health()` handler (checks peer count only)

olegshmuelov:
## Summary

- Add `Healthy()` method to `p2pNetwork` that checks network readiness and discovery bootstrap state
- Register p2p with the existing `nodeprobe.Prober` so the 60s health loop detects silent discovery failures introduced by PR #2730's Fatal-to-Error migration
- Generalize prober fatal message from "ethereum node(s)" to "node(s)" now that p2p is included

## Test

- Table-driven unit tests for `Healthy()`: not ready, ready+healthy, discovery failed, canceled context
- Lint clean (`make lint` passes with 0 issues)

Closes #2738


cli/operator/prober.go
```diff
@@ -23,14 +24,14 @@ const (
...
 func ensureEthereumNodesHealthy(ctx context.Context, logger *zap.Logger, p *nodeprobe.Prober) {
 	probeCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	if err := p.ProbeAll(probeCtx); err != nil {
-		logger.Fatal(ethereumNodesUnhealthyFatalErrorMsg, zap.Error(err))
+		logger.Fatal(nodesUnhealthyFatalErrorMsg, zap.Error(err))
 	}
 
 	logger.Info("ethereum node(s) are healthy")
```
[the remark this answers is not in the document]
olegshmuelov: ensureEthereumNodesHealthy is called at node.go:589 before p2p or event syncer are registered — only CL + EL are in the prober at that point. The log is accurate. The rename was intentionally scoped to startNodeProber fatal message, which is the loop that probes all nodes including p2p.

network/p2p/p2p.go
```diff
@@ -459,6 +465,7 @@ func (n *p2pNetwork) bootstrapDiscovery(connector chan peer.AddrInfo) {
 		})
 	}, 3)
 	if err != nil {
+		n.discoveryFailed.Store(true)
 		n.logger.Error("could not setup discovery", zap.Error(err))
 		return
 	}
```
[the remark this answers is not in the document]
olegshmuelov: Bootstrap() returns nil on context cancellation — the discover() loop exits on ctx.Done(), then Bootstrap returns nil. So tasks.Retry sees success and the if err != nil block is never entered during shutdown. The flag cannot be set by context cancellation.

Even hypothetically: Healthy() checks ctx.Err() first, so a poisoned flag would be masked by context.Canceled — probeNode treats that as benign. No spurious Fatal.

network/p2p/p2p.go
```diff
@@ -468,6 +475,21 @@ func (n *p2pNetwork) isReady() bool {
...
+// Healthy reports whether the p2p network is operating normally.
+// It satisfies the nodeprobe health-check interface.
+func (n *p2pNetwork) Healthy(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if !n.isReady() {
+		return fmt.Errorf("p2p network not ready")
+	}
+	if n.discoveryFailed.Load() {
```
nkryuchkov, line 489: This question is related to https://github.com/ssvlabs/ssv/pull/2730 rather than this PR, but does it make sense to keep the ssv node running if discovery fails?

cli/operator/node.go
```diff
@@ -652,6 +652,7 @@ var StartNodeCmd = &cobra.Command{
 			if err := p2pNetwork.Start(); err != nil {
 				logger.Fatal("failed to start network", zap.Error(err))
 			}
+			nodeProber.AddNode(p2pNodeName, p2pNetwork.(p2pv1.HealthChecker), proberHealthcheckTimeout, proberRetriesMax, proberRetryDelay)
```
nkryuchkov, line 658: Adding the `p2p` module makes the initial name worse, because it no longer works with nodes only. It looks out of scope of this PR, though

nkryuchkov (dismissed):

network/p2p/p2p.go
olegshmuelov: Yes, but not indefinitely. The prober gives it ~2 minutes of grace (5 retries × 10s delay + 60s tick), then Fatals. This is the bridge between #2730's "don't Fatal from goroutines" and "don't stay alive when broken."

The latch is intentionally one-way: if all 3 `tasks.Retry` attempts failed, the underlying problem won't self-resolve. Process restart is the right recovery path.

cli/operator/node.go
olegshmuelov: Agreed, the naming has been drifting since `eventSyncerNodeName`. Opened #2750 for a rename pass.

network/p2p/p2p.go
nkryuchkov: Should we set `discoveryFailed` to `false` on success, then? Or did I misunderstand it?

y0sher (dismissed):

network/p2p/p2p.go
olegshmuelov: No, `bootstrapDiscovery` runs once at startup. If all 3 retries fail, the flag is set and stays set (process will restart via prober). If any retry succeeds, the flag was never set. There's no path where it's `true` and then bootstrap succeeds later.

cli/operator/node.go
momosh-ssv, line 658: What do you think about using the two-value form here? `p2pNetwork` is typed as `network.P2PNetwork`, which doesn't embed `HealthChecker` — so this panics at runtime if the concrete type ever changes (mocks, future refactor). The existing assertions on lines 665-666 have the same issue, but this is a new one.

Alternatively, adding `HealthChecker` to `network.P2PNetwork` would make the assertion unnecessary.

network/p2p/p2p.go
```diff
@@ -70,6 +70,11 @@ type HostProvider interface {
 	Host() host.Host
 }
 
+// HealthChecker provides health checking capability for the node prober.
+type HealthChecker interface {
+	Healthy(ctx context.Context) error
+}
```
momosh-ssv, line 76: Worth considering a compile-time interface satisfaction check:

```go
var _ HealthChecker = (*p2pNetwork)(nil)
```

This catches signature drift at build time rather than as a runtime panic from the type assertion in `node.go`.

momosh-ssv (commented):
Two minor points not yet raised — both around type assertion safety.

cli/operator/node.go
olegshmuelov: The pattern is consistent with `PeersIndexProvider` and `HostProvider` on lines 665-666, these are intentionally separate from `P2PNetwork`. Not every implementation needs them, only the concrete type exposes them. Promoting to the interface would force mocks to implement `Healthy()` unnecessarily.

Two-value form doesn't help meaningfully, if the assertion fails, the node can't start either way.

momosh-ssv (approved):

ljuba-ssv (approved):

nkryuchkov (approved):

cli/operator/prober.go
```diff
@@ -23,14 +24,14 @@ const (
 	proberRetryDelay         = 10 * time.Second
 )
 
-const ethereumNodesUnhealthyFatalErrorMsg = "ethereum node(s) are not healthy"
+const nodesUnhealthyFatalErrorMsg = "node(s) are not healthy"
 
 func ensureEthereumNodesHealthy(ctx context.Context, logger *zap.Logger, p *nodeprobe.Prober) {
```
iurii-ssv, line 29: I guess we need to remove the "ethereum" from other places in this file to be consistent (seems like `ensureEthereumNodesHealthy` is the only one left)

cli/operator/prober.go
iurii-ssv, line 37: should remove `ethereum`

iurii-ssv (commented):
LGTM, just couple clean-ups to keep things consistent

cli/operator/prober.go
olegshmuelov: agreed on that.
pls see. https://github.com/ssvlabs/ssv/pull/2748#discussion_r2987972309

cli/operator/prober.go
olegshmuelov: ditto

iurii-ssv (approved):

iurii-ssv:
To cross-link with our dashboard: `F-ssv-267`

PR #2751, opened 2026-03-25
network/streams: cap stream reads to prevent OOM (iurii suggestions)

iurii-ssv:
Aligns the way `Request` and `HandleStream` funcs work with streams.

### DIFF
--- a/network/streams/controller.go
+++ b/network/streams/controller.go
@@ -54,35 +54,36 @@ func (n *streamCtrl) Request(logger *zap.Logger, peerID peer.ID, protocol protoc
 	ctx, cancel := context.WithTimeout(n.ctx, n.dialTimeout)
 	defer cancel()
 
-	s, err := n.host.NewStream(ctx, peerID, protocol)
+	stream, err := n.host.NewStream(ctx, peerID, protocol)
 	if err != nil {
 		return nil, err
 	}
 
-	requestsSentCounter.Add(n.ctx, 1, metric.WithAttributes(protocolIDAttribute(s.Protocol())))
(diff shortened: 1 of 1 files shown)


network/streams/controller.go
```diff
@@ -54,35 +54,36 @@ func (n *streamCtrl) Request(logger *zap.Logger, peerID peer.ID, protocol protoc
...
 	if err != nil {
 		if errors.Is(err, ErrStreamMessageTooLarge) {
-			n.observeOversizedPayload(logger, peerID, stream.Protocol(), "response")
+			n.observeOversizedPayload(logger, peerID, s.Protocol(), "response")
 		}
 		return nil, errors.Wrap(err, "could not read stream msg")
 	}
 
-	responsesReceivedCounter.Add(n.ctx, 1, metric.WithAttributes(protocolIDAttribute(stream.Protocol())))
+	responsesReceivedCounter.Add(n.ctx, 1, metric.WithAttributes(protocolIDAttribute(s.Protocol())))
```
momosh-ssv, line 86: Worth considering using `stream.Protocol()` here (and on line 81) instead of `s.Protocol()` — `HandleStream` consistently uses the raw `stream` for all `Protocol()` calls, and line 62 in this same function does too. Functionally identical today since `streamWrapper` doesn't override `Protocol()`, but keeping it uniform avoids a subtle divergence if that ever changes.

iurii-ssv: Good point, I think it's cleaner if we use our wrapper everywhere - https://github.com/ssvlabs/ssv/pull/2751/changes/755f1e592766440c8fe93aa87c58be0f66ce1090

momosh-ssv (approved):
no issues here, gj!

nkryuchkov (approved):

PR #2752, opened 2026-03-26
chore(ci) - Update actions base image to node:24

vaclav-ssvlabs:
## Summary
GitHub is enforcing Node.js 24 for GitHub Actions runtime and will block workflows that rely on older Node runtimes.

This PR updates CI workflows to align with the requirement.

## Changes
- Update reusable action references (`uses:`) to latest released versions.
- Update workflow Node container images (`node:<version>`) to `node:24` where they were below 24.

### DIFF
--- a/.github/workflows/spec-alignment.yml
+++ b/.github/workflows/spec-alignment.yml
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6.0.2
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6.3.0
         with:
           go-version: "1.24.x"
 
@@ -26,7 +26,7 @@ jobs:
 
       - name: Upload output.diff
         if: failure()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7.0.0
         with:
           name: output.diff
           path: ./scripts/spec-alignment/output.diff
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6.0.2
         with:
           lfs: true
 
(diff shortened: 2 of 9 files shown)


.github/workflows/unit-test.yml
```diff
@@ -16,23 +16,23 @@ jobs:
...
+      - uses: actions/checkout@v6.0.2
         with:
           lfs: true
 
       - name: Setup make
         run: sudo apt-get update && sudo apt-get install make
 
       - name: Setup Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6.3.0
```
momosh-ssv, line 27: Should we also update `actions/checkout@v4` on line 19 of this file to `@v6.0.2`? It looks like checkout was updated in 3 workflows (dutytracer-bench, spec-alignment, ssvsigner-e2e-test) but still references `@v4` here and in 4 other files (full-test, spec-test, spec-test-raceless, lint). Could this cause those workflows to break once GitHub enforces Node 24?

.github/workflows/spec-alignment.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Upload output.diff
         if: failure()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7.0.0
```
momosh-ssv, line 29: Should we also be checking whether `actions/stale@v9.1.0` in `stale.yml` needs a bump for Node 24 compatibility? Might be worth verifying so it does not get missed.

momosh-ssv (commented):
Nice catch getting ahead of the Node 24 enforcement. A few things I noticed below — mainly around completeness of the migration.

julienh-ssv (approved):

.github/workflows/unit-test.yml
iurii-ssv: Seems like it was updated already in this file ?

image

.github/workflows/spec-alignment.yml
iurii-ssv: Seems it does need a bump indeed:

image

iurii-ssv (commented):
I think it's better to open this PR against **stage** branch @vaclav-ssvlabs (we'll merge **stage** -> **main** before the `June 2nd, 2026` deadline you've quoted)

.github/workflows/spec-alignment.yml
vaclav-ssvlabs: I see it is part of the PR already

vaclav-ssvlabs:
closing in favor of https://github.com/ssvlabs/ssv/pull/2759

vaclav-ssvlabs:
@iurii-ssv @momosh-ssv can you review that new PR

Issue #2753, opened 2026-03-26
fix(qbft): RoundTimer first-round timeout silently dropped due to late callback registration


olegshmuelov:
## Problem

The `RoundTimer` is constructed with a nil callback (`done=nil`), armed for round 1 during `Instance.Start()`, and only receives its timeout callback when the runner calls `OnTimeout()` afterward. If the first-round timeout fires in that window, the callback is nil and the timeout is silently dropped.

### Call sequence

1. `roundtimer.New(ctx, ..., nil)` — timer created with nil callback (`operator/validator/controller.go:1027`, `:1088`)
2. `Controller.StartNewInstance()` → `Instance.Start()` — arms round 1 (`protocol/v2/qbft/instance/instance.go:110`)
3. Runner registers callback (`protocol/v2/ssv/runner/timer.go:20`)

For late-start duties where `RoundTimeout()` returns near-zero, the timer fires between steps 2 and 3.

### Impact

The instance misses its first round timeout, delaying the round bump when the proposer is offline. Recovery depends on receiving round-change messages from peers who did timeout.

## Possible approaches

**A. Pending-timeout replay inside RoundTimer** — If `done == nil` when the timer fires, store the pending round. When `OnTimeout()` is later called, replay if the context is still active and the round still matches. Contained to `roundtimer/timer.go` and tests.

**B. Restructure startup sequence** — Pass the callback through `StartNewInstance()` → `Start()`, or add an atomic "register callback + arm" API on `RoundTimer`. Cleaner long-term but touches controller, instance, runner, and spectests.

PR #2754, opened 2026-03-26
fix(qbft): stop timer leak in RoundTimer.TimeoutForRound

olegshmuelov:
## Summary

- Replace leaky goroutine+channel pattern in `RoundTimer.TimeoutForRound` with `time.AfterFunc` — each call now stops the previous timer and schedules a new one, eliminating timer and goroutine leaks (up to ~12 per QBFT instance)
- Move round update inside mutex to linearize with timer replacement, snapshot round+done under same RLock in callback, call done after releasing lock
- Remove `waitForRound` method entirely

## Finding

F-ssv-001 — RoundTimer: Timer Field Never Stored Back — Goroutine and Timer Leak

## Follow-up

Pre-existing done=nil registration race discovered during review: #2753

## Test

- 15 existing tests pass unchanged
- Added `timer_stored_and_reused` — guards the defect class directly (would fail on old code where `t.timer` stayed nil)
- Added `context_cancelled_before_arm` — verifies early return skips timer creation, no callback fires
- Added `context_cancelled_after_arm` — verifies callback is suppressed when context cancelled after arming
- 27 total tests, all passing across 4 roles

### DIFF
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -141,7 +141,8 @@ func (t *RoundTimer) OnTimeout(done OnRoundTimeoutF) {
 	t.done = done
 }
 
-// TimeoutForRound stops any running timer and schedules a callback for the given round.
+// TimeoutForRound stops any running timer, waits for any in-flight callback to
+// finish, and schedules a new callback for the given round.
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
 	// Optimistic early-exit: a narrow window exists between this check and
 	// timer creation where ctx could be canceled; the callback re-checks
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -149,6 +149,9 @@ func (t *RoundTimer) Round() specqbft.Round {
 
 // TimeoutForRound stops any running timer and schedules a callback for the given round.
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
+	// Optimistic early-exit: a narrow window exists between this check and
+	// timer creation where ctx could be canceled; the callback re-checks
+	// ctx.Err() to handle that case.
 	if t.ctx.Err() != nil {
 		return
 	}
@@ -158,6 +161,8 @@ func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Roun
 	t.mtx.Lock()
 	atomic.StoreUint64(&t.round, uint64(round))
 	if t.timer != nil {
+		// Stop prevents future fires; if it returns false the callback goroutine
+		// may already be running — it will be suppressed by the round-number check.
 		t.timer.Stop()
 	}
 	t.timer = time.AfterFunc(timeout, func() {
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -3,7 +3,6 @@ package roundtimer
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
@@ -48,7 +47,7 @@ type RoundTimer struct {
 	// result holds the result of the timer
 	done OnRoundTimeoutF
 	// round is the current round of the timer
-	round uint64
+	round specqbft.Round
 	// timeoutOptions holds the timeoutOptions for the timer
 	timeoutOptions TimeoutOptions
 	// role is the role of the instance
@@ -142,11 +141,6 @@ func (t *RoundTimer) OnTimeout(done OnRoundTimeoutF) {
 	t.done = done
 }
 
-// Round returns a round.
-func (t *RoundTimer) Round() specqbft.Round {
-	return specqbft.Round(atomic.LoadUint64(&t.round)) // #nosec G115
-}
-
 // TimeoutForRound stops any running timer and schedules a callback for the given round.
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
 	// Optimistic early-exit: a narrow window exists between this check and
@@ -159,10 +153,7 @@ func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Roun
 	timeout := t.RoundTimeout(height, round)
 
 	t.mtx.Lock()
-	// round is read lock-free via Round(), so accesses stay atomic.
-	// The store happens while t.mtx is held only to keep round updates
-	// linearized with timer/done/deferred state changes.
-	atomic.StoreUint64(&t.round, uint64(round))
+	t.round = round
 	if t.timer != nil {
 		// Stop prevents future fires; if it returns false the callback goroutine
 		// may already be running — it will be suppressed by the round-number check.
@@ -175,13 +166,13 @@ func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Roun
 			return
 		}
 		t.mtx.RLock()
-		currentRound := t.Round()
-		done := t.done
-		t.mtx.RUnlock()
-		if currentRound != round || done == nil {
+		defer t.mtx.RUnlock()
+		if t.round != round {
 			return
 		}
-		done(round)
+		if t.done != nil {
+			t.done(round)
+		}
 	})
 	t.mtx.Unlock()
 }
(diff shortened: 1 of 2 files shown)


protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -147,40 +147,36 @@ func (t *RoundTimer) Round() specqbft.Round {
...
-		timer = time.NewTimer(timeout)
-	} else {
-		timer.Stop()
-		// draining the channel of existing timer
-		select {
-		case <-timer.C:
-		default:
-		}
+	t.mtx.Lock()
+	atomic.StoreUint64(&t.round, uint64(round))
```
momosh-ssv: I'm thinking about picking one synchronization strategy for `t.round` here? Right now it is written with `atomic.StoreUint64` under `t.mtx.Lock()`, but read via `atomic.LoadUint64` in `Round()` without any lock. The atomic store under the mutex is redundant — either the mutex serializes writes (and all reads should also lock), or the atomic is sufficient on its own (and the mutex is unnecessary for this field).

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -147,40 +147,36 @@ func (t *RoundTimer) Round() specqbft.Round {
...
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
-	atomic.StoreUint64(&t.round, uint64(round))
+	// Optimistic early-exit: a narrow window exists between this check and
+	// timer creation where ctx could be canceled; the callback re-checks
+	// ctx.Err() to handle that case.
+	if t.ctx.Err() != nil {
+		return
+	}
+
 	timeout := t.RoundTimeout(height, round)
```
momosh-ssv, line 154: Worth considering: `RoundTimeout` returns `time.Until(dutyStartTime.Add(...))`, which can be negative if the duty start is in the past (late consensus start, clock skew). `time.AfterFunc` with a negative duration fires the callback immediately — while the Lock on line 161 is still held. 

This seems safe (callback blocks on RLock until Unlock on 181), and matches old behavior, but the immediate-fire path is an undocumented assumption. 
Should we do a short test for the negative-timeout case and/or an explicit clamp with a comment would make the intent clear?

protocol/v2/qbft/roundtimer/timer.go
olegshmuelov: Intentional, `Round()` is a lock-free reader via `atomic.LoadUint64`, the store happens under the lock only because we're already locked for timer/done state. Added a clarifying comment.

protocol/v2/qbft/roundtimer/timer.go
olegshmuelov: This is the late-start scenario tracked in #2753. `AfterFunc` fires immediately but blocks on `RLock` until we release the write lock, so no deadlock. Added an inline comment here, and a `TestNegativeTimeout` in the stacked #2755 (needs the `OnTimeout(height, done)` API).

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: WDYT of simplifying it by getting rid of this public `Round()` method (which I think allows us to also get rid of `atomic` then) ?

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -147,40 +147,41 @@ func (t *RoundTimer) Round() specqbft.Round {
...
 		}
-	}
+		t.mtx.RLock()
+		currentRound := t.Round()
+		done := t.done
+		t.mtx.RUnlock()
+		if currentRound != round || done == nil {
+			return
+		}
+		done(round)
```
iurii-ssv: From what I understand, this `currentRound != round` covers the case when "we are about to call `done` for Round N, but it might be already kinda late to run that `done` due to Round N+1 being already upon us ... so we quit early" ... but if we've already called `done` func for Round N then it won't be aborted anymore even if Round N+1 is already upon us -> which means the "`done` call for Round N+1" might execute before the "`done` call for Round N" (unless the implementor of `done` explicitly thought of that scenario and handled it explicitly with locks of their own ... which isn't a good contract to have),

it seems [**stage** branch calls `done` with the mutex held](https://github.com/ssvlabs/ssv/blob/36273e32ff90f4284e5805029cc8f40002e0493a/protocol/v2/qbft/roundtimer/timer.go#L178-L182) - I assume to prevent this from ever happening, thoughts ?

y0sher:
## Staging Test Report (operators 69-72, initial ~20 min soak)

**Branch:** `fix/roundtimer-goroutine-leak`
**Baseline:** operators 29-32 running `stage`

### Results: PASS (initial) — needs longer soak to confirm leak fix

### Goroutines: Favorable per-validator ratio

| Node Group | Goroutines | Validators | Uptime | Goroutines/Validator |
|---|---|---|---|---|
| **PR (69-72)** | ~31,290 | 2,870 | ~20 min | **~10.9** |
| **Baseline (29-32)** | ~18,655 | 1,601 | 25-54 hrs | **~11.7** |

PR nodes have 79% more validators but a *lower* per-validator goroutine ratio (10.9 vs 11.7). Goroutine count was stable across two measurements ~3 min apart — no immediate growth.

### Memory: Lower per-validator

| Node | Memory (MB) |
|---|---|
| ssv-node-69 (PR) | 900 |
| ssv-node-70 (PR) | 970 |
| ssv-node-71 (PR) | 945 |
| ssv-node-72 (PR) | 930 |
| ssv-node-29 (baseline) | 1,186 |
| ssv-node-30 (baseline) | 1,360 |
| ssv-node-31 (baseline) | 1,269 |
| ssv-node-32 (baseline) | 1,299 |

PR nodes use ~900-970 MB vs baseline ~1,186-1,360 MB despite having **79% more validators**.

### Functional: PASS

- Duties completing with 100% success, consensus in round 1
- One isolated round-change event at slot 2700194 right after startup (warm-up, not timer issue)
- No panics, no timer-related errors

### Errors: PASS (all pre-existing)

- `"could not publish p2p message: validation ignored"` — transient P2P, not timer-related
- `"failed to find event by ID"` — network-wide event sync issue (same on baseline nodes)

### Caveat

**~20 min uptime is too short to conclusively validate a goroutine leak fix.** The leak manifests as goroutines accumulating over hours/days with each new round. Will re-check in 2-4 hours to confirm goroutine count stays stable rather than growing.

### TODO

- [ ] Re-check goroutine count in 2-4 hours (should remain ~31,290 if fix works; would grow on unfixed code)
- [ ] Re-check memory after extended uptime
- [ ] Compare goroutine growth rate: PR nodes vs baseline

---

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: > unless the implementor of `done` explicitly thought of that scenario and handled it explicitly with locks of their own ... which isn't a good contract to have

Just seems like we don't gain much by complicating this contract. Does it make sense to fire Round 2 timeout before Round 1 timeout callback completed ?

A related consideration, `TimeoutForRound` blocks on that mutex ... so waiting for `done` to complete (with the mutex held) would mean the "scheduling of round timeout" is a blocking call ... which is undesirable but probably fine.

It seems there should be 2 different mechanisms/locks:
- 1st is for the scheduling of timeouts (perhaps an intent sent on a non-blocking channel)
- 2nd is for the serialized execution of those `done` funcs (to ensure Round 1 timeout always completes before Round 2 timeout, etc.)

Anyway, just recorded some thoughts, maybe not worth the implementation complexity.

protocol/v2/qbft/roundtimer/timer.go
olegshmuelov: Good call. Removed `Round()` and dropped `sync/atomic`, `t.round` is now `specqbft.Round` read/written only under the mutex. One sync model.

protocol/v2/qbft/roundtimer/timer.go
olegshmuelov: You're right. Moved `done(round)` back under RLock, matching stage. The call chain is non-reentrant (`done` only enqueues a timeout event), so no deadlock risk. Serialization preserved.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -142,45 +141,38 @@ func (t *RoundTimer) OnTimeout(done OnRoundTimeoutF) {
 	t.done = done
 }
 
-// Round returns a round.
-func (t *RoundTimer) Round() specqbft.Round {
-	return specqbft.Round(atomic.LoadUint64(&t.round)) // #nosec G115
-}
-
-// TimeoutForRound times out for a given round.
+// TimeoutForRound stops any running timer and schedules a callback for the given round.
```
iurii-ssv: Nit: could update func description to expand a bit on the way this func works:
```go
// TimeoutForRound stops any running timer (or waits for the already running timeout-callback to complete) 
// and schedules a callback for the given round.
```

iurii-ssv (dismissed):
GJ!

iurii-ssv (approved):

momosh-ssv (approved):

PR #2755, opened 2026-03-26
fix(qbft): defer RoundTimer arming until callback is registered

### LINKED ISSUE #2753: fix(qbft): RoundTimer first-round timeout silently dropped due to late callback registration

olegshmuelov:
## Summary

- Defer `RoundTimer` arming when the timeout callback is missing or stale from a previous duty, preventing silently dropped first-round timeouts on late-start duties
- Add height-aware `OnTimeout(height, done)` so `TimeoutForRound` can distinguish the current duty's callback from a stale one — deferred requests are replayed under the lock via `armLocked` once the correct callback is registered
- Stop and nil stale timers in the deferred path to suppress in-flight callbacks from previous duties

Closes #2753

## Finding

#2753 — RoundTimer first-round timeout silently dropped due to late callback registration

## Test

- 20 new deferred-arming tests across all 4 roles: first-duty nil callback, stale callback on duty N+1, stale timer stopped, deferred replaced by new round, context canceled before OnTimeout
- 28 existing tests pass unchanged
- `golangci-lint`: 0 issues

### DIFF
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -196,9 +196,10 @@ func (t *RoundTimer) OnTimeout(height specqbft.Height, callback OnRoundTimeoutF)
 	}
 }
 
-// TimeoutForRound stops any running timer and schedules a callback for the given round.
-// If the callback is not yet registered for this duty's height (nil or stale from a
-// previous duty), the request is deferred until OnTimeout registers the correct callback.
+// TimeoutForRound stops any running timer, waits for any in-flight callback to
+// finish, and schedules a new callback for the given round. If the callback is
+// not yet registered for this duty's height (nil or stale from a previous duty),
+// the request is deferred until OnTimeout registers the correct callback.
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
 	if t.ctx.Err() != nil {
 		return
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -149,6 +149,9 @@ func (t *RoundTimer) Round() specqbft.Round {
 
 // TimeoutForRound stops any running timer and schedules a callback for the given round.
 func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
+	// Optimistic early-exit: a narrow window exists between this check and
+	// timer creation where ctx could be canceled; the callback re-checks
+	// ctx.Err() to handle that case.
 	if t.ctx.Err() != nil {
 		return
 	}
@@ -158,6 +161,8 @@ func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Roun
 	t.mtx.Lock()
 	atomic.StoreUint64(&t.round, uint64(round))
 	if t.timer != nil {
+		// Stop prevents future fires; if it returns false the callback goroutine
+		// may already be running — it will be suppressed by the round-number check.
 		t.timer.Stop()
 	}
 	t.timer = time.AfterFunc(timeout, func() {
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -39,14 +39,27 @@ type TimeoutOptions struct {
 	slow           time.Duration
 }
 
+// deferredTimeout stores a TimeoutForRound request that arrived before the
+// correct callback was registered for this duty's height. OnTimeout replays
+// it once the callback is set.
+type deferredTimeout struct {
+	height specqbft.Height
+	round  specqbft.Round
+}
+
 // RoundTimer helps to manage current instance rounds.
 type RoundTimer struct {
 	mtx *sync.RWMutex
 	ctx context.Context
 	// timer is the underlying time.Timer
 	timer *time.Timer
-	// result holds the result of the timer
+	// done is the timeout callback for the current duty
 	done OnRoundTimeoutF
+	// doneHeight is the height for which done was registered
+	doneHeight specqbft.Height
+	// deferred stores a pending arm request when TimeoutForRound is called
+	// before the correct callback is registered for this duty
+	deferred *deferredTimeout
 	// round is the current round of the timer
 	round uint64
 	// timeoutOptions holds the timeoutOptions for the timer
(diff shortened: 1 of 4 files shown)


momosh-ssv (commented):
The fix seems sound — two suggestions below.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -39,14 +39,27 @@ type TimeoutOptions struct {
...
 type RoundTimer struct {
 	mtx *sync.RWMutex
 	ctx context.Context
 	// timer is the underlying time.Timer
 	timer *time.Timer
-	// result holds the result of the timer
+	// done is the timeout callback for the current duty
 	done OnRoundTimeoutF
+	// doneHeight is the height for which done was registered
+	doneHeight specqbft.Height
```
momosh-ssv: Should we consider removing the `done` parameter from `New` now that `OnTimeout(height, done)` is the canonical registration path?

`New` sets `done` but leaves `doneHeight` at zero. This works because production callers pass `nil` and tests happen to use `FirstHeight == 0`, but if anyone passes a non-nil `done` for a non-zero height, `TimeoutForRound` will enter the deferred path, nil out the provided callback, and nobody calls `OnTimeout` to replay it — silently losing the timeout.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -148,35 +195,31 @@ func (t *RoundTimer) Round() specqbft.Round {
...
-	// Optimistic early-exit: a narrow window exists between this check and
-	// timer creation where ctx could be canceled; the callback re-checks
-	// ctx.Err() to handle that case.
 	if t.ctx.Err() != nil {
 		return
 	}
 
-	timeout := t.RoundTimeout(height, round)
-
 	t.mtx.Lock()
```
momosh-ssv, line 208: Worth considering `defer t.mtx.Unlock()` here for consistency with `OnTimeout`. Two manual unlock paths (lines 219, 224) work correctly today but reduce resilience to future edits — an added early return could forget the unlock.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -134,12 +146,48 @@ func (t *RoundTimer) RoundTimeout(height specqbft.Height, round specqbft.Round)
...
 
 	t.done = done
+	t.doneHeight = height
+	d := t.deferred
+	t.deferred = nil
+
+	// Replay the deferred arm now that the correct callback is registered.
+	// Done under the lock to linearize against concurrent TimeoutForRound
+	// calls from message processing (proposal, round-change, timeout bump).
+	if d != nil && d.height == height && t.ctx.Err() == nil {
```
iurii-ssv: Probably cleaner to have this `t.ctx.Err() == nil` be the very 1st thing `OnTimeout` func check

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: Nit: could `done`->`callback` and `doneHeight`->`callbackHeight` for clarity

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -148,40 +196,34 @@ func (t *RoundTimer) Round() specqbft.Round {
...
 
-	timeout := t.RoundTimeout(height, round)
-
 	t.mtx.Lock()
+	defer t.mtx.Unlock()
+
 	// round is read lock-free via Round(), so accesses stay atomic.
 	// The store happens while t.mtx is held only to keep round updates
 	// linearized with timer/done/deferred state changes.
 	atomic.StoreUint64(&t.round, uint64(round))
```
iurii-ssv: We are updating `t.round` unconditionally here (regardless of what `height` it belongs to), this seems dangerous to me, can't we overwrite the `t.round` that was stored there for the current height with a round-value from some previous height (due to a delayed `TimeoutForRound` that was targeting the duty for the previous slot) ?

iurii-ssv (commented):
An improvement over what we had for sure, 

**Discussion**: don't like that `RoundTimer` is duty-height dependent though (maybe we could spawn 1 per duty instead ?) ... Would be nice to make it more abstract (the fact that it works with the "current duty" is very error-prone since "current" can mean slot N or slot N+1 at the slot boundary). Makes me feel like `RoundTimer` needs to enforce monotonic `doneHeight` increases somewhere at the very least (unless it's already done implicitly somehow).

protocol/v2/qbft/roundtimer/timer.go
olegshmuelov: Good catch, real correctness issue. Fixed:
- `t.round` only set inside `armLocked`, not unconditionally in `TimeoutForRound`
- Monotonic height guards in both `OnTimeout` and `TimeoutForRound`
- `callbackHeight` advances in the deferred path, closing the window between `TimeoutForRound(H+1)` and `OnTimeout(H+1)`
- New test `testStaleDuringDeferredWindow` covers the exact interleaving

olegshmuelov:
> An improvement over what we had for sure,
> 
> **Discussion**: don't like that `RoundTimer` is duty-height dependent though (maybe we could spawn 1 per duty instead ?) ... Would be nice to make it more abstract (the fact that it works with the "current duty" is very error-prone since "current" can mean slot N or slot N+1 at the slot boundary). Makes me feel like `RoundTimer` needs to enforce monotonic `doneHeight` increases somewhere at the very least (unless it's already done implicitly somehow).

Agree that a per-duty timer would eliminate the height-tracking complexity entirely. It's the right direction, I'd like to take that on as a follow-up once the goroutine-leak base lands on stage.

For this PR: both `OnTimeout` and `TimeoutForRound` now enforce monotonic height increases, and `callbackHeight` advances in the deferred path too, so the stale-call surface is fully closed.

iurii-ssv (dismissed):
Looks correct to me, I would strongly recommend we refactor it further to at least implement "per duty timer" to keep the complexity of this code manageable (as a follow-up Issue/PR).

momosh-ssv (dismissed):
see no issues here, nice one!

ljuba-ssv (dismissed):

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -133,47 +145,87 @@ func (t *RoundTimer) RoundTimeout(height specqbft.Height, round specqbft.Round)
...
+	// calls from message processing (proposal, round-change, timeout bump).
+	if d != nil && d.height == height {
+		t.armLocked(d.height, d.round)
+	}
+}
+
+// TimeoutForRound stops any running timer and schedules a callback for the given round.
+// If the callback is not yet registered for this duty's height (nil or stale from a
+// previous duty), the request is deferred until OnTimeout registers the correct callback.
+func (t *RoundTimer) TimeoutForRound(height specqbft.Height, round specqbft.Round) {
```
iurii-ssv: Seems the new comment (1st line of it) overrides previous description that's a bit more comprehensive:
```
// TimeoutForRound stops any running timer, waits for any in-flight callback to
// finish, and schedules a new callback for the given round.
```

iurii-ssv (dismissed):
Just a minor thing

momosh-ssv (dismissed):

ljuba-ssv (dismissed):

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2756, opened 2026-03-26
duty-fetching: rework proposer (better retries & edge-cases)

iurii-ssv:
This PR is the continuation of https://github.com/ssvlabs/ssv/pull/2699 (which was merged into **duty-fetching-rework-proposer** branch so we can thoroughly test it before merging it into **stage** branch). cc @10234567Z FYI.

The general approach mirrors https://github.com/ssvlabs/ssv/pull/2712 pretty much exactly, with the only notable change added being: the handling of reorg events for proposer-duty was extended to cover the case of epoch-transition, see https://github.com/ssvlabs/ssv/pull/2699#discussion_r2976408498 for more details.

Closes https://github.com/ssvlabs/ssv/issues/2619 / `F-ssv-120` / https://github.com/ssvlabs/ssv-node-board/issues/676


operator/duties/proposer.go
```diff
@@ -84,81 +96,178 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-		return
+		err := h.fetchAndProcessDuties(ctx, logger, currentEpoch, currentSlot)
+		if err != nil {
+			logger.Error("fetching duties for the current epoch failed", zap.Error(err))
+			return
+		}
+		h.dutyFetchIntents[currentEpoch] = true // the intent has been fulfilled
+
+		logger.Debug("fetching duties for the current epoch succeeded")
 	}
```
momosh-ssv, line 266: Worth considering adding an OpenTelemetry span to `prepareCurrentEpoch` (and `prepareNextEpoch`). The old `processFetching` wrapped the fetch with an outer `proposer.fetch` span that captured status and events regardless of outcome. These new methods skip that, so the higher-level signal that distinguishes "we attempted a fetch" from "the intent check short-circuited" is lost. Only the inner `proposer.fetch_and_store` span from `fetchAndProcessDuties` survives.

operator/duties/proposer.go
```diff
@@ -84,81 +96,178 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+				zap.Uint64("current_slot", uint64(currentSlot)),
+			)
 
-		case <-h.indicesChange:
-			currentSlot := h.beaconConfig.EstimatedCurrentSlot()
-			currentEpoch := h.beaconConfig.EstimatedEpochAtSlot(currentSlot)
+			logger.Info("🔀 reorg event received",
+				zap.Any("event", reorgEvent),
+				zap.Bool("refetch_current_epoch_duties", refetchCurrentEpoch),
+				zap.Bool("refetch_next_epoch_duties", true),
```
momosh-ssv: Should we drop this field? `refetch_next_epoch_duties` is hardcoded to `true` so it never varies and adds no diagnostic signal to the log line.

operator/duties/proposer.go
```diff
@@ -84,81 +96,178 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+					if h.atLastSlotOfCurrentEpoch(currentSlot) {
+						delete(h.dutyFetchIntents, currentEpoch) // optimization: prune irrelevant intent
+						h.prepareNextEpoch(tickCtx, logger, currentEpoch, currentSlot)
+					} else {
+						h.prepareCurrentEpoch(tickCtx, logger, currentEpoch, currentSlot)
 					}
+				case <-time.After(time.Until(indicesChangeDeadline)):
+					// It's too late(risky) to handle indices change on the current slot, we'll do it on the next slot.
+				case <-tickCtx.Done():
+					return
```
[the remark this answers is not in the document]
10234567Z: Seems trivial considering worst case we are considering is like 4s~ as ``indicesChangeDeadline`` is basically ``slotStart + IntervalDuration`` -> ``slotStart + (slotDuration / intervalsPerSlot)`` -> ``slotStart + (12/3)`` -> ``slotStart + 4s``. Which get GC'd anyway once it gets fired.

This same logic we even have in ``scheduler.go`` , so I think this should be fine?

operator/duties/proposer.go
10234567Z: > Should we drop this field? `refetch_next_epoch_duties` is hardcoded to `true` so it never varies and adds no diagnostic signal to the log line.

Yeah makes sense, considering now on any kind of reorg, the next  epoch duties would be by default getting refetched.

operator/duties/proposer.go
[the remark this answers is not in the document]
iurii-ssv: > `time.After` allocates a timer that cannot be garbage-collected until it fires

I think this has been resolved in the more recent versions of Golang. But even if not, we are not creating too many of those timer-objects anyway (to result into a memory leak) ... so it's not a real issue.

operator/duties/proposer.go
iurii-ssv: Could remove, no strong opition.

operator/duties/proposer.go
```diff
@@ -84,78 +96,198 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+				h.processExecution(tickCtx, currentEpoch, currentSlot)
+
+				// Process intents (if any): fetch & prepare the duties for the next epoch.
+				h.prepareNextEpoch(tickCtx, logger, currentEpoch, currentSlot)
+
+				// Clean up the irrelevant data to prevent infinite memory growth at the 1st slot of the epoch.
+				if slotNumber == 1 && currentEpoch >= 1 {
+					h.duties.EraseEpochData(currentEpoch - 1)
+					delete(h.dutyFetchIntents, currentEpoch-1)
+				}
```
olegshmuelov, line 114: The old proposer handler always refetched duties at epoch start (`fetchFirst = true` set at the last slot of each epoch). The new intent-based mechanism trusts pre-fetched duties: if `prepareNextEpoch` fulfilled the intent for epoch N+1 mid-epoch, `prepareCurrentEpoch` on the first tick of epoch N+1 sees `fulfilled = true` and skips the refetch.

This creates a gap on epoch-boundary reorgs. The scheduler's epoch-transition check (`HandleHeadEvent`) only compares old `currentDutyDependentRoot` vs new `PreviousDutyDependentRoot`. It never inspects the new epoch's `CurrentDutyDependentRoot` at transition. So a boundary reorg that changes the last block of the previous epoch — changing the new epoch's `current_duty_dependent_root` (which proposer duties depend on) without changing `previous_duty_dependent_root` — emits no scheduler signal. Pre-fetched proposer duties remain stale.

The root cause is in the scheduler (pre-existing gap), but the old proposer code was immune because it always refetched at epoch start. The new intent mechanism exposes the gap.

**Impact:** One epoch of stale proposer duties on a boundary reorg. Missed proposal at worst, not slashing-critical. Boundary reorgs are rare on mainnet. Subsequent epochs are unaffected.

**Suggested fix** — force-invalidate the current epoch intent at epoch start:

```go
// Clean up the irrelevant data to prevent infinite memory growth at the 1st slot of the epoch.
if slotNumber == 1 && currentEpoch >= 1 {
    h.duties.EraseEpochData(currentEpoch - 1)
    delete(h.dutyFetchIntents, currentEpoch-1)
    // Force refetch at epoch start to cover boundary reorgs that the
    // scheduler cannot detect (CurrentDutyDependentRoot changed at
    // transition without PreviousDutyDependentRoot changing).
    h.dutyFetchIntents[currentEpoch] = false
}
```

One extra beacon node call per epoch (~6.4 min), negligible cost. Restores the old code's safety while keeping retry-on-failure.

operator/duties/proposer_test.go
```diff
@@ -388,97 +466,581 @@ func TestScheduler_Proposer_Reorg_Current(t *testing.T) {
...
+		waitForDutiesExecution(t, fetchDutiesCall, executeDutiesCall, timeout, expected)
+
+		// Stop scheduler & wait for graceful exit.
+		cancel()
+		require.NoError(t, scheduler.Wait())
+		ticker.WaitShutdown()
+	})
+}
+
+func TestScheduler_Proposer_Fetch_Execute_Next_Epoch_Duty(t *testing.T) {
```
olegshmuelov, line 1529: nit: `TestScheduler_Proposer_Fetch_Execute_Next_Epoch_Duty` has two steps labeled "STEP 2".

operator/duties/proposer.go
iurii-ssv: Has been addressed in https://github.com/ssvlabs/ssv/pull/2760

operator/duties/attester.go
```diff
@@ -187,13 +189,13 @@ func (h *AttesterHandler) HandleDuties(ctx context.Context) {
 				defer cancel()
 
 				// 1) Declare intents.
-				if !reorgEvent.Current {
-					// Reorg on the previous epoch means the duties for the current epoch might have changed, so
-					// we want to re-fetch them.
+				// Depending on whether the current or previous duty dependent root has changed, we might need to
+				// re-fetch the duties for the current epoch (if the current duty dependent root has changed) and/or
+				// next epoch (if the previous duty dependent root has changed) to ensure we have the up-to-date
+				// duties for all validators for both epochs.
```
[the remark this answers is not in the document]
iurii-ssv: Has been addressed in https://github.com/ssvlabs/ssv/pull/2760

operator/duties/proposer.go
iurii-ssv: > The new intent-based mechanism trusts pre-fetched duties: if `prepareNextEpoch` fulfilled the intent for epoch N+1 mid-epoch, `prepareCurrentEpoch` on the first tick of epoch N+1 sees `fulfilled = true` and skips the refetch.

This is correct, the intent needs to be fulfilled once (we record this to prevent unnecessary fetches).

Also, need to clarify the way we detect reorgs in `HandleHeadEvent`, giving it a fresh look I think I misunderstood how exactly it is supposed to function ... so I've made some adjustments there that should get us on the same page (hopefully clarifying the code involved) https://github.com/ssvlabs/ssv/pull/2767

> This creates a gap on epoch-boundary reorgs. The scheduler's epoch-transition check (`HandleHeadEvent`) only compares old `currentDutyDependentRoot` vs new `PreviousDutyDependentRoot`. It never inspects the new epoch's `CurrentDutyDependentRoot` at transition. So a boundary reorg that changes the last block of the previous epoch — changing the new epoch's `current_duty_dependent_root` (which proposer duties depend on) without changing `previous_duty_dependent_root` — emits no scheduler signal. Pre-fetched proposer duties remain stale.

Great find! I see what you mean now, it's a tricky one ... just to re-phrase it for posterity, what you are saying is basically: "the reorg-detection implemented in `HandleHeadEvent` is not strong enough to capture all possible reorg-scenarios that can happen upon the very 1st slot of an epoch", right ?

The suggested fix would work. I would however add another step to it - to act on that intent right away, instead of just declaring it as unfulfilled, that should prevent duty-fetch from happening at the start of the 2nd slot which would delay the duty execution due to `prepareCurrentEpoch` resolving that intent, for every epoch) ... Even then, it suggests to me that `AttesterHandler` would also completely miss a `ReorgEvent` (the one targeting the next epoch re-fetch) in a similar manner - which could be addressed by a similar work-around ... but I'm wondering if there is a way to address it in the `HandleHeadEvent` func (which is the root-cause of it all).

To address it in the `HandleHeadEvent` func, maybe we could keep track of the "last block root" so we could compare `CurrentDutyDependentRoot` against it during epoch-transitions ?

iurii-ssv: > To address it in the `HandleHeadEvent` func, maybe we could keep track of the "last block root" so we could compare `CurrentDutyDependentRoot` against it during epoch-transitions ?

I've sketched a PR to do that https://github.com/ssvlabs/ssv/pull/2768 (it's based on https://github.com/ssvlabs/ssv/pull/2767) - @olegshmuelov do you think this will work ?

olegshmuelov: Yes, this works. Fixing it at the scheduler level is cleaner, covers both proposer and attester with a single mechanism, and the handlers stay simple.

This also resolves the concern about acting on the intent immediately vs just declaring it unfulfilled: since the scheduler now emits a proper `ReorgEvent` at epoch transition, the handler's reorg path fires naturally and processes the intent on the same tick. No per-handler workaround needed.

I'll review #2767 and #2768.

operator/duties/proposer_test.go
iurii-ssv: https://github.com/ssvlabs/ssv/pull/2756/changes/82ef13b0feb2ba5a325abe4eca0737692d27d396

operator/duties/proposer.go
iurii-ssv: Merged both of the PRs mentioned above into the current branch, those should resolve this finding fully now

iurii-ssv:
@greptile could you re-review this PR ?

operator/duties/scheduler.go
```diff
@@ -363,65 +367,63 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
+			expectedCurrentDutyDependentRoot := s.currentDutyDependentRoot[:]
+			expectedPreviousDutyDependentRoot := s.previousDutyDependentRoot[:]
+			if currentEpoch > s.lastEpoch {
+				// Epoch transition case:
+				// - the root tracked in s.currentDutyDependentRoot now describes the previous epoch, hence becomes
+				//   the expected previous dependent root
+				// - we use the latest observed block-root from the previous epoch as the expected current dependent
+				//   root since it's the only thing we can compare the current dependent root we got with
+				epochTransition = true
+				expectedCurrentDutyDependentRoot = s.lastBlockRoot[:]
```
momosh-ssv, line 405: Might be worth noting that if the last slot of the old epoch was missed (no block produced), `lastBlockRoot` holds an earlier block's root — which won't match the beacon node's `CurrentDutyDependentRoot` for the new epoch, so this fires a spurious reorg.

The consequence is just an extra fetch (safe, not harmful), but maybe a short comment here would save the next reader from puzzling over why they see unexpected reorg logs after missed slots at epoch boundaries?

operator/duties/scheduler.go
```diff
@@ -363,65 +367,63 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
+			// We don't handle future events to keep things simple.
+			logger.Warn("got future head event from EL, most likely cause is clock-skew between SSV node and EL")
+			return
+		}
+
+		// Check for reorg & fire corresponding ReorgEvent if needed.
+		if s.lastEpoch != 0 {
+			zeroRoot := new(phase0.Root)[:]
+
+			epochTransition := false
```
momosh-ssv: What do you think about inlining this? `epochTransition` was removed from `ReorgEvent`, so this variable is only used in the debug log on line 412. Could just pass the `currentEpoch > s.lastEpoch` expression directly into the log field to avoid the dead-looking local.

operator/duties/scheduler.go
iurii-ssv: If the slot was truly missed (not just by our own EL - by the majority of Ethereum nodes), the `CurrentDutyDependentRoot` should actually refer to the slot prior to the missed last slot of the epoch (and it's recursive if multiple slots at the end of an epoch are missing) ... so "`lastBlockRoot` holds an earlier block's root" is the expected/desired behavior

to clarify for posterity, the spurious reorg can happen in case we got multiple blocks (events) for the same slot, and the last one of them (the one we record as `lastBlockRoot`) turned out to be different from what the majority of Ethereum nodes think it should be (which is what we'll receive in `CurrentDutyDependentRoot`, aka canonical root for the epoch) - in which case:
- we either never saw/received the block that ended up being the canonical root ... which would be a **valid** reorg
- or we got it, but overrode with another block that came next (within the same slot) ... which would be a **spurious** reorg

so yeah, spurious is possible but isn't harmful (since we should have plenty of time to handle the reorg by doing that duty-re-fetch before the next tick comes)

operator/duties/scheduler.go
iurii-ssv: I think it's better to replace the `if currentEpoch > s.lastEpoch` with `if epochTransition` instead actually - this would remove duplication, and clean it up a bit - https://github.com/ssvlabs/ssv/pull/2756/changes/921631161b19f8b8f0ed828a7275805a430006a8

operator/duties/scheduler.go
iurii-ssv: I've expanded that comment to add some details - https://github.com/ssvlabs/ssv/pull/2756/changes/921631161b19f8b8f0ed828a7275805a430006a8 https://github.com/ssvlabs/ssv/pull/2756/changes/56388fdcd928e8f64857d9234313e83c187252df

operator/duties/scheduler.go
```diff
@@ -363,65 +369,65 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
+			expectedCurrentDutyDependentRoot := s.currentDutyDependentRoot[:]
+			expectedPreviousDutyDependentRoot := s.previousDutyDependentRoot[:]
+			if epochTransition {
+				// Epoch transition case:
+				// - the root tracked in s.currentDutyDependentRoot now describes the previous epoch, hence becomes
+				//   the expected previous dependent root
+				// - we use the latest observed block-root from the previous epoch as the expected current dependent
+				//   root since it's the only thing we can compare the current dependent root we got with. This might
+				//   produce a spurious (false-positive) reorg in case when another event overrode the canonical root
+				//   observed (and recorded as lastBlockRoot) during pior event(s) - but it works OK for us regardless.
```
olegshmuelov: pior -> prior typo

operator/duties/scheduler.go
```diff
@@ -363,65 +369,65 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
+		}
+		if event.Slot > s.beaconConfig.EstimatedCurrentSlot() {
+			// We don't handle future events to keep things simple.
+			logger.Warn("got future head event from EL, most likely cause is clock-skew between SSV node and EL")
+			return
+		}
+
+		// Check for reorg & fire corresponding ReorgEvent if needed.
+		if s.lastEpoch != 0 {
+			zeroRoot := new(phase0.Root)[:]
```
olegshmuelov: nit: idiomatic Go version is simpler
```go
var zeroRoot phase0.Root
```

olegshmuelov (dismissed):
few nit comments

olegshmuelov (dismissed):
LGTM!

operator/duties/proposer.go
```diff
@@ -84,78 +96,200 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-		span.SetStatus(codes.Error, err.Error())
-		return
+	if fulfilled, ok := h.dutyFetchIntents[currentEpoch]; ok && !fulfilled {
+		logger.Debug("fetching duties for the current epoch")
+
+		err := h.fetchAndProcessDuties(ctx, logger, currentEpoch, currentSlot)
+		if err != nil {
+			logger.Error("fetching duties for the current epoch failed", zap.Error(err))
+			span.SetStatus(codes.Error, err.Error())
+			return
```
momosh-ssv, line 261: What happens if `fetchAndProcessDuties` fails on the last slot of the epoch? The intent stays unfulfilled (`false`), but the next tick is in the new epoch, which cleans up `currentEpoch-1` data (line 112-113) — so the unfulfilled intent gets deleted without ever being retried.

Separately, for next-epoch intents: `prepareNextEpoch` gates on `shouldFetchNextEpoch(currentSlot)`, so if a reorg-triggered fetch fails early in the epoch, the retry is deferred until the second half. Might be worth letting already-failed intents bypass that gate.

operator/duties/proposer.go
```diff
@@ -84,78 +96,200 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
-	if err := h.fetchAndProcessDuties(ctx, epoch); err != nil {
-		// Set empty duties to inform DutyStore that fetch for this epoch is done.
-		h.duties.Set(epoch, []dutystore.StoreDuty[eth2apiv1.ProposerDuty]{})
-		h.logger.Error("failed to fetch duties for current epoch", zap.Error(err))
-		span.SetStatus(codes.Error, err.Error())
-		return
+	if fulfilled, ok := h.dutyFetchIntents[currentEpoch]; ok && !fulfilled {
+		logger.Debug("fetching duties for the current epoch")
+
+		err := h.fetchAndProcessDuties(ctx, logger, currentEpoch, currentSlot)
```
momosh-ssv, line 257: I fear the old code's `h.duties.Set(epoch, []{})` on error was serving a purpose that's now lost. If a reorg sets the intent to `false` but the refetch fails here, the duty store still holds the pre-reorg duties. On the next tick, `processExecution` runs *before* `prepareCurrentEpoch` retries — so it can execute stale pre-reorg duties.

Maybe we should erase the epoch data when declaring the intent unfulfilled (e.g., in the reorg handler alongside `h.dutyFetchIntents[currentEpoch] = false`), so stale duties can't be picked up while a refetch is pending?

operator/duties/proposer.go
```diff
@@ -84,78 +96,200 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+				// to currentSlot+1 as that might be too short (hence setting it to currentSlot+2).
+				reorgCtx, cancel := context.WithDeadline(ctx, h.beaconConfig.SlotStartTime(currentSlot+2))
+				defer cancel()
 
-			h.indicesChanged = true
+				// 1) Declare intents.
+				if refetchCurrentEpoch {
+					h.dutyFetchIntents[currentEpoch] = false
+				}
+				h.dutyFetchIntents[nextEpoch] = false
```
momosh-ssv, line 190: Should we consider skipping the next-epoch invalidation when only `PreviousDutyDependentRootChanged` is set? Proposer duties depend on the *current* root, so a previous-only reorg shouldn't affect them. The unconditional invalidation is safe but adds an unnecessary beacon fetch during reorgs — which is exactly when the beacon node is already under pressure.

operator/duties/proposer_test.go
```diff
@@ -388,97 +466,591 @@ func TestScheduler_Proposer_Reorg_Current(t *testing.T) {
...
```
momosh-ssv, line 1529: Any reason we're not testing the fetch-failure + retry-on-next-tick path? Seems like a key behavioral change — the old code set empty duties on error, the new code leaves the intent unfulfilled for retry. A test that simulates a beacon node error on first fetch, verifies no duties execute, then succeeds on retry would validate this mechanism.

operator/duties/proposer.go
```diff
@@ -84,78 +96,200 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+					// When at epoch boundary, we only care about pre-fetching & preparing the duties for the next
+					// epoch (the current epoch will have been passed upon the next slot-tick). Otherwise, pre-fetch &
+					// prepare the duties for the current epoch.
+					if h.atLastSlotOfCurrentEpoch(currentSlot) {
+						delete(h.dutyFetchIntents, currentEpoch) // optimization: prune irrelevant intent
+						h.prepareNextEpoch(tickCtx, logger, currentEpoch, currentSlot)
+					} else {
+						h.prepareCurrentEpoch(tickCtx, logger, currentEpoch, currentSlot)
 					}
+				case <-time.After(time.Until(indicesChangeDeadline)):
```
momosh-ssv, line 141: Worth noting that by the time execution reaches this `select` (after `prepareCurrentEpoch` + `processExecution` + `prepareNextEpoch` + cleanup), `time.Until(indicesChangeDeadline)` can be negative — making `time.After` fire immediately, so indices changes are never checked on this tick.

That's the intended fallback ("too late"), but under sustained load (slow beacon, many validators), this could systematically defer indices changes every tick. Might be worth monitoring via the OTel spans to see how often this branch is hit in practice.

operator/duties/proposer.go
iurii-ssv: > What happens if fetchAndProcessDuties fails on the last slot of the epoch? The intent stays unfulfilled (false), but the next tick is in the new epoch, which cleans up currentEpoch-1 data (line 112-113) — so the unfulfilled intent gets deleted without ever being retried.

Say the current epoch is N,

for the current epoch (`fetchAndProcessDuties` fails inside of `prepareCurrentEpoch` on the last slot of the current epoch): on the next tick we are already in the epoch N+1 and we no longer care about duties for epoch N at all (we shouldn't even try to execute them) ... so cleaning up intents for past epochs is OK/desirable;

for the next epoch (`fetchAndProcessDuties` fails inside of `prepareNextEpoch` on the last slot of the current epoch): on the next tick we are already in the epoch N+1 and we retry fetching duties with `h.prepareCurrentEpoch(tickCtx, logger, currentEpoch, currentSlot)` until epoch N+1 ends

> Separately, for next-epoch intents: `prepareNextEpoch` gates on `shouldFetchNextEpoch(currentSlot)`, so if a reorg-triggered fetch fails early in the epoch, the retry is deferred until the second half. Might be worth letting already-failed intents bypass that gate.

there is no point trying to fetch duties for the next epoch **early** into current epoch, we'll have plenty of time for that during the late phase of the current epoch (which `shouldFetchNextEpoch` check is for) ... the important part is that reorg will **declare** the intent, so the intent itself will not be lost (and the deferred fulfillment is by design)

operator/duties/proposer.go
iurii-ssv: Your reasoning is sound, however, I think the code in this PR handles it better because:
- we call `h.processExecution` exactly 1 time per slot, this means we either execute pre-reorg duties or post-reorg duties (never both for the same slot)
- since **reorg** might be false-positive (we never know for sure until the slot gets pretty old) it's always better to execute some duties vs just skipping them - because there is a chance pre-reorg duties will turn out to be correct

so there is really no downside/danger to executing pre-reorg duties, as far as I know.

operator/duties/proposer.go
iurii-ssv: I believe `PreviousDutyDependentRootChanged` also means (pretty much always) `CurrentDutyDependentRootChanged` will be different (since in a blockchain every recent block depends on every other block from past history, assuming they are part of the same chain) ... so that translates into "The next epoch is always re-fetched on any reorg" which is implicitly equal to checking for `if CurrentDutyDependentRootChanged || PreviousDutyDependentRootChanged` (this would always be true for any reorg event a handler receives),

cc @olegshmuelov if you have anything to add, or correct me

operator/duties/proposer.go
iurii-ssv: That's correct, "slow beacon" could delay the handling of `indicesChange`, but if it does: 
1) we likely have a more pressing issue on our hands than updating validator-metadata 
2) I think delaying the handling for several slots (maybe several epochs even) is not critical since we should get `indicesChange` event well before it is "expected to be fully processed" (`ReactivateCluster` + SSV node startup seem to be the only issuers of this event)

that is to say, I don't think monitoring for the delays in this case is needed (probably not worth the additional complexity)

operator/duties/proposer.go
olegshmuelov: Correct. To add precision: block roots are hash-chained, so if the canonical root at epoch N-2 changed, every block in epoch N-1 has a different parent hash and its root must also change. `PreviousDutyDependentRootChanged` implies `CurrentDutyDependentRootChanged` as a cryptographic invariant. The "only Previous changed" scenario can't happen, so the unconditional refetch is effectively identical to gating on either flag.

operator/duties/proposer_test.go
iurii-ssv: Mostly because these tests are quite complex to write & maintain, so it wasn't clear whether it's valuable to invest time into that.

But now that the approach we've chosen in this PR has got some validation & reviews, we can probably add more tests ... will do + post an update here ...

operator/duties/proposer.go
```diff
@@ -70,12 +75,19 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
 
 		case <-next:
 			currentSlot := h.ticker.Slot()
-			next = h.ticker.Next()
+			next = h.ticker.Next() // advances h.ticker
```
nkryuchkov, line 78: out of scope of the PR, but should we rename `Next` to `Advance`?

operator/duties/proposer.go
```diff
@@ -84,78 +96,200 @@ func (h *ProposerHandler) HandleDuties(ctx context.Context) {
...
+				// Process intents (if any): fetch & prepare the duties for the current epoch.
+				h.prepareCurrentEpoch(tickCtx, logger, currentEpoch, currentSlot)
+
+				h.processExecution(tickCtx, currentEpoch, currentSlot)
+
+				// Process intents (if any): fetch & prepare the duties for the next epoch.
+				h.prepareNextEpoch(tickCtx, logger, currentEpoch, currentSlot)
+
+				// Clean up the irrelevant data to prevent infinite memory growth at the 1st slot of the epoch.
+				if slotNumber == 1 && currentEpoch >= 1 {
```
nkryuchkov, line 111: If you mean the first slot of an epoch, should this be `slotNumber == 0`?

nkryuchkov (dismissed):

operator/duties/proposer.go
iurii-ssv: `slotNumber` here is in range [1-32], not [0-31] - it's an "informal slot number" rather than strict slot index (it's used to build the `buildStr` as well),

I don't have a strong preference, we can use [0-31] range if you like it more.

operator/duties/proposer.go
iurii-ssv: Good idea, this would document the assumptions better, we can do it as a follow-up a bit later - https://github.com/ssvlabs/ssv/issues/2784

operator/duties/proposer_test.go
iurii-ssv: Added more tests (both for proposer & attester): https://github.com/ssvlabs/ssv/pull/2756/changes/b9100b3897d65d4684a28ae3a795407e6e829976 https://github.com/ssvlabs/ssv/pull/2756/changes/e24bead44046dcabdb2ee9407f2787efccce03ed https://github.com/ssvlabs/ssv/pull/2756/changes/3bb936863a7a36f87d7736b3c8dbf4456773aa97

operator/duties/proposer.go
nkryuchkov: Ah, I see it already was `+ 1` before your PR. Can we add a comment here explaining why we are comparing with `1`? Perhaps it's redundant but it might reduce any possible confusion

iurii-ssv: For this case I would say a comment would be a bit redundant since the `slotNumber` is defined just several lines above. If it was passed down as an argument from somewhere else - then indeed would be nice to document the assumption ... but even so I'm not sure how to phrase the comment to clarify the difference, I've changed it like this now: `at the 1st slot of the epoch` -> `at the very 1st slot of the epoch`.

**Update:** expanded the comment to clarify further:
```go
// Clean up the irrelevant data to prevent infinite memory growth at the very 1st slot of the epoch.
// Note, it doesn't have to be "the very 1st slot" exactly - it's just the most natural time to do it.
if slotNumber == 1 && currentPeriod >= 1 {
```

olegshmuelov (dismissed):

nkryuchkov (approved):

operator/duties/proposer.go
momosh-ssv: Maybe worth adding a single debug log (or `span.AddEvent`) on the deferred branch? Agreed monitoring infrastructure isn't worth the complexity, but right now the "too late" case is silent — if it ever manifests in production we'd have no signal to work from. ~2 lines, no real cost.

momosh-ssv (approved):
nice, gj!

operator/duties/proposer.go
iurii-ssv: The reason I'm not sure if it's worth adding is that it doesn't seem like "just 2 additional lines of code":
- if we enter `case <-time.After(time.Until(indicesChangeDeadline)):` - the func just returns (no code executes until the next tick)
- in order to "log that edge-case" we need to keep monitoring the channel until the next tick (and terminate our monitoring upon the next tick)
- to do that, we either need to add some additional code that calculates "how long to monitor for" and times out after that; Or we could spin up a go-routine just for logging

so I can add this kind of handling there, but given how unlikely/low-priority the "skipping of indicies change" is ... I would not recommend adding all this complex handling in there (simply because the code already contains a lot of other complex interactions in it - probably not worth complicating it further for something that's not super relevant)

PR #2762, opened 2026-03-29
Remove duplication in Runner delegation code

julienh-ssv:
**Fix for:**

F-ssv-034 P1
~600 lines of identical Runner delegation boilerplate across 5 runners
ssv · Code Quality — Code Duplication · Assigned: dev4 · Effort: —
Each Runner has ~15 identical single-line methods delegating to `r.BaseRunner.X()`: `GetRole()`, `GetLastHeight()`, `GetLastRound()`, `GetStateRoot()`, `SetTimeoutFunc()`, etc. Could be eliminated with Go struct embedding.

Impact: ~600 lines of pure boilerplate. Every new delegation method must be added to all 5 runners.

-------------------

**What changed:**

- Most changes are replacing r.BaseRunner.X forwarders with embedded *BaseRunner method promotion.
- One intentional semantic change is a correctness fix: Decode() now actually populates the receiver (previous json.Unmarshal(data, &b) / &r was wrong for pointer receivers). See protocol/v2/ssv/runner/runner.go:168.
- JSON/root stability is accounted for via custom MarshalJSON/UnmarshalJSON on runners where embedding would have changed the encoded shape (and ValCheck is intentionally not restored on decode).

### DIFF
--- a/protocol/v2/ssv/runner/aggregator.go
+++ b/protocol/v2/ssv/runner/aggregator.go
@@ -125,14 +125,14 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 
 	// reconstruct selection proof sig
 	span.AddEvent("reconstructing beacon signature", trace.WithAttributes(observability.BeaconBlockRootAttribute(root)))
-	fullSig, err := r.state().ReconstructBeaconSig(r.state().PreConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
+	fullSig, err := r.State.ReconstructBeaconSig(r.State.PreConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
 	if err != nil {
 		// If the reconstructed signature verification failed, fall back to verifying each partial signature
-		r.FallBackAndVerifyEachSignature(r.state().PreConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
+		r.FallBackAndVerifyEachSignature(r.State.PreConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
 		return fmt.Errorf("got pre-consensus quorum but it has invalid signatures: %w", err)
 	}
 
-	duty := r.state().CurrentDuty.(*spectypes.ValidatorDuty)
+	duty := r.State.CurrentDuty.(*spectypes.ValidatorDuty)
 	span.SetAttributes(
 		observability.CommitteeIndexAttribute(duty.CommitteeIndex),
 		observability.ValidatorIndexAttribute(duty.ValidatorIndex),
@@ -142,7 +142,7 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 	// to perform this aggregation duty or not
 	ok := r.IsAggregator(r.NetworkConfig.TargetAggregatorsPerCommittee, duty.CommitteeLength, fullSig)
 	if !ok {
-		r.state().Finished = true
+		r.State.Finished = true
 		r.measurements.EndDutyFlow()
 		recordTotalDutyDuration(ctx, r.measurements.TotalDutyTime(), spectypes.RoleAggregator, 0)
 		return nil
@@ -295,17 +295,17 @@ func (r *AggregatorRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 	root := roots[0]
 
 	span.AddEvent("reconstructing beacon signature", trace.WithAttributes(observability.BeaconBlockRootAttribute(root)))
-	sig, err := r.state().ReconstructBeaconSig(r.state().PostConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
+	sig, err := r.State.ReconstructBeaconSig(r.State.PostConsensusContainer, root, r.GetShare().ValidatorPubKey[:], r.GetShare().ValidatorIndex)
 	if err != nil {
 		// If the reconstructed signature verification failed, fall back to verifying each partial signature
-		r.FallBackAndVerifyEachSignature(r.state().PostConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
+		r.FallBackAndVerifyEachSignature(r.State.PostConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
 		return fmt.Errorf("got post-consensus quorum but it has invalid signatures: %w", err)
 	}
 	specSig := phase0.BLSSignature{}
 	copy(specSig[:], sig)
 
 	cd := &spectypes.ValidatorConsensusData{}
-	err = cd.Decode(r.state().DecidedValue)
+	err = cd.Decode(r.State.DecidedValue)
 	if err != nil {
 		return fmt.Errorf("could not decode consensus data: %w", err)
 	}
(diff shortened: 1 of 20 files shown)


protocol/v2/ssv/runner/aggregator.go
olegshmuelov: Nit: with `*BaseRunner` embedded, this forwarding method is redundant, the promoted `BaseRunner.OnTimeoutQBFT` already satisfies the `Runner` interface. Same applies to the identical method in `proposer.go`, `sync_committee_contribution.go`, `validator_registration.go`, `voluntary_exit.go`, and `committee.go`.
Can be removed for consistency with the other forwarding methods that were cleaned up.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -449,24 +444,20 @@ func (r *AggregatorRunner) GetNetwork() specqbft.Network {
...
+	for _, share := range r.Share {
 		return share
 	}
 	return nil
 }
 
 func (r *AggregatorRunner) state() *State {
-	return r.BaseRunner.State
+	return r.State
 }
```
olegshmuelov: With `*BaseRunner` embedded, `r.State` is directly accessible, this `state()` helper is now redundant delegation, which is exactly what this PR removes. Same applies to `proposer.go`, `sync_committee_contribution.go`, `validator_registration.go`, and `voluntary_exit.go`. Consider inlining `r.State` at call sites and removing these methods for consistency.

protocol/v2/ssv/runner/runner_decode_test.go
```diff
@@ -0,0 +1,145 @@
...
+	)
+	require.NoError(t, err)
+
+	data, err := r.Encode()
+	require.NoError(t, err)
+	require.Contains(t, string(data), "\"ValCheck\"")
+
+	var decoded AggregatorRunner
+	require.NoError(t, decoded.Decode(data))
+	decoded.NetworkConfig = cloneTestNetworkConfig()
```
olegshmuelov, line 70: Worth noting: decoded runners are intentionally partial, both `NetworkConfig` and `ValCheck` must be rehydrated after decode. This is pre-existing behavior, but now that you've formalized it with explicit `UnmarshalJSON` + tests, consider adding a brief doc comment on the `Runner` interface's `Decode` method (or on `BaseRunner.Decode`) stating this contract explicitly, so future callers don't assume a fully functional runner post-decode.

protocol/v2/ssv/runner/runner.go
```diff
@@ -124,6 +124,10 @@ func (b *BaseRunner) GetShares() map[phase0.ValidatorIndex]*spectypes.Share {
 	return b.Share
 }
 
+func (b *BaseRunner) HasRunningDuty() bool {
+	return b.hasRunningDuty()
```
nkryuchkov: Can we have only `HasRunningDuty`?

protocol/v2/ssv/runner/runner.go
```diff
@@ -149,6 +153,10 @@ func (b *BaseRunner) GetStateRoot() ([32]byte, error) {
 	return b.State.GetRoot()
 }
 
+func (b *BaseRunner) GetNetworkConfig() *networkconfig.Network {
```
nkryuchkov: Why do we need it? Can we just use `b.NetworkConfig`?

nkryuchkov (dismissed):

protocol/v2/ssv/runner/runner.go
iurii-ssv: This probably applies to other such getters (eg. `GetStateRoot()`) - we probably should remove the ones that aren't doing any "additional" adjustments, and access those struct-fields directly (unless it provides us a significantly shorter alias, such as `state()` was doing previuosly)

iurii-ssv (dismissed):
Nice!

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -449,63 +440,55 @@ func (r *AggregatorRunner) GetNetwork() specqbft.Network {
...
-	return r.BaseRunner.GetStateRoot()
-}
+	aux := &aggregatorRunnerJSON{}
+	if err := json.Unmarshal(data, aux); err != nil {
+		return err
+	}
 
-func (r *AggregatorRunner) SetTimeoutFunc(fn TimeoutF) {
-	r.BaseRunner.SetTimeoutFunc(fn)
+	r.BaseRunner = aux.BaseRunner
```
momosh-ssv, line 494: What happens if the JSON input has `"BaseRunner": null` or omits the key entirely? `aux.BaseRunner` would be nil, this line assigns nil to the embedded pointer, and any subsequent promoted method call (`HasRunningDuty()`, `GetRole()`, etc.) panics. Current tests pass because test data is always well-formed, but this is a real boundary that deserves a guard.

Maybe we could validate `aux.BaseRunner != nil` and return an error before assigning. Same applies to the identical `UnmarshalJSON` in `proposer.go`, `sync_committee_contribution.go`, `validator_registration.go`, and `voluntary_exit.go`.

protocol/v2/ssv/runner/runner.go
```diff
@@ -157,8 +165,17 @@ func (b *BaseRunner) Encode() ([]byte, error) {
 	return json.Marshal(b)
 }
 
+// Decode unmarshals persisted runner state into the receiver.
+//
+// Note: decoded runners are intentionally partial; runtime dependencies (e.g. `NetworkConfig`, `TimeoutF`, and
+// runner-specific value checkers) must be rehydrated by the caller after decode.
 func (b *BaseRunner) Decode(data []byte) error {
-	return json.Unmarshal(data, &b)
+	if b == nil {
```
momosh-ssv, line 175: I fear this nil guard is unreachable. In Go, calling a method on a nil pointer receiver dereferences the pointer before the method body executes — so `b.Decode(...)` on a nil `*BaseRunner` panics at the call site, never reaching this check. It gives a false sense of safety.

Maybe just remove it and document the non-nil precondition at call sites instead, or return a more descriptive error in the doc comment so callers know they must construct a non-nil `BaseRunner` before calling `Decode`.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -449,63 +440,55 @@ func (r *AggregatorRunner) GetNetwork() specqbft.Network {
...
-func (r *AggregatorRunner) GetRole() spectypes.RunnerRole {
-	return r.BaseRunner.GetRole()
-}
+func (r *AggregatorRunner) MarshalJSON() ([]byte, error) {
+	type aggregatorRunnerJSON struct {
+		BaseRunner *BaseRunner `json:"BaseRunner"`
+		// ValCheck is intentionally marshaled to preserve the historical runner state JSON shape
+		// (and thus runner state roots used by spec tests). It is a runtime-only dependency and
+		// is ignored on decode.
+		ValCheck any `json:"ValCheck"`
```
momosh-ssv, line 470: Worth considering: `ValCheck any` causes `json.Marshal` to serialize the concrete runtime type — which in production holds references to beacon clients, share maps, etc. Since the value is discarded on decode anyway, the wire bytes vary depending on runtime state with no benefit.

Should we marshal `ValCheck` as `nil` (or a fixed sentinel like `json.RawMessage("null")`) instead? That makes the output deterministic and avoids accidentally capturing runtime references. Same applies to `proposer.go` and `sync_committee_contribution.go`.

protocol/v2/ssv/runner/runner_decode_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	require.Equal(t, spectypes.RoleAggregator, decoded.GetRole())
+	require.False(t, decoded.HasRunningDuty())
+	require.Len(t, decoded.GetShares(), 1)
+	require.Nil(t, decoded.ValCheck)
```
momosh-ssv, line 79: Should we also verify state root roundtrip here (`Encode(decoded) == Encode(original)`)? That is the property that actually matters for spec test compliance. The `VoluntaryExit` decode test does check root equality — seems like `AggregatorRunner`, `ProposerRunner`, and `SyncCommitteeAggregatorRunner` should too.

momosh-ssv (commented):
Seems good, have 4 questions.

protocol/v2/ssv/runner/runner.go
iurii-ssv: > In Go, calling a method on a nil pointer receiver dereferences the pointer before the method body executes

Are you sure this is correct ? Always thought in Go you could call a methods on a typed-nil (unless the method is called/defined with a value-receiver, I guess)

julienh-ssv: In Go, calling a method with a pointer receiver on a typed-nil pointer is valid; the receiver can be nil inside the method. The guard is reachable and keeps the error clearer than relying on `json.Unmarshal(nil *T)`

julienh-ssv: @momosh-ssv Go only dereferences the pointer when calling a method with a **value receiver**, not a **pointer receiver**.

protocol/v2/ssv/runner/runner.go
julienh-ssv: fixing

iurii-ssv (approved):
Great!

olegshmuelov (approved):

ljuba-ssv (approved):

PR #2765, opened 2026-03-30
encrypt persisted p2p network key

nkryuchkov:
Fixes `F-ssv-185`.

## Summary

Encrypt the node's persisted P2P network private key instead of storing raw key bytes in the DB.

## What changed

- add encrypted `enc:` storage format for the persisted P2P network key
- keep backward compatibility with legacy plaintext DB entries
- lazily migrate legacy plaintext keys to encrypted storage when protection is available
- use local operator-key-derived encryption in local signer mode
- use `ssv-signer` `/v1/operator/encrypt` and `/v1/operator/decrypt` in remote signer mode
- keep legacy plaintext fallback with a warning when no compatible protection source exists
- fail startup clearly if the DB already contains an encrypted network key but the configured `ssv-signer` is too old to decrypt it
- update root `go.mod` to the latest `ssvsigner` version with operator encrypt/decrypt support

## Migration

- Existing plaintext DB entries remain supported.
- On first startup with a compatible protection source, the node lazily rewrites the stored P2P network key from legacy plaintext to encrypted `enc:` format.
- Local operator-key deployments encrypt automatically after upgrade.
- Remote-signer deployments encrypt automatically only when `ssv-signer` is upgraded first to a version that supports `/v1/operator/encrypt` and `/v1/operator/decrypt`.
- If `ssv-signer` is still old, the node keeps using legacy plaintext storage with a warning and does not break existing setups.
- Once the DB entry is migrated to `enc:` format, older node versions cannot read it and downgrading is not supported.

## Rollback to a previous version

If startup fails because the DB contains an encrypted persisted P2P network key, recover by setting the previous raw `NETWORK_PRIVATE_KEY`. If that key is not available, wipe/reset the DB and let a new P2P key be generated.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"bytes"
 	"context"
+	crand "crypto/rand"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
@@ -1127,7 +1128,10 @@ func probeRemoteNetworkKeyProtector(
 	ctx context.Context,
 	client *ssvsigner.Client,
 ) error {
-	probeKey := []byte("p2p-network-key-probe")
+	probeKey := make([]byte, 32)
+	if _, err := crand.Read(probeKey); err != nil {
+		return fmt.Errorf("generate remote data protector probe: %w", err)
+	}
 	encrypted, err := client.OperatorEncrypt(ctx, probeKey)
 	if err != nil {
 		return fmt.Errorf("probe remote data protector encrypt: %w", err)
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -446,7 +446,7 @@ var StartNodeCmd = &cobra.Command{
 		cfg.P2pNetworkConfig.MessageValidator = messageValidator
 		cfg.SSVOptions.ValidatorOptions.MessageValidator = messageValidator
 
-		p2pNetwork := setupP2P(logger, db)
+		p2pNetwork := setupP2P(cmd.Context(), logger, db, operatorPrivKey, ssvSignerClient)
 
 		cfg.SSVOptions.Context = cmd.Context()
 		cfg.SSVOptions.DB = db
@@ -1071,9 +1071,46 @@ func setupSSVNetwork(logger *zap.Logger) (*networkconfig.SSV, error) {
 	return ssvConfig, nil
 }
 
-func setupP2P(logger *zap.Logger, db basedb.Database) network.P2PNetwork {
-	istore := ssv_identity.NewIdentityStore(logger, db)
-	netPrivKey, err := istore.SetupNetworkKey(cfg.NetworkPrivateKey)
+func setupP2P(ctx context.Context, logger *zap.Logger, db basedb.Database, operatorPrivKey keys.OperatorPrivateKey, signerClient *ssvsigner.Client) network.P2PNetwork {
+	var protectFn func(context.Context, []byte) ([]byte, error)
+	var unprotectFn func(context.Context, []byte) ([]byte, error)
+
+	// The configured startup mode determines how the persisted P2P network key is protected.
+	if operatorPrivKey != nil {
+		encryptionKey, err := operatorPrivKey.EKMEncryptionKey()
+		if err != nil {
+			logger.Fatal("failed to derive operator-based network key protection secret", zap.Error(err))
+		}
+		protectFn = func(_ context.Context, plaintext []byte) ([]byte, error) {
+			return keys.EncryptPayload(encryptionKey, plaintext)
+		}
+		unprotectFn = func(_ context.Context, protectedValue []byte) ([]byte, error) {
+			return keys.DecryptPayload(encryptionKey, protectedValue)
+		}
+	} else if signerClient != nil {
+		err := probeRemoteNetworkKeyProtector(ctx, signerClient)
+		if err == nil {
+			protectFn = signerClient.OperatorEncrypt
+			unprotectFn = signerClient.OperatorDecrypt
+		} else if !errors.Is(err, ssvsigner.ErrOperatorDataProtectionUnsupported) {
+			logger.Fatal("failed to probe ssv-signer p2p network key protection", zap.Error(err))
+		} else {
+			hasEncryptedKey, hasEncryptedKeyErr := ssv_identity.HasEncryptedNetworkKey(db)
+			if hasEncryptedKeyErr != nil {
+				logger.Fatal("failed to inspect stored p2p network private key format", zap.Error(hasEncryptedKeyErr))
+			}
+			if hasEncryptedKey {
+				logger.Fatal("existing database contains an encrypted p2p network private key, but the configured ssv-signer cannot encrypt or decrypt it. Upgrade ssv-signer to a version that supports /v1/operator/encrypt and /v1/operator/decrypt, or restore the operator key and signing mode that originally encrypted this database", zap.Error(err))
+			}
+
+			logger.Warn("ssv-signer does not support remote p2p network key protection, falling back to local compatibility mode",
+				zap.Error(err),
+			)
+		}
+	}
+
+	istore := ssv_identity.NewIdentityStore(logger, db, protectFn, unprotectFn)
+	netPrivKey, err := istore.SetupNetworkKey(ctx, cfg.NetworkPrivateKey)
 	if err != nil {
 		logger.Fatal("failed to setup network private key", zap.Error(err))
 	}
@@ -1084,6 +1121,25 @@ func setupP2P(logger *zap.Logger, db basedb.Database) network.P2PNetwork {
 		logger.Fatal("failed to setup p2p network", zap.Error(err))
 	}
 	return n
+}
+
+func probeRemoteNetworkKeyProtector(
+	ctx context.Context,
+	client *ssvsigner.Client,
+) error {
+	probeKey := []byte("p2p-network-key-probe")
+	encrypted, err := client.OperatorEncrypt(ctx, probeKey)
+	if err != nil {
+		return fmt.Errorf("probe remote data protector encrypt: %w", err)
+	}
+	decrypted, err := client.OperatorDecrypt(ctx, encrypted)
+	if err != nil {
+		return fmt.Errorf("probe remote data protector decrypt: %w", err)
+	}
+	if !bytes.Equal(decrypted, probeKey) {
+		return errors.New("probe remote network key protector mismatch")
+	}
+	return nil
 }
 
 // syncContractEvents blocks until historical events are synced and then spawns a goroutine syncing ongoing events.
--- a/identity/store.go
+++ b/identity/store.go
@@ -1,10 +1,13 @@
 package p2p
 
 import (
+	"bytes"
+	"context"
 	"crypto/ecdsa"
+	"errors"
+	"fmt"
 
 	gcrypto "github.com/ethereum/go-ethereum/crypto"
-	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
 	"github.com/ssvlabs/ssv/observability/log"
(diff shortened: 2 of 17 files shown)


ssvsigner/keys/aes.go
```diff
@@ -0,0 +1,43 @@
...
+		return nil, err
+	}
+	nonceSize := gcm.NonceSize()
+	if len(nonceCipherText) < nonceSize {
+		return nil, errors.New("malformed ciphertext")
+	}
+
+	nonce, ciphertext := nonceCipherText[:nonceSize], nonceCipherText[nonceSize:]
+	return gcm.Open(nil, nonce, ciphertext, nil)
+}
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai The linked issue is closed, and the linter pipeline passed without this comment

ssvsigner/server.go
```diff
@@ -440,6 +444,74 @@ func (s *Server) handleSignOperator(ctx *fasthttp.RequestCtx) {
...
+	}
+
+	encrypted, err := keys.EncryptPayload(encryptionKey, payload)
+	if err != nil {
+		logger.Error("request failed", zap.Error(err))
+		s.writeJSONErr(ctx, logger, fasthttp.StatusInternalServerError, err)
+		return
+	}
+
+	logger.Info("request finished successfully")
```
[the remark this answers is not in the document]
nkryuchkov: @greptileai it will become inconsistent with other handlers then. I'd rather remove all of them at once if we really see them making too much noise

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
...
+	var (
+		privateKey *ecdsa.PrivateKey
+		found      bool
+		encrypted  bool
+		err        error
+	)
+	if skEncoded == "" {
+		obj, keyFound, err := s.db.Get(prefix, netKeyPrefix)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get privateKey: %w", err)
```
olegshmuelov, line 86: Both the DB fetch error (L86) and the key decode/decrypt error (L91) use the identical message `"failed to get privateKey"`. When decryption fails (wrong operator key, corrupt data), the operator sees `"failed to get privateKey: cipher: message authentication failed"` which doesn't indicate that decryption was the issue.

Suggestion:
```go
// L91 — differentiate from the DB error on L86
return nil, fmt.Errorf("decode network key: %w", err)
```

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
...
+		logger:      logger.Named(log.NameP2PStorage),
+		db:          db,
+		protectFn:   protectFn,
+		unprotectFn: unprotectFn,
 	}
 	return &es
 }
 
-func (s identityStore) GetNetworkKey() (*ecdsa.PrivateKey, bool, error) {
+func (s identityStore) GetNetworkKey(ctx context.Context) (*ecdsa.PrivateKey, bool, error) {
```
olegshmuelov, line 60: Pre-existing: `GetNetworkKey` silently swallows DB errors. When `db.Get` returns `found=false, err=<something>` (I/O failure, corruption), the `!found` check on L62 fires first and returns `nil, false, nil`, the error is lost.

ssvsigner/server_test.go
```diff
@@ -426,6 +427,53 @@ func (s *ServerTestSuite) TestOperatorSign() {
 	})
 }
 
+func (s *ServerTestSuite) TestOperatorDataProtection() {
```
olegshmuelov, line 430: The test covers encrypt's empty-payload and key-derivation-failure paths but not the symmetric decrypt paths:
- Empty payload sent to `PathOperatorDecrypt` (should return 400)
- `EKMEncryptionKey()` failure on the decrypt path (should return 500)

The handlers use identical code structure so the risk is low, but the asymmetry is noticeable and these guard conditions should be covered.

olegshmuelov (commented):
Two operational notes worth considering before merge:

**One-way migration:** Once the node encrypts the key (happens automatically on first startup with an operator key configured), downgrading to an older node version breaks startup with `"invalid private key, >=N"` — the old code passes `enc:` + ciphertext directly to `gcrypto.ToECDSA` which rejects the 64-byte blob as out of range. Worth documenting in release notes. The remote-signer path already has a clear fatal message for this; the local-key downgrade path does not. Consider making the migration log more explicit:

```go
s.logger.Info("migrated p2p network key to encrypted storage (this is irreversible — do not downgrade the node)")
```

**ssv-signer version coupling:** Operators using remote signers must upgrade ssv-signer BEFORE upgrading the node if they want encryption. If the signer is old, the node correctly falls back to plaintext with a warning, no breakage.

Consider adding a **Migration** section to the PR description so QA and reviewers are aware of the one-way upgrade behavior.

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
...
 	obj, found, err := s.db.Get(prefix, netKeyPrefix)
 	if !found {
```
momosh-ssv: What happens if `db.Get` returns `found=false` with a non-nil error here? 
I fear that `!found` check fires first and the error is lost. 

With the new migration path, a transient DB failure on startup would cause the node to generate and encrypt a new identity, permanently losing the original P2P key.

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
...
+			}
+			privateKey.Curve = gcrypto.S256() // temporary hack, so libp2p Secp256k1 is recognized as geth Secp256k1 in disc v5.1
+			found = true
+		}
 	}
 	if skEncoded == "" && found && privateKey != nil {
+		if !encrypted {
+			if s.hasProtectedStorage() {
+				s.logger.Info("migrating plaintext p2p network private key to encrypted storage")
+				if err := s.saveNetworkKey(ctx, privateKey); err != nil {
```
momosh-ssv, line 101: How safe is this if the process crashes or disk fills mid-`db.Set`? 
Seems that the migration is a destructive overwrite with no backup — the key record could be left corrupted. 
A backup-then-replace pattern (write to staging key, verify decrypt, replace in `db.Update`) would make this safer.

cli/operator/node.go
```diff
@@ -1071,9 +1071,46 @@ func setupSSVNetwork(logger *zap.Logger) (*networkconfig.SSV, error) {
...
-func setupP2P(logger *zap.Logger, db basedb.Database) network.P2PNetwork {
-	istore := ssv_identity.NewIdentityStore(logger, db)
-	netPrivKey, err := istore.SetupNetworkKey(cfg.NetworkPrivateKey)
+func setupP2P(ctx context.Context, logger *zap.Logger, db basedb.Database, operatorPrivKey keys.OperatorPrivateKey, signerClient *ssvsigner.Client) network.P2PNetwork {
+	var protectFn func(context.Context, []byte) ([]byte, error)
+	var unprotectFn func(context.Context, []byte) ([]byte, error)
+
+	// The configured startup mode determines how the persisted P2P network key is protected.
+	if operatorPrivKey != nil {
+		encryptionKey, err := operatorPrivKey.EKMEncryptionKey()
```
momosh-ssv, line 1081: Any reason both network key encryption and EKM storage encryption share the same derived AES-256 key? 

Because `EKMEncryptionKey()` uses HKDF with nil `info`, so a ciphertext from one context would decrypt in the other. Passing distinct `info` params (e.g., `"ssv-network-key-v1"` vs `"ssv-ekm-v1"`) gives domain separation for free.

ssvsigner/server.go
```diff
@@ -440,6 +444,75 @@ func (s *Server) handleSignOperator(ctx *fasthttp.RequestCtx) {
 	s.writeBytes(ctx, logger, signature)
 }
 
+func (s *Server) handleOperatorEncrypt(ctx *fasthttp.RequestCtx) {
```
momosh-ssv, line 447: Do we need to handle access control on these endpoints? Any client that can reach the ssv-signer port gets an encrypt/decrypt oracle over the operator-derived key.

cli/operator/node.go
```diff
@@ -1086,6 +1123,25 @@ func setupP2P(logger *zap.Logger, db basedb.Database) network.P2PNetwork {
 	return n
 }
 
+func probeRemoteNetworkKeyProtector(
+	ctx context.Context,
+	client *ssvsigner.Client,
+) error {
+	probeKey := []byte("p2p-network-key-probe")
```
momosh-ssv: Since this goes over the wire, using a random probe payload instead of the fixed `"p2p-network-key-probe"` string would avoid leaking a predictable plaintext/ciphertext pair on every startup. 

I think that AES-GCM is resistant to known-plaintext attacks so the risk is negligible, but `crypto/rand` costs nothing.

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
 	// netKeyPrefix = []byte("network-key/")
 	// operatorKeyPrefix is the prefix for operator key
 	// operatorKeyPrefix = []byte("operator-key/")
+	encryptedPrefix = []byte("enc:")
```
momosh-ssv, line 28: Could this lead to a false positive on a legitimate plaintext key? 
Maybe use a longer prefix like `"encrypted:"` would eliminate it entirely.

ssvsigner/keys/aes.go
```diff
@@ -0,0 +1,43 @@
...
+	block, err := aes.NewCipher(encryptionKey)
+	if err != nil {
+		return nil, err
+	}
+	gcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return nil, err
+	}
+	nonceSize := gcm.NonceSize()
+	if len(nonceCipherText) < nonceSize {
```
momosh-ssv: Might be worth tightening this to `<= nonceSize`. When `len == nonceSize`, `gcm.Open` gets an empty ciphertext and returns a confusing GCM auth error instead of the clearer "malformed ciphertext".

identity/store.go
```diff
@@ -22,72 +25,171 @@ var (
...
+		if keyFound {
+			privateKey, encrypted, err = decodeNetworkKey(ctx, obj.Value, s.unprotectFn)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get privateKey: %w", err)
+			}
+			privateKey.Curve = gcrypto.S256() // temporary hack, so libp2p Secp256k1 is recognized as geth Secp256k1 in disc v5.1
+			found = true
+		}
 	}
 	if skEncoded == "" && found && privateKey != nil {
```
momosh-ssv, line 97: What do you think about wrapping the read-migrate-write in `db.Update()` for atomicity? If two node processes start concurrently against the same DB (misconfiguration), both could race on the migration. 

No data loss since it's the same key material, but a transaction would be cleaner.

cli/operator/node.go
nkryuchkov: We don't pass `ssv-ekm-v1` now, and if we change it, we'll have to coordinate it with `ssvsigner`. And if we pass nil info to e2m but pass something to network key, it will be confusing and inconsistent.

If we had to implement this from scratch, I'd rather use info, but I don't think it adds many benefits in our case. Anyway, we are storing the network key as plaintext right now, and it's the only scope of changes, so I think not having info for it doesn't reduce security

ssvsigner/server.go
nkryuchkov: No, we already have `handleOperatorSign` without access control. `ssvsigner` is for experienced users and has to be set up properly to remain secure. We encourage all users to use TLS and known clients

identity/store.go
nkryuchkov: no, because of the key length

identity/store.go
nkryuchkov: I think we would need it if we had several write operations, but we have only one

identity/store.go
nkryuchkov: But we have just one `Set`, so this should be handled by the DB. I think adding backup-then-replace would overcomplicate the code and won't be a single operation

cli/operator/node.go
momosh-ssv: Ok, agree; thanks for clearing this up @nkryuchkov

olegshmuelov (approved):
LGTM!

momosh-ssv (approved):

y0sher (approved):

iurii-ssv:
@nkryuchkov could you write up a comment (or add it to PR description) about corresponding "disaster recovery plan" we can add to release notes ?

nkryuchkov:
> @nkryuchkov could you write up a comment (or add it to PR description) about corresponding "disaster recovery plan" we can add to release notes ?

@iurii-ssv added to the description

Roy-blox:
## QA Verification — Encrypted p2p key storage (PR #2765)

Verified on stage-hoodi nodes 101-104 (redeployed from pre-encryption branch `proposer-runner-tests` ~Mar 15 → `stage`).

**Migration fired correctly on all 4 nodes at startup** (2026-04-29 11:11–11:12 UTC):
```
INFO migrating plaintext p2p network private key to encrypted storage name=P2PStorage
```

**Peer count recovered** to pre-restart levels (~4–6 peers) within 4 minutes, confirming peer ID was preserved through migration. No legacy plaintext fallback observed.

**Result: PASS ✓**

Tracked in [QA-ssv-033](https://github.com/ssvlabs/ssv-node-board/issues/1034)

PR #2768, opened 2026-03-31
duty-fetching: rework reorg detection (1st slot)

iurii-ssv:
This PR extends https://github.com/ssvlabs/ssv/pull/2767 to cover the scenario described in https://github.com/ssvlabs/ssv/pull/2756#discussion_r3008801386

The summary is: 
- if reorg happens during the very 1st slot of some epoch N (aka during epoch-transition from N-1 to N), we currently don't detect it (unless the "previous dependent root" also happened to change) - and it results into SSV node potentially failing to re-fetch the up-to-date duties for the entire epoch (current epoch for proposer, next epoch for attester)
- this PR tracks the root of the latest event received from EL, which is consulted in that case of epoch-transition to detect those reorgs (it might produce some "false positive reorg signal" due to: the latest block we receive from EL does not necessarily always become the canonical block for the corresponding epoch ... but it shouldn't happen often, and it shouldn't interfere with the duty execution much since EL rarely arrive close to the end of a slot)

### DIFF
--- a/operator/duties/scheduler.go
+++ b/operator/duties/scheduler.go
@@ -367,18 +367,22 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
 // HandleHeadEvent handles the "head" events from the beacon node.
 func (s *Scheduler) HandleHeadEvent() func(ctx context.Context, event *eth2apiv1.HeadEvent) {
 	return func(ctx context.Context, event *eth2apiv1.HeadEvent) {
-		if event.Slot != s.beaconConfig.EstimatedCurrentSlot() {
-			// No need to process outdated events here. Future events shouldn't be possible, unless there is
-			// a very large clock-drift - for simplicity, we don't handle this case either.
-			return
-		}
-
 		currentSlot := event.Slot
 		currentEpoch := s.beaconConfig.EstimatedEpochAtSlot(currentSlot)
 		slotNumber := uint64(currentSlot)%s.beaconConfig.SlotsPerEpoch + 1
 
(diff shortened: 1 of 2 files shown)


operator/duties/scheduler.go
```diff
@@ -362,73 +367,63 @@ func (s *Scheduler) SlotTicker(ctx context.Context) {
...
 
 		buildStr := fmt.Sprintf("e%v-s%v-#%v", currentEpoch, currentSlot, slotNumber)
 		logger := s.logger.With(zap.String("epoch_slot_pos", buildStr))
 
+		if event.Slot < s.beaconConfig.EstimatedCurrentSlot() {
+			// No need to process outdated events here.
+			return
+		}
+		if event.Slot > s.beaconConfig.EstimatedCurrentSlot() {
+			// We don't handle future events to keep things simpple.
```
olegshmuelov, line 382: nit: simple typo

olegshmuelov (approved):
Using `lastBlockRoot` as proxy for expected `CurrentDutyDependentRoot` at epoch transition is the right approach, fixes the detection gap for both proposer and attester with a single scheduler-level mechanism.

One minor nit

operator/duties/scheduler.go
iurii-ssv: Fixed in https://github.com/ssvlabs/ssv/pull/2756/changes/82ef13b0feb2ba5a325abe4eca0737692d27d396 (in https://github.com/ssvlabs/ssv/pull/2756) to preserve the approval

PR #2769, opened 2026-04-01
duty-fetching: rework sync-committee (better retries & edge-cases)

iurii-ssv:
Applies the intent-based duty-fetching approach from #2712 (attester) and #2756 (proposer) to the **sync-committee** handler.

### What changes
- **Retries:** replace the `fetchCurrentPeriod` / `fetchNextPeriod` booleans with a `dutyFetchIntents` map (`period -> fulfilled`). Unfulfilled intents are retried on subsequent slot-ticks, so a single transient Beacon-node error no longer drops a whole period's sync-committee duties (the bug behind `F-ssv-120`).
- **Fetching:** current/next-period duties are fetched & prepared via `prepareCurrentPeriod` / `prepareNextPeriod`, each wrapped in a trace span, matching the attester/proposer handlers.
- **Indices-change:** handled inside the tick (early in the slot), re-fetching the affected periods immediately so the updated duties are ready for the very next tick.
- **Reorg:** a reorg can only affect the **next** period — current-period sync-committee membership is fixed a full period in advance — so we re-fetch just the next period, and only when the current duty dependent root changed.
- **Edge-cases & tests:** fix the `LastSlotOfSyncPeriod` doc comment, add `atLastSlotOrPastCurrentPeriod` (+ unit tests), and add scheduler tests covering the retry / reorg / indices-change / period-transition paths.

Sync-committee remains the only `FilterIndices` caller, so the `waitForInit` flag (which gates validator-store initialization) is kept.

Also backports a one-line comment-wording cleanup to the attester/proposer handlers for consistency.

Resolves `F-ssv-120` / https://github.com/ssvlabs/ssv-node-board/issues/1010


operator/duties/sync_committee.go
```diff
@@ -107,51 +111,86 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
...
+
+					// 2) Process certain intents immediately.
+					if currentSlot == h.beaconConfig.LastSlotOfSyncPeriod(currentPeriod) {
+						delete(h.dutyFetchIntents, currentPeriod) // optimization: prune irrelevant intent
+						h.prepareNextPeriod(tickCtx, logger, currentEpoch, currentPeriod, currentSlot, true)
+					} else {
+						h.prepareCurrentPeriod(tickCtx, logger, currentEpoch, currentPeriod, currentSlot, true)
+					}
+				case <-time.After(time.Until(indicesChangeDeadline)):
+					// It's too late(risky) to handle indices change on the current slot, we'll do it on the next slot.
```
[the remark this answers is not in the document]
iurii-ssv: Not worth optimizing, code-clarity is more important in this case given how infrequent those timer-allocations are

operator/duties/sync_committee.go
```diff
@@ -107,51 +111,86 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
...
 
-			// reset next period duties if in appropriate slot range
-			if h.shouldFetchNextPeriod(currentSlot) {
-				h.fetchNextPeriod = true
-			}
+				// 2) Process certain intents immediately.
+				if reorgEvent.Current {
+					h.prepareNextPeriod(reorgCtx, logger, currentEpoch, currentPeriod, currentSlot, true)
+				}
+			}()
```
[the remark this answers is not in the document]
iurii-ssv: Will add a clarifying comment once https://github.com/ssvlabs/ssv/pull/2756 has been merged (the comment-style should be similar to the comments in there)

operator/duties/sync_committee.go
```diff
@@ -107,51 +111,86 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
...
+					h.dutyFetchIntents[currentPeriod] = false
+					h.dutyFetchIntents[nextPeriod] = false
+
+					// 2) Process certain intents immediately.
+					if currentSlot == h.beaconConfig.LastSlotOfSyncPeriod(currentPeriod) {
+						delete(h.dutyFetchIntents, currentPeriod) // optimization: prune irrelevant intent
+						h.prepareNextPeriod(tickCtx, logger, currentEpoch, currentPeriod, currentSlot, true)
+					} else {
+						h.prepareCurrentPeriod(tickCtx, logger, currentEpoch, currentPeriod, currentSlot, true)
+					}
```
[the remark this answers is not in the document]
iurii-ssv: > The parallel handler (`attester.go`) calls the equivalent `prepareNextEpoch` in the same `else` branch. For consistency, consider also calling `h.prepareNextPeriod(...)` here:

This has been updated for `attester.go` to NOT issue that `prepareNextPeriod` call there, so it is consistent now.

iurii-ssv:
@greptile pls re-review

networkconfig/beacon.go
```diff
@@ -99,7 +99,8 @@ func (b *Beacon) FirstEpochOfSyncPeriod(period uint64) phase0.Epoch {
 	return phase0.Epoch(period * b.EpochsPerSyncCommitteePeriod)
 }
 
-// LastSlotOfSyncPeriod calculates the first epoch of the given sync period.
+// LastSlotOfSyncPeriod calculates the last slot of the given sync period for which
+// producing a sync committee message still makes sense.
 func (b *Beacon) LastSlotOfSyncPeriod(period uint64) phase0.Slot {
```
momosh-ssv: Might be worth renaming this to something like `LastActionableSlotOfSyncPeriod`? 

It returns `firstSlotOfNextPeriod - 2` rather than the period's literal last slot, so a future caller doing `LastSlotOfSyncPeriod(p) + 1` to get the next period's first slot would be off by one. 

The new doc comment helps, but the name on its own still reads like the final slot.

operator/duties/sync_committee.go
```diff
@@ -107,105 +112,211 @@ func (h *SyncCommitteeHandler) HandleDuties(ctx context.Context) {
...
+
+				indicesChangeDeadline := h.beaconConfig.SlotStartTime(currentSlot).Add(h.beaconConfig.IntervalDuration())
+				select {
+				case <-h.indicesChangeCh:
+					logger.Info("🔁 indices change received")
+
+					// 1) Declare intents.
+					// Some validator-related state has changed, so re-fetch the duties for the current and next
+					// period to keep them up to date for all validators.
+					h.dutyFetchIntents[currentPeriod] = false
```
momosh-ssv, line 152: Any reason the intent-declaration order is flipped here versus `attester.go` (which declares `nextEpoch` then `currentEpoch`)? 

No functional difference since these are just map writes, but given this handler is meant to mirror the attester, matching the order would make the two a touch easier to diff.

momosh-ssv (dismissed):

networkconfig/beacon.go
iurii-ssv: Good call — renamed to `LastActionableSlotOfSyncPeriod` and expanded the doc comment to make the off-by-one explicit (it returns one slot *before* the period's literal last slot, so a `+1` to reach the next period's first slot would indeed be wrong). Updated the only caller (`atLastSlotOrPastCurrentPeriod`) and the tests accordingly. Done.

operator/duties/sync_committee.go
iurii-ssv: Good catch. Rather than flip just sync-committee, I standardized on a single `(current, next)` declaration order across all three handlers — so `attester.go` and `proposer.go` now declare `currentEpoch` before `nextEpoch` (which already matched their `HandleInitialDuties` order and the "current and next" comment wording), and sync-committee keeps its existing order. All three mirror each other now. Purely cosmetic — these are independent map writes, so no behavior change.

momosh-ssv (dismissed):
great one, thanks

ovidiu-ssv-labs:
## Non-blocking findings

**1. Missing test: reorg → fetch failure → retry**

There are tests for retry on transient fetch failure (`TestScheduler_SyncCommittee_Retry_*`), but no test covering the reorg-triggered path: reorg fires → `prepareNextPeriod` fails → intent stays `false` → retries on subsequent tick. This is the most direct validation of `F-ssv-120` on the reorg code path, and the reorg handler is distinct enough (uses `reorgCtx`, no `shouldFetchNextPeriod` gate) that a regression there wouldn't be caught by existing tests.

---

**2. No `Reset` of stale next-period duties on reorg before re-fetch** — reorg case in `operator/duties/sync_committee.go`

The old code called `h.duties.Reset(nextPeriod)` before re-fetching so that if the subsequent fetch failed, stale committee assignments were cleared. The new code overwrites duties only on success — if the re-fetch keeps failing, validators continue executing with a stale membership list for the new period. The window is short (1–2 slots until retry succeeds), but it is a behavioural change from the old code. If the preference is "stale duties > missed duties", a comment explaining the deliberate departure from the old approach would be helpful.

---

**3. `tickCtx.Done()` can skip next-period intent registration at a period boundary** — `operator/duties/sync_committee.go`, step 3

Step 3 (registering `dutyFetchIntents[nextPeriod] = false` after a period rollover) sits inside the select block and is skipped when `tickCtx` fires. Under sustained beacon-node slowness (each tick taking >12s) coinciding with a period boundary, the new next-period intent is never registered and next-period duty pre-fetch is indefinitely deferred. Moving step 3 outside the select — or duplicating the intent registration inside the `tickCtx.Done` case before the return — would make this robust.

ovidiu-ssv-labs (dismissed):

ovidiu-ssv-labs (dismissed):

ovidiu-ssv-labs (approved):

momosh-ssv (approved):

PR #2771, opened 2026-04-02
runner: optimize away BaseRunner.mtx

iurii-ssv:
This PR removes the need for `BaseRunner.mtx` that complicated the code quite a bit,

the only real reason to have it were `onTimeout` func(s) that would check `HasRunningDuty()` and skip the issuance of corresponding timeout even if necessary (as [per @momosh-ssv discovery](https://github.com/ssvlabs/ssv/pull/2664#discussion_r2986732753)) ... now we do this instead:
- we do not perform the `HasRunningDuty()` check in the `onTimeout` func(s) themselves, we just issue the timeout event every time (it's pretty rare, so the overhead from the added events is negligible)
- these timeout events are processed within the same for-loop (`StartQueueConsumer` and `ConsumeQueue` funcs) with every other duty-related message/event, they are all processed **sequentially** (due to reading from same thread-safe queue, in the same for-loop)
- the `HasRunningDuty()` check is performed during the actual processing of those timeout events (see `case types.Timeout:`)

Additional notes: 
- Validator/Committee Runners/Queues are managed somewhat differently - hence why there are certain discrepancies with respect to how those are handled in this PR
- this PR also simplifies & standardizes runner-state / qbft-instance-state related checks, preserving old behavior (pure refactor) - figured it's a good time to do it, since we are changing the code in the area anyways

Related: `F-ssv-186` `F-ssv-054` https://github.com/ssvlabs/ssv-node-board/issues/764


protocol/v2/ssv/runner/runner.go
```diff
@@ -193,16 +201,7 @@ func (b *BaseRunner) SetHighestDecidedSlot(slot phase0.Slot) {
...
-	// start new state
-	// TODO nicer way to get quorum
-	state := NewRunnerState(quorum, duty)
-
-	// TODO: potentially incomplete locking of b.State. runner.Execute(duty) has access to
-	// b.State but currently does not write to it
-	b.mtx.Lock() // writes to b.State
-	b.State = state
-	b.mtx.Unlock()
+	b.State = NewRunnerState(quorum, duty)
```
nkryuchkov: do we need `baseSetupForNewDuty` and `hasDutyAssigned` after this change?

nkryuchkov (dismissed):

protocol/v2/ssv/runner/runner.go
iurii-ssv: I guess we can: 
- get rid of `baseSetupForNewDuty` (inline it)
- I'd keep `hasDutyAssigned` as the name itself clarifies the intent (inlining `b.State != nil` might be quite confusing)

https://github.com/ssvlabs/ssv/pull/2771/changes/fcff15c607b7b80bbaaaf2765688c797f0355176

iurii-ssv:
@greptile pls re-review

iurii-ssv:
@greptile could you re-review ?

iurii-ssv:
@greptile pls re-review

iurii-ssv:
> One P2 finding: HasRunningQBFTInstance now directly dereferences RunningInstance.State.Decided without the nil-check that the old IsDecided() provided, introducing a potential panic in deserialization edge cases where Instance.State could be nil. All other changes are correct refactors or positive behavioural improvements (e.g., removing the silent nil-CurrentDuty marshal path in runner_state.go, correctly returning nil instead of an error when a pruned committee runner's timeout arrives).
> 
> protocol/v2/ssv/runner/runner.go (HasRunningQBFTInstance nil-safety)

Regarding that concern, the behavior is intentional - the `HasRunningQBFTInstance` has an explicit comment about it:
```go
func (b *BaseRunner) HasRunningQBFTInstance() bool {
	// Note: RunningInstance.State cannot be nil for existing RunningInstance by construction.
	return b.hasDutyRunning() && b.HasStartedQBFTInstance() && !b.State.RunningInstance.State.Decided
}
```
I think `panic` is better in this case than any of:
- retuning an implicit `false` that would hide/mislead on the actual root cause (of the bug being present for some `RunningInstance.State` initialization we have somewhere)
- returning an `error` that needs to be handled - this would complicate a lot of call-sights ... so it's just not worth it (and it will also make it harder to spot the root cause - a bug in our code that needs fixing)

protocol/v2/ssv/runner/runner.go
```diff
@@ -96,21 +103,22 @@ type BaseRunner struct {
 	highestDecidedSlot phase0.Slot
 }
 
+func (b *BaseRunner) HasRunningDuty() bool {
```
nkryuchkov, line 109: This looks inconsistent with `hasDutyRunning`. They have a different word order. Why do we need `hasDutyRunning`? We can just use `HasRunningDuty`

iurii-ssv: Technically we don't need both (can just use `HasRunningDuty`), ideally we don't want to expose this functionality at all ... hence why we have [all these func defined as private, and also follow the same naming scheme](https://github.com/ssvlabs/ssv/blob/e232cdab3946dbba7bfdb60efe77149160633c89/protocol/v2/ssv/runner/runner.go#L471-L481) ... but the legacy design of Validator-Runner forces us to export it (Commitee-Runner does not force us to export it).

Long-term (after we've reworked Validator-Runner to be per-duty) - there won't be a need for a public `HasRunningDuty` func - and so we will be able to cleanly delete it (without having to adjust the rest of the code to use private func such as `hasDutyRunning`).

So, having `HasRunningDuty` func to delegate to `hasDutyRunning` seems like an OK trade-off to me, given we gain "clear visibility" into who exactly needs this func exported. Hence I would leave it as is.

iurii-ssv: Added clarifying comment https://github.com/ssvlabs/ssv/pull/2771/changes/070bb290f4191c0a3290c647d7b22005a2d66b17

Just to clarify, `Runner` may have to expose/present some of it's internal state to the outside world (such as maybe `HasStartedQBFTInstance`) - we just need to limit it as much as possible.

nkryuchkov (dismissed):

olegshmuelov (dismissed):
LGTM!

protocol/v2/ssv/validator/timer.go
```diff
@@ -86,33 +75,31 @@ func (v *Validator) createTimerMessage(identifier spectypes.MessageID, height sp
...
-			return
-		}
-
-		hasDuty := dr.HasRunningDuty()
-		if !hasDuty {
+		// If the relevant queue hasn't been initialized yet, there isn't a running duty we can issue a
+		// timeout for, in practice this should never happen - but we need to handle this just in case.
+		// This is also possible if the queue got pruned already (due to becoming old and irrelevant).
+		q := c.Queues[phase0.Slot(height)]
+		if q.Q == nil {
```
ljuba-ssv, line 85: nit: This checks whether the *queue* was pruned, but the comment says "likely was pruned" implying the runner. The old code checked `c.Runners[slot] == nil`. Now if the runner is pruned but the queue still exists, the timeout gets pushed into a queue whose runner is gone. It's handled gracefully downstream (the `ProcessMessage` change at committee.go:327 returns nil), so not a bug — but the comment could be more precise, e.g. "missing queue (slot may not have been initialized or was already pruned)."

protocol/v2/ssv/runner/runner_state.go
```diff
@@ -84,18 +84,15 @@ func (pcs *State) MarshalJSON() ([]byte, error) {
...
 
-	if pcs.CurrentDuty != nil {
-		if ValidatorDuty, ok := pcs.CurrentDuty.(*spectypes.ValidatorDuty); ok {
-			alias.ValidatorDuty = ValidatorDuty
-		} else if committeeDuty, ok := pcs.CurrentDuty.(*spectypes.CommitteeDuty); ok {
-			alias.CommitteeDuty = committeeDuty
-		} else {
-			return nil, errors.New("can't marshal because BaseRunner.State.CurrentDuty isn't ValidatorDuty or CommitteeDuty")
-		}
+	if ValidatorDuty, ok := pcs.CurrentDuty.(*spectypes.ValidatorDuty); ok {
```
ljuba-ssv, line 88: nit: The old code had `if pcs.CurrentDuty != nil { ... }` which would silently marshal State with no duty field when CurrentDuty was nil. The new code will hit the `else` branch and return an error. This is arguably better behavior, but it's a subtle behavioral change — worth a comment noting that CurrentDuty is guaranteed non-nil by `NewRunnerState` construction.

protocol/v2/ssv/runner/runner.go
```diff
@@ -96,21 +103,25 @@ type BaseRunner struct {
...
-	if b.HasRunningDuty() {
-		runningInstance = b.State.RunningInstance
-		if runningInstance != nil {
-			decided, _ := runningInstance.IsDecided()
-			return !decided
-		}
-	}
-	return false
+	// Note: RunningInstance.State cannot be nil for existing RunningInstance by construction.
+	return b.hasDutyRunning() && b.HasStartedQBFTInstance() && !b.State.RunningInstance.State.Decided
```
ljuba-ssv: nit (two things):

1. **Redundant check:** `hasDutyRunning()` already implies `hasDutyAssigned()` (== `State != nil`), and `HasStartedQBFTInstance()` checks `hasDutyAssigned()` again. Could simplify to:
```go
return b.hasDutyRunning() && b.State.RunningInstance != nil && !b.State.RunningInstance.State.Decided
```

2. **Dropped nil-guard:** The old `IsDecided()` had a nil-check on `Instance.State` before reading `Decided`. The comment says "RunningInstance.State cannot be nil for existing RunningInstance by construction" — this holds for `NewInstance` but deserialized instances (`UnmarshalJSON`) could theoretically have a null State from a malformed JSON fixture. Low risk since it's only reachable via spectests, but noting it.

protocol/v2/ssv/validator/timer.go
```diff
@@ -22,38 +22,27 @@ func (v *Validator) onTimeout(ctx context.Context, logger *zap.Logger, identifie
...
-			logger.Debug("❌ failed to decode timer msg", zap.Error(err))
+			logger.Error("❌ failed to decode timer msg", zap.Error(err))
 			return
 		}
 
 		if pushed := q.TryPush(dec); !pushed {
-			logger.Warn("❗️ dropping timeout message because the queue is full",
-				fields.RunnerRole(identifier.GetRoleType()),
-			)
+			logger.Error("❗️ dropping timeout message because the queue is full", fields.RunnerRole(identifier.GetRoleType()))
```
ljuba-ssv, line 45: nit: The old level was `Warn`. A full queue during heavy load is an operational signal (backpressure), not necessarily an error — `Warn` might be more appropriate here to avoid noise in error-level alerting. Same applies to line 102.

protocol/v2/ssv/runner/runner.go
```diff
@@ -81,8 +79,17 @@ type DoppelgangerProvider interface {
...
-	mtx            sync.RWMutex
-	State          *State
+	// State stores the current runner state, this state corresponds to 1 particular duty the runner is
+	// currently busy with at the moment. The BaseRunner is not responsible for synchronizing any updates
+	// State might need to record - the caller is responsible to ensure the updates/reads (these can happen
+	// whenever runner's method is called to process a p2p message, or an event) are applied sequentially,
+	// plus the caller is also responsible for ensuring there is no race with moving on to the next duty
+	// (the baseSetupForNewDuty call).
+	// Note, the current implementation achieves concurrent safety by making sure every State read/update
+	// is done by the same go-routine, handling all the messages in queue.SSVMessage (p2p messages and events)
```
ljuba-ssv, line 89: Ultra-minor: "go-routine" → "goroutine" (the canonical Go spelling, per the spec and standard library docs).

ljuba-ssv (dismissed):
Clean PR — the mutex removal is well-reasoned and the timeout-via-queue approach is solid. A few minor nits inline.

protocol/v2/ssv/runner/runner.go
iurii-ssv: > but deserialized instances (UnmarshalJSON) could theoretically have a null State from a malformed JSON fixture. Low risk since it's only reachable via spectests, but noting it.

This is only relevant for spec-tests, and given those pass - we are already good. I think it's best to panic in this case to find the issue early during development (rather than silently return some incorrect value) in this particular case.

protocol/v2/ssv/validator/timer.go
iurii-ssv: Well, yeah, I've changed this intentionally to Error since it is a "mission-critical error" (or at least a very undesirable one) ... `Warn` doesn't do it justice.

protocol/v2/ssv/runner/runner.go
iurii-ssv: Good point, but I think it's better to refactor this in a project-wide swoop - created an issue for it to address later: https://github.com/ssvlabs/ssv-node-board/issues/1007

protocol/v2/ssv/runner/runner_state.go
iurii-ssv: Added a note-comment for this block of code to document this.

protocol/v2/ssv/validator/timer.go
iurii-ssv: I'm not entirely sure what this comment suggests in terms of "rephrasing",

> Now if the runner is pruned but the queue still exists

both runner & queue are pruned together as a part of a single atomic operation (with `c.mtx` held), so this ^ doesn't really apply.

ljuba-ssv: very very small nit :) thought the comment implies it was just the runner that was getting pruned and not both, completely irrelevant

ljuba-ssv (dismissed):

ljuba-ssv (approved):

ljuba-ssv (approved):

olegshmuelov (approved):

Issue #2750, opened 2026-03-25
refactor: rename nodeprobe to reflect broader health-check scope


olegshmuelov:
## Context

`nodeprobe` was originally scoped to ethereum node health checks (CL/EL), but the naming was already stretched — `eventSyncerNodeName` is not an ethereum node. #2748 adding p2p discovery makes it clearly wrong. The package now probes arbitrary components but the naming still says "node."

## Scope

- Rename package `nodeprobe` → `probe` (or `healthprobe`)
- Rename `startNodeProber` → `startHealthProber`
- Rename `*NodeName` constants → `*Name` or `*ComponentName`
- Update `nodeProber` variable names and API handler fields
- ~6 files, mechanical rename, no logic changes

## Ref

https://github.com/ssvlabs/ssv/pull/2748#discussion_r2987858749

PR #2772, opened 2026-04-03
health-prober: rename to match responsibilities

iurii-ssv:
This PR aligns the names used for health-prober related things to its actual behavior (it now deals with generic components, rather than Ethereum nodes).

The SSV node API that allows to trigger component-probing wasn't changed in this PR due to:
- it seems the names exposed in there are fine as is (doesn't seem like breaking backwards compatibility worth the change)
- the probing of p2p-component isn't even exposed in that API

Resolves https://github.com/ssvlabs/ssv/issues/2750 / `F-ssv-267`


iurii-ssv:
@greptile pls re-review

api/handlers/node/node.go
```diff
@@ -22,31 +22,31 @@ type Node struct {
...
 	peersIndex peersIndex,
 	network p2pNetwork,
 	topicIndex topicIndex,
-	nodeProber *nodeprobe.Prober,
-	clNodeName string,
-	elNodeName string,
+	healthProber *hprobe.HealthProber,
+	clComponentName string,
+	elComponentName string,
 	eventSyncerNodeName string,
```
olegshmuelov: This parameter was the only one in `NewNode` not renamed. The struct field it maps to (`eventSyncerComponentName`) was renamed, so the assignment on L49 silently bridges old->new. Consistent naming:

Suggestion:
```go
eventSyncerComponentName string,
```

with the struct assignment becoming:
```go
eventSyncerComponentName: eventSyncerComponentName,
```

olegshmuelov (dismissed):
nit: `network/p2p/p2p.go:486` still references the old package name:

```go
// It satisfies the nodeprobe health-check interface.
```

Should be:
```go
// It satisfies the hprobe health-check interface.
```

olegshmuelov (approved):
LGTM!

ljuba-ssv (approved):

PR #2773, opened 2026-04-05
fix(p2p): reject bad peers early in InterceptPeerDial

olegshmuelov:
## Summary
- `InterceptPeerDial` unconditionally returned `true`, allowing outbound dials to known-bad peers
- Added `isBadPeer` check to reject bad peers before address resolution, avoiding wasted TCP connection resources
- Same check already exists in `InterceptAddrDial` (per-address) and `InterceptSecured` (post-handshake) — this moves rejection to the earliest possible stage

## Finding
F-ssv-128 — Connection gater InterceptPeerDial always returns true (P1/HIGH Security)

## Test
- Updated `TestConnGaterInterceptPeerDial` to cover bad-peer rejection and good-peer passthrough
- All connection gater tests pass with `-race`

### DIFF
--- a/network/peers/connections/conn_gater.go
+++ b/network/peers/connections/conn_gater.go
@@ -64,6 +64,10 @@ func NewConnectionGater(
 // to the addresses of that peer being available/resolved. Blocking connections
 // at this stage is typical for blacklisting scenarios
 func (n *connGater) InterceptPeerDial(id peer.ID) bool {
+	if n.isBadPeer(id) {
+		n.logger.Debug("preventing outbound dial to bad peer", fields.PeerID(id))
+		return false
+	}
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


network/peers/connections/conn_gater.go
```diff
@@ -64,6 +64,10 @@ func NewConnectionGater(
 // to the addresses of that peer being available/resolved. Blocking connections
 // at this stage is typical for blacklisting scenarios
 func (n *connGater) InterceptPeerDial(id peer.ID) bool {
+	if n.isBadPeer(id) {
+		n.logger.Debug("preventing outbound dial to bad peer", fields.PeerID(id))
+		return false
+	}
 	return true
 }
```
iurii-ssv: Nit: while we are here, I would "name" the returned `bool` value (to be `allow bool`) for `connGater` methods (`InterceptPeerDial` and `InterceptAddrDial` and others) - so it's clear what true/false means

iurii-ssv (dismissed):

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2775, opened 2026-04-05
refactor(qbft): per-duty RoundTimer — eliminate cross-duty state management

### LINKED ISSUE #2766: refactor(qbft): per-duty RoundTimer to eliminate cross-duty state management
## Summary

Refactor `RoundTimer` from a single shared long-lived instance to a per-duty scoped timer,
eliminating the deferred arming, height tracking, and stale-callback guards added in #2755.

## Motivation

The current `RoundTimer` is created once per QBFT controller and reused across all duties.
Because `TimeoutForRound` fires before the callback is registered (instance.Start calls it
before the runner sets the callback via `OnTimeout`), the timer needs:

- `callbackHeight` to track which duty owns the callback
- `deferredTimeout` to queue arms that arrive before the callback
- Monotonic height guards to reject stale calls from previous duties
- Replay logic in `OnTimeout` to re-arm deferred requests

This complexity exists solely because one object serves multiple duties with late-bound callbacks.

## Proposed Design

Create a fresh timer per duty with the callback wired at construction:

**Interface:**
```go
type Timer interface {
    TimeoutForRound(round specqbft.Round)  // no height — timer is scoped to one duty
    Stop()
}
```

**Construction:** The runner creates the timer (it has the callback) and injects it into the
instance before Start:
```go
timer := roundtimer.New(ctx, beaconConfig, role, height, callback)
```

**Key question:** How does the timer reach the instance? Options:
1. **Inject into instance.Start** — cleanest, but changes Start signature
2. **TimerFactory on Config** — `Config.NewTimer(height, callback) Timer` — keeps current wiring
3. **SetTimer on instance before Start** — minimal API change

Option 2 preserves the Config-based access pattern while enabling per-duty scoping.

## What Gets Eliminated

- `callbackHeight`, `deferred`, `deferredTimeout` struct
- `OnTimeout` method (callback at construction)
- Height parameter from all `TimeoutForRound` call sites (4 in instance/)
- Height guards and replay logic
- `registerTimeoutHandler` in runner

## Scope

- `protocol/v2/qbft/roundtimer/timer.go` — new per-duty constructor, simplified struct
- `protocol/v2/qbft/roundtimer/timer_test.go` — simplified tests (no deferred arming tests needed)
- `protocol/v2/qbft/config.go` — TimerFactory or remove Timer field
- `protocol/v2/qbft/instance/` — all 4 TimeoutForRound call sites drop height
- `protocol/v2/ssv/runner/timer.go` — timer creation replaces registerTimeoutHandler
- `protocol/v2/ssv/runner/runner.go` — wire timer before Start
- `operator/validator/controller.go` — factory instead of single timer
- Spec tests and mock timer adapt to new interface

## Context

Follow-up from #2755 review ([comment](https://github.com/ssvlabs/ssv/pull/2755#pullrequestreview-4029197192)).

olegshmuelov:
## Summary
- Refactor RoundTimer from shared-across-duties to per-duty scoped, eliminating deferred arming, height tracking, and stale-callback guards added in #2755
- Align node Timer interface with ssv-spec (`TimeoutForRound` without height)
- Fix `State.Finished` data race exposed by per-duty timer callbacks

## Why
The shared RoundTimer served multiple duties with late-bound callbacks, requiring complex cross-duty state management (deferred arming, height guards, replay logic). This was the root cause of the temporal coupling between `Instance.Start` and `registerTimeoutHandler`. Per-duty timers make this class of bugs structurally impossible.

## Context
Follow-up from the #2754 / #2755 review chain. No spec repo changes needed — ssv-spec v1.2.2 already defines `Timer.TimeoutForRound(round)` without a height parameter.

Closes #2766

## Test
- All existing tests pass with `-race`
- `golangci-lint` clean


protocol/v2/qbft/controller/controller.go
```diff
@@ -87,8 +88,8 @@ func (c *Controller) StartNewInstance(
 
 	newInstance := instance.NewInstance(logger, c.GetConfig(), c.CommitteeMember, c.Identifier, c.Height, c.OperatorSigner)
 	c.StoredInstances.addNewInstance(newInstance)
-	newInstance.Start(ctx, value, height, valueChecker)
 	c.forceStopAllInstanceExceptCurrent()
+	newInstance.Start(ctx, value, height, timer, valueChecker)
```
iurii-ssv: Isn't the `forceStopAllInstanceExceptCurrent` call meant to be done after the `newInstance.Start` ? If we reorder it like that - seems like we might end up with 2 running instances (I'm not sure if it has any negative implications, just as a surface-observation - looks like we shouldn't be doing it like that).

Could also refactor the `forceStopAllInstanceExceptCurrent` to just be `c.forceStopAllInstances` instead - that would be a cleaner solution overall, I guess.

protocol/v2/qbft/instance/instance.go
```diff
@@ -79,11 +80,22 @@ func (i *Instance) ForceStop() {
 	i.forceStop = true
 }
 
+// SetTimer replaces the timer. Used in tests for instances not started via Start.
+func (i *Instance) SetTimer(timer specqbft.Timer) {
+	i.timer = timer
+}
```
iurii-ssv: Seems like we don't really need the setter for `i.timer` and could just always initialize it with `NewInstance` instead ? If so, I would keep setters to the necessary minimum since it eventually becomes hard to track down "how various struct fields get initialized/updated" (just as a general convention I would follow).

protocol/v2/qbft/instance/instance.go
```diff
@@ -107,7 +119,8 @@ func (i *Instance) Start(
 	i.bumpToRound(specqbft.FirstRound)
 	i.State.Height = height
 	i.ValueChecker = valueChecker
-	i.config.GetTimer().TimeoutForRound(height, specqbft.FirstRound)
+	i.timer = timer
+	i.timer.TimeoutForRound(specqbft.FirstRound)
 	i.metrics.StartStage(stageProposal)
```
iurii-ssv: To my another comment about `SetTimer` setter - this is basically why I don't like setters because we basically have 3 different ways to initialize `Instance struct` fields (use constructor, set the field directly, use setter).

I think we could move all of these to be initialized in the `NewInstance` constructor:
```go
	i.StartValue = value
	i.bumpToRound(specqbft.FirstRound)
	i.State.Height = height
	i.ValueChecker = valueChecker
	i.timer = timer
```

protocol/v2/qbft/instance/proposal.go
```diff
@@ -33,8 +33,9 @@ func (i *Instance) uponProposal(ctx context.Context, logger *zap.Logger, msg *sp
 	msgRound := msg.QBFTMessage.Round
 
 	// A future justified proposal should bump us into future round and reset timer
-	if msgRound > i.State.Round {
-		i.config.GetTimer().TimeoutForRound(msg.QBFTMessage.Height, msgRound)
+	// timer is nil for decoded instances that haven't gone through Start
+	if msgRound > i.State.Round && i.timer != nil {
+		i.timer.TimeoutForRound(msgRound)
 	}
```
iurii-ssv, line 39: To my another comment about `SetTimer`, if you end up moving it into constructor - might want to double-check is this `&& i.timer != nil` is still relevant after that, and `that haven't gone through Start` part of the comment would become outdated too.

protocol/v2/qbft/instance/round_change.go
```diff
@@ -116,7 +116,10 @@ func (i *Instance) uponChangeRoundPartialQuorum(logger *zap.Logger, newRound spe
 	i.bumpToRound(newRound)
 	i.State.ProposalAcceptedForCurrentRound = nil
 
-	i.config.GetTimer().TimeoutForRound(i.State.Height, i.State.Round)
+	// timer is nil for decoded instances that haven't gone through Start
+	if i.timer != nil {
+		i.timer.TimeoutForRound(i.State.Round)
+	}
```
iurii-ssv: To my another comment about `SetTimer`, if you end up moving it into constructor - might want to double-check is this `i.timer != nil` is still relevant after that, and `that haven't gone through Start` part of the comment would become outdated too.

protocol/v2/qbft/instance/timeout.go
```diff
@@ -41,7 +41,10 @@ func (i *Instance) UponRoundTimeout(ctx context.Context, logger *zap.Logger) err
 	defer func() {
 		i.bumpToRound(newRound)
 		i.State.ProposalAcceptedForCurrentRound = nil
-		i.config.GetTimer().TimeoutForRound(i.State.Height, newRound)
+		// timer is nil for decoded instances that haven't gone through Start
+		if i.timer != nil {
+			i.timer.TimeoutForRound(newRound)
+		}
```
iurii-ssv: To my another comment about `SetTimer`, if you end up moving it into constructor - might want to double-check is this `i.timer != nil` is still relevant after that, and `that haven't gone through Start` part of the comment would become outdated too.

protocol/v2/ssv/runner/runner.go
```diff
@@ -90,7 +90,8 @@ type BaseRunner struct {
 	ssvtypes.OperatorSigner
 
 	// implementation vars
-	TimeoutF TimeoutF `json:"-"`
+	TimeoutF    TimeoutF           `json:"-"`
+	timerCancel context.CancelFunc `json:"-"`
```
iurii-ssv: It'd be best to replace the `// implementation vars` with something more meaningful, while we are at it.

protocol/v2/ssv/runner/timer.go
```diff
@@ -7,16 +7,18 @@ import (
...
+func (b *BaseRunner) createTimer(ctx context.Context, logger *zap.Logger, height specqbft.Height) specqbft.Timer {
+	if b.timerCancel != nil {
+		b.timerCancel()
 	}
+	ctx, cancel := context.WithCancel(ctx)
+	b.timerCancel = cancel
+	identifier := spectypes.MessageID(b.QBFTController.GetIdentifier())
+	callback := b.TimeoutF(ctx, logger, identifier, height)
+	return roundtimer.New(ctx, b.NetworkConfig.Beacon, b.RunnerRoleType, height, callback)
 }
```
iurii-ssv, line 24: We could probably improve the way we manage the lifecycle of round-timer (and corresponding `ctx`+`timerCancel`), something like:
- `RoundTimer` should have `Stop` method that would be called whenever the corresponding managing qbft-instance has been stopped (via `Instance.ForceStop` call, I guess)
- this should make sense because `Instance` is fully managing `RoundTimer` now, the `BaseRunner` doesn't need to know anything about `RoundTimer` (so we can get rid of `timerCancel`)
- we probably want to change `OnRoundTimeoutF` func-signature to accept `ctx` + `TimeoutF` to NOT accept `ctx`, to perform this refactoring (I would imagine)

protocol/v2/ssv/runner/runner.go
```diff
@@ -489,6 +488,19 @@ func (b *BaseRunner) hasDutyAssigned() bool {
...
+	defer b.mtx.Unlock()
+
+	if b.timerCancel != nil {
+		b.timerCancel()
+		b.timerCancel = nil
+	}
+	if b.State != nil {
+		b.State.Finished = true
+	}
+}
```
iurii-ssv, line 501: Regarding `b.mtx`, I have just finished https://github.com/ssvlabs/ssv/pull/2771 where we can basically get rid of it for good ... I'm not entirely sure what exact race you are referring to in PR description:

> * Fix `State.Finished` data race exposed by per-duty timer callbacks

because we basically "serialize/pipeline" every message/event that reads/changes duty-related state through the same thread-safe queue (so all those `.State.Finished` reads/writes should be perfectly serialized already), could you clarify ?

**Side-note:** if there really is a race related to .State.Finished to resolve here - we'd need to apply similar fix to the AggregatorCommitteeRunner (from `boole-fork` branch)

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -139,94 +116,42 @@ func (t *RoundTimer) RoundTimeout(height specqbft.Height, round specqbft.Round)
...
 
 	// Calculate the time until the duty should start plus the timeout duration
 	return time.Until(dutyStartTime.Add(timeoutDuration))
 }
 
 // armLocked stops any running timer and schedules a new AfterFunc for the
-// given height and round. Caller must hold t.mtx.
-func (t *RoundTimer) armLocked(height specqbft.Height, round specqbft.Round) {
+// given round. Caller must hold t.mtx.
+func (t *RoundTimer) armLocked(round specqbft.Round) {
```
iurii-ssv: Nit: inlining this func would make the code a bit simpler to read in the future (since we no longer need it)

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -139,94 +116,42 @@ func (t *RoundTimer) RoundTimeout(height specqbft.Height, round specqbft.Round)
...
 	// immediately but the callback blocks on RLock until we release mtx.
-	t.timer = time.AfterFunc(t.RoundTimeout(height, round), func() {
+	t.timer = time.AfterFunc(t.RoundTimeout(round), func() {
 		if t.ctx.Err() != nil {
 			return
 		}
 		t.mtx.RLock()
 		defer t.mtx.RUnlock()
-		if t.callbackHeight != height || t.round != round || t.callback == nil {
+		if t.round != round || t.callback == nil {
```
iurii-ssv: The `|| t.callback == nil` part seems no longer relevant ? 

I would also leave a brief comment explaining that `t.round != round` check is "an optimization, in case the duty has already moved on to the next round".

iurii-ssv (commented):
Nice cleanup! Left some questions/suggestions.

protocol/v2/qbft/controller/controller.go
olegshmuelov: Intentional reorder — two independent mechanisms stop old instances here:
1. `createTimer()` in `decide()` cancels the old timer context — old callbacks check `ctx.Err()` and bail
2. `ForceStop()` prevents late consensus messages from continuing the old instance

Setting `ForceStop` before `Start` means old instances reject messages sooner. No 2-running risk since `ForceStop` is a flag check, not blocking.

Re rename: `forceStopAllInstances` only works if the call moves before `addNewInstance`. Since it's after, it must skip the new instance at `c.Height`. Current name is accurate.

protocol/v2/ssv/runner/timer.go
olegshmuelov: Agree, Instance fully owning timer lifecycle is the clean end state. The main coupling to untangle: `finishDuty()` on BaseRunner currently cancels the timer via `timerCancel` when a duty decides successfully. Moving that to Instance means BaseRunner calls through to Instance to stop the timer — doable but changes the layer boundary. I'd keep this as a focused follow-up, especially with #2771 in flight.

protocol/v2/ssv/runner/runner.go
olegshmuelov: The race: `time.AfterFunc` fires the callback on its own goroutine. The callback calls `HasRunningDuty()` (in `Validator.onTimeout` and `Committee.onTimeout` in `validator/timer.go`), which reads `State.Finished` under `RLock`. Before this PR, the write side (`State.Finished = true` in each runner) had no lock — that's the race.

`finishDuty()` wraps the write in `mtx.Lock()`, properly pairing with `HasRunningDuty()`'s `RLock`. With the old shared timer it rarely triggered because `OnTimeout`'s height-check bailed stale callbacks before reaching `HasRunningDuty()`. Per-duty timers removed that bailout, widening the window. `-race` caught it.

If #2771 serializes the timer callback path through the queue, the mtx becomes redundant. Happy to rebase and drop it if #2771 lands first.

Re AggregatorCommitteeRunner on boole-fork — will check and apply if needed.

protocol/v2/qbft/instance/instance.go
olegshmuelov: Done, timer now injected at construction. `NewInstance` accepts `timer specqbft.Timer` (nil for skeleton/decoded instances). Removed it from `Start` params. `SetTimer` stays only for the JSON-decoded test path (2 call sites).

protocol/v2/qbft/instance/instance.go
olegshmuelov: Moved the construction-time fields (timer, removed redundant height/round re-set). Kept `StartValue` and `ValueChecker` in `Start`, those are duty input data only available at start time. This gives a clean split: constructor = identity + infrastructure, Start = arm timer + set duty data.

protocol/v2/qbft/instance/proposal.go
olegshmuelov: timerless instances are real (decided skeletons pass nil to constructor, decoded instances may not have SetTimer called). Updated the comment to reference the actual cases instead of "haven't gone through Start."

protocol/v2/qbft/instance/round_change.go
olegshmuelov: ditto

protocol/v2/qbft/instance/timeout.go
olegshmuelov: ditto

protocol/v2/ssv/runner/runner.go
olegshmuelov: Checked `boole-fork`. `AggregatorCommitteeRunner` has 4 direct `State.Finished = true` writes without locking, same race pattern. @nkryuchkov @y0sher  will the upcoming release will be merged` into `boole-fork` as well? If so this fix carries over automatically. If not, we need a standalone fix there.

iurii-ssv: > The race: `time.AfterFunc` fires the callback on its own goroutine. The callback calls `HasRunningDuty()` (in `Validator.onTimeout` and `Committee.onTimeout` in `validator/timer.go`), which reads `State.Finished` under `RLock`. Before this PR, the write side (`State.Finished = true` in each runner) had no lock — that's the race.

Oh yeah, that one is accounted for in https://github.com/ssvlabs/ssv/pull/2771 (as mentioned in PR description) - it's probably better to merge it first (before we merge https://github.com/ssvlabs/ssv/pull/2775) to avoid the need to keep `boole-fork` branch consistent with the fix from this PR.

nkryuchkov: > @nkryuchkov @y0sher will the upcoming release will be merged`into`boole-fork` as well? If so this fix carries over automatically. If not, we need a standalone fix there.

yes, it should be

protocol/v2/qbft/controller/controller.go
iurii-ssv: Ah, I see.

protocol/v2/ssv/runner/timer.go
iurii-ssv: Sounds good, leaving it up to you to create a task for this follow-up then!

iurii-ssv (approved):
Looks great to me!

protocol/v2/ssv/runner/timer.go
olegshmuelov: Thought about this more. The main complication: `createTimer()` needs to cancel the old timer BEFORE the new Instance exists (before `StartNewInstance`), so Instance can't own that cancellation.
The current model, creator cleans up, is actually straightforward. If #2771 serializes timer callbacks through the queue, the need for explicit cancellation diminishes and the lifecycle simplifies naturally. I'd revisit this after #2771 lands rather than redesigning the ownership now.

nkryuchkov (approved):

ljuba-ssv (approved):

PR #2777, opened 2026-04-06
runner: clarify queue state management

iurii-ssv:
While working on https://github.com/ssvlabs/ssv/pull/2771, I noticed that Validator and Committee queues were difficult to reason about because queue prioritization state was partially duplicated outside the runners.

This PR clarifies queue state management and fixes a committee queue prioritization bug.

## What changed

- Removed the committee-specific `queueContainer` wrapper and store only `queue.Queue` in `Committee.Queues`.
- Derive queue prioritization state from the current runner state on every pop instead of keeping a separate stored queue state.
- Clarified that `queue.State.Slot` doubles as QBFT height, since runner QBFT instances are started at `height = slot`.
- Updated validator and committee queue consumers to populate prioritization state consistently from runner getters.
- Added regression coverage for committee queue filtering past round 1.

## Behavior fix

Previously, `protocol/v2/ssv/validator/committee_queue.go` did not update the queue state round from the runner. As a result, the filter that skips prepare/commit messages before a proposal is accepted only worked against the default round value, rather than the runner’s actual current round.

This PR updates the committee queue consumer to use the runner’s current round before each pop, so prepare/commit messages for the current round are filtered correctly.

Additionally, stale lower-height commit prioritization now checks the decoded QBFT message type instead of the outer SSV message type, so lower-height commits retain their intended priority over lower-slot pre-consensus messages.

Related: `F-ssv-186`, `F-ssv-054`, https://github.com/ssvlabs/ssv-node-board/issues/764

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"bytes"
 	"context"
+	crand "crypto/rand"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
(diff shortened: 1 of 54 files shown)


olegshmuelov (approved):
LGTM

ljuba-ssv (approved):

PR #2779, opened 2026-04-07
Avoid retry loops when subscribing random subnets (F-ssv-183)

julienh-ssv:
F-ssv-183

## Summary
- replace retry-based random subnet selection with direct sampling from currently available subnets
- fail fast when the requested subnet count exceeds the number of non-active subnets
- add focused tests for the error and success paths in `SubscribeRandoms`

## Root cause
`SubscribeRandoms` retried indefinitely until it found a random subnet set with zero overlap against `currentSubnets`. When too few free subnets remained, that condition could become impossible, leaving the caller stuck in a tight loop.

## Impact
This change guarantees termination and returns a clear error when the request cannot be satisfied with the currently available subnets.

## Validation
- `go test ./network/p2p -run TestSubscribeRandoms -count=1`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run ./network/p2p/...`
- `git diff --check`


nkryuchkov (approved):

olegshmuelov (approved):

network/p2p/p2p_pubsub.go
```diff
@@ -87,43 +87,45 @@ func (n *p2pNetwork) SubscribeAll() error {
...
-		if !randSubnetAlreadyInUse {
-			// found a set of random subnets that we aren't yet using
-			break
+	availableSubnetsCount := commons.SubnetsCount - n.currentSubnets.ActiveCount()
+	if numSubnets > availableSubnetsCount {
+		return fmt.Errorf("not enough available subnets: requested %d, available %d", numSubnets, availableSubnetsCount)
+	}
+
+	availableSubnets := make([]uint64, 0, availableSubnetsCount)
+	for subnet := uint64(0); subnet < commons.SubnetsCount; subnet++ {
```
oleg-ssvlabs, line 107: nit: this could be simplified to:

`for subnet := range uint64(commons.SubnetsCount)`

network/p2p/p2p_pubsub.go
```diff
@@ -87,43 +87,45 @@ func (n *p2pNetwork) SubscribeAll() error {
...
+		availableSubnets[i], availableSubnets[j] = availableSubnets[j], availableSubnets[i]
+	})
+
+	randomSubnets := availableSubnets[:numSubnets]
+
 	for _, subnet := range randomSubnets {
-		err := n.topicsCtrl.Subscribe(commons.SubnetTopicID(uint64(subnet)))
+		err := n.topicsCtrl.Subscribe(commons.SubnetTopicID(subnet))
 		if err != nil {
 			return fmt.Errorf("could not subscribe to subnet %d: %w", subnet, err)
```
oleg-ssvlabs, line 123: I would probably look into changing how error handling is done here. 

The better way would be to continue iterating the subnet list and joining the errors, then updating the `persistentSubnets` with subnets that were subscribed successfully. With this approach `persistentSubnets` would be in sync with with subnets node is subscribed to and also node would actually attempt to subscribe to the number of requested subnets.

For now, it probably does not matter because `SubscribeRandoms` is only called once and crashes the node on error, so "best effort subscribe" does not make much sense, but if the caller would not crash on error, this suggestion could be a good idea.

oleg-ssvlabs (commented):
gj.
You can fix the nit comment. The other one is rather "potential improvement"

momosh-ssv (approved):

PR #2780, opened 2026-04-07
fix(exporter): return error on SubscribeAll failure

alok-ssv:
Addresses `F-ssv-053`

I dont see any point adding a unit test for this.

### DIFF
--- a/operator/node.go
+++ b/operator/node.go
@@ -196,8 +196,7 @@ func (n *Node) Start(ctx context.Context) error {
 		// For exporter, we want to connect to all subnets.
 		startValidators = func() error {
 			if err := n.net.SubscribeAll(); err != nil {
-				n.logger.Error("failed to subscribe to all subnets", zap.Error(err))
-				return err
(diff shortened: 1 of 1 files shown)


nkryuchkov (dismissed):

operator/node.go
```diff
@@ -195,10 +195,9 @@ func (n *Node) Start(ctx context.Context) error {
 	if n.exporterOptions.Enabled {
 		// For exporter, we want to connect to all subnets.
 		startValidators = func() error {
-			err := n.net.SubscribeAll()
-			if err != nil {
+			if err := n.net.SubscribeAll(); err != nil {
 				n.logger.Error("failed to subscribe to all subnets", zap.Error(err))
-				return nil
+				return err
```
iurii-ssv: Lets not do both: log and propagate, duplicate log line usually resulting from that in such cases can be pretty confusing (the only reason to log in such cases would probably be some additional bulky info we don't want to propagate with the error itself)

iurii-ssv (approved):

momosh-ssv (approved):

PR #2781, opened 2026-04-07
Handle unsupported keys and duty type mismatches (F-ssv-184)

julienh-ssv:
Fix for F-ssv-184
Multiple runner implementations use unchecked type assertions (`.(type)` without `ok` check) on the `CurrentDuty` field and `decidedValue`. Additionally, the P2P key conversion functions `ECDSAPrivFromInterface` and `ECDSAPubFromInterface` perform bare type assertions on peer-provided keys. If the underlying type doesn't match, this causes a panic that crashes the node.

The P2P key conversion case is particularly concerning because `ECDSAPubFromInterface` is called from the discovery path (`DiscV5Service.Node`) where it processes peer-provided public keys. If a peer presents a non-Secp256k1 key (e.g., Ed25519, which libp2p supports), the assertion panics.

Impact: Node crash, which could cause missed duties and potential slashing if the node doesn't restart promptly. The runner assertions are primarily internal consistency issues, but the P2P key conversion panics are directly exploitable by any peer presenting a non-Secp256k1 key.

Recommendation: Use checked type assertions (`value, ok := x.(Type)`) and return appropriate errors instead of panicking. Example:
```go
secpKey, ok := privkey.(*crypto.Secp256k1PrivateKey)
if !ok {
    return nil, fmt.Errorf("unsupported key type: expected Secp256k1, got %T", privkey)
}
```

Fix for https://github.com/ssvlabs/ssv-node-board/issues/762


nkryuchkov (dismissed):

protocol/v2/ssv/runner/committee.go
```diff
@@ -235,23 +235,23 @@ func (r *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Logg
 	r.measurements.EndConsensus()
 	recordConsensusDuration(ctx, r.measurements.ConsensusTime(), spectypes.RoleCommittee)
 
-	duty := r.State.CurrentDuty
+	committeeDuty, err := committeeDutyFromState(r.State)
+	if err != nil {
+		return err
+	}
+	slot := committeeDuty.DutySlot()
```
iurii-ssv: `slot`->`commiteeDutySlot` since `slot` can refer to many different things (message-slot, current-duty slot, estimated-current-slot, etc.).

But also, for consistency, we should either define this helper-var in every runner like that, or just inline `committeeDuty.DutySlot()` instead (TBH inlining in this case would probably be my choice)

protocol/v2/ssv/runner/committee.go
```diff
@@ -270,12 +270,17 @@ func (r *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Logg
...
-		beaconVote = decidedValue.(*spectypes.BeaconVote)
+		beaconVote *spectypes.BeaconVote
 		totalAttesterDuties,
 		totalSyncCommitteeDuties,
 		blockedAttesterDuties atomic.Uint32
 	)
 
+	beaconVote, err = beaconVoteFromEncoder(decidedValue)
+	if err != nil {
+		return err
```
iurii-ssv: Lets try to wrap errors instead of `return err` whenever possible, applies to multiple places in this PR.

It's kind of "out of scope of this PR", but I don't see how else we can add wrap errors consistently if we don't do it gradually over time.

protocol/v2/ssv/runner/committee.go
```diff
@@ -899,6 +904,10 @@ func (r *CommitteeRunner) expectedPostConsensusRootsAndBeaconObjects(ctx context
 	syncCommitteeMap = make(map[phase0.ValidatorIndex][32]byte)
 	beaconObjects = make(map[phase0.ValidatorIndex]map[[32]byte]any)
 	duty := r.State.CurrentDuty
+	committeeDuty, err := committeeDutyFromState(r.State)
+	if err != nil {
+		return nil, nil, nil, err
+	}
```
iurii-ssv: We can delete this `duty := r.State.CurrentDuty` line now, and just re-use `committeeDuty` below

protocol/v2/ssv/runner/type_assertions.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, fmt.Errorf("duty is not a ValidatorDuty: %T", duty)
+	}
+	if validatorDuty == nil {
+		return nil, fmt.Errorf("validator duty is nil")
+	}
+
+	return validatorDuty, nil
+}
+
+func validatorDutyFromState(state *State) (*spectypes.ValidatorDuty, error) {
```
iurii-ssv: For consistency/simplicity, `validatorDutyFromState` should be a method on `{{VALIDTOR_DUTY_TYPE}}Runner` (eg. `ProposerRunner`) and called `currentValidatorDuty`.

We could also place this method on `baseRunner` instead to avoid duplication, but I think duplication is fine in this case.

iurii-ssv (commented):
Great changes! I would ask for some updates though to keep things consistent.

protocol/v2/ssv/runner/type_assertions.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, fmt.Errorf("duty is not a CommitteeDuty: %T", duty)
+	}
+	if committeeDuty == nil {
+		return nil, fmt.Errorf("committee duty is nil")
+	}
+
+	return committeeDuty, nil
+}
+
+func committeeDutyFromState(state *State) (*spectypes.CommitteeDuty, error) {
```
iurii-ssv: For consistency/simplicity, `committeeDutyFromState` should be a method on `CommiteeRunner` and called `currentCommiteeDuty`

protocol/v2/ssv/runner/voluntary_exit.go
```diff
@@ -223,7 +233,12 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
 // Returns *phase0.VoluntaryExit object with current epoch and own validator index
 func (r *VoluntaryExitRunner) calculateVoluntaryExit() (*phase0.VoluntaryExit, error) {
 	epoch := r.NetworkConfig.EstimatedEpochAtSlot(r.State.CurrentDuty.DutySlot())
```
momosh-ssv: Could this still panic on nil `CurrentDuty`? Seems that `r.State.CurrentDuty.DutySlot()` runs before `validatorDutyFromState` can guard against a nil duty, which defeats the safety goal of this PR. Maybe swap the order — fetch `duty` first and use `duty.DutySlot()` for the epoch calc.

protocol/v2/ssv/runner/validator_registration.go
```diff
@@ -178,6 +183,11 @@ func (r *ValidatorRegistrationRunner) executeDuty(ctx context.Context, logger *z
 	// Reuse the existing span instead of generating new one to keep tracing-data lightweight.
 	span := trace.SpanFromContext(ctx)
 
+	validatorDuty, err := validatorDutyFromDuty(duty)
+	if err != nil {
+		return err
+	}
+
 	vr, err := r.buildValidatorRegistration(duty.DutySlot())
```
momosh-ssv: Any reason we're not using `validatorDuty.DutySlot()` here? Every other call site in this function switched to `validatorDuty`, so `duty.DutySlot()` reads as a leftover. Not a bug (the interface exposes `DutySlot`), just consistency.

protocol/v2/ssv/runner/committee.go
```diff
@@ -899,6 +904,10 @@ func (r *CommitteeRunner) expectedPostConsensusRootsAndBeaconObjects(ctx context
 	syncCommitteeMap = make(map[phase0.ValidatorIndex][32]byte)
 	beaconObjects = make(map[phase0.ValidatorIndex]map[[32]byte]any)
 	duty := r.State.CurrentDuty
```
momosh-ssv: Worth considering dropping this `duty` local entirely? It's also used at line 917 (`slot := duty.DutySlot()`); swapping that to `committeeDuty.DutySlot()` would let this line go away, which also addresses the earlier note about removing it.

iurii-ssv (dismissed):

nkryuchkov (dismissed):

network/commons/keys_test.go
oleg-ssvlabs, line 1: unit tests do not cover `nil` key inputs cases

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -132,7 +137,10 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 		return fmt.Errorf("got pre-consensus quorum but it has invalid signatures: %w", err)
 	}
 
-	duty := r.State.CurrentDuty.(*spectypes.ValidatorDuty)
+	duty, err := r.currentValidatorDuty()
+	if err != nil {
+		return err
```
oleg-ssvlabs: all returned errors from this methods are wrapped, so to be consistent, it is better to wrap this error as well. It relates to other changes in the file as well

nkryuchkov (dismissed):

protocol/v2/ssv/runner/proposer.go
```diff
@@ -137,7 +142,10 @@ func (r *ProposerRunner) ProcessPreConsensus(ctx context.Context, logger *zap.Lo
 		return fmt.Errorf("got pre-consensus quorum but it has invalid signatures: %w", err)
 	}
 
-	duty := r.State.CurrentDuty.(*spectypes.ValidatorDuty)
+	duty, err := r.currentValidatorDuty()
+	if err != nil {
```
oleg-ssvlabs, line 146: Better to wrap errors in this file as well (where it needs to be wrapped)

iurii-ssv (dismissed):

protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -155,7 +160,10 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
 		return nil
 	}
 
-	duty := r.State.CurrentDuty.(*spectypes.ValidatorDuty)
+	duty, err := r.currentValidatorDuty()
+	if err != nil {
```
oleg-ssvlabs, line 164: ditto

protocol/v2/ssv/runner/type_assertions_test.go
oleg-ssvlabs, line 1: `type_assertions_test.go` covers happy paths and wrong-type errors, but not the `nil` and typed-nil scenarios that the new helpers explicitly add. Please add cases for `nil` `Duty`/`Encoder`, typed-nil `validator`/`committee` duties, typed-nil `decided` values, and the `nil` `runner/state/current-duty` scenarios for `currentCommitteeDuty`.

oleg-ssvlabs (changes requested):
gj, left few comments

protocol/v2/ssv/runner/committee.go
iurii-ssv: We should use `committeeDutyFromDuty` func instead of this block of code here in `CommitteeRunner.StartNewDuty` to match what we do for Validator-Runners

protocol/v2/ssv/runner/type_assertions.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (b *BaseRunner) currentValidatorDuty() (*spectypes.ValidatorDuty, error) {
+	if b == nil {
+		return nil, fmt.Errorf("runner is nil")
+	}
+	if b.State == nil {
+		return nil, fmt.Errorf("runner state is nil")
+	}
+	if b.State.CurrentDuty == nil {
+		return nil, fmt.Errorf("current duty is nil")
```
momosh-ssv, line 34: Worth considering — many `r.State.CurrentDuty.DutySlot()` direct accesses still remain in `runner_validations.go`, `proposer.go` (`expectedPreConsensusRootsAndDomain` etc.), `sync_committee_contribution.go`, `committee.go`, `aggregator.go` and `validator_registration.go` that bypass these guards. And `runner.go:515` already documents `// assume CurrentDuty is not nil if state is not nil`, so the codebase already relies on that invariant.

I fear the in-between state (guards here, raw access elsewhere) is inconsistent — either extend the helper to those sites, or drop the `State == nil` / `CurrentDuty == nil` checks here since the invariant is already assumed upstream.

protocol/v2/ssv/runner/committee.go
```diff
@@ -235,23 +235,23 @@ func (r *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Logg
 	r.measurements.EndConsensus()
 	recordConsensusDuration(ctx, r.measurements.ConsensusTime(), spectypes.RoleCommittee)
 
-	duty := r.State.CurrentDuty
+	committeeDuty, err := r.currentCommitteeDuty()
+	if err != nil {
+		return fmt.Errorf("current committee duty: %w", err)
```
momosh-ssv, line 240: Could this be simplified? The helper already wraps the inner error with `current duty: %w`, so this ends up producing `current committee duty: current duty: duty is not a CommitteeDuty: ...`. Maybe drop one of the wraps — a bare `return err` here (or removing the inner wrap inside `currentCommitteeDuty`) would read cleaner. Same pattern applies at `expectedPostConsensusRootsAndBeaconObjects` (line ~907) and `voluntary_exit.go` `calculateVoluntaryExit`.

protocol/v2/ssv/runner/committee.go
```diff
@@ -270,12 +270,17 @@ func (r *CommitteeRunner) ProcessConsensus(ctx context.Context, logger *zap.Logg
 		signaturesCh = make(chan *spectypes.PartialSignatureMessage)
 		dutiesCh     = make(chan *spectypes.ValidatorDuty)
 
-		beaconVote = decidedValue.(*spectypes.BeaconVote)
+		beaconVote *spectypes.BeaconVote
```
momosh-ssv: Might be worth pulling `beaconVote` out of this var block — it's a forward declaration that's immediately overwritten a few lines below at L279. A standalone `beaconVote, err := beaconVoteFromEncoder(decidedValue)` would match the pattern the other runners use and read cleaner.

protocol/v2/ssv/runner/voluntary_exit.go
```diff
@@ -159,6 +164,11 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
 	// Reuse the existing span instead of generating new one to keep tracing-data lightweight.
 	span := trace.SpanFromContext(ctx)
 
+	validatorDuty, err := validatorDutyFromDuty(duty)
+	if err != nil {
+		return err
+	}
+
 	voluntaryExit, err := r.calculateVoluntaryExit()
```
momosh-ssv: Seems that `validatorDuty` here comes from the function argument while `calculateVoluntaryExit` independently re-fetches via `r.currentValidatorDuty()` (reading from state). They should match today, but having two sources of truth for the same duty in one function feels inconsistent. Maybe pass `validatorDuty` into `calculateVoluntaryExit` so there's only one path.

protocol/v2/ssv/runner/type_assertions_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	got, err := (&BaseRunner{State: &State{CurrentDuty: committeeDuty}}).currentCommitteeDuty()
+	require.NoError(t, err)
+	require.Same(t, committeeDuty, got)
+
+	_, err = (&BaseRunner{State: &State{CurrentDuty: &spectypes.ValidatorDuty{}}}).currentCommitteeDuty()
+	require.ErrorContains(t, err, "current duty: duty is not a CommitteeDuty")
+}
+
+func TestCalculateVoluntaryExitHandlesMissingCurrentDuty(t *testing.T) {
```
momosh-ssv: Couple of nits on this file:

- `TestCalculateVoluntaryExitHandlesMissingCurrentDuty` is really a `VoluntaryExitRunner` test — might fit better in a `voluntary_exit_test.go` for discoverability.
- `TestValidatorDutyFromState` / `TestCommitteeDutyFromState` above exercise `currentValidatorDuty` / `currentCommitteeDuty` — the `FromState` naming no longer matches what's under test.

network/commons/keys_test.go
julienh-ssv: Added nil-interface and typed-nil coverage for both `ECDSAPrivFromInterface` and `ECDSAPubFromInterface`.

protocol/v2/ssv/runner/aggregator.go
julienh-ssv: Wrapped the helper-returned errors in `aggregator.go` to match the rest of the file.

protocol/v2/ssv/runner/proposer.go
julienh-ssv: Wrapped the helper-returned errors in `proposer.go` for consistency with the surrounding flow.

protocol/v2/ssv/runner/sync_committee_contribution.go
julienh-ssv: Applied the same error-wrapping cleanup in `sync_committee_contribution.go`.

protocol/v2/ssv/runner/type_assertions_test.go
julienh-ssv: Expanded the helper tests for nil, typed-nil, and runner/state/current-duty edge cases, and added the `VoluntaryExitRunner` regression in `voluntary_exit_test.go`.

protocol/v2/ssv/runner/committee.go
julienh-ssv: Switched `CommitteeRunner.StartNewDuty` to use `committeeDutyFromDuty`.

protocol/v2/ssv/runner/type_assertions.go
julienh-ssv: Added `currentDutySlot()` and replaced the remaining raw `CurrentDuty.DutySlot()` reads across the runner package so the guard usage is consistent.

protocol/v2/ssv/runner/committee.go
julienh-ssv: Removed the double wrapping by simplifying the helper internals and keeping the contextual wrap at the call sites.

protocol/v2/ssv/runner/committee.go
julienh-ssv: Pulled `beaconVote` into a standalone assignment instead of the forward declaration in the var block.

protocol/v2/ssv/runner/voluntary_exit.go
julienh-ssv: Changed `calculateVoluntaryExit` to take the already-validated `validatorDuty`, so there is only one duty source in that flow.

protocol/v2/ssv/runner/type_assertions_test.go
julienh-ssv: Moved the `VoluntaryExitRunner`-specific regression into `voluntary_exit_test.go` and renamed the current-duty helper tests to match what they cover.

protocol/v2/ssv/runner/voluntary_exit.go
```diff
@@ -221,9 +236,13 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
...
-func (r *VoluntaryExitRunner) calculateVoluntaryExit() (*phase0.VoluntaryExit, error) {
-	epoch := r.NetworkConfig.EstimatedEpochAtSlot(r.State.CurrentDuty.DutySlot())
-	validatorIndex := r.State.CurrentDuty.(*spectypes.ValidatorDuty).ValidatorIndex
+func (r *VoluntaryExitRunner) calculateVoluntaryExit(duty *spectypes.ValidatorDuty) (*phase0.VoluntaryExit, error) {
+	if duty == nil {
+		return nil, fmt.Errorf("validator duty is nil")
+	}
+	epoch := r.NetworkConfig.EstimatedEpochAtSlot(duty.DutySlot())
+
+	validatorIndex := duty.ValidatorIndex
```
iurii-ssv: The formatting here is kinda hard to read, can we just inline these vars ?

iurii-ssv (dismissed):

protocol/v2/ssv/runner/voluntary_exit.go
julienh-ssv: Inlined the temporary fields in `calculateVoluntaryExit` to make the return block easier to scan.

iurii-ssv (approved):

oleg-ssvlabs (approved):
🥇

ljuba-ssv (approved):

PR #2785, opened 2026-04-08
dutystore: add regression coverage for independent duty-type locks

nkryuchkov:
## Summary

`F-ssv-160`

Adds a regression test for the duty store locking model.

The current `dutystore` implementation already keeps separate stores for attester, proposer, sync committee, and voluntary exit duties, with each store using its own `sync.RWMutex`. This PR adds coverage to make that behavior explicit and prevent regressions back to cross-duty-type lock contention.

## What changed

- Added `TestStoreDutyTypesUseIndependentLocks` in [duties_test.go](/Users/bloxmac/GolandProjects/ssv/operator/duties/dutystore/duties_test.go)
- The test holds the attester write lock and verifies that:
  - proposer reads still complete
  - sync committee reads still complete
  - voluntary exit writes still complete
- Added a small timeout helper used by the regression test

## Why

The `F-ssv-160` finding describes the duty store as if all duty types shared a single mutex. That does not match the current implementation on this branch. The new test documents the intended concurrency boundary and guards against future regressions.

## Verification

- Run `go test ./operator/duties/dutystore`

### DIFF
--- a/operator/duties/dutystore/duties_test.go
+++ b/operator/duties/dutystore/duties_test.go
@@ -1,6 +1,7 @@
 package dutystore
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nkryuchkov:
@greptileai review it again

operator/duties/dutystore/duties_test.go
```diff
@@ -56,3 +58,68 @@ func TestDutiesEraseEpochData(t *testing.T) {
...
+		}
+		return nil
+	})
+	assertCompletesWithin(t, "voluntary exit write", func() error {
+		store.VoluntaryExit.AddDuty(slot, pk)
+		if count := store.VoluntaryExit.GetDutyCount(slot, pk); count != 1 {
+			return fmt.Errorf("expected voluntary exit count 1, got %d", count)
+		}
+		return nil
+	})
```
olegshmuelov: Nit (non-blocking): The test holds the attester lock and verifies the other three types are independent — which catches the primary regression (someone introducing a shared lock). A more exhaustive version would hold each duty type's lock in turn (subtests per lock holder), covering the full N-to-N independence matrix. Optional enhancement if you want to be thorough, but the current coverage is sufficient for the stated goal.

operator/duties/dutystore/duties_test.go
```diff
@@ -56,3 +58,143 @@ func TestDutiesEraseEpochData(t *testing.T) {
...
+			unlock: func(store *Store) {
+				store.VoluntaryExit.mu.Unlock()
+			},
+			assertOpen: func(store *Store) error {
+				store.VoluntaryExit.AddDuty(slot, pk)
+				if count := store.VoluntaryExit.GetDutyCount(slot, pk); count != 1 {
+					return fmt.Errorf("expected voluntary exit count 1, got %d", count)
+				}
+				return nil
+			},
```
momosh-ssv, line 164: Might be worth making this assertion tolerate more than one call. The `count != 1` check assumes `voluntary exit` is never invoked as the `other` more than once per subtest — which is true today because `newStoreWithDuties()` is fresh per subtest and the outer loop skips the holder. 

Seems a little fragile if someone later reorders the loop or adds another call site. 
Maybe `count >= 1`, or pre-seed a duty and do a read-only check like the other types do.

momosh-ssv (dismissed):
looks good, gj!

iurii-ssv (dismissed):

iurii-ssv (approved):

olegshmuelov (approved):

ljuba-ssv (approved):

PR #2786, opened 2026-04-08
tests: resolve flaky p2p-discovery tests

iurii-ssv:
Saw this [pipeline fail recently](https://github.com/ssvlabs/ssv/actions/runs/24090131290/job/70273906810?pr=2756):
```
--- FAIL: TestFairMixIterator_PreForkEmpty (0.01s)
    forking_dv5_listener_test.go:177: 
        	Error Trace:	/home/runner/work/ssv/ssv/network/discovery/forking_dv5_listener_test.go:177
        	            				/home/runner/work/ssv/ssv/network/discovery/iterator_test.go:114
        	Error:      	expected next to be %v
        	Test:       	TestFairMixIterator_PreForkEmpty
        	Messages:   	true
FAIL
coverage: 65.4% of statements
FAIL	github.com/ssvlabs/ssv/network/discovery	1.598s
```
This PR bumps the deadlines these tests use up to `400ms` - which is experimentally verified stable/reliable sleep-amount, this should solve flakyness.

### DIFF
--- a/network/discovery/forking_dv5_listener_test.go
+++ b/network/discovery/forking_dv5_listener_test.go
@@ -167,5 +167,5 @@ func requireNextTimeout(t *testing.T, expected bool, iter enode.Iterator, timeou
 	case <-ctx.Done():
 	}
 
-	assert.Equalf(t, expected, got, "expected iter.Next to be %v", expected)
+	require.Equalf(t, expected, got, "expected iter.Next to be %v", expected)
 }
--- a/network/discovery/forking_dv5_listener_test.go
+++ b/network/discovery/forking_dv5_listener_test.go
@@ -65,7 +65,7 @@ func TestForkListener_RandomNodes(t *testing.T) {
 	assert.Equal(t, nodes[1], nodeFromPostForkListener)
 
 	// No more next
-	requireNextTimeout(t, false, iter, 10*time.Millisecond)
+	requireNextTimeout(t, false, iter, 400*time.Millisecond)
 }
 
 func TestForkListener_AllNodes(t *testing.T) {
--- a/network/discovery/forking_dv5_listener_test.go
+++ b/network/discovery/forking_dv5_listener_test.go
@@ -11,7 +11,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const iteratorTimeout = 5 * time.Millisecond
+const iteratorTimeout = 500 * time.Millisecond
 
 func TestForkListener_Create(t *testing.T) {
 	localNode := NewLocalNode(t)
--- a/network/discovery/forking_dv5_listener_test.go
+++ b/network/discovery/forking_dv5_listener_test.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"context"
 	"testing"
 	"time"
 
@@ -51,12 +52,12 @@ func TestForkListener_RandomNodes(t *testing.T) {
 
 	forkListener := NewForkingDV5Listener(zap.NewNop(), preForkListener, postForkListener, iteratorTimeout)
 
-	iter := forkListener.RandomNodes()
-	defer iter.Close()
+	iterator := forkListener.RandomNodes()
+	defer iterator.Close()
 	var nodes []*enode.Node
 	for i := 0; i < 2; i++ {
-		require.True(t, iter.Next())
-		nodes = append(nodes, iter.Node())
+		require.True(t, iterator.Next())
+		nodes = append(nodes, iterator.Node())
 	}
 
 	assert.Len(t, nodes, 2)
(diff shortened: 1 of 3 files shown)


network/discovery/forking_dv5_listener_test.go
```diff
@@ -65,7 +65,7 @@ func TestForkListener_RandomNodes(t *testing.T) {
 	assert.Equal(t, nodes[1], nodeFromPostForkListener)
 
 	// No more next
-	requireNextTimeout(t, false, iter, 10*time.Millisecond)
+	requireNextTimeout(t, false, iter, 400*time.Millisecond)
```
olegshmuelov: The "optimize" commit differentiated `expected==true` (400ms) vs `expected==false` (50ms) in `iterator_test.go`, but this `expected==false` check was left at 400ms. The iterator is already exhausted here, 50ms is sufficient to confirm no more nodes, and matches the pattern in the other file.

Suggestion:
```go
requireNextTimeout(t, false, iter, 50*time.Millisecond)
```

network/discovery/forking_dv5_listener_test.go
olegshmuelov, line 150: Pre-existing: when `deadline` fires and the main goroutine's `select` receives it first, main returns. The background goroutine is in `iter.Next()` or `time.Sleep`. When it resumes and tries `next <- ok`, the unbuffered channel has no reader. The goroutine's `case <-deadline:` can't fire (value already consumed by main). The goroutine blocks forever — a leak.

Not introduced by this PR. `MockIterator.Next()` is mutex-protected, `enode.FairMix` has internal sync, so the leaked goroutine doesn't cause data races. Cleaned up at test process exit. A proper fix would use `context.Context` for cancellation.

olegshmuelov: Pre-existing, extremely unlikely: `time.After(timeout)` sends once. The background goroutine's `case <-deadline:` at L159 and the main goroutine's `case <-deadline:` at L175 race for the same value. If the background goroutine wins, it exits and `defer close(next)` fires. Main then reads `false` from the closed channel infinitely (closed channels return zero value without blocking). Since `<-deadline` at L175 was consumed by L159, main has no exit path for `expected==false` cases — infinite tight loop.

In practice, the background goroutine spends most time in `time.Sleep(timeout/10)`, so main almost always receives the deadline first. Never observed in CI. Same fix as https://github.com/ssvlabs/ssv/pull/2786/changes#r3057275136 -  `context.Context` cancellation.

network/discovery/forking_dv5_listener_test.go
iurii-ssv: Missed that one, fixed now 👍

network/discovery/forking_dv5_listener_test.go
iurii-ssv: This should help https://github.com/ssvlabs/ssv/pull/2786/changes/486c436d0df14d7a9e683ba56515817581420d11

iurii-ssv: Think this should address it https://github.com/ssvlabs/ssv/pull/2786/changes/486c436d0df14d7a9e683ba56515817581420d11

iurii-ssv:
Spotted [another type of flaky pipelines related to these tests](https://github.com/ssvlabs/ssv/actions/runs/24252632599/job/70815841735?pr=2720), https://github.com/ssvlabs/ssv/pull/2786/changes/53d1490e4394d46c991904c8e4554fa72c0f097a should fix it I think

network/discovery/forking_dv5_listener_test.go
```diff
@@ -147,36 +148,24 @@ func TestForkListener_Close(t *testing.T) {
...
+	}()
+
+	got := false
+	select {
+	case next := <-result:
+		got = next
+	case <-ctx.Done():
 	}
+
+	assert.Equalf(t, expected, got, "expected iter.Next to be %v", expected)
```
olegshmuelov: Function is named `requireNextTimeout` but uses `assert.Equalf` (non-fatal). If the timeout path fires on an `expected=true` call (e.g., in `iterator_test.go`), the test continues and calls `iterator.Node()` while the abandoned goroutine is still inside `FairMix.Next()` — `m.cur` is accessed without synchronization, which is a data race. Practically impossible to trigger with MockIterator (microsecond return vs 400ms timeout), but `require.Equalf` matches the function name, stops the test on failure, and prevents the theoretical race.

Suggestion:
```go
require.Equalf(t, expected, got, "expected iter.Next to be %v", expected)
```

olegshmuelov (dismissed):
one minor suggestion, not blocking

olegshmuelov (approved):

nkryuchkov (approved):

momosh-ssv (approved):

PR #2787, opened 2026-04-08
network/topics: add topic validator timeout (F-ssv-126)

julienh-ssv:
F-ssv-126 P1 Triaged
No validation timeout on topic validators
ssv · Performance / DoS · Assigned: dev4 · Effort: —
When registering topic validators, the code explicitly comments out the validation timeout:
```go
// Optional: set a timeout for message validation
// opts = append(opts, pubsub.WithValidatorTimeout(time.Second))
```

Without a timeout, a slow or malicious message that triggers expensive validation (e.g., signature verification, committee lookups) can block the validation goroutine pool indefinitely. With `validateThrottle = 8192` goroutines, a sustained attack could exhaust all validation goroutines, causing legitimate messages to queue up and eventually be dropped.

Impact: Validation goroutine exhaustion. Legitimate messages are delayed or dropped, degrading duty execution quality.

Recommendation: Uncomment and set a reasonable validation timeout (e.g., 3-5 seconds). The message validation pipeline already has its own internal timeout handling.

Fixes https://github.com/ssvlabs/ssv-node-board/issues/682


iurii-ssv (dismissed):
Looks reasonable, linter needs fixing though.

iurii-ssv (dismissed):

olegshmuelov (changes requested):
The timeout value and API usage are correct, but `WithValidatorTimeout` is unenforceable right now — the validation pipeline never checks the context. In `validateSingleTopic` (go-libp2p-pubsub v0.15.0), validation runs synchronously and the throttle slot is held until return regardless of the deadline.

To make this effective, thread `ctx` through the validation chain and add early-out checks before the expensive operations:

1. `handlePubsubMessage(ctx, pmsg, time.Now())` — propagate ctx
2. `handleSignedSSVMessage(ctx, ...)` — propagate ctx
3. Check `ctx.Err()` before `signatureVerifier.VerifySignature` (the most expensive op)
4. Optionally check `ctx.Err()` before `validationMu.Lock()` (the mutex isn't context-aware, but we can bail before waiting)

No need to make the mutex itself context-aware — checking `ctx.Err()` before acquiring it is enough. That plus the check before `VerifySignature` covers the two expensive paths and makes the timeout effective.

Without this, the goroutine pool exhaustion described in F-ssv-126 isn't actually mitigated.

olegshmuelov (dismissed):
LGTM!

message/validation/validation_test.go
```diff
@@ -164,10 +164,119 @@ func Test_ValidateSSVMessage(t *testing.T) {
 
 		receivedAt := netCfg.SlotStartTime(slot)
 		topicID := commons.CommitteeTopicID(spectypes.CommitteeID(signedSSVMessage.SSVMessage.GetID().GetDutyExecutorID()[16:]))[0]
-		_, err = validator.handleSignedSSVMessage(signedSSVMessage, topicID, peerID, receivedAt)
+		_, err = validator.handleSignedSSVMessage(context.Background(), signedSSVMessage, topicID, peerID, receivedAt)
```
nkryuchkov: can we use `t.Context()`?

message/validation/validation_test.go
```diff
@@ -164,10 +164,119 @@ func Test_ValidateSSVMessage(t *testing.T) {
...
+
+		errCh := make(chan error, 1)
+		go func() {
+			_, err := validator.handleSignedSSVMessage(ctx, signedSSVMessage, topicID, peerID, receivedAt)
+			errCh <- err
+		}()
+
+		select {
+		case err := <-errCh:
+			require.Error(t, err)
```
nkryuchkov: Is this check redundant? The one below seems to cover it

nkryuchkov (dismissed):

iurii-ssv (dismissed):
A minor rearrangement would be nice: https://github.com/ssvlabs/ssv/pull/2799

message/validation/validation.go
```diff
@@ -122,9 +122,13 @@ func (mv *messageValidator) Validate(ctx context.Context, peerID peer.ID, pmsg *
 		return mv.validateSelf(pmsg)
 	}
 
+	if err := ctx.Err(); err != nil {
+		return mv.handleValidationError(ctx, peerID, nil, err)
```
momosh-ssv, line 126: Seems that when `ctx.Err()` propagates here, `handleValidationError` won't match the `Error` type, so it falls into the generic branch at `errors.go:150` — recording `err.Error()` (i.e. `"context deadline exceeded"`) as the metric label. If the 5s timeout fires frequently under load, that raw string blends in with legitimately invalid messages and makes it hard for operators to spot timeout pressure.

Maybe worth adding an explicit branch in `handleValidationError` for `errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)` with a dedicated metric label like `"validation_timeout"` and a debug log that calls it out.

network/topics/controller.go
```diff
@@ -42,6 +42,8 @@ type messageValidator interface {
 	ValidatorForTopic(topic string) func(ctx context.Context, p peer.ID, pmsg *pubsub.Message) pubsub.ValidationResult
 }
 
+const topicValidatorTimeout = 5 * time.Second
```
momosh-ssv, line 48: What do you think about adding a short comment here on the rationale behind the 5s value? I fear future readers (or someone tuning under load) won't know the tradeoff space — too tight means false timeouts on legitimate messages, too loose means the DoS window stays wide.

Something like:
```go
// topicValidatorTimeout bounds how long a single message validation may run.
// Budget: committee checks + up to N BLS signature verifications + lock wait.
// Timeouts surface as ValidationIgnore (message is dropped, not rejected).
const topicValidatorTimeout = 5 * time.Second
```

message/validation/validation.go
iurii-ssv: Not sure this is an issue, since "context deadline exceeded" is something that should in theory never really happen - so if it does, we'd want to see it and investigate. And it doesn't really "dilute" the rest of the recorded errors, does it ?

message/validation/validation_test.go
julienh-ssv: Switched the touched test calls over to `t.Context()` so the added timeout-path coverage follows the current test style.

message/validation/validation_test.go
julienh-ssv: Removed the redundant `require.Error` checks and kept the stronger `require.ErrorIs` assertions only.

message/validation/validation.go
julienh-ssv: Added explicit timeout/cancellation handling in `handleValidationError`, with dedicated discard reasons and debug logs so validation pressure is visible in metrics instead of blending into generic invalid-message labels.

network/topics/controller.go
julienh-ssv: Added a short comment on `topicValidatorTimeout` to document the tradeoff the 5s bound is protecting.

nkryuchkov (approved):

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2789, opened 2026-04-09
p2p: add more p2p logs (F-ssv-271)

julienh-ssv:
Add more p2p logs
F-ssv-271

fixes https://github.com/ssvlabs/ssv-node-board/issues/997

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -157,12 +157,12 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
 // if we reached peers limit, make sure to accept peers with more than 1 shared subnet,
 // which lets other components to determine whether we'll want to connect to this node or not.
 func (dvs *DiscV5Service) Bootstrap(handler HandleNewPeer) error {
-	const logFrequency = 50
+	const logWindowSize = 50
 
 	var (
-		skippedPeersWindow uint64
-		skippedByReason    = map[skipReason]uint64{}
-		skippedUnknown     uint64
+		skippedPeersTotalPerWindow   uint64
+		skippedByReasonPerWindow     = map[skipReason]uint64{}
+		skippedPeersUnknownPerWindow uint64
 	)
 
 	dvs.discover(
@@ -170,15 +170,15 @@ func (dvs *DiscV5Service) Bootstrap(handler HandleNewPeer) error {
 		func(e PeerEvent) {
 			err := dvs.checkPeer(dvs.ctx, e)
 			if err != nil {
-				skippedPeersWindow++
+				skippedPeersTotalPerWindow++
 				var skipErr *peerSkipError
 				if errors.As(err, &skipErr) {
-					skippedByReason[skipErr.reason]++
+					skippedByReasonPerWindow[skipErr.reason]++
 				} else {
-					skippedUnknown++
+					skippedPeersUnknownPerWindow++
 				}
-				if skippedPeersWindow >= logFrequency {
-					summaryFields := discoverySkipSummaryFields(skippedPeersWindow, skippedByReason, skippedUnknown)
+				if skippedPeersTotalPerWindow >= logWindowSize {
+					summaryFields := discoverySkipSummaryFields(skippedPeersTotalPerWindow, skippedByReasonPerWindow, skippedPeersUnknownPerWindow)
 					summaryFields = append(summaryFields,
 						zap.Error(err),
 						fields.PeerID(e.AddrInfo.ID),
@@ -186,9 +186,9 @@ func (dvs *DiscV5Service) Bootstrap(handler HandleNewPeer) error {
 						zap.Int("trimmed_recently_size", dvs.trimmedRecently.SlowLen()),
 					)
 					dvs.logger.Debug("discovery skipped peers summary", summaryFields...)
-					skippedPeersWindow = 0
-					skippedByReason = map[skipReason]uint64{}
-					skippedUnknown = 0
+					skippedPeersTotalPerWindow = 0
+					skippedByReasonPerWindow = map[skipReason]uint64{}
+					skippedPeersUnknownPerWindow = 0
 				}
 				return
 			}
@@ -220,17 +220,14 @@ func newPeerSkipError(reason skipReason, err error) error {
 
 func discoverySkipSummaryFields(skippedPeers uint64, skippedByReason map[skipReason]uint64, skippedUnknown uint64) []zap.Field {
 	fieldsOut := []zap.Field{
-		zap.Uint64("skipped_peers", skippedPeers),
+		zap.Uint64("skipped_peers_total", skippedPeers),
 	}
 	if skippedUnknown > 0 {
 		fieldsOut = append(fieldsOut, zap.Uint64("skipped_unknown_reason", skippedUnknown))
 	}
 
 	reasons := make([]string, 0, len(skippedByReason))
-	for reason, count := range skippedByReason {
-		if count == 0 {
-			continue
-		}
+	for reason := range skippedByReason {
 		reasons = append(reasons, string(reason))
 	}
 	sort.Strings(reasons)
@@ -247,6 +244,7 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 		var err error
 		pid, err = PeerID(e.Node)
 		if err != nil {
+			recordPeerSkipped(ctx, skipReasonInvalidPeerID)
 			return newPeerSkipError(skipReasonInvalidPeerID, errors.Wrap(err, "could not get peer ID from node record"))
 		}
 	}
@@ -255,6 +253,7 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 	peerDiscoveriesCounter.Add(ctx, 1)
 	nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyDomainType)
 	if err != nil {
+		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, errors.Wrap(err, "could not read domain type"))
 	}
 	if dvs.ssvConfig.DomainType != nodeDomainType {
@@ -265,6 +264,7 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 	// Get the peer's subnets, skipping if it has none.
 	peerSubnets, err := records.GetSubnetsEntry(e.Node.Record())
 	if err != nil {
+		recordPeerSkipped(ctx, skipReasonInvalidSubnets)
 		return newPeerSkipError(skipReasonInvalidSubnets, fmt.Errorf("could not read subnets: %w", err))
 	}
 	if commons.ZeroSubnets == peerSubnets {
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -156,6 +156,9 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
 // Bootstrap start looking for new nodes, note that this function blocks.
 // if we reached peers limit, make sure to accept peers with more than 1 shared subnet,
 // which lets other components to determine whether we'll want to connect to this node or not.
+//
+// All peer filtering is done inside checkPeer (not as discover() filters) so that
+// every skip reason is tracked with structured metrics and windowed summary logs.
 func (dvs *DiscV5Service) Bootstrap(handler HandleNewPeer) error {
 	const logWindowSize = 50
 
@@ -506,7 +509,12 @@ func (dvs *DiscV5Service) PublishENR() {
 	pings, errs := 0, 0
 	peerIDs := map[peer.ID]struct{}{}
 
-	// Publish ENR.
+	// Publish ENR by pinging random SSV nodes so they learn our updated record.
+	// Minimal filtering: we only require valid SSV nodes that aren't marked bad.
+	// We intentionally omit connection-state filters (alreadyConnected, recentlyTrimmed)
+	// because wider propagation is more important — pings are cheap and already-connected
+	// peers should also learn about our ENR update. Subnet filters are also omitted
+	// because ENR publication is a global broadcast, not subnet-specific.
 	dvs.discover(ctx, func(e PeerEvent) {
 		_, err := dvs.dv5Listener.Ping(e.Node)
 		if err != nil {
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sort"
 	"time"
 
 	"github.com/ethereum/go-ethereum/p2p/discover"
 	"github.com/ethereum/go-ethereum/p2p/discover/v5wire"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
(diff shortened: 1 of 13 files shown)


network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
 func (dvs *DiscV5Service) Bootstrap(handler HandleNewPeer) error {
-	// Log every 10th skipped peer.
-	const logFrequency = 10
-	var skippedPeers uint64 = 0
+	const logFrequency = 50
+
+	var (
+		skippedPeersWindow uint64
+		skippedByReason    = map[skipReason]uint64{}
+		skippedUnknown     uint64
```
iurii-ssv: Maybe for completeness
- `logFrequency`->`logWindowSize`
- `skippedPeersWindow`->`skippedPeersTotalPerWindow`
- `skippedUnknown`->`skippedPeersUnknownPerWindow`

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+		fieldsOut = append(fieldsOut, zap.Uint64("skipped_unknown_reason", skippedUnknown))
+	}
+
+	reasons := make([]string, 0, len(skippedByReason))
+	for reason, count := range skippedByReason {
+		if count == 0 {
+			continue
+		}
+		reasons = append(reasons, string(reason))
+	}
```
iurii-ssv: The `if count == 0 {` condition should never be `true` ? So I would omit it for simplicity

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+}
+
+func newPeerSkipError(reason skipReason, err error) error {
+	return &peerSkipError{reason: reason, err: err}
+}
+
+func discoverySkipSummaryFields(skippedPeers uint64, skippedByReason map[skipReason]uint64, skippedUnknown uint64) []zap.Field {
+	fieldsOut := []zap.Field{
+		zap.Uint64("skipped_peers", skippedPeers),
+	}
```
iurii-ssv, line 224: Nit: `skipped_peers`->`skipped_peers_total`

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+}
+
 func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
+	pid := e.AddrInfo.ID
+	if pid == "" {
+		var err error
+		pid, err = PeerID(e.Node)
+		if err != nil {
+			return newPeerSkipError(skipReasonInvalidPeerID, errors.Wrap(err, "could not get peer ID from node record"))
+		}
```
iurii-ssv, line 249: Seems like we need to add `recordPeerSkipped(ctx, skipReasonInvalidPeerID)` here too ?

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+	}
+
 	// Get the peer's domain type, skipping if it mismatches ours.
 	peerDiscoveriesCounter.Add(ctx, 1)
 	nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyDomainType)
 	if err != nil {
-		return errors.Wrap(err, "could not read domain type")
+		return newPeerSkipError(skipReasonInvalidDomainType, errors.Wrap(err, "could not read domain type"))
 	}
```
iurii-ssv, line 268: Seems like we need to add `recordPeerSkipped(ctx, skipReasonInvalidDomainType)` here too ?

network/p2p/p2p.go
```diff
@@ -387,6 +394,7 @@ func (n *p2pNetwork) peersTrimming() func() {
 			zap.Int("outbound_peers_before_trim", outboundBefore),
 			zap.Int("inbound_peers_after_trim", inboundAfter),
 			zap.Int("outbound_peers_after_trim", outboundAfter),
+			zap.Int("trimmed_recently_size", n.trimmedRecently.SlowLen()),
```
iurii-ssv, line 401: Let's also add the `zap.Bool("trim_inbound_only", trimInboundOnly),` here too

network/p2p/p2p.go
```diff
@@ -376,6 +376,13 @@ func (n *p2pNetwork) peersTrimming() func() {
 
 		inboundBefore, outboundBefore := n.connectionStats()
 		peersToTrim := n.choosePeersToTrim(maxPeersToDrop, trimInboundOnly)
+		if len(peersToTrim) == 0 {
+			n.logger.Debug("no peers selected for trimming",
+				zap.Bool("trim_inbound_only", trimInboundOnly),
+				zap.Int("connected_peers", len(n.host.Network().Peers())),
+			)
```
iurii-ssv, line 386: Should probably log more info here, like:
```go
			n.logger.Debug(
				"trimmed peers",
				zap.Int("inbound_peers", inboundBefore),
				zap.Int("outbound_peers", outboundBefore),
				zap.Bool("trim_inbound_only", trimInboundOnly),
				zap.Int("trimmed_recently_size", n.trimmedRecently.SlowLen()),
			)
```

network/p2p/p2p.go
```diff
@@ -441,6 +450,15 @@ func (n *p2pNetwork) choosePeersToTrim(trimCnt int, trimInboundOnly bool) map[pe
 			if connDir == p2pnet.DirOutbound && trimInboundOnly {
 				continue
 			}
+			peerSubnets, _ := n.idx.GetPeerSubnets(p)
+			sharedSubnets := ownSubnets.SharedSubnets(peerSubnets)
+			n.logger.Debug("selected peer for trimming",
+				fields.PeerID(p),
+				zap.Float64("peer_trim_score", peerScores[p]),
```
iurii-ssv: `peer_trim_score`->`peer_score` since there is no such thing as "trim score"

network/p2p/p2p_discovery.go
```diff
@@ -150,10 +165,60 @@ func (n *p2pNetwork) startDiscovery() error {
...
+			waited := time.Since(discoveredPeer.LastTry)
+			if waited < retryCooldownMin {
+				stats.cooldownBlockedCandidates++
+				return true
+			}
+			const retryCooldownMax = 300 * time.Second
+			cooldown := min(retryCooldownMax, retryCooldownMin*time.Duration(discoveredPeer.Tries))
+			peerRelevance := min(1, float64(waited)/float64(cooldown))
+			peerScore *= peerRelevance * peerRelevance
+		}
```
iurii-ssv: Can we create a helper-func that does this calculation (so we can dedup it) ?

network/p2p/p2p_discovery.go
```diff
@@ -9,11 +9,20 @@ import (
...
 	"github.com/ssvlabs/ssv/observability/log/fields"
 	"github.com/ssvlabs/ssv/utils/async"
 )
 
+type peerSelectionPoolStats struct {
+	poolSize                  int
+	readyCandidates           int
+	cooldownBlockedCandidates int
+	zeroScoreCandidates       int
+	positiveScoreCandidates   int
```
iurii-ssv: For consistencyt/completeness, since `cooldownBlockedCandidates` aren't scored:
- `zeroScoreCandidates`->`zeroScoreReadyCandidates`
- `positiveScoreCandidates`->`positiveScoreReadyCandidates`

and same change should apply to logged lines (`zero_score_candidates`, `positive_score_candidates`)

network/p2p/p2p_discovery.go
```diff
@@ -150,10 +165,60 @@ func (n *p2pNetwork) startDiscovery() error {
...
+			return
+		}
 		n.logger.Info("proposed discovered peers",
 			zap.Int("count", len(peersToConnect)),
+			zap.Int("pool_size", poolStats.poolSize),
+			zap.Int("ready_candidates", poolStats.readyCandidates),
+			zap.Int("cooldown_blocked_candidates", poolStats.cooldownBlockedCandidates),
+			zap.Int("zero_score_candidates", poolStats.zeroScoreCandidates),
+			zap.Int("positive_score_candidates", poolStats.positiveScoreCandidates),
 		)
```
iurii-ssv: Let's maybe create a helper-func to convert all these `poolStats` fields into logger-fields (to deduplicate all these)

iurii-ssv (commented):
Valuable stuff, left some suggestions to enhance it a bit further

network/p2p/observability.go
```diff
@@ -93,8 +93,10 @@ func recordPeerCountPerTopic(ctx context.Context, logger *zap.Logger, ctrl topic
 			subnetPeerCounts = append(subnetPeerCounts, len(peers))
 			if len(peers) == 0 {
 				deadSubnets++
-			} else if len(peers) <= 2 {
-				unhealthySubnets++
+			} else if len(peers) == 1 {
+				soloSubnets++
+			} else if len(peers) == 2 {
+				duoSubnets++
 			}
```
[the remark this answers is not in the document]
iurii-ssv: I don't think anybody relies on this.

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
 			}
 			handler(e)
 		},
 		defaultDiscoveryInterval,
 		dvs.ssvNodeFilter(),
-		dvs.sharedSubnetsFilter(1),
-		dvs.alreadyDiscoveredFilter(),
-		dvs.badNodeFilter(),
-		dvs.alreadyConnectedFilter(),
-		dvs.recentlyTrimmedFilter(),
```
iurii-ssv, line 185: Might as well do `ssvNodeFilter` check inside of `checkPeer` for completeness (also recording the corresponding metric)

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x doesn't match %x", nodeDomainType, dvs.ssvConfig.DomainType))
 	}
 
 	// Get the peer's subnets, skipping if it has none.
 	peerSubnets, err := records.GetSubnetsEntry(e.Node.Record())
 	if err != nil {
-		return fmt.Errorf("could not read subnets: %w", err)
+		return newPeerSkipError(skipReasonInvalidSubnets, fmt.Errorf("could not read subnets: %w", err))
 	}
 	if commons.ZeroSubnets == peerSubnets {
```
olegshmuelov, line 280: Missing `recordPeerSkipped(ctx, skipReasonInvalidSubnets)` here. Same pattern gap as invalidPeerID and invalidDomainType that @iurii-ssv flagged. All three new skip reasons need the OTel metric call for consistency.

Suggestion:
```go
peerSubnets, err := records.GetSubnetsEntry(e.Node.Record())
if err != nil {
    recordPeerSkipped(ctx, skipReasonInvalidSubnets)
    return newPeerSkipError(skipReasonInvalidSubnets, fmt.Errorf("could not read subnets: %w", err))
}
```

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
-		return errors.New("peer already discovered recently")
+		return newPeerSkipError(skipReasonNoSharedSubnets, fmt.Errorf("no shared subnets: own=%s peer=%s", dvs.subnets.StringHumanReadable(), peerSubnets.StringHumanReadable()))
+	}
+
+	if dvs.discoveredPeersPool.Has(pid) {
+		recordPeerSkipped(ctx, skipReasonAlreadyDiscovered)
+		return newPeerSkipError(skipReasonAlreadyDiscovered, errors.New("peer already discovered recently"))
+	}
+
+	dvs.subnetsIdx.UpdatePeerSubnets(pid, peerSubnets)
```
olegshmuelov: `UpdatePeerSubnets` was previously called immediately after subnet validation. Now it runs after bad/connected/trimmed/shared/discovered checks, meaning these peers won't have their subnet index refreshed.

For already-connected peers, this means subnet changes (e.g., peer starts subscribing to new subnets) won't be picked up during re-discovery — only through other code paths. This is probably fine since connected peers have their own subnet update mechanisms, but worth a comment noting the intentional change.

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+	if pid == "" {
+		var err error
+		pid, err = PeerID(e.Node)
+		if err != nil {
+			return newPeerSkipError(skipReasonInvalidPeerID, errors.Wrap(err, "could not get peer ID from node record"))
+		}
+	}
+
 	// Get the peer's domain type, skipping if it mismatches ours.
 	peerDiscoveriesCounter.Add(ctx, 1)
```
olegshmuelov, line 263: `peerDiscoveriesCounter` semantics changed. Previously, external filters (shared subnets, already discovered, bad, connected, recently trimmed) ran before `checkPeer`, so the counter only counted post-filter peers. Now only `ssvNodeFilter` runs before, so the counter counts significantly more peers.

This makes the counter more semantically correct (it measures "SSV peer discoveries" not "post-filter discoveries"), but any monitoring threshold or alert based on the old baseline will see a step change. Worth noting in the PR description.

network/discovery/dv5_service.go
```diff
@@ -155,75 +157,153 @@ func (dvs *DiscV5Service) Node(logger *zap.Logger, info peer.AddrInfo) (*enode.N
...
+				if errors.As(err, &skipErr) {
+					skippedByReason[skipErr.reason]++
+				} else {
+					skippedUnknown++
+				}
+				if skippedPeersWindow >= logFrequency {
+					summaryFields := discoverySkipSummaryFields(skippedPeersWindow, skippedByReason, skippedUnknown)
+					summaryFields = append(summaryFields,
+						zap.Error(err),
+						fields.PeerID(e.AddrInfo.ID),
```
olegshmuelov: The summary log aggregates 50 skipped peers but includes `zap.Error(err)` and `fields.PeerID(e.AddrInfo.ID)` from only the 50th peer (the one that triggered the window flush). This could be misleading — a reader might think the logged error is the primary skip reason when it's just the last one.

Consider either dropping the per-peer fields from the summary (the reason breakdown already tells the story) or explicitly labeling them as "last_error" / "last_peer_id".

network/p2p/p2p_discovery.go
```diff
@@ -150,10 +165,60 @@ func (n *p2pNetwork) startDiscovery() error {
...
 	})
 
 	return nil
 }
+
+func (n *p2pNetwork) peerSelectionPoolStats(ownSubnets commons.Subnets, currentSubnetPeers SubnetPeers) peerSelectionPoolStats {
```
olegshmuelov: `peerSelectionPoolStats` does a full `Range` over the discovery pool, calling `GetPeerSubnets` and `Score` for each peer. The actual peer selection loop (starting ~line 95) does the same iteration. This doubles the work per selection cycle.

For typical pool sizes (hundreds to low thousands) this is negligible. But if the pool grows large, the doubled iteration adds latency to the selection tick. Worth keeping in mind if pool sizes grow — could be optimized by collecting stats during the existing selection loop instead of a separate pass.

This is not blocking. The simplicity of a separate stats function has value.

iurii-ssv:
@julienh-ssv I've created a PR into this one to bundle minor suggestions together: https://github.com/ssvlabs/ssv/pull/2790 (and marked the related discussions I've had open as resolved)

network/p2p/p2p_discovery.go
iurii-ssv: Yeah, since this runs in "background" and the pool-size is unlikely to grow beyond several thousands (plus I think we trim it periodically) it's better to keep the "duplicate computation" (vs complicating this code)

network/discovery/dv5_service.go
iurii-ssv: Oh, missed that one, I've added the fix for it to https://github.com/ssvlabs/ssv/pull/2790 - so I'll resolve this discussion

network/discovery/dv5_service.go
iurii-ssv: Well spotted, I don't think we have "alerts" for this, but to keep it even more obvious I would maybe move this `peerDiscoveriesCounter.Add(ctx, 1)` out of this `DiscV5Service.checkPeer` into `DiscV5Service.discover` (right before the `handler(PeerEvent` call).

julienh-ssv: Kept the current semantics intentionally. With the `ssv` validation now inside `checkPeer`, this counter tracks SSV discovery candidates before later connection gating, which matches the metric name better. I did not find any alerting tied to the previous baseline. If we want the old post-filter view too, we can add a separate metric in a follow-up.

network/p2p/p2p_discovery.go
julienh-ssv: Agreed. I kept the separate stats pass for readability and only deduped the retry/cooldown scoring logic so we do not maintain that math twice. Given expected pool sizes and that this stays on the debug path, I think the extra scan is acceptable for now. If profiling ever shows it matters, we can fold stats collection into the main selection loop.

iurii-ssv (dismissed):
👍  Would also be nice to merge this clarification (can ship with this PR): https://github.com/ssvlabs/ssv/pull/2798

julienh-ssv:
> 👍 Would also be nice to merge this clarification (can ship with this PR): #2798

Merged. Thank you for opening PRs instead of commenting. It's more efficient. Really appreciated!

iurii-ssv (approved):

network/p2p/p2p.go
```diff
@@ -441,6 +454,15 @@ func (n *p2pNetwork) choosePeersToTrim(trimCnt int, trimInboundOnly bool) map[pe
 			if connDir == p2pnet.DirOutbound && trimInboundOnly {
 				continue
 			}
+			peerSubnets, _ := n.idx.GetPeerSubnets(p)
+			sharedSubnets := ownSubnets.SharedSubnets(peerSubnets)
+			n.logger.Debug("selected peer for trimming",
```
momosh-ssv, line 459: What do you think about moving this log outside the inner `pConns` loop? In the (rare) multi-connection case warned about at line 440, the same peer would be logged as "selected for trimming" once per connection, which could be confusing when reading logs. Emitting it once per peer (after the inner loop) would keep the output cleaner.

momosh-ssv (approved):

network/p2p/p2p.go
iurii-ssv: Technically, the `for _, pConn := range pConns` should always complete exactly 1 iteration total, as per the code above it:
```
		// we shouldn't have more than 1 connection per peer, but if we do we'd want a
		// warning about it logged, and we'd want to handle it to the best of our ability
		if len(pConns) > 1 {
			n.logger.Error(
				"choosePeersToTrim: encountered peer we have multiple open connections with (expected 1 at most)",
				zap.String("peer_id", p.String()),
				zap.Int("connections_count", len(pConns)),
			)
		}
```
I think the best way to go about addressing it would be to restructure this code a bit to also not do `trimCnt--` for the same peer multiple times in that same `for`-loop.

This is orthogonal to this PR however, I would address it separately later on - https://github.com/ssvlabs/ssv-node-board/issues/994

ljuba-ssv (approved):

PR #2793, opened 2026-04-10
chore(test) - implement exec client reorg unit test

oleg-ssvlabs:
## Summary

- Replace the empty `TestChainReorganizationLogs` in `eth/executionclient/execution_client_test.go` with a real test
- New test targets the actual reorg-handling branch in `fetchLogsInBatches` (`execution_client.go:228-235`) that filters `log.Removed == true`
- Uses a fake JSON-RPC server returning a canned `eth_getLogs` response with a mix of normal and `Removed=true` logs, avoiding the timing fragility of driving a full simulated beacon fork

## Why this approach

The old test was commented out and could not be ported cleanly — the `simulator.Backend` API, `bind.DeployContract` signature, and `BlockLogs` type have all changed since it was written. Reviving it would have tested `followDistance` arithmetic rather than the reorg-handling code, while inheriting engine-API timing fragility that would likely have made it flaky in CI.

The new approach unit-tests the exact branch the client uses to handle reorged logs. It is deterministic, has no timing dependency, and runs in under a second.

Closes https://github.com/ssvlabs/ssv-node-board/issues/536

### DIFF
--- a/eth/executionclient/execution_client_test.go
+++ b/eth/executionclient/execution_client_test.go
@@ -640,105 +640,99 @@ func TestFetchLogsInBatches(t *testing.T) {
 	})
 }
 
-// TestChainReorganizationLogs check that the client receives removed logs correctly.
-// Steps:
-//  1. Deploy the Callable contract.
-//  2. Set up an event subscription.
-//  3. Save the current block which will serve as parent for the fork.
-//  4. Send a transaction.
-//  5. Check that the event was included.
-//  6. Fork by using the parent block as ancestor.
-//  7. Mine two blocks to trigger a reorg.
-//  8. Check that the event was removed.
-
+// TestChainReorganizationLogs verifies that logs marked as Removed=true (which
+// indicates a chain reorganization on the execution layer) are filtered out by
+// fetchLogsInBatches and never delivered to consumers.
+//
+// The test drives the client against a fake JSON-RPC server that returns a
+// canned eth_getLogs response containing a mix of normal and Removed logs.
+// This isolates the reorg-handling branch in fetchLogsInBatches without the
+// timing fragility of driving a full simulated beacon fork.
 func TestChainReorganizationLogs(t *testing.T) {
-	// TODO: fix reorg test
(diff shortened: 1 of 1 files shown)


nkryuchkov (dismissed):

eth/executionclient/execution_client_test.go
```diff
@@ -640,105 +640,99 @@ func TestFetchLogsInBatches(t *testing.T) {
...
+	}
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req struct {
+			ID     json.RawMessage `json:"id"`
+			Method string          `json:"method"`
+		}
+		body, err := io.ReadAll(r.Body)
+		require.NoError(t, err)
+		require.NoError(t, json.Unmarshal(body, &req))
```
iurii-ssv: This is kinda hard to remember :) but I think we shouldn't be using assertions that result into `t.Fatal` calls (or similar) from go-routines **different** from the one that's running the test:
```go
// [T.FailNow], [T.Fatal], [T.Fatalf], [T.SkipNow], [T.Skip], or [T.Skipf]. Those methods, as well as
// the [T.Parallel] method, must be called only from the goroutine running the
// Test function.
//
// The other reporting methods, such as the variations of [T.Log] and [T.Error],
type T struct {
```
Here server could probably just respond with an error, but we also could send the stuff we want to validate on a channel that main-go-routine will read, and assert on.

iurii-ssv (changes requested):
Just one gotcha to fix

eth/executionclient/execution_client_test.go
oleg-ssvlabs: Yes, I actually remember that :)

fixed 8cc53e8dbaba26448395ea4d68debc98c7c21c35

iurii-ssv (approved):

nkryuchkov (approved):

ljuba-ssv (approved):

momosh-ssv (approved):

PR #2795, opened 2026-04-12
Swap gossipsub score index atomically

julienh-ssv:
## Context
Ticket: F-ssv-129

## Problem
`SetScores` rebuilt the internal map by first clearing it, then copying. Ticket claims that Between those steps concurrent readers could temporarily observe an empty map and treat bad peers as unscored.

## Changes
- Build a new scores map from the input, then swap it under the lock.
- Keep semantics the same (latest inspection cycle fully replaces the index) without an empty window.

## Outcome
- Reduce lock contention by copying without lock

## Tests
- `go test ./network/peers -run "Test(GetGossipScore|SetScores|HasBadGossipScore)"`

### DIFF
--- a/network/peers/gossip_score_index_test.go
+++ b/network/peers/gossip_score_index_test.go
@@ -56,17 +56,32 @@ func TestSetScores(t *testing.T) {
 	require.Equal(t, 0.0, score3)
 }
 
-func TestClear(t *testing.T) {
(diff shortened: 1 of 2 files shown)


julienh-ssv:
Closing as false positive since the benefit seems negligible.

PR #2796, opened 2026-04-12
optimize validator queue message keys (F-ssv-157)

julienh-ssv:
Ticket: F-ssv-157

This removes `fmt.Sprintf` / `fmt.Sprint`-based message key construction from the validator message queue hot path.

Changes
- Build message keys using `strings.Builder` + `strconv.Append*`.
- Hex-encode `spectypes.MessageID` directly to avoid `MessageID.String()` allocations.
- Replace operator ID formatting with a stable, allocation-light join.

Why
Keys are generated for every enqueue/dequeue/lookup; avoiding reflection/interface boxing and reducing allocations lowers latency under load.

### DIFF
--- a/protocol/v2/ssv/validator/msg_queue.go
+++ b/protocol/v2/ssv/validator/msg_queue.go
@@ -2,7 +2,9 @@ package validator
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
+	"strconv"
 	"strings"
 
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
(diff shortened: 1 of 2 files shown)


protocol/v2/ssv/validator/msg_queue.go
```diff
@@ -33,6 +35,37 @@ type messageProcessingState struct {
...
+	var buf [20]byte
+	out := strconv.AppendUint(buf[:0], v, 10)
+	_, _ = b.Write(out)
+}
+
+func writeInt64(b *strings.Builder, v int64) {
+	var buf [20]byte
+	out := strconv.AppendInt(buf[:0], v, 10)
+	_, _ = b.Write(out)
+}
```
nkryuchkov, line 50: Can you add a comment that 20 bytes is enough to hold the string representation of any 64-bit integer in base 10? It might look unclear why the size is exactly 20. Perhaps we could extract a const and use it in both functions

nkryuchkov (dismissed):

protocol/v2/ssv/validator/msg_queue.go
julienh-ssv: Sure why not, fixed in c096b0788

nkryuchkov (approved):

iurii-ssv (approved):
LGTM

protocol/v2/ssv/validator/msg_queue_test.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		{
+			name: "qbft/consensus",
+			msg: &queue.SSVMessage{
+				SSVMessage: &spectypes.SSVMessage{
+					MsgType: spectypes.SSVConsensusMsgType,
+					MsgID:   msgID,
+				},
+				SignedSSVMessage: &spectypes.SignedSSVMessage{
+					OperatorIDs: []spectypes.OperatorID{1, 22, 333},
```
momosh-ssv, line 82: Might be worth adding a qbft case with empty `OperatorIDs` — would pin the `[]` output from `writeOperatorIDs` and catch any future edge-case regression in the key format.

momosh-ssv (approved):

PR #2800, opened 2026-04-14
chore(cleanup) - remove unused constants

oleg-ssvlabs:
## Summary

Remove 8 unused constants flagged in the dead-code review, plus the orphan type/struct left over when the sync protocol constants are removed.

### Constants removed

- `ibft/storage/store.go` — `highestInstanceKey`, `instanceKey` (pre-Alan instance-based storage; legacy data cleanup uses `DropPrefix(oldPrefix)` so these suffix keys are unreferenced)
- `message/validation/const.go` — `allowedRoundsInPast` (only usage site was a `/* */` block comment at `consensus_validation.go:538`; comment reference also cleaned up; the TODO explaining why the check is disabled is kept)
- `network/topics/msg_id.go` — `MsgIDBadEncodedMessage`
- `observability/log/names.go` — `NameScoreInspector`, `NameDutyFetcher`
- `protocol/v2/p2p/network.go` — `LastDecidedProtocol`, `DecidedHistoryProtocol`

### Related cleanup (same file)

Removing `LastDecidedProtocol` / `DecidedHistoryProtocol` leaves their parent `SyncProtocol` type with no values, and the sibling `SyncResult` struct and `SyncResults` type have zero callers. All three are deleted along with the constants so nothing is left dangling.

Closes https://github.com/ssvlabs/ssv-node-board/issues/553


momosh-ssv (dismissed):

message/validation/consensus_validation.go
```diff
@@ -535,7 +535,7 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
 	}
 
 	// TODO: lowestAllowed is not supported yet because first round is non-deterministic now
-	lowestAllowed := /*estimatedRound - allowedRoundsInPast*/ specqbft.FirstRound
+	lowestAllowed := specqbft.FirstRound
```
iurii-ssv: Lets remove this TODO then as well while we are at it - since it becomes "incomplete". Or move the `estimatedRound - allowedRoundsInPast` part under that same TODO comment instead.

iurii-ssv (commented):
Just 1 suggestion

message/validation/consensus_validation.go
oleg-ssvlabs: good point.
fixed fe1c4b9bea45cc95c9c3167ce57daa02c3677180

iurii-ssv (approved):

message/validation/consensus_validation.go
```diff
@@ -534,8 +534,7 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
 		estimatedRound = currentEstimatedRound
 	}
 
-	// TODO: lowestAllowed is not supported yet because first round is non-deterministic now
-	lowestAllowed := /*estimatedRound - allowedRoundsInPast*/ specqbft.FirstRound
+	lowestAllowed := specqbft.FirstRound
```
nkryuchkov, line 537: I'm not sure about this one. We need to implement this check, but we had some issues with it in the past.

I think we need to revert this and enable the lowest allowed check in a separate PR

nkryuchkov (changes requested):

message/validation/consensus_validation.go
oleg-ssvlabs: Do you mean that it is better to keep the comment out `/*estimatedRound - allowedRoundsInPast*/` code?

nkryuchkov: Yes, with the TODO. It reminds me that we need to do it every time I see it 😄

We need a separate PR that fixes the check, and we will need to test it properly

nkryuchkov: @oleg-ssvlabs https://github.com/ssvlabs/ssv/pull/2807

oleg-ssvlabs: I see @nkryuchkov .
Since you’ve already implemented the change, maybe there’s no longer a need to keep this TODO? :)

nkryuchkov: @oleg-ssvlabs okay, let's not keep it

nkryuchkov (approved):

ljuba-ssv (approved):

PR #2802, opened 2026-04-14
chore(refactor) - blind pkg

oleg-ssvlabs:
## Summary

Extract two small helpers from `EnsureBlinded` (in `protocol/v2/blockchain/beacon/blind/blind.go`) to reduce copy-paste duplication across the Capella/Deneb/Electra/Fulu branches.

## Motivation

`EnsureBlinded` converts a full beacon block to its blinded form per fork. Each of the four fork branches was independently reimplementing:

1. **SSZ hash-tree-root computation for the transactions and withdrawals lists.** The tx type (`[]bellatrix.Transaction`) and withdrawal type (`[]*capella.Withdrawal`) are identical across all four forks, but each branch repeated the same `(&utilbellatrix.ExecutionPayloadTransactions{...}).HashTreeRoot()` / `(&utilcapella.ExecutionPayloadWithdrawals{...}).HashTreeRoot()` boilerplate with its own error-wrapping.

2. **Construction of a `deneb.ExecutionPayloadHeader`.** Deneb, Electra, and Fulu all use the exact same `deneb.ExecutionPayload` → `deneb.ExecutionPayloadHeader` conversion, with the exact same 16 field assignments. A field added to `deneb.ExecutionPayloadHeader` previously required touching three call sites.

Ticket CQ-5 originally recommended extracting "common blinding logic into a generic helper parameterized by fork version." I'm intentionally not going that far — the four blinded block types (`apiv1capella.BlindedBeaconBlock`, `apiv1deneb.BlindedBeaconBlock`, `apiv1electra.BlindedBeaconBlock` used for both Electra and Fulu) have different body field sets and don't share a Go interface. A unified helper would require reflection or heavily-constrained generics for no readability win, and any future fork that changes the body layout would need to fork the abstraction anyway. The per-fork `switch` branches stay as-is; only the mechanically duplicated parts are factored out.

## Changes

- `protocol/v2/blockchain/beacon/blind/blind.go`:
  - Introduce `computePayloadRoots(txs, withdrawals) (txRoot, wRoot, error)` — used by all four fork branches.
  - Introduce `buildDenebPayloadHeader(*deneb.ExecutionPayload) (*deneb.ExecutionPayloadHeader, error)` — used by Deneb, Electra, and Fulu.
  - Capella branch still constructs its own `capella.ExecutionPayloadHeader` (different type, no blob fields) but now uses `computePayloadRoots` for the roots.

Net: 291 → 256 lines. Deneb/Electra/Fulu payload-header construction goes from ~20 lines of field copies each to a single helper call.

Closes https://github.com/ssvlabs/ssv-node-board/issues/560

### DIFF
--- a/protocol/v2/blockchain/beacon/blind/blind.go
+++ b/protocol/v2/blockchain/beacon/blind/blind.go
@@ -8,12 +8,10 @@ import (
 	apiv1deneb "github.com/attestantio/go-eth2-client/api/v1/deneb"
 	apiv1electra "github.com/attestantio/go-eth2-client/api/v1/electra"
 	"github.com/attestantio/go-eth2-client/spec"
-
-	// no direct use of bellatrix package here
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/capella"
 	"github.com/attestantio/go-eth2-client/spec/deneb"
-
-	// no direct use of electra package here; kept for type refs via api/v1
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	utilbellatrix "github.com/attestantio/go-eth2-client/util/bellatrix"
 	utilcapella "github.com/attestantio/go-eth2-client/util/capella"
 	ssz "github.com/ferranbt/fastssz"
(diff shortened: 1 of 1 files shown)


protocol/v2/blockchain/beacon/blind/blind.go
```diff
@@ -289,3 +210,47 @@ func EnsureBlinded(p *api.VersionedProposal) (*api.VersionedProposal, ssz.Marsha
...
+	if err != nil {
+		return phase0.Root{}, phase0.Root{}, fmt.Errorf("compute withdrawals root: %w", err)
+	}
+	return txRoot, wRoot, nil
+}
+
+// buildDenebPayloadHeader converts a Deneb-shaped ExecutionPayload into its header form.
+// Deneb, Electra, and Fulu all use deneb.ExecutionPayload / deneb.ExecutionPayloadHeader,
+// so one builder serves all three forks.
+func buildDenebPayloadHeader(payload *deneb.ExecutionPayload) (*deneb.ExecutionPayloadHeader, error) {
```
momosh-ssv, line 232: Might be worth adding a nil guard on `payload` here — all three callers already check before calling, so it's safe today, but a one-liner `if payload == nil { return nil, fmt.Errorf("nil execution payload") }` would be cheap insurance against future misuse.

momosh-ssv (approved):

nkryuchkov (approved):

protocol/v2/blockchain/beacon/blind/blind.go
oleg-ssvlabs: this function is unexported(meaning it is tightly connected to local callers) and all local callers already do `nil` checks [here](https://github.com/ssvlabs/ssv/blob/c8b9bce43866805727bb1242e7bf387f048f475d/protocol/v2/blockchain/beacon/blind/blind.go#L108), [here](https://github.com/ssvlabs/ssv/blob/c8b9bce43866805727bb1242e7bf387f048f475d/protocol/v2/blockchain/beacon/blind/blind.go#L142) and [here](https://github.com/ssvlabs/ssv/blob/c8b9bce43866805727bb1242e7bf387f048f475d/protocol/v2/blockchain/beacon/blind/blind.go#L177), so I think it's OK not the `nil` guard here

PR #2804, opened 2026-04-14
refactor(runner): unification of runner ctor arguments

oleg-ssvlabs:
## Summary

Refactors all 6 runner constructors in `protocol/v2/ssv/runner/` to take a single options struct instead of positional arguments. Before this change, `NewProposerRunner` took 13 positional params and the others took 6–10.

Addresses ticket **CQ-7 / F-ssv-039** (`NewProposerRunner` has too many parameters). Expanded the scope to cover every runner since they share most dependencies.

Closes https://github.com/ssvlabs/ssv-node-board/issues/562

## Design

- **`BaseRunnerOptions`** (new, in `runner.go`) holds the 6 fields every runner needs: `NetworkConfig`, `Share`, `Beacon`, `Network`, `Signer`, `OperatorSigner`.
- **Per-role options structs** embed `BaseRunnerOptions` and add role-specific fields:
  - `ProposerRunnerOptions` — `QBFTController`, `DoppelgangerHandler`, `ValCheck`, `HighestDecidedSlot`, `Graffiti`, `ProposerDelay`
  - `AggregatorRunnerOptions` — `QBFTController`, `ValCheck`, `HighestDecidedSlot`
  - `SyncCommitteeAggregatorRunnerOptions` — `QBFTController`, `ValCheck`, `HighestDecidedSlot`
  - `CommitteeRunnerOptions` — `AttestingValidators`, `QBFTController`, `DutyGuard`, `DoppelgangerHandler`
  - `ValidatorRegistrationRunnerOptions` — `ValidatorRegistrationSubmitter`, `FeeRecipientProvider`, `GasLimit`
  - `VoluntaryExitRunnerOptions` — empty wrapper, kept for signature consistency
- Production caller `SetupRunners` constructs `baseOpts` once and reuses it across all cases.

## Drive-by cleanups

- Dropped the dead `logger *zap.Logger` parameter on `NewProposerRunner` — it was declared but never referenced.
- Dropped the now-unused `logger` parameter on `SetupRunners` (only caller updated in same change).
- Aligned `NewAggregatorRunner` with the other constructors: it now returns `(Runner, error)` instead of `(*AggregatorRunner, error)`. The two test-only sites that mutate the `IsAggregator` hook type-assert to `*AggregatorRunner`. (Separate commit.)

### DIFF
--- a/protocol/v2/ssv/runner/proposer.go
+++ b/protocol/v2/ssv/runner/proposer.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/ssvlabs/ssv/ssvsigner/ekm"
 
-	"github.com/ssvlabs/ssv/networkconfig"
 	"github.com/ssvlabs/ssv/observability"
 	"github.com/ssvlabs/ssv/observability/log/fields"
 	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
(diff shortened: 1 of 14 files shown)


nkryuchkov (approved):

protocol/v2/ssv/testing/runner.go
```diff
@@ -123,105 +123,78 @@ var ConstructBaseRunner = func(
...
+			BaseRunnerOptions:  baseOpts,
+			QBFTController:     contr,
+			ValCheck:           valCheck,
+			HighestDecidedSlot: TestingHighestDecidedSlot,
+		})
 		if err != nil {
 			return nil, err
 		}
-		rnr.IsAggregator = func(_ uint64, _ uint64, _ []byte) bool {
+		rnr.(*runner.AggregatorRunner).IsAggregator = func(_ uint64, _ uint64, _ []byte) bool {
```
momosh-ssv, line 157: Worth considering — this downcast is only safe because `NewAggregatorRunner` always returns a `*AggregatorRunner` wrapped in `Runner`. If we ever wrap the returned `Runner` (e.g. a decorator), the assertion panics. Maybe worth exposing `IsAggregator` via a method on the concrete type (e.g. `SetIsAggregator`) so tests don't rely on the downcast — or at least the comma-ok form here and at line 326.

momosh-ssv (approved):

iurii-ssv (approved):
LGTM

protocol/v2/ssv/runner/proposer.go
```diff
@@ -61,44 +60,45 @@ type ProposerRunner struct {
...
+		doppelgangerHandler: opts.DoppelgangerHandler,
+		ValCheck:            opts.ValCheck,
 		measurements:        newMeasurementsStore(),
-		graffiti:            graffiti,
+		graffiti:            opts.Graffiti,
 
-		proposerDelay: proposerDelay,
+		proposerDelay: opts.ProposerDelay,
 	}, nil
 }
```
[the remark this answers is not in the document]
oleg-ssvlabs: It’s a valid comment, although I would consider it more of a “nice to have,” since there were previously no nil checks and it should be clear that all parameters must be set.

protocol/v2/ssv/testing/runner.go
oleg-ssvlabs: Considering this is test-only code, this cast is OK to me. We have similar casts (e.g., *runner.CommitteeRunner) in other places in this file. If you think this is an issue, we can look into it in a separate PR.

PR #2807, opened 2026-04-15
message/validation: enable lowest allowed round check

nkryuchkov:
## Summary

Enable the lower bound of the estimated round window in message validation.

This restores rejecting non-decided messages that are too far behind the locally estimated round, while keeping the lower bound clamped to `specqbft.FirstRound` to avoid unsigned underflow.

The round spread check now runs after slot-time and slot-advance validation so existing `late slot` and `slot already advanced` errors keep the same precedence.

See also https://github.com/ssvlabs/ssv/pull/2811

Resolves https://github.com/ssvlabs/ssv-node-board/issues/1002


message/validation/consensus_validation.go
```diff
@@ -534,8 +544,9 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
 		estimatedRound = currentEstimatedRound
 	}
 
-	// TODO: lowestAllowed is not supported yet because first round is non-deterministic now
-	lowestAllowed := /*estimatedRound - allowedRoundsInPast*/ specqbft.FirstRound
+	lowestAllowed := lowestAllowedRound(estimatedRound)
```
iurii-ssv: We might want to re-think the way we handle this `lowestAllowed` calculation, the `currentEstimatedRound` doesn't take into account duty type, meaning it assumes:
- every round lasts `roundtimer.QuickTimeout` (2s)
- every round-change starts at `roundtimer.QuickTimeout` (2s) into slot

Which I think doesn't correspond to any "real" duty we have - the "real" current-round values are calculated by `RoundTimer.RoundTimeout` in a somewhat different manner ... so we **must align** validation vs those "real" calculations.

Also, for clarify, I would rename `estimatedRound` -> `estimatedRoundMsgReceivedAt`.

iurii-ssv (changes requested):
I think there is a discrepancy in the code-area around these changes we need to look into

message/validation/consensus_validation.go
iurii-ssv: For **stage** branch:
- `lowestAllowed` is always `specqbft.FirstRound` - so, unaffected by `currentEstimatedRound` miscalculations
- `highestAllowed` seems to be affected though ... but `currentEstimatedRound` typically returns "higher value than what the correct implementation would return" - so it just results in a more relaxed check than would be ideal, without breaking anything

nkryuchkov:
@greptileai review it again

nkryuchkov:
@greptileai review it again

message/validation/consensus_validation.go
```diff
@@ -410,22 +411,16 @@ func (mv *messageValidator) maxRound(role spectypes.RunnerRole) (specqbft.Round,
...
+}
 
-	sinceFirstSlowRound := sinceSlotStart - (time.Duration(roundtimer.QuickTimeoutThreshold) * roundtimer.QuickTimeout)
-	delta, err = casts.DurationToUint64(sinceFirstSlowRound / roundtimer.SlowTimeout)
-	if err != nil {
-		return 0, fmt.Errorf("failed to convert time duration to uint64: %w", err)
+func lowestAllowedRound(estimatedRound specqbft.Round) specqbft.Round {
+	if estimatedRound < specqbft.FirstRound+allowedRoundsInPast {
+		return specqbft.FirstRound
 	}
```
iurii-ssv, line 421: Should be `<=` here, otherwise this func might `return 0` (and 0 is basically a `NoRound`),

but even so, would be simpler to have a check like `if estimatedRound <= allowedRoundsInPast` instead

message/validation/consensus_validation.go
```diff
@@ -410,22 +411,16 @@ func (mv *messageValidator) maxRound(role spectypes.RunnerRole) (specqbft.Round,
...
-	}
+func (mv *messageValidator) currentEstimatedRound(role spectypes.RunnerRole, sinceSlotStart time.Duration) (specqbft.Round, error) {
+	return roundtimer.EstimatedRoundAt(role, mv.netCfg.SlotDuration, sinceSlotStart)
+}
 
-	sinceFirstSlowRound := sinceSlotStart - (time.Duration(roundtimer.QuickTimeoutThreshold) * roundtimer.QuickTimeout)
-	delta, err = casts.DurationToUint64(sinceFirstSlowRound / roundtimer.SlowTimeout)
-	if err != nil {
-		return 0, fmt.Errorf("failed to convert time duration to uint64: %w", err)
+func lowestAllowedRound(estimatedRound specqbft.Round) specqbft.Round {
```
iurii-ssv, line 418: I would rename `estimatedRound` -> `estimatedRoundMsgReceivedAt` for completeness

Actually, inlining this func results into simpler code even.

message/validation/consensus_validation.go
```diff
@@ -293,6 +285,15 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
 		return err
 	}
 
+	if len(signedSSVMessage.OperatorIDs) == 1 {
+		// Rule: Round must be between estimated round - allowedRoundsInPast and
+		// estimated round + allowedRoundsInFuture. Only for non-decided messages.
+		// Keep this after slot-time and slot-advance validation so late/old-slot messages preserve their original errors.
```
iurii-ssv, line 291: I would rename `estimatedRound` -> `estimatedRoundMsgReceivedAt` for completeness

message/validation/consensus_validation.go
```diff
@@ -523,22 +518,28 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
 	slotStartTime := mv.netCfg.SlotStartTime(phase0.Slot(consensusMessage.Height)) /*.
 	Add(mv.waitAfterSlotStart(role))*/ // TODO: not supported yet because first round is non-deterministic now
 
+	role := signedSSVMessage.SSVMessage.GetID().GetRoleType()
+
 	sinceSlotStart := time.Duration(0)
```
iurii-ssv, line 523: Nit: `sinceSlotStart`->`timeIntoSlot`

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -38,6 +39,46 @@ type TimeoutOptions struct {
...
+
+// EstimatedRoundAt returns the round that should be current for the given role
+// at the provided elapsed time since slot start. Keep this aligned with RoundTimeout.
+func EstimatedRoundAt(role spectypes.RunnerRole, slotDuration, sinceSlotStart time.Duration) (specqbft.Round, error) {
+	sinceRoundStart := sinceSlotStart - baseRoundStartDelay(role, slotDuration)
+	if sinceRoundStart <= 0 {
+		return specqbft.FirstRound, nil
+	}
+
+	return estimatedRoundAtSince(sinceRoundStart)
```
iurii-ssv, line 79: `estimatedRoundAtSince` -> `estimatedRoundAt` (to match `EstimatedRoundAt`)

message/validation/consensus_validation.go
```diff
@@ -523,22 +518,28 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
...
-	// TODO: lowestAllowed is not supported yet because first round is non-deterministic now
-	lowestAllowed := /*estimatedRound - allowedRoundsInPast*/ specqbft.FirstRound
-	highestAllowed := estimatedRound + allowedRoundsInFuture
-
-	role := signedSSVMessage.SSVMessage.GetID().GetRoleType()
+	lowestAllowed := specqbft.FirstRound
+	if role != spectypes.RoleProposer {
+		// Proposer round timeouts are still relative rather than slot-start-based,
+		// so we keep the lower bound relaxed until those calculations are aligned.
+		lowestAllowed = lowestAllowedRound(estimatedRoundMsgReceivedAt)
```
iurii-ssv, line 538: Let's also mention "where those calculations need to align":
```go
		// Proposer round timeouts are still relative to QBFT instance start times rather than absolute time-into-slot
		// values, so we keep the lower bound relaxed until that changes in rountimer package.
```

iurii-ssv (commented):
Should work better now, but the code is still a bit overly-complex - I would suggest a bunch of changes before merging https://github.com/ssvlabs/ssv/pull/2811 (addresses all my discussions from below, hence they are resolved).

message/validation/consensus_validation.go
nkryuchkov: @greptileai requested to change it to `<` in https://github.com/ssvlabs/ssv/pull/2807#discussion_r3085788868

iurii-ssv:
@greptileai review this again

iurii-ssv (dismissed):

message/validation/consensus_validation_test.go
```diff
@@ -5,100 +5,230 @@ import (
...
+			role:         spectypes.RoleCommittee,
+			timeIntoSlot: netCfg.SlotDuration/3 + roundtimer.QuickTimeout*5,
+			round:        specqbft.FirstRound + 3,
+		},
+		{
+			name:         "proposer still allows first round when slot-start estimate advanced",
+			role:         spectypes.RoleProposer,
+			timeIntoSlot: roundtimer.QuickTimeout * 5,
+			round:        specqbft.FirstRound,
+		},
```
olegshmuelov, line 211: The proposer cap is the main behavioral change in this PR and no test covers its reject path.

  ```go
  {
      name:         "proposer rejects round 3 regardless of estimated round",
      role:         spectypes.RoleProposer,
      timeIntoSlot: roundtimer.QuickTimeout * 5,
      round:        specqbft.FirstRound + 2,
      wantErr:      ErrEstimatedRoundNotInAllowedSpread,
  },
  {
      name:         "proposer allows round 2 at slot start",
      role:         spectypes.RoleProposer,
      timeIntoSlot: 0,
      round:        specqbft.FirstRound + 1,
  },
  ```

message/validation/consensus_validation.go
```diff
@@ -530,29 +518,37 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
...
+	// so adding allowedRoundsInFuture cannot get close to uint64 overflow.
+	highestAllowedRound := estimatedRoundMsgReceivedAt + allowedRoundsInFuture
+	// Proposer round timeouts are relative to QBFT instance start times rather than absolute time-into-slot
+	// values (until https://github.com/ssvlabs/ssv/issues/2429 is implemented), since we don't have any visibility
+	// into the actual QBFT instance state here - the best we can do for proposer role is to always permit messages
+	// with round 1 and round 2 (and none else) since these are the only rounds proposer-runner can meaningfully
+	// work with (round 3+ is too late for this duty type).
+	if role == spectypes.RoleProposer {
+		lowestAllowedRound = specqbft.FirstRound
+		highestAllowedRound = specqbft.FirstRound + 1
```
momosh-ssv: What do you think about the proposer upper bound being hard-capped at round 2? I fear this may be stricter than the rest of the codebase expects — `maxRound(RoleProposer)` returns 6, and `RoundTimeout` is designed to reach round 3+ on proposer failure. 

With this cap, if the primary proposer drops and the cluster legitimately needs to round-change to produce a block before the slot ends, single-signer round-change/prepare/commit messages at round 3+ will fail validation and never propagate. 
Worth aligning the two caps — either loosen this to match `maxRound(Proposer)`, or tighten `maxRound(Proposer)` to 2 so there's a single source of truth.

message/validation/consensus_validation.go
```diff
@@ -530,29 +518,37 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
...
+		lowestAllowedRound = estimatedRoundMsgReceivedAt - allowedRoundsInPast
+	}
+	// No overflow bug here: estimatedRoundMsgReceivedAt comes from elapsed slot time,
+	// so adding allowedRoundsInFuture cannot get close to uint64 overflow.
+	highestAllowedRound := estimatedRoundMsgReceivedAt + allowedRoundsInFuture
+	// Proposer round timeouts are relative to QBFT instance start times rather than absolute time-into-slot
+	// values (until https://github.com/ssvlabs/ssv/issues/2429 is implemented), since we don't have any visibility
+	// into the actual QBFT instance state here - the best we can do for proposer role is to always permit messages
+	// with round 1 and round 2 (and none else) since these are the only rounds proposer-runner can meaningfully
+	// work with (round 3+ is too late for this duty type).
```
momosh-ssv: Seems that this comment ("round 3+ is too late for this duty type") disagrees with `maxRound(RoleProposer) = 6` in the same file. 
Two caps with different values — maybe worth reconciling so future readers don't have to guess which one is authoritative.

message/validation/consensus_validation.go
```diff
@@ -303,6 +295,16 @@ func (mv *messageValidator) validateQBFTMessageByDutyLogic(
 		return err
 	}
 
+	if len(signedSSVMessage.OperatorIDs) == 1 {
+		// Rule: Round must be between `estimated round msg received at` - allowedRoundsInPast and
+		// `estimated round msg received at` + allowedRoundsInFuture. Only for non-decided messages.
+		// Keep this after slot-time and slot-advance validation so late/old-slot messages preserve
+		// their original errors.
```
momosh-ssv, line 303: Might be worth expanding this comment — the check now also runs after `validateBeaconDuty` and `validateDutyCount`, not just slot-time/slot-advance. 
All of those are `reject: false` so no semantic regression, but readers trying to reason about error precedence will land here first.

network/topics/msg_validator_test.go
```diff
@@ -30,6 +31,16 @@ import (
 
 func TestMsgValidator(t *testing.T) {
 	logger := zaptest.NewLogger(t)
+	beaconCfg := *networkconfig.TestNetwork.Beacon
+	// Pin slot 0 to start right now so that by the time `mv.Validate` runs the message is
+	// still comfortably inside round 1 for the committee role. Committee round 1 is accepted
+	// as long as timeIntoSlot stays below ~(slotDuration/3 + 3*QuickTimeout) ≈ 10s — any
+	// larger offset here would eat into that budget. A zero offset maximizes it.
+	beaconCfg.GenesisTime = time.Now()
```
momosh-ssv, line 39: Could this lead to flakes on slow CI? The budget you cite is ~8–10s and three subtests run serially here. A race-detector or sanitizer run on a noisy shared runner could straddle a round boundary. 
Might be worth injecting a frozen clock (e.g. stub `SlotStartTime`) rather than relying on wall-clock proximity.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -30,6 +30,82 @@ type TimeoutOptions struct {
...
+func EstimatedRoundAt(role spectypes.RunnerRole, slotDuration, timeIntoSlot time.Duration) (specqbft.Round, error) {
+	// Compute the round directly by inverting the piecewise-linear roundTimeoutOffset formula:
+	//   Quick phase (r <= T): offset(r) = headStart + r * quick
+	//   Slow phase  (r >  T): offset(r) = headStart + T * quick + (r - T) * slow
+	elapsed := timeIntoSlot - round1HeadStart(role, slotDuration)
+	if elapsed < 0 {
+		return specqbft.FirstRound, nil
+	}
+
+	quickEnd := casts.DurationFromUint64(uint64(QuickTimeoutThreshold)) * QuickTimeout
```
momosh-ssv, line 84: Any reason we're reading `QuickTimeout`/`QuickTimeoutThreshold`/`SlowTimeout` directly here, while `roundTimeoutForRound` reads from the `TimeoutOptions` receiver? A test that overrides `RoundTimer.timeoutOptions` won't be reflected here, so the two can silently diverge without the cross-validation tests noticing. Maybe route this through `defaultTimeoutOptions` so the dependency is explicit.

protocol/v2/qbft/roundtimer/timer_test.go
```diff
@@ -87,6 +87,287 @@ func TestTimeoutForRound(t *testing.T) {
...
+//   - offset exactly → just advanced to round r+1
+//   - offset + 1ns → still in round r+1
+//
+// This is the test that would have caught an off-by-one `<` vs `<=` in EstimatedRoundAt's loop,
+// or a wrong starting `r` — none of which the pre-existing tests directly exercised.
+func TestEstimatedRoundAtBoundaries(t *testing.T) {
+	// Use a realistic slot duration (12s) rather than the timer_test package's 600ms so the
+	// numbers line up with the real QuickTimeout (2s) and SlowTimeout (2m) that
+	// EstimatedRoundAt uses via defaultTimeoutOptions.
+	const realSlotDuration = 12 * time.Second
```
momosh-ssv: Might be worth deriving this from `networkconfig.TestNetwork.Beacon.SlotDuration` (or `setupTestBeaconConfig()`) instead of hard-coding 12s. 
If mainnet slot duration ever shifts (fork, new testnet), this test keeps passing against a stale assumption.

iurii-ssv:
@momosh-ssv @nkryuchkov most of the comments on this PR are related to https://github.com/ssvlabs/ssv/pull/2811 part of it, I'll address them soon.

message/validation/consensus_validation.go
iurii-ssv: In practice there is no need to do Round 3+, as the comment explains.

Regarding code structure, you are right, we need to tighten `maxRound(Proposer)` to 2 and allow all rounds in `roundBelongsToAllowedSpread` at the same time (as an exception for proposer duty-type, admitting that we cannot estimate it precisely), to make it unambiguous and simple.

**Update:** applied

message/validation/consensus_validation_test.go
iurii-ssv: Will be superseded by changes from https://github.com/ssvlabs/ssv/pull/2807#discussion_r3124738838

message/validation/consensus_validation.go
iurii-ssv: Resolving this one as a duplicate of the comment above - https://github.com/ssvlabs/ssv/pull/2807#discussion_r3116452736

message/validation/consensus_validation.go
iurii-ssv: I think we can just remove this part, the validation order for this particular thing is not important/relevant as long as spec-tests pass:

> 		// Keep this after slot-time and slot-advance validation so late/old-slot messages preserve
>		// their original errors.

network/topics/msg_validator_test.go
iurii-ssv: 10s seems like plenty of time for this test to finish, so I wouldn't worry about it unless we hit an actual flaky failure,

and it's kinda tangential to the functionality this PR implements (I basically just applied an 80% improvement for 20% of cost here, since I've passed this code by).

protocol/v2/qbft/roundtimer/timer_test.go
iurii-ssv: Good idea, should be fixed now

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: Good find, I was originally thinking we could use `TimeoutOptions` to keep the related stuff together, but then realized it's not really worth it (inlining is simpler) ... and forgot to inline it for `RoundTimer.RoundTimeout` ... removed the `TimeoutOptions` completely now.

iurii-ssv (dismissed):

iurii-ssv (dismissed):

iurii-ssv (approved):

momosh-ssv (approved):

PR #2808, opened 2026-04-15
[T-ssv-035] Buffer scheduler reorg events and test reorg flags

julienh-ssv:
## Summary
- buffer the scheduler reorg channel so a single head event can emit both previous-root and current-root reorg notifications without blocking
- add direct `HandleHeadEvent` tests for previous-epoch and current-epoch reorg flag emission
- add a regression test that verifies `HandleHeadEvent` still returns when no reorg consumer is draining the channel

## Why
`HandleHeadEvent` could emit up to two `ReorgEvent` values for a single head event, but the internal channel was unbuffered and had no direct coverage for the exact `Current` flag values. That left two gaps: the wrong reorg flag could silently steer handlers down the wrong refetch path, and the scheduler could block if the reorg channel was not being drained.

## Impact
This keeps scheduler reorg signaling non-blocking for the per-head-event burst that the code already supports, and it locks in the expected `ReorgEvent.Current` semantics with direct tests.

## Root Cause
The scheduler used an unbuffered internal `reorg` channel even though `HandleHeadEvent` can publish both a previous-root and a current-root reorg before any consumer necessarily receives the first event.

## Validation
- `go test ./operator/duties -run 'TestScheduler_HandleHeadEvent|TestScheduler_Regression_IndicesChangeStuck'`
- `go test ./operator/duties`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -v ./operator/duties/...`

## Ticket
Refs: `T-ssv-035`
Fixes: issue - Fixes ssvlabs/ssv-node-board#924

### DIFF
--- a/operator/duties/scheduler.go
+++ b/operator/duties/scheduler.go
@@ -34,6 +34,9 @@ const (
 	// blockPropagationDelay time to propagate around the nodes
 	// before kicking off duties for the block's slot.
 	blockPropagationDelay = 300 * time.Millisecond
+	// reorgChannelBuffer allows HandleHeadEvent to emit a reorg notification without blocking on
+	// downstream fanout startup or temporary consumer lag.
+	reorgChannelBuffer = 1
 )
 
 // DutiesExecutor is an interface for executing duties.
@@ -156,7 +159,7 @@ func NewScheduler(logger *zap.Logger, opts *SchedulerOptions) *Scheduler {
 		handlers: []dutyHandler{},
 
 		ticker:   opts.SlotTickerProvider(),
-		reorg:    make(chan ReorgEvent),
+		reorg:    make(chan ReorgEvent, reorgChannelBuffer),
 		waitCond: sync.NewCond(&sync.Mutex{}),
 	}
 
--- a/operator/duties/scheduler_test.go
+++ b/operator/duties/scheduler_test.go
@@ -7,6 +7,7 @@ import (
 	"testing/synctest"
 	"time"
 
+	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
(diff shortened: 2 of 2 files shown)


nkryuchkov (approved):

operator/duties/scheduler.go
```diff
@@ -34,6 +34,9 @@ const (
 	// blockPropagationDelay time to propagate around the nodes
 	// before kicking off duties for the block's slot.
 	blockPropagationDelay = 300 * time.Millisecond
+	// reorgChannelBuffer allows HandleHeadEvent to emit a reorg notification without blocking on
+	// downstream fanout startup or temporary consumer lag.
+	reorgChannelBuffer = 1
```
iurii-ssv, line 39: I would also link this issue in this comment to mention:
```go
// This is a short-term work-around for https://github.com/ssvlabs/ssv-node-board/issues/992
```

iurii-ssv (approved):
LGTM

operator/duties/scheduler.go
julienh-ssv: Hmm I don't think we need to, it will be in the git blame which the IDE can show line by line. I'm not a huge fan of putting github links in the source code itself. Git should be enough imho.

iurii-ssv: > I'm not a huge fan of putting github links in the source code itself. Git should be enough imho.

I think this is unavoidable given how heavily we leverage Github (it makes sense keep linking github PRs/issues since we already have lots of that in our existing code).

Technically git is sufficient, but we'll be losing/missing out on important info if we are reluctant to use github links.

ljuba-ssv (approved):

Issue #688, opened 2022-08-21
OOM termination for small nodes


amir-blox:
**Describe the bug**

Nodes with limited resources seems to experience OOM termination pretty frequently.

Seems like the issue is originated in the pubsub router, see [open issue](https://github.com/libp2p/go-libp2p-pubsub/issues/350)

**To Reproduce**

**Expected behavior**

**Screenshots**

**Node (please complete the following information):**
 - Operator Version: `v0.2.x`, `v0.3.x`

**Additional context**

PR #2810, opened 2026-04-15
Prevent msg ID handler overflows from dropping peers

julienh-ssv:
## Ticket
- `[F-ssv-132] #688 — MsgIDHandler.Add silently drops events when buffer is full`
- Under load, `MsgIDHandler.Add` could silently drop peer-to-message associations once its async buffer filled, weakening deduplication and peer scoring.

## Changes
- increased the async msg-id buffer from `32` to `1024` to absorb larger bursts
- changed the full-buffer path to fall back to a synchronous insert instead of dropping the mapping
- added a `msg_id_buffer_fallback` metric to make overload visible
- added focused tests for queued processing and the full-buffer fallback path

## Validation
- `go test ./network/topics -run 'TestMsgIDHandler' -count=1`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -v ./network/topics/...`

Fixes - Fixes ssvlabs/ssv-node-board#688

### DIFF
--- a/network/topics/msg_id_test.go
+++ b/network/topics/msg_id_test.go
@@ -1,0 +1,59 @@
+package topics
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgIDHandlerStartProcessesQueuedEvents(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	handler := &msgIDHandler{
+		ctx:    ctx,
+		added:  make(chan addedEvent, 1),
+		ids:    make(map[string]*msgIDEntry),
+		locker: &sync.Mutex{},
+		ttl:    time.Minute,
+	}
+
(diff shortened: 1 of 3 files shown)


network/topics/msg_id_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgIDHandlerStartProcessesQueuedEvents(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
```
nkryuchkov: can we use `t.Context()` instead?

iurii-ssv (dismissed):

network/topics/msg_id_test.go
julienh-ssv: Updated this test to use `t.Context()` and reran `go test ./network/topics -run "TestMsgIDHandler" -count=1` plus `make lint`.

iurii-ssv (approved):

nkryuchkov (approved):

ljuba-ssv (approved):

PR #2811, opened 2026-04-15
msg-validation: round-timeout validation refactoring

iurii-ssv:
Builds on top of https://github.com/ssvlabs/ssv/pull/2807 to clean up the message-validation code related to validating QBFT Rounds:
- fixes [an edge-case](https://github.com/ssvlabs/ssv/pull/2807#discussion_r3087705125) identified in https://github.com/ssvlabs/ssv/pull/2807
- provides additional comments, renames and restructures things for clarity
- increases tests coverage
- now hard-caps proposer messages to rounds 1..2, rounds 3+ aren't relevant in practice since the earliest round 3 can start is "4s into the slot" (while proposer-duty must finish at the latest by "4s into the slot")

Part of https://github.com/ssvlabs/ssv-node-board/issues/1002


iurii-ssv:
@greptile pls re-review

message/validation/consensus_validation.go
```diff
@@ -515,36 +508,37 @@ func (mv *messageValidator) roundBelongsToAllowedSpread(
...
+	highestAllowedRound := estimatedRoundMsgReceivedAt + allowedRoundsInFuture
+	// Proposer round timeouts are relative to QBFT instance start times rather than absolute time-into-slot
+	// values (until https://github.com/ssvlabs/ssv/issues/2429 is implemented), since we don't have any visibility
+	// into the actual QBFT instance state here - the best we can do for proposer role is to always permit messages
+	// with round 1 and round 2 (and none else) since these are the only rounds proposer-runner can meaningfully
+	// work with (round 3+ is too late for this duty type).
+	if role == spectypes.RoleProposer {
+		lowestAllowedRound = specqbft.FirstRound
+		highestAllowedRound = specqbft.FirstRound + 1
+	}
```
nkryuchkov, line 536: Will the code be simpler if we set these as initial values and then overwrite them if role is NOT proposer?

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -39,7 +38,36 @@ type TimeoutOptions struct {
...
+//
+//	Round 1     ends at  headStart + 1 * quick
+//	Round 2     ends at  headStart + 2 * quick
+//	...
+//	Round T     ends at  headStart + T * quick               (T = quickThreshold)
+//	Round T+1   ends at  headStart + T * quick + 1 * slow
+//	Round T+2   ends at  headStart + T * quick + 2 * slow
+//
+// Every role has its own dedicated headStart duration.
+func (o TimeoutOptions) roundTimeoutOffset(role spectypes.RunnerRole, slotDuration time.Duration, round specqbft.Round) time.Duration {
```
nkryuchkov, line 52: I'm not sure that 'offset' describes the duration very well

nkryuchkov (approved):

message/validation/consensus_validation.go
iurii-ssv: I think it's simpler to reason about the way it currently is - since it's the Proposer-case that diverges from the "default behavior".

Note, there is a "default behavior" and there is a "default lowestAllowedRound/highestAllowedRound for the default behavior".

So, using these proposal-specific values for the "default lowestAllowedRound/highestAllowedRound for the default behavior" is incorrect because that would make `highestAllowedRound = specqbft.FirstRound + 1` a default for Committee-duty type, for example (which is simply would be wrong).

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: WDYT about renaming `roundTimeoutOffset`->`roundTimeoutSlotOffset` / `roundTimeoutIntoSlot` ? Or do you have some other name in mind

Or maybe we could name it just `roundTimeout` to keep consistent/similar to `RoundTimer.RoundTimeout`

nkryuchkov: Maybe something like `RoundTimeoutAfter?`

iurii-ssv: `After` part doesn't quite describe it, how about `roundTimeoutForRound` ?

ljuba-ssv (approved):

protocol/v2/qbft/roundtimer/timer.go
iurii-ssv: Renamed to `roundTimeoutForRound`

PR #2812, opened 2026-04-16
queue: improve pressure admission and drop metrics

nkryuchkov:
## Summary

Addresses `F-ssv-175`.

This PR makes dropped messages observable instead of effectively silent.

Changes:
- add `ssv.queue.messages.dropped` counter with `queue_type`, `queue_id`, `drop_reason` attributes; recorded by queues themselves on `buffer_full`
- add `ssv.validator.router.messages.dropped` counter for the validator router; reasons: `buffer_full`, `context_canceled`
- factor queue-side metrics behind a small `queueObserver` interface so `queue.go` stays focused on queueing semantics
- fix a non-deterministic `select` in `messageRouter.Route` so a cancelled context wins over a buffered send (also makes `TestRouter_DropsWhenContextCanceled` deterministic)

## Why

`F-ssv-175` flagged that queues drop messages silently under load, with no visibility for operators. With these counters, drops are now graphable per-queue and per-reason, and the router-vs-queue split is preserved (different namespaces) so dashboards can attribute drops to the right stage.


nkryuchkov:
@greptileai review the latest changes

nkryuchkov:
@greptileai review the latest changes

protocol/v2/ssv/queue/admission.go
```diff
@@ -0,0 +1,55 @@
...
+		if body == nil {
+			return false, DropReasonMalformed
+		}
+		currentSlot := state.Slot
+		if currentSlot == 0 {
+			currentSlot = phase0.Slot(state.Height)
+		}
+		if body.Slot < currentSlot {
+			return false, DropReasonStaleSlot
+		}
```
olegshmuelov, line 56: `admission.go` rejects `PartialSignatureMessages` with `body.Slot < currentSlot`, but the runner accepts `[currentDutySlot-1, currentDutySlot]` — see `runner_validations.go:69` (`minSlot := b.State.CurrentDuty.DutySlot() - 1`). Validator role queues are reused across duties, so under 3/4-pressure during slot rollover this drops a valid late post-consensus partial-sig from the previous slot, costing the validator a final aggregation/submission. Regression from pre-PR behavior.

Suggestion:

```go
// Match runner's [currentDutySlot-1, currentDutySlot] acceptance window.
minAcceptableSlot := currentSlot
if currentSlot > 0 {
    minAcceptableSlot = currentSlot - 1
}
if body.Slot < minAcceptableSlot {
    return false, DropReasonStaleSlot
}
```

Please add a regression test covering this case.

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -58,14 +59,30 @@ func (v *Validator) EnqueueMessage(ctx context.Context, msg *queue.SSVMessage) {
 	v.mtx.RLock() // read v.Queues
 	defer v.mtx.RUnlock()
 	if q, ok := v.Queues[msg.MsgID.GetRoleType()]; ok {
+		queueID := queue.ValidatorMetricID(msg.MsgID.GetRoleType())
+		if accepted, dropReason := queue.ShouldAcceptUnderPressure(v.messageQueueState(msg.MsgID), msg, q.InboxLen(), q.InboxCap()); !accepted {
```
olegshmuelov: `ShouldAcceptUnderPressure` receives `q.InboxLen()`, but the real backlog lives in both the channel and `q.head` (after `readInbox` drains the chan into the linked list). When the consumer is running steadily, `InboxLen` returns near-zero while the total queue may hold hundreds of messages — so pressure admission never triggers precisely when the queue is most behind.

Suggestion: expose total depth and feed it to the admission signal:

```go
type Queue interface {
    Len() int        // total = chan + head
    InboxLen() int   // chan only
    InboxCap() int
}

// caller:
queue.ShouldAcceptUnderPressure(state, msg, q.Q.Len(), q.Q.InboxCap())
```

protocol/v2/ssv/queue/admission.go
```diff
@@ -0,0 +1,55 @@
...
+	if threshold >= capacity && capacity > 1 {
+		threshold = capacity - 1
+	}
+	return buffered >= threshold
+}
+
+// ShouldAcceptUnderPressure keeps current or newer work admissible when inbox pressure is high.
+// It rejects only obviously stale messages, leaving strict-full behavior unchanged.
+func ShouldAcceptUnderPressure(state *State, msg *SSVMessage, buffered, capacity int) (bool, string) {
+	if !IsUnderPressure(buffered, capacity) || state == nil || !state.HasRunningInstance || msg == nil {
```
olegshmuelov: The `!state.HasRunningInstance` clause in the L26 guard short-circuits to "accept" in two real windows: before `StartNewInstance` (duty accepted, consensus not started) and after `IsDecided() == true` (post-consensus partial signatures still being collected). In both, `state.Slot` and `state.Height` are still meaningful — stale-height QBFT and stale-slot partial-sigs are safe to reject even without a running instance.

Suggestion: drop `!state.HasRunningInstance` from the L26 guard, and gate only the round check on it (round is undefined without an instance; height/slot are not):

```go
func ShouldAcceptUnderPressure(state *State, msg *SSVMessage, buffered, capacity int) (bool, string) {
    if !IsUnderPressure(buffered, capacity) || state == nil || msg == nil {
        return true, ""
    }

    switch body := msg.Body.(type) {
    case *specqbft.Message:
        if body == nil {
            return false, DropReasonMalformed
        }
        if body.Height < state.Height {
            return false, DropReasonStaleHeight
        }
        if state.HasRunningInstance && body.Height == state.Height && body.Round < state.Round {
            return false, DropReasonStaleRound
        }
    case *spectypes.PartialSignatureMessages:
        // ... keep existing slot check, applying the currentSlot-1 allowance from the separate comment.
    }
    return true, ""
}
```

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -58,14 +59,30 @@ func (v *Validator) EnqueueMessage(ctx context.Context, msg *queue.SSVMessage) {
...
+			const eventMsg = "❗ dropping stale message because the queue is under pressure"
+			queue.RecordDroppedMessage(queue.ValidatorQueueMetricType, queueID, dropReason)
+			logger.Warn(eventMsg,
+				zap.String("drop_reason", dropReason),
+				zap.String("msg_type", message.MsgTypeToString(msg.MsgType)),
+				zap.String("msg_id", msg.MsgID.String()))
+			span.AddEvent(eventMsg, trace.WithAttributes(attribute.String("drop_reason", dropReason)))
+			span.SetStatus(codes.Error, eventMsg)
+			return
+		}
```
olegshmuelov, line 73: Admission only gates `Validator.EnqueueMessage`. Three control-message paths bypass it via direct `TryPush`: `duty_executor.go:31` (ExecuteDuty), `timer.go:53` (validator timeout), `timer.go:114` (committee timeout). Combined with the `HasRunningInstance` gap (separate comment on `admission.go:26`), a stale-traffic burst during the no-duty window can fill the queue and silently drop the next `ExecuteDuty` or timeout — costing the validator a duty.

Also: `Validator.messageQueueState` (`queue_validator.go:99`) returns `nil` when `DutyRunnerForMsgID` returns nil, which short-circuits admission to accept-all (`admission.go:26` checks `state == nil`). The committee version always builds a slot-only state — worth normalizing here:

```go
func (v *Validator) messageQueueState(msgID spectypes.MessageID, msgSlot phase0.Slot) *queue.State {
    state := &queue.State{
        Slot:   msgSlot,
        Height: specqbft.Height(msgSlot),
        Quorum: v.Operator.GetQuorum(),
    }
    r := v.DutyRunners.DutyRunnerForMsgID(msgID)
    if r == nil {
        return state // still gate on slot/height even without a runner
    }
    // ... existing population from runner
    return state
}
```

The immediate liveness gap is closed by the `admission.go:26` fix in the separate comment (drop `!state.HasRunningInstance` from the guard). Longer term, consider centralizing all queue inserts (including `ExecuteDuty` and timeouts) behind one admission-aware helper so future control paths can't bypass it.

protocol/v2/ssv/validator/committee_queue_test.go
```diff
@@ -1197,6 +1197,94 @@ func TestHandleMessageQueueFullAndDropping(t *testing.T) {
 	assert.Nil(t, qContainer.Q.Pop(finalPopCtx, queue.NewCommitteeQueuePrioritizer(qContainer.queueState), queue.FilterAny))
 }
 
+func TestHandleMessageDropsStaleRoundUnderPressure(t *testing.T) {
```
olegshmuelov, line 1200: The committee path has this test exercising the full admission flow end-to-end (runner lookup → admission reject → queue size assertion → metric record). The validator path has no equivalent — `TestShouldAcceptUnderPressure` only tests the predicate in isolation. Since this PR changes behavior on every validator role queue (attester/proposer/sync-committee-contrib/aggregator/cluster), please add a mirror integration test for `Validator.EnqueueMessage` modeled on this one.

protocol/v2/ssv/queue/queue.go
olegshmuelov: `warnIfInboxIsTooBig` fires `logger.Warn("queue is half-full")` on every `Push`/`TryPush` once `len > cap/2`, with no rate limiting. Under sustained pressure (which this PR targets) that's hundreds of identical Warn logs per second, drowning out higher-signal alerts. The structured drop metrics this PR adds now capture the same condition cleanly.

Suggestion: remove `warnIfInboxIsTooBig`, or replace with a sampled/rate-limited observation (or a gauge instead of a log line) if you want to keep the 50% signal for ops.

olegshmuelov (changes requested):

olegshmuelov (dismissed):
LGTM!

ljuba-ssv (dismissed):

protocol/v2/ssv/runner/runner.go
```diff
@@ -138,6 +139,17 @@ func (b *BaseRunner) GetRole() spectypes.RunnerRole {
...
+func (b *BaseRunner) GetCurrentDutySlot() (phase0.Slot, bool) {
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
+
+	if b.State == nil || b.State.CurrentDuty == nil {
+		return 0, false
+	}
+
+	return b.State.CurrentDuty.DutySlot(), true
+}
```
iurii-ssv: The `b.mtx` has been removed in the latest **stage** branch, I'm not sure why this func shows up here (it has also been removed in https://github.com/ssvlabs/ssv/pull/2771),

which also means if this PR relies on `GetCurrentDutySlot`, and we need to preserve it in some form - we would need to ensure the callers are not reading it concurrently (`b.State` is not thread-safe by itself) .

iurii-ssv (changes requested):
Requesting `git merge` before we actually merge this one, since it seems to have some sort of artifacts (and Github doesn't seems to care/notice)

protocol/v2/ssv/validator/queue_validator.go
```diff
@@ -120,6 +164,14 @@ func (v *Validator) StartQueueConsumer(
...
 			state.Height = r.GetLastHeight()
 			state.Round = r.GetLastRound()
+			if slot, ok := r.GetCurrentDutySlot(); ok {
+				state.Slot = slot
+			} else {
+				state.Slot = phase0.Slot(state.Height)
+			}
+			if state.Height == 0 && state.Slot > 0 {
+				state.Height = specqbft.Height(state.Slot)
+			}
```
iurii-ssv: This block of code is not obvious at all, needs a clarification comment at least

last time I looked into it - I realized setting the `state.Slot` here isn't really worth this complexity, at least until we can merge `state.Height` and `state.Slot` to be a single field (`state.Slot`)

And this block of code is seemingly duplicated couple of times across this PR too.

iurii-ssv: So, what I suggest is:
- we move this change out of this PR (I assume it's orthogonal to it)
- if the change is not orthogonal (and is rather required to implement the "admission" thingy), we should pause this PR for now
- we merge `state.Height` + `state.Slot` to be just `state.Slot`
- we maybe refactor the way message-`Queue` works with runner-`State` to cut down on the unnecessary entanglement/dependencies
- and only then, we implement/apply this change on top of that refactor to better understand & document what this code actually does

**Update:** I actually sketched a [PR that clarifies how exactly the `queue.State` works](https://github.com/ssvlabs/ssv/pull/2777) recently (and now applied "`state.Height` + `state.Slot` merged to be just `state.Slot`" there) - so we should merge it first and review how the added code from this PR fits on top

protocol/v2/ssv/queue/admission.go
```diff
@@ -0,0 +1,60 @@
...
+	if threshold >= capacity && capacity > 1 {
+		threshold = capacity - 1
+	}
+	return buffered >= threshold
+}
+
+// ShouldAcceptUnderPressure keeps current or newer work admissible when inbox pressure is high.
+// It rejects only obviously stale messages, leaving strict-full behavior unchanged.
+func ShouldAcceptUnderPressure(state *State, msg *SSVMessage, buffered, capacity int) (bool, string) {
+	if !IsUnderPressure(buffered, capacity) || state == nil || msg == nil {
```
iurii-ssv: Nit: it doesn't make much sense to have these defensive checks here `|| state == nil || msg == nil` - it kinda makes it hard to figure out which of the checks actually matter (and which ones are "safety guards, that are also unreachable code")

protocol/v2/ssv/validator/committee_queue.go
```diff
@@ -64,19 +65,54 @@ func (c *Committee) EnqueueMessage(ctx context.Context, msg *queue.SSVMessage) {
 
 	c.mtx.Lock()
 	q := c.getQueue(logger, slot)
+	state := c.messageQueueState(slot)
 	c.mtx.Unlock()
 
+	queueID := queue.CommitteeMetricID(slot)
+	if accepted, dropReason := queue.ShouldAcceptUnderPressure(state, msg, q.Q.Len(), q.Q.Cap()); !accepted {
```
iurii-ssv: Couple points:
- the `queue.ShouldAcceptUnderPressure` call happens only before we are about to make a `TryPush` (and never when we are doing the blocking `Push`) ... I'm not sure why that is, seems like it should apply to both
- I'm pretty sure `queue.ShouldAcceptUnderPressure` should not be  a separate func, but rather a method on the target `q` itself (possibly even a private/unexposed method, called implicitly inside of `TryPush`/`Push` funcs) ... fact that we need to also expose `RecordDroppedMessage` outside of the `queue` package strongly hints at that too

iurii-ssv (commented):
We could also "separate the concerns" a bit better by [moving the metrics code into a separate wrapper](https://github.com/ssvlabs/ssv/compare/queue-pressure...queue-pressure-iurii-suggestions?expand=1), if you like. Which would be even more beneficial if we [move the admission logic inside of the Queue impl](https://github.com/ssvlabs/ssv/pull/2812#discussion_r3123264379).

operator/validator/router.go
```diff
@@ -25,9 +26,16 @@ type messageRouter struct {
...
+	default:
+	}
+
+	select {
 	case r.ch <- message:
 	default:
+		recordRouterMessageDrop(context.Background(), queue.DropReasonBufferFull)
 		r.logger.Warn("message router buffer is full, dropping message")
 	}
 }
```
iurii-ssv, line 83: TBH I don't much like how we are referencing those `queue.DropReason...` here. 

To keep the concerns separated, `messageRouter` shouldn't know anything about `queue` - it works with `r.ch` instead (abstracting the queue away, if you will) ... so, by that logic, if we are now importing the `queue` package here, might as well ditch the `r.ch` as well, right ?

protocol/v2/ssv/validator/committee_queue.go
nkryuchkov: I'm removing this for now due to https://github.com/ssvlabs/ssv/pull/2812#discussion_r3123231935

operator/validator/observability.go
```diff
@@ -64,3 +76,9 @@ func recordValidatorStatus(ctx context.Context, count uint32, status validatorSt
 		metric.WithAttributes(validatorStatusAttribute(status)),
 	)
 }
+
+func recordRouterMessageDrop(ctx context.Context, reason string) {
```
momosh-ssv: Any reason we're passing `context.Background()` from both call sites? The `ctx` parameter ends up unused in practice — and on the send path we have `Route`'s incoming ctx, which would carry tracing baggage / span context if we threaded it through. Maybe pass `Route`'s ctx on the send-path and keep `context.Background()` only on the cancel-path (where ctx is genuinely done), or drop the parameter and make this internal.

protocol/v2/ssv/queue/queue.go
```diff
@@ -86,21 +74,17 @@ func New(logger *zap.Logger, capacity int, opts ...Option) Queue {
 }
 
 func (q *priorityQueue) Push(msg *SSVMessage) {
-	q.recordInboxSize(int64(len(q.inbox)) + 1)
-
-	q.warnIfInboxIsTooBig()
-
 	q.inbox <- msg
+	q.observer.recordInboxSize(int64(len(q.inbox)))
```
momosh-ssv, line 77: Might be worth a comment here — `recordInboxSize` now reads `len(q.inbox)` after the send, so the consumer can drain before the read and the gauge will under-report peaks. The previous `+1` before send captured peak deterministically. For a 'current inbox size' metric I fear this can silently miss load. An async/observable gauge registered once that reads `len(q.inbox)` on collection would be both accurate and non-racy.

protocol/v2/ssv/queue/queue_observer.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	o.inboxSizeMetric.Record(
+		context.Background(),
+		inboxSize,
+		o.inboxSizeRecordOps...,
+	)
+}
+
+func (o metricsQueueObserver) recordDrop(reason string) {
+	if o.queueType == "" || o.queueID == "" {
```
momosh-ssv: Any reason we're keeping the `o.queueType == "" || o.queueID == ""` guard? Both are required positional params to `WithInboxSizeMetric`, so this can't fire in practice — and if a caller did pass empty strings, swallowing the drop silently feels worse than surfacing the misconfiguration. Seems like dead defensive code.

protocol/v2/ssv/queue/queue.go
```diff
@@ -43,6 +41,9 @@ type Queue interface {
 
 	// Len returns the number of messages in the queue.
 	Len() int
+
+	// Cap returns the inbox channel capacity.
+	Cap() int
```
momosh-ssv: Might be worth keeping `Cap()` unexported until there's an external caller — grep shows no usage outside this package. Same goes for `DroppedMessagesMetric` in `observability.go` (only referenced internally). Reduces interface surface for now.

momosh-ssv (dismissed):

protocol/v2/ssv/queue/observability.go
```diff
@@ -31,11 +33,22 @@ var (
...
+		),
+	)
 )
 
 const (
 	ValidatorQueueMetricType = "validator"
 	CommitteeQueueMetricType = "committee"
+
+	DropReasonBufferFull      = "buffer_full"
+	DropReasonContextCanceled = "context_canceled"
```
iurii-ssv: `DropReasonContextCanceled` is exported here but never referenced anywhere in the repo — `Push`/`TryPush` don't take a `ctx`, and the router uses its own private `routerDropReasonContextCanceled` constant in `operator/validator/observability.go`. Looks like a leftover from when admission had a context-cancel path. Suggest removing it (and `DropReasonBufferFull` could move to be unexported if `recordDroppedMessage` is the only caller — let me know if I'm missing a use).

operator/validator/router_test.go
```diff
@@ -62,3 +62,48 @@ func TestRouter(t *testing.T) {
...
+		SSVMessage: &spectypes.SSVMessage{
+			MsgType: spectypes.SSVConsensusMsgType,
+			MsgID:   spectypes.NewMsgID(networkconfig.TestNetwork.DomainType, []byte{1, 1, 1, 1, 1}, spectypes.RoleCommittee),
+			Data:    []byte("data"),
+		},
+	}
+
+	router.Route(ctx, msg)
+
+	require.Zero(t, router.Len())
```
iurii-ssv, line 88: Both new tests (`TestRouter_DropsWhenContextCanceled` here and `TestRouter_DropsWhenBufferFull` below) only assert the message wasn't pushed (`router.Len()`), not that the counter was incremented with the right reason. Since the whole point of this change is observability, a regression that silently disconnects `recordRouterMessageDrop` from these paths would still pass.

The queue side already wires a `metric.NewManualReader` in `TestPriorityQueue_TryPushInboxSizeMetricDoesNotExceedCapacityOnDrop` — same pattern would work here. Same comment applies to the queue side: there's no test asserting on `droppedMessagesMetric` either.

protocol/v2/ssv/queue/queue_observer.go
```diff
@@ -0,0 +1,62 @@
...
+
+type metricsQueueObserver struct {
+	inboxSizeMetric    metric.Int64Gauge
+	inboxSizeRecordOps []metric.RecordOption
+	queueType          string
+	queueID            string
+}
+
+// WithInboxSizeMetric configures queue-level observability for inbox size and dropped messages.
+func WithInboxSizeMetric(inboxSizeMetric metric.Int64Gauge, queueType, queueID string) Option {
```
iurii-ssv: Nit: the option now configures both inbox-size **and** drop telemetry (the doc-comment on L32 says as much), but the name still only mentions inbox size. Iurii's [suggestion branch](https://github.com/ssvlabs/ssv/compare/queue-pressure...queue-pressure-iurii-suggestions) renamed it to `WithQueueMetrics` — worth picking up, especially since there's only one caller anyway.

protocol/v2/ssv/queue/observability.go
```diff
@@ -51,3 +64,15 @@ func CommitteeMetricID(slot phase0.Slot) string {
 	slotInEpoch := slot % 32
 	return fmt.Sprintf("%d", slotInEpoch)
 }
+
+func recordDroppedMessage(queueType, queueID, reason string) {
```
iurii-ssv: Minor perf nit: builds three `attribute.String`s and a `metric.WithAttributes(...)` per call. The sibling `inboxSizeRecordOps` is pre-built once at queue construction. Drops should be rare in steady state, but the scenario this PR exists to surface is exactly the one where they aren't — under a sustained drop spiral this allocates on every dropped message.

Pre-building a per-observer drop-attr template the same way `inboxSizeRecordOps` is built (one per reason, or per `(queueType, queueID)` with the reason added per-call) would match the existing pattern. Non-blocking.

iurii-ssv (commented):
https://github.com/ssvlabs/ssv/pull/2842 addresses all the findings brought up in this review

iurii-ssv (approved):
Looks clean, I'd throw in this additional test coverage though https://github.com/ssvlabs/ssv/pull/2844

momosh-ssv (approved):

PR #2813, opened 2026-04-16
Synchronize access to current subnets

julienharbulot:
T-ssv-028 — Race Condition in UpdateSubnets

Original ticket summary:
- `network/p2p/p2p.go:466-551` updates `n.currentSubnets` without synchronization while `peersTrimming()` and `ActiveSubnets()` read it concurrently.
- Risk: torn reads of subnet state can lead to incorrect peer trimming or discovery registration.

Changes:
- synchronize reads and writes of `currentSubnets` with a dedicated lock
- use stable subnet snapshots in trim / active-subnet / random-subnet paths
- add a focused concurrent-access regression test

Fixes - Fixes ssvlabs/ssv-node-board#917

### DIFF
--- a/network/p2p/p2p_setup.go
+++ b/network/p2p/p2p_setup.go
@@ -257,7 +257,7 @@ func (n *p2pNetwork) setupPeerServices() error {
 }
 
 func (n *p2pNetwork) ActiveSubnets() p2pcommons.Subnets {
-	return n.currentSubnets
+	return n.currentSubnetsSnapshot()
 }
 
 func (n *p2pNetwork) FixedSubnets() p2pcommons.Subnets {
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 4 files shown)


olegshmuelov (approved):
LGTM

network/p2p/p2p_setup.go
```diff
@@ -257,7 +257,7 @@ func (n *p2pNetwork) setupPeerServices() error {
 }
 
 func (n *p2pNetwork) ActiveSubnets() p2pcommons.Subnets {
-	return n.currentSubnets
+	return n.currentSubnetsSnapshot()
```
nkryuchkov, line 260: `ActiveSubnets` and `currentSubnetsSnapshot` are doing the same thing, can we merge them into one method?

nkryuchkov (approved):

ljuba-ssv (approved):
LGTM — tight, minimal fix for the documented `currentSubnets` race.

One follow-up worth opening a ticket for: `persistentSubnets` has the same goroutine-crossing pattern and is still unprotected. `SubscribeRandoms` writes it (and carries a `// This method isn't thread-safe` comment acknowledging the issue), while `SubscribedSubnets()`, `subscribeToFixedSubnets`, and discovery setup read it from other goroutines. Not in scope here, but the same `RWMutex` + accessor pattern introduced in this PR would drop in cleanly.

PR #2815, opened 2026-04-16
operator: add exporter network key regression tests

nkryuchkov:
Fixes https://github.com/ssvlabs/ssv-node-board/issues/1008.

## Summary

Add regression tests for the intended exporter behavior around persisted P2P network keys.

## What changed

- add tests that verify exporter keeps the persisted network key plaintext on a fresh DB
- add tests that verify exporter still keeps it plaintext even if `OPERATOR_KEY` is passed
- add tests that verify exporter still keeps it plaintext even if `SSVSigner.Endpoint` is passed
- add a test that verifies exporter fails clearly if it encounters an already-encrypted persisted network key without a protector

## Why

Exporter is not supposed to use signing config. These tests lock in that contract so we do not accidentally reintroduce exporter-side network key encryption again.

## Operational note

This does not recover already-encrypted exporter DBs. Those still require operational cleanup:
- wipe/reset the DB

### DIFF
--- a/cli/operator/node_test.go
+++ b/cli/operator/node_test.go
@@ -63,6 +63,58 @@ func Test_warnIfSSVAPIAddressUnset(t *testing.T) {
 	})
 }
 
+func Test_warnIfExporterSigningConfigProvided(t *testing.T) {
+	originalCfg := cfg
+	t.Cleanup(func() { cfg = originalCfg })
+
+	t.Run("does not warn when exporter has no signing config", func(t *testing.T) {
+		cfg = config{}
+
+		core, recorded := observer.New(zapcore.WarnLevel)
+		logger := zap.New(core)
+
+		warnIfExporterSigningConfigProvided(logger)
+
+		require.Len(t, recorded.All(), 0)
+	})
+
+	t.Run("warns when exporter is given operator key", func(t *testing.T) {
+		cfg = config{
+			OperatorPrivateKey: "operator-key",
+		}
+
+		core, recorded := observer.New(zapcore.WarnLevel)
+		logger := zap.New(core)
+
+		warnIfExporterSigningConfigProvided(logger)
+
+		logs := recorded.All()
+		require.Len(t, logs, 1)
+		require.Equal(t, zapcore.WarnLevel, logs[0].Level)
+		require.Equal(t, "exporter mode ignores operator signing configuration", logs[0].Message)
+		require.EqualValues(t, len(cfg.OperatorPrivateKey), logs[0].ContextMap()["operator_private_key_len"])
+	})
+
+	t.Run("warns when exporter is given ssv-signer config", func(t *testing.T) {
+		cfg = config{
+			SSVSigner: SSVSignerConfig{
+				Endpoint: "https://signer.example",
+			},
+		}
+
+		core, recorded := observer.New(zapcore.WarnLevel)
+		logger := zap.New(core)
+
+		warnIfExporterSigningConfigProvided(logger)
+
+		logs := recorded.All()
+		require.Len(t, logs, 1)
+		require.Equal(t, zapcore.WarnLevel, logs[0].Level)
+		require.Equal(t, "exporter mode ignores operator signing configuration", logs[0].Message)
+		require.Equal(t, cfg.SSVSigner.Endpoint, logs[0].ContextMap()["ssv_signer_endpoint"])
+	})
+}
+
 func Test_ssvAPIListenAddress(t *testing.T) {
 	t.Parallel()
 
--- a/cli/operator/node_test.go
+++ b/cli/operator/node_test.go
@@ -471,18 +471,64 @@ func Test_resolveExporterP2PNetworkKeyProtector(t *testing.T) {
 		require.Len(t, encryptionKey, 32)
 	})
 
+	t.Run("ignores incomplete keystore config when operator key is available", func(t *testing.T) {
+		privateKey, err := keys.GeneratePrivateKey()
+		require.NoError(t, err)
+
+		cfg = config{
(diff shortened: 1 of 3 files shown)


nkryuchkov:
@greptileai please review this

nkryuchkov:
@greptileai review this again

nkryuchkov:
@greptileai review this again

nkryuchkov:
@greptileai review this again

cli/operator/node_test.go
```diff
@@ -63,6 +65,118 @@ func Test_warnIfSSVAPIAddressUnset(t *testing.T) {
...
+		store := ssv_identity.NewIdentityStore(logger, db, nil, nil)
+		_, err = store.SetupNetworkKey(ctx, "")
+		require.NoError(t, err)
+
+		encrypted, err := ssv_identity.HasEncryptedNetworkKey(db)
+		require.NoError(t, err)
+		require.False(t, encrypted)
+	})
+
+	t.Run("exporter ignores ssv-signer config and still stores plaintext", func(t *testing.T) {
```
olegshmuelov: These two subtests don't lock in the contract their names claim. They set `cfg.OperatorPrivateKey` and `cfg.SSVSigner.Endpoint`, then construct the store via `ssv_identity.NewIdentityStore(logger, db, nil, nil)` and call `SetupNetworkKey` directly. Nothing reads back the cfg — the production wiring that actually decides whether to install protectors lives in `cli/operator/node.go:setupP2P` (lines 1121-1170), where `protectFn`/`unprotectFn` are derived from `operatorPrivKey` and `signerClient`, not from `cfg` at this site. The test would pass identically even if a future refactor reintroduced exporter-side encryption (the regression these tests are meant to catch).

A few options:

1. Refactor the protector decision out of `setupP2P` into a small pure function that takes `(exporterEnabled bool, operatorPrivKey, signerClient)` → `(protectFn, unprotectFn)`, and exercise it. Best long-term.
2. At minimum, rename the subtests to honestly reflect what they assert — e.g. `"identity store with no protectors stays plaintext"` — and drop the dead `cfg.OperatorPrivateKey` / `cfg.SSVSigner.Endpoint` setup that misleads the reader.
3. Add an inline comment pointing at the wiring file/line that this test is _trying_ to lock down, and a `// TODO` that the wiring path itself isn't covered by this test.

cli/operator/node_test.go
```diff
@@ -63,6 +65,118 @@ func Test_warnIfSSVAPIAddressUnset(t *testing.T) {
 	})
 }
 
+func TestExporterNetworkKeyStorage(t *testing.T) {
```
olegshmuelov: issue (non-blocking): Test names oversell what's actually verified.

All four `TestExporterNetworkKeyStorage` subtests construct `IdentityStore` with explicit nil protectors, bypassing the wiring in `setupP2P` (lines 1121-1170) where `protectFn`/`unprotectFn` are decided. The `cfg.OperatorPrivateKey` set in `exporter ignores operator key…` and the `cfg.SSVSigner.Endpoint` set in `exporter ignores ssv-signer config…` are never read by any code path here. A regression that wires up a protector in exporter mode (the bug class this is meant to catch) would still pass green.

The fix is to extract the protector-decision logic out of `setupP2P` into `decideNetworkKeyProtectors(exporterEnabled, operatorPrivKey, signerClient) (protectFn, unprotectFn, error)` and exercise that. The current `setupP2P` is hard to unit-test (creates libp2p, calls `logger.Fatal`); a 20-line extraction is enough.

cli/operator/node_test.go
```diff
@@ -63,6 +65,118 @@ func Test_warnIfSSVAPIAddressUnset(t *testing.T) {
...
+		store := ssv_identity.NewIdentityStore(logger, db, nil, nil)
+		_, err = store.SetupNetworkKey(ctx, "")
+		require.NoError(t, err)
+
+		encrypted, err := ssv_identity.HasEncryptedNetworkKey(db)
+		require.NoError(t, err)
+		require.False(t, encrypted)
+	})
+
+	t.Run("exporter fails clearly on encrypted db key", func(t *testing.T) {
```
olegshmuelov: suggestion (non-blocking): `exporter fails clearly on encrypted db key` duplicates `identity/store_test.go:404`.

Same encrypted-write-then-read-with-nil-protectors setup, same `require.ErrorContains(..., "network key is encrypted but no compatible network key protector is configured")` assertion. Driving via `SetupNetworkKey` vs `GetNetworkKey` and using `keys.GeneratePrivateKey()` vs a static test key don't exercise new behavior. Worth dropping this subtest and adding `t.Run("via SetupNetworkKey", ...)` inside the existing identity-package test instead — cheaper, and the coverage lives next to the code.

cli/operator/node_test.go
```diff
@@ -63,6 +65,118 @@ func Test_warnIfSSVAPIAddressUnset(t *testing.T) {
 	})
 }
 
+func TestExporterNetworkKeyStorage(t *testing.T) {
+	originalCfg := cfg
+	t.Cleanup(func() { cfg = originalCfg })
```
olegshmuelov: nit: Outer `cfg` snapshot/restore leaks subtest pollution.

Each subtest fully replaces `cfg = config{...}` without its own cleanup. Works today because subtests run sequentially, but will break the moment anyone adds `t.Parallel()` to them. Push the snapshot/restore down into each subtest, or add a small `withCfg(t, config{...})` helper.

nkryuchkov:
@greptileai review this again

iurii-ssv (approved):
LGTM

ljuba-ssv (approved):

PR #2816, opened 2026-04-20
Add tests for beacon aggregator aggregate selection proof

julienharbulot:
## Ticket
`T-ssv-036` — No Tests for Beacon Aggregator API

Original issue:
- `beacon/goclient/aggregator.go` had no coverage for `SubmitAggregateSelectionProof`
- Risk called out incorrect fork handling for Electra aggregate attestations (`attData.Index = 0` vs pre-Electra)
- Risk also called out context cancellation during `waitTwoThirdsIntoSlot`

## Changes
- add fork-aware tests covering `Phase0` through `Fulu` for `SubmitAggregateSelectionProof`
- assert pre-Electra uses the committee index and Electra/Fulu use index `0`
- add a cancellation test proving the method exits during the 2/3-slot wait before beacon-node requests
- harden fork selection to derive from the requested slot consistently

Fixes - Fixes ssvlabs/ssv-node-board#925

### DIFF
--- a/beacon/goclient/aggregator_test.go
+++ b/beacon/goclient/aggregator_test.go
@@ -1,0 +1,254 @@
+package goclient
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/attestantio/go-eth2-client/api"
+	"github.com/attestantio/go-eth2-client/mock"
+	"github.com/attestantio/go-eth2-client/spec"
+	"github.com/attestantio/go-eth2-client/spec/electra"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/jellydator/ttlcache/v3"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/networkconfig"
+)
+
+type aggregatorClientMock struct {
+	mock.Service
+}
+
+func (*aggregatorClientMock) SubmitProposal(context.Context, *api.SubmitProposalOpts) error {
+	return nil
+}
+
+func TestSubmitAggregateSelectionProof_UsesForkSpecificAggregateAndProof(t *testing.T) {
+	t.Parallel()
+
+	slotSig := make([]byte, 96)
+	for i := range slotSig {
+		slotSig[i] = byte(i + 1)
+	}
+
+	validatorIndex := phase0.ValidatorIndex(42)
+	committeeIndex := phase0.CommitteeIndex(7)
+
+	testCases := []struct {
+		name          string
+		version       spec.DataVersion
+		epoch         phase0.Epoch
+		expectedIndex phase0.CommitteeIndex
+	}{
+		{
+			name:          "phase0 uses committee index",
+			version:       spec.DataVersionPhase0,
+			epoch:         0,
+			expectedIndex: committeeIndex,
+		},
+		{
+			name:          "altair uses committee index",
+			version:       spec.DataVersionAltair,
+			epoch:         1,
+			expectedIndex: committeeIndex,
+		},
+		{
+			name:          "bellatrix uses committee index",
+			version:       spec.DataVersionBellatrix,
+			epoch:         2,
+			expectedIndex: committeeIndex,
+		},
+		{
+			name:          "capella uses committee index",
+			version:       spec.DataVersionCapella,
+			epoch:         3,
+			expectedIndex: committeeIndex,
+		},
+		{
+			name:          "deneb uses committee index",
+			version:       spec.DataVersionDeneb,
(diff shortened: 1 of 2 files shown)


nkryuchkov (dismissed):

beacon/goclient/aggregator_test.go
```diff
@@ -0,0 +1,254 @@
...
+			version:       spec.DataVersionElectra,
+			epoch:         5,
+			expectedIndex: 0,
+		},
+		{
+			name:          "fulu uses zero index",
+			version:       spec.DataVersionFulu,
+			epoch:         6,
+			expectedIndex: 0,
+		},
```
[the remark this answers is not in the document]
julienh-ssv: Addressed in a7833c53b by removing the redundant Fulu case from the table-driven fork coverage and keeping the Electra case for the shared post-Electra aggregate path.

beacon/goclient/aggregator_test.go
```diff
@@ -0,0 +1,254 @@
...
+	switch version {
+	case spec.DataVersionElectra, spec.DataVersionFulu:
+		proof, ok := gotProof.(*electra.AggregateAndProof)
+		require.True(t, ok)
+		require.Equal(t, validatorIndex, proof.AggregatorIndex)
+		require.Equal(t, slotSig, proof.SelectionProof[:])
+		require.Equal(t, expectedIndex, proof.Aggregate.Data.Index)
+	default:
+		proof, ok := gotProof.(*phase0.AggregateAndProof)
+		require.True(t, ok)
```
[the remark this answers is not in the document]
julienh-ssv: Addressed in a7833c53b by switching the aggregate-and-proof type assertions to require.Truef so failures report the expected and actual types.

beacon/goclient/aggregator_test.go
```diff
@@ -0,0 +1,248 @@
...
+			version:       spec.DataVersionDeneb,
+			epoch:         4,
+			expectedIndex: committeeIndex,
+		},
+		{
+			name:          "electra uses zero index",
+			version:       spec.DataVersionElectra,
+			epoch:         5,
+			expectedIndex: 0,
+		},
```
olegshmuelov, line 83: suggestion: Table sets `attData.Slot = slot`, so the same test passes under the old `attData.Slot`-based fork derivation and the new `slot`-based one. The actual defense this PR ships is not exercised.

  Add a case where the beacon node returns an `attData` whose `Slot` sits in a pre-Electra epoch while the duty `slot` is post-Electra, and assert `attData.Index` is forced to `0`. That case fails against the pre-fix implementation and passes now — which is the behavioral invariant worth locking in.

  ```go
  {
      name:            "electra duty with pre-electra attData.Slot forces zero index",
      version:         spec.DataVersionElectra,
      epoch:           5,
      attDataSlotEpoch: 4, // beacon node returned stale pre-Electra slot
      expectedIndex:   0,
  },
  ```

beacon/goclient/aggregator_test.go
```diff
@@ -0,0 +1,248 @@
...
+		errCh := make(chan error, 1)
+
+		go func() {
+			_, _, err := client.SubmitAggregateSelectionProof(ctx, 0, 1, 128, 42, make([]byte, 96))
+			errCh <- err
+		}()
+
+		time.Sleep(cfg.IntervalDuration())
+		cancel()
+		time.Sleep(time.Millisecond)
```
olegshmuelov: suggestion: Drop this sleep. Under `synctest.Test`, the bubble only advances synthetic time when every goroutine is durably blocked, and the `err := <-errCh` below is itself a durable block that serializes the send → receive without needing a time advance. The line is a coincidence-wait that obscures the determinism the test depends on.

  ```suggestion
  	cancel()
  ```

beacon/goclient/aggregator_test.go
julienharbulot: Addressed in 763382e48 by adding a regression case where the duty slot is Electra but the beacon node returns a stale pre-Electra attestation slot; the test now proves we derive fork handling from the duty slot and still force index 0.

beacon/goclient/aggregator_test.go
julienharbulot: Addressed in 763382e48 by removing the extra post-cancel sleep; the synctest path now relies on the blocking receive for deterministic synchronization.

iurii-ssv (approved):
LGTM, seems like we could make Fulu tests a bit more complete/tight:

The new fork table never includes a `spec.DataVersionFulu` case, even though `SubmitAggregateSelectionProof` has a dedicated Fulu branch and the PR description says coverage runs through Fulu. As written, the newest fork still relies on the Electra case as a proxy, so a regression specific to the Fulu path would slip through.

ljuba-ssv (approved):

PR #2817, opened 2026-04-20
[T-ssv-038] Add tests for concurrent beacon head events

julienharbulot:
## Ticket
- `T-ssv-038`
- Original issue: Beacon event handler concurrency behavior in `beacon/goclient/events.go` was not covered by tests, specifically concurrent head-event deduplication via `lastProcessedEventSlot` and non-blocking drops when subscriber channels are full.
- Fixes - Fixes ssvlabs/ssv-node-board#927

## Changes
- add a concurrency-focused test that exercises overlapping head events and verifies only the newest slot is broadcast and cached
- add an overflow test that verifies a full subscriber channel drops the broadcast without losing internal head-cache updates

## Validation
- `go test -tags "blst_enabled lfs" -race ./beacon/goclient -run 'TestSubscribeToHeadEvents|TestNewEventHandler'`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -v ./beacon/goclient/...`


beacon/goclient/events_test.go
```diff
@@ -77,6 +80,114 @@ func TestSubscribeToHeadEvents(t *testing.T) {
...
+				handler(&apiv1.Event{
+					Topic: string(eventTopicHead),
+					Data: &apiv1.HeadEvent{
+						Slot:  slot,
+						Block: block,
+					},
+				})
+			}(i)
+		}
+		wg.Wait()
```
olegshmuelov, line 136: issue (non-blocking): no start barrier — goroutines can serialize and the subtest stays green even if `lastProcessedEventSlotLock` is deleted. Gate the launch behind a release channel so all 32 race the handler at once.

  ```go
  start := make(chan struct{})
  var wg sync.WaitGroup
  for i := 0; i < 32; i++ {
      wg.Add(1)
      go func(i int) {
          defer wg.Done()
          <-start
          slot, block := phase0.Slot(10), initialRoot
          if i%2 == 0 {
              slot, block = 11, nextRoot
          }
          handler(&apiv1.Event{
              Topic: string(eventTopicHead),
              Data:  &apiv1.HeadEvent{Slot: slot, Block: block},
          })
      }(i)
  }
  close(start)
  wg.Wait()
  ```

beacon/goclient/events_test.go
```diff
@@ -77,6 +80,114 @@ func TestSubscribeToHeadEvents(t *testing.T) {
...
+		assert.Equal(t, nextRoot, nextEvent.Block)
+
+		headAtSlot10 := client.headCache.Get(10)
+		require.NotNil(t, headAtSlot10)
+		assert.Equal(t, initialRoot, headAtSlot10.Value())
+
+		headAtSlot11 := client.headCache.Get(11)
+		require.NotNil(t, headAtSlot11)
+		assert.Equal(t, nextRoot, headAtSlot11.Value())
+	})
```
olegshmuelov, line 151: suggestion (non-blocking): scenario only races slot=10 vs slot=11, so a monotonicity regression between two *advancing* slots (e.g. 11 vs 12) wouldn't be caught. Add a sibling subtest that races two advancing slots after the prime and asserts only the highest reaches the subscriber.

beacon/goclient/events_test.go
julienh-ssv: Addressed in f8bef990c by gating the goroutines behind a start barrier so the concurrent subtest actually races the handler.

beacon/goclient/events_test.go
julienh-ssv: Addressed in f8bef990c by adding a sibling subtest that primes slot 10, races advancing slots 11 and 12, and asserts only the highest slot reaches the subscriber.

beacon/goclient/events_test.go
```diff
@@ -77,6 +80,185 @@ func TestSubscribeToHeadEvents(t *testing.T) {
...
+						Slot:  slot,
+						Block: block,
+					},
+				})
+			}(i)
+		}
+		close(start)
+		wg.Wait()
+
+		require.Len(t, eventsCh, 1)
```
momosh-ssv, line 194: Could this lead to flakes? Seems that the broadcast and `headCache.Set` happen after `lastProcessedEventSlotLock` is released in `events.go`, so if a slot=11 goroutine wins the gate first and a slot=12 goroutine acquires the lock next (12>11 passes), both can reach `eventsCh` and this assertion would see 2 events. Maybe drain the channel and assert the final event is slot=12, or feed slot=12 then slot=11 sequentially and assert slot=11 is dropped by monotonicity.

beacon/goclient/events_test.go
```diff
@@ -77,6 +80,185 @@ func TestSubscribeToHeadEvents(t *testing.T) {
 	})
 }
 
+func TestNewEventHandler(t *testing.T) {
```
iurii-ssv, line 83: The new subtest at [beacon/goclient/events_test.go#L153-L218](https://github.com/ssvlabs/ssv/blob/f8bef990c245450c06d96f0017afa976a9eb29ad/beacon/goclient/events_test.go#L153-L218) asserts behavior that the current handler in [beacon/goclient/events.go#L181-L209](https://github.com/ssvlabs/ssv/blob/f8bef990c245450c06d96f0017afa976a9eb29ad/beacon/goclient/events.go#L181-L209) does not guarantee. When `11` wins the lock before `12`, both are strictly increasing slots and both are broadcast, so the single-message expectation at [beacon/goclient/events_test.go#L206](https://github.com/ssvlabs/ssv/blob/f8bef990c245450c06d96f0017afa976a9eb29ad/beacon/goclient/events_test.go#L206) is flaky. I reproduced this repeatedly on [f8bef99](https://github.com/ssvlabs/ssv/commit/f8bef990c245450c06d96f0017afa976a9eb29ad) with `go test ./beacon/goclient -run 'TestNewEventHandler' -count=100`; the failure consistently reported 2 items in the channel instead of 1.

iurii-ssv (commented):
**Cleanup Notes**
- [beacon/goclient/events_test.go#L84-L259](https://github.com/ssvlabs/ssv/blob/f8bef990c245450c06d96f0017afa976a9eb29ad/beacon/goclient/events_test.go#L84-L259) repeats the same minimal client/channel setup three times; a small helper would make the intent easier to read once the flaky case is corrected.
- If you want deterministic coverage for the ordering contract in [beacon/goclient/events.go#L181-L209](https://github.com/ssvlabs/ssv/blob/f8bef990c245450c06d96f0017afa976a9eb29ad/beacon/goclient/events.go#L181-L209), the test should control the relative ordering explicitly instead of relying on scheduler timing after `close(start)`.

beacon/goclient/events_test.go
julienharbulot: Addressed in 5b69f1ca1 by replacing the racy 11-vs-12 concurrent assertion with a deterministic monotonicity check: after priming slot 10, the test feeds slot 12 and then slot 11 sequentially and verifies the stale 11 is dropped.

beacon/goclient/events_test.go
julienharbulot: Addressed in 5b69f1ca1 by changing that subtest to assert only behavior guaranteed by 
'the handler: once slot 12 is processed, a later slot 11 is rejected by monotonicity and never reaches the subscriber or cache.

iurii-ssv (approved):

olegshmuelov (approved):

momosh-ssv (approved):

PR #2819, opened 2026-04-20
tests: (p2p, exporter/api) robust port allocation

iurii-ssv:
This PR started as a fix for 3 test-flakiness issues; while in there I
also folded in some directly-related refactors and dead-code removal so
the same code wouldn't need touching again immediately afterwards.

================================================================================

**Latent bugs fixed in passing (in `*wsSession.sendMsg`)**

While renaming `conn` → `wsSession` I noticed two pre-existing bugs
in `sendMsg` and fixed them in the same pass:

1. `w.Close` was not called on `w.Write` error, leaking the frame
   writer. Masked in practice because `WriteLoop` tears down the whole
   session on any `sendMsg` error.
2. The write deadline used `pingTimeout` (60s, a read-side ping
   constant) instead of `s.writeTimeout` (the configured per-session
   write timeout, ~3s in the exporter). Effectively made write
   deadlines much looser than intended.

Both are unrelated to the flakiness work but are part of this PR.

================================================================================

**Issue 1)**

Two root causes made ./network/p2p tests flaky when run concurrently
with others, forcing the Makefile to invoke them in a separate `go test`
command:

1. TOCTOU in RandomTCPPort: tests pre-checked port availability via
    DialTimeout and then later asked libp2p to bind to that port. Under
    -p 16 -parallel 256 CPU pressure, another test could grab the port
    in the gap. Use TCP/UDP port 0 so the kernel picks free ports
    atomically at bind time.

2. mDNS service tag collision: all test processes advertised on the
    shared "ssv.discovery" tag, so concurrently-running test groups
    cross-discovered each other's peers. Plumb a per-LocalNet random tag
    (ssv.test.<hex>) through `p2p.Config.MdnsDiscoveryTag` and into
    `discovery.NewLocalDiscovery`'s new `serviceTag` parameter; empty
    falls back to `LocalDiscoveryServiceTag`, so production is unchanged.
    The same per-process random tag is also applied to `network/topics/controller_test.go`'s
    `newPeers`, which had the same shared-default-tag exposure.

With both fixed, ./network/p2p can run alongside the rest of the suite,
so merge the unit-test target's two `go test` invocations (and coverage
file stitching) back into a single command.

================================================================================

**Issue 2)**

`getRandomPort` picked a random port in [8001,14000) and probed it with
`DialTimeout`. Between the probe and the server's later bind, another
process could grab the port — intermittent CI failures.

Listening on ":0" resolves that.

================================================================================

**Issue 3)**

Resolves `TestP2pNetwork_MessageValidation` flakiness (see https://github.com/ssvlabs/ssv/actions/runs/24669605966/job/72136402325?pr=2819) by adjusting its expectations:
- it previously would validate the amount of received messages (wrong, because once the peer-scores cross a certain threshold due to rejects - network starts dropping messages coming from those peers ... hence we can't assert on it, and "how many messages are dropped" is not identifiable/deterministic)
- the strict ordering assertion flaked because peers with identical rejected-rates (nodes 0 and 1) differ only within scoring noise

Replace both with a rejected-rate invariant: peers with a strictly lower
rejected-rate must not be ranked below peers with a higher rate. Ties within
the same rate group are allowed.

To keep regression power where the rate check alone is too loose (nodes 0
and 1 both have rate=0), also assert a bucket invariant on each observer:
accepted-only ≥ ignored-only ≥ rejected-only. Without it, a regression
that stops rewarding accepts or starts rewarding ignores would slip through.

The rate+bucket invariants are polled via `assert.Eventually` (15s ceiling,
100ms tick) — fast machines exit in a few ticks, slow CI gets headroom,
and on timeout the test falls through to diagnostic asserts that surface
which invariant on which observer failed.

================================================================================

**Refactors that came along**

- `api/server` and `exporter/api`: `Run()` → `Start(addr) (boundAddr, <-chan error, error)`. Start binds before returning so callers using `":0"` learn the kernel-assigned port; the returned channel emits a non-graceful serve error and is closed silently on ctx-driven shutdown. Both call sites (`cli/operator/node.go`, `operator/node.go`'s `startWSServer`) now fatal-exit only on non-graceful serve errors — restoring the supervisor-restart shape. Previously, `cli/operator/node.go` Fatal'd on *any* `Run()` error including `http.ErrServerClosed` from graceful shutdown, and `operator/node.go` silently swallowed the error in a goroutine, leaving the process running without an accept loop.

- `network/discovery`: drop the polymorphic `NewService`; expose `NewDiscV5Service` and `NewLocalDiscovery` directly so callers pick the one they want. `NewLocalDiscovery` gains a `serviceTag` param (consumed by Issue 1).

- `exporter/api/broadcaster`: document the `broadcasted.Send` contract (must not block) and rework fan-out around it. `Broadcast` now marshals once, snapshots the connection set under the lock, and fans out outside it — no per-message goroutines, no per-conn worker, lock-hold bounded by map-copy so a misbehaving `Send` can't hold up `Register`/`Deregister` or the next `Broadcast`. `*wsSession.Send` enforces the contract: a non-blocking `select+default` on `s.send` (cap 256); on overflow it tears the session down via `Close` so the slow client reconnects with a fresh, consistent view rather than silently missing decideds. `FromFeed` subscribes synchronously and pumps inline until ctx cancels.

- `exporter/api/server`: `registerHandler` wraps each handler with a ctx watcher that closes the upgraded websocket on server ctx cancel, so an idle `/query` connection blocked in `ReadJSON` doesn't survive shutdown. `http.Server.Shutdown` doesn't close upgraded websockets, so without this the handler goroutine would leak past shutdown.

- `exporter/api/ws_session` (renamed from `conn`): clarifies ownership — `*wsSession` owns its ctx + send queue; the underlying `*websocket.Conn` lifecycle belongs to `wsServer`'s wrappedHandler (which closes it on its own defer). `Close` is just `cancelCtx` — idempotent, safe from any goroutine; `WriteLoop`, `ReadLoop`, `handleStream` defers, and `Send`-overflow all converge on it. `pingLoop` swaps `time.NewTimer` + unconditional `<-t.C` for a ticker + select on ctx, so a quiet session no longer waits up to `pingInterval` (~48s) past ctx cancel.

- Dead code removed: `LocalNet.WithBootnode`, `LocalNetOptions.UseDiscv5`, the exporter/api `getRandomPort` + `checkPort` helpers, `testutils.udpRand`, the entire `network/testing/net.go` (`RandomTCPPort` / `UDPPortsRandomizer` + their helpers — zero callers after the above), an unused `time.NewTimer` in `WriteLoop`'s `withPing` branch, and the `MockMessageValidator.Total*` counters + `roleBroadcasts` tracker that fed the removed count assertion in Issue 3.

- Misc: `wsServer` logs the actually-registered endpoints instead of a hard-coded list; the per-LocalNet mDNS tag uses `math/rand` (test-only uniqueness, no crypto need).

================================================================================

Resolves https://github.com/ssvlabs/ssv-node-board/issues/941


iurii-ssv:
@greptile pls re-review

cli/operator/node.go
```diff
@@ -693,12 +693,9 @@ var StartNodeCmd = &cobra.Command{
 				hexporter.NewExporter(logger, storageMap, collector, nodeStorage.ValidatorStore()),
 				cfg.ExporterOptions.Enabled && cfg.ExporterOptions.Mode == exporter.ModeArchive,
 			)
-			go func() {
-				err := apiServer.Run()
-				if err != nil {
-					logger.Fatal("failed to start API server", zap.Error(err))
-				}
-			}()
+			if _, err := apiServer.Start(cfg.SSVOptions.Context); err != nil {
```
olegshmuelov: issue (blocking): Bind errors still fatal, but `httpServer.Serve` failures are now Error-logged only — the supervisor restart on accept-loop death is gone. Same shape in `n.ws.Start` at [`operator/node.go:440`](https://github.com/ssvlabs/ssv/blob/daed33b494b39406bc53648773c4180494151a00/operator/node.go#L440). Either return a serve-completion channel and keep fatal-on-exit, or flip readiness on serve exit.

exporter/api/broadcaster.go
```diff
@@ -36,26 +37,26 @@ func newBroadcaster(logger *zap.Logger) Broadcaster {
...
-
-	for {
-		select {
-		case msg := <-cn:
-			go func(msg Message) {
+// FromFeed subscribes to the given feed synchronously so any subsequent Send
+// on the feed is guaranteed to reach this broadcaster, then spawns a pump
+// goroutine that forwards messages to registered connections until ctx is
+// canceled.
+func (b *broadcaster) FromFeed(ctx context.Context, msgFeed *event.Feed) {
```
olegshmuelov, line 52: `event.Feed.Send` blocks when the 512-buffer fills; decided publishers call it synchronously. The old `go b.Broadcast(msg)` per-message drained it instantly — now `Marshal` + fan-out runs inline before the next read. Push the broadcast onto an async worker so stream clients can't back-pressure decideds.

network/p2p/p2p_validation_test.go
```diff
@@ -272,33 +238,26 @@ func TestP2pNetwork_MessageValidation(t *testing.T) {
...
 		require.Equal(t, len(vNet.Nodes)-1, len(peers), "node %d", node.Index)
-		for i, validOrder := range validOrders {
-			valid := true
-			for j, peer := range peers {
-				if peer.index != validOrder[j] {
-					valid = false
-					break
+		for i := 0; i < len(peers); i++ {
+			for j := i + 1; j < len(peers); j++ {
+				if rejectedRate[peers[i].index] > rejectedRate[peers[j].index] {
```
olegshmuelov: Both accepted-only and ignored-only nodes have rate=0, so the new check treats them as interchangeable. A regression that stops rewarding accepts or starts rewarding ignores now passes. Add an explicit accepted-bucket ≥ ignored-bucket ≥ rejected-bucket assertion alongside the rate check.

iurii-ssv: Should be addressed in latest commits now.

cli/operator/node.go
iurii-ssv: Well spotted, fixed in latest commits

exporter/api/broadcaster.go
iurii-ssv: Upon further investigation, this issue goes deeper actually, while "The old `go b.Broadcast(msg)` per-message drained it instantly" did indeed not have this issue, but it was doing either/or:
- spawning potentially unlimited number of go-routines (meaning, a slow client(s) could cause server to go OOM)
- "dropping messages on the floor" if the client couldn't keep up

neither of which is ideal, not sure if it cause any real issues in practice ... but since I've started on it already, I've reworked it a bit:

> The new design specifically prevents back-pressure to the publisher: each c.Send is non-blocking (cap-256 select+default), and a slow client that overflows its buffer is Closed so the next Broadcast skips it. The broadcaster goroutine therefore never blocks on any client, so the 512-buffer in front of it drains at marshal-speed, so outFeed.Send from the decided publisher never blocks for long. Slow clients reconnect with a fresh consistent view rather than dragging the publisher down.

network/topics/controller_test.go
```diff
@@ -380,7 +380,7 @@ func newPeers(ctx context.Context, logger *zap.Logger, t *testing.T, n int, msgV
 func newPeer(ctx context.Context, logger *zap.Logger, t *testing.T, msgValidator validation.MessageValidator, msgID bool, scoreInspector pubsub.ExtendedPeerScoreInspectFn) *P {
 	h, err := libp2p.New(libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/0"))
 	require.NoError(t, err)
-	ds, err := discovery.NewLocalDiscovery(ctx, logger, h)
+	ds, err := discovery.NewLocalDiscovery(ctx, logger, h, "")
```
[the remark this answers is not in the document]
iurii-ssv: Addressed with a random hex tag instead (per-call `rand.Uint64()`) — same idea, but also safe across processes: the same-named test running in two `go test` workers would still collide on a `t.Name()`-based tag, whereas a per-process random tag won't.

iurii-ssv:
@greptile pls review

exporter/api/server.go
```diff
@@ -102,22 +130,17 @@ func (ws *wsServer) RegisterHandler(name, endPoint string, handler func(conn *we
...
 	}
 
 	otelHandler := otelhttp.NewHandler(http.HandlerFunc(wrappedHandler), name)
 	ws.router.Handle(endPoint, otelHandler)
+	ws.endpoints = append(ws.endpoints, endPoint)
 }
 
 // handleQuery receives query message and respond async
 func (ws *wsServer) handleQuery(conn *websocket.Conn) {
+	defer func() { _ = conn.Close() }()
```
julienharbulot: This closes the websocket once `handleQuery` returns, but should the handler also return when `ws.ctx` is canceled? Right now shutdown only logs at the ctx check below and then blocks in `ReadJSON` again. `http.Server.Shutdown` does not close upgraded websocket connections, and query connections are not wrapped in `conn` like the stream path, so an idle `/query` websocket can survive ctx-driven shutdown and leak the handler goroutine/connection. The stream path is ctx-aware via `conn.Close`: query should probably close/return on server ctx cancellation too?

iurii-ssv: Good catch — this was real. Moved the cleanup up into `RegisterHandler`'s wrapper instead of patching `handleQuery` directly so every registered endpoint inherits the same shutdown semantics: a watcher goroutine closes the upgraded `*websocket.Conn` on `ws.ctx.Done()`, unblocking any in-flight `ReadJSON` (or future blocking I/O) so the handler returns. Pinned in `TestHandleQuery_ServerCtxCancelClosesIdleConn`.

iurii-ssv:
@greptile pls re-review

iurii-ssv:
@greptile pls re-review

julienharbulot (dismissed):
LGTM

exporter/api/ws_session.go
```diff
@@ -0,0 +1,247 @@
...
+
+// isCloseError determines whether the given error is of CloseError type
+func isCloseError(err error) bool {
+	var closeError *websocket.CloseError
+	return errors.As(err, &closeError)
+}
+
+// pingMsg construct a ping message
+func pingMsg(cid string) []byte {
+	return []byte{cid[0], cid[1], cid[3], cid[4]}
```
momosh-ssv, line 251: Might be worth a length guard here — `pingMsg` indexes `cid[0..4]`, so it panics if `cid` is shorter than 5 chars. Production seems safe, and the new session test uses `withPing=false`, but any future `withPing=true` session constructed with a short id would panic on the first ping. 
A `if len(cid) < 5 { return []byte(cid) }` guard would remove the trap.

iurii-ssv: Good catch — fixed in [6bdf438](https://github.com/ssvlabs/ssv/pull/2819/commits/6bdf43820abe194004154bbb91da2ca68611a6cb), also added a TestPingMsg regression test pinning no-panic behavior across empty / short / boundary / realistic ids.

julienharbulot (approved):

momosh-ssv (approved):

PR #2820, opened 2026-04-20
operator/validator: reduce router fan-out

nkryuchkov:
## Summary

Addresses `ssvlabs/ssv-node-board#728`.

Reduces the validator network router fan-out from `2048` to `16` in `operator/validator`.

## Why

The task called out excessive contention from spawning `2048` goroutines that all read from the same router channel. That fan-out creates unnecessary scheduling and channel contention overhead.

I benchmarked the router path with different fan-out values using `BenchmarkRouterFanout` and compared repeated runs for the current pod shape. The relevant production-like signal was a node running with:

- `cpu.max = 360000 100000`
- `go_sched_gomaxprocs_threads = 4`

For that shape, `16` had the best median throughput among the tested candidates.

## Changes

- set `operator/validator` router fan-out to `16`
- add a short code comment documenting why `16` was chosen
- add `BenchmarkRouterFanout` to make future tuning measurable

## Validation

- `go test ./operator/validator/...`
- repeated benchmark runs with `go test ./operator/validator -run '^$' -bench '^BenchmarkRouterFanout' ...`


nkryuchkov:
@greptileai review it again

nkryuchkov:
@greptileai review it again

operator/validator/controller_bench_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	logger := zap.NewNop()
+	ctx, cancel := context.WithCancel(context.Background())
+	state := &benchmarkState{}
+
+	ctrl := &Controller{
+		logger:        logger,
+		ctx:           ctx,
+		networkConfig: networkconfig.TestNetwork,
+		validatorsMap: validators.New(ctx),
```
olegshmuelov: issue (non-blocking): Bench covers the exporter fallback only.

`validatorsMap` is empty and `ExporterOptions.Enabled=true` (L83), so nothing routes through `Validator.EnqueueMessage` / `Committee.EnqueueMessage` — the production hot path, including the per-committee write lock.

Populate validators/committees and route through those branches before trusting "40 is best."

operator/validator/controller_bench_test.go
```diff
@@ -0,0 +1,105 @@
...
+		messageRouter: newMessageRouter(logger),
+		messageWorker: worker.NewWorker(logger, &worker.Config{
+			Ctx:          ctx,
+			WorkersCount: benchmarkWorkerCount,
+			Buffer:       1 << 20,
```
olegshmuelov: suggestion (non-blocking): Bench diverges from production downstream.

Buffer is `1<<20` vs the prod default `65536`, and the handler set at L87 is a counter instead of `handleWorkerMessages`. That masks any worker-side backpressure a lower router fan-out would hit first in prod.

Match the prod buffer and report drops alongside `msgs/s`.

```go
messageWorker: worker.NewWorker(logger, &worker.Config{
    Ctx:          ctx,
    WorkersCount: benchmarkWorkerCount,
    Buffer:       65536, // match production MsgWorkerBufferSize
}),
```

operator/validator/controller.go
```diff
@@ -53,7 +53,10 @@ import (
 //go:generate go tool -modfile=../../tool.mod mockgen -package=mocks -destination=./mocks/controller.go -source=./controller.go
 
 const (
-	networkRouterConcurrency = 2048
+	// 40 was the best median fan-out in BenchmarkRouterFanout for the current
+	// pod shape, where the Go runtime rounds the 3.6 CPU cgroup quota to
+	// GOMAXPROCS=4.
+	networkRouterConcurrency = 40
```
olegshmuelov: nit: `WorkersCount` and `QueueBufferSize` are yaml/env-tunable; this isn't.

Consider promoting it to a `MsgRouterConcurrency` field on `ControllerOptions` (default = `networkRouterConcurrency`) so a different pod shape doesn't need a new release.

olegshmuelov (commented):
issue (blocking): No metric for router drops or channel fill — flagging on [`operator/validator/router.go:25-33`](https://github.com/ssvlabs/ssv/blob/3d29c7a9dfa97fabc1502e463fb68b4fdc0b4ab4/operator/validator/router.go#L25-L33) (file is outside this PR's diff).

The only overload signal today is the warn log in `Route()`. Add a dropped-message counter and a `len(r.ch)` gauge before rolling out cluster-wide — otherwise we can't verify the new fan-out or catch a bad pod shape.

```go
default:
    routerDroppedTotal.Inc()
    r.logger.Warn("message router buffer is full, dropping message")
```

also one nit: title says 48, constant is 40 — worth retitling/squashing so `git log` matches.

nkryuchkov:
@greptileai review this again

operator/validator/router.go
```diff
@@ -22,16 +22,36 @@ type messageRouter struct {
...
+		return msg, true
+	}
+}
+
+func (r *messageRouter) Len() int {
+	return len(r.ch)
+}
+
+func (r *messageRouter) recordBufferFill(ctx context.Context) {
+	routerBufferFillGauge.Record(ctx, int64(len(r.ch)))
```
momosh-ssv, line 67: Could this lead to claw-back of the contention we just reduced? `routerBufferFillGauge.Record` fires on every message in `Route()` (success + drop) and `Receive()` — roughly two sync-gauge calls per routed message on the hot path. Might be worth switching to an async observable gauge polled on the exporter tick, or at least sampling every Nth call.

operator/validator/controller.go
```diff
@@ -86,9 +87,10 @@ type ControllerOptions struct {
 	ProposerDelay                  time.Duration
 
 	// worker flags
-	WorkersCount    int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
-	QueueBufferSize int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
-	GasLimit        uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	WorkersCount         int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
+	QueueBufferSize      int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
+	MsgRouterConcurrency int    `yaml:"MsgRouterConcurrency" env:"MSG_ROUTER_CONCURRENCY" env-default:"16" env-description:"Number of goroutines draining the network message router"`
```
momosh-ssv: Any reason we're not collapsing the default to a single source? `defaultNetworkRouterConcurrency = 16` and `env-default:"16"` encode the same number twice. Since cleanenv always populates the tag, the `<= 0` fallback only kicks in for test paths — seems easy for the two 16s to drift on a future tuning pass.

network/network.go
```diff
@@ -18,8 +18,8 @@ type DecodedSSVMessage interface {
 
 // MessageRouter is accepting network messages and route them to the corresponding (internal) components
 type MessageRouter interface {
-	// Route routes the given message, this function MUST NOT block
-	Route(ctx context.Context, message DecodedSSVMessage)
+	// Route routes the given message without blocking and reports whether it was accepted.
+	Route(ctx context.Context, message DecodedSSVMessage) bool
 }
```
momosh-ssv, line 23: What do you think about reverting the `bool` return? The only production caller (`network/p2p/p2p_pubsub.go:245`) discards it, and drop accounting (counter + warn log) already happens inside `Route()`. Seems the interface churn only serves the benchmark, which could read `messageRouter.Route` directly since it lives in the same package.

operator/validator/controller.go
```diff
@@ -53,7 +53,8 @@ import (
 //go:generate go tool -modfile=../../tool.mod mockgen -package=mocks -destination=./mocks/controller.go -source=./controller.go
 
 const (
-	networkRouterConcurrency = 2048
+	// Chosen from BenchmarkRouterFanout to keep fan-out low on one shared channel.
```
momosh-ssv: Might be worth keeping the pod-shape rationale in the comment. The previous revision called out GOMAXPROCS=4 from the 3.6 CPU cgroup; the current phrasing no longer tells a future tuner when 16 stops being the right value.

operator/validator/router.go
```diff
@@ -22,16 +22,36 @@ type messageRouter struct {
 	ch     chan network.DecodedSSVMessage
 }
 
-func (r *messageRouter) Route(ctx context.Context, message network.DecodedSSVMessage) {
+func (r *messageRouter) Route(ctx context.Context, message network.DecodedSSVMessage) bool {
 	select {
 	case <-ctx.Done():
 		r.logger.Warn("context canceled, dropping message")
```
momosh-ssv: Seems that on shutdown under load, each buffered producer hits this branch and logs at Warn. Pre-existing behaviour, but since the file is in scope — maybe drop this to Debug and reserve Warn for the buffer-full branch.

operator/validator/controller.go
```diff
@@ -301,42 +309,40 @@ func (c *Controller) GetValidatorStats() (uint64, uint64, uint64, error) {
 func (c *Controller) handleRouterMessages() {
 	ctx, cancel := context.WithCancel(c.ctx)
```
momosh-ssv: Any reason we're deriving ctx here? `cancel` only runs via `defer`, so this is equivalent to passing `c.ctx` straight into `Receive`.

operator/validator/controller.go
```diff
@@ -520,7 +516,7 @@ func (c *Controller) InitValidators() ([]*validator.Validator, error) {
 // StartNetworkHandlers init msg worker that handles network messages
 func (c *Controller) StartNetworkHandlers() {
 	c.network.UseMessageRouter(c.messageRouter)
-	for i := 0; i < networkRouterConcurrency; i++ {
+	for i := 0; i < c.msgRouterConcurrency; i++ {
```
momosh-ssv, line 522: Should we consider guarding against `msgRouterConcurrency <= 0` here? Collapsing to the env-default tag fixed the duplicate-`16` drift, but it also removed the only safety net — if a future caller (or test) constructs `ControllerOptions{}` without cleanenv populating the tag, this loop spawns zero workers and the router silently stalls. A one-line `max(c.msgRouterConcurrency, 16)` (or a Fatal on `<= 0`) would keep a single source of truth without the foot-gun.

nkryuchkov: It looks like the opposite of what you requested in https://github.com/ssvlabs/ssv/pull/2820#discussion_r3122927284

momosh-ssv: Yeah, agreed. 
My original ask was to collapse the two `16`s into one source — either drop the `constant + <= 0` fallback and trust the `env-default` tag, or drop the tag and use the constant. 
This change just routes through the field, but both defaults are still there, so the drift risk I flagged I think still isn't really addressed.

iurii-ssv, line 522: **Blocking — ties into the unresolved thread above.**

With the constant gone and the default living only on the `env-default:"16"` tag, this loop spawns zero goroutines for any caller that constructs `ControllerOptions{}` programmatically without going through `cleanenv` (e.g. the existing tests in `eth/eventsyncer/event_syncer_test.go` and `eth/ethtest/utils_test.go`). The router then silently accumulates messages until the buffer fills and starts dropping everything.

Production is fine because `cli/operator/node.go` reads through `cleanenv.ReadConfig`, but a future test or integration site that exercises `StartNetworkHandlers()` will hit a silent stall.

Two cheap fixes that preserve the single-source-of-truth goal:
- fail fast: `if c.msgRouterConcurrency <= 0 { c.logger.Fatal("msgRouterConcurrency must be > 0") }`
- or normalize once in `NewController`: `if options.MsgRouterConcurrency <= 0 { options.MsgRouterConcurrency = 16 }`

I'd lean toward fail-fast since silent stalls in a message router are nasty to diagnose.

operator/validator/controller.go
```diff
@@ -86,9 +82,12 @@ type ControllerOptions struct {
...
 
 	// worker flags
-	WorkersCount    int    `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
-	QueueBufferSize int    `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
-	GasLimit        uint64 `yaml:"ExperimentalGasLimit" env:"EXPERIMENTAL_GAS_LIMIT" env-description:"Gas limit for MEV block proposals (must match across committee, otherwise MEV fails). Do not change unless you know what you're doing"`
+	WorkersCount    int `yaml:"MsgWorkersCount" env:"MSG_WORKERS_COUNT" env-default:"256" env-description:"Number of message processing workers"`
+	QueueBufferSize int `yaml:"MsgWorkerBufferSize" env:"MSG_WORKER_BUFFER_SIZE" env-default:"65536" env-description:"Size of message worker queue buffer"`
+	// Chosen from BenchmarkRouterFanout for the current pod shape, where the
+	// 3.6 CPU cgroup quota rounds to GOMAXPROCS=4.
+	MsgRouterConcurrency int    `yaml:"MsgRouterConcurrency" env:"MSG_ROUTER_CONCURRENCY" env-default:"16" env-description:"Number of goroutines draining the network message router"`
```
iurii-ssv: nit: the field comment captures the GOMAXPROCS=4 rationale (good), but the `env-description` is what surfaces in `--help` and is the only thing an operator on different hardware will read. Consider expanding it to something like `"...; 16 is tuned for ~4 vCPU pods, override on larger hosts"` so a 16-vCPU bare-metal operator doesn't silently inherit a pessimal value.

operator/validator/router.go
```diff
@@ -18,20 +21,48 @@ func newMessageRouter(logger *zap.Logger) *messageRouter {
...
 	case <-ctx.Done():
-		r.logger.Warn("context canceled, dropping message")
+		r.logger.Debug("context canceled, dropping message")
+		return false
 	case r.ch <- message:
+		r.recordBufferFill(ctx, false)
+		return true
 	default:
+		routerDroppedCounter.Add(ctx, 1)
+		r.recordBufferFill(ctx, true)
```
iurii-ssv: nit: forcing the gauge record on every drop (vs the sampled path on success/receive) means during a sustained drop storm the gauge gets hammered while always reading `bufSize`. The counter (line 42) already tells you the drop rate; the gauge value adds no new info while the buffer is full. Consider letting the sampled path handle this — drop the `force=true` and just call `recordBufferFill(ctx, false)`.

operator/validator/controller_bench_test.go
```diff
@@ -0,0 +1,262 @@
...
+					}
+				}()
+			}
+
+			producers.Wait()
+			require.Eventually(b, func() bool {
+				return state.processed.Load() == int64(b.N)
+			}, 30*time.Second, time.Millisecond)
+
+			elapsed := time.Since(start)
```
iurii-ssv, line 83: nit: `elapsed` is captured *after* `require.Eventually` returns, so the throughput metrics (`router_msgs/s`, `processed_msgs/s`) are correctly end-to-end. But `b.ResetTimer`/`b.StopTimer` are also active, which means `ns/op` will reflect a different number than the custom `_msgs/s` metrics. That's fine for comparing fan-outs via `b.ReportMetric`, but a future reader who eyeballs `ns/op` will get a misleading number. A one-line comment near `b.ResetTimer()` saying "ns/op is not the throughput signal — use the reported _msgs/s metrics" would save someone confusion later.

iurii-ssv (approved):
LGTM

momosh-ssv (approved):

PR #2823, opened 2026-04-20
Fail p2p topic subscription tests instead of hanging

julienharbulot:
T-ssv-045 — Infinite Polling Loops Without Timeout

The original ticket flagged unbounded sleep-polling in test code that can leave CI hanging indefinitely when the expected condition never happens.

Changes in this PR:
- replace the unbounded topic-peer polling loop in `network/p2p/p2p_test.go` with a bounded `require.Eventually` assertion
- keep the test behavior the same on success while failing fast with a clear timeout instead of hanging forever
- leave `exporter/api/server_test.go` unchanged because it already uses a timeout-based polling helper

Validation:
- `go test ./exporter/api`
- `go test ./network/p2p -run '^TestStartReturnsErrorWhenAlreadyStarted$'`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run ./network/p2p`

Fixes - Fixes ssvlabs/ssv-node-board#934


iurii-ssv (dismissed):

ljuba-ssv (dismissed):

julienharbulot:
Addressed the Greptile helper-traceability suggestion in 11e40242a by adding `t.Helper()` to `createNetworkAndSubscribe`, so assertion failures now point back to the calling test.

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2826, opened 2026-04-21
qbft: clarify & simplify QBFT Instance lifecycle

iurii-ssv:
While working on https://github.com/ssvlabs/ssv/pull/2771 I've noticed the way we manage QBFT `Instance` is way overcomplicated (vs what it should be) + we can simplify the lifecycle of `RoundTimer` (to also make it more precise) - this PR does just that.

Related to https://github.com/ssvlabs/ssv/issues/2766 / https://github.com/ssvlabs/ssv-node-board/issues/764


protocol/v2/qbft/instance/instance.go
```diff
@@ -146,6 +149,10 @@ func (i *Instance) Start(
 	span.SetStatus(codes.Ok, "")
 }
 
+func (i *Instance) Stop() {
+	i.RoundTimer.Stop()
+}
```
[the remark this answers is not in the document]
iurii-ssv: In production this isn't supposed to happen (JSON-decoded instances are only used in spec-tests),

for consistency with those usages in `proposal.go`, `timeout.go` I've removed the nil-checks there too since those are no longer needed/relevant.

protocol/v2/qbft/controller/decided.go
```diff
@@ -2,61 +2,90 @@ package controller
...
-	} else if decided, _ := inst.IsDecided(); !decided {
+		c.RecentInstances.addNewInstance(inst)
+		if msg.QBFTMessage.Height > c.CurrentInstanceHeight {
+			// If the new instance turned out to be "from the future", it means other nodes in our cluster are
+			// already ahead working with an instance we haven't gotten to. We accept it as our current (latest
+			// known) height.
+			// NOTE: we could also stop & trim older instances that became irrelevant now - but spectests do not
+			// expect this, and we are doing it in `Controller.StartNewInstance` periodically anyway.
+			c.CurrentInstanceHeight = msg.QBFTMessage.Height
+		}
```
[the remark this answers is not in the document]
iurii-ssv: While this is possible, in practice there is an additional `didDecideCorrectly` guard that executes after to "catch" such incorrect/duplicate `Signal "newly decided" to the caller.` cases ... so it's outside of the scope of this PR to improve the handling for it, and I don't see much reason to create a task for it either since we are unlikely to have to rewrite this code much further to actually stumble into this being a problem.

iurii-ssv:
@greptile pls review

protocol/v2/qbft/instance/marshalutils.go
olegshmuelov, line 31: suggestion (non-blocking): document the `NewInstance`-first precondition.

  `roundTimer` is unexported, so `Decode` doesn't restore it — callers must `NewInstance(..., roundTimerF)` first. All current callers do, but a future `&Instance{}; inst.Decode(...)` would nil-panic on `MarkDecided`/`MarkIrrelevant`.

  ```go
  // UnmarshalJSON overlays spec-test state; the receiver must already be
  // constructed via NewInstance — roundTimer is not restored.
  func (i *Instance) UnmarshalJSON(data []byte) error {
  ```

protocol/v2/qbft/instance/instance.go
```diff
@@ -283,5 +309,5 @@ func (i *Instance) bumpToRound(round specqbft.Round) {
 
 // CanProcessMessages will return true if instance can process messages
 func (i *Instance) CanProcessMessages() bool {
```
olegshmuelov, line 313: question: should `CanProcessMessages` also reject `State.Decided`?

  `MarkDecided` says it renders the instance "read-only", but `CanProcessMessages` only gates on `markedIrrelevant`. The controller already filters decided messages upstream, so the live path is fine — but a direct `ProcessMsg` on a decided instance (spec-test pattern) still dispatches `uponProposal`/`uponRoundChange` and mutates state. Either add the check here, or soften the `MarkDecided` docstring.

  ```go
  func (i *Instance) CanProcessMessages() bool {
      return !i.markedIrrelevant && !i.State.Decided && i.State.Round < i.config.GetCutOffRound()
  }
  ```

iurii-ssv: The flow is: 
- an `Instance` can become **decided**, which basically means qbft has finished; and that's what "read-only" refers to (no QBFT-related progress is done afterward)
- but even after an `Instance` has **decided**, we want to record all the messages relevant to it (to log them, and whatnot) - which is why `CanProcessMessages` DOESN'T drop messages when  `State.Decided == true` (we accept and store the message, but don't do any QBFT-related state updates)

I guess we can rephrase "read-only" to be something like "no QBFT-related progress is done afterward" to clarify it - applied in https://github.com/ssvlabs/ssv/pull/2826/changes/c15c584f79df8b07702eeb0d019ad8db982c1a0d

iurii-ssv:
> protocol/v2/qbft/roundtimer/timer.go – Stop() should hold t.mtx when accessing t.timer for full correctness.

Addressed that in https://github.com/ssvlabs/ssv/pull/2826/changes/ea7e4342c67f8e4e6cbd6342d610e37a27139e5b, @greptile pls re-review

protocol/v2/qbft/controller/controller.go
```diff
@@ -77,36 +84,61 @@ func (c *Controller) StartNewInstance(
...
 
 	return newInstance, nil
 }
 
-func (c *Controller) forceStopAllInstanceExceptCurrent() {
-	for _, i := range c.StoredInstances {
-		if i.State.Height != c.Height {
-			i.ForceStop()
-		}
+// markInstancesIrrelevant marks all recent instances as irrelevant.
```
olegshmuelov: nit: rename left the doc comment on the wrong name . The NOTE at L106-L107 also still references `markOlderInstancesIrrelevant`, worth fixing in the same change since that NOTE documents the load-bearing ordering hazard.

  ```suggestion
  // markRecentInstancesIrrelevant marks all recent instances as irrelevant.
  ```

  And L106-L107:

  ```go
  // NOTE: the c.markRecentInstancesIrrelevant() call comes first because `addNewInstance` might evict an instance
  //       from c.RecentInstances, meaning c.markRecentInstancesIrrelevant() will miss it if called after.
  ```

olegshmuelov (approved):
LGTM

ljuba-ssv (approved):

PR #2833, opened 2026-04-27
Replace flaky test sleeps with explicit waits - T-ssv-047

julienharbulot:
T-ssv-047

Issue summary:
This ticket tracks flaky and unnecessarily slow tests caused by sleep-based synchronization in the `ssv` test suite. The main offenders were fixed-duration waits in the ETH E2E and P2P validation tests, plus randomized sleep-based interleaving in the hashmap concurrency test.

Changes:
- replaced fixed sleeps in `eth/ethtest/eth_e2e_test.go` with `require.Eventually` and `require.Never` checks against the actual state transitions under test
- replaced the fixed post-broadcast wait in `network/p2p/p2p_validation_test.go` with condition-based polling for validator counters and peer-score ordering
- removed randomized sleeps from `utils/hashmap/hashmap_test.go` and used explicit concurrency coordination instead

Validation:
- `go test ./utils/hashmap -run TestIssue1682 -count=1`
- `go test ./eth/ethtest -run TestEthExecLayer -count=1`
- `go test ./network/p2p -run TestGetMaxPeers -count=1`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -v ./network/p2p ./eth/ethtest ./utils/hashmap`
- `go test ./network/p2p -run TestP2pNetwork_MessageValidation -count=1 -v -timeout 2m` still times out locally in mDNS peer discovery during virtual network setup, before the edited assertions run

Fixes - Fixes ssvlabs/ssv-node-board#936

### DIFF
--- a/eth/ethtest/eth_e2e_test.go
+++ b/eth/ethtest/eth_e2e_test.go
@@ -229,7 +229,16 @@ func TestEthExecLayer(t *testing.T) {
 
 		// Step 4 Liquidate Cluster
 		{
-			taskExecutor.EXPECT().LiquidateCluster(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+			liquidatedDone := make(chan struct{}, 1)
+			taskExecutor.EXPECT().LiquidateCluster(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
+				func(ethcommon.Address, []uint64, []*ssvtypes.SSVShare) error {
+					select {
+					case liquidatedDone <- struct{}{}:
+					default:
+					}
+					return nil
+				},
+			).AnyTimes()
 
 			clusterLiquidate := NewTestClusterLiquidatedInput(common)
 			clusterLiquidate.prepare([]*ClusterLiquidatedEventInput{
@@ -244,18 +253,11 @@ func TestEthExecLayer(t *testing.T) {
 			testEnv.CloseFollowDistance(&blockNum)
 
 			clusterID := ssvtypes.ComputeClusterIDHash(testAddrAlice, []uint64{1, 2, 3, 4})
-			require.Eventually(t, func() bool {
-				shares := nodeStorage.Shares().List(nil, registrystorage.ByClusterIDHash(clusterID))
-				if len(shares) != 5 {
-					return false
-				}
-				for _, s := range shares {
-					if !s.Liquidated {
-						return false
-					}
-				}
-				return true
-			}, asyncWaitTimeout, asyncPollTick)
+			select {
+			case <-liquidatedDone:
+			case <-time.After(asyncWaitTimeout):
+				t.Fatal("timed out waiting for cluster liquidation task")
+			}
 
 			shares := nodeStorage.Shares().List(nil, registrystorage.ByClusterIDHash(clusterID))
 			require.NotEmpty(t, shares)
@@ -268,7 +270,16 @@ func TestEthExecLayer(t *testing.T) {
 
 		// Step 5 Reactivate Cluster
 		{
-			taskExecutor.EXPECT().ReactivateCluster(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+			reactivatedDone := make(chan struct{}, 1)
+			taskExecutor.EXPECT().ReactivateCluster(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
+				func(ethcommon.Address, []uint64, []*ssvtypes.SSVShare) error {
+					select {
+					case reactivatedDone <- struct{}{}:
+					default:
+					}
+					return nil
+				},
+			).AnyTimes()
 
 			clusterID := ssvtypes.ComputeClusterIDHash(testAddrAlice, []uint64{1, 2, 3, 4})
 
@@ -292,18 +303,11 @@ func TestEthExecLayer(t *testing.T) {
 			clusterReactivated.produce()
 			testEnv.CloseFollowDistance(&blockNum)
 
-			require.Eventually(t, func() bool {
-				clusterShares := nodeStorage.Shares().List(nil, registrystorage.ByClusterIDHash(clusterID))
-				if len(clusterShares) != 5 {
-					return false
-				}
-				for _, s := range clusterShares {
-					if s.Liquidated {
-						return false
-					}
-				}
-				return true
-			}, asyncWaitTimeout, asyncPollTick)
+			select {
+			case <-reactivatedDone:
+			case <-time.After(asyncWaitTimeout):
+				t.Fatal("timed out waiting for cluster reactivation task")
+			}
 
 			shares = nodeStorage.Shares().List(nil, registrystorage.ByClusterIDHash(clusterID))
 			require.NotEmpty(t, shares)
--- a/eth/ethtest/eth_e2e_test.go
+++ b/eth/ethtest/eth_e2e_test.go
@@ -1,6 +1,7 @@
 package ethtest
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"math/big"
@@ -27,6 +28,11 @@ var (
 
 // E2E tests for ETH package
 func TestEthExecLayer(t *testing.T) {
+	const (
+		asyncWaitTimeout = 15 * time.Second
+		asyncPollTick    = 50 * time.Millisecond
+	)
+
 	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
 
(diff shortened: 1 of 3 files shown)


network/p2p/p2p_validation_test.go
```diff
@@ -272,24 +262,51 @@ func TestP2pNetwork_MessageValidation(t *testing.T) {
...
-		for i, validOrder := range validOrders {
-			valid := true
-			for j, peer := range peers {
-				if peer.index != validOrder[j] {
-					valid = false
-					break
-				}
-			}
-			if valid {
+		require.True(t, peerScoresMatchAnyOrder(peers, validOrders), "node %d, peers %v", node.Index, peers)
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 7496ba550 by capturing the passing peer-score snapshot inside `require.Eventually` and reusing that same snapshot for the final assertion instead of re-reading asynchronously updated scores.

utils/hashmap/hashmap_test.go
```diff
@@ -490,21 +491,30 @@ func TestIssue1682(t *testing.T) {
...
-						time.Sleep(time.Duration(randSeed().Intn(200)) * time.Millisecond)
+						attempted.Done()
 						if !found {
+							<-releaseSubscribed
+							runtime.Gosched()
 							m.Set(cmtID, validatorStatusSubscribed)
 						}
-					}()
+					}(cmtID)
 				}
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 7496ba550 by adding a clarifying comment that `attempted` tracks the post-`GetOrSet` barrier while `wg` still tracks full goroutine completion.

eth/ethtest/eth_e2e_test.go
```diff
@@ -125,23 +131,14 @@ func TestEthExecLayer(t *testing.T) {
...
-					return
-				case <-stopChan:
-					return
-				default:
-					time.Sleep(100 * time.Millisecond)
-				}
-			}
+		defer func() {
+			syncCancel()
+			require.NoError(t, <-syncOngoingErrCh)
```
momosh-ssv: Should we use `assert.NoError` here instead of `require.NoError`? If an earlier `require.*` in the subtest already triggered `t.FailNow()`, this deferred require can fire a second failure that masks the original cause.

eth/ethtest/eth_e2e_test.go
```diff
@@ -228,7 +229,16 @@ func TestEthExecLayer(t *testing.T) {
...
+			liquidatedDone := make(chan struct{}, 1)
+			taskExecutor.EXPECT().LiquidateCluster(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(
+				func(ethcommon.Address, []uint64, []*ssvtypes.SSVShare) error {
+					select {
+					case liquidatedDone <- struct{}{}:
+					default:
+					}
+					return nil
+				},
+			).AnyTimes()
```
momosh-ssv: What do you think about `.Times(1)` here (and at the `ReactivateCluster` mock below) instead of `AnyTimes()`? The buffer-1 channel + `default` drop silently discards extra invocations — works today since the handler runs once per event, but the combination is subtle. If exactly one call is expected, `.Times(1)` makes that explicit.

eth/ethtest/eth_e2e_test.go
```diff
@@ -319,14 +343,14 @@ func TestEthExecLayer(t *testing.T) {
 			setFeeRecipient.produce()
 			testEnv.CloseFollowDistance(&blockNum)
 
-			// Wait until the state is changed
-			time.Sleep(time.Millisecond * 5000)
+			require.Eventually(t, func() bool {
+				feeRecipient, err := nodeStorage.GetFeeRecipient(testAddrAlice)
+				return err == nil && bytes.Equal(testAddrBob.Bytes(), feeRecipient[:])
+			}, asyncWaitTimeout, asyncPollTick)
```
momosh-ssv, line 353: Might be worth capturing `feeRecipient` inside the `Eventually` closure and asserting on the captured value? The re-read on the next line is the same TOCTOU shape that was fixed in `p2p_validation_test.go` — benign here since no further events are injected, but inconsistent. Same applies to Step 1 (lines 156–163).

utils/hashmap/hashmap_test.go
```diff
@@ -490,21 +491,31 @@ func TestIssue1682(t *testing.T) {
...
-					go func() {
+					attempted.Add(1)
+					go func(cmtID string) {
 						defer wg.Done()
-						time.Sleep(time.Duration(randSeed().Intn(2000)) * time.Millisecond)
+						<-start
 						_, found := m.GetOrSet(cmtID, validatorStatusSubscribing)
-						time.Sleep(time.Duration(randSeed().Intn(200)) * time.Millisecond)
+						// `attempted` tracks the "reached GetOrSet" barrier, while `wg` still tracks full goroutine completion.
+						attempted.Done()
```
momosh-ssv: Worth considering deferring `attempted.Done()` like `wg.Done()`? If `GetOrSet` ever panics (theoretical, but possible under `-race` with corrupt state), `attempted.Wait()` will hang forever rather than fail the test.

eth/ethtest/eth_e2e_test.go
julienharbulot: Addressed in 2f4dddfda by switching the deferred sync check to `assert.NoError`, so a prior `require.*` failure in the subtest is not masked by a second `FailNow` from the cleanup path.

eth/ethtest/eth_e2e_test.go
julienharbulot: Addressed in 2f4dddfda by capturing the successful values inside the `Eventually` closures and asserting on those captured snapshots instead of re-reading immediately afterward.

eth/ethtest/eth_e2e_test.go
julienharbulot: Addressed in 2f4dddfda by tightening both cluster-task mocks to `.Times(1)` and removing the silent default-drop path, so the expected single invocation is explicit.

utils/hashmap/hashmap_test.go
julienharbulot: Addressed in 2f4dddfda by moving `attempted.Done()` into a defer around the `GetOrSet` call, preserving the post-`GetOrSet` barrier while avoiding a hang if that call ever panics.

momosh-ssv (approved):
really a gj!

iurii-ssv:
@greptile pls re-review

iurii-ssv (approved):
LGTM, got a bunch of AI-based suggestions that seem relevant (could be a separate follow-up PR):

### 1. Failure diagnostics regression (medium)

`require.Eventually` failures show only "Condition never satisfied" — the structured error messages from the old code are lost. When this fails on CI months from now, the operator will only see a generic timeout.

- [eth_e2e_test.go: Step 1](https://github.com/ssvlabs/ssv/pull/2833/files#diff-eth/ethtest/eth_e2e_test.goR158-R166) — single closure tests `err`, `nextNonce`, and `len(shares)` — three distinct failure modes collapsed into one.
- [p2p_validation_test.go: counter check](https://github.com/ssvlabs/ssv/pull/2833/files#diff-network/p2p/p2p_validation_test.goR197-R217) — the old loop produced `"node %d accepted %d messages (expected %d)"`; the new closure just returns `false`.

**Suggestion:** Use `require.EventuallyWithT(t, func(c *assert.CollectT) {...}, ...)` — it lets you write per-condition `assert.Equal` calls inside the closure, and on timeout it surfaces *which* assertion last failed.

### 2. Hashmap test — narrowed concurrency coverage (low/medium)

The original [TestIssue1682](https://github.com/ssvlabs/ssv/blob/cf4462108/utils/hashmap/hashmap_test.go#L497-L507) had `GetOrSet`, `Set`, and the main goroutine's `Range` workload all racing freely (random sleeps spread the three operations across overlapping windows).

The new structure (in the PR diff) serializes phases:
1. close(start) → all goroutines call `GetOrSet` concurrently
2. `attempted.Wait()` blocks main until *all* `GetOrSet` calls have returned
3. close(releaseSubscribed) → winners proceed to `Set`
4. *Then* the `Range` loop starts in the main goroutine

This means **`Range` no longer interleaves with `GetOrSet`** — only with `Set`. The original bug being guarded against (silent entry loss in `cornelk/hashmap`, now `sync.Map`-backed) is still covered, but the test is strictly less aggressive.

**Suggestion:** Consider moving the `done`/`Range` loop to start *before* `attempted.Wait()` and `close(releaseSubscribed)`. That preserves `Range`-vs-`GetOrSet` interleaving while still removing the sleep-driven flakiness. The barrier ordering (`start` → `attempted` → `releaseSubscribed`) doesn't depend on when Range starts.

### 3. Deferred table rendering shows nothing on `Eventually` timeout (low)

In [p2p_validation_test.go](https://github.com/ssvlabs/ssv/pull/2833/files#diff-network/p2p/p2p_validation_test.goR237-R245), `peers` is populated *only inside* the `Eventually` closure. If `Eventually` times out, `peers` is `nil`, and the [deferred table](https://github.com/ssvlabs/ssv/pull/2833/files#diff-network/p2p/p2p_validation_test.goR251-R266) renders empty — losing the diagnostic value the table was added for.

**Suggestion:** After `Eventually`, do a best-effort `peers, _ = sortedPeerScores(node)` (overwriting only if currently empty) so the table always shows the last observed scores.

### 4. Redundant post-Eventually assertions (low)

Several blocks capture a value inside the closure and re-assert it after:

```go
require.Eventually(t, func() bool {
    feeRecipient, err := nodeStorage.GetFeeRecipient(testAddrAlice)
    if err != nil || !bytes.Equal(testAddrBob.Bytes(), feeRecipient[:]) {
        return false
    }
    updatedFeeRecipient = feeRecipient
    return true
}, ...)
require.Equal(t, testAddrBob.Bytes(), updatedFeeRecipient[:])
```

Same pattern at [Step 1's `syncedNonce`](https://github.com/ssvlabs/ssv/pull/2833/files#diff-eth/ethtest/eth_e2e_test.goR159-R168) and [Step 7's `updatedFeeRecipient`](https://github.com/ssvlabs/ssv/pull/2833/files#diff-eth/ethtest/eth_e2e_test.goR346-R355). The post-assertion is trivially true if `Eventually` succeeded — it's noise. `EventuallyWithT` would let you keep one clear `require.Equal` and drop the captures.


### 5. Minor — `runtime.Gosched()` is a hint, not a guarantee

The [`runtime.Gosched()` before `m.Set`](https://github.com/ssvlabs/ssv/pull/2833/files#diff-utils/hashmap/hashmap_test.goR518) is harmless, but shouldn't be relied on for interleaving. The test's correctness doesn't depend on it (the channel barriers do the real coordination), so leaving it is fine — just don't lean on it as semantic.

PR #2834, opened 2026-04-27
Improve beacon mock error-path coverage - T-ssv-050

julienharbulot:
T-ssv-050

Original ticket summary:
The shared beacon mock in `beacon/goclient/mocks/server.go` only returned happy-path JSON and panicked on unexpected requests, which left `429`, `500`, malformed JSON, and timeout handling effectively untested.

Changes:
- extended the shared beacon mock to support configurable status codes, headers, delayed responses, and deterministic error responses for unexpected paths
- added focused tests for the mock server itself and beacon client error paths covering `429`, `500`, malformed JSON, and timeout behavior

Fixes - Fixes ssvlabs/ssv-node-board#939

### DIFF
--- a/beacon/goclient/error_paths_test.go
+++ b/beacon/goclient/error_paths_test.go
@@ -1,6 +1,7 @@
 package goclient
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"testing"
@@ -18,10 +19,10 @@ func Test_specForClient_errorPaths(t *testing.T) {
 	t.Parallel()
 
 	tt := []struct {
-		name     string
-		response mocks.Response
-		wantErr  string
-		timeout  time.Duration
+		name           string
+		response       mocks.Response
+		wantErr        string
+		requestTimeout time.Duration
 	}{
 		{
 			name: "rate limited",
@@ -31,7 +32,6 @@ func Test_specForClient_errorPaths(t *testing.T) {
 				mocks.WithHeader("Retry-After", "1"),
 			),
 			wantErr: "GET failed with status 429",
-			timeout: 100 * time.Millisecond,
 		},
 		{
 			name: "internal server error",
@@ -40,7 +40,6 @@ func Test_specForClient_errorPaths(t *testing.T) {
 				mocks.WithStatusCode(http.StatusInternalServerError),
 			),
 			wantErr: "GET failed with status 500",
-			timeout: 100 * time.Millisecond,
 		},
 		{
 			name: "timeout",
@@ -48,8 +47,8 @@ func Test_specForClient_errorPaths(t *testing.T) {
 				nil,
 				mocks.WithDelay(200*time.Millisecond),
 			),
-			wantErr: "context deadline exceeded",
-			timeout: 50 * time.Millisecond,
+			wantErr:        "context deadline exceeded",
+			requestTimeout: 50 * time.Millisecond,
 		},
 	}
 
@@ -66,9 +65,16 @@ func Test_specForClient_errorPaths(t *testing.T) {
 			defer server.Close()
 
 			client := &GoClient{log: zap.NewNop()}
-			provider := newHTTPService(t, server.URL, tc.timeout)
+			provider := newHTTPService(t, server.URL)
 
-			_, err := client.specForClient(t.Context(), provider)
+			ctx := t.Context()
+			if tc.requestTimeout > 0 {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, tc.requestTimeout)
+				defer cancel()
+			}
+
+			_, err := client.specForClient(ctx, provider)
 			require.ErrorContains(t, err, tc.wantErr)
 		})
 	}
@@ -86,23 +92,26 @@ func Test_genesisForClient_malformedJSON(t *testing.T) {
 	defer server.Close()
 
 	client := &GoClient{log: zap.NewNop()}
-	provider := newHTTPService(t, server.URL, 100*time.Millisecond)
+	provider := newHTTPService(t, server.URL)
 
 	_, err := client.genesisForClient(t.Context(), provider)
 	require.ErrorContains(t, err, "invalid character")
 }
 
-func newHTTPService(t *testing.T, addr string, timeout time.Duration) *eth2clienthttp.Service {
+func newHTTPService(t *testing.T, addr string) *eth2clienthttp.Service {
 	t.Helper()
 
 	service, err := eth2clienthttp.New(
 		t.Context(),
 		eth2clienthttp.WithAddress(addr),
 		eth2clienthttp.WithLogLevel(zerolog.Disabled),
-		eth2clienthttp.WithTimeout(timeout),
+		eth2clienthttp.WithTimeout(time.Second),
 		eth2clienthttp.WithReducedMemoryUsage(true),
 	)
 	require.NoError(t, err)
 
-	return service.(*eth2clienthttp.Service)
+	provider := service.(*eth2clienthttp.Service)
+	require.Eventually(t, provider.IsActive, time.Second, 10*time.Millisecond)
+
+	return provider
 }
--- a/beacon/goclient/error_paths_test.go
+++ b/beacon/goclient/error_paths_test.go
@@ -101,7 +101,6 @@ func newHTTPService(t *testing.T, addr string, timeout time.Duration) *eth2clien
 		eth2clienthttp.WithLogLevel(zerolog.Disabled),
 		eth2clienthttp.WithTimeout(timeout),
 		eth2clienthttp.WithReducedMemoryUsage(true),
-		eth2clienthttp.WithAllowDelayedStart(true),
 	)
 	require.NoError(t, err)
 
--- a/beacon/goclient/error_paths_test.go
+++ b/beacon/goclient/error_paths_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	eth2api "github.com/attestantio/go-eth2-client/api"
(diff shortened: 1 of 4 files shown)


beacon/goclient/mocks/server.go
```diff
@@ -6,44 +6,142 @@ import (
...
+				return
 			}
 		}
 
-		w.Header().Set("Content-Type", "application/json")
-		if _, err := w.Write(resp); err != nil {
-			panic(fmt.Sprintf("got error writing response: %v", err))
+		if resp.Delay > 0 {
+			time.Sleep(resp.Delay)
 		}
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 409f98199 by making mock response delays respect request cancellation via `r.Context().Done()`, so timed-out requests no longer keep the handler sleeping until the full delay elapses.

beacon/goclient/mocks/server.go
```diff
@@ -6,44 +6,145 @@ import (
...
 	}))
 }
 
+func responseForPath(mockResponses map[string]json.RawMessage, path string) Response {
+	if resp, ok := mockResponses[path]; ok {
+		return NewResponse(resp)
+	}
+
+	return NewResponse(
+		json.RawMessage(fmt.Sprintf(`{"error":"unexpected request","path":%q}`, path)),
```
momosh-ssv: What do you think about marshalling this body via `json.Marshal` rather than `%q`? Seems that Go-quoting and JSON string escaping diverge on control bytes, so a path with awkward characters could produce invalid JSON despite the `application/json` header. Building the object with `json.Marshal(map[string]string{...})` keeps it correct for any path.

beacon/goclient/mocks/server.go
```diff
@@ -6,44 +6,145 @@ import (
...
+			w.Header().Add(key, value)
+		}
+	}
+
+	if w.Header().Get("Content-Type") == "" {
+		w.Header().Set("Content-Type", "application/json")
+	}
+
+	statusCode := resp.StatusCode
+	if statusCode == 0 {
```
momosh-ssv, line 129: Should we consider treating a zero `StatusCode` as a misuse rather than silently defaulting to 200? I fear a handler that mistakenly returns `Response{}` with `nil` error surfaces as a successful empty-body 200 to the client, quietly masking the test bug. A panic on zero — or requiring `NewResponse` for construction — would make accidental zero-values visible.

beacon/goclient/mocks/server.go
```diff
@@ -6,44 +6,145 @@ import (
...
+
+	_, _ = w.Write(resp.Body)
+}
+
 func ServerResponses() map[string]json.RawMessage {
 	var responses map[string]json.RawMessage
+
 	f, err := os.Open("./mocks/mock-beacon-responses.json")
+	if err != nil {
+		f, err = os.Open("./mock-beacon-responses.json")
```
momosh-ssv: Could we anchor this file lookup to the source directory via `runtime.Caller`/`filepath.Join` instead of relying on cwd? Seems that the `./mocks/...` → `./...` fallback hides a real fragility — any future test invoked from a third location panics with a confusing `os.Open` error. Anchoring to the package dir makes it cwd-independent.

beacon/goclient/error_paths_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func newHTTPService(t *testing.T, addr string) *eth2clienthttp.Service {
+	t.Helper()
+
+	service, err := eth2clienthttp.New(
+		t.Context(),
+		eth2clienthttp.WithAddress(addr),
+		eth2clienthttp.WithLogLevel(zerolog.Disabled),
+		eth2clienthttp.WithTimeout(time.Second),
+		eth2clienthttp.WithReducedMemoryUsage(true),
```
momosh-ssv, line 125: Might be worth a short comment here noting that `WithReducedMemoryUsage(true)` is load-bearing: it skips the spec/genesis prefetch in `eth2clienthttp.New`, which is what keeps the cache cold so each test actually re-hits the mock. Without it, `IsActive` could warm the cache and the 429/500/timeout cases would silently miss the network.

beacon/goclient/error_paths_test.go
```diff
@@ -0,0 +1,117 @@
...
+		requestTimeout time.Duration
+	}{
+		{
+			name: "rate limited",
+			response: mocks.NewResponse(
+				json.RawMessage(`{"message":"rate limited"}`),
+				mocks.WithStatusCode(http.StatusTooManyRequests),
+				mocks.WithHeader("Retry-After", "1"),
+			),
+			wantErr: "GET failed with status 429",
```
momosh-ssv: Any reason we're not asserting on the typed `*api.Error` from go-eth2-client here? I fear `"GET failed with status 429"` couples us to the upstream error wording — if the lib rephrases, every case in this table breaks at once with no signal about what really regressed. A structural check on `StatusCode` would be more durable.

momosh-ssv (commented):
One small suggestion: maybe worth adding a 503 case alongside 429/500. Beacon nodes commonly return 503 during fork transitions, and the new harness makes it cheap to cover.

beacon/goclient/error_paths_test.go
```diff
@@ -0,0 +1,117 @@
...
+				json.RawMessage(`{"message":"boom"}`),
+				mocks.WithStatusCode(http.StatusInternalServerError),
+			),
+			wantErr: "GET failed with status 500",
+		},
+		{
+			name: "timeout",
+			response: mocks.NewResponse(
+				nil,
+				mocks.WithDelay(200*time.Millisecond),
```
iurii-ssv: Seems like this `200*time.Millisecond` value here might cause flakyness, we should use at least `400*time.Millisecond` (from some experiments I've ran), but maybe even `1000*time.Millisecond` wouldn't hurt for cases where "waiting is cheap".

beacon/goclient/mocks/server.go
```diff
@@ -6,44 +6,145 @@ import (
...
+	return func(resp *Response) {
+		resp.StatusCode = statusCode
+	}
+}
+
+func WithHeader(key, value string) ResponseOption {
+	return func(resp *Response) {
+		if resp.Headers == nil {
+			resp.Headers = make(http.Header)
+		}
```
iurii-ssv: That `nil`-check here seems redundant (if we really really wanted to sanity-check ourselves on not "calling this method twice" - we should just `panic` ... but that's not really common to do these sorts of assertions, I don't think, since it makes the code harder to read & reason about)

beacon/goclient/error_paths_test.go
```diff
@@ -0,0 +1,117 @@
...
+				defer cancel()
+			}
+
+			_, err := client.specForClient(ctx, provider)
+			require.ErrorContains(t, err, tc.wantErr)
+		})
+	}
+}
+
+func Test_genesisForClient_malformedJSON(t *testing.T) {
```
iurii-ssv, line 97: [`genesis_test.go`](https://github.com/ssvlabs/ssv/blob/stage/beacon/goclient/genesis_test.go#L114-L136) already exercises malformed-JSON for genesis — indirectly, through `New()` timing out. The new [`Test_genesisForClient_malformedJSON`](https://github.com/ssvlabs/ssv/pull/2834/files#diff-beacon-goclient-error-paths-test-goR83-R99) is a strict improvement (faster, more targeted), so this isn't a blocker — but I'd consider deleting the older "error" subtest to avoid two tests covering the same ground.

iurii-ssv (commented):
LGTM, mostly minor suggestions

beacon/goclient/mocks/server.go
julienharbulot: Addressed in ef542b7e8 by marshalling the unexpected-request payload with `json.Marshal`, so the mock now emits valid JSON for any path.

beacon/goclient/mocks/server.go
julienharbulot: Addressed in ef542b7e8 by treating zero `StatusCode` as misuse and panicking instead of silently defaulting to 200.

beacon/goclient/mocks/server.go
julienharbulot: Addressed in ef542b7e8 by anchoring `mock-beacon-responses.json` lookup to the package directory via `runtime.Caller` and `filepath.Join`.

beacon/goclient/error_paths_test.go
julienharbulot: Addressed in ef542b7e8 by documenting why `WithReducedMemoryUsage(true)` is load-bearing for these mock-backed error-path tests.

beacon/goclient/error_paths_test.go
julienharbulot: Addressed in ef542b7e8 by switching the 429/500 assertions to structural `*api.Error` status-code checks, and I added a 503 case while touching that table.

beacon/goclient/error_paths_test.go
julienharbulot: Addressed in ef542b7e8 by increasing the timeout-case mock delay from `200ms` to `400ms`.

beacon/goclient/mocks/server.go
julienharbulot: Addressed in ef542b7e8 by removing the redundant `nil` check in `WithHeader`.

beacon/goclient/error_paths_test.go
julienharbulot: Addressed in ef542b7e8 by removing the older malformed-genesis subtest from `genesis_test.go`, leaving the more targeted error-path coverage.

iurii-ssv (approved):

momosh-ssv (approved):
All review feedback addressed in ef542b7e8 — thanks!

PR #2835, opened 2026-04-27
qbft: clarify round-changes

iurii-ssv:
See Greptile summary below.

Closes https://github.com/ssvlabs/ssv-node-board/issues/1004

Closes ssvlabs/ssv-node-board#1052

### DIFF
--- a/protocol/v2/qbft/instance/observability.go
+++ b/protocol/v2/qbft/instance/observability.go
@@ -31,7 +31,6 @@ type roundChangeReason string
 const (
 	reasonTimeout       roundChangeReason = "timeout"
 	reasonPartialQuorum roundChangeReason = "partial-quorum"
-	reasonJustified     roundChangeReason = "justified"
 )
 
 var (
--- a/protocol/v2/qbft/instance/proposal.go
+++ b/protocol/v2/qbft/instance/proposal.go
@@ -35,13 +35,13 @@ func (i *Instance) uponProposal(ctx context.Context, logger *zap.Logger, msg *sp
 	i.metrics.EndStage(ctx, currentRound)
 	i.metrics.StartStage(stagePrepare)
 
-	i.State.ProposalAcceptedForCurrentRound = msg
-
 	// A future justified proposal should move us into the future round, hence we try to bump the round here.
 	// Always move on to the message-round. The round-change message broadcast is a best-effort thing, the QBFT
 	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
 	// for whatever reason.
 	i.bumpToRound(msgRound)
+
+	i.State.ProposalAcceptedForCurrentRound = msg
 
 	r := qbft.HashDataRoot(msg.SignedMessage.FullData)
 	prepare, err := i.CreatePrepare(msgRound, r)
--- a/protocol/v2/qbft/instance/proposal.go
+++ b/protocol/v2/qbft/instance/proposal.go
@@ -10,6 +10,7 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 	"go.uber.org/zap"
 
+	"github.com/ssvlabs/ssv/protocol/v2/qbft"
 	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
 )
 
@@ -28,24 +29,21 @@ func (i *Instance) uponProposal(ctx context.Context, logger *zap.Logger, msg *sp
 
 	logger.Debug("📬 got proposal message")
 
-	i.State.ProposalAcceptedForCurrentRound = msg
-
+	currentRound := i.State.Round
 	msgRound := msg.QBFTMessage.Round
 
-	// A future justified proposal should bump us into future round and reset the round timer.
-	if msgRound > i.State.Round {
-		i.roundTimer.TimeoutForRound(msgRound)
-	}
-	i.bumpToRound(msgRound)
-
-	i.metrics.EndStage(ctx, msgRound)
+	i.metrics.EndStage(ctx, currentRound)
 	i.metrics.StartStage(stagePrepare)
 
-	r, err := specqbft.HashDataRoot(msg.SignedMessage.FullData)
-	if err != nil {
-		return fmt.Errorf("could not hash input data: %w", err)
-	}
+	i.State.ProposalAcceptedForCurrentRound = msg
 
+	// A future justified proposal should move us into the future round, hence we try to bump the round here.
+	// Always move on to the message-round. The round-change message broadcast is a best-effort thing, the QBFT
+	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
+	// for whatever reason.
+	i.bumpToRound(msgRound)
+
+	r := qbft.HashDataRoot(msg.SignedMessage.FullData)
 	prepare, err := i.CreatePrepare(msgRound, r)
 	if err != nil {
 		return fmt.Errorf("could not create prepare msg: %w", err)
@@ -83,10 +81,7 @@ func (i *Instance) isValidProposal(msg *specqbft.ProcessingMessage) error {
 	}
 
 	// verify full data integrity
-	r, err := specqbft.HashDataRoot(msg.SignedMessage.FullData)
-	if err != nil {
-		return fmt.Errorf("could not hash input data: %w", err)
-	}
+	r := qbft.HashDataRoot(msg.SignedMessage.FullData)
 	if !bytes.Equal(msg.QBFTMessage.Root[:], r[:]) {
 		return spectypes.NewError(spectypes.RootHashInvalidErrorCode, "H(data) != root")
 	}
@@ -189,10 +184,7 @@ func (i *Instance) isProposalJustification(
 	}
 
 	// proposed fullData must equal highest prepared fullData
-	r, err := specqbft.HashDataRoot(fullData)
-	if err != nil {
-		return fmt.Errorf("could not hash input data: %w", err)
-	}
+	r := qbft.HashDataRoot(fullData)
 	if !bytes.Equal(r[:], rcMsg.QBFTMessage.Root[:]) {
 		return errors.New("proposed data doesn't match highest prepared")
 	}
@@ -233,16 +225,13 @@ func (i *Instance) ProposerForRound(round specqbft.Round) spectypes.OperatorID {
                         extractSignedPrepares(prepares));
 */
 func (i *Instance) CreateProposal(fullData []byte, roundChanges, prepares []*specqbft.ProcessingMessage) (*spectypes.SignedSSVMessage, error) {
-	r, err := specqbft.HashDataRoot(fullData)
-	if err != nil {
-		return nil, fmt.Errorf("could not hash input data: %w", err)
-	}
+	r := qbft.HashDataRoot(fullData)
 
-	roundChangeSignedMessages := make([]*spectypes.SignedSSVMessage, 0)
+	roundChangeSignedMessages := make([]*spectypes.SignedSSVMessage, 0, len(roundChanges))
 	for _, msg := range roundChanges {
 		roundChangeSignedMessages = append(roundChangeSignedMessages, msg.SignedMessage)
 	}
-	prepareSignedMessages := make([]*spectypes.SignedSSVMessage, 0)
+	prepareSignedMessages := make([]*spectypes.SignedSSVMessage, 0, len(prepares))
 	for _, msg := range prepares {
 		prepareSignedMessages = append(prepareSignedMessages, msg.SignedMessage)
 	}
--- a/protocol/v2/qbft/instance/round_change.go
+++ b/protocol/v2/qbft/instance/round_change.go
@@ -115,7 +115,6 @@ func (i *Instance) uponChangeRoundPartialQuorum(ctx context.Context, logger *zap
 	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
 	// for whatever reason.
 	i.bumpToRound(newRound)
-	i.State.ProposalAcceptedForCurrentRound = nil
 
 	startValueRoot := qbft.HashDataRoot(i.StartValue)
 	logger = logger.With(zap.String("qbft_start_value_root", hex.EncodeToString(startValueRoot[:])))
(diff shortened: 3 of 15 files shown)


protocol/v2/qbft/instance/proposal.go
```diff
@@ -28,24 +29,23 @@ func (i *Instance) uponProposal(ctx context.Context, logger *zap.Logger, msg *sp
...
+
 	// A future justified proposal should bump us into future round and reset the round timer.
+	// Always move on to the message-round, the round-change message broadcast is a best-effort thing, the QBFT
+	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
+	// for whatever reason.
+	i.bumpToRound(msgRound)
+	i.State.ProposalAcceptedForCurrentRound = msg
 	if msgRound > i.State.Round {
 		i.roundTimer.TimeoutForRound(msgRound)
 	}
```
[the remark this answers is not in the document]
iurii-ssv: Good catch, that's a typo, fixed in https://github.com/ssvlabs/ssv/pull/2835/changes/a3532cee41dbdd9dc908a8179b9d18904f91f698

iurii-ssv:
@greprile pls re-review

iurii-ssv:
> Notable behavioral change: in `UponRoundTimeout`, when a node reaches `CutOffRound-1`, the round is now bumped to `CutOffRound` _before_ `Broadcast` is called, causing the broadcast to fail with "instance is no longer considered relevant" — whereas previously the round-change message was successfully broadcast then the round was bumped via a deferred call. The test `TestUponRoundTimeoutStopsProcessingAfterReachingCutOffRound` explicitly validates this new behavior.

This ^ is slight behavior improvement/correction, since there is no point to broadcast a message with CutOffRound as it will simply be ignored/dropped by other nodes.

> Only a single P2 observation about lost observability on the reasonJustified metric.

That `reasonJustified` metric was incorrect btw, it was recorded "as if" it was a round-change ... while by the time that branch executes, we are already in that new round, and just proposing. So the original implementation/classification for this metric was a bit mislead by the complex handling code.

protocol/v2/qbft/instance/round_change.go
```diff
@@ -66,68 +70,60 @@ func (i *Instance) uponRoundChange(
 
 		proposal, err := i.CreateProposal(
 			valueToPropose,
-			i.State.RoundChangeContainer.MessagesForRound(prevRound), // TODO - might be optimized to include only necessary quorum
+			i.State.RoundChangeContainer.MessagesForRound(i.State.Round), // TODO - might be optimized to include only necessary quorum
```
julienharbulot, line 83: I think this leader path also needs to move the instance to the justified round before building the proposal? Here's the thinking: `hasReceivedProposalJustificationForLeadingRound` explicitly accepts future-round quorums (`newRound > i.State.Round`), but this branch still calls `CreateProposal` while `i.State.Round` is the old round, and `CreateProposal` uses `i.State.Round` as the proposal round. In a local repro with state round 1 and a quorum of round-change messages for round 2, the leader broadcasts a round-1 proposal and stays at round 1. We should bump to `justifiedRoundChangeMsg.QBFTMessage.Round` before `CreateProposal` and use `MessagesForRound` for that round.

julienharbulot: For context, simpler to explain in code:

```go
func TestUponRoundChangeAsLeaderBroadcastsFutureRoundProposalOnJustifiedQuorum(t *testing.T) {
        env := newInstanceTestEnv(t, 1)
        env.setLeader(1)
        env.inst.State.Round = 1
        env.inst.StartValue = []byte("start-value")

        env.addMessages(
                env.inst.State.RoundChangeContainer,
                env.roundChange(2, 2, specqbft.NoRound, [32]byte{}, nil, nil),
                env.roundChange(2, 3, specqbft.NoRound, [32]byte{}, nil, nil),
        )

        err := env.inst.uponRoundChange(
                context.Background(),
                zap.NewNop(),
                env.roundChange(2, 4, specqbft.NoRound, [32]byte{}, nil, nil),
        )
        require.NoError(t, err)

        msg := env.broadcastedProcessingMessage(0)
        require.Equal(t, specqbft.ProposalMsgType, msg.QBFTMessage.MsgType)
        require.Equal(t, specqbft.Round(2), msg.QBFTMessage.Round)
        require.Equal(t, specqbft.Round(2), env.inst.State.Round)
  }
```

protocol/v2/qbft/instance/timeout.go
```diff
@@ -4,53 +4,45 @@ import (
...
-	// Remember to track the impact of this change and revert/modify if necessary.
-	defer func() {
-		i.bumpToRound(newRound)
-		i.State.ProposalAcceptedForCurrentRound = nil
-		i.roundTimer.TimeoutForRound(newRound)
-	}()
+	// Always move on to the next round. The round-change message broadcast is a best-effort thing, the QBFT
+	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
+	// for whatever reason.
+	i.bumpToRound(newRound)
```
momosh-ssv, line 39: Could this lead to a liveness regression at cutoff? The bump now runs before `Broadcast`, so when `prevRound + 1 == CutOffRound` the bump succeeds but `i.Broadcast` calls `IsRelevant()` which returns false (`State.Round < CutOffRound` no longer holds), and the round-change is silently dropped. A node that hits cutoff one tick before its peers will not send the RC that could have helped them advance. The new test on lines 62-74 actually encodes this behavior. Maybe broadcast before bumping (revert to the defer-style ordering), or relax `IsRelevant` for `Broadcast` so the cutoff round can still emit its final RC?

protocol/v2/qbft/instance/round_change.go
```diff
@@ -53,6 +54,9 @@ func (i *Instance) uponRoundChange(
 	}
 
 	if justifiedRoundChangeMsg != nil {
+		i.metrics.EndStage(ctx, i.State.Round)
```
momosh-ssv, line 57: Should we consider only advancing the stage after the broadcast actually succeeds? `EndStage` / `StartStage` now run before `CreateProposal` / `Broadcast` here (and similarly at lines 110-112 for the partial-quorum path). I fear if `CreateProposal` or `MarshalJustifications` errors, the metric reports `stageProposal` while nothing was emitted, which can make dashboards misleading.

protocol/v2/qbft/instance/observability.go
```diff
@@ -31,7 +31,6 @@ type roundChangeReason string
 const (
 	reasonTimeout       roundChangeReason = "timeout"
 	reasonPartialQuorum roundChangeReason = "partial-quorum"
```
momosh-ssv, line 33: What do you think about keeping `reasonJustified`? Once the leader-path round-bump issue (open thread on `round_change.go`) is addressed and a real round change happens in the justified path, there will be no labeled metric for it. Operators currently can distinguish 'round changed because of justified RC quorum' from 'timeout' / 'partial-quorum' — losing that makes it harder to diagnose why a cluster keeps round-changing.

protocol/v2/qbft/messages.go
```diff
@@ -0,0 +1,8 @@
+package qbft
+
+import "crypto/sha256"
+
+// HashDataRoot hashes input data to root.
+func HashDataRoot(data []byte) [32]byte {
+	return sha256.Sum256(data)
```
momosh-ssv, line 7: Worth considering a guard against silent divergence from spec? The new helper is byte-identical to `specqbft.HashDataRoot` today and `messages_test.go` checks it on a few inputs at fix-time. Seems that nothing in `protocol/v2/qbft/spectest/` exercises this code path though, so a future ssv-spec hash change would let nodes hash differently from spec while `make spec-test` still passes. Maybe keep this as a thin wrapper (`r, _ := specqbft.HashDataRoot(data); return r`) or assert equivalence against the spec helper in CI?

y0sher:
@momosh-ssv @Roy-blox @iurii-ssv 
this change QBFT code in critical paths, lets finish with code review and test thoroughly if we want it in the version. we'll want some mileage in staging env.

protocol/v2/qbft/messages.go
iurii-ssv: Spec-divergence risk is covered by [`TestHashDataRootMatchesSpec`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/messages_test.go#L10-L21) — it asserts equivalence against `specqbft.HashDataRoot` on several inputs, so if the spec ever changes its hash output, `make unit-test` fails before anything ships.

protocol/v2/qbft/instance/round_change.go
iurii-ssv: Good catch — fix pushed (7ddb771f1). `bumpToRound(justifiedRoundChangeMsg.QBFTMessage.Round)` now runs before `CreateProposal`, so the proposal carries the justified round and `MessagesForRound(i.State.Round)` returns the right justifications.

That said, after investigating, this future-round leader case turns out to be structurally unreachable via natural message flow: the `RoundChangeContainer` is populated only inside [`uponRoundChange:34`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/round_change.go#L34) (single call site), partial-quorum (f+1) is strictly less than full quorum (2f+1), and [`processMsgF.Run()`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/instance.go#L197) serializes processing — so the (f+1)-th RC always triggers the partial-quorum branch and bumps `State.Round` before the (2f+1)-th arrives. Your reproducer bypasses that via direct `addMessages`.

Fix is still worth keeping — closes a latent inconsistency between `hasReceivedProposalJustificationForLeadingRound` (which accepts future rounds) and the surrounding code (which assumed current round). Test added with a comment explaining the deliberate container preload.

protocol/v2/qbft/instance/timeout.go
iurii-ssv: Intentional. Round timers are deterministic and started in lockstep at the start of a duty — once a node hits `CutOffRound`, peers running the same schedule are at most one round-tick behind and will reach it on their own shortly. The marginal value of broadcasting one extra RC at cutoff is essentially zero, and we preferred the cleaner bump-before-broadcast ordering for consistency with the partial-quorum and justified paths.

protocol/v2/qbft/instance/round_change.go
iurii-ssv: Keeping consistent with the "best-effort broadcast" framing this PR uses throughout: the protocol stage advances when we move on (we've already built the new message and the state machine has progressed); broadcast failures are tracked independently. Gating stage advancement on a successful broadcast would create a different inconsistency — on broadcast failure the instance would still be in the new round but the dashboard would show the old stage.

protocol/v2/qbft/instance/observability.go
iurii-ssv: Fair concern. After digging in, this metric was effectively double-counting in pre-PR code: by the time the leader-on-justified-quorum branch fired, `State.Round` had already been advanced (and recorded as a round change) via `reasonTimeout` or `reasonPartialQuorum`. The reasoning:

1. `RoundChangeContainer` is populated only inside [`uponRoundChange:34`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/round_change.go#L34) — single call site (grep-verified).
2. Inside `uponRoundChange`, [partial-quorum is checked](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/round_change.go#L95) after `AddFirstMsgForSignerAndRound` on every call.
3. Partial-quorum threshold (f+1) is strictly less than full quorum (2f+1) by definition.
4. [`processMsgF.Run()`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/instance.go#L197) serializes processing.

So as RCs stream in, the (f+1)-th always triggers `uponChangeRoundPartialQuorum` → bumps `State.Round` — before the (2f+1)-th arrives and the leader branch fires. By the time the leader branch executes, `bumpToRound(newRound)` is always a no-op (`State.Round == newRound` already). No real round change happens in this branch.

Soft confirmation: if this future-round leader case *were* a real scenario in normal flow, ssv-spec tests would have a scenario for it — they don't. Leaving the metric removed.

(Julien's `bumpToRound` fix in 7ddb771f1 is still worth keeping for defense-in-depth against future refactors, but the metric would just be a perpetual zero.)

iurii-ssv:
Note: force-pushed rebased commits as my and Momo's interleaving `git merge` produced an awkward commit history (pulling https://github.com/ssvlabs/ssv/commit/4f78c80c47aaded9b9733aef2cd8a2dd5c3c3ba0 in to this PR, which is fixed now)

momosh-ssv:
Deferring to **v2.4.4**.

Not enough time remaining in the v2.4.3 window to soak this properly — touches the consensus path, so we want a full stage-soak rather than rushing the cut. Rolling to the next minor.

Tracking issue moved to milestone v2.4.4: ssvlabs/ssv-node-board#1052.

julienharbulot (approved):
All review comments addressed, lgtm

protocol/v2/qbft/instance/observability.go
momosh-ssv: Ok, got it 👌 
One thing though: this leans on `bumpToRound` fix from [7ddb771](https://github.com/ssvlabs/ssv/commit/7ddb771f1949033bdbcd6ab085757b5901b0bc58), but I can't find that bump or the future-round test in the current branch — seems the cleanup squash kept only a same-round test?
Could we confirm it was dropped intentionally (since the case is unreachable), or restore it?

iurii-ssv: That `7ddb771` got rebased to [`5ff4c098`](https://github.com/ssvlabs/ssv/commit/5ff4c098a9a76c2da1aa72385e83311bd4348714) when the branch was synced with `stage` (same change, new SHA — that's why the old link doesn't resolve here). Both pieces are live on the branch:
- the bump at [`round_change.go:68`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/round_change.go#L68)
- the future-round test [`TestUponRoundChangeAsLeaderBroadcastsFutureRoundProposalOnJustifiedQuorum`](https://github.com/ssvlabs/ssv/blob/qbft-correct-instance-message-retries/protocol/v2/qbft/instance/round_change_test.go#L230) (alongside the same-round one). Kept as defense-in-depth, as discussed.

momosh-ssv: ok then, all good for me

momosh-ssv (approved):

PR #2838, opened 2026-05-04
identity/store: rollback p2p network key storage to plaintext

nkryuchkov:
## Summary

Roll back persisted P2P network key storage to plaintext for stage rollback safety.

Closes https://github.com/ssvlabs/ssv-node-board/issues/1038

## What changed

- keep support for reading encrypted persisted P2P network keys
- when an encrypted key is successfully read during `SetupNetworkKey`, rewrite it back to plaintext in DB
- stop migrating plaintext keys to encrypted storage
- stop writing new persisted P2P network keys in encrypted format
- keep the decrypt path in place temporarily so already-migrated stage DBs can be downgraded safely before the actual revert

## Why

PR [#2765](https://github.com/ssvlabs/ssv/pull/2765) was already merged into `stage`, so some stage nodes may already have persisted P2P keys stored as `enc:` in DB.

A direct revert would be unsafe for those nodes. This bridge PR downgrades stored keys back to plaintext first, so the later revert can happen without breaking nodes that already migrated their DB.

## Scope

- this PR is intentionally a temporary rollback bridge
- exporter behavior is unchanged
- the follow-up step is to let this deploy and rewrite DBs, then revert the original encryption change

## Testing

- `go test ./cli/operator ./identity`

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -1119,12 +1119,12 @@ func setupSSVNetwork(logger *zap.Logger) (*networkconfig.SSV, error) {
 }
 
 func setupP2P(ctx context.Context, logger *zap.Logger, db basedb.Database, exporterEnabled bool, operatorPrivKey keys.OperatorPrivateKey, signerClient *ssvsigner.Client) network.P2PNetwork {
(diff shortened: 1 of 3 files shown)


nkryuchkov:
@greptileai review it again

iurii-ssv (approved):
LGTM

julienharbulot (approved):

momosh-ssv (approved):

PR #2841, opened 2026-05-05
Revert network key encryption, fix findings in original code

nkryuchkov:
## Summary

Revert [#2765](https://github.com/ssvlabs/ssv/pull/2765) and [#2838](https://github.com/ssvlabs/ssv/pull/2838).

## Why

We decided to move persisted P2P network key encryption to a future release to reduce rollout risk.

[#2838](https://github.com/ssvlabs/ssv/pull/2838) was a temporary rollback bridge that downgraded already-migrated stage DBs back to plaintext. Now that it has been deployed, we can remove both:
- the original encrypted-storage change from [#2765](https://github.com/ssvlabs/ssv/pull/2765)
- the temporary plaintext-rollback bridge from [#2838](https://github.com/ssvlabs/ssv/pull/2838)

## What changed

- revert persisted P2P network key encryption
- revert remote/local network-key protection wiring
- revert the temporary downgrade-to-plaintext bridge
- restore the previous plaintext-only persisted P2P key behavior
- keep current `stage` changes unrelated to `#2765` and `#2838`
- fix two revert regressions in `identity/store.go`:
  - return DB errors before treating the key as missing
  - return decode errors before touching `pk.Curve`
- restore the malformed-ciphertext guard in `ssvsigner/ekm/signer_storage.go` to `<= nonceSize`

## Testing

- `go test ./identity ./cli/operator`
- `go test ./ekm` in `ssvsigner/`
- `make golangci-lint`
- `make deadcode-lint`


nkryuchkov:
@greptileai review this again

iurii-ssv (approved):
LGTM, safe to merge as revert has been applied across both stage & prod envs.

identity/store.go
```diff
@@ -1,13 +1,10 @@
...
 
 import (
-	"bytes"
-	"context"
 	"crypto/ecdsa"
-	"errors"
-	"fmt"
 
 	gcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/pkg/errors"
```
momosh-ssv, line 7: Hey, just to check something — was switching back to `pkg/errors` here intentional? #2825 just replaced it with stdlib repo-wide, and after this revert lands `identity/store.go` looks like the only file still importing `github.com/pkg/errors` (now `// indirect` in go.mod). Happy to keep it as faithful-revert if you'd rather, just figured worth flagging.

PR #2843, opened 2026-05-06
Track pubsub topic validator receives - F-ssv-1044

julienharbulot:
## Ticket

F-ssv-1044

The ticket reports an observability gap: existing SSV validation counters show what happened after SSV validation runs, but they do not show whether a pubsub message reached the node's topic validator at all. That leaves delivery/topic/mesh issues ambiguous from validation/decoding issues.

## Changes

- Added `ssv_p2p_pubsub_messages_received_total` via OTel instrument `ssv.p2p.pubsub.messages.received`.
- Records the counter at the libp2p topic validator entrypoint, before delegating to SSV message validation.
- Labels the metric only by bounded `topic`.
- Added a focused metric test for the counter value and topic attribute.

## Validation

- `go test ./network/topics -run TestRecordPubsubMessageReceived -count=1 -timeout=30s`
- `go test ./network/topics -run '^$' -count=1 -timeout=30s`
- `GOWORK=off go tool -modfile=tool.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run ./network/topics`

Fixes - Fixes ssvlabs/ssv-node-board#1044


iurii-ssv:
@greptile pls re-review

network/topics/observability.go
```diff
@@ -12,6 +14,9 @@ import (
 const (
 	observabilityName      = "github.com/ssvlabs/ssv/network/topics"
 	observabilityNamespace = "ssv.p2p.messages"
+
+	pubsubObservabilityNamespace = "ssv.p2p.pubsub.messages"
+	pubsubTopicAttributeKey      = "ssv.p2p.pubsub.topic"
```
iurii-ssv, line 19: **Attribute-key fragmentation for "topic name"** (informational — not blocking)

This new key labels exactly the same logical thing (a libp2p topic name) as two existing keys in the codebase:

- `ssv.p2p.message.topic` — used by `inboundMessageCounter`/`outboundMessageCounter` in this very file (`messageTopicAttribute`, line 55).
- `ssv.p2p.topic.name` — used by `peersPerTopicGauge` in `network/p2p/observability.go:91`.

With this PR there are now three keys for the same value across the P2P metrics surface. The most natural correlation an operator wants — "received → inbound → peers per topic" — cannot be done with a single `group by topic` in PromQL; they'll need to remember which key applies to which counter.

I understand the rationale (the new `ssv.p2p.pubsub.*` namespace deserves its own attribute prefix), so I'm not suggesting a concrete change here — it's a design call. But it would be worth either reusing one of the existing keys, or opening a follow-up to consolidate all three under a single canonical key (e.g. `ssv.p2p.topic.name`).

network/topics/observability.go
```diff
@@ -29,13 +34,23 @@ var (
 			metric.WithUnit("{message}"),
 			metric.WithDescription("total number of outbound(broadcasted) messages")))
 
+	pubsubMessagesReceivedCounter = metrics.New(
+		meter.Int64Counter(
+			observability.InstrumentName(pubsubObservabilityNamespace, "received"),
+			metric.WithUnit("{message}"),
+			metric.WithDescription("total number of messages received by the pubsub topic validator")))
```
iurii-ssv: **Description is ambiguous next to `inboundMessageCounter`**

The description here ("total number of messages received by the pubsub topic validator") and `inboundMessageCounter`'s description ("total number of inbound messages", line 29) sound nearly identical to anyone reading the metric registry — yet the entire point of this PR is that the two counters measure different things (pre- vs. post-validation). Tightening the wording so the distinction is visible at a glance:

```suggestion
			metric.WithDescription("total number of messages delivered to the pubsub topic validator, before SSV validation runs (compare with ssv_p2p_messages_in_total for the post-validation rate)")))
```

network/topics/observability.go
```diff
@@ -46,3 +61,7 @@ func messageTypeAttribute(value uint64) attribute.KeyValue {
 		Value: observability.Uint64AttributeValue(value),
 	}
 }
+
+func recordPubsubMessageReceived(ctx context.Context, topic string) {
```
iurii-ssv, line 68: **Minor: worth a 1-line comment to flag pre-validation timing**

A future reader touching the wrapper in `controller.go` might assume this counts only successful deliveries. Calling out the timing explicitly here means they don't have to chase the call site to find out:

```suggestion
// recordPubsubMessageReceived is called from the topic validator wrapper before the inner SSV
// validator runs, so the counter increments for every message libp2p hands to the validator
// regardless of validation outcome (accept/ignore/reject/timeout).
func recordPubsubMessageReceived(ctx context.Context, topic string) {
```

network/topics/observability_test.go
```diff
@@ -0,0 +1,48 @@
...
+	var rm metricdata.ResourceMetrics
+	require.NoError(t, reader.Collect(t.Context(), &rm))
+
+	for _, scopeMetrics := range rm.ScopeMetrics {
+		for _, metric := range scopeMetrics.Metrics {
+			if metric.Name != "ssv.p2p.pubsub.messages.received" {
+				continue
+			}
+
+			sum, ok := metric.Data.(metricdata.Sum[int64])
```
iurii-ssv: **Nit: loop variable `metric` shadows the imported `metric` package**

From `go.opentelemetry.io/otel/sdk/metric` (line 8). Compiles only because the package isn't referenced inside the loop body, but it's an avoidable foot-gun if anyone later edits this block. Renaming to `m`:

```suggestion
		for _, m := range scopeMetrics.Metrics {
			if m.Name != "ssv.p2p.pubsub.messages.received" {
				continue
			}

			sum, ok := m.Data.(metricdata.Sum[int64])
```

iurii-ssv (commented):
Leaving a few inline notes. Most are minor; the only judgement call is the attribute-key fragmentation across P2P metrics — flagged as informational since it touches design rather than mechanics.

If helpful, I've put the mechanical fixes (description, helper comment, test shadowing rename) into a draft PR you can merge into this branch as-is: **https://github.com/ssvlabs/ssv/pull/2862**. Cherry-pick or close as you prefer.

network/topics/observability_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+)
+
+func TestRecordPubsubMessageReceived(t *testing.T) {
+	reader := metric.NewManualReader()
+	provider := metric.NewMeterProvider(metric.WithReader(reader))
+	previousProvider := otel.GetMeterProvider()
+	otel.SetMeterProvider(provider)
```
momosh-ssv, line 16: Should we consider moving the provider swap into a `TestMain`, like `protocol/v2/ssv/queue/main_test.go` does? 

Seems that setting the global provider mid-test leaves the package-level counters re-pointed for whatever runs after, and restoring the previous (delegating) provider in cleanup may not actually rebind them. 

Setting a `ManualReader`-backed provider once before `m.Run()` would match the existing pattern and avoid the order dependence.

network/topics/controller.go
```diff
@@ -308,7 +308,13 @@ func (ctrl *topicsCtrl) setupTopicValidator(name string) error {
 
 		opts := []pubsub.ValidatorOpt{pubsub.WithValidatorTimeout(topicValidatorTimeout)}
 
-		err = ctrl.ps.RegisterTopicValidator(name, ctrl.msgValidator.ValidatorForTopic(name), opts...)
+		validator := ctrl.msgValidator.ValidatorForTopic(name)
+		wrappedValidator := func(ctx context.Context, p peer.ID, pmsg *pubsub.Message) pubsub.ValidationResult {
```
momosh-ssv, line 312: Might be worth a test that goes through the wrapper itself — the new test calls `recordPubsubMessageReceived` directly, so the wiring here stays unverified. 

Registering the validator and pushing one message through it, then asserting the counter, would close that gap.

PR #2854, opened 2026-05-18
feat(p2p): add highlighted peer observability

julienharbulot:
## Summary

Adds targeted observability for attack-simulator traffic by letting an SSV node highlight one or more configured libp2p peers, including peers configured by their secp256k1 public key. The stage attack-simulator public key can now be passed through `P2P_HIGHLIGHTED_PEERS`, with an optional `P2P_HIGHLIGHTED_PEER_LABEL` such as `attack-simulator`.

The implementation wires a shared highlighted peer observer through P2P setup and message validation so we can follow traffic from the attack simulator across connection, stream, pubsub, peer score, wire validation, and SSV-level validation paths.

## What changed

- Added `network/peers/peertrace` to parse highlighted peer IDs/public keys and emit consistent logs and metrics.
- Added config/env support for `P2P_HIGHLIGHTED_PEERS` and `P2P_HIGHLIGHTED_PEER_LABEL`.
- Added highlighted peer logs for connection handling, stream handling, pubsub validation, delivered pubsub messages, pubsub trace events, and peer score inspection.
- Added SSV-level validation observer events so highlighted messages report accepted/ignored/rejected decisions with reason, role, SSV message type, slot, duty executor ID, signers, and QBFT fields when available.
- Added counters under `ssv.p2p.highlighted_peer` for events, pubsub validation outcomes, and SSV validation outcomes.
- Expanded pubsub trace metadata for message counts/topics/IDs and IHAVE/IWANT/GRAFT/PRUNE control metadata, which helps identify barrage and mesh/control-plane attacks.
- Documented the attack-simulator public key in `config/config.example.yaml`.

## Why

When running attack-simulator against a stage SSV node, it was difficult to tell whether attack traffic reached the node, whether it passed wire-level pubsub validation, and how far it progressed through SSV message validation. Highlighting the simulator peer gives us a low-noise view of simulator-originated traffic and enough context to understand whether each attack is affecting gossip, scoring, stream handling, or SSV validation.

## Validation

- `gopls` diagnostics on edited Go files: clean
- `git diff --check`: clean
- ok  	github.com/ssvlabs/ssv/message/validation	(cached)
ok  	github.com/ssvlabs/ssv/network/peers/peertrace	(cached)
ok  	github.com/ssvlabs/ssv/network/topics	(cached)
- ok  	github.com/ssvlabs/ssv/cli/operator	(cached) [no tests to run]
ok  	github.com/ssvlabs/ssv/network/p2p	(cached) [no tests to run]
- `go test ./network/streams ./network/peers/connections ./network/peers/peertrace ./message/validation -timeout 60s`

Note: `go_vulncheck ./...` could not run locally because the scanner binary is built with Go 1.25 while this checkout requires Go 1.26. No dependency changes are included in this PR.


network/topics/scoring.go
```diff
@@ -160,6 +157,13 @@ func scoreInspector(logger *zap.Logger,
 			if peerScores.Score < -1000 {
 				fields = append(fields, zap.Bool("low_score", true))
 			}
+			peerObserver.Observe(context.Background(), logger, "pubsub_peer_score", pid, fields...)
+
+			if inspections%logFrequency != 0 {
+				// Don't log yet.
+				continue
+			}
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 07594efc7 by gating score field construction on either a log cycle or a highlighted-peer match. Non-highlighted peers now skip the extra field allocation between log cycles, while highlighted peers still emit pubsub_peer_score observations; covered by TestScoreInspectorObservesHighlightedPeersBetweenLogCycles.

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,274 @@
...
+	if !ok {
+		return
+	}
+
+	logFields := []zap.Field{
+		zap.Bool("p2p_highlight", true),
+		zap.String("p2p_highlight_label", o.label),
+		zap.String("p2p_highlight_event", "ssv_message_validated"),
+		zap.String("peer_id", match.ID.String()),
+		zap.String("peer_source", match.Source),
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 07594efc7 by passing the validator logger through the SSVValidationObserver interface into ObserveSSVValidation, instead of using zap.L(). Added a logger-backed observer test for this path.

network/peers/connections/conn_handler.go
```diff
@@ -47,7 +49,12 @@ func NewConnHandler(
 	connIdx peers.ConnectionIndex,
 	peerInfos peers.PeerInfoIndex,
 	discoveredPeersPool *ttl.Map[peer.ID, discovery.DiscoveredPeer],
+	peerObservers ...*peertrace.Observer,
 ) ConnHandler {
+	var peerObserver *peertrace.Observer
+	if len(peerObservers) > 0 {
+		peerObserver = peerObservers[0]
+	}
```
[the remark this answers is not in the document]
julienharbulot: Addressed in 07594efc7 by replacing the variadic observer parameter with a single *peertrace.Observer and updating the call sites/tests so the compiler enforces the single-observer contract.

message/validation/errors.go
```diff
@@ -181,10 +185,47 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
...
 
 	recordRejectedMessage(ctx, loggerFields.Role, valErr.Text())
+	mv.observeSSVValidation(ctx, peerID, loggerFields, SSVValidationRejected, valErr.Text(), valErr)
 	return pubsub.ValidationReject
 }
 
-func (mv *messageValidator) handleValidationSuccess(ctx context.Context, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
+func (mv *messageValidator) handleValidationSuccess(ctx context.Context, peerID peer.ID, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
 	recordAcceptedMessage(ctx, messageRole(decodedMessage))
+	mv.observeSSVValidation(ctx, peerID, mv.buildLoggerFields(decodedMessage), SSVValidationAccepted, "valid", nil)
```
alok-ssv, line 202: buildLoggerFields will happen for every message even when highlighting is disabled. I think its better to avoid calling this function if `mv.observer == nil` to avoid any performance regression.

alok-ssv (dismissed):

iurii-ssv:
@greptile pls re-review

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/libp2p/go-libp2p/core/peer"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.uber.org/zap"
+
+	ssvvalidation "github.com/ssvlabs/ssv/message/validation"
+	"github.com/ssvlabs/ssv/observability"
+	"github.com/ssvlabs/ssv/observability/metrics"
+	ssvmessage "github.com/ssvlabs/ssv/protocol/v2/message"
```
iurii-ssv: Layering nit: `peertrace` (a low-level p2p observability package) now imports `message/validation` (for `SSVValidationEvent`) and `protocol/v2/message` (for `MsgTypeToString` / `QBFTMsgTypeToString`). The `SSVValidationObserver` interface lives in `message/validation` but its sole implementer is here. There's no Go cycle today, but the conceptual direction is inverted and a future need for `message/validation` to itself observe via `peertrace` would close the loop.

Consider either moving `SSVValidationEvent`/`SSVValidationObserver` (it's a pure DTO + 1-method interface) to `peertrace` itself or to a neutral package, or having `cli/operator/node.go` provide a tiny adapter so `peertrace.Observer` doesn't need to know SSV-protocol types directly.

network/peers/connections/conn_gater.go
```diff
@@ -88,38 +121,48 @@ func (n *connGater) InterceptAddrDial(id peer.ID, multiaddr ma.Multiaddr) (allow
...
 
-	if !n.validateDial(remoteAddr) {
+	allowed, reason := n.validateDial(remoteAddr)
+	if !allowed {
 		// Yield this goroutine to allow others to run in-between connection attempts.
 		runtime.Gosched()
 
-		n.logger.Debug("connection rejected due to IP rate limit", zap.String("remote_addr", remoteAddr.String()))
+		n.observeDecision(connectionGaterPhaseAccept, connectionGaterDecisionReject, reason, "", libp2pnetwork.DirInbound)
+		n.logger.Debug("connection rejected by connection gater",
```
iurii-ssv: Operator-visible behavior change: the previous log line for this branch was `"connection rejected due to IP rate limit"`. It's now `"connection rejected by connection gater"` with a `reason` field (which can be `ip_rate_limit`, `invalid_remote_addr`, …).

Any dashboards/alerts that key on the old exact message will silently stop matching. Worth either:
- keeping the old message and just adding the `reason` field, or
- calling this out in the PR body / release notes so operators can update alerting.

Same applies to the `InterceptAccept` `"connection rejected due to inbound limit"` line just above — that one is preserved, which is good; the asymmetry is the issue.

network/topics/controller.go
```diff
@@ -308,10 +317,33 @@ func (ctrl *topicsCtrl) setupTopicValidator(name string) error {
 
 		opts := []pubsub.ValidatorOpt{pubsub.WithValidatorTimeout(topicValidatorTimeout)}
 
-		err = ctrl.ps.RegisterTopicValidator(name, ctrl.msgValidator.ValidatorForTopic(name), opts...)
+		validator := ctrl.msgValidator.ValidatorForTopic(name)
+		wrappedValidator := func(ctx context.Context, p peer.ID, pmsg *pubsub.Message) pubsub.ValidationResult {
+			recordPubsubMessageReceived(ctx, name)
```
iurii-ssv: Scope nit: `recordPubsubMessageReceived` (and its backing `ssv.p2p.pubsub.messages.received` counter) is incremented for **every** pubsub message, regardless of whether `PeerObserver` is enabled. That's a reasonable general-purpose metric, but the PR is titled "add highlighted peer observability" and the body doesn't mention this new counter.

Either:
- mention it explicitly in the PR description, or
- split it into its own commit/PR so it can be reviewed (cardinality, naming, ownership) on its own merits.

Minor by itself, but worth not letting it sneak in unannounced.

network/topics/pubsub.go
```diff
@@ -197,16 +199,16 @@ func NewPubSub(
 		psOpts = append(psOpts, pubsub.WithDirectPeers(cfg.StaticPeers))
 	}
 
-	if cfg.TraceLog {
-		psOpts = append(psOpts, pubsub.WithEventTracer(newTracer(logger)))
+	if cfg.TraceLog || cfg.PeerObserver.Enabled() {
```
iurii-ssv, line 202: Side-effect for existing `PubsubTrace=true` operators: the tracer is now instantiated when `PeerObserver.Enabled()` (good), but `tracer.go` also unconditionally adds `messageCount` / `messageTopics` / `messageIDs` / `graftCount` / `graftTopics` / `pruneCount` / `pruneTopics` / `prunePeerCount` to every `SEND_RPC` / `DROP_RPC` / `RECV_RPC` trace entry — including for operators who only have `PubsubTrace=true` and have **not** opted into highlighting.

That materially grows their existing trace log volume. Two options:
1. Gate the new metadata on `pst.peerObserver.Enabled()` (since the unconditional `pubsub event` log already had a slimmer schema before).
2. Leave as-is and call out the expanded payload in the PR description / release notes so operators running `PUBSUB_TRACE=true` aren't surprised.

Probably (2) is fine — just want it to be intentional.

network/topics/tracer.go
```diff
@@ -150,7 +179,30 @@ func (pst *psTracer) log(logger *zap.Logger, evt *ps_pb.TraceEvent) {
 	default:
 		return
 	}
-	logger.Debug("pubsub event", fields...)
+	if highlightedPeer != "" {
+		pst.peerObserver.Observe(context.Background(), logger, "pubsub_trace_"+strings.ToLower(evt.GetType().String()), highlightedPeer, fields...)
```
iurii-ssv: `context.Background()` here (and at `network/topics/scoring.go:168`) bypasses OTEL context propagation / cancellation for the highlighted-peer metric increments.

The tracer doesn't currently store a context, but it could capture the topics controller's `ctx` at construction time (the scoring inspector likewise has it available on the closure). Cheap fix, keeps shutdown semantics consistent with the rest of the metrics in this PR (e.g. `topicsCtrl.peerObserver.Observe(ctrl.ctx, ...)` in `controller.go`).

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,282 @@
...
+		zap.String("p2p_highlight_label", o.label),
+		zap.String("p2p_highlight_event", event),
+		zap.String("peer_id", match.ID.String()),
+		zap.String("peer_source", match.Source),
+	}
+	if match.PublicKeyHex != "" {
+		highlightFields = append(highlightFields, zap.String("peer_public_key", match.PublicKeyHex))
+	}
+	highlightFields = append(highlightFields, fields...)
+	logger.Info("p2p highlighted peer event", highlightFields...)
```
iurii-ssv, line 167: Inconsistent nil-logger handling: `ObserveSSVValidation` defensively swaps in `zap.NewNop()` when `logger == nil`, but `Observe` (here) and `ObserveValidation` call `logger.Info(...)` directly and would panic.

In practice all current callers pass a non-nil logger, but the API surface is now inconsistent. Either:
- add the same `if logger == nil { logger = zap.NewNop() }` guard here, or
- drop the guard in `ObserveSSVValidation` and document that callers must pass a non-nil logger.

Pick one to make the contract clear.

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,282 @@
...
+	logFields := []zap.Field{
+		zap.Bool("p2p_highlight", true),
+		zap.String("p2p_highlight_label", o.label),
+		zap.String("p2p_highlight_event", "ssv_message_validated"),
+		zap.String("peer_id", match.ID.String()),
+		zap.String("peer_source", match.Source),
+		zap.String("ssv_validation_result", event.Outcome),
+		zap.String("ssv_validation_reason", event.Reason),
+		zap.String("role", event.Role.String()),
+		zap.Int32("role_id", int32(event.Role)),
```
iurii-ssv: Nit: `spectypes.RunnerRole` is `uint64`, but this field is logged as `Int32`. The cast silently truncates if a role ID ever exceeds 2^31 (won't happen in practice, but the type mismatch is gratuitous). Suggest `zap.Uint64("role_id", uint64(event.Role))` for symmetry with the `slot` field above.

iurii-ssv (commented):
A few additional observations beyond what's already been raised by Greptile, mostly: (a) one operator-visible log-message change worth calling out, (b) one out-of-scope new metric, (c) a couple of small consistency / correctness nits.

cli/operator/node.go
```diff
@@ -446,13 +447,24 @@ var StartNodeCmd = &cobra.Command{
...
+		cfg.P2pNetworkConfig.PeerObserver = highlightedPeerObserver
+
 		messageValidator := validation.New(
 			networkConfig,
 			nodeStorage.ValidatorStore(),
 			nodeStorage,
 			dutyStore,
 			signatureVerifier,
 			validation.WithLogger(logger),
+			validation.WithSSVValidationObserver(highlightedPeerObserver),
```
momosh-ssv: Could this lead to the disabled-observer guard never firing? 
`peertrace.New` returns a nil `*peertrace.Observer` when no peers are configured (the default), and passing that concrete nil into `WithSSVValidationObserver` (which takes the `SSVValidationObserver` interface) stores a typed-nil — so `mv.observer == nil` in `handleValidationSuccess` stays false even with highlighting off. 

I fear that quietly bypasses the c4ef00dbd guard, so `buildLoggerFields` (plus a `validatorStore` lookup at debug level) runs on every accepted message in the default config — the hot validation path. Maybe guard the option at the call site so the interface stays truly nil when disabled:

```go
opts := []validation.Option{validation.WithLogger(logger)}
if highlightedPeerObserver != nil {
    opts = append(opts, validation.WithSSVValidationObserver(highlightedPeerObserver))
}
```

network/topics/controller.go
```diff
@@ -308,10 +317,33 @@ func (ctrl *topicsCtrl) setupTopicValidator(name string) error {
...
+			result := validator(ctx, p, pmsg)
+			ctrl.peerObserver.ObserveValidation(ctx, ctrl.logger, p, name, validationResultString(result), len(pmsg.Data),
```
momosh-ssv: Might be worth gating this on `ctrl.peerObserver.Enabled()` first. Seems that `validationResultString(result)` and the `zap.Int` field get built for every pubsub message before `ObserveValidation` no-ops internally via `Match` — cheap, but it's per-message work unrelated to highlighting when the feature is off.

momosh-ssv (dismissed):

cli/operator/node.go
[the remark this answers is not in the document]
julienharbulot: Fixed in bf26928d7 by wrapping the peertrace observer with an SSV-validation adapter that returns a nil interface when highlighting is disabled. That keeps mv.observer truly nil in the default config and avoids the accepted-message hot-path work.

network/peers/peertrace/observer.go
julienharbulot: Fixed in bf26928d7 by moving the SSV-validation-specific adapter into network/peers/peertrace/ssvvalidation. The base peertrace package no longer imports message/validation or protocol/v2/message.

network/peers/connections/conn_gater.go
julienharbulot: Fixed in bf26928d7. The ip_rate_limit rejection path now keeps the previous "connection rejected due to IP rate limit" message and still includes the structured reason field.

network/topics/controller.go
julienharbulot: Fixed in bf26928d7 by removing the general pubsub received counter and its test from this PR, so the change stays scoped to highlighted-peer observability.

network/topics/pubsub.go
julienharbulot: Fixed in bf26928d7. Expanded RPC metadata is now appended only for highlighted peers, so existing PUBSUB_TRACE=true logs keep the slimmer schema unless highlighting is enabled for that peer.

network/topics/tracer.go
julienharbulot: Fixed in bf26928d7 by passing the pubsub/controller context into the tracer and score inspector for highlighted-peer metric/log paths instead of using context.Background().

network/peers/peertrace/observer.go
julienharbulot: Fixed in bf26928d7 by adding nil logger handling to Observer.Observe; ObserveValidation goes through that path, so the observer API is consistent.

network/peers/peertrace/observer.go
julienharbulot: Fixed in bf26928d7 by logging role_id with zap.Uint64, matching the RunnerRole type.

cli/operator/node.go
julienharbulot: Fixed in bf26928d7 by using an adapter that returns a nil SSVValidationObserver when highlighting is disabled, keeping the disabled-observer guard effective in the default config.

network/topics/controller.go
julienharbulot: Fixed in c27fd6f95 by checking ctrl.peerObserver.Enabled() before building validationResultString/zap fields and calling ObserveValidation.

network/topics/controller.go
```diff
@@ -286,6 +290,11 @@ func (ctrl *topicsCtrl) listen(sub *pubsub.Subscription) error {
 				messageTopicAttribute(topicNameFull),
 				messageTypeAttribute(uint64(m.MsgType)),
 			))
+			ctrl.peerObserver.Observe(ctrl.ctx, logger, "pubsub_message_delivered", msg.ReceivedFrom,
```
iurii-ssv: Perf, default path: the variadic `zap` fields (and the slice backing them) are evaluated *before* `Observe` no-ops internally, so this allocates per delivered gossip message on **every** node — even with highlighting disabled, since a nil/non-matching observer only returns early once called. This sits in the `listen()` hot loop.

Gate on a match first, e.g. `if _, ok := ctrl.peerObserver.Match(msg.ReceivedFrom); ok { ... }`. The same principle applies to the wrapped topic validator below — it is `Enabled()`-gated, so the disabled case is fine, but with the feature on it still builds per-message fields for all peers.

Fixed in #2881.

message/validation/errors.go
```diff
@@ -180,10 +184,50 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
...
 	return pubsub.ValidationReject
 }
 
-func (mv *messageValidator) handleValidationSuccess(ctx context.Context, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
+func (mv *messageValidator) handleValidationSuccess(ctx context.Context, peerID peer.ID, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
 	recordAcceptedMessage(ctx, messageRole(decodedMessage))
+	if mv.observer == nil {
+		return pubsub.ValidationAccept
+	}
+	mv.observeSSVValidation(ctx, peerID, mv.buildLoggerFields(decodedMessage), SSVValidationAccepted, "valid", nil)
```
iurii-ssv, line 202: Follow-up to the earlier `mv.observer == nil` guard (which handles the disabled case): with highlighting *enabled*, this still runs `buildLoggerFields` for every accepted message from **every** peer — and at debug log level that includes a validator-store lookup per message via `addDutyIDField`. Stage nodes (the target deployment for the attack simulator) typically run debug, and accepted messages are the hot path.

Consider a cheap `Interested(peer.ID) bool` on `SSVValidationObserver` so the peer match happens *before* field/event construction instead of at the end inside `ObserveSSVValidation`.

Implemented in #2881.

network/topics/tracer.go
```diff
@@ -35,6 +46,7 @@ func (pst *psTracer) log(logger *zap.Logger, evt *ps_pb.TraceEvent) {
 	fields := []zap.Field{
 		zap.String("type", evt.GetType().String()),
 	}
+	var highlightedPeer peer.ID
```
iurii-ssv: Perf (observer-only mode): when `PeerObserver` is enabled without `PubsubTrace`, the tracer gets registered, and `log()` eagerly builds fields — `peer.IDFromBytes`, hex encodings, the slice — for **every** pubsub event of **every** peer, only to drop them (no `Observe` match, `traceLog` false).

Extracting the event peer first and returning early (`if !pst.traceLog && !highlighted { return }`) keeps "highlight one peer" from taxing the whole gossip pipeline.

Done in #2881.

network/topics/tracer.go
```diff
@@ -84,25 +101,31 @@ func (pst *psTracer) log(logger *zap.Logger, evt *ps_pb.TraceEvent) {
...
 		fields = append(fields, zap.String("topic", msg.GetTopic()))
 	case ps_pb.TraceEvent_SEND_RPC:
 		msg := evt.GetSendRPC()
 		pid, err := peer.IDFromBytes(msg.GetSendTo())
 		if err == nil {
 			fields = append(fields, zap.String("targetPeer", pid.String()))
+			highlightedPeer = pid
 		}
-		if meta := msg.GetMeta(); meta != nil {
+		if meta := msg.GetMeta(); meta != nil && pst.highlighted(highlightedPeer) {
```
iurii-ssv: Regression for existing `PubsubTrace=true` operators: gating this whole block on `pst.highlighted(...)` also removed the **long-standing** IHAVE/IWANT and subscription metadata from regular-peer trace entries (here and in the DROP_RPC/RECV_RPC cases below). The earlier review round asked to keep the *new* metadata (messages/GRAFT/PRUNE) out of regular entries — but the pre-existing fields should stay, otherwise dashboards built on today's trace schema lose data.

#2881 restores the old schema for regular peers and keeps the verbose additions highlighted-only (with a test asserting both).

network/peers/connections/conn_gater.go
```diff
@@ -143,15 +192,53 @@ func (n *connGater) InterceptUpgraded(conn libp2pnetwork.Conn) (allow bool, reas
...
+	}
+
+	_, highlighted := n.peerObserver.Match(id)
+	recordConnectionGaterDecision(ctx, phase, decision, reason, direction, highlighted)
+
+	if id == "" {
+		return
+	}
+
+	logFields := make([]zap.Field, 0, 4+len(extraFields))
```
iurii-ssv, line 225: `highlighted` is already computed a few lines up, but `logFields` is still built for every gating decision with a non-empty peer ID — and only `Observe` (which re-matches internally and drops them for regular peers) consumes it. Dial/accept decisions fire constantly during discovery churn and floods, so this is steady waste.

`if !highlighted { return }` right after the metric record skips it (the `id == ""` check is subsumed, `Match` returns false for an empty ID).

Fixed in #2881.

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,233 @@
...
+		zap.String("p2p_highlight_label", o.label),
+		zap.String("p2p_highlight_event", event),
+		zap.String("peer_id", match.ID.String()),
+		zap.String("peer_source", match.Source),
+	}
+	if match.PublicKeyHex != "" {
+		highlightFields = append(highlightFields, zap.String("peer_public_key", match.PublicKeyHex))
+	}
+	highlightFields = append(highlightFields, fields...)
+	logger.Info("p2p highlighted peer event", highlightFields...)
```
iurii-ssv, line 167: Log volume: every highlighted-peer event logs at Info, and across trace events + delivered messages + pubsub validation + SSV validation that is roughly 3–5 lines per message from the highlighted peer. The stated use case is pointing this at a peer that *floods* the node — so unbounded logging becomes the bottleneck/disk-filler exactly during the experiment.

Suggest rate-limiting log emission while keeping the counters exact (metrics already capture volume; logs degrade to a sample). #2881 adds a 10/s, burst-100 limiter shared by `Observe` and the ssvvalidation adapter, counting metrics before the limiter so they stay accurate.

Or maybe we should just downgrade this from `Info` to `Debug` and we'll be fine ... I'm not 100% sure. Gotta Look into it some more.

cli/operator/network.go
```diff
@@ -55,7 +55,8 @@ func setupSSVNetwork(logger *zap.Logger, cfg *config) (*networkconfig.SSV, error
 
 		// Shallow-copy here before updating DomainType below — otherwise the update mutates the underlying global
 		// value ssvConfig is pointing to.
-		ssvConfig = new(*ssvConfig)
+		cfgCopy := *ssvConfig
```
iurii-ssv: Unrelated change: this rewrites the Go 1.26 `new(*ssvConfig)` expression (valid on `stage`; the repo requires Go 1.26) into the classic two-line copy. Semantically identical — presumably a local-toolchain artifact. Suggest dropping it to keep the PR scoped.

Reverted in #2881.

network/p2p/config.go
```diff
@@ -58,6 +59,14 @@ type Config struct {
 	PubSubTrace bool `yaml:"PubSubTrace" env:"PUBSUB_TRACE" env-description:"Enable pubsub debug tracing in logs"`
 	// DiscoveryTrace is a flag to turn on/off discovery tracing in logs
 	DiscoveryTrace bool `yaml:"DiscoveryTrace" env:"DISCOVERY_TRACE" env-description:"Enable discovery debug tracing in logs"`
+	// HighlightedPeers is a list of libp2p peer IDs that should be highlighted in p2p logs and metrics.
+	HighlightedPeers string `yaml:"HighlightedPeers" env:"P2P_HIGHLIGHTED_PEERS" env-default:"" env-description:"Comma, semicolon, or whitespace-separated libp2p peer IDs to highlight in p2p logs and metrics. 0x-prefixed secp256k1 public keys are still accepted for backwards compatibility, but P2P_HIGHLIGHTED_PEER_KEYS is preferred for keys"`
```
iurii-ssv: Wording nit: "still accepted for backwards compatibility" — this option is introduced in this very PR, so there is no released behavior to stay compatible with (the same phrase appears in the `peertrace.Config.Peers` doc). If it refers to an earlier internal stage build, say so explicitly; otherwise "also accepted for convenience" avoids implying history that doesn't exist.

Reworded in #2881.

network/peers/peertrace/observer.go
```diff
@@ -0,0 +1,233 @@
...
+	Source       string
+}
+
+type Observer struct {
+	label string
+	peers map[peer.ID]Peer
+}
+
+func New(cfg Config) (*Observer, error) {
+	tokens := splitPeerList(strings.Join([]string{cfg.Peers, cfg.PeerKeys}, " "))
```
iurii-ssv: Because both config fields are joined into one token list before parsing, a libp2p peer ID placed in `PeerKeys` is silently accepted as a peer ID — contradicting the field's name, description, and env-var docs. Parsing `PeerKeys` strictly (`parsePublicKey` only, error otherwise) catches that misconfiguration at startup instead of masking it.

Done in #2881 (with a test).

message/validation/errors.go
```diff
@@ -152,17 +152,20 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
...
 		recordIgnoredMessage(ctx, loggerFields.Role, validationCanceledReason)
+		mv.observeSSVValidation(ctx, peerID, loggerFields, SSVValidationIgnored, validationCanceledReason, err)
 		logger.Debug("ignoring message due to validation cancellation", zap.Error(err))
 		return pubsub.ValidationIgnore
 	}
 
 	var valErr Error
 	if !errors.As(err, &valErr) {
 		recordIgnoredMessage(ctx, loggerFields.Role, err.Error())
+		mv.observeSSVValidation(ctx, peerID, loggerFields, SSVValidationIgnored, err.Error(), err)
```
iurii-ssv: Metric cardinality: `err.Error()` flows into the new `ssv_validations` counter's reason attribute via the observer — arbitrary error strings make an unbounded OTel label. `recordIgnoredMessage` on the line above has the same pre-existing pattern, so no objection there, but for a newly introduced metric it is worth bounding the reason to a fixed value and letting the raw text travel on the event/log only (the event's `Error` field already carries it).

#2881 passes a fixed `unexpected_error` reason here.

message/validation/errors.go
```diff
@@ -180,10 +184,50 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
...
+		Outcome:        outcome,
+		Reason:         reason,
+		Role:           loggerFields.Role,
+		SSVMessageType: loggerFields.SSVMessageType,
+		Slot:           loggerFields.Slot,
+		DutyExecutorID: loggerFields.DutyExecutorID,
+		Signers:        loggerFields.Signers,
+	}
+	if loggerFields.Consensus != nil {
+		consensus := *loggerFields.Consensus
```
iurii-ssv: Consistency nit: `Consensus` is deep-copied here, but `DutyExecutorID`/`Signers` a few lines up alias the validator's slices — half-defensive copying. The observer consumes the event synchronously, so either stance works; pick one.

#2881 drops the copy and instead documents on `SSVValidationEvent` that events (including their reference fields) are only valid for the duration of the call.

iurii-ssv (commented):
Second pass over the current head, the inline comments below are **new items only**, in two themes plus polish:
- what the feature costs when it's **off** (default-path allocations in the gossip delivery loop), and
- what happens when it's **on** — per-message work for *regular* peers ([message/validation/errors.go](https://github.com/ssvlabs/ssv/blob/9a6b30f75d7b1bfa1c31d34b25bfc4480c8935be/message/validation/errors.go), [network/topics/tracer.go](https://github.com/ssvlabs/ssv/blob/9a6b30f75d7b1bfa1c31d34b25bfc4480c8935be/network/topics/tracer.go)), plus unbounded Info logging from the highlighted peer during the very flood it is meant to observe.

**Suggested resolution: #2881** — it is based on this branch and targets it, and implements every comment below + adds more tests; merging it resolves this review wholesale. Feel free to cherry-pick instead if you disagree with any individual piece. But review it carefully either way.

momosh-ssv:
Waiting on #2881 to land on this branch, and will re-review then

network/peers/peertrace/observer.go
iurii-ssv: Documenting the rationale and design discussion behind the rate-limiting suggestion (implemented in #2881), since "why add logs we then drop?" is a fair question:

**Why limit at all.** Counters and logs have different jobs here: counters answer *how much* (and are never limited), logs answer *what exactly* — topics, reasons, message IDs, QBFT fields. During a barrage the per-message lines are massively redundant: at ~5k msg/s from the simulator with 3–5 highlight log paths per message that's ~20k Info lines/s, on the order of a GB of JSON per minute. At that point logging *is* the experiment — we'd be measuring zap and the disk rather than the node's attack response — and a peer that knows it is highlighted gets a log-amplification lever. A 10/s sample still yields ~600 detailed exemplars per minute (more than anyone reads), while normal-rate debugging keeps full fidelity (burst 100 covers the interesting onset of each phase). So the qualitative signal survives sampling; the quantitative signal was never the logs' job.

**Where the current shape is admittedly blunt.** The limiter in #2881 is one global bucket per observer, so a pubsub-message flood can starve the *rare* event types — gater decisions, peer-score snapshots, stream events — which were never a volume problem and are exactly the lines you don't want to lose. Drops are also invisible (only inferable from counter-vs-log deltas), and the rate is a constant.

**Suggested refinement** (happy to push to #2881 if we agree):
1. **Per-event-type buckets** instead of one global one — a barrage then drowns only its own category; gater/score/stream lines keep flowing. (zap's built-in sampler keys on the message string, so it'd only help if the event name moved into the message; a small keyed-limiter map gets the same effect without reshaping the log schema.)
2. **Make drops visible** — a `logs_suppressed` counter so sampling is explicit.
3. **Make the rate configurable** — e.g. `P2P_HIGHLIGHTED_PEER_LOG_RATE`, default as today, `0` = unlimited, so a deliberately low-rate experiment can opt into full fidelity.

The defensible alternative is no limiter at all plus experiment discipline — but then I'd still keep a cap on the three per-message paths (pubsub validation, delivery, trace), since those are the ones that can hurt the node under the very flood it's meant to observe.

network/p2p/config.go
```diff
@@ -58,6 +59,15 @@ type Config struct {
...
+	// HighlightedPeers is a list of libp2p peer IDs that should be highlighted in p2p logs and metrics.
+	HighlightedPeers string `yaml:"HighlightedPeers" env:"P2P_HIGHLIGHTED_PEERS" env-default:"" env-description:"Comma, semicolon, or whitespace-separated libp2p peer IDs to highlight in p2p logs and metrics. 0x-prefixed secp256k1 public keys are also accepted for convenience, but P2P_HIGHLIGHTED_PEER_KEYS is preferred for keys"`
+	// HighlightedPeerKeys is a list of secp256k1 public identity keys whose derived libp2p peers should be highlighted in p2p logs and metrics.
+	HighlightedPeerKeys string `yaml:"HighlightedPeerKeys" env:"P2P_HIGHLIGHTED_PEER_KEYS" env-default:"" env-description:"Comma, semicolon, or whitespace-separated 0x-prefixed secp256k1 public identity keys to highlight in p2p logs and metrics"`
+	// HighlightedPeerLabel is attached to every highlighted peer log and metric.
+	HighlightedPeerLabel string `yaml:"HighlightedPeerLabel" env:"P2P_HIGHLIGHTED_PEER_LABEL" env-default:"highlighted-peer" env-description:"Label attached to highlighted peer logs and metrics"`
```
iurii-ssv, line 67: Maybe we should have a dedicated sub-section (eg. `HighlightedPeers`) where `Label`, `IDs`, `Pubkeys` would be defined (instead of just dumping it all the this level in config-hierarchy)

network/p2p/config.go
```diff
@@ -58,6 +59,15 @@ type Config struct {
...
+	// PeerObserver observes traffic for highlighted peers. The operator setup builds it
+	// from HighlightedPeers and HighlightedPeerKeys; nil disables highlighting.
+	PeerObserver *peertrace.Observer `yaml:"-"`
```
iurii-ssv, line 70: I don't think this member belongs on the config-struct, config should only facilitate yaml/env config parsing ... even if we already have something else on it - we shouldn't stuff even more on it ...

Lets see if we can refactor that ?

network/topics/pubsub.go
```diff
@@ -197,16 +199,16 @@ func NewPubSub(
...
+		psOpts = append(psOpts, pubsub.WithEventTracer(newTracer(ctx, logger, cfg.TraceLog, cfg.PeerObserver)))
```
iurii-ssv, line 203: We'll need similar handling for `Libp2pTrace` once https://github.com/ssvlabs/ssv/pull/2882 is in, I'm guessing

I'm not however sure whether `cfg.PeerObserver.Enabled() == true` should automatically turn on either of `Libp2pTrace` / `TraceLog` - those log way more data SSV node logs in DEBUG mode ...

iurii-ssv (commented):
Another review batch

alok-ssv (approved):

network/p2p/config.go
momosh-ssv: What do you think about doing this now rather than later? 
Since this hasn't shipped yet there's no migration cost — a nested `HighlightedPeers: {Label, IDs, PeerKeys}` block, keeping the existing `P2P_HIGHLIGHTED_*` env names on the leaf fields so the stage setup keeps working.

network/p2p/config.go
momosh-ssv: +1 on the direction. 

Seems that the cheap version fits this PR: `p2pv1.New` has a single production call site (`cli/operator/network.go:95`), so the observer could move to an explicit constructor param instead of the struct — that line plus a few lifecycle tests. 

The broader Config split (`Router`, `MessageValidator`, `NodeStorage` are in the same boat) feels like a separate follow-up though — maybe worth an issue?

network/topics/pubsub.go
momosh-ssv: I fear auto-enabling would be the wrong default — those traces dwarf DEBUG volume, as you say. 

Though seems the current wiring already avoids it: observer-only mode registers the tracer but `tracer.go` early-returns for non-highlighted peers, so only highlighted entries get emitted (rate-limited since #2881). 

Keeping the flags orthogonal and doing the same for `Libp2pTrace` once #2882 lands sounds right to me.

network/p2p/config.go
iurii-ssv: > What do you think about doing this now rather than later?

Yes, that's why I'm raising it :) Once we merge it - we'll need to support it, no way back. So, lets take the time to think it through @julienh-ssv @alok-ssv (how it fits into existing config structure, whether anything related needs to be refactored first to accommodate our changes in the best way possible, etc.).

message/validation/observer.go
```diff
@@ -0,0 +1,43 @@
...
+	SSVMessageType spectypes.MsgType
+	Slot           phase0.Slot
+	DutyExecutorID []byte
+	Signers        []spectypes.OperatorID
+	Consensus      *ConsensusFields
+	Error          string
+}
+
+// SSVValidationObserver receives SSV-level validation decisions.
+type SSVValidationObserver interface {
```
iurii-ssv, line 37: We have a bunch of types with `SSV...` prefix (or things like `ObserveSSVValidation`) in this PR, I wonder if it adds anything - seems like it is mostly redundant (adds clutter) ?

PR #2858, opened 2026-05-20
operator/duties: decouple validator-registration wire slot from local execution gate

iurii-ssv:
## What changed

Apply the same wire/execution decoupling pattern from #2857 (voluntary-exit) to event-driven validator-registration. The single `validatorRegistrationSlotsToPostpone` (= 4) becomes three constants:

- **`validatorRegistrationDutySlotsToPostpone` (= 4)** — the deterministic duty slot that goes on the wire and feeds the signed `ValidatorRegistration.Timestamp` (via `EpochStartTime(EstimatedEpochAtSlot(...))`). The value has been 4 since this constant was introduced; treated as a network-wide invariant going forward.
- **`validatorRegistrationSchedulingSlack` (= 4)** — the `slack` term in the execution offset below; absorbs per-operator timing variance once a registration event clears the EL follow distance. Independent of `validatorRegistrationDutySlotsToPostpone` despite sharing the value 4.
- **`validatorRegistrationExecutionSlotsToPostpone` (= `FollowDistance + slack` = 12)** — purely local "earliest broadcast" gate. Mirrors voluntary-exit to shrink the retry-window race where a fast-EL operator broadcasts before slow-EL peers have started a runner for the duty.

Event-driven duties now enter an `eventQueue` and are drained by the existing `processExecution` ticker hook (the periodic path is unchanged — its slot is already the current slot, with no race to defer past). `Scheduler.ExecuteDuties` special-case extended to skip the late-execution warning and slot-delay histogram for validator-registration too (same reason as voluntary-exit: `duty.Slot` is the shared duty slot, not the execution slot).

Also fixes the stale "validator exit events" copy-paste reference in `blockSlot`'s docstring, and rewrites the misleading "ensure the duty is scheduled with a slot number never in the past" comment that was a lie even before this PR (with the EL log stream's 8-block follow distance, `blockSlot + 4` was always in the past on receipt).

## Why

Validator-registration is structurally exposed to the same EL-streaming-lag issue as voluntary-exit (flagged as a cleanup on #2857's follow-up): the existing `+4` is less than `FollowDistance = 8`, so the dispatched duty's `Slot` is always in the past. Three independent effects on `ValidatorRegistration`:

1. **Signed `Timestamp` cross-operator agreement**: every operator must derive the same epoch from `dutySlot` so the signed object hashes the same way for BLS aggregation. The wire constant (still 4) was deterministic before and stays deterministic now — but the wire role is explicit and the constant is guarded by a pin test.
2. **Cross-operator broadcast race**: a fast-EL operator's partial-sig can arrive at a slow-EL peer before that peer has started a runner, surfacing as `ErrNoDutyAssigned` → `RetryableError`. The queue retries for ~1 slot; beyond that, the message is dropped (and only the periodic VRSubmitter loop catches up). The execution gate defers our broadcast to the same `blockSlot + FollowDistance + slack` mark that voluntary-exit uses, reducing the race window.
3. **Wire-format invariant**: keeping the slot pinned at `4` means a node running this code remains validateable by peers running any prior code (the `Timestamp` is derived from the same epoch).

Unlike voluntary-exit, validator-registration's inbound validation does not lock to a per-slot `dutyStore` key (the dutyLimit is a constant 2), so the race is less catastrophic — the periodic `VRSubmitter` loop will resubmit eventually. The gate is here for consistency with the voluntary-exit pattern and the modest race-window reduction; the docstring on `validatorRegistrationExecutionSlotsToPostpone` calls that out so a future reader sees the trade-off.

## Impact

- Event-driven registrations now broadcast at `blockSlot + FollowDistance + 4` (~144 s after the originating EL block) instead of immediately on receipt at ~`blockSlot + FollowDistance` (~96 s) — i.e. ~48 s of added latency. This adds latency to fee-recipient propagation to relays; in practice the periodic submitter (every slot, hash-distributed across the epoch) was already the eventual-consistency mechanism, so the visible delay is bounded.
- `ValidatorRegistration.Timestamp` reflects the epoch of `blockSlot + 4` — same as before this PR.
- Periodic registration path is untouched: same `Slot = currentSlot`, same firing behaviour.
- Hidden side-effect fix: pre-2858 passed `dutyExecutionDeadline(dutySlot)` = `SlotStartTime(blockSlot + 5)` to `ExecuteDuties` for event-driven registrations. With `FollowDistance = 8` making `currentSlot ≥ blockSlot + 8` by the time the EL log stream surfaces the event, that deadline was already in the past, so the runner's `context.WithDeadline(...)` was immediately canceled. Post-2858 uses `dutyExecutionDeadline(currentTickSlot)`, giving the runner a real ~12 s wall-clock window.
- `scheduler.executions` counter still tracks validator-registration dispatches; the `scheduler.slot_ticker_delay.duration` histogram no longer receives the (intentionally meaningless) ~96-second values for event-driven registrations. Periodic registrations would have recorded ~0 values; those are also skipped, but operationally that's the same as recording zero (the histogram point carries no signal in either case).

## Verification

- `go test ./operator/duties -count=1`
- New `event-driven duty deferred until execution gate` subtest ticks one slot before the gate (expects no execution) then at the gate (expects execution), pinning the deferred-broadcast semantics.
- New `multiple event-driven duties drained together` subtest covers the queue draining behaviour when more than one event is pending past the gate.
- New `TestValidatorRegistrationDutySlotPinned` asserts `validatorRegistrationDutySlotsToPostpone == 4`, surfacing any literal drift as a visible diff during PR review.

## Notes

- Stacked on #2857 (voluntary-exit wire/execution decoupling). Best merged after #2857; otherwise the scheduler/observability hooks this PR extends won't exist on `stage`.


iurii-ssv:
Re [Greptile review](https://github.com/ssvlabs/ssv/pull/2858#issuecomment-4496154083), both P2s addressed:

1. **Stale `dutyExecutionDeadline` docstring** — already fixed in [`72ee06453`](https://github.com/ssvlabs/ssv/pull/2858/commits/72ee06453), pre-review. Greptile's "Last reviewed commit" was [`86617b362`](https://github.com/ssvlabs/ssv/pull/2858/commits/86617b362), one commit behind. Current state mirrors the same clarification I applied to voluntary-exit in [`4473aad85`](https://github.com/ssvlabs/ssv/pull/2857/commits/4473aad85): `slot` is the firing slot (current ticker slot at dispatch), not `duty.Slot` (the wire slot held in the past for event-driven registrations).

2. **No dispatch-time log for drained event-driven duties** — fixed in [`f1ea140ee`](https://github.com/ssvlabs/ssv/pull/2858/commits/f1ea140ee). Added a count-summary log in `processExecution` right after the queue drain, mirroring voluntary-exit's `"executed voluntary exit duties"` pattern. The per-duty descriptor stays in the enqueue log (`"🛠 scheduled validator registration duty for execution"`) — the dispatch log is just `slot + count`, sufficient to correlate the two ends of the deferred-broadcast flow in operator logs.

iurii-ssv:
@greptile pls re-review

iurii-ssv:
@greptile pls review

operator/duties/validator_registration.go
```diff
@@ -97,18 +165,22 @@ func (h *ValidatorRegistrationHandler) HandleDuties(ctx context.Context) {
...
-				ValidatorIndex: regDescriptor.ValidatorIndex,
-				PubKey:         regDescriptor.ValidatorPubkey,
-				Slot:           dutySlot,
-			}}, h.dutyExecutionDeadline(dutySlot))
-			h.logger.Debug("validator registration duty sent",
-				zap.Uint64("slot", uint64(dutySlot)),
+			dutySlot := blockSlot + validatorRegistrationDutySlotsToPostpone
+			earliestExecutionSlot := blockSlot + validatorRegistrationExecutionSlotsToPostpone
+
+			h.eventQueue = append(h.eventQueue, &queuedRegistration{
```
momosh-ssv, line 175: Should we de-dup on enqueue here? 
Unlike voluntary-exit (effectively one-shot per validator), registrations are recurring.
That means redundant partial-sig broadcasts plus queue memory proportional to event churn. 

Maybe key on (ValidatorIndex, dutySlot) and keep the latest descriptor — or, if duplicates are fine because the periodic loop and the runner's dutyLimit=2 already bound things, worth a short comment saying so.

iurii-ssv: Good call-out — went with your Option B (documented, no code de-dup); comment added at the enqueue site in 4b5a80a.

Reasoning:
- The enqueued duty carries no fee recipient (the runner reads the current one at execution time via `feeRecipientProvider.GetFeeRecipient`), so multiple entries sharing a `(ValidatorIndex, dutySlot)` produce byte-identical registrations — redundant rather than conflicting.
- They're bounded downstream by the receiver's `dutyLimit=2`/epoch and reconciled by the periodic `VRSubmitter` loop; the queue drains every slot (items live ≤ `validatorRegistrationExecutionSlotsToPostpone`), and the producer (`Controller.UpdateFeeRecipient`) already throttles to `scheduledValidatorRegsLimit` per owner.

One caveat that steered me away from the code path: a de-dup keyed on `ValidatorIndex` **alone** would be unsafe — "keep latest" is local-queue-timing-dependent, so operators could diverge on `dutySlot` and break partial-sig aggregation. Your exact key `(ValidatorIndex, dutySlot)` is safe, but it only collapses two events for the same owner in the *same EL block* (rare), so the extra code didn't seem worth the surface. I documented the determinism constraint inline so a future "simplify to per-index" doesn't quietly reintroduce it.

julienharbulot (approved):
LGTM after fixing momo's comments

alok-ssv (approved):
LGTM

momosh-ssv (approved):

operator/duties/validator_registration.go
momosh-ssv: ok, nice.
Im down with this approach

PR #2675, opened 2026-01-27
sync-committee-contribution: use the correct root(s) when calculating SyncCommitteeSubnetID

iurii-ssv:
As per this find - https://github.com/ssvlabs/ssv/pull/2674#discussion_r2732719892 - we need to find correct/reliable way to calculate `SyncCommitteeSubnetID`

PR #2859, opened 2026-05-22
v2.4.3: verify & test QA findings

iurii-ssv:
## Summary

Address findings from the v2.4.3 ↔ Anchor v1.2.3 cross-client interop QA work. One small production change in the SCC-aggregator runner that canonicalizes `Contributions` ordering, plus five new test files (one integration-level) and a shared runner test-kit refactor, pinning behavior at the boundaries we investigated.

## Production change

**Canonicalize SCC-aggregator contribution ordering** — `protocol/v2/ssv/runner/sync_committee_contribution.go`.

`ProcessPreConsensus` now reduces the pre-consensus quorum roots to `(subnet, selection-proof)` pairs (a new `subnetSelectionProof` slice — pairing the two values in one slice makes the `(subnet, proof)` pairing unbreakable by construction), sorts them ascending by subnet via `sortBySubnet`, then splits them back into the parallel `selectionProofs`/`subnets` slices at the `GetSyncCommitteeContribution` call boundary.

Sorting by subnet alone is a total order here, so the (unstable) `slices.SortFunc` needs no tiebreaker: each pre-consensus root is the hash of `SyncAggregatorSelectionData{Slot, SubcommitteeIndex: subnet}`, so two sync-committee indices that map to the same subnet collapse to a single root — the `subnets` slice never holds duplicates.

Without this, the upstream `roots` slice from `basePreConsensusMsgProcessing` (in `ssv-spec`) is collected via `slices.Collect(maps.Keys(...))` — random per-process map iteration. Two SSV nodes running the same duty produce identical sets but in different orders, so the `Contributions` SSZ root proposed by the QBFT leader varies depending on which node leads. The fix produces a deterministic, spec-canonical (ascending `SubcommitteeIndex`) order matching `ssv-spec`'s canonical test fixtures and Anchor's producer.

The ordering non-determinism stems from `roots` being collected via `slices.Collect(maps.Keys(quorumRoots))` (map-iteration order) in `basePreConsensusMsgProcessing` — that is the underlying source, and it predates #2675. #2675 (Jan 2026) was a **correctness fix**, not the cause: it fixed a proof↔subnet *mispairing* bug (flagged in the [#2674 review](https://github.com/ssvlabs/ssv/pull/2674#discussion_r2732719892)) where, for validators holding 2+ sync-committee positions in different subnets, a selection proof for subnet A was paired with subnet B's contribution → the beacon node rejected the aggregate. Pre-#2675 the output happened to be deterministically ordered (stable `ValidatorSyncCommitteeIndices` order) but mispaired; #2675 made the pairing correct via a `root → ValidatorSyncCommitteeIndex` map, which surfaced the pre-existing map-order non-determinism in the output. The non-determinism has stayed invisible to `ssv-spec` CI because `TestingBeaconNode.GetSyncCommitteeContribution` discards its `subnetIDs` argument and always returns the canonical pre-sorted fixture. This PR sorts the paired output ascending-by-subnet — making it correctly paired *and* deterministic, completing the fix originally suggested in the #2674 review (at the output stage rather than the roots stage).

**Backward compatible** with older SSV versions in production: receivers (old or new SSV, Anchor) only verify SSZ-decodability of `Contributions`, not ordering. A sorted-emit proposer interoperates with any follower regardless of version. The change only affects what bytes an SSV node emits when it's the QBFT leader; it doesn't change what SSV accepts.

## Tests

Five new test files plus a supporting refactor, all passing under `-race`:

1. **`protocol/v2/ssv/runner/sync_committee_contribution_test.go`** — `TestSortBySubnet`, 6 sub-tests covering: sort ascending + preserve pairing, determinism across 100 repeated runs, spec-canonical order, empty slice, single element, already-sorted. (The paired-slice design removed the old two-parallel-slices length invariant, so there is no longer a mismatched-length panic path to test.)

2. **`protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_test.go`** — `TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall`, an integration-level guard that drives a full pre-consensus quorum through `ProcessPreConsensus` and asserts the subnets actually handed to `GetSyncCommitteeContribution` are in ascending (spec-canonical) order. This pins the fix at the wiring boundary — a helper-only test can't catch a dropped or misplaced sort call, and `ssv-spec`'s runner tests can't either (their `TestingBeaconNode.GetSyncCommitteeContribution` discards `subnetIDs` and returns a pre-sorted fixture).

3. **`protocol/v2/ssv/runner/sync_committee_contribution_ssz_test.go`** — `TestContributionsSSZEncodingMatchesSpec`, pins `ssv-spec`'s current SSZ encoding for `Contributions`: variable-length-list framing (per-element 4-byte offsets) on a list whose element type is fixed-length 256 bytes. This is non-canonical SSZ but is what the spec defines; Anchor's `ContributionWrapper` matches. Regression guard: any future `ssv-spec` change to canonical-SSZ framing breaks this test in CI before a wire-format mismatch can reach production.

4. **`protocol/v2/blockchain/beacon/blind/blind_fulu_electra_equiv_test.go`** — `TestEnsureBlinded_FuluProducesElectraEquivalentBytes`, verifies that SSV's reuse of `apiv1electra.BlindedBeaconBlock` for Fulu produces SSZ bytes byte-identical to the equivalent Electra-versioned blinded block. Confirmed at the Lighthouse source level: `BeaconBlockBodyFulu` and `BeaconBlockBodyElectra` differ in exactly one type label (`Payload::Fulu` vs `Payload::Electra`), and for blinded blocks that label resolves to `ExecutionPayloadHeader{Fulu,Electra}` superstruct variants with byte-identical SSZ schemas. Regression guard: any future Fulu-specific divergence in `blind.go` surfaces here before breaking interop.

5. **`protocol/v2/ssv/value_check_test.go`** — `TestVoteCheckerSourceTargetEpoch`, pins `value_check.go:47`'s current `source.Epoch >= target.Epoch` behavior. Two sub-tests: genesis case (`source == target == 0`, currently rejected — pinned with a `TODO(B1 fix)` marker to flip when the spec fix lands) and the `source > target` regression-coverage case (unchanged by the spec fix). The SSV-Go one-character mirror fix is deliberately held back until spec-maintainer direction is confirmed on the upstream spec issue (linked below), to avoid SSV drifting from the canonical spec unilaterally.

**Supporting refactor (no production behavior change):** extracts the shared single-validator runner scaffolding into `protocol/v2/ssv/runner/runner_testkit_test.go` (`newRunnerTestKit`, `testingRoundTimerF`) and rewires `proposer_test.go` and `committee_test.go` onto it, so the new SCC integration test reuses the same setup instead of duplicating it.

## Related upstream issues

- [ssvlabs/ssv-spec#631](https://github.com/ssvlabs/ssv-spec/issues/631) — spec issue for the `source >= target` over-strict check (rejects spec-legal genesis-epoch attestations). The SSV-Go mirror fix at `value_check.go:47` will land in a follow-up once #631 is resolved.
- [sigp/anchor#1050](https://github.com/sigp/anchor/issues/1050) — Anchor `InvalidPartialSignatureTypeCount` outbound-drop, root cause of the cross-client SCC-aggregator pattern observed during QA. Anchor's producer-side fix in flight at [sigp/anchor#1052](https://github.com/sigp/anchor/pull/1052).
- [sigp/anchor#1051](https://github.com/sigp/anchor/issues/1051) — Anchor `SlotAlreadyAdvanced` outbound-drop at slot boundaries (companion finding).


protocol/v2/ssv/value_check_test.go
```diff
@@ -0,0 +1,68 @@
...
+		err := checker.CheckValue(encodeBeaconVote(t, 0, 0))
+		require.ErrorContains(t, err, "source >= target")
+		// TODO(B1 fix): once the fix lands, this assertion should flip — at minimum
+		// the source/target gate should no longer fire on (0, 0).
+	})
+
+	t.Run("source.Epoch > target.Epoch is rejected (unchanged by the B1 fix)", func(t *testing.T) {
+		checker := NewVoteChecker(nil, 0, []phase0.BLSPubKey{}, &spectypes.BeaconVote{})
+		err := checker.CheckValue(encodeBeaconVote(t, 1, 0))
+		require.ErrorContains(t, err, "source >= target")
```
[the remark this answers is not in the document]
iurii-ssv: This is intentional and deliberate — the SSV-Go one-character mirror fix (`>=` → `>`) is held back until [`ssvlabs/ssv-spec#631`](https://github.com/ssvlabs/ssv-spec/issues/631) gets spec-maintainer direction, to avoid SSV-Go drifting from the canonical spec unilaterally. The same rule lives in `ssv-spec` and is what *both* SSV-Go and Anchor inherit; fixing it in SSV-Go alone would create a temporary spec-vs-impl divergence that the spec issue is meant to resolve in one go.

The `TODO(B1 fix)` marker is the pin that flags the flip needed once the spec PR lands — at that point the production code change and the test-assertion flip will land atomically in a follow-up PR.

Production impact of holding back: near-zero on mainnet (no fresh-genesis bootstrapping), but every fresh-genesis cross-client interop test will continue to surface this in epoch 0. That's the same trade-off the docstring describes.

Keeping as-is.

protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -626,3 +632,33 @@ func (r *SyncCommitteeAggregatorRunner) GetRoot() ([32]byte, error) {
...
+// is built via `slices.Collect(maps.Keys(...))`, which Go randomises per process. Two SSV
+// nodes can then produce different Contributions SSZ roots for the same logical
+// contribution set. The spec's de-facto canonical ordering is ascending SubcommitteeIndex
+// (see ssv-spec/types/testingutils/beacon_node_sync_committee.go test fixtures).
+func sortContributionsBySubnet(subnets []uint64, selectionProofs []phase0.BLSSignature) {
+	if len(subnets) != len(selectionProofs) {
+		panic(fmt.Sprintf("sortContributionsBySubnet: paired slices must be the same length (subnets=%d, proofs=%d)", len(subnets), len(selectionProofs)))
+	}
+	sort.Stable(subnetSorter{subnets: subnets, selectionProofs: selectionProofs})
+}
```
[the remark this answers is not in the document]
iurii-ssv: The panic guards a paired-slice invariant that is enforced one branch above by the caller:

```go
for _, root := range roots {
    ...
    selectionProofs = append(selectionProofs, blsSigSelectionProof)
    subnets = append(subnets, subnet)
}
```

Each iteration appends to both slices in lockstep. The only way they can have different lengths is a programmer-introduced refactor that breaks the invariant — i.e. a bug, not a runtime condition.

The Go idiom for this kind of check is `panic` (programmer-invariant violation → crash early, log line points at the broken caller). Returning an error would force every caller to handle a case that can't legitimately fire, and could mask the underlying bug if a future caller just logs-and-continues instead of crashing.

That said — agreed the panic is fragile if the caller ever evolves (e.g. an early `continue` skips one append). Two ways to harden:

1. Move the invariant check up to be alongside the append loop, where the relationship is locally visible.
2. Lift the loop into a small helper that returns `(subnets, selectionProofs)` and guarantees the pairing by construction (no separate slices to keep in sync).

Neither feels worth the churn for an invariant that's locally obvious — the loop is 3 lines apart from the sort call. Keeping the panic, with a clear panic message that names the invariant.

protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -144,6 +145,11 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
 		subnets = append(subnets, subnet)
 	}
 
+	// Sort paired (subnets, selectionProofs) by ascending subnet so the resulting
+	// Contributions slice has a deterministic, spec-canonical order. See
+	// sortContributionsBySubnet for the full rationale.
+	sortContributionsBySubnet(subnets, selectionProofs)
```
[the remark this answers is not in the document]
iurii-ssv: This is based on a misreading of the runner model — duplicate subnets can't occur in this code path. Walking through it:

[`SyncCommitteeAggregatorRunner`](https://github.com/ssvlabs/ssv/blob/v2.4.3-qa-findings/protocol/v2/ssv/runner/sync_committee_contribution.go) is **per-validator** (one `share` per runner via `r.GetShare()`). The runner builds its `rootToSyncCommitteeIdx` mapping at [`sync_committee_contribution.go:527-551`](https://github.com/ssvlabs/ssv/blob/v2.4.3-qa-findings/protocol/v2/ssv/runner/sync_committee_contribution.go#L527-L551):

```go
r.rootToSyncCommitteeIdx = make(map[phase0.Root]phase0.ValidatorIndex)
for _, vIdx := range validatorDuty.ValidatorSyncCommitteeIndices {
    subnet := r.GetBeaconNode().SyncCommitteeSubnetID(phase0.CommitteeIndex(vIdx))
    data := &altair.SyncAggregatorSelectionData{
        Slot:              validatorDuty.DutySlot(),
        SubcommitteeIndex: subnet,
    }
    msg, err := signBeaconObject(ctx, r, ..., data, ...)
    ...
    r.rootToSyncCommitteeIdx[msg.SigningRoot] = phase0.ValidatorIndex(vIdx)
}
```

The `SyncAggregatorSelectionData` being signed contains **only** `Slot` and `SubcommitteeIndex` — no `vIdx`. So if a single validator has multiple `ValidatorSyncCommitteeIndices` falling into the same subnet (e.g. positions 5, 50, 100 all in subnet 0 = floor(idx/128)), the SSZ signing root is **byte-identical** across those iterations.

That means:
- The map only retains the last-iterated `vIdx` for that root (single entry).
- The pre-consensus quorum container deduplicates partial signatures by root — only one `root` ends up in the `roots` slice for that subnet.
- The downstream loop at [line 137-145](https://github.com/ssvlabs/ssv/blob/v2.4.3-qa-findings/protocol/v2/ssv/runner/sync_committee_contribution.go#L137-L145) produces one `(subnet, proof)` entry per unique subnet for this validator.

No duplicate subnets, no ties for `sort.Stable` to leave under-determined. The fix is sufficient as-is.

That said, the suggestion has merit as defensive future-proofing — if the runner architecture ever changes to multi-validator (e.g. some future committee-level SCC-aggregator), duplicates could appear. Happy to add a tiebreak on the BLS sig bytes if you'd prefer, but flagging that it's hardening against a hypothetical the current model doesn't produce.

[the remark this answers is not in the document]
iurii-ssv: One framing correction for the record, since this thread touches the ordering non-determinism: I'd earlier described #2675 as having "introduced" the non-determinism. That's not accurate. The non-determinism's actual source is `roots` being collected via `slices.Collect(maps.Keys(quorumRoots))` (map-iteration order), which predates #2675.

#2675 was a **correctness fix** — it fixed a proof↔subnet mispairing bug ([flagged here](https://github.com/ssvlabs/ssv/pull/2674#discussion_r2732719892)) where validators holding 2+ sync-committee positions in different subnets would pair a selection proof for subnet A with subnet B's contribution, causing the beacon node to reject the aggregate. Pre-#2675 the output was deterministically ordered but mispaired; #2675 made the pairing correct and in doing so surfaced the pre-existing map-order non-determinism. This PR's sort completes the fix the #2674 reviewer originally suggested (deterministic order), now applied at the output stage. Net: #2675 was the right call; this PR removes its residual cosmetic cost.

iurii-ssv:
@greptile pls re-review

protocol/v2/ssv/runner/sync_committee_contribution.go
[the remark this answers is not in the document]
iurii-ssv: Simplified panic away in https://github.com/ssvlabs/ssv/pull/2859/changes/753b94a508f0a66c0a5ece15165b2e1856420b50

protocol/v2/blockchain/beacon/blind/blind_fulu_electra_equiv_test.go
```diff
@@ -0,0 +1,146 @@
...
+		Body:          makeBody(),
+	}
+
+	electraProposal := &api.VersionedProposal{
+		Version: spec.DataVersionElectra,
+		Electra: &apiv1electra.BlockContents{Block: blk},
+	}
+	fuluProposal := &api.VersionedProposal{
+		Version: spec.DataVersionFulu,
+		Fulu:    &apiv1fulu.BlockContents{Block: blk},
```
momosh-ssv, line 96: Could this be near-tautological as written? 
Both `EnsureBlinded` branches construct the same concrete type (`apiv1electra.BlindedBeaconBlock`), and the test feeds the same `blk` pointer into both proposals — so byte-equality looks guaranteed by construction (`MarshalSSZ` is the same method on the same type). 

It would catch a regression where the Fulu branch switches to a different struct type, but a field-level divergence inside the shared type would stay invisible. 
Might be worth building the Fulu body from an independently-constructed value so a field-copy mistake in one branch surfaces — or softening the doc comment to state plainly what it actually guards ("Fulu branch still returns the Electra type").

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	require.Equal(t, 1, testBeacon.getContributionCalls,
+		"GetSyncCommitteeContribution should be called exactly once, when pre-consensus quorum is reached")
+
+	// TestingContributionProofIndexes = {0,1,2} map to subnets {0,1,2}; with the fix
+	// the runner hands them to the beacon node in ascending order regardless of the
+	// random map-iteration order of the upstream roots slice. Without the sort this
+	// assertion would observe a random permutation and fail.
+	require.Equal(t, []uint64{0, 1, 2}, testBeacon.capturedSubnets,
```
momosh-ssv: Worth considering whether asserting the exact `{0,1,2}` set couples the test to undocumented spec-fixture invariants — it holds only because the fixture's `IsSyncCommitteeAggregator` passes all three proofs and `SyncCommitteeSubnetID` maps {0,1,2}→{0,1,2}. 

A future ssv-spec bump that changes fixture selection would fail this opaquely. The adjacent `slices.IsSorted` check is the property that actually matters for the determinism fix — maybe lean on that as the primary assertion and keep `{0,1,2}` as a sanity check with a comment pointing at the fixture source.

protocol/v2/ssv/runner/runner_testkit_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// newRunnerTestKit builds the common single-validator runner scaffolding for role,
+// wiring bn as the runner's beacon node. If cfg is nil a fresh cloneTestNetworkConfig()
+// is used, so parallel tests can mutate beacon timing without racing on the shared global.
+func newRunnerTestKit(t *testing.T, role spectypes.RunnerRole, bn beacon.BeaconNode, cfg *networkconfig.Network) *runnerTestKit {
+	t.Helper()
+
+	if cfg == nil {
+		cfg = cloneTestNetworkConfig()
```
momosh-ssv, line 42: Might be worth a note here that `cloneTestNetworkConfig` is a shallow clone — it's safe today only because tests mutate value fields (beacon timing) on the copy. 

A future test that mutates a pointer/map/slice field on the cloned `Beacon`/`Network` would alias the shared global and could race under `-race`. No impact now; flagging so the constraint is visible to the next person who reaches for this kit.

protocol/v2/blockchain/beacon/blind/blind_fulu_electra_equiv_test.go
iurii-ssv: Good catch on the docstring overselling — tightened in 051a81e70. Two clarifications:

The test isn't fully tautological: the Electra and Fulu arms of `EnsureBlinded` are *separately duplicated* field-copy blocks ([blind.go#L151-L173](https://github.com/ssvlabs/ssv/blob/051a81e70/protocol/v2/blockchain/beacon/blind/blind.go#L151-L173) / [#L186-L208](https://github.com/ssvlabs/ssv/blob/051a81e70/protocol/v2/blockchain/beacon/blind/blind.go#L186-L208)), not a shared helper — a field dropped or reordered in one arm but not the other *does* change the bytes and trips this test.

The "build Fulu from an independent value" route isn't reachable here: `apiv1fulu.BlockContents.Block` is typed `*electraspec.BeaconBlock` (which is *why* the same `blk` assigns to both proposals), so there's no distinct Fulu type to construct from.

So I went with your second option: the docstring now states only what it pins (the two branches stay byte-equal) and explicitly flags that it does **not** re-verify the Lighthouse-level claim. Also dropped the redundant `t.Logf` trio.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_test.go
iurii-ssv: Agreed — reworked in 051a81e70. `slices.IsSorted` is now the headline assertion (the actual determinism invariant the fix guarantees), and `{0,1,2}` is kept as an explicitly-labeled exact-set sanity check that points at the fixture source. A future ssv-spec fixture bump now trips that line with an "update the expectation" signal instead of looking like the sort regressed.

protocol/v2/ssv/runner/runner_testkit_test.go
iurii-ssv: Good call — addressed in 051a81e70 rather than deferred. I relocated `cloneTestNetworkConfig` into the shared runner test-kit (it's used package-wide now, not just by the proposer test) and gave it a [doc-comment](https://github.com/ssvlabs/ssv/blob/051a81e70/protocol/v2/ssv/runner/runner_testkit_test.go#L91-L104) that spells out exactly this constraint: only the `Beacon` value and its `Forks` map are deep-copied, and any other pointer/map/slice field still aliases the global — so a test that mutates one must extend the clone first.

julienharbulot (approved):
LGTM after fixing momo's comments

momosh-ssv (approved):
all good now, gj!

PR #2860, opened 2026-05-26
api, network/peers: prevent panic on peer NodeInfo without metadata

iurii-ssv:
## Summary

A peer sending a NodeInfo envelope without a Metadata block (only ForkVersion + NetworkID, per [`UnmarshalRecord`'s <3-entries path](https://github.com/ssvlabs/ssv/blob/stage/network/records/node_info.go#L117-L119)) lands in the peers index with `Metadata == nil`. The [`/v1/node/peers`](https://github.com/ssvlabs/ssv/blob/stage/api/handlers/node/node.go#L152) and `/v1/node/health` handlers then nil-deref on `nodeInfo.Metadata.NodeVersion` in `(*Node).peers`, panicking once per probe per such peer.

Mainnet is currently hit by one peer at `51.161.197.214` (libp2p ID `16Uiu2HAmUiHi…CYgo`) that connects to every operator and sends 2-entry NodeInfo records on every handshake. Local stage code never produces 2-entry records (`setupPeerServices` always sets `self.Metadata`), so this is a remote client — exact identity unknown without the `agent_version` field this PR also adds.

The unguarded dereference at `node.go:152` has existed since [`60f386fec4`](https://github.com/ssvlabs/ssv/commit/60f386fec4) (Jul 2023) — it's a latent bug, and the recent panics are because of a new misbehaving peer in the cluster, not a regression on our side.

## Core fix

- **`api/handlers/node/node.go`**: guard `nodeInfo.Metadata.NodeVersion` with a nil check so the API can't panic regardless of what's in the index. The peer still appears in the response with `Version` left empty.
- **`network/peers/connections/handshaker.go`**: reject `nil-Metadata` in `verifyTheirNodeInfo` before `SetNodeInfo`, so such records never get stored. The handshake error propagates through the existing `defer-updatePeerInfo` path; `conn_handler` then disconnects the peer (same outcome as the existing subnet-share rejection, just earlier).
- **AgentVersion logging**: new `peerAgentVersion()` helper queries libp2p's identify-protocol `AgentVersion` from the peerstore; added as `agent_version` to both the existing `"Verified handshake nodeinfo"` INFO log and the new rejection WARN log. Lets us identify misbehaving clients in production.

## Consistency / completeness

- `Identity` handler gets the same nil-Metadata guard for symmetry with the `peers` handler.
- Existing `if Metadata != nil` checks at `handshaker.updateNodeSubnets` and `p2p/observability.go` get `// invariant:` comments documenting why they remain after the handshake-side rejection.
- **`api/server`**: replace `chi.Recoverer` with a zap-based `middlewareRecover` that logs panics at ERROR level with `method`/`path`/`remote_addr`/`panic`/`stack` as structured fields. `chi.Recoverer` writes the stack to stderr via `debug.PrintStack` and can't be correlated with the triggering request in log aggregators — which is what made the original investigation slow. Matches `chi`'s `http.ErrAbortHandler` semantics.
- **`network/peers/connections/mock`**: `MockPeerstore.Get` gains a `MockValues map[peer.ID]map[string]any` for per-key lookups, so handshake `AgentVersion` tests can exercise the real identify-protocol path. Legacy "return peer.ID for any key" fallback preserved for existing tests.

## Tests

| Test | What it covers |
|---|---|
| `TestPeers_SkipsNilMetadata` | `peers` + `health` handlers don't panic on nil-Metadata NodeInfo |
| `TestIdentity_SkipsNilMetadata` | `identity` handler symmetric behavior |
| `TestVerifyTheirNodeInfo_RejectsNilMetadata` | Handshake rejects + doesn't store nil-Metadata + handshake error propagates |
| `TestPeerAgentVersion` | 4 sub-cases: nil net, non-string value, configured string (via `MockValues`), peerstore error |
| `TestMiddlewareRecover` | Panic capture + 500 + `http.ErrAbortHandler` rethrow + passthrough |

Pre-existing `TestHandshakeTestData/happy_flow` continues to pass with `agent_version` added to its observed log.

## Verification

Locally on the touched packages:

- ✅ `go build ./...` and `go vet ./...`
- ✅ `gofmt` clean
- ✅ `golangci-lint run` — 0 issues
- ✅ `openapi-lint` — 0 issues
- ✅ All tests pass: `./api/handlers/node/...`, `./api/server/...`, `./network/peers/connections/...`

`./network/p2p/...` has a pre-existing flaky goroutine-leak issue with `libp2p/zeroconf` unrelated to this change — confirmed it reproduces on clean `stage` as well.

## Operational follow-up

Independent of this PR — the offending peer (`51.161.197.214`, libp2p ID `16Uiu2HAmUiHi…CYgo`) is currently active on mainnet. With this PR merged we still won't connect to it (subnet-share check rejects anyway), and we'll now have its `agent_version` in the WARN log for identification. If the client identifies itself meaningfully via libp2p `identify`, that may give us enough to determine whether it's a third-party SSV-compatible client, an old release, or something adversarial.

---

- [ ] E2E passed - [action log](put link here)

### DIFF
--- a/api/handlers/node/mock_test.go
+++ b/api/handlers/node/mock_test.go
@@ -140,7 +140,10 @@ type MockPeersIndex struct {
 }
 
 func (m *MockPeersIndex) Self() *records.NodeInfo {
-	return m.self
+	// Mirror the real peersIndex.Self which returns a clone — this ensures
+	// the Identity handler tests exercise the same Clone path that runs in
+	// production, including its nil-Metadata behavior.
+	return m.self.Clone()
 }
 
 func (m *MockPeersIndex) NodeInfo(id peer.ID) *records.NodeInfo {
--- a/api/handlers/node/mock_test.go
+++ b/api/handlers/node/mock_test.go
@@ -140,9 +140,9 @@ type MockPeersIndex struct {
 }
 
 func (m *MockPeersIndex) Self() *records.NodeInfo {
-	// Mirror the real peersIndex.Self which returns a clone — this ensures
-	// the Identity handler tests exercise the same Clone path that runs in
-	// production, including its nil-Metadata behavior.
+	// Mirror the real peersIndex.Self which returns a clone — keeps the mock
+	// faithful to production semantics so test callers can't accidentally
+	// mutate the stored NodeInfo through the returned pointer.
 	return m.self.Clone()
 }
 
--- a/api/handlers/node/node.go
+++ b/api/handlers/node/node.go
@@ -53,15 +53,9 @@ func NewNode(
 func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
 	nodeInfo := h.peersIndex.Self()
 	resp := identityJSON{
-		PeerID: h.network.LocalPeer(),
-	}
-	// invariant: setupPeerServices initializes self.Metadata at startup, so on
-	// the live path nodeInfo.Metadata is always non-nil. The guard mirrors the
-	// `peers` handler below and keeps the response well-formed even if the
-	// invariant is weakened in the future.
-	if nodeInfo != nil && nodeInfo.Metadata != nil {
-		resp.Subnets = nodeInfo.Metadata.Subnets
-		resp.Version = nodeInfo.Metadata.NodeVersion
+		PeerID:  h.network.LocalPeer(),
+		Subnets: nodeInfo.Metadata.Subnets,
+		Version: nodeInfo.Metadata.NodeVersion,
 	}
 	for _, addr := range h.network.ListenAddresses() {
 		resp.Addresses = append(resp.Addresses, addr.String())
--- a/api/handlers/node/node.go
+++ b/api/handlers/node/node.go
@@ -53,9 +53,15 @@ func NewNode(
 func (h *Node) Identity(w http.ResponseWriter, r *http.Request) error {
 	nodeInfo := h.peersIndex.Self()
 	resp := identityJSON{
-		PeerID:  h.network.LocalPeer(),
-		Subnets: nodeInfo.Metadata.Subnets,
-		Version: nodeInfo.Metadata.NodeVersion,
+		PeerID: h.network.LocalPeer(),
+	}
+	// invariant: setupPeerServices initializes self.Metadata at startup, so on
+	// the live path nodeInfo.Metadata is always non-nil. The guard mirrors the
+	// `peers` handler below and keeps the response well-formed even if the
+	// invariant is weakened in the future.
+	if nodeInfo != nil && nodeInfo.Metadata != nil {
+		resp.Subnets = nodeInfo.Metadata.Subnets
+		resp.Version = nodeInfo.Metadata.NodeVersion
 	}
 	for _, addr := range h.network.ListenAddresses() {
 		resp.Addresses = append(resp.Addresses, addr.String())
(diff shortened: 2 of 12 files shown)


iurii-ssv:
@greptile pls re-review

Roy-blox:
## QA — stage verification (partial; mainnet canary still required)

**This is not a full hardening pass.** Two of the four claims this PR addresses can only be verified end-to-end against the misbehaving mainnet peer at `51.161.197.214`. Stage-side testing confirms the fix doesn't regress and the observable behavior change lands; the rejection and recovery paths still need to fire against a real bad peer.

### Setup
- Branch: `fix/api-nil-metadata-panic`, SHA `3588054c`
- Sacrificial node: op 76 on stage-hoodi
- Started on fix at `2026-05-26T14:26:39Z`, restored to stage default after verification

### Results

| Claim | Check | Verdict |
|---|---|---|
| C-API-1 | `/v1/node/peers` + `/v1/node/health` respond non-5xx | `PASS-hardening` (log-oracle): 110 `/v1/node/health` requests in 10m, 100% status=200, zero 5xx |
| C-API-2 | Handshake rejection of nil-Metadata NodeInfo | `PARTIAL-surface`: 0 hits on stage (peer not present). `TestVerifyTheirNodeInfo_RejectsNilMetadata` unit test is the primary evidence; mainnet canary needed |
| C-API-3 | Recovery middleware logs structured fields | `PARTIAL-surface`: 0 `API server recovered from panic` events in 7d on op 76 (nothing on stage triggers this code path); mainnet canary needed |
| C-API-4 | `agent_version` field present in handshake logs | `PASS-hardening`: 4/4 handshakes on op 76 in 5m include the field |
| regression | Other API endpoints unaffected | PASS: SSV API bound (`Serving SSV API addr=:16000`), zero 5xx in served-request logs, no error-level entries |
| regression | Op 76 duty execution healthy | PASS: 15m post-restart — 72 COMMITTEE finishes, 5 AGGREGATOR, 72 SYNC_COMMITTEE_CONTRIBUTION, zero panics, zero FATALs |

### Direct evidence — C-API-4 observable behavior change

```
# op 13 (stage HEAD, no fix) — line ends here:
... metadata={...} networkID=0x00003114

# op 76 (fix branch) — line appends new field:
... metadata={...} networkID=0x00003114 agent_version=SSV-Node:untagged-0916f7c1952a565d31b6af0dbac4ac07fed97aad
```

### What we couldn't verify on stage

- **The actual nil-Metadata NodeInfo rejection path (C-API-2)** — `setupPeerServices` always sets `self.Metadata` on stage, so no stage peer emits 2-entry records. The only known source is the mainnet peer at `51.161.197.214`. The PR's unit tests (`TestVerifyTheirNodeInfo_RejectsNilMetadata`, `TestPeers_SkipsNilMetadata`, `TestIdentity_SkipsNilMetadata`) cover the code paths; runtime confirmation needs mainnet.
- **The new panic-recovery middleware format (C-API-3)** — nothing on stage panics this code path. `TestMiddlewareRecover` covers the format in-process; runtime confirmation requires either an actual panic in production or a synthetic trigger that we didn't set up.

### Harness limitations
- Synthetic API load probe (`curl` against op's pod IP) blocked by vnet routing — same limitation as the C13 (SSVAPIAddress) test. Compensated by log-oracle: scanning request-handled and error-level entries from the API server's own logs over the test window. Recommended follow-up for both tests: `kubectl exec` from an adjacent pod to close the external-probe loop.

### Sign-off

- C-API-1 ✓, C-API-4 ✓, both regression checks ✓ — stage side is clear.
- C-API-2 and C-API-3 remain `PARTIAL-surface` until a mainnet canary observes the rejection WARN and (if it ever fires) the structured panic-recovery log against the real bad peer.

**Recommended next step:** deploy fix to one mainnet sensor, watch for ≥1 `nil-Metadata` rejection WARN + `agent_version` field identifying `51.161.197.214` over 24h. That closes C-API-2 to `PASS-hardening` and gives us the identity of the misbehaving client.

iurii-ssv:
@greptile pls re-review

momosh-ssv (approved):

ljuba-ssv (approved):

PR #2861, opened 2026-05-26
observability: restart-resilient counters + proposal-parent labels

iurii-ssv:
## Summary

Phase 1 of ssvlabs/ssv-node-board#1069.

- **Restart-resilient sparse counters**: new `observability/metrics.RegisterSparseCounter` + `EmitBaselines` so PromQL `increase()`/`rate()` return correct values after process restart. Applied to 26 sparse counters across 10 packages (the `proposal_parent` and `attestation_data` sets in `beacon/goclient` plus 16 vulnerable counters from a broader audit). Single `EmitBaselines` call in `node.go` startup.
- **Per-attribute-set baselines** via `RegisterLabeledBaseline`: lets packages pre-emit baselines for the specific label combinations they'll use at runtime, so per-label PromQL queries (e.g. `rate({beacon_client="..."})`) also work after restart. Applied to the four `proposal_parent` counters for each configured beacon address.
- **`operator_id` as OTel resource attribute**: split `observability.Initialize` into `InitializeLogger` (early) and `Initialize` (late). New `WithResourceAttributes` option carries late-known facts like `operator_id` from `operatorDataStore`. Every metric and trace is now automatically labeled with `ssv.operator_id` (when the operator is on-chain).
- **`proposal_parent` per-call labels**: `verifyProposalParent` now receives `bestClient` from `getProposalParallel` and labels all four counter sites with `ssv.beacon.client`. Plus two quick wins: `slot == 0` underflow guard and `Warn → Info` for the mismatch log (which was alert noise during normal fork resolution — see Phase-0 evidence below).

## Phase-0 evidence behind the design

From 30 days of stage data analyzed in #1069 — 28,975 verifications, 16 mismatches across 16 unique operators (uniform distribution, not concentrated).

**The metric is a noisy proxy, not a true-staleness signal.** Detailed analysis of the 2026-05-24 fork event showed that the operator whose BN was actually on the stale fork *did not* fire the mismatch warning — its cache agreed with its (stale) BN. The three operators on the *winning* fork fired the warning, because their caches still held the earlier head root while their BNs had already updated. So the metric systematically flags operators with the *fresher* BN view, not the staler one. Acting on this signal directly (tier-1/2/3 refetch logic, SIP-83-style leader selection) would harm fork resolution rather than help it.

Other observations from the same window:
- 100% of mismatches were slot-1 (in-flight build race during fork, not deeply-stale beacons)
- Mismatch rate is **0.055%** — even this noisy proxy fires rarely; true proposal-staleness is likely far lower

This PR ships only the observability + cache-improvement subset; no behavior changes to the duty hot path.

## Limitations / known follow-ups

- **`beacon_client` label not yet applied to the 8 `attestation_data` counters** — the `simpleAttestationData` path uses `gc.multiClient` which doesn't expose which underlying client returned the data. Needs separate design.
- **`ssvsigner` sparse counters (3 of them) not covered** — `ssvsigner` is a separate Go module and would need a small duplicate of the baseline mechanism.
- **Per-attribute-set baselines covered only for `proposal_parent`** — other counters labeled at runtime (e.g. by validator role) still need each package to register its own labeled baseline via `RegisterLabeledBaseline`. The infrastructure is in place; just hasn't been wired up for those counters yet.

## Incidental cleanups

- `observability/metrics` and `observability/traces` package loggers default to `zap.NewNop()` so they don't panic on a nil pointer if `InitLogger` hasn't been called yet (latent bug — instruments are created during package init, before `main()`).
- `operator/validator` `validators.removed` description corrected from "total number of validator errors" copy-paste.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test` on all packages touched (observability, beacon/goclient, network/streams, network/peers/connections, network/discovery, message/validation, eth/executionclient, eth/eventhandler, operator/validator, protocol/v2/ssv/runner, protocol/v2/qbft/instance, cli/operator) — all green
- [x] Broader `go test ./operator/... ./protocol/... ./api/... ./exporter/...` — all green
- [ ] **Staging deploy**: confirm `ssv_cl_proposal_parent_*_total` series include `ssv.operator_id` and `ssv.beacon.client` labels
- [ ] **Staging deploy**: restart a node, confirm `increase(ssv_cl_proposal_parent_verify_total{cluster="stage"}[30d])` returns non-zero (proves unlabeled baseline emission works)
- [ ] **Staging deploy**: restart a node, confirm `increase(ssv_cl_proposal_parent_verify_total{beacon_client="..."}[30d])` returns non-zero for each configured beacon (proves per-label baseline emission works)
- [ ] **Staging deploy**: confirm the mismatch log is at `info` level, not `warn`
- [ ] **Staging deploy**: confirm metrics emitted *before* `operatorDataStore` is ready are silently dropped (no errors in startup logs)

### DIFF
--- a/beacon/goclient/proposer_test.go
+++ b/beacon/goclient/proposer_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	eth2apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+	apiv1electra "github.com/attestantio/go-eth2-client/api/v1/electra"
 	"github.com/attestantio/go-eth2-client/spec"
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
@@ -128,37 +129,37 @@ func createProposalBeaconServer(t *testing.T, options beaconProposalServerOption
 	return server, serverGotRequests
 }
 
-// Create a safe proposal response using ssv-spec utilities (called once during server setup)
+// createProposalResponseSafe builds a JSON-encoded beacon proposal response with the
+// requested slot + fee recipient. It deep-clones the shared spec-testing fixture before
+// mutating: TestingBeaconBlockV / TestingBlindedBeaconBlockV return wrappers whose inner
+// *electra.BlockContents / *electra.BlindedBeaconBlock fields point at package-level
+// singletons (TestingBlockContentsElectra, TestingBlindedBeaconBlockElectra). Mutating
+// those directly leaks state across every test in the binary; SSZ round-trip gives us a
+// per-call independent copy.
 func createProposalResponseSafe(slot phase0.Slot, feeRecipient bellatrix.ExecutionAddress, blinded bool) []byte {
 	if blinded {
-		// Get a blinded block from ssv-spec testing utilities
-		versionedBlinded := spectestingutils.TestingBlindedBeaconBlockV(spec.DataVersionElectra)
-		block := versionedBlinded.ElectraBlinded
+		shared := spectestingutils.TestingBlindedBeaconBlockV(spec.DataVersionElectra).ElectraBlinded
+		sszBytes, _ := shared.MarshalSSZ()
+		block := new(apiv1electra.BlindedBeaconBlock)
+		_ = block.UnmarshalSSZ(sszBytes)
 
-		// Modify the fields we need for our test
 		block.Slot = slot
 		block.Body.ExecutionPayloadHeader.FeeRecipient = feeRecipient
 
-		// Wrap in response structure
-		response := map[string]any{
-			"data": block,
-		}
+		response := map[string]any{"data": block}
 		data, _ := json.Marshal(response)
 		return data
 	}
 
-	// Get a regular block from ssv-spec testing utilities
-	versioned := spectestingutils.TestingBeaconBlockV(spec.DataVersionElectra)
-	blockContents := versioned.Electra
+	shared := spectestingutils.TestingBeaconBlockV(spec.DataVersionElectra).Electra
+	sszBytes, _ := shared.MarshalSSZ()
+	blockContents := new(apiv1electra.BlockContents)
+	_ = blockContents.UnmarshalSSZ(sszBytes)
 
-	// Modify the fields we need for our test
 	blockContents.Block.Slot = slot
 	blockContents.Block.Body.ExecutionPayload.FeeRecipient = feeRecipient
 
-	// Wrap in response structure
-	response := map[string]any{
-		"data": blockContents,
-	}
+	response := map[string]any{"data": blockContents}
 	data, _ := json.Marshal(response)
 	return data
 }
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -147,16 +147,26 @@ var StartNodeCmd = &cobra.Command{
 			}
 		}
 
-		observabilityOptions := []observability.Option{
-			observability.WithLogger(
-				cfg.LogLevel,
-				cfg.LogLevelFormat,
-				cfg.LogFormat,
-				cfg.LogFilePath,
-				cfg.LogFileSize,
-				cfg.LogFileBackups,
-			),
+		if err := observability.InitializeLogger(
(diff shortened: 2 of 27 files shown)


observability/metrics/baseline.go
```diff
@@ -0,0 +1,113 @@
...
+// and produces extra samples in the scrape window. Prefer the once-at-startup pattern.
+func EmitBaselines(ctx context.Context) {
+	// The sparse-counters loop holds the lock during Add(ctx, 0). This is safe because
+	// metric.Int64Counter.Add is a leaf operation in the OTel SDK — it does not call back
+	// into user code and so cannot re-enter RegisterSparseCounter. If a future counter
+	// implementation grows callbacks (e.g. observable instruments with user callbacks),
+	// switch to the snapshot-then-invoke pattern used for labeled baselines below.
+	sparseCountersMu.Lock()
+	for _, c := range sparseCounters {
+		c.Add(ctx, 0)
```
momosh-ssv, line 99: Might be worth a closer look here: `EmitBaselines` emits `Add(ctx, 0)` with no attributes, but it seems most of the counters registered as sparse are only ever incremented *with* a per-call attribute set — `connected`/`disconnected`/`filtered` carry `direction`, `events_failed` the event name, `rounds_changed` role/round/reason, `submissions.failed` the role, `router.messages.dropped` the reason, etc. 

For those, the zero baseline lands on a phantom `{}` series that production never touches, so the real labeled series still start cold after a restart — the exact `increase()`/`rate()` gap this is meant to close. The `proposal_parent` counters handle this correctly via `RegisterLabeledBaseline`. 

Maybe we either extend that pattern to the labeled counters, or explicitly de-scope them and tighten the docs so the mechanism's current reach is clear? The genuinely-unlabeled ones (`multi_client.method_errors`, validator `errors`, `attestation_data.*`) do work as-is.

cli/operator/node.go
```diff
@@ -147,43 +147,33 @@ var StartNodeCmd = &cobra.Command{
...
+			log.Fatal("could not initialize logger", zap.Error(err))
 		}
+
+		logger := zap.L()
+		defer ssvlog.CapturePanic(logger)
+
+		// Metric and trace provider initialization is deferred until later in startup
+		// (after operatorDataStore is set up) so that operator_id can be baked into the
+		// OTel resource attributes — every emitted metric/trace is then automatically
+		// labeled with the operator. Metrics emitted before that point are dropped, which
```
momosh-ssv, line 167: Worth considering a short note guarding this trade-off: the dropped-metrics window looks sound today (the only counter I found emitting before `Initialize` is `beaconNodeStatusGauge`, a re-emitted gauge, so the first dropped sample is harmless). 

But it rests on an unasserted invariant — that nothing one-shot fires before `Initialize`. A one-liner here (or near `goclient.New`) warning future refactors not to emit one-shot counters before observability is initialized would keep this from silently breaking.

beacon/goclient/proposer_test.go
```diff
@@ -128,37 +129,37 @@ func createProposalBeaconServer(t *testing.T, options beaconProposalServerOption
...
+// per-call independent copy.
 func createProposalResponseSafe(slot phase0.Slot, feeRecipient bellatrix.ExecutionAddress, blinded bool) []byte {
 	if blinded {
-		// Get a blinded block from ssv-spec testing utilities
-		versionedBlinded := spectestingutils.TestingBlindedBeaconBlockV(spec.DataVersionElectra)
-		block := versionedBlinded.ElectraBlinded
+		shared := spectestingutils.TestingBlindedBeaconBlockV(spec.DataVersionElectra).ElectraBlinded
+		sszBytes, _ := shared.MarshalSSZ()
+		block := new(apiv1electra.BlindedBeaconBlock)
+		_ = block.UnmarshalSSZ(sszBytes)
```
momosh-ssv, line 144: Might be worth `require.NoError` on the marshal/unmarshal here. 

The deep-clone nicely fixes the shared-singleton mutation, but if `UnmarshalSSZ` ever errored we'd silently get a zero-valued block and a confusing downstream failure — a guard would localize any future fixture breakage.

PR #2862, opened 2026-05-28
Suggested fixes for review feedback on #2843

iurii-ssv:
Suggested resolutions for the inline review feedback on #2843. Each commit-included change matches one inline comment in that review — merge as-is, cherry-pick, or close if you'd rather address differently.

## Changes

- **Tighten `ssv_p2p_pubsub_messages_received_total` description** so an operator reading the metric registry can immediately tell it is the *pre-validation* counterpart of `ssv_p2p_messages_in_total`.
- **Add a brief comment on `recordPubsubMessageReceived`** noting it is called from the validator wrapper before the inner validator runs, so all outcomes (accept/ignore/reject/timeout) are counted — useful context for the next reader.
- **Rename test loop variable** that shadowed the imported `go.opentelemetry.io/otel/sdk/metric` package (lint-only nit).

## Not included

The cross-counter attribute-key fragmentation (`ssv.p2p.pubsub.topic` vs `ssv.p2p.message.topic` vs `ssv.p2p.topic.name`) is left for you to decide on — that one is a design call rather than a mechanical fix.

## Test plan

- [x] `go build ./network/topics/...`
- [x] `go test -run TestRecordPubsubMessageReceived ./network/topics/ -count=1`

### DIFF
--- a/network/topics/observability.go
+++ b/network/topics/observability.go
@@ -38,7 +38,7 @@ var (
 		meter.Int64Counter(
 			observability.InstrumentName(pubsubObservabilityNamespace, "received"),
 			metric.WithUnit("{message}"),
-			metric.WithDescription("total number of messages received by the pubsub topic validator")))
+			metric.WithDescription("total number of messages delivered to the pubsub topic validator, before SSV validation runs (compare with ssv_p2p_messages_in_total for the post-validation rate)")))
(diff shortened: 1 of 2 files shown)


momosh-ssv (approved):

PR #2863, opened 2026-05-29
fix: standardize weighted attestation-data error handling and log levels

iurii-ssv:
## Summary

`weightedAttestationData` fans out an `AttestationData` request to **every** configured beacon node and succeeds as long as *one* returns data — so individual client failures are tolerated and routinely recovered from. Two things didn't reflect that:

1. **Over-loud levels.** Per-client errors, `"hard timeout reached"`, and the final `"no attestations received"` were all logged at **ERROR**, producing noisy/misleading errors for benign conditions. The most common on mainnet:
   ```
   GET failed with status 500: {"code":500,"message":"UNHANDLED_ERROR:
   HeadBlockNotFullyVerified { beacon_block_root: 0x…, execution_status: Optimistic(0x…) }"}
   ```
   That's a beacon node correctly refusing to produce attestation data while its head is still **optimistically synced** (the EL hasn't validated the payload) — transient and expected per the [optimistic sync spec](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md). And the committee runner already logs the returned error at ERROR (`operator/validator/controller.go`), so the in-function `"no attestations received"` ERROR was a **duplicate**.

2. **Inconsistent error aggregation.** It collected per-client errors into a one-off `zap.Errors` log field and returned a generic `"no attestations received"`, whereas the **weighted proposal path** (`proposer.go`) aggregates with `errors.Join` and returns a detailed error.

## Change

Standardize on the proposal idiom and right-size the levels so **ERROR means "the duty failed"**:

- **Aggregate per-client errors with `errors.Join`** and **return** a detailed wrapped error (`"all N clients failed to get attestation data for slot S, encountered errors: %w"`) — identical in shape to the proposal path. The per-client detail now rides along to whoever logs the error.
- **Drop the in-function ERROR log** — the duty layer already logs the returned error once at ERROR, now *with* the per-client detail.
- **Per-client errors → DEBUG**; **`"hard timeout reached"` → WARN** (a blown deadline is noteworthy but not itself a failure: a late response can still arrive and the fetch succeed).
- **Unify** the two per-client error messages → `"error fetching attestation data"`.

Net result: the attestation path is now consistent with the weighted-proposal path, and a tolerated/recoverable condition no longer surfaces as ERROR.

## Testing

- `go build ./beacon/goclient/...`, `go vet ./beacon/goclient/`, `gofmt` — clean.
- `go test ./beacon/goclient/` (full package) — pass. Existing weighted error/hard-timeout assertions updated to the new aggregated error, plus a new **multiple-clients all-error** case asserting the aggregated `errors.Join` output (`"all 3 clients failed … encountered errors:"`).


julienharbulot (approved):
Reviewed the weighted attestation-data error handling changes, the related proposer nil-error guard, and the focused goclient tests. The change keeps the selection/control flow intact while moving tolerated per-client failures out of ERROR logs and preserving actionable failure details in the returned aggregated error.

momosh-ssv (approved):

PR #2866, opened 2026-05-29
cli/operator: refactor node startup into a thin, testable flow

iurii-ssv:
## Summary

Refactors the `cli/operator` startup path from a single ~650-line `Run` closure into a thin, testable `runNode()` → `newNode()` → `start()` flow over a `node` struct — fixing a confirmed p2p-lifecycle bug, making startup fail-fast, propagating errors instead of `Fatal`-ing mid-flow, and adding in-process startup tests. Lands as a sequence of **atomic, individually-gated commits** (reviewable one-by-one); behavior-preserving except for the explicit fixes/fail-fast changes called out below.

The first commit is the original config-consolidation foundation (`config.resolveAndValidate`); the rest build the startup refactor on top.

## Why

The operator boot path was a single ~650-line closure in `node.go` with 40+ `logger.Fatal` exits and 13 scattered `ExporterOptions.Enabled` branches interleaving three modes (operator / exporter-standard / exporter-archive). It was hard to read, impossible to unit-test, and hid a real p2p-lifecycle bug (below).

## What changed (by theme)

**0. Config consolidation (foundation).** New `cli/operator/config.go` owns the config declarations + a single `resolveAndValidate` entry point (signing + proposer-delay validation → returned errors, with signing-source context attached as structured fields on the one fatal). No functional change.

**1. p2p-lifecycle fix (the bug).** Moved the `GetValidatorStats` wiring + `Setup`/`Start` + health-registration out of the `if DynamicMaxPeers` block so they always run. Previously, `DynamicMaxPeers=false` left the network never `Setup`/`Start`'d → `operator.Node.Start` failed `Subscribe*` with `ErrNetworkIsNotReady`, **and** a nil `GetValidatorStats` made pubsub peer-scoring silently fall back to fake hard-coded stats. (Repro needs the env var `P2P_DYNAMIC_MAX_PEERS=false` — a YAML `false` is clobbered to the default; see #2868.)

**2. Structure.** Extracted everything after bootstrap into `runNode(ctx, cfg, logger)` (in `node.go`); introduced a `node` struct with `newNode()` (wiring) / `Close()` (p2p+EL+db teardown, now CLI-owned) / `start()` (long-lived services). `runNode()` is a ~110-line orchestrator. The startup helpers are organized by concern into focused files — `keys.go`, `db.go`, `network.go`, `eventsync.go` (config-domain helpers in `config.go`); `p2pNetwork.Close()` is made robust on partial-setup states. The cobra command + observability bootstrap live in `start_node.go`, keeping `node.go` cobra-free.

**3. Fail-fast.** `resolveAndValidate` runs first, so config errors surface before any network I/O; a typed `nodeMode` enum + `resolved` reject an invalid `EXPORTER_MODE` up front (was a late Fatal).

**4. Error propagation.** Converted the synchronous-path `Fatal`s to returned errors that flow to a single top-level `Fatal` in `start_node.go` (`startupError` carries structured log fields).

**5. Mode clarity.** Split the 3-mode wiring into named leaf builders — `resolveOperatorIdentity`, `openNodeDB`, `buildKeyManager`, `buildDoppelganger` — dispatched through a single `resolved.isExporter()` predicate, so `newNode()` reads as one linear spine instead of scattered `if Enabled` branches.

**6. Tests.** In-process smoke tests drive `newNode()`/`start()` with stub beacon/EL clients across all three modes; a CLI-level regression covers the p2p-lifecycle fix (mutation-verified).

## Behavior-change callouts (otherwise behavior-preserving)

- **p2p-lifecycle fix** (theme 1) — intentional.
- **Invalid `EXPORTER_MODE`** now fails fast at startup (was a late collector-switch Fatal).
- **Config validation** (e.g. invalid `ProposerDelay`) now surfaces before network I/O.
- **EL client now connects before db/key-loading/`validateConfig`** — it is built in `runNode()` before `newNode()`. Benign: the CL already connected before local validation in both old and new code, and the process exits on failure either way.
- **Startup-failure logging consolidated** — all **synchronous** startup failures now surface as one top-level `"could not start node"` line with the specific failure in the `error` field, instead of each step logging its own `Fatal` message. Log alerting that greps the message text for a specific failure string (e.g. `"failed to create beacon go-client"`) should match the `error` field instead. (Background-goroutine failures still log their own line — see #2867.)
- Config-consolidation folds the distinct signing/proposer-delay fatals into one `"invalid configuration"` fatal (the specific reason is in the `error` field); the "both keystore and password files" fatal additively gains the four signing-source fields.

## Deferred follow-up

The in-goroutine `Fatal`s for the long-lived background services (metrics + SSV-API servers, health prober, ongoing event-sync) are **not** converted here — that needs an errgroup + graceful HTTP shutdown that is cross-package and overlaps #2819. Tracked in **#2867**.

## Test plan

- [x] `gofmt`, `go build ./...`, `go vet ./cli/operator/...` clean.
- [x] `golangci-lint run ./cli/operator/...` → 0 issues.
- [x] `go test -race ./cli/operator/...` passes — incl. the smoke + regression tests across all three modes (operator ±doppelganger, exporter-standard, exporter-archive) plus the `newNode` error-cleanup (db-close) test.
- [x] Each commit is individually gated (gofmt/build/vet/test/lint) and bisectable.

---

Known follow-ups (out of scope here):

- The errgroup / graceful-shutdown lifecycle work — **#2867** (sequence after #2819).
- The cleanenv `env-default:"true"` bool footgun — a YAML `false` is clobbered to the default (affects `DynamicMaxPeers` / `PubSubScoring`) — **#2868**.
- Unify the `KeyStore.*` log-field names across `signingLogFields` / `warnExporterSigning`.
- The MEV/block-fetch PR will pull `resolveAndValidate` ahead of beacon-client construction and resolve block-fetch path/deadline *into* the typed `resolved` (rather than scattering it across `goclient`).


cli/operator/config_test.go
```diff
@@ -0,0 +1,151 @@
...
+		{name: "operator private key", mutate: func(c *config) { c.OperatorPrivateKey = "key" }, wantPK: true},
+		{name: "keystore missing password -> error", mutate: func(c *config) { c.KeyStore.PrivateKeyFile = "pk" },
+			wantErr: "both keystore and password files"},
+		{name: "keystore missing key -> error", mutate: func(c *config) { c.KeyStore.PasswordFile = "pw" },
+			wantErr: "both keystore and password files"},
+		{name: "ssv-signer + private key -> error", mutate: func(c *config) {
+			c.SSVSigner.Endpoint = "http://signer:9000"
+			c.OperatorPrivateKey = "key"
+		}, wantErr: "cannot enable both remote signing"},
+		{name: "keystore + private key -> error", mutate: func(c *config) {
```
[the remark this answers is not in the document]
iurii-ssv: Good catch — added the `ssv-signer + keystore` case to `Test_resolveSigning` in 3b68b36ce, so all three branches of the `usingSSVSigner && (usingKeystore || usingPrivKey)` conflict are now exercised.

cli/operator/config.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	if c.OperatorPrivateKey != "" {
+		res.usingPrivKey = true
+	}
+
+	if res.usingSSVSigner && (res.usingKeystore || res.usingPrivKey) {
+		return resolved{}, fmt.Errorf("cannot enable both remote signing (SSVSigner.Endpoint) and local signing (PrivateKeyFile/OperatorPrivateKey)")
+	} else if res.usingKeystore && res.usingPrivKey {
+		return resolved{}, fmt.Errorf("cannot enable both OperatorPrivateKey and PrivateKeyFile")
+	}
```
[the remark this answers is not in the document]
iurii-ssv: Restored in 3b68b36ce: `resolveAndValidate` now wraps a signing error with the configured sources (`SSVSigner.Endpoint`, `KeyStore.PrivateKeyFile`/`PasswordFile`, and `OperatorPrivateKey set=<bool>` — the key value itself is never logged). The triage context is back in the consolidated startup `Fatal`, as error text rather than structured fields; `Test_resolveAndValidate_signingErrorContext` covers it.

iurii-ssv:
@greptile pls re-review

iurii-ssv:
@greptile pls re-review

cli/operator/network.go
```diff
@@ -0,0 +1,172 @@
...
+
+	nodeType := "light"
+	if cfg.SSVOptions.ValidatorOptions.FullNode {
+		nodeType = "full"
+	}
+
+	logger.Info("setting ssv network",
+		zap.Any("config", ssvConfig),
+		zap.String("nodeType", nodeType),
+		zap.String("registryContract", ssvConfig.RegistryContractAddr.String()),
```
momosh-ssv, line 76: What happens if a node is configured with neither `NetworkName` nor `CustomNetwork`? Seems that `ssvConfig` stays nil here, so `RegistryContractAddr.String()` (and `DomainType[:]` above when `CustomDomainType` is set) would panic rather than fail cleanly. 

Maybe worth an early `if ssvConfig == nil { return nil, fmt.Errorf("no ssv network configured (set NetworkName or CustomNetwork)") }` guard so this fails fast in line with the refactor's goal.

cli/operator/node.go
```diff
@@ -77,1277 +45,594 @@ import (
...
+		return nil, err
 	}
-
-	operatorPrivKey, err := keys.PrivateKeyFromBytes(operatorPrivKeyBytes)
-	if err != nil {
-		return nil, nil, fmt.Errorf("could not extract operator private key from bytes: %w", err)
+	validatorProvider := nodeStorage.ValidatorStore().WithOperatorID(operatorDataStore.GetOperatorID)
+	var validatorRegistrationSubmitter runner.ValidatorRegistrationSubmitter
+	if !res.isExporter() {
+		validatorRegistrationSubmitter = runner.NewVRSubmitter(ctx, logger, networkConfig.Beacon, consensusClient, validatorProvider)
```
momosh-ssv, line 278: Do we need to stop this goroutine on teardown? `NewVRSubmitter` launches a goroutine bound to `ctx`, but `node.Close()` only tears down p2p/EL/db and never cancels `ctx` — so if `newNode` fails after this point (or `Close` is called), it keeps running until the caller cancels `ctx`. 

I fear in prod it's only masked by the top-level `Fatal`→`os.Exit`, and the smoke tests already work around it by canceling ctx manually. 
This looks distinct from #2867 (which covers `start()`'s background services, not this construction-phase goroutine). 
Might be worth either having `runNode` own a `context.WithCancel` cancelled in the `Close` defer, or documenting that `Close` is resource-only and lifecycle goroutines belong to the caller's ctx.

cli/operator/network.go
iurii-ssv: Good catch — fixed. `setupSSVNetwork` now fails fast when neither is set, via an `else` on the network-resolution chain, so `ssvConfig` can't be nil past that point (no panic from the `CustomDomainType` write or the `RegistryContractAddr` log):

```go
} else {
    return nil, fmt.Errorf("no ssv network configured: set Network (e.g. \"mainnet\") or CustomNetwork")
}
```

(0b76c109)

cli/operator/node.go
iurii-ssv: Addressed — went with your first option.

`newNode` now derives the node's own ctx (a child of runNode's) and stores it; `node.Close()` cancels it. So the VRSubmitter — and the other construction-phase ctx-bound goroutines (in exporter modes, the duty-trace collector and slot-pruning) — stop on teardown rather than relying on the top-level `Fatal`→`os.Exit`. The smoke tests dropped their manual `cancel()` workaround as a result (`Close()` owns it now). `runNode` keeps the parent ctx for goclient/EL (goclient has no `Close()`, so ctx cancellation is its only shutdown), and yes — distinct from #2867. (0b76c109)

One thing that surfaced while verifying this: the validator controller's `ttlcache` cleanup goroutines aren't ctx-bound either — they stop only via `Stop()`, which nothing ever called (a pre-existing leak). Follow-up f7fe8e6dc adds `Controller.Stop()` and calls it from `Close()`, so teardown is now actually complete.

momosh-ssv: > One thing that surfaced while verifying this...

Great call; I wasn't sure about the immediate action, since it was pre-existing, and was thinking about dedicating a separate PR for this

momosh-ssv (dismissed):

cli/operator/node.go
```diff
@@ -77,1277 +45,616 @@ import (
...
-		!bytes.Equal(legacyHash, storedHash) {
-		// Prevent the node from running with a different key.
-		return fmt.Errorf("operator private key is not matching the one encrypted the storage")
+	if len(n.cfg.LocalEventsPath) == 0 {
+		healthProber.AddComponent(eventSyncerComponentName, eventSyncer, proberHealthcheckTimeout, proberRetriesMax, proberRetryDelay)
 	}
 
-	return nil
-}
+	go startHealthProber(n.ctx, n.logger, healthProber)
```
alok-ssv, line 533: The startHealthProber calls log.Fatal terminating the process. Is this expected? The PR says we are avoiding this?

Maybe better to log the error and keep retrying?

cli/operator/prober.go
```diff
@@ -24,17 +25,18 @@ const (
 	proberRetryDelay         = 10 * time.Second
 )
 
-const componentsUnhealthyFatalErrorMsg = "component(s) are not healthy"
+const componentsUnhealthyErrorMsg = "component(s) are not healthy"
 
-func ensureComponentsHealthy(ctx context.Context, logger *zap.Logger, p *hprobe.HealthProber) {
+func ensureComponentsHealthy(ctx context.Context, logger *zap.Logger, p *hprobe.HealthProber) error {
```
alok-ssv, line 30: nit: We should be able to use this function in startHealthProber

cli/operator/node.go
```diff
@@ -77,1280 +45,619 @@ import (
...
+				elComponentName,
+				eventSyncerComponentName,
+			),
+			&hvalidators.Validators{
+				Shares: n.nodeStorage.Shares(),
+			},
+			hexporter.NewExporter(n.logger, n.storageMap, n.collector, n.nodeStorage.ValidatorStore()),
+			n.mode == modeExporterArchive,
+		)
+		_, apiServeErr, err := apiServer.Start(cfg.SSVOptions.Context)
```
julienharbulot: I think this should use `n.cfg.SSVOptions.Context` or `n.ctx` instead of the package-level `cfg`, because this path is now meant to run off the injected `*config`?

In the CLI path today `n.cfg` appears to point at the same global `cfg`, so production may be fine, but keeping this reference makes the new `node` flow depend on hidden global state. If `newNode`/`start` is used with a test config or any non-global config and `SSVAPIPort > 0`, the API server could start with the wrong context, possibly nil, and shutdown would be coupled to the package global rather than the node lifecycle.

cli/operator/config.go
```diff
@@ -0,0 +1,319 @@
...
+		if c.KeyStore.PrivateKeyFile == "" || c.KeyStore.PasswordFile == "" {
+			return resolved{}, fmt.Errorf("both keystore and password files must be provided if using keystore")
+		}
+		res.usingKeystore = true
+	}
+	if c.OperatorPrivateKey != "" {
+		res.usingPrivKey = true
+	}
+
+	if res.usingSSVSigner && (res.usingKeystore || res.usingPrivKey) {
```
julienharbulot, line 208: What do you think about rejecting the non-exporter case where none of `SSVSigner.Endpoint`, `KeyStore`, or `OperatorPrivateKey` is configured?

`resolveSigning` currently accepts that as a valid `resolved{}`. Later, `resolveOperatorIdentity` can reach `operatorPrivKey.Public().Base64()` with a nil `OperatorPrivateKey`, so this startup path can still panic instead of returning the new single top-level startup error. Since this PR centralizes signing validation and moves config checks earlier, this seems like a good place to make “exactly one signing source for operator mode” explicit.

alok-ssv (commented):
I asked AI to also review. It found some issue which is not part of the diff. Can you check if this should be fixed?

```
[P1] Delay the cutoff bump until after the timeout round-change is sent — /Users/Alok/dev/ssv/protocol/v2/qbft/instance/timeout.go:36-55
    When newRound reaches the cutoff round, calling bumpToRound before CreateRoundChange/Broadcast makes the instance irrelevant first, so Broadcast now rejects the final timeout-driven round change instead of sending it. In the common
    CutOffRound == currentRound + 1 case this turns a normal timeout transition into an error and prevents the duty from completing cleanly.
```

cli/operator/node.go
iurii-ssv: Good catch — fixed in 1b31e2c45. `start()` now passes `n.ctx` to `apiServer.Start(...)`, matching the rest of the method (e.g. `n.operatorNode.Start(n.ctx)` just below) and the node-owned lifecycle.

You're right that it's harmless in the CLI path today — `n.cfg` is the same global, and `newNode` sets `cfg.SSVOptions.Context = n.ctx`, so they resolve to the same object — but the bare `cfg` was a leftover from merging #2819's `apiServer.Start(cfg.SSVOptions.Context)` into the refactored node flow, and it reintroduced exactly the hidden-global coupling this refactor is removing. With `n.ctx` the API server's lifetime is bound to the node instead of the package global.

While here, I put the metrics server on the same footing in the same commit — it was the other startup HTTP server still on the old fire-and-forget `Start`/`Fatal` pattern, so `metrics.Handler.Start` now takes the node ctx (non-blocking, returns a `serveErr` channel) and shuts down on `Close()` too.

cli/operator/config.go
iurii-ssv: Done in 1b31e2c45 — `resolveSigning` now rejects the no-source case, so "exactly one signing source for operator mode" is explicit (the conflict checks reject >1, this rejects 0):

```go
if !res.usingSSVSigner && !res.usingKeystore && !res.usingPrivKey {
    return resolved{}, fmt.Errorf("no operator signing configured: set one of SSVSigner.Endpoint, KeyStore (PrivateKeyFile + PasswordFile), or OperatorPrivateKey")
}
```

It's only reached in operator mode (exporter mode takes the `warnExporterSigning` path and never calls `resolveSigning`), and the error flows through the same `signingConfigError` / `signingLogFields` wrapping as the conflict errors — so it surfaces as the single top-level startup error instead of the nil-interface panic in `resolveOperatorIdentity` you spotted. Covered by flipping the `Test_resolveSigning` "nothing set" case to expect the error (and the `resolveAndValidate` tests that used a keyless `config{}` now set a source).

iurii-ssv:
@alok-ssv thanks again for the extra pass — the finding is real, and the mechanism stands: in `UponRoundTimeout`, `i.bumpToRound(newRound)` runs *before* `i.Broadcast(...)`; `Broadcast` rejects when `!IsRelevant()` (`State.Round < CutOffRound`), so at the boundary `prevRound == CutOffRound-1` (CutOffRound = 12) the bump to round 12 makes the instance irrelevant before the broadcast, the final round-change is dropped, and the call returns an error. ssv-spec avoids this by deferring the bump.

Correction to my earlier "I'll fix it in a follow-up": I opened #2872 to mirror the spec, but after implementing it I don't think it earns a behavior change, so I'm closing it as won't-fix and keeping current behavior:

- It only differs at round 11→12 — after ~11 consecutive timeouts, at the cluster-wide give-up point. Round == CutOffRound is a dead round: reaching it makes the instance irrelevant, so no node can *decide* there. A round-change for the cutoff round has no liveness value — at most it nudges a lagging peer to give up a tick sooner.
- The only observable effects are the un-sent terminal RC and a Debug-level error + error-status span on an already-abandoned duty. The error isn't retryable, so there's no retry churn. No healthy duty is affected.
- The eager-bump ordering also matches our partial-quorum path. It does diverge from ssv-spec's deferred-bump-in-timeout, but the divergence is behaviorally inert at the give-up boundary.

[I'll leave a short comment in `timeout.go`](https://github.com/ssvlabs/ssv/pull/2873) marking the cutoff-boundary behavior as intentional so it doesn't get re-flagged. Pre-existing either way (not introduced by this PR's diff).

cli/operator/node.go
iurii-ssv: Good catch on the inconsistency. The distinction: this PR's Fatal-avoidance targets the startup/teardown path — owning Close() rather than bypassing it. This particular Fatal is a different beast: a runtime liveness watchdog. If CL/EL (or event-syncer/p2p) stays unhealthy for minutes, crashing so the orchestrator restarts us — potentially onto a healthy endpoint — is the intended behavior. It's pre-existing (identical on stage); the refactor only relocated it.

That said, you're right it shouldn't *stay* a Fatal long-term — the restart *behavior* is correct, but the *mechanism* (os.Exit from a background goroutine, skipping Close()) is exactly what we're eliminating elsewhere. The proper fix is a graceful watchdog-triggered shutdown: return into the node errgroup → Close() → non-zero exit, so the orchestrator still restarts — same machinery as the other background-service Fatals. "Log and retry forever" isn't quite the fix either: it drops the liveness property (a node idling with a dead EL is a silent failure).

Marked it with a `// TODO(#2867)` here and folded the design into #2867. Keeping behavior unchanged in this PR.

cli/operator/prober.go
iurii-ssv: Agreed in spirit — though it's contingent on the Fatal question above. Once startHealthProber returns an error instead of Fatal-ing, the two share their probe plumbing and a shared helper falls out naturally. One caveat: literal reuse isn't quite right — `ensureComponentsHealthy` carries a 30s timeout and a per-success "all healthy" info-log that'd spam every 60s at runtime. So the clean version is a thin `probe(ctx, p, timeout)` core that both wrap, each keeping its own policy (one-shot fail-fast vs periodic watchdog). Tracked together with the Fatal conversion in #2867.

julienharbulot (approved):

alok-ssv (approved):

momosh-ssv (approved):

PR #2869, opened 2026-06-01
executionclient: dual-transport EL client (WS subscriptions + HTTP queries) — works around Besu WS event-loop bug

Roy-blox:
## Summary

Split the EL client transport so heavy request/response calls (`eth_getLogs`, `eth_blockNumber`, `eth_getBlockByNumber`, `eth_syncing`, `eth_chainId`) route to an optional **HTTP** endpoint while subscriptions (`eth_subscribe`-family: `SubscribeFilterLogs`, `SubscribeNewHead`) stay on the existing **WebSocket** endpoint.

New env var: `ETH_1_QUERY_ADDR` (optional, semicolon-separated, positionally paired with `ETH_1_ADDR`). When unset, behavior is byte-identical to today's single-transport client — fully backward compatible.

## Background — the Besu bug we're working around

Operators paired with **Besu 26.5.0** (and likely earlier 26.x) saw a sustained outage where SSV nodes entered a continuous crash-loop. Effectiveness dropped to 0% for affected clusters on staging. A multi-day investigation traced it to a single Besu commit:

> [hyperledger/besu#9848](https://github.com/hyperledger/besu/pull/9848) — *"Stream debug_traceBlock* responses directly to avoid OOM on large blocks"* (commit `b2cbdeae01`, merged 2026-04-08)

That PR added `StreamBackpressure.awaitDrain` (parks the calling thread on a `CountDownLatch` for up to 60s) and wired it into the WebSocket `JsonResponseStreamer.write` path. The Javadoc on `StreamBackpressure` explicitly states:

> *Safe to call from `executeBlocking` worker threads only — must never be called from the Vertx event loop.*

But the PR refactored only the **new** streaming method path (`debug_traceBlock*`) to use `executeBlocking`. The **pre-existing** non-streaming WebSocket reply path was left unchanged:

```java
// ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketMessageHandler.java:206-214
private void replyToClient(final ServerWebSocket websocket, final Object result) {
  traceResponse(result);
  try {
    JSON_OBJECT_WRITER.writeValue(new JsonResponseStreamer(websocket), result);  // ← runs on event loop
  } catch (IOException ex) { ... }
}
```

`replyToClient` is reached from `.onSuccess(...)` callbacks on the `Future` produced by `vertx.executeBlocking(...)` — which by Vert.x semantics resolve on the **event loop**, not on a worker. Jackson's writes funnel through `JsonResponseStreamer.write` → `StreamBackpressure.awaitDrain` → `CountDownLatch.await(60s)` **on the event loop**, blocking every other RPC consumer on that Besu instance: the consensus client's engine API, SSV health probes, and every other SSV node sharing the backend.

We observed `vert.x-eventloop-thread-1 has been blocked for 30000+ ms` warnings firing **continuously** on three staging Besu instances. The cascade: SSV node issues a fat `eth_getLogs` over WebSocket → Besu's event loop wedges → SSV health probe times out → SSV pod FATALs → restart → re-issues the same query → wedges Besu again. Op 1 alone accumulated **700+ restarts in three days** before this fix.

Stack trace from production for confirmation:

```
at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:276)
at org.hyperledger.besu.ethereum.api.jsonrpc.StreamBackpressure.awaitDrain(StreamBackpressure.java:52)
at org.hyperledger.besu.ethereum.api.jsonrpc.websocket.JsonResponseStreamer.write(JsonResponseStreamer.java:61)
at com.fasterxml.jackson.core.json.UTF8JsonGenerator._flushBuffer(UTF8JsonGenerator.java:2293)
... (Jackson serialization of an eth_getLogs response) ...
```

## Why fix this on the SSV side at all

The Besu issue is fundamentally a Besu bug. An upstream fix should be filed there. But:

1. **We can't wait.** Operators are seeing 0% effectiveness *today*. Upstream PR cycles are slow.
2. **The bug only triggers on the WS transport.** The HTTP request handler uses a different code path (`JsonRpcObjectExecutor`) which, while it has a latent variant of the same pattern, has a much larger write queue and isn't shared with the engine API thread pool. In practice, HTTP avoids the symptom entirely for SSV's workload.
3. **`eth_subscribe` is unsupported on HTTP** — pure HTTP isn't a viable substitute for the WS transport SSV needs for `SubscribeFilterLogs` / `SubscribeNewHead`. So we can't just flip everything to HTTP.
4. The **right architectural move** is to use each transport for what it's good at: WS for streaming server-push, HTTP for request/response. This PR does that.

## Why this doesn't harm other EL clients

| EL | Effect of the split |
|---|---|
| **Besu** | Heavy responses leave the buggy code path. Subscriptions (small messages) stay on WS where they must. Sidesteps the bug. |
| **Geth** | No behavior change. Geth handles both transports cleanly. `eth_getLogs` over HTTP is *recommended* by Geth maintainers for large queries (lower per-request overhead). |
| **Erigon** | Same as Geth. |
| **Nethermind** | Same as Geth. |
| **Lodestar/Lighthouse/etc.** | Out of scope — this is the **execution** client, not the consensus client. |

Subscriptions still go where they always did. Queries can go to either transport. With `ETH_1_QUERY_ADDR` unset, behavior is identical to the current single-client path — every existing deployment continues to work unchanged.

### Original commit (`d0551d64e`)

`eth/executionclient/eth_client.go` — holds two `*ethclient.Client` instances (`sub`, `query`). Each method routes to the appropriate one via `queryOrSub()` (falls back to `sub` when `query` is nil for backward compatibility):

| Method | Transport |
|---|---|
| `SubscribeFilterLogs` | `sub` (WS-required) |
| `SubscribeNewHead` | `sub` (WS-required) |
| `FilterLogs` | `query` (preferred HTTP) |
| `BlockNumber` | `query` |
| `HeaderByNumber` | `query` |
| `SyncProgress` | `query` |
| `ChainID` | `query` |

`eth/executionclient/config.go` — adds `QueryAddr` (env: `ETH_1_QUERY_ADDR`).
`eth/executionclient/options.go` — adds `WithQueryAddr` / `WithQueryAddrsMulti`.
`eth/executionclient/execution_client.go` — dual-dial in `connect()`; both clients closed cleanly on shutdown.
`eth/executionclient/multi_client.go` — propagates positionally-paired query addrs to each underlying single client.
`cli/operator/node.go` — wires `cfg.ExecutionClient.QueryAddr` through to `executionclient.New` / `NewMulti`.

### Post-review hardening (`9539d3a40`)

After internal review surfaced four gaps in the initial commit, the follow-up commit addresses each:

1. **Chain-ID parity at connect** — `connect()` now fetches `ChainID` from both transports when query is configured and **fails closed** on mismatch (with both addrs and chain IDs in the error). Without this, an operator typoing `ETH_1_QUERY_ADDR` to point at the wrong network would silently let SSV ingest registry events from chain A while subscriptions track chain B — a slashing-adjacent divergence that `streamLogsToChan` only warns about belatedly. The existing `MultiClient.assertSameChainID` validates *across* clients via `queryOrSub()` — it never observed the sub-transport's chain ID. Now each single client internally validates its own pair first.

2. **Sub-transport health probe** — `Healthy()` previously called `SyncProgress()` which routes through `queryOrSub()`, so when query was configured only the HTTP transport was probed. A WS-down/HTTP-up partial failure would report healthy while `SubscribeFilterLogs` / `SubscribeNewHead` silently stalled, and `MultiClient` would never failover. Added `ethClient.PingSub()` (a cheap `eth_blockNumber` against the sub transport) and call it from `Healthy()` after the existing sync-distance check. No-op in single-transport mode; the existing healthInvalidationInterval cache covers both probes.

3. **Config validation + WS-only WARN** — `cli/operator/node.go` now trims whitespace on every `ETH_1_ADDR` and `ETH_1_QUERY_ADDR` entry, and **fatals** when `ETH_1_QUERY_ADDR` is set but its entry count doesn't match `ETH_1_ADDR` (a silent count mismatch would leave some clients falling back to WS — the exact situation `ETH_1_QUERY_ADDR` exists to prevent). Empty entries within `ETH_1_QUERY_ADDR` are still permitted as a positional opt-out for non-Besu backends in mixed deployments; we WARN at startup listing the affected `ETH_1_ADDR` positions so operators see partial coverage.

4. **Honest observability** — `recordRequest` previously logged `client.Address()` — hardcoded to `nodeAddr` (the WS URL) regardless of which transport actually served the call, so `FilterLogs` over HTTP would log `client_addr=ws://...:8546`. Misleading. Now: `recordRequest` takes the effective addr as a string + variadic extra `zap.Field`s; `ethClient` gains `SubAddr()`, `QueryAddr()`, and `QueryTransport()` helpers; dial labels split into `DialContext.sub` / `DialContext.query`; the new chain-ID parity calls labeled `eth_chainId.sub` / `eth_chainId.query`; every recordRequest call site passes the effective addr plus `transport=http|ws`. Metric `ServerAddress` attribute follows the log (intentional — the metric was lying before, now accurate; slight cardinality increase per route).

   Caveat: `errSingleClient` still wraps errors with `ec.nodeAddr` (the operator-configured address). Operator-facing error messages should name the address the operator put in config; logs/metrics now show truth.

Diff: **10 files, +305 / −46 lines** total across both commits.

## Evidence — staging deploy on clusters 1-4 and 93-96

Both clusters were pointed at `stage-besu-lh-hoodi-1` (the buggy Besu instance) and were at 0% / volatile 50% effectiveness respectively for 11+ days under the old code.

Deployed with `ETH_1_QUERY_ADDR=http://stage-besu-lh-hoodi-1-execution-http.ethereum-clients.svc:8545`.

| | Before (old code, same Besu) | After (dual-transport, same Besu) |
|---|---|---|
| Pod restarts (3 days) | ~700 per op | **0** |
| FATAL/15min | 7–11 | **0** |
| Cluster 1-4 effectiveness | 0% for 11 days | **96–98%** within one epoch |
| Cluster 93-96 effectiveness | ~55% volatile | **97%** steady |
| Besu `awaitDrain` warns / 5min | ~300 | ~190 (residual from other clusters still on old code) |

The same Besu instance whose WebSocket event loop was blocked continuously for days is now serving these clusters at near-perfect effectiveness — because the heavy requests no longer hit the buggy code path.

### Mixed-backend deploy validation (post-review)

To validate the WS-only WARN path, cluster 1-4 was redeployed with a `MultiClient` configuration: **Besu first with HTTP query addr** + **Geth second with no query addr** (blank position).

`ETH_1_ADDR`:
```
ws://stage-besu-lh-hoodi-1-execution.ethereum-clients.svc:8546;ws://stage-geth-lh-hoodi-1-execution.ethereum-clients.svc:8546
```

`ETH_1_QUERY_ADDR`:
```
http://stage-besu-lh-hoodi-1-execution-http.ethereum-clients.svc:8545;
```
(trailing semicolon → blank second position)

Captured logs from Loki:

```
INFO  connecting EL(s)
  addresses=[
    'ws://stage-besu-lh-hoodi-1-execution.ethereum-clients.svc:8546',
    'ws://stage-geth-lh-hoodi-1-execution.ethereum-clients.svc:8546'
  ]
  query_addresses=[
    'http://stage-besu-lh-hoodi-1-execution-http.ethereum-clients.svc:8545',
    ''
  ]
  request_timeout=10s sync_distance_tolerance=5

WARN  ETH_1_QUERY_ADDR set but some positions are blank — those EL backends
      will use WS for queries and may trigger Besu's StreamBackpressure
      event-loop block on large eth_getLogs responses. Set an HTTP endpoint
      at each blank position to opt those backends into the dual-transport
      path.
  ws_only_addrs=['ws://stage-geth-lh-hoodi-1-execution.ethereum-clients.svc:8546']
```

The Geth backend (position 2, no HTTP query addr) is correctly identified, named in the structured `ws_only_addrs` field, and explained with the underlying reason. All 4 pods rolled cleanly on the new image with `ready=true, restarts=0`.

## Current limitation — this fix does NOT solve everything

**It only sidesteps the WS event-loop block. If the Besu instance is broken in other ways, the fix can't help.**

We are currently seeing this on cluster 9-12. Ops 9 and 12 are healthy on the new image. Ops **10 and 11** still FATAL with `failed to sync historical registry events`. Direct empirical probe against Besu's HTTP shows:

```
eth_getLogs(0x327d4..0x3289b, SSV contract) → {"error":{"code":-32603,"message":"Internal error"}}
eth_getLogs(0x50000..0x5007f, SSV contract) → OK (0 events)
eth_getLogs(0x80000..0x8007f, SSV contract) → OK
eth_getLogs(0x100000..0x10007f, SSV contract) → OK
eth_getLogs(0x200000..0x20007f, SSV contract) → Internal error
eth_getLogs(0x2c0000..0x2c007f, SSV contract) → OK (4 events)
eth_syncing → false
eth_getBlockByNumber(0x327d4) → block header returned correctly
```

So this Besu instance has **interspersed corrupted receipt ranges**: block headers exist but receipts/logs for certain block ranges return `Internal error`. The corruption isn't a clean pruning boundary — it's specific damaged segments.

### Why this hits ops 10 and 11 specifically

Forensics from `kubectl get pvc`:

| Op | PVC age | Created | `last_processed_block` |
|---|---|---|---|
| 9 | 94 days | 2026-02-26 16:41 | 2,931,328 (current) |
| **10** | **2 days 22h** | **2026-05-29 14:35:47** | **0** |
| **11** | **2 days 22h** | **2026-05-29 14:36:21** | **0** |
| 12 | 94 days | 2026-02-26 16:43 | 2,931,328 (current) |

Ops 10 and 11 got brand-new PVCs **36 seconds apart** on 2026-05-29. The previous volumes (with populated event-syncer DBs) were lost. They're now trying to do a fresh historical sync from contract-deploy block 206,804 — which hits the first corrupted receipt range on Besu, returns `Internal error`, FATALs, restarts, repeats.

With ops 10/11 down, the 4-op cluster can't reach 3-of-4 QBFT quorum, so cluster 9-12 is at 0% effectiveness despite our code change working correctly. Two ops on the new image, all four needed.

### Estimated cause of the PVC loss

The StatefulSets in question use `storageClassName: local-storage` — **disk on the specific Kubernetes node**, not network-attached. With local-storage:

- The PV is pinned to a node via node affinity.
- If that node is drained, replaced, or has a disk failure, the PV becomes unreachable.
- The PVC eventually re-binds to a fresh empty volume on a different node.

The 36-second gap between ops 10 and 11's new PVC timestamps strongly suggests a **single node-level event** (drain, hardware failure, K8s upgrade) on 2026-05-29 ~14:35 UTC took down both at the same time. Ops 9 and 12 weren't on the affected node, so their PVCs are intact.

### Estimated cause of the Besu receipt corruption

Besu reports `eth_syncing: false` and serves block headers for the affected ranges — so it thinks it's healthy. But the receipts are missing/corrupted in **interspersed** ranges (not a clean prefix). Most likely:

- **Failed snap-sync segment**: during initial sync (snap mode), receipt segments are downloaded in parallel from peers. If a segment download was interrupted or written partially before a crash, Besu's Bonsai storage doesn't have great recovery semantics for partial-state corruption.
- **Interrupted write during a prior crash**: the Besu instance has been restarted several times over the past 6 days due to the WS event-loop issue cascading into pod restarts.

This is **separate from the WS bug**. It's a Besu data-integrity issue and needs a Besu-side fix (re-sync from scratch, restore from snapshot, or wait for upstream Besu to improve recovery semantics).

### Practical options for cluster 9-12

1. Point cluster 9-12 at an EL whose receipts aren't corrupted (e.g., one of the `geth-lh-hoodi-*` pairs). Two-line env-var change in the deploy. Ops 10/11 do the normal historical sync against a healthy EL.
2. Repair the Besu instance (re-sync from scratch or restore from a known-good snapshot).

The PR itself does not address either — they're operational fixes outside the scope of this code change. We just want to be explicit that the dual-transport split is **necessary but not sufficient** when the EL has data integrity issues.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./eth/executionclient/... ./cli/operator/...` clean
- [x] Existing executionclient tests still build & pass (no behavior change when `QueryAddr` is unset)
- [x] Deployed to staging clusters 1-4 and 93-96 against the buggy Besu — observed 0% → 97% effectiveness recovery in one epoch
- [x] Verified per-call routing empirically (`SyncProgress` succeeded in 2ms against the wedged-WS Besu — only possible if the call traveled HTTP)
- [x] Verified mixed-backend WARN path: redeployed 1-4 with Besu+HTTP and Geth-blank; WARN fired at startup naming the Geth backend in `ws_only_addrs`
- [ ] Reviewer to confirm: any caller still wired through `ec.client` (the wrapped `ethClient`) routes correctly via the new `queryOrSub()` shim

## Follow-ups (not in this PR)

- File upstream Besu issue (with patch suggestion: wrap `replyToClient`'s `JSON_OBJECT_WRITER.writeValue(...)` in `vertx.executeBlocking`, mirroring the streaming path in the same file). Same fix applies to `JsonRpcObjectExecutor.handleJsonObjectResponse` on the HTTP path.
- Add a runtime guard in `StreamBackpressure.awaitDrain` that throws if invoked from an event-loop context — would have caught PR #9848's regression in tests.
- Add a regression test that drives a large response with a slow-draining WS client and asserts the event loop is never parked.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -370,21 +370,37 @@ var StartNodeCmd = &cobra.Command{
 			logger.Fatal("no execution node address provided")
 		}
 
+		// Optional positionally-paired HTTP endpoints for request/response calls
+		// (eth_getLogs etc.) — keeps the heavy responses off WebSocket and works
+		// around Besu's StreamBackpressure event-loop block. Empty leaves the
+		// legacy single-transport behavior.
+		var queryAddrList []string
+		if cfg.ExecutionClient.QueryAddr != "" {
+			queryAddrList = strings.Split(cfg.ExecutionClient.QueryAddr, ";")
+		}
+
 		logger.Info("connecting EL(s)",
 			fields.Addresses(executionAddrList),
+			zap.Strings("query_addresses", queryAddrList),
 			zap.Duration("request_timeout", cfg.ExecutionClient.ConnectionTimeout),
 			zap.Uint64("sync_distance_tolerance", cfg.ExecutionClient.SyncDistanceTolerance),
 		)
 
 		var executionClient executionclient.Provider
 		if len(executionAddrList) == 1 {
+			singleOpts := []executionclient.Option{
+				executionclient.WithLogger(logger),
+				executionclient.WithReqTimeout(cfg.ExecutionClient.ConnectionTimeout),
+				executionclient.WithSyncDistanceTolerance(cfg.ExecutionClient.SyncDistanceTolerance),
+			}
+			if len(queryAddrList) > 0 && queryAddrList[0] != "" {
+				singleOpts = append(singleOpts, executionclient.WithQueryAddr(queryAddrList[0]))
+			}
 			ec, err := executionclient.New(
 				cmd.Context(),
 				executionAddrList[0],
 				ssvNetworkConfig.RegistryContractAddr,
-				executionclient.WithLogger(logger),
-				executionclient.WithReqTimeout(cfg.ExecutionClient.ConnectionTimeout),
-				executionclient.WithSyncDistanceTolerance(cfg.ExecutionClient.SyncDistanceTolerance),
+				singleOpts...,
 			)
 			if err != nil {
 				logger.Fatal("could not connect to execution client", zap.Error(err))
@@ -392,13 +408,19 @@ var StartNodeCmd = &cobra.Command{
 
 			executionClient = ec
 		} else {
+			multiOpts := []executionclient.OptionMulti{
+				executionclient.WithLoggerMulti(logger),
+				executionclient.WithReqTimeoutMulti(cfg.ExecutionClient.ConnectionTimeout),
+				executionclient.WithSyncDistanceToleranceMulti(cfg.ExecutionClient.SyncDistanceTolerance),
+			}
+			if len(queryAddrList) > 0 {
+				multiOpts = append(multiOpts, executionclient.WithQueryAddrsMulti(queryAddrList))
+			}
 			ec, err := executionclient.NewMulti(
 				cmd.Context(),
 				executionAddrList,
 				ssvNetworkConfig.RegistryContractAddr,
-				executionclient.WithLoggerMulti(logger),
-				executionclient.WithReqTimeoutMulti(cfg.ExecutionClient.ConnectionTimeout),
-				executionclient.WithSyncDistanceToleranceMulti(cfg.ExecutionClient.SyncDistanceTolerance),
+				multiOpts...,
 			)
 			if err != nil {
 				logger.Fatal("could not connect to execution client", zap.Error(err))
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -366,7 +366,10 @@ var StartNodeCmd = &cobra.Command{
 		}
 
 		executionAddrList := strings.Split(cfg.ExecutionClient.Addr, ";")
-		if len(executionAddrList) == 0 {
+		for i := range executionAddrList {
+			executionAddrList[i] = strings.TrimSpace(executionAddrList[i])
+		}
+		if len(executionAddrList) == 0 || executionAddrList[0] == "" {
 			logger.Fatal("no execution node address provided")
 		}
 
@@ -374,9 +377,41 @@ var StartNodeCmd = &cobra.Command{
 		// (eth_getLogs etc.) — keeps the heavy responses off WebSocket and works
 		// around Besu's StreamBackpressure event-loop block. Empty leaves the
 		// legacy single-transport behavior.
+		//
+		// Empty entries within the list are permitted and mean "use WS only for
+		// this position" — useful for mixed backend setups where only some ELs
+		// are Besu. MultiClient.connect handles the per-entry fallback.
 		var queryAddrList []string
 		if cfg.ExecutionClient.QueryAddr != "" {
 			queryAddrList = strings.Split(cfg.ExecutionClient.QueryAddr, ";")
+			for i := range queryAddrList {
+				queryAddrList[i] = strings.TrimSpace(queryAddrList[i])
+			}
+			// Pair-by-position: a silent length mismatch would leave some
+			// clients falling back to WS — the very situation
+			// ETH_1_QUERY_ADDR exists to avoid. Fail loud on misconfig.
+			if len(queryAddrList) != len(executionAddrList) {
+				logger.Fatal("ETH_1_QUERY_ADDR entry count must match ETH_1_ADDR — fix the config or unset ETH_1_QUERY_ADDR",
+					zap.Int("query_addr_count", len(queryAddrList)),
+					zap.Int("node_addr_count", len(executionAddrList)),
+				)
+			}
+			// Surface any positions that fall back to WS so operators see
+			// at startup that the Besu workaround isn't covering every
+			// backend. Not fatal — explicitly supported, but worth knowing
+			// because large eth_getLogs on these positions may still hit
+			// Besu's StreamBackpressure event-loop block.
+			var wsOnlyAddrs []string
+			for i, q := range queryAddrList {
+				if q == "" {
+					wsOnlyAddrs = append(wsOnlyAddrs, executionAddrList[i])
+				}
+			}
+			if len(wsOnlyAddrs) > 0 {
+				logger.Warn("ETH_1_QUERY_ADDR set but some positions are blank — those EL backends will use WS for queries and may trigger Besu's StreamBackpressure event-loop block on large eth_getLogs responses. Set an HTTP endpoint at each blank position to opt those backends into the dual-transport path.",
+					zap.Strings("ws_only_addrs", wsOnlyAddrs),
+				)
+			}
 		}
 
 		logger.Info("connecting EL(s)",
--- a/eth/executionclient/eth_client.go
+++ b/eth/executionclient/eth_client.go
@@ -25,17 +25,21 @@ import (
 // call falls back to subClient — preserving the pre-split single-client
 // behavior and keeping the change backward compatible.
 type ethClient struct {
-	sub   *ethclient.Client
-	query *ethclient.Client
+	sub      *ethclient.Client
+	subAddr  string // URL dialed for sub (always set)
+	query    *ethclient.Client
+	queryAddr string // URL dialed for query (empty when query is nil)
 
 	// reqTimeout specifies the default timeout to be used with every ethclient.Client call.
 	reqTimeout time.Duration
 }
 
-func newEthClient(sub, query *ethclient.Client, reqTimeout time.Duration) *ethClient {
+func newEthClient(sub *ethclient.Client, subAddr string, query *ethclient.Client, queryAddr string, reqTimeout time.Duration) *ethClient {
 	return &ethClient{
 		sub:        sub,
+		subAddr:    subAddr,
 		query:      query,
+		queryAddr:  queryAddr,
 		reqTimeout: reqTimeout,
 	}
 }
@@ -50,11 +54,59 @@ func (c *ethClient) queryOrSub() *ethclient.Client {
 	return c.sub
 }
 
+// SubAddr returns the URL of the subscription transport (the address dialed
+// for eth_subscribe — always WS or IPC).
+func (c *ethClient) SubAddr() string {
+	return c.subAddr
+}
+
+// QueryAddr returns the URL that actually serves query-routed calls
+// (FilterLogs, BlockNumber, HeaderByNumber, SyncProgress, ChainID): the
+// query client's address when dual-transport is configured, the sub client's
+// address otherwise. Use this when recording observability for query-routed
+// calls so logs/metrics report the URL the request truly traveled.
+func (c *ethClient) QueryAddr() string {
+	if c.query != nil {
+		return c.queryAddr
+	}
+	return c.subAddr
+}
+
+// QueryTransport returns "http" when dual-transport is configured (query
+// client present) and "ws" otherwise. Pair with QueryAddr() so observability
+// records can be filtered by transport regardless of operator URL choice.
+func (c *ethClient) QueryTransport() string {
+	if c.query == nil {
+		return "ws"
+	}
+	return "http"
+}
+
 func (c *ethClient) SyncProgress(ctx context.Context) (*ethereum.SyncProgress, error) {
 	reqCtx, cancel := context.WithTimeout(ctx, c.reqTimeout)
 	defer cancel()
 
 	return c.queryOrSub().SyncProgress(reqCtx)
+}
+
+// PingSub performs a cheap round-trip against the subscription transport only.
+// In single-transport mode (query == nil) this is a no-op — SyncProgress on
+// the sub client itself already covers liveness. When dual-transport is in
+// use, Healthy()'s SyncProgress probe only touches the query (HTTP) client,
+// so without this probe a WS-down / HTTP-up partial failure would report
+// healthy while SubscribeFilterLogs / SubscribeNewHead silently stall, and
+// MultiClient would never failover.
+//
+// Uses BlockNumber as the probe: tiny response (a uint64), can't trigger
+// the Besu WS-backpressure path we're avoiding for queries.
+func (c *ethClient) PingSub(ctx context.Context) error {
+	if c.query == nil {
+		return nil
+	}
+	reqCtx, cancel := context.WithTimeout(ctx, c.reqTimeout)
+	defer cancel()
+	_, err := c.sub.BlockNumber(reqCtx)
+	return err
 }
 
 func (c *ethClient) BlockNumber(ctx context.Context) (uint64, error) {
--- a/eth/executionclient/eth_client.go
+++ b/eth/executionclient/eth_client.go
@@ -10,72 +10,100 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
-// ethClient is a wrapper around ethclient.Client to add a default timeout for every call we
-// make with ethclient.Client. This allows us to manage timeouts for ethclient.Client in one
-// place such that we don't accidentally forget to set a timeout for some ethclient.Client call.
+// ethClient wraps two ethclient.Client instances and routes each call to the
+// transport best suited for it:
+//   - sub:   used only for streaming endpoints (eth_subscribe family). These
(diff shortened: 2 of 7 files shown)


eth/executionclient/execution_client.go
```diff
@@ -526,17 +533,34 @@ func (ec *ExecutionClient) streamLogsToChan(
...
+		reqStart = time.Now()
+		queryClient, err = ethclient.DialContext(reqCtx, ec.queryAddr)
+		recordRequest(ctx, ec.logger, "DialContext", ec, time.Since(reqStart), err)
+		if err != nil {
+			subClient.Close()
+			return ec.errSingleClient(fmt.Errorf("ethclient dial query addr: %w", err), "dial")
+		}
+	}
+
+	ec.client = newEthClient(subClient, queryClient, ec.reqTimeout)
```
momosh-ssv: Should we add a chain-ID consistency check between the two transports here? 
The only chain-ID assertion at startup (`multi_client.go` `assertSameChainID`) runs through `ChainID()` → `queryOrSub()`, so it validates the *query* client and never the sub client — and the single-client `New()` path has no chain-ID check at all. 

I fear if an operator points `ETH_1_ADDR` (WS) at one chain and the paired `ETH_1_QUERY_ADDR` (HTTP) at another, startup succeeds and we ingest registry events from the wrong chain while subscriptions track the other — the kind of discrepancy `streamLogsToChan` already warns can be slashing-adjacent. Seems like fetching `ChainID()` from both `subClient` and `queryClient` directly when `queryClient != nil` and failing connect on mismatch would close that window for one RPC at startup.

eth/executionclient/eth_client.go
```diff
@@ -10,72 +10,100 @@ import (
...
+	}
+	return c.sub
+}
+
 func (c *ethClient) SyncProgress(ctx context.Context) (*ethereum.SyncProgress, error) {
 	reqCtx, cancel := context.WithTimeout(ctx, c.reqTimeout)
 	defer cancel()
 
-	return c.client.SyncProgress(reqCtx)
+	return c.queryOrSub().SyncProgress(reqCtx)
```
momosh-ssv, line 89: Do we need to worry that health checks no longer cover the WS transport once a query addr is set? 

`Healthy()` → `SyncProgress()` routes through `queryOrSub()`, so when `query` is configured we only ever probe HTTP. 

I fear if the WS endpoint dies but HTTP stays up, the node reports healthy while `SubscribeNewHead` (pinned to `sub`) silently stalls — surfacing only via the 2-minute subscription timeout, and in MultiClient the client stays “healthy” so there's no failover. 
Maybe `Healthy()` should also probe `sub` (a cheap call) when `query != nil` and treat either transport down as unhealthy.

cli/operator/node.go
```diff
@@ -370,35 +370,57 @@ var StartNodeCmd = &cobra.Command{
 			logger.Fatal("no execution node address provided")
 		}
 
+		// Optional positionally-paired HTTP endpoints for request/response calls
+		// (eth_getLogs etc.) — keeps the heavy responses off WebSocket and works
+		// around Besu's StreamBackpressure event-loop block. Empty leaves the
+		// legacy single-transport behavior.
+		var queryAddrList []string
+		if cfg.ExecutionClient.QueryAddr != "" {
+			queryAddrList = strings.Split(cfg.ExecutionClient.QueryAddr, ";")
```
momosh-ssv, line 386: Might be worth validating the split count against `executionAddrList` here. 

As-is, a mismatch is silent: more query addrs than node addrs → extras ignored; fewer → trailing clients quietly fall back to WS so the Besu workaround just doesn't apply to them, with no warning. 

Whitespace also isn't trimmed, so `a; b` would dial `" b"`. Seems safer to `Fatal` on `len(queryAddrList) != len(executionAddrList)` (when non-empty) with a clear message, and `TrimSpace` each entry.

eth/executionclient/execution_client.go
```diff
@@ -526,17 +533,34 @@ func (ec *ExecutionClient) streamLogsToChan(
...
 	if err != nil {
 		return ec.errSingleClient(fmt.Errorf("ethclient dial: %w", err), "dial")
 	}
 
-	ec.client = newEthClient(c, ec.reqTimeout)
+	var queryClient *ethclient.Client
+	if ec.queryAddr != "" && ec.queryAddr != ec.nodeAddr {
+		reqStart = time.Now()
+		queryClient, err = ethclient.DialContext(reqCtx, ec.queryAddr)
+		recordRequest(ctx, ec.logger, "DialContext", ec, time.Since(reqStart), err)
```
momosh-ssv: Could we make the two dials distinguishable in metrics/logs? 

Both `DialContext` calls share the same `"DialContext"` route label, and `recordRequest` always tags `ec.nodeAddr` (the WS addr) regardless of which transport served the call — so we can't tell from observability whether the sub or query dial failed, or whether a `FilterLogs` actually went over HTTP vs fell back to WS. 
Given the silent fall-back paths above, that's exactly the signal operators would need when the split misbehaves. Maybe distinct labels (`DialContext.sub` / `DialContext.query`) plus a transport attribute when the query client serves a call.

eth/executionclient/execution_client.go
Roy-blox: mmmm I see what you're saying, it's a bit of an edge case but for sure worth adding

eth/executionclient/eth_client.go
Roy-blox: yeah makes sense, I'll add a check for the HTTP as well to get the healthy status, might require a better solution in the future but for now it should be good

cli/operator/node.go
Roy-blox: I'll fix the trimming and add a log to clarify that if we're working with multi client for example but don't have an http address assigned to the second/third EL it'll notify the user that we're only using WS for all EL interactions (shouldn't be an issue with non besu clients)

eth/executionclient/execution_client.go
Roy-blox: mmm this is absolutely right, seems that it was an issue before the change and it will need some more changes than expected but I believe it's better to bite the bullet now and do the changes in this PR rather than postpone it for a later version and have the observability issue of not knowing the fail origin (http or ws), I'll add a solution for it as well

Roy-blox:
Closing this PR.

After further review, this dual-transport approach turned out to be significantly more complex than the problem warrants. The Besu team has already addressed the WS event-loop bug on their side, so there's no need for us to carry this workaround in our codebase.

We'll rely on Besu's own fix rather than introducing the additional complexity of a dual-transport EL client here.

PR #2872, opened 2026-06-03
qbft/instance: defer round bump in UponRoundTimeout so the cutoff-boundary round-change is broadcast

iurii-ssv:
## What

At the terminal timeout (round `CutOffRound-1` → `CutOffRound`), `UponRoundTimeout` bumped `State.Round` **before** broadcasting the round change. The bump made the instance irrelevant (`State.Round == CutOffRound`, since `IsRelevant()` requires `State.Round < CutOffRound`), so the subsequent `Broadcast` rejected the message with an `InstanceStoppedProcessingMessagesErrorCode` ("instance is no longer considered relevant"). `UponRoundTimeout` then returned a spurious error and the final timeout-driven round-change was never sent.

The fix defers the bump (mirroring ssv-spec [`qbft/timeout.go`](https://github.com/ssvlabs/ssv-spec/blob/main/qbft/timeout.go) `UponRoundTimeout`, which bumps in a `defer`), so the round change is created and broadcast while `State.Round` is still `prevRound` (`< CutOffRound`, i.e. still relevant). Using `defer` also preserves the existing intentional *"always move on to the next round; the round-change broadcast is best-effort"* behavior (commit f977043b2): the bump still runs even when `CreateRoundChange`/`Broadcast` return early with an error.

## Impact

Low. This only occurs at the last round after ~11 consecutive timeouts, when the duty is already being abandoned cluster-wide. It does **not** affect healthy duties. It is a real deviation from the ssv-spec reference that produced a spurious error return plus a missing final round-change broadcast.

## Tests

- Adds `TestUponRoundTimeoutBroadcastsRoundChangeAtCutOffBoundary` — covers the `CutOffRound-1` → `CutOffRound` boundary (the real `CutOffRound` = 12, i.e. round 11→12): asserts the round-change **is** broadcast and `UponRoundTimeout` returns `nil`. The existing ssv-spec timeout tests cover rounds 1/2/3/5/15 but not this boundary, which is why it slipped through.
- Reworks `TestUponRoundTimeoutStopsProcessingAfterReachingCutOffRound` — it previously encoded the buggy behavior (error on the round that bumps *into* the cutoff). It now tests the genuine "stop" case: already *at* the cutoff round → no-op error, no broadcast, round unchanged.
- Updates `TestUponRoundTimeoutBumpsRound`'s in-broadcast assertions to reflect the spec ordering (round change created/broadcast while still in the previous round, before the deferred bump).

`make spec-test` (qbft + ssv spectest, `-race`) and `go test ./protocol/v2/qbft/...` pass.

### DIFF
--- a/protocol/v2/qbft/instance/timeout.go
+++ b/protocol/v2/qbft/instance/timeout.go
@@ -24,6 +24,17 @@ func (i *Instance) UponRoundTimeout(ctx context.Context, logger *zap.Logger) err
 	prevRound := i.State.Round
 	newRound := prevRound + 1
 
+	// Always move on to the next round. The round-change message broadcast is a best-effort thing, the QBFT
+	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
+	// for whatever reason — hence the bump is deferred, so it still runs even when CreateRoundChange/Broadcast
+	// below return early with an error.
+	//
+	// Deferring also keeps State.Round at prevRound while we create and broadcast the round change. Bumping
+	// eagerly would advance the instance into the cut-off round at the boundary (prevRound == CutOffRound-1),
+	// at which point Broadcast rejects the message as no-longer-relevant and this final round-change — the one
+	// the spec expects us to send — would never go out.
+	defer i.bumpToRound(newRound)
+
 	i.metrics.EndStage(ctx, prevRound)
 	i.metrics.StartStage(stageRoundChange)
 	i.metrics.RecordRoundChange(ctx, prevRound, reasonTimeout)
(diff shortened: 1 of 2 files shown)


iurii-ssv:
Closing as won't-fix.

The behavior here is real and off-spec, but it only differs at the round 11→12 cutoff boundary — the cluster-wide give-up point. Round == CutOffRound is a dead round (reaching it makes the instance irrelevant, so no node can decide there), so the dropped round-change has no liveness value; the only effects are an un-sent terminal RC and a non-retryable Debug-level error/trace. The eager-bump ordering also matches our partial-quorum path. Not worth a behavior change on a degenerate path.

Detail + acknowledgment of the original finding: https://github.com/ssvlabs/ssv/pull/2866#issuecomment-4612482365. Will add a clarifying comment in `timeout.go` so the boundary behavior reads as intentional.

PR #2873, opened 2026-06-03
qbft/instance: document the intentional cutoff-boundary round-change drop

iurii-ssv:
Comment-only. Documents the intentional cutoff-boundary behavior in `UponRoundTimeout` so it stops getting re-flagged (it's been independently raised more than once).

At the cutoff boundary (`State.Round == CutOffRound-1`), the eager `bumpToRound` advances the instance into `CutOffRound` before `Broadcast`, so the final round-change is dropped and `UponRoundTimeout` returns an error. This is intentional and inert: `CutOffRound` is the cluster-wide give-up round — no instance can decide at or past it — so the dropped round-change has no liveness value. (ssv-spec defers the bump and would broadcast it; the divergence is behaviorally inert at this give-up boundary.)

Adds a clarifying comment at the bump and in `TestUponRoundTimeoutStopsProcessingAfterReachingCutOffRound` (whose setup is the boundary case, not the already-at-cutoff case its name suggests).

**No behavior change.**

## Context
- Corrected acknowledgment of the finding: https://github.com/ssvlabs/ssv/pull/2866#issuecomment-4612482365
- The behavior-change alternative, closed won't-fix: #2872

### DIFF
--- a/protocol/v2/qbft/instance/timeout.go
+++ b/protocol/v2/qbft/instance/timeout.go
@@ -36,6 +36,12 @@ func (i *Instance) UponRoundTimeout(ctx context.Context, logger *zap.Logger) err
 	// Always move on to the next round. The round-change message broadcast is a best-effort thing, the QBFT
 	// cluster as a whole can progress further even if our round-change message cannot be created/broadcast
 	// for whatever reason.
+	//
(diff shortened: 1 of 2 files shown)


momosh-ssv (approved):

alok-ssv (approved):

PR #2874, opened 2026-06-04
refactor: replace []byte(fmt.Sprintf) with fmt.Appendf

solunolab:
I modified the return statement to utilize fmt.Appendf instead of fmt.Sprintf for creating the formatted byte slice. 
 
 This change enhances performance by directly appending the formatted string to the byte slice, avoiding the need for intermediate memory allocation created by fmt.Sprintf. 
 
 The resulting code is cleaner and more efficient, maintaining the same functionality while improving resource usage.
 
 More info can see https://github.com/golang/go/issues/47579

### DIFF
--- a/beacon/goclient/attest_test.go
+++ b/beacon/goclient/attest_test.go
@@ -556,7 +556,7 @@ func createBeaconServer(t *testing.T, options beaconServerResponseOptions) (*htt
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
-			resp := []byte(fmt.Sprintf(`{
+			resp := fmt.Appendf(nil, `{
 				"execution_optimistic": false,
 				"finalized": false,
 				"data": {
@@ -573,7 +573,7 @@ func createBeaconServer(t *testing.T, options beaconServerResponseOptions) (*htt
 					"root": "0x2922d4d36529c39ae7c463bc0a18f434d616954bdc0a38f7c24e0847a181de15",
 					"canonical": true
 				}
-			}`, options.SlotReturnedFromHeaderEndpoint))
+			}`, options.SlotReturnedFromHeaderEndpoint)
 			if _, err := w.Write(resp); err != nil {
 				w.WriteHeader(http.StatusInternalServerError)
 				return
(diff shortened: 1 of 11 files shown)


julienharbulot (approved):
Reviewed the diff and ran the targeted package tests successfully. No findings.

iurii-ssv (commented):
Thanks for the cleanup, and welcome! 🙏 The `fmt.Appendf(nil, …)` swap is correct and does avoid the extra `string` allocation + copy — nice.

One small thing, if you don't mind: in `utils/boot_node/node.go` the first write in the loop got converted, but the two a few lines below were left on the old pattern (I think because of the trailing `+"\n"` concatenation):

```go
write(w, []byte(fmt.Sprintf("UDP Port: %d", n.UDP())+"\n"))
write(w, []byte(fmt.Sprintf("TCP Port: %d", n.TCP())+"\n\n"))
```

Would you mind folding the suffix into the format string so the whole function stays consistent?

```go
write(w, fmt.Appendf(nil, "UDP Port: %d\n", n.UDP()))
write(w, fmt.Appendf(nil, "TCP Port: %d\n\n", n.TCP()))
```

One more thought, if you're up for it: could you try enabling `perfsprint` in `.golangci.yaml` and running `make lint` to see if it works? If it does cleanly, it'd let us keep this consistent automatically going forward 🙂 (no worries if it surfaces more than expected — happy to scope that separately). Everything else looks good. Thanks again! 🙌

solunolab:
> Thanks for the cleanup, and welcome! 🙏 The `fmt.Appendf(nil, …)` swap is correct and does avoid the extra `string` allocation + copy — nice.
> 

> 
> One more thought, if you're up for it: could you try enabling `perfsprint` in `.golangci.yaml` and running `make lint` to see if it works? If it does cleanly, it'd let us keep this consistent automatically going forward 🙂 (no worries if it surfaces more than expected — happy to scope that separately). Everything else looks good. Thanks again! 🙌


@iurii-ssv Thank you for your review!

I have updated the two remaining writes in utils/boot_node/node.go to use fmt.Appendf with the newline suffix folded into the format string.

I also tried enabling perfsprint and ran make lint. It works, but it surfaces 355 existing findings across the repo, so I left that out of this PR to keep the scope focused. Happy to handle that separately if desired.

iurii-ssv (approved):
LGTM

momosh-ssv (approved):
Verified the refactor: is semantically equivalent to `[]byte(fmt.Sprintf(…))` while skipping the intermediate string allocation + copy. Solid and clean.

Great job, thank you very much.

PR #2875, opened 2026-06-05
ssvsigner, runner: pin fixed-fork signing domains (voluntary exit + validator registration)

iurii-ssv:
## Summary

Voluntary exits signed via the remote signer (ssv-signer → Web3Signer) put the **current** fork in `fork_info` while computing the signing root over the **Capella** domain that [EIP-7044](https://eips.ethereum.org/EIPS/eip-7044) permanently pins for exits. The request was internally inconsistent. It worked only because a network-aware Web3Signer applies EIP-7044 itself and ignores the fork we send — a misconfigured one rejected it and the exit never signed.

This PR makes the remote request self-consistent (mirroring the already-correct local signer) and generalizes the fix to the fixed-fork signing domains.

## Root cause

SSV computes the exit signing root over the Capella domain on both paths, and always sends it as `signing_root`. Web3Signer recomputes its own root from `fork_info` + the exit and rejects with HTTP 400 on mismatch ([`Eth2SignForIdentifierHandler`](https://github.com/Consensys/web3signer/blob/master/core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java)):

- **Network-aware signer:** applies EIP-7044, uses Capella regardless of our `fork_info` → recomputed root matches → signs. This masked the inconsistency.
- **Misconfigured signer:** does not apply EIP-7044 for the exit's epoch and falls back to the fork we sent (current) → current-fork root → mismatch → **HTTP 400, exit never signs**.

The misconfiguration is easy to hit because Web3Signer's `--network` is **optional and silently defaults to `mainnet`**. On a young network like Hoodi (genesis March 2025), an operator who simply omits `--network` gets the mainnet fork schedule; Hoodi's current epoch is far below mainnet's Deneb epoch (269568), so the exit resolves to a **pre-Deneb** milestone, EIP-7044 is not applied, and Web3Signer derives the exit domain from the fork we send. That is the Hoodi incident — no exotic misconfiguration required, just a missing flag.

The local signer builds the exit domain from Capella directly, so it was never affected.

## Changes

- **Pin Capella for voluntary exits (the fix).** `prepareSignRequest` overrides `fork_info` with the Capella fork for `DomainVoluntaryExit`. The request is now self-consistent regardless of the signer's network-awareness: a correct signer still applies EIP-7044, and a signer that fell back to our fork now derives the Capella domain instead of rejecting. Removes SSV's reliance on the signer's EIP-7044 override.
- **Pin the genesis builder fork for validator registration (parity).** `DomainApplicationBuilder` is the other fixed-fork domain (genesis fork version + empty genesis validators root). Web3Signer derives this domain itself, so this is a no-op for a correct signer — purely request self-consistency and parity with the exit pin (no known incident).
- **Actionable error on exit signing failure.** A failed voluntary-exit signing is wrapped with a hint pointing at the Web3Signer `--network`, in `VoluntaryExitRunner.executeDuty` — the caller that already knows the duty type, so the generic signing helper stays domain-agnostic. Scoped to exits (the rare, high-signal case) so frequent registrations don't get a misleading hint.
- **Refactor.** Deduplicate the partial-signature verification helper (`ssvtypes.VerifyBeaconPartialSignature`) and the duplicate-signature resolution (`PartialSigContainer.ResolveDuplicateSignature`) into shared code — both were copied between `BaseRunner` and `CommitteeObserver`. Remove the dead `MockBeaconNetwork`.

## QA notes

**Reproduction** — Hoodi, remote signer, with `--network` omitted on Web3Signer (so it silently defaults to mainnet):

| | Voluntary exit | Attestations / aggregations (same window) |
|---|---|---|
| **Before** | duty dispatched to all operators never completes; Web3Signer returns **HTTP 400** (signing-root mismatch) | signing works fine |
| **After** | exit is signed and broadcast; duty completes, validator exits | signing works fine |

The "attestations succeed but the exit 400s" split is the tell that you are hitting this and not a generic signer outage: every other duty derives its domain from the fork the node sends, so only the EIP-7044 exit domain is affected.

**Automated coverage (added/strengthened):**

- Unit (`RemoteKeyManager`): assert the request carries the **Capella** fork for exits and the **genesis** fork + empty root for registrations, at a post-Capella (Fulu-era) slot so the wrong fork is observable — verified these fail without the fix.
- Unit (`ssvtypes.VerifyBeaconPartialSignature`): correct / wrong-root / unknown-signer.
- Unit (`PartialSigContainer.ResolveDuplicateSignature`): keep-valid-existing / replace-invalid-with-valid / drop-both-invalid.
- e2e (`ssvsigner/e2e/signing`): local, remote, and Web3Signer-direct agree on the Capella-domain exit and genesis-builder-domain registration signatures with the clock in the Fulu era. Run against Web3Signer 25.11.0 + Postgres + ssv-signer.
- `go build` / `go vet` (both modules), unit tests, the e2e suite, and `gofmt` / `goimports` all clean.

## Notes

- For a correctly network-configured Web3Signer this is a no-op for exits (it already applies EIP-7044); in that case the proximate cause is the Web3Signer config. The change removes SSV's reliance on signer network-awareness either way.
- "Wrong/missing `--network`" spans two buckets: a signer that resolves the exit epoch **pre-Deneb** falls back to our fork (fixed by the pin — the case hit on Hoodi today), while a signer network-aware for the **wrong** network applies EIP-7044 with that network's Capella version and still rejects (not fixed by the pin; the `--network` error hint is the only part that surfaces it).

### DIFF
--- a/protocol/v2/ssv/runner/voluntary_exit.go
+++ b/protocol/v2/ssv/runner/voluntary_exit.go
@@ -190,7 +190,9 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
 		spectypes.DomainVoluntaryExit,
 	)
 	if err != nil {
-		return fmt.Errorf("could not sign VoluntaryExit object: %w", err)
+		// EIP-7044 pins exits to the Capella domain; a remote signer that disagrees on the
+		// fork (e.g. a wrong Web3Signer --network) rejects exits while other duties still sign.
+		return fmt.Errorf("could not sign voluntary exit (the remote signer may be misconfigured, e.g. a wrong Web3Signer --network): %w", err)
 	}
 
 	msgs := &spectypes.PartialSignatureMessages{
--- a/protocol/v2/ssv/runner/voluntary_exit.go
+++ b/protocol/v2/ssv/runner/voluntary_exit.go
@@ -161,7 +161,7 @@ func (r *VoluntaryExitRunner) expectedPreConsensusRootsAndDomain() ([]ssz.HashRo
 
 // expectedPostConsensusRootsAndDomain an INTERNAL function, returns the expected post-consensus roots to sign
 func (r *VoluntaryExitRunner) expectedPostConsensusRootsAndDomain(context.Context) ([]ssz.HashRoot, phase0.DomainType, error) {
-	return nil, [4]byte{}, errors.New("no post consensus roots for voluntary exit")
+	return nil, spectypes.DomainError, errors.New("no post consensus roots for voluntary exit")
 }
 
 func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logger, duty spectypes.Duty) error {
@@ -190,9 +190,9 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
 		spectypes.DomainVoluntaryExit,
 	)
 	if err != nil {
-		// EIP-7044 pins exits to the Capella domain; a remote signer that disagrees on the
-		// fork (e.g. a wrong Web3Signer --network) rejects exits while other duties still sign.
-		return fmt.Errorf("could not sign voluntary exit (the remote signer may be misconfigured, e.g. a wrong Web3Signer --network): %w", err)
+		// EIP-7044 pins exits to the Capella domain, so a signer that disagrees on the fork
+		// rejects exits while other duties still sign (e.g. a remote Web3Signer with the wrong --network).
+		return fmt.Errorf("could not sign voluntary exit (if signing remotely, check the signer config, e.g. the Web3Signer --network): %w", err)
 	}
 
 	msgs := &spectypes.PartialSignatureMessages{
--- a/ssvsigner/e2e/signing/voluntary_exit_test.go
+++ b/ssvsigner/e2e/signing/voluntary_exit_test.go
@@ -1,0 +1,84 @@
+package signing
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/ssvlabs/ssv/ssvsigner/e2e"
+)
+
+// VoluntaryExitSigningTestSuite covers voluntary-exit signing, which (per EIP-7044)
+// must always use the Capella fork domain regardless of the current fork.
+type VoluntaryExitSigningTestSuite struct {
+	e2e.E2ETestSuite
+}
+
+// TestVoluntaryExitSigningConsistency verifies that LocalKeyManager, RemoteKeyManager
+// (via SSV-Signer + Web3Signer), and Web3Signer-direct all produce the same voluntary
+// exit signature, computed over the Capella domain — even though "now" is in the Fulu
+// era and the current fork differs from Capella. This mirrors the Hoodi incident where
+// the remote path was sending the current fork instead of Capella, and guards the
+// Capella fork pinning in RemoteKeyManager.prepareSignRequest.
+func (s *VoluntaryExitSigningTestSuite) TestVoluntaryExitSigningConsistency() {
+	// Place the chain clock in the Fulu era (mainnet Fulu @ epoch 411392) so the
+	// current fork (0x06000000) is well past Capella (0x03000000). Web3Signer is run
+	// with --network=mainnet, so at this epoch it applies the EIP-7044 override and
+	// signs exits with the Capella fork version.
+	testCurrentEpoch := phase0.Epoch(411392 + 256)
+	s.GetEnv().SetTestCurrentEpoch(testCurrentEpoch)
+
+	ctx, cancel := context.WithTimeout(s.GetContext(), 60*time.Second)
+	defer cancel()
+
+	validatorKeyPair := s.AddValidator(ctx)
+
+	exitEpoch := testCurrentEpoch
+	exitSlot := s.GetEnv().GetBeaconConfig().FirstSlotAtEpoch(exitEpoch)
+
+	voluntaryExit := &phase0.VoluntaryExit{
+		Epoch:          exitEpoch,
+		ValidatorIndex: 1,
+	}
+
+	// EIP-7044: exits are signed over the Capella domain, not the current fork's.
+	domain, err := s.CalculateVoluntaryExitDomain()
+	s.Require().NoError(err, "Failed to calculate voluntary exit domain")
+
+	s.T().Logf("✍️ Signing voluntary exit (epoch %d, current fork = Fulu) over the Capella domain", exitEpoch)
+
+	localSig, localRoot, err := s.GetEnv().GetLocalKeyManager().SignBeaconObject(
+		ctx, voluntaryExit, domain, validatorKeyPair.BLSPubKey, exitSlot, spectypes.DomainVoluntaryExit)
+	s.Require().NoError(err, "Local key manager failed to sign voluntary exit")
+	s.Require().NotEmpty(localSig)
+	s.Require().NotEmpty(localRoot)
+
+	remoteSig, remoteRoot, err := s.GetEnv().GetRemoteKeyManager().SignBeaconObject(
+		ctx, voluntaryExit, domain, validatorKeyPair.BLSPubKey, exitSlot, spectypes.DomainVoluntaryExit)
+	s.Require().NoError(err, "Remote key manager failed to sign voluntary exit")
+	s.Require().NotEmpty(remoteSig)
+	s.Require().NotEmpty(remoteRoot)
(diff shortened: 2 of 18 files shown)


ssvsigner/e2e/suite.go
```diff
@@ -119,6 +147,14 @@ func (s *E2ETestSuite) SignWeb3Signer(
 		req.Type = web3signer.TypeBlockV2
 		req.BeaconBlock = beaconBlockData
 
+	case spectypes.DomainVoluntaryExit:
+		data, ok := obj.(*phase0.VoluntaryExit)
+		if !ok {
+			return nil, phase0.Root{}, errors.New("could not cast obj to VoluntaryExit")
+		}
+		req.Type = web3signer.TypeVoluntaryExit
+		req.VoluntaryExit = data
```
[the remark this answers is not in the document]
iurii-ssv: Good catch — addressed in 0c6bf4a88. Added a comment documenting that this helper intentionally forwards the current fork (not Capella) to exercise Web3Signer's own EIP-7044 override, whereas the production RemoteKeyManager pins Capella via voluntaryExitForkInfo().

iurii-ssv:
@greptile pls re-review

protocol/v2/ssv/runner/voluntary_exit.go
```diff
@@ -190,7 +190,9 @@ func (r *VoluntaryExitRunner) executeDuty(ctx context.Context, logger *zap.Logge
 		spectypes.DomainVoluntaryExit,
 	)
 	if err != nil {
-		return fmt.Errorf("could not sign VoluntaryExit object: %w", err)
+		// EIP-7044 pins exits to the Capella domain; a remote signer that disagrees on the
+		// fork (e.g. a wrong Web3Signer --network) rejects exits while other duties still sign.
+		return fmt.Errorf("could not sign voluntary exit (the remote signer may be misconfigured, e.g. a wrong Web3Signer --network): %w", err)
```
julienharbulot: nit: This runner is signer-agnostic, so this error can also wrap LocalKeyManager failures where there’s no remote signer or Web3Signer `--network` to fix. Maybe you could keep the EIP-7044 hint but make it a bit more generic around signer/domain mismatch, or move the Web3Signer-specific hint closer to the remote signer path? Otherwise local signing/slashing/storage failures on voluntary exits might point operators at the wrong thing.

julienharbulot (dismissed):
No issues found, one small suggestion non-blocking.

protocol/v2/ssv/runner/voluntary_exit.go
iurii-ssv: Good point — `executeDuty` is signer-agnostic, so the unconditional "remote signer" hint could misdirect local-signer failures (and the non-signer `signBeaconObject` errors too: nil network config, beacon-domain fetch, unknown validator index). Made the Web3Signer hint conditional — `if signing remotely, check the signer config, e.g. the Web3Signer --network` — rather than moving it into the remote path, so the generic signing helper stays domain-agnostic and the hint stays at the duty-aware caller.

Addressed in c35b7c448 (which also folds in a tiny unrelated readability cleanup flagged in review: the post-consensus stub now returns `spectypes.DomainError` like its pre-consensus sibling and `CommitteeRunner`, instead of a bare `[4]byte{}` — never read in the error path, so it's a no-op).

ssvsigner/e2e/signing/voluntary_exit_test.go
```diff
@@ -0,0 +1,84 @@
...
+// TestVoluntaryExitSigningConsistency verifies that LocalKeyManager, RemoteKeyManager
+// (via SSV-Signer + Web3Signer), and Web3Signer-direct all produce the same voluntary
+// exit signature, computed over the Capella domain — even though "now" is in the Fulu
+// era and the current fork differs from Capella. This mirrors the Hoodi incident where
+// the remote path was sending the current fork instead of Capella, and guards the
+// Capella fork pinning in RemoteKeyManager.prepareSignRequest.
+func (s *VoluntaryExitSigningTestSuite) TestVoluntaryExitSigningConsistency() {
+	// Place the chain clock in the Fulu era (mainnet Fulu @ epoch 411392) so the
+	// current fork (0x06000000) is well past Capella (0x03000000). Web3Signer is run
+	// with --network=mainnet, so at this epoch it applies the EIP-7044 override and
```
momosh-ssv, line 35: Might be worth a note that this suite only exercises the EIP-7044-correct path: with `--network=mainnet` Web3Signer applies the override itself, so the roots match regardless of the Capella pin. 

The actual incident — a signer that falls back to our `fork_info` — isn't reproduced here; the Capella assertion in `remote_key_manager_test.go` is what guards the fix. Just so a future reader doesn't take this as reproducing the bug. Non-blocking.

momosh-ssv (dismissed):

alok-ssv (dismissed):
LGTM!

julienharbulot (dismissed):
previous comments fixed, looks good

ssvsigner/e2e/signing/voluntary_exit_test.go
iurii-ssv: Right — with `--network=mainnet` Web3Signer applies the EIP-7044 override itself at this epoch, so the suite passes with or without the pin; the doc comment overstated its role. Reworded it in 219749bf4dfaedd82917b7edb04f258f86193fdb to say the suite only exercises the EIP-7044-correct path and to point at the `SignVoluntaryExit` Capella assertion in `remote_key_manager_test.go` as what guards the pin.

alok-ssv (approved):

ovidiu-ssv-labs:
## QA — stage-hoodi, operators 5-8 (`COMMITTEE-5_6_7_8`)

Operator 5 runs **ssv-signer → Web3Signer**. Operators 6, 7, 8 use the local key manager.

---

### Scenario 1 — Before: correctly configured Web3Signer (`--network=hoodi`), `stage` branch
**Validator 838944**

Web3Signer applies EIP-7044 on its own and ignores the wrong fork SSV sends. All 4 operators signed and submitted. Bug was masked — not reproducible in this configuration.

---

### Scenario 2 — Before: misconfigured Web3Signer (no `--network`), `stage` branch
**Validator 839167**

With `--network` omitted, Web3Signer defaults to mainnet. Hoodi's current epoch is below mainnet's Deneb threshold, so EIP-7044 is not applied and the signer falls back to the current fork sent by SSV — which is not Capella. The signing root mismatches and Web3Signer rejects with **HTTP 400**.

- **Operator 5**: signing failed, exit dropped with `POST failed with status 400: invalid type: null, expected struct VoluntaryExit`
- **Operators 6, 7, 8**: signed and submitted successfully
- Exit was submitted by the 3-of-4 threshold — validator exited, but operator 5 did not contribute

> **Observability note:** In addition to the DEBUG-level HTTP 400, a WARN-level log is also emitted by the SSV node when ssv-signer returns a 500 back upstream: `response error for https://127.0.0.1:8090/v1/validators/sign/0xaa47c6ab…: unexpected status: 500`. A misconfigured Web3Signer is therefore visible even at INFO log level.

✅ **Bug confirmed.**

---

### Scenario 3 — After: misconfigured Web3Signer (no `--network`), fix branch
**Validator 838454**

With the fix, SSV pins the Capella fork in `fork_info` for `DomainVoluntaryExit`. The request is now self-consistent — even a signer that doesn't apply EIP-7044 derives the correct Capella domain from what SSV sends.

- **All 4 operators** (including operator 5): signed and submitted successfully, same reconstructed signature
- No HTTP 400 errors observed

✅ **Fix confirmed.**

---

### Scenario 4 — After: correctly configured Web3Signer (`--network=hoodi`), fix branch
**Validator 839544**

Verifies the fix doesn't break the happy path for a properly configured signer.

- **All 4 operators**: signed and submitted successfully

✅ **No regression.**

---

### Scenario 5 — After: node 6 P2P-silent, forced 3-of-4 consensus on nodes 5-7-8, fix branch
**Validator 838359**

Node 6 deployed with `P2P_SILENT_BROADCAST=true` — signs but broadcasts nothing. Forces the exit to be reconstructed from nodes 5, 7, 8 only, requiring node 5 to receive its own signed message back.

Node 5's quorum log: signers `[8, 7, 5]` — node 5's own partial signature arrived back via the self-receive path. All 4 nodes submitted the same reconstructed signature.

✅ **Self-receive on ssv-signer node works correctly.**

protocol/v2/ssv/runner/voluntary_exit.go
momosh-ssv: Thanks for catching this — the hint reads right now properly. Resolving.

ssvsigner/e2e/signing/voluntary_exit_test.go
momosh-ssv: Thanks — the reworded comment captures it exactly. Appreciate the quick turnaround — resolving.

momosh-ssv (approved):
gj!

ovidiu-ssv-labs (approved):

PR #2876, opened 2026-06-08
cli/operator,operator: converge startup Fatals into errgroup-coordinated shutdown

### LINKED ISSUE #2867: cli/operator: errgroup lifecycle — converge startup Fatals into a single handler
## Background

The startup refactor (#2866) reshaped the operator-node boot path into a thin `runNode()` → `newNode()` → `node.start()` flow (all in `cli/operator/`) and converted the **synchronous** startup steps from `logger.Fatal` to returned errors, surfaced once via the single top-level `Fatal` in `cli/operator/start_node.go`.

It deliberately left the **long-lived background goroutines** launched by `node.start()` (`cli/operator/node.go`) still calling `logger.Fatal` on failure. A background failure therefore `os.Exit`s deep in a goroutine — bypassing the CLI's `node.Close()` teardown (p2p + db) and the single top-level handler.

**What already landed (this branch).** Two pieces the original issue treated as pending are now done on the integration branch:

- **Both HTTP servers already have a ctx-aware shutdown.** The metrics handler (`Handler.Start` in `observability/metrics/handler.go`, via #2866) and the SSV API server (`Server.Start` in `api/server/server.go`, via #2819) each now bind, serve in the background, and run a `<-ctx.Done()` → `httpServer.Shutdown(ctx)` watcher; each returns the bound address plus a `serveErr` channel that emits a non-graceful serve error and is closed silently on a clean stop. So the "servers are abandoned on normal shutdown" problem is **already fixed** — on `ctx` cancel both shut down gracefully. (`origin/main` still has the old blocking `Handler.Start(mux, addr)` / `Server.Run`; the new ctx-aware shape lives on the #2866 + #2819 lineage this issue builds on.)

What remains is purely the **`Fatal` coordination**: the bind paths and the serve-loop goroutines still call `logger.Fatal`, and nothing joins the long-lived goroutines into a group — so any one failure crashes the process past `Close()` instead of returning up.

> Code references: `cli/operator` symbols (and the server `Start` methods) are from #2866 / #2819, not yet on `main`, so they're given by file + symbol name rather than `main` blob links.

## Goal

Make **"`Fatal` once"** literal within `cli/operator`: every failure path reachable from `start()` — whether a **synchronous** bind error or a **background-goroutine** error — returns up through `runNode()` to the single `Fatal` in `start_node.go`, with `node.Close()` running on the way out. The long-lived goroutines join an `errgroup` bound to the startup `ctx`; the first failure cancels the group and returns the error, while a plain `ctx` cancellation returns `nil`. The HTTP servers join via their existing `serveErr` channels (their `Shutdown` watcher already unblocks them on cancel).

## The `Fatal`s to convert (all in `cli/operator`)

Six sites, all reachable from `node.start()`:

| # | Site | Where | Kind |
|---|------|-------|------|
| 1 | metrics bind | `node.go` `start()` — `"failed to start metrics server"` | **synchronous** |
| 2 | metrics serve loop | `node.go` `start()` — `"metrics server serve loop exited"` | goroutine |
| 3 | SSV API bind | `node.go` `start()` — `"failed to start API server"` | **synchronous** |
| 4 | SSV API serve loop | `node.go` `start()` — `"API server serve loop exited"` | goroutine |
| 5 | ongoing event sync | `eventsync.go` `syncContractEvents`'s `SyncOngoing` goroutine — `"failed syncing ongoing registry events"` | goroutine |

## HTTP servers — already ctx-aware; only optional DRY left

Both servers already bind-and-serve in the background with a `<-ctx.Done()` → `Shutdown` watcher and a `serveErr` channel (metrics: `Handler.Start`, #2866; SSV API: `Server.Start`, #2819 — see Background). So **no further server-package change is functionally required** to join the errgroup: just consume the `serveErr` channel (#2 / #4 above).

What's left is **optional cleanup, not a blocker**: the two `Start` methods are near-identical, so a shared helper (e.g. `serveHTTP`) would DRY them. Mind one signature tension if you attempt it — a blocking `serveHTTP(ctx, srv) error` conflicts with the current `Start(ctx) (boundAddr, <-chan error, err)` shape, which returns the `:0`-resolved bound address *before* serving (tests depend on this). Reconcile with that shape rather than reintroducing a blocking signature.

The WS server (`wsServer.Start`, started inside `operator.Node.Start` in `operator/node.go`, with its own goroutine `Fatal`) is **out of scope** — it's operator-package, not `cli/operator`.

## Proposed approach

1. `g, gctx := errgroup.WithContext(ctx)` in `start()`.
2. `g.Go(...)` each long-lived service: the metrics + SSV-API serve loops (`return <-serveErr`), the health prober, and the ongoing event sync. Convert the two synchronous bind errors (#1, #3) to `return err`.
3. Run the node as a group member too — `g.Go(func() error { return n.operatorNode.Start(gctx) })` — then `return g.Wait()`. (Don't `return n.operatorNode.Start(...)` *before* `g.Wait()`: that skips the group wait when `Start` itself errors.) Note `operatorNode.Start` blocks on `dutyScheduler.Wait()` for **operator** nodes — only exporter-standard blocks on `<-ctx.Done()` — and its clean-cancel→`nil` behavior is **owned by the `operator` package, not this issue** (see Out of scope).
4. Have `syncContractEvents` **return a start-func** for `SyncOngoing` rather than spawning it + `Fatal` internally, so `start()` can errgroup it.
5. Convert `startHealthProber` to return an error: clean `gctx.Done()` → `nil` (exit 0); persistent unhealth → non-zero error so the node still restarts (see *Health prober*). Extract the shared `probe` core used by both it and `ensureComponentsHealthy`.
6. **(Optional)** DRY the two server `Start` methods behind a shared `serveHTTP` helper (see *HTTP servers*). The ctx-aware serve + shutdown itself is already done — this step is cleanup only.

## Normal-shutdown invariant + test

Every joined goroutine MUST return `nil` on plain `ctx` cancellation (not a fatal-worthy error) — otherwise a clean SIGINT logs a spurious fatal. Add a test that cancels `ctx` and asserts `runNode()` / `start()` returns `nil`.

This is **harder than it first looks**: the existing in-process smoke tests (`cli/operator/node_test.go`) deliberately stop at `newNode()` and never enter `start()`. Driving `start()` to completion does real p2p `Setup`/`Start` *and* runs `operatorNode.Start` (the duty scheduler). `recordingP2PNetwork` already stubs the network, but `operatorNode` is a concrete `*operator.Node` with no test seam. So either:

- introduce an `operatorNode` interface seam (count this as part of this issue's work), or
- scope the test to the errgroup helper / prober / `serveHTTP` in isolation rather than full `start()`.

## Done when

- No `logger.Fatal` reachable from `cli/operator`'s `start()` — synchronous *or* in a goroutine — except `start_node.go`'s single top-level handler.
- A background-service failure (or a synchronous bind error) returns through the errgroup → `runNode()` → the single `Fatal`, with `node.Close()` running.
- Health-prober trip on persistent unhealth still terminates the node (non-zero exit → restart), now via `Close()` rather than `os.Exit`; a clean cancel exits 0.
- SIGINT / ctx-cancel shuts the metrics + SSV-API HTTP servers down gracefully (already true) and `runNode()` returns `nil`.
- Normal-shutdown test passes; gofmt / build / vet / test / lint green.

---
_Follow-up to #2866. #2819 (server graceful-shutdown) is merged — consume its `serveErr` surface rather than re-implementing it._

### LINKED ISSUE #2880: hprobe: ProbeAll surfaces its own early-cancel context.Canceled in the joined error
### Summary

`ProbeAll` probes components in parallel and, the moment one fails, cancels the rest to quit early ([prober.go#L49](https://github.com/ssvlabs/ssv/blob/95d33d04a/hprobe/prober.go#L49) `ctx, cancel := context.WithCancel(ctx)`, then `cancel()` at [L64](https://github.com/ssvlabs/ssv/blob/95d33d04a/hprobe/prober.go#L64)). A sibling probe cut short that way returns `context.Canceled` (via `probeComponent`'s retry-delay `return ctx.Err()` at [L130](https://github.com/ssvlabs/ssv/blob/95d33d04a/hprobe/prober.go#L130)), and that error gets `errors.Join`'d into `ProbeAll`'s result ([L76–77](https://github.com/ssvlabs/ssv/blob/95d33d04a/hprobe/prober.go#L76)) alongside the genuine failure.

Because `errors.Is` on a `Join` is true if **any** branch matches, `ProbeAll`'s returned error can satisfy `errors.Is(err, context.Canceled)` even when a real component failure is present.

### Impact

A caller that uses `errors.Is(err, context.Canceled)` to treat cancellation as a clean stop would then **suppress a genuine failure** (false negative). Example: the EL is down (real failure) while the event-syncer is mid-retry; the EL's failure early-cancels the event-syncer, which returns `context.Canceled`; the joined result is `Join("EL: connection refused", "event-syncer: context canceled")`, so `errors.Is(…, context.Canceled)` is true and the real failure is masked.

Not an active bug for the current callers — `startHealthProber` keys off the parent ctx (`ctx.Err()`), not the error value, and `ensureComponentsHealthy` just propagates the error — so neither is fooled today. Filing as hygiene and to protect future callers.

### Inconsistency

`probeComponent` already swallows `errors.Is(err, context.Canceled)` from a component's `Healthy()` as "not an error" ([L111](https://github.com/ssvlabs/ssv/blob/95d33d04a/hprobe/prober.go#L111)). The same "a cancellation is not a verdict" rule should apply when joining sibling results.

### Proposed fix

Skip `context.Canceled` when joining:

```go
for err := range errsCh {
    if errors.Is(err, context.Canceled) {
        continue // sibling cut short by our own early-cancel, not a health verdict
    }
    errs = errors.Join(errs, err)
}
```

So `ProbeAll` reports only genuine failures, never its own quit-early cancellations.

iurii-ssv:
## Summary

Makes **"Fatal once" literal** for the operator-node boot path: every failure reachable from startup — a synchronous bind error or a long-lived background goroutine — converges on the single top-level `Fatal` in `cli/operator/start_node.go`, with `node.close()` (p2p + EL + db teardown) running on the way out. Previously these failures called `logger.Fatal` deep in goroutines, `os.Exit`-ing the process and bypassing teardown.

Implements #2867 and folds in the operator-package follow-up it deferred. Builds on the ctx-aware HTTP servers from #2866 and #2819.

### `cli/operator` — construct vs. run

The old monolithic `runNode` is split into two peer phases, called one after another:

- **`buildNode`** constructs the node graph: validates config up front, connects the CL/EL clients, and wires everything via `newNode` — which is now **pure construction and starts no goroutines**.
- **`runNode`** owns the lifecycle: it derives a `context.WithCancelCause` root, runs every long-lived service through a `spawn` helper (the first to fail cancels the root with its error as the *cause*), blocks on the first terminal event, then tears the node down — waiting for the services to unwind, then `node.close()` — within a single grace window (`shutdownGraceTimeout`), and returns the cause. A service that ignores cancellation can't hang shutdown past the window.
- **`node.start`** is the bring-up recipe: the synchronous steps (health gate, historical contract sync, metadata sync, network) plus `spawn` of the long-lived services. The background services formerly launched *inside* `newNode` (the VRSubmitter; in exporter modes the duty-trace collector / slot-pruning) now start here through `spawn`, so they bind to the supervised root and are awaited at teardown.
- The two synchronous bind `Fatal`s (metrics, SSV-API) and the prober / event-sync `Fatal`s become returned errors that surface as the terminal cause.

### `start-node` — signal handling

A two-stage `signal.Notify` handler (not `NotifyContext`, which can't express the escape hatch): the **first** SIGINT/SIGTERM cancels the node ctx for a graceful unwind (services stop, `node.close()` runs, exit 0); a **second** one force-exits via `Fatal`. The node previously had no signal handling at all (`cmd.Context()` was `context.Background()`), so a deliberate stop terminated abruptly and never ran teardown. The top-level handler classifies off the signal ctx: a deliberate stop exits 0 (even if an error coincided), a genuine failure `Fatal`s (non-zero, so the orchestrator restarts).

### `operator.Node.Start`

Gets its own `errgroup`: the WS serve-loop and `dutyScheduler.Wait` join as members; the scheduler-nil `Fatal`s become returned errors.

### prober (`hprobe` + `cli/operator`)

- `ProbeAll` returns when every probe finishes **or** ctx expires, whichever comes first — a `Healthy` impl that ignores its ctx can no longer wedge the round (and the bring-up gate / watchdog with it).
- Fixes #2880: a real failure that early-cancels a sibling no longer satisfies `errors.Is(err, context.Canceled)` (the false-negative hazard for callers); a regression test pins the exact scenario.
- The runtime watchdog returns an error (→ graceful close, non-zero exit) instead of `Fatal`-ing from its loop.

### `protocol/v2/ssv/runner`, `exporter/api`

- `NewVRSubmitter` splits its constructor from a `Start(ctx)` method instead of launching a goroutine from the constructor — so `newNode` stays goroutine-free.
- The WS server takes its ctx at `Start(ctx)` (symmetric with the metrics/SSV-API servers) rather than at construction.

## Behavior notes

- **Graceful shutdown:** SIGINT/SIGTERM → the lifecycle ctx is canceled → every long-lived service returns → `node.close()` runs → exit 0. A genuine background-service failure brings the node down through `node.close()` with a non-zero exit, so the orchestrator restarts it.
- **Grace-bounded teardown:** the whole teardown (services unwinding + `node.close()`) is bounded; past the window the node surfaces the terminal cause rather than hanging on something ignoring cancellation (leftovers are reaped at process exit).

Closes #2867.

### DIFF
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -67,6 +67,22 @@ func syncContractEvents(
 		eventsyncer.WithLogger(logger),
 	)
 
+	// load & parse local events yaml if exists, otherwise sync from contract
+	if len(cfg.LocalEventsPath) != 0 {
+		localEvents, err := localevents.Load(cfg.LocalEventsPath)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to load local events: %w", err)
+		}
+
+		if err := eventHandler.HandleLocalEvents(ctx, localEvents); err != nil {
+			return nil, nil, fmt.Errorf("error occurred while running event data handler: %w", err)
+		}
+
+		// No ongoing sync in local-events mode.
+		return eventSyncer, nil, nil
+	}
+
+	// Determine the block to resume contract sync from.
 	fromBlock, found, err := nodeStorage.GetLastProcessedBlock(nil)
 	if err != nil {
 		return nil, nil, fmt.Errorf("syncing registry contract events failed, could not get last processed block: %w", err)
@@ -80,82 +96,67 @@ func syncContractEvents(
 		fromBlock = new(big.Int).SetUint64(fromBlock.Uint64() + 1)
 	}
 
-	// Set only on the contract-sync path below; nil in local-events mode (no ongoing sync there).
-	var startOngoingSync func(context.Context) error
-
-	// load & parse local events yaml if exists, otherwise sync from contract
-	if len(cfg.LocalEventsPath) != 0 {
-		localEvents, err := localevents.Load(cfg.LocalEventsPath)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to load local events: %w", err)
-		}
-
-		if err := eventHandler.HandleLocalEvents(ctx, localEvents); err != nil {
-			return nil, nil, fmt.Errorf("error occurred while running event data handler: %w", err)
-		}
-	} else {
-		// Sync historical registry events.
-		logger.Debug("syncing historical registry events", zap.Uint64("fromBlock", fromBlock.Uint64()))
-		lastProcessedBlock, err := eventSyncer.SyncHistory(ctx, fromBlock.Uint64())
-		switch {
-		case errors.Is(err, executionclient.ErrNothingToSync):
-			// Nothing was synced, keep fromBlock as is.
-			logger.Info("finished syncing historical events, nothing to sync",
-				zap.Uint64("from_block", fromBlock.Uint64()),
-			)
-		case err == nil:
-			// Successfully synced up to a fresh block, advance fromBlock to the block after lastProcessedBlock.
-			logger.Info("finished syncing historical events to a fresh block",
-				zap.Uint64("from_block", fromBlock.Uint64()),
-				zap.Uint64("last_processed_block", lastProcessedBlock),
-			)
-			fromBlock = new(big.Int).SetUint64(lastProcessedBlock + 1)
-		default:
-			return nil, nil, fmt.Errorf("failed to sync historical registry events: %w", err)
-		}
-
-		// Print registry stats.
-		shares := nodeStorage.Shares().List(nil)
-		operators, err := nodeStorage.ListOperatorsAll(nil)
-		if err != nil {
-			logger.Error("failed to get operators", zap.Error(err))
-		}
-
-		operatorValidators := 0
-		liquidatedValidators := 0
-		operatorID := operatorDataStore.GetOperatorID()
-		if operatorDataStore.OperatorIDReady() {
-			for _, share := range shares {
-				if share.BelongsToOperator(operatorID) {
-					operatorValidators++
-				}
-				if share.Liquidated {
-					liquidatedValidators++
-				}
-			}
-		}
-		logger.Info("historical registry sync stats",
-			zap.Uint64("my_operator_id", operatorID),
-			zap.Int("operators", len(operators)),
-			zap.Int("validators", len(shares)),
-			zap.Int("liquidated_validators", liquidatedValidators),
-			zap.Int("my_validators", operatorValidators),
+	// Sync historical registry events.
+	logger.Debug("syncing historical registry events", zap.Uint64("fromBlock", fromBlock.Uint64()))
+	lastProcessedBlock, err := eventSyncer.SyncHistory(ctx, fromBlock.Uint64())
+	switch {
+	case errors.Is(err, executionclient.ErrNothingToSync):
+		// Nothing was synced, keep fromBlock as is.
+		logger.Info("finished syncing historical events, nothing to sync",
+			zap.Uint64("from_block", fromBlock.Uint64()),
 		)
+	case err == nil:
+		// Successfully synced up to a fresh block, advance fromBlock to the block after lastProcessedBlock.
+		logger.Info("finished syncing historical events to a fresh block",
+			zap.Uint64("from_block", fromBlock.Uint64()),
+			zap.Uint64("last_processed_block", lastProcessedBlock),
+		)
+		fromBlock = new(big.Int).SetUint64(lastProcessedBlock + 1)
+	default:
+		return nil, nil, fmt.Errorf("failed to sync historical registry events: %w", err)
+	}
 
-		// Sync ongoing registry events; the node must terminate if this stops. It can't operate
-		// without staying current with Ethereum events, and until reorg handling exists, restarting
-		// from persisted state is safer than continuing on possibly-incorrect state. A clean ctx
-		// cancellation returns nil; any other error is returned as a startupError carrying
-		// last_processed_block.
-		startOngoingSync = func(ctx context.Context) error {
-			if err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64()); err != nil && !errors.Is(err, context.Canceled) {
-				return startupError{
-					err:    fmt.Errorf("failed syncing ongoing registry events: %w", err),
-					fields: []zap.Field{zap.Uint64("last_processed_block", lastProcessedBlock)},
-				}
+	// Print registry stats.
+	shares := nodeStorage.Shares().List(nil)
+	operators, err := nodeStorage.ListOperatorsAll(nil)
+	if err != nil {
+		logger.Error("failed to get operators", zap.Error(err))
+	}
+
+	operatorValidators := 0
+	liquidatedValidators := 0
+	operatorID := operatorDataStore.GetOperatorID()
+	if operatorDataStore.OperatorIDReady() {
+		for _, share := range shares {
+			if share.BelongsToOperator(operatorID) {
+				operatorValidators++
 			}
-			return nil
+			if share.Liquidated {
+				liquidatedValidators++
+			}
 		}
+	}
+	logger.Info("historical registry sync stats",
+		zap.Uint64("my_operator_id", operatorID),
+		zap.Int("operators", len(operators)),
+		zap.Int("validators", len(shares)),
+		zap.Int("liquidated_validators", liquidatedValidators),
+		zap.Int("my_validators", operatorValidators),
+	)
+
+	// Sync ongoing registry events; the node must terminate if this stops. It can't operate
+	// without staying current with Ethereum events, and until reorg handling exists, restarting
+	// from persisted state is safer than continuing on possibly-incorrect state. A clean ctx
+	// cancellation returns nil; any other error is returned as a startupError carrying
+	// last_processed_block.
+	startOngoingSync := func(ctx context.Context) error {
+		if err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64()); err != nil && !errors.Is(err, context.Canceled) {
+			return startupError{
+				err:    fmt.Errorf("failed syncing ongoing registry events: %w", err),
+				fields: []zap.Field{zap.Uint64("last_processed_block", lastProcessedBlock)},
+			}
+		}
+		return nil
 	}
 
 	return eventSyncer, startOngoingSync, nil
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -20,10 +20,10 @@ import (
 	"github.com/ssvlabs/ssv/ssvsigner/ekm"
 )
 
-// syncContractEvents blocks until historical events are synced, then returns the event syncer along
-// with a start-func for ongoing event sync (nil in local-events mode). The caller runs the start-func
-// as a long-lived errgroup member so an ongoing-sync failure returns up to the single top-level Fatal
-// (with Close running) instead of os.Exit-ing the process from inside a goroutine.
+// syncContractEvents blocks until historical events are synced, then returns the event syncer plus a
+// start-func for ongoing event sync (nil in local-events mode). The caller runs the start-func as an
+// errgroup member, so an ongoing-sync failure surfaces up to the top-level Fatal instead of
+// os.Exit-ing from a goroutine.
 func syncContractEvents(
 	ctx context.Context,
 	logger *zap.Logger,
@@ -81,8 +81,7 @@ func syncContractEvents(
 		fromBlock = new(big.Int).SetUint64(fromBlock.Uint64() + 1)
 	}
 
-	// Set only in the contract-sync path below; stays nil in local-events mode, where there is no
-	// ongoing sync to run.
+	// Set only on the contract-sync path below; nil in local-events mode (no ongoing sync there).
 	var startOngoingSync func(context.Context) error
 
 	// load & parse local events yaml if exists, otherwise sync from contract
@@ -144,12 +143,11 @@ func syncContractEvents(
 			zap.Int("my_validators", operatorValidators),
 		)
 
-		// Sync ongoing registry events. Terminate the node if it stops: the node can't operate without
+		// Sync ongoing registry events; terminate the node if it stops. The node can't operate without
 		// staying current with Ethereum events, and until reorg handling exists, restarting from
 		// persisted state is safer than continuing on possibly-incorrect state. A clean ctx
-		// cancellation (normal shutdown) returns nil; any other failure returns up to the single
-		// top-level Fatal (with Close running). startupError carries last_processed_block so that
-		// context is preserved through the returned-error path.
+		// cancellation returns nil; any other error surfaces up to the top-level Fatal, with
+		// last_processed_block preserved via startupError.
 		startOngoingSync = func(ctx context.Context) error {
 			if err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64()); err != nil && !errors.Is(err, context.Canceled) {
 				return startupError{
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -21,9 +21,8 @@ import (
 )
 
 // syncContractEvents blocks until historical events are synced, then returns the event syncer plus a
-// start-func for ongoing event sync (nil in local-events mode). The caller runs the start-func as an
-// errgroup member, so an ongoing-sync failure surfaces up to the top-level Fatal instead of
-// os.Exit-ing from a goroutine.
+// start-func for ongoing event sync (nil in local-events mode). The start-func runs until a clean
+// ctx cancellation (→ nil) or a sync failure (→ a startupError carrying last_processed_block).
 func syncContractEvents(
 	ctx context.Context,
 	logger *zap.Logger,
@@ -143,11 +142,11 @@ func syncContractEvents(
 			zap.Int("my_validators", operatorValidators),
 		)
 
-		// Sync ongoing registry events; terminate the node if it stops. The node can't operate without
-		// staying current with Ethereum events, and until reorg handling exists, restarting from
-		// persisted state is safer than continuing on possibly-incorrect state. A clean ctx
-		// cancellation returns nil; any other error surfaces up to the top-level Fatal, with
-		// last_processed_block preserved via startupError.
+		// Sync ongoing registry events; the node must terminate if this stops. It can't operate
+		// without staying current with Ethereum events, and until reorg handling exists, restarting
+		// from persisted state is safer than continuing on possibly-incorrect state. A clean ctx
+		// cancellation returns nil; any other error is returned as a startupError carrying
+		// last_processed_block.
 		startOngoingSync = func(ctx context.Context) error {
 			if err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64()); err != nil && !errors.Is(err, context.Canceled) {
 				return startupError{
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -20,7 +20,10 @@ import (
 	"github.com/ssvlabs/ssv/ssvsigner/ekm"
 )
 
-// syncContractEvents blocks until historical events are synced and then spawns a goroutine syncing ongoing events.
+// syncContractEvents blocks until historical events are synced, then returns the event syncer along
+// with a start-func for ongoing event sync (nil in local-events mode). The caller runs the start-func
+// as a long-lived errgroup member so an ongoing-sync failure returns up to the single top-level Fatal
+// (with Close running) instead of os.Exit-ing the process from inside a goroutine.
 func syncContractEvents(
 	ctx context.Context,
 	logger *zap.Logger,
@@ -32,15 +35,15 @@ func syncContractEvents(
 	operatorDataStore operatordatastore.OperatorDataStore,
 	keyManager ekm.KeyManager,
 	doppelgangerHandler eventhandler.DoppelgangerProvider,
-) (*eventsyncer.EventSyncer, error) {
+) (*eventsyncer.EventSyncer, func(context.Context) error, error) {
 	eventFilterer, err := executionClient.Filterer()
 	if err != nil {
-		return nil, fmt.Errorf("failed to set up event filterer: %w", err)
+		return nil, nil, fmt.Errorf("failed to set up event filterer: %w", err)
 	}
 
 	eventParser, err := eventparser.New(eventFilterer)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create event parser: %w", err)
+		return nil, nil, fmt.Errorf("failed to create event parser: %w", err)
 	}
 
 	eventHandler, err := eventhandler.New(
@@ -55,7 +58,7 @@ func syncContractEvents(
 		eventhandler.WithLogger(logger),
 	)
 	if err != nil {
-		return nil, fmt.Errorf("failed to setup event data handler: %w", err)
+		return nil, nil, fmt.Errorf("failed to setup event data handler: %w", err)
 	}
 
 	eventSyncer := eventsyncer.New(
@@ -67,26 +70,30 @@ func syncContractEvents(
 
 	fromBlock, found, err := nodeStorage.GetLastProcessedBlock(nil)
 	if err != nil {
-		return nil, fmt.Errorf("syncing registry contract events failed, could not get last processed block: %w", err)
+		return nil, nil, fmt.Errorf("syncing registry contract events failed, could not get last processed block: %w", err)
 	}
 	if !found {
 		fromBlock = networkConfig.RegistrySyncOffset
 	} else if fromBlock == nil {
-		return nil, fmt.Errorf("syncing registry contract events failed, last processed block is nil")
+		return nil, nil, fmt.Errorf("syncing registry contract events failed, last processed block is nil")
 	} else {
 		// Start syncing from the next block.
 		fromBlock = new(big.Int).SetUint64(fromBlock.Uint64() + 1)
 	}
 
+	// Set only in the contract-sync path below; stays nil in local-events mode, where there is no
+	// ongoing sync to run.
+	var startOngoingSync func(context.Context) error
+
 	// load & parse local events yaml if exists, otherwise sync from contract
 	if len(cfg.LocalEventsPath) != 0 {
 		localEvents, err := localevents.Load(cfg.LocalEventsPath)
 		if err != nil {
-			return nil, fmt.Errorf("failed to load local events: %w", err)
+			return nil, nil, fmt.Errorf("failed to load local events: %w", err)
 		}
 
 		if err := eventHandler.HandleLocalEvents(ctx, localEvents); err != nil {
-			return nil, fmt.Errorf("error occurred while running event data handler: %w", err)
+			return nil, nil, fmt.Errorf("error occurred while running event data handler: %w", err)
 		}
 	} else {
 		// Sync historical registry events.
@@ -106,7 +113,7 @@ func syncContractEvents(
 			)
 			fromBlock = new(big.Int).SetUint64(lastProcessedBlock + 1)
 		default:
-			return nil, fmt.Errorf("failed to sync historical registry events: %w", err)
+			return nil, nil, fmt.Errorf("failed to sync historical registry events: %w", err)
 		}
 
 		// Print registry stats.
@@ -137,19 +144,22 @@ func syncContractEvents(
 			zap.Int("my_validators", operatorValidators),
 		)
 
-		// Sync ongoing registry events in the background. Crash if it stops: the node can't operate
-		// without staying current with Ethereum events, and until reorg handling exists, restarting
-		// from persisted state is safer than continuing on possibly-incorrect state.
-		go func() {
-			err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64())
-			if err != nil && !errors.Is(err, context.Canceled) {
-				logger.Fatal("failed syncing ongoing registry events",
-					zap.Uint64("last_processed_block", lastProcessedBlock),
-					zap.Error(err),
-				)
+		// Sync ongoing registry events. Terminate the node if it stops: the node can't operate without
+		// staying current with Ethereum events, and until reorg handling exists, restarting from
+		// persisted state is safer than continuing on possibly-incorrect state. A clean ctx
+		// cancellation (normal shutdown) returns nil; any other failure returns up to the single
+		// top-level Fatal (with Close running). startupError carries last_processed_block so that
+		// context is preserved through the returned-error path.
+		startOngoingSync = func(ctx context.Context) error {
+			if err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64()); err != nil && !errors.Is(err, context.Canceled) {
+				return startupError{
+					err:    fmt.Errorf("failed syncing ongoing registry events: %w", err),
+					fields: []zap.Field{zap.Uint64("last_processed_block", lastProcessedBlock)},
+				}
 			}
-		}()
+			return nil
+		}
 	}
 
-	return eventSyncer, nil
+	return eventSyncer, startOngoingSync, nil
 }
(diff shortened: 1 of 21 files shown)


cli/operator/prober.go
```diff
@@ -27,45 +27,54 @@ const (
...
+		if err := probe(ctx, p, probeFrequency); err != nil {
+			// A canceled ctx means we're shutting down, not that a component is unhealthy: the probe
+			// inherits ctx, so cancellation surfaces here as a probe error. Return nil (exit 0) rather
+			// than tripping the watchdog (non-zero exit).
+			if ctx.Err() != nil {
+				return nil
 			}
-		}()
+			return fmt.Errorf("%s: %w", componentsUnhealthyErrorMsg, err)
+		}
```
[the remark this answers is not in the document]
iurii-ssv: Intentionally keeping the `ctx.Err()` (parent-state) check rather than inspecting the error for `context.Canceled`/`context.DeadlineExceeded`: a wedged component (e.g. a hung EL) surfaces as a probe `DeadlineExceeded` *while the parent ctx is still live*, and that must trip the watchdog — keying off the error value would suppress exactly that case (a genuinely hung CL/EL would no longer restart the node).

On the race itself: if the ctx is canceled at this check it's either (a) another service already failed, in which case *its* error is the terminal cause the node exits with, so the process still exits non-zero and the orchestrator restarts it, or (b) a deliberate shutdown (SIGINT/Close), where exit 0 is correct. So a genuine, restart-worthy failure is never silently turned into a clean exit.

Added regression-guarding tests — `Test_startHealthProber_tripsOnUnhealthyComponent` (wedged component → `DeadlineExceeded` → must trip) and `Test_startHealthProber_returnsNilOnCtxCancel` (clean cancel → nil) — plus a clearer comment explaining the discriminator.

cli/operator/node.go
```diff
@@ -599,21 +615,63 @@ func (n *node) start() error {
...
+	}
+	if apiServeErr != nil {
+		g.Go(func() error {
 			if err := <-apiServeErr; err != nil {
-				n.logger.Fatal("API server serve loop exited", zap.Error(err))
+				return fmt.Errorf("API server serve loop exited: %w", err)
 			}
-		}()
+			return nil
+		})
```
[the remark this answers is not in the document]
iurii-ssv: The metrics/API serve loops can't wedge teardown: both servers are started with the lifecycle ctx, so their ctx-aware `Shutdown` watcher fires on cancel, and `http.Server.Serve` returns `ErrServerClosed` as soon as `Shutdown` closes the listener — independent of connection draining — which closes `serveErr`. So the plain receive always unblocks on shutdown (the in-server `Shutdown` already carries a 5s bound too).

(The WS server originally had an asymmetry here — its shutdown was bound to a construction-time ctx rather than the lifecycle one — since resolved by taking its ctx at `Start`; all three serve loops now behave identically.)

cli/operator/node.go
```diff
@@ -486,33 +497,39 @@ func (n *node) Close() error {
...
 		if err != nil {
-			n.logger.Fatal("failed to start metrics server", zap.Error(err))
+			return fmt.Errorf("failed to start metrics server: %w", err)
 		}
-		go func() {
-			if err := <-metricsServeErr; err != nil {
-				n.logger.Fatal("metrics server serve loop exited", zap.Error(err))
-			}
-		}()
+		metricsServeErr = serveErr
```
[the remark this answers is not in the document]
iurii-ssv: Good call — documented at the metrics `Start` site. (Note from a later round: the deferred-detection stance described here was deliberately reversed — see https://github.com/ssvlabs/ssv/pull/2876#discussion_r3383436278 — a service failure is now a terminal event that fail-fasts bring-up: the node exits on it regardless and the syncs are checkpointed, so deferring only postponed the restart.)

iurii-ssv:
@greptile pls re-review

operator/node.go
```diff
@@ -170,13 +171,17 @@ func New(logger *zap.Logger, opts Options, exporterOpts exporter.Options, slotTi
...
-	if err := n.startWSServer(); err != nil {
+	g, gctx := errgroup.WithContext(ctx)
+
+	wsServeErr, err := n.startWSServer()
+	if err != nil {
 		return fmt.Errorf("start WS server: %w", err)
 	}
 
-	// Start the duty scheduler in modes that use it.
+	// Start the duty scheduler in modes that use it. It joins gctx so a sibling failure (e.g. the WS
```
julienharbulot: I think this only becomes true after `runServices` is called. `dutyScheduler.Start(gctx)` can block in `HandleInitialDuties`, but the WS serve loop is not yet registered with the errgroup, so a WS serve failure during that startup window just sits buffered in `wsServeErr` and does not cancel `gctx` or stop scheduler startup.

What do you think about registering the WS serve-loop watcher before starting the scheduler, or otherwise make startup errors after `startWSServer` cancel this inner group? That would keep the invariant that a sibling service failure tears down the rest of operator startup instead of being observed only after the blocking scheduler startup completes.

iurii-ssv: You're right — thanks, the comment overstated it. The WS watcher isn't joined until `runServices`, so during `dutyScheduler.Start` (which blocks in `HandleInitialDuties`) a WS serve failure just buffers in `wsServeErr` and isn't acted on until `runServices`. Fixed the comment to describe the actual behavior.

On registering the watcher earlier / making post-`startWSServer` errors cancel the inner group — I dug into what that would actually change and decided against it:

- The only services started-early-but-joined-late are the **diagnostic HTTP servers** (WS here; metrics + SSV-API in `cli/operator`). The prober, ongoing event-sync, and the scheduler block aren't running during this window. So the only behavior that invariant adds is "a diagnostic-server crash aborts in-progress startup immediately."
- For a diagnostic server that's not a win — and in the cli layer it's a clear loss: it would let a rare metrics hiccup bounce a node mid-`SyncHistory` (a multi-hour, *resumable* sync), which is the opposite of what we want. The buffered path surfaces the crash correctly once startup completes (no masking), and the window is bounded (`HandleInitialDuties` self-limits to ~a slot), so the node always reacts.
- Doing it *properly* — so a sibling failure tears down startup without masking the real error (a naive register-early would surface `scheduler.Start`'s ctx error instead of the WS one) — means routing the whole blocking startup through the errgroup as a member. To stay consistent we'd do the same in `cli/operator`'s `start()`. Since it's net-negative for the diagnostic serve loops in both layers, I kept the buffered pattern (mirrors cli/operator) and made the comment accurate + explicit about the intent.

Happy to revisit if you think a service that's genuinely *critical during startup* (where immediate teardown would be the right call) is likely to land here.

iurii-ssv: I've pushed some more changes to simplify the code structure further (my original diff was overcomplicating things a bit for no good reason - inlining a bunch of things makes the code much more straightforward)

cli/operator/node.go
```diff
@@ -486,22 +487,36 @@ func (n *node) Close() error {
...
+//
+// Each service joins an errgroup on gctx as it is brought up; the first failure cancels gctx — which
+// tears down the others (and the ctx-aware HTTP servers) — and surfaces as the group error, while a
+// plain ctx cancellation has every service return nil (clean shutdown → nil). The synchronous
+// bring-up steps run on n.ctx, not gctx, so they abort only on actual shutdown: a long resumable step
+// (e.g. historical sync) is not torn down when a non-critical service (e.g. a diagnostic HTTP server)
+// crashes — that error still surfaces, at g.Wait. Every failure path returns up through runNode() to
+// the single top-level Fatal, with node.Close() running on the way out.
 func (n *node) start() error {
+	g, gctx := errgroup.WithContext(n.ctx)
```
momosh-ssv: Might be worth a note here that the whole early-return teardown relies on an invariant that isn't visible at this site: the group is derived from `n.ctx`, but nothing in `start()` cancels it. 

Every early `return err` below (e.g. `metadataSyncer.SyncAll`, `startNetwork`) depends on `runNode`'s deferred `n.Close()` -> `n.cancel()` to cancel `n.ctx` and reap the already-joined `g.Go` members. 

I'm worried that future changes would leak every joined serve goroutine on an early return. A one-line doc note (or a local `defer cancel()` guard) would make the dependency explicit.

cli/operator/start_node.go
```diff
@@ -56,7 +59,20 @@ var StartNodeCmd = &cobra.Command{
...
+		// (long-lived services return nil, node.Close() runs) instead of the process being terminated
+		// abruptly. Scoped to start-node so other subcommands keep cobra's default signal behavior.
+		ctx, stop := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+		defer stop()
+
+		if err := runNode(ctx, &cfg, logger); err != nil {
+			// A signal during the synchronous startup phase surfaces as an error rather than the
+			// errgroup's clean-cancel (nil) path; treat it as a deliberate stop, not a startup failure
+			// — exit 0 (letting the deferred observability shutdown run) instead of Fatal.
+			if ctx.Err() != nil {
```
momosh-ssv, line 86: Could this swallow a genuine startup failure? 
The guard keys only on whether a signal arrived, not on the nature of the error — so a real, restart-worthy failure (EL refused, DB corruption) that lands in the same window as a SIGTERM exits 0, and a supervisor won't restart it. 

In practice a signal during startup is almost always a deliberate stop (and usually the orchestrator sent it), so this is defensible — but it's strictly broader than "signal-caused failure." Maybe narrow the check to errors that wrap the signal ctx's `context.Canceled`, or add a one-line comment acknowledging that a coincident genuine error is intentionally treated as a clean stop.

cli/operator/node.go
iurii-ssv: Good catch — this went through a couple of iterations (a reinstated `defer cancel()`, then explicit bounded awaits) and has since settled in the terminal-cause supervisor: the lifecycle owner cancels the shared ctx with the first terminal error as its cause, then awaits the services and `Close` under a single grace window. The invariant you flagged no longer exists to be relied on — an early bring-up return simply becomes the cancellation cause, and stopping + awaiting the already-started services is sequenced by the supervisor itself, not by whoever returns first.

cli/operator/start_node.go
iurii-ssv: Agreed it's broad. I went with your comment option rather than narrowing to `errors.Is(err, context.Canceled)`, because the narrow check is fragile: not every bring-up step wraps the ctx error on cancel (some libraries convert a cancellation into a domain/transport error), so a genuine SIGTERM could then produce a non-`context.Canceled` error and spuriously `Fatal` a *deliberate* stop — the exact scary "could not start node" log this guard exists to suppress.

The broad `ctx.Err()` check reliably means "a signal arrived → operator asked to stop," and exiting 0 is correct regardless of a coincident error (whoever sent the signal isn't restarting on exit code). Added a note making that intentional breadth explicit.

operator/node.go
iurii-ssv: Update (latest PR state): circled back and did what you originally suggested — the WS serve-loop watcher now registers **early**, right where the server is started, instead of being deferred past the scheduler bring-up.

I'd held off to avoid masking a WS error with the scheduler bring-up's `context canceled`, but that masking requires a WS serve crash *during* startup, and `http.Server.Serve` effectively never returns a non-graceful error after a successful bind — so the deferral wasn't earning its keep against the member-far-from-its-creation placement (net runtime behavior is unchanged in practice).

But if you have a particular diff in mind of what your suggestion would look like (when applied) - I'll look into it some more @julienh-ssv

julienharbulot (dismissed):
Nice fix

cli/operator/node.go
```diff
@@ -526,12 +535,13 @@ func (n *node) start() error {
...
+	if startOngoingSync != nil {
 		healthProber.AddComponent(eventSyncerComponentName, eventSyncer, proberHealthcheckTimeout, proberRetriesMax, proberRetryDelay)
+		g.Go(func() error {
+			return startOngoingSync(gctx)
+		})
 	}
 
-	go startHealthProber(n.ctx, n.logger, healthProber)
-
 	if _, err := n.metadataSyncer.SyncAll(n.ctx); err != nil {
```
alok-ssv: The SyncAll and ensureNoMissingKeys still use n.ctx. Would it be better to use the group ctx so they can fail fast?

alok-ssv (dismissed):

cli/operator/node.go
momosh-ssv: Took a close look at this one. I'd vote to keep them on `n.ctx` but add a one-liner so it stops reading like an oversight. Since `gctx` derives from `n.ctx`, a sibling failure (metrics loop / ongoing-sync) cancels `gctx` but can't interrupt these synchronous steps — so switching to `gctx` would only fail-fast a few seconds sooner. Both are bounded anyway (SyncAll's `GetValidatorData` by the beacon `longTimeout`, `SyncHistory` is finite), then `g.Wait()` surfaces the sibling error and `Close()` runs right after. And the current default seems like the one you'd want: a diagnostic metrics-server blip shouldn't tear down a resumable historical/metadata sync mid-flight. So: keep `n.ctx`, add a note that it's deliberate and abort is deferred to `g.Wait()`.

cli/operator/prober.go
```diff
@@ -39,33 +39,39 @@ func ensureComponentsHealthy(ctx context.Context, logger *zap.Logger, p *hprobe.
...
-	const probeFrequency = 60 * time.Second
+// startHealthProber is a runtime liveness watchdog: it probes all components every probeFrequency
+// and, on persistent unhealth, returns an error so the node terminates gracefully (via Close,
+// non-zero exit) and the orchestrator restarts it; a clean ctx cancellation returns nil. Crashing on
+// persistent unhealth is intentional — a node idling with a dead EL while reporting healthy is a
+// silent failure, worse than a restart — so it must not degrade to log-and-retry.
+func startHealthProber(ctx context.Context, logger *zap.Logger, p *hprobe.HealthProber) error {
+	const (
+		probeFrequency = 60 * time.Second // how often a probe round runs
+		probeTimeout   = 60 * time.Second // max time for one probe round (ProbeAll)
```
momosh-ssv: Any reason `probeTimeout` and `probeFrequency` are both 60s? A probe round that consumes its full timeout (wedged EL/CL) leaves `ticker.C` already fired, so the next round starts with no gap — back-to-back probing of a struggling endpoint. It's benign today (only while already unhealthy, and the watchdog trips that same round), but coupling the two is a bit fragile: if a per-component timeout later drops below 60s you could get a tight retry storm right before the trip. Might be worth a smaller, decoupled `probeTimeout`, or a note that the equality is intentional.

cli/operator/prober_test.go
```diff
@@ -0,0 +1,56 @@
...
+	// Wedged component: blocks until its 20ms health-check ctx times out -> DeadlineExceeded. No
+	// retries, so the watchdog trips on the first tick.
+	p.AddComponent("el", wedgedComponent{}, 20*time.Millisecond, 0, 0)
+
+	err := startHealthProber(context.Background(), zap.NewNop(), p)
+	require.ErrorContains(t, err, componentsUnhealthyErrorMsg)
+}
+
+// Test_startHealthProber_returnsNilOnCtxCancel verifies a clean ctx cancellation (normal shutdown)
+// returns nil rather than tripping the watchdog.
```
momosh-ssv, line 72: Might be worth a third case: cancel the parent ctx while a component is mid-probe, so the probe returns a deadline/cancel error AND `ctx.Err() != nil` at the same time. That's the exact race the guard at prober.go:64 exists to resolve (don't mistake a shutdown for unhealth), and neither existing test fires both together. Would lock the guard's semantics against a future refactor.

cli/operator/node.go
iurii-ssv: Good point — though it was a deliberate trade-off (04361ada9): bring-up stayed on `n.ctx` so a member crash mid-bring-up wouldn't abort a long step with a bare `context canceled` — the early return never reaches `g.Wait()`, so the member's root-cause error would've been dropped.

Resolved both halves (the mechanism has since evolved into the terminal-cause supervisor, but the substance is the same): bring-up now runs on the same ctx as the services — incl. `ensureComponentsHealthy` and `syncContractEvents`, which had the same issue (historical sync is the longest blocking step of all) — so the first terminal failure makes every step fail fast, and the root cause is preserved as the ctx's cancellation cause, which is what ultimately gets logged. Teardown then awaits the services under a single grace window, so a failed bring-up no longer leaves them running behind it.

@momosh-ssv on your two axes (our replies crossed):
- *"only fail-fast a few seconds sooner"* — each step is bounded, but with `n.ctx` the node doesn't stop at the current step: it runs the **entire remaining bring-up** before anything surfaces — potentially hours of first-sync work if the crash lands during `SyncHistory`, then network setup, plus the API server and operator node getting `Start`'ed on an already-canceled ctx (odd, untested paths).
- *"a metrics blip shouldn't tear down a resumable sync"* — by this PR's design a service error is already terminal: the node exits regardless, so finishing the sync only postpones the restart; the syncs are checkpointed, so aborting loses at most the in-flight batch.

If you'd still rather have the bring-up isolated from service failures, happy to discuss — but the fail-fast now comes at no diagnostic cost.

cli/operator/prober.go
iurii-ssv: Addressed, in two parts:

- Back-to-back rounds are now impossible by construction: the loop waits `probeInterval` measured from round *end* (ticker dropped), so a long round always leaves a full quiet gap — and that stays true under any future re-tuning of the timeouts, instead of relying on `probeTimeout <= probeFrequency` staying in sync.
- `probeTimeout` keeps its 60s but the remaining coupling is now documented: it's the watchdog's own liveness guard (even a `Healthy` impl that ignores its ctx can't stall the loop), and it deliberately undercuts the components' nominal retry schedule (~110s with the standard prober params) — failing solidly for a whole round is exactly the unhealth the watchdog exists to trip on.

cli/operator/prober_test.go
iurii-ssv: Added. One subtlety surfaced while writing it: a polite component that returns `ctx.Err()` on cancel can't pin the guard — hprobe doesn't count a bare `context.Canceled` as a component failure, so the round doesn't fail and the watchdog exits through its `select`, never reaching the guard. The guard is load-bearing for a component that masks the cancellation as its own transport error — which is what real clients tend to produce (`connection reset` etc.) — so the test uses a `cancelMaskingComponent` doing exactly that: the round fails and `ctx.Err() != nil` land together, and the test pins the result to nil.

alok-ssv (dismissed):

cli/operator/node.go
iurii-ssv: The teardown invariant you flagged here is gone — replaced by an explicit supervisor.

There's no longer an errgroup whose members are joined inside `start()`, where an early `return err` leaks them unless something upstream remembers to cancel the ctx. The lifecycle is now [`runNode`](https://github.com/ssvlabs/ssv/blob/0e502790872c4e135617b7039585095ce98525c2/cli/operator/node.go#L172): it owns the cancel-cause root, spawns every long-lived service, blocks on the first terminal event, then awaits those services and closes the node itself — all in one place, bounded by a single grace window. A bring-up failure simply *becomes* the terminal cause (`cancel(err)`), and the same function sequences the unwind. Nothing is joined to a group that a future caller has to remember to reap, so the dependency you worried about no longer exists to be relied on.

It's actually stronger now than when I first replied: *every* long-lived goroutine runs through that supervised `spawn` — including the few the node constructor used to launch directly (the VRSubmitter, and in exporter modes the duty-trace collector / slot-pruning), which previously bound to a construction ctx and were never awaited. `newNode` now constructs only; the run phase starts everything.

cli/operator/node.go
momosh-ssv: Agreed — the supervisor settles this nicely. With the root cause preserved as the cancellation cause, the diagnostic cost that motivated keeping `n.ctx` is gone, and since a service error is terminal by design anyway (and the syncs checkpoint), finishing a sync would only postpone the restart. Happy with the fail-fast as it stands.

cli/operator/prober.go
momosh-ssv: Nice — measuring the gap from round end makes the quiet gap hold by construction rather than relying on the two timeouts staying in sync.
Resolving.

cli/operator/prober_test.go
momosh-ssv: That masking subtlety is a nice catch — a polite component that returns `ctx.Err()` never reaches the guard, so `cancelMaskingComponent` pins exactly the load-bearing case. Resolving.

momosh-ssv (dismissed):

operator/node.go
iurii-ssv: Follow-up: the remaining edge in this area — the WS errgroup member isn't awaited on `Start`'s early-return paths, and the `go …` background tasks below sit outside the group entirely — is tracked as a lifecycle cleanup in #2890.

julienharbulot (approved):
Reviewed the current diff with a focus on functional lifecycle behavior. The centralized supervisor is a good improvement over the previous goroutine-level Fatal paths: service failures now converge through one teardown path, shutdown is bounded, signal-driven stops are handled explicitly, and the hprobe changes cover the important wedged/cancelled probe cases. I also checked the updated WS server ctx handoff and validator-registration submitter lifecycle and don't see a blocking functional issue.

cli/operator/eventsync.go
```diff
@@ -67,26 +69,29 @@ func syncContractEvents(
...
 		if err != nil {
-			return nil, fmt.Errorf("failed to load local events: %w", err)
+			return nil, nil, fmt.Errorf("failed to load local events: %w", err)
 		}
 
 		if err := eventHandler.HandleLocalEvents(ctx, localEvents); err != nil {
-			return nil, fmt.Errorf("error occurred while running event data handler: %w", err)
+			return nil, nil, fmt.Errorf("error occurred while running event data handler: %w", err)
 		}
 	} else {
```
oleg-ssvlabs: nit: I think it would be easier to read if we add an early return within the `if` statement and remove `else`.
Something like:

```Go
if len(cfg.LocalEventsPath) != 0 {
   ...
   return eventSyncer, nil, nil
}
```

Then there wouldn't be a need to declare `startOngoingSync` variable so far from it actual use

cli/operator/node.go
```diff
@@ -198,28 +195,29 @@ type node struct {
...
-func newNode(ctx context.Context, cfg *config, logger *zap.Logger, res resolved, networkConfig *networkconfig.Network, consensusClient beaconClient, executionClient executionclient.Provider) (_ *node, err error) {
+func newNode(
+	ctx context.Context,
+	cfg *config,
+	logger *zap.Logger,
+	res resolved,
+	networkConfig *networkconfig.Network,
+	consensusClient beaconClient,
+	executionClient executionclient.Provider,
+) (_ *node, err error) {
```
oleg-ssvlabs, line 218: nit: I think it would be better not to use _named_ return values here

cli/operator/node.go
```diff
@@ -486,33 +479,47 @@ func (n *node) Close() error {
...
-// NOTE: the metrics/SSV-API server goroutines call logger.Fatal on failure — crashing the
-// process and bypassing Close() — a candidate for errgroup-based coordinated shutdown.
-func (n *node) start() error {
+// start brings the node up: it runs the synchronous bring-up steps and registers every long-lived
+// service via spawn, returning once the node is up (nil) or a bring-up step has failed (that
+// error). Everything binds to ctx — when it dies, bring-up aborts mid-step and the services stop;
+// the services' terminal errors travel through spawn, not through the return value.
+func (n *node) start(ctx context.Context, spawn func(func() error)) error {
+	// Background services wired (but not started) by newNode, mode-dependent.
+	if n.vrSubmitter != nil {
```
oleg-ssvlabs, line 488: I find it slightly difficult to follow which parts of the node run in each mode (`Exporter` vs `Operator`).

We check the mode during initialization and instantiate certain dependencies conditionally (e.g., `vrSubmitter` and `collector`), but we also perform mode checks later when launching slot pruning. The logic feels somewhat inconsistent - sometimes behavior is controlled through `mode` checks, and other times through `nil` checks, even though both ultimately depend on the node's mode.

This might be a good candidate for a future refactor. We already have quite a few mode-dependent code paths, so it could be cleaner to split the implementation into separate types, such as `OperatorNode` and `ExporterNode`. That would make the behavior of each mode much more explicit.

Taking it even further, I could imagine extracting the Exporter into a separate service/repository and exposing the shared functionality as a library from the Operator node. Of course, that would be a much larger refactor.

oleg-ssvlabs (dismissed):
gj. Just left couple of minor non blocking comments

cli/operator/node.go
iurii-ssv: The named `err` here is load-bearing rather than stylistic: the two error-path cleanup defers — `db.Close()` and `p2pNetwork.Close()` — read the returned `err` to decide whether to tear down what's already been opened when construction fails partway through. That's the standard Go "release-on-error" idiom, and it needs `err` named so the deferred closures can observe the final returned value.

The leading result is deliberately left blank (`_ *node`) to make it explicit that we never assign to it — only `err` carries state the defers care about. Dropping named returns entirely would mean reintroducing a manual closer-stack + success flag, which I think reads worse than this. So I'd prefer to keep it as-is here.

cli/operator/node.go
iurii-ssv: Fully agree with the direction. The mix of `mode` checks and `nil` checks for what's ultimately the same mode distinction isn't great, and making the modes explicit — separate `OperatorNode`/`ExporterNode` types, or even pulling the exporter out entirely — is the right long-term shape.

That's a meaningful refactor on its own, though, and this PR is already quite wide. Folding it in would extend the scope even further, so I'd rather not grow it here. Let's track it as a follow-up and do it properly in its own change.

cli/operator/eventsync.go
iurii-ssv: Should be addressed now.

oleg-ssvlabs (approved):

momosh-ssv (approved):
amazing piece of work, this one is!

PR #2877, opened 2026-06-09
message/validation: fix flaky pending-queued validator test + tidy test setup

iurii-ssv:
## Summary

Fixes a low-rate flaky failure in `Test_ValidateSSVMessage/"pending queued state validator"` and folds in related test-only cleanups in the same file. **No production code is touched.**

The flake surfaces on unrelated CI (e.g. the attempt-1 unit-test job of #2876), failing with a misleading error:

```
expected: "validator is not attesting, got pending_queued"            (ErrValidatorNotAttesting)
in chain: "incorrect topic, got topic 59 / base name 59, want [114]"  (ErrIncorrectTopic)
```

## Root cause — epoch-boundary race

- `generateShares` built the share with `ActivationEpoch = currentEpoch + 1`, sampled from the wall clock **at setup time**.
- The subtest expects rejection at `getCommitteeAndValidatorIndices` → `share.IsAttesting(EstimatedCurrentEpoch())`, which compares against the wall-clock epoch **at validation time**. A `pending_queued` validator is attesting iff `ActivationEpoch <= epoch`.
- `TestNetwork`'s epoch is 6m24s (12s slot × 32). If a run crosses an epoch boundary between setup and the assertion, `EstimatedCurrentEpoch()` reaches `ActivationEpoch`, the validator counts as attesting, the not-attesting branch is skipped, and validation falls through to the topic check.
- That topic check then fails because the subtest derived the topic from the share's **random** validator pubkey (`CommitteeID(execID[16:])`) rather than its committee — producing the confusing `ErrIncorrectTopic` symptom.

### Flaky-test fix (commit 1)
- Set `ActivationEpoch` to `currentEpoch + 100` (≈10.7h ahead, vs the 10-minute test timeout) so the validator stays `pending_queued`/not-attesting for the whole run regardless of boundary timing. Added a comment so it isn't reverted to `+1`.
- Normalized the four non-committee validator-state subtests (`liquidated`, `unknown state`, `pending queued state`, `no share metadata`) to derive the topic from `committeeID`. For a non-committee (validator-role) message the duty executor ID is the validator pubkey, so `CommitteeID(execID[16:])` is meaningless; the correct topic is the validator's committee. All four reject before the topic check, so behavior is unchanged.

### Follow-up cleanups (commit 2, no behavior change)
- **Helper:** extracted `saveShareWithRandomKey`, removing the repeated generate-key / copy-pubkey / save boilerplate from the five random-key shares in `generateShares`.
- **Topic audit:** found and normalized the two remaining non-committee subtests with the same wrong topic derivation — `partial zero signer ID` and `partial wrong RSA signature size` (both build a validator-pubkey `RoleAggregator` partial-sig message). Audited the other ~43 `CommitteeID(execID[16:])` derivations: all are committee-role messages (where `execID[16:]` genuinely is the committee ID) or the intentional `unknown committee ID` case — left as-is.
- **Rename:** `nonUpdatedMetadataNextEpoch` → `nonUpdatedMetadataFutureEpoch`, since its activation epoch is now further out than the next epoch.
- **Grammar:** "validator that in pending queued state" → "...that is in...".

## Test plan

- `go test -tags "blst_enabled lfs" -race -run 'Test_ValidateSSVMessage$' ./message/validation/ -count=5` — pass
- `go test -tags "blst_enabled lfs" -race ./message/validation/` — pass
- `go vet -tags "blst_enabled lfs" ./message/validation/` — clean

### DIFF
--- a/message/validation/validation_test.go
+++ b/message/validation/validation_test.go
@@ -123,7 +123,7 @@ func Test_ValidateSSVMessage(t *testing.T) {
 			shares.liquidated,
 			shares.inactive,
 			shares.nonUpdatedMetadata,
-			shares.nonUpdatedMetadataNextEpoch,
+			shares.nonUpdatedMetadataFutureEpoch,
 			shares.noMetadata,
 		} {
 			if bytes.Equal(share.ValidatorPubKey[:], pubKey) {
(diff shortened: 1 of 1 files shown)


julienharbulot (approved):
Reviewed for meaningful bug-level issues. The change is test-only, the validator-role topic setup now matches the production routing path, and the focused validation test passes locally.

momosh-ssv (approved):

alok-ssv (approved):

PR #2879, opened 2026-06-09
observability: surface silent duty failures + improve p2p logs

iurii-ssv:
## Motivation

A recent cluster-wide voluntary-exit failure (validators never exiting despite repeated on-chain requests) could not be diagnosed from node logs, because the voluntary-exit duty and the partial-signature path it depends on fail **silently**:

- A pre-consensus-only duty (voluntary exit, validator registration) that never reaches a partial-signature quorum just stops at the `!hasQuorum` early-return — no log at any level.
- Broadcast egress isn't confirmed in logs, and the gossipsub message-id (needed to correlate a message across nodes and against a libp2p pubsub trace) is never logged.
- Non-retryable duty-processing failures are logged at DEBUG, so a failed exit sign/broadcast/submission is invisible at INFO.
- Message acceptance at validation is metered but not logged.

This adds low-noise log points along that path so the next such failure is diagnosable from logs alone. Per review feedback, there is **no per-role or per-path special-casing**: every log point applies uniformly to all duties. The first two gaps get dedicated log points below; the last two deliberately do not — the deadline warning is the single escalated signal (drop details remain at DEBUG), and accepts stay unlogged since arrival is already provable from the runner's "got pre-consensus message" line.

## Changes

1. **`runner` — duty-completion deadline warning, every duty.** A pure completion watcher: when any duty hasn't completed (`markDutyFinished`, which every runner already funnels through) by the end of the current wall-clock slot, emit one `WARN`. It knows nothing about quorum/consensus phases and reads no runner state (released via a channel closed by `markDutyFinished`), so it's safe alongside the single-threaded message loop and correct when a runner is reused for a later duty. For beacon duties this complements QBFT round-timeout and submission logs; for voluntary exit / validator registration it is the *only* failure signal — previously these failed with no log at any level. The deadline is wall-clock slot end (not `duty.Slot` end) because a voluntary-exit envelope is stamped several slots in the past.

2. **`network` / `topics` — broadcast egress log, every duty.** Each broadcast logs at DEBUG the gossipsub message-id (the same id a receiver and the libp2p pubsub tracer assign → cross-node correlation) and the topic peer count (surfaces a sparse/empty mesh). `topics.MsgID` is exported so the broadcast path and gossipsub compute the id identically.

**Cleanup:** removes the dead `Error.silent` suppression switch in message validation — it was never set, so `Silent()` was always false and every dropped message was already logged. Removing it makes "every drop is logged" structurally true (the guarantee these changes rely on). Behavior-preserving.

## Notes

Observability-only (logging, one exported helper, and a behavior-identical replacement of libp2p's backoff connector); no happy-path behavior changes. Designed to *complement* libp2p `PUBSUB_TRACE` — the egress msg-id (#2) is what lets node logs and a pubsub trace be correlated.

## Follow-up: gossip-path visibility (added after the trace capture)

The pubsub-trace capture for this incident found the failure outside the cluster's own logs: the four operators had **zero direct connectivity to each other** (each held exactly 2 connections, both to all-subnet relay nodes), and those relays silently refused to forward voluntary-exit partial-sigs (no `dutyStore.VoluntaryExit` entry → `dutyLimit` 0 → `ErrTooManyDutiesPerEpoch` → `ValidationIgnore`). This PR also closes the observability gaps that made that invisible and hard to confirm:

**`network/p2p` — dial failures logged.** Outbound dial errors were swallowed by libp2p's `BackoffConnector` (visible only on libp2p's internal logger), so a node that could never reach its committee partners logged nothing. Replaced with an equivalent SSV-owned connector (same backoff-cache semantics, covered by a unit test) that logs `could not connect to discovered peer` at DEBUG with peer id, addrs and error.

**`network/p2p` — WARN on persistently unreachable proposed peers.** Discovery kept proposing the committee partners every selection round for hours with nothing to show for it. The selection loop now tracks first-proposal time per still-discovered peer and emits a rate-limited WARN (default: unconnected ≥10m, warn ≤1/5m) telling the operator to check firewall/NAT/advertised addresses.

**`network/topics` — payload message ids on trace RPC events.** The pubsub tracer logged only control ids (IHAVE/IWANT) on `SEND_RPC`/`RECV_RPC`/`DROP_RPC`, so "did the sender actually hand the message to the relay" had to be inferred. RPC events now include `msgIDs` (format locked by a unit test), making a message's forwarding path (or the hop where it vanished) directly observable across nodes.

**`eth/eventhandler` — silent exit-skip logged.** A `ValidatorExited` event for a share without beacon metadata silently produced no duty; now logged at DEBUG.

## Follow-up: duty success/finality split + outcome classifier (evolves change #1)

Change #1's deadline watcher reused `markDutyFinished` / `State.Finished` as its completion signal. But `State.Finished` is ssv-spec's **success** flag — "a full successful cycle (pre, consensus and post) with quorum" — and it is serialized into the spec-test post-duty-runner-state root. Reusing it overloaded *success* to also mean *watcher-finality*, and since the watcher was released only on the success / no-op paths, a duty that **failed terminally** (e.g. a beacon-node call 404-ing on a missed slot) or **never reached pre-consensus quorum** tripped the `did not complete before slot end` WARN as if it had silently stalled. On a testnet the WARN was dominated by `AGGREGATOR` / `SYNC_COMMITTEE_CONTRIBUTION` / `VALIDATOR_REGISTRATION`, not by the core committee duties.

This splits the two concepts and turns duty completion into a definitive, single-answer classifier.

**Success vs. finality.** `State.Finished` → `State.Succeeded` (kept serialized as `"Finished"`, so the spec-test state root is byte-identical — verified by the existing spec tests). `markDutyFinished` → `markDutySucceeded`; `hasDutyFinished` → `hasDutySucceeded`.

**One outcome per duty.** A per-duty watcher (`watchDutyOutcome`) is now the single reporter of a duty's terminal outcome — it records the `ssv.runner.duty.outcome{role, outcome}` counter (Prometheus `ssv_runner_duty_outcome_total`) and warns only for the outcomes worth attention:

| outcome | meaning | log |
|---|---|---|
| `succeeded` | full cycle, submitted to the beacon node | existing per-runner success log |
| `not_required` | completed with nothing to submit (e.g. not selected as aggregator) | — |
| `failed` | terminated by a non-recoverable error (carries the reason) | `⚠️ duty failed` |
| `stuck` | not concluded before the end of the current wall-clock slot | `⚠️ duty did not complete before slot end` |

Markers (`markDutySucceeded` / `markDutyNotRequired` / `markDutyFailed`) hand the outcome to the watcher over a buffered(1) channel; the watcher is the sole reporter, so each duty is classified exactly once (no double-count; a guard prefers a conclusion that lands right at the deadline). Terminal-failure paths in the one-shot runners (`aggregator`, `sync_committee_contribution`, `proposer`, `voluntary_exit`, `validator_registration`) and `executeDuty` failures in the duty-start path now report `failed` instead of tripping the deadline WARN; the committee runner keeps its partial-progress semantics (a committee duty resolves to `succeeded` or `stuck`). `not_required` still sets `State.Succeeded`, so the state root is unchanged.

Net effect: `stuck` is now reserved for genuinely silent stalls (the voluntary-exit failure that motivated this PR), terminal failures are reported as `failed` with their reason, and "what happened to this duty?" is answerable from one metric.

### DIFF
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -8,7 +8,6 @@ import (
 
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
-	spectypes "github.com/ssvlabs/ssv-spec/types"
 	"go.uber.org/zap"
 
 	"github.com/ssvlabs/ssv/observability/log/fields"
@@ -175,16 +174,7 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
 }
 
 func (mv *messageValidator) handleValidationSuccess(ctx context.Context, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
-	role := messageRole(decodedMessage)
-	recordAcceptedMessage(ctx, role)
-
-	// Accepts aren't logged in general (committee/attestation traffic would be a firehose), but for
-	// the rare, one-shot quorum duties an accept line confirms a peer's partial signature reached
-	// and passed validation — the question past voluntary-exit failures couldn't answer from logs.
-	if role == spectypes.RoleVoluntaryExit || role == spectypes.RoleValidatorRegistration {
-		loggerFields := mv.buildLoggerFields(decodedMessage)
-		mv.logger.With(loggerFields.AsZapFields()...).Debug("accepted message")
-	}
+	recordAcceptedMessage(ctx, messageRole(decodedMessage))
 
 	return pubsub.ValidationAccept
 }
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -8,6 +8,7 @@ import (
 
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
+	spectypes "github.com/ssvlabs/ssv-spec/types"
 	"go.uber.org/zap"
 
 	"github.com/ssvlabs/ssv/observability/log/fields"
@@ -20,7 +21,6 @@ type Error struct {
 	want     any
 	innerErr error
 	reject   bool
-	silent   bool
 }
 
 func (e Error) Error() string {
@@ -47,10 +47,6 @@ const (
 
 func (e Error) Reject() bool {
 	return e.reject
-}
-
-func (e Error) Silent() bool {
-	return e.silent
 }
 
 func (e Error) Text() string {
@@ -168,22 +164,29 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
 	}
 
 	if !valErr.Reject() {
-		if !valErr.Silent() {
-			logger.Debug("ignoring invalid message", zap.Error(valErr))
-		}
+		logger.Debug("ignoring invalid message", zap.Error(valErr))
 		recordIgnoredMessage(ctx, loggerFields.Role, valErr.Text())
 		return pubsub.ValidationIgnore
 	}
 
-	if !valErr.Silent() {
-		logger.Debug("rejecting invalid message", zap.Error(valErr))
-	}
-
+	logger.Debug("rejecting invalid message", zap.Error(valErr))
 	recordRejectedMessage(ctx, loggerFields.Role, valErr.Text())
 	return pubsub.ValidationReject
 }
 
 func (mv *messageValidator) handleValidationSuccess(ctx context.Context, decodedMessage *queue.SSVMessage) pubsub.ValidationResult {
-	recordAcceptedMessage(ctx, messageRole(decodedMessage))
+	role := messageRole(decodedMessage)
+	recordAcceptedMessage(ctx, role)
+
+	// Accepts are metered but not logged in general — committee/attestation traffic would make it
+	// a firehose. But for the rare, one-shot quorum duties (voluntary exit, validator
+	// registration), an explicit accept line at the validation boundary lets you confirm a peer's
+	// partial signature actually reached and passed validation (with the duty id, slot and
+	// signer) — the exact question that couldn't be answered from logs in past exit failures.
+	if role == spectypes.RoleVoluntaryExit || role == spectypes.RoleValidatorRegistration {
+		loggerFields := mv.buildLoggerFields(decodedMessage)
+		mv.logger.With(loggerFields.AsZapFields()...).Debug("accepted message")
+	}
+
 	return pubsub.ValidationAccept
 }
--- a/message/validation/errors.go
+++ b/message/validation/errors.go
@@ -178,11 +178,9 @@ func (mv *messageValidator) handleValidationSuccess(ctx context.Context, decoded
 	role := messageRole(decodedMessage)
 	recordAcceptedMessage(ctx, role)
 
-	// Accepts are metered but not logged in general — committee/attestation traffic would make it
-	// a firehose. But for the rare, one-shot quorum duties (voluntary exit, validator
-	// registration), an explicit accept line at the validation boundary lets you confirm a peer's
-	// partial signature actually reached and passed validation (with the duty id, slot and
-	// signer) — the exact question that couldn't be answered from logs in past exit failures.
+	// Accepts aren't logged in general (committee/attestation traffic would be a firehose), but for
+	// the rare, one-shot quorum duties an accept line confirms a peer's partial signature reached
+	// and passed validation — the question past voluntary-exit failures couldn't answer from logs.
 	if role == spectypes.RoleVoluntaryExit || role == spectypes.RoleValidatorRegistration {
 		loggerFields := mv.buildLoggerFields(decodedMessage)
 		mv.logger.With(loggerFields.AsZapFields()...).Debug("accepted message")
--- a/network/p2p/p2p_connector.go
+++ b/network/p2p/p2p_connector.go
@@ -14,28 +14,30 @@ import (
 	"github.com/ssvlabs/ssv/observability/log/fields"
 )
 
+type connCacheData struct {
+	nextTry time.Time
+	start   libp2pdiscbackoff.BackoffStrategy
+}
+
 // backoffConnector connects to proposed peers, skipping peers that are still within their backoff
 // period. It is a drop-in replacement for libp2p's discovery/backoff.BackoffConnector with one
 // difference: dial failures are logged to the node's logger. The libp2p implementation reports
 // them only to its own internal logger, so a node that cannot reach its proposed peers (firewall,
 // NAT, wrong advertised address) leaves no trace of that in SSV logs.
 type backoffConnector struct {
+	logger *zap.Logger
+
 	host       host.Host
-	logger     *zap.Logger
-	cache      *lru.TwoQueueCache[peer.ID, *connCacheData]
 	connTryDur time.Duration
 	backoff    libp2pdiscbackoff.BackoffFactory
-	mux        sync.Mutex
-}
 
-type connCacheData struct {
-	nextTry time.Time
-	strat   libp2pdiscbackoff.BackoffStrategy
+	cacheMux sync.Mutex
+	cache    *lru.TwoQueueCache[peer.ID, *connCacheData]
 }
 
 func newBackoffConnector(
-	host host.Host,
 	logger *zap.Logger,
+	host host.Host,
 	cacheSize int,
 	connTryDur time.Duration,
 	backoff libp2pdiscbackoff.BackoffFactory,
@@ -46,11 +48,11 @@ func newBackoffConnector(
 	}
 
 	return &backoffConnector{
-		host:       host,
 		logger:     logger,
-		cache:      cache,
+		host:       host,
 		connTryDur: connTryDur,
 		backoff:    backoff,
+		cache:      cache,
 	}, nil
 }
 
@@ -68,20 +70,20 @@ func (c *backoffConnector) Connect(ctx context.Context, peerCh <-chan peer.AddrI
 				continue
 			}
 
-			c.mux.Lock()
+			c.cacheMux.Lock()
 			if cached, ok := c.cache.Get(pi.ID); ok {
 				now := time.Now()
 				if now.Before(cached.nextTry) {
-					c.mux.Unlock()
+					c.cacheMux.Unlock()
 					continue
 				}
-				cached.nextTry = now.Add(cached.strat.Delay())
+				cached.nextTry = now.Add(cached.start.Delay())
 			} else {
-				cached = &connCacheData{strat: c.backoff()}
-				cached.nextTry = time.Now().Add(cached.strat.Delay())
+				cached = &connCacheData{start: c.backoff()}
+				cached.nextTry = time.Now().Add(cached.start.Delay())
 				c.cache.Add(pi.ID, cached)
 			}
-			c.mux.Unlock()
+			c.cacheMux.Unlock()
 
 			go func(pi peer.AddrInfo) {
 				ctx, cancel := context.WithTimeout(ctx, c.connTryDur)
--- a/network/p2p/p2p_connector.go
+++ b/network/p2p/p2p_connector.go
@@ -1,0 +1,102 @@
+package p2pv1
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	libp2pdiscbackoff "github.com/libp2p/go-libp2p/p2p/discovery/backoff"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/observability/log/fields"
+)
+
+// backoffConnector connects to proposed peers, skipping peers that are still within their backoff
+// period. It is a drop-in replacement for libp2p's discovery/backoff.BackoffConnector with one
+// difference: dial failures are logged to the node's logger. The libp2p implementation reports
+// them only to its own internal logger, so a node that cannot reach its proposed peers (firewall,
+// NAT, wrong advertised address) leaves no trace of that in SSV logs.
+type backoffConnector struct {
+	host       host.Host
+	logger     *zap.Logger
+	cache      *lru.TwoQueueCache[peer.ID, *connCacheData]
+	connTryDur time.Duration
+	backoff    libp2pdiscbackoff.BackoffFactory
+	mux        sync.Mutex
+}
+
+type connCacheData struct {
+	nextTry time.Time
+	strat   libp2pdiscbackoff.BackoffStrategy
+}
+
+func newBackoffConnector(
+	host host.Host,
+	logger *zap.Logger,
+	cacheSize int,
+	connTryDur time.Duration,
+	backoff libp2pdiscbackoff.BackoffFactory,
+) (*backoffConnector, error) {
+	cache, err := lru.New2Q[peer.ID, *connCacheData](cacheSize)
+	if err != nil {
+		return nil, err
+	}
+
+	return &backoffConnector{
+		host:       host,
+		logger:     logger,
+		cache:      cache,
+		connTryDur: connTryDur,
+		backoff:    backoff,
+	}, nil
+}
+
+// Connect attempts to connect to the peers passed in by peerCh. Will not connect to peers if they
+// are within the backoff period.
+func (c *backoffConnector) Connect(ctx context.Context, peerCh <-chan peer.AddrInfo) {
+	for {
+		select {
+		case pi, ok := <-peerCh:
+			if !ok {
+				return
+			}
+
+			if pi.ID == c.host.ID() || pi.ID == "" {
+				continue
+			}
+
+			c.mux.Lock()
+			if cached, ok := c.cache.Get(pi.ID); ok {
+				now := time.Now()
+				if now.Before(cached.nextTry) {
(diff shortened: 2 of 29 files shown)


protocol/v2/ssv/runner/runner.go
```diff
@@ -248,7 +255,50 @@ func (b *BaseRunner) baseStartNewNonBeaconDuty(ctx context.Context, logger *zap.
...
+	if err := runner.executeDuty(ctx, logger, duty); err != nil {
+		return err
+	}
+
+	// Pre-consensus-only duties (voluntary exit, validator registration) have no consensus or
+	// post-consensus phase and broadcast their partial signature exactly once, so a failure to
+	// collect a partial-signature quorum is otherwise completely silent (the runner just stops
+	// at the !hasQuorum early-return). Watch the deadline and emit one operator-visible warning
+	// if the duty never completes.
+	b.watchNonBeaconDutyDeadline(ctx, logger, duty, quorum)
```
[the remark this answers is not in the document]
iurii-ssv: Good catch — applied in c372639dd (note: rebased onto `stage`, so the SHA changed from my earlier reply). This can't happen today (both voluntary exit and validator registration complete in `ProcessPreConsensus` after quorum, never synchronously inside `executeDuty`), but the guard makes it robust against single-operator setups / future refactors: the watcher now only starts when `b.hasDutyRunning()` is true, so a duty already finished by the time we'd start it is skipped rather than left with an unreleasable channel.

network/p2p/p2p_pubsub.go
```diff
@@ -51,23 +52,35 @@ func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedS
...
 		}
+
+		// Egress confirmation. gossip_msg_id is the id gossipsub itself assigns (and the libp2p
+		// pubsub tracer logs), so a single message can be followed across nodes and against a
+		// trace. topic_peers surfaces a near-empty/sparse topic mesh, a prime suspect when a
+		// one-shot message (e.g. a voluntary-exit partial-sig) fails to reach peers.
+		topicPeers, _ := n.topicsCtrl.Peers(topic)
+		n.logger.Debug("📤 broadcast message to topic",
+			fields.MessageID(msg.SSVMessage.MsgID),
+			zap.String("gossip_msg_id", hex.EncodeToString([]byte(topics.MsgID(encodedMsg)))),
```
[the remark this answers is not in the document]
iurii-ssv: The hex form is intentional, and it matches the tracer operators actually use here: SSV's pubsub event tracer (`network/topics/tracer.go`) logs `hex.EncodeToString(msg.GetMessageID())` — the same hex form — and that's the `PUBSUB_TRACE` path. So for SSV↔SSV correlation and against the SSV pubsub-event log, no decode is needed.

You're right that a raw `go-libp2p-pubsub` `JSONTracer`/`PBTracer` emits the id in binary and would need a `hex.DecodeString` first — that's a secondary, opt-in path. I reworded the code comment in c372639dd to reference SSV's pubsub event tracer (hex) instead of "the libp2p pubsub tracer" to remove the ambiguity.

protocol/v2/ssv/runner/runner.go
```diff
@@ -248,7 +255,52 @@ func (b *BaseRunner) baseStartNewNonBeaconDuty(ctx context.Context, logger *zap.
...
+	return nil
+}
+
+// watchNonBeaconDutyDeadline warns if a pre-consensus-only duty hasn't finished by its deadline.
+// Completion is signaled by markDutyFinished closing nonBeaconDeadlineDone, not by reading runner
+// state, so it's safe alongside the single-threaded message loop. Each duty gets its own channel so
+// a reused runner still reports an earlier duty that never completed.
+func (b *BaseRunner) watchNonBeaconDutyDeadline(ctx context.Context, logger *zap.Logger, duty *spectypes.ValidatorDuty, quorum uint64) {
+	done := make(chan struct{})
+	b.nonBeaconDeadlineDone = done
```
momosh-ssv: Might be worth a word on the orphaned-goroutine case here. 

If a reused runner starts a new duty, this overwrites `nonBeaconDeadlineDone` with a fresh channel — but if the previous duty reached quorum yet failed reconstruction/submission (so `markDutyFinished` never ran), that earlier watcher is never released via `done`. 

It does self-terminate on its own already-expired `timer.C`, so it's harmless, but the current comment ("each duty gets its own channel so a reused runner still reports an earlier duty") implies intent without covering the orphan. 

A one-liner noting the old watcher is reaped by its own timer would save a future reader the trace.

protocol/v2/ssv/runner/runner.go
```diff
@@ -248,7 +255,52 @@ func (b *BaseRunner) baseStartNewNonBeaconDuty(ctx context.Context, logger *zap.
...
+
+		select {
+		case <-done:
+			return // duty finished successfully
+		case <-ctx.Done():
+			return // node/validator shutting down
+		case <-timer.C:
+		}
+
+		logger.Warn("⚠️ pre-consensus duty did not reach quorum or submit by deadline",
```
momosh-ssv: Do we want to distinguish quorum-failure from submission-failure here? 

`markDutyFinished` only runs on full success, so this WARN also fires when quorum *was* reached but reconstruction/submission failed. 

Those point an operator at different things (peer mesh vs beacon node), but the single message lumps them together. Not blocking — just wondering if a separate signal for "quorum reached, submit failed" is worth it down the line.

message/validation/errors.go
```diff
@@ -168,22 +164,27 @@ func (mv *messageValidator) handleValidationError(ctx context.Context, peerID pe
...
-	recordAcceptedMessage(ctx, messageRole(decodedMessage))
+	role := messageRole(decodedMessage)
+	recordAcceptedMessage(ctx, role)
+
+	// Accepts aren't logged in general (committee/attestation traffic would be a firehose), but for
+	// the rare, one-shot quorum duties an accept line confirms a peer's partial signature reached
+	// and passed validation — the question past voluntary-exit failures couldn't answer from logs.
+	if role == spectypes.RoleVoluntaryExit || role == spectypes.RoleValidatorRegistration {
+		loggerFields := mv.buildLoggerFields(decodedMessage)
+		mv.logger.With(loggerFields.AsZapFields()...).Debug("accepted message")
```
momosh-ssv: Tiny consistency nit: `handleValidationError` just above hoists `logger := mv.logger.With(...)` into a local, whereas this inlines `mv.logger.With(...).Debug(...)`. Same base logger so behavior is identical — purely cosmetic if you want them to read the same.

protocol/v2/ssv/runner/runner.go
iurii-ssv: Added the one-liner — the `watchNonBeaconDutyDeadline` doc now covers the orphan: starting a new duty overwrites the channel, and the earlier watcher (whose duty failed mid-flight, so `markDutyFinished` never ran) warns for its own duty and is reaped by its own timer.

protocol/v2/ssv/runner/runner.go
iurii-ssv: Documented the triage on the WARN instead of splitting the signal: when quorum *was* reached but reconstruction/submission failed, that error already logs with specifics the moment it happens (the queue's could-not-handle drop log), so a deadline WARN with no preceding error line for the duty ⇒ quorum never formed. Distinguishing inside the WARN itself would mean sharing quorum state with the watcher goroutine, which deliberately reads no runner state to stay race-free — not worth it while the adjacent log answers it.

Heads-up: in the same push we removed per-role special-casing across the PR (the queue drop-log WARN raise is reverted to uniform DEBUG), so this deadline WARN is now the single escalated signal for these duties.

message/validation/errors.go
iurii-ssv: Resolved by deletion — the per-role accept-log block is gone entirely as part of dropping role special-casing from the PR. Arrival of a peer's partial-sig is already provable one layer up from the runner's "📬 got pre-consensus message" DEBUG line, which is what the voluntary-exit investigation actually used.

protocol/v2/ssv/runner/runner.go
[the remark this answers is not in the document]
iurii-ssv: Superseded in the latest push: the watcher (now `watchDutyDeadline`, generalized to all duties) is started *before* `executeDuty`, so a duty that completes synchronously closes its own channel — the `hasDutyRunning` guard is gone along with the edge case it guarded.

protocol/v2/ssv/runner/runner.go
iurii-ssv: Update: in the latest push the watcher became a pure completion watcher applied to every duty (no quorum/consensus knowledge), so the WARN text is now the generic "duty did not complete before slot end" and the triage comment moved out of the code. The triage itself still holds: a submission/reconstruction failure logs its specific error when it happens; a deadline WARN with no preceding error for the duty means quorum never formed.

iurii-ssv:
@greptile pls re-review

protocol/v2/ssv/runner/runner.go
iurii-ssv: Small follow-up for accuracy: after the generalization the function is `watchDutyDeadline` — the orphan note carried over to its doc unchanged (previous duty's watcher warns for its own duty and is reaped by its own timer).

protocol/v2/ssv/runner/runner.go
```diff
@@ -247,8 +256,46 @@ func (b *BaseRunner) baseStartNewNonBeaconDuty(ctx context.Context, logger *zap.
...
+	go func() {
+		select {
+		case <-finishedOnTime:
+			return // duty completed
+		case <-ctx.Done():
+			return // node/validator shutting down
+		case <-time.After(time.Until(onTimeDeadline)):
+		}
+
+		logger.Warn("⚠️ duty did not complete before slot end")
```
momosh-ssv: What happens if a committee duty ends in partial success? 

Seems that `CommitteeRunner` calls `markDutyFinished` only when `HasSubmittedAllValidatorDuties` is true, so the partial-success path leaves the watcher pending — we'd log `✔️finished duty processing (partial success)` at Info and then this WARN for the same duty. 

For a cluster with one persistently failing validator, I fear that's a contradictory Info+WARN pair every slot. Maybe that's intended (the duty genuinely didn't fully complete) — just want to make sure it's a deliberate choice. Not blocking.

momosh-ssv (dismissed):

protocol/v2/ssv/runner/runner.go
iurii-ssv: Yeah, the way "Finished" currently works is a mess (it conflates all of success/failure/stuck, and even produces false positives).

That's why I'll push some more changes to this PR to make it make sense (otherwise it's a useless metric). Not much value merging this PR without it.

iurii-ssv: Good question — worth pinning down, because the restructuring changed the model but not this case. Short version: **a committee duty that ends in partial success is still reported as `stuck`**, exactly as you described.

The watcher no longer keys off a single `markDutyFinished`; each duty now reports exactly one outcome via `watchDutyOutcome`:
- `succeeded` / `not_required` — full completion
- `failed` — a terminal error, with the reason (`⚠️ duty failed`)
- `stuck` — never reached a terminal state by wall-clock slot end (the old WARN, now `⚠️ duty did not complete before slot end (likely stuck)` + `ssv_runner_duty_outcome_total{outcome="stuck"}`)

For `CommitteeRunner`:
- `markDutySucceeded()` is still gated on `HasSubmittedAllValidatorDuties` (matches ssv-spec, which sets `State.Finished` only there).
- The restructuring added a defer that reports `failed` on terminal post-quorum **errors** (expected-roots failure, submission errors, no-valid-duties).
- Partial success returns `nil` (no error) → it hits neither path. The runner stays open, keeps submitting as later post-consensus quorums arrive; if a validator's quorum never forms by slot end → `stuck`.

So yes — for a cluster with one persistently-failing validator you still get, per slot and for the same duty: `✔️finished duty processing (partial success)` (Info) + `⚠️ duty did not complete before slot end (likely stuck)` (WARN).

It's deliberate in that the duty genuinely didn't fully complete — the under-submitted validator(s) really are stuck and we don't want to hide that. The restructuring's goal was narrower: stop reporting genuine **errors** (e.g. a beacon-node 404) as `stuck`. Partial-success isn't an error, so it's correctly *not* reclassified to `failed`.

The contradictory-looking Info+WARN every slot is still a fair concern, though ... but TBH I wound't worry about it at this point (we can get to it and improve at some future point, it's pretty rare case anyways).

momosh-ssv: Thanks for the detailed breakdown — matches my reading of the new code. Agreed partial success genuinely is not a terminal state, and with `outcome=stuck` carrying the metric the under-submitted validators stay visible. Fine to leave the Info+WARN pairing for a future pass.

protocol/v2/ssv/runner/aggregator.go
```diff
@@ -120,6 +120,14 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 		return nil
 	}
 
+	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
+	// so a terminal failure below won't be retried.
+	defer func() {
+		if err != nil {
+			r.markDutyFailed(err)
```
momosh-ssv, line 127: Any reason we're not guarding these one-shot defers with `!errors.Is(err, context.Canceled)` like the committee runner's? 

Seems that during shutdown a submission aborted mid-flight could be reported as `⚠️ duty failed: context canceled` (and counted under `outcome=failed`), depending on which select branch the watcher takes. 

Not blocking — shutdown-only noise.

momosh-ssv (dismissed):

protocol/v2/ssv/runner/aggregator.go
iurii-ssv: Good catch — it could indeed fire: the watcher's `select` picks a ready branch at random, so on shutdown it's ~50/50 whether a `failed{context.Canceled}` conclusion or `<-ctx.Done()` wins.

I fixed it at the classifier rather than per-defer — [`markDutyFailed`](https://github.com/ssvlabs/ssv/blob/observability/quorum-duty-debug-logs/protocol/v2/ssv/runner/runner.go#L653) now drops a `context.Canceled` reason:

```go
func (b *BaseRunner) markDutyFailed(reason error) {
	if errors.Is(reason, context.Canceled) {
		return
	}
	b.concludeDuty(dutyOutcomeFailed, reason)
}
```

Why there:

- A cancellation isn't a duty *failure* — it means the duty was **abandoned** (we stopped trying), not attempted-and-failed. Keeping `"failed"` for genuine, attributable failures is `markDutyFailed`'s own contract, not knowledge about callers/shutdown.
- Filtering at the source means the bogus conclusion is **never produced**, so there's nothing to race in the watcher — no coin-flip to neutralize.
- It's uniform: the [committee runner's](https://github.com/ssvlabs/ssv/blob/observability/quorum-duty-debug-logs/protocol/v2/ssv/runner/committee.go#L517) inline `!errors.Is(err, context.Canceled)` guard is now redundant and removed, and the [`baseStartNewDuty`](https://github.com/ssvlabs/ssv/blob/observability/quorum-duty-debug-logs/protocol/v2/ssv/runner/runner.go#L249) / [`baseStartNewNonBeaconDuty`](https://github.com/ssvlabs/ssv/blob/observability/quorum-duty-debug-logs/protocol/v2/ssv/runner/runner.go#L267) start paths — which also call `markDutyFailed` on a shutdown `context.Canceled` from `executeDuty` — are covered too.

A *genuine* failure that happens to race shutdown is still reported (it's a real failure); only the `context.Canceled` non-failures are dropped. Regression tests `markDutyFailed drops a context.Canceled reason` (deterministic) and `does not report a duty aborted by cancellation` (watcher-level) both fail if the filter is removed.

network/p2p/p2p_connector.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type connCacheData struct {
+	nextTry time.Time
+	start   libp2pdiscbackoff.BackoffStrategy
+}
+
+// backoffConnector connects to proposed peers, skipping peers that are still within their backoff
+// period. It is a drop-in replacement for libp2p's discovery/backoff.BackoffConnector with one
+// difference: dial failures are logged to the node's logger. The libp2p implementation reports
```
alok-ssv, line 24: Is this needed after #2882 ?

alok-ssv (approved):
The timeout-driven stuck classification is reasonable, but worth monitoring for noise near slot boundaries on heavily loaded nodes.

network/p2p/p2p_connector.go
iurii-ssv: Kind of, but not really. The #2882 is "full log output" for deep debugging sessions, while this is very useful to have in general (so that most of p2p connectivity issues can be inspected/answered just by looking at the DEBUG logs emitted by SSV node when it runs in `LOG_LEVEL=debug` mode).

So #2882 logs probably supersede these logs, but enabling #2882 logs is a measure of last resort & mostly feasible for testing environments (or short debugging sessions) only since the logs volume it produces is huge. While running SSV node with `LOG_LEVEL=debug` is pretty much recommended (and we do try and keep debug-logs output of SSV node within reasonable limits).

protocol/v2/ssv/runner/aggregator.go
momosh-ssv: Nice — centralizing this in `markDutyFailed` reads cleaner than per-defer guards, and covers the start paths too. Tests look good, resolving.

momosh-ssv (approved):

PR #2882, opened 2026-06-11
network/p2p: route go-libp2p logs through SSV logger via Libp2pTrace flag

iurii-ssv:
## What

Adds a `Libp2pTrace` config flag (`LIBP2P_TRACE`, default off) that hooks go-libp2p's logging (`github.com/ipfs/go-log/v2`) into the SSV `zap` logger and applies an `error,swarm2=debug,basichost=debug` level policy.

This is the SSV-config equivalent of running with `GOLOG_LOG_LEVEL=error,swarm2=debug,basichost=debug`, with two advantages:

- libp2p logs are rendered through SSV's own encoder/output (same format, same file sink) instead of go-log's default colorized stderr.
- verbosity is driven by SSV config (yaml/env), so the `GOLOG_LOG_LEVEL` environment variable is no longer needed for this.

When the flag is off, libp2p logging is left completely untouched (existing default / `GOLOG_LOG_LEVEL` behavior), mirroring how `PubSubTrace` works.

## How

`go-log/v2` is itself zap-based: every subsystem logger writes into a single shared core and is gated by a per-subsystem `AtomicLevel` via `zap.IncreaseLevel` (effective level = `max(coreLevel, subsystemLevel)`).

- `observability/log`: extracted `assembleCore` so the SSV logger and the libp2p core share the same encoder and writers (notably a single file syncer — no double log-rotation). Added `HookLibp2pLogging`, which:
  - calls `golog.SetupLogging` with `Level: error` and `SubsystemLevels: {swarm2: debug, basichost: debug}` to seed the per-subsystem levels (this registers them even for subsystem loggers not yet created, so it's init-order independent);
  - builds a debug-floor core sharing the SSV encoder/writers and installs it via `golog.SetPrimaryCore`. The per-subsystem levels do the actual gating.
- `network/p2p`: new `Libp2pTrace` field on the p2p `Config`; `Setup()` calls the hook when it's enabled.

The flip side: `go-log/v2` moves from indirect to direct in `go.mod`.

## Testing

- `TestHookLibp2pLogging` asserts `swarm2`/`basichost` log at debug while other subsystems stay error-only, exercised through the real go-log cores.
- `go build ./...`, `go vet`, and the `observability/log` package tests pass.

## Note on altitude

The hook lives in `observability/log` (keeps the core-building private and reused) and is invoked from p2p `Setup()` alongside the existing `PubSubTrace`/`TraceLog` wiring.

### DIFF
--- a/observability/log/global.go
+++ b/observability/log/global.go
@@ -6,8 +6,10 @@ import (
 	"log"
 	"os"
 	"runtime/debug"
+	"sync"
 	"time"
 
+	golog "github.com/ipfs/go-log/v2"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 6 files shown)


observability/log/global.go
```diff
@@ -48,54 +49,115 @@ func SetGlobal(levelName string, levelEncoderName string, logFormat string, file
...
+	// regardless of package init ordering.
+	golog.SetupLogging(golog.Config{
+		Level: golog.LevelError,
+		SubsystemLevels: map[string]golog.LogLevel{
+			"swarm2":    golog.LevelDebug,
+			"basichost": golog.LevelDebug,
+		},
+		// No Stdout/Stderr/File: the core SetupLogging builds writes nowhere and
+		// is immediately replaced by the SSV core below.
 	})
```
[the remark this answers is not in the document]
iurii-ssv: `SetupLogging` reads no env vars — it builds its output set purely from the `cfg.Stderr/Stdout/File/URL` struct fields ([setup.go#L117-L138](https://github.com/ipfs/go-log/blob/v2.8.1/setup.go#L117-L138)). With all of them at their zero values, `outputPaths` is empty and `zap.Open()` with no paths returns `zapcore.AddSync(io.Discard)` ([zap writer.go#L93-L95](https://github.com/uber-go/zap/blob/v1.27.0/writer.go#L93-L95)) — so the core genuinely writes to `io.Discard`; nothing reaches stderr.

`GOLOG_OUTPUT` is only consulted in `configFromEnv()` ([setup.go#L347](https://github.com/ipfs/go-log/blob/v2.8.1/setup.go#L347)), which feeds the package's `init()` call `SetupLogging(configFromEnv())` — not the explicit `Config` passed here. So the "writes nowhere" comment is accurate and there's no stderr leak in the window before `SetPrimaryCore`. Keeping it as-is.

observability/log/libp2p_test.go
```diff
@@ -0,0 +1,25 @@
...
+	for _, subsystem := range []string{"swarm2", "basichost"} {
+		core := golog.Logger(subsystem).Desugar().Core()
+		require.True(t, core.Enabled(zapcore.DebugLevel), "%s should log at debug", subsystem)
+	}
+
+	// Any other subsystem stays at the error default: errors pass, debug is dropped.
+	other := golog.Logger("some-other-subsystem").Desugar().Core()
+	require.True(t, other.Enabled(zapcore.ErrorLevel))
+	require.False(t, other.Enabled(zapcore.DebugLevel))
+}
```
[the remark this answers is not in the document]
iurii-ssv: Good point — the existing test only exercises loggers created after the hook. The retrofit path does hold (subsystem loggers are built on the shared, mutable `loggerCore`, and `SetupLogging` mutates the per-subsystem `AtomicLevel`s in place, so already-created loggers pick up the new level), but it deserves its own test.

Adding `TestHookLibp2pLoggingRetrofitsExistingLogger`: it creates the `swarm2` logger *before* the hook, pins it to `error`, then asserts the hook raises that already-created logger to `debug` — covering the `init()`-order path. Verified passing.

iurii-ssv:
@greptile pls re-review

julienharbulot (dismissed):
Reviewed the logging hook and p2p integration. This looks good to me: the new flag is opt-in and leaves the existing go-log/GOLOG behavior untouched when disabled, the hook runs before the libp2p host setup so it applies to the relevant startup diagnostics, and the go-log integration uses the shared primary core plus subsystem atomic levels so already-created libp2p loggers are covered. Reusing the SSV encoder and file/stdout sinks also avoids opening a separate log file path for libp2p output.

observability/log/global.go
```diff
@@ -48,54 +50,124 @@ func SetGlobal(levelName string, levelEncoderName string, logFormat string, file
...
+		return fmt.Errorf("global logger not initialized")
 	}
 
-	lv2 := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
-		return true // debug log returns all logs
+	// Seed go-log's per-subsystem levels and default the rest to error. Setting
+	// them via SetupLogging (rather than SetLogLevel) registers the levels even
+	// for subsystem loggers that haven't been created yet, so the policy holds
+	// regardless of package init ordering.
+	golog.SetupLogging(golog.Config{
```
momosh-ssv, line 161: Worth considering: `SetupLogging` resets every existing subsystem level to `Level` before applying `SubsystemLevels`, so an operator combining `LIBP2P_TRACE=true` with their own `GOLOG_LOG_LEVEL` would have the env settings silently overridden. Seems fine for a debug flag, but maybe a line in the flag's doc comment noting it takes full ownership of go-log's level policy?

observability/log/global.go
```diff
@@ -48,54 +50,124 @@ func SetGlobal(levelName string, levelEncoderName string, logFormat string, file
...
+		CallerKey:        "caller",
+		EncodeCaller:     zapcore.ShortCallerEncoder,
+		EncodeDuration:   zapcore.StringDurationEncoder,
+		NameKey:          "name",
+		ConsoleSeparator: "\t",
+	}
+
+	var fileSyncer zapcore.WriteSyncer
+	if fileOptions != nil {
+		fileSyncer = zapcore.AddSync(fileOptions.writer())
```
momosh-ssv, line 73: Should we wrap this in `zapcore.Lock` like `stdoutSyncer`? The SSV core and the libp2p core now share this one lumberjack instance — lumberjack's internal mutex makes it safe today, but I fear the asymmetry could bite if this writer is ever swapped for one without internal locking.

observability/log/libp2p_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"testing"
+
+	golog "github.com/ipfs/go-log/v2"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
+)
+
+func TestHookLibp2pLogging(t *testing.T) {
```
momosh-ssv, line 23: Might be worth a `t.Cleanup` restoring go-log's state — these tests leave `swarm2`/`basichost` pinned to debug and the primary core replaced, which could surprise a future test in this package that touches go-log.

observability/log/global.go
iurii-ssv: Good call — documented it. Added a note that the flag takes full ownership of go-log's level policy and overrides any `GOLOG_LOG_LEVEL`, both on the [`Libp2pTrace` field](https://github.com/ssvlabs/ssv/blob/64fd1c9dc71ee2673ac0b47e9dea2ee3a5571f8c/network/p2p/config.go#L61-L64) and at the [`SetupLogging` call](https://github.com/ssvlabs/ssv/blob/64fd1c9dc71ee2673ac0b47e9dea2ee3a5571f8c/observability/log/global.go#L151-L155). The reset-to-error is intentional (the flag's contract is exactly `error,swarm2=debug,basichost=debug`); the doc just makes that explicit.

observability/log/global.go
iurii-ssv: I'd lean against the lock here. `lumberjack.Logger.Write` is internally mutex-guarded ([lumberjack.go](https://github.com/natefinch/lumberjack/blob/v2.0.0/lumberjack.go#L188-L191)), and each `zapcore.Core.Write` emits one fully-encoded entry per call — so the SSV and libp2p cores sharing this one instance is already race-free *and* non-interleaving; a `zapcore.Lock` would just be redundant double-locking.

The asymmetry with `stdoutSyncer` is real though, and your instinct to question it is fair: that one needs the lock precisely because raw `os.Stdout` gives no such guarantee. I added a [comment](https://github.com/ssvlabs/ssv/blob/64fd1c9dc71ee2673ac0b47e9dea2ee3a5571f8c/observability/log/global.go#L68-L70) spelling out why the file syncer is safe without it, so the difference doesn't read as an oversight. Good with that?

observability/log/libp2p_test.go
iurii-ssv: Added — a [`resetGoLog(t)`](https://github.com/ssvlabs/ssv/blob/64fd1c9dc71ee2673ac0b47e9dea2ee3a5571f8c/observability/log/libp2p_test.go#L11-L19) helper that uses `t.Cleanup` to reset all subsystem levels back to error (and discard the primary core) after each hook test, so the raised `swarm2`/`basichost` levels and the replaced core don't leak into other tests in this package.

observability/log/global.go
momosh-ssv: Looks great, thanks — the config.example.yaml note is a nice touch on top.

observability/log/global.go
momosh-ssv: Yep, good with that — checked lumberjack's `Write` and it does hold the mutex for the whole call, and each zap core write lands as a single `Write`, so the lock would indeed be redundant.

observability/log/libp2p_test.go
momosh-ssv: Thanks — and the end-to-end file-sink test is a nice bonus, that was the one path the original tests skipped.

momosh-ssv (approved):

alok-ssv (approved):
Enabling Libp2pTrace intentionally overrides GOLOG_LOG_LEVEL behavior. That matches the implementation/comments; it may be worth emphasizing in release notes/operator docs to avoid surprises.

Issue #2429, opened 2025-08-08
proposer: QBFT round timeout based off of slot start time


iurii-ssv:
QBFT round timeouts are set in [`RoundTimeout` func](https://github.com/ssvlabs/ssv/blob/2976c7fe2cd59127c7b2f2f05bd8a31aa986d934/protocol/v2/qbft/roundtimer/timer.go#L102-L139) and are different for every duty type.

For proposer-duty round 1 deadlines (times out) at **2s after QBFT instance start**, while for committee-duty round 1 times out at **2s after slot start**. This is an important difference because anchoring the deadline at duty start can be problematic for 2 reasons:
- different Operators end up with the different deadline times, and that can even lead to QBFT being unable to change round (round 1 -> round 2, for example)
- for proposer-duty it typically takes 1s-1.5s to perform RANDAO + Fetch-Block steps before the corresponding runner it starts QBFT instance, this results in the deadline for round 1 being set to 3s-3.5s and given that Ethereum block needs to be proposed and spread out through most of the network in under 4s that leaves almost no time for round 2 (and block-propagation) in case it is needed

Thus it would be better if we could allocate the 4s time-chunk after slot start the proposer-duty has at its disposal fairly between round 1 and round 2 by basing round timeout value off of slot start time. Since proposer-duty can realistically do 1 round-change at most we don't need to worry about implementing complex calculations for round 3 and beyond. I suggest we set round 1 timeout at **2.5s after slot start**:
- it typically takes 1s-1.5s to perform RANDAO + Fetch-Block steps
- that leaves 1s-1.5s for QBFT round 1
- and in case we go into QBFT round 2 we have up to 1.5s to complete it (note, we should complete it well before 4s mark to have enough time to also complete QBFT post-consensus phase where the block gets signed and after that we need to broadcast signed block into Ethereum network)

Another idea would be to set round 1 timeout at **2.8s after slot start** to "favor" round 1 and accommodate more aggressive [`ProposerDelay` values](https://github.com/ssvlabs/ssv/blob/stage/docs/MEV_CONSIDERATIONS.md).

**IMPORTANT** to note, we might want to classify this change as **"fork"** (it will not work very well with the SSV nodes of older versions in the same cluster due to round-timeouts targeting different times - **2s since slot start** vs **2s since QBFT instance start**)

Prior mentions/discussions of this feature can be found here:
[...]

PR #2883, opened 2026-06-11
docs: QBFT decision timing analytics for proposer and committee duties

iurii-ssv:
Adds two analytics documents under `docs/analytics/` quantifying how QBFT decision timing relates to on-chain outcomes on Ethereum mainnet, based on 35 days of network-wide data (2026-05-07 → 2026-06-10). Docs-only change, no code.

## `docs/analytics/proposer/` — block proposals

44,838 proposer duties, 99.93% exporter-trace coverage, every miss verified against the canonical chain.

- Round 2 is rare (0.29%), but round-2 decisions miss **11.4%** of blocks vs **0.049%** for round-1 (~230×).
- The driver is wall-clock decision time vs the 4s attestation deadline: ~0.05% missed when QBFT decides by 3s, 30% at 3.5–4.0s, **100% (8/8) past 4.0s**. A round-2 decision completing by ~3.5s still lands 98% of the time.
- On-time round-1 misses are a separate failure class (relay/submission/orphaning, not consensus timing).
- Cross-references `docs/MEV_CONSIDERATIONS.md` (`ProposerDelay` headroom: p50 decision 1.40s, p99 2.25s) and [#2429](https://github.com/ssvlabs/ssv/issues/2429).

## `docs/analytics/attester/` — committee duties (attestations)

Stratified sample of 36 epochs (216,243 committee duties, 3.89M validator-attestations) + complete cluster-failure forensics over the full window (7,732 events). The conclusion is the **opposite** of the proposer story:

- Attestation inclusion is insensitive to decision time: **0 missed attestations across the 2,826 round-2-decided duties** (19,463 attestations), 0 across round-3+, 0 in duties decided 7–8.5s into the slot.
- **86% of missed attestations come from duties whose QBFT never decided**; full-window forensics: 82% never decided, 13% decided on time and failed at submission, only 15 of 7,611 events decided ≥7s. Five chronic clusters account for 82% of all cluster-failure events — reliability is an operator-health problem, not a timing problem.
- Lateness costs quality, not inclusion: first-possible-block inclusion drops 99% → 57% across 7–8s, and the 4s no-block fallback path costs head-vote correctness (~96.4%), which round-2 retries heal (99.6%+) — round changes improve head votes.
- Miss data guarded against the monitor's known whole-committee false-negative mode via a full-window block-gap scan (exactly one gap slot found in 35 days, the previously documented incident; affected events excluded).

## Contents

- `docs/analytics/proposer/README.md` + miss-rate-vs-decision-time chart
- `docs/analytics/attester/README.md` + four charts: miss sources, quality vs decision time, and round-1-only / round-2-only conditioned variants
- A note in the attester doc explaining why round changes are ~7x more common for committee duties than proposals (instance-relative vs slot-aligned timers, slot-correlated late-block bursts, per-slot vs per-validator weighting)
- Step-by-step methodology in both Reproduction sections: pipeline scripts, endpoints and server caps, sampling design (stratified one-epoch-per-day, fixed seed), metric definitions, and re-run constraints
- The two docs cross-link each other

Data sources for both: e2m duty monitor (duty enumeration/outcomes), exporter duty traces (QBFT rounds and message timing, network-wide P2P observation), public beacon API (canonical-chain verification).


docs/analytics/attester/README.md
```diff
@@ -0,0 +1,107 @@
...
+- **False-negative guard**: the monitor's known whole-committee false-negative mode is triggered by block-ingestion gaps in its database. A full-window scan comparing its block table against the canonical chain (via a public beacon API) found **exactly one** gap slot in 35 days — the previously documented incident — affecting zero sampled misses; forensics events near it were excluded.
+- **Decision time** = arrival of the earliest decided-quorum message at the exporter, minus slot start (P2P skew typically ≤300ms).
+
+### Caveats
+
+- The ≥8s region is thin in the sample (28 duties / 128 attestations); conclusions there lean on the full-window forensics, which point the same way.
+- Zero observed misses in round 2/3+ bounds the rate by sample size, not to literal zero.
+- Scattered misses of 1–4 validators within decided duties exist (they are the 0.0011%) and are not part of the ≥5-validator forensics set.
```
momosh-ssv: Might be worth reconciling this with the TL;DR — 41 misses from just four duties averages ~10 per duty, so at least one of those duties must have had ≥5 validators missing together. 

Seems that contradicts both the "1–4 validators" bound and these being outside the ≥5-validator forensics set; maybe the bound or the duty count needs adjusting.

docs/analytics/attester/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+This is the committee-duty (attestation) companion to the [proposer analysis](../proposer/README.md), asking the same question for attestations: how does the timing of the committee QBFT decision relate to whether the cluster's attestations land on chain?
+
+## TL;DR — the opposite of the proposer story
+
+For block proposals, decision time is destiny (100% missed past 4.0s into the slot). For committee duties, decision time is **almost irrelevant to inclusion**:
+
+- **0 of the 19,463 attestations from the 2,826 round-2-decided duties were missed** (CI95 ≤ 0.02%); likewise 0 of 128 from round-3+ decisions, and 0 of 355 from the 61 duties decided 7–8.5s into the slot. Round-1 decisions miss 0.0011%. Even the handful of duties decided past 10s were included, ~13 slots late.
+- Of the 1,500 missed validator-attestations in the sample, 97% came from duties that never produced a decision: 86% where consensus messages were observed but a decide quorum never formed, and 12% where the exporter saw no consensus activity for the duty at all. The remaining 3% (41 misses, from just four duties) completed consensus early — between 2.3s and 5.1s — and failed after it.
```
momosh-ssv: Worth considering rounding these consistently — 86% and 12% read as the decomposition of the 97% but sum to 98. The underlying fractions (85.5% + 11.7% = 97.3%) each round up; maybe 86% + 11%, or one decimal place, would avoid the double-take.

docs/analytics/attester/README.md
```diff
@@ -0,0 +1,107 @@
...
+- **Outcomes**: SSV Labs' internal duty monitor (e2m) per-validator attestation rows — included/missed, inclusion slot (delay), head-vote correctness — for the sampled epochs; `success=false` enumeration over the full window (335,077 misses = 0.041% of ~808M validator-duties) for the forensics.
+- **False-negative guard**: the monitor's known whole-committee false-negative mode is triggered by block-ingestion gaps in its database. A full-window scan comparing its block table against the canonical chain (via a public beacon API) found **exactly one** gap slot in 35 days — the previously documented incident — affecting zero sampled misses; forensics events near it were excluded.
+- **Decision time** = arrival of the earliest decided-quorum message at the exporter, minus slot start (P2P skew typically ≤300ms).
+
+### Caveats
+
+- The ≥8s region is thin in the sample (28 duties / 128 attestations); conclusions there lean on the full-window forensics, which point the same way.
```
momosh-ssv: What do you think about phrasing this as the round-3+ region instead? 

The numbers match the round-3+ row exactly, but the TL;DR mentions round-2 duties decided up to 8.5s, so the ≥8s region presumably isn't only round-3+.

docs/analytics/attester/README.md
iurii-ssv: Went with `86% + 11%` so the split sums to the 97% (and `86 + 11 + 3 = 100`). The 11% is `176/1500 = 11.7%` nudged down so the arithmetic reads clean — happy to switch to one decimal place instead if you'd rather not round that way.

docs/analytics/attester/README.md
iurii-ssv: Good point: `28 duties / 128 attestations` is the round-3+ count exactly, and the ≥8s *time* region also picks up late round-2 duties, so that label was wrong for these numbers. Relabeled the caveat to `Round-3+`.

docs/analytics/attester/README.md
iurii-ssv: Reconciled: `four duties` is the correct figure, so the caveat was the loose part. Dropped the "1–4 validators" bound and the "not part of the forensics set" claim; it now notes that those four duties (the ones with ≥5 validators missing together) *do* appear in the full-window forensics under "decided round 1 (on time), failed downstream" — i.e. the sample misses and the forensics set overlap rather than partition.

PR #2884, opened 2026-06-11
exporter: archive-mode duty-trace collector shutdown panic + flaky peer-score test

iurii-ssv:
After rebasing onto `stage` (#2876, errgroup-coordinated shutdown), the slot-pruning half of this PR is now handled upstream, so the remaining scope is the **archive-mode duty-trace collector** shutdown panic plus a flaky peer-score test. (Originally this also fixed the exporter-standard slot-pruning panic in `cli/operator/node.go`; see §2.) Surfaced while investigating CI on #2882, split out here and independent of it.

## 1. `operator/dutytracer` — `panic: pebble: closed` on shutdown (exporter-archive)

In archive mode the duty-trace collector writes to the pebble DB from goroutines that outlived `Collector.Start`: the eviction loop, the schedule filler/worker, and a detached `go c.collectLateMessage(…)` per late message. On shutdown the ctx was canceled and the DB closed out from under any still mid-write → `pebble: closed` panic.

Fix — give the collector ownership of its own goroutines so `Start` becomes a real join point:
- `Collector.Start` now joins the schedule filler + worker before returning, then joins in-flight late-message retries via `stopLateCollect()`.
- Late-message retries are tracked by a `WaitGroup`; `collectLateAsync` refuses to spawn new ones once shutdown has begun (`lateClosed` under `lateMu`), which keeps `lateWG.Add` from racing `lateWG.Wait`.

The net effect: `Start(ctx)` blocks until all its writers have stopped, so any supervisor that waits for `Start` to return before closing the DB is race-free.

## 2. Relationship to `stage` / #2876

#2876 (now in `stage`) replaced the node's ad-hoc shutdown with an errgroup-coordinated `supervise`: long-lived services — including slot-pruning's `PruneContinuously` and the duty-trace collector — are spawned under a `WaitGroup`, and shutdown runs `wg.Wait()` before `node.close()` closes the DB. That structurally fixes the **exporter-standard slot-pruning** panic this PR originally addressed in `cli/operator/node.go`, so after merging `stage` this PR no longer touches `node.go` — that half is provided upstream.

The collector fix in §1 is still required and is load-bearing under this model: `supervise`'s `wg.Wait()` only closes the archive-mode race if `Collector.Start(ctx)` truly blocks until its internal writers stop, which is exactly what §1 establishes.

## 3. `network/p2p` — flaky `TestP2pNetwork_MessageValidation`

The test waits for live GossipSub peer scores to settle into rate/bucket/strict-separation invariants via `assert.Eventually`. On loaded CI runners the 15s ceiling wasn't enough decay cycles, so it intermittently timed out (e.g. scores `0.14` vs `0.28`, not yet separated). Raised the ceiling to 30s; the happy path still returns as soon as the invariants hold, so green runs aren't slowed. A stability stopgap, not a fix for the underlying lack of peer-score determinism.

## Testing
- `operator/dutytracer` passes under `-race`, including `TestCollector_stopLateCollect` covering the late-message shutdown coordination: it waits for an in-flight retry, drops post-close spawns, and stress-races concurrent spawns against shutdown.
- `go build ./...` / `go vet ./...` clean across the merged tree.
- `TestP2pNetwork_MessageValidation` can't be run to completion locally (mDNS local discovery doesn't function in this environment — it hangs identically on `stage`), so the timeout change is verified by reasoning, not a local green run.

## Follow-up
- #2889 — make `Prune` / `removeSlotsOlderThan` honor ctx (the initial slot GC is currently uncancellable); contained to `ibft/storage/store.go`.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -54,24 +54,29 @@ import (
 	"github.com/ssvlabs/ssv/storage/basedb"
 )
 
-func runNode(ctx context.Context, cfg *config, logger *zap.Logger) error {
-	// runNode owns the parent ctx for the clients it builds directly (goclient, execution) before
-	// newNode exists. goclient has no Close() of its own — ctx cancellation is its only shutdown —
-	// so cancel on return to release it once the node has stopped and we unwind.
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+// shutdownGraceTimeout bounds the whole graceful teardown — services unwinding plus the node close —
+// once the first terminal event has landed. Past it teardown gives up and surfaces the terminal
+// cause instead of hanging forever on something ignoring the cancellation convention; the leftovers
+// are reaped at process exit. 15s is generous against the designed sub-second unwind and leaves
+// room ahead of orchestrator kill grace.
+const shutdownGraceTimeout = 15 * time.Second
 
+// buildNode builds the node graph: it validates config up front (before any network I/O, so a
+// misconfigured node fails fast), connects the CL/EL clients — binding them to ctx, whose
+// cancellation is goclient's only shutdown — and wires everything together via newNode. The returned
+// *node owns the resources newNode opened; node.close releases them at teardown.
+func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, error) {
 	// Validate the configuration up front, before any network I/O, so a misconfigured node
 	// fails fast (e.g. an invalid ProposerDelay is rejected before the beacon client is built,
 	// rather than after connecting to the beacon). resolveAndValidate only reads cfg.
 	res, err := cfg.resolveAndValidate(logger)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	ssvNetworkConfig, err := setupSSVNetwork(logger, cfg)
 	if err != nil {
-		return fmt.Errorf("could not setup network: %w", err)
+		return nil, fmt.Errorf("could not setup network: %w", err)
 	}
 
 	logger.Info("connecting CL(s)",
@@ -82,14 +87,14 @@ func runNode(ctx context.Context, cfg *config, logger *zap.Logger) error {
 
 	cliopt, err := goclient.NewOptions(cfg.ConsensusClient, cfg.ProposerDelay)
 	if err != nil {
-		return startupError{
+		return nil, startupError{
 			err:    fmt.Errorf("failed to create beacon client options: %w", err),
 			fields: []zap.Field{fields.Address(cfg.ConsensusClient.BeaconNodeAddr)},
 		}
 	}
 	consensusClient, err := goclient.New(ctx, logger, cliopt)
 	if err != nil {
-		return startupError{
+		return nil, startupError{
 			err:    fmt.Errorf("failed to create beacon go-client: %w", err),
 			fields: []zap.Field{fields.Address(cfg.ConsensusClient.BeaconNodeAddr)},
 		}
@@ -107,7 +112,7 @@ func runNode(ctx context.Context, cfg *config, logger *zap.Logger) error {
 		}
 	}
 	if len(executionAddrList) == 0 {
-		return fmt.Errorf("no execution node address provided")
+		return nil, fmt.Errorf("no execution node address provided")
 	}
 
 	logger.Info("connecting EL(s)",
@@ -127,7 +132,7 @@ func runNode(ctx context.Context, cfg *config, logger *zap.Logger) error {
 			executionclient.WithSyncDistanceTolerance(cfg.ExecutionClient.SyncDistanceTolerance),
 		)
 		if err != nil {
-			return fmt.Errorf("could not connect to execution client: %w", err)
+			return nil, fmt.Errorf("could not connect to execution client: %w", err)
 		}
 
 		executionClient = ec
@@ -141,7 +146,7 @@ func runNode(ctx context.Context, cfg *config, logger *zap.Logger) error {
 			executionclient.WithSyncDistanceToleranceMulti(cfg.ExecutionClient.SyncDistanceTolerance),
 		)
 		if err != nil {
-			return fmt.Errorf("could not connect to execution client: %w", err)
+			return nil, fmt.Errorf("could not connect to execution client: %w", err)
 		}
 
 		executionClient = ec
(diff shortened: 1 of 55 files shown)


cli/operator/node.go
```diff
@@ -461,16 +473,21 @@ func newNode(ctx context.Context, cfg *config, logger *zap.Logger, res resolved,
...
-// calls validatorCtrl.Stop() (the controller's ttlcache cleanup loops, which aren't ctx-bound), then
+// (stopping the VRSubmitter and, in exporter modes, the duty-trace collector / slot-pruning), waits
+// for the slot-pruning goroutines to exit so they can't touch the DB after it closes, and calls
+// validatorCtrl.Stop() (the controller's ttlcache cleanup loops, which aren't ctx-bound), then
 // closes the CLI-owned resources (p2p network and execution client, then the db that p2p depends on).
 // newNode is the only constructor, so all these fields are always set.
 func (n *node) Close() error {
 	n.cancel()
+	if n.stopSlotPruning != nil {
+		n.stopSlotPruning()
```
momosh-ssv: Should we consider giving the archive-mode collector the same treatment? 

Seems that `go collector.Start(ctx, slotTickerProvider)` (line 412) is still unjoined — it evicts duty traces to the DB on every slot tick (and spawns `startScheduleFiller`/`runScheduleWorker`, which also write), so after `n.cancel()` a mid-evict collector can hit the closed DB the same way the prune loops did. 

Maybe a `stopCollector` join here, before the closes, would cover it.

cli/operator/node.go
```diff
@@ -461,16 +473,21 @@ func newNode(ctx context.Context, cfg *config, logger *zap.Logger, res resolved,
 		collector:           collector,
 		p2pNetwork:          p2pNetwork,
 		operatorNode:        operatorNode,
+		stopSlotPruning:     stopSlotPruning,
```
momosh-ssv: Might be worth asserting `stopSlotPruning != nil` in `Test_newNode_wiresExporterNode` for the standard-mode case — since `Close()` nil-guards the field, I fear a refactor that drops this assignment would still pass tests and quietly reintroduce the panic.

cli/operator/node.go
iurii-ssv: Good catch — addressed in the latest push.

The collector turned out to be the same class as the prune race: [`Start`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/operator/dutytracer/collector.go#L132) is launched with `go` and writes to the same pebble DB from three places that were never joined before `db.Close()` — the eviction loop, the schedule filler/worker, and the detached late-message retries spawned by [`Collect`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/operator/dutytracer/collector.go#L547).

So rather than only wrapping the `go collector.Start`, I gave the collector ownership of all of them: `Start` now waits for the schedule helpers and any in-flight late-message retries before it returns, and `Collect` routes late retries through a tracked spawn ([`collectLateAsync`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/operator/dutytracer/collector.go#L559)) that's refused once the join begins ([`stopLateCollect`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/operator/dutytracer/collector.go#L575)), so `Add` can't race `Wait`. [`node.Close()`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/cli/operator/node.go#L499) then drives it through a `stopCollector` func mirroring `stopSlotPruning` ([`startCollector`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/cli/operator/node.go#L743): child ctx + WaitGroup), called before `db.Close()` and on the assembly-failure path.

cli/operator/node.go
iurii-ssv: Done — [`Test_newNode_wiresExporterNode`](https://github.com/ssvlabs/ssv/blob/fix-prune-shutdown-and-scoring-flake/cli/operator/node_test.go#L186-L195) now asserts both stop funcs per mode. They're mode-complementary: `stopSlotPruning` is non-nil in standard and `stopCollector` non-nil in archive, each nil in the other mode. So a refactor that drops either assignment fails the test instead of slipping past `Close()`'s nil-guard.

alok-ssv (dismissed):

cli/operator/node.go
```diff
@@ -664,28 +697,63 @@ func setupOperatorDataStore(
 	return operatordatastore.New(operatorData), nil
 }
 
-func initSlotPruning(ctx context.Context, stores *ibftstorage.ParticipantStores, slotTickerProvider slotticker.Provider, slot phase0.Slot, retain uint64) {
-	var wg sync.WaitGroup
+// initSlotPruning runs the initial stale-slot GC and then starts the continuous
+// pruning loops on a child of ctx. It returns a stop function that cancels those
+// loops and waits for them to exit; callers must invoke it before closing the DB,
+// otherwise an in-flight prune can run against a closed DB and panic.
+func initSlotPruning(ctx context.Context, stores *ibftstorage.ParticipantStores, slotTickerProvider slotticker.Provider, slot phase0.Slot, retain uint64) (stop func()) {
```
momosh-ssv: Worth considering for a follow-up — seems that `Prune` → `removeSlotsOlderThan` never checks `ctx`, so this initial GC blocks `newNode` unboundedly even if startup is cancelled mid-way. Pre-existing rather than introduced here (and the new stop func can't help — it hasn't been returned yet). Maybe make `removeSlotsOlderThan` honor ctx in a follow-up.

iurii-ssv: You're right, and it's a real gap. I checked #2876 (the errgroup lifecycle work): it restructures the caller (`initSlotPruning` → `startSlotPruning`, moved into `start()`) but doesn't touch `ibft/storage/store.go` — so [`Prune`](https://github.com/ssvlabs/ssv/blob/stage/ibft/storage/store.go#L46) still ignores its ctx and [`removeSlotsOlderThan`](https://github.com/ssvlabs/ssv/blob/stage/ibft/storage/store.go#L115) still loops without a cancellation check. The initial `wg.Wait()` stays uncancellable there too.

Tracked in #2889 with the proposed fix (thread ctx into `removeSlotsOlderThan`, have `Prune` pass it through — contained to `ibft/storage/store.go`, so it conflicts with neither this PR nor #2876) and a merge-ordering note. Keeping it out of this PR as you suggested.

iurii-ssv:
@greptile pls re-review

Roy-blox (dismissed):
Approving after investigating the shutdown panic paths end-to-end.

**Fix 1 — `initSlotPruning` / exporter-standard:**
The `PruneContinuously` goroutines were launched on the node ctx with no join point before `db.Close()`. A prune iteration already past its `ctx.Done()` check could call into a closed pebble DB and panic (`pebble: closed`). The new child-ctx + `stop func()` pattern ensures all prune loops are cancelled and joined before the DB closes. LIFO deferred-close ordering is safe since the child ctx cancel doesn't depend on parent resources.

**Fix 2 — dutytracer collector / exporter-archive:**
Late-message goroutines were fire-and-forget after `Collector.Start` returned. The `WaitGroup` + `lateClosed` guard correctly prevents new spawns once shutdown begins and waits for in-flight retries to drain. `stopLateCollect()` is wired into both the `Close()` path and the `newNode` assembly-failure path, covering all shutdown routes.

CI green across all 7 checks as of Jun 14 (lint, unit, spec-tests, signer E2E, codecov). Merging.

oleg-ssvlabs (dismissed):
lgtm

momosh-ssv (dismissed):
real great job now!

oleg-ssvlabs (approved):

iurii-ssv:
@greptile pls re-review

momosh-ssv (approved):

PR #839, opened 2023-02-07
Sync committee duties improvements&fixes

olegshmuelov:

PR #777, opened 2022-12-22
Multi duties

nivBlox:

PR #2885, opened 2026-06-12
fix: fetch head block root for sync-committee contributions

iurii-ssv:
## Problem

When the duty slot's block proposal is missed (or arrives later than 1/3 into the slot), the `SYNC_COMMITTEE_CONTRIBUTION` duty fails for the entire cluster:

```
could not get sync committee contribution: multi-client request -> BeaconBlockRoot: fetch beacon block root: GET failed with status 404: {"code":404,"message":"Block not found for id '3252140'"}
```

Observed on Hoodi at slot [3252140](https://dora.hoodi.ethpandaops.io/slot/3252140) (missed proposal, finalized). `GetSyncCommitteeContribution` resolves the contribution's `beacon_block_root` via `GET /eth/v1/beacon/blocks/{slot}/root`, and the beacon API answers 404 for an empty slot — a correct and permanent answer, so the duty terminally fails whenever a cluster validator is a sync-committee aggregator for a slot with no block. The failure is deterministic across all operators (each queries its own node for a block that doesn't exist).

## Fix

Query `head` instead of the slot number. Per the consensus spec, sync-committee messages at slot N sign the head root as seen at slot N — for an empty slot that's an earlier block's root. The committee runner already behaves this way (sync messages sign the QBFT-decided beacon vote, i.e. attestation data's head vote), so in the common case `head` matches what the cluster's own messages signed — though a freshly-queried `head` can diverge from that decided root under a reorg or a lagging node. #839 made the identical change for the sync-committee message path; the contribution path kept the by-slot lookup (since #777) and was never aligned.

## Backward compatibility

Fully backward compatible — safe to roll into a live, partially-upgraded cluster with no coordinated upgrade:

- **Beacon API:** `head` and `<slot>` are both standard `block_id` values for `GET /eth/v1/beacon/blocks/{block_id}/root`; `head` is supported by every consensus client. No new endpoint or client-version requirement.
- **Mixed-version clusters (no consensus split):** the fetched contribution is wrapped in `ValidatorConsensusData` and run through QBFT, where only the leader's value is agreed and followers validate it. The contribution value-check (`syncCommitteeContributionChecker` → `checkValidatorConsensusData`) and the spec's `ValidatorConsensusData.Validate()` only verify duty type / pubkey / validator index / epoch and that the contributions decode — neither pins nor re-derives the `beacon_block_root`. A follower therefore accepts whatever root the leader proposed, whether it came from `head` or by-slot, so new and old operators interoperate either way. On an empty slot an old leader still 404s, but a QBFT round change can hand leadership to an upgraded operator — so a mixed cluster is never worse than today and gets the full fix once operators upgrade. No wire/message-format, signing-domain, or consensus-value-layout change.
- **No storage/format changes.** The only Go signature change (`waitOneThirdIntoSlot`) is unexported with a single in-package caller.

## Cleanups in the touched flow

- `waitOneThirdIntoSlot` now mirrors `waitTwoThirdsIntoSlot`: context cancellation surfaces as an error instead of letting the block-root request fire with a cancelled context.
- The 2/3-wait error path returns `DataVersionNil` rather than a literal `0` (i.e. `DataVersionUnknown`), matching the sentinel used by the function's other error paths.
- Dropped a loop-var copy made obsolete by Go 1.22 semantics.
- Completed/fixed stale doc comments in the file (truncated `GetSyncCommitteeContribution` doc, `IsSyncCommitteeAggregator` typo, vague submit doc, leftover narration comments).

## Tests

- Updated `TestGetSyncCommitteeContributionPreservesInputOrdering`, whose mock previously pinned the by-slot request shape, to assert `head`.
- Added `TestGetSyncCommitteeContributionFetchesHeadRoot`: the mock 404s any by-slot lookup and only answers `head`, so it reproduces the original failure (and fails if the query ever regresses to by-slot).

### DIFF
--- a/beacon/goclient/sync_committee_contribution.go
+++ b/beacon/goclient/sync_committee_contribution.go
@@ -58,6 +58,9 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 	// Resolve the contribution root from head rather than by slot: sync-committee messages
 	// sign the head root as seen during the duty slot, which for a missed or late proposal
 	// is an earlier block's root — a by-slot lookup would 404 there and fail the duty.
+	// In the common case head matches the root the cluster's own messages signed; those
+	// sign the QBFT-decided head vote, which can diverge from a freshly-queried head under
+	// a reorg or a lagging node.
 	start := time.Now()
 	beaconBlockRootResp, err := gc.multiClient.BeaconBlockRoot(ctx, &api.BeaconBlockRootOpts{
 		Block: "head",
@@ -66,11 +69,8 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 	if err != nil {
 		return nil, DataVersionNil, errMultiClient(fmt.Errorf("fetch beacon block root: %w", err), "BeaconBlockRoot")
 	}
-	if beaconBlockRootResp == nil {
-		return nil, DataVersionNil, errMultiClient(fmt.Errorf("beacon block root response is nil"), "BeaconBlockRoot")
-	}
-	if beaconBlockRootResp.Data == nil {
-		return nil, DataVersionNil, errMultiClient(fmt.Errorf("beacon block root response data is nil"), "BeaconBlockRoot")
+	if err := checkPtrResponse(beaconBlockRootResp, "beacon block root"); err != nil {
+		return nil, DataVersionNil, errMultiClient(err, "BeaconBlockRoot")
 	}
 
 	blockRoot := beaconBlockRootResp.Data
@@ -96,11 +96,8 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 			if err != nil {
 				return errMultiClient(fmt.Errorf("fetch sync committee contribution: %w", err), "SyncCommitteeContribution")
 			}
-			if syncCommitteeContrResp == nil {
-				return errMultiClient(fmt.Errorf("sync committee contribution response is nil"), "SyncCommitteeContribution")
-			}
-			if syncCommitteeContrResp.Data == nil {
-				return errMultiClient(fmt.Errorf("sync committee contribution response data is nil"), "SyncCommitteeContribution")
+			if err := checkPtrResponse(syncCommitteeContrResp, "sync committee contribution"); err != nil {
+				return errMultiClient(err, "SyncCommitteeContribution")
 			}
 
 			contribution := syncCommitteeContrResp.Data
--- a/beacon/goclient/sync_committee_contribution.go
+++ b/beacon/goclient/sync_committee_contribution.go
@@ -18,12 +18,11 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
-// IsSyncCommitteeAggregator returns tru if aggregator
+// IsSyncCommitteeAggregator reports whether the given selection proof makes the validator
+// a sync-committee aggregator.
 func (gc *GoClient) IsSyncCommitteeAggregator(proof []byte) bool {
-	// Hash the signature.
 	hash := sha256.Sum256(proof)
 
-	// Keep the signature if it's an aggregator.
 	cfg := gc.BeaconConfig()
 
 	// as per spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#aggregation-selection
@@ -35,12 +34,13 @@ func (gc *GoClient) IsSyncCommitteeAggregator(proof []byte) bool {
 	return binary.LittleEndian.Uint64(hash[:8])%modulo == 0
 }
 
-// SyncCommitteeSubnetID returns sync committee subnet ID from subcommittee index
+// SyncCommitteeSubnetID returns the sync-committee subnet ID for the given subcommittee index.
 func (gc *GoClient) SyncCommitteeSubnetID(index phase0.CommitteeIndex) uint64 {
 	return uint64(index) / (gc.BeaconConfig().SyncCommitteeSize / gc.BeaconConfig().SyncCommitteeSubnetCount)
 }
 
-// GetSyncCommitteeContribution returns
+// GetSyncCommitteeContribution fetches one contribution per requested subnet, each paired
+// with its selection proof, all for the head block root as of the duty slot.
 func (gc *GoClient) GetSyncCommitteeContribution(
 	ctx context.Context,
 	slot phase0.Slot,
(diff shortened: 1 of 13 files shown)


beacon/goclient/sync_committee_contribution.go
```diff
@@ -51,13 +51,18 @@ func (gc *GoClient) GetSyncCommitteeContribution(
...
+	}
 
-	scDataReqStart := time.Now()
+	// Resolve the contribution root from head rather than by slot: sync-committee messages
+	// sign the head root as seen during the duty slot, which for a missed or late proposal
+	// is an earlier block's root — a by-slot lookup would 404 there and fail the duty.
+	start := time.Now()
 	beaconBlockRootResp, err := gc.multiClient.BeaconBlockRoot(ctx, &api.BeaconBlockRootOpts{
-		Block: fmt.Sprint(slot),
+		Block: "head",
```
momosh-ssv, line 66: Any reason we re-query the local node's `head` here rather than reusing the QBFT-decided `beaconVote.BlockRoot` the committee runner already signs the sync messages against (committee.go ~973)? 

In the common case they're identical, but on a reorg — or if `multiClient` routes this read to a node lagging the one that served the attestation data — `head` could resolve to a different root than the cluster's messages signed, and reusing the decided root would also drop a round-trip. 

Head is the canonical choice and the contribution aggregates the whole subnet anyway, so this may well be immaterial — mainly curious whether threading the decided root through was considered.

beacon/goclient/sync_committee_test.go
```diff
@@ -300,6 +301,60 @@ func TestGetSyncCommitteeContributionPreservesInputOrdering(t *testing.T) {
 	}
 }
 
+// TestGetSyncCommitteeContributionFetchesHeadRoot is a regression test for the
+// empty-slot 404 bug: the contribution root must be resolved from "head", never the
+// duty slot. When the slot's proposal is missed, a by-slot block-root lookup 404s and
+// terminally fails the duty for the whole cluster, whereas "head" resolves to the
+// previous block's root — exactly what the cluster's sync-committee messages signed.
```
momosh-ssv: Might be worth softening "exactly what the cluster's sync-committee messages signed" — the messages sign the QBFT-decided `beaconVote.BlockRoot`, while this path resolves `head` independently, so in rare reorg/lagging-node cases the two could differ. 

Something like "the root the cluster's messages signed in the common case" would read a touch more precisely.

beacon/goclient/sync_committee_contribution.go
iurii-ssv: Good question — using `head` rather than threading the decided root was deliberate, for two reasons:

1. **The decided root isn't in scope at this call site.** This fetch runs in the `SyncCommitteeAggregatorRunner`'s pre-consensus phase — [`GetSyncCommitteeContribution` is called](https://github.com/ssvlabs/ssv/blob/62fe6fa8a8e9dfdc9bbde3e80c2c3ad8a69e4a5f/protocol/v2/ssv/runner/sync_committee_contribution.go#L186) before that runner's own [`decide`](https://github.com/ssvlabs/ssv/blob/62fe6fa8a8e9dfdc9bbde3e80c2c3ad8a69e4a5f/protocol/v2/ssv/runner/sync_committee_contribution.go#L204), and its QBFT decides on the *contributions*, not a `BeaconVote`. The `beaconVote.BlockRoot` you're pointing at lives in the separate [`CommitteeRunner`](https://github.com/ssvlabs/ssv/blob/62fe6fa8a8e9dfdc9bbde3e80c2c3ad8a69e4a5f/protocol/v2/ssv/runner/committee.go#L985) — a distinct runner/duty with its own QBFT instance. So reusing it wouldn't be a local change; it'd mean coupling the aggregator to the committee runner's decided value for the slot (and handling the cases where that duty hasn't decided yet / isn't running).
2. **`head` is the spec-canonical choice and keeps this aligned with the message path.** #839 put the sync-committee *message* path on `head`; this brings the contribution path in line. Since the contribution aggregates the whole subnet, even when a freshly-queried `head` differs from the decided root (reorg / lagging node) `head` is the better bet on what the subnet actually signed, and the aggregator needn't include its own message.

You're right it drops a round-trip and would be marginally more consistent on the cluster's own messages — happy to file a follow-up if we want to thread the decided root through, but it felt out of scope for a minimal, backward-compatible bugfix.

beacon/goclient/sync_committee_test.go
iurii-ssv: Agreed, it overstated it. Softened to "in the common case the root the cluster's sync-committee messages signed" in the test comment, and added the same caveat to the production comment in `GetSyncCommitteeContribution` (and the PR description): the cluster's messages sign the QBFT-decided head vote, which can diverge from a freshly-queried `head` under a reorg or lagging node.

beacon/goclient/sync_committee_contribution.go
momosh-ssv: > happy to file a follow-up if we want to 

No need, since the intentional divergence is now well documented.
And this is the right outcome, when you look at it – resolve-and-move-on

momosh-ssv (approved):
real gj!

ovidiu-ssv-labs:
## Non-blocking finding

**Silent empty contribution on reorg / late block** — `beacon/goclient/sync_committee_contribution.go:66`

The head root is snapshotted at t₁/₃ (after `waitOneThirdIntoSlot`) and then reused at t₂/₃ when the `SyncCommitteeContribution` RPCs fire. On a reorg or a late block arriving in that ~4s window, the snapshotted root can diverge from the root that sync-committee messages actually signed (`beaconVote.BlockRoot`, QBFT-decided by the committee runner at ~t₀). The beacon node assembles the contribution from pooled messages for the root it receives — if no messages exist for that root, it returns a contribution with zero `aggregation_bits`. The duty completes without error but earns no reward.

The PR comment acknowledges this trade-off, and it's strictly better than the hard 404 it replaces. Two suggestions for follow-up:

1. **Pin the known degradation in a test** — a test case where `head` returns a root with no pooled messages and asserting `AggregationBits == 0` would make the accepted trade-off explicit and prevent future confusion about whether it's a bug.
2. **Longer-term fix** — accepting `beaconBlockRoot phase0.Root` as an explicit argument to `GetSyncCommitteeContribution` (letting the runner pass the QBFT-decided vote root directly) would eliminate the divergence window entirely, though it requires an interface change.

ovidiu-ssv-labs (approved):

PR #2888, opened 2026-06-12
fix: aggregate for the attestation data root the cluster attested with

iurii-ssv:
## Problem

`AGGREGATOR` duties occasionally fail with the beacon node reporting it has nothing to aggregate:

```
failed to submit aggregate and proof: multi-client request -> AggregateAttestation: fetch aggregate attestation: GET failed with status 404: {"code":404,"message":"No aggregated attestation for slot=3252713, dataRoot=0x22f764f9d98f16164e69a1abeb030d313c678711e6227a35246962211d772585, committeeIndex=21"}
```

Observed on Hoodi at slot 3252713 — a normal, finalized slot, so the aggregate genuinely didn't exist *for the requested root*. `SubmitAggregateSelectionProof` re-derives attestation data at 2/3 of the slot via `GetAttestationData` and requests an aggregate for its root. That data is this node's local view (typically served from the attestation-data cache populated at ~1/3 of the slot) — not necessarily what the cluster's QBFT consensus decided and attested with. When head views diverge mid-slot, the re-derived root can match nothing anyone attested with, and the beacon node correctly answers 404, failing the duty for the whole cluster.

## Fix

Per the consensus spec, the aggregate must match the aggregator's **own attestation data**. The decided data already flows through goclient in `SubmitAttestations`, so:

- `SubmitAttestations` now remembers the hash-tree-root of each submitted attestation's data, keyed by (slot, committee) — the committee is in the key because pre-Electra the committee index is part of the data (post-Electra the index is 0, per EIP-7549).
- `SubmitAggregateSelectionProof` requests the aggregate for that remembered root. The beacon node holds at least our own attestation matching it, so a matching aggregate must exist.
- If no attested root is known (attestation failed or hasn't landed by 2/3 of the slot), it falls back to the previous re-derivation behavior unchanged.

Known accepted imprecision: if two clusters on the same node share a (slot, committee) and their QBFT instances decided different roots (requires diverging leader views in the same slot), last submission wins — the requested root is still one this node attested with, so an aggregate still exists.

## Tests

- New `TestSubmitAggregateSelectionProof_PrefersAttestedDataRoot` covers the Electra (committee bits) and pre-Electra (data index) paths, asserting the aggregate is requested for the submitted root without re-deriving attestation data.
- The existing fork-matrix test now exercises the fallback path (no prior submission) with unchanged assertions.

- [ ] E2E passed - [action log](put link here)

### DIFF
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -29,24 +29,32 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
 		return nil, 0, fmt.Errorf("wait for 2/3 of slot: %w", err)
 	}
 
-	attData, _, err := gc.GetAttestationData(ctx, slot)
-	if err != nil {
-		return nil, DataVersionNil, fmt.Errorf("fetch attestation data: %w", err)
-	}
+	// Aggregate for the root of the attestation data this node actually submitted for this
+	// duty (the cluster-decided value): the beacon node then holds at least our own
+	// attestation matching it. Re-deriving the data locally can yield a root nobody
+	// attested with, which the node answers with a 404 (no matching aggregate).
+	root, found := gc.attestedDataRoot(slot, committeeIndex)
+	if !found {
+		// No record of our own attestation (it failed or hasn't landed yet) — fall back
+		// to re-deriving the root from this node's view of the slot.
+		attData, _, err := gc.GetAttestationData(ctx, slot)
+		if err != nil {
+			return nil, DataVersionNil, fmt.Errorf("fetch attestation data: %w", err)
+		}
 
-	// Explicitly set Index field as beacon nodes may return inconsistent values.
-	// EIP-7549: For Electra and later, index must always be 0, pre-Electra uses committee index.
-	config := gc.getBeaconConfig()
-	dataVersion, _ := config.ForkAtEpoch(config.EstimatedEpochAtSlot(slot))
-	attData.Index = 0
-	if dataVersion < spec.DataVersionElectra {
-		attData.Index = committeeIndex
-	}
+		// Explicitly set Index field as beacon nodes may return inconsistent values.
+		// EIP-7549: For Electra and later, index must always be 0, pre-Electra uses committee index.
+		config := gc.getBeaconConfig()
+		dataVersion, _ := config.ForkAtEpoch(config.EstimatedEpochAtSlot(slot))
+		attData.Index = 0
+		if dataVersion < spec.DataVersionElectra {
+			attData.Index = committeeIndex
+		}
 
-	// Get aggregate attestation data.
-	root, err := attData.HashTreeRoot()
-	if err != nil {
-		return nil, DataVersionNil, fmt.Errorf("fetch attestation data root: %w", err)
+		root, err = attData.HashTreeRoot()
+		if err != nil {
+			return nil, DataVersionNil, fmt.Errorf("fetch attestation data root: %w", err)
+		}
 	}
 
 	aggDataReqStart := time.Now()
--- a/beacon/goclient/aggregator_test.go
+++ b/beacon/goclient/aggregator_test.go
@@ -2,6 +2,7 @@ package goclient
 
 import (
 	"context"
+	"errors"
 	"sync/atomic"
 	"testing"
 	"testing/synctest"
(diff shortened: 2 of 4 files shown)


beacon/goclient/aggregator_test.go
```diff
@@ -156,6 +161,72 @@ func TestSubmitAggregateSelectionProof_UsesForkSpecificAggregateAndProof(t *test
...
+	committeeIndex := phase0.CommitteeIndex(7)
+
+	testCases := []struct {
+		name    string
+		version spec.DataVersion
+		epoch   phase0.Epoch
+	}{
+		{name: "electra committee comes from committee bits", version: spec.DataVersionElectra, epoch: 5},
+		{name: "pre-electra committee comes from data index", version: spec.DataVersionPhase0, epoch: 0},
+	}
```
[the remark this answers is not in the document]
iurii-ssv: Good catch — added a `DataVersionFulu` case (epoch 6) to `TestSubmitAggregateSelectionProof_PrefersAttestedDataRoot`, so the `att.Fulu.CommitteeBits` branch is now exercised end-to-end alongside Electra and pre-Electra.

beacon/goclient/aggregator.go
```diff
@@ -29,24 +29,32 @@ func (gc *GoClient) SubmitAggregateSelectionProof(
...
 
-	attData, _, err := gc.GetAttestationData(ctx, slot)
-	if err != nil {
-		return nil, DataVersionNil, fmt.Errorf("fetch attestation data: %w", err)
-	}
+	// Aggregate for the root of the attestation data this node actually submitted for this
+	// duty (the cluster-decided value): the beacon node then holds at least our own
+	// attestation matching it. Re-deriving the data locally can yield a root nobody
+	// attested with, which the node answers with a 404 (no matching aggregate).
+	root, found := gc.attestedDataRoot(slot, committeeIndex)
```
momosh-ssv, line 36: What happens if the aggregate request is served by a client that never accepted our attestations? 

In parallel-submissions mode the root is remembered once at least one client accepts, but `AggregateAttestation` goes through `multiClient`, so it could land on a node that missed our submission and only learns it via gossip. 

Seems fine in practice — just checking this corner was considered.

iurii-ssv: Yes — considered, and it's a non-issue in practice:

- In parallel-submissions mode `multiClientSubmit` fans out to *all* clients, so every healthy node already holds our attestation directly; the `AggregateAttestation` call lands on the active one.
- It's requested at 2/3 of the slot, ~one interval after the ~1/3 submission, so gossip has propagated attestations network-wide by then.
- Most importantly it's strictly better than the old re-derivation regardless: worst case (a node that missed our submission *and* hasn't seen it via gossip) we get the same 404 we'd have gotten before — no regression — while the common case is fixed.

That leaves exactly the corner noted under "accepted imprecision" in the description. If we ever actually see it, the cheap fix is to fall back to the re-derived root on a 404, but I'd rather not add the extra round-trip speculatively.

momosh-ssv: Thanks — the "worst case is the same 404 as before, no regression" point is the reassuring part. 
Agreed it's not worth the speculative round-trip. Resolving.

momosh-ssv (dismissed):
nice one!

ovidiu-ssv-labs:
## Non-blocking finding

**`multiClientSubmit` returns `nil` on zero submissions — `rememberAttestedDataRoots` fires with stale data** — `beacon/goclient/attest.go`

`multiClientSubmit` has three exit paths:
- submissions > 0 → `return nil` (at least one client succeeded)
- submissions == 0, err != nil → `return errMultiClient(...)` (all clients errored)
- submissions == 0, err == nil → `return nil` (e.g. `gc.clients` is empty)

In the third case `SubmitAttestations` sees `nil` and calls `rememberAttestedDataRoots`, populating the cache with a root that was never actually submitted. A downstream `SubmitAggregateSelectionProof` would then request an aggregate for a root no client attested with, reproducing the very 404 this PR fixes.

This path currently requires `gc.clients` to be empty (which shouldn't happen in production), but it's a latent invariant violation. A comment on `multiClientSubmit` stating it only returns `nil` when at least one submission succeeded, or an explicit `submissions.Load() == 0` guard at the call site before the cache write, would make the contract clear.

ovidiu-ssv-labs (dismissed):

ovidiu-ssv-labs (approved):

momosh-ssv (approved):

PR #2892, opened 2026-06-15
fix(exporter): populate dutyStore in all modes — stop blackholing exit/proposer/sync-contribution messages

### LINKED ISSUE #2886: Exporter doesn't relay or record voluntary-exit messages (both modes); standard mode also drops proposer & sync-contribution messages
## Summary

The Exporter is expected to act as a relay hub: it subscribes to all 128 subnets and is an attractive mesh peer. However, message validation (shared by all node types) consults `dutyStore` for several rules, and on exporters parts of that store are never populated. Messages failing those rules are **Ignored** — and in gossipsub an ignored message is **not forwarded**, so the exporter silently blackholes entire message classes instead of relaying them:

| Message class | Operator | Exporter `standard` (default) | Exporter `archive` |
|---|---|---|---|
| Voluntary exit | ✅ relayed | ❌ **always ignored** | ❌ **always ignored** |
| Proposer (RANDAO/consensus/post-consensus) | ✅ | ❌ ignored (except first-slot-of-epoch RANDAO) | ✅ |
| Sync-committee contribution | ✅ | ❌ always ignored | ✅ |
| Committee (attestations) | ✅ | ⚠️ over-capped for clusters with <16 validators that have a sync-committee member | ✅ |
| Aggregator / validator registration | ✅ | ✅ | ✅ |

Since ignored messages also never reach the message router, the exporter's **own recorded data** is missing for the same classes (no proposer / sync-contribution participations in standard mode, no exit traces in any mode).

## Why ignoring = blackholing

- Only `ValidationAccept` causes gossipsub to forward a message to mesh peers / offer it via IHAVE; `Ignore` and `Reject` both drop it. All "I don't have data confirming this duty" errors are Ignore ([errors.go#L74-L98](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/errors.go#L74-L98)).
- The seen-cache (TTL ≈ 6.4 min, [pubsub.go#L136](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/network/topics/pubsub.go#L136)) is marked **before** validation, so copies of the same message arriving later from other peers are not re-validated — one Ignore is permanent per message.
- Flood-publish is not enabled, so the origin operator publishes only to its mesh (D≈8); propagation genuinely depends on intermediate nodes re-forwarding.

## Root cause

Message validation is constructed identically for all modes, sharing one `dutystore.New()` ([cli/operator/node.go#L293-L307](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/cli/operator/node.go#L293-L307)). Three validation rules depend on it:

1. **Voluntary exit duty count** — limit is `dutyStore.VoluntaryExit.GetDutyCount(slot, pk)` ([common_checks.go#L85-L91](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/common_checks.go#L85-L91)); with an empty store the limit is 0 and every exit message fails with `too many duties per epoch`.
2. **Proposer / sync-contribution duty existence** — `ErrNoDuty` when the duty isn't in the store ([common_checks.go#L121-L159](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/common_checks.go#L121-L159)), checked on both the partial-sig and consensus paths.
3. **Committee duty limit boost** — the sync-committee lookup that raises the per-epoch cap to 32 ([common_checks.go#L96-L114](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/common_checks.go#L96-L114)) finds nothing, so small clusters in a sync committee get over-capped at `2×V` duty-slots per epoch.

Who populates the store, per mode:

- **Standard mode skips the duty scheduler entirely** (`shouldRunDutyScheduler`, [operator/node.go#L81-L83](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/node.go#L81-L83), introduced in #2711) → nothing is ever populated → all three rules above misfire.
- **Archive mode runs a fetch-only scheduler** (all-shares provider, prefetching beacon, no-op executor — [operator/node.go#L107-L136](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/node.go#L107-L136)), primarily because the duty tracer needs scheduled-duty info. That incidentally fixes rules 2 and 3, **but not voluntary exits**: `VoluntaryExitHandler` is excluded under "These handlers only execute duties and are not needed in exporter mode" ([scheduler.go#L168-L182](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/duties/scheduler.go#L168-L182)). The comment is wrong for this handler — besides executing own-validator exits it is also the only component that populates `dutyStore.VoluntaryExit` ([voluntary_exit.go#L139-L186](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/duties/voluntary_exit.go#L139-L186)) consumed by validation on **all** node types.

As a result, in both exporter modes nobody consumes `validatorExitCh`, and every `ValidatorExited` contract event ends with a 2-slot send timeout and an error log ([task_executor.go#L157-L181](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/validator/task_executor.go#L157-L181)).

## How to confirm on a live exporter

- Log `failed to schedule voluntary exit duty!` shortly after each `ValidatorExited` event (both modes); `exporter standard mode: skipping duty scheduler` at startup (standard).
- Metric `ssv.p2p.message.validations.ignored` with `discard_reason="too many duties per epoch"` for the voluntary-exit role (and, in standard mode, `discard_reason="no duty for this epoch"` pegged at ~100% for proposer / sync-contribution roles).

## Related fragilities (same rule family, all node types)

- The exit dutyCount rule ties gossip relaying to local EL event sync. #2851 / #2857 fixed the operator-side broadcast-timing race, but any node lagging more than `FollowDistance + slack` still silently drops in-flight exits.
- `dutyStore` is in-memory while event-sync progress is persisted: a node restarting between the `ValidatorExited` event and the exit broadcast never re-registers the duty and permanently ignores that exit.
- Exit messages for a **liquidated** validator are ignored network-wide (liveness checks in [validation.go#L297-L318](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/validation.go#L297-L318)) — "exit because I'm being liquidated" is a plausible user flow.
- A second exit for the same validator within one epoch is dropped: a per-epoch observed count is compared against a per-slot allowed count ([common_checks.go#L55-L83](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/message/validation/common_checks.go#L55-L83) vs [dutystore/voluntary_exit.go#L20-L30](https://github.com/ssvlabs/ssv/blob/dfc54bfcd671ebf3707496b1dc7dfa58e8b1876c/operator/duties/dutystore/voluntary_exit.go#L20-L30)).

Historical note: before #1468 the exit limit was a flat 2 per epoch (like validator registration); the rule comments in both validation files still document that behavior while the code now does the strict per-slot lookup.

## Proposed fix (layered)

1. **Populate `dutyStore.VoluntaryExit` in the event path, not the scheduler.** The exit "duty" is authorized by an EL event that every node type already processes; compute `dutySlot = blockSlot + 4` where the event/task is handled and call `AddDuty` synchronously there. Keep `VoluntaryExitHandler` purely as the own-validator execution queue (the descriptor can carry the precomputed slot, removing the duplicate `HeaderByNumber` lookup). Fixes exit relaying/recording on operators and both exporter modes, and removes the 2-slot channel-send timeout that can currently drop an operator's **own** exit execution under load. Pruning of old slots needs a new home (prune-on-write window or a small janitor).
2. **Standard mode: run the validation-relevant duty handlers.** Validation only consults `dutyStore.Proposer` and `dutyStore.SyncCommittee` (the attester store is used for execution/tracing, not validation), so a slim fetch-only scheduler with just the Proposer + SyncCommittee handlers restores correct relaying and standard-mode data completeness while still avoiding the expensive attester/committees prefetch that motivated #2711. (Blunt alternative: give standard mode the same fetch-only scheduler as archive.)
3. **Restart resilience:** re-scan recent `ValidatorExited` events (last ~2 epochs of blocks) on startup to repopulate the in-memory store, or persist the small exit-duty map.
4. **Spec-level discussion:** revisit the strict existence check from #1468. Exit messages already require valid RSA signatures from the validator's committee operators, so the marginal anti-spam value over a flat per-validator cap (as registration uses) is small, while the strict check created this whole class of liveness coupling (exporter gaps, restart gap, EL-lag races). Relaxing it would make exit propagation independent of every node's local EL view; needs coordinated rollout.
5. **Cleanups:** fix the misleading "only execute duties" comment in `scheduler.go`; update the stale "2 for … voluntary exit" rule comments in `partial_validation.go` / `consensus_validation.go`; remove or consume the dead `ValidatorExitCh` / `ValidatorRegistrationCh` wiring in exporter-mode `SchedulerOptions`.

## References

- #2711 (standard mode stopped running the duty scheduler)
- #1468 (introduced the strict exit-duty existence check)
- #2851, #2857, #2858 (recent operator-side fixes to the same rule family for exits/registrations)

All code links are pinned to the current `stage` head at the time of writing.

Roy-blox:
## Problem

Closes #2886.

Three message classes were silently dropped (gossipsub `Ignored`) by all exporters:

| Message class | Operator | Exporter standard | Exporter archive |
|---|---|---|---|
| Voluntary exit | ✅ | ❌ blackholed | ❌ blackholed |
| Proposer (RANDAO/consensus/post-consensus) | ✅ | ❌ blackholed | ✅ |
| Sync-committee contribution | ✅ | ❌ blackholed | ✅ |

`Ignored` in gossipsub means the message is **not forwarded to mesh peers** and is permanently cached as seen, so it can never be re-validated. The exporter became a propagation sink for these message classes.

### Root causes

1. **Standard-mode exporter**: `shouldRunDutyScheduler` returned `false` (#2711) — no scheduler ran, `dutyStore.{Proposer,SyncCommittee,VoluntaryExit}` stayed empty → `ErrNoDuty` / `ErrTooManyDutiesPerEpoch` on every incoming message of these types.

2. **Archive-mode exporter**: Scheduler ran but `VoluntaryExitHandler` was excluded under the comment _"only executes duties"_ — missed that it also populates `dutyStore.VoluntaryExit` which message validation reads for the dutyCount check. Every `ValidatorExited` EL event hit the 2-slot channel-send timeout logged as `"failed to schedule voluntary exit duty!"`.

## Fix

**Two targeted changes, 3 files.**

### `operator/node.go` — always run the duty scheduler

`shouldRunDutyScheduler` now returns `true` unconditionally. Standard-mode exporters get the same AllShares provider, PrefetchingBeacon, and NoopExecutor already used by archive mode, so no duty is ever executed on an exporter.

### `operator/duties/scheduler.go` — restructure handler registration

| Handler | Before | After |
|---|---|---|
| Proposer | operator + exporter | operator + exporter ✓ |
| SyncCommittee | operator + exporter | operator + exporter ✓ |
| **VoluntaryExit** | **operator only** | **all modes** ✓ |
| Attester | operator + exporter | operator + archive-exporter |
| Committee | operator only | operator only ✓ |
| ValidatorRegistration | operator only | operator only ✓ |

Attester is dropped from standard-mode exporter: the attester store is not consulted by message validation, and fetching duties for all network validators is expensive without the duty-tracing benefit that motivates it in archive mode.

As a side effect, `VoluntaryExitHandler` now consumes `validatorExitCh` in both exporter modes, eliminating the 2-slot channel-send timeout error.

## Testing

- `go build ./operator/...` — clean
- `go vet ./operator/... ./operator/duties/...` — clean  
- `TestShouldRunDutyScheduler` updated and passing (standard mode now returns `true`)
- All existing tests pass under `go test ./operator/... ./operator/duties/...`

## Verification on a live exporter

After deploying, the following should disappear from exporter logs:
- `"failed to schedule voluntary exit duty!"` (both modes)
- Metric `ssv_p2p_message_validations_ignored_total` with `discard_reason="no duty for this epoch"` pegged at ~100% for proposer / sync-contribution roles (standard mode)
- Metric `ssv_p2p_message_validations_ignored_total` with `discard_reason="too many duties per epoch"` for voluntary-exit role (both modes)


Roy-blox:
## Test evidence & verification

### Unit test added (commit `bafbcea4`)

`TestNewScheduler_HandlerRegistration` — table-driven test covering all three scheduler modes. Confirms that:

| Mode | Handlers registered |
|------|-------------------|
| Operator | PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT, ATTESTER, CLUSTER, VALIDATOR_REGISTRATION |
| Standard exporter | PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT |
| Archive exporter | PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT, ATTESTER |

All 112 existing `operator/duties/...` tests still pass.

---

### Live stage evidence (pre-fix)

Queried `ssv_p2p_message_validations_ignored_total` across all environments right now:

| Container | Role | Reason | Drops |
|-----------|------|--------|-------|
| `hoodi-ssv-node-exporter-ovh` | VOLUNTARY_EXIT | too many duties per epoch | **4,724** |
| `hoodi-ssv-node-exporter-ovh-2` | VOLUNTARY_EXIT | too many duties per epoch | **4,724** |
| `mainnet-ssv-node-exporter-1-ovh` | VOLUNTARY_EXIT | too many duties per epoch | 1,172 |
| `mainnet-ssv-node-exporter-2-ovh` | VOLUNTARY_EXIT | too many duties per epoch | 1,172 |
| `ssv-node-exporter` | VOLUNTARY_EXIT | too many duties per epoch | 1,044 |
| `mainnet-ssv-node-exporter-1-ovh` | PROPOSER | no duty for this epoch | 377 |
| `mainnet-ssv-node-exporter-2-ovh` | PROPOSER | no duty for this epoch | 377 |
| `ssv-node-exporter` | PROPOSER | no duty for this epoch | 18 |

Bug confirmed on every exporter node, all environments (hoodi / mainnet / sepolia).

---

### Why I'm >98% confident the fix is correct

The failure mode is deterministic:

1. Exporter receives a VOLUNTARY_EXIT gossip message
2. `message/validation/common_checks.go` calls `dutyStore.VoluntaryExit.GetDutyCount(slot, pk)`
3. Without `VoluntaryExitHandler`, the store is **never populated** → count = 0
4. The check fails with `ErrTooManyDutiesPerEpoch` (0 allowed, 1 seen)
5. GossipSub sees `ValidationIgnore` → **permanently blackholes the message** (cached as seen, never retried)

The fix adds `VoluntaryExitHandler` (and `ProposerHandler` / `SyncCommitteeHandler` which already ran but now always run) to the standard exporter. These handlers call `AddDuty()` when they fetch duties from the beacon, which populates the stores so validation passes.

The only way this could fail to fix the issue is if the beacon never sends VoluntaryExit/Proposer/SyncCommittee duties to the exporter — but those are fetched from the Ethereum beacon chain, same as for operator nodes. The fix is mechanically equivalent to "start running the handlers that were previously skipped."

After deploy: `ssv_p2p_message_validations_ignored_total{discard_reason="too many duties per epoch", role="VOLUNTARY_EXIT"}` counter should stop growing on all exporter containers.

Roy-blox:
## Live deployment test — final evidence

Two nodes deployed to stage-hoodi for 15 minutes, identical config except branch:

| Node | Branch | Config |
|------|--------|--------|
| ssv-node-97/98 | `fix/exporter-duty-store-blackhole` | `EXPORTER=true, EXPORTER_MODE=standard` |
| ssv-node-99/100 | `stage` (unfixed) | `EXPORTER=true, EXPORTER_MODE=standard` |

The control nodes (99/100) logged the dead branch explicitly:
```
exporter standard mode: skipping duty scheduler  name=Operator
```

The fix nodes (97/98) logged:
```
starting duty scheduler  name=DutyScheduler
fetching duties for the current epoch  handler=PROPOSER
```

### Drop rates after 15 minutes

| Node | Role | Reason | Rate |
|------|------|--------|------|
| ssv-node-99 (stage) | SYNC_COMMITTEE_CONTRIBUTION | **no duty for this epoch** | **10.8/s** |
| ssv-node-100 (stage) | SYNC_COMMITTEE_CONTRIBUTION | **no duty for this epoch** | **9.3/s** |
| ssv-node-99 (stage) | PROPOSER | **no duty for this epoch** | **0.033/s** |
| ssv-node-100 (stage) | PROPOSER | **no duty for this epoch** | **0.033/s** |
| ssv-node-97 (fix) | SYNC_COMMITTEE_CONTRIBUTION | no duty for this epoch | **0/s** |
| ssv-node-98 (fix) | SYNC_COMMITTEE_CONTRIBUTION | no duty for this epoch | **0/s** |
| ssv-node-97 (fix) | PROPOSER | no duty for this epoch | **0/s** |
| ssv-node-98 (fix) | PROPOSER | no duty for this epoch | **0/s** |

`AGGREGATOR` and `COMMITTEE` "too few signers" drops appear on **all** nodes at similar rates — those are unrelated to this fix (different validation path, not a duty-store check).

VOLUNTARY_EXIT: no drops on either set during this window (no validator exits on hoodi testnet in the last ~30 min). The mechanism is identical to SYNC_COMMITTEE_CONTRIBUTION — the duty store is populated by the handler or it's empty — so the same fix covers it. The live exporter totals (4724 drops on `hoodi-ssv-node-exporter-ovh`) stand as historical evidence.

**Confidence: >99%.** The fix eliminates the duty-store drops entirely and the control group demonstrates the bug is real and active.

Roy-blox:
## QA Testing

### What was tested

Standard-mode exporter nodes — before the fix the duty scheduler was skipped entirely, leaving `dutyStore.Proposer`, `dutyStore.SyncCommittee`, and `dutyStore.VoluntaryExit` permanently empty. Message validation consults those stores for every incoming gossip message; empty stores cause `ValidationIgnore` (GossipSub permanently blackholes the message — never forwarded, never retried).

### How to reproduce

Deploy two nodes to stage-hoodi with `EXPORTER=true EXPORTER_MODE=standard`:
- one from `stage` (unfixed)
- one from this branch (fixed)

Wait ~15 minutes, then compare:

```promql
rate(ssv_p2p_message_validations_ignored_total{
  container=~"ssv-node-99|ssv-node-100",
  ssv_p2p_discard_reason="no duty for this epoch"
}[10m])

rate(ssv_p2p_message_validations_ignored_total{
  container=~"ssv-node-97|ssv-node-98",
  ssv_p2p_discard_reason="no duty for this epoch"
}[10m])
```

**Expected result (unfixed):** `SYNC_COMMITTEE_CONTRIBUTION` drops at ~10/s, `PROPOSER` drops at ~0.033/s.
**Expected result (fixed):** both rates = 0.

Alternatively, check the startup logs:
- Unfixed: `exporter standard mode: skipping duty scheduler`
- Fixed: `starting duty scheduler` followed by `fetching duties for the current epoch handler=PROPOSER`

### What was observed on 2026-06-15

| Node | Branch | SYNC_COMMITTEE_CONTRIBUTION "no duty" | PROPOSER "no duty" |
|------|--------|--------------------------------------|-------------------|
| ssv-node-97 | this branch | **0/s** | **0/s** |
| ssv-node-98 | this branch | **0/s** | **0/s** |
| ssv-node-99 | stage (unfixed) | 10.8/s | 0.033/s |
| ssv-node-100 | stage (unfixed) | 9.3/s | 0.033/s |

VOLUNTARY_EXIT was not observable in this window (no validator exits on hoodi during the test period). Historical evidence: `hoodi-ssv-node-exporter-ovh` has 4,724 accumulated `VOLUNTARY_EXIT` drops with `too many duties per epoch` on the current unfixed exporter.

### Unit test

`TestNewScheduler_HandlerRegistration` in `operator/duties/scheduler_test.go` — asserts the exact handler set for each mode without needing a running node:

```
operator:          PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT, ATTESTER, CLUSTER, VALIDATOR_REGISTRATION
standard exporter: PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT
archive exporter:  PROPOSER, SYNC_COMMITTEE, VOLUNTARY_EXIT, ATTESTER
```

operator/duties/scheduler.go
```diff
@@ -165,19 +169,30 @@ func NewScheduler(logger *zap.Logger, opts *SchedulerOptions) *Scheduler {
...
-	// When adding a new handler here, ensure it supports both modes.
+	// Proposer, SyncCommittee, and VoluntaryExit are needed in all modes.
+	// Proposer and SyncCommittee populate dutyStore entries consulted by message validation on all
+	// node types. VoluntaryExit populates the per-(slot,pk) duty count also consulted by validation;
+	// in exporter mode all exit descriptors have OwnValidator=false so no exit is ever executed.
 	s.dutyHandlers = append(s.dutyHandlers,
-		NewAttesterHandler(dutyStore.Attester, opts.ExporterMode),
 		NewProposerHandler(dutyStore.Proposer, opts.ExporterMode),
 		NewSyncCommitteeHandler(dutyStore.SyncCommittee, opts.ExporterMode),
+		NewVoluntaryExitHandler(dutyStore.VoluntaryExit, opts.ValidatorExitCh),
```
momosh-ssv: Might be worth adding a symmetric guard for VoluntaryExit. 

Proposer and SyncCommittee both short-circuit with `if h.exporterMode { return }` in `processExecution`, but `VoluntaryExitHandler` has no such in-handler guard now that it runs in exporter mode — its exporter safety rests entirely on `OwnValidator=false` (empty `OperatorData` ⇒ operatorID 0) plus the `NoopExecutor`. 

That holds today, but the contract is implicit and untested, so a future change to how `OwnValidator` is set, or to the executor wiring, could silently let an exporter sign an exit. 

A cheap exporter-mode guard in `processExecution`, or a test asserting an exporter never queues an `OwnValidator=true` exit, would lock it down.

operator/duties/scheduler_test.go
```diff
@@ -693,3 +693,61 @@ func TestScheduler_HandleHeadEvent_DoesNotBlockWithoutReorgConsumer(t *testing.T
 		PreviousDutyDependentRootChanged: true,
 	}}, drainReorgEvents(s.reorgCh))
 }
+
+func TestNewScheduler_HandlerRegistration(t *testing.T) {
```
momosh-ssv, line 697: What do you think about extending this to cover the safety invariant, not just registration? 

It asserts the handler set per mode, which is great, but nothing about `exporterMode` propagation into the handlers or the `NoopExecutor` wiring — the test would still pass if someone wired a real executor into an exporter. 

A focused assertion at the node/scheduler boundary (exporter mode ⇒ no-op executor, or handlers carry `exporterMode=true`) would guard the property that actually keeps an exporter from signing.

operator/duties/scheduler.go
Roy-blox: Good call, added it.  now has an  field and  guards before calling .

operator/duties/scheduler_test.go
Roy-blox: Extended the test to assert exporterMode propagates into each handler type and that Scheduler.exporterMode is set correctly.

operator/node.go
```diff
@@ -103,38 +99,36 @@ func New(logger *zap.Logger, opts Options, exporterOpts exporter.Options, slotTi
...
-			DutyExecutor:            dutyExecutor,
-			IndicesChgCh:            opts.ValidatorController.IndicesChangeChan(),
-			ValidatorRegistrationCh: opts.ValidatorController.ValidatorRegistrationChan(),
-			ValidatorExitCh:         opts.ValidatorController.ValidatorExitChan(),
-			DutyStore:               opts.DutyStore,
-			SlotTickerProvider:      slotTickerProvider,
-			P2PNetwork:              opts.P2PNetwork,
-			ExporterMode:            exporterOpts.Enabled,
-		})
+	if exporterOpts.Enabled {
```
iurii-ssv, line 108: Non-blocking (behavior change + docs).

This is the core fix, and it's correct — but note it also means standard-mode exporters run the duty scheduler for the **first time**. Beyond unblocking forwarding, they'll now fetch proposer (per-epoch) and sync-committee (per-period, all indices) duties via the `AllShares` provider + `PrefetchingBeacon` — the same path archive mode already uses, minus the attester handler. The added beacon load is bounded and proven (archive mode does strictly more), but it's a behavioral change for standard-exporter operators.

`docs/release-notes_exporter_v2.0.0-rc.1.md` currently states standard mode "keeps the legacy behavior". Consider a short note (there, or wherever exporter behavior is tracked) that standard exporters now schedule duties and forward voluntary-exit / proposer / sync-contribution messages instead of dropping them.

iurii-ssv (commented):
Good set of changes, LGTM overall.

The only thing that's off/misleading is: PR description says "Closes https://github.com/ssvlabs/ssv/issues/2886", while #2886 is in fact a much bigger design discussion/issue overall - this PR takes more of a symptomatic approach that is still nice to have short-term ... but I'd keep #2886 open (classify this PR as "part of https://github.com/ssvlabs/ssv/issues/2886"") as we might run into it in the future as well.

**How this PR maps to #2886**

This PR resolves the *relaying/recording symptoms*, but it implements a subset of the layered plan in #2886. Suggest changing `Closes #2886` → `Part of #2886` (or filing follow-ups) so the remaining items don't get auto-closed.

✅ **Addressed**
- Standard-mode proposer & sync-contribution blackholing
- Voluntary-exit blackholing in **both** exporter modes
- Committee over-cap for small clusters with a sync-committee member — *incidentally* fixed by populating `dutyStore.SyncCommittee` (not named or tested here)
- `ValidatorExitCh` is now consumed → removes the 2-slot send-timeout symptom
- Cleanup: the misleading "only execute duties" comment

⏳ **Left for #2886 (not in this PR)**
- **Part 1** — populate `dutyStore.VoluntaryExit` in the EL event path / make the handler execution-only. This PR instead keeps `AddDuty` in the handler and just runs it in all modes, so the channel coupling + per-event `HeaderByNumber` lookup remain (residual burst-timeout risk).
- **Part 3** — restart resilience: the store is in-memory, so an exit observed just before a restart is still permanently ignored.
- **Part 4** — spec-level relax of the strict #1468 existence check.
- **Related fragilities** — EL-lag drop (> `FollowDistance`+slack), liquidated-validator exits ignored network-wide, 2nd exit/epoch dropped.
- **Cleanups** — stale "2 for voluntary exit" comments in `partial_validation.go` / `consensus_validation.go`; dead `ValidatorRegistrationCh` still passed-but-unconsumed in exporter mode.

PR #2895, opened 2026-06-16
feat(exporter): remove standard mode, add optional duty-trace retention

y0sher:
## Summary

Follow-up to #2892. Removes the exporter's **standard** mode — exporters now always run **full duty tracing** (the former *archive* behavior) — and adds an optional, default-off retention window for the on-disk duty-trace store.

> Stacked on #2892 (`fix/exporter-duty-store-blackhole`) so the archive voluntary-exit fix is included. Base will retarget to `stage` once #2892 merges.

## Why

`EXPORTER_MODE` defaulted to `standard`, yet standard was a strict — and historically broken — subset of archive:
- It served only the participant-store-backed `/v1/exporter/decideds` endpoint (no trace endpoints).
- It was the source of the gossip-message blackhole in #2886 (proposer/sync/exit messages dropped because no duty scheduler ran).

Every deployed exporter (prod mainnet, stage mainnet, all hoodi exporters) already runs `archive`. Verified live: the production mainnet exporter handles all duty classes archive does; archive is a **functional superset** of standard (same `decideds` endpoint + the trace endpoints). So removing standard deletes a real footgun (a default that silently blackholes) without losing any capability.

## What changed

- **`exporter.Options`**: drop `Mode` / `ModeStandard` / `ModeArchive`. Replace the standard-only `RetainSlots` with **`RetainEpochs`** (`EXPORTER_RETAIN_EPOCHS`, default `0` = retain indefinitely — matches today's archive behavior).
- **`cli/operator`**: node-mode enum collapses to `operator` / `exporter`; `resolveMode` can no longer fail; the duty-trace collector + read API are always wired for exporters; the standard-only participant-store pruning is removed.
- **`duties.Scheduler`**: drop the `ArchiveMode` option — the Attester handler is registered for every exporter (full tracing), unchanged for operators.
- **`validator.Controller`**: route non-committee messages to the trace collector whenever one is wired (exporter), else the legacy participant path (operator) — nil-safe via `traceCollector != nil`.
- **`dutytracer`**: new `DutyTraceStore.PruneSlot(slot)` (single prefix-drop per key group) and a retention loop in `Collector.Start` that prunes trace data older than `RetainEpochs`. Pruning is capped per tick and **seeds forward on first run** — enabling retention does not retroactively sweep pre-existing history.

## Compatibility

- A stale `EXPORTER_MODE` / `RetainSlots` in existing configs is simply ignored (unknown fields), so nodes keep booting.
- The effective default flips from the (broken) `standard` to full tracing — which is what the fleet already runs.
- Out of scope (deliberately): the operator/legacy participant-store path and its `/v1/exporter/decideds` handler are untouched; only the **exporter** stops using standard mode.

## Tests

- Updated: handler-registration (`operator/duties`), mode resolution + exporter wiring (`cli/operator`).
- Added: `TestPruneSlot` (store removes all trace kinds for a slot, leaves adjacent slots intact) and `TestCollector_pruneExpired` (retention cursor: seed-forward, per-tick advance, per-tick cap).
- `go build ./...`, `go vet`, and affected unit suites pass.

## Verification status

Unit-tested and built clean. Live deploy verification on a stage exporter is **pending** (vnet tunnel was down at PR-creation time) — will confirm boot (full tracing), trace endpoints, health, and that the now-ignored `EXPORTER_MODE` in the deploy template still boots cleanly.

### DIFF
--- a/cli/operator/config.go
+++ b/cli/operator/config.go
@@ -68,9 +68,8 @@ const maxSafeProposerDelay = 1000 * time.Millisecond
 type nodeMode int
 
 const (
-	modeOperator         nodeMode = iota // not an exporter
-	modeExporterStandard                 // exporter, standard tracing
-	modeExporterArchive                  // exporter, archive tracing (pre-consensus + consensus)
+	modeOperator nodeMode = iota // not an exporter
+	modeExporter                 // exporter (full duty tracing: pre-consensus + consensus + post-consensus)
 )
 
 // resolved carries config-derived state computed by resolveAndValidate (not operator-provided):
@@ -136,11 +135,7 @@ func (c *config) resolveAndValidate(logger *zap.Logger) (resolved, error) {
 
 	// Resolve the operating mode last so a doubly-misconfigured node still surfaces the signing
 	// or proposer-delay error first.
-	m, err := resolveMode(c.ExporterOptions)
-	if err != nil {
-		return resolved{}, err
-	}
-	res.mode = m
+	res.mode = resolveMode(c.ExporterOptions)
 
 	return res, nil
 }
@@ -218,21 +213,14 @@ func (c *config) resolveSigning() (resolved, error) {
 	return res, nil
 }
 
-// resolveMode derives the node's operating mode from the exporter options, rejecting an
-// unrecognized EXPORTER_MODE up front (fail-fast). A non-exporter node is always modeOperator,
-// regardless of the (then-irrelevant) EXPORTER_MODE.
-func resolveMode(opts exporter.Options) (nodeMode, error) {
+// resolveMode derives the node's operating mode from the exporter options. An enabled exporter
+// always runs in full duty-tracing mode; a non-exporter node is modeOperator. (The legacy
+// standard/archive EXPORTER_MODE distinction was removed — exporters are always full tracers.)
+func resolveMode(opts exporter.Options) nodeMode {
 	if !opts.Enabled {
-		return modeOperator, nil
+		return modeOperator
 	}
-	switch opts.Mode {
-	case exporter.ModeStandard:
-		return modeExporterStandard, nil
-	case exporter.ModeArchive:
-		return modeExporterArchive, nil
-	default:
-		return modeOperator, fmt.Errorf("invalid exporter mode %q (must be %q or %q)", opts.Mode, exporter.ModeStandard, exporter.ModeArchive)
-	}
+	return modeExporter
 }
 
 // warnExporterSigning warns when signing configuration is provided in exporter mode (where it
--- a/cli/operator/config.go
+++ b/cli/operator/config.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/ilyakaznacheev/cleanenv"
@@ -64,7 +65,7 @@ const maxSafeProposerDelay = 1000 * time.Millisecond
 
 // nodeMode is the resolved operating mode of the node, derived once from ExporterOptions by
 // resolveAndValidate so startup can dispatch on a typed value instead of re-deriving the mode
-// from ExporterOptions.Enabled / .Mode at each site.
+// from ExporterOptions.Enabled at each site.
 type nodeMode int
 
 const (
@@ -136,6 +137,9 @@ func (c *config) resolveAndValidate(logger *zap.Logger) (resolved, error) {
 	// Resolve the operating mode last so a doubly-misconfigured node still surfaces the signing
 	// or proposer-delay error first.
 	res.mode = resolveMode(c.ExporterOptions)
+	if res.mode != modeOperator {
+		warnDeprecatedExporterEnv(logger)
+	}
 
 	return res, nil
 }
@@ -221,6 +225,19 @@ func resolveMode(opts exporter.Options) nodeMode {
 		return modeOperator
 	}
 	return modeExporter
+}
+
+// warnDeprecatedExporterEnv flags removed exporter env vars (set via a pre-existing deployment) so
+// operators notice a stale value is now ignored: exporters always run full duty tracing, and
+// retention moved from slots to EXPORTER_RETAIN_EPOCHS. Best-effort — only env vars are checked,
+// not equivalent YAML keys (cleanenv silently drops unknown fields).
+func warnDeprecatedExporterEnv(logger *zap.Logger) {
+	for _, env := range []string{"EXPORTER_MODE", "EXPORTER_RETAIN_SLOTS"} {
+		if v, ok := os.LookupEnv(env); ok {
+			logger.Warn("ignoring removed exporter config option; exporters always run full duty tracing — use EXPORTER_RETAIN_EPOCHS for retention",
+				zap.String("env", env), zap.String("value", v))
+		}
+	}
 }
 
 // warnExporterSigning warns when signing configuration is provided in exporter mode (where it
--- a/operator/dutytracer/collector.go
+++ b/operator/dutytracer/collector.go
@@ -127,9 +127,7 @@ func (c *Collector) Start(ctx context.Context, tickerProvider slotticker.Provide
 	// Start a single worker to process schedule writes asynchronously.
 	go c.runScheduleWorker(ctx)
 
-	// lastPrunedSlot tracks retention progress; seeded lazily on the first eligible tick so we
-	// enforce retention going forward without an expensive backfill of pre-existing history.
-	var lastPrunedSlot phase0.Slot
+	pruner := &tracePruner{store: c.store, logger: c.logger}
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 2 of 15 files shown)


y0sher:
## Live verification — stage mainnet exporter (node 1002)

Deployed `feat/exporter-remove-standard-add-retention` (commit `40d2563`) to the stage mainnet exporter (`ssv-node-exporter-mainnet-1`, ~52K mainnet validators) via the `bot-stage-ssv-exporter-mainnet-template`, which still sets `EXPORTER_MODE=archive` — now an ignored field.

| Check | Result |
|---|---|
| Boots on new commit | ✅ `starting SSV-Node:untagged-40d2563…` |
| Full-tracing config | ✅ `exporter enabled (full duty tracing) retain_epochs=0 retain_slots=0` |
| Back-compat (stale `EXPORTER_MODE=archive` ignored) | ✅ boots cleanly, no error |
| Handler set | ✅ `PROPOSER`, `SYNC_COMMITTEE`, `VOLUNTARY_EXIT`, `ATTESTER` (one each) |
| No legacy mode log / no `skipping duty scheduler` | ✅ absent |
| Duty tracer collecting | ✅ `name=dutytracer … quorum reached after flush beacon_role=ATTESTER signers_count=3` |
| panic / FATAL | ✅ none |
| Peers / health | ✅ connected, warming up (fresh DB) |

`dropping buffered pending signatures during eviction` appears during the first minutes — a dutytracer warm-up artifact (evicting slots whose lifecycle started before the node did; unchanged eviction code). It is absent on the long-running prod archive exporter. Expected to subside once the node is past warm-up; steady-state resource use should match the archive exporter it replaces (the workload is identical minus the removed dead code).

Retention (`RetainEpochs`/pruning) is covered by unit tests (`TestPruneSlot`, `TestCollector_pruneExpired`); default `0` = retain indefinitely was confirmed live.

y0sher:
## Update — retention refactor + steady-state verification (commit `69cd7d8`)

Retention logic was refactored into a self-contained, fully-tested `tracePruner`. Structure is leaner, but the production safeguards are intentionally **kept** (instrumented + edge-case-guarded, not trimmed for brevity):

- underflow guard (`currentSlot <= retainSlots`) — no unsigned slot wraparound early in chain life;
- seed-forward on first activation — enabling retention on a live chain (slot ~14.5M) does not backfill millions of historical slots;
- per-tick cap (`maxPrunePerTick`) — a large boundary jump after a pause can't stall the eviction loop;
- cursor catch-up — a coalesced/late tick reclaims skipped slots instead of leaking them;
- debug log on prune + per-op duration metric via `DutyTraceStoreMetrics`.

`TestTracePruner` covers: disabled, underflow, seed-forward (no backfill), steady-state single-slot prune, and capped catch-up. The in-memory-cursor limitation (slots expiring during downtime aren't reclaimed after restart) is documented in-code as an accepted bound for best-effort disk retention.

### Live re-verification — stage mainnet exporter (node 1002), ~52K mainnet validators

| Check | Result |
|---|---|
| Boots on `69cd7d8` | ✅ |
| Full-tracing config | ✅ `exporter enabled (full duty tracing) retain_epochs=0 retain_slots=0` |
| Handler set | ✅ PROPOSER + SYNC_COMMITTEE + VOLUNTARY_EXIT + ATTESTER |
| **Steady-state RSS (52 min uptime)** | ✅ **5.80 GB** — in line with the archive exporter it replaces (prod 6.44 GB, stage 6.0–6.3 GB) |
| Peers | ✅ 510 |
| CL/event sync | ✅ 100% |
| Warm-up `dropping buffered` noise | ✅ cleared (0 in last 5m) |
| Non-benign ERROR logs (last 5m) | ✅ 0 |

Steady-state resource use confirms the expectation: removing standard mode and making the exporter a full tracer costs the same as the archive mode already deployed fleet-wide — the workload is identical minus the deleted dead code.

y0sher:
## Independent Codex review + fixes (commit `0319747`)

Ran an independent adversarial review (Codex / gpt-5.5). Verdict was "needs-work, no blockers." Findings and how they were addressed:

**Fixed**
- **Pruned-slot resurrection (correctness):** after `PruneSlot` deleted a slot, a late message took the disk-miss path in `getOrCreate{Validator,Committee}Trace` and re-saved an empty trace — resurrecting a pruned slot. Reachable when `RetainEpochs` is small enough that the floor lands inside the message-validation window. Fixed with a single retained-slot **floor** (`retentionState`, owned by `Collector`): the collection + schedule paths consult `expired()` and drop via a new `errExpiredSlot` sentinel (`Collect` drops it silently). This also folds the old `tracePruner` cursor into one source of truth (Codex's suggested simplification).
- **Stale config silently ignored:** startup now logs a deprecation warning when removed `EXPORTER_MODE` / `EXPORTER_RETAIN_SLOTS` env vars are set. *(Verified live — the deploy template still sets `EXPORTER_MODE=archive`, now flagged and ignored.)*
- **Honest retention semantics:** `EXPORTER_RETAIN_EPOCHS` is documented as best-effort, forward-only (history aging out during downtime isn't reclaimed after restart).
- **Simplification:** `PruneSlot` drops scheduled data with a single `sd+slot` prefix instead of iterating roles.
- **Tests:** `TestRetentionState` now covers the floor/`expired` guard; new `TestCollector_dropsExpiredSlot` proves the hot path refuses expired slots; `TestPruneSlot` covers multiple validator roles. Stale comments fixed.

**Assessed, not changed**
- Codex flagged the schedule worker as a second resurrection vector; it's effectively unreachable (the filler only enqueues `current`/`current-1`, far newer than the floor), but the cheap `expired()` guard was added there too as defense-in-depth.
- True cross-restart disk bound (persisted prune cursor) deferred as a follow-up — over-building for a default-off feature; semantics now documented honestly.

**Re-verified live (node 1002, commit `0319747`):** boots in full-tracing mode, correct handler set, deprecation warning fired, no panics; steady-state resource use matches the prior commit (~5.8 GB, == archive baseline). `go build ./...`, vet, gofmt, and the affected unit suites all pass.

iurii-ssv:
@greptile pls re-review

iurii-ssv (approved):
LGTM

operator/dutytracer/collector.go
```diff
@@ -523,6 +614,9 @@ func (c *Collector) Collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 		go c.collectLateMessage(ctx, msg, verifySig)
 		return nil
 	}
+	if errors.Is(err, errExpiredSlot) {
+		return nil // message is for a slot outside the retention window; drop silently
```
momosh-ssv, line 618: Might be worth emitting a Debug log or a `dropped_expired` counter here rather than dropping silently. 

As-is, a message dropped because its slot aged out is indistinguishable from a too-tight `EXPORTER_RETAIN_EPOCHS` quietly discarding live traffic — there's no signal to tell correct retention from a misconfiguration eating real data. 

The same goes for the `errExpiredSlot` path in `collectLateMessage`.

Non-blocking btw.

momosh-ssv (approved):

operator/dutytracer/collector.go
iurii-ssv: Good call on the observability — a `dropped_expired` counter (plus a Debug line in the late path, where the `err = nil` currently suppresses the deferred error log) is cheap.

One thing on the framing though: I don't think this can actually mask a too-tight window eating *live* traffic. Retention is opt-in and off by default (`EXPORTER_RETAIN_EPOCHS=0` → `floor` stays `0`, so `expired()` is always false and nothing is ever dropped), and when it's enabled the expired-drop only fires for messages older than the *entire* window. Live traces arrive within a few slots; to drop anything recent you'd have to set the window to ~1 epoch, which is an obviously-broken value rather than a subtle misconfig.

So I'd add it for debuggability of the retention feature itself — catching a floor-computation regression, or just confirming pruning behaves — rather than as a guard against silently discarding real data, since that's already prevented by the off-by-default + whole-window design.

cli/operator/config.go
ovidiu-ssv-labs, line 85:
```suggestion
// isExporter reports whether the node runs as an exporter (always in former archive mode) rather than as an
```

ovidiu-ssv-labs: Small nitpick for consistency, very optional.

ovidiu-ssv-labs (approved):

ovidiu-ssv-labs:
**Stage-hoodi deploy test — no regression detected ✅**

Deployed PR branch (`e0ded53e`) on nodes 101-102 with `EXPORTER=true,EXPORTER_MODE=archive`, alongside a stage control on 103-104 (same env vars, without PR 2892/2895 fixes). Ran for ~20h.

**VOLUNTARY_EXIT (live exit event, validator 838944)**
- **101-102 (PR branch):** `added voluntary exit task to pipeline` — duty store populated, P2P messages accepted.
- **103-104 (control):** `failed to schedule voluntary exit duty!` + 4 partial signatures blackholed with `too many duties per epoch, want <=0` — confirms pre-fix behavior reproduced on the control side.

**Message validation drops**
No `no duty for this epoch` or `too many duties per epoch` drops on any of the 4 nodes over the full 20h window. All observed ignores are benign exporter gossip noise (`got decided message with too few signers`), symmetric across both groups.

**Duty scheduler**
All 4 nodes started the scheduler and fetched ATTESTER + PROPOSER duties. No node emitted `exporter standard mode: skipping duty scheduler`.

**Peer connectivity:** 86-88 peers per node (consistent with fleet exporter baseline of ~87-94).

PR #2897, opened 2026-06-16
fix(cli/operator): convert logger.Fatal goroutines to errgroup (#2867)

### LINKED ISSUE #2867: cli/operator: errgroup lifecycle — converge startup Fatals into a single handler

Roy-blox:
## Summary

Closes #2867.

All `logger.Fatal` calls inside background goroutines in `cli/operator` are replaced with errgroup-propagated errors. The first service failure now cancels the rest and surfaces as a return value from `start()`, which already ends at the single `logger.Fatal("could not start node", ...)` in `start_node.go`.

**Changes:**
- **`node.go` `start()`**: wrap with `errgroup.WithContext(n.ctx)`, replace all 4 Fatal goroutines (metrics bind/serve, API bind/serve) with `return err` / `g.Go(...)`, use `gctx` throughout all ctx-accepting calls, register `operatorNode.Start` in the group, return `g.Wait()`
- **`prober.go` `startHealthProber()`**: signature `→ error`; unhealthy probe now returns `fmt.Errorf(...)` instead of calling `logger.Fatal`
- **`eventsync.go` `syncContractEvents()`**: return `(*EventSyncer, func() error, error)`; ongoing-sync goroutine + Fatal removed; caller receives a start-func it `g.Go`s

## Test plan

- [ ] CI build passes
- [ ] Deploy to staging nodes, verify startup commit hash matches PR head
- [ ] Duties running at round 1 with no errors
- [ ] No `logger.Fatal` calls remaining in background goroutines in `cli/operator`

### DIFF
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -20,7 +20,9 @@ import (
 	"github.com/ssvlabs/ssv/ssvsigner/ekm"
 )
 
-// syncContractEvents blocks until historical events are synced and then spawns a goroutine syncing ongoing events.
+// syncContractEvents blocks until historical events are synced and returns the EventSyncer.
+// It also returns a start-func for ongoing event sync (nil if LocalEventsPath is set),
+// which the caller must run (e.g. via errgroup.Go) to keep the node current with the contract.
 func syncContractEvents(
 	ctx context.Context,
 	logger *zap.Logger,
@@ -32,15 +34,15 @@ func syncContractEvents(
 	operatorDataStore operatordatastore.OperatorDataStore,
 	keyManager ekm.KeyManager,
 	doppelgangerHandler eventhandler.DoppelgangerProvider,
-) (*eventsyncer.EventSyncer, error) {
+) (*eventsyncer.EventSyncer, func() error, error) {
 	eventFilterer, err := executionClient.Filterer()
 	if err != nil {
-		return nil, fmt.Errorf("failed to set up event filterer: %w", err)
+		return nil, nil, fmt.Errorf("failed to set up event filterer: %w", err)
 	}
 
 	eventParser, err := eventparser.New(eventFilterer)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create event parser: %w", err)
+		return nil, nil, fmt.Errorf("failed to create event parser: %w", err)
 	}
 
 	eventHandler, err := eventhandler.New(
@@ -55,7 +57,7 @@ func syncContractEvents(
 		eventhandler.WithLogger(logger),
 	)
 	if err != nil {
-		return nil, fmt.Errorf("failed to setup event data handler: %w", err)
+		return nil, nil, fmt.Errorf("failed to setup event data handler: %w", err)
 	}
 
 	eventSyncer := eventsyncer.New(
@@ -67,26 +69,28 @@ func syncContractEvents(
 
 	fromBlock, found, err := nodeStorage.GetLastProcessedBlock(nil)
 	if err != nil {
-		return nil, fmt.Errorf("syncing registry contract events failed, could not get last processed block: %w", err)
+		return nil, nil, fmt.Errorf("syncing registry contract events failed, could not get last processed block: %w", err)
 	}
 	if !found {
 		fromBlock = networkConfig.RegistrySyncOffset
 	} else if fromBlock == nil {
-		return nil, fmt.Errorf("syncing registry contract events failed, last processed block is nil")
+		return nil, nil, fmt.Errorf("syncing registry contract events failed, last processed block is nil")
 	} else {
 		// Start syncing from the next block.
 		fromBlock = new(big.Int).SetUint64(fromBlock.Uint64() + 1)
 	}
 
+	var ongoingSync func() error
+
 	// load & parse local events yaml if exists, otherwise sync from contract
 	if len(cfg.LocalEventsPath) != 0 {
 		localEvents, err := localevents.Load(cfg.LocalEventsPath)
 		if err != nil {
-			return nil, fmt.Errorf("failed to load local events: %w", err)
+			return nil, nil, fmt.Errorf("failed to load local events: %w", err)
 		}
 
 		if err := eventHandler.HandleLocalEvents(ctx, localEvents); err != nil {
-			return nil, fmt.Errorf("error occurred while running event data handler: %w", err)
+			return nil, nil, fmt.Errorf("error occurred while running event data handler: %w", err)
 		}
 	} else {
 		// Sync historical registry events.
@@ -106,7 +110,7 @@ func syncContractEvents(
 			)
 			fromBlock = new(big.Int).SetUint64(lastProcessedBlock + 1)
 		default:
-			return nil, fmt.Errorf("failed to sync historical registry events: %w", err)
+			return nil, nil, fmt.Errorf("failed to sync historical registry events: %w", err)
 		}
 
 		// Print registry stats.
@@ -137,19 +141,18 @@ func syncContractEvents(
 			zap.Int("my_validators", operatorValidators),
 		)
 
-		// Sync ongoing registry events in the background. Crash if it stops: the node can't operate
-		// without staying current with Ethereum events, and until reorg handling exists, restarting
-		// from persisted state is safer than continuing on possibly-incorrect state.
-		go func() {
-			err := eventSyncer.SyncOngoing(ctx, fromBlock.Uint64())
+		// Return a start-func for ongoing event sync. The caller (start()) registers it with errgroup
+		// so a failure propagates as a clean error rather than a Fatal. The node can't operate without
+		// staying current with Ethereum events, so any non-canceled failure exits the process.
+		ongoingFromBlock := fromBlock.Uint64()
+		ongoingSync = func() error {
+			err := eventSyncer.SyncOngoing(ctx, ongoingFromBlock)
 			if err != nil && !errors.Is(err, context.Canceled) {
-				logger.Fatal("failed syncing ongoing registry events",
-					zap.Uint64("last_processed_block", lastProcessedBlock),
-					zap.Error(err),
-				)
+				return fmt.Errorf("failed syncing ongoing registry events (from_block=%d): %w", ongoingFromBlock, err)
 			}
-		}()
+			return nil
+		}
 	}
 
-	return eventSyncer, nil
+	return eventSyncer, ongoingSync, nil
 }
--- a/cli/operator/eventsync.go
+++ b/cli/operator/eventsync.go
@@ -147,7 +147,7 @@ func syncContractEvents(
 		ongoingFromBlock := fromBlock.Uint64()
 		ongoingSync = func() error {
 			err := eventSyncer.SyncOngoing(ctx, ongoingFromBlock)
-			if err != nil && !errors.Is(err, context.Canceled) {
+			if err != nil && ctx.Err() == nil {
 				return fmt.Errorf("failed syncing ongoing registry events (from_block=%d): %w", ongoingFromBlock, err)
 			}
 			return nil
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -14,6 +14,7 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 
 	hexporter "github.com/ssvlabs/ssv/api/handlers/exporter"
 	hnode "github.com/ssvlabs/ssv/api/handlers/node"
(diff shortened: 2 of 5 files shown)


Roy-blox:
## QA Report: staging deploy ✅

| Field | Value |
|-------|-------|
| **Environment** | stage-hoodi |
| **Nodes** | 13-16 |
| **Sensor** | `ssv-stage-fix-errgroup-lifecycle-2867-r-53ee54-1` |
| **Deployed at** | 2026-06-16T12:38Z |

### 1. Commit verified
All 4 nodes started on commit `725368c9c7a73945c3deda1e246e0d6dba75cdae` (PR head).
```
2026-06-16T12:38:12Z ssv-node-13 INFO  starting SSV-Node:untagged-725368c9c7a73945c3deda1e246e0d6dba75cdae
2026-06-16T12:38:12Z ssv-node-14 INFO  starting SSV-Node:untagged-725368c9c7a73945c3deda1e246e0d6dba75cdae
2026-06-16T12:38:12Z ssv-node-15 INFO  starting SSV-Node:untagged-725368c9c7a73945c3deda1e246e0d6dba75cdae
2026-06-16T12:38:12Z ssv-node-16 INFO  starting SSV-Node:untagged-725368c9c7a73945c3deda1e246e0d6dba75cdae
```

### 2. Duty execution
Duties completing at round 1 on all nodes:
```
ssv-node-14 INFO  ✔️successfully finished duty processing runner_role=AGGREGATOR slot=3283293 consensus_rounds=1 total_consensus_time=0.062s total_duty_time=7.9s
```

### 3. Errors
Only benign P2P gossip noise (`validation ignored`) — same pattern as baseline. No application errors, no panics, no `Fatal` logs.

### 4. Metrics

| Operator | Peers | CL Sync | Consensus Avg (AGGREGATOR, 5m) |
|----------|-------|---------|-------------------------------|
| 13 | 7 | synced | 0.0065s |
| 14 | 8 | synced | 0.0077s |
| 15 | — | synced | 0.0066s |
| 16 | 7 | synced | 0.0081s |

Consensus at 6-13ms across duty types — normal for stage.

### 5. errgroup-specific checks
- No `FATAL` entries in any node's logs after startup — confirms Fatal goroutines are gone
- Nodes ran for 10+ minutes without any unexpected restarts or exits
- Health prober, ongoing event sync, metrics server, and `operatorNode.Start` all running as errgroup members with no premature cancellations observed

Roy-blox:
**Greptile's concerns — addressed in follow-up commit `fc70bdf`:**

**1. `startHealthProber` ctx.Err() guard — valid, fixed.**

The concern is correct: if a component's `Healthy()` implementation returns a non-context error when its network connection is torn down (e.g. `io.EOF` instead of `context.Canceled`), `ProbeAll` would return that error, and the old code would propagate it through the errgroup to `start_node.go`'s `logger.Fatal("could not start node")` — making a clean shutdown look like a startup failure.

Fixed: added `if ctx.Err() != nil { return nil }` before the error return in `startHealthProber`. If the context is already cancelled when ProbeAll fails, we return nil — the errgroup already has the root cause (or it's a clean shutdown). Added a test (`Test_startHealthProber_ctxCancelMasksProbeFail`) that covers this path.

**2. `ongoingSync` `context.DeadlineExceeded` — valid, fixed.**

The original `!errors.Is(err, context.Canceled)` filter misses `context.DeadlineExceeded`. Changed to `ctx.Err() == nil` which is both shorter and complete: if the context is done for any reason (canceled or deadline exceeded), the error is a shutdown artifact and we return nil.

cli/operator/node.go
```diff
@@ -529,15 +532,18 @@ func (n *node) start() error {
...
 	}
+	if ongoingSync != nil {
+		g.Go(ongoingSync)
+	}
 
-	go startHealthProber(n.ctx, n.logger, healthProber)
+	g.Go(func() error { return startHealthProber(gctx, n.logger, healthProber) })
 
-	if _, err := n.metadataSyncer.SyncAll(n.ctx); err != nil {
+	if _, err := n.metadataSyncer.SyncAll(gctx); err != nil {
```
momosh-ssv, line 541: What do you think about these early returns racing a backgrounded failure? 

Once `ongoingSync`/prober/metrics are `g.Go`'d, a failure cancels `gctx`, so this path returns `failed to sync metadata on startup: context canceled` while the real cause sits in the un-waited errgroup — the funnel Fatal then misattributes the crash. 

Maybe `return errors.Join(err, g.Wait())` on these early returns so the originating error wins?

cli/operator/node.go
```diff
@@ -599,21 +605,26 @@ func (n *node) start() error {
...
-	}
-	if err := n.operatorNode.Start(n.ctx); err != nil {
-		return fmt.Errorf("failed to start SSV node: %w", err)
+			return nil
+		})
 	}
 
-	return nil
+	g.Go(func() error {
+		if err := n.operatorNode.Start(gctx); err != nil {
```
momosh-ssv, line 621: Should we consider that `operatorNode.Start` still calls `logger.Fatal` internally (WS serve loop, `operator/node.go:445`)? 

Running inside this `g.Go`, it `os.Exit`s past the errgroup and `Close()` on a mid-run WS failure — the same pattern #2867 set out to remove, just one layer down. 

Worth a follow-up, or threading that `serveErr` out like metrics/API do?

cli/operator/prober_test.go
```diff
@@ -0,0 +1,60 @@
...
+// Test_startHealthProber_ctxCancelMasksProbeFail verifies the ctx.Err() guard: when ctx is
+// canceled at the same time a probe fails (e.g. the network stack tears down components before
+// the errgroup context fully propagates), startHealthProber returns nil rather than a misleading
+// "components unhealthy" error that would cause a Fatal log on normal shutdown.
+func Test_startHealthProber_ctxCancelMasksProbeFail(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // cancel before starting — simulates shutdown racing a probe
+
+	prober := hprobe.NewHealthProber(zap.NewNop())
+	prober.AddComponent("cl", simpleComponent{err: errors.New("broken")}, 100*time.Millisecond, 0, 0)
```
momosh-ssv, line 56: Might be worth strengthening this case — the stub's `Healthy` ignores ctx and returns the same error regardless, so the test passes via the guard but doesn't actually prove it distinguishes a real probe failure from a cancel. 

A component that respects ctx would return `context.Canceled` (swallowed to nil in `probeComponent`), so `ProbeAll` returns nil and the guard is never reached. A ctx-respecting stub would exercise the realistic teardown race the comment describes.

cli/operator/eventsync.go
```diff
@@ -137,19 +141,18 @@ func syncContractEvents(
...
-					zap.Error(err),
-				)
+		// Return a start-func for ongoing event sync. The caller (start()) registers it with errgroup
+		// so a failure propagates as a clean error rather than a Fatal. The node can't operate without
+		// staying current with Ethereum events, so any non-canceled failure exits the process.
+		ongoingFromBlock := fromBlock.Uint64()
+		ongoingSync = func() error {
+			err := eventSyncer.SyncOngoing(ctx, ongoingFromBlock)
+			if err != nil && ctx.Err() == nil {
+				return fmt.Errorf("failed syncing ongoing registry events (from_block=%d): %w", ongoingFromBlock, err)
```
momosh-ssv, line 151: Seems that the old Fatal logged `last_processed_block` while this now logs `from_block` (= `last_processed_block + 1`). 

Minor, but operators grepping for `last_processed_block` on this failure won't find it anymore — maybe include both fields?

PR #2899, opened 2026-06-18
Sync boole-fork with main + stage consensus fixes

momosh-ssv:
## What this is

**Step 1 of 2 toward full stage parity** for `boole-fork`. It syncs boole-fork's **consensus layer** with `stage`, keeping boole-fork's functionality paramount throughout. It does **not** yet bring in stage's infrastructure refactor — that's step 2 (see *Deferred*).

Base for this PR is `boole-fork`.

## How it's composed

1. **Merge `main` into `boole-fork`** — reconciles the 22-file consensus collision where both branches independently rewrote the same code (QBFT round-timer, controller, message validation, committee queues, duties). Because `main` already carries everything stage inherited from it, this pulls in the bulk of stage's consensus work. boole's behaviour wins wherever the two designs conflict.

2. **5 cherry-picked stage consensus commits** (original authorship preserved):
   - `#2858` — decouple validator-registration wire slot from local execution gate
   - `#2863` — standardize weighted attestation-data error handling / log levels
   - `#2864` — goclient multi-client init comments
   - `#2879` *(partial)* — duty-outcome model (`markDutySucceeded`/`NotRequired`/`Failed` + `concludeDuty`/`watchDutyOutcome`), with boole's `AggregatorCommitteeRunner` mapped onto the new markers. **The bundled p2p/topics infra (connector, libp2p logging, gossip tracer) is excluded.**
   - `#2859` *(partial)* — deterministic sync-committee contribution ordering (`sortBySubnet`). **Stage's `newRunnerTestKit` + new test files excluded.**
   - `#2877` — fix flaky pending-queued validator test (the real fix: `ActivationEpoch +1` → `+100` so an epoch-boundary-crossing run can't see the validator as attesting) + `saveShareWithRandomKey` test tidy, reconciled onto boole's test (boole's topic/`leaderTestCtx` setup kept; stage's `commons.CommitteeTopicID` churn dropped).

## Deliberately skipped (boole-paramount)

- `#2873` (qbft round-change) — would revert boole's intentional "bump round in a defer, after broadcast" design.
- `#2835` (qbft justified-round bump) — documented in its own commits as a no-op on an unreachable path.

## Deferred to step 2 — infra reconciliation (owners)

All recoverable from `origin/stage`, nothing lost:
- p2p connector subsystem + `NewDiscV5Service`/`SSVConfig` discovery API (`#2882`, parts of `#2879`)
- cli `node.go` split + errgroup shutdown (`#2866`, `#2876`)
- p2p/exporter test-port allocation (`#2819`)
- the test-kit-dependent test additions from `#2879`/`#2859`

These collide with boole-fork's own p2p/cli/network-key rework and need the p2p/cli owners to reconcile.

## Validation

- `go build ./...` clean.
- Pass: consensus/validation/duties/runner/validator/qbft + **full spec-test suite**.
- **Known pre-existing issues, not introduced here:** `aggregator_test.go` electra-duty/pre-electra-slot case (fails identically on `origin/boole-fork`), and flaky `TestTopicManager` (green on re-run).

## Known follow-up (from review, non-blocking)

Found one **observability** gap (no consensus-safety impact): boole's `AggregatorCommitteeRunner` (boole-only, so #2879's template didn't cover it) doesn't emit `markDutyFailed` on terminal post-quorum failures in `ProcessPreConsensus`/`ProcessPostConsensus`, so genuine failures are later classified as "stuck" rather than "failed". 

A blanket `defer markDutyFailed(err)` (the sibling `AggregatorRunner` pattern) is **not** safe here — this runner has a post-quorum "keep waiting for more messages" path that returns a non-nil (already-recovered) error, so a blanket defer would mis-mark an in-progress duty as failed. 

The correct fix is per-terminal-return `markDutyFailed` calls, best done as a follow-up.

## Next step

Step 2: reconcile stage's infra refactor (above) into boole-fork to reach full stage parity, after which the eventual `boole-fork → stage` PR becomes clean.


protocol/v2/ssv/runner/sync_committee_contribution.go
```diff
@@ -140,12 +148,15 @@ func (r *SyncCommitteeAggregatorRunner) ProcessPreConsensus(ctx context.Context,
...
-		selectionProofs = append(selectionProofs, blsSigSelectionProof)
-		subnets = append(subnets, subnet)
+		pairs = append(pairs, subnetSelectionProof{subnet: subnet, selectionProof: blsSigSelectionProof})
 	}
 
-	if len(selectionProofs) == 0 {
-		r.finishDuty()
+	// Sort by ascending subnet so the resulting Contributions slice has a
+	// deterministic, spec-canonical order. See sortBySubnet for the full rationale.
+	sortBySubnet(pairs)
```
iurii-ssv, line 156: `sortBySubnet` canonicalizes the *inputs*, but the resulting `Contributions` order is only deterministic if `GetSyncCommitteeContribution` (called below, ~L186) returns contributions in the same order as the `selectionProofs`/`subnets` it's handed. That's a reasonable assumption and matches upstream #2859 — but the regression test that pins it is exactly the part excluded from this partial cherry-pick, so on boole-fork the cross-node SSZ-root guarantee is currently **unverified in-tree**.

Since divergence here is consensus-relevant (two nodes disagreeing on the `Contributions` root → silent contribution failure), consider not deferring a *minimal* standalone test of `sortBySubnet` ordering (one that doesn't depend on stage's runner test-kit) to step 2. Non-blocking.

iurii-ssv (commented):
**High-level review — consensus-layer sync** (deferring the step-2 p2p/cli infra reconciliation; CI disregarded per scope).

I'm intentionally **not duplicating** findings already on record (and I agree they're non-blocking): the `AggregatorCommitteeRunner` `markDutyFailed` observability gap (PR description + greptile's summary), and greptile's two notes (un-namespaced `drop_reason`; the 2048→16 `MsgRouterConcurrency` default). One additional, non-blocking note inline.

protocol/v2/ssv/runner/sync_committee_contribution.go
momosh-ssv: Good call — agreed that's worth closing now rather than waiting for step 2, given it's consensus-relevant. I've added a standalone `TestSortBySubnet` (`protocol/v2/ssv/runner/sync_committee_contribution_test.go`, `package runner`) that pins the ascending-subnet ordering and asserts the `(subnet, proof)` pairing survives the sort, plus empty/nil safety. 

It's a pure-function test with no dependency on stage's runner test-kit, so it lands cleanly on boole-fork in-tree. The full end-to-end #2859 regression test still rides in with step 2. Pushed in `9476098dc`.

momosh-ssv:
@greptile do another review now

iurii-ssv (approved):

PR #2900, opened 2026-06-19
beacon/goclient: fork-gate aggregate index by duty slot (fix electra/pre-electra edge case)

momosh-ssv:
## Problem

`GoClient.computeAttDataRootAndVersion` (used by `SubmitAggregateSelectionProof`) decides EIP-7549 committee-index zeroing from the fork at **`attData.Slot`** — i.e. the slot the *beacon node returned*, the very source the adjacent comment flags as one that "may return inconsistent values":

```go
version, _ := gc.beaconConfig.BeaconForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(attData.Slot))
attData.Index = 0
if version < spec.DataVersionElectra {
    attData.Index = committeeIndex
}
```

For an **Electra duty** where the node returns attestation data stamped with a **pre-Electra slot**, `version` resolves pre-Electra, so the committee index is left non-zero. The resulting `AttestationDataRoot` sent in the `AggregateAttestation` request then **diverges from what peers compute** for the same duty — a real aggregate-attestation correctness issue around fork boundaries, not just a test artifact.

## Fix

Gate the fork decision on the requested **duty `slot`** (authoritative — it's the slot we're aggregating for) rather than the beacon-returned `attData.Slot`.

- **No behavior change in the normal case** (`attData.Slot == slot`).
- Corrects only the divergent edge case.

5-line change in `beacon/goclient/aggregator.go`.

## Verification

- `TestSubmitAggregateSelectionProof_UsesForkSpecificAggregateAndProof` — all 7 subtests pass (the `electra_duty_with_pre_electra_attestation_slot_uses_zero_index` subtest was failing before).
- Full `beacon/goclient` package green, no regressions.

## Note

This is a **pre-existing failure** on `boole-fork` (the subtest fails on the untouched base). It's currently reddening the `ssv (pre)` / `ssv (post)` checks on #2899 (the consensus-sync PR). Landing this fix in `boole-fork` clears that red — #2899 doesn't touch this code, so its `ssv (pre)/(post)` go green once the base is fixed.

### DIFF
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -107,7 +107,7 @@ func (gc *GoClient) waitTwoThirdsIntoSlot(ctx context.Context, slot phase0.Slot)
 	}
 }
 
-func (gc *GoClient) computeAttDataRootAndVersion(
+func (gc *GoClient) computeAttestationDataRoot(
 	ctx context.Context,
 	slot phase0.Slot,
 	committeeIndex phase0.CommitteeIndex,
@@ -122,7 +122,8 @@ func (gc *GoClient) computeAttDataRootAndVersion(
 	// Decide the fork from the requested duty slot, not attData.Slot — the latter is what the
 	// beacon node returned (the same source the comment above warns "may return inconsistent values"),
 	// whereas the aggregate is for our duty's slot, which is authoritative.
-	version, _ := gc.beaconConfig.BeaconForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(slot))
+	cfg := gc.getBeaconConfig()
+	version, _ := cfg.BeaconForkAtEpoch(cfg.EstimatedEpochAtSlot(slot))
 	attData.Index = 0
 	if version < spec.DataVersionElectra {
 		attData.Index = committeeIndex
@@ -140,7 +141,7 @@ func (gc *GoClient) fetchVersionedAggregate(
 	slot phase0.Slot,
 	committeeIndex phase0.CommitteeIndex,
 ) (*spec.VersionedAttestation, spec.DataVersion, error) {
-	root, err := gc.computeAttDataRootAndVersion(ctx, slot, committeeIndex)
+	root, err := gc.computeAttestationDataRoot(ctx, slot, committeeIndex)
 	if err != nil {
 		return nil, DataVersionNil, errMultiClient(fmt.Errorf("compute attestation root: %w", err), "AggregateAttestation")
 	}
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -118,8 +118,11 @@ func (gc *GoClient) computeAttDataRootAndVersion(
 	}
 
 	// Explicitly set Index field as beacon nodes may return inconsistent values.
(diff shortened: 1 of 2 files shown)


beacon/goclient/aggregator.go
```diff
@@ -118,8 +118,11 @@ func (gc *GoClient) computeAttDataRootAndVersion(
 	}
 
 	// Explicitly set Index field as beacon nodes may return inconsistent values.
-	// EIP-7549: Electra+ uses Index=0; pre-Electra uses committee index
-	version, _ := gc.beaconConfig.BeaconForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(attData.Slot))
+	// EIP-7549: Electra+ uses Index=0; pre-Electra uses committee index.
+	// Decide the fork from the requested duty slot, not attData.Slot — the latter is what the
+	// beacon node returned (the same source the comment above warns "may return inconsistent values"),
+	// whereas the aggregate is for our duty's slot, which is authoritative.
+	version, _ := gc.beaconConfig.BeaconForkAtEpoch(gc.getBeaconConfig().EstimatedEpochAtSlot(slot))
```
iurii-ssv: Minor / non-blocking: this statement mixes a raw field read (`gc.beaconConfig`) with the thread-safe accessor (`gc.getBeaconConfig()`) in the same expression. It's safe today since `beaconConfig` is written once at init (the constructor blocks on `beaconConfigInit` before returning, and reconnects only `AssertSame`), but for consistency with the accessor used elsewhere — and to drop one redundant `RLock` — consider snapshotting once:

```suggestion
	cfg := gc.getBeaconConfig()
	version, _ := cfg.BeaconForkAtEpoch(cfg.EstimatedEpochAtSlot(slot))
```

iurii-ssv: Minor / pre-existing (about the enclosing `computeAttDataRootAndVersion`): the function returns only `root` — the `version` it computes is discarded — so the `...AndVersion` suffix is misleading. Could rename to `computeAttestationDataRoot`. Not introduced by this PR; flagging only because the line is being touched.

iurii-ssv (approved):
Informational: the fix is backward-compatible across mixed old/new node clusters — this root is used only as the local query key to each node's own beacon node; it never crosses the SSV network as a consensus value (value-checks on the proposed `AggregateAndProof` are structural only), so divergence here cannot split consensus.

Two minor, non-blocking notes inline.

_Context (no change needed): the root is still hashed over the returned `attData` including `attData.Slot`, so this fixes the index component specifically; for spec-compliant nodes `attData.Slot == slot` and the two are equivalent anyway._

beacon/goclient/aggregator.go
momosh-ssv: Good call — snapshotted the config once now so both calls go through the same getBeaconConfig() accessor, consistent with the usage just above and dropping the redundant `RLock`. 

Done in bda85bf5d.

momosh-ssv: Agreed, the suffix was misleading — renamed to computeAttestationDataRoot since it only returns the root. 

Folded it in here while the line was being touched (bda85bf5d).

iurii-ssv (approved):
LGTM

PR #2906, opened 2026-06-25
network/p2p: de-flake TestP2pNetwork_MessageValidation + test-harness fixes

iurii-ssv:
## Problem

`TestP2pNetwork_MessageValidation` intermittently times out on CI (it surfaced again on a recent Unit-tests run). It waits for live GossipSub peer scores to settle into ordering invariants via a fixed wall-clock `assert.Eventually`, which races the scoring subsystem against the CI scheduler — under contention the same 30s budget buys less actual protocol progress. Two different failure modes were folded into that one wait:

- **peer population** — a node's score map never reached `nodeCount-1` peers, so the snapshot the assertion needs stayed partial;
- **score settling** — every observer had to satisfy the invariants in the *same* poll, a coincidence that grows unlikely under load.

The earlier mitigation raised the settle budget 15s → 30s, but that only ever addressed the second mode (and only probabilistically), which is why it resurfaced — most recently as a `len != nodeCount-1` population timeout, which a settle-budget bump can't fix.

## Fix (the de-flake)

Stop racing the wall clock; gate on the actual preconditions instead.

1. **Readiness gate** — before broadcasting, wait until every node's scorer tracks all `nodeCount-1` peers. This is a *monotonic* precondition (a peer stays in the score map once present; `RetainScore` ≈ one epoch ≫ the test), so waiting on it is robust, and it removes the partial-snapshot mode by construction.
2. **Sustained signal** — broadcast continuously rather than in a fixed burst, so the scoring signal is held while we wait for it to settle, instead of having to snapshot a transient post-burst window.
3. **Per-node latching** — latch each observer the moment it first holds a complete, valid snapshot, instead of requiring all observers valid in one poll.

The latching loop runs entirely on the test goroutine, so the `atomic.Pointer` + `assert.Eventually` goroutine-race handling the old code needed is removed too. Its inter-poll wait `select`s on `ctx.Done()` so a cancelled test context exits promptly.

## Folded-in test-harness fixes

Two pre-existing issues in the same `network/p2p` test harness, each its own commit:

- **`t.Fatalf` from a non-test goroutine** — `createVirtualNet`'s `PeerScoreInspector` runs on gossipsub's scoring goroutine, where `Fatalf`/`FailNow` doesn't actually fail the test (it just `Goexit`s that goroutine). Switched to `t.Errorf` (documented goroutine-safe).
- **Retry cleanup wedged on mDNS `Close()`** — `CreateAndStartLocalNet` closes a failed attempt's nodes before retrying; that close was sequential and unbounded, so a node whose `Close()` blocks (zeroconf's mDNS `Server.Shutdown` can wedge on a multicast socket write) hung the whole retry. Now closes concurrently with a timeout. **Only the failure path is affected** — a successful start closes nothing, so this is dead code on a healthy run (CI).

## Testing

`go vet` and `go test -c` clean. The de-flake change was validated on CI — **4/4** unit-test runs green under the full `-parallel 256` contention that caused the original flake. The mDNS-Close fix was verified locally (the retry loop now makes progress across attempts instead of wedging on the first `Close()`). The p2p suite uses mDNS-based local discovery, which doesn't form a network in my local environment, so the test's runtime is otherwise validated via CI.

### DIFF
--- a/network/p2p/p2p_validation_test.go
+++ b/network/p2p/p2p_validation_test.go
@@ -19,7 +19,6 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/sourcegraph/conc/pool"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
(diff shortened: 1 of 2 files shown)


network/p2p/p2p_validation_test.go
```diff
@@ -378,51 +398,56 @@ func TestP2pNetwork_MessageValidation(t *testing.T) {
...
-		// below can surface which invariant on which observer failed.
-		fresh := make(map[NodeIndex][]peerScore, len(vNet.Nodes))
-		for _, node := range vNet.Nodes {
-			fresh[node.Index] = snapshotForNode(node)
+		if len(settledSnap) == len(vNet.Nodes) || time.Now().After(deadline) {
+			break
 		}
-		settled.Store(&fresh)
+		time.Sleep(100 * time.Millisecond)
 	}
```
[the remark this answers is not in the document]
iurii-ssv: Good catch — addressed. The inter-poll wait now `select`s on `ctx.Done()`, and the loop is guarded with `for ctx.Err() == nil`, so a cancelled test context wakes the poll immediately and exits the loop during teardown (matching the ctx-aware `select` the broadcaster goroutine in this test already uses).

One heads-up on the suggested snippet for the next reader: a bare `break` inside the `select` breaks only the `select`, not the enclosing `for`, so the loop would keep iterating after cancellation — which is why I guarded the loop condition instead of relying on a `break` in the select.

iurii-ssv:
@greptile pls re-review

iurii-ssv:
@greptile pls re-review

network/p2p/p2p_validation_test.go
```diff
@@ -492,17 +523,21 @@ func createVirtualNet(
...
 				return
 			}
+			// This callback runs on gossipsub's scoring goroutine, so failures
+			// are reported with Errorf — Fatalf/FailNow must be called from the
+			// test goroutine. selfPeer/peerID are always one of this local net's
+			// virtual nodes, so these checks only guard a harness regression.
 			node := vn.NodeByPeerID(selfPeer)
 			if node == nil {
-				t.Fatalf("self peer not found (%s)", selfPeer)
+				t.Errorf("self peer not found (%s)", selfPeer)
```
momosh-ssv: What happens if the inspector fires after the test body returns? Switching to `t.Errorf` fixes the off-goroutine `Fatalf`/`FailNow` problem, but `t.Errorf` is itself only safe until the test returns. At scorer shutdown go-libp2p-pubsub does `defer ps.inspectScores()`, which spawns a detached `go ps.inspect(...)`; `p2pNetwork.Close()` cancels that ctx but doesn't await the goroutine, so it can still be scheduling after `vNet.Close()` and the rest of the test have returned. 

If it then hits one of these `NodeByPeerID == nil` branches, `t.Errorf` panics the whole test binary ("Log in goroutine after Test has completed") — turning the de-flake into a nastier intermittent failure.

The odds are low since these are "can't happen" guards on a self-contained local net, so not a blocker. But maybe safer to not touch `t` from this goroutine at all — record the message into an `atomic.Pointer[string]` here and assert it once on the test goroutine after `broadcasters.Wait()`?

iurii-ssv: Great catch — you're right, and I'd underweighted exactly this. The detached `go ps.inspect(...)` at scorer shutdown can outlive the test, and `t.Errorf` there would panic the binary ("Log in goroutine after Test has completed") — trading the flake for a crash.

Done it your way: the inspector now records the first failure into `VirtualNet.inspectorErr` (`atomic.Pointer[string]`) instead of touching `t`, and the test goroutine surfaces it via `t.Fatalf` after `broadcasters.Wait()`. A post-test firing is then just a harmless atomic store. Thanks for the careful read!

momosh-ssv (approved):
outstanding job here!

ovidiu-ssv-labs (approved):

Issue #2902, opened 2026-06-24
Track: boole-fork ⇄ stage reconciliation backlog (follow-up to #2899)


iurii-ssv:
## Context

#2899 ("Sync boole-fork with main + stage consensus fixes") was **step 1 of 2** toward `boole-fork` ⇄ `stage` parity. It merged (stale) `main` and cherry-picked a selection of `stage` consensus commits dated **up to ~June 13**. A review afterward found that several duty/consensus fixes landed on `stage` in the **June 19–22** window — after #2899's snapshot but before it merged — and were neither included nor listed as deferred.

This issue tracks **everything on `stage` (plus review follow-ups) not yet reconciled into `boole-fork`**, so we can check items off over time.

**Completeness:** this list accounts for **all 24 commits** on `stage` that are not in `boole-fork`. Of those, 4 were fully cherry-picked by #2899 (#2877, #2864, #2863, #2858 — not tracked here), and 2 were partially picked (#2879, #2859); the partial picks were verified to have dropped **only** p2p/topics infra and test-kit code — their duty cores are present — so the dropped remainders are tracked under **B**. The other 18 are tracked below.

Buckets:
- **A** — duty fixes missed by the snapshot (undocumented in #2899). High priority.
- **B** — work #2899 explicitly deferred / only partially cherry-picked.
- **C** — decisions to re-confirm at parity time (deliberately skipped, boole-paramount).
- **D** — minor cleanups from review.

> Where an item touches code `boole-fork` rewrote, it needs a **manual port, not a clean cherry-pick** — flagged per item.

---

### A. Duty fixes missed by the #2899 snapshot (high priority — not documented in #2899)

- [x] **#2888 — aggregator: aggregate for the attestation-data root the cluster attested with.** Today the aggregator requests the aggregate for a *locally re-derived* root instead of the root the cluster's QBFT actually attested with → beacon node 404 → **whole-cluster aggregator duty fails on head divergence** (lost aggregator rewards; intermittent). ⚠️ Must be reconciled with the boole-only **#2900**, which edits the *same* `computeAttestationDataRoot` for a *different* bug (fork-version selection). The two are mutually divergent — keep #2900's authoritative-slot fork-gating **and** add #2888's attested-root cache (+ its zero-client `multiClientSubmit` guard) and regression tests. Files: `beacon/goclient/aggregator.go`, `attest.go`, `goclient.go`.
[...]

PR #2907, opened 2026-06-29
fix(beacon): fetch head block root for sync-committee contributions (#2885)

momosh-ssv:
Bucket A of #2902 — ports stage #2885 onto `boole-fork`.

**Bug:** `GetSyncCommitteeContribution` resolves the contribution root with a by-slot block-root lookup (`Block: fmt.Sprint(slot)`), which 404s whenever the duty slot's proposal is missed or arrives late → the SYNC_COMMITTEE_CONTRIBUTION duty fails for the whole cluster.

**Fix:** query `"head"`. Sync-committee messages sign the head root as seen during the duty slot, so for an empty/late slot the correct root is the previous block's — a head query returns it, a by-slot query 404s. Mirrors the message-path fix (#839) that the contribution path was left behind by.

Follow-ons from the same upstream commit: `waitOneThirdIntoSlot` surfaces ctx cancellation as an error; `DataVersionNil` on the 2/3-wait error path; manual nil-checks → `checkPtrResponse` (added on boole in #2905); dropped an obsolete loop-var copy. Adds a regression test that 404s any by-slot lookup and only answers `head`.

**Validation:** `go build ./...` clean (except the pre-existing `spectest fixRunnerForRun`); `go test ./beacon/goclient/...` green incl. the new regression test; gofmt clean.

### DIFF
--- a/beacon/goclient/sync_committee_contribution.go
+++ b/beacon/goclient/sync_committee_contribution.go
@@ -51,7 +51,7 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 		return nil, DataVersionNil, fmt.Errorf("mismatching number of selection proofs and subnet IDs")
 	}
 
-	if err := gc.waitOneThirdIntoSlot(ctx, slot); err != nil {
+	if err := gc.waitIntoSlot(ctx, slot, 1); err != nil {
 		return nil, DataVersionNil, fmt.Errorf("wait for 1/3 of slot: %w", err)
 	}
 
@@ -75,7 +75,7 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 
 	blockRoot := beaconBlockRootResp.Data
 
-	if err := gc.waitTwoThirdsIntoSlot(ctx, slot); err != nil {
+	if err := gc.waitIntoSlot(ctx, slot, 2); err != nil {
 		return nil, DataVersionNil, fmt.Errorf("wait for 2/3 of slot: %w", err)
 	}
 
@@ -128,22 +128,4 @@ func (gc *GoClient) SubmitSignedContributionAndProof(
 	}
 
 	return nil
-}
-
-// waitOneThirdIntoSlot waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after slot start time)
-func (gc *GoClient) waitOneThirdIntoSlot(ctx context.Context, slot phase0.Slot) error {
-	config := gc.getBeaconConfig()
-	delay := config.IntervalDuration()
-	finalTime := config.SlotStartTime(slot).Add(delay)
-	wait := time.Until(finalTime)
-	if wait <= 0 {
-		return nil
-	}
-
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-time.After(wait):
-		return nil
-	}
 }
--- a/beacon/goclient/sync_committee_contribution.go
+++ b/beacon/goclient/sync_committee_contribution.go
@@ -80,6 +80,10 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 	}
 
 	// Fetch sync committee contributions for each subnet in parallel.
+	// NOTE: this read and the head-root read above are not pinned to the same beacon
+	// node, and the multiClient does not fail over on a 4xx; if scores reorder in
+	// between, this call can land on a node that hasn't imported blockRoot and 404.
+	// Narrow (reorg/lagging-node window); tracked in ssvlabs/go-eth2-client#16.
 	var (
 		contributions = make(spectypes.Contributions, len(subnetIDs))
 		g             errgroup.Group
--- a/beacon/goclient/sync_committee_contribution.go
+++ b/beacon/goclient/sync_committee_contribution.go
@@ -18,12 +18,11 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
-// IsSyncCommitteeAggregator returns tru if aggregator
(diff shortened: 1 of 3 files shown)


beacon/goclient/sync_committee_contribution.go
```diff
@@ -80,28 +85,24 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 		g             errgroup.Group
 	)
 	for i := range subnetIDs {
-		index := i
 		g.Go(func() error {
 			start := time.Now()
 			syncCommitteeContrResp, err := gc.multiClient.SyncCommitteeContribution(ctx, &api.SyncCommitteeContributionOpts{
```
iurii-ssv, line 94: Separate from the head-vs-signed-root divergence already noted at the head query above: the two `multiClient` calls here aren't pinned to the same node, and a 4xx isn't failed over.

`doCall` re-runs `activeClientsSortedByScore()` independently per call and returns immediately on a 4xx (`case ... statusCodeFamily(apiErr.StatusCode) == 4: return res, err` — no failover), verified in `ssvlabs/go-eth2-client` `multi/client.go`. With the old by-slot lookup the resolved root was a fact every synced node agreed on, so a node split between the two calls was harmless. With `"head"` the root is each node's subjective fork-choice head.

Scenario: `BeaconBlockRoot{head}` resolves root R on node A (top score); before this contribution call A is briefly de-scored, so the call lands on node B which hasn't imported R (or has no sync-committee messages for R) → 404 → no failover → the duty errors for this operator. It's narrow — needs score reordering between the two back-to-back calls *and* the second node lacking R — so not a blocker, but it's new exposure the by-slot path didn't have. Options: pin both calls to one client, or make a contribution 4xx failover-eligible.

beacon/goclient/sync_committee_contribution.go
```diff
@@ -130,18 +131,19 @@ func (gc *GoClient) SubmitSignedContributionAndProof(
 }
 
 // waitOneThirdIntoSlot waits until one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after slot start time)
-func (gc *GoClient) waitOneThirdIntoSlot(ctx context.Context, slot phase0.Slot) {
+func (gc *GoClient) waitOneThirdIntoSlot(ctx context.Context, slot phase0.Slot) error {
```
iurii-ssv: Cleanup: now that this function returns `error`, it's byte-for-byte identical to `waitTwoThirdsIntoSlot` (`beacon/goclient/aggregator.go`) except the interval multiplier (`Add(delay)` vs `Add(2 * oneInterval)`). Worth collapsing into a single `waitIntoSlot(ctx, slot, intervals int) error` — the in-flight `IntervalDuration(slot)` Gloas migration otherwise has to be applied in both copies (and historically one was already missed).

beacon/goclient/sync_committee_contribution.go
```diff
@@ -51,27 +51,32 @@ func (gc *GoClient) GetSyncCommitteeContribution(
 		return nil, DataVersionNil, fmt.Errorf("mismatching number of selection proofs and subnet IDs")
 	}
 
-	gc.waitOneThirdIntoSlot(ctx, slot)
+	if err := gc.waitOneThirdIntoSlot(ctx, slot); err != nil {
```
iurii-ssv: The new behavior introduced here — `waitOneThirdIntoSlot` returning `ctx.Err()`, wrapped as `"wait for 1/3 of slot: %w"` — has no test coverage: both contribution tests use a past slot that hits the `wait <= 0` early-return, so this error path never executes. A future refactor could revert to the old silent return and the suite would stay green. A minimal test (future slot + already-cancelled ctx, asserting the wrapped error and `DataVersionNil`) would lock it in.

iurii-ssv (commented):
Nice! Three inline notes: one correctness consideration (distinct from the head-divergence point the bot already raised at the head query) and two low-severity quality items.

beacon/goclient/sync_committee_contribution.go
momosh-ssv: Confirmed — verified `doCall` re-sorts by score per call and returns on 4xx without failover, so with `"head"` the two reads can land on different nodes and the second 404s with no retry.

 Strictly new exposure vs. the old by-slot path, but narrow (needs a score reorder between the back-to-back calls *and* the second node lacking the root), and it's a library-side gap — `multi.Service` exposes no single-client pinning. 
 
 Left it out of this minimal bugfix and tracked it in ssvlabs/go-eth2-client#16 (pin a sequence to one client, or make the read 4xx failover-eligible). Added a `NOTE` at the call site pointing there in 7157d44.

beacon/goclient/sync_committee_contribution.go
momosh-ssv: Done in d6ece6c — collapsed both into `waitIntoSlot(ctx, slot, intervals int)` (intervals=1 → ⅓, intervals=2 → ⅔), updated all three call sites, and removed the two copies. The future `IntervalDuration(slot)` Gloas migration now only has one place to touch.

beacon/goclient/sync_committee_contribution.go
momosh-ssv: Added `TestGetSyncCommitteeContributionPropagatesCanceledContext` in 14755e2 — future slot so the wait actually blocks, already-canceled context, asserting the wrapped `"wait for 1/3 of slot"` error, `context.Canceled`, `DataVersionNil`, and that the beacon node is never queried. Locks the path in against a silent-return regression.

iurii-ssv (approved):
LGTM

PR #2909, opened 2026-06-29
fix(beacon): aggregate for the root the cluster attested with (#2888)

momosh-ssv:
Bucket A of #2902 — ports stage #2888 onto `boole-fork`, reconciled with boole's #2900.

**Bug:** the aggregator re-derives attestation data at 2/3 of the slot and requests the aggregate for *that* root — this node's local view, not necessarily what the cluster's QBFT decided and attested with. On head divergence the re-derived root matches nothing anyone attested with → beacon node **404 → the whole cluster's AGGREGATOR duty fails** (lost rewards, intermittent).

**Fix:** remember the root of the attestation data this node actually submitted (`SubmitAttestations`), keyed by `(slot, committee)`, and request the aggregate for that root — the beacon node holds at least our own attestation matching it, so an aggregate must exist. Fall back to re-deriving when uncached (attestation failed / not landed yet).

**Boole reconciliation (the part that differs from stage):**
- The attested-root preference is injected in `fetchVersionedAggregate`, which boole shares between `SubmitAggregateSelectionProof` (AggregatorRunner) **and** `GetAggregateAttestation` (boole's `AggregatorCommitteeRunner`). So **both** boole aggregator paths are fixed — stage's #2888 only covered the submit path.
- `computeAttestationDataRoot` keeps **#2900**'s duty-slot fork-gating untouched, as the fallback.
- Guards `multiClientSubmit` against zero clients (would panic on `WithMaxGoroutines(0)` / silently return nil and cache a never-submitted root).

**Review:** deep-reviewed (consensus-correctness) — APPROVE, no must/should-fix; committee-index cache keying traced end-to-end. Ported the full upstream regression suite + added `TestGetAggregateAttestation_PrefersAttestedDataRoot` for boole's second path. #2900's `UsesForkSpecificAggregateAndProof` test still green. `go build/test/vet/gofmt` clean (except pre-existing `spectest fixRunnerForRun`).

### DIFF
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -146,14 +146,14 @@ func (gc *GoClient) fetchVersionedAggregate(
 	// aggregate must exist. Re-deriving the data locally can yield a root nobody attested with,
 	// which the beacon node answers with a 404 (no matching aggregate).
 	//
-	// NOTE (boole adaptation): fetchVersionedAggregate is shared by both SubmitAggregateSelectionProof
-	// (AggregatorRunner) and GetAggregateAttestation (AggregatorCommitteeRunner). Both paths have the
-	// identical 404 risk, so injecting here fixes both — stage's #2888 only fixed the submit path.
+	// A cache hit only guarantees that *some* beacon node accepted the attestation. The
+	// AggregateAttestation request below is routed to the single highest-scored client and does
+	// not fail over on 4xx, so it can still 404 if that node hasn't ingested the attestation yet
+	// (gossip backfill before 2/3 of the slot makes this rare).
 	root, found := gc.attestedDataRoot(slot, committeeIndex)
 	if !found {
 		// No record of our own attestation (it failed or hasn't landed yet) — fall back to
-		// re-deriving the root from this node's view of the slot. computeAttestationDataRoot
-		// preserves #2900's authoritative-slot fork-gating (BeaconForkAtEpoch(EstimatedEpochAtSlot(slot))).
+		// re-deriving the root from this node's view of the slot, fork-gated by the duty slot.
 		var err error
 		root, err = gc.computeAttestationDataRoot(ctx, slot, committeeIndex)
 		if err != nil {
--- a/beacon/goclient/aggregator.go
+++ b/beacon/goclient/aggregator.go
@@ -141,9 +141,24 @@ func (gc *GoClient) fetchVersionedAggregate(
 	slot phase0.Slot,
 	committeeIndex phase0.CommitteeIndex,
 ) (*spec.VersionedAttestation, spec.DataVersion, error) {
-	root, err := gc.computeAttestationDataRoot(ctx, slot, committeeIndex)
-	if err != nil {
-		return nil, DataVersionNil, errMultiClient(fmt.Errorf("compute attestation root: %w", err), "AggregateAttestation")
+	// Prefer the root of the attestation data this node actually submitted (the cluster-decided
+	// value): the beacon node holds at least our own attestation matching it, so a matching
+	// aggregate must exist. Re-deriving the data locally can yield a root nobody attested with,
+	// which the beacon node answers with a 404 (no matching aggregate).
+	//
+	// NOTE (boole adaptation): fetchVersionedAggregate is shared by both SubmitAggregateSelectionProof
+	// (AggregatorRunner) and GetAggregateAttestation (AggregatorCommitteeRunner). Both paths have the
+	// identical 404 risk, so injecting here fixes both — stage's #2888 only fixed the submit path.
+	root, found := gc.attestedDataRoot(slot, committeeIndex)
+	if !found {
+		// No record of our own attestation (it failed or hasn't landed yet) — fall back to
+		// re-deriving the root from this node's view of the slot. computeAttestationDataRoot
+		// preserves #2900's authoritative-slot fork-gating (BeaconForkAtEpoch(EstimatedEpochAtSlot(slot))).
+		var err error
+		root, err = gc.computeAttestationDataRoot(ctx, slot, committeeIndex)
+		if err != nil {
+			return nil, DataVersionNil, errMultiClient(fmt.Errorf("compute attestation root: %w", err), "AggregateAttestation")
+		}
 	}
 
 	start := time.Now()
--- a/beacon/goclient/attest.go
+++ b/beacon/goclient/attest.go
@@ -581,7 +581,7 @@ func (gc *GoClient) SubmitAttestations(ctx context.Context, attestations []*spec
 
 	// Only reached when at least one client accepted the attestations, so the beacon node
 	// holds them — remember their data roots for the aggregator flow.
-	gc.rememberAttestedDataRoots(attestations)
+	gc.rememberAttestedDataRoots(ctx, attestations)
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


beacon/goclient/aggregator.go
```diff
@@ -141,9 +141,24 @@ func (gc *GoClient) fetchVersionedAggregate(
...
-		return nil, DataVersionNil, errMultiClient(fmt.Errorf("compute attestation root: %w", err), "AggregateAttestation")
+	// Prefer the root of the attestation data this node actually submitted (the cluster-decided
+	// value): the beacon node holds at least our own attestation matching it, so a matching
+	// aggregate must exist. Re-deriving the data locally can yield a root nobody attested with,
+	// which the beacon node answers with a 404 (no matching aggregate).
+	//
+	// NOTE (boole adaptation): fetchVersionedAggregate is shared by both SubmitAggregateSelectionProof
+	// (AggregatorRunner) and GetAggregateAttestation (AggregatorCommitteeRunner). Both paths have the
+	// identical 404 risk, so injecting here fixes both — stage's #2888 only fixed the submit path.
+	root, found := gc.attestedDataRoot(slot, committeeIndex)
```
iurii-ssv, line 153: Observation (low severity, not a regression). A cache hit guarantees that *some* beacon node accepted the attestation, but the subsequent `AggregateAttestation` is routed by the multi-client to the single highest-scored node, and `doCall` does not fail over on 4xx. So if that node rejected / hasn't yet ingested the attestation while a lower-scored peer accepted it, the cache hits but the request can still 404 with no failover. Gossip backfill before 2/3 of the slot makes this rare, and the old always-re-derive path had the same single-node exposure — so this isn't a regression. Might be worth one sentence in the comment acknowledging the hit path still assumes the *queried* node holds it.

beacon/goclient/attest.go
```diff
@@ -550,27 +556,93 @@ func (gc *GoClient) multiClientSubmit(
...
+
+// rememberAttestedDataRoots caches the root of the attestation data that was just submitted,
+// per (slot, committee). The aggregator flow prefers this root over re-deriving the data:
+// the beacon node holds at least our own attestation matching it, so an aggregate must
+// exist, while a re-derived root may match nothing anyone attested with.
+func (gc *GoClient) rememberAttestedDataRoots(attestations []*spec.VersionedAttestation) {
+	for _, att := range attestations {
+		data, err := att.Data()
+		if err != nil {
+			gc.log.Debug("could not extract attestation data", zap.Error(err))
```
iurii-ssv: These three extraction failures are swallowed with `Debug` + `continue`, so a populated-but-empty cache silently degrades to the 404-prone re-derivation in `fetchVersionedAggregate` — indistinguishable from a legitimate cache miss. Since these errors shouldn't happen for an attestation we just successfully submitted, consider `Warn` + a counter so a "submitted but un-rememberable" entry is observable rather than only a disconnected DEBUG line.

beacon/goclient/attest.go
```diff
@@ -550,27 +556,93 @@ func (gc *GoClient) multiClientSubmit(
...
+			continue
+		}
+		committee, err := attestationCommitteeIndex(att, data)
+		if err != nil {
+			gc.log.Debug("could not extract attestation committee index", zap.Error(err))
+			continue
+		}
+		root, err := data.HashTreeRoot()
```
iurii-ssv, line 625: `rememberAttestedDataRoots` runs once per submitted attestation, but the committee runner submits one attestation per validator and validators in the same committee share a single `AttestationData` (for Electra the data is identical across *all* committees too). So a large committee recomputes the same `HashTreeRoot` and rewrites the same `(slot, committee)` key N times to produce one entry. Minor — dedupe per distinct `(slot, committee)` if you end up touching this.

beacon/goclient/attest.go
```diff
@@ -550,27 +556,93 @@ func (gc *GoClient) multiClientSubmit(
...
+	item := gc.attestedDataRootCache.Get(attestedDataRootKey{slot: slot, committee: committee})
+	if item == nil {
+		return phase0.Root{}, false
+	}
+	return item.Value(), true
+}
+
+// attestationCommitteeIndex extracts the committee an attestation belongs to: from the
+// committee bits for Electra and later (EIP-7549), from the data's Index field before.
+func attestationCommitteeIndex(att *spec.VersionedAttestation, data *phase0.AttestationData) (phase0.CommitteeIndex, error) {
```
iurii-ssv: The pinned `ssvlabs/go-eth2-client` fork already exposes `VersionedAttestation.CommitteeIndex()` with identical semantics (pre-Electra → `Data.Index`; Electra/Fulu → the single committee bit, erroring on 0 or >1 bits). Calling `att.CommitteeIndex()` in `rememberAttestedDataRoots` lets you drop this helper, the `//nolint:gosec`, the extra `data` parameter, and `TestAttestationCommitteeIndex` (which re-tests library behavior).

beacon/goclient/aggregator.go
```diff
@@ -141,9 +141,24 @@ func (gc *GoClient) fetchVersionedAggregate(
 	slot phase0.Slot,
 	committeeIndex phase0.CommitteeIndex,
 ) (*spec.VersionedAttestation, spec.DataVersion, error) {
-	root, err := gc.computeAttestationDataRoot(ctx, slot, committeeIndex)
-	if err != nil {
-		return nil, DataVersionNil, errMultiClient(fmt.Errorf("compute attestation root: %w", err), "AggregateAttestation")
+	// Prefer the root of the attestation data this node actually submitted (the cluster-decided
```
iurii-ssv, line 144: This comment block embeds ephemeral `#2888`/`#2900` and "stage vs boole" references that will dangle once the branches merge/squash. Suggest keeping the *why* (prefer the root we actually attested with; fall back to re-deriving with duty-slot fork-gating) and dropping the cross-branch/issue-number archaeology.

iurii-ssv (commented):
Overall looks solid, just some consistency / cleanup comments to consider.

beacon/goclient/aggregator.go
momosh-ssv: Good call — added that caveat to the comment. The hit path now notes a cache hit only guarantees *some* node accepted the attestation, while `AggregateAttestation` routes to the single highest-scored client with no 4xx failover, so it can still 404 if that node hasn't ingested it yet. Done in c320b0a.

beacon/goclient/attest.go
momosh-ssv: Agreed, a silent degrade to the 404-prone re-derivation shouldn't be invisible. Switched all three extraction failures to `Warn` and added a counter (`ssv.cl.attestation_data.remember_root_failed`) that should stay at zero. Done in da1e3cc.

beacon/goclient/attest.go
momosh-ssv: Done — added a `seen` set so we hash and store each distinct `(slot, committee)` root only once per submission batch. da1e3cc.

beacon/goclient/attest.go
momosh-ssv: Nice — confirmed the pinned fork's `CommitteeIndex()` has identical semantics (pre-Electra → `Data.Index`; Electra/Fulu → single committee bit, erroring on 0 or >1). Switched `rememberAttestedDataRoots` to `att.CommitteeIndex()` and dropped the helper, the `//nolint:gosec`, the `data` param, and `TestAttestationCommitteeIndex`. da1e3cc.

beacon/goclient/aggregator.go
momosh-ssv: Agreed they'd dangle post-merge. Trimmed the comment to keep only the *why* (prefer the root we attested with; fall back to re-deriving fork-gated by the duty slot) and dropped the cross-branch/issue archaeology. c320b0a.

iurii-ssv (approved):
Gj!

PR #2910, opened 2026-06-29
fix(duties): retry-capable sync-committee duty-fetching (#2769)

momosh-ssv:
Bucket A of #2902 — ports stage #2769 onto `boole-fork` (hand-port onto boole's diverged sync-committee handler).

**Bug:** the sync-committee handler used `fetchCurrentPeriod`/`fetchNextPeriod` boolean flags. A single transient beacon-node error at a sync-committee period boundary could drop a **whole period (~256 epochs) of duties**.

**Fix:** replace the booleans with a per-period `dutyFetchIntents` map (period → fulfilled). An unfulfilled intent is retried on later slot-ticks, so a transient error no longer drops the period. This mirrors the intent model boole's `attester`/`proposer` handlers already use.

**Boole reconciliation (the part not in upstream):**
- Boole's `BooleForkAtSlot` gate is preserved: **pre-fork** runs the Alan `processExecution`; **post-fork** it only fetches + stores duties and feeds **both** the Committee and AggregatorCommittee handlers (which own execution). Kept boole's `netCfg` naming.
- Reorg re-fetches **only** the next period (current-period membership is fixed a period ahead) and does **not** Reset before re-fetch (serve stale + retry rather than drop on a failed re-fetch — strictly safer than the old code).

**Supporting:** `base_handler` gains `atLastSlotOrPastCurrentPeriod`; `attester`/`proposer` register the next-epoch intent **before** the indices-change select (a deadline-overrunning tick still records it); `networkconfig` rename `LastSlotOfSyncPeriod`→`LastActionableSlotOfSyncPeriod` (name/doc only, semantics unchanged).

**Review:** deep-reviewed for duty-scheduling correctness — **APPROVE**; the reviewer traced retry/transition/reorg/cleanup/deadline paths and confirmed no duty can be dropped, duplicated, or mis-timed. Ported the upstream retry/reorg/indices test suite + added `TestScheduler_SyncCommittee_Boole_Fork_Fetches_But_Does_Not_Execute` to guard the boole post-fork gate (previously uncovered — the whole duties suite ran pre-Boole). `go build/test/vet/gofmt` clean (except pre-existing `spectest fixRunnerForRun`).

### DIFF
--- a/networkconfig/beacon.go
+++ b/networkconfig/beacon.go
@@ -99,8 +99,10 @@ func (b *Beacon) FirstEpochOfSyncPeriod(period uint64) phase0.Epoch {
 	return phase0.Epoch(period * b.EpochsPerSyncCommitteePeriod)
 }
 
-// LastSlotOfSyncPeriod calculates the first epoch of the given sync period.
-func (b *Beacon) LastSlotOfSyncPeriod(period uint64) phase0.Slot {
+// LastActionableSlotOfSyncPeriod calculates the last slot of the given sync period for which
+// producing a sync committee message still makes sense. Note this is one slot before the period's
+// literal last slot (a message produced during that final slot would never be included).
+func (b *Beacon) LastActionableSlotOfSyncPeriod(period uint64) phase0.Slot {
 	lastEpoch := b.FirstEpochOfSyncPeriod(period+1) - 1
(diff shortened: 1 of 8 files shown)


iurii-ssv (approved):
Nice!

PR #2911, opened 2026-06-29
fix(ssvsigner): pin fixed-fork signing domains for the remote signer (#2875)

momosh-ssv:
Bucket A of #2902 — **scoped** port of stage #2875 onto `boole-fork`: the remote-signing fix only.

**Bug ("Hoodi incident"):** voluntary exits are signed over the EIP-7044-pinned **Capella** domain, but the remote signer (ssv-signer → Web3Signer) sent the *current* fork in `fork_info`. A Web3Signer with a wrong/missing `--network` resolves the exit epoch pre-Deneb, recomputes a mismatched root, and rejects with **HTTP 400 → the exit never signs**. Liveness bug for remote-signing operators (you confirmed boole needs remote signing); local signing was already correct.

**Fix:** `prepareSignRequest` pins the fixed fork — Capella for `DomainVoluntaryExit`, genesis-builder for `DomainApplicationBuilder` — so the remote request is self-consistent with the locally-computed `signing_root` and matches the local signer. A missing-fork lookup **errors out** (never pins a zero-value `Fork`/wrong domain). New `ForkAtVersion` on the beacon config + interface does the fixed-fork lookup. Plus a signer-agnostic `--network` hint on exit-signing failure.

**Deliberately NOT ported (the bundled refactor):** #2875 also dedups `VerifyBeaconPartialSignature`/`ResolveDuplicateSignature` into shared code, touching `runner.go`, `runner_signatures.go`, `committee_observer.go`, `partial_sig_container.go`, `types/crypto.go` — boole's heavily-diverged consensus core. That cleanup is unrelated to remote signing and collides with boole; left out by design.

**Review:** deep-reviewed for signing-domain correctness — **APPROVE**. The reviewer verified by source that the pinned domain equals what the local node computes for both domains (`Forks[Capella]`+gvr for exits; `Forks[Phase0].CurrentVersion`+zeroed gvr for registration), and that the `ok=false` path errors rather than signing over a wrong domain. `MockBeaconNetwork` left as-is (didn't expand dead code). `go build/test/vet/gofmt` clean (except pre-existing `spectest fixRunnerForRun`); e2e signing tests added (Docker-gated).

### DIFF
--- a/networkconfig/beacon.go
+++ b/networkconfig/beacon.go
@@ -166,6 +166,11 @@ func (b *Beacon) BeaconForkAtEpoch(epoch phase0.Epoch) (spec.DataVersion, *phase
 	return version, &fork
 }
 
+func (b *Beacon) ForkAtVersion(version spec.DataVersion) (phase0.Fork, bool) {
+	fork, ok := b.Forks[version]
+	return fork, ok
+}
+
 func (b *Beacon) AssertSame(other *Beacon) error {
 	if b.Name != other.Name {
 		return fmt.Errorf("different Name")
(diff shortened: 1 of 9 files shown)


iurii-ssv (approved):
Looks correct.

PR #2912, opened 2026-06-29
observability(runner): classify AggregatorCommitteeRunner terminal failures as failed (#2902)

momosh-ssv:
Bucket B of #2902 — the observability follow-up raised in the #2899 review.

**Gap:** boole's `AggregatorCommitteeRunner` never calls `markDutyFailed` on terminal post-quorum failures, so the duty-outcome watcher classifies genuine failures as **"stuck"** rather than **"failed"**.

**Fix:** `r.markDutyFailed(err)` at each terminal failure return — **6 sites**:

`ProcessPreConsensus`
- `expectedPreConsensusRoots` error
- defensive `default:` — unexpected role type in pre-consensus metadata (deterministic invariant violation)

`ProcessPostConsensus`
- `expectedPostConsensusRootsAndBeaconObjects` error
- empty beacon objects → `ErrNoValidDutiesToExecute` (terminal, non-recoverable: `committee_queue` drops the message and terminates the runner)
- defensive `default:` — unexpected role type in post-consensus (deterministic invariant violation)
- submit / construct / missing-object `executionErr` (reconstruction already succeeded, so the root keeps quorum and is never re-reported → the duty never concludes on its own)

**Deliberately NOT marked** (a blanket `defer` is unsafe — these are re-entrant or still-recoverable):
- the re-entrant pre-consensus window — `Validate()` / `decide()` and the keep-waiting `return anyErr` (duty still progressing; this window re-enters on every pre-consensus message until consensus starts)
- `waitTwoThirdsIntoSlot` + listener `ctx.Done()` returns — reachable only on shutdown within a short wait; `markDutyFailed` drops `context.Canceled`, and the duty deadline (~1 epoch out) is unreachable there, so there is nothing to record
- the recoverable post-consensus `PostConsensusQuorumWithInvalidSignatures` return — `FallBackAndVerifyEachSignature` can drop the root below quorum so a later partial-sig message re-crosses quorum and retries; discriminated from the terminal submit/construct case via `errors.As` + `.Code`
- the `markDutyNotRequired` / success paths

All exclusions are documented inline.

**Review:** deep-reviewed for duty-outcome correctness — **APPROVE**. Verified no marked site can fire while the duty is still recoverable (the pre-consensus re-entry marks were dropped for exactly this reason), and that the `executionErr` split correctly excludes only the recoverable invalid-sigs sub-case.

Regression coverage added in 64fa002: a new `AggregatorCommitteeRunner` post-consensus test env (mirroring `newCommitteeRunnerEnv`) plus a beacon wrapper that injects a submit error, covering both branches of the split — submit failure concludes `failed`, recoverable invalid-sigs does not.

`go build/test/vet/gofmt` clean (except the pre-existing `spectest fixRunnerForRun`).

### DIFF
--- a/protocol/v2/ssv/runner/aggregator_committee.go
+++ b/protocol/v2/ssv/runner/aggregator_committee.go
@@ -588,7 +588,13 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 
 	// Else, if some aggregators or contributors were selected (even with an error for others), proceed to consensus
 	if err := consensusData.Validate(); err != nil {
-		r.markDutyFailed(err)
+		// Not markDutyFailed: this pre-consensus window is re-entrant until consensus starts
+		// (guarded by HasStartedConsensus). basePreConsensusMsgProcessing reports a new quorum
+		// per root, not once per duty, and a Validate error is non-retryable so the committee
+		// queue drops the message but keeps the runner alive. A transient GetAggregateAttestation
+		// failure can leave consensusData empty here, yet a later message whose root reaches quorum
+		// can still build valid data and start consensus — concluding "failed" now would mask a
+		// duty that still completes.
 		return fmt.Errorf("invalid aggregator committee consensus data: %w", err)
 	}
 
@@ -600,7 +606,11 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 		consensusData,
 		r.ValCheck,
 	); err != nil {
-		r.markDutyFailed(err)
+		// Not markDutyFailed: same re-entrant pre-consensus window as the Validate check above.
+		// decide sets RunningInstance only on success, so a failure here leaves HasStartedConsensus
+		// false and a later new-quorum message can start consensus and complete the duty. The
+		// message drop is non-terminal (runner survives), so concluding "failed" now would risk a
+		// false failure on a duty that still succeeds.
 		return fmt.Errorf("failed to start consensus: %w", err)
 	}
 
--- a/protocol/v2/ssv/runner/aggregator_committee.go
+++ b/protocol/v2/ssv/runner/aggregator_committee.go
@@ -358,6 +358,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 
 	aggregatorMap, contributionMap, err := r.expectedPreConsensusRoots(ctx, logger)
 	if err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("could not get expected pre-consensus roots: %w", err)
 	}
 
@@ -507,8 +509,10 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 				}
 
 			default:
-				// This should never happen as we build rootToMetadata ourselves with valid roles
-				return fmt.Errorf("unexpected role type in pre-consensus metadata: %v", metadata.Role)
+				// deterministic terminal invariant violation → classify as failed, not stuck
+				err := fmt.Errorf("unexpected role type in pre-consensus metadata: %v", metadata.Role)
+				r.markDutyFailed(err)
+				return err
 			}
 		}
 	}
@@ -536,6 +540,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 	if len(aggregatorSelections) > 0 {
 		// Wait once per duty before fetching aggregate attestations (spec: 2/3 into slot).
 		if err := r.waitTwoThirdsIntoSlot(ctx, duty.DutySlot()); err != nil {
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
 			return err
 		}
 
@@ -582,6 +588,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 
 	// Else, if some aggregators or contributors were selected (even with an error for others), proceed to consensus
 	if err := consensusData.Validate(); err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("invalid aggregator committee consensus data: %w", err)
 	}
 
@@ -593,6 +601,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 		consensusData,
 		r.ValCheck,
 	); err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("failed to start consensus: %w", err)
 	}
 
@@ -788,6 +798,8 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	aggregatorMap, contributionMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 	}
 	if len(beaconObjects) == 0 {
@@ -925,7 +937,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 		for {
 			select {
 			case <-ctx.Done():
-				return ctx.Err()
+				// terminal post-quorum failure → classify as failed, not stuck
+				err := ctx.Err()
+				r.markDutyFailed(err)
+				return err
 			case err := <-errCh:
 				executionErr = err
 			case signatureResult, ok := <-signatureCh:
@@ -972,7 +987,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 					contributionsToSubmit[signatureResult.validatorIndex][root] = signedContrib
 
 				default:
-					return fmt.Errorf("unexpected role type in post-consensus: %v", role)
+					// deterministic terminal invariant violation → classify as failed, not stuck
+					err := fmt.Errorf("unexpected role type in post-consensus: %v", role)
+					r.markDutyFailed(err)
+					return err
 				}
 			}
 		}
@@ -1043,6 +1061,7 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	}
 
 	if executionErr != nil {
+		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later post-consensus message re-enters the submit loops and retries the pending roots (already-submitted roots are skipped via HasSubmitted), so concluding "failed" here would prematurely mask a duty that still completes.
 		return executionErr
 	}
 
--- a/protocol/v2/ssv/runner/aggregator_committee.go
+++ b/protocol/v2/ssv/runner/aggregator_committee.go
@@ -540,8 +540,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 	if len(aggregatorSelections) > 0 {
 		// Wait once per duty before fetching aggregate attestations (spec: 2/3 into slot).
 		if err := r.waitTwoThirdsIntoSlot(ctx, duty.DutySlot()); err != nil {
-			// terminal post-quorum failure → classify as failed, not stuck
-			r.markDutyFailed(err)
+			// Only reachable on shutdown (ctx canceled) within this short wait — markDutyFailed
+			// would drop a context.Canceled reason anyway, so there is nothing to record here.
 			return err
 		}
 
@@ -588,7 +588,6 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 
 	// Else, if some aggregators or contributors were selected (even with an error for others), proceed to consensus
 	if err := consensusData.Validate(); err != nil {
-		// terminal post-quorum failure → classify as failed, not stuck
 		r.markDutyFailed(err)
 		return fmt.Errorf("invalid aggregator committee consensus data: %w", err)
 	}
@@ -601,7 +600,6 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 		consensusData,
 		r.ValCheck,
 	); err != nil {
-		// terminal post-quorum failure → classify as failed, not stuck
 		r.markDutyFailed(err)
 		return fmt.Errorf("failed to start consensus: %w", err)
 	}
@@ -803,6 +801,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 	}
 	if len(beaconObjects) == 0 {
+		// Empty post-quorum (all beacon objects failed to build) is terminal and non-recoverable:
+		// committee_queue drops the message and terminates the runner on this error. Classify as
+		// failed (matching CommitteeRunner) rather than leaving the watcher to report a false stuck.
+		r.markDutyFailed(ErrNoValidDutiesToExecute)
 		return ErrNoValidDutiesToExecute
 	}
 
@@ -937,10 +939,9 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 		for {
 			select {
 			case <-ctx.Done():
-				// terminal post-quorum failure → classify as failed, not stuck
-				err := ctx.Err()
-				r.markDutyFailed(err)
-				return err
+				// Only reachable on shutdown (ctx canceled) — the listener completes in ms and the
+				// duty deadline is ~1 epoch out — and markDutyFailed drops context.Canceled anyway.
+				return ctx.Err()
 			case err := <-errCh:
 				executionErr = err
 			case signatureResult, ok := <-signatureCh:
@@ -1061,7 +1062,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	}
 
 	if executionErr != nil {
-		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later post-consensus message re-enters the submit loops and retries the pending roots (already-submitted roots are skipped via HasSubmitted), so concluding "failed" here would prematurely mask a duty that still completes.
+		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later
+		// post-consensus message re-enters the submit loops and retries the pending roots
+		// (already-submitted roots are skipped via HasSubmitted), so concluding "failed" here
+		// would prematurely mask a duty that still completes.
 		return executionErr
 	}
 
--- a/protocol/v2/ssv/runner/aggregator_committee.go
+++ b/protocol/v2/ssv/runner/aggregator_committee.go
@@ -1072,10 +1072,17 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	}
 
 	if executionErr != nil {
-		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later
-		// post-consensus message re-enters the submit loops and retries the pending roots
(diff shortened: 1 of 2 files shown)


protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -788,6 +798,8 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	aggregatorMap, contributionMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 	}
 	if len(beaconObjects) == 0 {
```
iurii-ssv, line 813: `ErrNoValidDutiesToExecute` (returned just below) is a terminal, non-recoverable post-quorum exit, but it's left unmarked — so the duty-outcome watcher never classifies it.

Unlike the `executionErr` path (which falls through to the committee queue's `default` branch that drops the message yet keeps the runner alive, so it can recover), `ErrNoValidDutiesToExecute` has its own handler in `committee_queue.go` (`case errors.Is(err, runner.ErrNoValidDutiesToExecute)`) that drops the message **and terminates the runner**. With no marker sent, the watcher records no terminal outcome: the duty is either left unrecorded (the scheduler's deferred `cancel()` fires before the watcher's end-of-slot deadline) or reported as a false `stuck` — the phantom signal this PR removes elsewhere.

It's also inconsistent with the sibling `CommitteeRunner`, whose `ProcessPostConsensus` uses a named return + a post-quorum `defer markDutyFailed`, so its identical `return ErrNoValidDutiesToExecute` is recorded as `failed`.

Suggest marking this path — `markDutyNotRequired()` if "nothing valid to submit" is the intended clean outcome, or `markDutyFailed(...)` to match `CommitteeRunner`. (The PR description groups this with the recoverable paths, but it isn't recoverable.)

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -536,6 +540,8 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 	if len(aggregatorSelections) > 0 {
 		// Wait once per duty before fetching aggregate attestations (spec: 2/3 into slot).
 		if err := r.waitTwoThirdsIntoSlot(ctx, duty.DutySlot()); err != nil {
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
```
iurii-ssv: Minor: this mark can't actually record `failed`. `waitTwoThirdsIntoSlot` only returns `ctx.Err()`, and it waits just ~2/3 into the slot, whereas the duty context's deadline is ~1 epoch out (`context.WithDeadline(s.ctx, SlotStartTime(slot + SLOTS_PER_EPOCH + 1))` in `operator/duties/scheduler.go`). So during this short wait the only way `ctx` is done is shutdown → `context.Canceled`, which `markDutyFailed` intentionally drops. The `DeadlineExceeded → failed` case from the PR description isn't reachable here, so this is effectively inert — consider dropping it or adjusting the rationale.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -925,7 +937,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 		for {
 			select {
 			case <-ctx.Done():
-				return ctx.Err()
+				// terminal post-quorum failure → classify as failed, not stuck
+				err := ctx.Err()
+				r.markDutyFailed(err)
```
iurii-ssv: Same as the `waitTwoThirdsIntoSlot` site above: effectively inert. The listener completes in milliseconds (BLS reconstruction) and the duty ctx's deadline is ~1 epoch out, so the only realistic `ctx.Done()` trigger here is shutdown → `context.Canceled`, which `markDutyFailed` drops. Even on a (not-reachable-in-time) `DeadlineExceeded`, the watcher's end-of-slot `stuck` timer would already have concluded the duty, making this a no-op.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -1043,6 +1061,7 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	}
 
 	if executionErr != nil {
+		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later post-consensus message re-enters the submit loops and retries the pending roots (already-submitted roots are skipped via HasSubmitted), so concluding "failed" here would prematurely mask a duty that still completes.
```
iurii-ssv: nit: the reasoning here is correct and worth keeping, but it's a single ~290-char line while the other multi-line comments in this file wrap to ~100 cols — consider wrapping. Also the `// terminal post-quorum failure → classify as failed, not stuck` comment is repeated at 6 sites; a single explanatory comment would cut the duplication.

iurii-ssv (commented):
Looks solid overall, a few notes inline. The main one: the `ErrNoValidDutiesToExecute` path is terminal and non-recoverable (it terminates the runner in `committee_queue.go`) yet left unmarked, so it surfaces as a false `stuck`/unrecorded outcome and is inconsistent with `CommitteeRunner`, which records it as `failed`. The other two are low-severity (two ctx-error marks are effectively inert; one comment-style nit).

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Good catch — you're right that this is terminal and non-recoverable: `committee_queue.go` has a dedicated `ErrNoValidDutiesToExecute` case that drops the message *and* terminates the runner, so leaving it unmarked surfaced as a false `stuck`/unrecorded outcome. I went with `markDutyFailed` (over `markDutyNotRequired`) to stay consistent with `CommitteeRunner`, and because reaching here means we had aggregator selections and post-quorum yet every beacon object failed to build — that's an anomaly, not a clean "nothing to submit" (those cases mark `notRequired` earlier at the pre-consensus / no-aggregator exits). Fixed in 456ab7c.

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Agreed, it's inert — `waitTwoThirdsIntoSlot` only returns `ctx.Err()`, and within that short wait the only trigger is shutdown → `context.Canceled`, which `markDutyFailed` drops. Removed the mark and replaced it with a comment explaining why there's nothing to record here. 456ab7c.

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Same reasoning — ms-long listener, duty ctx deadline ~1 epoch out, so the only realistic trigger is shutdown `context.Canceled` (dropped by `markDutyFailed`). Removed it and reverted to a plain `return ctx.Err()` with a note. 456ab7c.

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Done — wrapped the `executionErr` rationale to ~100 cols, and deduped the repeated `terminal post-quorum failure → classify as failed, not stuck` note down to one occurrence per function (the distinct "invariant violation" comments stay as-is). 456ab7c.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -582,6 +588,7 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 
 	// Else, if some aggregators or contributors were selected (even with an error for others), proceed to consensus
 	if err := consensusData.Validate(); err != nil {
+		r.markDutyFailed(err)
```
iurii-ssv: This `markDutyFailed` (and the `decide()` one just below) isn't actually a terminal site — it sits in a **re-entrant** pre-consensus window, so it can conclude a duty `failed` that then goes on to complete successfully.

`basePreConsensusMsgProcessing` reports `hasNewQuorum` *per root reaching quorum for the first time*, not once per duty. Until consensus actually starts (the `HasStartedConsensus()` guard at the top of `ProcessPreConsensus`), this `Validate()`/`decide()` path is re-entered whenever a *different* validator's selection-proof root quorums on a later message. The `Validate` error also isn't a `RetryableError`, so it lands in the committee-queue `default` branch, which drops the message but **keeps the runner alive** — the same "can recover" property we relied on to justify the `executionErr` exclusion.

And `Validate()` here isn't deterministically terminal: in the selection loop just above, a failed `GetAggregateAttestation` (beacon RPC) `continue`s without appending, so if every selected aggregator's fetch fails, `consensusData` is left empty and `Validate()` returns `AggCommConsensusDataNoValidator` ("no validators assigned").

Concrete sequence — a duty with ≥2 aggregator validators whose selection roots reach quorum on separate messages (possible with divergent operator validator sets / staggered message arrival):
1. msg A: V1's root quorums, `IsAggregator(V1)` is true, but `GetAggregateAttestation` fails transiently → empty `consensusData` → `Validate()` fails → `markDutyFailed` consumes `dutyConcluded` (watcher reports `failed`); message dropped, runner survives.
2. msg B: V2's root quorums, fetch succeeds → `decide()` starts consensus → duty submits → eventually `markDutySucceeded`, but `concludeDuty` is now a no-op since `dutyConcluded == nil`.

Net: a duty that actually submitted to the beacon chain is permanently recorded `failed` with a `⚠️ duty failed` warning — the exact phantom signal this PR removes elsewhere. It's the same recoverability argument made for `executionErr`; it applies here too, just via a transient `GetAggregateAttestation` rather than a transient submit.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -593,6 +600,7 @@ func (r *AggregatorCommitteeRunner) ProcessPreConsensus(
 		consensusData,
 		r.ValCheck,
 	); err != nil {
+		r.markDutyFailed(err)
```
iurii-ssv: Same re-entry concern as the `Validate()` mark above: this is reached on every pre-consensus message that brings a *new* root to quorum (until `HasStartedConsensus()`), and the error isn't retryable so the runner survives the message drop. If `decide()` fails on one new-quorum batch but a later batch starts consensus and the duty completes, the eventual `markDutySucceeded` is dropped because `dutyConcluded` was already consumed here → duty mis-reported as `failed`. `decide()` isn't necessarily deterministic across re-entries either, since each invocation rebuilds `consensusData` from only that invocation's new roots.

iurii-ssv (commented):
Couple more things we probably want to tend to

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Confirmed, and fixed.
You're right this isn't a terminal site. 

Dropped the mark and left a comment explaining why this window is non-terminal. 7d05eca

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Same fix here. 
`decide()` sets `RunningInstance` only after `StartNewInstance` succeeds (runner.go:612), so a failure leaves `HasStartedConsensus()` false and the pre-consensus window stays re-entrant.

A later new-quorum batch can start consensus and complete the duty, at which point the already-consumed `dutyConcluded` makes `markDutySucceeded` a no-op. Non-retryable error → runner survives the message drop, so this is recoverable just like the `Validate()` case above. 

Removed the mark with a matching comment. 7d05eca

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -1043,6 +1072,10 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 	}
 
 	if executionErr != nil {
+		// Not markDutyFailed: a transient submit/reconstruct error can recover — a later
```
iurii-ssv: The `transient submit/reconstruct error can recover` rationale holds for the reconstruct sub-case but not the submit sub-case, so this exclusion looks too broad.

`executionErr` aggregates two different failure kinds:

- **Reconstruct failure** (invalid sigs → `PostConsensusQuorumWithInvalidSignatures`): `FallBackAndVerifyEachSignature` can drop the root below quorum, so a later message can re-establish quorum and re-enter — recoverable, correctly excluded. ✅
- **Submit failure** (`SubmitSignedAggregateSelectionProof` / `SubmitSignedContributionAndProof` erroring, ~L1019 / ~L1049): reconstruction already succeeded, so the `(validator, root)` keeps its post-consensus quorum. `basePartialSigMsgProcessing` (`runner.go`) reports a root only on `!quorumReachedPreviously && hasQuorum` (the first crossing), so a root that kept quorum is never re-reported. A later message then yields `hasQuorum == false` → the early `return nil` fires, the submit loops (`for i, root := range roots`) never revisit that root, and `RecordSubmission` (only called on success) never runs → `HasSubmitted` / `HasSubmittedAllDuties` stay false. ❌

Net: a rejected beacon submission is never retried and the duty never concludes, so the watcher records it as `stuck` — the phantom signal this PR removes elsewhere. `CommitteeRunner`'s post-quorum `defer markDutyFailed` records the identical `executionErr` as `failed`, so the two runners now diverge on arguably the most operationally-important terminal failure. (The deterministic `could not find beacon/ssz object` and `constructSignedAggregateAndProof` paths that also feed `executionErr` are likewise non-recoverable.)

Consider splitting the handling — `markDutyFailed` on the submit/construct/missing-object errors, and keep excluding only the quorum-lost reconstruct case.

iurii-ssv (commented):
One more thing to check out

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Good catch — confirmed and split. You're right the submit sub-case isn't recoverable: reconstruction already succeeded, so the `(validator, root)` keeps its post-consensus quorum, and since `basePartialSigMsgProcessing` reports a root only on its first quorum crossing (`runner.go:534`, `!quorumReachedPreviously && hasQuorum`), a later message never re-reports it → the submit loop never revisits it → false `stuck`. Only the reconstruct case genuinely recovers (`FallBackAndVerifyEachSignature` can drop the root below quorum so a later message re-crosses and re-enters).

So `markDutyFailed` now fires on the submit/construct/missing-object errors, and the exclusion is narrowed to just the `PostConsensusQuorumWithInvalidSignatures` case. Discriminated via `errors.As` + `.Code` rather than `errors.Is` — the spec's `Error.Is` is code-blind (matches any `*spectypes.Error`), so a code sentinel with `errors.Is` would misroute. 58d8206

On the `CommitteeRunner` divergence you flagged: it actually has the inverse problem — its blanket `defer markDutyFailed` marks the *recoverable* invalid-sigs case as failed too (its errCh push is an untagged `fmt.Errorf`). Fixing that separately so this PR stays aggregator-scoped.

iurii-ssv (approved):
One final nice-to-have thing to reconcile: `ProcessPostConsensus` scatters four explicit `markDutyFailed` calls where the sibling [`CommitteeRunner.ProcessPostConsensus`](https://github.com/ssvlabs/ssv/blob/58d82065d9dbde72c5e363a79839f57a69ad65f5/protocol/v2/ssv/runner/committee.go#L491) achieves the same with a single named-return `defer`, armed right after the pre-quorum `return nil` ([`committee.go#L514-L518`](https://github.com/ssvlabs/ssv/blob/58d82065d9dbde72c5e363a79839f57a69ad65f5/protocol/v2/ssv/runner/committee.go#L514-L518)). Three of the four aggregator sites (`#L810`, `#L817`, `#L1003`) are exactly what that `defer` covers automatically; only the recoverable invalid-sigs early-return would need to stay explicit (returned before the defer arms, or by clearing `err`).

The PR body's rejection of "a blanket defer" really only applies to the re-entrant `ProcessPreConsensus` — the post-consensus half could mirror `CommitteeRunner` with a single carve-out. Benefit: the two sibling runners stop diverging on the same concern in the same method, and a future terminal `return` added here can't silently regress to "stuck." Fine to keep as-is if you'd rather not churn it.

(Minor: the PR description says the `waitTwoThirdsIntoSlot`/listener paths "conclude failed on a slot/tick `DeadlineExceeded`", but the code — correctly — does *not* mark them, since that deadline is unreachable during those waits. The code is right; just the description is slightly ahead of it.)

protocol/v2/ssv/runner/aggregator_committee.go
momosh-ssv: Follow-up, two things:

**Regression coverage** added in 64fa002. 
There was no test env driving `AggregatorCommitteeRunner` through post-consensus, so I built one (`newAggregatorCommitteeRunnerEnv`, mirroring `newCommitteeRunnerEnv`) plus a beacon wrapper that injects a submit error, and covered both branches of this split:
- submit failure → duty concluded `failed`
- reconstruct-invalid-sigs (`PostConsensusQuorumWithInvalidSignatures`) → NOT concluded (still recoverable)

**Correcting my note above on the `CommitteeRunner` divergence**: I was half-right. 
Digging in, the error-*code* difference is actually spec-mandated, not a bug: the spectest fixtures expect the committee role to emit `ReconstructSignatureErrorCode` and only the aggregator role to emit `PostConsensusQuorumWithInvalidSignatures` (tagging committee the same way breaks 18 `TestSSVMapping` cases). 

The real issue on that side was behavioral `CommitteeRunner`'s blanket `defer markDutyFailed` concluded the *recoverable* invalid-sigs case as `failed` too (the inverse of this PR).

Fixed separately in #2918: keeps the spec-correct code, just discriminates on it via `errors.As` + `.Code` and drops the blanket defer for explicit per-return marks.

momosh-ssv:
> One final nice-to-have thing to reconcile: ProcessPostConsensus scatters four explicit markDutyFailed calls where the sibling [CommitteeRunner.ProcessPostConsensus](https://github.com/ssvlabs/ssv/blob/58d82065d9dbde72c5e363a79839f57a69ad65f5/protocol/v2/ssv/runner/committee.go#L491)...

Thanks for bearing with me, but I'd lean toward not mirroring the `CommitteeRunner` defer here. Counterintuitively, that's the direction that removes the divergence.

#2918 actually removes that blanket defer `markDutyFailed` from `CommitteeRunner`, because it concluded the recoverable `PostConsensusQuorumWithInvalidSignatures` case as failed (the inverse bug of this PR). 

The aggregator has the same carve-out. So adopting the defer here would converge onto a pattern we're deleting from the sibling for being error-prone around exactly this case.
Once #2918 lands, both runners converge on explicit per-return marks.

Happy to revisit, but my inclination is to leave it explicit. And you're right on the description nit, I'll tighten the PR body to match the code.

iurii-ssv:
Agreed — keeping the explicit per-return marks is the right call here. With #2918 dropping the blanket defer on the committee side, explicit marks are the pattern the two runners converge on, and the marks + narrow invalid-sigs carve-out in this PR are already exactly that.

One optional, non-blocking thought: a one-line invariant note at the top of the post-quorum section (e.g. "every terminal post-quorum return must `markDutyFailed`; recoverable returns must document why") would cheaply guard the residual risk the defer idea was aimed at — a future terminal `return` silently regressing to a false `stuck`. Take it or leave it.

Nothing outstanding from my side — thanks for the thorough back-and-forth on this one.

iurii-ssv:
Correcting my note above, since the #2918 side evolved right after I posted it: with the reconstruct error now tagged recoverable at the `errCh` push site (a local sentinel wrapping everything the reconstruct goroutine pushes, which is post-fallback and recoverable by construction), the receive-side ambiguity that made the blanket defer error-prone is gone — and `CommitteeRunner` is collapsing its explicit marks back into a single sentinel-filtered defer after all. So "explicit marks are the pattern the two runners converge on" no longer describes where the committee side landed.

That doesn't change anything for this PR: the explicit marks here are behaviorally correct as-is, and unifying the aggregator runners onto the same push-site tagging (+ the filtered-defer treatment) is already scoped in #2919 — which is also where the sibling-divergence concern this thread started with ultimately gets resolved.

PR #2913, opened 2026-06-29
chore: comment/build cleanups + fmt.Appendf (#2874, #2902 Bucket D)

### LINKED ISSUE #2902: Track: boole-fork ⇄ stage reconciliation backlog (follow-up to #2899)

momosh-ssv:
Low-risk cleanups, **no behavior change** — closes #2902 Bucket D and the #2874 cosmetic item.

- **qbft/roundtimer** — rewrite the stale `RoundTimeout` doc (described an old base/additional-timeout scheme that no longer exists).
- **qbft/instance** — correct the `bumpToRound` comment (it no longer sends round metrics; it just sets the round).
- **spectest build quirk** — rename `committee_msg_processing_type.go` → `committee_msg_processing_type_test.go`. Its types (`CommitteeSpecTest`/`MultiCommitteeSpecTest`) are referenced only from tests, but as a *non-test* file it referenced the test-only `fixRunnerForRun`, so `go build ./...` failed on the package. **`go build ./...` is now clean** (this was the long-standing pre-existing build error noted across the recent PRs).
- **#2874** — replace `[]byte(fmt.Sprintf(...))` with `fmt.Appendf(nil, ...)` (35 occurrences across 13 files; mostly tests + `utils/boot_node/node.go`). Behavior-identical; applied to boole's actual occurrences (not stage's literal file list, since some files differ).

`go build ./...` clean · `go test`/`vet`/`gofmt` green on touched packages.


iurii-ssv (approved):
LGTM

PR #2914, opened 2026-06-29
fix(dutytracer): join collector + slot-pruning before DB close on shutdown (#2884)

momosh-ssv:
Bucket B of #2902 — ports stage #2884 onto `boole-fork` (defensive: harmless if boole doesn't run exporter/archive mode, a real crash-fix if it does).

**Bug:** the duty-trace collector and the slot-pruner spawn goroutines that write to the pebble DB. On shutdown the DB was closed **without waiting** for them, so an in-flight evict/schedule/late-retry write (collector, archive mode) or prune (slot-pruner, standard mode) could hit a closed DB → **panic (`pebble: closed`)**.

**Two-part fix** (the in-package drain is *inert* without the caller-side join — caught in deep review):
- **`operator/dutytracer/collector.go`** — track the detached late-retry goroutines (`lateWG`/`lateMu`/`lateClosed`); `Start` joins the schedule filler/worker then the late retries on `ctx.Done()` before returning; `collectLateAsync` refuses new spawns once shutdown began. Deadlock-free (the late path never enqueues to `scheduleJobs`); no `WaitGroup` Add-during-Wait race.
- **`cli/operator/node.go`** — actually **join** those lifetimes before `db.Close()`: the collector goroutine is made joinable (`collectorDone`) and `initSlotPruning` returns a join func; both are joined **inline** after `operatorNode.Start` returns, before the deferred `db.Close()` fires. This also closes the **standard-mode** `PruneContinuously` race (upstream #2884's first half).

**Review:** deep-reviewed for concurrency/shutdown correctness — **APPROVE**. Verified: joins run before `db.Close()` (inline, not deferred); no shutdown hang (the collector/pruner use `cmd.Context()`, provably Done by the join point in both modes); no `WaitGroup` Add-during-Wait race; `initSlotPruning` signature change has one caller + nil-guards. `go test -race ./operator/dutytracer/...` clean. The bundled p2p flaky-test fix was not ported (boole's test diverges).

Note: `go build ./...` has the pre-existing `spectest fixRunnerForRun` quirk on this branch's base — fixed by #2913.

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -511,10 +511,13 @@ var StartNodeCmd = &cobra.Command{
 			})
 		}
 
+		// joinSlotPruning waits for all PruneContinuously goroutines to return.
+		// It is non-nil only in exporter-standard mode; must be called before db.Close().
+		var joinSlotPruning func()
 		if cfg.ExporterOptions.Enabled && cfg.ExporterOptions.Mode == exporter.ModeStandard {
 			retain := cfg.ExporterOptions.RetainSlots
 			threshold := cfg.SSVOptions.NetworkConfig.EstimatedCurrentSlot()
-			initSlotPruning(cmd.Context(), storageMap, slotTickerProvider, threshold, retain)
+			joinSlotPruning = initSlotPruning(cmd.Context(), storageMap, slotTickerProvider, threshold, retain)
 		}
 
 		cfg.SSVOptions.ValidatorOptions.StorageMap = storageMap
@@ -544,6 +547,9 @@ var StartNodeCmd = &cobra.Command{
 
 		// Exporter duty tracing
 		var collector *dutytracer.Collector
+		// collectorDone is closed when the collector goroutine returns; non-nil only
+		// in archive mode. Joined before db.Close() to prevent writes to a closed DB.
+		var collectorDone chan struct{}
 		if cfg.ExporterOptions.Enabled {
 			switch cfg.ExporterOptions.Mode {
 			case exporter.ModeArchive:
@@ -556,7 +562,11 @@ var StartNodeCmd = &cobra.Command{
 					dstore, networkConfig.Beacon, decidedStreamPublisherFn,
 					dutyStore)
 
-				go collector.Start(cmd.Context(), slotTickerProvider)
+				collectorDone = make(chan struct{})
+				go func() {
+					defer close(collectorDone)
+					collector.Start(cmd.Context(), slotTickerProvider)
+				}()
 				cfg.SSVOptions.ValidatorOptions.DutyTraceCollector = collector
 				cfg.SSVOptions.ExporterRead = exporter2.NewExporter(
 					logger,
@@ -719,6 +729,16 @@ var StartNodeCmd = &cobra.Command{
 		}
 		if err := operatorNode.Start(cfg.SSVOptions.Context); err != nil {
 			logger.Fatal("failed to start SSV node", zap.Error(err))
+		}
+
+		// Join background goroutines before the deferred db.Close() runs.
+		// Both collector.Start and PruneContinuously return promptly on ctx.Done(),
+		// so these joins complete quickly after the shutdown signal is received.
+		if collectorDone != nil {
+			<-collectorDone
+		}
+		if joinSlotPruning != nil {
+			joinSlotPruning()
 		}
 	},
 }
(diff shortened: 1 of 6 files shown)


operator/dutytracer/collector.go
```diff
@@ -654,12 +684,36 @@ func (c *Collector) computeRoleRoots(ctx context.Context, slot phase0.Slot, in [
 func (c *Collector) Collect(ctx context.Context, msg *queue.SSVMessage, verifySig func(*spectypes.PartialSignatureMessages) error) error {
 	err := c.collect(ctx, msg, verifySig)
 	if errors.Is(err, errInFlight) {
-		go c.collectLateMessage(ctx, msg, verifySig)
+		c.collectLateAsync(ctx, msg, verifySig)
```
iurii-ssv, line 687: This now drains the **async** late-retry (`collectLateAsync` -> `lateWG`), but that path only runs for the `errInFlight` case — a *second, concurrent* late message for the same key. The **first/primary** late message is handled by the synchronous `c.collect(...)` on the line above (685), which does the DB work itself:

- `getOrCreate{Committee,Validator}Trace` reads from disk when `slot <= lastEvictedSlot` (collector.go:215 / :270), and
- when `late`, writes via `c.store.SaveCommitteeDuty` / `SaveValidatorDuty` (collector.go:949 / :1009 / :1075 / :1117) -> `DutyTraceStore` -> `pdb.DB.Set` (storage/pebble/pebble.go:43), which panics `pebble: closed` after `db.Close()`.

`Collect()` runs on the message-worker goroutine (`handleWorkerMessages`, operator/validator/controller.go:400), spawned by `go w.startWorker` (protocol/v2/queue/worker/message_worker.go:65) and never joined. The new `collectorDone` join in cli/operator/node.go doesn't cover it, so if the worker is mid-`collect` on a late message when ctx cancels, the synchronous `Save*` can land after `db.Close()` — the exact panic this PR fixes. Note the async retry is the *less* common path; the synchronous write here is the one left unprotected.

cli/operator/node.go
```diff
@@ -720,6 +730,16 @@ var StartNodeCmd = &cobra.Command{
...
+
+		// Join background goroutines before the deferred db.Close() runs.
+		// Both collector.Start and PruneContinuously return promptly on ctx.Done(),
+		// so these joins complete quickly after the shutdown signal is received.
+		if collectorDone != nil {
+			<-collectorDone
+		}
+		if joinSlotPruning != nil {
+			joinSlotPruning()
+		}
```
iurii-ssv, line 748: These joins cover the collector's *own* goroutines (evict loop, schedule filler/worker, late-retry) and the slot-pruner — but not the validator controller's message worker, which also writes to the pebble DB synchronously via `Collect()` -> `collect()` -> `c.store.Save*` for late messages (see the comment on `Collect` in operator/dutytracer/collector.go).

`worker.Close()` only cancels its ctx with no wait (protocol/v2/queue/worker/message_worker.go:106) and isn't called here, and the router/worker goroutines started by `StartNetworkHandlers` (`go w.startWorker`, message_worker.go:65; `go c.handleRouterMessages`, operator/validator/controller.go:543) are bare `go` with no join. So an in-flight late write on the worker can still race this deferred `db.Close()` -> `pebble: closed`.

If this isn't handled by another bucket, the consistent fix is to quiesce the worker here before `db.Close()`: have `worker.Close()` wait for the in-flight `process` (a `WaitGroup` around `startWorker`) and call it (plus join the router handlers) after `operatorNode.Start` returns, alongside these joins.

iurii-ssv (commented):
Seems there is 1 residual gap: the **primary** late-write path into the duty-trace store is the *synchronous* `Collect()` call running on the validator controller's message-worker goroutine, and that goroutine is never joined before `db.Close()`. So in exporter **archive** mode the same `pebble: closed` panic this PR targets is still reachable on shutdown. See the comments below for details.

operator/dutytracer/collector.go
momosh-ssv: Agreed — the async path was the *less* common one, and the synchronous `c.collect(...)` above it (the `getOrCreate*` reads + late `Save*` writes) was the real exposure. Rather than push more drain logic into the collector, I quiesced the goroutine that actually runs `Collect`: the message worker. As of d8fe6994f it tracks its `startWorker` goroutines with a `WaitGroup`, and `Worker.Close()` cancels then `Wait()`s for any in-flight `process()` to return. So by the time we close the DB, no goroutine can be mid-`collect`/`Save*`.

cli/operator/node.go
momosh-ssv: Done in d8fe6994f, along the lines you suggested. Added `Controller.StopNetworkHandlers()` (→ `messageWorker.Close()`), and `node.go` now calls it *before* the `collectorDone`/`joinSlotPruning` joins and the deferred `db.Close()`. `Worker.Close()` waits for in-flight `process()` via a `WaitGroup` (I left the queue open on purpose — closing it would let `startWorker` read a typed-nil from the closed channel and hand it to the handler). Stopping the worker first also freezes new late-retry spawns before the collector drains its `lateWG`, so the two interlock cleanly.

On the router goroutines (`handleRouterMessages`): I left those unjoined deliberately — they stop on `c.ctx` cancellation and only enqueue, never touch the DB, so they aren't part of the write-after-close hazard. Happy to add an explicit join too if you'd prefer the symmetry.

iurii-ssv (approved):
Looks good!

PR #2918, opened 2026-07-01
observability(runner): classify CommitteeRunner terminal post-quorum failures as failed, not stuck

momosh-ssv:
## What

`CommitteeRunner.ProcessPostConsensus` used a blanket `defer { if err != nil { markDutyFailed(err) } }` on its named return, which concluded the duty **failed** on *any* non-nil return — including the genuinely recoverable reconstruct-invalid-signatures case.

`FallBackAndVerifyEachSignature` can drop a root below quorum, so a later partial-signature message re-crosses quorum and re-enters the loop and the duty completes. But the blanket defer had already consumed the single-shot `dutyConcluded` channel, so the eventual `markDutySucceeded` became a no-op — and a duty that **actually completed** was mis-reported as `failed`. This is the inverse of the false-`stuck` signal fixed for `AggregatorCommitteeRunner` in #2912.

## Fix

- Remove the blanket defer.
- Add explicit `r.markDutyFailed(...)` before **every** terminal post-quorum error return (expected-roots error, `ErrNoValidDutiesToExecute`, `ctx.Done()`, `SubmitAttestations`, the `currentDutySlot`/`Data` failures, `SubmitSyncMessages`).
- Exclude **only** the recoverable reconstruct-invalid-sigs case, discriminated via `errors.As` + `.Code` (the spec's `Error.Is` is code-blind, so `errors.Is` against a code would misroute).

### Note on the error code
The reconstruct error keeps its spec-mandated `ReconstructSignatureErrorCode` — the `errCh` push is left **untagged**. The spec fixtures expect the committee role to emit `ReconstructSignatureErrorCode` and only the *aggregator* role to emit `PostConsensusQuorumWithInvalidSignatures`, so the two runners discriminate on different codes by design. (An initial attempt to tag it like the aggregator broke 18 `TestSSVMapping` spectests — caught and reverted.)

## Testing
- `go build` / `go vet` / `gofmt` clean
- `go test ./protocol/v2/ssv/runner/...` ✅
- `go test ./protocol/v2/ssv/spectest/...` ✅ (all `TestSSVMapping` pass)

Companion to #2912 (same behavioral fix on `AggregatorCommitteeRunner`); both target `boole-fork`.

### DIFF
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -511,9 +511,10 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	attestationMap, committeeMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
+		err = fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 		// terminal post-quorum failure → classify as failed, not stuck
 		r.markDutyFailed(err)
-		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
+		return err
 	}
 	if len(beaconObjects) == 0 {
 		// terminal post-quorum failure → classify as failed, not stuck
@@ -524,7 +525,7 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 	attestationsToSubmit := make(map[phase0.ValidatorIndex]*spec.VersionedAttestation)
 	syncCommitteeMessagesToSubmit := make(map[phase0.ValidatorIndex]*altair.SyncCommitteeMessage)
 
-	var executionErr error
+	var recoverableErr, terminalErr error
 
 	span.SetAttributes(observability.BeaconBlockRootCountAttribute(len(roots)))
 	// For each root that got at least one quorum, find the duties associated to it and try to submit
@@ -646,7 +647,14 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 				r.markDutyFailed(err)
 				return err
 			case err := <-errCh:
-				executionErr = err
+				// errCh currently carries only the reconstruct-invalid-sigs error, which is recoverable;
+				// classify defensively by code so any other error arriving here is still treated as terminal.
+				var specErr *spectypes.Error
+				if errors.As(err, &specErr) && specErr.Code == spectypes.ReconstructSignatureErrorCode {
+					recoverableErr = err
+				} else {
+					terminalErr = err
+				}
 			case signatureResult, ok := <-signatureCh:
 				if !ok {
 					break listener
@@ -654,12 +662,12 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 
 				validatorObjects, exists := beaconObjects[signatureResult.validatorIndex]
 				if !exists {
-					executionErr = fmt.Errorf("could not find beacon object for validator index: %d", signatureResult.validatorIndex)
+					terminalErr = fmt.Errorf("could not find beacon object for validator index: %d", signatureResult.validatorIndex)
 					continue
 				}
 				sszObject, exists := validatorObjects[root]
 				if !exists {
-					executionErr = fmt.Errorf("could not find ssz object for root: %s", root)
+					terminalErr = fmt.Errorf("could not find ssz object for root: %s", root)
 					continue
 				}
 
@@ -678,7 +686,7 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 					att := sszObject.(*spec.VersionedAttestation)
 					att, err = specssv.VersionedAttestationWithSignature(att, signatureResult.signature)
 					if err != nil {
-						executionErr = fmt.Errorf("could not insert signature in versioned attestation")
+						terminalErr = fmt.Errorf("could not insert signature in versioned attestation")
 						continue
 					}
 
@@ -825,19 +833,23 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 		}
 	}
 
-	if executionErr != nil {
+	// A terminal error on any root wins over a recoverable one. executionErr used to be a single
+	// last-write-wins variable, so when a genuine failure and a recoverable reconstruct error happened
+	// in the same round, classification depended on goroutine/root ordering. Splitting the two keeps it
+	// deterministic — a real failure is always recorded as failed, never mislabeled as recoverable/stuck.
+	if terminalErr != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(terminalErr)
+		return terminalErr
+	}
+	if recoverableErr != nil {
 		// Reconstruct-invalid-sigs is recoverable: FallBackAndVerifyEachSignature can drop the root
 		// below quorum, so a later partial-sig message re-crosses quorum and re-enters this loop to
 		// retry pending roots (already-submitted roots are skipped via HasSubmitted). ReconstructBeaconSig
-		// tags this with ReconstructSignatureErrorCode (spec expects code 32 for the committee role; the
-		// aggregator role uses PostConsensusQuorumWithInvalidSignatures instead). Not markDutyFailed.
-		var specErr *spectypes.Error
-		if errors.As(executionErr, &specErr) && specErr.Code == spectypes.ReconstructSignatureErrorCode {
-			return executionErr
-		}
-		// terminal post-quorum failure → classify as failed, not stuck
-		r.markDutyFailed(executionErr)
-		return executionErr
+		// tags this with ReconstructSignatureErrorCode (code 32 in the pinned ssv-spec). Not markDutyFailed.
+		// Note: the post-Boole AggregatorCommitteeRunner tags the analogous condition with
+		// PostConsensusQuorumWithInvalidSignatures (code 81) instead.
+		return recoverableErr
 	}
 
 	if r.HasSubmittedAllValidatorDuties(attestationMap, committeeMap) {
--- a/protocol/v2/ssv/runner/committee.go
+++ b/protocol/v2/ssv/runner/committee.go
@@ -505,27 +505,19 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 		return nil
 	}
 
-	// We have quorum and are committed to submitting. Pre-quorum waiting, full success
-	// (markDutySucceeded) and partial progress all return nil, so this only fires on a terminal
-	// post-quorum error — report it as failed instead of letting it fall through to a false "stuck".
-	// Unlike the consensus phase, a failure here is final: submission is the duty's last step.
-	// Shutdown (context cancellation) needs no special-casing — markDutyFailed drops a context.Canceled
-	// reason, so a submission aborted by shutdown isn't recorded as a failure.
-	defer func() {
-		if err != nil {
-			r.markDutyFailed(err)
-		}
-	}()
-
 	r.measurements.EndPostConsensus()
 	recordPostConsensusDuration(ctx, r.measurements.PostConsensusTime(), spectypes.RoleCommittee)
 
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	attestationMap, committeeMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 	}
 	if len(beaconObjects) == 0 {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(ErrNoValidDutiesToExecute)
 		return ErrNoValidDutiesToExecute
 	}
 
@@ -648,7 +640,11 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 		for {
 			select {
 			case <-ctx.Done():
-				return ctx.Err()
+				// terminal post-quorum failure → classify as failed, not stuck; markDutyFailed
+				// drops context.Canceled internally, so shutdown isn't recorded as a failure.
+				err := ctx.Err()
+				r.markDutyFailed(err)
+				return err
 			case err := <-errCh:
 				executionErr = err
 			case signatureResult, ok := <-signatureCh:
@@ -718,17 +714,26 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 			recordFailedSubmission(ctx, spectypes.BNRoleAttester)
 			const errMsg = "could not submit attestations"
 			aLogger.Error(errMsg, zap.Error(err))
-			return fmt.Errorf("%s: %w", errMsg, err)
+			err = fmt.Errorf("%s: %w", errMsg, err)
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
+			return err
 		}
 
 		currentDutySlot, err := r.currentDutySlot()
 		if err != nil {
-			return fmt.Errorf("current duty slot: %w", err)
+			err = fmt.Errorf("current duty slot: %w", err)
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
+			return err
 		}
 		recordSuccessfulSubmission(ctx, int64(len(attestations)), r.NetworkConfig.EstimatedEpochAtSlot(currentDutySlot), spectypes.BNRoleAttester)
 		attData, err := attestations[0].Data()
 		if err != nil {
-			return fmt.Errorf("could not get attestation data: %w", err)
+			err = fmt.Errorf("could not get attestation data: %w", err)
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
+			return err
 		}
 		const eventMsg = "✅ successfully submitted attestations"
 		span.AddEvent(eventMsg, trace.WithAttributes(
@@ -769,14 +774,20 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 			recordFailedSubmission(ctx, spectypes.BNRoleSyncCommittee)
 			const errMsg = "could not submit sync committee messages"
 			scLogger.Error(errMsg, zap.Error(err))
-			return fmt.Errorf("%s: %w", errMsg, err)
+			err = fmt.Errorf("%s: %w", errMsg, err)
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
+			return err
 		}
 
 		syncMsgsCount := len(syncCommitteeMessages)
 		if syncMsgsCount <= math.MaxUint32 {
 			currentDutySlot, err := r.currentDutySlot()
 			if err != nil {
-				return fmt.Errorf("current duty slot: %w", err)
+				err = fmt.Errorf("current duty slot: %w", err)
+				// terminal post-quorum failure → classify as failed, not stuck
+				r.markDutyFailed(err)
+				return err
 			}
 			recordSuccessfulSubmission(
 				ctx,
@@ -788,7 +799,10 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 
 		currentDutySlot, err := r.currentDutySlot()
 		if err != nil {
-			return fmt.Errorf("current duty slot: %w", err)
+			err = fmt.Errorf("current duty slot: %w", err)
+			// terminal post-quorum failure → classify as failed, not stuck
+			r.markDutyFailed(err)
+			return err
 		}
 		const eventMsg = "✅ successfully submitted sync committee"
 		span.AddEvent(eventMsg, trace.WithAttributes(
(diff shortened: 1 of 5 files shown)


protocol/v2/ssv/runner/committee.go
```diff
@@ -812,6 +826,17 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
 	}
 
 	if executionErr != nil {
+		// Reconstruct-invalid-sigs is recoverable: FallBackAndVerifyEachSignature can drop the root
+		// below quorum, so a later partial-sig message re-crosses quorum and re-enters this loop to
+		// retry pending roots (already-submitted roots are skipped via HasSubmitted). ReconstructBeaconSig
+		// tags this with ReconstructSignatureErrorCode (spec expects code 32 for the committee role; the
+		// aggregator role uses PostConsensusQuorumWithInvalidSignatures instead). Not markDutyFailed.
```
iurii-ssv: This comment is factually wrong on two points:

1. `ReconstructSignatureErrorCode` is **34**, not 32. (The code uses the named constant, so runtime behavior is correct — only the number in the comment is wrong.)
2. `PostConsensusQuorumWithInvalidSignatures` exists nowhere in ssv-spec (v1.2.2) or this repo. `AggregatorRunner.ProcessPostConsensus` calls the **same** `ReconstructBeaconSig`, which yields the **same** `ReconstructSignatureErrorCode` — the two runners do not "discriminate on different codes by design."

Because of this, the aggregator still wraps the identical case in its blanket `defer markDutyFailed` with no reconstruct exception ([aggregator.go#L316-L333](https://github.com/ssvlabs/ssv/blob/9d80e2f2f52dab389929d607edaf8af75f77a812/protocol/v2/ssv/runner/aggregator.go#L316-L333)), so its recoverable reconstruct-invalid-signatures path is still mis-reported as `failed` — the very bug this PR fixes for the committee. Suggest (a) correcting the comment, and (b) applying the same fix to the aggregator, ideally via a shared `isRecoverableReconstructError` helper on `BaseRunner` so the two runners stay consistent.

protocol/v2/ssv/runner/committee.go
```diff
@@ -812,6 +826,17 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
+		var specErr *spectypes.Error
+		if errors.As(executionErr, &specErr) && specErr.Code == spectypes.ReconstructSignatureErrorCode {
```
iurii-ssv: `executionErr` retains only the **last** error written by the listener loop across all roots. If, in the same post-consensus round, one root yields the code-tagged reconstruct error (34) and another yields a genuinely non-recoverable `executionErr` (e.g. `could not find beacon object for validator index` or `could not insert signature in versioned attestation`), the classification here — recoverable (return, unconcluded) vs `markDutyFailed` — depends on non-deterministic goroutine/select ordering.

If the code-tagged error happens to be written last, `errors.As` matches and a real non-recoverable failure is silently not recorded as failed (later surfacing as a deadline/"stuck" mislabel). The removed blanket defer classified all such cases deterministically as failed. Low severity, but worth considering treating the reconstruct error as recoverable only when it is the sole error type.

iurii-ssv (commented):
A couple of observations on the reconstruct-error handling. The core fix (removing the blanket defer, explicit per-site `markDutyFailed`, and excepting the recoverable reconstruct case) is mechanically sound and the `errors.As` traversal reaches the code-tagged `*spectypes.Error` correctly. Comments below are about an incorrect code comment (which also points to a latent same-bug in the aggregator) and one classification edge.

protocol/v2/ssv/runner/committee.go
```diff
@@ -505,27 +505,19 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
 	r.measurements.EndPostConsensus()
 	recordPostConsensusDuration(ctx, r.measurements.PostConsensusTime(), spectypes.RoleCommittee)
 
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	attestationMap, committeeMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
 		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
 	}
```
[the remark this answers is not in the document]
momosh-ssv: Good catch — fixed in `657721c4b`. The error is now wrapped first, and the wrapped value is passed to both `markDutyFailed` and the `return`, so monitoring keeps the full "could not get expected post consensus roots and beacon objects" context.

protocol/v2/ssv/runner/committee.go
momosh-ssv: Thanks, you're right that the comment was misleading, clarified in `657721c4b`. 

One thing worth flagging on the specifics: this looks to be a spec-version mismatch. We pin `ssv-spec v1.2.3-0.20260305184636`, where `ReconstructSignatureErrorCode` is `32` (it was `34` back in v1.2.2), and `PostConsensusQuorumWithInvalidSignatures` does exist there (code `81`, used by `AggregatorCommitteeRunner`). So the `32` in the comment was actually correct for our pinned spec — but I've reworded it to stop conflating the two aggregator runners.

On the aggregator fix itself I agreed, and I've split it out into #2919 to keep this PR scoped to `CommitteeRunner`. 

Two distinct gaps there: the legacy `AggregatorRunner` has the identical blanket-`defer markDutyFailed` bug on the same code-32 path (though it's only wired pre-Boole), while the live post-Boole `AggregatorCommitteeRunner` never calls `markDutyFailed` at all; so it needs the shared-helper treatment you suggested (handling both code 32 and 81).

protocol/v2/ssv/runner/committee.go
momosh-ssv: Agreed, this was a real ordering hazard. 
Fixed in `657721c4b` by splitting the single `executionErr` into `terminalErr` / `recoverableErr`.

A genuine terminal error on any root now always wins over a recoverable reconstruct error, so classification no longer depends on goroutine/root ordering. The recoverable path is gated on `ReconstructSignatureErrorCode` at receipt, and anything else off `errCh` is treated as terminal.

protocol/v2/ssv/runner/committee.go
```diff
@@ -648,22 +641,33 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
 			case err := <-errCh:
-				executionErr = err
+				// errCh currently carries only the reconstruct-invalid-sigs error, which is recoverable;
+				// classify defensively by code so any other error arriving here is still treated as terminal.
+				var specErr *spectypes.Error
+				if errors.As(err, &specErr) && specErr.Code == spectypes.ReconstructSignatureErrorCode {
+					recoverableErr = err
+				} else {
+					terminalErr = err
+				}
```
iurii-ssv: The defensive default here inverts recoverability for a real subclass: `ReconstructBeaconSig` can fail **before** the coded check ever runs. Only `types.VerifyReconstructedSignature` attaches `ReconstructSignatureErrorCode`; the earlier `types.ReconstructSignatures` step (BLS point `Deserialize`/`Recover`) returns a plain wrapped error with **no spec code**. Message validation checks partial-sig structure but not point-deserializability, so 96 garbage bytes from a byzantine/buggy operator take exactly that path.

Sequence: garbage partial sig → quorum forms → reconstruct fails with an *uncoded* error → classified `terminalErr` → `markDutyFailed` consumes the single-shot `dutyConcluded`. Meanwhile the goroutine has already run `FallBackAndVerifyEachSignature` (it runs on *any* reconstruct error), which removes the bad sig; a later partial-sig re-crosses quorum, the duty completes, and `markDutySucceeded` is a no-op → false `failed` — the exact bug class this PR fixes. Note the asymmetry: the easiest byzantine input (garbage bytes failing `Deserialize`) hits the uncoded path, while only a well-formed-but-wrong point reaches the coded verify error — yet both are recovered identically by the fallback.

Suggest classifying at the push site instead: everything the reconstruct goroutine sends to `errCh` is by construction post-fallback and recoverable. A local, non-spec sentinel avoids the spectest breakage described in the PR body, because it preserves both the message text and (via `Unwrap`) the reachability of the code-tagged `*spectypes.Error`, so the committee role still observably emits code 32:

```go
type recoverableReconstructError struct{ err error }

func (e recoverableReconstructError) Error() string { return e.err.Error() }
func (e recoverableReconstructError) Unwrap() error { return e.err }
```

Push: `errCh <- recoverableReconstructError{fmt.Errorf("%s: %w", eventMsg, err)}`; receive: `errors.As` on the sentinel. Future producers that push without the sentinel still default to terminal, so the defensive property is kept — and the comment above ("errCh currently carries only the reconstruct-invalid-sigs error") becomes accurate, whereas today deserialize/recover failures flow through it too.

Side note for #2919: the `isRecoverableReconstructError` helper suggested there discriminates by spec codes (32/81) and would inherit this same blind spot into all three runners — worth switching that plan to push-site classification as well.

protocol/v2/ssv/runner/committee.go
```diff
@@ -505,34 +505,27 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
 	recordPostConsensusDuration(ctx, r.measurements.PostConsensusTime(), spectypes.RoleCommittee)
 
 	// Get validator-root maps for attestations and sync committees, and the root-beacon object map
 	attestationMap, committeeMap, beaconObjects, err := r.expectedPostConsensusRootsAndBeaconObjects(ctx, logger)
 	if err != nil {
-		return fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
+		err = fmt.Errorf("could not get expected post consensus roots and beacon objects: %w", err)
+		// terminal post-quorum failure → classify as failed, not stuck
+		r.markDutyFailed(err)
+		return err
```
iurii-ssv: Structural suggestion (behavior-preserving): with the recoverable case now discriminable, the nine explicit `markDutyFailed` sites — each carrying the same one-line comment — collapse back into the original single defer plus one exclusion:

```go
defer func() {
	var rec recoverableReconstructError // or the errors.As + .Code check
	if err != nil && !errors.As(err, &rec) {
		r.markDutyFailed(err)
	}
}()
```

Checked equivalence path-by-path: the recoverable return carries its discriminator (excluded); the `ctx.Done()` return still flows through `markDutyFailed`'s internal `context.Canceled` filter (as the removed comment used to note); pre-quorum returns precede the defer; and the `terminalErr`/`recoverableErr` split — which is a genuine improvement — stays as is.

Besides ~30 fewer lines, the main value is future-proofing: any error return later added to this ~390-line function gets classified automatically instead of silently regressing to a false "stuck" (the bug class now being fixed for the second time, counting the aggregator follow-up). The tripled `currentDutySlot` error boilerplate in the sync-committee section collapses along with it.

This also dovetails with #2919, which already plans a shared recoverability helper plus a committee refactor to use it — landing the predicate here means the follow-up only has to extend it to the aggregator runners.

iurii-ssv (commented):
Two follow-ups on the reconstruct-error classification (inline below): the receive-side code gate misses uncoded reconstruct failures (BLS point deserialize/recover errors carry no spec code), and a structural suggestion to fold the per-site `markDutyFailed` calls back into a single defer now that the recoverable case is discriminable.

Would also be good to pin the fix with a regression test: a quorum containing one non-deserializable partial signature must not conclude the duty `failed`, and a subsequent valid quorum must still report `succeeded`.

protocol/v2/ssv/runner/committee.go
momosh-ssv: You're right, and this is the more important of the two.
Thanks for tracing it all the way through. Confirmed against the pinned spec: `ReconstructSignature` fails uncoded at the nil-checks and at `ReconstructSignatures` (BLS `Deserialize`/`Recover`), and only `VerifyReconstructedSignature` attaches `ReconstructSignatureErrorCode`. 

So the easiest byzantine input (96 garbage bytes) took the uncoded path → classified `terminalErr` → `markDutyFailed`, even though the goroutine had already run `FallBackAndVerifyEachSignature` and the duty could still recover. Exactly the bug class this PR set out to fix.

Fixed in a3d22d031 by switching to push-site classification, as you suggested: a `recoverableReconstructError` sentinel wraps the error where it's pushed onto `errCh` (post-fallback, recoverable by construction). 
The receive site now discriminates on the tag via `errors.As`, so both coded and uncoded reconstruct failures are covered, while anything pushed without the tag still defaults to terminal. `Unwrap` keeps the wrapped chain (incl. any code-tagged `*spectypes.Error`) reachable, so the committee role still observably emits the code and the spectests stay green.

Good call on #2919 too! 
I'll switch that plan to push-site classification rather than discriminating by spec codes (32/81), so the shared helper doesn't inherit the same blind spot.

protocol/v2/ssv/runner/committee.go
momosh-ssv: Agreed, with the recoverable case now discriminable by tag, the nine explicit `markDutyFailed` sites collapse cleanly. Done in a3d22d031.

I restored the single defer right after the quorum check (as in the original), so the two pre-quorum returns still return `nil` untouched, and added the one exclusion:

```go
defer func() {
	if err != nil && !isRecoverableReconstructError(err) {
		r.markDutyFailed(err)
	}
}()
```

Checked the paths match your equivalence analysis: `ctx.Done()` still flows through `markDutyFailed`'s internal `context.Canceled` filter; the recoverable return carries its tag and is excluded; the tripled `currentDutySlot` boilerplate collapses too. 

The `terminalErr`/`recoverableErr` split stays as-is — terminal still wins deterministically, and both just flow into the defer now. Now any error return added here later gets classified automatically instead of regressing to a false stuck.

protocol/v2/ssv/runner/committee.go
iurii-ssv: Heads-up — the fix commit these replies reference doesn't seem to have reached the remote yet: the PR head here is still the pre-review one, and GitHub can't resolve the cited SHA anywhere in the repo. Mind checking the push went through? I'll verify the sentinel + defer changes once they're visible.

One thing to fold in while it's pending, since the replies don't mention it: the regression test from the review summary — a quorum containing one non-deserializable partial signature must not conclude the duty `failed`, and a subsequent valid quorum must still conclude `succeeded`. That pins both the push-site classification and the defer exclusion in one test (and #2912 already has the env pattern to mirror).

iurii-ssv: Push is visible now — verified the implementation end-to-end and it checks out:

- The tag is applied at the [push site](https://github.com/ssvlabs/ssv/blob/a3d22d031efcd282614d7914006b0d313c414380/protocol/v2/ssv/runner/committee.go#L637) (post-fallback, so recoverable by construction) and covers both the coded verify-failure and the uncoded `Deserialize`/`Recover` sub-cases; untagged errors still default to terminal at the [receive gate](https://github.com/ssvlabs/ssv/blob/a3d22d031efcd282614d7914006b0d313c414380/protocol/v2/ssv/runner/committee.go#L666).
- The [sentinel](https://github.com/ssvlabs/ssv/blob/a3d22d031efcd282614d7914006b0d313c414380/protocol/v2/ssv/runner/error.go#L64-L75) delegates `Error()` so the emitted message is unchanged, and `Unwrap()` keeps the code-tagged `*spectypes.Error` reachable via `errors.As` — and CI confirms it: both spec-test jobs (plus lint/unit) are green on the new head.

The regression test from the summary is the one thing still outstanding — the new commit predates my note above, so just re-flagging: a quorum containing one non-deserializable partial sig must not conclude the duty `failed`, and a subsequent valid quorum must still conclude `succeeded`.

protocol/v2/ssv/runner/committee.go
iurii-ssv: Audited the collapse path-by-path on the pushed commit — the equivalence holds:

- The [defer](https://github.com/ssvlabs/ssv/blob/a3d22d031efcd282614d7914006b0d313c414380/protocol/v2/ssv/runner/committee.go#L517-L521) arms right after the pre-quorum `return nil`, so the early message-processing failure return keeps its pre-PR unmarked semantics, and the named `err` return means every later `return <expr>` — including the shadowed `err :=` sites — is visible to it.
- All post-arming returns check out: the terminal ones (expected-roots, `ErrNoValidDutiesToExecute`, `ctx.Err()` with `context.Canceled` filtered inside `markDutyFailed`, both submits, the `currentDutySlot`/attestation-data helpers, `terminalErr`) flow into the mark; `recoverableErr` carries its tag and is excluded; the reconstruct goroutine's internal `return`s never touch the outer named return; and `ErrNoValidDutiesToExecute` is still returned bare, so the `errors.Is` handling in `committee_queue` is unaffected.

LGTM — with the regression test from the other thread, this is complete on my end.

iurii-ssv:
**Summary & next steps**

All review threads are resolved and verified, and CI is fully green on the current head: recoverable reconstruct failures are tagged at the push site (covering both the coded verify-failure and the uncoded deserialize/recover cases), untagged errors still default to terminal, the per-site `markDutyFailed` calls are collapsed into a single sentinel-filtered defer, and regression tests pin both classification branches.

Pre-merge checklist — complete:
- [x] Regression test pinning the fix — landed and passing: recoverable garbage-sig quorum → not `failed`, later valid quorum → `succeeded`; terminal submit failure → `failed`; the tag never hides the reconstruct error code from `errors.As`.
- [x] Lint — fixed (`QF1008` + the `beacon` variable shadowing rename); every check is green.

LGTM from my side — ready to merge.

Follow-ups after merge, tracked in #2919:
- [x] Helper plan switched from receive-side code discrimination (32/81) to reusing this PR's push-site tag (per the update comment there).
- [ ] Pre-existing `errCh` drain hardening (a buffered error can go unread when `signatureCh` closes first) — drain after the listener breaks; applies to both `committee.go` and `aggregator_committee.go`.
- [ ] Reciprocal cross-reference comment on the aggregator side.
- [ ] Refresh the `AggregatorCommitteeRunner` item — #2912 already adds the missing marks + the code-81 exclusion; what remains is the deterministic terminal-vs-recoverable split, the legacy `AggregatorRunner` fix, and unifying all runners on the shared tag.

protocol/v2/ssv/runner/committee.go
iurii-ssv: Checked the new commit — the coverage is exactly what was asked, and it verifies cleanly:

- The classification test pins both branches end-to-end: the garbage sig keeps its structural length, so it enters the container and crosses quorum optimistically, exercising the *uncoded* deserialize path; it surfaces a recoverable error without concluding the duty, and signer 4 then re-crosses quorum and concludes `succeeded` (with the submit-failure branch concluding `failed` symmetrically).
- Nice touch in `error_test.go` asserting the tag doesn't hide the coded `*spectypes.Error` from `errors.As` — that pins the spectest-observability property directly.

One thing to fix — lint failed on this push with a single trivial staticcheck issue:

> `protocol/v2/ssv/runner/committee_postconsensus_classification_test.go:37:13: QF1008: could remove embedded field "BaseRunner" from selector`

i.e. `observeConclusion` should read `env.runner.dutyConcluded = concluded`. The remaining jobs (spec-test ×2, test) were still running as of this writing; everything else is green. Once lint is fixed and CI completes, I'll tick the box in the summary above.

ovidiu-ssv-labs:
The fix is correct, well-documented, and the tests pass under `-race`, including `TestSSVMapping` — **no blockers from me**.

Below are three **minor, non-blocking** observations. None need to happen in this PR; #1 in particular is a reasonable candidate for a separate follow-up issue.

### 1. Pre-existing `errCh` drain race (not introduced here) — follow-up candidate
`committee.go` listener loop: when `signatureCh` closes (`ok=false`) while `errCh` still holds a buffered error, `select` picks a ready case pseudo-randomly and can take the `signatureCh`-closed branch → `break listener`, leaving the buffered error unread.
- For the *recoverable* case this is harmless (unread → returns `nil` → still not marked failed).
- For a *terminal* error it would resurface the exact false-`stuck` this PR fights.

This race exists identically in the original `executionErr` code and in `aggregator_committee.go`, so it is **not a regression** in this PR. But since the whole point of this change is reliable classification, draining `errCh` after the listener breaks would harden the guarantee. Worth a separate issue rather than expanding this PR's scope.

### 2. Divergence from `AggregatorCommitteeRunner`
`CommitteeRunner` now classifies via the `recoverableReconstructError` tag wrapper + a `terminalErr`/`recoverableErr` split, whereas the sibling `AggregatorCommitteeRunner` (#2912) uses `spectypes.WrapError(PostConsensusQuorumWithInvalidSignatures, …)` + a single `executionErr`. The divergence is **justified** (the PR body explains the spec-fixture constraint that made tagging with the aggregator code break 18 spectests), so no code change requested — just noting that a one-line cross-reference comment in each runner pointing at the other would save a future reader the archaeology.

### 3. `beacon` identifier shadowing in the test helper
In `committee_test.go` (`newCommitteeRunnerEnvInternal`), the local variable `beacon` shadows the newly-imported `beacon` package. It compiles fine (the `beacon.BeaconNode` param type resolves in package scope) and there's already a comment acknowledging it — purely a readability nit; a distinct local name (e.g. `defaultBeacon`) would remove the double-take.

_Optional cleanups only — happy to see this merge as-is... once lint also passes_ 😅

momosh-ssv:
Thanks @ovidiu-ssv-labs, much appreciated. Where each landed:

- **#3 (`beacon` shadowing):** fixed in e0032b046 and renamed to `defaultBeacon`. (Same commit fixed the `QF1008` lint red, so CI is green now.)
- **#2 (cross-reference comment):** added the forward note in `committee.go` at the `terminalErr`/`recoverableErr` split (474fe8073), explaining why it classifies by tag while `AggregatorCommitteeRunner` uses code 81. 
Folded the reciprocal aggregator-side note into [#2919] so both sides get documented when that runner moves to the same tag approach.
- **#1 (pre-existing `errCh` drain race):** agree it's out of scope here and not a regression and folded it into [#2919][fold-comment], since the identical pattern lives in `aggregator_committee.go` too. Draining `errCh` after the listener breaks is the hardening you described.

While there I also flagged that [#2919]'s original code-based-helper direction is superseded by this PR's push-site tag, so the follow-up reuses it rather than reintroducing the uncoded-error blind spot.

[#2919]: https://github.com/ssvlabs/ssv/issues/2919
[fold-comment]: https://github.com/ssvlabs/ssv/issues/2919#issuecomment-4868181249

iurii-ssv (approved):
Well done!

PR #2926, opened 2026-07-03
observability(runner): classify aggregator terminal post-quorum failures as failed (#2919)

### LINKED ISSUE #2919: runner: classify aggregator terminal post-quorum failures as failed (follow-up to #2918)
### Context

PR #2918 fixed `CommitteeRunner.ProcessPostConsensus` so that terminal post-quorum failures are classified as `failed` (not "stuck"), while the recoverable reconstruct-invalid-signatures path (`ReconstructSignatureErrorCode`, code 32 in the pinned ssv-spec) is deliberately left un-marked so a later partial-sig can retry.

During review (@iurii-ssv), it was noted the aggregator runners have the analogous problem. This issue tracks bringing them in line. It was intentionally left out of #2918 to keep that PR scoped to `CommitteeRunner`.

### Two distinct gaps

**1. Legacy `AggregatorRunner` (`protocol/v2/ssv/runner/aggregator.go`) — identical bug**
- `ProcessPostConsensus` (~L314) uses a blanket `defer { if err != nil { markDutyFailed(err) } }`.
- It calls `ReconstructBeaconSig` (~L327), which on failure returns `ReconstructSignatureErrorCode` (the **same** recoverable, code-32 condition committee handles), then falls back via `FallBackAndVerifyEachSignature`.
- Result: the recoverable reconstruct-invalid-sigs path is mis-marked as a terminal `failed`, exactly the bug #2918 fixed for committee.
- Note: this runner is only wired **pre-Boole** (`!NetworkConfig.BooleFork()`), so impact is limited to pre-Boole networks.

**2. Live post-Boole `AggregatorCommitteeRunner` (`protocol/v2/ssv/runner/aggregator_committee.go`) — never marks failed at all**
- `ProcessPostConsensus` (L766) has **no** `markDutyFailed` call anywhere; on `executionErr` it simply `return executionErr` (L1045-1046).
- Consequently *every* terminal post-quorum failure on the live aggregator path is currently left unconcluded and later surfaces as a deadline/"stuck" mislabel.
- Its recoverable condition is tagged `PostConsensusQuorumWithInvalidSignatures` (code 81), not code 32 — so the classification logic must handle both codes.
- ⚠️ The fix here is **not** "mark everything failed": that would re-introduce the exact false-`failed` bug #2918 fixed for committee. It must add `markDutyFailed` for terminal errors **while excluding the recoverable code-81 path** (leave that un-marked so a later partial-sig can retry).
- It also shares committee's last-write-wins `executionErr` hazard, at even more sites — `executionErr` is overwritten (not accumulated) at L930, L938, L944, L953, L990, L1020 — so terminal-vs-recoverable classification is order-dependent there too.

### Suggested direction

Introduce a shared `isRecoverableReconstructError` helper on `BaseRunner` that recognises **both** `ReconstructSignatureErrorCode` (32) and `PostConsensusQuorumWithInvalidSignatures` (81), then wire the "terminal → markDutyFailed, recoverable → skip" classification into:
- `CommitteeRunner` (refactor to use the helper),
- legacy `AggregatorRunner`,
- `AggregatorCommitteeRunner` (add the missing `markDutyFailed` path — for terminal errors only, excluding the recoverable code-81 path per the note above).

Also apply the same deterministic terminal-vs-recoverable split #2918 introduced for committee (separate `terminalErr`/`recoverableErr` instead of a single last-write-wins `executionErr`) to `AggregatorCommitteeRunner` (and `AggregatorRunner` where applicable).

### Acceptance / testing
- `go test ./protocol/v2/ssv/spectest/...` must pass. Mind `TestSSVMapping` — its fixtures pin the per-runner spec error codes, so the helper must preserve each runner's existing emitted code (32 vs 81), not unify them.
- `go test ./protocol/v2/ssv/runner/...` for the runner-level behavior.
- Add/extend coverage for: a terminal error winning over a concurrent recoverable error in the same round (deterministic classification), and the recoverable path staying un-marked so a retry can conclude the duty.

### References
- Fix for committee: #2918 (commit `657721c4b`)
- Review thread: https://github.com/ssvlabs/ssv/pull/2918#discussion_r3506739870

momosh-ssv:
Follow-up to #2918. Brings the aggregator runners in line with the committee fix so terminal post-quorum failures are concluded as `failed` (instead of surfacing later as a false "stuck"), while recoverable reconstruct-invalid-sigs stay un-marked so a later partial-sig can retry.

Closes #2919.

## Changes

- **`aggregator.go` (legacy `AggregatorRunner`, pre-Boole)** — wrap the pre- and post-consensus reconstruct-invalid-sigs returns in `recoverableReconstructError` and guard both defers with `!isRecoverableReconstructError`, mirroring #2918. This runner previously reproduced the exact blanket-`markDutyFailed` bug committee had.

- **`aggregator_committee.go`** — replace the single last-write-wins `executionErr` with a deterministic `terminalErr`/`recoverableErr` split (terminal wins regardless of goroutine/root ordering); classify the recoverable path by the `PostConsensusQuorumWithInvalidSignatures` (code 81) at the errCh receive site; drain `errCh` after the listener closes so a buffered error can't be skipped by the `signatureCh`-close race.
  - The aggregator deliberately keeps classifying by **code 81** (not the sentinel tag): it force-wraps every reconstruct error with that code at the sole push site, so it has no uncoded-BLS blind spot, and tagging here would break its spectest fixtures. The divergence from committee's tag approach is documented from both sides.

- **`committee.go`** — add the same defensive `errCh` drain (the close-race is pre-existing here too).

## Review input folded in (from #2918)

- **@ovidiu-ssv-labs** — the pre-existing `errCh` drain race (buffered error skippable when `signatureCh` closes) and the reciprocal cross-reference comment documenting the code-vs-tag divergence.
- **@iurii-ssv** — confirmed the deterministic terminal-vs-recoverable classification direction and the deadline-as-terminal semantics.

## Testing

- `go test ./protocol/v2/ssv/runner/... -race` — green
- `go test ./protocol/v2/ssv/spectest/...` incl. `TestSSVMapping` — green (per-runner emitted codes 32/81 preserved)

New/extended runner tests cover: legacy-aggregator terminal vs recoverable classification; deterministic terminal-wins-over-concurrent-recoverable; the `errCh` drain postcondition; tag-not-code classification; and the recoverable-invalid-sigs → retry → **succeeds** flow for `AggregatorCommitteeRunner`.

> Note: the legacy `AggregatorRunner`'s recoverable→retry-succeeds path is covered structurally by the equivalent `CommitteeRunner` test (shared BaseRunner conclusion machinery) rather than re-tested explicitly, since it's pre-Boole-only and its spec harness is awkward. Happy to add it if reviewers prefer full symmetry.

### DIFF
--- a/protocol/v2/ssv/runner/aggregator.go
+++ b/protocol/v2/ssv/runner/aggregator.go
@@ -120,8 +120,13 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 		return nil
 	}
 
-	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
-	// so a terminal failure below won't be retried.
+	// We have quorum and are committed to completing this duty here, so this defer reports a terminal
+	// failure below as failed rather than letting it surface as a false "stuck".
+	// The one exception is a recoverable BLS-reconstruction failure: the fallback drops the offending
+	// partial sig(s) below quorum, so a later partial-sig message re-crosses quorum and re-enters here —
+	// those are tagged recoverableReconstructError and must not be recorded as failed (hence the guard).
+	// Shutdown (context cancellation) needs no special-casing — markDutyFailed drops a context.Canceled
+	// reason, so a duty aborted by shutdown isn't recorded as a failure.
 	defer func() {
 		if err != nil && !isRecoverableReconstructError(err) {
 			r.markDutyFailed(err)
@@ -309,8 +314,13 @@ func (r *AggregatorRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 		return nil
 	}
 
-	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
-	// so a terminal failure below won't be retried.
+	// We have quorum and are committed to completing this duty here, so this defer reports a terminal
+	// failure below as failed rather than letting it surface as a false "stuck".
+	// The one exception is a recoverable BLS-reconstruction failure: the fallback drops the offending
+	// partial sig(s) below quorum, so a later partial-sig message re-crosses quorum and re-enters here —
+	// those are tagged recoverableReconstructError and must not be recorded as failed (hence the guard).
+	// Shutdown (context cancellation) needs no special-casing — markDutyFailed drops a context.Canceled
+	// reason, so a duty aborted by shutdown isn't recorded as a failure.
 	defer func() {
 		if err != nil && !isRecoverableReconstructError(err) {
 			r.markDutyFailed(err)
--- a/protocol/v2/ssv/runner/aggregator.go
+++ b/protocol/v2/ssv/runner/aggregator.go
@@ -123,7 +123,7 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
 	// so a terminal failure below won't be retried.
 	defer func() {
-		if err != nil {
+		if err != nil && !isRecoverableReconstructError(err) {
 			r.markDutyFailed(err)
 		}
 	}()
@@ -140,7 +140,7 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 	if err != nil {
 		// If the reconstructed signature verification failed, fall back to verifying each partial signature
 		r.FallBackAndVerifyEachSignature(r.State.PreConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
-		return fmt.Errorf("got pre-consensus quorum but it has invalid signatures: %w", err)
+		return recoverableReconstructError{fmt.Errorf("got pre-consensus quorum but it has invalid signatures: %w", err)}
 	}
 
 	duty, err := r.currentValidatorDuty()
@@ -312,7 +312,7 @@ func (r *AggregatorRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
 	// so a terminal failure below won't be retried.
 	defer func() {
-		if err != nil {
+		if err != nil && !isRecoverableReconstructError(err) {
 			r.markDutyFailed(err)
 		}
 	}()
@@ -328,7 +328,7 @@ func (r *AggregatorRunner) ProcessPostConsensus(ctx context.Context, logger *zap
 	if err != nil {
 		// If the reconstructed signature verification failed, fall back to verifying each partial signature
 		r.FallBackAndVerifyEachSignature(r.State.PostConsensusContainer, root, r.GetShare().Committee, r.GetShare().ValidatorIndex)
-		return fmt.Errorf("got post-consensus quorum but it has invalid signatures: %w", err)
+		return recoverableReconstructError{fmt.Errorf("got post-consensus quorum but it has invalid signatures: %w", err)}
 	}
 	specSig := phase0.BLSSignature{}
 	copy(specSig[:], sig)
--- a/protocol/v2/ssv/runner/aggregator_committee.go
+++ b/protocol/v2/ssv/runner/aggregator_committee.go
@@ -822,7 +822,14 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
 		return bytes.Compare(roots[i][:], roots[j][:]) < 0
 	})
 
-	var executionErr error
+	// Unlike the sibling CommitteeRunner (which tags recoverable reconstruct failures with the
+	// recoverableReconstructError sentinel), this runner classifies them by the
+	// PostConsensusQuorumWithInvalidSignatures spec code: every reconstruct error is force-wrapped
+	// with that code at the push site below, so there is no uncoded-BLS blind spot here, and tagging
+	// instead would break this runner's spectest fixtures. The divergence is deliberate — see the
+	// reciprocal note in committee.go. A single last-write-wins error made terminal-vs-recoverable
+	// classification depend on goroutine/root ordering; splitting keeps it deterministic (terminal wins).
+	var terminalErr, recoverableErr error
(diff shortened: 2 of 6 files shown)


protocol/v2/ssv/runner/aggregator.go
```diff
@@ -123,7 +123,7 @@ func (r *AggregatorRunner) ProcessPreConsensus(ctx context.Context, logger *zap.
 	// We have quorum and are committed to completing this duty here. The quorum above fires only once,
 	// so a terminal failure below won't be retried.
 	defer func() {
-		if err != nil {
+		if err != nil && !isRecoverableReconstructError(err) {
```
iurii-ssv: nit: this comment now tells only half the story — the guard exists precisely because the recoverable case *is* retried (the fallback drops the offending signer(s) below quorum, so a later partial-sig message re-crosses quorum and re-enters here). Consider carrying over the one-line exception note from the `CommitteeRunner` defer comment ([committee.go#L508-L521](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/committee.go#L508-L521)), here and on the identical post-consensus defer ([aggregator.go#L311-L317](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/aggregator.go#L311-L317)).

protocol/v2/ssv/runner/committee.go
```diff
@@ -712,6 +712,26 @@ func (r *CommitteeRunner) ProcessPostConsensus(ctx context.Context, logger *zap.
...
+			case err := <-errCh:
+				if isRecoverableReconstructError(err) {
+					recoverableErr = err
+				} else {
+					terminalErr = err
+				}
+			default:
+				break drainErrCh
+			}
+		}
```
iurii-ssv, line 732: The greptile comment on `aggregator_committee.go` flagged the listener/drain classification copy-paste there, but the same two-site duplication exists in this file too — the listener receive ([committee.go#L664-L674](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/committee.go#L664-L674)) and this drain share only the `isRecoverableReconstructError` predicate, not the classify-and-assign block. If that comment gets addressed with a per-function `classify := func(err error)` closure over `terminalErr`/`recoverableErr`, the same shape fits here so the two receive sites can never drift apart.

iurii-ssv (commented):
Couple minor cleanups, plus it seems similar approach could be applied elsewhere as well: **`proposer.go` and `sync_committee_contribution.go` still carry the exact bug class this PR removes from the legacy aggregator** — blanket `if err != nil { r.markDutyFailed(err) }` defers ([proposer.go#L134-L138](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/proposer.go#L134-L138), [proposer.go#L358-L362](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/proposer.go#L358-L362), [sync_committee_contribution.go#L110-L114](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/sync_committee_contribution.go#L110-L114), [sync_committee_contribution.go#L333-L337](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/sync_committee_contribution.go#L333-L337)) combined with recoverable post-fallback reconstruct returns ([proposer.go#L146-L151](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/proposer.go#L146-L151), [proposer.go#L370-L375](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/proposer.go#L370-L375), [sync_committee_contribution.go#L363-L370](https://github.com/ssvlabs/ssv/blob/fix/2919-aggregator-terminal-failed/protocol/v2/ssv/runner/sync_committee_contribution.go#L363-L370)). A byzantine partial sig there still concludes the duty `failed` even though a later message could retry. Scoping this PR to the aggregators per #2919 is fine — but the description should say whether a follow-up covers proposer/contribution, since the false-`failed` now knowingly remains in those two runners.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -953,21 +960,28 @@ func (r *AggregatorCommitteeRunner) ProcessPostConsensus(
...
+				// PostConsensusQuorumWithInvalidSignatures code; anything arriving without that code is terminal.
+				var specErr *spectypes.Error
+				if errors.As(err, &specErr) && specErr.Code == spectypes.PostConsensusQuorumWithInvalidSignatures {
+					recoverableErr = err
+				} else {
+					terminalErr = err
+				}
 			case signatureResult, ok := <-signatureCh:
 				if !ok {
 					break listener
```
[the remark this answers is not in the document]
momosh-ssv: Good cal, done in`9614121c0`.
Matches the same treatment applied to `committee.go` in `18729e0c9`.

protocol/v2/ssv/runner/committee.go
momosh-ssv: Agreed, addressed in 18729e0c9.

Folded the listener receive and the drain into a single `classify := func(err error)` closure here too, so the two sites can't drift. Did the same for `AggregatorCommitteeRunner` in 9614121c0.

protocol/v2/ssv/runner/aggregator.go
momosh-ssv: You're right, the old comment only covered the terminal half. 
Carried over the recoverable-exception note from the `CommitteeRunner` defer to both defers here (pre- and post-consensus) in a617835e5. 

It now spells out that the guard exists because the recoverable case *is* retried: the fallback drops the offending signer below quorum and a later partial-sig re-crosses quorum and re-enters.

iurii-ssv (approved):
Nice!

PR #2927, opened 2026-07-06
convergence(unit 2): networking + message/validation fork-gating

momosh-ssv:
## Unit 2 of the boole-fork → stage convergence: networking + message/validation fork-gating

Third slice of the [Option C convergence](https://github.com/ssvlabs/ssv/issues/2902) onto `integration/boole-convergence`, stacked on units 0 (#2924) and 1 (#2925), both already merged. Ports boole-fork's fork-aware pubsub layer onto stage, **stage-idiomatically** (keeps stage's `uint64` subnets — not boole's `Subnet` type).

### What it delivers
- **Publish** — `BroadcastAtSlot(msg, slot)` on the `protocolp2p.Broadcaster` interface; routes to the Alan (`ssv.v2.<subnet>`) or Boole (`/ssv/<net>/boole/<subnet>`) topic by `BooleForkAtSlot(slot)`. Legacy `Broadcast` self-resolves the slot. The interface addition ripples (mechanically) through the qbft controller and runners.
- **Subscribe** — dual Alan+Boole subscription during `InBooleTransitionWindow`; Alan dropped after the window.
- **Receive** — slot-gated `validateTopicAtSlot` + `validateDomainAtSlot` (after decode; slot from `consensusMessage.Height` / `partialSignatureMessages.Slot`), replacing the pre-decode Alan-only topic check; pre-decode domain check relaxed to a `{DomainType, NextDomainType}` allowlist with the exact per-slot pin enforced post-decode.
- **Pre-wiring gate** — `networkconfig` `Forks` is now a pointer, so an absent `forks:` block defaults `Boole` to *never* (`MaxUint64`) rather than fork-at-genesis.

### Review status
Two independent `code-reviewer-deep` passes — **receive-side/security** and **networking/subscription** — both returned **no blockers, no majors**. Verified: topic/domain strings byte-match the publish side; domain relaxation + ordering match the boole-fork reference; transition-window state machine has no message-loss gap and no Alan leak; qbft/runner ripple passes the correct duty slot everywhere. One MAJOR (latent, gated to unit 5) was found and fixed: `committeeRole()` now includes `RoleAggregatorCommittee` (was publish/receive-asymmetric). Two defensive items fixed (empty-committee guard; documented `persistentSubnets` startup invariant).

### Testing
Local gates green: `go build ./...`, `go vet ./...` (main **+ ssvsigner** module), `scripts/deadcode.sh`, `golangci-lint` (changed packages), and the `message/validation` + `network/commons` + subnet unit tests — incl. new post-fork Boole receive tests and a `committeeRole` symmetry guard.

> ⚠️ The libp2p integration tests (`TestP2pNetwork_*`, `network/topics.TestTopicManager`) can't run in the dev sandbox (no mdns/multicast) — **CI is the real networking gate** for those.

### Deferred follow-ups (non-blocking)
- A `-race`-verified mutex around `persistentSubnets` (documented invariant for now; the tests that would exercise it need CI).
- `operator/validator/metadata/syncer.go` still uses the Alan-only subnet for a sync-ordering heuristic (not the ENR authority) — a later unit.

_The first commit is titled "WIP … pre-rebase snapshot" — that's the consolidated networking draft, now fully reviewed; kept as-is to avoid history rewrite._

### DIFF
--- a/message/validation/validation.go
+++ b/message/validation/validation.go
@@ -246,7 +246,9 @@ func (mv *messageValidator) validateTopicAtSlot(committeeInfo CommitteeInfo, top
 	if mv.netCfg.BooleForkAtSlot(slot) {
 		expectedTopic = commons.BooleTopic(mv.netCfg.SSV.Name, commons.BooleCommitteeSubnet(committeeInfo.committee))
 	} else {
-		expectedTopic = commons.GetTopicFullName(commons.SubnetTopicID(commons.AlanCommitteeSubnet(committeeInfo.committeeID)))
+		// CommitteeTopicID(cid)[0] == SubnetTopicID(AlanCommitteeSubnet(cid)), which is what
+		// BroadcastAtSlot publishes on the Alan side, so the full names byte-match.
+		expectedTopic = commons.GetTopicFullName(commons.CommitteeTopicID(committeeInfo.committeeID)[0])
 	}
 
 	// Rule: Check if message was sent in the correct topic
--- a/message/validation/validation.go
+++ b/message/validation/validation.go
@@ -4,10 +4,10 @@ package validation
 // validator.go contains main code for validation and most of the rule checks.
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"fmt"
-	"slices"
 	"sync"
 	"time"
 
@@ -197,7 +197,7 @@ func (mv *messageValidator) handleSignedSSVMessage(
 		return decodedMessage, err
 	}
 
-	if err := mv.committeeChecks(signedSSVMessage, committeeInfo, topic); err != nil {
+	if err := mv.committeeChecks(signedSSVMessage, committeeInfo); err != nil {
 		return decodedMessage, err
 	}
 
@@ -212,14 +212,14 @@ func (mv *messageValidator) handleSignedSSVMessage(
 
 	switch signedSSVMessage.SSVMessage.MsgType {
 	case spectypes.SSVConsensusMsgType:
-		consensusMessage, err := mv.validateConsensusMessage(ctx, signedSSVMessage, committeeInfo, receivedFrom, receivedAt)
+		consensusMessage, err := mv.validateConsensusMessage(ctx, signedSSVMessage, committeeInfo, topic, receivedFrom, receivedAt)
 		decodedMessage.Body = consensusMessage
 		if err != nil {
 			return decodedMessage, err
 		}
 
 	case spectypes.SSVPartialSignatureMsgType:
-		partialSignatureMessages, err := mv.validatePartialSignatureMessage(ctx, signedSSVMessage, committeeInfo, receivedFrom, receivedAt)
+		partialSignatureMessages, err := mv.validatePartialSignatureMessage(ctx, signedSSVMessage, committeeInfo, topic, receivedFrom, receivedAt)
 		decodedMessage.Body = partialSignatureMessages
 		if err != nil {
 			return decodedMessage, err
@@ -232,19 +232,43 @@ func (mv *messageValidator) handleSignedSSVMessage(
 	return decodedMessage, nil
 }
 
-func (mv *messageValidator) committeeChecks(signedSSVMessage *spectypes.SignedSSVMessage, committeeInfo CommitteeInfo, topic string) error {
-	if err := mv.belongsToCommittee(signedSSVMessage.OperatorIDs, committeeInfo.committee); err != nil {
-		return err
+func (mv *messageValidator) committeeChecks(signedSSVMessage *spectypes.SignedSSVMessage, committeeInfo CommitteeInfo) error {
+	// Note: the "correct topic" rule is enforced later, per-slot and fork-aware, by
+	// validateTopicAtSlot once the message slot is known (see consensus/partial validation).
+	return mv.belongsToCommittee(signedSSVMessage.OperatorIDs, committeeInfo.committee)
+}
+
+// validateTopicAtSlot enforces that the message arrived on the topic that matches the
+// committee's subnet for the fork active at the given slot. The expected topic strings
+// mirror the publish side (p2pNetwork.BroadcastAtSlot) exactly.
+func (mv *messageValidator) validateTopicAtSlot(committeeInfo CommitteeInfo, topic string, slot phase0.Slot) error {
+	var expectedTopic string
+	if mv.netCfg.BooleForkAtSlot(slot) {
+		expectedTopic = commons.BooleTopic(mv.netCfg.SSV.Name, commons.BooleCommitteeSubnet(committeeInfo.committee))
+	} else {
+		expectedTopic = commons.GetTopicFullName(commons.SubnetTopicID(commons.AlanCommitteeSubnet(committeeInfo.committeeID)))
 	}
 
 	// Rule: Check if message was sent in the correct topic
-	messageTopics := commons.CommitteeTopicID(committeeInfo.committeeID)
-	topicBaseName := commons.GetTopicBaseName(topic)
-	if !slices.Contains(messageTopics, topicBaseName) {
+	if expectedTopic != topic {
 		e := ErrIncorrectTopic
-		e.got = fmt.Sprintf("topic %v / base name %v", topic, topicBaseName)
-		e.want = messageTopics
+		e.got = fmt.Sprintf("%v @ %v", topic, slot)
+		e.want = expectedTopic
 		return e
+	}
+
+	return nil
+}
+
+// validateDomainAtSlot enforces that the message domain matches the fork active at the
+// given slot (Alan before the Boole fork, Boole after).
+func (mv *messageValidator) validateDomainAtSlot(msgID spectypes.MessageID, slot phase0.Slot) error {
+	expectedDomain := mv.netCfg.DomainTypeAtSlot(slot)
+	if msgDomain := msgID.GetDomain(); !bytes.Equal(msgDomain, expectedDomain[:]) {
+		err := ErrWrongDomain
+		err.got = hex.EncodeToString(msgDomain)
+		err.want = hex.EncodeToString(expectedDomain[:])
+		return err
 	}
 
 	return nil
--- a/network/commons/subnets.go
+++ b/network/commons/subnets.go
@@ -2,14 +2,19 @@ package commons
 
 import (
 	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"math"
 	"math/big"
 	"math/bits"
+	"strconv"
 	"strings"
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
+
+	"github.com/ssvlabs/ssv/networkconfig"
 )
 
 const (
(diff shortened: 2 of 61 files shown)


network/p2p/p2p_pubsub.go
```diff
@@ -168,43 +217,91 @@ func (n *p2pNetwork) Subscribe(pk spectypes.ValidatorPK) error {
...
+		return p2pprotocol.ErrNetworkIsNotReady
+	}
+	if subnet >= commons.SubnetsCount {
+		return fmt.Errorf("invalid subnet %d", subnet)
+	}
+	topic := commons.GetTopicFullName(commons.SubnetTopicID(subnet))
+	if useBoole {
+		topic = commons.BooleTopic(n.cfg.NetworkConfig.SSV.Name, subnet)
+	}
+	if err := n.topicsCtrl.Subscribe(topic); err != nil {
```
iurii-ssv, line 287: **Blocker — topic names get double-prefixed (and Boole names mangled).**

This passes an already-*full* topic name into `topicsCtrl.Subscribe`: the Alan branch builds `commons.GetTopicFullName(commons.SubnetTopicID(subnet))` = `ssv.v2.5`, the Boole branch builds `commons.BooleTopic(...)` = `/ssv/<net>/boole/5`. But `network/topics/controller.go` `Subscribe`/`Broadcast`/`Unsubscribe`/`Peers` still call `commons.GetTopicFullName(name)` **internally**, so the effective topics become `ssv.v2.ssv.v2.5` and `ssv.v2./ssv/<net>/boole/5`.

The same full-name-into-controller pattern appears in `BroadcastAtSlot` (L81/L95), `unsubscribeSubnet` (L300/L304), and `committeeTopicSetForCurrentEpoch` (L336-337) feeding `subscribeCommittee`/`Unsubscribe`.

CI confirms the breakage: the `ssv` job's `TestP2pNetwork_SubscribeBroadcast` receives 0 messages (`"0" is not greater than or equal to "2"`), and the logs show topics like `ssv.v2.ssv.v2.64`. `TestMsgValidator/valid_consensus_msg` fails for the same reason.

Fix: make the topics controller treat the passed name as already-full (drop its internal `GetTopicFullName`), since the Boole naming scheme (`/ssv/<net>/boole/<subnet>`) can't be expressed as an `ssv.v2.<base>` base name anyway.

network/p2p/p2p.go
```diff
@@ -707,125 +764,187 @@ func (n *p2pNetwork) getMaxPeers(topic string) int {
...
-//   - calculate subnetPeersExcluding (currently connected subnets IF that candidate peer is excluded/disconnected)
-//   - use SubnetPeers.Score to calculate the final peer-score for each peer (based on: subnetPeersExcluding, desired
-//     set of subnets, subnets this peer is connected to)
+//   - snapshot ownAlanSubnets, ownBooleSubnets, ownSubnetPeers, peerObservedSubnets
+//   - for each candidate peer:
+//   - build subnetPeersExcluding by decrementing the (fork, subnet) slots the peer actually serves
+//   - Score the peer against the excluding-counts
 func (n *p2pNetwork) buildPeerTrimScores(peerIDs []peer.ID) map[peer.ID]float64 {
-	ownSubnets := n.SubscribedSubnets()
+	ownAlanSubnets, ownBooleSubnets := n.subscribedSubnetsForCurrentEpoch()
```
iurii-ssv, line 781: **Blocker — nil-pointer panic in the trim path.**

`buildPeerTrimScores` now calls `subscribedSubnetsForCurrentEpoch()`, which dereferences `n.cfg.NetworkConfig.EstimatedCurrentSlot()` (p2p_pubsub.go:171). `newTrimTestNetwork` in `p2p_trimming_test.go` constructs the `p2pNetwork` without `cfg`, so `TestBuildPeerTrimScores_CharacterizesDeadSoloDuoScoring` panics with an invalid-memory-address deref (visible in the red `ssv` CI job).

Set `cfg` (with a `NetworkConfig`) in the trim-test network constructor; optionally guard the deref so this path degrades instead of panicking.

network/p2p/p2p_pubsub.go
```diff
@@ -38,6 +39,21 @@ func (n *p2pNetwork) UseMessageRouter(router network.MessageRouter) {
 
 // Broadcast publishes the message to all peers in subnet
 func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedSSVMessage) error {
+	decodedMsg, err := queue.DecodeSignedSSVMessage(msg)
```
iurii-ssv, line 42: Minor / efficiency: legacy `Broadcast` fully decodes the message (`queue.DecodeSignedSSVMessage` + `Slot()`) solely to resolve the slot, discards the decoded result, then `BroadcastAtSlot` independently re-serializes it via `msg.Encode()` (L65). So every plain-`Broadcast` call pays a decode it throws away. Callers that already know the slot (qbft controller/instance use `Height`) correctly skip this by calling `BroadcastAtSlot` directly; for the remaining `Broadcast` callers, consider having `BroadcastAtSlot` reuse the already-decoded message, or extract the slot without a full decode.

message/validation/validation.go
```diff
@@ -232,24 +232,50 @@ func (mv *messageValidator) handleSignedSSVMessage(
...
+	return mv.belongsToCommittee(signedSSVMessage.OperatorIDs, committeeInfo.committee)
+}
+
+// validateTopicAtSlot enforces that the message arrived on the topic that matches the
+// committee's subnet for the fork active at the given slot. The expected topic strings
+// mirror the publish side (p2pNetwork.BroadcastAtSlot) exactly.
+func (mv *messageValidator) validateTopicAtSlot(committeeInfo CommitteeInfo, topic string, slot phase0.Slot) error {
+	var expectedTopic string
+	if mv.netCfg.BooleForkAtSlot(slot) {
+		expectedTopic = commons.BooleTopic(mv.netCfg.SSV.Name, commons.BooleCommitteeSubnet(committeeInfo.committee))
```
iurii-ssv, line 247: Minor / efficiency: post-fork this recomputes `commons.BooleCommitteeSubnet(committeeInfo.committee)` — a SHA-256 per operator plus a `big.Int` mod — on **every** received consensus/partial-signature message, on the validation hot path. The value is already cached elsewhere: `storage.Committee.BooleSubnet`/`AlanSubnet` and `SSVShare.BooleCommitteeSubnet()`/`AlanCommitteeSubnet()`. `getCommitteeAndValidatorIndices`/`newCommitteeInfo` currently drop it. Consider threading the cached Alan/Boole subnet through `CommitteeInfo` and using it here.

iurii-ssv (commented):
Two blocking issues (both reproduced by the currently-red `ssv` CI job), plus two minor efficiency notes. Skipped the three items already raised by the Greptile bot (pre-fork Boole topic enumeration, transition Alan-topic unsubscribe leak, stale advertised handshake domain).

network/p2p/p2p_pubsub.go
momosh-ssv: Fixed in b0d1de881 — exactly as suggested: dropped the internal `GetTopicFullName` from the topics controller (`Subscribe`/`Broadcast`/`Unsubscribe`/`Peers`/`Topics`/`Close`) so it treats the passed name as already-full, and callers pass the full Alan/Boole topic. The `ssv` job is green now (a01c9cfa3 also cleaned up the network/topics test fixtures that still used base names).

network/p2p/p2p.go
momosh-ssv: Fixed in b0d1de881 — `newTrimTestNetwork` now sets `cfg: &Config{NetworkConfig: networkconfig.TestNetwork}`, so `buildPeerTrimScores` → `subscribedSubnetsForCurrentEpoch` no longer nil-derefs. Trim tests pass.

network/p2p/p2p_pubsub.go
momosh-ssv: Good catch. It only affects the legacy `Broadcast` path — qbft/runners call `BroadcastAtSlot` directly and skip the decode. Tracking as a follow-up rather than reworking the broadcast signatures at the base of the stack.

message/validation/validation.go
momosh-ssv: Agreed the post-fork recompute is wasteful. I tried threading the cached `Committee.BooleSubnet`/`SSVShare.BooleCommitteeSubnet()` through `CommitteeInfo`, but it makes correctness depend on every `Committee` construction site populating the cache — a zero/mismatched value silently rejects messages (it broke a post-fork test whose mock committee had an unset subnet). Since it only bites post-fork (not live yet), deferring to a dedicated, well-tested pass rather than the base of the stack.

network/commons/subnets.go
```diff
@@ -51,23 +60,80 @@ func GetTopicBaseName(topicName string) string {
...
-		topics[i] = GetTopicFullName(SubnetTopicID(i))
+// Topics returns the available topics for the given network's fork state: Alan-only pre-fork,
+// Alan+Boole during the transition window, Boole-only after.
+func Topics(netCfg *networkconfig.Network) []string {
+	topics := make([]string, 0, SubnetsCount*2)
+	for subnet := uint64(0); subnet < SubnetsCount; subnet++ {
+		if !netCfg.BooleFork() {
+			topics = append(topics, GetTopicFullName(SubnetTopicID(subnet)))
+		}
+		topics = append(topics, BooleTopic(netCfg.SSV.Name, subnet))
```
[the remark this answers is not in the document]
momosh-ssv: `Topics()` feeds the pubsub whitelist, not subscription (which is fork-gated separately via `subscribedSubnetsForCurrentEpoch`). Pre-whitelisting Boole topics lets a node started before the fork accept Boole messages at the fork slot without a restart, and matches boole-fork. Whitelisting ≠ joining, so there is no pre-fork Boole traffic.

network/p2p/p2p_pubsub.go
```diff
@@ -221,17 +318,38 @@ func (n *p2pNetwork) Unsubscribe(pk spectypes.ValidatorPK) error {
...
-	for _, topic := range topics {
+	for topic := range n.committeeTopicSetForCurrentEpoch(share) {
 		if err := n.topicsCtrl.Unsubscribe(topic, false); err != nil {
 			return err
 		}
 	}
-	n.subscribedCommittees.Delete(string(cmtid[:]))
+
+	cid := share.CommitteeID()
+	n.subscribedCommittees.Delete(string(cid[:]))
```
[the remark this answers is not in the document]
momosh-ssv: The periodic `UpdateSubnets` reconciliation handles this: post-window `subscribedSubnetsForCurrentEpoch` returns Alan=Zero, so `DiffSubnets` yields `removedAlanSubnets` and unsubscribes them on the next tick. So a transiently-leaked Alan topic is swept — verified no persistent leak.

network/p2p/p2p_setup.go
```diff
@@ -203,7 +203,7 @@ func (n *p2pNetwork) setupPeerServices() error {
 	if err != nil {
 		return err
 	}
-	d := n.cfg.NetworkConfig.DomainType
+	d := n.cfg.NetworkConfig.CurrentDomainType()
 	domain := "0x" + hex.EncodeToString(d[:])
 	self := records.NewNodeInfo(domain)
```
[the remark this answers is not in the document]
momosh-ssv: The advertised `NodeInfo` domain uses `CurrentDomainType()` at startup — a pre-existing discovery-layer pattern (stage already advertised a startup domain), unchanged in intent here and matching boole-fork. Flagging as a broader discovery-layer item rather than a unit-2 change.

iurii-ssv (approved):
LGTM

PR #2928, opened 2026-07-06
Exporter refactoring #3: Move websocket query bridge into exporter v1 API

julienharbulot:
## Summary

This PR continues the exporter refactor by moving the legacy websocket `/query` bridge out of `operator` and into `exporter/v1/api`, where the legacy exporter websocket API is owned.

There is no intended behavior change. The goal is clearer package ownership with a small, reviewable diff.

Actually diff is smaller than it looks because the code in `exporter/v1/api/query_bridge.go` was simply moved from `operator/node.go`.

## What Changed

  - Moved websocket `/query` dispatch and archive-read bridging into `exporter/v1/api`.
  - Moved the related websocket query tests with that code.
  - Changed `operator` to only start a preconfigured websocket server. This makes package boundaries clearer.
  - Kept `cli/operator` as the composition root that wires storage, exporter reads, and websocket handlers together.

  ## Why This Is Better

  Before this change, `operator` both started the websocket server and understood exporter websocket query semantics. That made runtime
  orchestration depend on exporter API details.

  After this change:

  - legacy websocket query behavior lives under `exporter/v1`
  - `operator` no longer imports exporter API code
  - root `exporter` remains independent from `operator`
  - future exporter cleanup can happen with less cross-package coupling

  ## Validation

  - `make lint`
  - `go test -tags "blst_enabled lfs" ./exporter/v1/... ./operator ./cli/operator`
  - `make docker-unit-test`

### DIFF
--- a/exporter/v1/api/query_bridge.go
+++ b/exporter/v1/api/query_bridge.go
@@ -1,0 +1,179 @@
+package api
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/hashicorp/go-multierror"
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+	"go.uber.org/zap"
+
+	exportercore "github.com/ssvlabs/ssv/exporter"
+	dutytracer "github.com/ssvlabs/ssv/exporter/dutytracer"
+	exporterstore "github.com/ssvlabs/ssv/exporter/store"
+	"github.com/ssvlabs/ssv/ibft/storage"
+	"github.com/ssvlabs/ssv/observability/log/fields"
+	"github.com/ssvlabs/ssv/protocol/v2/message"
+)
+
+type traceDecidedsReader interface {
+	TraceDecidedsCore(*exportercore.DecidedsQuery) (*exportercore.TraceDecidedsResult, *multierror.Error)
+}
+
+type validatorIndexReader interface {
+	ValidatorIndex(spectypes.ValidatorPK) (phase0.ValidatorIndex, bool)
+}
+
+// HandleQueryRequests dispatches websocket query messages to either the legacy
+// participant-store path or the archive exporter-core compatibility path.
+func (h *Handler) HandleQueryRequests(store *storage.ParticipantStores, exporterRead traceDecidedsReader, validators validatorIndexReader, domain spectypes.DomainType, nm *NetworkMessage) {
+	if nm.Err != nil {
+		nm.Msg = Message{
+			Type: TypeError,
+			Data: []string{fmt.Sprintf("could not parse network message: %v", nm.Err)},
+		}
+	}
+	h.logger.Debug("got incoming export request",
+		zap.String("type", string(nm.Msg.Type)))
+
+	switch nm.Msg.Type {
+	case TypeDecided:
+		// In exporter archive mode we serve decided queries via exporter core.
+		// Fall back to legacy qbft storage when exporter isn't wired.
+		if exporterRead != nil {
+			h.handleDecidedViaExporter(exporterRead, validators, domain, nm)
+			break
+		}
+		h.HandleParticipantsQuery(store, nm, domain)
+	case TypeError:
+		h.HandleErrorQuery(nm)
+	default:
+		h.HandleUnknownQuery(nm)
+	}
+}
+
+func (h *Handler) handleDecidedViaExporter(exporterRead traceDecidedsReader, validators validatorIndexReader, domain spectypes.DomainType, nm *NetworkMessage) {
+	res := Message{Type: nm.Msg.Type, Filter: nm.Msg.Filter}
+
+	pkBytes, err := hex.DecodeString(nm.Msg.Filter.PublicKey)
+	if err != nil {
+		h.logger.Warn("failed to decode validator public key", zap.Error(err))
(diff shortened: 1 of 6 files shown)


exporter/v1/api/query_bridge.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+	h.logger.Debug("got incoming export request",
+		zap.String("type", string(nm.Msg.Type)))
+
+	switch nm.Msg.Type {
+	case TypeDecided:
+		// In exporter archive mode we serve decided queries via exporter core.
+		// Fall back to legacy qbft storage when exporter isn't wired.
+		if exporterRead != nil {
```
iurii-ssv, line 38: **Typed-nil interface: this `!= nil` check is always true in non-archive modes, so this path panics instead of falling back to legacy qbft storage.**

`exporterRead` is the `traceDecidedsReader` interface parameter, but the caller in `cli/operator/node.go` passes the concrete `var exporterRead *exportercore.Exporter`, which is assigned only in `modeExporterArchive`. In `modeExporterStandard` / `modeOperator` it stays a **nil `*Exporter` wrapped in a non-nil interface**, so `exporterRead != nil` evaluates to `true`, `handleDecidedViaExporter` runs, and `exporterRead.TraceDecidedsCore(coreQuery)` (line 97) dereferences the nil pointer → panic in the WS query goroutine.

Before this refactor this was a `*Node` method checking the concrete field `n.exporterRead != nil` — a correct check that fell through to `HandleParticipantsQuery`. So this regresses the legacy participant-store path from "works" to "panics" whenever `WsAPIPort` is set outside archive mode. The moved tests don't catch it because `newWSQueryHarness` always builds a non-nil `coreExporter`.

**Suggested fix:** drop the `traceDecidedsReader` interface and take the concrete `*exportercore.Exporter`. The interface buys nothing here — this package already imports `exportercore`, and the test passes a real `*exportercore.Exporter` rather than a mock — and the concrete type makes the nil check behave correctly:

```go
func (h *Handler) HandleQueryRequests(store *storage.ParticipantStores, exporterRead *exportercore.Exporter, validators validatorIndexReader, domain spectypes.DomainType, nm *NetworkMessage) {
```
```go
func (h *Handler) handleDecidedViaExporter(exporterRead *exportercore.Exporter, validators validatorIndexReader, domain spectypes.DomainType, nm *NetworkMessage) {
```

then delete the now-unused `traceDecidedsReader` interface. The existing test compiles unchanged.

Alternatively, keep the interface but have the caller pass an untyped `nil` when the exporter isn't wired (`if exporterRead != nil { …exporterRead… } else { …nil… }`) — that fixes the crash too, but leaves the same footgun for the next caller.

julienharbulot: Great catch, thank you. I’m going to mark this as draft while my local exporter finishes resyncing, then I’ll fix this and run the manual verification pass before un-drafting it again.

julienharbulot: Done

julienharbulot:
Tested manually in standard and archive mode.

iurii-ssv (dismissed):
Good stuff!

**Non-blocking cleanup suggestion:** the not-found filtering carried over into `query_bridge.go` is provably a no-op at this call site, so the move is a good opportunity to drop it rather than carry it into a fresh file.

- `TraceDecidedsCore` already strips duty-not-found errors as its final step ([exporter/decided.go#L73-L74](https://github.com/ssvlabs/ssv/blob/stage/exporter/decided.go#L73-L74) — "by design, not found duties are expected and not considered as API errors"), and both of its early-return paths emit only `*ValidationError`. So no `dutytracer.ErrNotFound` / `store.ErrNotFound` can ever reach the [`filterOutDutyNotFoundErrors(errs)` call in the bridge](https://github.com/ssvlabs/ssv/blob/refactor/exporter-v1-websocket-query-bridge/exporter/v1/api/query_bridge.go#L97) — it never filters anything.
- The moved [`isNotFoundError` / `filterOutDutyNotFoundErrors`](https://github.com/ssvlabs/ssv/blob/refactor/exporter-v1-websocket-query-bridge/exporter/v1/api/query_bridge.go#L158-L175) also duplicate the unexported helpers in [exporter/exporter.go#L52-L67](https://github.com/ssvlabs/ssv/blob/stage/exporter/exporter.go#L52-L67). The pre-move code justified the copy with "mirrors exportercore.filterOutDutyNotFoundErrors to make future WS refactoring simpler" — this PR is that refactoring, so the copies can just go.

Behavior-identical replacement in `handleDecidedViaExporter` (nil `errs` is already handled by the `ErrorOrNil()` guard):

```go
var unexpectedErr error
if errs.ErrorOrNil() != nil {
	for _, e := range errs.Errors {
		// Preserve legacy WS leniency: treat validation errors as "no messages".
		if isExporterValidationError(e) {
			continue
		}
		unexpectedErr = e
	}
}
```

…then delete `isNotFoundError` + `filterOutDutyNotFoundErrors` and the now-unused `dutytracer` / `exporterstore` imports (~25 lines). Since removal is provably behavior-preserving, it still fits this PR's "no behavior change" contract — but equally fine as a follow-up in the series if you'd rather keep this one purely mechanical.

momosh-ssv (dismissed):

iurii-ssv (approved):

momosh-ssv (approved):

PR #2930, opened 2026-07-06
convergence(unit 4): fork-aware applyDynamicMaxPeers

momosh-ssv:
## Unit 4 of the boole-fork → stage convergence: fork-aware dynamic max-peers

Fifth slice, **stacked on unit 3** (#2929). Scope turned out much smaller than anticipated because stage is already ahead on most of what unit 4 was expected to port.

### The one real fork delta
`applyDynamicMaxPeers` (the peer-budget heuristic that raises `MaxPeers` to match how many committee subnets an operator serves) counted **Alan subnets only**. Boole makes it fork-aware: count each validator's **Boole** committee subnet, plus its **Alan** subnet before/through the transition (`!booleFork`), so the peer budget covers both subnet sets while the node is subscribed to both. Uses the `SSVShare.BooleCommitteeSubnet()/AlanCommitteeSubnet()` helpers added in unit 2. This also resolves the nit deferred from the unit-2 review.

### Already stage-ahead (nothing to port — verified)
- **#2866** (node.go split into thin/testable flow + errgroup shutdown) — stage already refactored node.go this way.
- **#2891** (config defaults in code / `ApplyDefaults`) — stage has `ApplyDefaults` across beacon/goclient, cli/config, eth/executionclient, network/p2p, exporter/config; **boole has none**.
- **Exporter `fixedSubnets=AllSubnets` default** — present at `node.go:364`.
- **#2914 teardown race** — stage re-solved it via **ctx-cancel-before-close** (documented in `newNode`/`close()`), rather than boole's `messageWorker.Close()`. Sound; no change needed.

### Testing
`go build ./...`, `go vet ./...` (main **+ ssvsigner**), `scripts/deadcode.sh`, `gofmt -l`, `golangci-lint ./cli/...`, and `go test ./cli/... -count=1` — all green.

_Small, low-risk change (a peer-count heuristic — not consensus/data-format), so no separate deep-review pass; it's the same fork-aware subnet pattern already reviewed in unit 2's context._


cli/operator/node.go
```diff
@@ -622,12 +622,20 @@ func (n *node) applyDynamicMaxPeers() {
...
+	// Alan and Boole subnets are in play, so budget peers for both.
+	booleFork := n.networkConfig.BooleForkAtSlot(n.networkConfig.EstimatedCurrentSlot())
 	for _, v := range myValidators {
-		subnet := networkcommons.AlanCommitteeSubnet(v.CommitteeID())
-		if !mySubnets.IsSet(subnet) {
+		if subnet := v.BooleCommitteeSubnet(); !mySubnets.IsSet(subnet) {
 			mySubnets.Set(subnet)
 			myActiveSubnets++
 		}
+		if !booleFork {
```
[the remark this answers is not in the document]
momosh-ssv: This matches boole-fork exactly (same `if !booleFork` gate on the Alan-subnet count). The effect is confined to the peer-count *heuristic*: during the single-slot transition window the node subscribes to both Alan and Boole committee topics but this counts only the Boole set, so `MaxPeers` can be sized slightly low for that one slot. Since it only bites a node that (re)starts inside the one-slot window and `MaxPeers` self-corrects as peers churn, keeping it faithful to the deployed reference rather than diverging here. Noting it as a candidate for a dedicated follow-up if we want the budget to track `subscribedSubnetsForCurrentEpoch` (count Alan when `!booleFork || InBooleTransitionWindow`).

cli/operator/node.go
```diff
@@ -622,12 +622,20 @@ func (n *node) applyDynamicMaxPeers() {
 	myValidators := n.nodeStorage.ValidatorStore().OperatorValidators(n.operatorDataStore.GetOperatorID())
 	mySubnets := networkcommons.Subnets{}
 	myActiveSubnets := 0
+	// Post-Boole-fork committees live on Boole subnets; before/through the transition both the
+	// Alan and Boole subnets are in play, so budget peers for both.
+	booleFork := n.networkConfig.BooleForkAtSlot(n.networkConfig.EstimatedCurrentSlot())
 	for _, v := range myValidators {
-		subnet := networkcommons.AlanCommitteeSubnet(v.CommitteeID())
-		if !mySubnets.IsSet(subnet) {
+		if subnet := v.BooleCommitteeSubnet(); !mySubnets.IsSet(subnet) {
```
iurii-ssv: `BooleCommitteeSubnet()` returns `UnknownSubnetId` (`math.MaxUint64`) for an empty committee. Because `Subnets.Set`/`IsSet` both no-op for indices ≥ `SubnetsCount` (128), this block would run `myActiveSubnets++` without ever recording the subnet in `mySubnets` — and each subsequent empty-committee validator repeats it, inflating the count and `MaxPeers`.

Guarded in practice (real shares are non-empty, per the helper's own comment), and the previous `AlanCommitteeSubnet(v.CommitteeID())` path was always `mod 128` so it couldn't hit this — it's newly introduced by switching to `BooleCommitteeSubnet`. The two-bitmap refactor suggested below removes it for free, since `Set` silently drops the out-of-range index.

cli/operator/node.go
```diff
@@ -622,12 +622,20 @@ func (n *node) applyDynamicMaxPeers() {
...
 			mySubnets.Set(subnet)
 			myActiveSubnets++
 		}
+		if !booleFork {
+			if alanSubnet := v.AlanCommitteeSubnet(); !mySubnets.IsSet(alanSubnet) {
```
iurii-ssv: `mySubnets` is a single 128-entry bitmap shared across both branches, but Boole and Alan subnets are distinct gossip topics (`/ssv/<net>/boole/<n>` vs `ssv.v2.<n>`) with independent meshes. If a validator's Alan subnet number coincides with a Boole subnet number already set by *any* validator, `!mySubnets.IsSet(alanSubnet)` is false, this branch skips `myActiveSubnets++`, and the budget misses a subnet the node is genuinely subscribed to — sizing `MaxPeers` below the "cover both subnet sets" intent.

This is a different mechanism from the one-slot subsequent-window under-budget already discussed on this PR: it under-counts even in the pre-fork period when both branches run.

Counting the two sets independently avoids both this and the `UnknownSubnetId` case noted above:

```go
var booleSubnets, alanSubnets networkcommons.Subnets
	booleSubnets.Set(v.BooleCommitteeSubnet())
	if !booleFork {
		alanSubnets.Set(v.AlanCommitteeSubnet())
	}
}
myActiveSubnets := booleSubnets.ActiveCount() + alanSubnets.ActiveCount()
```

iurii-ssv (commented):
Couple things to check out

cli/operator/node.go
momosh-ssv: Good catch — confirmed. Folded into the two-bitmap refactor: `Set` drops the out-of-range `UnknownSubnetId`, so an empty committee can no longer `active++` without recording. Added a test pinning it. Fixed in 4d006b0f7.

cli/operator/node.go
momosh-ssv: Agreed, real pre-fork under-count independent of the one-slot window. Now counting Boole and Alan in separate bitmaps and summing `ActiveCount()`, so a subnet-number collision no longer dedupes across the two topic sets. Added a test that searches for an actual Boole/Alan collision and asserts it counts 2. Fixed in 4d006b0f7.

iurii-ssv (approved):
LGTM, just maybe record Greptile suggestions as those seem deferred for some kind of a followup PR (vs outright rejected) - so we can address those too later on.

PR #2931, opened 2026-07-06
convergence(unit 5a): fork-aware round-robin proposer

momosh-ssv:
## Unit 5a of the boole-fork → stage convergence: fork-aware round-robin proposer

First slice of unit 5 (runner/qbft), **stacked on unit 4** (#2930). Foundation layer: the **consensus-critical** QBFT proposer/leader selection, fork-gated.

### What it does
- **`protocol/v2/qbft/round_robin_proposer.go`**: `RoundRobinProposer` changes signature to `(height, round, committee, netCfg)` and adds **ETH-epoch variability** (so 4-operator committees don't repeat leaders every epoch, since 32%4=0) + **committee sorting** (proposer independent of input order). The old implementation is preserved as `RoundRobinProposerPreBooleFork`. Ported exactly from boole-fork.
- **`operator/validator/controller.go`** (`buildController.ProposerF`): fork-gated — `BooleForkAtSlot(state.Height) ? RoundRobinProposer(new) : RoundRobinProposerPreBooleFork`. (The `NewAggregatorCommitteeRunner` construction stays for 5c.)
- **`message/validation/consensus_validation.go`** leader check: fork-gated identically, and the unit-2 helper `roundRobinProposer` is renamed `roundRobinProposerPreBooleFork`. This keeps validation and the actual proposer in agreement per-fork — otherwise post-fork proposals would be rejected as "signer not leader".

### Scope boundary
Foundation only. Aggregator-committee helpers (the `value_check` checker, `msg.go`/`runner_role.go` role bits) are **deferred to their consumers in 5b/5c** to avoid landing unreachable code (deadcode-lint).

### Testing
Co-lands `round_robin_proposer_test.go` (pre-fork matches stage logic; post-fork epoch offset; committee-order independence). The unit-3 post-fork acceptance test now builds a proposal signed by the fork-aware leader (stronger coverage). Full matrix green: `go build ./...`, `go vet` (main **+ ssvsigner**), `deadcode`, `gofmt`, `golangci-lint`, and qbft / message-validation / operator-validator tests.

> ⚠️ **Consensus-critical** (proposer/leader selection). A `code-reviewer-deep` pass is being run in parallel; findings will be addressed as follow-up commits.

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1048,8 +1048,12 @@ func SetupCommitteeRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
 			ProposerF: func(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
-				leader := qbft.RoundRobinProposer(state, round)
-				return leader
+				if options.NetworkConfig.BooleForkAtSlot(phase0.Slot(state.Height)) {
+					committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
+					return qbft.RoundRobinProposer(state.Height, round, committee, options.NetworkConfig)
+				}
+
+				return qbft.RoundRobinProposerPreBooleFork(state, round)
 			},
 			Network:     options.Network,
 			CutOffRound: roundtimer.CutOffRound,
@@ -1113,8 +1117,12 @@ func SetupRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
 			ProposerF: func(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
-				leader := qbft.RoundRobinProposer(state, round)
-				return leader
+				if options.NetworkConfig.BooleForkAtSlot(phase0.Slot(state.Height)) {
+					committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
+					return qbft.RoundRobinProposer(state.Height, round, committee, options.NetworkConfig)
+				}
+
+				return qbft.RoundRobinProposerPreBooleFork(state, round)
 			},
 			Network:     options.Network,
 			CutOffRound: roundtimer.CutOffRound,
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1048,12 +1048,8 @@ func SetupCommitteeRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
 			ProposerF: func(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
-				if options.NetworkConfig.BooleForkAtSlot(phase0.Slot(state.Height)) {
-					committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
-					return qbft.RoundRobinProposer(state.Height, round, committee, options.NetworkConfig)
-				}
-
-				return qbft.RoundRobinProposerPreBooleFork(state, round)
+				committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
+				return qbft.Proposer(state.Height, round, committee, options.NetworkConfig)
 			},
 			Network:     options.Network,
 			CutOffRound: roundtimer.CutOffRound,
@@ -1117,12 +1113,8 @@ func SetupRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
 			ProposerF: func(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
-				if options.NetworkConfig.BooleForkAtSlot(phase0.Slot(state.Height)) {
-					committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
-					return qbft.RoundRobinProposer(state.Height, round, committee, options.NetworkConfig)
-				}
-
-				return qbft.RoundRobinProposerPreBooleFork(state, round)
+				committee := ssvtypes.OperatorIDsFromOperators(state.CommitteeMember.Committee)
+				return qbft.Proposer(state.Height, round, committee, options.NetworkConfig)
 			},
 			Network:     options.Network,
 			CutOffRound: roundtimer.CutOffRound,
--- a/protocol/v2/qbft/round_robin_proposer.go
+++ b/protocol/v2/qbft/round_robin_proposer.go
@@ -1,15 +1,16 @@
 package qbft
 
 import (
+	"sort"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
-// RoundRobinProposer returns the proposer for the round.
-// Each new height starts with the first proposer and increments by 1 with each following round.
-// Each new height has a different first round proposer which is +1 from the previous height.
-// First height starts with index 0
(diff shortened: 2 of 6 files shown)


operator/validator/controller.go
```diff
@@ -1048,8 +1048,12 @@ func SetupCommitteeRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
 			ProposerF: func(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
-				leader := qbft.RoundRobinProposer(state, round)
-				return leader
+				if options.NetworkConfig.BooleForkAtSlot(phase0.Slot(state.Height)) {
```
iurii-ssv: This fork-gated branch now exists in three identical copies: here, in `SetupRunners` below, and in `validateQBFTLogic` (message validation). The PR's key invariant is that validation and the actual proposer agree per fork — but with three hand-synced copies that agreement holds by convention only: a future edit to one gate (fork predicate or committee derivation) that misses the others desyncs validation from the runners, and post-fork proposals get rejected as `ErrSignerNotLeader`.

Consider a single fork-aware entry point next to the two implementations in the `qbft` package, e.g. `qbft.Proposer(height, round, committee []spectypes.OperatorID, netCfg)` that does the `BooleForkAtSlot` branch internally (the pre-fork body works on `[]OperatorID` just as well — message validation already carries such a copy, and per-caller committee order is preserved either way). All three call sites then collapse to one call and the agreement becomes structural.

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +19,37 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
+	committee []spectypes.OperatorID,
+	netCfg networkConfig,
+) spectypes.OperatorID {
+	// Ensure committee is sorted to avoid proposer depending on input order.
+	if !sort.SliceIsSorted(committee, func(i, j int) bool { return committee[i] < committee[j] }) {
+		sorted := make([]spectypes.OperatorID, len(committee))
+		copy(sorted, committee)
+		sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
+		committee = sorted
+	}
```
iurii-ssv, line 45: The repo is on Go 1.26 and already uses the stdlib `slices` package widely, so this can drop the reflection-based `sort` calls and the hand-written less funcs (this runs on the per-message validation hot path):

```go
// Ensure committee is sorted to avoid proposer depending on input order.
if !slices.IsSorted(committee) {
	committee = slices.Clone(committee)
	slices.Sort(committee)
}
```

(with `sort` → `slices` in imports)

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -1,15 +1,16 @@
...
 	specqbft "github.com/ssvlabs/ssv-spec/qbft"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 )
 
-// RoundRobinProposer returns the proposer for the round.
-// Each new height starts with the first proposer and increments by 1 with each following round.
-// Each new height has a different first round proposer which is +1 from the previous height.
-// First height starts with index 0
-func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.OperatorID {
+// Deprecated - only for pre-Boole fork
```
iurii-ssv: Go tooling (staticcheck/gopls) only recognizes the deprecation marker in the `Deprecated: ` form (with a colon) — worth matching it so remaining callers get flagged automatically when the pre-fork path is eventually removed:

```suggestion
// Deprecated: only for pre-Boole fork.
```

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +19,37 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
 	return state.CommitteeMember.Committee[index].OperatorID
 }
+
+type networkConfig interface {
+	EstimatedEpochAtSlot(slot phase0.Slot) phase0.Epoch
+}
+
+// RoundRobinProposer returns the proposer for the round.
+// Each new height starts with the first proposer and increments by 1 with each following round.
+// Each new height has a different first round proposer which is +1 from the previous height.
```
iurii-ssv: This carried-over sentence is no longer exact for the fork-aware version: when the height crosses an epoch boundary, the first-round proposer advances by +2 (height increment plus epoch-offset increment) — e.g. slots 31→32 with a 4-operator committee jump from index 3 to index 1. Since this is consensus-critical, the comment should describe the actual behavior:

```suggestion
// Each new height has a different first round proposer which is +1 from the previous height
// (+2 when the height crosses an epoch boundary, as the epoch offset advances as well).
```

iurii-ssv (approved):
LGTM, just potential improvements to consider (could be a follow-up PR)

operator/validator/controller.go
momosh-ssv: Done in b5c95d35b — collapsed all three copies into a single `qbft.Proposer(height, round, committee []OperatorID, netCfg)` that does the `BooleForkAtSlot` branch internally. Both `ProposerF` closures in controller.go and the leader check in `validateQBFTLogic` now call it, so validation and the runners agree on the leader by construction. The pre-Boole body is folded into an unexported `roundRobinProposerPreBooleFork([]OperatorID)` in the same file. Verified behavior-preserving: `OperatorIDsFromOperators(Committee)` preserves order and the pre-fork branch runs the identical math on the same ordered slice.

protocol/v2/qbft/round_robin_proposer.go
momosh-ssv: Done in b5c95d35b — dropped the reflection-based `sort` for stdlib `slices.IsSorted`/`slices.Clone`/`slices.Sort` (`OperatorID` is a `uint64`, so it satisfies `cmp.Ordered`).

protocol/v2/qbft/round_robin_proposer.go
momosh-ssv: Done in b5c95d35b — applied your suggested wording describing the +2 first-round advance across an epoch boundary.

protocol/v2/qbft/round_robin_proposer.go
momosh-ssv: Addressed in b5c95d35b by removing the exported pre-fork function entirely — the pre-Boole path is now an unexported helper reached only through `qbft.Proposer`, so there is no exported symbol left to deprecate. The single entry point makes the eventual removal a one-liner (drop the `BooleForkAtSlot` branch and the helper).

protocol/v2/qbft/round_robin_proposer.go
```diff
@@ -18,3 +19,37 @@ func RoundRobinProposer(state *specqbft.State, round specqbft.Round) spectypes.O
...
+) spectypes.OperatorID {
+	// Ensure committee is sorted to avoid proposer depending on input order.
+	if !sort.SliceIsSorted(committee, func(i, j int) bool { return committee[i] < committee[j] }) {
+		sorted := make([]spectypes.OperatorID, len(committee))
+		copy(sorted, committee)
+		sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
+		committee = sorted
+	}
+
+	ethEpoch := uint64(netCfg.EstimatedEpochAtSlot(phase0.Slot(height)))
```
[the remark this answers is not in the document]
momosh-ssv: This divide (`EstimatedEpochAtSlot` → `slot / SlotsPerEpoch`) is the same one the pre-fork path already performs and matches boole-fork. `SlotsPerEpoch` is a required, non-zero field populated when the network config is loaded, so a zero value would be a misconfiguration that fails much earlier — not a value reachable from a live proposer selection. Keeping it faithful to the reference rather than adding a guard on the hot path.

iurii-ssv (approved):
👍

PR #2936, opened 2026-07-09
Exporter refactoring #4: Decouple validator message tracing from exporter

julienharbulot:
This PR continues the exporter refactoring by removing exporter-specific types from the validator controller.

## What changed
- Replaced `operator/validator` dependencies on exporter config and `dutytracer.Collector` with a local `ExporterMode` flag and `MessageTraceHandler` callback.
- Kept archive exporter construction in `cli/operator`, where exporter components are already assembled, and pass only the callback into the validator controller.
- Updated validator common options and nearby tests to use the narrower boundary.

## Why
The validator controller only needs to know whether it is running in exporter mode and how to trace consensus / partial-signature messages for validators this node does not run. It should not depend on exporter implementation packages.

This keeps exporter details in exporter wiring, preserves the existing standard/archive behavior, and makes the diff mostly mechanical for review.

## Validation
- `go test -tags "blst_enabled lfs" ./operator/validator ./protocol/v2/ssv/validator ./cli/operator ./eth/ethtest ./eth/eventsyncer ./eth/eventhandler`
- `go test -tags "blst_enabled lfs" ./cli/operator ./operator/validator`
- `make lint`

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -18,8 +18,6 @@ import (
 	"github.com/ssvlabs/ssv/ssvsigner/ekm"
 
 	"github.com/ssvlabs/ssv/doppelganger"
-	exporterconfig "github.com/ssvlabs/ssv/exporter/config"
-	dutytracer "github.com/ssvlabs/ssv/exporter/dutytracer"
 	"github.com/ssvlabs/ssv/ibft/storage"
 	"github.com/ssvlabs/ssv/message/validation"
 	"github.com/ssvlabs/ssv/network"
@@ -52,6 +50,13 @@ import (
 
 //go:generate go tool -modfile=../../tool.mod mockgen -package=mocks -destination=./mocks/controller.go -source=./controller.go
 
+// MessageTraceHandler traces consensus and partial-signature messages observed for validators this node does not run.
+type MessageTraceHandler func(
+	ctx context.Context,
+	msg *queue.SSVMessage,
+	verifySig func(*spectypes.PartialSignatureMessages) error,
+) error
+
 // ControllerOptions for creating a validator controller
 type ControllerOptions struct {
 	Context                        context.Context
@@ -63,6 +68,7 @@ type ControllerOptions struct {
 	Network                        P2PNetwork
 	Beacon                         beaconprotocol.BeaconNode
 	FullNode                       bool `yaml:"FullNode" env:"FULLNODE" env-description:"Store complete message history instead of just latest messages"`
+	ExporterMode                   bool
 	BeaconSigner                   ekm.BeaconSigner
 	OperatorSigner                 ssvtypes.OperatorSigner
 	OperatorDataStore              operatordatastore.OperatorDataStore
@@ -70,7 +76,7 @@ type ControllerOptions struct {
 	ValidatorRegistrationSubmitter runner.ValidatorRegistrationSubmitter
 	NewDecidedHandler              qbftcontroller.NewDecidedHandler
 	DutyRoles                      []spectypes.BeaconRole
-	DutyTraceCollector             *dutytracer.Collector
+	MessageTraceHandler            MessageTraceHandler
 	StorageMap                     *storage.ParticipantStores
 	ValidatorStore                 registrystorage.ValidatorStore
 	MessageValidator               validation.MessageValidator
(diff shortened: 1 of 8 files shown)


julienharbulot:
**Claude Code review:**

> Verdict: yes, this improves the codebase — the refactor is correct and well-contained
>
> What the change does well. Before, operator/validator imported both exporter/config and exporter/dutytracer, and protocol/v2/ssv/validator imported exporter/config — the validator layer knew about exporter internals. Now both packages are exporter-free: the controller depends on a MessageTraceHandler func type it defines itself (the consumer-side abstraction, idiomatic Go), and the protocol options carry a plain ExporterMode bool. Exporter knowledge now lives only at the composition root (cli/operator, operator/node.go), which is exactly where config-driven wiring belongs. The diff is mechanical and small, and node_test.go gained assertions that pin the new wiring per mode. gopls reports no diagnostics, the generated mocks reference none of the removed types, and all NewController/NewCommonOptions call sites are updated.
> 
> Behavior preservation — verified, no bugs found. I checked each semantic substitution:
> 
> - valOpts.ExporterMode = res.isExporter() is equivalent to the old ExporterOptions.Enabled: resolveMode maps Enabled=false to modeOperator and rejects invalid modes before newNode runs, so isExporter() ⇔ Enabled at that point.
> - The new c.messageTraceHandler != nil check in handleWorkerMessages is equivalent to the old Mode == ModeArchive check: the handler is wired only in modeExporterArchive, and handleWorkerMessages is only reachable via messageWorker.TryEnqueue, which sits inside the ExporterMode branch of handleRouterMessages — same gate as before. It's actually more robust: the old code would nil-panic on c.traceCollector.Collect if archive mode ever ran without a collector; the new code falls through to handleNonCommitteeMessages.
> - Collector.Collect's signature matches MessageTraceHandler exactly, and the c.ctx (vs the ctx parameter) usage is unchanged.

julienharbulot:
Manually tested in archive and standard modes.

iurii-ssv (dismissed):
LGTM

operator/validator/controller.go
```diff
@@ -398,12 +404,10 @@ func (c *Controller) handleWorkerMessages(ctx context.Context, msg network.Decod
 		ncv = item.Value()
 	}
 
-	if c.validatorCommonOpts.ExporterOptions.Mode == exporterconfig.ModeArchive {
-		// use new exporter functionality
-		return c.traceCollector.Collect(c.ctx, ssvMsg, ncv.VerifySig)
+	if c.messageTraceHandler != nil {
```
momosh-ssv, line 407: Might be worth adding a small targeted test for this new nil-dispatch. `TestHandleNonCommitteeMessages` swaps in a stub via `UseHandler`, so neither branch here (handler set -> dispatches to it; handler nil -> falls back to `handleNonCommitteeMessages`) is actually exercised through `handleWorkerMessages`. Not a regression — the old `Mode == ModeArchive` check had the same gap — but this is the exact boundary the PR introduces, so it seems like the natural place to lock it in.

ovidiu-ssv-labs (commented):
**Finding 1 · [MINOR] No direct unit test exercises the messageTraceHandler branch in handleWorkerMessages** — `operator/validator/controller_test.go:172`

handleWorkerMessages' new dispatch logic (`if c.messageTraceHandler != nil { return c.messageTraceHandler(c.ctx, ssvMsg, ncv.VerifySig) }`, controller.go:407-409) is the actual runtime contract this PR introduces at the controller boundary, but it isn't exercised by a controller-level unit test with a real (or stub) handler. 

**Suggested fix:** Add a test alongside TestHandleNonCommitteeMessages that sets `ctr.messageTraceHandler` to a stub function recording its (ctx, msg, verifySig) args and a sentinel return, drives a consensus/partial-signature message through `handleRouterMessages`/`handleWorkerMessages` (or calls `handleWorkerMessages` directly), and asserts the stub was called exactly once with the expected message and that `handleNonCommitteeMessages`'s side effects (e.g. `ncv.SaveRoots`/`ncv.ProcessMessage`) did NOT run. Example sketch:
```go
func TestHandleWorkerMessages_UsesMessageTraceHandler(t *testing.T) {
    var got *queue.SSVMessage
    ctr := setupController(t, logger, MockControllerOptions{
        validatorCommonOpts: &validator.CommonOptions{},
    })
    ctr.messageTraceHandler = func(ctx context.Context, msg *queue.SSVMessage, _ func(*spectypes.PartialSignatureMessages) error) error {
        got = msg
        return nil
    }
    err := ctr.handleWorkerMessages(t.Context(), someSSVMessage)
    require.NoError(t, err)
    require.Same(t, someSSVMessage, got)
}
```

iurii-ssv (approved):

julienharbulot:
Signed the latest commit [Test validator message trace dispatch](https://github.com/ssvlabs/ssv/pull/2936/commits/f6c87dac6aafff0ac3decf157729ebd81a60cada) and force pushed. No content changed. (@iurii-ssv's approval still valid because content unchanged.) -- Iurii if you see this comment, please re-run the CI and assign the PR to me 🙏

momosh-ssv (approved):
Thanks for adding the dispatch test. Verified it locks in the handler branch and passes locally. LGTM.

PR #2937, opened 2026-07-09
convergence(unit 5c-exporter): aggregator-committee duty-trace query layer

momosh-ssv:
## convergence(unit 5c-exporter): aggregator-committee duty-trace query layer

Adds the **reader** half of the aggregator-committee duty-trace feature (the producer/runner, disk-store role-keying, SSZ model and migration_9 landed in units 3/5c). Ported from `boole-fork` PR #2683 (commit `d405c1a36`), adapted onto stage's `exporter/dutytracer` + `exporter/traces` structure (boole keeps it under `operator/dutytracer` as `package validator`).

### What changed
- **Collector** (`exporter/dutytracer/collector.go`): re-key `committeeTraces`/`inFlightCommittee` by `committeeTraceKey{committeeID, runnerRole}` so a `Committee` and an `AggregatorCommittee` duty for the same committee+slot no longer collide in cache. Ports the AC post-consensus root classification (`computeAggregatorCommitteePostConsensusRoles`, `getAggregatorSelectionRoot`, root-scoped quorum via `signersByRoot`/`classifyRootForPending`).
- **Store** (`store.go`): `GetCommitteeDuty`/`GetCommitteeDuties` now take `RunnerRole`; `GetCommitteeDecideds` keeps `BeaconRole` at the boundary and converts internally (signer-bucket filtering). **Eviction** groups flushed duties by role and persists under the correct per-role disk prefix.
- **Exporter core + HTTP API**: `roles` query filter + `role` response field on committee-traces; new `RunnerRole`/`RunnerRoleSlice` api type; `RunnerRoleFromString` + the `AGGREGATOR_COMMITTEE` case in `RunnerRoleToString`. OpenAPI regenerated.

### Scope / deferred
- `networkConfig` threading into `NewExporter` (validator-path fork gating only) is **deliberately deferred** — `NewExporter` signature unchanged.

### Review
- **code-reviewer-deep: PASS** — ported classification logic byte-faithful to `d405c1a36`; keying collision-free; role axes (`RunnerRole` vs `BeaconRole`) distinct; per-role eviction correct; caller fix preserves behavior.
- Coverage raised on the classification paths (0→71–88%) + a regression guard that Committee and AggregatorCommittee traces coexist for the same committeeID+slot.

### Known / faithful-to-reference (non-blocking)
1. ~~`RunnerRoleFromString` accepts non-committee role strings in the filter → they match nothing~~ **Fixed in `bf5b32e7d`:** the committee-traces `roles` filter now binds through `CommitteeRunnerRoleFromString` (COMMITTEE / AGGREGATOR_COMMITTEE only), so a non-committee role is rejected at the boundary as a 400 instead of surfacing the store's `unsupported committee runner role` error as a 500.
2. `computeAggregatorCommitteePostConsensusRoles` indexes `Contributors[i]` — relies on the AC consensus-data SSZ invariant (not introduced here).
3. `GetCommitteeDuties` can duplicate a duty during the eviction window — pre-existing in the base.

Stacks directly on `integration/boole-convergence`; no dependency on unit 5d.

### DIFF
--- a/api/handlers/exporter/committee_model.go
+++ b/api/handlers/exporter/committee_model.go
@@ -10,7 +10,7 @@ import (
 	"github.com/ssvlabs/ssv/api"
 	exportercore "github.com/ssvlabs/ssv/exporter"
 	"github.com/ssvlabs/ssv/exporter/traces"
-	obsutils "github.com/ssvlabs/ssv/observability/utils"
+	"github.com/ssvlabs/ssv/protocol/v2/message"
 )
 
 type CommitteeIDLengthError struct {
@@ -116,7 +116,7 @@ func toCommitteeTrace(t *traces.CommitteeDutyTrace) CommitteeTrace {
 	return CommitteeTrace{
 		// consensus trace
 		Slot:          uint64(t.Slot),
-		Role:          obsutils.FormatRunnerRole(t.Role),
+		Role:          message.RunnerRoleToString(t.Role),
 		Consensus:     toRounds(t.Rounds),
 		Decideds:      toDecideds(t.Decideds),
 		SyncCommittee: toCommitteePost(t.SyncCommittee),
--- a/api/handlers/exporter/committee_model.go
+++ b/api/handlers/exporter/committee_model.go
@@ -10,6 +10,7 @@ import (
 	"github.com/ssvlabs/ssv/api"
 	exportercore "github.com/ssvlabs/ssv/exporter"
 	"github.com/ssvlabs/ssv/exporter/traces"
+	obsutils "github.com/ssvlabs/ssv/observability/utils"
 )
 
 type CommitteeIDLengthError struct {
@@ -23,6 +24,8 @@ type CommitteeTracesRequest struct {
 	From uint64 `json:"from" format:"int64" minimum:"0"`
 	// To is the ending slot (inclusive).
 	To uint64 `json:"to" format:"int64" minimum:"0"`
+	// Roles is a comma-separated list of committee runner roles to include.
+	Roles api.RunnerRoleSlice `json:"roles" swaggertype:"array,string" enums:"COMMITTEE,AGGREGATOR_COMMITTEE"`
 	// CommitteeIDs is a comma-separated list of committee IDs (hex, 64 chars per ID).
 	CommitteeIDs api.HexSlice `json:"committeeIDs" swaggertype:"array,string" format:"hex" minLength:"64" maxLength:"64" pattern:"^[0-9a-f]{64}$"`
 }
@@ -46,6 +49,7 @@ func toCommitteeTracesQuery(r *CommitteeTracesRequest) (*exportercore.CommitteeT
 	q := &exportercore.CommitteeTracesQuery{
 		From:         r.From,
 		To:           r.To,
+		Roles:        toRunnerRoles(r.Roles),
 		CommitteeIDs: r.parseCommitteeIds(),
 	}
 
@@ -78,14 +82,18 @@ func toCommitteeTraceResponse(result *exportercore.CommitteeTracesResult, errs *
 type CommitteeTrace struct {
 	// Slot is the duty slot for this committee trace.
 	Slot uint64 `json:"slot" format:"int64"`
+	// Role is the committee runner role for this duty (e.g., COMMITTEE).
+	Role string `json:"role" example:"COMMITTEE" enums:"COMMITTEE,AGGREGATOR_COMMITTEE"`
 	// Consensus lists per-round QBFT messages observed for this committee.
 	Consensus []Round `json:"consensus"`
 	// Decideds lists decided messages emitted for this duty.
 	Decideds []Decided `json:"decideds"`
 
-	// SyncCommittee contains post-consensus messages for sync-committee duties.
+	// SyncCommittee contains post-consensus messages for sync-committee duties,
+	// including sync committee contribution signers.
 	SyncCommittee []CommitteeMessage `json:"sync_committee"`
-	// Attester contains post-consensus messages for attester duties.
+	// Attester contains post-consensus messages for attester duties,
+	// including aggregator signers.
 	Attester []CommitteeMessage `json:"attester"`
 
 	// CommitteeID is the 32-byte committee identifier (hex).
@@ -108,6 +116,7 @@ func toCommitteeTrace(t *traces.CommitteeDutyTrace) CommitteeTrace {
 	return CommitteeTrace{
 		// consensus trace
 		Slot:          uint64(t.Slot),
+		Role:          obsutils.FormatRunnerRole(t.Role),
 		Consensus:     toRounds(t.Rounds),
 		Decideds:      toDecideds(t.Decideds),
 		SyncCommittee: toCommitteePost(t.SyncCommittee),
--- a/exporter/committee.go
+++ b/exporter/committee.go
@@ -25,7 +25,7 @@ func (e *Exporter) CommitteeTracesCore(request *CommitteeTracesQuery) (*Committe
 	cids := request.CommitteeIDs
 	for s := request.From; s <= request.To; s++ {
 		slot := phase0.Slot(s)
-		duties, err := e.getCommitteeDutiesForSlot(slot, cids)
+		duties, err := e.getCommitteeDutiesForSlot(slot, cids, request.Roles...)
 		all = append(all, duties...)
 		errs = multierror.Append(errs, err)
 	}
@@ -46,23 +46,30 @@ func validateCommitteeRequest(request *CommitteeTracesQuery) error {
 	return nil
 }
 
-func (e *Exporter) getCommitteeDutiesForSlot(slot phase0.Slot, committeeIDs []spectypes.CommitteeID) ([]*traces.CommitteeDutyTrace, error) {
+func (e *Exporter) getCommitteeDutiesForSlot(slot phase0.Slot, committeeIDs []spectypes.CommitteeID, roles ...spectypes.RunnerRole) ([]*traces.CommitteeDutyTrace, error) {
 	if len(committeeIDs) == 0 {
-		duties, err := e.traceStore.GetCommitteeDuties(slot)
-		return duties, err
+		return e.traceStore.GetCommitteeDuties(slot, roles...)
+	}
+
+	if len(roles) == 0 {
+		roles = []spectypes.RunnerRole{
+			spectypes.RoleCommittee,
+			spectypes.RoleAggregatorCommittee,
+		}
 	}
 
 	duties := make([]*traces.CommitteeDutyTrace, 0, len(committeeIDs))
 
 	var errs *multierror.Error
 	for _, cmtID := range committeeIDs {
-		duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID)
-		if err != nil {
-			e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
-			errs = multierror.Append(errs, err)
-			continue
+		for _, role := range roles {
+			duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID, role)
+			if err != nil {
+				errs = multierror.Append(errs, err)
+				continue
+			}
+			duties = append(duties, duty)
 		}
-		duties = append(duties, duty)
 	}
 	return duties, errs.ErrorOrNil()
 }
--- a/exporter/committee.go
+++ b/exporter/committee.go
@@ -65,6 +65,11 @@ func (e *Exporter) getCommitteeDutiesForSlot(slot phase0.Slot, committeeIDs []sp
 		for _, role := range roles {
 			duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID, role)
 			if err != nil {
+				// not-found is expected (we probe every role per committee) and is
+				// filtered upstream; only log genuine storage failures.
+				if !isNotFoundError(err) {
+					e.logger.Error("error getting committee duty", zap.Error(err), fields.Slot(slot), fields.CommitteeID(cmtID))
+				}
 				errs = multierror.Append(errs, err)
 				continue
 			}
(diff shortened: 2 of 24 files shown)


api/handlers/exporter/committee_model.go
```diff
@@ -108,6 +116,7 @@ func toCommitteeTrace(t *traces.CommitteeDutyTrace) CommitteeTrace {
 	return CommitteeTrace{
 		// consensus trace
 		Slot:          uint64(t.Slot),
+		Role:          obsutils.FormatRunnerRole(t.Role),
```
iurii-ssv: Nit (maintainability, non-blocking): the response `role` value is produced here by `obsutils.FormatRunnerRole` (i.e. `spectypes.RunnerRole.String()` with `_RUNNER` trimmed), whereas the `roles` request filter and the documented OpenAPI enums go through `message.RunnerRoleToString` / `RunnerRoleFromString`. They happen to agree today (`COMMITTEE` / `AGGREGATOR_COMMITTEE`), so this isn't a bug — but the response vocabulary and the filter vocabulary are now two independent code paths that must stay in lockstep for a client to round-trip a returned `role` back into `roles`. Consider serializing this field via `message.RunnerRoleToString` so there's a single source of truth.

exporter/dutytracer/collector.go
```diff
@@ -752,31 +886,61 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
+					if rolesByRoot, err := c.computeAggregatorCommitteePostConsensusRoles(ctx, slot, msg.SignedSSVMessage.FullData); err == nil {
+						trace.aggPostConsensusRoles = rolesByRoot
+						trace.flushPending()
+						// Check quorum for all validators after flushing pending signatures.
+						// This ensures quorum detection happens immediately when signatures that
+						// arrived before the proposal are reclassified into role buckets.
+						c.checkQuorumAfterFlush(c.logger, committeeID, slot, trace)
+					} else {
+						// CRITICAL: If we fail to compute role roots, pending signatures will be dropped.
+						pendingCount := 0
```
iurii-ssv: Nit (DRY, non-blocking): this `pendingCount` tally plus the `CRITICAL: ... pending signatures will be dropped` log is identical to the `RoleCommittee` arm above — only the log message differs. Consider extracting the count into a small helper (e.g. `countPending(trace.pendingByRoot)`) shared by both arms so the two copies can't drift.

iurii-ssv (commented):
Just couple minor suggestions

ovidiu-ssv-labs (commented):
The core feature is well-engineered and correct: the {committeeID, runnerRole} cache keying is genuinely collision-free, the RunnerRole/BeaconRole conversions are consistent in both directions, per-role eviction persists under the right disk prefix, and the ported AC classification logic is faithful and well-tested (all affected packages build and tests pass). The one real defect is that the new `roles` query filter accepts parseable-but-non-committee role strings and turns them into an HTTP 500 rather than the benign 'matches nothing' the PR claims.  [verdict: with_fixes]

> 1 finding — inline on the diff.

protocol/v2/message/msg.go
```diff
@@ -46,39 +56,63 @@ func QBFTMsgTypeToString(mt specqbft.MessageType) string {
...
-	case "VOLUNTARY_EXIT":
+	case roleVoluntaryExit:
 		return spectypes.BNRoleVoluntaryExit, nil
 	default:
 		return 0, fmt.Errorf("unknown role: %s", s)
 	}
 }
 
+// RunnerRoleFromString returns RunnerRole from string.
+func RunnerRoleFromString(s string) (spectypes.RunnerRole, error) {
```
ovidiu-ssv-labs, line 79: **Finding 1 · [IMPORTANT] roles filter 500s on parseable-but-non-committee roles; PR's 'matches nothing' claim is wrong**

The PR lists as non-blocking quirk #1: "RunnerRoleFromString accepts non-committee role strings in the filter -> they match nothing." I traced this end-to-end and it does not match nothing — it returns an HTTP 500.

`RunnerRoleFromString` (msg.go:79) accepts seven strings, five of which are NOT committee runner roles: `AGGREGATOR` (ssvtypes.RoleAggregator = RunnerRole(1)), `PROPOSER` (2), `SYNC_COMMITTEE_CONTRIBUTION` (3), `VALIDATOR_REGISTRATION`, `VOLUNTARY_EXIT`. `RunnerRoleSlice.Bind` (api/types.go:149) uses this directly with no committee-only restriction, and `validateCommitteeRequest` only checks From<=To — so e.g. `GET /v1/exporter/traces/committee?from=N&to=N&roles=AGGREGATOR` binds successfully.

That value then flows to the store: `getCommitteeDutiesForSlot` -> `GetCommitteeDuties(slot, RunnerRole(1))` -> the disk store loops the role and calls `makeCommitteeSlotRolePrefix` -> `CommitteeRunnerRoleToPrefix(RunnerRole(1))` which returns `fmt.Errorf("unsupported committee runner role: %s", role)` (exporter/store/store.go:44). That error is appended to the multierror and returned. It is NOT a NotFound error, so `filterOutDutyNotFoundErrors` (exporter/committee.go:34) does not strip it. With no matching traces, `CommitteeTraces` hits `len(result.Traces) == 0 && errs.ErrorOrNil() != nil` and returns `http.StatusInternalServerError` (api/handlers/exporter/committee_http.go:45-47). The committeeID path 500s identically via `GetCommitteeDuty`.

This is newly reachable, not pre-existing: the `roles` runner-role filter, `RunnerRoleFromString`, and `RunnerRoleSlice` are all introduced by this PR (the base committee-traces API had no runner-role filter). The 'faithful to reference' framing doesn't apply on this stage base. It's also untested: the HTTP tests use `mockTraceStore`, whose `GetCommitteeDutiesFunc`/`GetCommitteeDutyFunc` never reproduce the `CommitteeRunnerRoleToPrefix` error, so the 500 path has no coverage. It's a client-triggerable 5xx on user input (should be a 400 or an empty 200), not a security issue, but the PR's stated behavior for this quirk is materially incorrect and should be fixed or the claim corrected.

Not covered by the PR's tests. The PROPOSER+500 cases in `exporter_test.go` are the decideds/validator endpoints (BeaconRole, where PROPOSER is legitimate) using mocked stores with forced errors. The committee-traces handler tests also mock dutyTraceStore, so the real `CommitteeRunnerRoleToPrefix` rejection path never runs in a handler test. The behavior is unverified either way.

**Suggested fix:** Constrain the committee-traces filter to committee runner roles at the boundary so bad input is a 400 (or silently empty), never a 500. Options:

1. Validate in the HTTP adapter (`toRunnerRoles` / `toCommitteeTracesQuery`) and reject anything other than RoleCommittee / RoleAggregatorCommittee, returning a 400 with the offending value.
2. Add a committee-only parser (e.g. `CommitteeRunnerRoleFromString`) used by `RunnerRoleSlice.Bind` for this endpoint, rejecting the five non-committee strings.
3. At minimum, treat the `CommitteeRunnerRoleToPrefix` error as an empty result (skip the role) rather than surfacing it, and add a test that passes `roles=AGGREGATOR` and asserts 400/empty — not 500.

Also update the PR's quirk-1 note: the behavior is a 500, not 'matches nothing'.

momosh-ssv: Good catch, and you're right it's a 500 not "matches nothing".

I traced it the same way (`Bind` → `GetCommitteeDuty` → `CommitteeRunnerRoleToPrefix` error → unfiltered multierror → 500). Fixed in bf5b32e7d.

The filter now binds through a committee-only `CommitteeRunnerRoleFromString` (COMMITTEE / AGGREGATOR_COMMITTEE), so non-committee roles are rejected at the boundary as a 400.
Covering both the query and JSON-body paths. Added a handler test asserting 400 (not 500) for `roles=AGGREGATOR`, and corrected the quirk-1 note in the PR description.

exporter/dutytracer/collector.go
```diff
@@ -883,21 +1046,36 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
...
+				if err != nil {
+					c.logger.Warn("failed to compute aggregator selection root",
+						zap.Error(err),
+						fields.Slot(verifyCtx.slot),
+						fields.CommitteeID(verifyCtx.committeeID))
+				} else {
+					aggSelectionRoot = root
+					aggSelectionRootReady = true
+				}
+			}
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 6be42347b.

Added the `verifyCtx.runnerRole == RoleAggregatorCommittee` guard so an `AggregatorCommitteePartialSig` misrouted to a `RoleCommittee` trace can't trigger the binary selection-root classification.

exporter/committee.go
```diff
@@ -46,23 +46,30 @@ func validateCommitteeRequest(request *CommitteeTracesQuery) error {
...
+		for _, role := range roles {
+			duty, err := e.traceStore.GetCommitteeDuty(slot, cmtID, role)
+			if err != nil {
+				errs = multierror.Append(errs, err)
+				continue
+			}
+			duties = append(duties, duty)
 		}
-		duties = append(duties, duty)
 	}
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 5b7e03ae5.

Restored the per-callsite `logger.Error`, guarded by `isNotFoundError` so the expected not-found probes (we now query every role per committee) stay quiet and only genuine storage failures are logged.

exporter/dutytracer/collector.go
momosh-ssv: Fixed in 6be42347b.

Extracted a shared `countPending(trace.pendingByRoot)` helper so the two arms can't drift.

api/handlers/exporter/committee_model.go
momosh-ssv: Fixed in a70ff30c0.

The response `role` is now serialized via `message.RunnerRoleToString`, so it shares a single source of truth with the `roles` filter vocabulary and round-trips cleanly.

exporter/dutytracer/collector.go
```diff
@@ -883,21 +1032,36 @@ func (c *Collector) collect(ctx context.Context, msg *queue.SSVMessage, verifySi
 		}
 
 		// process partial sig for committee
-		if verifyCtx.runnerRole == spectypes.RoleCommittee {
-			trace, late, err := c.getOrCreateCommitteeTrace(verifyCtx.slot, verifyCtx.committeeID)
+		if isCommitteeRunnerRole(verifyCtx.runnerRole) {
+			var aggSelectionRoot phase0.Root
+			aggSelectionRootReady := false
+			if verifyCtx.runnerRole == spectypes.RoleAggregatorCommittee && pSigMessages.Type == spectypes.AggregatorCommitteePartialSig {
+				root, err := c.getAggregatorSelectionRoot(ctx, verifyCtx.slot)
```
iurii-ssv, line 1039: **Coverage gap (test, non-blocking).** The pre-consensus AggregatorCommittee selection-proof path isn't exercised end-to-end. Every new AC test drives `PostConsensusPartialSig`; nothing sends a `spectypes.AggregatorCommitteePartialSig` through `Collect()`. As a result this `getAggregatorSelectionRoot` wiring, the binary `aggSelectionRoot` classification in `processPartialSigCommittee`, and its quorum branch in `checkAndPublishQuorum` are only covered indirectly — `getAggregatorSelectionRoot` has a unit test, but not its integration into classification/quorum. A pre-consensus analog of `TestCollector_AggregatorCommitteeDuty_PostConsensusQuorum` (send selection-proof partials, assert Attester vs SyncCommittee bucketing + per-root quorum) would close it.

exporter/dutytracer/store.go
```diff
@@ -128,112 +129,130 @@ func (c *Collector) getValidatorDutyFromDiskIndex(role spectypes.BeaconRole, slo
...
+		out = append(out, spectypes.RoleAggregatorCommittee)
+	}
+	return out
+}
+
 func (c *Collector) GetAllCommitteeDecideds(slot phase0.Slot, roles ...spectypes.BeaconRole) ([]ParticipantsRangeIndexEntry, error) {
 	var errs *multierror.Error
 
-	duties, err := c.GetCommitteeDuties(slot, roles...)
+	runnerRoles := committeeRunnerRolesForBeaconRoles(roles...)
```
iurii-ssv, line 239: **Minor (non-blocking, behavior shift vs base).** When `roles` contains only non-committee-backed beacon roles (e.g. `PROPOSER`), `committeeRunnerRolesForBeaconRoles` returns empty → `GetCommitteeDuties(slot, runnerRoles...)` runs with no role filter and returns every duty, and the `hasSignersForRoles` filter below passes them all (a non-committee role maps to no signer bucket, so it's skipped and the func returns `true`). Net effect: a non-committee role yields decideds for *all* committee duties — and, unlike base (which only ever queried `RoleCommittee`), now also `AggregatorCommittee` duties. Harmless for real callers (the input is nonsensical), just noting the semantics widened. If you want to be strict, returning empty when `len(roles) > 0 && len(runnerRoles) == 0` would reject it.

iurii-ssv (approved):
A couple of additional observations from a fresh review pass — both non-blocking.

exporter/dutytracer/store.go
momosh-ssv: Good catch.
Confirmed the widening: with only non-committee beacon roles, `committeeRunnerRolesForBeaconRoles` returns empty and `GetCommitteeDuties` runs unfiltered, returning all committee + AC duties. 

Fixed in d4283e238: both `GetAllCommitteeDecideds` and `GetCommitteeDecideds` now return empty when a role filter was requested but maps to no committee runner role, instead of falling through to a no-filter query. Added a `roles=PROPOSER` → empty subtest.

exporter/dutytracer/collector.go
momosh-ssv: Agreed, the pre-consensus selection-proof path was only covered indirectly. Added `TestCollector_AggregatorCommitteeDuty_PreConsensusQuorum` in fa28cd370: it drives `AggregatorCommitteePartialSig` partials through `Collect()`, asserts selection-root signatures bucket under Attester (aggregator) vs other roots under SyncCommittee (contribution), and verifies per-root quorum publishes once the threshold is met for each role.

iurii-ssv (approved):

ovidiu-ssv-labs (approved):

PR #2938, opened 2026-07-09
convergence(unit 5d): un-skip aggregator-committee + SyncCommitteeAggregatorProof spec vectors

momosh-ssv:
## convergence(unit 5d): un-skip aggregator-committee + SyncCommitteeAggregatorProof spec vectors

Closes the temporary spec-test coverage regression from front-load-0. The AC runner + producer landed in 5c; this unit is the spectest-harness wiring that makes the previously-**skipped** aggregator-committee and SyncCommitteeAggregatorProof vectors actually **execute and pass**. Ported from `boole-fork`'s `protocol/v2/ssv/spectest/` (DEFAULT-only; the `//go:build alan_spec` variants are deliberately out of scope). **No production code changed** — spectest package only.

### What changed
- Delete integration's skip-counter machinery (`skipped*Vectors`, `isAggregatorCommitteeDutyMap`); the AC + SyncCommitteeAggregatorProof vectors now run.
- Port boole's harness: `helpers.go`, `sync_committee_aggregator_proof_type{,_default}.go`, `error_code_map{,_runtime}.go`, `runner_fixture.go`, and the `ssv_mapping_test.go` / `msg_processing_type.go` / `committee_msg_processing_type.go` / `util.go` / `value_checker.go` modifications.
- Resolve the line-699 TODO: `fixCommitteeForRun` now populates both `c.Runners` and `c.AggregatorRunners` (reading `CommitteeRunners`/`AggregatorCommitteeRunners` from fixtures with fallbacks).

### Two deliberate divergences from the boole source
1. **`QBFTController.IdentifierFn` → `Identifier`** — boole uses an identifier-rotation feature integration doesn't have; substituted integration's field.
2. **Kept the STRICT post-root assertion** — boole's `RunAsPartOfMultiTest` silently downgrades a post-root mismatch to a non-failing `logger.Error`. We did **not** port that; `require.EqualValues` remains the operative check in every run path. This is intentionally stricter than the source.

### One skipped vector (with TODO)
`pre consensus post decided / sync committee aggregator selection proof` is skipped via an exact single-vector match. Root cause (verified by building both branches + instrumenting the runner): the fixture hardcodes a real-mainnet Altair slot (epoch 74240), but `TestNetwork` uses a toy fork schedule (Phase0=0…Fulu=6, byte-identical in boole), so the runner derives the wrong `DataVersion`. **boole only "passes" this vector because of the masked downgrade above** — our strict assertion exposed it. It's a test-fixture/toy-schedule artifact, **not a production bug** (production `ForkAtEpoch` uses real epochs). Follow-up: regenerate the fixture / reconcile the toy schedule.

### Review
- **code-reviewer-deep: PASS, no blockers.** Verified: hard-failure post-root/error assertions in all run paths (no vacuous-pass), skip is exactly one vector, harness logic-faithful to boole, wrong-`DataVersion` is test-only, error-code remap narrow, no deadcode.
- Validation: `make spec-test` + `-race` green; `git grep skipped|isAggregatorCommitteeDutyMap = 0`; build/vet/gofmt/deadcode/golangci-lint clean.

### Known / inherited-from-boole (non-blocking, tracked)
1. `assertRootsRelaxed` checks only the count of broadcasted beacon roots, not content (`CompareBroadcastedBeaconMsgs` commented out upstream) — restore when fork schedules are reconciled.
2. `sync_committee_aggregator_proof_type_default.go` calls `r.GetStateRoot()` unconditionally (panic-fails, never vacuous-pass) — a `require.NotNil` would make it legible.

Stacks on `integration/boole-convergence`; independent of PR #2937 (5c-exporter) — zero file overlap.


protocol/v2/ssv/spectest/error_code_map.go
```diff
@@ -0,0 +1,11 @@
+//go:build !alan_spec
```
iurii-ssv: **`spectest` no longer compiles under `-tags alan_spec`.**

This file (plus `error_code_map_runtime.go`, `sync_committee_aggregator_proof_type_default.go`, `ssv_mapping_default_test.go`) is `//go:build !alan_spec` and holds the *only* definitions of `adjustActualError`, `adjustExpectedErrorCode`, `adjustActualErrorForRunner`, `adjustActualErrorForRole`, and `RunSyncCommitteeAggProof`. Those symbols are called *unconditionally* from the no-build-tag files (`msg_processing_type.go`, `committee_msg_processing_type.go`, `value_checker.go`, `ssv_mapping_test.go`), and there is no `alan_spec` sibling defining them — so `go build -tags alan_spec ./protocol/v2/ssv/spectest` fails with undefined-symbol errors. (At merge-base every spectest file was tag-free, so this PR introduces the incompatibility.)

Currently harmless — nothing builds this package with `alan_spec` (spec-test uses `-tags blst_enabled`; the only `alan_spec`-tagged file repo-wide is `networkconfig/test-network_alan.go`) — but it's a compile break, not merely "won't run," the moment the alan spec-test build is wired up. Cheap mitigation: add an `error_code_map_alan.go` (+ a `RunSyncCommitteeAggProof` alan stub) with identity / no-op impls, or file a tracking issue for the alan-side port. (The identity shims here only earn their keep once that counterpart exists.)

protocol/v2/ssv/spectest/error_code_map_runtime.go
```diff
@@ -0,0 +1,40 @@
...
+	if role == ssvtypes.RoleSyncCommitteeContribution {
+		return adjustWrongBeaconRole(err)
+	}
+	return err
+}
+
+func adjustWrongBeaconRole(err error) error {
+	var specErr *spectypes.Error
+	if errors.As(err, &specErr) && specErr.Code == spectypes.WrongBeaconRoleTypeErrorCode {
+		return spectypes.WrapError(spectypes.QBFTValueInvalidErrorCode, err)
```
iurii-ssv, line 45: This remaps `WrongBeaconRoleTypeErrorCode` → `QBFTValueInvalidErrorCode` for the sync-committee-contribution path at the *test* layer. That papers over a real difference between the error code our runner returns and the one the spec expects, and it can also hide a genuine wrong-role regression in this path. Preferable to reconcile the production error code (or document precisely why the divergence is expected and safe) rather than leave a silent test-side remap; at minimum keep it behind an explicit TODO / tracked issue.

protocol/v2/ssv/spectest/helpers.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+	spectestingutils "github.com/ssvlabs/ssv-spec/types/testingutils"
+	"golang.org/x/exp/maps"
+
+	"github.com/ssvlabs/ssv/networkconfig"
+)
+
+func keySetFromShares(shares map[phase0.ValidatorIndex]*spectypes.Share) *spectestingutils.TestKeySet {
+	for _, share := range shares {
```
iurii-ssv, line 22: `keySetFromShares` returns the keyset of an arbitrary share (map iteration order is nondeterministic). Fine while every fixture shares a single operator keyset, but it would silently pick a random one if a fixture ever carried heterogeneous keysets. A one-line comment stating that assumption would prevent a subtle future surprise.

protocol/v2/ssv/spectest/msg_processing_type.go
```diff
@@ -153,13 +141,8 @@ func (test *MsgProcessingSpecTest) runPreTesting(ctx context.Context, logger *za
...
 			}
 			if test.DecidedSlashable && IsQBFTProposalMessage(msg) {
-				consensusMsg, err := specqbft.DecodeMessage(msg.SSVMessage.Data)
-				if err != nil {
-					panic(err)
-				}
-				slot := phase0.Slot(consensusMsg.Height)
 				for _, validatorShare := range test.Runner.GetShares() {
-					test.Runner.GetSigner().(*ekm.TestingKeyManagerAdapter).AddSlashableSlot(validatorShare.SharePubKey, slot)
+					test.Runner.GetSigner().(*ekm.TestingKeyManagerAdapter).AddSlashableSlot(validatorShare.SharePubKey, spectestingutils.TestingDutySlot)
```
iurii-ssv, line 148: This switched the slashable slot from the decoded proposal `Height` to the constant `spectestingutils.TestingDutySlot`. Equivalent for single-slot duties, but it's a semantic narrowing — worth a short comment (or confirm no multi-slot `DecidedSlashable` vector relies on the per-message height).

protocol/v2/ssv/spectest/ssv_mapping_test.go
```diff
@@ -186,34 +144,30 @@ func prepareTest(
...
-			// TODO(convergence unit 5): enable AggregatorCommittee vectors.
-			return nil
-		}
 
 		return &runnable{
 			name: typedTest.TestName(),
 			test: func(t *testing.T) {
+				subtests := test.(map[string]any)["Tests"].([]any)
+				for _, subtest := range subtests {
+					typedTest.Tests = append(typedTest.Tests, newRunnerDutySpecTestFromMap(t, subtest.(map[string]any)))
```
iurii-ssv, line 172: Minor: subtests here are decoded lazily inside the run closure, whereas `MultiMsgProcessingSpecTest` decodes them eagerly in `prepareTest`. Both are correct; the inconsistency is mildly confusing — a one-line comment on the intent (or aligning the two) would help the next reader.

iurii-ssv (commented):
I've skipped the items already raised by the Greptile bot: the unconditional `r.GetStateRoot()` nil-deref, the empty-`AggregatorRunners` nil-deref in `RunAsPartOfMultiTest`, and the `BeaconAggregatorsMap` parallel-slice length. The two substantive notes below are the `alan_spec` compile break and the test-side error-code remap; the rest are low-severity nits.

ovidiu-ssv-labs (approved):
Apart from Iurii's and Greptile's findings, looks good to me.

protocol/v2/ssv/spectest/sync_committee_aggregator_proof_type_default.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if test.ExpectedError != "" {
+		require.EqualError(t, lastErr, test.ExpectedError)
+	} else {
+		require.NoError(t, lastErr)
+	}
+
+	postRoot, err := r.GetStateRoot()
+	require.NoError(t, err)
+	require.EqualValues(t, test.PostDutyRunnerStateRoot, hex.EncodeToString(postRoot[:]))
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in d8498454c.

protocol/v2/ssv/spectest/msg_processing_type.go
```diff
@@ -208,6 +200,16 @@ func (test *MsgProcessingSpecTest) RunAsPartOfMultiTest(t *testing.T, logger *za
...
 		committee = c.CommitteeMember.Committee
+	case *runner.AggregatorCommitteeRunner:
+		var runnerInstance *runner.AggregatorCommitteeRunner
+		for _, runner := range c.AggregatorRunners {
+			runnerInstance = runner
+			break
+		}
+		actualRunner = runnerInstance
+		network = runnerInstance.GetNetwork().(*protocoltesting.TestingNetwork)
+		beaconNetwork = runnerInstance.GetBeaconNode().(*protocoltesting.BeaconNodeWrapped)
```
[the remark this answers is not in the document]
momosh-ssv: Added `require.NotNil(t, runnerInstance)` to both the `CommitteeRunner` and `AggregatorCommitteeRunner` branches so an empty map fails legibly rather than nil-panicking. 

Fixed in d8498454c.

protocol/v2/ssv/spectest/error_code_map.go
momosh-ssv: Thanks for the precise write-up.
Confirmed and fixed in 68922a6a9. 

Added `alan_spec` sibling files (`error_code_map_alan.go`, `error_code_map_runtime_alan.go`, `sync_committee_aggregator_proof_type_alan.go`) with identity/stub impls so `go build`/`vet -tags alan_spec ./protocol/v2/ssv/spectest` compiles again. 

The alan spec-test suite itself is still not wired up, so these shims just guard the compile until the alan-side port lands.

protocol/v2/ssv/spectest/helpers.go
momosh-ssv: Added a comment in d2b9b59be stating the single-shared-keyset assumption (all current fixtures share one operator keyset; heterogeneous keysets would pick nondeterministically via map order).

protocol/v2/ssv/spectest/msg_processing_type.go
momosh-ssv: Confirmed.
Added a comment in d8498454c noting the single-slot assumption so a future multi-slot vector relying on per-message height does not silently break.

protocol/v2/ssv/spectest/ssv_mapping_test.go
momosh-ssv: Added a one-line comment in d2b9b59be: subtests decode lazily inside the run closure (vs `MultiMsgProcessingSpecTest`'s eager decode in `prepareTest`) so a decode failure attributes to the individual subtest run rather than to setup.

protocol/v2/ssv/spectest/error_code_map_runtime.go
momosh-ssv: Agreed it should not be a silent remap. 
Added an explicit TODO in d2b9b59be documenting that this is a test-layer reconciliation of a divergence between the code our runner returns and the one the spec expects on the sync-committee-contribution path, and that it should be reconciled in production (or documented as safe) and removed. 

Keeping it behind the TODO rather than reconciling the prod code in this test-only PR.
Shout if you'd prefer the latter here.

protocol/v2/ssv/spectest/error_code_map.go
```diff
@@ -0,0 +1,11 @@
...
+
+package spectest
+
+func adjustActualError(err error) error {
```
iurii-ssv, line 8: Nit / follow-up on the alan_spec compile fix: `adjustActualError` and `adjustExpectedErrorCode` are identity in **both** build variants (here and in `error_code_map_alan.go`), so for this pair the build split is currently a no-op — the only build-conditional behavior lives in the `*ForRunner` / `*ForRole` runtime pair. The comment in `error_code_map_alan.go` ("The DEFAULT (!alan_spec) build remaps a handful of error codes") reads as if the default build remaps *these two*, which it does not. Consider collapsing this pair into a single tag-free file, and dropping the now no-op `adjustExpectedErrorCode(...)` wrap on `PartialSigContainerTest` in `ssv_mapping_test.go`. Non-blocking.

protocol/v2/ssv/spectest/msg_processing_type.go
```diff
@@ -187,12 +173,21 @@ func (test *MsgProcessingSpecTest) runPreTesting(ctx context.Context, logger *za
 	return v, c, lastErr
 }
 
+func (test *MsgProcessingSpecTest) BeaconAggregatorsMap() map[phase0.CommitteeIndex]bool {
+	aggregatorsMap := make(map[phase0.CommitteeIndex]bool)
+	for i, idx := range test.BeaconAggregators {
+		aggregatorsMap[idx] = test.BeaconAggregatorsValues[i]
```
iurii-ssv, line 179: Minor robustness: this indexes `BeaconAggregatorsValues[i]` while ranging over `BeaconAggregators`, so a fixture where the two arrays differ in length panics with an opaque index-out-of-range instead of a legible test failure. They are parsed as two independent slices in `msgProcessingSpecTestFromMap`, so nothing structurally guarantees equal length. A length check at parse time (where `t` is in scope) would fail clearly. Non-blocking.

protocol/v2/ssv/spectest/error_code_map.go
momosh-ssv: Good catch. Fixed in 650bd3682.

protocol/v2/ssv/spectest/msg_processing_type.go
momosh-ssv: Agreed, the two slices aren't structurally length-locked. 
Added an equal-length `require.Equal` check in `msgProcessingSpecTestFromMap` (where `t` is in scope) so a mismatched fixture fails legibly instead of panicking with index-out-of-range. 

Fixed in 650bd3682.

iurii-ssv (approved):

PR #2939, opened 2026-07-10
convergence: aggregator-committee message-validation test coverage

momosh-ssv:
## convergence: aggregator-committee message-validation test coverage

Closes the coverage hole unit 5b left: `validRoleAtSlot`'s three Boole-gated branches (`signed_ssv_message.go`) had **zero dedicated tests** — only the always-valid roles were hit indirectly. Test-only; the production branches already landed in 5b. Blocker (`commons.Subnet` rename) is moot — adapted to integration's `uint64`-subnet idiom.

### Added (4 subtests, `message/validation/validation_test.go`)
1. `RoleAggregatorCommittee` **pre-fork** → `ErrInvalidRole` (AC only exists post-Boole).
2. old `RoleAggregator` **post-fork** → `ErrInvalidRole` (superseded by AC).
3. old `RoleSyncCommitteeContribution` **post-fork** → `ErrInvalidRole`.
4. `RoleAggregatorCommittee` **late-slot** → `ErrLateSlotMessage`.

Subtest #4 signs its proposal with the fork-aware `RoundRobinProposer` leader (post-Boole leader selection is round-robin — a static op-1 signature would fail the leader-check before reaching the late-slot check), mirroring the existing post-fork committee-consensus pattern in the same file. No new helpers.

Test-only (+79/−0, single file). build/gofmt/golangci-lint green; the 4 subtests run+pass.

PR #2940, opened 2026-07-10
convergence: fork-aware exporter validator-path duty routing (NewExporter networkConfig)

momosh-ssv:
## convergence: fork-aware exporter validator-path duty routing

Completes the deferred follow-up from 5c-exporter (PR #2937). Threads `networkConfig` into `NewExporter` so the exporter's validator-trace API routes **AGGREGATOR** and **SYNC_COMMITTEE_CONTRIBUTION** duties correctly across the Boole fork. Today they always route to the validator path (permanently Alan); post-Boole they must route to the committee path.

### Changes
- `exporter/validator.go`: replace fork-unaware `isCommitteeDuty` with `isCommitteeDutyAtSlot`/`isCommitteeDutyInRange` methods that gate the two aggregator-family roles on `BooleForkAtSlot` (ATTESTER/SYNC_COMMITTEE stay fork-invariant → committee path always; other roles → validator path). `validateValidatorRequest` becomes a method. Ported boole's `CommitteeSignerBucketForBeaconRole` bucket-switch for membership gating (so aggregator-family filtering is correct post-Boole, not the old loose `else`).
- `exporter/exporter.go` + `api/handlers/exporter/exporter.go`: `networkConfig` added as the last `NewExporter` param (struct field + pass-through).
- `cli/operator/node.go`: both call sites pass the node's `networkConfig`.

### Behavior
Additive — **pre-Boole is byte-identical to today**; only post-Boole routing of AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION changes (validator → committee path). Read-only API surface; no consensus/duty-execution impact.

### Review & tests
- **code-reviewer-deep: PASS, no blockers** — routing logic byte-identical to boole, pre-fork preserved, membership switch faithful (no regression), no nil-`networkConfig` surface (all constructors pass a real value).
- New `TestExporterValidatorTraces_ForkGating` (8 subtests) + `_ValidationSymmetric` (4) assert **which store** is hit pre-vs-post-Boole (committee vs validator) via call-flags, correct signer-bucket + `CommitteeID`, and ATTESTER/SYNC_COMMITTEE fork-invariance. Per-test config is a clone (`Forks.Boole=0`) that doesn't mutate shared `TestNetwork`.
- Validation: build, vet (main + ssvsigner), deadcode, gofmt, golangci-lint, tests all green.

Stacks on `integration/boole-convergence`; no overlap with #2939 (that's `message/validation/`).

### DIFF
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -2565,6 +2565,112 @@ func TestExporterValidatorTraces_ForkGating_ValidationSymmetric(t *testing.T) {
 	}
 }
 
+// TestExporterValidatorTraces_ForkGating_CrossForkRange proves the behavior of a slot range
+// straddling the Boole fork boundary (from pre-Boole, to post-Boole): validation is evaluated
+// at the range's upper bound, so aggregator-family roles require pubkeys/indices, and with
+// indices provided each slot routes independently — validator path before the boundary,
+// committee path from it onward.
+func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
+	const booleEpoch = phase0.Epoch(5)
+	idx := phase0.ValidatorIndex(1)
+	var committeeID spectypes.CommitteeID
+	committeeID[0] = 7
+
+	ssvCopy := *networkconfig.TestNetwork.SSV
+	ssvCopy.Forks.Boole = booleEpoch
+	netCfg := *networkconfig.TestNetwork
+	netCfg.SSV = &ssvCopy
+
+	booleSlot := netCfg.FirstSlotAtEpoch(booleEpoch)
+	require.GreaterOrEqual(t, uint64(booleSlot), uint64(10), "boole fork slot too low for the range below")
+	from := uint64(booleSlot) - 10 // pre-Boole
+	to := uint64(booleSlot) + 10   // post-Boole
+
+	roles := []struct {
+		name              string
+		signerDataBuilder func(slot phase0.Slot) *traces.CommitteeDutyTrace
+	}{
+		{
+			name: "AGGREGATOR",
+			signerDataBuilder: func(slot phase0.Slot) *traces.CommitteeDutyTrace {
+				return &traces.CommitteeDutyTrace{
+					Slot:        slot,
+					CommitteeID: committeeID,
+					Attester:    []*traces.SignerData{{Signer: 99, ValidatorIdx: []phase0.ValidatorIndex{idx}}},
+				}
+			},
+		},
+		{
+			name: "SYNC_COMMITTEE_CONTRIBUTION",
+			signerDataBuilder: func(slot phase0.Slot) *traces.CommitteeDutyTrace {
+				return &traces.CommitteeDutyTrace{
+					Slot:          slot,
+					CommitteeID:   committeeID,
+					SyncCommittee: []*traces.SignerData{{Signer: 99, ValidatorIdx: []phase0.ValidatorIndex{idx}}},
+				}
+			},
+		},
+	}
+
+	for _, role := range roles {
+		t.Run(role.name+" without filters requires pubkeys/indices", func(t *testing.T) {
+			exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), &netCfg)
+
+			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
+				"from":  from,
+				"to":    to,
+				"roles": []string{role.name},
+			}))
+			req.Header.Set("Content-Type", "application/json")
+			rec := httptest.NewRecorder()
+
+			require.Error(t, exp.ValidatorTraces(rec, req))
+		})
+
+		t.Run(role.name+" with indices routes each slot by its own fork state", func(t *testing.T) {
+			store := newMockTraceStore()
+
+			validatorSlots := make(map[phase0.Slot]struct{})
+			committeeSlots := make(map[phase0.Slot]struct{})
+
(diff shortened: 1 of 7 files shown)


exporter/validator.go
```diff
@@ -247,6 +249,37 @@ func (e *Exporter) buildValidatorSchedule(req *ValidatorTracesQuery, indices []p
...
+		// Boole: AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION are committee duties.
+		return e.networkConfig.BooleForkAtSlot(slot)
+	}
+	return true
+}
+
+// isCommitteeDutyInRange reports whether the given beacon role resolves to a
+// committee-backed duty anywhere in the given (inclusive) slot range, evaluated
+// against the fork at the range's upper bound.
+func (e *Exporter) isCommitteeDutyInRange(role spectypes.BeaconRole, from, to phase0.Slot) bool {
```
iurii-ssv: nit (simplification): `isCommitteeDutyInRange` is equivalent to `isCommitteeDutyAtSlot(role, to)`.

The only behavioral difference is the `from > to` early-return, but that case is already rejected in `validateValidatorRequest` (`if request.From > request.To { return ... }`) before this is ever called, so the guard is dead and `from` is otherwise unused. The rest of the body is identical to `isCommitteeDutyAtSlot` with `slot = to`.

Consider dropping this method and calling `e.isCommitteeDutyAtSlot(role, phase0.Slot(request.To))` at the single call site — removes the duplicated role-resolution logic and one more place to keep in sync if the routing rule changes.

api/handlers/exporter/exporter_test.go
```diff
@@ -5,7 +5,9 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
+
```
iurii-ssv: nit: this stray blank line splits the stdlib import block into an accidental sub-group (`"net/http"` / blank / `"net/http/httptest"`). `gofmt` tolerates it, but it reads as an unintended group boundary — please remove it.

iurii-ssv (approved):
LGTM, just couple consistency/cleanup nits.

exporter/validator.go
momosh-ssv: Good catch, you're right. 
Fixed in e3b785f02.

Dropped `isCommitteeDutyInRange` and the call site now uses `isCommitteeDutyAtSlot(role, to)` with a comment explaining why `to` is the evaluation point.

api/handlers/exporter/exporter_test.go
momosh-ssv: Fixed in 4e3d8e252 — removed the stray blank line.

exporter/validator.go
```diff
@@ -247,6 +249,37 @@ func (e *Exporter) buildValidatorSchedule(req *ValidatorTracesQuery, indices []p
...
+	}
+	runnerRole, ok := ssvtypes.CommitteeRunnerRoleForBeaconRole(role)
+	if !ok {
+		return false
+	}
+	if runnerRole == spectypes.RoleAggregatorCommittee {
+		return e.networkConfig.BooleForkAtSlot(to)
+	}
+	return true
 }
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in e9cf0896a.
Added `TestExporterValidatorTraces_ForkGating_CrossForkRange` covering a range straddling the fork boundary: without filters the request is rejected (validation at the range's upper bound), and with indices each slot routes by its own fork state.

exporter/validator.go
```diff
@@ -121,12 +121,13 @@ func (e *Exporter) getValidatorCommitteeDutiesForRoleAndSlot(role spectypes.Beac
...
 
-	// This helper is only reached for committee-backed beacon roles (see isCommitteeDuty),
-	// which today always resolve to the RoleCommittee runner role.
+	// This helper is only reached for committee-backed beacon roles (see isCommitteeDutyAtSlot),
+	// which resolve to either the RoleCommittee or RoleAggregatorCommittee runner role.
 	runnerRole, ok := ssvtypes.CommitteeRunnerRoleForBeaconRole(role)
 	if !ok {
 		return nil, fmt.Errorf("unexpected committee-backed beacon role: %s", role.String())
 	}
+	bucket, _ := ssvtypes.CommitteeSignerBucketForBeaconRole(role)
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 56c8d95c7.

The bucket lookup now errors on `!ok`, mirroring the `CommitteeRunnerRoleForBeaconRole` check just above, so a divergence between the two role maps fails visibly instead of bypassing membership gating.

exporter/exporter.go
```diff
@@ -24,14 +25,16 @@ type Exporter struct {
...
-func NewExporter(logger *zap.Logger, participantStores *ibftstorage.ParticipantStores, traceStore dutyTraceStore, validators registrystorage.ValidatorStore) *Exporter {
+func NewExporter(logger *zap.Logger, participantStores *ibftstorage.ParticipantStores, traceStore dutyTraceStore, validators registrystorage.ValidatorStore, networkConfig *networkconfig.Network) *Exporter {
 	return &Exporter{
 		participantStores: participantStores,
 		traceStore:        traceStore,
 		validators:        validators,
 		logger:            logger,
+		networkConfig:     networkConfig,
 	}
 }
```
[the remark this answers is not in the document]
momosh-ssv: I fear a nil check just for `networkConfig` would be inconsistent here.
`NewExporter` doesn't guard any of its other dependencies (`logger`, `traceStore`, `validators` would panic just as fast on nil), and that matches constructor style across the codebase. All call sites pass a real config, and a nil would surface on the first test run, so I'd rather keep the constructor uniform.

PR #2941, opened 2026-07-10
Boole → stage convergence — final consolidation

momosh-ssv:
## Boole → stage convergence — final consolidation

Merges the entire `boole-fork` protocol-fork layer into `stage`, assembled unit-by-unit on `integration/boole-convergence`. This is the **consolidation merge** — every unit below was already implemented, deep-reviewed, and merged into `integration` as its own PR; this PR lands them in `stage` as one coherent set.

### What's included
- **Front-load-0 + unit 1** — ssv-spec v1.2.3 pin + compat shim; networkconfig fork API (`SSVForks`, `BooleForkAtSlot`, `DomainTypeAtSlot`).
- **Unit 2** (#2927) — fork-gated networking + message validation (dual Alan/Boole topics, slot-gated topic/domain acceptance).
- **Unit 3** (#2929) — role-aware exporter duty traces + migration_9 + SSZ model.
- **Unit 4** (#2930) — fork-aware `applyDynamicMaxPeers`.
- **Unit 5a** (#2931) — fork-aware round-robin proposer (single `qbft.Proposer` entry point).
- **Unit 5b** (#2932) — fork-aware runners, scheduler & message validation.
- **Unit 5c** (#2933) — `AggregatorCommitteeRunner` + consensus wiring.
- **Unit 5c-exporter** (#2937) — aggregator-committee duty-trace query layer.
- **Unit 5d** (#2938) — un-skip aggregator-committee + SyncCommitteeAggregatorProof spec vectors (kept STRICT post-root assertions — stricter than boole).
- **Follow-ups** (flowing in): AC message-validation coverage (#2939) + fork-aware exporter validator-path routing (#2940).

### Status
- ✅ `stage` synced into this branch — its tip is the merge-base, the `operator/node.go` conflict is resolved in-branch, and the PR merges cleanly.
- ✅ #2939 + #2940 merged — their changes are included in this diff.
- **Behavioral note:** `local-testnet` now activates Boole from genesis — a flip for local dev clusters (see #2978, item 6).
- Merge method: **merge commit**.

### Follow-up (tracked, not in this PR)
- Regenerate the one skipped Altair spec vector (`sync committee aggregator selection proof`) against a reconciled fork schedule (#2964).
- Non-blocking hardening + polish from the full-pass review: #2978.


iurii-ssv:
@greptile can you review ?

operator/validator/controller.go
```diff
@@ -1050,8 +1081,8 @@ func SetupCommitteeRunners(
 			BeaconSigner: options.Signer,
 			Domain:       options.NetworkConfig.DomainType,
```
iurii-ssv: **Consensus MsgID domain is not fork-aware — post-Boole consensus messages will be rejected.**

The QBFT `Config.Domain` here and the controller `identifier` a few lines below (`identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, ...)`) both use the static `DomainType` (Alan). That identifier becomes `state.ID` in `StartNewInstance`, and via spec `Sign()` it is the wire `SSVMessage.MsgID` of **every** consensus message (proposal/prepare/commit/round-change/decided).

Post-fork the receiver's `validateDomainAtSlot` expects `DomainTypeAtSlot(slot) == NextDomainType` (Boole), so every post-fork consensus message this node emits carries the Alan domain and is dropped with `ErrWrongDomain` — consensus can't reach quorum. This is exactly the scenario asserted in `message/validation/validation_test.go` ("committee consensus rejected with alan domain post-fork" -> `ErrWrongDomain`).

The partial-signature path is already fork-aware (`runner.go:354`, `proposer.go:303`, `committee.go:403` use `DomainTypeAtSlot(slot)`), so within a single post-fork duty the partial sigs would be Boole while consensus is Alan.

Not caught by CI: all real networks pin `Boole=MaxUint64`, spec-tests use consistent spec domains on both ends, and the MV unit tests hand-craft the receiver input with the Boole domain — none drive the production sender into the production validator post-fork.

Suggested fix: derive the identifier/`Domain` from `DomainTypeAtSlot(duty.DutySlot())` (the slot is available in this per-duty factory) and add a `Boole:0` sender->receiver integration test.

operator/validator/controller.go
```diff
@@ -1115,8 +1167,8 @@ func SetupRunners(
...
```
iurii-ssv: Same fork-aware-domain issue as the committee/AC `buildController` above (see the detailed comment there). Note these validator runners are long-lived across the fork, so the fix here likely needs the controller identifier/`Domain` refreshed at duty start (when the slot is known) rather than fixed at construction with the static `DomainType`.

network/p2p/p2p_pubsub.go
```diff
@@ -38,6 +39,21 @@ func (n *p2pNetwork) UseMessageRouter(router network.MessageRouter) {
 
 // Broadcast publishes the message to all peers in subnet
 func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedSSVMessage) error {
+	decodedMsg, err := queue.DecodeSignedSSVMessage(msg)
```
iurii-ssv: Minor (perf): `Broadcast` fully SSZ-decodes every message here just to recover the slot, then `BroadcastAtSlot` re-encodes it. Correct, but it's an extra decode on the partial-sig broadcast path. The remaining `Broadcast` callers (`runner.go:379`, aggregator runners) already hold the duty slot and could call `BroadcastAtSlot` directly.

network/commons/subnets.go
```diff
@@ -51,23 +60,80 @@ func GetTopicBaseName(topicName string) string {
...
 	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), bigIntSubnetsCount)
 	return subnet.Uint64()
 }
 
+// BooleCommitteeSubnet returns the subnet for the given committee, calculated as (lowestHash % SubnetsCount).
+// This determines which subnet a committee lands on post-Boole-fork. An empty committee has no
+// subnet; it returns UnknownSubnetId defensively rather than panicking (real shares are non-empty).
+func BooleCommitteeSubnet(committee []spectypes.OperatorID) uint64 {
+	if len(committee) == 0 {
+		return UnknownSubnetId
```
iurii-ssv, line 94: Minor/defensive: `BooleCommitteeSubnet` returns `UnknownSubnetId` for an empty committee, but `BooleTopic` (unlike `SubnetTopicID`) does not special-case `UnknownSubnetId` — it would emit `/ssv/<net>/boole/<UnknownSubnetId>`. Only reachable for an empty committee (real shares are non-empty), but the Alan/Boole topic naming diverges for that sentinel.

networkconfig/network.go
```diff
@@ -20,9 +24,76 @@ func (n Network) String() string {
...
+func (n Network) inBooleSubsequentWindowWithSlots(slot phase0.Slot, windowSlots phase0.Slot) bool {
+	// If Boole is at genesis there is no transition/unsubscription window to apply; without
+	// this guard we would treat slot 0 as being inside the window.
+	if n.SSV.Forks.Boole == 0 {
+		return false
+	}
+
+	// Avoid FirstSlotAtEpoch overflow when Boole is beyond the representable epoch range;
+	// without this guard the multiplication would wrap and could treat small slots as in-window.
+	maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
```
iurii-ssv, line 123: Latent: `math.MaxUint64 / n.SlotsPerEpoch` panics if `SlotsPerEpoch == 0`. Unreachable for the built-in beacon configs (always 32), but a malformed custom config turns this overflow guard into a divide-by-zero panic. A `SlotsPerEpoch == 0` guard would harden it.

message/validation/consensus_validation.go
```diff
@@ -17,7 +17,7 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
 	"github.com/ssvlabs/ssv/protocol/v2/message"
-	"github.com/ssvlabs/ssv/protocol/v2/qbft"
+	qbft "github.com/ssvlabs/ssv/protocol/v2/qbft"
```
iurii-ssv: Nit: redundant import alias — the package is already named `qbft`, so this is equivalent to a plain import (staticcheck QF1008, which this repo has been cleaning up).

momosh-ssv:
Heads-up before this lands: there's a **P0 Boole blocker** in the current diff — the subscription filter rejects **every** Boole committee topic (`/ssv/<net>/boole/<subnet>`), so post-Boole every node either crash-loops at startup or can't broadcast QBFT messages. Filed as #2943, reproduced live on stage-hoodi at HEAD `19e06375`.

Root cause is `CanSubscribe` in `network/topics/sub_filter.go` — its "of interest" gate only recognizes the Alan `ssv.v2.` prefix, so it drops Boole topics before the whitelist is consulted.

Fix is up in **#2944** (targets `integration/boole-convergence`, so it'll fold into this PR's diff once merged). It's verified end-to-end on stage: a post-Boole committee on the fix submits attestations with zero filter errors, while the unfixed control keeps failing. Suggest gating this PR's merge on #2944.

protocol/v2/ssv/runner/aggregator_committee.go
```diff
@@ -0,0 +1,1817 @@
...
+		for _, selection := range aggregatorSelections {
+			// Check if attestation for committee index was already included
+			for _, idx := range consensusData.AggregatorsCommitteeIndexes {
+				if idx == uint64(selection.duty.CommitteeIndex) {
+					// If so, just add to aggregators and return
+					consensusData.Aggregators = append(consensusData.Aggregators, spectypes.AssignedAggregator{
+						ValidatorIndex: selection.duty.ValidatorIndex,
+						SelectionProof: selection.selectionProof,
+						CommitteeIndex: uint64(selection.duty.CommitteeIndex),
+					})
```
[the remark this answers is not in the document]
y0sher: sounds like a bug that is hard to test, maybe only UT can cover it. but definitely I would look into it and see if this valid @momosh-ssv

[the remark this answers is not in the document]
momosh-ssv: Looked into it, and it seems this one is a false positive. 

The "already included" branch is logic-identical to the spec's own runner (`ssv/aggregator_committee.go`, same append-only-to-`Aggregators` handling), and `AggregatorCommitteeConsensusData.Validate()` actually requires this shape: it checks set-equality between `Aggregators[].CommitteeIndex` and `AggregatorsCommitteeIndexes`, and the spec's own spectest (`InvalidAggregatorValidationDuplicateCommitteeIndex`) asserts that appending matching entries to all three slices — what's suggested here — must fail validation with a duplicate-index error. 

Post-consensus pairing is also safe: `GetAggregateAndProofs` builds one proof per aggregator in order, so two aggregators for one committee each submit their own signed copy of the shared aggregate, as the beacon chain expects. 

On the UT point, agreed it's worth covering: the shape rules are only tested in the ssv-spec repo, nothing on our side drives this branch. Filed #2950 to add a focused UT (two aggregators, one committee) as a follow-up.

operator/validator/controller.go
momosh-ssv: Fixed in 94d9d02fb (PR #2948, merged into this branch).

The controller now resolves its identifier per-height via `newIdentifierFn`/`IdentifierFn`, so post-fork consensus messages carry the Boole domain from `DomainTypeAtSlot`. Covered by `identifier_fn_test.go` and `controller_fork_test.go` (send-path assertions for `StartNewInstance` and the decided fast-forward). `Config.Domain` is indeed still static, but seems `GetSignatureDomainType` has no callers left at all.
Likely a candidate for removal in a follow-up cleanup.

operator/validator/controller.go
momosh-ssv: Fixed in 94d9d02fb together with the committee-side one.

Since `IdentifierFn` resolves at the instance's height rather than at construction, the long-lived validator runners pick up the fork-correct domain per duty without needing a refresh at duty start. I traced the remaining static-domain identifiers (`validator.go`, `startup.go`): they only feed role-keyed queue lookups and timeout events resolved by height, so none of them reach the wire.

network/p2p/p2p_pubsub.go
momosh-ssv: Fixed in b850e02ce.

Switched the four slot-holding callers (`runner.go`, `aggregator.go`, both `aggregator_committee.go` sites) to `BroadcastAtSlot`; they already had the duty slot in hand for the fork-aware MsgID, so the extra SSZ decode is gone from the partial-sig path. `Broadcast` stays as the slot-recovering fallback for the interface.

network/commons/subnets.go
momosh-ssv: Fixed in 644829f3b.

`BooleTopic` now mirrors `SubnetTopicID` and emits the `unknown` sentinel. Worth noting it was slightly more than a naming divergence: `ParseTopicSubnet` would happily parse `/boole/18446744073709551615` back as a "valid" subnet, while the sentinel fails parsing cleanly on both forks.

networkconfig/network.go
momosh-ssv: Fixed in 2543af1f2.

`SlotsPerEpoch == 0` now returns early as "not in window" before the division.

message/validation/consensus_validation.go
momosh-ssv: Fixed in 26cc4404a.

message/validation/partial_validation.go
```diff
@@ -241,7 +276,7 @@ func validatePartialSignatureMessageLimit(
 	signerState *SignerStateForSlotRound,
 ) error {
 	switch m.Type {
-	case spectypes.RandaoPartialSig, spectypes.SelectionProofPartialSig, spectypes.ContributionProofs,
+	case spectypes.RandaoPartialSig, ssvtypes.SelectionProofPartialSig, ssvtypes.ContributionProofs,
 		spectypes.ValidatorRegistrationPartialSig, spectypes.VoluntaryExitPartialSig:
```
iurii-ssv: **`AggregatorCommitteePartialSig` is missing from the pre-consensus duplicate-limit switch — AC duplicates lose the reject-to-punish path and get a misleading error.**

Its counterpart [`RecordPartialSignatureMessage`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/message/validation/seen_msg_types.go#L83) counts `AggregatorCommitteePartialSig` as pre-consensus, but this switch omits it. So the *second* AC selection-proof message from a signer for a slot (first one creates the signer state, so the limit check only starts firing from the second) falls through to the `default` branch — `fmt.Errorf("unexpected partial signature message type: %d")` — an untyped error that [`handleValidationError` maps to Ignore](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/message/validation/errors.go#L159-L163).

Consequences:
- same-peer logical duplicates are only ignored instead of rejected-to-punish, unlike every legacy pre-consensus type (Randao/SelectionProof/ContributionProofs/registration/exit);
- the error text claims the type is unexpected even though [`validPartialSigMsgType`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/message/validation/partial_validation.go#L347-L357) and [`partialSignatureTypeMatchesRole`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/message/validation/partial_validation.go#L362-L381) both accept it.

```suggestion
	case spectypes.RandaoPartialSig, ssvtypes.SelectionProofPartialSig, ssvtypes.ContributionProofs,
		spectypes.ValidatorRegistrationPartialSig, spectypes.VoluntaryExitPartialSig,
		spectypes.AggregatorCommitteePartialSig:
```

I couldn't find a test driving this path (second AC pre-consensus message with existing signer state) — worth adding one alongside, asserting `ErrTooManyPartialSigMessage` with `reject: true` for the same-peer case.

networkconfig/test-network.go
```diff
@@ -66,12 +69,35 @@ var TestNetwork = &Network{
...
 }
+
+func init() {
+	// Tests sometimes need to validate behavior both pre- and post-fork. This env var allows CI to
+	// flip the effective fork activation for the global TestNetwork without changing code.
+	//
+	// Supported values:
+	// - "pre":  disables the Boole fork (sets it to max epoch)
+	// - "post": enables the Boole fork from genesis (epoch 0)
+	switch strings.ToLower(strings.TrimSpace(os.Getenv("SSV_TEST_BOOLE_FORK"))) {
```
iurii-ssv, line 95: **The post-fork matrix never actually runs in CI.** Nothing sets `SSV_TEST_BOOLE_FORK` in the [Makefile](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/Makefile#L50-L70) or the workflows (and nothing builds with `-tags alan_spec`), so `make unit-test` / `make spec-test` only ever exercise the default pre-Boole `TestNetwork`; post-fork behavior is covered only by tests that construct explicit post-fork configs.

Given the blast radius of this fork, a second CI pass with `SSV_TEST_BOOLE_FORK=post` for the unit + spec suites seems worth the CI minutes — the switch is already wired here, it just has no caller. (Related: the [`alan_spec` stub of `RunSyncCommitteeAggProof`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/protocol/v2/ssv/spectest/sync_committee_aggregator_proof_type_alan.go#L11-L18) is dead in CI today — fine as scaffolding per its comment, just noting the two matrix mechanisms are both currently unused.)

cli/operator/node.go
```diff
@@ -653,6 +649,23 @@ func (n *node) applyDynamicMaxPeers() {
 	}
 }
 
+// operatorActiveSubnets returns the number of distinct committee subnets the operator's validators
+// participate in. Boole and Alan subnets are tallied in separate bitmaps: they are independent gossip
+// topics (/ssv/<net>/boole/<n> vs ssv.v2.<n>) even when their subnet numbers coincide, so a shared
+// bitmap would under-count on collision. Before/through the Boole transition both Alan and Boole
+// subnets are counted; post-fork only Boole. BooleCommitteeSubnet may return UnknownSubnetId for an
+// empty committee, but Set drops out-of-range indices, so it never inflates the count.
+func operatorActiveSubnets(validators []*ssvtypes.SSVShare, booleFork bool) int {
```
iurii-ssv: **This changes `MaxPeers` sizing on networks where the fork is *unscheduled*.** `booleFork` is false whenever `Forks.Boole == MaxUint64` — i.e. mainnet/hoodi/holesky/sepolia today — so the tally includes Boole subnets from day one, and `applyDynamicMaxPeers` can roughly double the per-subnet peer budget on production networks before any fork epoch is set. (Same pattern as [`commons.Topics()`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/network/commons/subnets.go#L128-L139) whitelisting all 128 Boole topics unconditionally.)

If the intent is for this PR to be a no-op on production until activation, consider gating the Boole tally on the fork actually being scheduled (e.g. `Forks.Boole != math.MaxUint64`) or on the transition window. If the early overprovisioning is deliberate, worth stating that in the comment so it doesn't read as an accident.

network/commons/subnets.go
```diff
@@ -38,36 +47,93 @@ func SubnetTopicID(subnet uint64) string {
...
 	bigInt.Mod(bigInt, bigIntSubnetsCount)
 }
 
-// Topics returns the available topics for this fork.
-func Topics() []string {
-	topics := make([]string, SubnetsCount)
-	for i := uint64(0); i < SubnetsCount; i++ {
-		topics[i] = GetTopicFullName(SubnetTopicID(i))
+// Topics returns the available topics for the given network's fork state: Alan-only pre-fork,
+// Alan+Boole during the transition window, Boole-only after.
```
iurii-ssv: Doc/code mismatch: the comment says "Alan-only pre-fork, Alan+Boole during the transition window", but the loop appends Boole topics unconditionally — pre-fork (long before any window) this returns Alan+Boole, and post-fork Boole-only. The behavior itself makes sense for a subscription-filter whitelist frozen at pubsub construction (a node started pre-fork must accept Boole subscriptions once the window opens, without a pubsub rebuild) — the comment should just say what the code does:

```suggestion
// Topics returns the topic whitelist for the given network's fork state: Alan+Boole before the
// fork (so a node started pre-fork can accept Boole subscriptions once the transition window
// opens, without a pubsub rebuild), Boole-only after.
```

protocol/v2/ssv/validator/duty_executor.go
```diff
@@ -28,7 +32,7 @@ func (v *Validator) ExecuteDuty(ctx context.Context, logger *zap.Logger, duty *s
 		return fmt.Errorf("decode duty execute msg: %w", err)
 	}
 
-	if pushed := v.Queues[duty.RunnerRole()].TryPush(dec); !pushed {
+	if pushed := v.Queues[role].TryPush(dec); !pushed {
```
iurii-ssv: Hardening: unguarded nil-queue push. `v.Queues` only has entries for roles with runners ([`NewValidator` iterates `options.DutyRunners`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/protocol/v2/ssv/validator/validator.go#L81-L92)), and validators built post-fork skip the legacy Aggregator/SyncCommitteeContribution runners — so a map miss here yields a nil `queue.Queue` interface and `TryPush` panics.

Today this is unreachable: the scheduler's dispatch gate and this role resolution use the same slot (legacy aggregator execution is gated `!BooleForkAtSlot(currentSlot)` in `attester.go`), so a legacy-role duty can't reach a post-fork-built validator short of clock regression across the boundary. But the sibling paths already guard exactly this: [`queue_validator.go`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/protocol/v2/ssv/validator/queue_validator.go#L60) uses the `, ok` form and [`timer.go`](https://github.com/ssvlabs/ssv/blob/b850e02ce674cea14b3647a09e47406141482cde/protocol/v2/ssv/validator/timer.go#L31-L38) nil-checks with a comment. An `if q, ok := v.Queues[role]; !ok { return fmt.Errorf(...) }` here would close the panic class for a one-line cost.

go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/sourcegraph/conc v0.3.0
 	github.com/spf13/cobra v1.8.1
 	github.com/ssvlabs/eth2-key-manager v1.5.6
-	github.com/ssvlabs/ssv-spec v1.2.2
+	github.com/ssvlabs/ssv-spec v1.2.3-0.20260305184636-289c93aa4c12
```
iurii-ssv, line 43: The spec pin is a pseudo-version (untagged commit). Fine for `stage`, but before this progresses further it's worth pinning a tagged `ssv-spec` release (v1.2.3 proper) — that also keeps this and `ssvsigner/go.mod` (same pseudo-version today) from silently drifting apart on future bumps.

iurii-ssv (commented):
Next batch

ovidiu-ssv-labs (commented):
Two items from the boole-fork → stage convergence that look like they didn't carry over from `boole-fork`. Both are gaps relative to `boole-fork` (stage never had them either), so they don't show up as changed lines in this diff — noting them here.

**1. `subscriptionRequestLimit` is too low for the transition window** — see the inline comment on `network/topics/pubsub.go`. It's `128 + 1`, but needs `commons.SubnetsCount*2 + 1`: during the Boole transition both the Alan and Boole topic sets are live, so heavily-subscribed peers (exporters, or nodes covering >64 subnets) exceed 129 and have their entire subscription RPC dropped by the sub-filter, degrading mesh formation for that window.

**2. The alan-spec spec-test run was dropped.** `boole-fork` ran the qbft/ssv spectest suites twice: the default build against the Boole vectors (ssv-spec v1.2.3), and `make spec-test-alan` (`-tags alan_spec`, `go.spec.alan.mod`) against the alan/mainnet vectors (ssv-spec v1.2.2), via the `spec-test-alan.yml` workflow. This branch drops `go.spec.alan.mod`/`.sum`, the `spec-test-alan` / `spec-test-alan-deps` Makefile targets, and the workflow — so the node's **pre-fork (alan) spec compliance is no longer verified in CI**, even though it runs pre-Boole on mainnet today. (Against the Boole vectors the assertions here are actually *stricter* than boole-fork's — hard `PostDutyRunnerStateRoot` equality rather than a logged mismatch — so this is specifically about losing the alan vector run.) It isn't in the description's intentional-changes list, so worth either restoring the second run or explicitly calling out the drop.

network/topics/pubsub.go
```diff
@@ -127,8 +127,8 @@ func NewPubSub(
 	}
 
 	// Set up a SubFilter with a whitelist of known topics.
-	sf := newSubFilter(logger, subscriptionRequestLimit)
-	for _, topic := range commons.Topics() {
+	sf := newSubFilter()
```
ovidiu-ssv-labs, line 132: `subscriptionRequestLimit` (line 28) is still `128 + 1`, carried over from the pre-fork code. During the Boole transition window every subnet has **two** live topics (Alan + Boole), so a node can legitimately be subscribed to up to `SubnetsCount*2` (256) topics — an exporter (`SubscribeAll`), or any node covering more than ~64 subnets, will exceed 129.

When such a peer advertises its subscriptions, `FilterIncomingSubscriptions` (`sub_filter.go:47`) returns `ErrTooManySubscriptions` and go-libp2p-pubsub drops the **entire** RPC: the peer's subscriptions never register, no mesh forms on the shared topics, and gossip to/from it is silently dropped for exactly the window where both topic sets need to be live.

`boole-fork` sized this as `commons.SubnetsCount*2 + 1` for this reason. Suggest restoring that (and updating the "128 subnets + 1" comment on line 27):

```go
subscriptionRequestLimit = commons.SubnetsCount*2 + 1
```

I think the likelihood of affecting an operator is low at the moment, but it will surely affect the exporters during the fork transition period. To what I understand, the effects may be limited to peers that connect after exporter is fully subscribed: exporter’s initial hello packet may contain all current subs (~256), and nodes with your `subscriptionRequestLimit=129` will ignore that hello subscription list.

message/validation/partial_validation.go
momosh-ssv: Fixed in 2dcb87a58.

Added the missing case, and took your suggestion on the test: the new UT drives a second AC pre-consensus message post-Boole and asserts ErrTooManyPartialSigMessage with reject for the same peer and ignore for another peer.

networkconfig/test-network.go
momosh-ssv: Partially done in 9920d1d21.

The spec suites now run a second time with SSV_TEST_BOOLE_FORK=post (verified green). The unit suite turned out to be not quite ready for the flip: Test_ValidateSSVMessage and TestMsgValidator hand-craft Alan topics/domains against the global TestNetwork, so I pinned both explicitly pre-fork (502c7a755, 41b2dab01 — post-fork stays covered by their own postBooleCfg subtests). 

The remaining unknown is network/p2p: I couldn't verify it locally in either mode (its socket-heavy tests hang on my machine even without the flip), and by inspection TestP2pNetwork_SubscribeBroadcast builds MsgIDs from the static netCfg.DomainType, which post-flip receivers should reject. 

Seems safer to give the p2p fixtures their own fork-agnostic pass in a follow-up and enable the unit matrix then, rather than risk a red job here.

cli/operator/node.go
momosh-ssv: Fixed in 8f02a34ce.

Went with the gate rather than the comment: the Boole tally now only counts when the fork is scheduled, via a new `networkconfig.BooleForkScheduled() (Forks.Boole != MaxUint64)`. 

Since scheduling ships in a release, a restart re-tallies before Boole subnets ever matter, so production `MaxPeers` stays unchanged until then. `TestOperatorActiveSubnets` got an unscheduled-fork case.

network/commons/subnets.go
momosh-ssv: Fixed in 3dc1271c6.

Took your wording as-is; the behavior was indeed intentional for a whitelist frozen at pubsub construction.

protocol/v2/ssv/validator/duty_executor.go
momosh-ssv: Fixed in bbfda4670.

Guarded with the ok-form like `queue_validator.go`, returning an error instead of panicking on a role with no runner.

go.mod
momosh-ssv: Agreed on the principle, but seems we can't do it yet, so a pseudo-version is the only way to pin this commit. Both go.mod and ssvsigner/go.mod carry the same pseudo-version today, so there's no drift to worry about right now. Happy to bump both to v1.2.3 proper the moment it's tagged.

Worth a follow-up reminder issue maybe?

network/topics/pubsub.go
momosh-ssv: Fixed in d03370ccb.

Restored the boole-fork sizing (commons.SubnetsCount*2 + 1) and updated the comment. Your read on the blast radius matches mine: an exporter's initial subscription RPC (~256 topics) would have been dropped wholesale by peers running the 129 limit, exactly during the window both topic sets are live.

protocol/v2/ssv/runner/aggregator_committee.go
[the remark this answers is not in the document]
momosh-ssv: Follow-up closed: the UT from #2950 landed via #2951 and is in this branch now — it covers the exact two-aggregators-one-committee shape discussed here, asserting the spec-required slice layout and per-aggregator proofs. I'd resolve this one.

momosh-ssv:
@ovidiu-ssv-labs both items from your review are addressed now. The subscription limit is fixed in d03370ccb (inline thread). The alan spec-test run is restored in 2d9d26e94: go.spec.alan.mod/.sum, the spec-test-alan Makefile targets, the workflow, and the build-tag scaffolding are back and adapted to the converged tree.

The qbft side needed the v1.2.2→v1.2.3 error-code remapping and a pre-fork proposer shim, the ssv side its alan error-code map plus two assert paths that had lost the remap seam. Verified locally: the alan run is green, and the default (Boole) suites stay green.

One residual gap worth flagging: `RunSyncCommitteeAggProof` keeps its `alan_spec` stub (skips). The Boole-fork's implementation relied on `BaseValidator`, which no longer exists on this branch; porting it to the committee-based flow felt like its own change, so I'd take that as a follow-up.

network/topics/pubsub.go
```diff
@@ -127,8 +129,8 @@ func NewPubSub(
...
+	for _, topic := range commons.Topics(cfg.NetworkConfig) {
```
iurii-ssv, line 133: **Non-blocking:** this whitelist is computed once at pubsub construction, so a node that starts pre-fork and runs across Boole activation keeps accepting incoming peer subscriptions to Alan topics indefinitely — our own Alan subscriptions get torn down by the [`UpdateSubnets`](https://github.com/ssvlabs/ssv/blob/integration/boole-convergence/network/p2p/p2p.go#L567) reconciliation, but the subscription filter stays two-fork wide until a restart.

Impact is limited to what peers may advertise to us, so fine to leave as-is; if we ever want the filter to tighten once the transition window closes, deregistering the Alan topics from the whitelist (e.g. from the `UpdateSubnets` loop) would do it.

network/p2p/p2p_pubsub.go
```diff
@@ -51,33 +67,34 @@ func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedS
...
 
-		// topicPeers surfaces a sparse/empty mesh — a prime suspect when a one-shot message fails to reach peers.
-		topicPeers, _ := n.topicsCtrl.Peers(topic)
-		n.logger.Debug("📤 broadcast message to topic",
-			fields.MessageID(msg.SSVMessage.MsgID),
-			zap.String("gossip_msg_id", hex.EncodeToString([]byte(topics.MsgID(encodedMsg)))),
-			fields.Topic(topic),
-			zap.Int("topic_peers", len(topicPeers)),
-		)
+	if err := n.topicsCtrl.Broadcast(topic, encodedMsg, n.cfg.RequestTimeout); err != nil {
```
iurii-ssv, line 120: **Nit (observability):** the rewrite drops the old per-broadcast debug log ("📤 broadcast message to topic") with the gossip msg ID and `topic_peers` count. That peer count was the prime signal for diagnosing a sparse/empty mesh when a one-shot message fails to reach peers. Now that the topic is resolved in a single place, consider re-adding it here.

network/p2p/p2p_pubsub.go
```diff
@@ -145,16 +160,50 @@ func (n *p2pNetwork) SubscribeRandoms(numSubnets int) error {
...
+	case n.cfg.NetworkConfig.InBooleTransitionWindow(currentSlot):
+		alanSubnets = n.persistentSubnets
+		booleSubnets = n.persistentSubnets
+	case n.cfg.NetworkConfig.BooleForkAtSlot(currentSlot):
+		booleSubnets = n.persistentSubnets
+	default:
+		alanSubnets = n.persistentSubnets
+	}
+
+	n.subscribedCommittees.Range(func(encodedCommittee string, statusAndSubnet statusWithSubnet) bool {
```
iurii-ssv, line 224: **Nit:** the switch inside this `Range` re-evaluates `InBooleTransitionWindow`/`BooleForkAtSlot` (both time-derived) once per subscribed committee. Hoisting the three-way decision above the `Range` — mirroring the persistent-subnets switch just above — avoids the repeated checks and also keeps the persistent and committee sides consistent if the fork/window state flips mid-iteration (harmless today, but a mixed Alan/Boole snapshot is avoidable for free).

operator/duties/committee.go
```diff
@@ -24,25 +25,76 @@ type CommitteeHandler struct {
...
+) *CommitteeHandler {
+	return &CommitteeHandler{
+		isAggregator: false,
+		attDuties:    attDuties,
+		syncDuties:   syncDuties,
 	}
+}
 
-	return h
+func NewAggregatorCommitteeHandler(
```
iurii-ssv: **Nit:** `NewCommitteeHandler` / `NewAggregatorCommitteeHandler` differ only in the `isAggregator` flag. A single constructor taking the flag (or the runner role) would keep the pair from drifting apart as the handler grows.

networkconfig/network.go
```diff
@@ -20,9 +24,89 @@ func (n Network) String() string {
 }
 
 const alanForkName = "alan"
+const boolePriorWindowEpochs = phase0.Epoch(1)    // epochs before Boole to subscribe to both topic sets
+const booleSubsequentWindowSlots = phase0.Slot(1) // slots after Boole to keep accepting old-topic messages
 
 // StorageName returns a config name used to make sure the stored network doesn't differ.
 // It combines the network name with fork name.
 func (n Network) StorageName() string {
 	return fmt.Sprintf("%s:%s", n.SSV.Name, alanForkName) // TODO: decide what forks change DB fork name
```
iurii-ssv: **Non-blocking:** `StorageName()` still pins the `alan` fork name per this TODO. Fine as a deliberate choice for the convergence, but the decision needs an owner before a Boole epoch actually ships in a release — flagging so it doesn't get lost.

network/topics/pubsub.go
momosh-ssv: Fixed in 52ce5a946.

Went with your suggestion: added `Controller.DeregisterTopics`, and the `UpdateSubnets` loop now sweeps the Alan topic set from the whitelist once the transition window closes. Turned out `Unsubscribe` already deregisters topics on teardown, so the sweep only needs to cover the construction-time registrations for subnets we never subscribed to.

network/p2p/p2p_pubsub.go
momosh-ssv: Fixed in 692e307ae.

Restored the log at the single broadcast site in `BroadcastAtSlot`, with the gossip msg ID and `topic_peers` count as before. Thanks for catching this, it was dropped unintentionally.

network/p2p/p2p_pubsub.go
momosh-ssv: Fixed in 4a77516e7.

Hoisted both checks above the `Range`; the persistent and committee sides now share one snapshot of the fork state.

operator/duties/committee.go
momosh-ssv: Fixed in 37008d61e.

Collapsed the pair into a single `NewCommitteeHandler` taking the `isAggregator` flag, call sites updated.

networkconfig/network.go
momosh-ssv: Agreed it should not get lost.

Filed #2956 so the decision has an owner before any Boole epoch ships. Folded the v1.2.3 spec-tag reminder from the go.mod thread into it as well.

go.mod
momosh-ssv: Follow-up filed: #2956 carries the reminder to move both go.mod files to the tagged v1.2.3 once it exists.

networkconfig/test-network.go
iurii-ssv: Thanks — the spec side is confirmed done: [`spec-test-boole-post`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/.github/workflows/spec-test.yml#L52-L92) runs green in CI, and [`spec-test-alan`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/.github/workflows/spec-test-alan.yml) is back. Leaving this open for the unit half.

Two things worth recording before this gets deferred:

**1. The remaining gap grew after your 07-20 assessment.** You named `network/p2p` as the one unknown, but the fork-aware metadata syncer landed after that and brought its own post-fork fixture failure. Reproduced at HEAD `db1d575`:

```
go test ./operator/validator/metadata/                          -> ok
SSV_TEST_BOOLE_FORK=post go test ./operator/validator/metadata/ -> FAIL
    --- FAIL: TestSyncer_Stream/SendsUpdatesAndStopsOnContextCancel
        syncer_test.go:475: Did not receive update in time
```

Cause is the fixture, not the code: [`share1`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/operator/validator/metadata/syncer_test.go#L421-L430) is built with an empty `Share.Committee`, so post-fork [`selfSubnets`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/operator/validator/metadata/syncer.go#L336-L356) takes the Boole arm and `BooleCommitteeSubnet()` returns `UnknownSubnetId`. Both `Subnets.Set` and `IsSet` bounds-guard that away, so the share sets no bit and then matches nothing in [`shareInOwnSubnets`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/operator/validator/metadata/syncer.go#L359-L369) — filtered out, no update, timeout. Populating `Committee` on the fixture (or pinning the test pre-fork like `Test_ValidateSSVMessage`/`TestMsgValidator` got) should be it.

For what it's worth, the rest of the flip looks healthier than feared — `message/validation`, `networkconfig`, `operator/validator`, `protocol/v2/ssv/...` all pass under `SSV_TEST_BOOLE_FORK=post` locally. (`network/topics` `TestTopicManager/happy_flow` fails for me in *both* modes, so that one is local mdns flakiness, not the flip.)

**2. There's no follow-up issue for this.** #2952 doesn't list the unit matrix, and nothing else does — so as it stands the deferral has no owner and the `alan_spec` / `SSV_TEST_BOOLE_FORK` unit path stays permanently unexercised. Could you fold it into #2952 (or file its own) covering: fork-agnostic `network/p2p` fixtures, the metadata-syncer fixture above, then enabling the unit-test matrix?

networkconfig/ssv.go
```diff
@@ -94,20 +103,38 @@ func (s *SSV) unmarshalFromConfig(aux marshaledConfig) error {
...
 
+	// If the config has no "forks" block at all (e.g. a stale custom-network YAML/JSON
+	// predating the Boole fork), default Boole to "never activates" rather than letting it
+	// zero-value to epoch 0 (fork-at-genesis). An explicit "forks: {Boole: 0}" is still
+	// honored verbatim. This protects custom-network operators from silently jumping to
+	// post-fork behavior when they haven't opted in.
+	forks := SSVForks{Boole: math.MaxUint64}
+	if aux.Forks != nil {
+		forks = *aux.Forks
+	}
```
iurii-ssv, line 142: **`Forks: {}` still activates Boole at genesis.** The `*SSVForks` pointer closes the "no forks block" case, but not "forks block present, field omitted" — and `SSVForks` has exactly one field, so an empty block is the easy shape to write. Verified against this commit:

```
absent       -> Boole=18446744073709551615   unscheduled  ✓
Forks:       -> Boole=18446744073709551615   unscheduled  ✓
Forks: {}    -> Boole=0                      fork at genesis
Forks:
  Boole: 0   -> Boole=0                      fork at genesis (intended)
```

`Forks: {}` is indistinguishable from an explicit `Boole: 0` here, so a hand-written custom-network config with an empty forks block silently opts into fork-at-genesis — with a `NextDomainType` that may well be unset too. That's the same failure mode [item 3 of #2952](https://github.com/ssvlabs/ssv/issues/2952) set out to prevent: [`generate-config`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/cli/generate_config.go#L128-L130) no longer *emits* that shape, but nothing stops a hand-written or previously-generated config from carrying it.

Making the marshaled `Boole` a pointer as well would fold "present but unset" into the same default as absent:

```go
type marshaledForks struct {
	Boole *phase0.Epoch `json:"Boole,omitempty" yaml:"Boole,omitempty"`
}
```

Minor while here: `SSVForks`'s json tag is `Boole`, whereas every sibling in `marshaledConfig` is snake_case (`domain_type`, `next_domain_type`, `registry_sync_offset`, …).

operator/validator/metadata/syncer.go
```diff
@@ -333,22 +333,38 @@ func (s *Syncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
...
+// slot ownSubnets was built from, so both sides resolve to the same fork phase.
+func (s *Syncer) shareInOwnSubnets(share *ssvtypes.SSVShare, ownSubnets networkcommons.Subnets, currentSlot phase0.Slot) bool {
+	switch {
+	case s.netCfg.InBooleTransitionWindow(currentSlot):
+		return ownSubnets.IsSet(share.AlanCommitteeSubnet()) || ownSubnets.IsSet(share.BooleCommitteeSubnet())
+	case s.netCfg.BooleForkAtSlot(currentSlot):
+		return ownSubnets.IsSet(share.BooleCommitteeSubnet())
+	default:
+		return ownSubnets.IsSet(share.AlanCommitteeSubnet())
+	}
```
iurii-ssv, line 378: **The transition-window arm over-matches — `ownSubnets` is a single bitmap holding the Alan ∪ Boole union.**

[`selfSubnets`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/operator/validator/metadata/syncer.go#L336-L356) sets both my Alan and my Boole subnet numbers into one flat 128-bit `Subnets`. This arm then ORs both of the share's lookups against that union, so a share matches whenever *either* of its subnet numbers collides with *any* of mine, in either namespace — a share whose Alan subnet is 7 matches because one of my *Boole* subnets happens to be 7.

That's the safe direction (a superset — nothing that should be synced gets dropped), so it isn't a correctness bug. The cost is that the sync candidate set roughly doubles for the duration of the window.

Flagging it mainly because [`operatorActiveSubnets`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/cli/operator/node.go#L656-L670) hit exactly this collision and deliberately went the other way, with a comment spelling out why:

> Boole and Alan subnets are tallied in separate bitmaps: they are independent gossip namespaces that happen to share a 0..127 index space, so a single bitmap would under-count on collision.

Either mirror that here (thread two `Subnets` through from `selfSubnets`), or add a line saying the over-match is accepted and bounded to the window — so the two places don't read as disagreeing about the same hazard.

observability/utils/format.go
```diff
@@ -5,10 +5,12 @@ import (
 	"strings"
 
 	spectypes "github.com/ssvlabs/ssv-spec/types"
+
+	ssvtypes "github.com/ssvlabs/ssv/protocol/v2/types"
 )
 
 func FormatRunnerRole(runnerRole spectypes.RunnerRole) string {
-	return strings.TrimSuffix(runnerRole.String(), "_RUNNER")
+	return strings.TrimSuffix(ssvtypes.RunnerRoleToString(runnerRole), "_RUNNER")
```
iurii-ssv, line 17: **There are now three `RunnerRole` → string mappers that have to stay in lockstep.**

This one (trim-based, delegating to `types.RunnerRoleToString`), [`types.RunnerRoleToString`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/protocol/v2/types/runner_role.go#L81-L90) (deprecated-role shim + spec `String()` fallback), and [`message.RunnerRoleToString`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/protocol/v2/message/msg.go#L117-L136) (a hand-maintained switch with its own string constants). This one and `message.`'s produce byte-identical output for every role — by completely different mechanisms.

That's worth noting because #2955 was precisely this bug class, and the fix landed in one of the three. The externally-visible surface happens to be safe: I checked every role `.String()` call site under `api/handlers/exporter/` and `exporter/` and they're all `spectypes.BeaconRole`, which still knows all of its values. But the next role added or deprecated has two independent places to get right, and only one of them fails loudly.

Non-blocking — a follow-up collapsing them (or at minimum a cross-reference in each doc comment) would do.

network/topics/controller.go
```diff
@@ -206,17 +200,23 @@ func (ctrl *topicsCtrl) Broadcast(topicName string, data []byte, timeout time.Du
...
 	return nil
 }
 
+// DeregisterTopics removes the given topics from the subscription-filter whitelist.
+func (ctrl *topicsCtrl) DeregisterTopics(names ...string) {
+	whitelist := ctrl.subFilter.(Whitelist)
+	for _, name := range names {
+		whitelist.Deregister(name)
+	}
+}
```
iurii-ssv, line 214: **Unchecked type assertion on new public interface surface.** `ctrl.subFilter.(Whitelist)` panics if `subFilter` is ever anything but the concrete `*subFilter`.

Not a regression — [`Subscribe`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/network/topics/controller.go#L168) and [`Unsubscribe`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/network/topics/controller.go#L230) already do the same. But those are internal call sites, whereas `DeregisterTopics` is new `Controller` API, and [`SubFilter`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/network/topics/sub_filter.go#L12-L18) deliberately keeps `// Whitelist` commented out of the interface, so nothing enforces the relationship at compile time. A comma-ok that no-ops is essentially free here.

If the assertion is meant to be a hard invariant, the cleaner fix is the other direction: promote `Whitelist` into `SubFilter` and delete all three assertions.

iurii-ssv (commented):
Another review-round of things to check.

FYI @momosh-ssv, I also posted a followup on https://github.com/ssvlabs/ssv/pull/2941#discussion_r3636355145 thread from the round before - probably want to check out as well.

network/p2p/p2p_pubsub.go
```diff
@@ -38,6 +41,41 @@ func (n *p2pNetwork) UseMessageRouter(router network.MessageRouter) {
 
 // Broadcast publishes the message to all peers in subnet
 func (n *p2pNetwork) Broadcast(msgID spectypes.MessageID, msg *spectypes.SignedSSVMessage) error {
+	msgSlot, err := broadcastMessageSlot(msg)
+	if err != nil {
+		return fmt.Errorf("could not resolve message slot: %w", err)
+	}
+
+	return n.BroadcastAtSlot(msg, msgSlot)
```
iurii-ssv, line 52: **Not a defect — a labelling request.** This method now has **zero production callers**. Every real broadcast path goes to `BroadcastAtSlot`: the QBFT instance resolves the slot from its own height in [`Instance.Broadcast`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/protocol/v2/qbft/instance/instance.go#L169-L176), and the runners were switched over earlier in this PR. The only remaining callers repo-wide are [`protocol/v2/testing/network.go`](https://github.com/ssvlabs/ssv/blob/db1d575381562b5664930944089b93e45f94516b/protocol/v2/testing/network.go#L34) and test doubles.

It can't simply be deleted — it satisfies spec's `p2p.Broadcaster`, which the pinned spec embeds into `ssv.Network` and which reaches us via `protocolp2p.Network` → `network.P2PNetwork`. So it has to stay.

Two things worth saying at the declaration, since neither is discoverable from here today:

1. That it's spec-interface conformance surface, not a live path — otherwise it reads as a peer alternative to `BroadcastAtSlot`, and the next person with a slot in hand gets no signal about which to reach for.
2. That its contract is now **narrower** than what it replaced: `queue.DecodeSignedSSVMessage` also resolved `SSVEventMsgType`, whereas `broadcastMessageSlot` returns `ErrUnknownMessageType` for it. That divergence is documented on the `default` branch below, but a caller looking at `Broadcast` won't see it.

Something like:

```go
// Broadcast publishes the message to all peers in subnet, resolving the slot from the message
// body. It exists to satisfy spec's p2p.Broadcaster; no production path calls it — prefer
// BroadcastAtSlot, which every caller with a duty slot in hand already uses. Note this handles
// only consensus and partial-signature messages (see broadcastMessageSlot).
```

Raising it because [item 6 of #2952](https://github.com/ssvlabs/ssv/issues/2952) framed this as a "cold path" — it's colder than that. The optimisation itself is fine; it just reads as if it were on a live path.

networkconfig/ssv.go
momosh-ssv: Fixed in 9c7d9d1f0.

Went with the pointer shape you suggested, via a dedicated `marshaledForks` wire type — `Forks: {}`, `Forks:` (null), and an omitted `Boole` all fold into the same never-activates default as an absent block now, and the UT covers each of your four shapes. Took the snake_case json tag along (`boole`); checked that nothing persists the JSON form (`storageCompatToken` is a plain constant), so seems there's no compat concern.

networkconfig/test-network.go
momosh-ssv: The metadata-syncer fixture is fixed in a359f55cd — `share1` now carries a committee, so it maps to a real subnet in both modes, and `./operator/validator/metadata` passes locally with and without `SSV_TEST_BOOLE_FORK=post`. Good catch on the empty-committee shape.

On the missing owner: agreed. #2952 is closed, so I filed #2964 as the dedicated owner — fork-agnostic `network/p2p` fixtures, a full-suite post-flip sweep, then enabling the unit matrix (with the `RunSyncCommitteeAggProof` alan stub folded in as an optional item).

operator/validator/metadata/syncer.go
momosh-ssv: Fixed in cb1eb0828.

Went with the documentation route: `selfSubnets` now spells out that the union can only over-match (a superset, bounded to the window) and cross-references the two-bitmap choice in `operatorActiveSubnets`, so the two sites no longer read as disagreeing. Threading two bitmaps through the syncer felt like real churn for a transient window cost — happy to revisit if you feel differently.

network/topics/controller.go
momosh-ssv: Fixed in 07fe816db.

Took your "other direction": `Whitelist` is promoted into `SubFilter` (the commented-out embed finally made it in), and all three assertions are gone — the relationship is compile-time enforced now.

network/p2p/p2p_pubsub.go
momosh-ssv: Fixed in abaf48003.

Adopted your wording essentially as-is at the declaration: spec's `p2p.Broadcaster` conformance surface, prefer `BroadcastAtSlot`, and the narrower consensus/partial-sig-only contract.

observability/utils/format.go
momosh-ssv: Fixed in 33f8057b6.

Went with your suggested minimum for now: each of the three mappers carries a cross-reference recording the lockstep constraint (and noting only `message.`'s switch fails loudly on drift). Collapsing them felt like its own change — I'd take that as a follow-up.

network/topics/pubsub.go
```diff
@@ -127,8 +129,8 @@ func NewPubSub(
...
 		sf.(Whitelist).Register(topic)
```
iurii-ssv: Leftover from the `Whitelist` → `SubFilter` promotion: `sf` is a `SubFilter`, which now embeds `Whitelist`, so this assertion can never fail and the call can be direct:

```suggestion
		sf.Register(topic)
```

(Same cosmetic pattern in `sub_filter_test.go`'s `sf.(Whitelist).Register(...)` calls.)

networkconfig/ssv.go
```diff
@@ -59,25 +65,37 @@ func (s *SSV) String() string {
...
 		Name:                    s.Name,
 		DomainType:              s.DomainType[:],
+		NextDomainType:          s.NextDomainType[:],
 		RegistrySyncOffset:      s.RegistrySyncOffset,
 		RegistryContractAddr:    s.RegistryContractAddr,
 		Bootnodes:               s.Bootnodes,
 		DiscoveryProtocolID:     s.DiscoveryProtocolID[:],
 		TotalEthereumValidators: s.TotalEthereumValidators,
-		Forks:                   s.Forks,
+		Forks:                   &marshaledForks{Boole: &boole},
```
iurii-ssv: Non-blocking, marshal-side counterpart of the unmarshal fix: `marshal()` always emits a non-nil `Boole`, so any config with an *unscheduled* fork — every built-in network today — serializes with the sentinel spelled out: `String()`, logs, and re-marshaled YAML/JSON all show `boole: 18446744073709551615`. Round-trip stays correct (an explicit epoch is honored verbatim), so this is purely cosmetic — but skipping the field when it's `math.MaxUint64` would keep the magic number out of printed/generated configs and make marshal symmetric with the unmarshal default:

```go
forks := &marshaledForks{}
if s.Forks.Boole != math.MaxUint64 {
	boole := s.Forks.Boole
	forks.Boole = &boole
}
```

message/validation/common_checks.go
```diff
@@ -38,11 +40,11 @@ func (mv *messageValidator) messageEarliness(slot phase0.Slot, receivedAt time.T
...
 	switch role {
-	case spectypes.RoleProposer, spectypes.RoleSyncCommitteeContribution:
+	case spectypes.RoleProposer, ssvtypes.RoleSyncCommitteeContribution:
 		ttl = 1 + LateSlotAllowance
-	case spectypes.RoleCommittee, spectypes.RoleAggregator:
+	case spectypes.RoleCommittee, spectypes.RoleAggregatorCommittee, ssvtypes.RoleAggregator:
 		ttl = mv.maxStoredSlots()
-	case spectypes.RoleValidatorRegistration, spectypes.RoleVoluntaryExit:
+	default:
 		return 0
```
iurii-ssv, line 52: **VR/VE messages are never late here.** `RoleValidatorRegistration` / `RoleVoluntaryExit` fall through to `return 0`, so the only time bound on them is the early-message check — while [`ErrLateSlotMessage`'s text](https://github.com/ssvlabs/ssv/blob/33f8057b6ef5c8957aef569a658452de5bdf610e/message/validation/errors.go#L76) ("+3(else)") and the [slot-timing rule in ssvlabs/knowledge-base#2](https://github.com/ssvlabs/knowledge-base/blob/849c3356489e15a6f6d723cb03316baae4c24556/p2p/MessageValidation/Rules.md?plain=1#L966) both claim a +3-slot window for these roles.

Pre-existing behavior, not this PR's regression — but with knowledge-base#2 codifying the ruleset now, it's the right moment to pick a side: either bound them (`ttl = 1 + LateSlotAllowance`, like proposer) or keep the exemption deliberate and fix the error text + doc to record it. Practical exposure is limited either way (`ErrSlotAlreadyAdvanced` and the per-epoch duty caps still apply), so non-blocking.

message/validation/partial_validation.go
iurii-ssv, line 1: **Recording a doc↔impl divergence** (file-level comment — the lines predate this PR's hunks): `validatePartialSignatureMessageSemantics` deliberately skips the `ValidatorIndex ∈ committee` assertion for committee roles ([L144-L154](https://github.com/ssvlabs/ssv/blob/33f8057b6ef5c8957aef569a658452de5bdf610e/message/validation/partial_validation.go#L144-L154)) — per the in-code rationale, operators aren't assumed synced on each other's validator sets. The [rule in ssvlabs/knowledge-base#2](https://github.com/ssvlabs/knowledge-base/blob/849c3356489e15a6f6d723cb03316baae4c24556/p2p/MessageValidation/Rules.md?plain=1#L870) requires it for every role. Volume stays capped by min(2V, V+512) / min(5V, V+4·512) and the per-index 2/5 limits, but the indices themselves are unconstrained for committee duties.

Non-blocking — raising it so the relaxation gets an explicit owner: either the doc records it (flagged there as well), or we conclude the check is feasible after all and add it here.

y0sher (dismissed):

message/validation/seen_msg_types.go
iurii-ssv, line 1: Nit, pre-existing: the header comment on line 3 still says "message_counts.go" — stale filename reference left from an old rename.

iurii-ssv (dismissed):
LGTM, posted some minor suggestions we can address / look into as a followup.

network/topics/pubsub.go
momosh-ssv: Fixed in 99fe781ae.

`pubsub.go` now calls sf.Register directly, and the same assertions are dropped from `sub_filter_test.go`.

networkconfig/ssv.go
momosh-ssv: Fixed in e81b3154a.

Marshal now skips Boole when it's the MaxUint64 sentinel, so unscheduled forks serialize with the field absent, symmetric with the unmarshal default. Explicit epochs are still emitted verbatim.

message/validation/common_checks.go
momosh-ssv: Good catch.
Went with recording the exemption rather than bounding: 207790071 gives VR/VE an explicit case in messageLateness with a comment marking it deliberate, and fixes ErrLateSlotMessage's text to stop claiming +3 for them. 

I fear bounding them here would start rejecting currently-valid traffic network-wide, so if we want that it should be its own deliberate change. The knowledge-base rule still needs the matching correction.
I'll follow up on ssvlabs/knowledge-base#2.

message/validation/partial_validation.go
momosh-ssv: Agreed this needs an explicit owner.
I'd say the doc is the right place to record the relaxation: the in-code rationale (operators aren't assumed synced on each other's validator sets) still holds, and the duty/count caps bound the exposure. 

Since you've flagged it on knowledge-base#2 as well, maybe we settle it there and keep this thread as the code-side pointer?

message/validation/seen_msg_types.go
momosh-ssv: Fixed in 11ec0d844.

migrations/migration_9_committee_duty_role_field.go
```diff
@@ -0,0 +1,281 @@
...
+				CommitteeID:  spectypes.CommitteeID(legacyTrace.CommitteeID),
+				OperatorIDs:  convertOperatorIDsV1(legacyTrace.OperatorIDs),
+				ProposalData: legacyTrace.ProposalData,
+				SyncCommittee: convertSignerDataV1(
+					legacyTrace.SyncCommittee,
+				),
+				Attester: convertSignerDataV1(legacyTrace.Attester),
+			}
+
+			value, err := trace.MarshalSSZ()
```
iurii-ssv, line 139: **The migration's _target_ type isn't frozen.**

The legacy source side is correctly frozen (`migration_9_CommitteeDutyTraceV1` plus its own generated encoding), but the write side marshals into the **live** `traces.CommitteeDutyTrace` and derives the key byte from the **live** `estore.CommitteeRunnerRoleToPrefix`. If a later PR adds a field to `CommitteeDutyTrace`, reorders one, or changes the role→byte map, migration_9 silently starts emitting a format that a future migration_10 won't expect it to have produced — and nothing in CI would notice, because the migration's own tests assert round-trips through whatever the live type happens to be.

A full freeze (a `migration_9_CommitteeDutyTraceV2` mirroring the V1 model) is the textbook fix, but it costs another ~1.5k lines of generated encoding for a one-shot migration. The cheap guard that catches the same failure mode is a golden-bytes test — it fails the instant either dependency changes shape, and tells the next person what to do about it:

```go
// migration_9 writes the LIVE traces.CommitteeDutyTrace and the LIVE role→byte map.
// If this fails, one of them changed shape: freeze migration_9's target into a V2 model
// rather than letting an old-DB upgrade silently emit the newest format.
func TestMigration9TargetLayoutIsFrozen(t *testing.T) {
	got, err := (&traces.CommitteeDutyTrace{ /* fixed fixture */ }).MarshalSSZ()
	require.NoError(t, err)
	require.Equal(t, "0x…", hexutil.Encode(got))

	b, err := estore.CommitteeRunnerRoleToPrefix(spectypes.RoleCommittee)
	require.NoError(t, err)
	require.Equal(t, byte(0x00), b)
}
```

Golden bytes rather than a `SizeSSZ` assertion — they also catch a field reordering that happens to preserve length.

migrations/migration_9_committee_duty_role_field.go
```diff
@@ -0,0 +1,281 @@
...
+
+	estore "github.com/ssvlabs/ssv/exporter/store"
+	traces "github.com/ssvlabs/ssv/exporter/traces"
+	"github.com/ssvlabs/ssv/storage/basedb"
+)
+
+// migrationBatchSize caps how many rewrites are held in memory before being flushed in a
+// single Update transaction. A mainnet exporter can have tens of millions of legacy "cd"
+// records, so we must not accumulate them all at once.
+const migrationBatchSize = 5000
```
iurii-ssv, line 22: **`flush()` can hit badger's `ErrTxnTooBig`.**

Each flush puts up to 5000 Sets + 5000 Deletes into a single `opt.Db.Update` transaction. Badger's limit is 15% of `MemTableSize` — ~9.6MB under the `badger.DefaultOptions` this repo opens with — which works out to roughly 960 bytes average per record before the transaction is rejected outright. A committee duty trace carrying a few hundred `Attester`/`SyncCommittee` `SignerData` entries clears that comfortably, and the failure mode is the node refusing to start.

This is unreachable today rather than a live bug: only archive exporters write `"cd"` keys, and `openNodeDB` puts exporters on pebble, where batches are memory-bound. But the migration is backend-agnostic, the badger test case runs on tiny fixtures, and nothing else would catch it if that mapping ever changes.

Lowering the count alone doesn't close it — a single record can reach ~4MB via `ProposalData` (`ssz-max:"4194532"`), so even a two-record batch can blow a count-based limit. Byte-budgeted batching is backend-agnostic and bounds the worst case properly:

```suggestion
// migrationBatchSize caps how many rewrites are held in memory before being flushed in a
// single Update transaction. A mainnet exporter can have tens of millions of legacy "cd"
// records, so we must not accumulate them all at once.
const migrationBatchSize = 5000

// migrationBatchBytes caps the accumulated payload per flush. badger's transaction limit is
// 15% of MemTableSize (~9.6MB under badger.DefaultOptions), and a single CommitteeDutyTrace can
// reach ~4MB via ProposalData, so flushing at 2MB keeps the worst case (threshold plus one
// oversized record) well under the cap. Pebble batches are memory-bound; this costs nothing there.
const migrationBatchBytes = 2 << 20
```

with the counter tracked alongside `pending` and reset in `flush()`:

```go
pendingBytes += len(rewrite.value) + len(rewrite.newKey) + len(rewrite.oldKey)
if len(pending) >= migrationBatchSize || pendingBytes >= migrationBatchBytes {
	return flush()
}
```

Minor, while in this comment block: "we must not accumulate them all at once" only holds on pebble. Badger's `GetAll` materializes every key for the prefix up front via `listRawKeys` before the first callback fires, so the batching bounds accumulated *values* only. Accurate for the path that actually has data — just worth not implying it's true of both backends.

message/validation/partial_validation.go
iurii-ssv: Agreed — `knowledge-base#2` is the right place to settle it, and I'm satisfied the relaxation itself is correct.

Digging in, the case for it is stronger than "operators aren't assumed synced on each other's validator sets". [`buildCommittee`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L645-L661) appends `share.ValidatorIndex` unconditionally, while the index maps just above it are gated on [`HasBeaconMetadata()`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L355) — so between a contract event landing and beacon metadata sync completing, the committee's `Indices` carries a literal `0` in place of the real index. Enforcing `ValidatorIndex ∈ Indices` for committee roles would reject legitimate post-consensus messages for exactly those validators, and since a rejection costs the sender gossip score, the nodes furthest behind would penalize the peers they most need to catch up from. Self-reinforcing, and clearly worse than the exposure it would close.

I also looked at the obvious middle path — reject only indices the node *positively knows* belong to another committee, allow unknown ones, which would be monotone in sync state. `ValidatorByIndex` already exists on the store, so it's a cheap O(1) lookup and would only need adding to validation's narrow `validatorStore` interface. It doesn't hold up: validators change committees whenever the operator set changes, so a node holding a stale index→committee mapping would reject legitimate messages from the *new* committee. Same class of failure, just rarer — not worth a per-message-per-index hot-path lookup to buy a partial check.

Two asks before we close this out:

1. **The doc should record the blast radius, not just the exemption.** "Committee roles are exempt" won't be reviewable in a year. The precise version: the signer is already authenticated ([`belongsToCommittee`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/message/validation/validation.go#L203) plus signature verification), so this is an insider capability rather than something open to arbitrary peers — a committee member can attach indices outside its own committee to its own messages, and the caps bound volume, not content. Where that content actually lands is the exporter: these indices populate `SignerData.ValidatorIdx`, so the practical exposure is mis-attributed participation in exporter query results, nothing consensus-affecting.

2. **Put the code-side pointer in the code.** This thread stops being anywhere anyone looks once a 248-file PR merges. A one-line comment at the [relaxation site](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/message/validation/partial_validation.go#L144-L154) referencing `knowledge-base#2` survives the merge and lands in front of the next person reading that code.

Resolving on that basis.

message/validation/validation.go
```diff
@@ -291,7 +331,7 @@ func (mv *messageValidator) getCommitteeAndValidatorIndices(msgID spectypes.Mess
 			return CommitteeInfo{}, ErrNoValidators
```
iurii-ssv, line 337: **`ErrNoValidators` is unreachable** — pre-existing, not this PR's doing, but this line is in the diff so it's a convenient moment.

I traced every writer to `byCommitteeID`: [L378](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L378) (add), [L456](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L456) (remove), [L525](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L525) (update). All three store a committee produced by `buildCommittee` from at least one share, and `buildCommittee` [appends exactly one index per share](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L660). The one path that could produce an empty committee is short-circuited: `handleShareRemoved` computes `committeeRemoved := len(newCommittee.Shares) == 0` and `delete`s the entry outright, so `Committee()` returns not-found and you get `ErrNonExistentCommitteeID` above instead. Net: every committee reachable here satisfies `len(Indices) == len(Shares) >= 1`.

Worth flagging rather than leaving alone, because the check reads as if it guards "this committee has no *known* validator indices" — and it doesn't. `Indices` carries a literal `0` for any share whose beacon metadata hasn't arrived yet (see my comment on the `Indices` field). So anyone who "fixes" `buildCommittee` to skip metadata-less shares would silently make this branch live, and it would then reject *every* message for a committee that's mid-metadata-sync.

Either drop the branch, or keep it and mark it explicitly as a defensive assertion with a note that `Indices` is share-cardinality rather than known-index-cardinality. Silent dead code that becomes a network-wide reject on an innocuous refactor is the outcome worth avoiding.

registry/storage/validatorstore.go
```diff
@@ -57,6 +59,35 @@ type Committee struct {
 	Operators []spectypes.OperatorID
 	Shares    []*types.SSVShare
 	Indices   []phase0.ValidatorIndex
```
iurii-ssv: **`Indices` deserves a doc comment** — pre-existing behavior, but this hunk is the natural place for it.

`Indices` is share-cardinality, not known-index cardinality. [`buildCommittee`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L660) appends `share.ValidatorIndex` unconditionally, and [`HasBeaconMetadata()`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/protocol/v2/types/ssvshare.go#L103) is `Status != ValidatorStateUnknown` — so a share added by a `ValidatorAdded` event contributes a literal `0` here until beacon metadata sync populates its index. Entries stay positionally aligned with `Shares`, which [`removeShareFromCommittee`'s corrupt-state check](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/registry/storage/validatorstore.go#L582) depends on.

This reads exactly like an oversight — I assumed it was one — but it's load-bearing in two places, which is precisely why it needs saying out loud:

- [`dutyLimit`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/message/validation/common_checks.go#L107-L124) for `RoleCommittee` returns `min(slotsPerEpoch, 2*len(validatorIndices))`. Filtering the zeros out would shrink the count and *tighten* the duty limit against legitimate traffic during the sync window.
- The `len(committee.Indices) == 0` → `ErrNoValidators` check in message validation would become live and reject every message for a committee mid-metadata-sync.

```suggestion
	// Indices holds one entry per share in Shares, positionally aligned with it — not one
	// entry per *known* validator index. buildCommittee appends share.ValidatorIndex
	// unconditionally, so a share added by a contract event contributes a literal 0 until
	// beacon metadata sync populates its real index. Consumers must not assume every entry
	// is a live index, and must not filter the zeros out: the cardinality is relied upon by
	// the committee duty limit (min(slotsPerEpoch, 2*len)) and by the len()==0 check in
	// message validation, both of which would tighten against legitimate traffic if it shrank.
	Indices []phase0.ValidatorIndex
```

(gofmt-checked — the comment line starts a new alignment group, so single-space `Indices []phase0.ValidatorIndex` is the formatted result.)

One genuine, if cosmetic, consequence while we're here: [`buildValidatorToCommitteeIndex`](https://github.com/ssvlabs/ssv/blob/11ec0d844a6cb71ac66808009e8e86db355ff168/exporter/dutytracer/collector.go#L1747) builds `map[index]committeeID`, so every committee holding a metadata-less share writes `result[0]` and index 0 aliases to whichever committee is iterated last. It's only consulted for indices that actually appear in a beacon duty schedule, so it'd take validator index 0 genuinely being an SSV validator to bite. Latent, not live — noting it so it isn't rediscovered as a mystery later.

migrations/migration_9_committee_duty_role_field.go
momosh-ssv: Fixed in 14319945d.

Flush now also triggers at 2MB of accumulated payload alongside the 5000-record cap, and the batch comment stops implying key-side bounds on badger. Took your constants and reasoning pretty much verbatim. 
Thanks for doing the `MemTableSize` math.

migrations/migration_9_committee_duty_role_field.go
momosh-ssv: Fixed in 8f3b577fb.

Added `TestMigration9TargetLayoutIsFrozen`, pinning golden bytes for a fixture that exercises the nested consensus traces, signer data and proposal data, plus the RoleCommittee prefix byte. Agreed the full V2 freeze is not worth ~1.5k generated lines for a one-shot migration; the golden test fails the moment either dependency moves and tells the finder to freeze the target then.

message/validation/partial_validation.go
momosh-ssv: Both asks handled. The code-side pointer is in aa984343d — one line at the relaxation site referencing ssvlabs/knowledge-base#2. For the blast radius, I have carried your framing over to knowledge-base#2 so it lands in the doc rather than staying buried in this thread — the doc edit itself stays with the kb author as agreed.

registry/storage/validatorstore.go
momosh-ssv: Fixed in 7de8ab0f9.

Took your comment essentially verbatim, just split `Indices` out of the field group so it reads as its own block. Also ack on the `buildValidatorToCommitteeIndex` index-0 aliasing, and agreed it is latent (it takes validator index 0 genuinely being an SSV validator to bite), so I would leave it recorded in this thread rather than touch the collector here.

message/validation/validation.go
momosh-ssv: Fixed in 7de8ab0f9.

Went with keep-and-mark over dropping it. I fear removing a validation-path branch while the PR sits at LGTM is more churn than the dead check costs, and as a marked assertion it documents exactly the trap you traced: the comment spells out that `Indices` is share-cardinality, so a `buildCommittee` refactor that skips metadata-less shares cannot silently turn this into a network-wide reject.

iurii-ssv (dismissed):

iurii-ssv (approved):
One request before merging: refresh the stale PR description:
- The "⚠️ one conflict to resolve in `operator/node.go`" note is outdated — `stage` has already been synced into this branch (its current tip is the merge-base), the conflict is resolved in-branch, and the PR now merges cleanly.
- The "⏳ #2939 + #2940 not yet merged into `integration`" caveat is likewise outdated — both are merged and their changes are included in this diff, so the "merge only once they're in" condition is satisfied.
- While editing: consider also noting that `local-testnet` now activates Boole from genesis (a behavioral flip for local dev clusters — see #2978, item 6).

Some more of non-blocking hardening findings from this review are tracked in #2978 (custom-config `NextDomainType` validation being the one to land before a fork is scheduled on a shared network).

y0sher (approved):

PR #2944, opened 2026-07-13
fix(p2p): make CanSubscribe fork-aware so Boole topics aren't rejected (#2943)

### LINKED ISSUE #2943: Boole: subscription filter rejects all Boole committee topics (CanSubscribe not fork-aware) — total post-Boole outage
## Summary

On `integration/boole-convergence`, an operator **cannot subscribe to or broadcast on any Boole-fork committee topic** (`/ssv/<network>/boole/<subnet>`). libp2p rejects it with `topic is not allowed by the subscription filter`. This breaks **all** post-Boole operation (attestations, sync committee, aggregations, validator registrations, QBFT consensus). Pre-Boole (Alan `ssv.v2.<subnet>` topics) is unaffected.

**Severity: P0 — fork blocker.** Reproduced at HEAD `19e06375` on stage-hoodi (2026-07-13).

## Root cause

The subscription filter's "is this topic of interest" gate was never updated for the Boole topic-naming scheme. `network/topics/sub_filter.go`:

```go
func (sf *subFilter) CanSubscribe(topic string) bool {
    if commons.GetTopicBaseName(topic) == topic {
        // not of the same fork
        return false
    }
    return sf.Whitelisted(topic)
}
```

`GetTopicBaseName` only strips the **Alan** prefix (`network/commons/subnets.go`):

```go
const topicPrefix = "ssv.v2"

func GetTopicBaseName(topicName string) string {
    return strings.TrimPrefix(topicName, topicPrefix+".")   // only "ssv.v2."
}
```

A Boole topic (`topicRoot = "/ssv"`, `BooleTopic = "/ssv/<net>/boole/<subnet>"`) does **not** start with `ssv.v2.`, so `TrimPrefix` returns it unchanged → `GetTopicBaseName(topic) == topic` is **true** → `CanSubscribe` returns **false** for every Boole topic. Because `FilterIncomingSubscriptions` also runs through `CanSubscribe`, the node refuses its own committee topic.

Note the inconsistency: `ParseTopicSubnet` (`network/commons/subnets.go`) **was** updated to understand both the Alan (`ssv.v2.`) and Boole (`/ssv/.../boole/`) schemes — but `GetTopicBaseName` / `CanSubscribe` were not. The fork-awareness is only half-applied.

## Impact — two failure modes (both observed)

- **(a) FATAL crash-loop at startup:** the boot-time "subscribe to 1 random subnet" step draws a Boole subnet → rejected → `FATAL could not start node` → CrashLoopBackOff.
- **(b) Running-node consensus halt:** nodes that started fine drop every QBFT round-change / prepare / commit broadcast (`could not broadcast msg: topic is not allowed by the subscription filter`) → `QBFT instance is decided` goes to zero. No duties complete.

## Reproduction (stage-hoodi, 2026-07-13)

- **Always-post-Boole (fork epoch 0):** 0 successful duties since startup; some nodes crash-loop (mode a), the rest halt consensus (mode b).
- **Mixed committees** (some members post-Boole, some pre-Boole): committees with a post-Boole majority are fully stuck; committees carried by a pre-Boole majority keep working (the post-Boole members can't broadcast).
- **Live fork transition:** a committee healthy pre-fork went to a complete flatline the instant it crossed the fork boundary:

  | Window (around the fork slot / activation time) | `QBFT instance is decided` | `successfully submitted` |
  |---|---|---|
  | ~12 min before fork | 616 | 616 |
  | ~5 min after fork | 0 | 0 |

  Last successful submission was one slot before the fork; `topic is not allowed by the subscription filter` began at the fork activation instant (~850 occurrences/node and ongoing).

This is specifically the subscription filter rejecting the **correct** post-fork topic — there are **no** `wrong domain` errors, so it is distinct from the earlier proposer/validator-registration domain-stamping bug (#2915).

## Suggested fix

Make the "of interest" check fork-aware. `ParseTopicSubnet` already recognizes both the Alan and Boole topic schemes — reuse it in `CanSubscribe`: treat a topic as of-interest if `ParseTopicSubnet` returns a valid subnet (Alan **or** Boole), then apply the whitelist as today. Add unit tests for `CanSubscribe` with Boole topics, plus a post-Boole subscription test so this can't regress.

momosh-ssv:
## What

Fixes #2943 — the subscription filter rejected **every** Boole-fork committee topic (`/ssv/<network>/boole/<subnet>`), causing a total post-Boole outage on `integration/boole-convergence`.

## Root cause

`network/topics/sub_filter.go` gated topics with:

```go
if commons.GetTopicBaseName(topic) == topic { // GetTopicBaseName only strips the Alan "ssv.v2." prefix
    return false // "not of the same fork"
}
```

A Boole topic has no `ssv.v2.` prefix, so `GetTopicBaseName` returns it unchanged → the guard treats it as a foreign fork and returns `false` **before the whitelist is consulted** — even though `Topics()` correctly seeds Boole topics. `ParseTopicSubnet` was already taught both schemes; `CanSubscribe` was not. The fork-awareness was only half-applied.

Because `FilterIncomingSubscriptions` also runs through `CanSubscribe`, the node refused its own committee topic.

## Fix

Gate on `commons.ParseTopicSubnet` (understands Alan **and** Boole subnet topics); recognized topics still pass through the dynamic whitelist as before. Foreign/malformed topics remain rejected.

```go
func (sf *subFilter) CanSubscribe(topic string) bool {
    if _, _, err := commons.ParseTopicSubnet(topic); err != nil {
        return false // not a recognized subnet topic
    }
    return sf.Whitelisted(topic)
}
```

## Tests

- New `TestSubFilter_CanSubscribeBoole`: Boole topic accepted only when whitelisted; other Boole subnet rejected; Alan + Boole coexist; malformed / foreign-fork topics rejected outright.
- Existing `TestSubFilter` still passes (verified all assertions hold under the new gate).

`go build ./network/topics/` OK; `go test ./network/topics/ -run TestSubFilter` PASS. (`TestTopicManager/happy_flow` fails in my sandbox due to blocked mDNS peer discovery — confirmed it fails identically on the unmodified base, so it's pre-existing/environmental, not from this change.)

## Verification (before this fix, on stage-hoodi 2026-07-13)

Reproduced live on the Boole-fork deployment: postfork/forking nodes emit ongoing `topic is not allowed by the subscription filter` (self-subscribe + broadcast); prefork (Alan) nodes are healthy. After merge, a post-Boole committee should show zero filter errors and resumed duty submissions — I'll redeploy + verify.

### DIFF
--- a/network/topics/sub_filter_test.go
+++ b/network/topics/sub_filter_test.go
@@ -6,12 +6,10 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/ssvlabs/ssv/network/commons"
-	"github.com/ssvlabs/ssv/observability/log"
 )
 
 func TestSubFilter(t *testing.T) {
-	l := log.TestLogger(t)
-	sf := newSubFilter(l, 2)
+	sf := newSubFilter()
 
 	require.False(t, sf.CanSubscribe("xxx"))
 	require.False(t, sf.CanSubscribe(commons.GetTopicFullName("xxx")))
@@ -26,8 +24,7 @@ func TestSubFilter(t *testing.T) {
 // total post-Boole outage (self-subscribe crash-loop + QBFT broadcast failures).
 func TestSubFilter_CanSubscribeBoole(t *testing.T) {
 	const networkName = "hoodi-stage"
-	l := log.TestLogger(t)
-	sf := newSubFilter(l, 2)
+	sf := newSubFilter()
 
 	booleTopic := commons.BooleTopic(networkName, 51)
 
--- a/network/topics/sub_filter_test.go
+++ b/network/topics/sub_filter_test.go
@@ -39,6 +39,10 @@ func TestSubFilter_CanSubscribeBoole(t *testing.T) {
 	// ...and a different, unregistered Boole subnet stays rejected.
 	require.False(t, sf.CanSubscribe(commons.BooleTopic(networkName, 52)))
 
+	// A well-formed Boole topic for a *different* network still parses, so it's not
+	// rejected at the gate — only the (per-network) whitelist turns it down.
+	require.False(t, sf.CanSubscribe(commons.BooleTopic("other-network", 51)))
+
 	// Alan and Boole topics coexist in the same filter.
 	sf.(Whitelist).Register(commons.GetTopicFullName("1"))
 	require.True(t, sf.CanSubscribe(commons.GetTopicFullName("1")))
--- a/network/topics/sub_filter_test.go
+++ b/network/topics/sub_filter_test.go
@@ -47,7 +47,7 @@ func TestSubFilter_CanSubscribeBoole(t *testing.T) {
 	// Malformed / foreign-fork topics are rejected outright, even if "whitelisted",
 	// because they don't parse as a known subnet topic.
 	for _, bad := range []string{
-		"/ssv/" + networkName + "/boole/",     // missing subnet
+		"/ssv/" + networkName + "/boole/",      // missing subnet
 		"/ssv/" + networkName + "/otherfork/1", // unknown fork segment
 		"/ssv/" + networkName + "/boole/abc",   // non-numeric subnet
 	} {
--- a/network/topics/sub_filter_test.go
+++ b/network/topics/sub_filter_test.go
@@ -19,3 +19,39 @@ func TestSubFilter(t *testing.T) {
 	require.True(t, sf.CanSubscribe(commons.GetTopicFullName("1")))
 	require.False(t, sf.CanSubscribe(commons.GetTopicFullName("2")))
 }
+
+// TestSubFilter_CanSubscribeBoole guards against regressing #2943: CanSubscribe must
+// recognize Boole-fork topics ("/ssv/<network>/boole/<subnet>"), not just Alan topics.
+// Before the fix, every Boole topic was rejected as "not of the same fork", causing a
+// total post-Boole outage (self-subscribe crash-loop + QBFT broadcast failures).
+func TestSubFilter_CanSubscribeBoole(t *testing.T) {
+	const networkName = "hoodi-stage"
+	l := log.TestLogger(t)
+	sf := newSubFilter(l, 2)
+
+	booleTopic := commons.BooleTopic(networkName, 51)
+
+	// A valid Boole topic is still gated by the whitelist: rejected until registered...
+	require.False(t, sf.CanSubscribe(booleTopic))
+	sf.(Whitelist).Register(booleTopic)
+	require.True(t, sf.CanSubscribe(booleTopic))
+
+	// ...and a different, unregistered Boole subnet stays rejected.
+	require.False(t, sf.CanSubscribe(commons.BooleTopic(networkName, 52)))
+
+	// Alan and Boole topics coexist in the same filter.
+	sf.(Whitelist).Register(commons.GetTopicFullName("1"))
+	require.True(t, sf.CanSubscribe(commons.GetTopicFullName("1")))
+	require.True(t, sf.CanSubscribe(booleTopic))
+
+	// Malformed / foreign-fork topics are rejected outright, even if "whitelisted",
+	// because they don't parse as a known subnet topic.
+	for _, bad := range []string{
+		"/ssv/" + networkName + "/boole/",     // missing subnet
+		"/ssv/" + networkName + "/otherfork/1", // unknown fork segment
+		"/ssv/" + networkName + "/boole/abc",   // non-numeric subnet
+	} {
+		sf.(Whitelist).Register(bad)
+		require.Falsef(t, sf.CanSubscribe(bad), "expected %q to be rejected", bad)
+	}
+}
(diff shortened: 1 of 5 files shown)


momosh-ssv:
### Verified on stage-hoodi (2026-07-13)

Deployed this branch with the Boole fork forced active (`hoodi-stage` `Boole: 0`, test-only) to a fresh committee (ops **57-60**), against Ovidiu's still-broken post-Boole nodes (**1-4**) as a control — identical Boole-active config, the only difference is this `CanSubscribe` fix.

| Check (last 10m) | Fix nodes 57-60 | Control 1-4 (no fix) |
|---|---|---|
| Commit running | `070cc13` (fix + Boole:0) | buggy build |
| `topic is not allowed by the subscription filter` | **0** | 2250 `could not broadcast msg` |
| Startup crash-loop (`could not start node`) | **0** (clean single start) | crash-looping |
| Boole topics (`/ssv/hoodi-stage/boole/*`) | subscribing & scoring ✅ | rejected |
| Duties completed post-Boole | **172 attestations submitted** ✅ | 0 |

Both failure modes from #2943 are resolved: nodes start cleanly (mode a) and complete the full QBFT consensus → broadcast → beacon-submission path on Boole committee topics (mode b). Test deployment has been rolled back and the throwaway `Boole:0` branch deleted.

network/topics/sub_filter_test.go
```diff
@@ -19,3 +19,39 @@ func TestSubFilter(t *testing.T) {
...
+
+	booleTopic := commons.BooleTopic(networkName, 51)
+
+	// A valid Boole topic is still gated by the whitelist: rejected until registered...
+	require.False(t, sf.CanSubscribe(booleTopic))
+	sf.(Whitelist).Register(booleTopic)
+	require.True(t, sf.CanSubscribe(booleTopic))
+
+	// ...and a different, unregistered Boole subnet stays rejected.
+	require.False(t, sf.CanSubscribe(commons.BooleTopic(networkName, 52)))
```
iurii-ssv, line 37: Optional (test coverage): the three cases in the loop below all fail the *parse* gate. The one branch not exercised is a topic that **parses** but is still rejected — a well-formed Boole topic for a *different network*. Adding it documents that network identity is enforced by the whitelist, not by `ParseTopicSubnet`:

```suggestion
	require.False(t, sf.CanSubscribe(commons.BooleTopic(networkName, 52)))

	// A well-formed Boole topic for a *different* network still parses, so it's not
	// rejected at the gate — only the (per-network) whitelist turns it down.
	require.False(t, sf.CanSubscribe(commons.BooleTopic("other-network", 51)))
```

iurii-ssv (commented):
One inline suggestion below (optional test coverage).

Separately — pre-existing and **out of scope** for this fix, flagging for a possible follow-up: in `network/topics/sub_filter.go`, `newSubFilter`'s `logger` parameter and the `subFilter.subsLimit` field are both accepted/stored but never read. `FilterIncomingSubscriptions` enforces the limit via the package-level `subscriptionRequestLimit`, not `subsLimit`:

```go
type subFilter struct {
	whitelist *dynamicWhitelist
	subsLimit int // never read
}

func newSubFilter(logger *zap.Logger, subsLimit int) SubFilter { // logger unused
	return &subFilter{
		whitelist: newWhitelist(),
		subsLimit: subsLimit,
	}
}
```

Both could be dropped to simplify the constructor (touches only the one caller in `pubsub.go`). Not something to change in this PR.

iurii-ssv (approved):
LGTM, aside from the failing linter + minor suggestions

network/topics/sub_filter_test.go
momosh-ssv: Good call — added in aa13e3176, took the suggestion as-is. Seems worth having it documented in the test that network identity is enforced by the whitelist rather than the parse gate.

network/topics/sub_filter_test.go
```diff
@@ -19,3 +19,39 @@ func TestSubFilter(t *testing.T) {
...
+
+	// Alan and Boole topics coexist in the same filter.
+	sf.(Whitelist).Register(commons.GetTopicFullName("1"))
+	require.True(t, sf.CanSubscribe(commons.GetTopicFullName("1")))
+	require.True(t, sf.CanSubscribe(booleTopic))
+
+	// Malformed / foreign-fork topics are rejected outright, even if "whitelisted",
+	// because they don't parse as a known subnet topic.
+	for _, bad := range []string{
```
ovidiu-ssv-labs, line 50: **Non-blocking (test hardening):** this loop covers missing/non-numeric subnet and an unknown-fork segment, but not a *well-formed Boole topic for a different network* — e.g. `/ssv/mainnet/boole/51` while running on `hoodi-stage`. That topic parses successfully via `ParseTopicSubnet` and is rejected only by the whitelist, which is exactly the network-isolation guarantee worth locking in with an explicit assertion. Nice-to-have, not a merge blocker.

ovidiu-ssv-labs (approved):

network/topics/sub_filter_test.go
momosh-ssv: Seems that this landed just before your review; the exact case was added in aa13e3176 off iurii's suggestion above: a well-formed `BooleTopic("other-network", 51)` that parses fine and is turned down only by the whitelist. I fear the review diff was pinned to d135749d3, which predates it. 

Shout if you'd rather also see the literal `/ssv/mainnet/...` spelling covered.

Issue #2915, opened 2026-06-30
bug: PROPOSER and VALIDATOR_REGISTRATION pre-consensus messages use wrong SSV domain after Boole fork


ovidiu-ssv-labs:
## Summary

After the Boole fork activates, `PROPOSER` and `VALIDATOR_REGISTRATION` duties fail silently: all operators sign and broadcast their pre-consensus partial signatures (RANDAO / validator registration), but the messages are rejected by the local pubsub validator before they ever leave the node. Pre-consensus never reaches quorum, and the duty times out.

`COMMITTEE` duties (attestations, sync committee) are **not affected**.

---

## Symptoms

Observed on stage-hoodi, duty `PROPOSER-e105697-s3382317-v839840` (operators 25–28, all on post-fork config):

1. All four operators log `signing and broadcasting randao partial sig`
2. Each operator's own broadcast is immediately rejected:
   ```
   msg="ignoring invalid message" runner_role="PROPOSER" ssv_message_type="partial_signature"
   error="wrong domain, got 00003114, want 00003115"
   ```
3. The publish fails:
   ```
   msg="could not publish p2p message" topic="/ssv/hoodi-stage/boole/24" error="validation ignored"
   ```
4. No operator ever receives a peer's pre-consensus message.
5. Slot ends with `⚠️ duty did not complete before slot end (likely stuck)` on all operators.

---

## Root Cause

`BaseRunner.signAndBroadcastPartialSigMsgs` in `protocol/v2/ssv/runner/runner.go:351` creates the `MsgID` using the **static, fork-unaware** `DomainType`:

```go
msgID := spectypes.NewMsgID(b.NetworkConfig.DomainType, validatorPubKey, b.RunnerRoleType)
```

After the Boole fork activates, the message validator calls `DomainTypeAtSlot(slot)`, which returns `NextDomainType` (`00003115`). The message was stamped with `DomainType` (`00003114`), so the domain check fails and the message is dropped with `ValidationIgnore` — before it is ever propagated to peers.

### Why COMMITTEE is not affected

`CommitteeRunner` has its own broadcast implementation (`committee.go:402`) that correctly uses `DomainTypeAtSlot`:

```go
MsgID: spectypes.NewMsgID(
    r.NetworkConfig.DomainTypeAtSlot(r.State.CurrentDuty.DutySlot()), // fork-aware ✓
    r.QBFTController.CommitteeMember.CommitteeID[:],
    r.RunnerRoleType,
),
```

[...]

PR #2948, opened 2026-07-14
fix(qbft): resolve controller identifier per-height so post-Boole consensus carries the correct domain (#2947)

### LINKED ISSUE #2947: Boole: QBFT controller identifier stamps static pre-Boole domain post-fork — all post-Boole consensus fails
## Summary

Post-Boole, **every QBFT consensus message (proposal / prepare / commit / round-change) is stamped with the pre-Boole SSV domain `0x00003114`** instead of the post-Boole `0x00003115`. The node's own message validator rejects them (`wrong domain, got 00003114, want 00003115`) → the publish is dropped (`could not publish p2p message: validation ignored`) → QBFT never reaches quorum → **all post-Boole consensus duties fail, starting with attestations (every slot).**

**Severity: P0 — fork blocker.** Reproduced at HEAD `310bd7b6` (integration/boole-convergence) on stage-hoodi, 2026-07-13. This surfaced only after the subscription-filter fix (#2943) landed — that bug was masking this one (messages never reached validation before).

## Root cause

The QBFT controller's `Identifier` is built **once at controller construction** from the raw static `NetworkConfig.DomainType`, and is never re-resolved per slot/height. `operator/validator/controller.go:1091` (committee / aggregator-committee) and `:1177` (proposer / aggregator / sync-committee-contribution / voluntary-exit):

```go
identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, options.Operator.CommitteeID[:], role)
qbftCtrl := qbftcontroller.NewController(identifier[:], ...)
```

`DomainType` is `0x00003114` for hoodi-stage (`networkconfig/hoodi-stage.go:12`; `NextDomainType` = `0x00003115` at `:13`). Note it reads the static field directly — not even the fork-aware `CurrentDomainType()`.

That static identifier then flows unchanged into every consensus broadcast:
- `protocol/v2/qbft/controller/controller.go:108-116` — `StartNewInstance(... height ...)` passes the stored `c.Identifier` and **ignores `height`/slot entirely**.
- `protocol/v2/qbft/instance/proposal.go:255` — `CreateProposal` sets `Identifier: i.State.ID` (= `c.Identifier`); same for `round_change.go` / `prepare.go` / `commit.go`.
- `protocol/v2/qbft/instance/instance.go:169-175` — broadcasts via `BroadcastAtSlot(msg, slot)`, but the MsgID's first 4 bytes (the domain) are already `0x00003114`.

A static field cannot be patched in place: a controller created pre-fork must emit the Boole domain post-fork within its own lifetime, so the domain must be derived from the instance height/slot, not frozen at construction.

## Regression (not pre-existing)

An earlier Boole implementation resolved the identifier **per height**, so the domain rotated correctly at the fork boundary:

```go
// prior (working) form
identifierFn := func(height specqbft.Height) []byte {
    msgID := spectypes.NewMsgID(options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(height)), options.Operator.CommitteeID[:], role) // fork-aware
    return msgID[:]
}
qbftCtrl := qbftcontroller.NewController(identifierFn, ...)
```

The controller there held `IdentifierFn func(height specqbft.Height) []byte` and called it in `StartNewInstance` and in the receive-side identifier match. The convergence work made every *other* domain path fork-aware — message validation (#2927), proposer selection, and the pre-consensus partial-sig path (#2915) — but **did not port the per-height QBFT-controller identifier**, keeping the Alan-era static one. That omission is the regression: the prior branch passed a live post-Boole transition; this one cannot.

## Blast radius

**Broken post-Boole (any runner with a QBFT consensus phase — all use the static identifier at `controller.go:1091`/`:1177`):** COMMITTEE (attestation — fails every slot, observed), AGGREGATOR_COMMITTEE, PROPOSER, AGGREGATOR, SYNC_COMMITTEE_CONTRIBUTION, VOLUNTARY_EXIT.

**Not affected:** VALIDATOR_REGISTRATION (no consensus phase), and partial-signature / post-consensus broadcasts — those are already fork-aware via `DomainTypeAtSlot(slot)` (`protocol/v2/ssv/runner/runner.go:354` (fixed in #2915), `committee.go:402-404`, `aggregator_committee.go:752-753,1710-1711`). This is why validator registrations and post-consensus submission still succeed — the break is specific to the **QBFT consensus** identifier. (Any `AGGREGATOR_COMMITTEE` "finishes" observed live are no-op/sparse, not correct consensus — it is equally vulnerable per the code.)

## Validator side is correct — the sender stamps the wrong domain

`message/validation/validation.go:281-290` (`validateDomainAtSlot`) expects `mv.netCfg.DomainTypeAtSlot(slot)`, which returns `NextDomainType` (`0x00003115`) post-Boole (`networkconfig/network.go:36-41`). The receiver is right; the sender is wrong.

## Reproduction (stage-hoodi, 2026-07-13, on a build that already includes the #2943 fix)

- **Live fork transition (committee 33-36)** — clean before/after at fork slot 3479136 / 17:17:12 UTC:

  | | before fork (~8 min) | after fork (~10 min) |
  |---|---|---|
  | `QBFT instance is decided` | 432 | **0** |
  | `successfully submitted attestations` | 160 | **0** |

  First `wrong domain, got 00003114, want 00003115` at 17:17:14.20Z (2.2s after the fork): `COMMITTEE-33_34_35_36 qbft_message_type=proposal qbft_round=1 runner_role=COMMITTEE ssv_message_type=consensus`.
- **Always-post-Boole nodes:** 0 COMMITTEE completions; thousands of wrong-domain hits per 10 min on own-committee `proposal`/`round_change`, driving round-changes to 11+ with no decide.
- **Pre-Boole control:** 0 wrong-domain, attestations flowing — confirms the failure is Boole-specific.
- Distinct from #2943 (subscription filter, fixed) and #2915 (pre-consensus partial sigs).

## Suggested fix

Restore per-height domain resolution in the QBFT controller identifier:
1. Give `protocol/v2/qbft/controller.Controller` an `IdentifierFn func(height specqbft.Height) []byte` instead of the static `Identifier []byte`; have `NewController` accept it; use `c.IdentifierFn(height)` in `StartNewInstance` and in the receive-side identifier match.
2. At `operator/validator/controller.go:1091` and `:1177`, build it fork-aware:
   ```go
   identifierFn := func(h specqbft.Height) []byte {
       id := spectypes.NewMsgID(options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(h)), <committeeID|pubkey>, role)
       return id[:]
   }
   ```

**Test gap:** existing fork-aware tests only exercise the **receiver** (`message/validation`). Add a **sender**-side test: a controller spanning the fork boundary must produce, at a post-Boole height, a QBFT MsgID whose domain equals `DomainTypeAtSlot(slot)` and passes `validateDomainAtSlot`.

momosh-ssv:
## What

Fixes #2947 — post-Boole, every QBFT consensus message (proposal/prepare/commit/round-change) was stamped with the **pre-Boole** SSV domain instead of the post-Boole one. The node's own message validator rejected them (`wrong domain, got 00003114, want 00003115`) → publish dropped → QBFT never reached quorum → **all post-Boole consensus duties failed**. This surfaced only after the subscription-filter fix (#2944) landed — that bug was masking this one.

## Root cause

The QBFT controller's identifier was built **once at construction** from the static `NetworkConfig.DomainType` and frozen for the controller's whole lifetime, which spans the fork boundary:

- `operator/validator/controller.go` (committee + proposer `buildController`): `spectypes.NewMsgID(options.NetworkConfig.DomainType, …)` — the static field, not the fork-aware `DomainTypeAtSlot`.
- `protocol/v2/qbft/controller/controller.go`: `Identifier []byte` (static), passed unchanged into `StartNewInstance` → `instance.NewInstance` (which ignored the in-scope `height`) and used verbatim in the `ProcessMsg` receive-side match.

The receiver (`message/validation` `validateDomainAtSlot`) correctly expects `DomainTypeAtSlot(slot)`. The convergence work made message-validation, proposer-selection, and the pre-consensus partial-sig path (#2915) fork-aware, but did **not** port the per-height QBFT-controller identifier. That omission is the regression.

## Fix

Restore per-height domain resolution. Inject `IdentifierFn func(height specqbft.Height) []byte` that resolves via `DomainTypeAtSlot(phase0.Slot(height))`, and use it on:
- **send** — `StartNewInstance` (`c.identifierAtHeight(height)`),
- **receive** — `ProcessMsg` identifier match (`c.identifierAtHeight(msg.QBFTMessage.Height)`),
- **decided fast-forward** — `UponDecided` (`inst == nil` branch).

`NewController`'s signature is unchanged — `IdentifierFn` is set as a field after construction, so spec-test shims compile as-is. The static `Identifier` is kept as a serialized/diagnostic fallback (tagged `json:"-"`, so `Encode`/`GetRoot` round-trips are unaffected). When `IdentifierFn` is nil, behavior is byte-identical to before.

`height == slot` for all consensus runners (single `StartNewInstance` caller in `BaseRunner.StartNewConsensus`, passing `specqbft.Height(slot)`).

## Tests

New `protocol/v2/qbft/controller/controller_fork_test.go`:
- **send**: a fork-spanning controller yields the pre-Boole domain at pre-fork heights and the post-Boole domain at post-fork heights; the assertion fails against the unpatched (frozen-identifier) code.
- **receive**: `ProcessMsg` accepts a message whose domain matches its height on each side of the fork, and rejects one whose domain doesn't match its height.
- **fallback**: nil `IdentifierFn` is byte-identical to the static `Identifier` across heights.

`go build ./...`, `go vet`, and `go test ./protocol/v2/qbft/... ./operator/validator/... ./message/validation/... ./protocol/v2/ssv/spectest/... ./protocol/v2/ssv/runner/...` all pass.

## Verification (ssv-mini, Boole fork active, distinct domains 0x00000000 → 0x00000001 at epoch 3)

Drove a live 4-node committee across the fork:
- **`wrong domain` rejections post-fork: 0** (was thousands on the unpatched build).
- **COMMITTEE (attestation) consensus decides at round 1**; attestations submitted post-fork.
- **Subscription-filter errors: 0** (confirms #2944 too).

## Note / follow-up (not addressed here)

During verification, `AGGREGATOR_COMMITTEE` consensus was observed to never decide before **round 5** (with recurring `could not publish … validation ignored` on the Boole topic), though aggregates still submit. This is independent of the domain fix (0 `wrong domain`) and of #2939/#2940 (both merged). Flagging for a separate investigation/issue.

### DIFF
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1088,8 +1088,15 @@ func SetupCommitteeRunners(
 			CutOffRound: roundtimer.CutOffRound,
 		}
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, options.Operator.CommitteeID[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		committeeID := options.Operator.CommitteeID
+		identifierFn := func(height specqbft.Height) []byte {
+			domain := options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(height))
+			id := spectypes.NewMsgID(domain, committeeID[:], role)
+			return id[:]
+		}
+		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
+		qbftCtrl := qbftcontroller.NewController(identifier, options.Operator, config, options.OperatorSigner, options.FullNode)
+		qbftCtrl.IdentifierFn = identifierFn
 		return qbftCtrl
 	}
 
@@ -1174,8 +1181,15 @@ func SetupRunners(
 			CutOffRound: roundtimer.CutOffRound,
 		}
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, share.ValidatorPubKey[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], operator, config, options.OperatorSigner, options.FullNode)
+		validatorPubKey := share.ValidatorPubKey
+		identifierFn := func(height specqbft.Height) []byte {
+			domain := options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(height))
+			id := spectypes.NewMsgID(domain, validatorPubKey[:], role)
+			return id[:]
+		}
+		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
+		qbftCtrl := qbftcontroller.NewController(identifier, operator, config, options.OperatorSigner, options.FullNode)
+		qbftCtrl.IdentifierFn = identifierFn
 		return qbftCtrl
 	}
 
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1061,6 +1061,16 @@ func (c *Controller) ReportValidatorStatuses(ctx context.Context) {
 	}
 }
 
+// newIdentifierFn returns a per-height QBFT identifier resolver for the given executor
+// (committee ID or validator public key), carrying the fork-correct SSV domain for the
+// height's slot.
+func newIdentifierFn(cfg *networkconfig.Network, executorID []byte, role spectypes.RunnerRole) func(specqbft.Height) []byte {
+	return func(height specqbft.Height) []byte {
+		id := spectypes.NewMsgID(cfg.DomainTypeAtSlot(phase0.Slot(height)), executorID, role)
+		return id[:]
+	}
+}
+
 func SetupCommitteeRunners(
 	ctx context.Context,
 	options *validator.Options,
@@ -1089,11 +1099,7 @@ func SetupCommitteeRunners(
 		}
 
 		committeeID := options.Operator.CommitteeID
-		identifierFn := func(height specqbft.Height) []byte {
-			domain := options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(height))
-			id := spectypes.NewMsgID(domain, committeeID[:], role)
-			return id[:]
-		}
+		identifierFn := newIdentifierFn(options.NetworkConfig, committeeID[:], role)
 		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
 		qbftCtrl := qbftcontroller.NewController(identifier, options.Operator, config, options.OperatorSigner, options.FullNode)
 		qbftCtrl.IdentifierFn = identifierFn
@@ -1182,11 +1188,7 @@ func SetupRunners(
 		}
 
 		validatorPubKey := share.ValidatorPubKey
-		identifierFn := func(height specqbft.Height) []byte {
-			domain := options.NetworkConfig.DomainTypeAtSlot(phase0.Slot(height))
-			id := spectypes.NewMsgID(domain, validatorPubKey[:], role)
-			return id[:]
-		}
+		identifierFn := newIdentifierFn(options.NetworkConfig, validatorPubKey[:], role)
 		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
 		qbftCtrl := qbftcontroller.NewController(identifier, operator, config, options.OperatorSigner, options.FullNode)
 		qbftCtrl.IdentifierFn = identifierFn
--- a/operator/validator/controller.go
+++ b/operator/validator/controller.go
@@ -1100,7 +1100,7 @@ func SetupCommitteeRunners(
 
 		committeeID := options.Operator.CommitteeID
 		identifierFn := newIdentifierFn(options.NetworkConfig, committeeID[:], role)
-		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
+		identifier := identifierFn(specqbft.FirstHeight) // domain at slot 0; used as Controller.Identifier (diagnostics only, not persisted)
 		qbftCtrl := qbftcontroller.NewController(identifier, options.Operator, config, options.OperatorSigner, options.FullNode)
 		qbftCtrl.IdentifierFn = identifierFn
 		return qbftCtrl
@@ -1189,7 +1189,7 @@ func SetupRunners(
 
 		validatorPubKey := share.ValidatorPubKey
 		identifierFn := newIdentifierFn(options.NetworkConfig, validatorPubKey[:], role)
-		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
+		identifier := identifierFn(specqbft.FirstHeight) // domain at slot 0; used as Controller.Identifier (diagnostics only, not persisted)
 		qbftCtrl := qbftcontroller.NewController(identifier, operator, config, options.OperatorSigner, options.FullNode)
 		qbftCtrl.IdentifierFn = identifierFn
 		return qbftCtrl
--- a/protocol/v2/qbft/controller/controller.go
+++ b/protocol/v2/qbft/controller/controller.go
@@ -34,7 +34,7 @@ type Controller struct {
 	// IdentifierFn, when non-nil, resolves the QBFT message identifier at a given height.
 	// It is used on both send (StartNewInstance) and receive (ProcessMsg) paths to produce
 	// the fork-correct SSV domain for the height/slot. When nil, falls back to Identifier.
-	// JSON-tagged as "-" because funcs are not JSON-serialisable; Controller is only marshalled
+	// JSON-tagged as "-" because funcs are not JSON-serialisable; Controller is only marshaled
 	// in spec-test fixtures and is never persisted and rehydrated in production.
 	IdentifierFn func(height specqbft.Height) []byte `json:"-"`
 
--- a/protocol/v2/qbft/controller/controller.go
+++ b/protocol/v2/qbft/controller/controller.go
@@ -31,6 +31,13 @@ type NewDecidedHandler func(msg qbftstorage.Participation)
 type Controller struct {
 	Identifier []byte
 
+	// IdentifierFn, when non-nil, resolves the QBFT message identifier at a given height.
+	// It is used on both send (StartNewInstance) and receive (ProcessMsg) paths to produce
+	// the fork-correct SSV domain for the height/slot. When nil, falls back to Identifier.
+	// JSON-tagged as "-" because funcs are not JSON-serialisable; Controller is only marshalled
+	// in spec-test fixtures and is never persisted and rehydrated in production.
+	IdentifierFn func(height specqbft.Height) []byte `json:"-"`
+
 	// LatestInstanceHeight is the height of the latest Instance Controller spun up. That latest instance
 	// is the "relevant" one currently driving consensus.
 	// JSON-tagged as "Height" to preserve compatibility with existing spec-test fixtures.
@@ -45,6 +52,16 @@ type Controller struct {
 	NewDecidedHandler NewDecidedHandler       `json:"-"`
 	config            qbft.IConfig
 	fullNode          bool
+}
+
+// identifierAtHeight returns the QBFT message identifier for the given height.
+// When IdentifierFn is set it delegates to it (producing the fork-correct domain);
+// otherwise it falls back to the static Identifier field.
+func (c *Controller) identifierAtHeight(height specqbft.Height) []byte {
+	if c.IdentifierFn != nil {
+		return c.IdentifierFn(height)
+	}
+	return c.Identifier
 }
 
 func NewController(
@@ -110,7 +127,7 @@ func (c *Controller) StartNewInstance(
 		logger,
 		c.GetConfig(),
 		c.CommitteeMember,
-		c.Identifier,
+		c.identifierAtHeight(height),
 		height,
 		c.OperatorSigner,
 		roundTimerF,
@@ -143,7 +160,7 @@ func (c *Controller) ProcessMsg(
 		return nil, fmt.Errorf("could not create ProcessingMessage from signed message: %w", err)
 	}
 
-	if !bytes.Equal(c.Identifier, msg.QBFTMessage.Identifier) {
+	if !bytes.Equal(c.identifierAtHeight(msg.QBFTMessage.Height), msg.QBFTMessage.Identifier) {
 		return nil, spectypes.NewError(spectypes.MessageIdentifierInvalidErrorCode, "message doesn't belong to Identifier")
 	}
 
--- a/protocol/v2/qbft/controller/controller.go
+++ b/protocol/v2/qbft/controller/controller.go
@@ -29,6 +29,10 @@ type NewDecidedHandler func(msg qbftstorage.Participation)
 
 // Controller is a QBFT coordinator responsible for starting and following the entire life cycle of multiple QBFT Instances
 type Controller struct {
+	// Identifier is the static QBFT message identifier fixed at construction time, carrying
+	// the domain at slot 0. When IdentifierFn is set, live traffic switches domains per
+	// height, so this field no longer matches what's on the wire post-fork — resolve via
+	// identifierAtHeight instead of reading this field (or GetIdentifier) directly.
 	Identifier []byte
 
 	// IdentifierFn, when non-nil, resolves the QBFT message identifier at a given height.
@@ -205,7 +209,9 @@ func (c *Controller) UponExistingInstanceMsg(ctx context.Context, logger *zap.Lo
 	return decidedMsg, nil
 }
 
-// GetIdentifier returns QBFT Identifier, used to identify messages
+// GetIdentifier returns the static QBFT Identifier fixed at construction time (domain at
+// slot 0). It does not reflect per-height fork domains — use identifierAtHeight for the
+// identifier actually carried on the wire at a given height.
 func (c *Controller) GetIdentifier() []byte {
 	return c.Identifier
 }
--- a/protocol/v2/qbft/controller/controller_fork_test.go
+++ b/protocol/v2/qbft/controller/controller_fork_test.go
@@ -103,7 +103,7 @@ func TestController_IdentifierAtHeight(t *testing.T) {
 		ctrl := NewController(identifier, member, nil, nil, false)
 		ctrl.IdentifierFn = identifierFn
 
-		slotsPerEpoch := phase0.Slot(networkconfig.TestNetwork.Beacon.SlotsPerEpoch)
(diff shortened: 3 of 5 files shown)


protocol/v2/qbft/controller/controller_fork_test.go
```diff
@@ -0,0 +1,297 @@
...
+	member := spectestingutils.TestingCommitteeMember(ks)
+	role := spectypes.RoleCommittee
+
+	// Build a post-Boole config: Boole active from genesis (epoch 0).
+	postBooleSSV := *networkconfig.TestNetwork.SSV
+	postBooleSSV.Forks = networkconfig.SSVForks{Boole: 0}
+	postBooleCfg := &networkconfig.Network{Beacon: networkconfig.TestNetwork.Beacon, SSV: &postBooleSSV}
+
+	// Build a pre-Boole config: Boole disabled (epoch MaxUint64).
+	preBooleCfg := networkconfig.TestNetwork // Boole = MaxUint64 by default
```
iurii-ssv: This `preBooleCfg` depends on the package-global `TestNetwork` still having `Boole == MaxUint64`, but `networkconfig/test-network.go`'s `init()` flips that global to `Boole = 0` when `SSV_TEST_BOOLE_FORK=post`. In that mode the assertion below (`require.Equal(t, preBooleCfg.DomainType[:], msgID.GetDomain(), ...)`) fails: `DomainTypeAtSlot` returns `NextDomainType`, while `DomainType` stays Alan.

Latent today (no CI job sets the var), but that env switch exists specifically for a post-fork test matrix, and every other config in this file (`postBooleCfg`, `midBooleCfg`) is built explicitly. Suggest making this one explicit too so it can't be mutated out from under the test:

```go
preBooleSSV := *networkconfig.TestNetwork.SSV
preBooleSSV.Forks = networkconfig.SSVForks{Boole: phase0.Epoch(math.MaxUint64)}
preBooleCfg := &networkconfig.Network{Beacon: networkconfig.TestNetwork.Beacon, SSV: &preBooleSSV}
```

(add `"math"` to the imports). Note the `preBooleCfg` in `TestController_NilIdentifierFn_...` is already safe — it reads the `DomainType` field directly rather than going through `DomainTypeAtSlot`.

operator/validator/controller.go
```diff
@@ -1088,8 +1088,15 @@ func SetupCommitteeRunners(
 			CutOffRound: roundtimer.CutOffRound,
 		}
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, options.Operator.CommitteeID[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		committeeID := options.Operator.CommitteeID
+		identifierFn := func(height specqbft.Height) []byte {
```
iurii-ssv: Non-blocking cleanup: this `identifierFn` closure is duplicated near-verbatim in `SetupRunners` (~L1185), differing only in the executor ID (`CommitteeID` here vs `ValidatorPubKey` there). Could be factored into a small shared helper:

```go
func newIdentifierFn(cfg *networkconfig.Network, executorID []byte, role spectypes.RunnerRole) func(specqbft.Height) []byte {
	return func(height specqbft.Height) []byte {
		id := spectypes.NewMsgID(cfg.DomainTypeAtSlot(phase0.Slot(height)), executorID, role)
		return id[:]
	}
}
```

The defensive local copies of `committeeID` / `validatorPubKey` are good and carry over (pass `committeeID[:]`).

protocol/v2/qbft/controller/controller_fork_test.go
```diff
@@ -0,0 +1,297 @@
...
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+	spectestingutils "github.com/ssvlabs/ssv-spec/types/testingutils"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/ssvlabs/ssv/networkconfig"
+)
+
+// TestController_IdentifierAtHeight verifies that the per-height identifier resolves to the
+// fork-correct SSV domain. Pre-Boole heights must yield DomainType (Alan, 0x...3114) and
```
iurii-ssv: Nit: these literal domains (`0x...3114` / `0x...3115`) are the real-network values; `TestNetwork` actually uses `{0,0,JatoNetworkID,0x2}` / `{0,0,JatoNetworkID,0x3}`. The assertions compare against `cfg.DomainType` / `cfg.NextDomainType`, so they're correct regardless — but the literals in the prose can mislead. Consider dropping them or citing the TestNetwork values.

ovidiu-ssv-labs:
**Important — the send-path regression itself has no real-code-path test.**

The new `controller_fork_test.go` exercises the receive path for real (`TestController_ProcessMsg_ForkDomainCheck` drives `ProcessMsg` and asserts both accept and `MessageIdentifierInvalidErrorCode` reject across the fork). But the **send** path — `StartNewInstance` (`controller.go`) and the `UponDecided` `inst == nil` branch (`decided.go`), which is what the regression actually broke — is only covered through the `identifierAtHeight` helper in isolation (`TestController_IdentifierAtHeight`).

Concretely: if someone reverted `c.identifierAtHeight(height)` back to `c.Identifier` in `StartNewInstance` (or in `UponDecided`) but left the helper intact, **every test in the file would still pass** — the exact regression this PR fixes would slip straight through.

Suggest adding a test that calls `StartNewInstance` at a post-fork height and asserts the resulting `instance.State.ID` (or an outgoing message's `Identifier`) carries the Boole domain, closing the coverage gap on the code the PR exists to fix.

And linting.

iurii-ssv (commented):
Looks correct, just some small improvements to consider + linter fails ATM

ovidiu-ssv-labs (changes requested):
Need to consider whether the test I mentioned in the comment above needs some fixing.

protocol/v2/qbft/controller/controller_fork_test.go
momosh-ssv: Good catch — fixed in 7a9a46006, `preBooleCfg` is now built from an explicit copy of `TestNetwork.SSV` with `Boole` pinned to `MaxUint64`, so the env switch can't mutate it out from under the test.

operator/validator/controller.go
momosh-ssv: Fixed in beb1851d4 — factored into a shared `newIdentifierFn` helper as suggested, both call sites now use it (the defensive local copies carry over via `committeeID[:]` / `validatorPubKey[:]`).

protocol/v2/qbft/controller/controller_fork_test.go
momosh-ssv: Fixed in 60f1a629a — dropped the real-network literals from the comment so it can't mislead against the TestNetwork values.

momosh-ssv:
> Need to consider whether the test I mentioned in the comment above needs some fixing.

Fixed in 95e7c7d4b.
Added `TestController_StartNewInstance_ForkDomain` and `TestController_UponDecided_ForkDomain`, which drive the real send paths and assert the resulting `instance.State.ID` carries the Boole domain post-fork (and Alan pre-fork), with an explicit assertion that it does not fall back to the static controller Identifier. Reverting either call site to `c.Identifier` now fails these tests. Lint is green as of 03129c05a.

operator/validator/controller.go
```diff
@@ -1088,8 +1098,11 @@ func SetupCommitteeRunners(
 			CutOffRound: roundtimer.CutOffRound,
 		}
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, options.Operator.CommitteeID[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], options.Operator, config, options.OperatorSigner, options.FullNode)
+		committeeID := options.Operator.CommitteeID
+		identifierFn := newIdentifierFn(options.NetworkConfig, committeeID[:], role)
+		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
```
iurii-ssv: Nit: “pre-fork domain” isn’t quite accurate — `identifierFn(specqbft.FirstHeight)` yields the domain *at slot 0*, which on a Boole-at-genesis network (ssv-mini, or `SSV_TEST_BOOLE_FORK=post` runs) is the *post*-fork domain. The “diagnostics only” part is the point; suggest wording it by what it is:

```suggestion
		identifier := identifierFn(specqbft.FirstHeight) // domain at slot 0; used as Controller.Identifier (diagnostics only, not persisted)
```

operator/validator/controller.go
```diff
@@ -1174,8 +1187,11 @@ func SetupRunners(
 			CutOffRound: roundtimer.CutOffRound,
 		}
 
-		identifier := spectypes.NewMsgID(options.NetworkConfig.DomainType, share.ValidatorPubKey[:], role)
-		qbftCtrl := qbftcontroller.NewController(identifier[:], operator, config, options.OperatorSigner, options.FullNode)
+		validatorPubKey := share.ValidatorPubKey
+		identifierFn := newIdentifierFn(options.NetworkConfig, validatorPubKey[:], role)
+		identifier := identifierFn(specqbft.FirstHeight) // pre-fork domain; used as Controller.Identifier (diagnostics only, not persisted)
```
iurii-ssv: Same nit as the committee variant above:

```suggestion
		identifier := identifierFn(specqbft.FirstHeight) // domain at slot 0; used as Controller.Identifier (diagnostics only, not persisted)
```

operator/validator/controller.go
```diff
@@ -1061,6 +1061,16 @@ func (c *Controller) ReportValidatorStatuses(ctx context.Context) {
 	}
 }
 
+// newIdentifierFn returns a per-height QBFT identifier resolver for the given executor
+// (committee ID or validator public key), carrying the fork-correct SSV domain for the
+// height's slot.
+func newIdentifierFn(cfg *networkconfig.Network, executorID []byte, role spectypes.RunnerRole) func(specqbft.Height) []byte {
```
iurii-ssv, line 1067: Extending @ovidiu-ssv-labs’s earlier coverage point: the send-path tests added since close the gap *inside the controller package*, but this operator-level wiring is still untested — the controller tests build their own in-package closures mirroring this helper, so nothing asserts that `SetupCommitteeRunners`/`SetupRunners` actually attach an `IdentifierFn` that flips domains at the fork. Concretely: deleting either `qbftCtrl.IdentifierFn = identifierFn` assignment (or having this helper use the static `DomainType`) still passes the entire suite.

Non-blocking, but a small test in `operator/validator` — either directly on `newIdentifierFn` with a fork-spanning config, or asserting the built controller’s `IdentifierFn` resolves pre/post-fork domains — would close the last uncovered link in the chain.

protocol/v2/qbft/controller/controller.go
```diff
@@ -31,6 +31,13 @@ type NewDecidedHandler func(msg qbftstorage.Participation)
 type Controller struct {
 	Identifier []byte
 
+	// IdentifierFn, when non-nil, resolves the QBFT message identifier at a given height.
+	// It is used on both send (StartNewInstance) and receive (ProcessMsg) paths to produce
+	// the fork-correct SSV domain for the height/slot. When nil, falls back to Identifier.
+	// JSON-tagged as "-" because funcs are not JSON-serialisable; Controller is only marshaled
+	// in spec-test fixtures and is never persisted and rehydrated in production.
+	IdentifierFn func(height specqbft.Height) []byte `json:"-"`
```
iurii-ssv, line 43: Nit (future-proofing): once `IdentifierFn` is set, the static `Identifier` permanently holds the slot-0 domain while live traffic switches domains per height — so any post-fork reader of `Identifier`/`GetIdentifier()` (today only the spectest runner fixture) gets a value that no longer matches what’s on the wire. Worth a sentence on the `Identifier` field or `GetIdentifier()` warning callers off it and pointing at `identifierAtHeight`, so the accessor can’t quietly reintroduce this exact bug later.

protocol/v2/qbft/controller/controller_fork_test.go
```diff
@@ -0,0 +1,464 @@
...
+)
+
+// TestController_IdentifierAtHeight verifies that the per-height identifier resolves to the
+// fork-correct SSV domain. Pre-Boole heights must yield the config's DomainType (Alan) and
+// post-Boole heights must yield its NextDomainType (Boole).
+//
+// The test also confirms the regression: without IdentifierFn the controller uses the static
+// Identifier frozen at construction time, causing post-fork heights to still carry the pre-fork
+// domain.
+func TestController_IdentifierAtHeight(t *testing.T) {
```
iurii-ssv, line 29: Non-blocking: most of this function now duplicates coverage that lives elsewhere. The first three subtests exercise a *locally defined* closure (mirroring `newIdentifierFn`) plus the trivial `identifierAtHeight` delegation, and the two nil-fallback subtests duplicate `TestController_NilIdentifierFn_ByteIdenticalToStatic`. The load-bearing guards are `TestController_StartNewInstance_ForkDomain`, `TestController_ProcessMsg_ForkDomainCheck` and `TestController_UponDecided_ForkDomain`, which drive the real code paths. This could shrink to just the fork-boundary subtest (or go entirely) — ~150 fewer lines, same protection.

iurii-ssv (commented):
Second pass over the latest revision (send-path tests + `newIdentifierFn` factoring). Inline comments below are all minor.

operator/validator/controller.go
momosh-ssv: Fixed in 104433e1b — took your wording; the comment now says "domain at slot 0" instead of "pre-fork domain".

operator/validator/controller.go
momosh-ssv: Fixed in 104433e1b — same wording applied to the validator variant.

operator/validator/controller.go
momosh-ssv: Fixed in 19be1d0f5 — added `TestNewIdentifierFn_ForkDomain` in `operator/validator`, driving `newIdentifierFn` with a fork-spanning config (Boole at epoch 10) for both committee and validator executors, asserting the produced identifiers match `NewMsgID` with `DomainType` before the fork slot and `NextDomainType` at it. That closes the "helper resolves the wrong domain" half at the operator level; asserting the `qbftCtrl.IdentifierFn` assignment itself survives would mean driving `SetupCommitteeRunners`/`SetupRunners` end-to-end — happy to take that as a follow-up if you think it's worth the scaffolding.

protocol/v2/qbft/controller/controller.go
momosh-ssv: Fixed in 104433e1b — added doc comments on both the `Identifier` field and `GetIdentifier` stating they hold the slot-0 domain fixed at construction and pointing readers at `identifierAtHeight`.

protocol/v2/qbft/controller/controller_fork_test.go
momosh-ssv: Fixed in b8b41dd7c — shrunk it to just the fork-boundary check (now reusing the shared `forkTestHarness`) and dropped the closure-only and duplicate nil-fallback subtests; −131 lines, same protection.

ovidiu-ssv-labs (approved):

iurii-ssv (approved):
Good job!

PR #2949, opened 2026-07-14
fix(roundtimer): give RoleAggregatorCommittee a 2/3-slot head start

momosh-ssv:
## What

The merged aggregator-committee runner (`RoleAggregatorCommittee`, new in the Boole convergence) never starts consensus at round 1 — it consistently starts at **round ~5** every slot, wasting the entire quick-round budget and generating round-change churn (`could not publish … validation ignored` on the committee topic).

## Root cause

`round1HeadStart` (`protocol/v2/qbft/roundtimer/timer.go`) grants a per-role head start on the slot-relative round timer:

- `RoleCommittee` → 1/3 slot (block available)
- `RoleAggregator`, `RoleSyncCommitteeContribution` → 2/3 slot (attestations available to aggregate)
- everything else → `default: 0`

`RoleAggregatorCommittee` has **no case**, so it falls to `0`. But the aggregator-committee aggregates attestations, so its consensus can only start ~2/3 into the slot. Both the round timer (`RoundTimeout`) and the message validator (`EstimatedRoundAt`, `message/validation/consensus_validation.go`) derive from `round1HeadStart`, so at ~2/3-slot they compute:

```
EstimatedRoundAt(2/3-slot) = FirstRound + (2/3-slot - 0) / QuickTimeout = round 5   (12s slot, 2s quick)
```

i.e. consensus starts already at round 5 every slot. The legacy `RoleAggregator` got a 2/3-slot head start and started at round 1; the merged role lost it.

## Fix

Add `RoleAggregatorCommittee` to the 2/3-slot branch of `round1HeadStart`, mirroring the legacy aggregator. One-line change; both the timer and the validator pick it up (single source of truth).

## Tests

`protocol/v2/qbft/roundtimer/timer_test.go`: added `RoleAggregatorCommittee` cases to `TestRoundTimeoutOffset` (round 1 → `8s + quick`, round 8 → `8s + quickPhase`), `TestEstimatedRoundAt` (keeps round 1 at 2/3-slot), and the boundary/role sweeps. Confirmed the new offset assertions **fail against the unpatched code** (round 1: expected 10s, got 2s).

`go build ./...`, `go test ./protocol/v2/qbft/roundtimer/... ./message/validation/... ./protocol/v2/ssv/validator/... ./protocol/v2/ssv/runner/...` all pass.

## Verification (ssv-mini, Boole fork active)

Live 4-node committee, hot-swapped to a build carrying this fix (on top of the #2947 domain fix, needed for post-Boole consensus at all):

| | before fix | after fix |
|---|---|---|
| `AGGREGATOR_COMMITTEE` decide round | **5** (every slot) | **1** (every slot, e22 s715–722) |
| `validation ignored` publish errors | ~1 / slot | ~0 |
| `COMMITTEE` (attestation) decide round | 1 | 1 (unchanged) |

## Relationship to other fixes

Independent of #2947 (QBFT controller fork domain) — that fix makes post-Boole consensus *possible*; this one makes the aggregator-committee's consensus *efficient*. Both target `integration/boole-convergence`.

### DIFF
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -132,9 +132,9 @@ func New(ctx context.Context, beaconConfig *networkconfig.Beacon, role spectypes
 // RoundTimeout calculates the timeout duration for a specific role, height, and round.
 //
 // Timeout Rules:
-// - For roles BNRoleAttester and BNRoleSyncCommittee, the base timeout is 1/3 of the slot duration.
-// - For roles BNRoleAggregator and BNRoleSyncCommitteeContribution, the base timeout is 2/3 of the slot duration.
-// - For role BNRoleProposer, the timeout is either quickTimeout or slowTimeout, depending on the round.
+// - For RoleCommittee, the base timeout (Round 1 head start) is 1/3 of the slot duration.
+// - For RoleAggregator, RoleSyncCommitteeContribution and RoleAggregatorCommittee, it is 2/3 of the slot duration.
+// - For RoleProposer, the timeout is either quickTimeout or slowTimeout, depending on the round.
 //
 // Additional Timeout:
 // - For rounds less than or equal to quickThreshold, the additional timeout is 'quick' seconds.
--- a/protocol/v2/qbft/roundtimer/timer.go
+++ b/protocol/v2/qbft/roundtimer/timer.go
@@ -48,8 +48,9 @@ func roundTimeoutForRound(role spectypes.RunnerRole, slotDuration time.Duration,
 
 // round1HeadStart returns the extra time, on top of Round 1's normal quick timeout, that
 // Round 1 is allowed to run for a given role. Committee gets 1/3 of the slot as head start
-// (time for the block to become available); aggregator and sync-committee-contribution get
-// 2/3 of the slot (time for attestations to arrive); proposer gets zero.
+// (time for the block to become available); aggregator, aggregator-committee and
+// sync-committee-contribution get 2/3 of the slot (time for attestations to arrive before
+// aggregating); proposer gets zero.
 //
 // Note: this is NOT the time at which Round 1 -> Round 2 transitions — that transition actually
 // happens at `slotStart + round1HeadStart + QuickTimeout`, because Round 1 still needs to run its
@@ -58,7 +59,7 @@ func round1HeadStart(role spectypes.RunnerRole, slotDuration time.Duration) time
 	switch role {
 	case spectypes.RoleCommittee:
 		return slotDuration / 3
-	case ssvtypes.RoleAggregator, ssvtypes.RoleSyncCommitteeContribution:
+	case ssvtypes.RoleAggregator, ssvtypes.RoleSyncCommitteeContribution, spectypes.RoleAggregatorCommittee:
 		return slotDuration / 3 * 2
 	default:
 		return 0
--- a/protocol/v2/qbft/roundtimer/timer_test.go
+++ b/protocol/v2/qbft/roundtimer/timer_test.go
@@ -123,9 +123,10 @@ func TestEstimatedRoundAt(t *testing.T) {
 			want:         specqbft.FirstRound,
 		},
 		{
-			// Regression for the missing RoleAggregatorCommittee head start: with head start 0
-			// this returned round 5 at two-thirds into the slot (when aggregation data arrives),
-			// starting aggregator-committee consensus mid-round instead of at round 1.
+			// Basic behavioral parity with the aggregator case above. NOTE: at this 600ms slot the
+			// head-start difference (0 vs 400ms) is below one QuickTimeout (2s), so this case alone
+			// does NOT discriminate patched vs unpatched — see the realistic-slot regression assertion
+			// after the loop (and TestRoundTimeoutOffset) for the actual guard.
 			name:         "aggregator-committee keeps first round until two-third slot delay passes",
 			role:         spectypes.RoleAggregatorCommittee,
 			timeIntoSlot: slotDuration / 3 * 2,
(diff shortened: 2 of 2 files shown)


protocol/v2/qbft/roundtimer/timer_test.go
```diff
@@ -121,6 +122,15 @@ func TestEstimatedRoundAt(t *testing.T) {
...
 		},
+		{
+			// Regression for the missing RoleAggregatorCommittee head start: with head start 0
+			// this returned round 5 at two-thirds into the slot (when aggregation data arrives),
+			// starting aggregator-committee consensus mid-round instead of at round 1.
+			name:         "aggregator-committee keeps first round until two-third slot delay passes",
+			role:         spectypes.RoleAggregatorCommittee,
+			timeIntoSlot: slotDuration / 3 * 2,
+			want:         specqbft.FirstRound,
+		},
```
ovidiu-ssv-labs, line 134: This case doesn't exercise the regression it documents, and passes with **and** without the fix.

With `slotDuration = 600ms`, `timeIntoSlot = slotDuration/3*2 = 400ms` and `QuickTimeout = 2s`, `EstimatedRoundAt` returns round 1 either way:
- unpatched (head start 0): `1 + floor(400ms / 2s)` = **1**
- patched (head start 400ms): `1 + floor(0 / 2s)` = **1**

So it's non-discriminating. The comment also claims *"with head start 0 this returned round 5 at two-thirds into the slot"* — that only holds for a **12s** slot; at 600ms it never reaches round 5. The `round 5` narrative is misleading for this case's own parameters.

Suggestion: rewrite this case with a realistic 12s slot (like `TestRoundTimeoutOffset` does) so it actually returns round 5 unpatched / round 1 patched, or drop the `round 5` comment and note that `TestRoundTimeoutOffset` is the real regression guard. (The pre-existing aggregator case just above shares the 600ms non-discriminating property, but doesn't make the false `round 5` claim.)

protocol/v2/qbft/roundtimer/timer_test.go
```diff
@@ -208,6 +222,7 @@ func TestEstimatedRoundAtBoundaries(t *testing.T) {
 		{"committee", spectypes.RoleCommittee},
 		{"aggregator", ssvtypes.RoleAggregator},
 		{"sync_committee_contribution", ssvtypes.RoleSyncCommitteeContribution},
+		{"aggregator_committee", spectypes.RoleAggregatorCommittee},
```
ovidiu-ssv-labs, line 238: Minor: this and the sibling table in `TestEstimatedRoundAtMatchesRoundTimeout` (line 350) are internal-consistency checks — both the expected offset (via `roundTimeoutForRound`) and the actual (`EstimatedRoundAt`) derive from the same `round1HeadStart`, so they'd pass even if the head-start value were wrong (both wrong consistently). Good as invariant guards for the invert-arithmetic relationship, but worth not over-crediting as regression coverage for this bug — the actual guard is `TestRoundTimeoutOffset`.

ovidiu-ssv-labs (commented):
Two test-hygiene notes on the round-timer coverage. The production fix itself looks correct and single-source-of-truth (both `RoundTimeout` and `EstimatedRoundAt` derive from `round1HeadStart`), and `TestRoundTimeoutOffset` genuinely guards it (fails on the unpatched code). The comments below are only about the `EstimatedRoundAt` test cases.

protocol/v2/qbft/roundtimer/timer_test.go
momosh-ssv: Good catch, you're right, that case was non-discriminating and the round-5 comment was false at a 600ms slot. 

Fixed in f14acca: reworded that table case as a plain behavioral check (with a note that the sub-QuickTimeout head-start delta is invisible there), and added a standalone assertion at a realistic 12s slot — `EstimatedRoundAt(RoleAggregatorCommittee, 12s, 8s)` — which resolves to round 1 with the fix and round 5 without it. Confirmed it fails against unpatched code (expected 1, got 5).

protocol/v2/qbft/roundtimer/timer_test.go
momosh-ssv: Agreed, these are invariant guards for the invert-arithmetic relationship, not regression coverage for this bug (both sides derive from `round1HeadStart`, so they'd stay consistent even with a wrong head start). 

The real guards are `TestRoundTimeoutOffset` and the new realistic-slot `EstimatedRoundAt` assertion. I've left these as-is for the invariant they do cover; happy to drop the aggregator-committee rows from them if you'd prefer to avoid over-implying coverage.

protocol/v2/qbft/roundtimer/timer.go
```diff
@@ -48,8 +48,9 @@ func roundTimeoutForRound(role spectypes.RunnerRole, slotDuration time.Duration,
 
 // round1HeadStart returns the extra time, on top of Round 1's normal quick timeout, that
 // Round 1 is allowed to run for a given role. Committee gets 1/3 of the slot as head start
-// (time for the block to become available); aggregator and sync-committee-contribution get
-// 2/3 of the slot (time for attestations to arrive); proposer gets zero.
+// (time for the block to become available); aggregator, aggregator-committee and
+// sync-committee-contribution get 2/3 of the slot (time for attestations to arrive before
```
iurii-ssv, line 52: **Optional docs nit (non-blocking):** now that this comment calls out `aggregator-committee`, the sibling `RoundTimeout` doc block further down (currently ~lines 133–136) reads a bit inconsistently — it still describes the head starts in the old beacon-role vocabulary and doesn't mention the new role:

```go
// - For roles BNRoleAggregator and BNRoleSyncCommitteeContribution, the base timeout is 2/3 of the slot duration.
```

Suggestion: while you're already refreshing timing docs in this file, consider updating that block in the same pass to reference `RoleAggregatorCommittee` (and the `RunnerRole` naming) so both comments stay in sync. Purely a documentation tidy-up — no behavioral impact.

iurii-ssv (approved):
LGTM, just 1 minor suggestion

ovidiu-ssv-labs (approved):

protocol/v2/qbft/roundtimer/timer.go
momosh-ssv: Good point, thanks.
Updated in d65cb943e: the `RoundTimeout` doc block now uses the `RunnerRole` names the switch actually operates on (`RoleCommittee` 1/3, `RoleAggregator`/`RoleSyncCommitteeContribution`/`RoleAggregatorCommittee` 2/3, `RoleProposer` quick/slow), consistent with the `round1HeadStart` comment.

PR #2951, opened 2026-07-15
test(runner): cover duplicate-committee aggregator selection (#2950)

### LINKED ISSUE #2950: Add UT covering AggregatorCommitteeRunner duplicate-committee aggregator selection
## Context

In PR #2941 review, greptile flagged the "already included" branch in `protocol/v2/ssv/runner/aggregator_committee.go` (two selected aggregators for the same beacon attestation committee append only to `Aggregators`, not to `AggregatorsCommitteeIndexes`/`AggregatedAttestations`) as a consensus-data desync. Verification showed it's a false positive: the shape is spec-mandated — `AggregatorCommitteeConsensusData.Validate()` requires set-equality and rejects duplicate committee indexes (spec spectest `InvalidAggregatorValidationDuplicateCommitteeIndex`), and `GetAggregateAndProofs` pairs each aggregator with the shared aggregate via its own `CommitteeIndex`.

## Problem

Nothing in this repo's CI exercises that branch: the shape rules are enforced only by `Validate()` at runtime and by tests living in the ssv-spec repo (the `aggregatorcommitteeconsensusdata` type spectests are not part of our spec-test suite). A refactor of the selection loop that breaks the invariant would first surface as `Validate()` rejections in production, not a red build.

## Task

Add a focused unit test: two validators selected as aggregators for the same beacon committee →
- one entry in `AggregatorsCommitteeIndexes` / `AggregatedAttestations`,
- two entries in `Aggregators`,
- `Validate()` passes,
- `GetAggregateAndProofs` yields two proofs sharing the same aggregate.

Raised by @y0sher in https://github.com/ssvlabs/ssv/pull/2941#discussion_r3571346910

momosh-ssv:
## What

Adds `TestAggregatorCommitteeRunnerPreConsensus_DuplicateCommitteeAggregators` — a focused UT for the `selectionLoop` branch in `protocol/v2/ssv/runner/aggregator_committee.go` where two validators are selected as aggregators for the **same** beacon attestation committee.

## Why

Raised in #2941 review (greptile flagged the branch as a consensus-data desync; verification showed it's the spec-mandated shape — see the [thread](https://github.com/ssvlabs/ssv/pull/2941#discussion_r3569359367)). Nothing in this repo's CI exercised the branch: the shape rules were enforced only by `Validate()` at runtime and by tests in the ssv-spec repo (whose `aggregatorcommitteeconsensusdata` type spectests don't run here, and whose runner spectests never use the multiple-aggregators duty fixture). A refactor breaking the invariant would have surfaced as production `Validate()` rejections instead of a red build.

## How

Drives the real pre-consensus path: `TestingAggregatorCommitteeDutyMultipleAggregators` (validators 1 & 2, one committee) → feed only pre-consensus selection-proof messages → decode the proposed value off `State.RunningInstance.StartValue` and assert:

- 2 `Aggregators` entries (validators 1 & 2, same `CommitteeIndex`)
- exactly 1 `AggregatorsCommitteeIndexes` + 1 `AggregatedAttestations` entry
- `Validate()` passes
- `GetAggregateAndProofs()` yields 2 proofs wrapping the same aggregate attestation

Closes #2950

### DIFF
--- a/protocol/v2/ssv/runner/aggregator_committee_test.go
+++ b/protocol/v2/ssv/runner/aggregator_committee_test.go
@@ -430,27 +430,36 @@ func TestAggregatorCommitteeRunnerPreConsensus_DuplicateCommitteeAggregators(t *
 	// not two independently-fetched copies), which is a stronger check than comparing the raw
 	// AggregatedAttestations bytes above. Compare the Aggregate sub-field's own hash tree root, not the
 	// whole AggregateAndProof's root (that also embeds the per-validator AggregatorIndex/SelectionProof,
-	// which legitimately differ between the two aggregators). proofs[i].Version is derived from the
-	// runner's own fork schedule (not the `version` constant used to build the fixtures), so this stays
-	// version-agnostic instead of asserting a hardcoded fork.
-	aggregateRoot := func(p *spec.VersionedAggregateAndProof) [32]byte {
+	// which legitimately differ between the two aggregators — the whole-proof roots are asserted
+	// distinct below for exactly that reason). proofs[i].Version is derived from the runner's own fork
+	// schedule (not the `version` constant used to build the fixtures), so this stays version-agnostic
+	// instead of asserting a hardcoded fork.
+	proofRoots := func(p *spec.VersionedAggregateAndProof) (aggregateRoot, wholeRoot [32]byte) {
 		t.Helper()
 		switch p.Version {
 		case spec.DataVersionElectra:
 			require.NotNil(t, p.Electra)
-			root, err := p.Electra.Aggregate.HashTreeRoot()
+			aggregateRoot, err := p.Electra.Aggregate.HashTreeRoot()
 			require.NoError(t, err)
-			return root
+			wholeRoot, err := p.Electra.HashTreeRoot()
+			require.NoError(t, err)
+			return aggregateRoot, wholeRoot
 		case spec.DataVersionFulu:
 			require.NotNil(t, p.Fulu)
-			root, err := p.Fulu.Aggregate.HashTreeRoot()
+			aggregateRoot, err := p.Fulu.Aggregate.HashTreeRoot()
 			require.NoError(t, err)
-			return root
+			wholeRoot, err := p.Fulu.HashTreeRoot()
+			require.NoError(t, err)
+			return aggregateRoot, wholeRoot
 		default:
 			t.Fatalf("unexpected aggregate-and-proof version in test: %s", p.Version)
-			return [32]byte{}
+			return [32]byte{}, [32]byte{}
 		}
 	}
-	require.Equal(t, aggregateRoot(proofs[0]), aggregateRoot(proofs[1]),
+	aggregateRoot0, wholeRoot0 := proofRoots(proofs[0])
+	aggregateRoot1, wholeRoot1 := proofRoots(proofs[1])
+	require.Equal(t, aggregateRoot0, aggregateRoot1,
 		"both aggregators' proofs must wrap the same aggregate attestation")
+	require.NotEqual(t, wholeRoot0, wholeRoot1,
+		"whole-proof roots must differ via per-validator AggregatorIndex/SelectionProof; identical roots would mean both entries collapsed into the same proof")
 }
--- a/protocol/v2/ssv/runner/aggregator_committee_test.go
+++ b/protocol/v2/ssv/runner/aggregator_committee_test.go
@@ -366,3 +366,91 @@ func TestAggregatorCommitteeRunnerProcessPostConsensus_DrainsBufferedErrCh(t *te
 		require.False(t, env.runner.State.Succeeded, "iteration %d", iter)
 	}
 }
+
+// TestAggregatorCommitteeRunnerPreConsensus_DuplicateCommitteeAggregators is the regression test for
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/runner/aggregator_committee_test.go
```diff
@@ -366,3 +366,91 @@ func TestAggregatorCommitteeRunnerProcessPostConsensus_DrainsBufferedErrCh(t *te
...
+// selectionLoop in the aggregator-committee pre-consensus path (aggregator_committee.go) must append a
+// second AssignedAggregator entry to consensusData.Aggregators for the repeat committee, while
+// AggregatorsCommitteeIndexes and AggregatedAttestations keep exactly one entry per unique committee
+// (fetched/marshaled once, not once per aggregator). This shape is otherwise only enforced by
+// consensusData.Validate() at runtime and by ssv-spec tests, not by anything in this repo.
+func TestAggregatorCommitteeRunnerPreConsensus_DuplicateCommitteeAggregators(t *testing.T) {
+	ctx := t.Context()
+	const version = spec.DataVersionElectra
+
+	base := protocoltesting.NewTestingBeaconNodeWrapped().(*protocoltesting.BeaconNodeWrapped)
```
iurii-ssv: Nit (optional): the `.(*protocoltesting.BeaconNodeWrapped)` type assertion is unnecessary here. `NewTestingBeaconNodeWrapped()` already returns `beacon.BeaconNode`, and `base` is only passed to `newAggregatorCommitteeRunnerEnv`, which takes that interface — so `base := protocoltesting.NewTestingBeaconNodeWrapped()` would suffice. The sibling tests need the assertion because they wrap `base` in `faultyAggregateSubmitBeacon{BeaconNodeWrapped: base}`; this test uses the beacon unwrapped. Harmless as-is, just redundant.

protocol/v2/ssv/runner/aggregator_committee_test.go
```diff
@@ -366,3 +366,91 @@ func TestAggregatorCommitteeRunnerProcessPostConsensus_DrainsBufferedErrCh(t *te
...
+	// version-agnostic instead of asserting a hardcoded fork.
+	aggregateRoot := func(p *spec.VersionedAggregateAndProof) [32]byte {
+		t.Helper()
+		switch p.Version {
+		case spec.DataVersionElectra:
+			require.NotNil(t, p.Electra)
+			root, err := p.Electra.Aggregate.HashTreeRoot()
+			require.NoError(t, err)
+			return root
+		case spec.DataVersionFulu:
```
iurii-ssv, line 447: Clarifying note (Greptile already flagged this `Fulu` branch as "dead code that future forks would hit as a fatal default"): this is intended and needs no change. The fixtures are Electra-anchored, so this branch is deliberate future-proofing, and the `default: t.Fatalf` below is the desired test behavior — a future fork reaching here should fail loudly so someone extends the switch, rather than silently passing on an unhandled version. No action needed.

protocol/v2/ssv/runner/aggregator_committee_test.go
```diff
@@ -366,3 +366,91 @@ func TestAggregatorCommitteeRunnerProcessPostConsensus_DrainsBufferedErrCh(t *te
...
+			require.NotNil(t, p.Fulu)
+			root, err := p.Fulu.Aggregate.HashTreeRoot()
+			require.NoError(t, err)
+			return root
+		default:
+			t.Fatalf("unexpected aggregate-and-proof version in test: %s", p.Version)
+			return [32]byte{}
+		}
+	}
+	require.Equal(t, aggregateRoot(proofs[0]), aggregateRoot(proofs[1]),
```
iurii-ssv: Optional strengthening: since the comment above notes the full `AggregateAndProof` roots "legitimately differ" (per-validator `AggregatorIndex`/`SelectionProof`), you could additionally assert `require.NotEqual` on the two whole-proof roots. That locks in that the per-aggregator fields are genuinely distinct — guarding against a hypothetical regression where both proofs collapse to identical content — and complements the "same `Aggregate` root" check here. Low priority.

iurii-ssv (approved):
Just minor suggestions/nits

ovidiu-ssv-labs (approved):

protocol/v2/ssv/runner/aggregator_committee_test.go
momosh-ssv: Good catch, fixed in 9c4c829ed.

Since this test never wraps the beacon node, `base` now stays as the plain `beacon.BeaconNode` interface.

protocol/v2/ssv/runner/aggregator_committee_test.go
momosh-ssv: Agreed on all points.
The Fulu branch is deliberate future-proofing and the fatal default is exactly the behavior we want when a new fork shows up. Leaving as is.

protocol/v2/ssv/runner/aggregator_committee_test.go
momosh-ssv: Nice suggestion, added in efd3e091d. 

The closure now returns both the Aggregate sub-root and the whole-proof root from the same version switch, and the test asserts the whole-proof roots differ while the aggregate roots match.

iurii-ssv (approved):

Issue #2952, opened 2026-07-20
Track: boole-convergence residual work (follow-ups to #2941; supersedes #2902)


iurii-ssv:
## Context

#2902 tracked the `stage` → `boole-fork` reconciliation backlog. That premise is obsolete: #2941 converges in the opposite direction (`boole-fork` → `stage` via `integration/boole-convergence`), and a full item-by-item verification (see [#2902 comment](https://github.com/ssvlabs/ssv/issues/2902#issuecomment-5022076286)) confirmed all duty fixes and infra items are present on the convergence branch. This issue aggregates **everything that remains**: five newly-found convergence gaps (boole-side fork adaptations #2941 did not carry over, found by diffing `boole-fork` against the convergence branch), plus the still-open cleanups carried over from #2902.

**Completeness:** a full hunk-level audit of the boole-fork delta (every functional file boole changed that #2941 did not touch) has been completed — the `FormatRunnerRole` duty-ID regression and the message-size-caps item below are its only new findings; everything else classified as absorbed, superseded, or cosmetic (see the triage comment below).

**None of the "before fork ships" items bite today** — every production network pins `Forks.Boole` to `MaxUint64` — so they don't block #2941. They must land before a Boole epoch is ever scheduled in a release.

## Regression active immediately after #2941 merges (not fork-gated)

[...]

PR #2957, opened 2026-07-21
boole-convergence follow-ups: pre-fork batch (#2952 items 1, 3, 4, 5, 6)

### LINKED ISSUE #2955: AGGREGATOR & SYNC_COMMITTEE_CONTRIBUTION duties log runner_role=UNDEFINED (and UNDEFINED duty_id / ssv.runner.role metric) on boole-convergence
## Summary

On the `integration/boole-convergence` line, pre-fork **aggregator** and **sync-committee-contribution** duties resolve their `runner_role` to `UNDEFINED` in logs and in the `ssv.runner.role` OTel/Prometheus attribute. The duties themselves execute and submit normally — this is an **observability regression, not a consensus break** — but it makes these duty types unidentifiable in logs and silently merges all their metrics into an `UNDEFINED` bucket.

Introduced by `84800c28b` ("convergence: bump ssv-spec to v1.2.3 + port boole's compat shim (FRONT-LOAD-0)"): the spec bump removed `RoleAggregator`/`RoleSyncCommitteeContribution` from `spectypes.RunnerRole.String()`, a local compat shim restored the wire values for routing/execution, but the observability string-formatting choke point was never pointed at the shim.

## Environment

- Branch: `deploy/boole-prefork-2007` (cut from `integration/boole-convergence`), commit `9920d1d21`
- Network: stage-hoodi, **pre-fork regime** (`Forks.Boole = MaxUint64`, `domain_type=0x00003114`) — so behavior should match `stage`
- Cluster: operators 25–28

## Observed behavior

Every log line for these duty types carries `runner_role=UNDEFINED` and `duty_id=UNDEFINED-e<epoch>-s<slot>-v<validator>`:

```
🔧 executing validator duty                           runner_role=UNDEFINED
signing and broadcasting selection proof partial sig  runner_role=UNDEFINED
🎯 got pre-consensus quorum                            runner_role=UNDEFINED
ℹ️ starting QBFT instance                              runner_role=UNDEFINED
submitted aggregate and proof                         runner_role=UNDEFINED
✔️successfully finished duty processing                runner_role=UNDEFINED  duty_id=UNDEFINED-...-v840066
```

`COMMITTEE` and `VALIDATOR_REGISTRATION` roles resolve correctly. Only the per-validator aggregator / sync-committee-contribution roles are affected.

A giveaway of the split code path: in `message/validation` "ignoring invalid message" logs, the **same** message shows a correct `duty_id=AGGREGATOR-...` next to a broken `runner_role=UNDEFINED` — two different role→string mappers on one line.

**Duties are functionally healthy** (measured on operators 25–28, boole-prefork build vs the prior `stage` build on the same nodes):

| Signal | stage build | boole-prefork build |
|---|---|---|
| `submitted aggregate and proof` | 29.3/min | 34.9/min |
| `submitted sync committee` | 25.1/min | 28.0/min |

## Root cause

The ssv-spec v1.2.3 bump left `RunnerRole.String()` with cases only for `RoleCommittee(0)`, `RoleProposer(2)`, `RoleValidatorRegistration(4)`, `RoleVoluntaryExit(5)`, `RoleAggregatorCommittee(6)` → anything else (including the deprecated `1`/`3`) returns `"UNDEFINED"`.

The compat shim re-declares the pre-fork values:

```go
// protocol/v2/types/runner_role.go:8-9
RoleAggregator                = spectypes.RunnerRole(1) // Deprecated
RoleSyncCommitteeContribution = spectypes.RunnerRole(3) // Deprecated
```

and the pre-fork runners are constructed with them (`protocol/v2/ssv/runner/aggregator.go:78`, `sync_committee_contribution.go:64`). Because these are plain `spectypes.RunnerRole` values (no distinct type), `.String()` dispatches to the spec method that no longer knows them.

The single unpatched choke point:

```go
// observability/utils/format.go:10-12
func FormatRunnerRole(runnerRole spectypes.RunnerRole) string {
    return strings.TrimSuffix(runnerRole.String(), "_RUNNER")
}
```

It feeds:

- `observability/log/fields/fields.go` — the `runner_role` zap field (used by `loggerWithDutyContext` in `operator/duties/scheduler.go`, so every runner log line)
- `observability/log/fields/duty_id.go` — the `duty_id` string
- `observability/attributes.go:41-43` — the `ssv.runner.role` OTel/metric attribute
- `protocol/v2/ssv/queue/observability.go:54-56` — `ValidatorMetricID()`

**Why `duty_id` looks correct in some logs but not others — there are two independent `duty_id` builders, and only one consults the shim.** Message-validation builds its `duty_id` with the shim-aware `ssvtypes.RunnerRoleToString` (`message/validation/logger_fields.go:107`), so its logs show the correct `AGGREGATOR-...` / `SYNC_COMMITTEE_CONTRIBUTION-...`. The runner/scheduler path builds `duty_id` with `BuildDutyID` (`observability/log/fields/duty_id.go:13`), which routes through the broken `FormatRunnerRole` and yields `UNDEFINED-...`. Worse, within a *single* message-validation line the two fields disagree: `AsZapFields()` (`logger_fields.go:39`) emits `runner_role` via `fields.RunnerRole(d.Role)` → `FormatRunnerRole` (broken → `UNDEFINED`) two lines away from the `duty_id` it built with `RunnerRoleToString` (correct → `AGGREGATOR`). So `duty_id=AGGREGATOR-... runner_role=UNDEFINED` on the same line is the signature of these two mappers diverging.

A correctly-ported mapper therefore already exists (`protocol/v2/types/runner_role.go:81-90`, `RunnerRoleToString`); the fix is to route `FormatRunnerRole` through the same logic so all consumers agree.

## Impact

- **Not consensus-breaking** — QBFT MsgID identifiers, message-validation routing (`case ssvtypes.RoleAggregator:` …), and `Queues[role]` lookup all switch on the raw integer enum (intact), so duties run and submit normally.
- **Observability blind spot** — `ssv.runner.role`-keyed dashboards/alerts (message-validation accept/reject/ignore counters, duty-scheduling metrics, pre/post-consensus duration, QBFT instance metrics, queue metrics) merge all pre-fork aggregator + sync-committee traffic into `UNDEFINED`. Logs for these duty types are also unfilterable by role.

## Suggested fix

Make the shared choke point aware of the deprecated values, delegating to the already-correct mapper:

```go
// observability/utils/format.go
func FormatRunnerRole(runnerRole spectypes.RunnerRole) string {
    switch runnerRole {
    case ssvtypes.RoleAggregator:
        return "AGGREGATOR"
    case ssvtypes.RoleSyncCommitteeContribution:
        return "SYNC_COMMITTEE_CONTRIBUTION"
    default:
        return strings.TrimSuffix(runnerRole.String(), "_RUNNER")
    }
}
```

No import cycle (`observability/attributes.go` already imports `protocol/v2/types`). Longer term, consolidate the three near-duplicate role→string mappers (`protocol/v2/types.RunnerRoleToString`, `protocol/v2/message.RunnerRoleToString`, spec `.String()`) onto one. Can be removed once the runner-merge (`TODO(convergence unit 5)` in `observability/log/fields/fields.go:289`) lands and legacy roles 1/3 are no longer constructed.

## Not affected / notes

- Exporter role decoding uses a separate, correct mapper — unaffected.
- The `ignoring invalid message … got decided message with too few signers` noise seen alongside these duties is normal decided-message dedup, present at the same rate on `stage`; unrelated to this bug.

momosh-ssv:
Follow-ups to #2941, tracked in #2952. Five independent items, one commit each:

- **Item 1 / fixes #2955** — `observability`: `FormatRunnerRole` now routes through the shim-aware `ssvtypes.RunnerRoleToString`, so pre-fork AGGREGATOR / SYNC_COMMITTEE_CONTRIBUTION duties no longer render as `UNDEFINED` (colliding duty IDs, merged `ssv.runner.role` metrics). Regression tests cover all roles + `BuildDutyID` non-collision.
- **Item 3** — `cli`: `generate-config` no longer emits fork-at-genesis configs; `Forks.Boole` defaults to `MaxUint64`, `NextDomainType` to the default network's value, with new `--ssv-boole-fork-epoch` / `--ssv-next-domain` flags (ported from boole-fork).
- **Item 4** — `network/discovery`: the `next_domaintype` ENR key now carries the real `NextDomainType` (both in `PublishENR` and the initial local-node decoration). The static-domain peer filter is kept per the resolution recorded in #2952; decision documented in code comments.
- **Item 5** — `message/validation`: `maxPartialSignatureMessages` 1000 → 5048. The old value under-sized `maxEncodedPartialSignatureSize` (151,220) vs the boole spec's AggregatorCommittee worst case (727,000 bytes / 5048 msgs), so a large valid post-fork message would have been rejected with `ErrSSVDataTooBig` pre-decode. A new test-only guard pins the hand-computed caps to the spec's `maxmsgsize` values (that spec package isn't importable from prod code — it pulls in `testing`/testify, which is presumably why the convergence dropped boole's derivation).
- **Item 6** — `network/p2p`: legacy `Broadcast` resolves the slot with a minimal per-type decode instead of the full `queue.DecodeSignedSSVMessage`, preserving exact slot/error semantics (cross-checked in tests).

Items 2 (fork-aware metadata syncer) and 7 (validation hot-path subnet cache) land separately.

**Note on `broadcastMessageSlot` semantics (item 6):** unlike the replaced `queue.DecodeSignedSSVMessage`, it does not handle `SSVEventMsgType` (falls through to `ErrUnknownMessageType`). Event messages are internal loopback and are never network-broadcast, so `Broadcast` cannot receive one — exact slot/error parity holds for the two wire types that can actually be broadcast.

### DIFF
--- a/cli/generate_config.go
+++ b/cli/generate_config.go
@@ -4,10 +4,12 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log"
+	"math"
 	"math/big"
 	"os"
 	"strings"
 
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
@@ -40,10 +42,12 @@ var (
 	operatorPrivateKey      string
 	metricsAPIPort          int
 	ssvDomain               string
+	ssvNextDomain           string
 	ssvRegistrySyncOffset   uint64
 	ssvRegistryContractAddr string
 	ssvBootnodes            string
 	ssvDiscoveryProtocolID  string
+	ssvBooleForkEpoch       uint64
 )
 
 type SSVConfig struct {
@@ -80,6 +84,11 @@ var generateConfigCmd = &cobra.Command{
 			log.Fatalf("Failed to decode network domain: %v", err)
 		}
 
+		parsedNextDomain, err := hex.DecodeString(strings.TrimPrefix(ssvNextDomain, "0x"))
+		if err != nil {
+			log.Fatalf("Failed to decode next network domain: %v", err)
+		}
+
 		parsedDiscoveryProtocolID, err := hex.DecodeString(strings.TrimPrefix(ssvDiscoveryProtocolID, "0x"))
 		if err != nil {
 			log.Fatalf("Failed to decode discovery protocol ID: %v", err)
@@ -105,10 +114,14 @@ var generateConfigCmd = &cobra.Command{
 		config.MetricsAPIPort = metricsAPIPort
 		config.SSV.CustomNetwork = &networkconfig.SSV{
 			DomainType:           spectypes.DomainType(parsedDomain),
+			NextDomainType:       spectypes.DomainType(parsedNextDomain),
 			RegistrySyncOffset:   new(big.Int).SetUint64(ssvRegistrySyncOffset),
 			RegistryContractAddr: ethcommon.HexToAddress(ssvRegistryContractAddr),
 			Bootnodes:            bootnodes,
 			DiscoveryProtocolID:  parsedDiscoveryProtocolIDArr,
+			Forks: networkconfig.SSVForks{
+				Boole: phase0.Epoch(ssvBooleForkEpoch),
+			},
 		}
 
 		data, err := yaml.Marshal(&config)
@@ -140,11 +153,14 @@ func init() {
 
 	ssvDomainDefault := "0x" + hex.EncodeToString(defaultNetwork.DomainType[:])
 	generateConfigCmd.Flags().StringVar(&ssvDomain, "ssv-domain", ssvDomainDefault, "SSV domain type")
+	ssvNextDomainDefault := "0x" + hex.EncodeToString(defaultNetwork.NextDomainType[:])
+	generateConfigCmd.Flags().StringVar(&ssvNextDomain, "ssv-next-domain", ssvNextDomainDefault, "SSV next domain type")
 	generateConfigCmd.Flags().Uint64Var(&ssvRegistrySyncOffset, "ssv-registry-sync-offset", defaultNetwork.RegistrySyncOffset.Uint64(), "SSV registry sync offset")
 	generateConfigCmd.Flags().StringVar(&ssvRegistryContractAddr, "ssv-registry-contract-addr", defaultNetwork.RegistryContractAddr.String(), "SSV registry contract addr")
 	generateConfigCmd.Flags().StringVar(&ssvBootnodes, "ssv-bootnodes", strings.Join(defaultNetwork.Bootnodes, sliceSeparator), "SSV bootnodes (comma-separated)")
 	ssvDiscoveryProtocolIDDefault := "0x" + hex.EncodeToString(defaultNetwork.DiscoveryProtocolID[:])
 	generateConfigCmd.Flags().StringVar(&ssvDiscoveryProtocolID, "ssv-discovery-protocol-id", ssvDiscoveryProtocolIDDefault, "SSV discovery protocol ID")
+	generateConfigCmd.Flags().Uint64Var(&ssvBooleForkEpoch, "ssv-boole-fork-epoch", math.MaxUint64, "Epoch at which the Boole fork occurs in the network")
 
 	RootCmd.AddCommand(generateConfigCmd)
 }
--- a/cli/generate_config.go
+++ b/cli/generate_config.go
@@ -83,10 +83,16 @@ var generateConfigCmd = &cobra.Command{
 		if err != nil {
 			log.Fatalf("Failed to decode network domain: %v", err)
 		}
+		if len(parsedDomain) != 4 {
+			log.Fatalf("Invalid network domain length: expected 4 bytes, got %d", len(parsedDomain))
+		}
 
 		parsedNextDomain, err := hex.DecodeString(strings.TrimPrefix(ssvNextDomain, "0x"))
 		if err != nil {
 			log.Fatalf("Failed to decode next network domain: %v", err)
+		}
+		if len(parsedNextDomain) != 4 {
+			log.Fatalf("Invalid next network domain length: expected 4 bytes, got %d", len(parsedNextDomain))
 		}
 
 		parsedDiscoveryProtocolID, err := hex.DecodeString(strings.TrimPrefix(ssvDiscoveryProtocolID, "0x"))
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -44,8 +44,8 @@ const (
 const (
 	partialSignatureSize           = 96
 	partialSignatureMsgSize        = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
-	maxPartialSignatureMessages    = 1000
-	partialSigMsgTypeSize          = 8 // uint64
+	maxPartialSignatureMessages    = 5048 // post-fork worst case (boole RoleAggregatorCommittee), see ssv-spec types/spectest/tests/maxmsgsize.MaxSizePartialSignatureMessages
+	partialSigMsgTypeSize          = 8    // uint64
 	maxPartialSignatureMsgsSize    = partialSigMsgTypeSize + slotSize + maxPartialSignatureMessages*partialSignatureMsgSize
 	maxEncodedPartialSignatureSize = maxPartialSignatureMsgsSize + maxPartialSignatureMsgsSize/encodingOverheadDivisor + 4
 )
--- a/message/validation/const.go
+++ b/message/validation/const.go
@@ -42,10 +42,13 @@ const (
 )
 
 const (
-	partialSignatureSize           = 96
-	partialSignatureMsgSize        = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
-	maxPartialSignatureMessages    = 5048 // post-fork worst case (boole RoleAggregatorCommittee), see ssv-spec types/spectest/tests/maxmsgsize.MaxSizePartialSignatureMessages
-	partialSigMsgTypeSize          = 8    // uint64
+	partialSignatureSize    = 96
+	partialSignatureMsgSize = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
+	// maxPartialSignatureMessages is the post-fork worst case (boole RoleAggregatorCommittee). The
+	// count derives from the inner ssv-spec types/spectest/tests/maxmsgsize.MaxSizePartialSignatureMessages;
+	// the drift guard in const_test.go checks the full envelope, MaxSizeSSVMessageFromPartialSignatureMessages.
+	maxPartialSignatureMessages    = 5048
+	partialSigMsgTypeSize          = 8 // uint64
 	maxPartialSignatureMsgsSize    = partialSigMsgTypeSize + slotSize + maxPartialSignatureMessages*partialSignatureMsgSize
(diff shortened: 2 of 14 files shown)


message/validation/const.go
```diff
@@ -44,8 +44,8 @@ const (
 const (
 	partialSignatureSize           = 96
 	partialSignatureMsgSize        = partialSignatureSize + rootSize + operatorIDSize + validatorIndexSize
-	maxPartialSignatureMessages    = 1000
-	partialSigMsgTypeSize          = 8 // uint64
+	maxPartialSignatureMessages    = 5048 // post-fork worst case (boole RoleAggregatorCommittee), see ssv-spec types/spectest/tests/maxmsgsize.MaxSizePartialSignatureMessages
```
iurii-ssv: Doc nit: this comment cites `maxmsgsize.MaxSizePartialSignatureMessages` (726,932 — the inner `PartialSignatureMessages` SSZ size that the `5048` count derives from), but the drift guard in `const_test.go` actually pins `maxEncodedPartialSignatureSize` against `MaxSizeSSVMessageFromPartialSignatureMessages` (727,000 — the full SSV-message envelope). Both are legitimate references, but citing the same constant the guard checks (or spelling out that "5048 comes from the inner size, the guard checks the outer envelope") would save whoever re-derives these after a future spec bump some head-scratching.

network/p2p/p2p_pubsub.go
```diff
@@ -40,18 +41,36 @@ func (n *p2pNetwork) UseMessageRouter(router network.MessageRouter) {
...
-	msgSlot, err := decodedMsg.Slot()
+	msgSlot, err := broadcastMessageSlot(msg)
 	if err != nil {
 		return fmt.Errorf("could not resolve message slot: %w", err)
 	}
 
 	return n.BroadcastAtSlot(msg, msgSlot)
 }
 
+// broadcastMessageSlot resolves the slot of a message that's about to be broadcast, decoding only
```
iurii-ssv, line 52: Two low-priority notes on this helper:

1. **Behavioral divergence for event messages.** The replaced `queue.DecodeSignedSSVMessage(msg).Slot()` also handled `SSVEventMsgType` (it would decode the `EventMsg` and return its slot); this switch instead falls through to `ErrUnknownMessageType` for that type. It's harmless — `EventMsg` is internal loopback and is never network-broadcast, so `Broadcast` can't actually receive one — and arguably more correct, but it does mean the commit's "exact slot/error semantics" holds only for the two wire types that can be broadcast. Might be worth a word in the commit/PR body so the divergence is on the record.

2. Completing greptile's note on the consensus branch above: the partial-signature branch (line 66) carries the same mislabel — the error says `SignedPartialSignatureMessage` while it's decoding a plain `PartialSignatureMessages`. Both strings are copied verbatim from `queue.ExtractMsgBody`, so if grep-parity with the source was the intent that's a fair reason to leave them; otherwise worth fixing both together.

iurii-ssv (commented):
Light follow-up pass — two low-priority polish notes inline. I skipped findings already covered by greptile: the --ssv-next-domain slice→[4]byte panic on wrong-length input (cli/generate_config.go) is a real catch, and that note already accounts for the pre-existing --ssv-domain gap, so I've nothing to add there.

network/p2p/p2p_pubsub.go
```diff
@@ -40,18 +41,36 @@ func (n *p2pNetwork) UseMessageRouter(router network.MessageRouter) {
...
+// the message body type actually sent over the wire (consensus or partial-signature), rather than
+// building a full queue.DecodedSSVMessage wrapper that Broadcast has no other use for.
+func broadcastMessageSlot(msg *spectypes.SignedSSVMessage) (phase0.Slot, error) {
+	switch msg.SSVMessage.MsgType {
+	case spectypes.SSVConsensusMsgType:
+		qbftMsg := &specqbft.Message{}
+		if err := qbftMsg.Decode(msg.SSVMessage.Data); err != nil {
+			return 0, fmt.Errorf("failed to decode SignedMessage: %w", err)
+		}
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 560e890b9.

cli/generate_config.go
```diff
@@ -80,6 +84,11 @@ var generateConfigCmd = &cobra.Command{
 			log.Fatalf("Failed to decode network domain: %v", err)
 		}
 
+		parsedNextDomain, err := hex.DecodeString(strings.TrimPrefix(ssvNextDomain, "0x"))
+		if err != nil {
+			log.Fatalf("Failed to decode next network domain: %v", err)
+		}
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 115dd94b8.

message/validation/const.go
momosh-ssv: Good catch, fixed in 2c68d81e3. 

The comment now spells out that 5048 derives from the inner `MaxSizePartialSignatureMessages` while the drift guard pins the full-envelope `MaxSizeSSVMessageFromPartialSignatureMessages`.

network/p2p/p2p_pubsub.go
momosh-ssv: Both points addressed. The `SSVEventMsgType` divergence is now on the record in 8f7656917.

A comment on the `default:` branch plus a note in the PR description. And 560e890b9 fixes both error labels; I don't think grep-parity is worth keeping there, since `queue.ExtractMsgBody`'s own labels are a pre-alan leftover anyway.

iurii-ssv (approved):

ovidiu-ssv-labs (commented):
Clean, well-tested bundle. All five items are correct: item 1 fixes the legacy-role UNDEFINED collision at the single FormatRunnerRole choke point (all named consumers covered, no duty_id collision); item 3's generate-config defaults and round-trip are right; item 4 propagates NextDomainType to both (and only) ENR writers with no peer-filter interaction; item 5's 5048 bump matches the spec derivation with a drift-proof guard; item 6 preserves slot/error semantics for both wire message types. Only two minor, non-blocking notes.  [verdict: yes]

**Finding · [MINOR] duty_executor.go still renders legacy runner roles as UNDEFINED via raw role.String() in an error message** — `protocol/v2/ssv/validator/duty_executor.go:40`

Item 1 fixes the *observability* choke point (`FormatRunnerRole`), which is exactly the set of consumers #2955 enumerated. However, `duty_executor.go:40` builds an error string with the raw spec mapping:

```go
q, ok := v.Queues[role]
if !ok {
    return fmt.Errorf("no queue for role %s", role.String())
}
```

Because `role` here is a `spectypes.RunnerRole` and the surrounding comment explicitly anticipates the legacy `Aggregator(1)` / `SyncCommitteeContribution(3)` values ("validators built post-fork have no legacy Aggregator/SyncCommitteeContribution runners"), this is precisely the case where `role.String()` returns `"UNDEFINED"` — the exact class of bug item 1 addresses, just in an error string rather than a metric/log field. It's out of #2955's stated scope and low impact (a diagnostic message, not a metric label collision), but since this PR is doing the pass to eliminate legacy-role `UNDEFINED` rendering, routing this one through the shim-aware formatter too would make the error actionable ("no queue for role AGGREGATOR" vs "no queue for role UNDEFINED").

**Suggested fix:** Use the shim-aware formatter here as well:

```go
return fmt.Errorf("no queue for role %s", ssvtypes.RunnerRoleToString(role))
// or observability/utils.FormatRunnerRole(role)
```

Optional / follow-up — not required to merge.

ovidiu-ssv-labs (approved):
Looks good, with the minor finding above. You can decide whether to fix it here or as follow-up.

momosh-ssv:
> However, duty_executor.go:40 builds an error string with the raw spec mapping...

Good catch, fixed in 19427577b. 

While touching it, seems that `StartQueueConsumer` in queue_validator.go had the same raw `String()` in its queue-not-found error, so I routed that one through the shim in the same commit too.

ovidiu-ssv-labs (approved):
Nicely fixed.

PR #2958, opened 2026-07-21
operator/validator/metadata: make syncer subnet selection fork-aware (#2952 item 2)

momosh-ssv:
Item 2 of #2952 (follow-up to #2941).

The metadata syncer's `selfSubnets` and both share-relevance filters (`SyncAll`, `nextBatchFromDB`) computed subnets exclusively via the Alan mapping. Post-fork the node's live topics are Boole subnets (~99% of committees map differently), so metadata would be kept fresh for the wrong neighbor set: committees sharing our Boole subnets but not our Alan subnets would never sync → their `ValidatorIndex`/metadata stays unset → reject-level message validation (`ErrNoValidators`) misfires on their gossip on our own topics.

This ports boole-fork's fork-aware switch verbatim (the resulting `syncer.go` is byte-identical to boole-fork's): `*networkconfig.Network` is injected into the `Syncer`, and the subnet mapping follows the fork phase — Alan pre-fork / **both** inside the transition window / Boole post-fork. The transition-window arm is deliberately evaluated before `BooleForkAtSlot`: the subsequent window is post-fork, and the flipped order would silently drop the old-topic neighbor set there.

- Pre-fork selection is byte-identical to the previous inline `SetCommitteeSubnet` logic (`Forks.Boole = MaxUint64` always hits the Alan arm), so this is behavior-neutral on today's networks.
- Exporters (`fixedSubnets = all`) are unaffected in every phase (covered by a test).
- New three-phase tests use per-test `TestNetwork` copies with `Forks.Boole` set explicitly, so the `SSV_TEST_BOOLE_FORK` env override can't perturb them.
- No mock changes (`netCfg` is a concrete type); `make generate` is diff-clean.

### DIFF
--- a/operator/validator/metadata/syncer.go
+++ b/operator/validator/metadata/syncer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"maps"
-	"math/big"
 	"slices"
 	"time"
 
@@ -14,6 +13,7 @@ import (
 	spectypes "github.com/ssvlabs/ssv-spec/types"
 
 	networkcommons "github.com/ssvlabs/ssv/network/commons"
+	"github.com/ssvlabs/ssv/networkconfig"
 	"github.com/ssvlabs/ssv/observability/log"
 	"github.com/ssvlabs/ssv/observability/log/fields"
 	"github.com/ssvlabs/ssv/protocol/v2/blockchain/beacon"
@@ -45,6 +45,7 @@ const (
 
 type Syncer struct {
 	logger            *zap.Logger
+	netCfg            *networkconfig.Network
 	shareStorage      shareStorage
 	validatorStore    selfValidatorStore
 	beaconNode        beacon.BeaconNode
@@ -66,6 +67,7 @@ type selfValidatorStore interface {
 
 func NewSyncer(
 	logger *zap.Logger,
+	netCfg *networkconfig.Network,
 	shareStorage shareStorage,
 	validatorStore selfValidatorStore,
 	beaconNode beacon.BeaconNode,
@@ -74,6 +76,7 @@ func NewSyncer(
 ) *Syncer {
 	u := &Syncer{
 		logger:            logger.Named(log.NameShareMetadataSyncer),
+		netCfg:            netCfg,
 		shareStorage:      shareStorage,
 		validatorStore:    validatorStore,
 		beaconNode:        beaconNode,
@@ -102,14 +105,11 @@ func WithSyncInterval(interval time.Duration) Option {
 // and triggers a full metadata synchronization for them.
 // It returns a mapping of validator public keys to their updated metadata.
 func (s *Syncer) SyncAll(ctx context.Context) (beacon.ValidatorMetadataMap, error) {
-	subnetsBuf := new(big.Int)
-	ownSubnets := s.selfSubnets(subnetsBuf)
+	ownSubnets := s.selfSubnets()
 
 	// Load non-liquidated shares.
 	shares := s.shareStorage.List(nil, registrystorage.ByNotLiquidated(), func(share *ssvtypes.SSVShare) bool {
-		networkcommons.SetCommitteeSubnet(subnetsBuf, share.CommitteeID())
-		subnet := subnetsBuf.Uint64()
-		return ownSubnets.IsSet(subnet)
+		return s.shareInOwnSubnets(share, ownSubnets)
 	})
 	if len(shares) == 0 {
 		s.logger.Info("could not find non-liquidated own subnets validator shares on initial metadata retrieval")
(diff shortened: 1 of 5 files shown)


operator/validator/metadata/syncer.go
```diff
@@ -334,21 +330,35 @@ func (s *Syncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
 
 // selfSubnets calculates the operator's subnets by adding up the fixed subnets and the active committees
 // it recvs big int buffer for memory reusing, if is nil it will allocate new
```
iurii-ssv: **Stale doc comment.** This line still documents the removed `buf *big.Int` parameter — "it recvs big int buffer for memory reusing, if is nil it will allocate new" — but `selfSubnets()` no longer takes a buffer. Suggest deleting this line; the preceding line stands on its own.

operator/validator/metadata/syncer.go
```diff
@@ -334,21 +330,35 @@ func (s *Syncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
...
-
 	mySubnets := s.fixedSubnets
-
+	currentSlot := s.netCfg.EstimatedCurrentSlot()
 	// Compute the new subnets according to the active committees/validators.
 	myValidators := s.validatorStore.SelfValidators()
 	for _, v := range myValidators {
-		networkcommons.SetCommitteeSubnet(localBuf, v.CommitteeID())
-		mySubnets.Set(localBuf.Uint64())
+		switch {
```
iurii-ssv, line 345: **Minor (efficiency/readability).** `currentSlot` is snapshotted once above, so `InBooleTransitionWindow(currentSlot)` / `BooleForkAtSlot(currentSlot)` are invariant across the loop, yet the switch re-evaluates both for every validator. Deciding the fork phase once before the loop and switching on that would be slightly cleaner and cheaper.

This also dovetails with the independent `EstimatedCurrentSlot()` snapshots in `selfSubnets` vs `shareInOwnSubnets` flagged separately: computing the fork phase once per sync operation and threading it through both would resolve both points together.

iurii-ssv (commented):
Just couple minor things.

operator/validator/metadata/syncer.go
```diff
@@ -334,21 +330,35 @@ func (s *Syncer) sleep(ctx context.Context, d time.Duration) (slept bool) {
...
+	currentSlot := s.netCfg.EstimatedCurrentSlot()
+	switch {
+	case s.netCfg.InBooleTransitionWindow(currentSlot):
+		return ownSubnets.IsSet(share.AlanCommitteeSubnet()) || ownSubnets.IsSet(share.BooleCommitteeSubnet())
+	case s.netCfg.BooleForkAtSlot(currentSlot):
+		return ownSubnets.IsSet(share.BooleCommitteeSubnet())
+	default:
+		return ownSubnets.IsSet(share.AlanCommitteeSubnet())
+	}
+}
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 2d407d5c1.

operator/validator/metadata/syncer_test.go
```diff
@@ -788,3 +798,158 @@ func TestSyncer_Sleep(t *testing.T) {
...
+	defer ctrl.Finish()
+
+	logger := zap.NewNop()
+
+	share1 := &ssvtypes.SSVShare{
+		Share:      spectypes.Share{ValidatorPubKey: spectypes.ValidatorPK{0x1}, Committee: []*spectypes.ShareMember{{Signer: 1}}},
+		Liquidated: false,
+	}
+	share2 := &ssvtypes.SSVShare{
+		Share:      spectypes.Share{ValidatorPubKey: spectypes.ValidatorPK{0x2}, Committee: []*spectypes.ShareMember{{Signer: 2}}},
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 927b683a1.

operator/validator/metadata/syncer.go
momosh-ssv: Good catch, fixed in f5985aad1, dropped the leftover buffer line.

operator/validator/metadata/syncer.go
momosh-ssv: Done in 2d407d5c1.

Went with the combined approach you suggested: the slot is snapshotted once per sync pass and threaded through both selfSubnets and shareInOwnSubnets, and the fork-phase checks are hoisted above the validator loop.

iurii-ssv (approved):
LGTM aside from the pipeline failures we probably need to check out

ovidiu-ssv-labs (commented):
All good.

ovidiu-ssv-labs (approved):

Issue #2956, opened 2026-07-21
Decide whether the Boole fork changes the StorageName DB fork suffix


momosh-ssv:
## Context

`networkconfig/network.go` still pins the storage fork name to `alan`:

```go
func (n Network) StorageName() string {
	return fmt.Sprintf("%s:%s", n.SSV.Name, alanForkName) // TODO: decide what forks change DB fork name
}
```

This was a deliberate choice for the Boole → stage convergence (#2941): keeping the suffix stable means existing DBs stay valid across the fork. But the decision needs an owner before a Boole epoch actually ships in a release — flagged by @iurii-ssv in https://github.com/ssvlabs/ssv/pull/2941#discussion_r3614078202 so it doesn't get lost.

## To decide

- Does the Boole fork warrant a new DB fork name (forcing a resync/migration), or does the schema stay compatible so `alan` should simply be renamed/kept as a network-level constant?
- If it stays, remove the TODO and document the policy for future forks.

## Related

- While touching version pins: `go.mod` and `ssvsigner/go.mod` carry the same `ssv-spec` pseudo-version; both should move to the tagged `v1.2.3` once it exists (https://github.com/ssvlabs/ssv/pull/2941#discussion_r3586621076).

momosh-ssv:
Looked into this — seems the deciding constraint is that `StorageName()` feeds `ConfigLock.NetworkName` (`cli/operator/config.go` → `operator/storage/config_lock.go`), so changing the suffix fails every existing node at startup ("database must be removed or reinitialized") — and it would do so at binary-upgrade time, not at fork activation, since the name is static. Given the convergence keeps the schema dual-fork-aware on the same DB, I fear a suffix change would just force a fleet-wide resync for no benefit.

So my suggestion: keep `alan` byte-for-byte as an opaque compatibility token and document the policy — the suffix changes only when a fork actually invalidates the stored layout and a coordinated resync is intended. Opened #2959 doing exactly that (const rename + policy comment, TODO removed).

On the version pins: `v1.2.3` doesn't exist as a tag on ssv-spec yet (latest is `v1.2.2`), so moving both go.mods off the pseudo-version is blocked until someone tags the spec repo.

iurii-ssv:
Thanks for picking this up and making the call, @momosh-ssv — agree with keeping `alan` for the convergence. As long as Boole leaves the stored layout intact there's no reason to force a resync, and dropping the TODO is the right move.

[...]

PR #2959, opened 2026-07-21
networkconfig: document the storage fork-name policy (#2956)

momosh-ssv:
Addresses the decision half of #2956.

`StorageName()` feeds `ConfigLock.NetworkName`, so any change to the suffix makes every existing node fail startup with "network mismatch… database must be removed or reinitialized" — and it does so at **binary-upgrade** time, not fork time, since the name is static. The Boole convergence deliberately keeps the schema dual-fork-aware on the same DB (that's what makes #2941 an in-place upgrade), so forcing a network-wide resync buys nothing.

This keeps the `alan` string byte-for-byte, renames the const to `storageForkName`, and replaces the TODO with the policy: the suffix changes only when a fork invalidates the stored layout and a coordinated resync is intended.

The `go.mod`/`ssvsigner/go.mod` pin to a tagged `v1.2.3` (also tracked in #2956) remains blocked: the tag doesn't exist on ssv-spec yet (latest is `v1.2.2`).

### DIFF
--- a/networkconfig/network.go
+++ b/networkconfig/network.go
@@ -23,14 +23,21 @@ func (n Network) String() string {
 	return string(jsonBytes)
 }
 
-const alanForkName = "alan"
+// storageForkName is the fork suffix baked into StorageName. It is an opaque
+// compatibility token, not a statement of the network's current fork: changing it
+// makes every existing node fail startup with a config-lock mismatch until the DB
+// is removed, so it must only change when a fork invalidates the stored data
+// layout and a coordinated, network-wide resync is intended. The Boole fork keeps
+// the schema dual-fork-aware on the same DB, so the historical "alan" value stays.
+const storageForkName = "alan"
+
 const boolePriorWindowEpochs = phase0.Epoch(1)    // epochs before Boole to subscribe to both topic sets
 const booleSubsequentWindowSlots = phase0.Slot(1) // slots after Boole to keep accepting old-topic messages
 
 // StorageName returns a config name used to make sure the stored network doesn't differ.
 // It combines the network name with fork name.
 func (n Network) StorageName() string {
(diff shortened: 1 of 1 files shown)


iurii-ssv:
Nice — the const doc reads much better than a bare TODO, and keeping `alan` is the right call. One wording suggestion, plus a naming nit.

> …it must only change when a fork invalidates the stored data layout…

This points at the wrong tool. A layout change is normally a **migration**, not a token bump: `migration_5_change_share_format_from_gob_to_ssz` converted the share format gob→ssz in place with `StorageName()` untouched and no resync, and `migration_4` rewrote the stored network name in place. Migrations run before the config-lock check, so that path is always available.

So the token isn't the lever for "layout changed" — it's the escape hatch for the narrower "old data is **unrecoverable** and we want a deliberate network-wide wipe." Wording it that way steers the next person to a migration first, and closes the trap where someone changes the layout, writes no migration, and forgets to bump the token → the guard silently passes and a new reader reads old data. (Accidental edits to the value itself are already caught by the `testnet:alan` expectations in `node_test.go`; this judgment call is the remaining gap, hence worth getting precise.)

Minor/optional: the const is still `...ForkName` while the doc says it's "not a statement of the network's current fork" — `storageCompatToken` or `storageLayoutTag` would fit better. Same framing lingers in `StorageName()`'s own godoc ("combines the network name with fork name"), worth aligning if you rename. Non-blocking.

momosh-ssv:
Good point on migrations being the default path.
You're right that `migration_5` is exactly the precedent, and they run before the config-lock check. 

Reworded in a6225ac94: the doc now steers layout changes to a migration first and frames the token bump as the escape hatch for unrecoverable data and intentional network-wide resync. Took the rename too. Went with `storageCompatToken`, and aligned the `StorageName()` godoc.

iurii-ssv (approved):

ovidiu-ssv-labs (approved):

PR #2960, opened 2026-07-21
protocol/v2/ssv/runner: restore SCC pre-consensus subnet-ordering flow test (#2952 item 8)

momosh-ssv:
Item 8 of #2952 (follow-up to #2941). Test-only — no production code touched.

The convergence deleted stage's runner test kit and the flow-level test asserting that the SyncCommitteeContribution pre-consensus path calls the beacon node with **sorted subnets**; helper-level coverage (`TestSortBySubnet`) survived, but it can't catch a dropped/misplaced `sortBySubnet` call at the `ProcessPreConsensus` call site — which is where the non-determinism actually lives (`slices.Collect(maps.Keys(...))` is randomized per range).

The restored test drives a full pre-consensus quorum through `ProcessPreConsensus` on the current runner scaffolding (a capture wrapper over the shared testing beacon node) and asserts `GetSyncCommitteeContribution` receives ascending subnets. Two deliberate hardenings over stage's original fixture:

- all **4** mainnet sync-committee subnets instead of 3 (1/24 instead of 1/6 odds per draw that random iteration accidentally comes out sorted), and
- the quorum is re-driven on 25 fresh runners per test run.

Mutation-checked: with `sortBySubnet(pairs)` commented out, the test failed 5/5 runs (e.g. `got [2 3 0 1]`); restored, the full runner package passes with `-count=3`.

### DIFF
--- a/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
+++ b/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
@@ -80,8 +80,8 @@ var syncCommitteeContributionValidatorSyncCommitteeIndices = []spectypes.Validat
 // Because Go's map-iteration randomization is per-range (not fixed for the process), a single
 // draw could in principle land on already-sorted order and let a missing sortBySubnet call slip
 // through undetected; the sub-test below re-drives the whole flow on fresh runners several times to
-// make that false negative astronomically unlikely, and the mutation check documented in this
-// package's test-authoring report empirically confirms the guard trips when the sort is removed.
+// make that false negative astronomically unlikely. Deleting the sortBySubnet call from
+// ProcessPreConsensus empirically trips this guard (e.g. "got [2 3 0 1]").
 func TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall(t *testing.T) {
 	t.Parallel()
 
--- a/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
+++ b/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
@@ -61,7 +61,7 @@ func (b *syncCommitteeContributionPreConsensusCaptureBeacon) GetSyncCommitteeCon
 // non-deterministic map iteration in basePreConsensusMsgProcessing happens to already come out
 // sorted (1/4! per draw instead of 1/3!), which matters because the test below also re-drives the
 // quorum across several fresh runners to make an accidental false-negative in the mutation check
-// (see TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall) vanishingly unlikely.
+// vanishingly unlikely.
 var syncCommitteeContributionValidatorSyncCommitteeIndices = []spectypes.ValidatorSyncCommitteeIndex{0, 128, 256, 384}
 
 // TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall is the runner-flow-level
--- a/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
+++ b/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
@@ -85,6 +85,9 @@ var syncCommitteeContributionValidatorSyncCommitteeIndices = []spectypes.Validat
 func TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall(t *testing.T) {
 	t.Parallel()
 
+	ctx := context.Background()
+	logger := zap.NewNop()
+
 	const draws = 25
 	for i := range draws {
 		testBeacon := newSyncCommitteeContributionPreConsensusCaptureBeacon()
@@ -97,10 +100,8 @@ func TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall(t *
 			ValidatorIndex:                spectestingutils.TestingValidatorIndex,
 			ValidatorSyncCommitteeIndices: syncCommitteeContributionValidatorSyncCommitteeIndices,
 		}
-		require.NoError(t, runner.StartNewDuty(context.Background(), zap.NewNop(), duty, keySet.Threshold))
+		require.NoError(t, runner.StartNewDuty(ctx, logger, duty, keySet.Threshold))
 
-		ctx := context.Background()
-		logger := zap.NewNop()
 		for operatorID := spectypes.OperatorID(1); operatorID <= keySet.Threshold; operatorID++ {
 			msg := spectestingutils.PreConsensusContributionProofWithValidatorSyncCommitteeIndices(
 				keySet.Shares[operatorID], keySet.Shares[operatorID], operatorID, operatorID,
--- a/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
+++ b/protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
@@ -1,0 +1,179 @@
+package runner
+
+import (
+	"context"
+	"slices"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
```diff
@@ -0,0 +1,179 @@
...
+// boundary where the non-determinism actually lives: a helper-only test can't catch a dropped or
+// misplaced sortBySubnet call at the ProcessPreConsensus call site; this one can. Re-added per
+// ssvlabs/ssv#2952 item 8 after the Boole convergence (#2941) deleted stage's runner test kit and
+// the flow-level test that depended on it.
+//
+// Because Go's map-iteration randomization is per-range (not fixed for the process), a single
+// draw could in principle land on already-sorted order and let a missing sortBySubnet call slip
+// through undetected; the sub-test below re-drives the whole flow on fresh runners several times to
+// make that false negative astronomically unlikely, and the mutation check documented in this
+// package's test-authoring report empirically confirms the guard trips when the sort is removed.
```
iurii-ssv: nit (doc): this references "this package's test-authoring report," but there's no such file in the repo, so it reads as a dangling pointer for future maintainers. The mutation evidence belongs in the PR description (where it already lives). Suggest dropping this clause or inlining the concrete result instead, e.g. "removing `sortBySubnet` makes this test fail (`got [3 0 1 2]`)".

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// syncCommitteeContributionValidatorSyncCommitteeIndices spans all 4 mainnet sync-committee
+// subnets (subnetSize = 512/4 = 128, see ssv-spec TestingBeaconNode.SyncCommitteeSubnetID), so the
+// pre-consensus quorum below produces 4 distinct (subnet, selection-proof) pairs. Using all 4
+// subnets (rather than stage's original 3-subnet fixture) shrinks the odds that the
+// non-deterministic map iteration in basePreConsensusMsgProcessing happens to already come out
+// sorted (1/4! per draw instead of 1/3!), which matters because the test below also re-drives the
+// quorum across several fresh runners to make an accidental false-negative in the mutation check
+// (see TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall) vanishingly unlikely.
```
iurii-ssv: nit (doc, trivial): this parenthetical points forward to `TestSyncCommitteeAggregatorProcessPreConsensusSortsSubnetsForBeaconCall` — the test defined immediately below in this same file — so the cross-reference is circular. The sentence already explains the 1/4!-per-draw reasoning; the `(see ...)` could just be dropped.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
```diff
@@ -0,0 +1,179 @@
...
+		duty := &spectypes.ValidatorDuty{
+			Type:                          spectypes.BNRoleSyncCommitteeContribution,
+			PubKey:                        spectestingutils.TestingValidatorPubKey,
+			Slot:                          spectestingutils.TestingDutySlot,
+			ValidatorIndex:                spectestingutils.TestingValidatorIndex,
+			ValidatorSyncCommitteeIndices: syncCommitteeContributionValidatorSyncCommitteeIndices,
+		}
+		require.NoError(t, runner.StartNewDuty(context.Background(), zap.NewNop(), duty, keySet.Threshold))
+
+		ctx := context.Background()
```
iurii-ssv: nit (trivial): `ctx`/`logger` are recreated here, but line 100 already built `context.Background()` / `zap.NewNop()` inline for `StartNewDuty`. Both are immutable — declaring a single `ctx`/`logger` once (e.g. just above the loop) and reusing them for `StartNewDuty` and `ProcessPreConsensus` would drop the duplication.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
```diff
@@ -0,0 +1,179 @@
...
+		require.Equal(t, 1, testBeacon.getContributionCalls,
+			"draw %d: GetSyncCommitteeContribution should be called exactly once, when pre-consensus quorum is reached", i)
+
+		// Sortedness is the invariant the determinism fix guarantees: without the sort the runner
+		// would hand the beacon node a random permutation of the upstream map-iteration order.
+		require.Truef(t, slices.IsSorted(testBeacon.capturedSubnets),
+			"draw %d: subnets passed to GetSyncCommitteeContribution must be ascending (spec-canonical), got %v",
+			i, testBeacon.capturedSubnets)
+		require.Equal(t, []uint64{0, 1, 2, 3}, testBeacon.capturedSubnets, "draw %d: exact-set sanity check", i)
+		require.Len(t, testBeacon.capturedSelectionProofs, len(testBeacon.capturedSubnets), "draw %d", i)
```
iurii-ssv, line 130: Clarifying note on the bot's pairing comment above (not a second report of it): the (subnet, proof) pairing here is guaranteed *by construction*, so the risk is lower than "a broken `sortBySubnet` that reordered subnets without dragging proofs" implies. `pairs` is a single `[]subnetSelectionProof` (subnet + proof in one struct) and `sortBySubnet` sorts that one slice; the parallel `selectionProofs`/`subnets` are split out index-aligned *after* the sort. Producing a mispairing would take rewriting the pairing into two slices, not a sort bug.

Also, the suggested "verify each proof's first byte" won't work at this level: unlike `TestSortBySubnet`'s synthetic `mk()` proofs, these are real reconstructed BLS selection-proof signatures whose bytes don't encode the subnet. If an end-to-end pairing check is still wanted, the tractable route is to verify each captured proof against the `SyncAggregatorSelectionData{Slot, SubcommitteeIndex: subnet}` signing root for its paired subnet.

iurii-ssv (commented):
Minor / non-blocking nits on an otherwise solid test-only PR.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
momosh-ssv: Good catch. That report was a working artifact that never landed in the repo. Fixed in 02a90aeb.

Dropped the reference and inlined the concrete result instead (re-ran the mutation to confirm: removing `sortBySubnet` fails with `got [2 3 0 1]`).

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
momosh-ssv: Fixed in 8b8b206d.

Dropped the circular parenthetical; the 1/4! reasoning stands on its own.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
momosh-ssv: Fixed in 1193ad14.

`ctx`/`logger` are now declared once above the draws loop and reused for both calls.

protocol/v2/ssv/runner/sync_committee_contribution_preconsensus_flow_test.go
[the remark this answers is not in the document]
momosh-ssv: Fixed in dc1c6fea.

momosh-ssv: Agreed on both points.
The pairing is by construction today, and byte-inspection can't work on real signatures. 

That said, the signing-root verification route you suggested turned out cheap (~15 lines), so I added it in dc1c6fea anyway: it makes the guard survive a future refactor of the pairs → parallel-slices split, and I verified it trips on a deliberate mispairing mutation.

ovidiu-ssv-labs (approved):

iurii-ssv (approved):

PR #2961, opened 2026-07-21
message/validation: cached committee subnets on the topic hot path (#2952 item 7)

momosh-ssv:
Item 7 of #2952 (follow-up to #2941) — the last "before fork ships" convergence gap.

`validateTopicAtSlot` recomputed the committee subnet on **every** message: a SHA-256 per operator post-fork (`BooleCommitteeSubnet`) and a `big.Int` mod pre-fork (`CommitteeTopicID`). Two commits, sequenced:

1. **`registry/storage`** — `Committee.BooleSubnet`/`AlanSubnet` were caller-populated fields (set only by `buildCommittee` under a `len` guard; ~20 test fixtures never set them). Replaced with memoized accessor methods (`atomic.Pointer`, mirroring the shipped `SSVShare` pattern) computed from `Operators`/`ID` on first access — an unpopulated cache is now unrepresentable. `BroadcastAtSlot` topic selection and pubsub scoring switch to the methods. No serialization of `Committee` exists anywhere (verified), so the field removal has no schema ripple.
2. **`message/validation`** — `CommitteeInfo` carries `subnet`/`subnetAlan` as **required** constructor parameters (deliberate divergence from boole-fork's zero-fillable struct-literal constructors), fed from the `Committee`/`SSVShare` memoized accessors. The hot path only reads two `uint64`s.

Receive-side topic checks now key off the same subnet source as `BroadcastAtSlot`, so publish and receive cannot diverge.

**Numbers:** post-fork topic check 331ns/6 allocs → 80ns/2 allocs; pre-fork 127ns/5 → 72ns/2 (benchmarks included).

**Test coverage** (the issue's ⚠️ was "correctness couples to every construction site"):
- property tests: cached accessor == independent `commons.*` recompute for random committees, on both `Committee` and `CommitteeInfo`;
- accept/reject through the real entry point, right + wrong topics, pre- and post-fork, for **both** the committee path and the previously-untested validator path (proposer);
- degenerate empty-committee guard-chain proof (`ErrSignerNotInCommittee` fires before topic validation — the unknown-subnet topic is unreachable);
- publish/receive topic parity at the exact fork-boundary slots (last pre-fork + first post-fork slot, live-clock window config);
- `-race` clean; alan topic-string byte-equivalence (`CommitteeTopicID(cid)[0] == SubnetTopicID(AlanCommitteeSubnet(cid))`) relied on and asserted.

### DIFF
--- a/message/validation/validation_bench_test.go
+++ b/message/validation/validation_bench_test.go
@@ -1,0 +1,140 @@
+package validation
+
+import (
+	"math"
+	"math/big"
+	"testing"
+
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/stretchr/testify/require"
+
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+
+	"github.com/ssvlabs/ssv/network/commons"
+	"github.com/ssvlabs/ssv/networkconfig"
+)
+
+// oldValidateTopicAtSlot recomputes the subnet from scratch (SHA-256 over the committee for
+// Boole, big.Int mod for Alan) on every call, mirroring the pre-optimization implementation.
+// It exists only in this test-only benchmark file to give an apples-to-apples before/after
+// comparison against the memoized-subnet validateTopicAtSlot.
+func oldValidateTopicAtSlot(netCfg *networkconfig.Network, committeeID spectypes.CommitteeID, committee []spectypes.OperatorID, topic string, slot phase0.Slot) error {
+	var expectedTopic string
+	if netCfg.BooleForkAtSlot(slot) {
+		expectedTopic = commons.BooleTopic(netCfg.SSV.Name, commons.BooleCommitteeSubnet(committee))
+	} else {
+		expectedTopic = commons.GetTopicFullName(commons.CommitteeTopicID(committeeID)[0])
+	}
+
+	if expectedTopic != topic {
+		e := ErrIncorrectTopic
+		e.got = topic
+		e.want = expectedTopic
+		return e
+	}
+
+	return nil
+}
+
+func benchCommitteeInfoAndNetCfg(boolePostFork bool) (*messageValidator, CommitteeInfo, string, phase0.Slot) {
+	ssvCfg := *networkconfig.TestNetwork.SSV
+	if boolePostFork {
+		ssvCfg.Forks = networkconfig.SSVForks{Boole: 0}
+	} else {
+		ssvCfg.Forks = networkconfig.SSVForks{Boole: math.MaxUint64}
+	}
+	netCfg := &networkconfig.Network{Beacon: networkconfig.TestNetwork.Beacon, SSV: &ssvCfg}
+
+	committee := []spectypes.OperatorID{1, 2, 3, 4}
+	var committeeID spectypes.CommitteeID
+	copy(committeeID[:], big.NewInt(12345).Bytes())
+
+	booleSubnet := commons.BooleCommitteeSubnet(committee)
+	alanSubnet := commons.AlanCommitteeSubnet(committeeID)
+	committeeInfo := newCommitteeInfo(committeeID, committee, nil, booleSubnet, alanSubnet)
+
+	slot := netCfg.FirstSlotAtEpoch(1)
+
+	var topic string
+	if boolePostFork {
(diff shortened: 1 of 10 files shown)


network/commons/testutils.go
```diff
@@ -0,0 +1,14 @@
+package commons
+
+import (
+	"fmt"
+
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+)
+
+// CommitteeTopicID returns the Alan-side committee topic ID. Test-only: production code
```
iurii-ssv: Nit (wording): this file is `testutils.go`, not `*_test.go`, so `CommitteeTopicID` is still compiled into the production `commons` package — it isn't excluded from the build. Keeping it here is the right call, since external test packages (`network/topics`, `message/validation`) can't reference another package's `_test.go` symbols. But 'Test-only' reads like 'not shipped'; consider rewording to something like 'only referenced by tests' to avoid that impression.

registry/storage/validatorstore.go
```diff
@@ -53,12 +55,37 @@ type SelfValidatorStore interface {
...
-	AlanSubnet  uint64
+	ID        spectypes.CommitteeID
+	Operators []spectypes.OperatorID
+	Shares    []*types.SSVShare
+	Indices   []phase0.ValidatorIndex
+
+	// booleCommitteeSubnet is a cached value for committee subnet (post-Boole-fork) so we don't recompute it every time.
+	booleCommitteeSubnet atomic.Pointer[uint64]
+	// alanCommitteeSubnet is a cached value for committee subnet (Alan fork) so we don't recompute it every time.
+	alanCommitteeSubnet atomic.Pointer[uint64]
```
iurii-ssv, line 66: Minor (doc): these two caches have different empty-committee semantics — `BooleCommitteeSubnet()` returns the `UnknownSubnetId` sentinel when `Operators` is empty, whereas `AlanCommitteeSubnet()` derives from `ID` and returns a real subnet regardless. That's actually more correct than the old `buildCommittee` guard (which suppressed both), and it's covered by `TestCommittee_SubnetMemoization_EmptyCommittee`, but a one-line note here would spare a future reader the double-take.

message/validation/validation_bench_test.go
```diff
@@ -0,0 +1,140 @@
...
+			t.Fatal(err)
+		}
+	})
+	oldAllocs := testing.AllocsPerRun(100, func() {
+		if err := oldValidateTopicAtSlot(mv.netCfg, committeeInfo.committeeID, committeeInfo.committee, topic, slot); err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	require.Less(t, newAllocs, oldAllocs)
```
iurii-ssv, line 139: Nit (optional): this asserts `newAllocs < oldAllocs` (relative). Robust today since the old path's SHA-256 reliably allocates, but the invariant you're really guarding is 'the hot path no longer allocates to compute the subnet' — an absolute bound (e.g. `require.LessOrEqual(t, newAllocs, 2)`, matching the '2 allocs' figure in the PR description) expresses that directly and won't drift if the old path's allocations ever change.

iurii-ssv (approved):
Just couple nits.

network/commons/testutils.go
momosh-ssv: Fixed in ed6139766.

Reworded to "Only referenced by tests" so it no longer reads as excluded from the build.

registry/storage/validatorstore.go
momosh-ssv: Fixed in ab819883a.

Added a note on BooleCommitteeSubnet calling out the empty-committee divergence from `AlanCommitteeSubnet`.

message/validation/validation_bench_test.go
momosh-ssv: I'd lean towards keeping the relative check here. 
`oldValidateTopicAtSlot` is a frozen copy inside the test file, so I don't think it can drift on its own. The only thing that could move its allocation count is a Go runtime/stdlib change, and that would shift an absolute bound too (fmt/sha256 alloc counts have changed between Go releases before). 

Seems the relative assertion pins exactly the invariant in the test name without risking spurious breakage on Go upgrades. Happy to add an absolute cap alongside if you feel strongly, though.

ovidiu-ssv-labs (approved):

PR #2969, opened 2026-07-30
eth/eventhandler: deflake slashing-protection stamp assertions

momosh-ssv:
`TestHandleBlockEventsStream` asserted the stored highest-attestation epochs and highest-proposal slot against `EstimatedCurrentSlot()` **re-evaluated at assertion time**. The stamp itself is written during event processing (`BumpSlashingProtectionTxn` reads the clock then), so a slot boundary crossing between stamp and assertion fails the equality spuriously — seen on CI as `expected: 0x64 / actual: 0x65` in the `ClusterLiquidated` subtest ([run](https://github.com/ssvlabs/ssv/actions/runs/29833681337)); the whole test takes ~25s on CI, so with 12s slots a crossing is routine.

Fix: capture the wall-clock slot window around the event processing that writes the stamp, and assert the stored values fall inside `[from, to]` (with `source == target − 1` preserved). Applied to both racy subtests (`ClusterLiquidated`, whose stamp happens back in the second `ValidatorAdded` subtest, and `ClusterReactivated`). The third reactivation subtest's `Greater`-than assertions span a deliberate ~900-slot genesis jump and are race-safe, so they're untouched.

Test-only; no production code changed. Part of a broader CI-flake cleanup (issue to follow).


eth/eventhandler/event_handler_test.go
```diff
@@ -862,13 +870,10 @@ func TestHandleBlockEventsStream(t *testing.T) {
...
 		require.NotNil(t, highestAttestation)
 
-		require.Equal(t, highestAttestation.Source.Epoch, netCfgVarEpoch.EstimatedEpochAtSlot(netCfgVarEpoch.EstimatedCurrentSlot())-1)
-		require.Equal(t, highestAttestation.Target.Epoch, netCfgVarEpoch.EstimatedEpochAtSlot(netCfgVarEpoch.EstimatedCurrentSlot()))
-
 		highestProposal, found, err := eh.keyManager.(*ekm.LocalKeyManager).RetrieveHighestProposal(phase0.BLSPubKey(sharePubKey))
 		require.NoError(t, err)
 		require.True(t, found)
-		require.Equal(t, highestProposal, netCfgVarEpoch.EstimatedCurrentSlot())
+		requireSlashingProtectionStampInWindow(t, netCfgVarEpoch, highestAttestation, highestProposal, validator3StampedFrom, validator3StampedTo)
```
iurii-ssv, line 878: Nit (non-blocking): unlike the ClusterReactivated case below, this window (`validator3StampedFrom`/`To`) is captured two subtests up in "ValidatorAdded again and nonce is bumped", so the assertion's validity here quietly relies on nothing re-bumping validator 3 in between. It does hold — the intervening add targets validator 4, and exit/remove/liquidate don't bump SP — but a one-line pointer here (e.g. "stamped back in the ValidatorAdded subtest; liquidation must not touch it") would help a reader who lands on this assertion first. The var-declaration comment explains the *why-window* but not the *where-stamped*.

eth/eventhandler/event_handler_test.go
```diff
@@ -1689,6 +1694,21 @@ func generateSharesData(validatorData *testValidatorData, operators []*testOpera
...
+// stamped from a slot within [from, to] — the wall-clock window captured around the event
+// processing that wrote it. Comparing against the clock re-read at assertion time is racy:
+// a slot boundary crossing in between fails exact-equality checks spuriously.
+func requireSlashingProtectionStampInWindow(t *testing.T, netCfg *networkconfig.Network, highestAttestation *phase0.AttestationData, highestProposal phase0.Slot, from, to phase0.Slot) {
+	fromEpoch := netCfg.EstimatedEpochAtSlot(from)
+	toEpoch := netCfg.EstimatedEpochAtSlot(to)
+
+	require.GreaterOrEqual(t, highestAttestation.Target.Epoch, fromEpoch)
+	require.LessOrEqual(t, highestAttestation.Target.Epoch, toEpoch)
+	require.Equal(t, highestAttestation.Target.Epoch-1, highestAttestation.Source.Epoch)
```
iurii-ssv: Nit (non-blocking): `highestAttestation.Target.Epoch-1` underflows to a huge `uint64` if `Target.Epoch == 0`, which would turn a legitimate failure into a confusing one rather than a clean assertion diff. Unreachable with the current genesis offsets (target epochs are ~3 and ~31 here) and it faithfully mirrors the original `…-1` check, so purely defensive — `require.Equal(t, highestAttestation.Source.Epoch+1, highestAttestation.Target.Epoch)` expresses the same invariant without the wrap.

iurii-ssv (dismissed):
Two non-blocking nits.

eth/eventhandler/event_handler_test.go
momosh-ssv: Good point.
Added the pointer in 9662e12c5, noting the window comes from the "ValidatorAdded again and nonce is bumped" subtest and that liquidation must not touch it.

eth/eventhandler/event_handler_test.go
momosh-ssv: Agreed, the flipped form is strictly safer. Fixed in 04140cc3a.

nkryuchkov (approved):

y0sher (approved):

Issue #2968, opened 2026-07-30
Boole convergence (#2941) — deferred polish items from full-pass review


momosh-ssv:
Non-blocking items surfaced during the #2941 full-pass review. None are correctness/security bugs (those were fixed in #2941); these are consistency, doc, and drift-risk cleanups worth doing as a batch.

> **Note:** two items originally here are covered in more detail by the reviewer's tracking issue #2978 and have been moved there to avoid duplication — the custom-config `next_domain_type` non-rotation (#2978 item 1) and gating the Boole-topic whitelist on `BooleForkScheduled()` (#2978 item 2). This issue tracks the remainder.

### Exporter API
- `exporter/validator.go:68-75` — a validator-traces range straddling the Boole fork is rejected wholesale: validation evaluates the fork at `request.To` while dispatch is per-slot. A dashboard polling a fixed window starts 400-ing when the window's tail crosses the fork. Decide: evaluate at `From`, or split the range and reject only the post-fork portion.
- `/committee` with no `roles` filter can now return two traces per `(slot, committeeID)` (one per committee runner role), distinguished only by the new `role` field. Additive-field back-compat holds; cardinality back-compat does not. Note it in the OpenAPI description or default the endpoint to `COMMITTEE` unless `roles` is supplied.

### Fork-gating consistency
- `operator/validator/controller.go` — the legacy Alan `RoleAggregator`/`RoleSyncCommitteeContribution` runners are gated on `BooleFork()` evaluated once at share-add time, not per duty slot. Dormant post-fork (harmless), but for ~34 slots after the fork two node populations behave differently on the same pre-fork `RoleAggregator` message. Consider slot-gating for symmetry, or note it retires with the Alan topics.

### Config / misc
- Duplicated aggregator-selection modulo/sha256 logic between `beacon/goclient/aggregator.go` and `protocol/v2/ssv/runner/aggregator.go` — must stay bit-identical or pre/post-Boole aggregator selection diverges. Extract a shared helper.
- No startup sanity check on a garbage custom `forks.boole` epoch (e.g. a typo'd huge-but-not-`MaxUint64` value, or an epoch before genesis) — it silently produces odd fork timing on custom networks. Add a startup validation/warning. (The related `next_domain_type` non-rotation case is #2978 item 1.)
[...]

PR #2971, opened 2026-07-30
network: make the pubsub topic whitelist slot-aware (#2968 item 3)

momosh-ssv:
Item 3 of #2968.

`commons.Topics` gated the Alan topic set on the coarse `BooleFork()` epoch snapshot while subscriptions derive per-slot from the transition-window predicate — so a node **started inside the ~34-slot subsequent window** whitelisted Boole-only yet legitimately subscribes to Alan topics for late pre-fork committee traffic, working only incidentally because `topicsCtrl.Subscribe` self-registers.

`Topics` now takes the evaluation slot explicitly and includes the Alan set whenever Alan traffic can still be valid at that slot (`!BooleForkAtSlot(slot) || InBooleTransitionWindow(slot)`), always including Boole; the sole caller (`NewPubSub`) passes `EstimatedCurrentSlot()`. The startup-snapshot design is unchanged — the snapshot just now matches the subscription policy at the same instant.

Deep review notes (passed): the include-Alan condition is logically identical to all three per-slot subscription-derivation sites in `network/p2p`; the whitelist gates subscriptions/peer-announcements (mesh membership), not message validation, and the widening is confined to the subsequent window where the fork design requires Alan traffic to flow; a node started past the window gets Boole-only with no Alan leak (`Boole==0` guard verified). Five-phase table test covers pre-fork steady state, prior window, subsequent window (the bug), the exact first slot past the window (the boundary that closes Alan), and well post-fork.

### DIFF
--- a/network/commons/subnets_test.go
+++ b/network/commons/subnets_test.go
@@ -214,11 +214,11 @@ func TestDiffSubnets(t *testing.T) {
 	require.Equal(t, 128-62, added.ActiveCount()+removed.ActiveCount())
 }
 
-// TestTopics covers the four Boole-transition phases the pubsub startup whitelist must track in
+// TestTopics covers the five Boole-transition phases the pubsub startup whitelist must track in
 // lockstep with per-slot subscription derivation (network/p2p's subscribedSubnetsForCurrentEpoch):
 // pre-fork steady state, the prior window, the subsequent window (the bug this guards against —
 // a node started in this window must still whitelist Alan topics for late pre-fork committee
-// traffic), and well post-fork.
+// traffic), the first slot past the subsequent window, and well post-fork.
 func TestTopics(t *testing.T) {
 	booleCfg := func(booleEpoch phase0.Epoch) *networkconfig.Network {
 		cfg := *networkconfig.TestNetwork
@@ -300,6 +300,8 @@ func TestTopics(t *testing.T) {
 	t.Run("first slot past the subsequent window is Boole-only", func(t *testing.T) {
 		cfg := booleCfg(cur + 1)
 		// One slot past the window end [F, F+SlotsPerEpoch+2) — the boundary that closes Alan.
+		// The +2 mirrors networkconfig's unexported booleSubsequentWindowLateSlots; if that margin
+		// ever changes, the InBooleTransitionWindow assertion below fails loudly.
 		slot := cfg.FirstSlotAtEpoch(cur+1) + phase0.Slot(cfg.SlotsPerEpoch) + 2
 		require.True(t, cfg.BooleForkAtSlot(slot))
 		require.False(t, cfg.InBooleTransitionWindow(slot), "expected slot just past the subsequent window")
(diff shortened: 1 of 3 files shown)


network/commons/subnets_test.go
```diff
@@ -210,6 +214,113 @@ func TestDiffSubnets(t *testing.T) {
 	require.Equal(t, 128-62, added.ActiveCount()+removed.ActiveCount())
 }
 
+// TestTopics covers the four Boole-transition phases the pubsub startup whitelist must track in
```
iurii-ssv: Nit: the comment says *four* Boole-transition phases, but `TestTopics` has five subtests — the `first slot past the subsequent window is Boole-only` boundary case is a fifth. Suggest updating to *five* to match the PR description. (Minor: these are sequential `t.Run` subtests rather than a table-driven test, if you also want to reconcile the description's "table test" wording.)

network/commons/subnets_test.go
```diff
@@ -210,6 +214,113 @@ func TestDiffSubnets(t *testing.T) {
...
+		for topic := range booleTopicsSet(cfg.SSV.Name) {
+			want[topic] = struct{}{}
+		}
+		require.Equal(t, want, got)
+	})
+
+	t.Run("first slot past the subsequent window is Boole-only", func(t *testing.T) {
+		cfg := booleCfg(cur + 1)
+		// One slot past the window end [F, F+SlotsPerEpoch+2) — the boundary that closes Alan.
+		slot := cfg.FirstSlotAtEpoch(cur+1) + phase0.Slot(cfg.SlotsPerEpoch) + 2
```
iurii-ssv, line 305: Nit: this `+ 2` — and the `[F, F+SlotsPerEpoch+2)` in the comment just above — hard-codes the `booleSubsequentWindowLateSlots` margin, which is unexported in `networkconfig` and so can't be referenced from here. It's self-guarding: if that margin ever changed, the `require.False(cfg.InBooleTransitionWindow(slot), ...)` on the next lines would fail loudly rather than pass silently. Still, a one-line comment tying this literal back to `booleSubsequentWindowLateSlots` would save a future reader the cross-package hunt.

iurii-ssv (dismissed):

network/commons/subnets_test.go
momosh-ssv: Fixed in 86489a4ee, the comment now says five and names the past-the-window boundary case explicitly.

network/commons/subnets_test.go
momosh-ssv: Fixed in 86489a4ee.
Added a comment tying the +2 to networkconfig's unexported booleSubsequentWindowLateSlots and noting the InBooleTransitionWindow assertion is the loud tripwire if the margin changes.

 Kept the literal rather than exporting the constant, seems not worth widening the API for a test.

y0sher (approved):

ljuba-ssv (approved):

PR #2972, opened 2026-07-30
networkconfig: single-source the aggregator-selection formula (#2968 item 5)

momosh-ssv:
Item 5 of #2968. Pure refactor — zero behavioral change.

`beacon/goclient` (pre-fork path) and `protocol/v2/ssv/runner` (post-fork AggregatorCommittee path) each hand-rolled the same selection computation — sha256(slotSig), first 8 bytes little-endian, mod `max(1, committeeCount/targetAggregatorsPerCommittee)`, `== 0` — which must stay bit-identical or pre/post-Boole aggregator selection diverges (a consensus-level split on who aggregates).

Extract `networkconfig.IsAggregatorSelected` and delegate both sites. The runner keeps its mockable `IsAggregator` field (default now wraps the shared helper); the pooled hasher is dropped (no remaining users; selection runs once per aggregator duty, not per message).

Deep review notes (passed): byte-for-byte equivalence verified in both directions against the pre-change implementations (hash input, digest slicing, endianness, clamp-before-mod order, truncating division); with `AggregatorCommitteeRunner` already routing through `goclient.IsAggregator`, **every** selection path now converges on the single helper, making drift structurally impossible; test seam (`protocol/v2/ssv/testing/runner.go` overrides) intact; parity test covers the clamp-to-1 case.

### DIFF
--- a/networkconfig/aggregator.go
+++ b/networkconfig/aggregator.go
@@ -6,7 +6,7 @@ import (
 )
 
 // IsAggregatorSelected returns true if the given slot signature selects the validator as an
-// aggregator for a committee of committeeCount, given targetAggregatorsPerCommittee.
+// aggregator for a committee of committeeLength members, given targetAggregatorsPerCommittee.
 //
 // This is the shared implementation of the beacon-chain aggregator-selection check, used by both
 // beacon/goclient and protocol/v2/ssv/runner. It must stay bit-identical across pre/post-Boole
@@ -18,8 +18,8 @@ import (
 //	 committee = get_beacon_committee(state, slot, index)
 //	 modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
 //	 return bytes_to_uint64(hash(slot_signature)[0:8]) % modulo == 0
-func IsAggregatorSelected(targetAggregatorsPerCommittee, committeeCount uint64, slotSig []byte) bool {
-	modulo := committeeCount / targetAggregatorsPerCommittee
+func IsAggregatorSelected(targetAggregatorsPerCommittee, committeeLength uint64, slotSig []byte) bool {
+	modulo := committeeLength / targetAggregatorsPerCommittee
 	if modulo == 0 {
 		// Modulo must be at least 1.
 		modulo = 1
--- a/networkconfig/aggregator.go
+++ b/networkconfig/aggregator.go
@@ -1,0 +1,30 @@
+package networkconfig
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+)
+
+// IsAggregatorSelected returns true if the given slot signature selects the validator as an
+// aggregator for a committee of committeeCount, given targetAggregatorsPerCommittee.
+//
+// This is the shared implementation of the beacon-chain aggregator-selection check, used by both
+// beacon/goclient and protocol/v2/ssv/runner. It must stay bit-identical across pre/post-Boole
+// call sites.
+//
+// Spec pseudocode definition:
+//
+//	def is_aggregator(state: BeaconState, slot: Slot, index: CommitteeIndex, slot_signature: BLSSignature) -> bool:
+//	 committee = get_beacon_committee(state, slot, index)
+//	 modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
+//	 return bytes_to_uint64(hash(slot_signature)[0:8]) % modulo == 0
+func IsAggregatorSelected(targetAggregatorsPerCommittee, committeeCount uint64, slotSig []byte) bool {
+	modulo := committeeCount / targetAggregatorsPerCommittee
+	if modulo == 0 {
+		// Modulo must be at least 1.
+		modulo = 1
+	}
+
+	h := sha256.Sum256(slotSig)
+	return binary.LittleEndian.Uint64(h[:8])%modulo == 0
+}
--- a/networkconfig/aggregator.go
+++ b/networkconfig/aggregator.go
@@ -18,11 +18,16 @@ import (
 //	 committee = get_beacon_committee(state, slot, index)
 //	 modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
 //	 return bytes_to_uint64(hash(slot_signature)[0:8]) % modulo == 0
+//
+// A zero targetAggregatorsPerCommittee (never the case in practice — the spec constant is 16)
+// is clamped to modulo 1 rather than panicking on division.
 func IsAggregatorSelected(targetAggregatorsPerCommittee, committeeLength uint64, slotSig []byte) bool {
-	modulo := committeeLength / targetAggregatorsPerCommittee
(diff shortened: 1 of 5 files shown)


networkconfig/aggregator.go
```diff
@@ -0,0 +1,30 @@
...
+// beacon/goclient and protocol/v2/ssv/runner. It must stay bit-identical across pre/post-Boole
+// call sites.
+//
+// Spec pseudocode definition:
+//
+//	def is_aggregator(state: BeaconState, slot: Slot, index: CommitteeIndex, slot_signature: BLSSignature) -> bool:
+//	 committee = get_beacon_committee(state, slot, index)
+//	 modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
+//	 return bytes_to_uint64(hash(slot_signature)[0:8]) % modulo == 0
+func IsAggregatorSelected(targetAggregatorsPerCommittee, committeeCount uint64, slotSig []byte) bool {
```
iurii-ssv: **Naming: `committeeCount` is a misnomer — consider `committeeLength` (or `committeeSize`).**

This helper is now the single source of truth, and both callers pass a committee *length* (`len(committee)` in the spec pseudocode just above), not a count of committees:

- the runner passes [`duty.CommitteeLength`](https://github.com/ssvlabs/ssv/blob/32b3904a1e8ebab07af391c840057aef7c67b26b/protocol/v2/ssv/runner/aggregator.go#L160)
- `AggregatorCommitteeRunner` passes [`vDuty.CommitteeLength`](https://github.com/ssvlabs/ssv/blob/32b3904a1e8ebab07af391c840057aef7c67b26b/protocol/v2/ssv/runner/aggregator_committee.go#L475)
- `beacon/goclient`'s own wrapper even names the argument [`committeeLength`](https://github.com/ssvlabs/ssv/blob/32b3904a1e8ebab07af391c840057aef7c67b26b/beacon/goclient/aggregator.go#L24)

So as written, goclient forwards its `committeeLength` straight into a parameter named `committeeCount`. Renaming here matches the `len(committee)` in the doc comment and removes the inconsistency; ideally the runner's mockable [`IsAggregator` field signature](https://github.com/ssvlabs/ssv/blob/32b3904a1e8ebab07af391c840057aef7c67b26b/protocol/v2/ssv/runner/aggregator.go#L55) is aligned too. Pure naming — no behavioral change.

networkconfig/aggregator.go
```diff
@@ -0,0 +1,30 @@
...
+	modulo := committeeCount / targetAggregatorsPerCommittee
```
iurii-ssv: **Very minor / pre-existing:** this division panics if `targetAggregatorsPerCommittee == 0`. In practice the target is a spec constant (default 16) and never zero, and both prior implementations shared this property, so it's not a regression. But now that the value arrives as a free parameter on the canonical helper, a one-line guard (or a documented precondition on the exported function) would harden the single source. Non-blocking.

protocol/v2/ssv/runner/aggregator_test.go
```diff
@@ -31,6 +34,42 @@ func Test_isAggregatorFn(t *testing.T) {
 	}
 }
 
+// Test_isAggregatorFn_MatchesSharedHelper asserts the runner's default IsAggregator func is a
+// thin, bit-identical wrapper over the shared networkconfig.IsAggregatorSelected helper — the
+// same helper beacon/goclient.GoClient.IsAggregator delegates to — across a spread of committee
+// sizes (including committeeCount < target, which clamps modulo to 1).
+func Test_isAggregatorFn_MatchesSharedHelper(t *testing.T) {
```
iurii-ssv, line 46: **This assertion is currently a tautology.** `isAggregatorFn()` now returns `networkconfig.IsAggregatorSelected` verbatim, so `got := isAggFn(...)` and `want := networkconfig.IsAggregatorSelected(...)` invoke the *same* function value — `require.Equal` can never fail here.

Its only real value is as a future regression guard (catching a change that stops the runner delegating to the shared helper). That's legitimate, but worth either a one-line comment stating that intent explicitly, or dropping the test given the delegation is now a single-line `return`.

networkconfig/aggregator_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/binary"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// computeExpectedIsAggregatorSelected is an independent recompute of the aggregator-selection
+// formula, mirroring the documented spec pseudocode so a bug in IsAggregatorSelected (e.g. wrong
+// endianness or hash) fails this test rather than silently drifting.
+func computeExpectedIsAggregatorSelected(targetAggregatorsPerCommittee, committeeCount uint64, slotSig []byte) bool {
```
iurii-ssv: **Optional test strengthening.** `computeExpectedIsAggregatorSelected` is a line-for-line reimplementation of the production formula, so `TestIsAggregatorSelected` asserts that two identical copies agree — it catches a typo in one copy but not a conceptual error present in both (e.g. wrong endianness or slice bounds replicated in each). Consider adding a handful of golden vectors (fixed `slotSig` + params → hardcoded expected `bool`) to pin actual known-good output. Optional.

iurii-ssv (commented):
Some minor suggestions.

One subjective note not worth its own thread: networkconfig is a slightly odd home for a pure beacon-spec function that never reads a NetworkConfig — a beacon/spec-helpers package might read more naturally — but it's cycle-free and both callers already import it, so it's defensible.

networkconfig/aggregator.go
momosh-ssv: Fixed in af9abfc3c.
Renamed to `committeeLength` across the helper, the runner's mockable `IsAggregator` field + `isAggregatorFn`, and the tests, so goclient's `committeeLength` no longer forwards into a mismatched name.

networkconfig/aggregator.go
momosh-ssv: Fixed in 9c3465159.
Added the guard: a zero target now clamps to modulo 1 rather than panicking, failing open to "selected" so aggregation duties are still attempted under a misconfiguration. 

Documented on the exported function and covered with a test case.

protocol/v2/ssv/runner/aggregator_test.go
momosh-ssv: Fixed in b348f7b4d.
Kept the test since pinning the delegation is exactly what this PR is for, and added an explicit comment stating it exists purely as a regression guard against the runner's default diverging from the shared helper.

networkconfig/aggregator_test.go
momosh-ssv: Fixed in 28fa276ff.
Added `TestIsAggregatorSelected_GoldenVectors` with hardcoded known-good outputs (both selected and not-selected outcomes across moduli 1, 3, 8, 100, plus the zero-target clamp), and reworded the mirror's doc comment so it no longer overclaims catching shared conceptual errors.

y0sher (approved):

iurii-ssv (approved):

PR #2973, opened 2026-07-30
boole-convergence polish: gating docs, startup fork validation, spectest bound (#2968 items 4, 6, 7)

momosh-ssv:
Items 4, 6 and 7 of #2968, one commit each:

- **Item 4 (docs only)** — the `BooleFork()` gate on the legacy Alan `RoleAggregator`/`RoleSyncCommitteeContribution` runners is now documented as a deliberate share-add-time snapshot: the cross-population asymmetry is bounded to the Boole subsequent window (`SlotsPerEpoch + booleSubsequentWindowLateSlots`; 34 slots on mainnet), affects only already-decided pre-fork stragglers, and the runners retire with the Alan topics. Decision recorded on the issue: not worth restructuring the one-shot runner-map construction to slot-gate a dormant path.
- **Item 6** — `Network.Validate()`, called right after the `Network` is assembled at startup: hard error when a scheduled `forks.boole` epoch would overflow `FirstSlotAtEpoch` (or `SlotsPerEpoch` is zero — the review caught that the overflow guard alone would panic on that same malformed-config class); operator-actionable warning when Boole is scheduled but `NextDomainType == DomainType` (the exact signature of a forgotten `next_domain_type`, which would "activate" the fork with zero observable domain change).
- **Item 7** — the alan-spec error-code remap bound is derived from `TimeoutInstanceErrorCode + 2` (= 72, the real v1.2.2 maximum) instead of the disconnected magic 79, so a spec bump adding codes can't silently widen the remap. Builds under `-tags alan_spec` (that leg doesn't run in normal CI).

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -105,12 +105,8 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
 		SSV:    ssvNetworkConfig,
 		Beacon: consensusClient.BeaconConfig(),
 	}
-	warnings, err := networkConfig.Validate()
-	if err != nil {
+	if err := networkConfig.Validate(); err != nil {
 		return nil, fmt.Errorf("invalid network config: %w", err)
-	}
-	for _, warning := range warnings {
-		logger.Warn(warning)
 	}
 
 	var executionAddrList []string
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -105,6 +105,13 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
 		SSV:    ssvNetworkConfig,
 		Beacon: consensusClient.BeaconConfig(),
 	}
+	warnings, err := networkConfig.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("invalid network config: %w", err)
+	}
+	for _, warning := range warnings {
+		logger.Warn(warning)
+	}
 
 	var executionAddrList []string
 	for _, addr := range strings.Split(cfg.ExecutionClient.Addr, ";") {
--- a/networkconfig/network.go
+++ b/networkconfig/network.go
@@ -94,10 +94,7 @@ func (n Network) Validate() error {
 	}
 
 	if n.BooleForkScheduled() {
-		// Mirrors the overflow guard in inBooleSubsequentWindowWithSlots: FirstSlotAtEpoch(Boole)
-		// would overflow if the fork epoch is beyond the representable slot range.
-		maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
-		if n.SSV.Forks.Boole > maxEpoch {
+		if maxEpoch := n.maxEpochConvertibleToSlot(); n.SSV.Forks.Boole > maxEpoch {
 			return fmt.Errorf("boole fork epoch %d overflows slot conversion (max supported epoch %d)", n.SSV.Forks.Boole, maxEpoch)
 		}
 
@@ -121,6 +118,13 @@ func (n Network) Validate() error {
 	}
 
 	return nil
+}
+
+// maxEpochConvertibleToSlot returns the highest epoch whose first slot still fits the slot
+// range; FirstSlotAtEpoch would overflow beyond it. Callers must rule out a zero SlotsPerEpoch
+// first, or the division panics.
+func (n Network) maxEpochConvertibleToSlot() phase0.Epoch {
+	return phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
 }
 
 // InBooleTransitionWindow checks if the slot is in the Boole transition window,
@@ -162,8 +166,7 @@ func (n Network) inBooleSubsequentWindowWithSlots(slot phase0.Slot, windowSlots 
 
 	// Avoid FirstSlotAtEpoch overflow when Boole is beyond the representable epoch range;
 	// without this guard the multiplication would wrap and could treat small slots as in-window.
-	maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
-	if n.SSV.Forks.Boole > maxEpoch {
+	if n.SSV.Forks.Boole > n.maxEpochConvertibleToSlot() {
 		return false
 	}
 
--- a/networkconfig/network.go
+++ b/networkconfig/network.go
@@ -103,13 +103,17 @@ func (n Network) Validate() (warnings []string, err error) {
 		}
 
 		// unmarshalFromConfig defaults NextDomainType to DomainType when the field is absent, so
-		// equality here is exactly the signature of a scheduled fork that forgot to set
-		// NextDomainType: it would "activate" with zero observable domain change.
-		if n.NextDomainType == n.DomainType {
-			warnings = append(warnings, fmt.Sprintf(
+		// equality while the fork is still ahead is exactly the signature of a scheduled fork that
+		// forgot to set NextDomainType: it would "activate" with zero observable domain change, and
+		// a restart is guaranteed before activation (see BooleForkScheduled), so refusing to start
+		// is the last safe moment to catch it. Once the fork is active, equality is legitimate
+		// steady state — a config written post-fork may set DomainType to the post-fork domain and
+		// omit NextDomainType.
+		if n.NextDomainType == n.DomainType && !n.BooleFork() {
+			return nil, fmt.Errorf(
 				"boole fork is scheduled at epoch %d but NextDomainType equals DomainType: the fork would activate with no observable domain change",
 				n.SSV.Forks.Boole,
-			))
+			)
 		}
 	}
 
--- a/networkconfig/network.go
+++ b/networkconfig/network.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"time"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	spectypes "github.com/ssvlabs/ssv-spec/types"
@@ -109,7 +110,11 @@ func (n Network) Validate() (warnings []string, err error) {
 		// is the last safe moment to catch it. Once the fork is active, equality is legitimate
 		// steady state — a config written post-fork may set DomainType to the post-fork domain and
 		// omit NextDomainType.
-		if n.NextDomainType == n.DomainType && !n.BooleFork() {
+		// The wall-clock read must be guarded: EstimatedCurrentSlot panics while the clock is
+		// still before GenesisTime, and before genesis a scheduled fork is by definition not
+		// yet active.
+		booleForkActive := !time.Now().Before(n.GenesisTime) && n.BooleFork()
+		if n.NextDomainType == n.DomainType && !booleForkActive {
 			return nil, fmt.Errorf(
 				"boole fork is scheduled at epoch %d but NextDomainType equals DomainType: the fork would activate with no observable domain change",
 				n.SSV.Forks.Boole,
--- a/networkconfig/network.go
+++ b/networkconfig/network.go
@@ -83,16 +83,14 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
 }
 
 // Validate checks the assembled network configuration for internal inconsistencies that fork
-// activation logic can't catch on its own. It returns non-fatal warnings for configurations that
-// are almost certainly a mistake, and an error for configurations that would panic or misbehave.
-// It stays logger-free by design: the caller decides how to surface warnings (e.g. logger.Warn)
-// and whether/how to abort on error.
-func (n Network) Validate() (warnings []string, err error) {
+// activation logic can't catch on its own, returning an error for configurations that would
+// panic or misbehave. It stays logger-free by design: the caller decides whether/how to abort.
+func (n Network) Validate() error {
 	// A zero SlotsPerEpoch is a malformed config regardless of fork scheduling: slot/epoch
 	// conversions divide by it unconditionally (e.g. Beacon.EstimatedEpochAtSlot), so it would
 	// panic on the first duty long before any fork logic runs.
 	if n.SlotsPerEpoch == 0 {
-		return nil, fmt.Errorf("slots per epoch must be positive")
+		return fmt.Errorf("slots per epoch must be positive")
 	}
 
 	if n.BooleForkScheduled() {
@@ -100,7 +98,7 @@ func (n Network) Validate() (warnings []string, err error) {
 		// would overflow if the fork epoch is beyond the representable slot range.
 		maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
 		if n.SSV.Forks.Boole > maxEpoch {
-			return nil, fmt.Errorf("boole fork epoch %d overflows slot conversion (max supported epoch %d)", n.SSV.Forks.Boole, maxEpoch)
+			return fmt.Errorf("boole fork epoch %d overflows slot conversion (max supported epoch %d)", n.SSV.Forks.Boole, maxEpoch)
 		}
 
 		// unmarshalFromConfig defaults NextDomainType to DomainType when the field is absent, so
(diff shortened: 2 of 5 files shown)


operator/validator/controller.go
```diff
@@ -1235,6 +1244,15 @@ func SetupRunners(
...
 			// Post-Boole, sync committee contribution duties route through the merged
 			// AggregatorCommitteeRunner (committee-scoped) instead of this legacy per-validator runner.
+			//
+			// Gating on BooleFork() here is a deliberate snapshot taken at share-add time: once a
+			// share is added, this runner set is fixed until the share is re-added, so a share added
+			// shortly before the fork keeps its legacy runner across the transition. The resulting
+			// cross-population asymmetry (some operators still running the legacy runner while others
+			// have moved to the committee runner) is bounded to the Boole subsequent window
+			// (SlotsPerEpoch + booleSubsequentWindowLateSlots; 34 slots on mainnet — see InBooleTransitionWindow)
+			// and only affects already-decided pre-fork aggregator stragglers — no correctness impact.
```
iurii-ssv: Copy-paste from the `RoleAggregator` block: this is the `RoleSyncCommitteeContribution` case, so 'aggregator stragglers' is imprecise — these are sync-committee-contribution duties. Suggest 'sync committee contribution stragglers' (or 'aggregator/contribution stragglers'). Doc-only nit; the identical line at 1233 in the aggregator block is correct as-is.

protocol/v2/qbft/spectest/error_code_map_alan.go
```diff
@@ -4,6 +4,13 @@ package qbft
 
 import spectypes "github.com/ssvlabs/ssv-spec/types"
 
+// maxShiftableLegacyErrorCode is the highest v1.2.2 error code eligible for the -2 shift below.
+// spectypes.TimeoutInstanceErrorCode is the last error code that existed in the spec before the
+// Boole-era additions (AggComm*/Committee* codes), which have no v1.2.2 counterpart and thus were
+// never subject to the shift. Its v1.2.2 value is its current (v1.2.3+) value plus the 2-code
+// offset described below: 70 + 2 = 72.
+const maxShiftableLegacyErrorCode = int(spectypes.TimeoutInstanceErrorCode) + 2
```
iurii-ssv, line 12: Minor: the `int(...)` conversion is redundant — the spec's `error.go` declares the enum as `UnmarshalSSZErrorCode int = iota + 1`, so `TimeoutInstanceErrorCode` is already typed `int`. Harmless/defensive, fine to keep. The derivation itself is correct: `TimeoutInstanceErrorCode == 70`, so the bound is 72 and every `AggComm*`/`Committee*` code sits above it — the old `79` only affected inputs 73–79, which never existed in v1.2.2 numbering, so the change is behavior-preserving.

networkconfig/network_test.go
```diff
@@ -202,6 +202,83 @@ func TestBooleForkAtSlot(t *testing.T) {
 	}
 }
 
+func TestNetworkValidate(t *testing.T) {
```
iurii-ssv, line 205: Good boundary coverage (zero-slots, overflow at `maxEpoch+1`, the warning, clean scheduled/unscheduled). Two optional cases to fully close the gaps: (1) `Boole == 0` (genesis fork) — untested, and `BooleForkScheduled()` treats 0 as scheduled, so it hits a distinct path; (2) the exact `Boole == maxEpoch` boundary asserting no error (the overflow case only covers `maxEpoch+1`).

iurii-ssv (dismissed):
A few minor, non-blocking nits.

operator/validator/controller.go
momosh-ssv: Fixed in 8e85bcce2, reworded to "sync committee contribution stragglers" in the SCC block.

protocol/v2/qbft/spectest/error_code_map_alan.go
momosh-ssv: You're right that it's a no-op today.

Keeping it deliberately, though: if a future spec release ever names the enum type (`type ErrorCode int`), the bare constant would make `maxShiftableLegacyErrorCode` that named type and the comparison against `code int` would stop compiling. 

The explicit conversion keeps this file immune to that. Happy to drop it if you feel it's noise.

networkconfig/network_test.go
momosh-ssv: Both added in 0a6d483f4 .
`clean_genesis_fork` (Boole == 0, goes through the full scheduled-validation branch) and `clean_at_max_epoch_boundary` (Boole == MaxUint64/32, asserting no error at the exact boundary).

networkconfig/network.go
```diff
@@ -81,6 +81,41 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
...
+// It stays logger-free by design: the caller decides how to surface warnings (e.g. logger.Warn)
+// and whether/how to abort on error.
+func (n Network) Validate() (warnings []string, err error) {
+	if n.BooleForkScheduled() {
+		// Guard the division below the way inBooleSubsequentWindowWithSlots does — but as a hard
+		// error: a zero SlotsPerEpoch is exactly the malformed-config class Validate exists to
+		// catch, and would otherwise panic here.
+		if n.SlotsPerEpoch == 0 {
+			return nil, fmt.Errorf("slots per epoch must be positive when a boole fork is scheduled")
+		}
```
nkryuchkov: I think we should always check this, not only if the Boole fork is scheduled. Setting this to 0 is a misconfiguration

momosh-ssv: Good point, agreed.
Fixed in 4b8bf87bb. Hoisted the check above the scheduled-fork branch so it always runs (EstimatedEpochAtSlot divides by SlotsPerEpoch unconditionally anyway), and added a test case covering the unscheduled-fork path.

networkconfig/network.go
```diff
@@ -81,6 +81,41 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
...
+		// would overflow if the fork epoch is beyond the representable slot range.
+		maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
+		if n.SSV.Forks.Boole > maxEpoch {
+			return nil, fmt.Errorf("boole fork epoch %d overflows slot conversion (max supported epoch %d)", n.SSV.Forks.Boole, maxEpoch)
+		}
+
+		// unmarshalFromConfig defaults NextDomainType to DomainType when the field is absent, so
+		// equality here is exactly the signature of a scheduled fork that forgot to set
+		// NextDomainType: it would "activate" with zero observable domain change.
+		if n.NextDomainType == n.DomainType {
```
y0sher: shouldn't the domain be updated on a fork? this sounds like an error not warning.

y0sher: unless the fork has already happened

momosh-ssv: Agreed👍

Fixed in 0fe941fea: equality is now a startup error while the scheduled fork is still ahead (a restart ships before activation, so that's the last safe moment to catch a forgotten NextDomainType), and it passes clean once the fork is already active, since a post-fork config may legitimately collapse both fields to the new domain. Covered both sides with tests.

y0sher (dismissed):
lgtm, ut failing.

networkconfig/network.go
```diff
@@ -81,6 +81,45 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
 	return n.BooleForkAtEpoch(n.EstimatedEpochAtSlot(slot))
 }
 
+// Validate checks the assembled network configuration for internal inconsistencies that fork
+// activation logic can't catch on its own. It returns non-fatal warnings for configurations that
+// are almost certainly a mistake, and an error for configurations that would panic or misbehave.
+// It stays logger-free by design: the caller decides how to surface warnings (e.g. logger.Warn)
+// and whether/how to abort on error.
+func (n Network) Validate() (warnings []string, err error) {
```
iurii-ssv: The `warnings []string` return is never populated: `Validate()` only ever returns `nil` warnings or an `err`. As a result the `for _, warning := range warnings { logger.Warn(...) }` loop in `cli/operator/node.go` never runs, and `expectedWarnings` in the new test is 0 in every case (so `require.Len(t, warnings, ...)` only ever asserts "empty").

The error-vs-warning decision for the `NextDomainType == DomainType` case was already settled as an error (see y0sher's thread) — that part is fine. But it leaves the whole warnings channel (this return value, the node.go loop, the test field) vestigial. Either wire up a real warning case and assert `expectedWarnings: 1`, or drop the unused plumbing so the signature doesn't advertise a code path that can't happen.

networkconfig/network.go
```diff
@@ -81,6 +81,45 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
...
+		}
+
+		// unmarshalFromConfig defaults NextDomainType to DomainType when the field is absent, so
+		// equality while the fork is still ahead is exactly the signature of a scheduled fork that
+		// forgot to set NextDomainType: it would "activate" with zero observable domain change, and
+		// a restart is guaranteed before activation (see BooleForkScheduled), so refusing to start
+		// is the last safe moment to catch it. Once the fork is active, equality is legitimate
+		// steady state — a config written post-fork may set DomainType to the post-fork domain and
+		// omit NextDomainType.
+		if n.NextDomainType == n.DomainType && !n.BooleFork() {
```
iurii-ssv: `!n.BooleFork()` reaches `Beacon.EstimatedSlotAtTime(time.Now())`, which panics when `GenesisTime` is in the future. `Validate()` is documented to return an error (not panic) for malformed configs, yet for a not-yet-genesis network with a scheduled Boole and `NextDomainType == DomainType`, this line panics instead of returning the intended error.

The `&&` short-circuit spares the common distinct-domain path, so exposure is narrow (scheduled fork + equal domains + pre-genesis beacon) — but that's exactly the misconfiguration this check exists to catch cleanly. Consider determining "fork not yet active" without the wall clock, or guarding the future-genesis case.

networkconfig/network.go
```diff
@@ -81,6 +81,45 @@ func (n Network) BooleForkAtSlot(slot phase0.Slot) bool {
...
+	// conversions divide by it unconditionally (e.g. Beacon.EstimatedEpochAtSlot), so it would
+	// panic on the first duty long before any fork logic runs.
+	if n.SlotsPerEpoch == 0 {
+		return nil, fmt.Errorf("slots per epoch must be positive")
+	}
+
+	if n.BooleForkScheduled() {
+		// Mirrors the overflow guard in inBooleSubsequentWindowWithSlots: FirstSlotAtEpoch(Boole)
+		// would overflow if the fork epoch is beyond the representable slot range.
+		maxEpoch := phase0.Epoch(math.MaxUint64 / n.SlotsPerEpoch)
```
iurii-ssv: Minor (maintainability): this `maxEpoch := math.MaxUint64 / n.SlotsPerEpoch` overflow guard is duplicated verbatim in `inBooleSubsequentWindowWithSlots` below. A small shared helper (e.g. `booleEpochOverflowsSlotRange()`) would keep the two in lockstep — the cross-reference comment helps but doesn't prevent drift.

cli/operator/node.go
```diff
@@ -105,6 +105,13 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
 		SSV:    ssvNetworkConfig,
 		Beacon: consensusClient.BeaconConfig(),
 	}
+	warnings, err := networkConfig.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("invalid network config: %w", err)
```
iurii-ssv, line 109: Nit: the sibling failure paths in `buildNode` return `startupError{err, fields}` to attach structured zap fields; this one returns a bare `fmt.Errorf`. Minor local-convention inconsistency — fine if intentional.

iurii-ssv (commented):
A bunch of small suggestions.

momosh-ssv:
> ut failing

It's a flake ❄ will fix in followups.
Thanks for the reviews, guys

networkconfig/network.go
momosh-ssv: Fixed in 7c32997aa — dropped the plumbing. With the equality case settled as an error there was no remaining warning producer, so Validate now returns a plain error and the node.go loop + test field are gone.

networkconfig/network.go
momosh-ssv: Nice catch, thanks. Fixed in 49c492755 — the wall-clock read is now guarded: before genesis a scheduled fork is by definition not yet active, so the check reports the misconfiguration as a proper error instead of panicking. Added a pre-genesis test case covering exactly that path.

networkconfig/network.go
momosh-ssv: Fixed in 401a1cc76 — extracted a shared maxEpochConvertibleToSlot() helper used by both sites, so the bound cannot drift. Kept it value-returning (rather than a bool predicate) so Validate error message can still report the max supported epoch.

cli/operator/node.go
momosh-ssv: It's intentional — seems the split in buildNode is that startupError carries a structured field worth preserving (the beacon address on the goclient paths), while the config-shaped failures nearby (setupSSVNetwork, the empty execution-address check) return bare errors. There's no obvious field to attach here, so I leaned on the latter. Happy to wrap it if you'd rather have the consistency.

networkconfig/network_test.go
```diff
@@ -202,6 +202,131 @@ func TestBooleForkAtSlot(t *testing.T) {
...
+			nextDomainType: domainAlan,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			beacon := *TestNetwork.Beacon
+			beacon.SlotsPerEpoch = test.slotsPerEpoch
+			slotsSinceGenesis := uint64(test.currentEpoch) * test.slotsPerEpoch
+			beacon.GenesisTime = time.Now().Add(-time.Duration(slotsSinceGenesis) * beacon.SlotDuration)
```
iurii-ssv: Nit (optional, maintainability) — non-blocking: this per-case beacon setup duplicates `beaconAtEpoch` just below (same `*TestNetwork.Beacon` copy + `GenesisTime = now - slotsSinceGenesis*SlotDuration`). It can't reuse `beaconAtEpoch` as-is because this test varies `SlotsPerEpoch` (including 0) and needs the `preGenesis` override — but a small `beaconAt(epoch, slotsPerEpoch)` helper (keeping the pre-genesis branch here) would fold the genesis math back into one place. Fine to leave if you'd rather not churn the helper.

networkconfig/network_test.go
```diff
@@ -202,6 +202,131 @@ func TestBooleForkAtSlot(t *testing.T) {
...
+			preGenesis:     true,
+			slotsPerEpoch:  32,
+			domainType:     domainAlan,
+			nextDomainType: domainAlan,
+			expectErr:      true,
+		},
+		{
+			name:           "clean_next_domain_equals_domain_after_fork",
+			boole:          10,
+			currentEpoch:   20,
```
iurii-ssv, line 277: Nit (optional, coverage) — non-blocking: the domain-equality cases cover `currentEpoch` before the fork (`before_fork` = 5) and well after (`after_fork` = 20), but not the exact activation epoch `currentEpoch == boole` (= 10), where `BooleFork()` flips on `>=`. This is a different axis from the fork-*value* boundaries already added (`clean_genesis_fork`, `clean_at_max_epoch_boundary`) — it's the wall-clock activation edge. Low value, since `TestBooleForkAtSlot`'s `at_fork` case already pins the `>=` semantics, so genuinely fine to skip; flagging only to fully close the transition edge.

iurii-ssv (dismissed):
LGTM (couple nits below)

ovidiu-ssv-labs (dismissed):

networkconfig/network_test.go
momosh-ssv: Fixed in 880817ced.

Added beaconAt(epoch, slotsPerEpoch) holding the genesis math, with beaconAtEpoch delegating to it; the preGenesis override stays in the test body as you suggested.

networkconfig/network_test.go
momosh-ssv: Fixed in f5aa89621.

Added clean_next_domain_equals_domain_at_fork (currentEpoch == boole) asserting equal domains already pass at the exact activation epoch, closing the wall-clock edge.

y0sher (approved):

ljuba-ssv (approved):

Issue #2964, opened 2026-07-28
Enable the SSV_TEST_BOOLE_FORK=post unit-test matrix (deferred from #2941)


momosh-ssv:
## Context

#2941 added a second **spec-test** CI pass against the post-fork config (`spec-test-boole-post`, green), but the **unit-test** half of the `SSV_TEST_BOOLE_FORK` matrix was deferred: some unit-test fixtures are not fork-agnostic, so `make unit-test` only ever runs in the default pre-Boole mode. #2952 is closed and doesn't cover this — this issue is the owner for the deferral, per the [review thread](https://github.com/ssvlabs/ssv/pull/2941#discussion_r3636355145).

State as of the 07-23 sweep (thanks @iurii-ssv): `message/validation`, `networkconfig`, `operator/validator`, and `protocol/v2/ssv/...` already pass under `SSV_TEST_BOOLE_FORK=post` locally; the gaps below are what remains.

## Work items

- [ ] **1. Fork-agnostic `network/p2p` fixtures.** `TestP2pNetwork_SubscribeBroadcast` builds MsgIDs from the static `netCfg.DomainType`, which post-flip receivers reject. Either derive domains per-slot in the fixtures (mirroring production `DomainTypeAtSlot`) or pin the affected tests explicitly pre-fork the way `Test_ValidateSSVMessage` / `TestMsgValidator` were (502c7a755, 41b2dab01). Note these socket-heavy tests hang on some local machines (mdns), so verification effectively happens in CI.
- [ ] **2. Full-suite sweep under `SSV_TEST_BOOLE_FORK=post`.** Run the entire unit suite in post mode and fix or explicitly pin whatever else surfaces. (The metadata-syncer fixture — the other known failure from the review thread — was already fixed in #2941 itself.)
- [ ] **3. Enable the matrix.** Add an `SSV_TEST_BOOLE_FORK=post` unit-test job to CI, mirroring how the spec-test workflow gained its second pass (9920d1d21).
- [ ] **4. (Optional, same theme) Port `RunSyncCommitteeAggProof` to the alan-spec run.** The `alan_spec` build keeps a skipping stub because the boole-fork implementation relied on the removed `BaseValidator`; porting it to the committee-based flow would close the last skipped suite in the alan matrix leg.

## Why it matters

[...]

PR #2974, opened 2026-07-30
Enable the SSV_TEST_BOOLE_FORK=post unit-test matrix (#2964)

momosh-ssv:
Items 1–3 of #2964.

- **Fork-agnostic p2p fixtures** — `TestP2pNetwork_SubscribeBroadcast`'s fixtures built MsgIDs from the static `netCfg.DomainType`, which post-flip receivers reject (the LocalNet nodes run on the global `TestNetwork`, so under the flip they're post-fork on the Boole domain while the fixture carried Alan). The three fixture sites now derive the domain per-slot via `DomainTypeAtSlot`, exactly like production (`p2p_setup.go`). Chosen over pinning the test pre-fork: unlike the validation/topics pins, p2p has no separate post-fork subtest, so pinning would leave the matrix with zero post-fork p2p coverage.
- **Full-suite sweep** — all 112 packages were run locally under `SSV_TEST_BOOLE_FORK=post`: everything passes except `network/p2p`/`network/topics`, which hang **identically in default mode** locally (the known mdns env-hang, not a fork gap) and are verified by this new CI leg on Linux. Notably the spec-test packages (LFS fixtures present) also pass post-mode locally.
- **CI matrix leg** — new `ssv-boole-post` job in `unit-test.yml`, mirroring how `spec-test.yml` gained `spec-test-boole-post`: same job with `SSV_TEST_BOOLE_FORK: post`, its own cache key (falling back to the primary's), and **no codecov upload** (the primary job owns the core-flagged report; `make unit-test` already excludes spectest, so nothing double-runs).

Item 4 (the `RunSyncCommitteeAggProof` alan-spec port) is deliberately not included — it's a ~half-day non-mechanical port (the alan build predates the committee/aggregator-committee duty split); leaving it tracked on the issue.

Why this matters (from the issue): until this leg exists, the global-config flip path stays permanently unexercised in CI — exactly the gap that let the metadata-syncer fixture regress unnoticed.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -51,6 +51,43 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  # Second matrix leg: the same unit suite with the global test config flipped post-Boole-fork,
+  # mirroring how spec-test.yml gained its spec-test-boole-post pass. No codecov upload — the
+  # primary ssv job owns the core-flagged coverage report.
+  ssv-boole-post:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6.0.2
+        with:
+          lfs: true
+
+      - name: Setup Go
+        id: setup-go
+        uses: actions/setup-go@v6.3.0
+        with:
+          go-version-file: go.mod
+          # Keep setup-go cache disabled and use explicit actions/cache below so keying/restore
+          # behavior is consistent across workflows.
+          cache: false
+
+      - name: Restore Go cache (ssv boole-post)
+        uses: actions/cache@v5.0.4
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ steps.setup-go.outputs.go-version }}-unit-ssv-boole-post-${{ hashFiles('go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-${{ steps.setup-go.outputs.go-version }}-unit-ssv-boole-post-
+            ${{ runner.os }}-go-${{ steps.setup-go.outputs.go-version }}-unit-ssv-
+            ${{ runner.os }}-go-${{ steps.setup-go.outputs.go-version }}-
+
+      - name: Run unit tests (post-Boole fork config)
+        run: make unit-test
+        env:
+          SSV_TEST_BOOLE_FORK: post
+
   ssvsigner:
     runs-on: ubuntu-latest
 
--- a/network/p2p/p2p_trimming_test.go
+++ b/network/p2p/p2p_trimming_test.go
@@ -2,9 +2,12 @@ package p2pv1
 
 import (
 	"context"
+	"math"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -47,8 +50,14 @@ func (c *testTopicsController) Peers(topicName string) ([]peer.ID, error) {
 
 func (c *testTopicsController) Topics() []string {
 	// Mirror the real controller, which lists subscribed topics by their full pubsub name.
+	// Entries already carrying a full Boole-fork name (e.g. "/ssv/testnet/boole/11") pass
+	// through unchanged; bare subnet strings are wrapped in the Alan-style full name.
 	full := make([]string, len(c.topics))
 	for i, t := range c.topics {
+		if strings.Contains(t, "/") {
+			full[i] = t
+			continue
+		}
 		full[i] = commons.GetTopicFullName(t)
 	}
 	return full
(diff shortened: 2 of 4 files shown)


momosh-ssv:
Root cause of the remaining `ssv-boole-post` failure found and fixed in e4810b2fa.

The test's `roundLeader` helper hand-rolled the **pre-fork** round-robin proposer formula, while the validator's leader check delegates to the fork-aware `qbft.Proposer` — whose post-fork variant adds an `ethEpoch` offset to the round-robin index. So for ~3 of every 4 epochs the fixtures signed committee proposals with the wrong leader, and every non-sender node rejected them with `ErrSignerNotLeader` (senders bypass validation via `WithSelfAccept`, which is why exactly the non-sender routers hit zero). Pre-fork the two formulas are bit-identical, hence the asymmetry. `roundLeader` now delegates to `qbft.Proposer`.

Also wired `validation.WithLogger` into the LocalNet's validators — verdicts previously went to the default nop logger, which is why this took log archaeology instead of one glance.

Worth noting for a follow-up (doesn't affect pass/fail): three of the six fixtures never validated in *either* fork — the proposer message uses round 4 (> maxRound 2) with no registered duty, the SCC fixture is a deprecated role post-fork, and VoluntaryExit is a consensus message for a role with no consensus. Making them fork-appropriate (proposer at round ≤ 2 with a seeded duty; `RoleAggregatorCommittee` committee-style message post-fork; VE as a partial-sig message or dropped) would make the test's coverage honest — tracked as an optional item on #2964.

network/p2p/p2p_trimming_test.go
```diff
@@ -47,8 +50,14 @@ func (c *testTopicsController) Peers(topicName string) ([]peer.ID, error) {
 
 func (c *testTopicsController) Topics() []string {
 	// Mirror the real controller, which lists subscribed topics by their full pubsub name.
+	// Entries already carrying a full Boole-fork name (e.g. "/ssv/testnet/boole/11") pass
+	// through unchanged; bare subnet strings are wrapped in the Alan-style full name.
 	full := make([]string, len(c.topics))
 	for i, t := range c.topics {
+		if strings.Contains(t, "/") {
```
iurii-ssv, line 59: This `strings.Contains(t, "/")` split handles the two shapes the tests actually feed it (bare subnet `"0"` → wrapped Alan name; `/ssv/<net>/boole/N` → passthrough), but an _Alan-full_ name (`ssv.v2.11`, no slash) would fall through to the `else` and get double-wrapped to `ssv.v2.ssv.v2.11`. Nothing feeds it that shape today, so this is purely a readability ask: a one-line note that the helper accepts only bare-subnet or Boole-full inputs (never an already-wrapped Alan name) would keep a future caller from getting silently mangled output.

network/p2p/testutils.go
```diff
@@ -292,6 +297,28 @@ func NewLocalNet(ctx context.Context, logger *zap.Logger, options LocalNetOption
...
 
+// fixedTestSubnets returns the persistent subnet set used by local test networks: two fixed
+// subnets (64, 90) unrelated to any share, plus - for every configured share's committee - both
+// its Alan-fork subnet (CommitteeID-hash based) and its Boole-fork subnet (lowest-operator-hash
+// based). persistentSubnets is a raw, fork-agnostic bit vector (see initCfg), so covering both
+// mappings here is what keeps the committee's subnet persistently subscribed on either side of
+// the Boole fork, rather than depending solely on the later Subscribe(vpk) path.
+func fixedTestSubnets(shares []*ssvtypes.SSVShare) p2pcommons.Subnets {
+	subnets := p2pcommons.ZeroSubnets
+	subnets.Set(64)
```
iurii-ssv, line 310: `Set(64)` / `Set(90)` carry over from the old `00…0100000400000400` constant, but now that `114` is derived dynamically from the share these two are the only remaining magic numbers. The doc comment calls them "unrelated to any share" — a word on _why_ two arbitrary extra subnets are pinned (e.g. to exercise subscription to non-committee subnets) would save the next reader a git-blame trip.

.github/workflows/unit-test.yml
```diff
@@ -51,6 +51,43 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  # Second matrix leg: the same unit suite with the global test config flipped post-Boole-fork,
+  # mirroring how spec-test.yml gained its spec-test-boole-post pass. No codecov upload — the
+  # primary ssv job owns the core-flagged coverage report.
+  ssv-boole-post:
```
iurii-ssv, line 57: For this leg to actually prevent the regression #2964 targets (the global-config flip path silently breaking), `ssv-boole-post` needs to be added to the repo's branch-protection **required status checks** — otherwise it runs on every PR but a red result won't block merge. The same likely applies to `spec-test-boole-post`. Worth a follow-up with whoever owns the repo settings, since gating is the whole point of the PR.

.github/workflows/unit-test.yml
```diff
@@ -51,6 +51,43 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6.0.2
```
iurii-ssv, line 61: Greptile flagged these `@vX.Y.Z` action refs for immutable-SHA pinning. For context: the primary `ssv` job and every other job in this workflow already use the same floating tags, so pinning only this new job would leave the file inconsistent without meaningfully improving supply-chain posture. If we want that hardening it should be applied file-wide (ideally repo-wide via Dependabot action pinning), not singled out here — as written this job just mirrors the existing convention.

network/p2p/testutils.go
```diff
@@ -204,14 +204,18 @@ func (ln *LocalNet) NewTestP2pNetwork(ctx context.Context, nodeIndex uint64, key
 	cfg := NewNetConfig(keys, ln.Bootnode, 0, 0, options.Nodes)
 	cfg.Ctx = ctx
 	cfg.MdnsDiscoveryTag = ln.mdnsTag
-	cfg.Subnets = "00000000000000000100000400000400" // calculated for topics 64, 90, 114; PAY ATTENTION for future test scenarios which use more than one eth-validator we need to make this field dynamically changing
+	testSubnets := fixedTestSubnets(options.Shares)
+	cfg.Subnets = testSubnets.StringHex() // PAY ATTENTION for future test scenarios which use more than one eth-validator we need to make this field dynamically changing
```
iurii-ssv: Minor: this `PAY ATTENTION ... make this field dynamically changing` comment is now partly stale — subnets *are* derived from the shares via `fixedTestSubnets(options.Shares)`. The only remaining limitation is the single-committee assumption, so the comment could be trimmed to say that rather than implying the field is still fully static.

iurii-ssv (dismissed):
None-blocking suggestions.

.github/workflows/unit-test.yml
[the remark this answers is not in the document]
momosh-ssv: Not pinning here — the new job mirrors the floating-tag convention every existing job in this workflow uses; repo-wide pinning would be the right venue for that hardening.

network/p2p/p2p_trimming_test.go
momosh-ssv: Fixed in 2be1af406.
Added a note that the helper accepts only bare-subnet or Boole-full inputs, and that an already-wrapped Alan name would get double-wrapped.

network/p2p/testutils.go
momosh-ssv: Fixed in cc7c79b81.
Documented that 64/90 are bit positions carried over from the legacy fixture constant this function replaced, kept so every node also stays subscribed to a couple of subnets with no local committee.

.github/workflows/unit-test.yml
momosh-ssv: Agreed.
Without required-check gating the leg is informational only. Seems that's a repo-settings change needing admin access, so I'll raise adding ssv-boole-post (and spec-test-boole-post) to the required checks as a follow-up on #2964.

.github/workflows/unit-test.yml
momosh-ssv: Agreed on all points.
The job just mirrors the file's existing convention. If we ever want that hardening, repo-wide pinning (e.g. Dependabot's action-pinning) seems like the right venue rather than one job.

network/p2p/testutils.go
momosh-ssv: Fixed in cc7c79b81.
I dropped the comment entirely rather than trimming it: fixedTestSubnets loops over all shares and their committees, so I couldn't find a remaining single-committee assumption to warn about. 

Happy to restore a trimmed note if I'm missing where it still bites.

nkryuchkov (approved):

y0sher (approved):

iurii-ssv (approved):

PR #2975, opened 2026-07-30
exporter: fork-straddling trace ranges + /committee cardinality doc (#2968 items 1, 2)

momosh-ssv:
Items 1 and 2 of #2968, one commit each.

**Item 1** — `validateValidatorRequest` evaluated the committee-duty fork gate at `request.To`, so a validator-traces window whose tail crossed the Boole fork was rejected wholesale: a dashboard polling a fixed window starts 400-ing the moment its tail crosses the fork, losing the still-servable pre-fork slots. The gate now evaluates at `request.From`, and each post-fork committee-duty slot lacking pubkeys/indices is reported as a **non-fatal per-slot note** (`ErrPostForkCommitteeDutyNote`, `errors.Is`-able) in the response `Errors` instead of being silently empty. The HTTP handler treats a response whose only errors are such notes as a partial 200 — review caught that without this, a straddling range with zero pre-fork traces (common: aggregator duties are sparse) would have 500'd. `from`-already-post-fork and unfiltered ATTESTER/SYNC_COMMITTEE requests reject exactly as before.

**Item 2** — docs-only: the `/committee` OpenAPI description now states that absent a `roles` filter the endpoint returns one trace per (slot, committeeID, **role**) — up to two rows per (slot, committeeID) post-fork, distinguished by the `role` field. Additive-field back-compat holds; cardinality back-compat does not, hence the note. Regenerated via `make openapi`, diff limited to description text.

Covered by fork-phase tests (per-test `TestNetwork` copies with `Forks.Boole` pinned), including HTTP-level cases for the straddling 200-partial path and the genuine-error-still-500 path.

### DIFF
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -2640,6 +2640,44 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 			}
 		})
 
+		t.Run(role.name+" without filters serves pre-fork data alongside post-fork notes", func(t *testing.T) {
+			store := newMockTraceStore()
+			// the unfiltered pre-fork path reads GetValidatorDuties per slot;
+			// post-fork slots are skipped with a note and must never reach it.
+			store.GetValidatorDutiesFunc = func(r spectypes.BeaconRole, slot phase0.Slot) ([]*traces.ValidatorDutyTrace, error) {
+				require.Less(t, uint64(slot), uint64(booleSlot), "unfiltered validator path used at post-Boole slot")
+				return []*traces.ValidatorDutyTrace{{Slot: slot, Role: r, Validator: idx}}, nil
+			}
+
+			exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), &netCfg)
+
+			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
+				"from":  from,
+				"to":    to,
+				"roles": []string{role.name},
+			}))
+			req.Header.Set("Content-Type", "application/json")
+			rec := httptest.NewRecorder()
+
+			require.NoError(t, exp.ValidatorTraces(rec, req))
+			require.Equal(t, http.StatusOK, rec.Code)
+
+			var resp ValidatorTracesResponse
+			require.NoError(t, json.Unmarshal(rec.Body.Bytes(), &resp))
+
+			// the partial-coverage contract: real pre-fork traces in Data while
+			// the post-fork tail is reported as a note in Errors, in one response.
+			require.Len(t, resp.Data, int(uint64(booleSlot)-from), "expected one trace per pre-fork slot")
+			for _, item := range resp.Data {
+				assert.Less(t, uint64(item.Slot), uint64(booleSlot), "post-fork slot leaked into data")
+				assert.Equal(t, role.name, item.Role)
+			}
+			require.NotEmpty(t, resp.Errors, "expected the post-fork note to surface alongside data")
+			for _, msg := range resp.Errors {
+				require.Contains(t, msg, "committee duty post-fork")
+			}
+		})
+
 		t.Run(role.name+" with indices routes each slot by its own fork state", func(t *testing.T) {
 			store := newMockTraceStore()
 
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -625,6 +625,16 @@ func newTestExporterForV2WithNetwork(traceStore *mockTraceStore, validators stor
 	return NewExporter(zap.NewNop(), nil, traceStore, validators, netCfg)
 }
 
+// booleNetwork returns a clone of TestNetwork with the Boole fork pinned at the
+// given epoch. The SSV config is copied too so the shared TestNetwork is never mutated.
+func booleNetwork(epoch phase0.Epoch) *networkconfig.Network {
+	ssvCopy := *networkconfig.TestNetwork.SSV
+	ssvCopy.Forks.Boole = epoch
+	netCfg := *networkconfig.TestNetwork
+	netCfg.SSV = &ssvCopy
+	return &netCfg
+}
+
 func buildJSONBody(t *testing.T, payload map[string]any) *strings.Reader {
 	t.Helper()
 	b, err := json.Marshal(payload)
@@ -2450,12 +2460,7 @@ func TestExporterValidatorTraces_ForkGating(t *testing.T) {
 				return &traces.CommitteeDutyTrace{Slot: s, CommitteeID: id}, nil
 			}
 
-			ssvCopy := *networkconfig.TestNetwork.SSV
-			ssvCopy.Forks.Boole = tt.booleEpoch
-			netCfg := *networkconfig.TestNetwork
-			netCfg.SSV = &ssvCopy
-
-			exp := newTestExporterForV2WithNetwork(store, validatorStore, &netCfg)
+			exp := newTestExporterForV2WithNetwork(store, validatorStore, booleNetwork(tt.booleEpoch))
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 				"from":    uint64(slot),
@@ -2539,12 +2544,7 @@ func TestExporterValidatorTraces_ForkGating_ValidationSymmetric(t *testing.T) {
 			}
 			validatorStore := newMockValidatorStore()
 
-			ssvCopy := *networkconfig.TestNetwork.SSV
-			ssvCopy.Forks.Boole = tt.booleEpoch
-			netCfg := *networkconfig.TestNetwork
-			netCfg.SSV = &ssvCopy
-
-			exp := newTestExporterForV2WithNetwork(store, validatorStore, &netCfg)
+			exp := newTestExporterForV2WithNetwork(store, validatorStore, booleNetwork(tt.booleEpoch))
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 				"from":  uint64(100),
@@ -2576,11 +2576,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 	var committeeID spectypes.CommitteeID
 	committeeID[0] = 7
 
-	ssvCopy := *networkconfig.TestNetwork.SSV
-	ssvCopy.Forks.Boole = booleEpoch
-	netCfg := *networkconfig.TestNetwork
-	netCfg.SSV = &ssvCopy
-
+	netCfg := booleNetwork(booleEpoch)
 	booleSlot := netCfg.FirstSlotAtEpoch(booleEpoch)
 	require.GreaterOrEqual(t, uint64(booleSlot), uint64(10), "boole fork slot too low for the range below")
 	from := uint64(booleSlot) - 10 // pre-Boole
@@ -2614,7 +2610,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 
 	for _, role := range roles {
 		t.Run(role.name+" without filters returns a partial response with post-fork notes", func(t *testing.T) {
-			exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), &netCfg)
+			exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), netCfg)
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 				"from":  from,
@@ -2649,7 +2645,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 				return []*traces.ValidatorDutyTrace{{Slot: slot, Role: r, Validator: idx}}, nil
 			}
 
-			exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), &netCfg)
+			exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), netCfg)
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 				"from":  from,
@@ -2696,7 +2692,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 				return role.signerDataBuilder(s), nil
 			}
 
-			exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), &netCfg)
+			exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), netCfg)
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 				"from":    from,
@@ -2731,18 +2727,14 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 func TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces(t *testing.T) {
 	const booleEpoch = phase0.Epoch(5)
 
-	ssvCopy := *networkconfig.TestNetwork.SSV
-	ssvCopy.Forks.Boole = booleEpoch
-	netCfg := *networkconfig.TestNetwork
-	netCfg.SSV = &ssvCopy
-
+	netCfg := booleNetwork(booleEpoch)
 	booleSlot := netCfg.FirstSlotAtEpoch(booleEpoch)
 	require.GreaterOrEqual(t, uint64(booleSlot), uint64(2), "boole fork slot too low for the range below")
 	from := uint64(booleSlot) - 2 // pre-Boole
 	to := uint64(booleSlot) + 2   // post-Boole
 
 	t.Run("only post-fork notes and no pre-fork traces -> 200 with empty data", func(t *testing.T) {
-		exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), &netCfg)
+		exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), netCfg)
 
 		req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 			"from":  from,
@@ -2769,7 +2761,7 @@ func TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces(t *testing.T) {
 		store.GetValidatorDutiesFunc = func(role spectypes.BeaconRole, slot phase0.Slot) ([]*traces.ValidatorDutyTrace, error) {
 			return nil, fmt.Errorf("forced error on GetValidatorDuties")
 		}
-		exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), &netCfg)
+		exp := newTestExporterForV2WithNetwork(store, newMockValidatorStore(), netCfg)
 
 		req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
 			"from":  from,
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -2485,7 +2485,7 @@ func TestExporterValidatorTraces_ForkGating(t *testing.T) {
 }
 
 // TestExporterValidatorTraces_ForkGating_ValidationSymmetric proves that validateValidatorRequest
-// (via isCommitteeDutyAtSlot at the range's upper bound) mirrors the same fork-gated routing decision for aggregator-family
+// (via isCommitteeDutyAtSlot at the range's lower bound) mirrors the same fork-gated routing decision for aggregator-family
 // roles: pre-Boole no pubkeys/indices are required, post-Boole they are (mirroring committee duties).
 func TestExporterValidatorTraces_ForkGating_ValidationSymmetric(t *testing.T) {
 	tests := []struct {
@@ -2567,9 +2567,9 @@ func TestExporterValidatorTraces_ForkGating_ValidationSymmetric(t *testing.T) {
 
 // TestExporterValidatorTraces_ForkGating_CrossForkRange proves the behavior of a slot range
 // straddling the Boole fork boundary (from pre-Boole, to post-Boole): validation is evaluated
-// at the range's upper bound, so aggregator-family roles require pubkeys/indices, and with
-// indices provided each slot routes independently — validator path before the boundary,
-// committee path from it onward.
+// at the range's lower bound, so unfiltered aggregator-family requests are accepted and served
+// partially — post-fork slots are reported as non-fatal notes — and with indices provided each
+// slot routes independently: validator path before the boundary, committee path from it onward.
 func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 	const booleEpoch = phase0.Epoch(5)
 	idx := phase0.ValidatorIndex(1)
@@ -2741,7 +2741,15 @@ func TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces(t *testing.T) {
 		req.Header.Set("Content-Type", "application/json")
 		rec := httptest.NewRecorder()
 
-		require.Error(t, exp.ValidatorTraces(rec, req))
+		err := exp.ValidatorTraces(rec, req)
+		require.Error(t, err)
+
+		var apiErr *api.ErrorResponse
+		require.ErrorAs(t, err, &apiErr)
+		require.Equal(t, http.StatusInternalServerError, apiErr.Code,
+			"a genuine store failure must not be masked by the post-fork note exemption")
+		require.Contains(t, apiErr.Message, "forced error on GetValidatorDuties",
+			"the genuine error, not a post-fork note, must surface to the caller")
 	})
 }
 
--- a/api/handlers/exporter/exporter_test.go
+++ b/api/handlers/exporter/exporter_test.go
@@ -2613,7 +2613,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 	}
 
 	for _, role := range roles {
-		t.Run(role.name+" without filters requires pubkeys/indices", func(t *testing.T) {
+		t.Run(role.name+" without filters returns a partial response with post-fork notes", func(t *testing.T) {
 			exp := newTestExporterForV2WithNetwork(newMockTraceStore(), newMockValidatorStore(), &netCfg)
 
 			req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
@@ -2624,7 +2624,20 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 			req.Header.Set("Content-Type", "application/json")
 			rec := httptest.NewRecorder()
 
-			require.Error(t, exp.ValidatorTraces(rec, req))
+			// the pre-fork portion of the range legitimately yields zero traces (no
+			// mock data), so the post-fork "requires pubkeys/indices" notes must not
+			// be treated as a hard failure: expect 200 with empty data and the notes
+			// surfaced in Errors.
+			require.NoError(t, exp.ValidatorTraces(rec, req))
+			require.Equal(t, http.StatusOK, rec.Code)
+
+			var resp ValidatorTracesResponse
+			require.NoError(t, json.Unmarshal(rec.Body.Bytes(), &resp))
+			require.Empty(t, resp.Data)
+			require.NotEmpty(t, resp.Errors)
+			for _, msg := range resp.Errors {
+				require.Contains(t, msg, "committee duty post-fork")
+			}
 		})
 
 		t.Run(role.name+" with indices routes each slot by its own fork state", func(t *testing.T) {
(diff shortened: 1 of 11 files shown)


exporter/validator.go
```diff
@@ -36,6 +43,14 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
 	for s := request.From; s <= request.To; s++ {
 		slot := phase0.Slot(s)
 		for _, role := range request.Roles {
+			if e.isCommitteeDutyAtSlot(role, slot) && len(indices) == 0 {
```
iurii-ssv: **[nit]** `isCommitteeDutyAtSlot(role, slot)` is evaluated twice per (slot, role): here in the guard, and again a few lines below when choosing `providerFunc`. For filtered requests (indices present) the guard's result is always discarded by `&& len(indices) == 0`, so it's redundant work over the whole range. Compute it once:

```go
isCommittee := e.isCommitteeDutyAtSlot(role, slot)
if isCommittee && len(indices) == 0 {
    // … note; continue
}
providerFunc := e.getValidatorDutiesForRoleAndSlot
if isCommittee {
    providerFunc = e.getValidatorCommitteeDutiesForRoleAndSlot
}
```

Also reads a little cleaner.

api/handlers/exporter/exporter_test.go
```diff
@@ -2613,7 +2613,7 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
 	}
 
 	for _, role := range roles {
-		t.Run(role.name+" without filters requires pubkeys/indices", func(t *testing.T) {
+		t.Run(role.name+" without filters returns a partial response with post-fork notes", func(t *testing.T) {
```
iurii-ssv, line 2612: The fork gate moved from the upper to the **lower** bound, but two test doc-comments still describe it as the upper bound and are now inaccurate:

- This test's doc comment just above (L2568–2571): _"validation is evaluated at the range's upper bound, so aggregator-family roles require pubkeys/indices …"_ — it's now evaluated at `from`, and unfiltered aggregator-family straddling requests return a partial 200 with notes instead of being rejected.
- `TestExporterValidatorTraces_ForkGating_ValidationSymmetric` doc comment (L2488): _"via isCommitteeDutyAtSlot at the range's upper bound"_.

Both test bodies still pass (they use single-side networks), so this is comment-only staleness — just worth updating the wording to match the new lower-bound behavior.

iurii-ssv (commented):
Check out the Greptile comment above + left two more minor nits.

api/handlers/exporter/exporter_test.go
```diff
@@ -2671,6 +2684,67 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
...
+
+		req := httptest.NewRequest(http.MethodPost, "/traces/validator", buildJSONBody(t, map[string]any{
+			"from":  from,
+			"to":    to,
+			"roles": []string{"AGGREGATOR"},
+		}))
+		req.Header.Set("Content-Type", "application/json")
+		rec := httptest.NewRecorder()
+
+		require.Error(t, exp.ValidatorTraces(rec, req))
```
ovidiu-ssv-labs: **Finding 3 · [MINOR] Pin the status code in the "genuine error still yields 500" test**

This subtest is the *only* guard on the safety net for the newly-loosened error path — it is what proves `onlyPostForkCommitteeDutyNotes` doesn't swallow real store failures. But it asserts only that *an* error came back, not *which* one:

```go
t.Run("genuine error alongside notes still yields 500", func(t *testing.T) {
    ...
    require.Error(t, exp.ValidatorTraces(rec, req))
})
```

`ValidatorTraces` returns `*api.ErrorResponse` from `toApiError` for both the 400 branch (`validator_http.go:43`) and the 500 branch (`validator_http.go:50`). `require.Error` is satisfied by either. If a future refactor caused the forced `GetValidatorDuties` failure to be wrapped in a `*exportercore.ValidationError`, `isValidationError` would fire first and the handler would return a **400** — and this test would still pass, green, with a name claiming it verified 500.

Note also that `rec.Code` is useless here: the handler *returns* the error rather than writing it, so nothing touches the recorder and `rec.Code` stays at the `httptest` default of 200. The status has to be read off the returned `*api.ErrorResponse.Code` (`api/errors.go:11`).

**Suggested fix:** ```go
err := exp.ValidatorTraces(rec, req)
require.Error(t, err)

var apiErr *api.ErrorResponse
require.ErrorAs(t, err, &apiErr)
require.Equal(t, http.StatusInternalServerError, apiErr.Code,
    "a genuine store failure must not be masked by the post-fork note exemption")
require.Contains(t, apiErr.Message, "forced error on GetValidatorDuties")
```

The `Message` assertion is worth adding too — it pins that the *genuine* error, not just one of the notes, is what surfaced to the caller.

**Review by ovidiu-ssv-labs — COMMENTED** [review]
The core mechanism is sound — evaluating the fork gate at From and emitting per-slot non-fatal notes is the right shape, the sentinel/errors.Is design is clean, and the test-network cloning is genuinely leak-free. But the PR documents the smaller back-compat deviation (/committee cardinality) while leaving the larger one (/traces/validator 400 -> 200-with-empty-data) undocumented, the test guarding the new 500 escape hatch doesn't actually assert the status code, and /decideds remains fork-blind for the same two roles this PR just made fork-aware on /traces/validator.  [verdict: with_fixes]



**Finding 1 · [IMPORTANT] Document the /traces/validator 400 -> 200 contract change in the OpenAPI description** — `api/handlers/exporter/validator_http.go:15`

**The PR documents the smaller back-compat break and leaves the larger one undocumented.**

Item 2 of this PR adds an OpenAPI note to `/committee` because "cardinality back-compat does not hold." That is the right instinct — but item 1 changes a *status code*, which is a strictly harder break for clients, and gets no doc change at all. `validator_http.go:15` still reads:

```go
// @Description Returns consensus, decided, and message traces for the requested validator duties.
```

**The behavior change, concretely.** For `roles=[AGGREGATOR]` (or `SYNC_COMMITTEE_CONTRIBUTION`) with no `pubkeys`/`indices`:

| request window | before | after |
|---|---|---|
| fully pre-Boole | 200 + data | 200 + data (unchanged) |
| straddles Boole | **400** | **200**, pre-fork data only, post-fork slots reported as strings in `errors` |
| fully post-Boole | 400 | 400 (unchanged) |

**Why it matters.** The 400 was a *contract signal*: "you must supply indices/pubkeys or this data is not retrievable." A client that treated non-2xx as "my query is wrong, fix it" now receives a 200 and, in the sparse-aggregator case that `TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces` covers, `data: []` with a populated `errors` array. A dashboard that only reads `data` will render an empty chart and report success — silently missing every post-fork aggregator duty in the window.

There is also an undocumented cliff: a window starting one slot before the fork returns 200-with-notes, while the same window shifted one slot forward returns 400. Nothing tells an API consumer that.

**Additional wrinkle:** `ValidatorTracesResponse.Errors` is a flat `[]string` (`validator_model.go:63`). Post-fork notes and genuine store failures land in the same untyped array, so a client cannot programmatically distinguish "partial coverage, expected" from "the store is broken." The sentinel exists in Go (`ErrPostForkCommitteeDutyNote`) but is flattened to a string at the API boundary.

**Suggested fix:** Extend the godoc description and regenerate via `make openapi`, mirroring what item 2 did for `/committee`:

```go
// @Description Returns consensus, decided, and message traces for the requested validator duties.
// @Description For AGGREGATOR and SYNC_COMMITTEE_CONTRIBUTION the fork state is evaluated at 'from': a range whose
// @Description 'from' is post-Boole and that supplies no 'pubkeys'/'indices' is rejected with 400. A range whose
// @Description 'from' is pre-Boole is accepted and served partially — the post-Boole slots are omitted from 'data'
// @Description and reported per slot in 'errors' with the text "committee duty post-fork". Such a response is a 200
// @Description even when 'data' is empty; clients must inspect 'errors' to detect partial coverage, and should
// @Description supply 'indices'/'pubkeys' or use /traces/committee to retrieve the post-fork portion.
```

If you want the stronger fix, promote the notes out of the flat `errors` array into a typed field (e.g. `"partial": [{"slot": N, "role": "AGGREGATOR", "reason": "post_fork_committee_duty"}]`) so clients can branch on it without string matching. Given this is a new field, additive back-compat holds.

**Finding 2 · [IMPORTANT] /decideds is still fork-blind for AGGREGATOR / SYNC_COMMITTEE_CONTRIBUTION** — `exporter/decided.go:41`

**Adjacent gap — pre-existing, not introduced here, but this PR makes it the last fork-blind routing site in the package.**

After this PR, `isCommitteeDutyAtSlot` (`exporter/validator.go:278`) is the *only* Boole-aware routing point in `exporter/`.

`TraceDecidedsCore` routes on a hardcoded role switch with no slot/fork input:

```go
// exporter/decided.go:41-46
switch role {
case spectypes.BNRoleAttester, spectypes.BNRoleSyncCommittee:
    roleParticipantsIdx, roleErrs = e.getCommitteeDecidedsForRole(slot, indices, role)
default:
    roleParticipantsIdx, roleErrs = e.getValidatorDecidedsForRole(slot, indices, role)
}
```

**Mechanism.** Post-Boole, aggregator and sync-committee-contribution duties execute under the `RoleAggregatorCommittee` runner (`protocol/v2/ssv/runner/aggregator_committee.go:86`), and the observer stores their traces on the committee path (`protocol/v2/ssv/validator/committee_observer.go:117` gates `RoleCommittee || RoleAggregatorCommittee` onto the committee-ID branch). So post-fork there is no `ValidatorDutyTrace` for `BNRoleAggregator` to find. Yet the `default` arm sends it to `getValidatorDecidedsForRole` -> `GetAllValidatorDecideds(BNRoleAggregator, slot)` -> `c.store.GetValidatorDuties(role, slot)` (`exporter/dutytracer/store.go:421-424`), which reads the validator-duty index and returns nothing.

**The strongest evidence that this is a bug and not intended:** the store layer is *already* built to serve these roles from the committee side. `committeeRunnerRolesForBeaconRoles` explicitly maps aggregator-family beacon roles to `RoleAggregatorCommittee`, and `GetCommitteeDecideds` defaults to `[]RunnerRole{RoleCommittee, RoleAggregatorCommittee}`. There is even a dedicated test, `TestCollector_GetCommitteeDecideds_RoleFiltering`, asserting `GetCommitteeDecideds(slot, index, spectypes.BNRoleAggregator)` works. **Nothing in production calls it** — the `decided.go:42` switch never routes `BNRoleAggregator` to the committee path.

**Impact.** Post-Boole, `/v1/exporter/decideds` with `roles=AGGREGATOR` or `SYNC_COMMITTEE_CONTRIBUTION` returns an empty `participants` list with a 200. Silent data loss — no error, no note, no 400. Worse than the `/traces/validator` case this PR just fixed, because there the user at least gets a note. Any duty-syncer or dashboard consuming decideds for these roles will show zero participation post-fork and conclude the operators stopped performing the duty.

**Scope.** This is outside items 1-2 of #2968 and not in this diff, so not blocking — but it is the natural item 3 of the same umbrella and should be filed before Boole activates, not after.

**Suggested fix:** Route the decideds switch through the same fork-aware predicate the traces path now uses, so there is exactly one place that knows about Boole:

```go
for _, role := range request.Roles {
    for s := request.From; s <= request.To; s++ {
        slot := phase0.Slot(s)

        var roleParticipantsIdx []dutytracer.ParticipantsRangeIndexEntry
        var roleErrs *multierror.Error

        if e.isCommitteeDutyAtSlot(role, slot) {
            roleParticipantsIdx, roleErrs = e.getCommitteeDecidedsForRole(slot, indices, role)
        } else {
            roleParticipantsIdx, roleErrs = e.getValidatorDecidedsForRole(slot, indices, role)
        }
        ...
```

This is behavior-preserving pre-Boole (`isCommitteeDutyAtSlot` returns true unconditionally for ATTESTER/SYNC_COMMITTEE and false for the aggregator family pre-fork), and correct post-Boole. Note the loop nesting in `decided.go` is role-outer/slot-inner, the inverse of `validator.go`, so `isCommitteeDutyAtSlot` must be evaluated inside the slot loop as above. Add a straddling-range test mirroring `TestValidatorTracesCore_StraddlingFork`.

If you'd rather keep this PR tight, file it as item 3 of #2968 and land it separately — but before Boole activates on mainnet.

> 3 findings approved — 2 could not be inlined (target line not in this diff), 1 inlined below.

exporter/validator.go
```diff
@@ -36,6 +43,14 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
...
+				// request validation only gates on 'from': a window whose tail crosses
+				// Boole reaches here for its post-fork slots without pubkeys/indices,
+				// so report the gap as a non-fatal note instead of silently skipping it.
+				errs = multierror.Append(errs, fmt.Errorf("%w: slot %d: role %s is a committee duty post-fork, please provide either pubkeys or indices to filter the duty for a specific validators subset or use the /committee endpoint to query all the corresponding duties", ErrPostForkCommitteeDutyNote, slot, role.String()))
```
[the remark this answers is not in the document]
momosh-ssv: Fixed in 82535ff — post-fork notes are now aggregated into one per role. The missing range bound / MaxUint64 wrap is pre-existing across all exporter range endpoints, tracked in #2986.

exporter/validator.go
momosh-ssv: Good point — fixed in 8e3d984.

api/handlers/exporter/exporter_test.go
momosh-ssv: Fixed in 9de8c0c — both doc comments now describe the lower-bound gating, and the CrossForkRange one also mentions the partial-200 behavior.

api/handlers/exporter/exporter_test.go
momosh-ssv: Fixed in 9de8c0c — asserting the returned *api.ErrorResponse now: Code pinned to 500 and Message pinned to the forced store error. Good catch on rec.Code staying at the httptest default, that would have bitten us.

momosh-ssv:
@ovidiu-ssv-labs re the two findings without inline threads:

**Finding 1**: fixed in 32c9d96 — extended the /traces/validator OpenAPI description with the partial-coverage contract, mirroring the /committee note. One wording change vs. the suggestion: since 82535ff the notes are aggregated into one per role covering the post-fork tail, rather than one per slot. The typed `partial` field is a good idea — left it out to keep this PR tight, happy to do it as an additive follow-up.

**Finding 2**: agreed on all points — filed as #2987 (item 3 of the #2968 umbrella) so it lands before Boole activates.

exporter/validator.go
```diff
@@ -47,6 +69,15 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
 		}
 	}
 
+	for _, role := range request.Roles {
+		noteFrom, ok := postForkNoteFrom[role]
+		if !ok {
+			continue
+		}
+		delete(postForkNoteFrom, role) // guard against duplicate roles in the request
+		errs = multierror.Append(errs, fmt.Errorf("%w: slots %d-%d: role %s is a committee duty post-fork, please provide either pubkeys or indices to filter the duty for a specific validators subset or use the /committee endpoint to query all the corresponding duties", ErrPostForkCommitteeDutyNote, noteFrom, request.To, role.String()))
```
iurii-ssv: [nit] The tail of this note ("please provide either pubkeys or indices to filter the duty ... or use the /committee endpoint ...") is duplicated verbatim from the 400 message in `validateValidatorRequest` (the `role %s is a committee duty` return just below). Consider hoisting the shared sentence into a package-level const so the note and the validation error can't drift apart.

api/handlers/exporter/exporter_test.go
```diff
@@ -2671,6 +2684,75 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
...
+		}))
+		req.Header.Set("Content-Type", "application/json")
+		rec := httptest.NewRecorder()
+
+		require.NoError(t, exp.ValidatorTraces(rec, req))
+		require.Equal(t, http.StatusOK, rec.Code)
+
+		var resp ValidatorTracesResponse
+		require.NoError(t, json.Unmarshal(rec.Body.Bytes(), &resp))
+		require.Empty(t, resp.Data)
```
iurii-ssv, line 2752: Coverage gap: every 200 case in these fork tests asserts `require.Empty(t, resp.Data)` (here, the CrossForkRange "without filters" subtest, and `TestValidatorTracesCore_StraddlingFork`). The shape that is never exercised is the actual partial-coverage contract - an unfiltered straddling range whose **pre-fork** slots return real traces **and** whose post-fork tail emits a note, i.e. `resp.Data` non-empty *while* `resp.Errors` carries the note. The "with indices" subtest hits the committee path but via the filtered route (which never emits a note) and only asserts routing. Worth one case that seeds pre-fork trace data so data + note coexist.

exporter/validator.go
```diff
@@ -66,11 +97,12 @@ func (e *Exporter) validateValidatorRequest(request *ValidatorTracesQuery) error
...
 	// either PubKeys or Indices are required for committee duty roles.
-	// Fork state is evaluated at the range's upper bound: if any slot in
-	// [from, to] is post-Boole, the 'to' slot is too.
+	// Fork state is evaluated at the range's lower bound so that a window
+	// whose tail crosses Boole still serves its pre-fork portion; the
+	// post-fork tail is reported as a non-fatal note in the per-slot loop.
 	if len(request.PubKeys) == 0 && len(request.Indices) == 0 {
 		for _, role := range request.Roles {
-			if e.isCommitteeDutyAtSlot(role, phase0.Slot(request.To)) {
+			if e.isCommitteeDutyAtSlot(role, phase0.Slot(request.From)) {
```
iurii-ssv, line 110: Follow-on to the [note-amplification thread](https://github.com/ssvlabs/ssv/pull/2975#discussion_r3682757966) / #2986 — not re-raising the endpoint-wide range bound, but one nuance that disposition doesn't capture: moving this gate from `request.To` to `request.From` **regresses** a specific input class from a clean 400 into a hang.

An unfiltered committee-duty request (`roles:[AGGREGATOR]`, no `indices`/`pubkeys`) with a **pre-Boole `from`** and **`to = math.MaxUint64`** used to be rejected here with 400 (the old gate evaluated `to`, which is post-Boole). It's now accepted and falls into the inclusive `for s := request.From; s <= request.To; s++` loop, where `s++` wraps past `MaxUint64` and `s <= request.To` stays true forever — an infinite loop pinning a goroutine at 100% CPU, with no `ctx`/timeout escape.

The DoS itself is pre-existing (already reachable via any filtered request, or an unfiltered `PROPOSER`, with `to = MaxUint64`), so this is **low-severity and non-blocking** — leaving the general fix to #2986 is fine. But since this PR is what removes the 400 that shielded the unfiltered path, a cheap local guard in `validateValidatorRequest` (reject `to == math.MaxUint64`, or `to - from` above a sane cap, next to the existing `from > to` check) would stop this PR from regressing the case while the endpoint-wide fix waits.

iurii-ssv (commented):
Couple more things to check out.

exporter/validator.go
momosh-ssv: Good call.
Hoisted the shared sentence into a `committeeDutyFilterHint` const used by both the note and the validation error. 

Fixed in 36e901d53.

api/handlers/exporter/exporter_test.go
momosh-ssv: Fixed in 5b4c59a1e.
Added an unfiltered straddling subtest (both roles) that seeds pre-fork traces via GetValidatorDuties, asserting non-empty `data` alongside the post-fork note in `errors`, plus a guard that the unfiltered validator path is never consulted at post-Boole slots.

exporter/validator.go
momosh-ssv: Fair point.
Even though the hang is pre-existing on the filtered paths, this PR did remove the 400 that shielded the unfiltered one. Fixed in 36f79684a: `validateValidatorRequest` now rejects `to == MaxUint64` next to the existing `from > to` check, with a test. 

The endpoint-wide range cap stays with #2986.

y0sher:
@ovidiu-ssv-labs please lets report here when we tested this PR and exporter data against pre and post fork operators. when you approve lets merge

ovidiu-ssv-labs:
## Test report: build with the PR against build without the PR
### All tests PASSED - with mentions

I tested this PR on the hoodi-stage exporter. I used two builds on the same pod and the same volume. Both builds use Boole epoch 117422. The fork slot is 3757504.

| | With the PR | Without the PR |
|---|---|---|
| Build | `1c85cf90` | `f77f5095` |
| Branch | `deploy/boole-exporter-forking-2008` | `deploy/boole-stage-exporter-forking-2108` |

The two builds use the same database. Therefore the row counts are comparable.

**Result: 41 test cases. 3 cases differ but expected. 38 cases are the same.**

### The three cases that differ

| Case | Request | With the PR | Without the PR |
|---|---|---|---|
| VAL-02 | Slots 3757490 to 3757510. Role `AGGREGATOR`. No filter. | `200`. 3163 rows. 1 note. | `400` |
| GET-04 | The same request as a GET | `200`. 3163 rows. 1 note. | `400` |
| VAL-03 | The same range. Two aggregator roles. | `200`. 3205 rows. 2 notes. | `400` |

Without the PR, the exporter refuses the request:

```
role AGGREGATOR is a committee duty, please provide either pubkeys or indices …
```

With the PR, the exporter serves the rows before the fork. It adds one note for each role:

```
committee duty post-fork requires pubkeys or indices: slots 3757504-3757510:
role AGGREGATOR is a committee duty post-fork, …
```

The note gives the exact slots that the exporter did not serve. VAL-03 shows that the exporter groups the notes. The range holds 7 slots after the fork and the request holds 2 roles. The exporter returns 2 notes. It does not return 14 notes.

One control case shows the value of the change. Without the PR, the same range with `indices` returns `200` and 1 row. The rows were always available. The exporter refused the request only.

### The 38 cases that are the same

No behaviour changed outside the three cases above. These cases include:

- Requests that must fail: a range after the fork with no filter, role `ATTESTER` across the fork, role `SYNC_COMMITTEE` across the fork, an empty role list, `from` more than `to`, and an incorrect role name for the committee route. All give `400` on both builds, with the same message.
- The `/v1/exporter/traces/committee` route: 7 cases, all the same. After the fork, a request with no role filter gives two rows for each slot and committee. One row is `COMMITTEE`. One row is `AGGREGATOR_COMMITTEE`.
- The `/v1/exporter/decideds` route: 8 cases, all the same.
- The filter paths: `pubkeys`, `indices`, an incorrect key length, and an unknown key. All the same.
- The GET routes: the query parameter binders give the same results as the JSON body.

### Status of each change in the PR

| Change | Status |
|---|---|
| Serve the part of the range before the fork | **Confirmed.** VAL-02 and GET-04. |
| Return `200` with notes in place of `500` | **Partly confirmed.** The tests show the correct path. The test for a real storage error needs a fault. A unit test covers it. |
| Group the notes, one for each role | **Confirmed.** VAL-03. |
| Refuse a `to` value of `MaxUint64` | **Confirmed by hand, not in the matrix.** With the PR the exporter returns `400`. Without the PR the exporter does not stop. **Important: the new test uses `==` only. A `to` value of `MaxUint64 - 1` is still permitted, and the exporter does not stop. Issue #2986 remains open and necessary.** |
| Evaluate `isCommitteeDutyAtSlot` one time for each slot and role | **Not testable.** The change has no effect that a test can see. |
| Document the row count for `/committee` | **Behaviour confirmed.** The exporter does not serve the OpenAPI file, so I compared the text in the repository only. |

No regression found.

### Note on additional tests

The 41 cases also cover the three exporter routes for their own sake, before the fork and after the fork. These tests found four items that this PR does not cause and does not correct. The most important item: the `/v1/exporter/decideds` route ignores the fork for aggregator roles. It returns no rows and no error after the fork. I record all four items in #2996.

exporter/validator.go
```diff
@@ -61,17 +98,25 @@ func (e *Exporter) validateValidatorRequest(request *ValidatorTracesQuery) error
 		return fmt.Errorf("'from' must be less than or equal to 'to'")
 	}
 
+	// the per-slot loop is inclusive of 'to', so the maximum uint64 value
+	// would wrap the counter and never terminate. An endpoint-wide range
+	// bound is tracked in #2986; this only rejects the guaranteed hang.
+	if request.To == math.MaxUint64 {
```
iurii-ssv: Clarifying note on the endpoint-wide `MaxUint64` disposition tracked in #2986 (raised in the note-amplification thread) — the general fix stays there, but one part of it is now trivially closable in this PR:

This `to == math.MaxUint64` guard closes the guaranteed hang, but only for `/traces/validator`. The two sibling endpoints run the **identical** inclusive loop with no such guard:

- `/traces/committee` — `for s := request.From; s <= request.To; s++` (`exporter/committee.go:26`, and again at `:95`); `validateCommitteeRequest` only checks `From > To`.
- `/traces/decided` — same loop at `exporter/decided.go:35`; `validateDecidedRequest` only checks `From > To`.

So `POST /v1/exporter/traces/committee` (or `/decided`) with `to = 18446744073709551615` still wraps the `uint64` counter (`s++` overflows `MaxUint64 → 0`, `0 <= MaxUint64` stays true) and spins forever.

The range-**size** cap rightly stays in #2986, but the `to == MaxUint64` guaranteed-hang guard is a proven one-liner. Worth hoisting it into shared range validation (or copying these two lines into the other two validators) so all three endpoints are covered now, rather than leaving two of the three hangable. Non-blocking.

api/handlers/exporter/exporter_test.go
```diff
@@ -2671,6 +2722,75 @@ func TestExporterValidatorTraces_ForkGating_CrossForkRange(t *testing.T) {
...
+// TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces covers the fix for a
+// fork-straddling AGGREGATOR/SYNC_COMMITTEE_CONTRIBUTION request without
+// pubkeys/indices whose pre-fork slots legitimately yield zero traces (e.g.
+// sparse aggregator duties): the response must be 200 with empty traces and
+// the post-fork notes surfaced, not a 500. A genuine error alongside those
+// notes must still yield 500.
+func TestExporterValidatorTraces_ForkGating_ZeroPreForkTraces(t *testing.T) {
+	const booleEpoch = phase0.Epoch(5)
+
+	ssvCopy := *networkconfig.TestNetwork.SSV
```
iurii-ssv: [nit, test-only] This new test re-inlines the clone-and-pin fork fixture:

```go
ssvCopy := *networkconfig.TestNetwork.SSV
ssvCopy.Forks.Boole = booleEpoch
netCfg := *networkconfig.TestNetwork
netCfg.SSV = &ssvCopy
```

which is now repeated 4× in this file (here, plus L2453, L2542, L2579). A small local helper — e.g. `booleNetwork(epoch phase0.Epoch)` returning the pinned `netCfg` — would collapse all four and keep the fork-fixture wiring in one place.

(The `preBooleNetwork()` / `postBooleNetwork()` / `straddlingNetwork()` helpers in `exporter/validator_test.go` can't be reused here — that's the separate `github.com/ssvlabs/ssv/exporter` package — so a local one is needed.) Non-blocking.

iurii-ssv (dismissed):
Just couple minor nits

ovidiu-ssv-labs (dismissed):

exporter/validator.go
momosh-ssv: Fixed in 3831c7abc - hoisted the range check into a shared `validateSlotRange` in exporter/validation.go and wired it into all three validators, so /traces/committee and /decideds now reject `to == MaxUint64` too. Added MaxUint64 cases for both. Range-size cap stays in #2986.

api/handlers/exporter/exporter_test.go
momosh-ssv: Fixed in e22aeb914 - added a local `booleNetwork(epoch)` next to `newTestExporterForV2WithNetwork` and collapsed all four sites.

exporter/validator.go
```diff
@@ -47,6 +74,15 @@ func (e *Exporter) ValidatorTracesCore(request *ValidatorTracesQuery) (*Validato
 		}
 	}
 
+	for _, role := range request.Roles {
+		noteFrom, ok := postForkNoteFrom[role]
+		if !ok {
+			continue
+		}
+		delete(postForkNoteFrom, role) // guard against duplicate roles in the request
+		errs = multierror.Append(errs, fmt.Errorf("%w: slots %d-%d: role %s is a committee duty post-fork, %s", ErrPostForkCommitteeDutyNote, noteFrom, request.To, role.String(), committeeDutyFilterHint))
```
iurii-ssv: **[nit, non-blocking]** The surfaced note text repeats itself. The `%w`-wrapped sentinel already reads _"committee duty post-fork requires pubkeys or indices"_, and this format then appends _"… role %s is a committee duty post-fork, "_ + `committeeDutyFilterHint` (_"please provide either pubkeys or indices …"_). So the string a client sees in `errors[]` ends up as:

> committee duty post-fork requires pubkeys or indices: slots 40-42: role AGGREGATOR is a committee duty post-fork, please provide either pubkeys or indices to filter the duty …

i.e. `committee duty post-fork` appears twice and `pubkeys or indices` three times. Since this is user-facing (it's what a dashboard renders), consider trimming the sentinel to a bare marker (e.g. `committee duty post-fork`) and letting the format string carry the actionable hint once. `errors.Is` and the `require.Contains(…, "committee duty post-fork")` assertions are unaffected.

iurii-ssv (dismissed):
Looks good.

ljuba-ssv (dismissed):

exporter/validator.go
momosh-ssv: Fixed in 440ffb736 — the sentinel is now the bare marker "committee duty post-fork" and the format string carries the hint once, so the note reads:

> committee duty post-fork: slots 3757504-3757510: role AGGREGATOR, please provide either pubkeys or indices to filter the duty for a specific validators subset or use the /committee endpoint to query all the corresponding duties

`errors.Is` matching and the `require.Contains(…, "committee duty post-fork")` assertions are untouched, and the OpenAPI description already documented the bare marker.

iurii-ssv (approved):

y0sher (approved):

Issue #2970, opened 2026-07-30
CI: recurring flaky tests in the ssv unit-test job (executionclient, eventhandler, utils/ttl)


momosh-ssv:
The `ssv` (unit-test) CI job fails sporadically on tests unrelated to the PR under test. During the #2952 follow-up PRs (July 21), the job failed on 5+ runs across #2957/#2959/#2961 with identical code passing on rerun — each PR needed 1–3 reruns to go green. Aggregating the known offenders so this stops being re-diagnosed per PR:

## 1. `eth/executionclient` — `TestStreamLogs`, `TestSimSSV`

Timing-sensitive simulator/websocket tests; fail with `timed out waiting for event: context deadline exceeded` on loaded runners, pass on rerun with identical code.

- `TestStreamLogs/successfully_streams_logs`: [run 29821833499](https://github.com/ssvlabs/ssv/actions/runs/29821833499), [run 29833681337](https://github.com/ssvlabs/ssv/actions/runs/29833681337) — failure message `err: context deadline exceeded, streamedLogsCount: 25`, i.e. logs were streaming but not fast enough for the deadline.
- `TestSimSSV`: [run 29821402829](https://github.com/ssvlabs/ssv/actions/runs/29821402829) — `timed out waiting for event`.

`TestStreamLogs` was already reported once in #2234 (closed) — it has regressed to flaky. Likely fix direction: scale the deadlines to the environment or poll-with-timeout instead of a fixed budget.

## 2. `eth/eventhandler` — `TestHandleBlockEventsStream` slot race

`ClusterLiquidated`/`ClusterReactivated` subtests asserted stored slashing-protection values against `EstimatedCurrentSlot()` re-evaluated at assertion time; any slot boundary crossing between the stamp (written during event processing) and the assertion fails with an off-by-one (`expected 0x64 / actual 0x65`, [run 29833681337](https://github.com/ssvlabs/ssv/actions/runs/29833681337)). This one is a genuine test bug, not runner load — **fix: #2969** (window-based assertions).

## 3. `utils/ttl` — `TestHashMap_parallel`

[...]

PR #2976, opened 2026-07-30
eth/executionclient: hoist the shared test env budget into a named const (#2970)

momosh-ssv:
Follow-up to the `eth/executionclient` deflake for #2970 (`TestStreamLogs`, `TestSimSSV`).

The substantive fix — widening all 21 `setupTestEnv` context budgets from 5s to 30s — already landed on `stage` through another path: #2977 (cherry-picked onto `integration/boole-convergence`) → #2941 (merged into `stage`). After merging `origin/stage` back into this branch, those commits cancel out of the diff.

What this PR still carries is the review follow-up from that work: hoisting the repeated `30*time.Second` into a single package-level `setupTestEnvTimeout` const next to `setupTestEnv`, switching all 21 call sites to it, and folding the two scattered inline rationale comments into the const's doc comment. Future tuning becomes a one-liner and the rationale lives in one place. `TestChainReorganizationLogs` keeps its own local 5s `testTimeout` as before.

No behavior change — locally: `go test ./eth/executionclient` clean.

Remaining #2970 item: the `utils/ttl` `TestHashMap_parallel` goroutine-after-test panic.


eth/executionclient/execution_client_test.go
```diff
@@ -908,7 +914,7 @@ func TestFilterLogs(t *testing.T) {
 	logger := zaptest.NewLogger(t)
 
 	t.Run("successfully filters logs", func(t *testing.T) {
-		env := setupTestEnv(t, 5*time.Second)
+		env := setupTestEnv(t, 30*time.Second)
```
iurii-ssv: `TestFilterLogs` is unrelated to the two flaky streaming tests (`TestStreamLogs`, `TestSimSSV`) named in the PR description, yet it's also bumped 5s → 30s here. In fact **all 21 `setupTestEnv` call sites** are widened — the description's "the budget at the two call sites goes 5s → 30s, other tests keep theirs" (and Greptile's matching summary) undercounts the scope. The only site left at 5s is `TestChainReorganizationLogs`, which uses a separate local `testTimeout` const rather than `setupTestEnv`.

The blanket bump is defensible — every one of these tests shares the same `env.ctx`-as-deadline pattern and the same jitter exposure — but please either correct the description to say all `setupTestEnv` sites were widened, or narrow the diff to the two tests the description names.

eth/executionclient/execution_client_test.go
```diff
@@ -163,7 +163,7 @@ func TestFetchHistoricalLogs(t *testing.T) {
 	logger := zaptest.NewLogger(t)
 
 	t.Run("successfully fetches historical logs within follow distance", func(t *testing.T) {
-		env := setupTestEnv(t, 5*time.Second)
+		env := setupTestEnv(t, 30*time.Second)
```
iurii-ssv: Nit / maintainability: `30*time.Second` is now repeated verbatim at all 21 `setupTestEnv` call sites. Since the PR already touches every one, this is a natural moment to hoist it into a package-level `const setupTestEnvTimeout = 30 * time.Second` — future tuning becomes a one-liner and the rationale lives in one place instead of the two scattered inline comments. There's local precedent: `TestChainReorganizationLogs` already declares a named `testTimeout` const.

iurii-ssv (dismissed):
Minor suggestions.

eth/executionclient/execution_client_test.go
momosh-ssv: Good catch.
The description went stale: it was written for the first commit and 7efa9e94 then deliberately widened the remaining sites (same `env.ctx`-as-deadline pattern, same jitter exposure). 

Updated the description to say all 21 `setupTestEnv` sites are now at 30s, with only `TestChainReorganizationLogs` keeping its local 5s const.

eth/executionclient/execution_client_test.go
momosh-ssv: Fixed in 940cbb27a.
Hoisted `setupTestEnvTimeout = 30 * time.Second` next to `setupTestEnv`, switched all 21 call sites to it, and folded the two scattered inline comments into the const's doc comment.

nkryuchkov (changes requested):
The diff shows that the initial duration was 30s. Has the timeout increase been merged in another PR? Should we update this PR's title and description?

momosh-ssv:
> Has the timeout increase been merged in another PR? Should we update this PR's title and description?

Good catch, yes.
The timeout increase reached stage through another path: #2977 carried the same widening commits onto integration/boole-convergence, which then merged into stage via #2941. After I merged origin/stage back into this branch, those commits cancelled out of the diff, so the "before" state here already shows 30s.

Updated the title and description to reflect what this PR still carries: only the follow-up const hoist (setupTestEnvTimeout) from iurii's review, in 940cbb27a.

nkryuchkov (approved):

ljuba-ssv (approved):

y0sher (approved):

PR #2980, opened 2026-08-03
network/discovery: stop undecodable packets from wedging discv5

iurii-ssv:
## Problem

Undecodable discv5 packets could wedge discovery. go-ethereum's post-fork listener forwards packets it can't decode to an `Unhandled` channel with a **blocking send** (`p2p/discover/v5_udp.go`, `handlePacket`), on the same goroutine that drives its UDP reader. The old `SharedUDPConn` served the pre-fork listener's reads directly off that channel, so a burst of junk — scan noise, stray discv4, a deliberate flood — could fill it, block the send, and stall the post-fork listener, which then stopped draining the real socket and the kernel dropped valid discv5 traffic.

## Fix

`SharedUDPConn` now runs an always-receptive `drain` goroutine that moves packets from `Unhandled` into an internal buffer, dropping and counting on overflow instead of blocking. The forwarding send never blocks beyond the handoff, so the socket keeps draining under any load. Reads are served from the buffer.

- `unhandledChanSize` (100) — handoff channel; small because `drain` is always receptive.
- `unhandledBufferSize` (1024) — burst-absorption buffer; overflow is a counted drop, not a stall.

Shutdown is ordered to neither wedge nor panic: stop the producer, close `Unhandled`, `WaitDrained`, then close the socket. `Close` is idempotent (`sync.Once`) — a double close of `Unhandled` would otherwise panic.

## Changes

- `shared_conn.go` — `drain` + buffer, idempotent `Close`, `WaitDrained`, `Dropped`.
- `dv5_service.go` — ordered/idempotent `Close`; error-path cleanup in `initDiscV5Listener`; wire up `NewSharedUDPConn`.
- `observability.go` — new counter `ssv.p2p.discovery.unhandled_packets.dropped`.
- `forking_dv5_listener.go` — corrected stale fork comments (no behaviour change).
- `shared_conn_test.go` / `service_test.go` — regression + cleanup tests: wedge, drain-while-closing, blocked-reader release, init cleanup (rebinds the port to prove the socket was freed), double `Close`.

## Notes

- **`initDiscV5Listener` cleanup (review).** The drain goroutine started before four error returns that leave `dv5Listener` nil, so the caller dropped the half-built service without `Close` — leaking the goroutine/socket (and post-fork listener on the last two paths). Now unwound in `Close` order. Never accumulated in practice: `Setup()` failure hits `logger.Fatal`.
- **`drain`'s ctx is metric-only, by design.** `Close` cancels before joining the producer, so stopping `drain` on ctx would reintroduce the wedge/panic.
- **Buffer stays 1024; pre-fork listener stays.** Measured, not assumed — over 6h, `sum by (ssv_fork) (rate(ssv_p2p_discovery_iterations_total[6h]))`:

  | cluster | post | pre |
  |---|---|---|
  | production-ovh (mainnet + hoodi) | 121.2/s | **121.2/s** |
  | stage | 1060.1/s | 104.6/s |
  | production | 38.5/s | 0.04/s |

  The pre-fork listener is live with a populated routing table and is the only path to nodes on the default protocol ID, so removing it risks cutting them off. Caveat: `ssv.fork` is attributed only on the iterations counter, not `peers.accepted`, so the data can't yet show whether any *accepted* peer needed it — that attribution is what would make removal decidable.

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -348,6 +348,13 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 	return nil
 }
 
+// listenV5 is indirected so tests can wrap or fail listener creation to
+// exercise initDiscV5Listener's error-path cleanup; it returns the Listener
+// interface, not *discover.UDPv5, so a wrapper can be substituted.
+var listenV5 = func(conn discover.UDPConn, ln *enode.LocalNode, cfg discover.Config) (Listener, error) {
+	return discover.ListenV5(conn, ln, cfg)
+}
+
 // initDiscV5Listener creates a new listener and starts it
 func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 	opts := discOpts.DiscV5Opts
@@ -414,7 +421,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		return err
 	}
 
-	dv5PostForkListener, err := discover.ListenV5(udpConn, localNode, *dv5PostForkCfg)
+	dv5PostForkListener, err := listenV5(udpConn, localNode, *dv5PostForkCfg)
 	if err != nil {
 		return fmt.Errorf("could not create discV5 listener: %w", err)
 	}
@@ -434,7 +441,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		return err
 	}
 
-	dv5PreForkListener, err := discover.ListenV5(sharedConn, localNode, *dv5PreForkCfg)
+	dv5PreForkListener, err := listenV5(sharedConn, localNode, *dv5PreForkCfg)
 	if err != nil {
 		return fmt.Errorf("could not create discV5 pre-fork listener: %w", err)
 	}
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -363,6 +363,16 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 	}
 	dvs.conn = udpConn
 
+	// Registered before anything else can fail, so every error path below
+	// releases the socket. Runs last of the deferred cleanups, by which point a
+	// listener may already have closed it.
+	defer func() {
+		if err != nil {
+			_ = udpConn.Close()
+			dvs.conn = nil
+		}
+	}()
+
 	localNode, err := dvs.createLocalNode(discOpts, ipAddr)
 	if err != nil {
 		return fmt.Errorf("could not create local node: %w", err)
@@ -376,7 +386,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 	}
 
 	// New discovery, with ProtocolID restriction, to be kept post-fork
-	unhandled := make(chan discover.ReadPacket, 100) // size taken from https://github.com/ethereum/go-ethereum/blob/v1.13.5/p2p/server.go#L551
+	unhandled := make(chan discover.ReadPacket, unhandledChanSize)
 	sharedConn := NewSharedUDPConn(dvs.ctx, udpConn, unhandled)
 	dvs.sharedConn = sharedConn
 
@@ -396,8 +406,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		_ = sharedConn.Close() // never returns an error
 		close(unhandled)
 		sharedConn.WaitDrained()
-		_ = udpConn.Close() // no-op once the listener above has closed it
-		dvs.sharedConn, dvs.conn = nil, nil
+		dvs.sharedConn = nil
 	}()
 
 	dv5PostForkCfg, err := opts.DiscV5Cfg(dvs.logger, WithProtocolID(protocolID), WithUnhandled(unhandled))
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -111,6 +111,10 @@ func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (
 		zap.Any("hostDNS", opts.HostDNS),
 	)
 	if err := dvs.initDiscV5Listener(opts); err != nil {
+		// initDiscV5Listener unwinds its own resources, but the context we
+		// derived above is ours to release on this failure path — the discarded
+		// service's Close (which would cancel it) never runs.
+		cancel()
 		return nil, err
 	}
 	return &dvs, nil
@@ -394,7 +398,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 
 	// New discovery, with ProtocolID restriction, to be kept post-fork
 	unhandled := make(chan discover.ReadPacket, unhandledChanSize)
-	sharedConn := NewSharedUDPConn(dvs.ctx, udpConn, unhandled)
+	sharedConn := NewSharedUDPConn(dvs.ctx, dvs.logger, udpConn, unhandled)
 	dvs.sharedConn = sharedConn
 
 	// Everything below can fail before dv5Listener is set, and the caller drops
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/p2p/discover"
@@ -80,6 +81,9 @@ type DiscV5Service struct {
 	subnets   commons.Subnets
 
 	publishLock chan struct{}
+
+	closeOnce sync.Once
+	closeErr  error
 }
 
 func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (*DiscV5Service, error) {
@@ -112,21 +116,37 @@ func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (
 	return &dvs, nil
 }
 
-// Close implements io.Closer
+// Close implements io.Closer. It is safe to call more than once; callers above
+// don't guard against that, and closing sharedConn.Unhandled twice would panic.
 func (dvs *DiscV5Service) Close() error {
+	dvs.closeOnce.Do(func() {
+		dvs.closeErr = dvs.close()
+	})
+	return dvs.closeErr
+}
+
+func (dvs *DiscV5Service) close() error {
 	if dvs.cancel != nil {
 		dvs.cancel()
 	}
-	if dvs.conn != nil {
-		if err := dvs.conn.Close(); err != nil {
-			return err
-		}
+	// Order matters. The post-fork listener produces into sharedConn.Unhandled,
+	// so it must be fully stopped before that channel is closed — closing it
+	// mid-send panics the producer. Stopping the listeners first is safe:
+	// SharedUDPConn.Close releases the pre-fork reader, and draining continues
+	// throughout so the post-fork listener cannot wedge on its way down.
+	if dvs.dv5Listener != nil {
+		dvs.dv5Listener.Close()
 	}
 	if dvs.sharedConn != nil {
 		close(dvs.sharedConn.Unhandled)
+		dvs.sharedConn.WaitDrained()
 	}
-	if dvs.dv5Listener != nil {
-		dvs.dv5Listener.Close()
+	if dvs.conn != nil {
+		// The post-fork listener owns this socket and closes it on shutdown, so
+		// finding it already closed is expected rather than an error.
+		if err := dvs.conn.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
+			return err
+		}
 	}
 	return nil
 }
@@ -357,7 +377,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) error {
 
 	// New discovery, with ProtocolID restriction, to be kept post-fork
 	unhandled := make(chan discover.ReadPacket, 100) // size taken from https://github.com/ethereum/go-ethereum/blob/v1.13.5/p2p/server.go#L551
-	sharedConn := &SharedUDPConn{udpConn, unhandled}
+	sharedConn := NewSharedUDPConn(dvs.ctx, udpConn, unhandled)
 	dvs.sharedConn = sharedConn
 
 	dv5PostForkCfg, err := opts.DiscV5Cfg(dvs.logger, WithProtocolID(protocolID), WithUnhandled(unhandled))
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -356,7 +356,13 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 // exercise initDiscV5Listener's error-path cleanup; it returns the Listener
 // interface, not *discover.UDPv5, so a wrapper can be substituted.
 var listenV5 = func(conn discover.UDPConn, ln *enode.LocalNode, cfg discover.Config) (Listener, error) {
-	return discover.ListenV5(conn, ln, cfg)
+	// Return an untyped nil on error, never a (*discover.UDPv5)(nil) wrapped in
+	// the interface, so a nil check on the result is meaningful for any caller.
+	listener, err := discover.ListenV5(conn, ln, cfg)
+	if err != nil {
+		return nil, err
+	}
+	return listener, nil
 }
 
 // initDiscV5Listener creates a new listener and starts it
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -120,8 +120,10 @@ func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (
 	return &dvs, nil
 }
 
-// Close implements io.Closer. It is safe to call more than once; callers above
-// don't guard against that, and closing sharedConn.Unhandled twice would panic.
+// Close implements io.Closer and is idempotent: a repeat call would panic
+// re-closing sharedConn.Unhandled, so the guard lives here. The sole production
+// caller (p2pNetwork.Close) already closes once, but tests and any future caller
+// reach this directly.
 func (dvs *DiscV5Service) Close() error {
 	dvs.closeOnce.Do(func() {
 		dvs.closeErr = dvs.close()
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -349,7 +349,7 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 }
 
 // initDiscV5Listener creates a new listener and starts it
-func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) error {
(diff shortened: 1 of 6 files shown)


network/discovery/shared_conn.go
```diff
@@ -1,33 +1,131 @@
...
+// Draining must outlive cancellation: DiscV5Service.Close cancels before it
+// joins the producer, so stopping on ctx.Done here would let the post-fork
+// listener wedge on a full channel or panic sending to a closed one. drain
+// stops only when Unhandled is closed, which its owner does after that join.
+func (s *SharedUDPConn) drain(ctx context.Context) {
+	defer s.drained.Done()
+	for packet := range s.Unhandled {
+		select {
+		case s.buffered <- packet:
+		default:
```
momosh-ssv, line 114: What do you think about dropping the oldest packet here instead of the newest? Under a sustained junk flood the buffer pins at 1024 and stays full, so the pre-fork listener works through a stale backlog while every fresh packet — including real responses to our own pings — gets discarded, and the staleness outlasts the flood itself. Something like `select { case <-s.buffered: default: }` before retrying the send would keep the freshest packet winning; the wedge fix comes from the non-blocking send, not the buffer depth, so this wouldn't weaken it.

network/discovery/shared_conn.go
```diff
@@ -1,33 +1,131 @@
...
+			s.dropped.Add(1)
+			recordUnhandledPacketDropped(ctx)
```
momosh-ssv: Might be worth a rate-limited warn on the transition into a dropping state — the counter is the only signal today and `Dropped()` has no production caller, so an operator watching pre-fork discovery degrade has nothing in the logs pointing at this path.

network/discovery/service_test.go
```diff
@@ -53,6 +53,12 @@ func TestDiscV5Service_Close(t *testing.T) {
 
 	err := dvs.Close()
 	assert.NoError(t, err)
+
+	// Callers above don't guard against closing twice, and the second pass
+	// would panic closing sharedConn.Unhandled again.
+	assert.NotPanics(t, func() {
```
momosh-ssv, line 59: Should we consider a test that pins the close() ordering itself? This covers idempotency, but if close() ever reverts to closing `Unhandled` before `dv5Listener.Close()` (the pre-PR order), every test still passes and the send-on-closed-channel panic only surfaces in production when junk arrives during shutdown. A fake post-fork listener that keeps sending on `unhandled` while its `Close()` runs would catch a reorder.

network/discovery/dv5_service.go
```diff
@@ -112,21 +116,37 @@ func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (
 	return &dvs, nil
```
momosh-ssv, line 120: Seems that `cancel` is never called when `initDiscV5Listener` fails just above, so the derived context stays registered in the parent for the node's lifetime. Maybe a defer-on-error (or `cancel()` before the early return) would round out the error-path cleanup — it looks like the one resource the cleanup commits didn't cover.

network/discovery/dv5_service.go
```diff
@@ -354,7 +374,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) error {
 
 	// New discovery, with ProtocolID restriction, to be kept post-fork
 	unhandled := make(chan discover.ReadPacket, 100) // size taken from https://github.com/ethereum/go-ethereum/blob/v1.13.5/p2p/server.go#L551
-	sharedConn := &SharedUDPConn{udpConn, unhandled}
+	sharedConn := NewSharedUDPConn(dvs.ctx, udpConn, unhandled)
 	dvs.sharedConn = sharedConn
```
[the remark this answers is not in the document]
iurii-ssv: Addressed by the error-path cleanup added after the commit this was reviewed against.

`initDiscV5Listener` now registers a deferred cleanup immediately after `NewSharedUDPConn`, so every error return below it — including the malformed-bootnode failure from `DiscV5Cfg` — closes the shared conn, closes `unhandled`, and calls `WaitDrained()` before returning. The drain goroutine is joined and the socket/channel are released rather than leaked.

Covered by `TestInitDiscV5Listener_CleansUpOnError` (the `malformed bootnode` case), which reproduces this exact scenario and fails against the pre-fix code.

network/discovery/shared_conn.go
iurii-ssv: Good call — switched to evicting the oldest. `drain` now routes each packet through `bufferPacket`, which on a full buffer receives one (the oldest) and retries the send, so the freshest packet always wins. The drop counter increments once per packet actually evicted — if a concurrent read frees a slot first, nothing is dropped. Added `TestSharedUDPConn_DropsOldestWhenFull` to pin the keep-freshest behavior; it fails against the old tail-drop.

network/discovery/shared_conn.go
iurii-ssv: Added a rate-limited `warnDropping` in `drain` — at most one line per `dropWarnInterval` (1m), logging the running dropped total — so the transition into shedding surfaces in the logs, not just the counter. `SharedUDPConn` now takes a `*zap.Logger` for this.

network/discovery/service_test.go
iurii-ssv: Added `TestDiscV5Service_CloseStopsProducerBeforeClosingUnhandled`: a fake post-fork listener keeps forwarding on `unhandled` until its `Close()` is called (mirroring how `UDPv5.Close` joins its goroutines before returning), and the test asserts `dvs.Close()` doesn't panic. Confirmed it's a real guard — reverting `close()` to close `Unhandled` before `dv5Listener.Close()` makes it panic with "send on closed channel".

network/discovery/dv5_service.go
iurii-ssv: Fixed — `NewDiscV5Service` now calls `cancel()` before returning on the `initDiscV5Listener` error path. You're right that this was the one resource the cleanup commits missed: `initDiscV5Listener` unwinds its own resources, but the derived context is owned by the constructor, and the discarded service's `Close()` (which would have cancelled it) never runs.

momosh-ssv (dismissed):

momosh-ssv (dismissed):

ovidiu-ssv-labs:
### Nothing pins the "`drain` must not stop on ctx" invariant

Reviewed `670cfb4..019c9ce`. The premise checks out line-for-line against go-ethereum v1.16.4 (`v5_udp.go:750` is a bare `t.unhandled <- up` with no default and no `closeCtx` case; `readNextCh` cap-1 genuinely gates `readLoop` on `handlePacket` returning; `UDPv5.Close` does join its goroutines, so the producer-stopped guarantee is real). I reproduced the old wedge in a throwaway module and it blocks at exactly 100 packets = `unhandledChanSize`, so the regression tests do bite. The concurrency in this PR is sound on every interleaving I could construct.

One gap worth closing before merge.

**`shared_conn.go:87`** — the ctx-is-metric-only decision is correct and load-bearing, and it's stronger than the PR body claims: because the `unhandled` send has no `closeCtx` escape, a ctx-sensitive `drain` wouldn't merely re-wedge discovery, it would deadlock **node shutdown** — `p2pNetwork.Close` cancels at `p2p.go:236` and only calls `disc.Close()` at `p2p.go:246`, so dispatch parks on a full channel and `postFork.Close()`'s `wg.Wait()` (`v5_udp.go:210`) never returns.

But nothing tests it. Every construction site passes `context.Background()` (`service_test.go:106`; `shared_conn_test.go:47,77,102,124,288,324`) and the close-order test installs `cancel: func() {}` (`service_test.go:112`), so the ctx is never cancelled anywhere in the suite. The 5-line comment at `shared_conn.go:82-86` is the only thing defending the invariant, and comments don't fail CI. The single most likely future refactor — adding the "obvious" `case <-ctx.Done(): return` to `drain` — reintroduces a shutdown deadlock **with a fully green suite**. I confirmed that: injecting exactly that into `drain` leaves `go test -race ./network/discovery/` passing (`ok 8.351s`).

**The fix that works.** I tried the two natural formulations that go through `Close()` and both are unreliable — worth recording so nobody spends the time twice:

- Making the existing `TestDiscV5Service_CloseStopsProducerBeforeClosingUnhandled` ctx use a real `context.WithCancel` + `dvs.cancel = cancel` does **not** catch it. `producingListener`'s goroutine selects on `<-p.stop` *alongside* the send (`service_test.go:84-88`), so it always escapes a full channel.
- Even with a producer whose send is unescapable (stop checked only *between* sends, mirroring `handlePacket`), it's a race: `cancel()` and `close(p.stop)` land microseconds apart at `dv5_service.go:134` and `:142`, so the producer usually exits before it can fill 100 slots. Passed 3/3 for me with the regression injected — a false negative.

What is deterministic is asserting the property directly, with no `Close()` and no producer-shutdown race — a mirror of `TestSharedUDPConn_ForwardingNeverBlocks` with an already-cancelled ctx:

```go
// TestSharedUDPConn_DrainsAfterCancel pins the invariant documented on drain:
// ctx is a metric context, not a stop signal. DiscV5Service.close cancels before
// it joins the producer, so a drain that honoured cancellation would let the
// post-fork listener park in its unescapable forwarding send — deadlocking
// shutdown in UDPv5.Close's wg.Wait, not just stalling discovery.
func TestSharedUDPConn_DrainsAfterCancel(t *testing.T) {
	unhandled := make(chan discover.ReadPacket, unhandledChanSize)
	ctx, cancel := context.WithCancel(context.Background())
	cancel()
	conn := NewSharedUDPConn(ctx, zap.NewNop(), nil, unhandled)

	const total = unhandledChanSize + unhandledBufferSize + 5000
	var sent atomic.Int64
	done := make(chan struct{})
	go func() {
		defer close(done)
		forwardBlocking(unhandled, total, &sent)
	}()

	select {
	case <-done:
	case <-time.After(10 * time.Second):
		t.Fatalf("forwarding blocked after %d/%d packets with a cancelled ctx: "+
			"drain must outlive cancellation or Close deadlocks joining the producer",
			sent.Load(), total)
	}

	close(unhandled)
	conn.WaitDrained()
}
```

Verified both directions in a worktree at `019c9ce`:

- regression injected (`case <-ctx.Done(): return` in `drain`) → **FAIL 3/3**, `blocked after 101/6124 packets` (100 channel + 1 in drain's hand)
- unmodified `019c9ce` → **PASS 5/5** under `-race`, 0.00s

Reuses `forwardBlocking` as-is, no new helpers, no sleeps, no timing assumptions.

<details>
<summary>Verification run on 019c9ce (unmodified)</summary>

`go build ./network/...`, `go vet ./network/discovery/`, `gofmt -l network/discovery/` — all clean.
`go test -race -count=1 ./network/discovery/` — `ok 8.688s`.
`-race -count=20 -run TestSharedUDPConn` — 20/20 on all 6 tests. `-race -count=10 -run TestInitDiscV5Listener_CleansUpOnError` — 30/30 subtests, the fixed-port rebind never flaked. `-race -shuffle=on` ×3 — clean.

network/discovery/shared_conn.go
```diff
@@ -1,33 +1,175 @@
...
+		total := s.dropped.Add(1)
+		recordUnhandledPacketDropped(ctx)
+		s.warnDropping(total)
+	}
+}
+
+// bufferPacket enqueues packet for the reader. When the buffer is full it evicts
+// the oldest packet so the freshest one wins — a real ping response must not be
+// starved by a backlog of junk during a flood. It reports whether a packet was
+// dropped to make room.
```
ovidiu-ssv-labs: **Minor:** the eviction rationale doesn't quite survive the FIFO read.

Evicting the oldest keeps the newest 1024, but `ReadFromUDPAddrPort` drains FIFO (L151), so a genuine ping response is *retained* yet sits behind up to 1023 stale packets. It's never worse than tail-drop, so the choice itself is fine — but "the freshest survives" oversells what 1024 + FIFO buys: with geth's `V5RespTimeout` at 700ms and the pre-fork listener consuming one packet per dispatch cycle, a persistently-full buffer can outlast the RPC anyway. A materially smaller depth (or a LIFO read) is what would actually favour freshness.

Not asking for a behaviour change — the wedge fix comes from the non-blocking send, as you note. Just worth a clause here so the next reader doesn't over-trust the rationale and conclude 1024 is protecting response latency.

network/discovery/shared_conn.go
```diff
@@ -1,33 +1,175 @@
...
+func (s *SharedUDPConn) bufferPacket(packet discover.ReadPacket) (dropped bool) {
+	for {
+		select {
+		case s.buffered <- packet:
+			return dropped
+		default:
+		}
+		// Full: evict the oldest and retry. A concurrent read may free a slot
+		// first, in which case nothing is dropped.
```
ovidiu-ssv-labs: **Minor:** this describes a path the code doesn't take.

The loop does retry, but the first pass goes send-fail → evict with no re-check in between, so the evict fires whenever `buffered` is non-empty. Interleaving: the send fails at 1024/1024, the pre-fork reader consumes one (slot now free), and `case <-s.buffered` *still* fires — a real packet is discarded and counted with space available.

"nothing is dropped" only holds when the buffer goes fully **empty** (the evict select's `default` fires and the loop retries the send), which is unreachable in the flood this comment is about. Harmless in aggregate — steady-state eviction count is still `P − R` — but a comment that contradicts the code is worse than none in a concurrency primitive.

Either retry the send once before evicting, or reword to say the newcomer displaces the oldest unconditionally when full.

network/discovery/dv5_service.go
```diff
@@ -107,26 +111,46 @@ func NewDiscV5Service(pctx context.Context, logger *zap.Logger, opts *Options) (
...
+		// service's Close (which would cancel it) never runs.
+		cancel()
 		return nil, err
 	}
 	return &dvs, nil
 }
 
-// Close implements io.Closer
+// Close implements io.Closer. It is safe to call more than once; callers above
+// don't guard against that, and closing sharedConn.Unhandled twice would panic.
```
ovidiu-ssv-labs: **Minor:** "callers above don't guard against that" isn't accurate — the only production caller does. `p2pNetwork.Close` wraps the whole teardown in `n.closeOnce.Do(...)` (`network/p2p/p2p.go:232`), so the double call can't arrive from there today.

The `sync.Once` is still worth keeping — a double close is a panic rather than a soft error, and tests plus any future caller reach this directly — so this is only about the stated reason. Something like "a second close would panic closing sharedConn.Unhandled, so guard it here rather than relying on callers" says the same thing without the claim about callers.

ovidiu-ssv-labs (commented):
Three Minor notes to go with the earlier comment — all comment/code mismatches, no behaviour change needed. Nothing here blocks; happy for them to ride in a follow-up. (All three are in code this PR introduces: base `shared_conn.go` had no buffer or eviction, and the idempotent `Close` is new.)

ovidiu-ssv-labs (dismissed):
Nothing blocks, but one more test is worth to be added as per my comment above.
You can decide whether to go ahead or not.

network/discovery/service_test.go
```diff
@@ -53,6 +53,72 @@ func TestDiscV5Service_Close(t *testing.T) {
...
+}
+
+// TestDiscV5Service_CloseStopsProducerBeforeClosingUnhandled pins the shutdown
+// order: close() must stop the listeners (the producers into Unhandled) before
+// closing Unhandled. If it ever reverts to closing Unhandled first, the still-live
+// producer panics with "send on closed channel" — which the idempotency test
+// above cannot catch, having no live producer.
+func TestDiscV5Service_CloseStopsProducerBeforeClosingUnhandled(t *testing.T) {
+	unhandled := make(chan discover.ReadPacket, unhandledChanSize)
+	sharedConn := NewSharedUDPConn(context.Background(), zap.NewNop(), nil, unhandled)
```
nkryuchkov: Can we use `t.Context()` here, or do we need an uncancellable context?

network/discovery/shared_conn_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	close(unhandled)
+	conn.WaitDrained()
+}
+
+// TestSharedUDPConn_DeliversPacketsToReader covers the normal path: what the
+// producer forwards is what the pre-fork listener reads.
+func TestSharedUDPConn_DeliversPacketsToReader(t *testing.T) {
+	unhandled := make(chan discover.ReadPacket, unhandledChanSize)
+	conn := NewSharedUDPConn(context.Background(), zap.NewNop(), nil, unhandled)
```
nkryuchkov: nit: `t.Context()`?

nkryuchkov (dismissed):

network/discovery/shared_conn.go
iurii-ssv: Reworded — dropped the "freshest wins" framing. The doc now says eviction bounds staleness but doesn't guarantee a prompt read: reads are FIFO, so a retained response can sit behind up to `unhandledBufferSize-1` older packets, and a persistently-full buffer can outlast `V5RespTimeout` regardless of depth — the wedge fix is the non-blocking send, not this buffer. Also aligned the type-level comment and the `DropsOldestWhenFull` test, which carried the same phrasing.

network/discovery/shared_conn.go
iurii-ssv: Reworded to match the code: the evict is unconditional while anything is buffered, so a newcomer displaces the oldest even if a concurrent read just freed a slot; `dropped` stays false only if the buffer emptied meanwhile (the evict's `default`), which the flood never reaches.

Kept the behaviour as-is — retrying the send once before evicting would remove the harmless over-evict but only adds a race window for no real gain, so not worth it. Agreed it isn't a behaviour change.

network/discovery/dv5_service.go
iurii-ssv: Fixed. Confirmed `p2pNetwork.Close` wraps teardown in `closeOnce.Do` (`network/p2p/p2p.go`), so "callers above don't guard" was wrong. The comment now says the `sync.Once` is there because a repeat close would panic re-closing `sharedConn.Unhandled`, and tests / any future caller reach `Close` directly even though the sole production caller already closes once. Also corrected the same claim mirrored in `TestDiscV5Service_Close`.

network/discovery/service_test.go
iurii-ssv: Switched to `t.Context()`. No uncancellable context needed here — `drain` treats ctx as metric-only and stops only when `Unhandled` closes, so cleanup-time cancellation is harmless. The one place that genuinely needs a pre-cancelled context is the new `TestSharedUDPConn_DrainsAfterCancel`, which keeps its own `context.WithCancel`.

network/discovery/shared_conn_test.go
iurii-ssv: Done — converted all the construction sites in both files to `t.Context()` for consistency (the file already used it elsewhere), not just this one.

iurii-ssv:
@ovidiu-ssv-labs Added `TestSharedUDPConn_DrainsAfterCancel` — thanks for writing it out, and for recording the two dead-ends so nobody re-walks them. Kept your direct formulation (already-cancelled ctx, no `Close()`, no producer-shutdown race), reusing `forwardBlocking` as-is.

Verified both directions on the pushed commit:
- inject `case <-ctx.Done(): return` into `drain` → **FAIL**, `blocked after 100/6124 packets` (100 channel + 1 in drain's hand)
- unmodified → **PASS** under `-race`, including `-count=15 -shuffle=on`

The test's doc note calls out why it deliberately keeps `context.Background()` where the rest of the suite moved to `t.Context()`.

iurii-ssv:
@greptile pls re-review

momosh-ssv (approved):

ovidiu-ssv-labs (approved):

PR #2981, opened 2026-08-05
network/discovery: accept NextDomainType in checkPeer for cross-client compatibility

### LINKED ISSUE #2967: Boole: discovery filter rejects Anchor ENRs post-fork (checkPeer not fork-aware) — cross-client edges only ever Anchor-initiated
## Summary

Post-Boole, go-ssv's discovery filter cannot match Anchor. Anchor advertises the active domain in its main `domaintype` key, while `checkPeer` compares that against the static `ssvConfig.DomainType`, so the comparison never succeeds. Anchor fixed the reverse direction in [sigp/anchor#1187](https://github.com/sigp/anchor/pull/1187), so cross-client edges do form, but **only Anchor can ever create one**.

We would like to ask you to revisit #2952 item 4 against a condition it did not weigh. Background and the full audit are in [sigp/anchor#1185](https://github.com/sigp/anchor/issues/1185).

## Evidence from hoodi-stage

`hi-h-sv-net2` (operator 417) restarted post-Boole on 2026-07-28 and stalled at 3 peers against `target_peers=69` with zero go-ssv peers. On 2026-07-29 00:18:46 UTC we put the #1187 build on that node alone, keeping operators 415/416/418 on v1.3.0 as a control since they still held warm pre-fork connections.

| | before | after |
|---|---:|---:|
| peers (in / out) | 3 (1 / 2) | 7 (1 / 6) |
| distinct go-ssv ENRs surfaced by discovery | 0 | 7 |
| connected go-ssv peers | 0 | 4 |

Deriving peer IDs from each discovered ENR's `secp256k1` entry and matching them against `Handshake completed` lines: of the 7 connections, 3 are Anchor and 4 are go-ssv, and **Anchor dialed all four go-ssv peers**. The single inbound connection is another Anchor node. We attempted 23 distinct dials in the window and no go-ssv node dialed us. No trusted or static peers are configured on these hosts, so all of it came through normal discv5.

Every go-ssv ENR carried `domaintype=00003114` with `next_domaintype=00003115`, and no Anchor ENR carried a next domain. So the stage fleet is populating `NextDomainType` correctly and #1187 consumes it.

## The ask

At `network/discovery/dv5_service.go:279` (PR #2941 head `7de8ab0f`), widen the comparison to accept either configured domain:

```go
// Accept a candidate on either configured domain, so peers that advertise the
// post-fork domain in their main key remain discoverable across the transition.
domainMatches := nodeDomainType == dvs.ssvConfig.DomainType ||
    (dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
        nodeDomainType == dvs.ssvConfig.NextDomainType)

if !domainMatches {
    recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
    return newPeerSkipError(skipReasonDomainTypeMismatch,
        fmt.Errorf("domain type %x matches neither %x nor %x",
            nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType))
}
```

- **No structural change.** `ssvConfig.NextDomainType` is on the same struct (`networkconfig/ssv.go:40`) and already used at `:498` and `:571`. One production call site, no `Options` change, no slot clock.
- **Strictly widens an accept-set**, so it cannot partition a homogeneous fleet, and no existing discovery test result changes.
- **The zero-value guard is load-bearing.** `UnmarshalJSON` defaults `NextDomainType` to `DomainType` only for parsed configs (`ssv.go:125-127`); the struct literals in the discovery tests leave it `[4]byte{}`, where a bare `||` would start accepting an all-zero domain.

It also makes `records.KeyNextDomainType` do something: today it is written at `:498` and `:571` and, as far as we can tell, never read anywhere in the non-test tree.

The fuller version is your own `isDomainCompatible` from `boole-fork`, which uses the runtime `CurrentDomainType()` and the transition window. That is the better end state, but it needs `DiscV5Service` to hold the whole `*networkconfig.Network` for the slot clock, so we are not asking for it here.

## Why we think it is worth revisiting

#2952 item 4 closed deliberately, with a condition attached:

> only port `isDomainCompatible` if full mixed-fleet interop during testnet migration is required

The case weighed there was interop with older `boole-fork` binaries during a migration. That is transient, and we agree it would not justify the change on its own. Anchor is a permanent second implementation that advertises the active domain rather than a static one, so this is the steady state rather than a window that closes.

## Mainnet exposure

Mainnet is pre-Boole, so both clients are on Alan, discovery is symmetric, and that symmetry currently supplies most of our connectivity:

| operator | peers | target | in | out | inbound share |
|---|---:|---:|---:|---:|---:|
| 2042 | 56 | 69 | 43 | 13 | 77% |
| 2043 | 70 | 72 | 56 | 14 | 80% |
| 2044 | 40 | 69 | 27 | 13 | 68% |
| 2045 | 60 | 69 | 45 | 15 | 75% |
| 948 | 22 | 63 | 18 | 4 | 82% |

Our nodes sustain only 13-15 outbound connections each, and 4 on operator 948. Anchor also stops running discovery once `connected_peers >= target_peers`, where the target is `min(60 + 3*subnets, 150)`, and operator 2043 is already at 70 of 72. On a densely peered network that combination leaves a restarted go-ssv node with no route back: it cannot discover us, and we may have stopped looking.

If Boole activates on mainnet with the current filter, the inbound path stops being replenished. Warm connections survive the transition, as they did on hoodi-stage in July, so it would look clean at first and then degrade with restarts and ordinary churn.

momosh-ssv:
Fixes #2967. Related: sigp/anchor#1185 (audit), sigp/anchor#1187 (Anchor-side fix).

## Problem

Post-Boole, `checkPeer` compares the peer's main `domaintype` ENR entry strictly against the static `ssvConfig.DomainType`. go-ssv keeps advertising the static current domain in its main key, while Anchor advertises the *active* domain — Boole post-fork. This is a permanent steady-state mismatch, not a transition window: go-ssv can never discover or dial Anchor peers, so cross-client edges form only when Anchor initiates. Live confirmation on stage-hoodi: `domainTypeMismatch` is the top discovery skip reason fleet-wide, and handshake logs show `want one of [0x00003115]`.

The handshake filter is already fork-aware, so discovery is the only blocker — widening it makes the connection succeed end-to-end.

## Change

`checkPeer` now accepts a peer whose advertised main domain equals either `DomainType` or `NextDomainType`. `NextDomainType` is only honored when non-zero: config parsing defaults it to `DomainType`, but struct-literal configs leave it zero, and an all-zero advertised domain must not be accepted.

Strictly widens the accept set — one production call site, no structural changes. The fork-aware handshake filter and per-slot message validation remain the authoritative fork enforcement.

## Tests

- ENR with main `domaintype == NextDomainType` (post-fork Anchor shape) passes
- ENR with main `domaintype == DomainType` still passes
- All-zero main domain rejected when `NextDomainType` is unset (zero-value guard)
- Unrelated domain (e.g. prod-hoodi's `0x00000503`) still rejected with `domainTypeMismatch`
- Peer's *next*-domain ENR key remains ignored (only the main key is compared)

## Post-merge verification (stage)

Restart a couple of stage nodes and confirm they establish outbound connections to the Anchor operators (415–418). sigp's acceptance criterion from sigp/anchor#1185: a post-fork restarted node forms ≥1 direct go-ssv→Anchor outbound edge.

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -16,6 +16,8 @@ import (
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
+	spectypes "github.com/ssvlabs/ssv-spec/types"
+
 	"github.com/ssvlabs/ssv/network/commons"
 	"github.com/ssvlabs/ssv/network/peers"
 	"github.com/ssvlabs/ssv/network/records"
@@ -266,19 +268,26 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 		return newPeerSkipError(skipReasonNotSSV, errors.New("node is not an SSV node"))
 	}
 
-	// Get the peer's domain type, skipping if it mismatches ours.
-	// Discovery filtering intentionally stays on the static current domain: it's symmetric
-	// within a converged fleet, and real fork enforcement happens in the fork-aware handshake
-	// filter and per-slot message validation, not here.
+	// Get the peer's domain type, skipping unless it matches our current or next domain.
+	// We advertise the static current domain in our main ENR key, but fork-aware clients
+	// (e.g. Anchor) advertise the active domain, which becomes the next domain after a fork —
+	// so a strict match against the current domain would reject them forever. Real fork
+	// enforcement happens in the fork-aware handshake filter and per-slot message validation,
+	// not here.
 	peerDiscoveriesCounter.Add(ctx, 1)
 	nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyDomainType)
 	if err != nil {
 		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, fmt.Errorf("could not read domain type: %w", err))
 	}
-	if dvs.ssvConfig.DomainType != nodeDomainType {
+	// Only accept NextDomainType when it's set: config parsing defaults it to DomainType,
+	// but a zero value must not make us accept peers advertising an all-zero domain.
+	domainMatches := nodeDomainType == dvs.ssvConfig.DomainType ||
+		(dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
+			nodeDomainType == dvs.ssvConfig.NextDomainType)
+	if !domainMatches {
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
-		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x doesn't match %x", nodeDomainType, dvs.ssvConfig.DomainType))
+		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType))
 	}
 
 	// Get the peer's subnets, skipping if it has none.
@@ -493,8 +502,8 @@ func (dvs *DiscV5Service) PublishENR() {
 		return
 	}
 	// KeyNextDomainType carries the real upcoming domain (not the current one) so that
-	// fork-aware (boole-style) binaries and future dynamic domain flips can rely on it,
-	// even though our own peer filter only ever compares against the static current domain.
+	// fork-aware (boole-style) binaries and future dynamic domain flips can rely on it;
+	// our own discovery filter accepts peers advertising either the current or next domain.
 	err = records.SetDomainTypeEntry(dvs.dv5Listener.LocalNode(), records.KeyNextDomainType, dvs.ssvConfig.NextDomainType)
 	if err != nil {
 		dvs.logger.Error("could not set next domain type", zap.Error(err))
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -283,14 +283,14 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 	// Only accept NextDomainType when it's set to a distinct value: config parsing defaults
 	// it to DomainType, and a zero value must not make us accept peers advertising an
 	// all-zero domain.
-	nextDomainAccepted := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
+	hasDistinctNextDomain := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
 		dvs.ssvConfig.NextDomainType != dvs.ssvConfig.DomainType
 	matchesDomain := nodeDomainType == dvs.ssvConfig.DomainType
-	matchesNextDomain := nextDomainAccepted && nodeDomainType == dvs.ssvConfig.NextDomainType
+	matchesNextDomain := hasDistinctNextDomain && nodeDomainType == dvs.ssvConfig.NextDomainType
 	if !matchesDomain && !matchesNextDomain {
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
 		var err error
-		if nextDomainAccepted {
+		if hasDistinctNextDomain {
 			err = fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType)
 		} else {
 			err = fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -289,9 +289,11 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 	matchesNextDomain := nextDomainAccepted && nodeDomainType == dvs.ssvConfig.NextDomainType
 	if !matchesDomain && !matchesNextDomain {
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
-		err := fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
+		var err error
 		if nextDomainAccepted {
 			err = fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType)
+		} else {
+			err = fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
 		}
 		return newPeerSkipError(skipReasonDomainTypeMismatch, err)
 	}
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -280,14 +280,20 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
 		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, fmt.Errorf("could not read domain type: %w", err))
 	}
-	// Only accept NextDomainType when it's set: config parsing defaults it to DomainType,
-	// but a zero value must not make us accept peers advertising an all-zero domain.
-	domainMatches := nodeDomainType == dvs.ssvConfig.DomainType ||
-		(dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
-			nodeDomainType == dvs.ssvConfig.NextDomainType)
-	if !domainMatches {
+	// Only accept NextDomainType when it's set to a distinct value: config parsing defaults
+	// it to DomainType, and a zero value must not make us accept peers advertising an
+	// all-zero domain.
+	nextDomainAccepted := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
+		dvs.ssvConfig.NextDomainType != dvs.ssvConfig.DomainType
+	matchesDomain := nodeDomainType == dvs.ssvConfig.DomainType
+	matchesNextDomain := nextDomainAccepted && nodeDomainType == dvs.ssvConfig.NextDomainType
+	if !matchesDomain && !matchesNextDomain {
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
-		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType))
+		err := fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
+		if nextDomainAccepted {
+			err = fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType)
+		}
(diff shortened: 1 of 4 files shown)


network/discovery/dv5_service.go
```diff
@@ -266,19 +268,26 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, fmt.Errorf("could not read domain type: %w", err))
 	}
-	if dvs.ssvConfig.DomainType != nodeDomainType {
+	// Only accept NextDomainType when it's set: config parsing defaults it to DomainType,
+	// but a zero value must not make us accept peers advertising an all-zero domain.
+	domainMatches := nodeDomainType == dvs.ssvConfig.DomainType ||
+		(dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
+			nodeDomainType == dvs.ssvConfig.NextDomainType)
+	if !domainMatches {
```
nkryuchkov: nit: Do you think two booleans might be a bit more readable here? 
```suggestion
	matchesDomain := nodeDomainType == dvs.ssvConfig.DomainType
	matchesNextDomain := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) && nodeDomainType == dvs.ssvConfig.NextDomainType

	if !matchesDomain && !matchesNextDomain {
```

nkryuchkov (dismissed):

network/discovery/dv5_service.go
```diff
@@ -266,19 +268,26 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
-		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x doesn't match %x", nodeDomainType, dvs.ssvConfig.DomainType))
+		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType))
```
iurii-ssv: nit (non-blocking): on the zero-guard path (`NextDomainType` unset), or when `NextDomainType == DomainType` (the config-parse default), this logs `domain type <x> matches neither 01020304 nor 00000000` (or `... nor 01020304`), which reads as if `00000000`/the duplicate were an accepted alternative. Consider only appending the `nor %x` clause when a distinct next domain is actually accepted. Cosmetic only — the accept/reject decision itself is correct.

network/discovery/dv5_service.go
```diff
@@ -266,19 +268,26 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 	nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyDomainType)
 	if err != nil {
 		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, fmt.Errorf("could not read domain type: %w", err))
 	}
-	if dvs.ssvConfig.DomainType != nodeDomainType {
+	// Only accept NextDomainType when it's set: config parsing defaults it to DomainType,
+	// but a zero value must not make us accept peers advertising an all-zero domain.
+	domainMatches := nodeDomainType == dvs.ssvConfig.DomainType ||
+		(dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
```
iurii-ssv: A future or custom struct-literal config that omits `NextDomainType` will silently fall back to strict current-domain discovery — cross-client edges won't form, with no error, since the zero-guard just skips this branch. A config-validation check, or a doc note that networks needing cross-client discovery must set `NextDomainType`, would prevent that footgun.

Minor: discovery accepts `NextDomainType` unconditionally, whereas the handshake gates its dual-domain accept on `InBooleTransitionWindow`. That's fine (the handshake stays authoritative); the only cost is a rare wasted dial if a peer advertises the next domain pre-fork.

iurii-ssv (commented):
Minor suggestions.

network/discovery/dv5_service.go
nkryuchkov: https://github.com/ssvlabs/ssv/pull/2981#discussion_r3720007169 should make it a bit easier to implement

network/discovery/dv5_service.go
momosh-ssv: Fixed in dc2f264ce.
Split into `matchesDomain` / `matchesNextDomain`. I added one extra condition to `matchesNextDomain` (next domain must differ from the current one) so the error message below can tell the two cases apart; behavior is unchanged.

network/discovery/dv5_service.go
momosh-ssv: Fixed in dc2f264ce.
The "nor %x" clause is now only appended when a distinct non-zero next domain is actually accepted; otherwise it's a plain "does not match %x". Updated the zero-guard test assertion to match.

network/discovery/dv5_service.go
momosh-ssv: Fixed in e4fad014b.
Went with the doc note: the `NextDomainType` field now documents that zero means strict current-domain discovery and that networks needing cross-client discovery must set it. 

I'd rather not add config validation here, since zero is a legitimate "no next domain" state that the guard exists to support. And agreed on the handshake staying authoritative. The occasional wasted dial pre-fork seems like an acceptable cost for simpler discovery.

nkryuchkov (dismissed):

network/discovery/dv5_service.go
```diff
@@ -266,19 +268,32 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
+	// it to DomainType, and a zero value must not make us accept peers advertising an
+	// all-zero domain.
+	nextDomainAccepted := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
+		dvs.ssvConfig.NextDomainType != dvs.ssvConfig.DomainType
+	matchesDomain := nodeDomainType == dvs.ssvConfig.DomainType
+	matchesNextDomain := nextDomainAccepted && nodeDomainType == dvs.ssvConfig.NextDomainType
+	if !matchesDomain && !matchesNextDomain {
 		recordPeerSkipped(ctx, skipReasonDomainTypeMismatch)
-		return newPeerSkipError(skipReasonDomainTypeMismatch, fmt.Errorf("domain type %x doesn't match %x", nodeDomainType, dvs.ssvConfig.DomainType))
+		err := fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
```
iurii-ssv: nit (non-blocking, feel free to ignore): building on the now-resolved wording thread above — the current form still constructs the `"does not match %x"` error unconditionally and then discards it whenever `nextDomainAccepted` is true. Since every real network config sets a distinct non-zero `NextDomainType`, `nextDomainAccepted` is true in production, so this first `fmt.Errorf` is thrown away on *every* domain-mismatch reject — and per the PR description `domainTypeMismatch` is the top discovery skip reason fleet-wide. An `if/else` avoids the throwaway and reads a touch cleaner:

```go
var err error
if nextDomainAccepted {
	err = fmt.Errorf("domain type %x matches neither %x nor %x", nodeDomainType, dvs.ssvConfig.DomainType, dvs.ssvConfig.NextDomainType)
} else {
	err = fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
}
return newPeerSkipError(skipReasonDomainTypeMismatch, err)
```

iurii-ssv (dismissed):
LGTM (1 optional nit below)

network/discovery/dv5_service.go
momosh-ssv: Fixed in ca41687a5.
Switched to if/else so only the applicable error is constructed.

network/discovery/dv5_service.go
```diff
@@ -266,19 +268,34 @@ func (dvs *DiscV5Service) checkPeer(ctx context.Context, e PeerEvent) error {
...
 	nodeDomainType, err := records.GetDomainTypeEntry(e.Node.Record(), records.KeyDomainType)
 	if err != nil {
 		recordPeerSkipped(ctx, skipReasonInvalidDomainType)
 		return newPeerSkipError(skipReasonInvalidDomainType, fmt.Errorf("could not read domain type: %w", err))
 	}
-	if dvs.ssvConfig.DomainType != nodeDomainType {
+	// Only accept NextDomainType when it's set to a distinct value: config parsing defaults
+	// it to DomainType, and a zero value must not make us accept peers advertising an
+	// all-zero domain.
+	nextDomainAccepted := dvs.ssvConfig.NextDomainType != (spectypes.DomainType{}) &&
```
nkryuchkov: `nextDomainAccepted` looks a bit unclear to be, where do we accept it from?
What about `nextDomainConfigured` or `nextDomainSet`?

network/discovery/dv5_service_test.go
```diff
@@ -49,7 +49,19 @@ func TestCheckPeer(t *testing.T) {
...
+				domainType:    &spectypes.DomainType{0x1, 0x2, 0x3, 0x6},
+				subnets:       mySubnets,
+				expectedError: nil,
+			},
+			{
+				name:          "unrelated domain type",
+				domainType:    &spectypes.DomainType{0x0, 0x0, 0x5, 0x3},
+				subnets:       mySubnets,
+				expectedError: errors.New("domain type 00000503 matches neither 01020304 nor 01020306"),
 			},
```
nkryuchkov, line 65: another case for this else clause would be great:
```go
		} else {
			err = fmt.Errorf("domain type %x does not match %x", nodeDomainType, dvs.ssvConfig.DomainType)
		}
```

networkconfig/ssv.go
```diff
@@ -35,8 +35,14 @@ type SSV struct {
...
-	Name                 string
-	DomainType           spectypes.DomainType
+	Name       string
+	DomainType spectypes.DomainType
+	// NextDomainType is the domain type expected to become active after the next fork.
+	// Discovery accepts peers advertising either DomainType or a distinct non-zero
+	// NextDomainType, so networks that need cross-client discovery around a fork must
+	// set it. Zero means no next domain (discovery is strict to DomainType). YAML/JSON
+	// parsing defaults it to DomainType when omitted; only configs built from struct
+	// literals can leave it zero.
```
nkryuchkov, line 47: nit: If we add such a big comment for `NextDomainType`, should we perhaps leave at least a few words for `DomainType` for similarity?

network/discovery/dv5_service.go
momosh-ssv: Fixed in 60662e5ee.
Renamed to `hasDistinctNextDomain`. I went with that over `nextDomainSet`/`nextDomainConfigured` since config parsing defaults `NextDomainType` to `DomainType`.

So it can be set yet still excluded by the `!= DomainType` condition, and "set"/"configured" would read misleadingly in exactly that case. Happy to adjust if you prefer another name.

network/discovery/dv5_service_test.go
momosh-ssv: Fixed in 089524d41.
Generalized the zero-guard test into a table (`TestCheckPeer_NoDistinctNextDomain`) so the else branch is now covered by three cases: zero next domain + zero peer domain, zero next domain + mismatched peer domain, and next domain equal to the current one + mismatched peer domain.

networkconfig/ssv.go
momosh-ssv: Fixed in db16f9696.
Added a short note on `DomainType` — active domain, advertised in the primary ENR entry, domain-separates message signatures.

iurii-ssv (approved):

ljuba-ssv (approved):

PR #2982, opened 2026-08-06
network: detect a wedged discv5 socket (operator health check + boot-node /healthz)

### LINKED ISSUE #2979: boot-node: discv5 can wedge silently — no health signal tied to discovery actually working
## Summary

A boot node can stop serving discovery while every external health signal still reports it as healthy. We recently had a boot node in this state for roughly 20 days without detection. Nothing in the boot node reports on whether discv5 is actually working, so the failure is invisible until a node somewhere has to cold-bootstrap and finds no peers.

## What happened

The boot node's discv5 UDP socket was wedged: the socket stayed bound, but the application had stopped draining it. Inbound `FINDNODE` requests queued in the kernel receive buffer and were eventually dropped, so every lookup against it returned nothing and its discv5 routing table stayed permanently empty.

Meanwhile the process was alive and the HTTP server on the TCP port kept answering normally, so from the outside the boot node looked fine.

The cluster it serves kept working the whole time on warm peer tables. When those nodes were eventually restarted for an unrelated reason, they cold-bootstrapped, found no peers, and the cluster partitioned — 20 days after the actual fault.

## Why it hid

`bootNode.Start()` runs the discv5 listener and the HTTP server as independent goroutines (`utils/boot_node/node.go`). `httpServer.ListenAndServe()` continuing to serve says nothing about whether discv5's read loop is still draining the UDP socket. Any liveness check pointed at the TCP port therefore passes while discovery is completely dead.

This is also why it was hard to diagnose from outside: TCP reachable, process up, logs clean (the boot node logs nothing after startup), ENR unchanged.

## Proposal

**Expose a health signal tied to discovery actually functioning, and make it fail closed.**

The boot node already has everything needed. The `/p2p` handler calls `listener.AllNodes()` and dumps the discv5 routing table. For a boot node specifically, an empty routing table is definitionally broken — its entire job is to know about other nodes.

Concretely:

- Add a health endpoint (or a status code on the existing one) that returns non-200 when the routing table has been empty for longer than some grace period, so a Kubernetes liveness probe restarts the pod automatically. A bounce is a complete fix for this failure mode, so self-healing is achievable rather than just alerting.
- The grace period needs to tolerate genuine cold start, where the table is legitimately empty for a short while.

This would have turned a 20-day silent outage into a self-correcting blip.

As a datapoint on how clearly this discriminates, comparing routing table sizes across our boot nodes at the time:

| boot node | `AllNodes()` |
|---|---|
| wedged | **0** |
| healthy | 72 |
| healthy | 102 |

## Related

- Separately, there was a reachable path by which the operator node's own discv5 stack could wedge with the same observable signature, via the blocking send to the `Unhandled` channel in the fork-listener setup. **(Fixed in #2980 — `SharedUDPConn` now drains `Unhandled` through an always-receptive goroutine that drops-and-counts on overflow instead of blocking.)**

## Resolution plan

Two things to separate: the *wedge* itself, and the *lack of detection* (this issue's core).

**Wedge — operator-node path addressed by #2980.** The operator's discv5 wedging via the blocking send to the `Unhandled` channel (see *Related* above) is fixed there, so an undecodable-packet flood can no longer stall the socket. The boot node was never exposed to this path — it runs a single `discover.ListenV5` with no `Unhandled` channel — so its wedge had a different cause, which is exactly why detection (not just this one mechanism) is what matters.

**Detection — missing on both node types.** The operator has the same blind spot as the boot node: `p2pNetwork.Healthy()` only checks readiness plus a discovery-*bootstrap*-failed flag, which a runtime wedge never trips (the bootstrap loop keeps running, just yielding nothing). The difference is that the operator already has a self-heal path — the `hprobe` watchdog restarts the node on persistent unhealth — while the boot node has none.

**Decided design — one shared signal, scoped actuation:**

- **Shared primitive:** a read-timestamp wrapper on the discv5 socket. "How long since we last drained a packet" is the direct, cause-agnostic wedge signal, and both node types own the conn they hand to `ListenV5`.
- **Operator:** feed staleness into `p2pNetwork.Healthy()` → the existing `hprobe` watchdog restarts the node. No routing-table check — a stale-but-populated table would mask a wedge.
- **Boot node:** a fail-closed health check that returns non-200 when the socket read is stale **or** the routing table has been empty past a cold-start grace → Kubernetes liveness restarts the pod. Empty-table is the boot node's definitional health (the 0-vs-72/102 datapoint above).
- **Starting grace values:** ~3 min read-staleness (both node types); ~10 min empty-table cold-start (boot node only).

**Work — implemented in #2982:**

1. Shared read-timestamp primitive (`TimedConn`).
2. Operator: wrap the post-fork socket, extend `Healthy()` → `hprobe` restart.
3. Boot node: fail-closed health endpoint.

iurii-ssv:
## Problem

A discv5 socket can wedge — stay bound but stop being drained — leaving discovery silently dead while the process still looks healthy. #2979 documents a boot node that sat like this for ~20 days undetected. Neither node type surfaces it today: the operator's `p2pNetwork.Healthy()` only checks a discovery-*bootstrap* flag (a runtime wedge never trips it, since the bootstrap loop keeps running and just yields nothing), and the boot node's HTTP handler always returns `200`.

#2980 removes one *cause* of the wedge on the operator (the blocking `Unhandled` send); this PR adds the missing *detection*, so a wedge from any cause becomes a self-correcting restart.

## Approach

One shared, cause-agnostic signal — "how long since the discv5 socket was last read" — with actuation scoped per node type.

- **`TimedConn`** (`network/discovery`): wraps the socket and stamps the last successful read; `StaleFor(d)` is the wedge signal, armed only once the socket has been read at least once (a node that never had inbound UDP isn't wedged — a restart can't fix that). Both node types wrap the conn they hand to `ListenV5`. Read staleness is also exported as an async gauge (`ssv.p2p.discovery.socket.read_staleness`, sampled at scrape time; `-1` = never read).
- **Operator**: the post-fork listener (the one that drains the socket) gets the wrapped conn; `DiscoveryStale` feeds `p2pNetwork.Healthy()`, and the existing `hprobe` watchdog restarts the node on a wedge — but only while the connected-peer set is degraded (< 10 peers). Staleness alone can't distinguish a wedged read loop from inbound UDP lost upstream, which a restart can't fix; a well-connected node logs an error and stays up, tripping the probe later only if its peers erode. No routing-table check here — a stale-but-populated table would mask it on an operator.
- **Boot node**: a fail-closed `/healthz` returns non-200 when the socket has gone unread while the routing table is populated (discv5 revalidates its peers, so a populated-but-unread socket is a wedge), or when the table has been empty past a cold-start grace while the socket is undrained. Empty-table is the boot node's definitional health (the 0-vs-72/102 datapoint in #2979) — with one deliberate exception: an empty table while the socket is actively drained is a config/compatibility mismatch (e.g. `DiscoveryProtocolID`) a restart can't fix, so it stays 200 with a loud error log instead of crash-looping the pod. Table state is sampled by an owned 1s ticker (probe-independent empty-table clock, cheap requests); both HTTP endpoints are GET/HEAD-only and the 503 body is fixed, with the reason logged.

Grace values: 3 min read-staleness (both node types); 10 min empty-table cold-start and 1s table sampling (boot node); 10-peer floor gating the operator restart.

## Notes

- **Was stacked on #2980** (the operator change wraps the socket in `initDiscV5Listener`, which #2980 rewrote); #2980 has merged and this PR targets `stage` directly.
- Out of scope (dropped from #2979 as not worth the effort): boot-node `/debug/pprof`, and a second `sepolia` boot node.
- Infra follow-up (not code in this repo): the boot node has no pod-restarting probe today (only an AWS LB TCP check, which is blind to the wedge); a k8s `livenessProbe` on `/healthz` is added by the downstream PRs below.

## Tests

- `TimedConn`: seeded-not-stale, never-read-never-stale, stale boundary (injected clock, no real sleeps), read-stamps, errored-read-doesn't-stamp, gauge flattening (`-1` sentinel).
- Operator: `DiscoveryStale`; `TestP2PNetwork_Healthy` covers the wedge across peer-set states — degraded / nil-host / floor-boundary fail the probe, wedged-but-well-connected passes it (no restart loop on an external UDP fault) — plus `ready with live discovery`; the existing nil-`disc` cases stay green. `TestInitDiscV5Listener_WrapsPostForkConn` pins the wiring the detector rests on.
- Boot node: `/healthz` across populated+fresh, populated+wedged, empty-within-grace (incl. the quiet-socket regression), and empty-past-grace undrained (503) vs drained (200 + error log, asserted via a zap observer); the sampler clock is probe-independent and `check` reads samples only; method filter (405, HEAD ok) and fixed 503 body.

Closes #2979. Full discovery + p2p + boot_node suites pass, discovery also under `-race`.

## Downstream infra PRs for the boot-node side of this (what makes the boot node self-healing)

- ssvlabs/charts#183 — adds the optional `livenessProbe` to the `boot-node` chart (inert by default).
- ssvlabs/gitops-production#881 — adopts it on the three boot nodes (draft; blocked on a boot-node image from this PR that serves `/healthz`).

Merge order: #2980 → this → build image → ssvlabs/charts#183 → ssvlabs/gitops-production#881.

### DIFF
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -78,6 +78,9 @@ type DiscV5Service struct {
 
 	conn       *net.UDPConn
 	sharedConn *SharedUDPConn
+	// socketConn wraps conn to time socket reads; the post-fork listener drains
+	// through it, so DiscoveryStale can detect a wedged socket.
+	socketConn *TimedConn
 
 	ssvConfig *networkconfig.SSV
 	subnets   commons.Subnets
@@ -157,6 +160,21 @@ func (dvs *DiscV5Service) close() error {
 		}
 	}
 	return nil
+}
+
+// DiscoveryStale reports whether the discv5 socket has gone unread for longer
+// than grace — the sign discovery has wedged. False before the listener is built
+// (or after a failed init), when there's nothing to judge.
+//
+// The operator health check polls this periodically, so it doubles as the
+// sampling point for the read-staleness gauge.
+func (dvs *DiscV5Service) DiscoveryStale(grace time.Duration) bool {
+	if dvs.socketConn == nil {
+		return false
+	}
+	age, ok := dvs.socketConn.ReadStaleness()
+	recordDiscoveryReadStaleness(dvs.ctx, int64(age.Seconds()))
+	return ok && age > grace
 }
 
 // Self returns self node
@@ -402,6 +420,12 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 	}
 	dvs.conn = udpConn
 
+	// Wrap the socket for liveness (see DiscoveryStale): only the post-fork
+	// listener drains it, so only it gets the wrapped conn — the pre-fork
+	// listener reads sharedConn's buffer, not the socket.
+	socketConn := NewTimedConn(udpConn)
+	dvs.socketConn = socketConn
+
 	// Registered before anything else can fail, so every error path below
 	// releases the socket. Runs last of the deferred cleanups, by which point a
 	// listener may already have closed it.
@@ -409,6 +433,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		if err != nil {
 			_ = udpConn.Close()
 			dvs.conn = nil
+			dvs.socketConn = nil
 		}
 	}()
 
@@ -453,7 +478,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		return err
 	}
 
-	dv5PostForkListener, err := listenV5(udpConn, localNode, *dv5PostForkCfg)
+	dv5PostForkListener, err := listenV5(socketConn, localNode, *dv5PostForkCfg)
 	if err != nil {
 		return fmt.Errorf("could not create discV5 listener: %w", err)
 	}
--- a/network/discovery/dv5_service.go
+++ b/network/discovery/dv5_service.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
@@ -81,6 +82,9 @@ type DiscV5Service struct {
 	// socketConn wraps conn to time socket reads; the post-fork listener drains
 	// through it, so DiscoveryStale can detect a wedged socket.
 	socketConn *TimedConn
+	// stalenessReg is the read-staleness gauge callback for socketConn,
+	// unregistered on close.
+	stalenessReg metric.Registration
 
 	ssvConfig *networkconfig.SSV
 	subnets   commons.Subnets
@@ -140,6 +144,10 @@ func (dvs *DiscV5Service) close() error {
 	if dvs.cancel != nil {
 		dvs.cancel()
 	}
+	if dvs.stalenessReg != nil {
+		// Stop reporting the read-staleness gauge for a closed service.
+		_ = dvs.stalenessReg.Unregister()
+	}
 	// Order matters. The post-fork listener produces into sharedConn.Unhandled,
 	// so it must be fully stopped before that channel is closed — closing it
 	// mid-send panics the producer. Stopping the listeners first is safe:
@@ -164,17 +172,14 @@ func (dvs *DiscV5Service) close() error {
 
 // DiscoveryStale reports whether the discv5 socket has gone unread for longer
 // than grace — the sign discovery has wedged. False before the listener is built
-// (or after a failed init), when there's nothing to judge.
-//
-// The operator health check polls this periodically, so it doubles as the
-// sampling point for the read-staleness gauge.
+// (or after a failed init), when there's nothing to judge. The read-staleness
+// gauge is sampled separately at scrape time (see observeDiscoveryReadStaleness),
+// keeping this a pure predicate.
 func (dvs *DiscV5Service) DiscoveryStale(grace time.Duration) bool {
 	if dvs.socketConn == nil {
 		return false
 	}
-	age, ok := dvs.socketConn.ReadStaleness()
-	recordDiscoveryReadStaleness(dvs.ctx, int64(age.Seconds()))
-	return ok && age > grace
+	return dvs.socketConn.StaleFor(grace)
 }
 
 // Self returns self node
@@ -512,6 +517,15 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 
 	dvs.dv5Listener = NewForkingDV5Listener(dvs.logger, dv5PreForkListener, dv5PostForkListener, 5*time.Second)
 	dvs.bootnodes = dv5PreForkCfg.Bootnodes // Just take bootnodes from one of the config since they're equal
+
+	// Registered last so the callback only outlives init if the service does.
+	// Registration fails only on instrument misuse and the gauge is optional,
+	// so a failure is logged, not returned.
+	if reg, regErr := observeDiscoveryReadStaleness(socketConn); regErr != nil {
+		dvs.logger.Warn("could not register discovery read-staleness gauge", zap.Error(regErr))
+	} else {
+		dvs.stalenessReg = reg
+	}
 
 	return nil
 }
--- a/network/discovery/timed_conn.go
+++ b/network/discovery/timed_conn.go
@@ -16,6 +16,13 @@ import (
 // shows up as a last-read timestamp that stops advancing. StaleFor turns that
 // into a liveness signal, used by both the operator and boot nodes.
 //
+// The signal is scoped to the socket. On the operator only the post-fork
+// listener reads it (through this wrapper); the pre-fork listener consumes
+// decode-rejected packets relayed via SharedUDPConn's buffer. A stall confined
+// to the pre-fork listener therefore moves neither this timestamp nor the
+// post-fork reader, whose relay into that buffer drops rather than blocks
+// (#2980). The boot node runs a single listener reading the socket directly.
+//
 // Only ReadFromUDPAddrPort is overridden; writes, Close and LocalAddr fall
 // through to the embedded conn.
 type TimedConn struct {
--- a/network/discovery/timed_conn.go
+++ b/network/discovery/timed_conn.go
@@ -1,0 +1,84 @@
+package discovery
+
+import (
+	"net"
+	"net/netip"
+	"sync/atomic"
+	"time"
+
+	"github.com/ethereum/go-ethereum/p2p/discover"
+)
+
+// TimedConn wraps a discv5 UDP socket and records when it was last read from.
+//
+// A discv5 listener drains its socket only through ReadFromUDPAddrPort, so a
+// wedged socket — bound but no longer read, leaving discovery silently dead —
+// shows up as a last-read timestamp that stops advancing. StaleFor turns that
+// into a liveness signal, used by both the operator and boot nodes.
+//
+// Only ReadFromUDPAddrPort is overridden; writes, Close and LocalAddr fall
+// through to the embedded conn.
+type TimedConn struct {
+	discover.UDPConn
+
+	lastReadUnixNano atomic.Int64
+	read             atomic.Bool      // set once the socket has been read at least once
+	now              func() time.Time // overridable in tests; nil means time.Now
+}
+
+var _ discover.UDPConn = (*TimedConn)(nil)
+
+// NewTimedConn wraps conn, seeding the last-read time so LastRead reports a
+// sensible value before the first read. Staleness only arms once the socket has
+// actually been read (see StaleFor).
+func NewTimedConn(conn *net.UDPConn) *TimedConn {
+	c := &TimedConn{UDPConn: conn}
+	c.lastReadUnixNano.Store(c.nowFn().UnixNano())
+	return c
+}
+
+func (c *TimedConn) nowFn() time.Time {
+	if c.now != nil {
+		return c.now()
+	}
+	return time.Now()
+}
+
+// ReadFromUDPAddrPort delegates and, on success, records the read time and marks
+// the socket as having been read. Errors leave both untouched, so a closed or
+// failing socket can't look freshly drained.
+func (c *TimedConn) ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error) {
+	n, addr, err = c.UDPConn.ReadFromUDPAddrPort(b)
+	if err == nil {
+		c.lastReadUnixNano.Store(c.nowFn().UnixNano())
+		c.read.Store(true)
+	}
+	return n, addr, err
+}
+
+// LastRead reports when a packet was last read from the socket.
+func (c *TimedConn) LastRead() time.Time {
+	return time.Unix(0, c.lastReadUnixNano.Load())
+}
+
+// ReadStaleness reports how long the socket has gone unread, and whether it has
+// ever been read. Before the first successful read there is nothing to measure,
+// so ok is false: a socket that has never delivered a packet is not wedged, it
+// has simply had no inbound discv5 traffic yet (unreachable bootnodes, blocked
+// UDP), which a restart cannot fix.
+func (c *TimedConn) ReadStaleness() (age time.Duration, ok bool) {
+	if !c.read.Load() {
+		return 0, false
+	}
+	return c.nowFn().Sub(c.LastRead()), true
+}
+
+// StaleFor reports whether the socket has wedged: it was being drained, then
+// went unread for longer than d. A socket that has never been read is never
+// stale — a wedge can only arise after draining has started, so requiring a
+// prior read rules out the false positive (a healthy node that has simply had
+// no inbound traffic yet) without missing any real wedge.
+func (c *TimedConn) StaleFor(d time.Duration) bool {
+	age, ok := c.ReadStaleness()
+	return ok && age > d
+}
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -545,6 +545,11 @@ func (n *p2pNetwork) isReady() bool {
 	return atomic.LoadInt32(&n.state) == stateReady
 }
 
+// discoveryStaleGrace is how long the discv5 socket may go unread before Healthy
+// treats discovery as wedged — comfortably above normal quiet on a live network,
+// where inbound traffic and query responses keep the socket busy.
+const discoveryStaleGrace = 3 * time.Minute
+
 // Healthy reports whether the p2p network is operating normally.
 // It satisfies the health-check interface from hprobe package.
 func (n *p2pNetwork) Healthy(ctx context.Context) error {
@@ -556,6 +561,16 @@ func (n *p2pNetwork) Healthy(ctx context.Context) error {
 	}
 	if n.discoveryFailed.Load() {
 		return fmt.Errorf("discovery bootstrap failed")
+	}
+	// A wedged discv5 socket leaves discovery silently dead while bootstrap keeps
+	// looping; surface it so the hprobe watchdog restarts the node. n.disc is nil
+	// in tests and briefly at startup, hence the guard.
+	if n.disc != nil && n.disc.DiscoveryStale(discoveryStaleGrace) {
+		// Surface the wedge in the log alongside the exit reason, so a restart
+		// isn't the only evidence that discovery died.
+		n.logger.Warn("discv5 socket wedged: not drained within grace, discovery stalled",
+			zap.Duration("grace", discoveryStaleGrace))
+		return fmt.Errorf("discv5 socket not drained for >%s (discovery wedged)", discoveryStaleGrace)
 	}
 	return nil
 }
--- a/network/p2p/p2p.go
+++ b/network/p2p/p2p.go
@@ -550,6 +550,16 @@ func (n *p2pNetwork) isReady() bool {
 // where inbound traffic and query responses keep the socket busy.
 const discoveryStaleGrace = 3 * time.Minute
 
+// discoveryStalePeerFloor decides what a stale discv5 socket means: below it
+// the node is effectively isolated and a restart is worth attempting; at or
+// above it the node keeps serving duties over its established connections.
+// The floor is needed because staleness alone can't tell a wedged read loop
+// from inbound UDP lost upstream (firewall/NAT changes) — a restart can't fix
+// the latter, and since any stray packet re-arms the check after boot,
+// restarting unconditionally would turn one external UDP fault into a
+// fleet-correlated restart loop.
+const discoveryStalePeerFloor = 10
+
 // Healthy reports whether the p2p network is operating normally.
 // It satisfies the health-check interface from hprobe package.
 func (n *p2pNetwork) Healthy(ctx context.Context) error {
@@ -563,14 +573,26 @@ func (n *p2pNetwork) Healthy(ctx context.Context) error {
 		return fmt.Errorf("discovery bootstrap failed")
 	}
 	// A wedged discv5 socket leaves discovery silently dead while bootstrap keeps
-	// looping; surface it so the hprobe watchdog restarts the node. n.disc is nil
-	// in tests and briefly at startup, hence the guard.
+	// looping; failing here makes the hprobe watchdog restart the node. Restart
+	// only when the wedge is actually biting — the peer set has degraded (see
+	// discoveryStalePeerFloor). n.disc is nil in tests and briefly at startup,
+	// hence the guard.
 	if n.disc != nil && n.disc.DiscoveryStale(discoveryStaleGrace) {
-		// Surface the wedge in the log alongside the exit reason, so a restart
-		// isn't the only evidence that discovery died.
-		n.logger.Warn("discv5 socket wedged: not drained within grace, discovery stalled",
-			zap.Duration("grace", discoveryStaleGrace))
-		return fmt.Errorf("discv5 socket not drained for >%s (discovery wedged)", discoveryStaleGrace)
+		peers := 0
+		if h := n.Host(); h != nil {
+			peers = len(h.Network().Peers())
+		}
+		if peers < discoveryStalePeerFloor {
+			// Surface the wedge in the log alongside the exit reason, so a restart
+			// isn't the only evidence that discovery died.
+			n.logger.Warn("discv5 socket wedged and peer set degraded, failing health check",
+				zap.Duration("grace", discoveryStaleGrace), zap.Int("peers", peers))
+			return fmt.Errorf("discv5 socket not drained for >%s with %d peers (discovery wedged)", discoveryStaleGrace, peers)
+		}
+		// This log is the only signal: the node stays up on its established peers
+		// and fails the probe only if they erode.
+		n.logger.Error("discv5 socket not drained within grace, holding restart while peer set is healthy",
+			zap.Duration("grace", discoveryStaleGrace), zap.Int("peers", peers))
 	}
 	return nil
 }
--- a/utils/boot_node/health.go
+++ b/utils/boot_node/health.go
@@ -1,0 +1,97 @@
+package bootnode
+
+import (
+	"fmt"
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+)
+
+const (
+	// bootReadStaleGrace is how long the socket may go unread before /healthz
+	// fails. A boot node is queried continuously, so a wedge stops reads well
+	// inside this.
+	bootReadStaleGrace = 3 * time.Minute
+
+	// bootEmptyTableGrace tolerates cold start, when the table is briefly empty
+	// before the boot node discovers peers.
+	bootEmptyTableGrace = 10 * time.Minute
+)
+
+// nodeLister is the slice of discovery.Listener the health check needs.
+type nodeLister interface {
+	AllNodes() []*enode.Node
+}
+
+// socketDrainState reports socket-read staleness; satisfied by *discovery.TimedConn.
+type socketDrainState interface {
+	StaleFor(time.Duration) bool
+}
+
+// bootNodeHealth backs /healthz. A boot node's whole job is discovery, so it's
+// broken if its socket has wedged or its routing table has been empty past cold
+// start; either fails /healthz closed so a liveness probe restarts the pod.
+type bootNodeHealth struct {
+	lister          nodeLister
+	socket          socketDrainState
+	readStaleGrace  time.Duration
+	emptyTableGrace time.Duration
+	now             func() time.Time
+	lastNonEmpty    atomic.Int64 // unix nanos of the last observation with a non-empty table
+}
+
+func newBootNodeHealth(lister nodeLister, socket socketDrainState) *bootNodeHealth {
+	h := &bootNodeHealth{
+		lister:          lister,
+		socket:          socket,
+		readStaleGrace:  bootReadStaleGrace,
+		emptyTableGrace: bootEmptyTableGrace,
+		now:             time.Now,
+	}
+	h.lastNonEmpty.Store(h.now().UnixNano())
+	return h
+}
+
+// check returns a reason when discovery looks broken, or nil when healthy.
+//
+// The socket-staleness check only applies while the routing table is populated:
+// discv5 revalidates those peers and should be reading their responses, so a
+// stale socket then means the read loop has wedged. An empty table generates no
+// such traffic, so staleness there is expected (cold start, or a quiet node with
+// no peers) and is judged only by the empty-table grace, which tolerates cold
+// start. lastNonEmpty is seeded at startup, so it catches both "never populated"
+// and "populated then emptied".
+//
+// One case slips past the fast path: a wedge present from boot. On restart discv5
+// loads seed nodes from the persistent enode DB straight into the table, so
+// AllNodes() can be >0 before the socket has been read once — and StaleFor stays
+// disarmed until that first read. Those seeds then fail revalidation and age out,
+// the table empties, and the empty-table grace trips instead. A runtime wedge
+// (socket drained, then stopped) is unaffected: a prior read has already armed
+// StaleFor.
+func (h *bootNodeHealth) check() error {
+	now := h.now()
+	if len(h.lister.AllNodes()) > 0 {
+		h.lastNonEmpty.Store(now.UnixNano())
+		if h.socket.StaleFor(h.readStaleGrace) {
+			return fmt.Errorf("discv5 socket not drained for >%s while routing table is populated", h.readStaleGrace)
+		}
+		return nil
+	}
+	if emptyFor := now.Sub(time.Unix(0, h.lastNonEmpty.Load())); emptyFor > h.emptyTableGrace {
+		return fmt.Errorf("discv5 routing table empty for >%s", h.emptyTableGrace)
+	}
+	return nil
+}
+
+func (h *bootNodeHealth) handler() http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		if err := h.check(); err != nil {
+			http.Error(w, err.Error(), http.StatusServiceUnavailable)
+			return
+		}
+		_, _ = w.Write([]byte("ok"))
+	}
+}
--- a/utils/boot_node/health.go
+++ b/utils/boot_node/health.go
@@ -1,12 +1,14 @@
 package bootnode
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"go.uber.org/zap"
 )
 
 const (
@@ -18,6 +20,10 @@ const (
 	// bootEmptyTableGrace tolerates cold start, when the table is briefly empty
 	// before the boot node discovers peers.
 	bootEmptyTableGrace = 10 * time.Minute
+
+	// bootHealthSamplePeriod is how often the routing table is sampled; health
+	// verdicts read the samples (see sample).
+	bootHealthSamplePeriod = time.Second
 )
 
 // nodeLister is the slice of discovery.Listener the health check needs.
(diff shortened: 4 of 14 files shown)


utils/boot_node/health.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	h.lastNonEmpty.Store(h.now().UnixNano())
+	return h
+}
+
+// check returns a reason when discovery looks broken, or nil when healthy.
+func (h *bootNodeHealth) check() error {
+	if h.socket.StaleFor(h.readStaleGrace) {
+		return fmt.Errorf("discv5 socket not drained for >%s", h.readStaleGrace)
+	}
```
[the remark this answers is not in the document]
iurii-ssv: Addressed by gating the socket-staleness check on a **populated** routing table: discv5 revalidates its table peers and reads their pong responses, so a stale socket while the table is populated is a genuine wedge. An empty table (cold start, or a genuinely quiet node with no peers) produces no such traffic, so it's judged only by the cold-start-tolerant empty-table grace — a healthy quiet node is no longer flagged. Added a regression test for the empty-table + stale-socket case.

utils/boot_node/node.go
```diff
@@ -169,20 +169,22 @@ func (n *bootNode) createListener(ipAddr string, port uint16, privateKey *ecdsa.
...
 	}
 
-	listener, err := discover.ListenV5(conn, localNode, discover.Config{
+	listener, err := discover.ListenV5(socketConn, localNode, discover.Config{
 		PrivateKey:   privateKey,
 		V5ProtocolID: &n.ssvConfig.DiscoveryProtocolID,
 	})
 	if err != nil {
-		n.logger.Fatal("Filed to create UDPv5 listener", zap.Error(err))
+		n.logger.Fatal("Failed to create UDPv5 listener", zap.Error(err))
```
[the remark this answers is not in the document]
iurii-ssv: Applied — lowercased to `failed to create UDPv5 listener`. (Its sibling fatals in this file are still capitalized; left as pre-existing, out of scope for this PR.)

network/p2p/p2p.go
```diff
@@ -544,6 +549,12 @@ func (n *p2pNetwork) Healthy(ctx context.Context) error {
 	if n.discoveryFailed.Load() {
 		return fmt.Errorf("discovery bootstrap failed")
 	}
+	// A wedged discv5 socket leaves discovery silently dead while bootstrap keeps
+	// looping; surface it so the hprobe watchdog restarts the node. n.disc is nil
+	// in tests and briefly at startup, hence the guard.
+	if n.disc != nil && n.disc.DiscoveryStale(discoveryStaleGrace) {
```
ovidiu-ssv-labs, line 580: **Finding 1 · [IMPORTANT but debatable] Arm the operator staleness check only after the first successful read — otherwise a cold start with unreachable bootnodes becomes a self-amplifying restart loop**

**Mechanism.** `NewTimedConn` seeds `lastReadUnixNano` at construction time (`network/discovery/timed_conn.go:34`), which happens inside `initDiscV5Listener` during `p2pNetwork.Setup()`. From that instant the 3-minute clock is running, and `StaleFor` cannot tell apart:

1. *the wedge* — the socket was being drained, then the read loop stopped; and
2. *never drained anything* — the socket is fine and the read loop is healthy, but no inbound UDP has arrived at all.

Both produce an identical `lastRead` that stops advancing. The boot node explicitly guards against this by gating on a populated routing table (`utils/boot_node/health.go:68`); the operator deliberately does not, so case (2) reads as a wedge.

**When case (2) actually happens.** An operator's socket stays fresh as long as its discv5 table has *any* peer, because those peers are revalidated every ~3s (geth `p2p/discover/common.go:72`, `table_reval.go:50-53`) and their pongs are reads. So the exposure is precisely: **an operator whose discv5 table is empty and whose FINDNODE/PING traffic gets no answer** — i.e. a node that has just (re)started while the bootnodes it is configured with are unreachable, or a node whose UDP is fully blocked by a firewall.

**Why it matters — the timing makes it a loop, not a blip.** Walking the actual bring-up path in `cli/operator/node.go`: `startNetwork` runs `p2pNetwork.Setup()` (line 668) and registers the p2p prober component (line 674), then `startHealthProber` is spawned (line 572) and probes immediately. So:

- `T+0` — socket bound, `lastRead` seeded, first probe passes.
- `T+3min` — `DiscoveryStale` flips true. This is a *level* signal, so it does not recover across `probeComponent`'s retries.
- `~T+5min` — the 6 attempts × 10s + 5 gaps × 10s burn down, `startHealthProber` returns an error, and the node exits non-zero for the orchestrator to restart (`cli/operator/prober.go:57-60`).
- restart → table empty again → still no inbound UDP → repeat.

The loop is self-amplifying: every restart throws away whatever table the node had, guaranteeing the next cycle also starts from empty. And it is correlated across the fleet — a bootnode outage (the very failure class #2979 is about) would now take down every operator that happens to restart during it, and keep them down, where previously they would have come up with degraded discovery and recovered on their own once a bootnode returned. A node with an established libp2p peer set that loses only UDP is also killed, dropping working TCP connections and its duty participation, even though the restart cannot fix an upstream network block.

**Note this does not weaken the detection you want.** #2980's wedge (blocking send on `Unhandled`) can only trigger *after* packets have been received and forwarded — the channel has to fill first. The same is true of the boot-node incident in #2979 ("the application *stopped* draining it"). So requiring at least one successful read before arming the check keeps every wedge in scope while removing the entire false-positive class.

**Suggested fix:** Track whether the socket has ever been read, and only treat staleness as a wedge once it has:

```go
// timed_conn.go
type TimedConn struct {
	discover.UDPConn

	lastReadUnixNano atomic.Int64
	reads            atomic.Uint64 // successful reads since start
	now              func() time.Time
}

func (c *TimedConn) ReadFromUDPAddrPort(b []byte) (int, netip.AddrPort, error) {
	n, addr, err := c.UDPConn.ReadFromUDPAddrPort(b)
	if err == nil {
		c.lastReadUnixNano.Store(c.nowFn().UnixNano())
		c.reads.Add(1)
	}
	return n, addr, err
}

// StaleFor reports a wedge: the socket was being drained and stopped. A socket
// that has never been read is not a wedge - it's a node that has never had
// inbound discv5 traffic (unreachable bootnodes, blocked UDP), which a restart
// cannot fix.
func (c *TimedConn) StaleFor(d time.Duration) bool {
	if c.reads.Load() == 0 {
		return false
	}
	return c.nowFn().Sub(c.LastRead()) > d
}
```

The boot node is unaffected (its empty-table rule already covers "never worked"). If you'd rather keep `StaleFor` purely time-based, expose `Reads()` and put the guard in `DiscV5Service.DiscoveryStale` instead. Either way, add a test case: seeded-but-never-read + clock advanced past grace → not stale.

network/discovery/dv5_service.go
```diff
@@ -433,7 +453,7 @@ func (dvs *DiscV5Service) initDiscV5Listener(discOpts *Options) (err error) {
 		return err
 	}
 
-	dv5PostForkListener, err := listenV5(udpConn, localNode, *dv5PostForkCfg)
+	dv5PostForkListener, err := listenV5(socketConn, localNode, *dv5PostForkCfg)
```
ovidiu-ssv-labs, line 486: **Finding 2 · [MINOR] The wiring that makes the whole feature work is untested, and it fails open**

Every test in this PR exercises `TimedConn`, `DiscoveryStale` and `check()` against hand-built structs. Nothing asserts the one fact the entire feature rests on: that the conn handed to `ListenV5` is the wrapped one.

If a future refactor of `initDiscV5Listener` passes `udpConn` instead of `socketConn` at `dv5_service.go:456` — easy to do, since both are in scope and the pre-fork call two lines down deliberately takes a *different* conn — then `dvs.socketConn` is still non-nil and still seeded, `StaleFor` still returns a plausible answer, and **every test in this PR still passes**. The detector just silently reports "healthy" forever. Same for the boot node at `utils/boot_node/node.go:179`. This is exactly the fail-open, silent-for-20-days failure mode #2979 is about, reintroduced one level up.

The package already has the machinery to pin this cheaply: `listenV5` is an injectable package-level var precisely so tests can wrap it, and `TestInitDiscV5Listener_CleansUpOnError` (`shared_conn_test.go:199`) already stubs it and counts calls.

**Suggested fix:** Add a wiring assertion using the existing stub pattern:

```go
// TestInitDiscV5Listener_WrapsPostForkConn pins the wiring the wedge detector
// depends on: the post-fork listener (the only one that drains the socket) must
// get the TimedConn, and the pre-fork one must get the SharedUDPConn.
func TestInitDiscV5Listener_WrapsPostForkConn(t *testing.T) {
	orig := listenV5
	t.Cleanup(func() { listenV5 = orig })

	var conns []discover.UDPConn
	listenV5 = func(conn discover.UDPConn, ln *enode.LocalNode, cfg discover.Config) (Listener, error) {
		conns = append(conns, conn)
		return orig(conn, ln, cfg)
	}

	dvs := testingDiscovery(t)
	t.Cleanup(func() { require.NoError(t, dvs.Close()) })

	require.Len(t, conns, 2)
	require.Same(t, dvs.socketConn, conns[0], "post-fork listener must read through TimedConn")
	require.IsType(t, &SharedUDPConn{}, conns[1])
}
```

And a one-line addition to the existing `TestNewDiscV5Service`: `assert.NotNil(t, dvs.socketConn)`. The boot node's `createListener` can be covered the same way by asserting the returned `*discovery.TimedConn` is non-nil.

ovidiu-ssv-labs (commented):
Sound, well-reasoned design: the socket-read timestamp is a genuinely cause-agnostic wedge signal, it is wired to the only listener that actually drains the socket, and greptile's 'quiet socket' P1 is properly fixed by the populated-table gate (verified against geth's 3s revalidation interval, not just the comment thread). The one thing worth changing before merge is the operator path, which cannot distinguish 'wedged' from 'never received any UDP' and will crash-loop a node that restarts while its bootnodes are unreachable; the rest are minor robustness/observability/test-wiring points.  [verdict: with_fixes]

network/discovery/timed_conn.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// LastRead reports when a packet was last read from the socket.
+func (c *TimedConn) LastRead() time.Time {
+	return time.Unix(0, c.lastReadUnixNano.Load())
+}
+
+// StaleFor reports whether the socket has gone unread for longer than d — the
+// signal that discovery has wedged.
+func (c *TimedConn) StaleFor(d time.Duration) bool {
```
ovidiu-ssv-labs, line 88: **Finding 3 · [MINOR] No metric or log for read staleness — the only evidence of a wedge is the process exiting**

The signal is consumed purely as a boolean at the moment of a restart decision. Nothing exports the underlying quantity, so:

- **You cannot alert before the kill.** A socket that has been unread for 2m50s is indistinguishable from one unread for 2s. There is no way to page on "discovery is degrading" — only to discover after the fact that a node restarted.
- **You cannot confirm the fix works.** After deploying this, there's no way to answer "how often does read-staleness approach the grace across the fleet?", which is exactly the data needed to tell whether 3 minutes is the right threshold or whether the false-positive class (see the other finding on the operator staleness check) is being hit in production.
- **Post-mortem evidence is thin.** On the operator, the wedge produces one error string on the way out; the pre-restart state (table size, last-read age, `SharedUDPConn.Dropped()`) is lost. Given #2979 went undiagnosed for 20 days, the diagnostic trail matters as much as the restart.

The package already has `network/discovery/observability.go` with registered instruments, and `SharedUDPConn` already exports `Dropped()`, so there's an established place for this.

**Suggested fix:** Export the age as an observable gauge in `network/discovery/observability.go` and record it wherever the health check reads it:

```go
discoveryReadStalenessGauge = observability.NewMetric(
	meter.Float64ObservableGauge(
		metricName("discovery.socket.read_staleness"),
		metric.WithUnit("s"),
		metric.WithDescription("seconds since the discv5 socket was last read"),
	),
)
```

Plus a `Warn` on the operator path when the check trips, carrying the table size alongside the staleness, so the exit reason and the surrounding state land in the same log line.

network/p2p/p2p.go
iurii-ssv: Done — `TimedConn` now tracks whether the socket has ever been read (a `read` flag set on every successful `ReadFromUDPAddrPort`), and `StaleFor` returns false until that first read.

A wedge can only arise after draining has started, so this keeps every real wedge in scope while removing the false-positive class: a node that has simply had no inbound UDP yet (unreachable bootnodes / blocked UDP), which a restart can't fix. Kept the guard inside `StaleFor` via a new `ReadStaleness()` returning `(age, ok)` so it stays the single source of truth — the boot node inherits it too, though its populated-table gate already covered the never-worked case.

Added the seeded-but-never-read regression test you suggested, and updated `TestDiscV5Service_DiscoveryStale`, which had encoded the old assumption.

network/discovery/dv5_service.go
iurii-ssv: Done — added `TestInitDiscV5Listener_WrapsPostForkConn`, using the existing injectable `listenV5` stub: it asserts the post-fork listener is handed the service's own `*TimedConn` (`socketConn`) and the pre-fork listener the `*SharedUDPConn`, so a refactor that passed the wrong conn now fails the suite instead of silently reporting healthy forever. Also added `assert.NotNil(t, dvs.socketConn)` to `TestNewDiscV5Service`.

Left the boot-node `createListener` path uncovered for now: it `Fatal`s and binds a real socket, so covering it cheaply would need a testability refactor (return errors + inject the listen fn, like `listenV5` here). Flagging as a follow-up rather than folding it into this PR.

network/discovery/timed_conn.go
iurii-ssv: Done, scoped to the operator. Added a synchronous `Int64Gauge` `ssv.p2p.discovery.socket.read_staleness` (seconds; 0 until the first read), sampled at the operator poll point in `DiscoveryStale`, plus a `Warn` in `p2pNetwork.Healthy()` when the check trips — so the wedge shows up in the logs and isn't only inferable from a restart.

Went with a synchronous gauge rather than an `ObservableGauge` to match the existing instruments in this package. No boot-node gauge: the boot-node binary wires no OTel exporter, so it would never be scraped — `/healthz` already surfaces that state.

utils/boot_node/health.go
```diff
@@ -0,0 +1,97 @@
...
+// such traffic, so staleness there is expected (cold start, or a quiet node with
+// no peers) and is judged only by the empty-table grace, which tolerates cold
+// start. lastNonEmpty is seeded at startup, so it catches both "never populated"
+// and "populated then emptied".
+//
+// One case slips past the fast path: a wedge present from boot. On restart discv5
+// loads seed nodes from the persistent enode DB straight into the table, so
+// AllNodes() can be >0 before the socket has been read once — and StaleFor stays
+// disarmed until that first read. Those seeds then fail revalidation and age out,
+// the table empties, and the empty-table grace trips instead. A runtime wedge
```
momosh-ssv: I fear the fallback described here doesn't hold under a dispatch-loop wedge. 
Table entries are only deleted in `tableRevalidation.handleResponse`, and revalidation itself goes through the wedged dispatch goroutine (`initCall` blocks on `t.callCh`, which that goroutine drains), so the seeds never age out, `AllNodes()` stays >0, `lastNonEmpty` keeps refreshing, and `/healthz` stays green indefinitely. 

The dispatch-wedge case is still caught (a prior read has armed `StaleFor`), but a stall where the socket never yields a read (kernel/conntrack-level) lands exactly in populated-table + disarmed-socket = healthy forever. Worth correcting the comment at minimum, or considering a self-probe: write a junk packet to the socket's own `LocalAddr()` and require `LastRead()` to advance, which covers it deterministically and independently of peers.

network/p2p/p2p.go
```diff
@@ -557,6 +562,16 @@ func (n *p2pNetwork) Healthy(ctx context.Context) error {
...
```
momosh-ssv, line 580: Could this lead to a restart loop on a healthy node that just has no inbound UDP? 
After a single successful read ever, 3 minutes of inbound silence fails `Healthy`: e.g. inbound UDP blocked upstream (firewall/NAT/conntrack change) while the node keeps operating fine over its established libp2p TCP peers; the restart drops every peer connection and can't fix the network. 

A total blackhole self-limits (the arm resets on restart), but a partial one the odd scan packet or NAT-refreshed reply arriving each boot re-arms every cycle and loops indefinitely. 
Maybe stamp writes in `TimedConn` too and only declare a wedge when reads AND writes are both stale: all v5 sends funnel through the same dispatch goroutine, so a real wedge freezes both, while an inbound blackhole leaves writes flowing (nursery bootnodes are re-pinged every refresh).

utils/boot_node/health.go
```diff
@@ -0,0 +1,97 @@
...
+// (socket drained, then stopped) is unaffected: a prior read has already armed
+// StaleFor.
+func (h *bootNodeHealth) check() error {
+	now := h.now()
+	if len(h.lister.AllNodes()) > 0 {
+		h.lastNonEmpty.Store(now.UnixNano())
```
momosh-ssv: Seems that the empty-table clock is driven by probe arrivals rather than table state: if probing pauses (kubelet restart, or the livenessProbe manifest lands in a later rollout than the image), the first resumed check compares now against a stale observation and can fail with no effective grace.

Restarting a healthy pod. Conversely, a table flapping in and out of empty faster than the probe period resets the clock every time, so the grace never accumulates. Might be worth sampling the table from a small owned ticker and having check() read only the sampled state.

network/discovery/dv5_service.go
```diff
@@ -159,6 +162,21 @@ func (dvs *DiscV5Service) close() error {
...
+// (or after a failed init), when there's nothing to judge.
+//
+// The operator health check polls this periodically, so it doubles as the
+// sampling point for the read-staleness gauge.
+func (dvs *DiscV5Service) DiscoveryStale(grace time.Duration) bool {
+	if dvs.socketConn == nil {
+		return false
+	}
+	age, ok := dvs.socketConn.ReadStaleness()
+	recordDiscoveryReadStaleness(dvs.ctx, int64(age.Seconds()))
```
momosh-ssv: Should we distinguish the never-read state in the gauge? 
Right now "read 200ms ago" and "never read since boot" both record 0 and never-read is precisely the state the health check deliberately ignores, so it's the one you'd want visible to a human for alerting.

 A sentinel (-1) or a companion read-ever metric would keep it distinguishable. Also, since this only records when Healthy() reaches the staleness check, the series goes quiet whenever isReady/discoveryFailed short-circuit earlier: i.e. exactly when discovery is broken.

network/discovery/timed_conn.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// TimedConn wraps a discv5 UDP socket and records when it was last read from.
+//
+// A discv5 listener drains its socket only through ReadFromUDPAddrPort, so a
+// wedged socket — bound but no longer read, leaving discovery silently dead —
+// shows up as a last-read timestamp that stops advancing. StaleFor turns that
+// into a liveness signal, used by both the operator and boot nodes.
+//
+// Only ReadFromUDPAddrPort is overridden; writes, Close and LocalAddr fall
```
momosh-ssv, line 26: Might be worth stating the scope here: only the post-fork listener reads through TimedConn, the pre-fork listener reads SharedUDPConn's buffer, so a pre-fork dispatch stall moves neither health signal. 

The known blocking cause is gone post-#2980, but as written the doc reads as if the wedge signal covers discovery generally.

utils/boot_node/health.go
```diff
@@ -0,0 +1,97 @@
...
+	if emptyFor := now.Sub(time.Unix(0, h.lastNonEmpty.Load())); emptyFor > h.emptyTableGrace {
+		return fmt.Errorf("discv5 routing table empty for >%s", h.emptyTableGrace)
+	}
+	return nil
+}
+
+func (h *bootNodeHealth) handler() http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		if err := h.check(); err != nil {
+			http.Error(w, err.Error(), http.StatusServiceUnavailable)
```
momosh-ssv: Do we need /healthz unauthenticated on the public ENR-advertised port?
 Every request takes the discv5 table mutex via AllNodes() and allocates the full node slice, there's no method filtering, and the error body discloses internal state. 
 
 Maybe memoize check() for ~1s, reject non-GET/HEAD with 405, and return a fixed body with the reason logged instead — or serve it on a pod-internal listener.

network/p2p/p2p.go
ovidiu-ssv-labs, line 580: **Finding 1 · [IMPORTANT] [still open] The operator's arming latch is permanent, so lost inbound UDP still produces a self-sustaining restart cycle**

**Status: independently verified as still open.** This confirms momosh-ssv's question on this line, with the mechanism traced end to end.

**Mechanism.** `read` is a one-way latch with no reset path:

```go
// network/discovery/timed_conn.go:50-57
func (c *TimedConn) ReadFromUDPAddrPort(b []byte) (n int, addr netip.AddrPort, err error) {
	n, addr, err = c.UDPConn.ReadFromUDPAddrPort(b)
	if err == nil {
		c.lastReadUnixNano.Store(c.nowFn().UnixNano())
		c.read.Store(true)   // set once, never cleared for the process lifetime
	}
	return n, addr, err
}
```

Once set, `ReadStaleness()` returns `ok=true` forever, so `Healthy()` at p2p.go:568 fails on any 3-minute inbound-UDP drought, regardless of cause. `startHealthProber` treats a failed round as terminal — it returns an error, the node exits non-zero, and the orchestrator restarts it.

**Why the fix for prior finding 1 does not cover this.** The never-read guard only exonerates a process that has never read a packet. A node running normally for hours has `read=true`. If inbound UDP then stops — upstream firewall/security-group change, NAT/conntrack rotation killing reply flows, an ISP-level UDP filter — the node keeps operating perfectly over its established libp2p TCP peer set, but exits within ~5 minutes.

**Why it self-sustains rather than self-limiting.** The stamp is taken at the socket read, before any decode: geth's `UDPv5.readLoop` reads every datagram on the port and only then attempts decode. So a port-scan probe, a stray discv4 packet, or a single NAT-refreshed reply is enough to flip `read` to true. Anything short of a perfect blackhole re-arms every boot, settling into a ~5-minute exit cycle.

**Why it matters.** Each cycle tears down every libp2p connection, forces full re-handshake/re-subscription, and risks missed attestation/sync-committee duties — direct penalty exposure on a validator node. The restart cannot fix an upstream UDP filter, so the loop is unbounded.

**On the suggested fix in the thread (stamp writes).** Does not discriminate — in the #2980 wedge, discv5 keeps sending (dispatch retries) even though reads are stale; in the blocked-inbound-UDP case, discv5 also keeps sending. Writes advancing with reads stale is the signature of both.

**What does discriminate:** not the socket — the impact. A genuine discv5 wedge progressively costs the node peers (churn with no replacement); an upstream UDP filter on a node with a warm, healthy TCP peer set does not.

**Suggested fix:** Require corroborating evidence that the node is actually impaired before restarting:

```go
if n.disc != nil && n.disc.DiscoveryStale(discoveryStaleGrace) {
	peers := len(n.host.Network().Peers())
	if peers < n.cfg.MinPeers {
		n.logger.Warn("discv5 socket wedged and peer set degraded", ...)
		return fmt.Errorf("discv5 socket not drained for >%s and only %d peers (discovery wedged)", discoveryStaleGrace, peers)
	}
	n.logger.Error("discv5 socket not drained, but peer set still healthy — not restarting", zap.Duration("grace", discoveryStaleGrace), zap.Int("peers", peers))
}
```

Alternative: raise discoveryStaleGrace substantially (discv5 revalidates at a 3s PingInterval, so even 30 minutes is a 600x margin) and cap self-inflicted restarts. Either way, add a test pinning that a permanent external UDP fault does not restart-loop.

utils/boot_node/health.go
```diff
@@ -0,0 +1,97 @@
...
+func (h *bootNodeHealth) check() error {
+	now := h.now()
+	if len(h.lister.AllNodes()) > 0 {
+		h.lastNonEmpty.Store(now.UnixNano())
+		if h.socket.StaleFor(h.readStaleGrace) {
+			return fmt.Errorf("discv5 socket not drained for >%s while routing table is populated", h.readStaleGrace)
+		}
+		return nil
+	}
+	if emptyFor := now.Sub(time.Unix(0, h.lastNonEmpty.Load())); emptyFor > h.emptyTableGrace {
```
ovidiu-ssv-labs: **Finding 2 · [IMPORTANT - debatable] Boot node: the empty-table branch ignores the socket signal, so a protocol-ID/config mismatch becomes a permanent CrashLoopBackOff**

**Mechanism.** `check()` splits on table population; the empty-table branch never consults `h.socket`:

```go
// utils/boot_node/health.go:74-87
if len(h.lister.AllNodes()) > 0 {
	h.lastNonEmpty.Store(now.UnixNano())
	if h.socket.StaleFor(h.readStaleGrace) { ... }   // socket consulted
	return nil
}
if emptyFor := now.Sub(...); emptyFor > h.emptyTableGrace {
	return fmt.Errorf("discv5 routing table empty for >%s", h.emptyTableGrace)  // socket ignored
}
```

So `AllNodes() == 0` for 10 minutes fails closed even when the socket is demonstrably being drained.

**Concrete break.** The boot node's table is populated exclusively by inbound traffic (no Bootnodes configured), and geth only adds an inbound peer after a completed handshake; a `V5ProtocolID` mismatch (wrong Network option, or a protocol-ID rollout landing on the boot node ahead of the fleet) makes every packet decode-fail before reaching the table. The socket is read normally (TimedConn is fresh), but AllNodes() stays 0 forever. /healthz then 503s every 10 minutes and the livenessProbe restarts the pod indefinitely — a restart cannot fix a config mismatch.

**Why this is worse than the pre-PR silent failure.** The boot node's only diagnostic surface is /p2p on the same HTTP server. Under CrashLoopBackOff that endpoint is unreachable for most of the cycle, so the operator is left with a flapping pod and no way to observe why the table is empty. A fresh deployment whose ENR hasn't been distributed yet hits the same flap, and CrashLoopBackOff backoff (up to 5 min) means it may be down at the moment the first cold-bootstrapping node arrives — the exact failure #2979 exists to prevent.

**Ruled out:** 'populated table full of dead peers during a network outage' does NOT false-positive — geth's tableRevalidation drains failed peers well inside the 3-minute read grace. Only the empty-table branch has this gap.

**Suggested fix:** Fold the socket signal into the empty-table branch — fail closed only when BOTH the table is empty AND the socket is undrained:

```go
if emptyFor := now.Sub(time.Unix(0, h.lastNonEmpty.Load())); emptyFor > h.emptyTableGrace {
	if age, everRead := h.socket.ReadStaleness(); everRead && age <= h.readStaleGrace {
		h.logger.Error("discv5 routing table empty while socket is being drained — check DiscoveryProtocolID / network config", zap.Duration("empty_for", emptyFor), zap.Duration("read_age", age))
		return nil
	}
	return fmt.Errorf("discv5 routing table empty for >%s and socket undrained", h.emptyTableGrace)
}
```

Needs socketDrainState widened to expose ReadStaleness() (time.Duration, bool) — *TimedConn already has it. Add two test cases: empty table + fresh socket past the empty grace (healthy, with error log), and empty table + never-read past the empty grace (503).

ovidiu-ssv-labs (commented):
🔁 Re-review vs c9d0cc1: 2 fixed, 1 partial, 0 still open, 3 new. All three prior findings were addressed — the never-read arming guard, the ListenV5 wiring test, and the staleness gauge + warn log are all in and the packages build and test green. Two things remain: the operator's arming latch is permanent, so momosh-ssv's restart-loop case (blocked inbound UDP after a prior read) is genuinely still open; and the boot node's empty-table branch ignores the socket signal, turning a discv5 protocol-ID/config mismatch into a permanent CrashLoopBackOff that also destroys the /p2p debug surface. The core design — a cause-agnostic socket-read timestamp wired to the only listener that actually drains the socket — is sound and well documented.  [verdict: with_fixes]

network/p2p/p2p.go
iurii-ssv: Done — adopted the impact gate. `Healthy()` now treats a stale socket as fatal only when the connected-peer count is below a floor (`discoveryStalePeerFloor = 10`); at or above it the node logs an error and stays up on its established connections, failing the probe later only if they erode. So a permanent external UDP fault on a working node no longer restart-loops — the loop needed the unconditional exit, and that path now requires corroborating impact.

Agreed on write-stamping: it doesn't discriminate — a stall confined to the read side keeps writes flowing and would slip past a reads-AND-writes-stale condition, and the blocked-inbound case keeps writes flowing too. Peer count keys on impact regardless of wedge shape.

There's no `cfg.MinPeers`, so the floor is a named constant with the rationale in its doc. Added the test you asked for — `discovery wedged but peer set healthy` → probe passes — plus a boundary case at floor−1 and a nil-host guard case.

iurii-ssv: Addressed via the peer-set gate (details in the sibling thread): staleness alone no longer restarts — only staleness plus a degraded peer set does — so the partial-blackhole loop on a node happily serving over TCP is gone.

Went with impact corroboration rather than write-stamping: all v5 sends funnel through the dispatch loop, so a genuine dispatch wedge freezes writes too (your point), but a stall confined to the read side would keep writes flowing and be missed by a reads-AND-writes condition — and an inbound blackhole also leaves writes flowing, so writes don't separate the two cases we need separated.

utils/boot_node/health.go
iurii-ssv: Partly agreed — the comment now states the residual precisely — but the named case (kernel/conntrack-level stall, socket never read) is caught: under it the dispatch loop is alive, so revalidation pings still time out via dispatch's own timers and failed peers are dropped (`livenessChecks /= 3` → delete at 0; DB seeds start at 0), the table empties within minutes, and the empty-table grace trips.

"Seeds never age out" requires dispatch itself to be wedged — but geth's read loop is one-packet-at-a-time flow-controlled (`readNextCh`), so any packet arriving before/as dispatch wedges arms the staleness check first. The genuinely uncaught combination is: dispatch wedged from boot **and** zero packets ever read **and** persistent-DB seeds. No known cause produces it since #2980, so it's documented as an accepted blind spot rather than closed with the self-probe — writing to `LocalAddr()` of a 0.0.0.0-bound socket is platform-fragile, and it adds a moving part on the hot discovery port for a no-known-cause case. Happy to revisit the self-probe as a follow-up if you feel strongly.

utils/boot_node/health.go
iurii-ssv: Done — table state is now sampled by an owned 1s ticker (`sample`/`start`), and `check()` reads only the sampled atomics. The empty-table clock advances with the table itself: a probe pause while the table was populated can no longer turn a stale observation into an instant past-grace verdict, and the 1s granularity removes the probe-instant aliasing on a flapping table.

Side benefit for your hardening thread: `/healthz` requests no longer touch the discv5 table mutex or allocate. Added `TestBootNodeHealth_ClockIsSampleDriven` and `TestBootNodeHealth_CheckReadsSamplesOnly`.

network/discovery/dv5_service.go
iurii-ssv: Done, both points. The gauge is now an `Int64ObservableGauge` registered against the socket at listener init and sampled at scrape time, so the series keeps reporting no matter what the health-check path does — including when `isReady`/`discoveryFailed` short-circuit before the staleness step (previously it went quiet exactly when discovery broke). Never-read reports `-1` instead of `0`, keeping the state the health check deliberately ignores visible for alerting (description updated). `DiscoveryStale` is a pure predicate again, and the callback registration is released on `Close`.

network/discovery/timed_conn.go
iurii-ssv: Done — the doc now scopes the signal: only the post-fork listener reads the socket through `TimedConn`; the pre-fork listener consumes decode-rejected packets relayed through `SharedUDPConn`'s buffer, so a pre-fork-only stall moves neither the timestamp nor the socket — and since the relay drops on a full buffer (#2980), it can't back up into the post-fork reader either.

utils/boot_node/health.go
iurii-ssv: Done, proportionately: non-GET/HEAD now gets 405 (on both `/healthz` and `/p2p`), the 503 body is a fixed `unhealthy` with the reason logged rather than disclosed, and the per-request cost is gone — `check()` reads sampled atomics only (see the sampler thread), so no memoization is needed.

Kept the port: the AWS LB check and the charts livenessProbe target it, and `/p2p` on the same mux — which dumps the entire table to anyone — is the materially larger pre-existing surface; worth its own follow-up if we want to lock that down.

utils/boot_node/health.go
iurii-ssv: Done — the empty-table branch now consults the socket: past the grace it fails closed only when the socket is undrained (never read, or read then gone stale); when the socket is actively drained it stays healthy and logs an error pointing at `DiscoveryProtocolID`/network config, since that state isn't restart-fixable and crash-looping would also take down the `/p2p` debug surface. `socketDrainState` widened to `ReadStaleness()`. Both suggested tests added — empty+drained past grace → 200 with the error log (asserted via a zap observer), empty+never-read past grace → 503 — plus empty+stale → 503.

One caveat to keep on record: a mismatch-broken boot node now stays green on `/healthz`, so anything routing on it would keep the node in rotation. If that matters we can split liveness from readiness (`/readyz`: empty table → out of rotation) as a follow-up; `/healthz` here carries k8s liveness semantics, which is what this PR's infra chain consumes.

network/p2p/p2p.go
momosh-ssv: Verified in 7639060c5: a stale socket fails the probe only below the 10 peer floor, otherwise it logs and stays up. Tests cover floor minus one, at floor, and nil host. Resolving.

utils/boot_node/health.go
momosh-ssv: Checked this against geth 1.16.4 and agree. Seeds load with zero liveness checks and the reval timeout fires from dispatch's own timer, so a read-side stall empties the table well inside the grace, and a dispatch wedge is armed by the first packet the read loop pulls before it blocks on packetInCh. The remaining combination needs zero packets ever, which #2979's empty-table datapoint also rules out for the incident shape. A loopback self-probe would be that same single arming packet, so it adds nothing. Comment reads right now, resolving.

utils/boot_node/health.go
momosh-ssv: Verified in 3644473fd: owned 1s sampler, check() reads the sampled atomics only, ClockIsSampleDriven and CheckReadsSamplesOnly pin it. Resolving.

network/discovery/dv5_service.go
momosh-ssv: Verified in 69d95f3a4: observable gauge registered at listener init, -1 before the first read, unregistered on close, DiscoveryStale is a pure predicate again. Resolving.

network/discovery/timed_conn.go
momosh-ssv: Verified in 69d95f3a4. Resolving.

utils/boot_node/health.go
momosh-ssv: Verified in 3644473fd: 405 on non GET/HEAD for both endpoints, fixed 503 body with the reason logged, no table access per request. Keeping the port makes sense while the LB check and the liveness probe target it. Resolving.

momosh-ssv (approved):

PR #2985, opened 2026-08-07
cli/operator: log the Boole fork schedule at boot

### LINKED ISSUE #2984: Log the SSV Boole fork schedule at boot (analogous to the Gloas 'fork scheduled' log)
## Context

External monitoring / test tooling needs to confirm, per node, which SSV epoch the **Boole** fork is scheduled at. The Gloas/ePBS fork already logs this at boot — `beacon/goclient/spec.go`: `gc.log.Info("Gloas (ePBS) fork scheduled", zap.Uint64("epoch", …))` — but there is **no equivalent for the SSV Boole fork**. The node knows the epoch at boot (`networkConfig.SSV.Forks.Boole`, used at runtime in `cli/operator/node.go`, `operator/duties/scheduler.go`, `operator/validator/controller.go`, …) but never logs the schedule.

This blocks a clean fork-awareness check in Aetheria's upcoming `(boole)` fork-transition suite (ssvlabs/aetheria#168): with no boot log, the suite has to *infer* the fork epoch from the first Boole topic subscription rather than read it directly from each node and cross-check agreement across the fleet (the way the gloas `(transition)` suite keys on `Gloas (ePBS) fork scheduled`).

## Ask

- Emit a single **INFO** line at startup, e.g. `Boole fork scheduled` with an `epoch` field (`SSVForks.Boole`), analogous to the Gloas log. Optionally include the current-vs-next domain types (`DomainType` / `NextDomainType`, `DomainTypeAtSlot`), which are the visible effect of the fork.
- A **DEBUG** twin for the disabled case (`Boole == math.MaxUint64`, i.e. unscheduled — the live-net default) would mirror the Gloas `fork not scheduled` pattern and flag a misconfiguration on a net where Boole *should* be scheduled.

## Why it helps

- Lets a harness/operator discover and cross-check the Boole fork epoch across the fleet (fork-awareness) instead of inferring it indirectly.
- Symmetric with the existing Gloas boot log, so tooling can treat both forks uniformly.
- Nice-to-have, not a blocker: Aetheria works by taking the epoch as configuration — but a boot log makes the fork-awareness assertion direct, cheap, and robust to config drift, and is generally useful for operators verifying a scheduled fork.

## Grounding

`networkconfig/ssv.go` (`SSVForks.Boole`), `networkconfig/network.go` (`BooleForkAtEpoch` / `DomainTypeAtSlot`), `beacon/goclient/spec.go` (the Gloas analog to mirror). Related: ssvlabs/aetheria#168, ssvlabs/ssv#2941 (Boole → stage convergence).

iurii-ssv:
Adds a single INFO line at operator boot — `Boole fork scheduled` with an `epoch` field — the SSV-fork analog of the beacon-side `Gloas (ePBS) fork scheduled` log. Operators and fleet tooling can then read each node's Boole epoch directly and cross-check agreement across the network, instead of inferring it from the first Boole topic subscription (which ssvlabs/aetheria#168's `(boole)` suite otherwise has to).

Networks with no Boole fork pin `Forks.Boole` to `math.MaxUint64` and log at DEBUG (`Boole fork not scheduled`) instead, flagging a misconfiguration on a net where Boole should be scheduled.

**Base:** `stage` — the Boole fork machinery isn't on `main` yet.

### Notes
- Epoch-only, mirroring the Gloas log: `fields.Epoch` emits the same `epoch` key, so tooling treats both forks uniformly. Domain types omitted — `setupSSVNetwork` already logs the full SSV config, so they'd be redundant here.
- Placed in `buildNode` (SSV-side), not `spec.go`: Boole is an SSV fork the beacon goclient doesn't know about. Uses the existing `BooleForkScheduled()` helper.
- No test — the Gloas analog has none, and exercising `buildNode` needs full CL/EL setup.

Closes #2984

### DIFF
--- a/cli/operator/node.go
+++ b/cli/operator/node.go
@@ -106,6 +106,16 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
 		Beacon: consensusClient.BeaconConfig(),
 	}
 
+	// Log the Boole fork schedule at boot — the SSV-fork analog of the beacon-side "Gloas (ePBS)
+	// fork scheduled" log — so operators and tooling can read each node's epoch and cross-check
+	// agreement across the fleet. Unscheduled networks pin Forks.Boole to math.MaxUint64, surfaced
+	// at DEBUG to flag a net where Boole should be scheduled but isn't.
+	if networkConfig.BooleForkScheduled() {
+		logger.Info("boole fork scheduled", fields.Epoch(networkConfig.SSV.Forks.Boole))
+	} else {
+		logger.Debug("boole fork not scheduled")
+	}
+
 	var executionAddrList []string
 	for _, addr := range strings.Split(cfg.ExecutionClient.Addr, ";") {
 		if addr = strings.TrimSpace(addr); addr != "" {


cli/operator/node.go
```diff
@@ -106,6 +106,16 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
 		Beacon: consensusClient.BeaconConfig(),
 	}
 
+	// Log the Boole fork schedule at boot — the SSV-fork analog of the beacon-side "Gloas (ePBS)
+	// fork scheduled" log — so operators and tooling can read each node's epoch and cross-check
+	// agreement across the fleet. Unscheduled networks pin Forks.Boole to math.MaxUint64, surfaced
+	// at DEBUG to flag a net where Boole should be scheduled but isn't.
+	if networkConfig.BooleForkScheduled() {
+		logger.Info("Boole fork scheduled", fields.Epoch(networkConfig.SSV.Forks.Boole))
```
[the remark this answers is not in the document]
iurii-ssv: Done — lowercased both to `boole fork scheduled` / `boole fork not scheduled`, matching `.cursorrules` and the rest of the `cli/operator` messages (`connecting CL(s)`, `setting ssv network`, …). The beacon-side `Gloas (ePBS) fork scheduled` log keeps its own casing in its own package, and the fork-awareness harness keys on the identical `epoch` field, so matching stays consistent either way.

cli/operator/node.go
```diff
@@ -106,6 +106,16 @@ func buildNode(ctx context.Context, cfg *config, logger *zap.Logger) (*node, err
...
 	}
 
+	// Log the Boole fork schedule at boot — the SSV-fork analog of the beacon-side "Gloas (ePBS)
+	// fork scheduled" log — so operators and tooling can read each node's epoch and cross-check
+	// agreement across the fleet. Unscheduled networks pin Forks.Boole to math.MaxUint64, surfaced
+	// at DEBUG to flag a net where Boole should be scheduled but isn't.
+	if networkConfig.BooleForkScheduled() {
+		logger.Info("boole fork scheduled", fields.Epoch(networkConfig.SSV.Forks.Boole))
+	} else {
+		logger.Debug("boole fork not scheduled")
```
ovidiu-ssv-labs, line 116: Maybe this will have to be raised to warn, otherwise after fork the node will have different behavior without knowing why (as long as log level stays info).

ovidiu-ssv-labs (approved):

cli/operator/node.go
iurii-ssv: Good point. I've kept it at DEBUG for now because a straight WARN would be noisy today: every real network pins `Boole: math.MaxUint64` (unscheduled by design — mainnet, holesky(-stage), hoodi(-stage)), so it'd fire on every node at every boot until Boole actually ships. And `BooleForkScheduled()` can't tell "misconfigured, should be scheduled" apart from "legitimately not scheduled yet" — both are just the sentinel.

Boot-time logging also can't really detect "after the fork the node behaves differently" — that's a runtime condition (network past the Boole epoch while this node's config still says unscheduled), so the loud signal belongs there rather than at boot. I'd keep this at DEBUG and add a runtime divergence WARN as a follow-up; the INFO `boole fork scheduled` line + fleet cross-checking the epoch is meant to be the boot-time detection path.

momosh-ssv (approved):

